PR #424, opened 2020-01-22
Add card_answered hook

glutanimate:
One use case would be in [Puppy Reinforcement](https://github.com/glutanimate/puppy-reinforcement/blob/master/src/puppy_reinforcement/views.py#L67), but this seems like a fairly popular point to latch on to in general: https://github.com/search?q=wrap%28sched.answercard&type=Code

Quick proof-of-concept (for demo purposes and debug console testing)

```python
from anki.hooks import card_answered

def onCardAnswered(card, ease):
    print(card, ease)

card_answered.append(onCardAnswered)
```

PR #430, opened 2020-01-24
Equip Reviewer._showAnswer with hooks covering common add-on usages

glutanimate:
Follow-up to #424

Demo code:

```python
from aqt import gui_hooks

def onHook(reviewer, card, ease):
   print(reviewer, card, ease)

def onFilter(ease_tuple, reviewer, card):
   print(ease_tuple, reviewer, card)
   return ease_tuple

gui_hooks.reviewer_did_answer_card.append(onHook)
gui_hooks.reviewer_will_answer_card.append(onFilter)
```

### DIFF
--- a/qt/aqt/gui_hooks.py
+++ b/qt/aqt/gui_hooks.py
@@ -650,6 +650,30 @@ class _ReviewDidUndoHook:
 review_did_undo = _ReviewDidUndoHook()
 
 
+class _ReviewerDidAnswerCardHook:
+    _hooks: List[Callable[["aqt.reviewer.Reviewer", Card, int], None]] = []
+
+    def append(self, cb: Callable[["aqt.reviewer.Reviewer", Card, int], None]) -> None:
+        """(reviewer: aqt.reviewer.Reviewer, card: Card, ease: int)"""
+        self._hooks.append(cb)
+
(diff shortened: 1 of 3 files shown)


glutanimate:
(Sorry, I need to adjust my `black` config.)

Continuing from the previous PR:

> Perhaps category 3 is better served catching the hotkeys instead of hooking into this function?

That would definitely be a viable approach as well. My only concern is that it would make it impossible to combine these "cat3" add-ons with something like [Customize Keyboard Shortcuts](https://ankiweb.net/shared/info/24411424).

BTW, I'm sure this feature request has come up a lot over the years, and for obvious reasons it's not an easy request to follow up on, but with development moving more rapidly now:

Do you think it would be feasible to implement some kind of native keyboard shortcut registry with add-on support and key reassignment capabilities? Finding an unassigned shortcut to use that doesn't conflict with Anki, the system, or other add-ons has only become more difficult over the years, so this would be a great boon.

I'm not sure how PyCharm handles this, but I think VS Code does a great job at handling keyboard assignments across its own definitions and those contributed by extensions for instance.

dae:
Looks good, thanks!

A shortcut manager is an option for the future, but not at the top of the priority list at the moment I'm afraid - I think it would take quite a bit of time to get right, and the multiple ways shortcuts are handled, and multiple screens that have different shortcuts, makes things complicated :-(

PR #431, opened 2020-01-27
Fix hooks modifying list inside for loop

lovac42:
If func throws an exception, or calls remHook, the list is modified inside the for loop causing the next task in hookFuncs to be skipped.
```
def onProfileLoaded():
    try:
        test_stuff()
    except:
        remHook("profileLoaded", onProfileLoaded)
        # ^Modifies hookFuncs inside the for loop

addHook('profileLoaded', onProfileLoaded)
```
https://github.com/krassowski/Anki-Night-Mode/issues/103

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -24,6 +24,7 @@ Erez Volk <erez.volk@gmail.com>
 zjosua <zjosua@hotmail.com>
 Arthur Milchior <arthur@milchior.fr>
 Yngve Hoiseth <yngve@hoiseth.net>
+lovac42 <https://github.com/lovac42>
 
 ********************
 
--- a/pylib/anki/hooks.py
+++ b/pylib/anki/hooks.py
@@ -446,7 +446,7 @@ def runHook(hook: str, *args) -> None:
     "Run all functions on hook."
     hookFuncs = _hooks.get(hook, None)
     if hookFuncs:
-        for func in hookFuncs:
+        for func in hookFuncs.copy():
             try:
                 func(*args)
             except:
@@ -457,7 +457,7 @@ def runHook(hook: str, *args) -> None:
 def runFilter(hook: str, arg: Any, *args) -> Any:
     hookFuncs = _hooks.get(hook, None)
     if hookFuncs:
-        for func in hookFuncs:
+        for func in hookFuncs.copy():
             try:
                 arg = func(arg, *args)
             except:


dae:
When an exception is throw, the loop is not iterated further, so I believe it's safe to mutate the list in that case. And I suspect there are not many add-ons that try to remove themselves from the hook that is currently running, so perhaps this would be better fixed in Night Mode instead of making every hook call require a copy?

Re the author check, when you make a merge in GitHub it'll use your hidden email - you can add that to the authors list as well by editing the contrib file on GitHub, and making some trivial change.

lovac42:
I don't see much of a performance lost, though I am not certain how hooks are used on mobile. Besides the Anki hooks, there are custom hooks setup by addons calling each other. And I wonder how many authors would be baffled by this should the situation arise.

The link was just an example, this is less about night mode, but more about future proofing the system, preventing issues before they occur, like interger overflow ;-). But I can understand your reasoning here, and I guess it doesn't matter as much as we switch over to the new style.

dae:
Let's take a wait and see approach - if you notice it's a recurring issue please let me know. Most of the current hooks are not performance sensitive at all, but hooks may be added to hotter code paths in the future, so it'd be nice if the copy could be avoided.

glutanimate:
@dae: This just bit me with an add-on in the same exact scenario: Attempting to hook into `profile_did_open` as a single-shot hook. I think this is a fairly common usage pattern for this hook, and at the moment there are quite a few add-ons out there who are probably misusing the hook completely (ignoring that things will break on multi-profile set-ups).

I'm note sure what the best solution here is – I'm not sure about the performance impact of copying the hook list for each hook (though it doesn't feel like it would make a huge difference). However, I do think that we need a solution for this, if only because I don't think the fact that Anki loops over hook subscribers is entirely obvious. Because of that, I think it's easy for add-on authors to be caught unaware of the implications of mutating the same hook while a hook subscriber is running.

dae:
How about adding another hook that fires at the end of main's __init__ (wrapping the fn passed to the timer), and updating the docs to warn users not to mutate the hook that is currently executing?

glutanimate:
I think that would be a good solution, yeah. Would you like me to submit a PR for the hook and if so would this be along the lines of what you had in mind?:


```python
...
if isWin:
    fn = self.setupProfileAfterWebviewsLoaded
else:
    fn = self.setupProfile

def on_window_init():
    fn()
    gui_hooks.main_window_did_init()

self.progress.timer(10, on_window_init, False, requiresCollection=False)
```

dae:
Yep, that looks good, thanks. Would you mind sending through a PR? I'll take care of updating the manual when I merge it, unless you feel like sending through a PR for that too.

glutanimate:
Just submitted a PR here for the hook and a bit of inline docs: https://github.com/ankitects/anki/pull/641

Re: the manual – yeah that would be great, thank you!

PR #433, opened 2020-01-31
Ints

Arthur-Milchior:
Each commit deals with one number, transforming it into a constant. So it's easy to check.
I also checked that I did add "f" before each string which became a f-string.


Arthur-Milchior:
Last commit simply reformat using black. I decided not to clean format at each step so that proof-reading the commit stay easy. Let me know if you prefer the formatting to be clean in each commit

dae:
Thank you for your hard work Arthur, though you've done a bit more than I was expecting! I will have a look through the changes and make some comments as necessary.

pylib/anki/importing/noteimp.py
```diff
@@ -55,6 +55,9 @@ def __init__(self) -> None:
 # 0: update if first field matches existing note
 # 1: ignore if first field matches existing note
 # 2: import even if first field matches existing note
+updateMode = 0
+ignoreMode = 1
+addMode = 2
```
dae: Constants should be SCREAMING_CASE

pylib/anki/consts.py
```diff
@@ -16,6 +16,7 @@
 
 # Card and queue types
 QUEUE_TYPE_NEW = 0
+QUEUE_TYPE_LRN = 1
```
dae: type and queue are related, but not interchangeable. For example, type=3 means a relearning card, but queue=3 means a learning or relearning card with a 24+ hour interval. For this reason, there should be separate constants for TYPE and QUEUE, as can be seen on the work the Agenty did already:  

CARD_TYPE_RELEARNING = 3
QUEUE_TYPE_PREVIEW = 4

dae:
The scheduling constants are probably also better placed in the sched*.py files rather than consts.py, as there are some that only apply to V2 for example. I've avoided changing the existing ones in consts.py to avoid breaking add-ons, but when introducing new constants, backwards compatibility is not a concern.

Arthur-Milchior:
Can you please let me know what you call a scheduling constant ? Only type and queue ?

pylib/anki/importing/noteimp.py
Arthur-Milchior: done

Arthur-Milchior:
One thing is currently not clear to me. I can copy paste the constants in sched.py and schedv2.py and give them different names when they have different meaning. But then, what should I do for the use of the constants in other files.
Since 0 always mean new, it's not really a problem if I use a constant imported from sched.py while sched2 is used.
However, I don't know by what I should replace the constant 3, in finder, in stats, where the meaning depends on the scheduler choosen

dae:
Yeah, I was referred to the QUEUE/TYPE constants.

In the cases where they don't differ, perhaps we could define them in sched.py and then import them for use into schedv2.py as well. That would make them available in both files, without having to define them twice.

IIRC, a queue of 3 means the same in both schedulers, and a type of 3 means the same in both schedulers, but a queue of 3 does not mean the same thing as a type of 3. So the other files should be ok? If you spot a problem while making the change, please let me know.

pylib/anki/consts.py
Arthur-Milchior: QUEUE and CARD are separated.
I should emphasize that this is really not a clear name. I used this constant name to follow constants already introduced in the code. However, it's strange that both CARD_TYPE_RELEARNING and QUEUE_TYPE_PREVIEW contain the word type, while only the former is related to the attribute "type" in the database/card object. QUEUE_PREVIEW would makes more sens I believe, in order to avoid confusion

Arthur-Milchior:
I find a problem which I should have expected. anki/sched.py imports anki.cards.Card
Which means that I can't tell anki.cards to import constants from anki.sched, it creates a loop of imports
One (ugly) solution would be to state that the constant is a value in the class Scheduler.
So I access it with self.col.sched.QUEUE_TYPE_REV
Otherwise, having constants in consts seems to make more sens here

pylib/anki/stats.py
```diff
@@ -103,13 +103,13 @@ class CollectionStats:
...
         self._stats = None
-        self.type = 0
+        self.type = CARD_TYPE_NEW
         self.width = 600
         self.height = 200
         self.wholeCollection = False
 
     # assumes jquery & plot are available in document
-    def report(self, type=0) -> str:
+    def report(self, type=CARD_TYPE_NEW) -> str:
```
dae: type is not a card type here

pylib/anki/importing/anki2.py
```diff
@@ -344,7 +344,7 @@ def _importCards(self) -> None:
             card[4] = intTime()
             card[5] = usn
             # review cards have a due date relative to collection
-            if card[7] in (2, 3) or card[6] == 2:
+            if card[7] in (2, 3) or card[6] == CARD_TYPE_LRN:
```
dae: This is wrong

pylib/anki/stats.py
```diff
@@ -605,7 +605,7 @@ def ivlGraph(self) -> str:
             totd.append((grp, tot / float(all) * 100))
         if self.type == CARD_TYPE_NEW:
             ivlmax = 31
-        elif self.type == 1:
+        elif self.type == CARD_TYPE_LRN:
```
dae: this is not a card type

pylib/anki/schedv2.py
```diff
@@ -1865,10 +1865,10 @@ def _emptyAllFiltered(self) -> None:
             f"""
 update cards set did = odid, queue = (case
 when type = {CARD_TYPE_LRN} then {QUEUE_TYPE_NEW}
-when type = {CARD_TYPE_RELEARNING} then 2
+when type = {CARD_TYPE_RELEARNING} then {CARD_TYPE_REV}
```
dae: shoud be queue, not type

pylib/anki/schedv2.py
```diff
@@ -1926,7 +1926,7 @@ def _moveManuallyBuried(self) -> None:
     # adding 'hard' in v2 scheduler means old ease entries need shifting
     # up or down
     def _remapLearningAnswers(self, sql: str) -> None:
-        self.col.db.execute(f"update revlog set %s and type in ({CARD_TYPE_NEW},2)" % sql)
+        self.col.db.execute(f"update revlog set %s and type in ({CARD_TYPE_NEW},{CARD_TYPE_REV})" % sql)
```
dae: these are revlog types, not card types

pylib/anki/stats.py
```diff
@@ -155,7 +155,7 @@ def todayStats(self) -> str:
 sum(case when ease = 1 then 1 else 0 end), /* failed */
 sum(case when type = {CARD_TYPE_NEW} then 1 else 0 end), /* learning */
 sum(case when type = {CARD_TYPE_LRN} then 1 else 0 end), /* review */
-sum(case when type = 2 then 1 else 0 end), /* relearn */
+sum(case when type = {CARD_TYPE_REV} then 1 else 0 end), /* relearn */
```
dae: revlog types not card types

pylib/anki/stats.py
```diff
@@ -551,13 +551,13 @@ def _done(self, num=7, chunk=1) -> Any:
 sum(case when type = {CARD_TYPE_NEW} then 1 else 0 end), -- lrn count
 sum(case when type = {CARD_TYPE_LRN} and lastIvl < 21 then 1 else 0 end), -- yng count
 sum(case when type = {CARD_TYPE_LRN} and lastIvl >= 21 then 1 else 0 end), -- mtr count
-sum(case when type = 2 then 1 else 0 end), -- lapse count
+sum(case when type = {CARD_TYPE_REV} then 1 else 0 end), -- lapse count
```
dae: revlog types

pylib/anki/stats.py
```diff
@@ -758,10 +758,10 @@ def _eases(self) -> Any:
         return self.col.db.all(
             f"""
 select (case
-when type in ({CARD_TYPE_NEW},2) then 0
+when type in ({CARD_TYPE_NEW},{CARD_TYPE_REV}) then 0
 when lastIvl < 21 then 1
-else 2 end) as thetype,
-(case when type in ({CARD_TYPE_NEW},2) and ease = 4 then %s else ease end), count() from revlog %s
+else {CARD_TYPE_REV} end) as thetype,
+(case when type in ({CARD_TYPE_NEW},{CARD_TYPE_REV}) and ease = 4 then %s else ease end), count() from revlog %s
```
dae: revlog types

pylib/anki/stats.py
```diff
@@ -926,21 +926,21 @@ def _lineTbl(self, i) -> str:
...
             % self._limit()
         )
 
     def _cards(self) -> Any:
         return self.col.db.first(
             f"""
 select
-sum(case when queue=2 and ivl >= 21 then 1 else 0 end), -- mtr
-sum(case when queue in ({QUEUE_TYPE_LRN},3) or (queue=2 and ivl < 21) then 1 else 0 end), -- yng/lrn
+sum(case when queue={QUEUE_TYPE_REV} and ivl >= 21 then 1 else 0 end), -- mtr
```
dae: revlog types

pylib/anki/consts.py
```diff
@@ -21,6 +21,9 @@
 QUEUE_TYPE_NEW = 0
 QUEUE_TYPE_LRN = 1
 QUEUE_TYPE_REV = 2
+QUEUE_TYPE_DAY_LRN = 3
+CARD_TYPE_RELEARNING = 3
+QUEUE_TYPE_DAY_LEARN_RELEARN = 3
```
dae: Could you merge the two duplicate QUEUE_TYPEs?

pylib/anki/stats.py
Arthur-Milchior: Corrected.
I also corrected the comment of method CollectionStats.report since it seems that the type indicated in comment are inconsistent with the 3 buttons in the footer

pylib/anki/schedv2.py
Arthur-Milchior: Corrected

pylib/anki/stats.py
Arthur-Milchior: Corrected in whole of stats

pylib/anki/consts.py
Arthur-Milchior: Done. Even if I must state I'm surprised. I thought that the queue 3 means two distinct things in sched v1 and v2; so two names would have make sens

pylib/anki/importing/anki2.py
Arthur-Milchior: I corrected the entire line. I don't know how I got this one so wrong

pylib/anki/stats.py
Arthur-Milchior: Corrected all of stats to a new constant

Arthur-Milchior:
Followed all of your correction. Thanks

pylib/anki/importing/anki2.py
```diff
@@ -356,13 +360,13 @@ def _importCards(self) -> None:
                 card[8] = card[14]
                 card[14] = 0
                 # queue
-                if card[6] == 1:  # type
-                    card[7] = 0
+                if card[6] == QUEUE_TYPE_LRN:  # type
```
dae: perhaps that one should be CARD_TYPE_LRN?

dae:
Thanks Arthur, apart from the comment above, I didn't notice any other issues, and I'll merge this before the first beta of 2.1.21

pylib/anki/importing/anki2.py
Arthur-Milchior: Corrected

Arthur-Milchior:
Is there anything pending ? I think I did all corrections. Please tell me if there is something I missed

dae:
It's all good to go, I'm just waiting on the 2.1.20 release before merging it, as it's a big change and makes sense to introduce at the start of a new beta testing period.

Arthur-Milchior:
Got it. Thanks for the explanation

PR #434, opened 2020-02-02
anki.exporting.AnkiExporter.deckIds

Arthur-Milchior:
This change would help me simplify the code of
https://ankiweb.net/shared/info/1983204951

It would allow my exporter to use the code of AnkiExporter while
selecting which decks I export, instead of exporting all decks as I
was doing until today. (I never realized I was exporting all deck
actually, until someone sent me a bug report)

### DIFF
--- a/pylib/anki/exporting.py
+++ b/pylib/anki/exporting.py
@@ -160,6 +160,12 @@ class AnkiExporter(Exporter):
     def __init__(self, col: _Collection) -> None:
         Exporter.__init__(self, col)
 
+    def deckIds(self) -> List[int]:
+        if not self.did:
+            return []
+        else:
+            return [self.did] + [x[1] for x in self.src.decks.children(self.did)]
+
     def exportInto(self, path: str) -> None:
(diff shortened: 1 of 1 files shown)


pylib/anki/exporting.py
```diff
@@ -160,6 +160,12 @@ class AnkiExporter(Exporter):
     def __init__(self, col: _Collection) -> None:
         Exporter.__init__(self, col)
 
+    def deckIds(self) -> typing.List[int]:
```
dae: The typing. prefix is unnecessary, but otherwise looks good.

Arthur-Milchior: Removed.
Some Typing remains in the code; such as typing.Union. A reason to keep them at some places and not other ?

dae: No reason to keep them - they were just added by a tool automatically, and nobody has gone through and manually removed the prefix and added the required import statement to the top.

dae:
Thanks!

PR #435, opened 2020-02-03
Extend Copy Debug Info and add human_version to manifest.json 

ijgnd:
relevant thread: https://anki.tenderapp.com/discussions/ankidesktop/38363-extend-copy-debug-info

### DIFF
--- a/qt/aqt/about.py
+++ b/qt/aqt/about.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 # License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 import platform
+import time
 
 import aqt.forms
 from anki.lang import _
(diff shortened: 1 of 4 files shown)


qt/aqt/about.py
```diff
@@ -32,10 +33,51 @@ def show(mw):
 
     # Copy debug info
     ######################################################################
+    def addon_fmt(addmgr, a):
```
dae: type hints would be better:

addmgr: AddonManager, a: AddonMeta

qt/aqt/addons.py
```diff
@@ -132,6 +133,7 @@ def from_json_meta(dir_name: str, json_meta: Dict[str, Any]) -> AddonMeta:
             min_point_version=json_meta.get("min_point_version", 0) or 0,
             max_point_version=json_meta.get("max_point_version", 0) or 0,
             branch_index=json_meta.get("branch_index", 0) or 0,
+            human_version=json_meta.get("human_version", "") or "",
```
dae: the 'or ""' should be removed, as it will cause the key to be saved if it's empty. The entries above only use or 0 because a previous beta was writing them as None.

dae:
Re the contributors failure, if you edit that file and make a trivial change using GitHub's editor, it should fix the problem.

ijgnd:
Thanks for the feedback.

qt/aqt/addons.py
```diff
@@ -132,6 +133,7 @@ def from_json_meta(dir_name: str, json_meta: Dict[str, Any]) -> AddonMeta:
             min_point_version=json_meta.get("min_point_version", 0) or 0,
             max_point_version=json_meta.get("max_point_version", 0) or 0,
             branch_index=json_meta.get("branch_index", 0) or 0,
+            human_version=json_meta.get("human_version", ""),
```
dae: Because write_addon_meta() checks for None, the second argument should be omitted here. I'll fix it after merging.

dae:
Thanks!

PR #437, opened 2020-02-05
correct missing french translation

Arthur-Milchior:
I took a loot at missing French translations. Translated what I thought was really missing, corrected the remaining in the code

### DIFF
--- a/pylib/anki/stats.py
+++ b/pylib/anki/stats.py
@@ -457,6 +457,10 @@ group by day order by day"""
             self._line(
                 i,
                 _("Average answer time"),
+                # T: For example, in the statistics line: " Average
+                # answer time: 16.8s (3.6 cards/minute)", then
+                # "%(a)0.1fs" represents "16.8s" and "%(b)s" represents
+                # "3.6 cards/minutes")
                 _("%(a)0.1fs (%(b)s)") % dict(a=(tot * 60) / total, b=text),
             )
         return self._lineTbl(i), int(tot)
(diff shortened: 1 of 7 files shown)


dae:
Could you check the formatting? I recommend you add a pre-push hook in git to automatically catch such problems before pushing, as described in README.contributing. It's helped me multiple times in the past :-)

dae:
(if the long comments need to be wrapped, I think you can use multiple #T: lines)

Arthur-Milchior:
Oh. I didn't know that multiple lines were acceptable. That's actually why I didn't do anything about this particular case.

dae:
You can check it works correctly by running qt/i18n/update-pot - it should add the split lines to anki.pot

Arthur-Milchior:
It indeed works.
Related question: is there a way to let random user have easy read-only access to crowdin's translation ?
For the sake fo consistency, ankidroid asks its translator to check anki's translation and use the same when possible. That was easy with previous translation method. But now, it means our translator need either to ask your permission to have access to translation, how to figure out how to read translation from anki source file. I fear that it's long and complex for translators. If we can just have a link to give them, that would be so much easier

dae:
I gave it a try, but it doesn't look like it's possible to provide read only access - the site asks users to sign up when they try to view a language. The translation files are on GitHub however, so you can  point users to https://github.com/ankitects/anki/tree/master/qt/i18n/translations/anki.pot and they can click on the language of interest, then use ctrl+f/cmd+f to search in the page for the string they're looking for.

dae:
(did you run black? check is still failing)

dae:
Thanks!

dae:
The new translation site allows people to view the translations without signing up: https://anki.tenderapp.com/discussions/announcements/138-more-updates-to-ankis-translations

Arthur-Milchior:
Thanks; I ported the information to ankidroid's wiki

PR #258, opened 2018-10-09
Allow add-on authors to more easily extend Anki's web views

glutanimate:
Extends all pertinent web views with runFilter hooks that enable add-on authors to easily prepend and append HTML to the body of the view, or add their own CSS or JS to the head element.

Also implements two runFilter hooks specific to the stats areas of the Overview and DeckBrowser, granting more control over these commonly modified parts of Anki's UI.

These changes should help alleviate some of the conflicts between add-ons that target the same HTML views, and also do away with some of the hacks that are currently in use to achieve similar
goals.

----

I've been meaning to submit something like this for a while now. What prompted me to finally work on these changes is an old issue that recently resurfaced in my efforts to refactor and port Review Heatmap to 2.1: Review Heatmap extends Anki's Overview and DeckBrowser screens with a custom element showcasing user activity over time. For the DeckBrowser screen this element is positioned right below the stats area. For the Overview screen it's right below the table element.

The main challenge I initially faced when modifying these parts of Anki's user interface was that they were also commonly targeted by other add-ons. So users installing an add-on like More Overview Stats would suddenly see their heatmap disappear, as the _table() function I was wrapping to add my heatmap would be overwritten by the other add-on. This forced me to move to a more hacky and unstable implementation in the form of completely [overwriting Overview._renderPage()](https://github.com/glutanimate/review-heatmap/blob/master/src/review_heatmap/main.py#L272-L316) and [Overview._body](https://github.com/glutanimate/review-heatmap/blob/master/src/review_heatmap/web.py#L319-L328). To this day this is how Review Heatmap works on the Deck screen. Needless to say, this in turn has been the cause of many incompatibilities with other add-ons, e.g. Night Mode.

My first hope with this PR is that the changes will allow add-on authors to modify Anki's web elements less intrusively. My second hope is that it will make the web part of add-on development more accessible in general.

To test these changes in practice I've also compiled a quick proof-of-concept add-on that is available [here](https://gist.github.com/glutanimate/b5c16c7e5bc3b96e7388af767bfd6080).

PR #438, opened 2020-02-06
Allow to add css/js to the editor easily

Arthur-Milchior:
At least four add-ons need to add css or js to the editor.
* Multi column note by HSSM
* Frozen field by Glutanimate
* Tex in editor by me
* Resize image in editor (by me, not yet published)
* Add-on 2064123047 merging the three first add-ons

One related problem each add-on have is that they need to add some
js/css to the editor. Which can be done mostly by changing the whole
method setupWeb. This is actually overkill, and allowing an extra
parameter to this method would solve this problem and allow for a more
beautiful add-on code.

This is why I suggest this extra parameter.

I want to emphasize that the default value here is the same as the one
in webview.py's stdHtml function.

I should not that an even more elegant solution would be to allow more
css/js file to be incorporated into the window. But that would lead to
more complex anki code and not much improvement in add-ons, so I don't
suggest spending time on it

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -93,7 +93,14 @@ class Editor:
         self.widget.setLayout(l)
         self.outerLayout = l
 
-    def setupWeb(self) -> None:
+    def setupWeb(self, head: str = "") -> None:
+        """Head -- some string to put in the html's head.
+
+
+        Allows add-ons to add css/js by redefining setupWeb and
+        appending it to the input head, then calling this method.
+
+        """
         self.web = EditorWebView(self.widget, self)
         self.web.title = "editor"
         self.web.allowDrops = True
@@ -167,6 +174,7 @@ class Editor:
             _html % (bgcol, bgcol, topbuts, _("Show Duplicates")),
             css=["editor.css"],
             js=["jquery.js", "editor.js"],
+            head=head,
         )
 
     # Top buttons


glutanimate:
Thanks for this thoughtful PR, Arthur! Seeing this reminded me that I still hadn't updated #258, and with Anki now sporting its new hook system, I thought this might be the perfect time to finally get this out of the way. If you have the time, I'd love to hear your thoughts in #439.

Arthur-Milchior:
You're welcome.
I understood that PR should be as small as possible. I would not really have considered doing such a big change as yours. I certainly agree that being able to give js/css files' directly into a filter seems really nice. I don't know whether it solves a problem I currently have and solve in a dirty way.
I often want to redefine a function from editor.js to edit it. For example saveField and setFormat. And what I do is either
```javascript
setTimeout(function(){saveField  = function(type){my new function}}},1000);
```
or define setFormat_ and monkey patch editor.py so that it calls setFormat_ instead of setFormat

My actual problem is that the content of "head" appears before the javascript files. So when it is executed, $ is not yet loaded.

And I guess there are still parts of Anki I don't really understand, since it's not immediately clear what your commit does. But I guess if it gets accepted I'll take a look at it; the concept seems cool

dae:
Extra hooks are welcome, as I'd like people to move away from monkey patching where possible.

Let's wait for Aristotelis' proposed changes to settle first, but changing webview's code to add the head string after the css/js files is an option in the future if it doesn't break anything.

Arthur-Milchior:
I can try to take a look at all my add ons and remove monkey patching, creating hooks, if that helps.
I have no way to know whether having head after js/css will change anything. Not without checking whether any add on rely on the fact that js/css is loaded after their head

dae:
Sure, would you like to start by sending me some patches for 3-5 hooks, and we can go from there? Please try and follow the naming format: https://github.com/ankitects/anki/blob/master/README.contributing#L59

Switching the order will probably be ok - please test it after 2.1.20 is released, and if you don't find any problems in your own testing we can try it in a beta.

dae:
Order switched

Arthur-Milchior:
Thank you. I guess I close this pull request now. No point in keeping it

PR #228, opened 2018-03-06
Extend Bridge class with new method that passes return value to JS

glutanimate:
While putting the final touches on [Mouseover Dictionary](https://github.com/glutanimate/mouseover-dictionary) and porting the add-on to Anki 2.1, I ran into a roadblock with setting up a py<->js interface. On Anki 2.0 [I had been using `addToJavaScriptWindowObject`](https://github.com/glutanimate/mouseover-dictionary/blob/master/src/mouseover_dictionary/main.py#L73-L99) to expose a custom Python object in JS, but that method was longer available with QtWebEngine.

I tried the usual by monkey-patching AnkiWebPage, or trying to register my Python object with the QWebChannel at runtime, but none of these attempts were fruitful. So eventually I decided to work on a PR to address this.

At its core, this commit simply reintroduces the `py.run` command that was available in Anki 2.0. It adds a new `run` method to the Bridge class – exposed as `pyrun` in JavaScript – for which the return value is passed back to the JavaScript caller. In cases where simply calling a Python handler through `pycmd` is not enough this will allow add-on authors to set up an easy-to-use JS-Py interface.

 Here's how I made use of that in my add-on, for instance: [Python](https://github.com/glutanimate/mouseover-dictionary/blob/master/src/mouseover_dictionary/main.py#L210-L214) and [JavaScript](https://github.com/glutanimate/mouseover-dictionary/blob/master/src/mouseover_dictionary/web.py#L169-L175).

PR #439, opened 2020-02-06
Allow add-on authors to easily inject their own content into Anki's web views

glutanimate:
**Overview**

This is a reworked version of #258 prompted by Arthur's thoughtful changes in #438.

Just like #258, this PR seeks to simplify the process of writing add-ons that modify Anki's web views. To do so it adds a series of easy-to-use filters that allow injecting custom HTML, CSS, and JS with as little inter-addon friction as possible.

**Details**

Web content filters are passed two arguments, a data class instance storing a number of modifiable fields (`ModifiableWebContent` type), and the original caller that initiated the web view update (i.e. an Anki object like a Reviewer or Editor instance).

Here is `ModifiableWebContent`'s in-line documentation, for your convenience:

```python
"""Data exchange class for web content that is modifiable by add-on filters

Attributes:
    body {str} -- HTML body
    head {str} -- HTML head
    addon_css {List[str]} -- List of subpaths, each pointing to an add-on-provided
                             CSS file
    addon_js {List[str]} -- List of subpaths, each pointing to an add-on-provided
                            JavaScript file

Important Notes:
    - When modifying the body or head of a web view, please make sure your changes
      only perform the minimum-required edits to make your add-on work. You should
      attempt to preserve as much of the original HTML as possible in order to
      avoid clashes with Anki's code or that of other add-ons. If possible, please
      only append or prepend your changes to the existing HTML. E.g.:
      
      > def on_view_will_set_web_content(web_content: ModifiableWebContent, view):
      >    web_content.body += "<your_html>"
      >    return web_content
    
    - The paths specified in `addon_css` and `addon_js` need to be accessible by
      Anki's media server. I.e., they should match the web export patterns you have
      specified via aqt.addons.AddonManager.setWebExports()
      
      E.g., to allow access to an `addon.js` and `addon.css` residing in a "web"
      subfolder in your add-on package, first register the corresponding web export:
      
      > from aqt import mw
      > mw.addonManager.setWebExports(__name__, r"web/.*(css|js)")
      
      Then append the subpaths to the corresponding web_content fields within the
      filter implemented by your add-on:
      
      > def on_view_will_set_web_content(web_content: ModifiableWebContent, view):
      >     web_content.addon_css.append(
      >         f"{mw.addonManager.addonFromModule(__name__)}/web/addon.css")
      >     web_content.addon_js.append(
      >         f"{mw.addonManager.addonFromModule(__name__)}/web/addon.js")
      >     return web_content
"""
```

Rather than calling the filters from each Anki view, as in #258, the changes in this PR move all calls to `AnkiWebView` (as suggested in the original PR's comments). The filter names are thus constructed dynamically, taking the form of `<caller>[_<view_name>]_will_set_web_content`.

The in-line docs in AnkiWebView dive more into this:

```python
"""Runs web content through addon-defined filters that are specific to
individual web views

Arguments:
    web_content {ModifiableWebContent} -- Data class storing web view HTML and
                                          addon-provided JS and CSS additions
    caller {Optional[Any]} -- Anki object that sets the web view's contents
                              (e.g. a Reviewer or Editor instance)
    view_name {Optional[str]} -- In case of web views that are reused across
                                 multiple GUI views – the name of that view
                                 (e.g. "deckbrowser", "overview", or "reviewer"
                                 for the main webview)

Notes:
    - Filter names are constructed dynamically from the web view title
      and, optionally, a `view_name` specified by the caller. The general
      form is: `<caller>[_<view_name>]_will_set_web_content`
      
      E.g.:
      - aqt.reviewer.Reviewer: `main_reviewer_will_set_web_content`
      - aqt.editor.Editor: `editor_will_set_web_content`
      
      For a list of all web content hooks please either see the
      "Web view content" section in qt/tools/genhooks_gui.py or
      look through the actual filters generated in gui_hooks.py

Returns:
    ModifiableWebContent -- Web content, with changes applied by add-ons
                            (if available)
"""
```

**Proof-of-concept**

If you'd like to give these changes a quick try, here's a small script I've written for Anki's debug console:

```python
from aqt import gui_hooks

def on_view_will_set_web_content(web_content, view):
    print(view)
    web_content.body += "<br><h1><font color=green>Hello World</font></h1>"
    web_content.head += "<script>console.log('hello world')</script>"
    return web_content

for view_name in ["editor", "main_deckbrowser", "main_overview", "main_reviewer", "main_reset", "stats", "top_toolbar", "bottom_toolbar", "previewer", "clayout_front", "clayout_back"]:
    hook = getattr(gui_hooks, f"{view_name}_will_set_web_content")
    hook.append(on_view_will_set_web_content)
```

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -1639,7 +1639,7 @@ where id in %s"""
             "reviewer.js",
         ]
         self._previewWeb.stdHtml(
-            self.mw.reviewer.revHtml(), css=["reviewer.css"], js=jsinc
+            self.mw.reviewer.revHtml(), css=["reviewer.css"], js=jsinc, caller=self
         )
         self._previewWeb.set_bridge_command(self._on_preview_bridge_cmd, "preview")
 
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -1374,7 +1374,7 @@ by clicking on one on the left."""
         d = CardInfoDialog(self)
         l = QVBoxLayout()
         l.setContentsMargins(0, 0, 0, 0)
-        w = AnkiWebView()
+        w = AnkiWebView(title="browser_card_info")
         l.addWidget(w)
         w.stdHtml(info + "<p>" + reps)
         bb = QDialogButtonBox(QDialogButtonBox.Close)
@@ -1543,7 +1543,7 @@ where id in %s"""
         self._previewWindow.silentlyClose = True
         vbox = QVBoxLayout()
         vbox.setContentsMargins(0, 0, 0, 0)
-        self._previewWeb = AnkiWebView()
+        self._previewWeb = AnkiWebView(title="previewer")
         vbox.addWidget(self._previewWeb)
         bbox = QDialogButtonBox()
 
@@ -2129,6 +2129,7 @@ update cards set usn=?, mod=?, did=? where id in """
         frm.fields.addItems(fields)
         self._dupesButton = None
         # links
+        frm.webView.title = "find_dupes"
         frm.webView.set_bridge_command(self.dupeLinkClicked, "find_dupes")
 
         def onFin(code):
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -215,10 +215,10 @@ class CardLayout(QDialog):
             "reviewer.js",
         ]
         pform.frontWeb.stdHtml(
-            self.mw.reviewer.revHtml(), css=["reviewer.css"], js=jsinc
+            self.mw.reviewer.revHtml(), css=["reviewer.css"], js=jsinc, caller=self
         )
         pform.backWeb.stdHtml(
-            self.mw.reviewer.revHtml(), css=["reviewer.css"], js=jsinc
+            self.mw.reviewer.revHtml(), css=["reviewer.css"], js=jsinc, caller=self
         )
         pform.frontWeb.set_bridge_command(self._on_bridge_cmd, "card_layout")
         pform.backWeb.set_bridge_command(self._on_bridge_cmd, "card_layout")
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -203,9 +203,9 @@ class CardLayout(QDialog):
 
     def setupWebviews(self):
         pform = self.pform
-        pform.frontWeb = AnkiWebView()
+        pform.frontWeb = AnkiWebView(title="clayout_front")
         pform.frontPrevBox.addWidget(pform.frontWeb)
-        pform.backWeb = AnkiWebView()
+        pform.backWeb = AnkiWebView(title="clayout_back")
         pform.backPrevBox.addWidget(pform.backWeb)
         jsinc = [
             "jquery.js",
--- a/qt/aqt/deckbrowser.py
+++ b/qt/aqt/deckbrowser.py
@@ -100,6 +100,8 @@ class DeckBrowser:
             self._body % dict(tree=tree, stats=stats, countwarn=self._countWarn()),
             css=["deckbrowser.css"],
             js=["jquery.js", "jquery-ui.js", "deckbrowser.js"],
+            caller=self,
+            view_name="deckbrowser",
         )
         self.web.key = "deckBrowser"
         self._drawButtons()
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -166,6 +166,7 @@ class Editor:
             _html % (bgcol, bgcol, topbuts, _("Show Duplicates")),
             css=["editor.css"],
             js=["jquery.js", "editor.js"],
+            caller=self,
         )
 
     # Top buttons
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -95,7 +95,6 @@ class Editor:
 
     def setupWeb(self) -> None:
         self.web = EditorWebView(self.widget, self)
-        self.web.title = "editor"
         self.web.allowDrops = True
         self.web.set_bridge_command(self.onBridgeCmd, "editor")
         self.outerLayout.addWidget(self.web, 1)
@@ -936,7 +935,7 @@ to a cloze type first, via Edit>Change Note Type."""
 
 class EditorWebView(AnkiWebView):
     def __init__(self, parent, editor):
-        AnkiWebView.__init__(self)
+        AnkiWebView.__init__(self, title="editor")
         self.editor = editor
         self.strip = self.editor.mw.pm.profile["stripHTML"]
         self.setAcceptDrops(True)
--- a/qt/aqt/gui_hooks.py
+++ b/qt/aqt/gui_hooks.py
@@ -151,6 +151,52 @@ class _AvPlayerWillPlayHook:
 av_player_will_play = _AvPlayerWillPlayHook()
 
 
+class _BottomToolbarWillSetWebContentFilter:
+    _hooks: List[
+        Callable[
+            ["aqt.webview.ModifiableWebContent", "aqt.toolbar.BottomBar"],
+            aqt.webview.ModifiableWebContent,
+        ]
+    ] = []
+
+    def append(
+        self,
+        cb: Callable[
+            ["aqt.webview.ModifiableWebContent", "aqt.toolbar.BottomBar"],
+            aqt.webview.ModifiableWebContent,
+        ],
+    ) -> None:
+        """(web_content: aqt.webview.ModifiableWebContent, bottom_toolbar: aqt.toolbar.BottomBar)"""
+        self._hooks.append(cb)
+
+    def remove(
+        self,
+        cb: Callable[
+            ["aqt.webview.ModifiableWebContent", "aqt.toolbar.BottomBar"],
+            aqt.webview.ModifiableWebContent,
+        ],
+    ) -> None:
+        if cb in self._hooks:
+            self._hooks.remove(cb)
+
+    def __call__(
+        self,
+        web_content: aqt.webview.ModifiableWebContent,
+        bottom_toolbar: aqt.toolbar.BottomBar,
+    ) -> aqt.webview.ModifiableWebContent:
+        for filter in self._hooks:
+            try:
+                web_content = filter(web_content, bottom_toolbar)
+            except:
+                # if the hook fails, remove it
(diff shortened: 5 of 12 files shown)


glutanimate:
A couple points for discussion:

- At the moment, `ModifiableWebContent` does not include the original list of JS and CSS files that Anki injects into each corresponding web view. That would be easy enough to change, but I can't think of any use cases that would require these lists to be add-on modifiable.
- This is more of a general point, but: Just like many other filters, the filters in this PR make use of a design pattern where a list is passed through multiple add-ons that incrementally append their changes to it.

    I feel like this could be a recipe for future add-on conflicts, as it's simple enough for (perhaps inexperienced) add-on authors to overwrite these lists rather than appending to them. Because there is no clearly defined API, it seems like a mistake like that could happen easily. In the long term, I think it would be nice if we could come up with a different solution that reduces the chances of that happening.

glutanimate:
Thanks for the feedback, Damien! I can definitely see your point about combining all of these into one hook with a context variable that subscribers can check. It certainly would make things easier for some add-ons, and it would also be nice to have the actual hook appear verbatim in the code rather than constructing it using something like a `getattr` trick. 

I did consider going with the single-hook approach initially, but my main reasons for opting for multiple hooks were:

- Having clear-cut type definitions on the caller (as you mention)
- Feeling iffy about having potentially dozens of add-ons be invoked on each stdHtml call. The performance impact would likely be negligible, but my main concern would be the increased error surface, as in: If one add-on misbehaves, you could suddenly have all of Anki's web views affected. 

    This is more of a concern with the filter pattern, as one link in the "filter chain" could end up breaking everything; but even with a regular hook, it still feels cleaner to me personally to have add-ons subscribe to a specific event. 
      
     It's true that some add-ons would slightly benefit from having a unified hook, but I think it's also arguable that there is a much larger number of add-ons out there that only need to modify one screen; and in cases where multiple web views need to be targeted, it honestly doesn't feel like a major effort on the add-on authors' side to simply subscribe to a couple more hooks.

    
To better illustrate where I'm coming from: There were multiple cases in the past where I had to deal with add-on conflicts arising from other add-ons mishandling return types or breaking these types of "chains" in some other way. The best example I can give are the multiple various patches to the link handler that add-ons used to perform without taking heed of preserving the return statement for other add-ons that needed it for their JS -> Py bridge. I mean, just take a look at the number of related PRs I had to submit to other add-ons here: https://github.com/ankitects/anki/pull/228 [1]

Basically, my point is that, I think it's best to be as explicit as possible with hooks, and try to find ways to have to put as little trust as possible into add-on authors handling things correctly. Because – let's face it – we all make mistakes (I've had my fair share of add-on compatibility mishaps early on), but it would be nice if those mistakes just affected my add-ons rather than resulting in a chain of erratic behavior that's very difficult to debug for other authors.

---

> In cases like this where the data being passed to the hook can be mutated directly, requiring the user to return the object they mutated is unnecessary

Good point! The return type started out as a regular tuple, and I completely forgot to switch to a regular hook once I changed that to the mutable type we now have.

> It would be nice if the context name passed to webview_did_receive_js_message was consistent with the names in this patch. In clayout.py you've picked different names. Are you aware of any add-ons that need to differentiate between the top and bottom webview?

No, you're right. I can't think of any at the moment. Will be updating that!

In its current state, the PR's context names also deviate for the main screens (e.g. `main_reviewer` instead of `webview_did_receive_js`'s plain `reviewer`). Would you like me to change these also?

-----

[1]: Slightly unrelated, but this is the first time I've looked at `webview_did_receive_js_message` properly, and to be honest: I would really feel a lot more comfortable if this was split up into multiple hooks for each web view in a similar fashion to above

 Just looking at the past experiences I've mentioned, it feels like a recipe for add-on conflicts – and because the same hook is now used globally across all JS -> Py bridges, it feels like the error surface is even greater than before. I think you did a great job at documenting how add-on authors *should* be handling this hook – and I think that will definitely make a difference  – but I still think we might be better served with some way of a stricter enforcement of good practices.

The other change I would highly suggest to implement – and this also factors into the decision to go with separate hooks – is to include a `caller` argument in the same manner as the hooks above. Looking at add-ons that currently use the "patched link handler" pattern, access to the parent Anki object (e.g. reviewer or editor) seems to be required in a lot of cases. In its current state, that can only be achieved with `webview_did_receive_js_message` if add-ons also attach to the corresponding Anki objects in some other way (e.g. by patching `__init__`).

dae:
I'm not really fond of the fact that in this implementation webview is calling other module's hooks on their behalf. If I see a hook starting with editor_, I expect to find it in the editor module. If the decision is made to use separate hooks for each module, this might either be better reworked so each module calls its own hook, or the hooks are named more like webview_will_set_web_content_for_editor. I still don't really like the latter solution though, because it means those method calls can't be typechecked.

Regarding some of the points you raised:

- I don't think 100 add-ons, most of which check the context and immediately return because it's not what they're interested in, would create a significant performance impact compared to the costs of IPC and DOM rendering.
- I appreciate your concerns about add-ons breaking things for other add-ons, but I don't think the situation is as bad as it was when it was a monkey-patch free-for-all. Take the existing webview_did_receive_js_message hook for example - if an add-on author makes a mistake and returns a truthy value indiscriminately, it will break every single link in Anki, as it will prevent the standard Anki handlers from firing. This is likely to be picked up by the author prior to releasing the add-on, and is easy for users to isolate as well, as it doesn't depend on a particular combination of add-ons. The same would apply in the web content case, if add-ons are replacing the head or body without basing their replacements on the existing text. It may even be worth passing the original css/js list in rather than using a separate list for add-ons, as that would break more obviously if someone overwrites the list.
- There is scope for instrumenting common hook paths in the future - a variable could enable debugging info printed out indicating which add-ons are mutating the state for a given call for example, and the code handling webview_did_receive_js_message could be modified so it printed the name of the function that decided not to proceed.
- Having webview_did_receive_js_message receive a caller sounds like a good idea.
- The reason I picked a string context for contextwebview_did_receive_js_message was to allow for add-on provided screens. But if the caller is being passed in, we could switch to an enum, and have a .Addon case where the caller instance can be checked to determine the specific add-on provided webview. The fewer string values, the better!

dae:
Instead of an enum, I've just created a few wrapper classes for the cases where one screen has webview hooks for different contexts:

https://github.com/ankitects/anki-addons/commit/9f75518cdf079e8e73d6b219a21a416a1992f911
https://github.com/ankitects/anki/commit/7fcb6b56722ad6dc0ef09450a4e5201ed0e16f4f

glutanimate:
> I'm not really fond of the fact that in this implementation webview is calling other module's hooks on their behalf. If I see a hook starting with editor_, I expect to find it in the editor module.

Same here, to be honest. I guess I was trying to go for a compromise between each module having its own hook and keeping the changes to a minimum. As you say, I think it would be best to go all the way in either direction (either having each module call its hook, or handling things the way `webview_did_receive_js_message` does).

Your points have definitely alleviated my concerns in regards to performance and add-on breakage. I agree that bad practices regarding Anki's existing web content would be fairly easy to spot by the add-on author / testers. However, handling of the add-on specific fields like `addon_css` and `addon_js` seems like a more tricky area, as breaking changes introduced by an add-on here would be more difficult to identify and trace back to its origin.

> It may even be worth passing the original css/js list in rather than using a separate list for add-ons, as that would break more obviously if someone overwrites the list.

I think this might be workable solution to this concern. But I'm not sure on the best way to implement this. Right now the split into `js` and `addon_js` (and the same for css) helps build the correct path in `AnkiWebView.webBundlePath`. In order to have both Anki web components and those served by add-ons included in the original css/js lists, my immediate thought would be to just qualify the paths to those files in the `AnkiWebView.stdHtml` calls more specifically, e.g. `_anki/jquery.js` instead of plain `jquery.js`. Would that be agreeable to you?

In that case I'd start reworking this PR to use the same approach as `webview_did_receive_js_message` as I now feel that's the simplest and most consistent way forward.

> There is scope for instrumenting common hook paths in the future - a variable could enable debugging info printed out indicating which add-ons are mutating the state for a given call for example, and the code handling webview_did_receive_js_message could be modified so it printed the name of the function that decided not to proceed.

This sounds amazing! I know there are probably much more pressing things to look into, but as an add-on author let me just say: With more and more add-on entry points moving towards hooks, I really think this can't come soon enough!

> Instead of an enum, I've just created a few wrapper classes for the cases where one screen has webview hooks for different contexts

Thanks, that looks great! I just left a couple comments regarding the browser web views.

(Really minor, but: Is there any chance the instance checks might slow things down? I know these can be comparatively slow for some more complex inheritance/abstraction scenarios. Though I guess this is once again a case where the delta would be simply dwarfed by IPC / DOM rendering.)

dae:
adding _anki is probably the easiest solution

Instance checks are about twice as slow as strings in the worst case, but we're talking 10ms vs 20ms for 100,000 checks on a modern machine, so it's basically negligible.

```python
from PyQt5.QtWidgets import QPushButton
from PyQt5.Qt import QApplication

app = QApplication([])

class Dummy(QPushButton):
    pass

class OtherDummy(QPushButton):
    pass

class PlainClass:
    pass

obj = OtherDummy()
str = "otherdummy"

from timeit import timeit

def instance_mismatch():
    isinstance(obj, Dummy)

def simple_instance_mismatch():
    isinstance(obj, PlainClass)

def instance_match():
    isinstance(obj, OtherDummy)

def type_mismatch():
    type(obj) == Dummy

def type_match():
    type(obj) == OtherDummy

def string_mismatch():
    str == "dummy"

def string_match():
    str == "otherdummy"

n = 100_000
print("i!=", timeit(instance_mismatch,number=n))
print("I!=",timeit(simple_instance_mismatch,number=n))
print("i==",timeit(instance_match, number=n))
print("t!=",timeit(type_mismatch, number=n))
print("t==",timeit(type_match, number=n))
print("s!=",timeit(string_mismatch, number=n))
print("s==",timeit(string_match, number=n))
```
```
(py514) dae@dip:~% $c/dtop/pyenv/bin/python speed.py
i!= 0.018241373999999977
I!= 0.01416395199999998
i== 0.00964160599999997
t!= 0.01411033500000003
t== 0.012979012999999984
s!= 0.008167562000000017
s== 0.008246688000000002
```

glutanimate:
Welp, thanks for going the extra mile and putting this to the test, Damien! That has definitely eliminated any (super minor) reservations I had towards instance checks :D

I reworked the PR as #445, so let's close this one in favor of the new one.

PR #440, opened 2020-02-07
Adding an option to an hook

Arthur-Milchior:
As stated in
https://github.com/ankitects/anki/pull/438#issuecomment-583168521 I'm
trying to change anki (as little as possible) to remove monkey
patching from my add-ons.

Add-on 1983204951 does monkey patching because it needs to take into
account whether the export window is called from browser or
not. Ideally, I'd like that, when I call the export window, I can add
an option. This option will then be passed to exporters_list_created
to decide how to edit the list. (In the present case, it means:
whether or not to add "export selected card)

Since exporters_list_created calls the legacy hook exportersList, I
can't directly edit it. So I prefer to add a new hook.

### DIFF
--- a/pylib/anki/exporting.py
+++ b/pylib/anki/exporting.py
@@ -408,7 +408,14 @@ class AnkiCollectionPackageExporter(AnkiPackageExporter):
 ##########################################################################
 
 
-def exporters() -> List[Tuple[str, Any]]:
+def exporters(options: Any = None) -> List[Tuple[str, Any]]:
+    """Options -- some option sent to hooks to decide how to update.
+    e.g. the string "browser" may be used to state that this was
+    called from the browser, and an option may be added to export
+    selected cards from browser.
+
+    """
+
     def id(obj):
         return ("%s (*%s)" % (obj.key, obj.ext), obj)
 
@@ -419,4 +426,5 @@ def exporters() -> List[Tuple[str, Any]]:
         id(TextCardExporter),
     ]
     hooks.exporters_list_created(exps)
+    hooks.exporters_list_created_with_options(exps, options)
     return exps
--- a/pylib/anki/hooks.py
+++ b/pylib/anki/hooks.py
@@ -185,6 +185,30 @@ class _ExportersListCreatedHook:
 exporters_list_created = _ExportersListCreatedHook()
 
 
+class _ExportersListCreatedWithOptionsHook:
+    _hooks: List[Callable[[List[Tuple[str, Any]], Any], None]] = []
+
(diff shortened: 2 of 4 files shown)


dae:
The existing new hook was only added in an earlier beta, so it would be safe to change it, remove legacy_hook, and call the old legacy hook manually. I'm not sure the pylib hook is the best place to be doing this however, and options: Any is not great, as different add-ons may pass in different types causing problems in the future.

Perhaps the easiest approach is just to integrate this functionality into Anki. Would you like to send me a patch that does that instead?

Regarding the sibling code, do you know of people actually taking advantage of the ability to export single cards from a note? I'm wondering whether 'export selected notes' would be more useful than 'export selected cards', since any missing cards just end up getting created later anyway.

Arthur-Milchior:
I thought about changing it. But the legacy was my problem. I didn't thought about calling it manually, that certainly would solve the problem.
I believe that integrating this feature would be great; that something which seems natural to use. With 3069 download, it's a sucesful add-on. And it's one I need to tell people to install each time I want to help them debugging a template without having them to share too many data

What do you mean by "sibling code" ? Do you mean the option to decide whether we want to export siblings of selected card ?
I have absolutely no feedback about the use of the addon, appart from the 15 comments on ankiweb. 
 Since I'm not going to do a user test for this feature, I'll just do wild guess (I could ask advice on the subreddit. Sometime it helps. But I don't trust the subreddit community to actually represents the average user)
So, I'll just list pro of card and notes:

Exporting cards: 
* you export what you select
* when missing siblings are generated, they get generated either in the same deck as selected cards, or in default deck (in the case where we selected two siblings in different decks, with a missing third sibling). This probably makes more sens to an average user. Indeed, if I select a card in deck A, I don't expect to find more card in deck B, which may occur if A has a not selected sibling in B.
* You don't export more data than expected, which is great for privacy. However, the only data which won't get exported this way is: deck of missing siblings, queue/type/due of missing siblings; so probably nothing important. 

Exporting note:
* no loss of data for user who don't understand difference between card and note
* no need to manually search for sibling of cards when exporting from deck if we want all notes

My intuition is to export note and not cards. And not to let user option to change the default, because that'll generate more confusion and "exporting cards" is probably not useful. 
Actually,the reason why I export "card" by default in my add-on is only that this was what did the first version, and I only later realized that exporting notes was better.
I think that at some point, I mentionned that a similar problem occurs when you export a deck containing a card and not all of those siblings. Should it get corrected too ?

dae:
Yep, I agree - let's just provide an export note option.

The missing cards in a deck export should probably be fixed at one point, but that can wait for now.

Arthur-Milchior:
Corrected

dae:
Closing in favour of #443

PR #441, opened 2020-02-08
Allow to change main window when overview/deckbrowser is shown

Arthur-Milchior:
This would allow to remove monkey patch to "Deck name in title 2.1". That's a 2.0 add-on I uploaded as 2.1, ad it seems quite successful. 1817 download.
(I'm trying to consider add-ons which are both successful and easy to consider with hooks)

### DIFF
--- a/qt/tools/genhooks_gui.py
+++ b/qt/tools/genhooks_gui.py
@@ -20,6 +20,17 @@ hooks = [
     # Reviewing
     ###################
     Hook(
+        name="overview_did_refresh",
+        args=["overview: aqt.overview.Overview"],
+        doc="""Allow to update the overview window. E.g. add the deck name in the
+        title.""",
+    ),
+    Hook(
+        name="deck_browser_did_render",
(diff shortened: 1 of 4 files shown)


qt/tools/genhooks_gui.py
```diff
@@ -19,6 +19,17 @@
...
     # Reviewing
     ###################
+    Hook(
+        name="overview_refreshed",
+        args=["overview: aqt.overview.Overview"],
+        doc="""Allow to update the overview window. E.g. add the deck name in the
+        title.""",
+    ),
+    Hook(
+        name="deckBrowser_rendered",
```
dae: please use deck_browser_did_render instead

and overview_did_refresh

otherwise looks good, thanks!

qt/tools/genhooks_gui.py
```diff
@@ -19,6 +19,17 @@
...
+        args=["overview: aqt.deckbrowser.DeckBrowser"],
```
dae: whoops, arg name is wrong here

dae:
Thanks!

PR #442, opened 2020-02-09
Type

Arthur-Milchior:
In order to type another PR I intended to make I thought it'd be nice to have more precise type for ids and other things. So this PR add them.
In one method, it change the name of a variable. Because during the execution, the type of the variable changed, which led to a typing error
I think you may want to reject/change the two last commits, everything else should not lead to a lot to a lot of discussion. Indeed, I added a type "DeckId" and a type "DConfId". Both are sometime int, sometime string. I decided to apply them everywhere before using the value. So that there is a consistency with type. Ankidroid is capable of dealing with deck id which are ints. I don't know for IOS and AnkiWeb

### DIFF
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -15,6 +15,8 @@ from anki.notes import Note
 from anki.sound import AVTag
 from anki.utils import intTime, joinFields, timestampID
 
+CardId = int
+
 # Cards
 ##########################################################################
 
@@ -32,9 +34,10 @@ class Card:
     timerStarted: Optional[float]
     lastIvl: int
     ord: int
+    id: CardId
 
     def __init__(
-        self, col: anki.collection._Collection, id: Optional[int] = None
+        self, col: anki.collection._Collection, id: Optional[CardId] = None
     ) -> None:
         self.col = col
         self.timerStarted = None
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -10,7 +10,7 @@ from typing import List, Optional
 import anki  # pylint: disable=unused-import
 from anki import hooks
 from anki.consts import *
-from anki.models import NoteType, Template
+from anki.models import FieldOrd, NoteType, Template
 from anki.notes import Note
 from anki.sound import AVTag
 from anki.utils import intTime, joinFields, timestampID
@@ -33,7 +33,7 @@ class Card:
     _note: Optional[Note]
     timerStarted: Optional[float]
     lastIvl: int
-    ord: int
+    ord: FieldOrd
     id: CardId
 
     def __init__(
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -35,6 +35,8 @@ class Card:
     lastIvl: int
     ord: FieldOrd
     id: CardId
+    odid: anki.decks.DeckId
+    did: anki.decks.DeckId
 
     def __init__(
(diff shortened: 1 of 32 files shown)


dae:
While it's great that you've done each of these in a separate commit, in big changes like these it might be more efficient if you start by submitting a pull request for just one or two things, and going from there based on feedback.

Aliasing a variable to a built in type doesn't buy us much, because the types can be used interchangeably. For example:

```python
(py514) dae@dip:~% mypy test.py
Success: no issues found in 1 source file
(py514) dae@dip:~% cat test.py
OneType = int
AnotherType = int

a: OneType = 5

def func(b: AnotherType) -> AnotherType:
    return b + 5

func(a)
```

So at the moment these changes don't make the code any safer.

To achieve type safety, you would need to use NewTypes: https://mypy.readthedocs.io/en/stable/more_types.html#newtypes

More work would be required to support this, as you'd need to wrap any plain type in the NewType object prior to passing it to a function that expects the NewType.

I'm not necessarily against doing this for the various IDs in the future, as they are an area that can be confused. But we would not be able to get the full benefit of such a change at the moment, as the note types and decks are currently accessed as JSON instead of in a properly typed structure. If you wanted to give NewTypes a try, perhaps Usn would be a good first step.

For the string values, I think adding separate types is probably more trouble than it's worth, and they're best left as just strings.

dae:
And re the places where the IDs are sometimes strings and sometimes ints, while I agree that it's really ugly and error prone and something I want to fix in the future, there is a big potential for introducing bugs if not changed carefully. From a brief scan of your change, it looks like it could lead to a deck conf being stored in the dict twice for example, once with a string ID and once with an integer ID.

dae:
Actually, I see you've converted the dict to int keys at load stage which should avoid that issue - assuming there's nowhere in the code that still assumes the keys should be strings.

Arthur-Milchior:
One of the reason of doing plenty of commits, apart from readability, is that you can decide which commits you keep and which one you don't like. Simple cherry-picking or interactive rebase could do it.
With the exception of lines which are modified twice; but usually those lines have the same kind of modification, so I assume you'll want both or reject both.
I can't tell that I did read the entire code, but I tested anki with this version and did see a single error. An add-on which assumed that the key was a string. That was an add-on of mine; I corrected it; it should never have dealt with the key directly; now it uses deck manager methods and there is no more problem.
The main idea is that no key should leave the deck manager as string. And the deck manager accepts string everywhere. 

I'll look at new types later.

dae:
The reason I'm suggesting you start small is so you can avoid wasted work - if you do 20 changes and I only cherry pick one, or the others all need changing based on the feedback from the first one, that's a lot of wasted effort.

I don't think we can rely on other parts of the code not using string keys - if the dict is changed to have integer keys, it really needs to be renamed at the same time to ensure no existing code accesses it in the wrong way. But I'd like to rework the deck and note type managers in the next few months, so it may make the most sense to wait until then before making such a change.

dae:
Just an update on this - the string key issue should be resolved once the deck handling is moved into a separate sql table as the deck conf has been.

I'm afraid I think it's best to say no to the rest of this PR, as it would need work to bring up to date, and the aliases are not adding any type safety. In the future when the code is refactored to use a proper typed object instead of untyped JSON, then it will make sense to use a more specific type. And for what it's worth, the Rust backend is already using NewTypes - the compiler will give an error if you accidentally pass a millisecond timestamp to a function that expects one in seconds, or a card ID to a function that expects a note ID for example.

PR #443, opened 2020-02-09
Export from browser in anki

Arthur-Milchior:
I must admit that the way information goes from aqt to anki seems strange to me. Creating the anki object first, then setting its value later, while the __init__ can take a parameter did (but this is never used)
I decided to follow current code and do the same for cids. 
As we discussed otherwhere, siblings are directly exported

### DIFF
--- a/pylib/anki/decks.py
+++ b/pylib/anki/decks.py
@@ -470,6 +470,9 @@ class DeckManager:
             dids.append(id)
         return self.col.db.list("select id from cards where did in " + ids2str(dids))
 
+    def for_card_ids(self, cids: List[int]) -> List[int]:
+        return self.col.db.list(f"select did from cards where id in {ids2str(cids)}")
+
     def _recoverOrphans(self) -> None:
         dids = list(self.decks.keys())
         mod = self.col.db.mod
--- a/pylib/anki/exporting.py
+++ b/pylib/anki/exporting.py
@@ -8,7 +8,7 @@ import shutil
 import unicodedata
 import zipfile
 from io import BufferedWriter
-from typing import Any, Dict, List, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 from zipfile import ZipFile
 
 from anki import hooks
--- a/pylib/anki/exporting.py
+++ b/pylib/anki/exporting.py
@@ -170,10 +170,10 @@ class AnkiExporter(Exporter):
         Exporter.__init__(self, col)
 
     def deckIds(self) -> List[int]:
-        if not self.did:
-            return []
-        else:
+        if self.did:
             return [self.did] + [x[1] for x in self.src.decks.children(self.did)]
+        else:
+            return []
 
     def exportInto(self, path: str) -> None:
         # sched info+v2 scheduler not compatible w/ older clients
--- a/pylib/anki/exporting.py
+++ b/pylib/anki/exporting.py
@@ -21,9 +21,17 @@ from anki.utils import ids2str, namedtmp, splitFields, stripHTML
 class Exporter:
     includeHTML: Union[bool, None] = None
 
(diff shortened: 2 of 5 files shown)


Arthur-Milchior:
Added an extra feature: if no card selected, the export window can't open (this occur only for empty search)

pylib/anki/collection.py
```diff
@@ -303,6 +303,12 @@ def beforeUpload(self) -> None:
     # Object creation helpers
     ##########################################################################
 
+    def siblings(self, cids):
```
dae: While not a bad place to put this, you could just use the existing browser .selectedNotesAsCards() method instead

pylib/anki/decks.py
```diff
@@ -470,6 +470,9 @@ def cids(self, did: int, children: bool = False) -> Any:
             dids.append(id)
         return self.col.db.list("select id from cards where did in " + ids2str(dids))
 
+    def of_cids(self, cids: List[int]) -> List[int]:
```
dae: for_card_ids() would be a bit clearer

pylib/anki/exporting.py
```diff
@@ -21,9 +21,18 @@
 class Exporter:
     includeHTML: Union[bool, None] = None
 
-    def __init__(self, col: _Collection, did: None = None) -> None:
+    did: Optional[int]
+    cids: Optional[List[int]]
```
dae: If my understanding is correct, if an argument to `__init__` specifies a type, it is not necessary to also define it at the class level. The cases where you see types specified at the class level in current code are usually cases where __init__ is untyped, or where the instance variable is being set outside of `__init__`

qt/aqt/exporting.py
```diff
@@ -38,7 +40,12 @@ def setup(self, did):
         self.frm.format.activated.connect(self.exporterChanged)
         self.exporterChanged(idx)
         # deck list
-        self.decks = [_("All Decks")] + sorted(self.col.decks.allNames())
+        if self.cids:
+            self.decks = [_("Browser's selection")]
```
dae: "Browser selection" would be more natural English, but why not just hide the widget completely when called from the Browse screen? Then you wouldn't need to think about the currentIndex() at all in the code below.

pylib/anki/collection.py
Arthur-Milchior: Oh right. I never saw it. Thanks

qt/aqt/exporting.py
Arthur-Milchior: Pardon my French. 
I'll change it this way. I didn't really give a thought about it and added to the code already existing.

Arthur-Milchior:
I left "Browser selection" to give the user a feedback. So they know what they export in case they are confused. I removed the remaoning options

pylib/anki/exporting.py
dae: Sorry if I wasn't clear - it'd be better to keep the type for cid in the `__init__` call, and remove the did/cid definitions from the class, as the constructor arg should be enough.

Otherwise looks good, thanks.

Arthur-Milchior: Corrected

dae:
Thanks Arthur!

Arthur-Milchior:
As far as I can tell, this was not added to the documentation. My english is not good enough for me to write it myself

dae:
Not every menu item is documented in the manual, and this feature is fairly self-explanatory, so I'm not sure it needs to be?

Arthur-Milchior:
I was looking at it because of this question: https://www.reddit.com/r/Anki/comments/f3pi29/my_first_feature_incorporated_in_anki/fhkdukj/?context=3 and also https://github.com/Arthur-Milchior/anki-export-from-browser/issues/7#event-3038505232 a little but
But yes, I believe it to be self explanatory

PR #444, opened 2020-02-12
Pre flush hooks

Arthur-Milchior:
In this PR I added two similar hooks. Applying some actions to card/notes before they get flushed.
Example of use cases are given in both commits.

I should note that in this case, I did not consider my most downloaded add-ons. But the one where I use monkey patching and that are used a lot are:
* enhance main window, 
which entirely change the way the deck browser is displayed. So it seems had to create ways to edit it. Unless I totally change the way nodes are computed in the scheduler (which would be a good thing according to me. I'm not a fan of using tuples to store information)
* open the same window multiple time. 
It seems hard to use hooks here, since this add-on entirely change the system used to open/close windows and actually replace the window manager class by a new one.

### DIFF
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -84,7 +84,7 @@ class Card:
         self._render_output = None
         self._note = None
 
-    def flush(self) -> None:
+    def _preFlush(self) -> none:
         self.mod = intTime()
         self.usn = self.col.usn()
         # bug check
@@ -95,6 +95,9 @@ class Card:
         ):
             hooks.card_odue_was_invalid()
         assert self.due < 4294967296
+
+    def flush(self) -> None:
+        self._preFlush()
         self.col.db.execute(
             """
 insert or replace into cards values
@@ -121,16 +124,8 @@ insert or replace into cards values
         self.col.log(self)
 
     def flushSched(self) -> None:
-        self.mod = intTime()
-        self.usn = self.col.usn()
+        self._preFlush()
         # bug checks
-        if (
-            self.queue == QUEUE_TYPE_REV
-            and self.odue
-            and not self.col.decks.isDyn(self.did)
-        ):
-            hooks.card_odue_was_invalid()
-        assert self.due < 4294967296
         self.col.db.execute(
             """update cards set
 mod=?, usn=?, type=?, queue=?, due=?, ivl=?, factor=?, reps=?,
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -84,7 +84,8 @@ class Card:
         self._render_output = None
         self._note = None
 
-    def _preFlush(self) -> none:
+    def _preFlush(self) -> None:
+        hooks.card_will_flush(self)
         self.mod = intTime()
         self.usn = self.col.usn()
         # bug check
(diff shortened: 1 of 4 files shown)


pylib/anki/cards.py
```diff
@@ -84,13 +84,17 @@ def load(self) -> None:
...
         self.mod = intTime()
         self.usn = self.col.usn()
-        # bug check
         if self.queue == 2 and self.odue and not self.col.decks.isDyn(self.did):
             hooks.card_odue_was_invalid()
         assert self.due < 4294967296
+
+    def flush(self) -> None:
+        self._preFlush()
+        # bug check
```
dae: comment is in wrong place

pylib/tools/genhooks.py
```diff
@@ -67,6 +67,16 @@
...
+    Hook(
+        name="note_is_being_flushed",
+        args=["note: Note"],
+        doc="Allow to change a note before adding it to the database.",
+    ),
+    Hook(
+        name="card_is_being_flushed",
+        args=["card: Card"],
+        doc="Allow to change a card before adding it to the database.",
+    ),
```
dae: please use 'card_will_flush' and 'note_will_flush' for consistency with the other hook names

pylib/tools/genhooks.py
```diff
@@ -67,6 +67,16 @@
         Your add-on can check filter_name to decide whether it should modify
         field_text or not before returning it.""",
     ),
+    Hook(
+        name="note_is_being_flushed",
+        args=["note: Note"],
+        doc="Allow to change a note before adding it to the database.",
```
dae: "before it is added/updated in the database"

pylib/anki/cards.py
Arthur-Milchior: Right. Corrected

pylib/tools/genhooks.py
Arthur-Milchior: I guess I don't really know what "flush" mean. I thought the process already started when the "flush" is called, and ends when the method "flush" returns.
But to be honest, I have no strong opinion on the hooks name. I just hope that I'll eventually be able to create correct hooks name without needing to change them after my PR

pylib/tools/genhooks.py
Arthur-Milchior: Right. Corrected

pylib/tools/genhooks.py
dae: The rule is pretty simple - if a hook is before some action, use 'will'. If the hook is after some action, use 'did'. :-)

'flush' comes from 'flush out', as in send/push out any pending changes that have not yet been saved to disk

It's probably not worth changing here (and I frequently make the same mistake myself), but in new methods, it'd be good if we could snake_case instead of camelCaps.

Could I trouble you to resolve the conflict that merging your earlier PR has caused? Thanks for that by the way :-)

Arthur-Milchior: You mean, for _preFlush ? 
I wanted cards and notes to be consistent in method name. 
Otherwise I used snake. Its make develop wich transform it into camlCase

dae: Yes, I meant _preFlush. Again, not a big deal, but better to use snake_case in future methods you add. Thanks for the patch!

PR #445, opened 2020-02-12
Allow add-on authors to easily inject their own content into Anki's web views – take 3

glutanimate:
Continued from #439 

Adds a `webview_will_set_content` hook that allows add-on authors to modify web view content before it gets rendered. Uses the same `context` passing approach as `webview_did_receive_js_message`.

(also preserves some of last PR's changes to how web view titles are set)

### DIFF
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -103,50 +103,52 @@ class AnkiWebPage(QWebEnginePage):  # type: ignore
 
 @dataclasses.dataclass
 class WebContent:
-    """Stores all dynamically modified content that a particular web view will be
-    populated with.
+    """Stores all dynamically modified content that a particular web view
+    will be populated with.
 
     Attributes:
         body {str} -- HTML body
         head {str} -- HTML head
-        css {List[str]} -- List of media server subpaths, each pointing to a CSS file
-        js {List[str]} -- List of media server subpaths, each pointing to a JS file
+        css {List[str]} -- List of media server subpaths,
+                           each pointing to a CSS file
+        js {List[str]} -- List of media server subpaths,
+                          each pointing to a JS file
 
     Important Notes:
-        - When modifying the attributes specified above, please make sure your changes
-        only perform the minimum requried edits to make your add-on work. You should
-        avoid overwriting or interfering with existing data as much as possible,
-        instead opting to append your own changes, e.g.:
+        - When modifying the attributes specified above, please make sure your
+        changes only perform the minimum requried edits to make your add-on work.
+        You should avoid overwriting or interfering with existing data as much
+        as possible, instead opting to append your own changes, e.g.:
         
             def on_webview_will_set_content(web_content: WebContent, context):
                 web_content.body += "<my_html>"
                 web_content.head += "<my_head>"
-                web_content.css.append("my_addon.css")
-                web_content.js.append("my_addon.js")
 
-        - The paths specified in `css` and `js` need to be accessible by Anki's media
-          server. Web components shipping with Anki are located under the `_anki`
-          subpath.
+        - The paths specified in `css` and `js` need to be accessible by Anki's
+          media server. All list members without a specified subpath are assumed
+          to be located under `/_anki`, which is the media server subpath used
+          for all web assets shipped with Anki.
           
-          Add-ons may expose their own web components by utilizing
-          aqt.addons.AddonManager.setWebExports(). Web exports registered in this
-          manner may then be accessed under the `_addons` subpath.
+          Add-ons may expose their own web assets by utilizing
+          aqt.addons.AddonManager.setWebExports(). Web exports registered
+          in this manner may then be accessed under the `/_addons` subpath.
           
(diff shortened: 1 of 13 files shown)


glutanimate:
Demo snippet:

```python
from aqt import gui_hooks

def on_webview_will_set_content(web_content, context):
    print(context)
    print(web_content)

gui_hooks.webview_will_set_content.append(on_webview_will_set_content)
```

A couple of things:

-  Did the `js` parameter being a string in https://github.com/ankitects/anki/pull/445/commits/c86e55d4513a4e2b829554072e61576820b7eacb serve any particular purpose that I might have overlooked?

- I just realized that  the "`_anki` prepending solution" might cause quite a few add-ons to break that currently re-implement existing `stdHtml` calls. I think it would be worthwhile to explore some alternatives. One approach we could go with is to assume that all members of the `css` and `js` lists are under `/_anki` by default, and only set the web bundle path to something else if the subpath in question starts with `/`. What do you think?

dae:
re js=" ", I don't remember, but if I had to guess, perhaps the code in the past was not checking 'js is None', but 'if not js', and the space was a way of avoiding the defaults being set.

And yeah, I think it is probably safer to assume _anki/ if the file doesn't start with '/' - could I trouble you to make that change?

qt/aqt/webview.py
```diff
@@ -96,14 +97,74 @@ def _onCmd(self, str):
...
+        - When modifying the attributes specified above, please make sure your changes
+        only perform the minimum requried edits to make your add-on work. You should
+        avoid overwriting or interfering with existing data as much as possible,
+        instead opting to append your own changes, e.g.:
+        
+            def on_webview_will_set_content(web_content: WebContent, context):
+                web_content.body += "<my_html>"
+                web_content.head += "<my_head>"
+                web_content.css.append("my_addon.css")
+                web_content.js.append("my_addon.js")
```
dae: I think having a bare filename here (and in the new hook's docs) might be a bit confusing - perhaps the css/js examples are best omitted from this example, or contain the full code below as well?

glutanimate:
> And yeah, I think it is probably safer to assume _anki/ if the file doesn't start with '/' - could I trouble you to make that change?

Sure thing! How would you feel about also adding support for remotely hosted JS and CSS? I.e. special treatment for list members matching `^http(s)://.*`.

qt/aqt/webview.py
glutanimate: Good point! I think removing the css/js examples is the best approach here, since they're somewhat redundant with the examples below anyway.

dae:
I'd prefer add-ons use local files, as that will be faster and won't break when the net is down.

qt/aqt/webview.py
glutanimate: ok, I've updated the examples both here and in the new hook's docs. Please let me know if there's anything further you would like me to tweak!

glutanimate:
Alright, all web asset paths should now be the same as before on `stdHtml`'s side.

> I'd prefer add-ons use local files, as that will be faster and won't break when the net is down.

Makes perfect sense. I was thinking along the lines of quickly testing things by throwing in a CDN-hosted script, but I could see how that might encourage bad practices.

dae:
LGTM, thanks!

Arthur-Milchior:
Thank you a lot @glutanimate for this work. I think it's going to be really useful. It was becoming a nightmare to deal with interactions with other add-ons dealing with the editor. Especially in the add-on I expect to release soon; allowing users to resize images in editor (I could not have imagined that it would be so complex !)

@dae do you have any idea when a version 21 with this code will be published ? I can't wait to use it

glutanimate:
@Arthur-Milchior glad you like these changes! Looking forward to seeing them put into practice, and your add-on (which looks really cool, by the way!) could definitely be a perfect testbed for them :)

dae:
Beta coming soon; it'll be at least a few weeks before it's ready for a stable release.

Arthur-Milchior:
While I'm editing add-on to port this feature, I realize there is one feature which is really missing. Being able to set some constants/variable/parameters.
There is multiple cases where most of the js code is fixed, appart from some value which is fixed by python. Often, the property comes from the add-on configuration and does not change until anki restarts. For example in "Edit card during review", the js files contains `%(var_name)s`. The file content is interpreted by python and var_name is replaced by its value. This become impossible to do directly with this feature (By the way, I don't like this way of coding, so I'm happy that it is disappearing)

Sometime, the constant is simply set (e.g. using web.eval()) and then JS believes use the constant values in the code. I clearly prefer coding this way. But it becames quite hard too. Because if code is evaluated in this hook, it seems it can not set constants (since the whole HTML page is not even set). And if add the constant initialization in the head, it will be executed after the js files, which means the js files can't use the variables in code executed directly.

I don't know how to solve this problem currently. But that's something that should probably be considered to improve this feature.

Appart from that, thank you a lot Glutanimate. It really makes some add-ons quite more elegant, even if it sometimes need rewriting quite some code

PR #446, opened 2020-02-13
small fix: store return value for hook in preview

ijgnd:
I started reading about the new hook system and I think something is missing here.


dae:
good catch, thanks!

PR #447, opened 2020-02-15
Add browser_will_build_tree filter

glutanimate:
Allows add-ons to populate the browser sidebar tree with their own items, and/or take over specific construction stages in their entirety.

This would enable [Hierarchical Tags](https://github.com/glutanimate/hierarchical-tags/blob/master/src/hierarchical_tags/browser.py), [Customize Sidebar](https://github.com/yukach/anki21-addon-customize-sidebar/blob/master/customize_sidebar.py#L160), and similar add-ons to modify the sidebar without having to monkey-patch the private `_*tree` browser methods.

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -11,6 +11,7 @@ import sre_constants
 import time
 import unicodedata
 from dataclasses import dataclass
+from enum import Enum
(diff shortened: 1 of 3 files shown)


dae:
Looks good, thanks!

PR #449, opened 2020-02-16
Add hooks for modifying HTML body sections in the deck browser and overview screens

glutanimate:
Allows add-on authors to specifically target and modify individual sections of the deck browser and overview HTML body at string composition time.

~~For the time being the changes in this PR only apply to the deck browser, but if they look good to you, I'd like to follow up with more of the same for the overview screen~~ Overview changes added

My goal with this PR is to offer add-on authors an easy way to modify individual sections of these screens without running into conflicts as the ones described in #258.

With #445 being merged, this might seem a bit redundant. However, the main advantage offered by the hooks in this PR is that they don't require add-on authors to perform error-prone regex replacements or DOM modifications through JS to insert their content at specific positions in each view.

My thinking is that both types of hooks serve different purposes: `webview_will_set_content` for large-scale modifications and/or extending web views with additional JS and CSS assets, and these section hooks for performing more targeted HTML modifications and additions.

**Sample add-on (updated)**

```python
from aqt import gui_hooks
from aqt.deckbrowser import DeckBrowser, DeckBrowserContent
from aqt.overview import Overview, OverviewContent

def on_deck_browser_will_render_content(
    deck_browser: DeckBrowser,
    content: DeckBrowserContent
):
    content.stats += "<div>my html</div>"

def on_overview_will_render_content(
    overview: Overview,
    content: OverviewContent
):
    content.table += "<div>my html</div>"

gui_hooks.deck_browser_will_render_content.append(
    on_deck_browser_will_render_content
)
gui_hooks.overview_will_render_content.append(
    on_overview_will_render_content
)

```

### DIFF
--- a/qt/aqt/deckbrowser.py
+++ b/qt/aqt/deckbrowser.py
@@ -5,7 +5,7 @@
 from __future__ import annotations
 
 from copy import deepcopy
-from enum import Enum
+from dataclasses import dataclass
 from typing import Any
 
 import aqt
@@ -24,10 +24,20 @@ class DeckBrowserBottomBar:
         self.deck_browser = deck_browser
 
 
-class DeckBrowserSection(Enum):
-    TREE = 0
-    STATS = 1
-    WARN = 2
+@dataclass
+class DeckBrowserContent:
+    """Stores sections of HTML content that the deck browser will be
+    populated with.
+    
+    Attributes:
+        tree {str} -- HTML of the deck tree section
+        stats {str} -- HTML of the stats section
+        countwarn {str} -- HTML of the deck count warning section
+    """
+
+    tree: str
+    stats: str
+    countwarn: str
 
 
 class DeckBrowser:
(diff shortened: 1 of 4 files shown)


glutanimate:
This is a bit tangential, but: How would you feel about adding more CSS classes and IDs to some of the elements in these web views? I think this could make it much easier for add-on authors to modify content via web technologies.

glutanimate:
Actually, giving these changes a bit more thought: It might make sense to refactor the three hook calls into one, with a dataclass bundling the different sections into one. This would reduce the number of calls, spare us the enum argument, and make things easier for authors, I feel. It would also mean that we could use a plain hook instead of a filter. What do you think?

dae:
Yeah, I think a dataclass would be simpler all round, and happy to accept additional classes/ids in the HTML as well.

glutanimate:
:+1: reworked the changes to use a dataclass instead, and added support for the overview screen

> happy to accept additional classes/ids in the HTML as well.

Great! Will follow up in a separate PR soon.

dae:
Thanks!

PR #450, opened 2020-02-17
Current card get selected in browser

Arthur-Milchior:
I'm pretty sure it was the way it worked before. I'm surprised that
it's not the case anymore.

If you open the browser from the reviewer, the current card get
selected if it exists. The current note is still entirely displayed.

Personally, I want to know easily which is the current card. Opening
the browser is the easiest way to do it; assuming I can see the
current card selected


dae:
I think it may have broken as part of the changes for Qt5. Thanks!

PR #359, opened 2019-12-05
list of decks contains default iff deck chooser contains default

Arthur-Milchior:
I don't exactly know whether it'd be considered as a bug or not. In a collection with a deck A with cards in it, the users does not see "default" in the deck browser. But they may see it in the study deck window. Which is useless since there is nothing to study anyway. So I ensured both windows where consistent.

I must remark that it also means that default can't be chosen when adding/moving a card. If you prefer Default to be still accessible in those cases, I can easily edit the code to allow it.

PR #452, opened 2020-02-19
No default deck in browser

Arthur-Milchior:
This PR is similar to https://github.com/ankitects/anki/pull/359
It deals with decks in browser. Both left window and filter

I realized a mistake in the previous PR (not a big one). If the default deck has no card but has child, it should have been displayed. Because in this case it's displayed in the deck browser, and you may want to review all children of default deck.

To help with this, I created two methods. One to check whether default deck should be displayed. One to check whether current deck should be displayed. It has three options. Show all deck. Don't show empty default deck (as it is now in reviewer). Don't show empty default deck without children (as in deck browser)
I changed the option of the method all. 0 and 1 means the same thing as false and true. 2 is the new value.

### DIFF
--- a/pylib/anki/decks.py
+++ b/pylib/anki/decks.py
@@ -92,6 +92,16 @@ defaultConf = {
     "usn": 0,
 }
 
+# How to select list of decks
+
+# difference between WITHOUT_EMPTY_DEFAULT and
+# WITHOUT_EMPTY_LEAF_DEFAULT is that the first contains default when
+# it has a child. This is required for deck tree.
+
+WITHOUT_EMPTY_DEFAULT = 0
+ALL_DECKS = 1
+WITHOUT_EMPTY_LEAF_DEFAULT = 2
+
 
 class DeckManager:
     decks: Dict[str, Any]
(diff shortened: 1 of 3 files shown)


dae:
Thanks Arthur. Since you're changing the type signature anyway, it probably makes more sense to use an enum here - I'll give it a tweak after applying it.

dae:
Actually, thinking about this more - is there a need for these two separate exclusion cases? Perhaps it would be simpler just to show the default deck either if it has cards or child decks in all cases? I'd like to get rid of the default deck entirely in the future, so making the handling code more complicated seems like we're moving in the wrong direction.

Arthur-Milchior:
I didn't want to remove the feature that already existed. That's why I put three settings. I would agree that the general rule "default is here if it has children or cards or there is no other deck" would be a good general rule and then keeping to using a Boolean is okay. This change the effect of the methods all/allNames in the case where default is empty and has children, I don't believe it's actually a problem

dae:
Yep, that's probably the best way forward - would you like to make another pull request? Sorry I merged the previous one prematurely.

PR #453, opened 2020-02-20
Add hooks for js->python cmd/link

Arthur-Milchior:
Examples of uses are given in the documentation


glutanimate:
Hmm, I might be missing something, but wouldn't `webview_did_receive_js_message` cover your use cases?

Arthur-Milchior:
Totally. I just never saw this hook

PR #454, opened 2020-02-20
Display add-on name in add-on configuration window title

glutanimate:


dae:
Will get to this one after the recent i18n changes have settled down a bit

glutanimate:
ok, thanks for letting me know!

Since we're on the topic of translations, can I pick your brains about something quickly? I've had some great contributions submitted recently regarding adding i18n to my add-ons (https://github.com/glutanimate/anki-addon-builder/pull/10). Image Occlusion would be the first add-on to be localized in that way (https://github.com/glutanimate/image-occlusion-enhanced/pull/125). 

The current implementation uses `gettext`, but with Anki moving towards Fluent, do you thing it would be worthwhile exploring in that way?

I guess I should probably preface this saying that I have very little clue about i18n systems – this will be my first dive into that area. Also, I'm fairly strapped for time (as usual), so if `gettext` is good enough of a solution, and the advantages Fluent offers aren't huge for smaller-scale projects, I would rather stick to that.

dae:
Fluent solves a bunch of pain points that gettext has, but the tradeoff is it's a young technology, and the tooling options are limited. Anki's Fluent integration isn't currently set up for external use, so you'd need to bundle the Python libraries, which would add a bunch of extra dependencies to your add-on - gettext may be easier at this point.

dae:
Ok, new docs here - feedback welcome!

https://ankitects.github.io/translating/#/anki/developers

glutanimate:
Thanks, Damien! Didn't have a lot of time over the last few days, but will be taking a look at the docs now and update the PR soon.

glutanimate:
ok, updated to use ftl!

dae:
Thanks!

PR #455, opened 2020-02-22
Hook when class is initialized

Arthur-Milchior:
It often occurs that add-on must redefine the `__init__` method in order to do one last action when class is loaded. This PR create a hook `class_name_did_init` which is called when a method init is called on the class class_name.

I decided to create a simple class which redefine `__init__`, in order to avoid code repetition. Actually, it allows to create a new line in each `__init__` AND adding one hook by class in pylib/tools/genhooks.py. The biggest downside is that any add-on dev' which looks for this hook won't directly see it in the class. So this need to be properly documented in the add-on creation manual

Creating a real meta class for HookOnInit would have been nice too. The unexpected problem here wath that each PyQt class have already their meta class. Which led to a meta class conflict when I added HookOnInit to any class which was a child of a QTClass. This is the reason why I simply used __init_subclass__, which lead to pylint saying W0231. I know that meta class can be merged, but I didn't want to venture in such complex python programming.

For example
```python
from anki import hooks
def init_card(card):
    print(card.id)
hooks.Card_did_init.append(init_card)
```
would print the id of each card created.

### DIFF
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -10,6 +10,7 @@ from typing import List, Optional
 import anki  # pylint: disable=unused-import
 from anki import hooks
 from anki.consts import *
+from anki.hooks import HookOnInit
 from anki.models import NoteType, Template
 from anki.notes import Note
 from anki.sound import AVTag
@@ -27,7 +28,7 @@ from anki.utils import intTime, joinFields, timestampID
 # - lrn queue: integer timestamp
 
 
-class Card:
+class Card(HookOnInit):
     _note: Optional[Note]
     timerStarted: Optional[float]
     lastIvl: int
@@ -36,6 +37,7 @@ class Card:
     def __init__(
         self, col: anki.collection._Collection, id: Optional[int] = None
     ) -> None:
+        # pylint: disable=W0231
         self.col = col
         self.timerStarted = None
         self._render_output: Optional[anki.template.TemplateRenderOutput] = None
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -22,6 +22,7 @@ from anki.consts import *
 from anki.db import DB
 from anki.decks import DeckManager
 from anki.errors import AnkiError
+from anki.hooks import HookOnInit
 from anki.lang import _, ngettext
 from anki.media import MediaManager
 from anki.models import ModelManager, NoteType, Template
@@ -62,7 +63,7 @@ defaultConf = {
 
 
 # this is initialized by storage.Collection
-class _Collection:
+class _Collection(HookOnInit):
     db: Optional[DB]
     sched: Union[V1Scheduler, V2Scheduler]
     crt: int
@@ -82,6 +83,7 @@ class _Collection:
         server: Optional["anki.storage.ServerData"] = None,
         log: bool = False,
     ) -> None:
+        # pylint: disable=W0231
         self.backend = backend
         self._debugLog = log
         self.db = db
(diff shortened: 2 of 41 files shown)


Arthur-Milchior:
I did have to change some imports in the hooks module; they were creating cyclic imports

glutanimate:
Hmm, I'm not a huge fan of having dynamically constructed hooks. To be fair, I did go with that approach in #439 initially, but in retrospect I'm glad we ended up with an explicit hook call in #258.

I just feel that dynamic hooks add more complexity for new developers, which I think even good docs can only be a partial solution to. They also make type annotations more difficult as you have to explicitly cast your types. And, this is more of a personal take, but it feels a bit janky to have all classes inherit from the same base class and have that pylint warning appear around the codebase everywhere (though I do understand your reasoning for going with that compromise).

More generally though, while I do see the need for giving add-on authors access to objects at their init stage in some cases, I think that the changes here are a bit too pre-emptive. Doing a quick grep through an archive of 2.1 add-ons, I only found ≈30 that monkey-patch an `__init__` method of some sort. Out of these, only one pertained to a non-aqt class, `aqt.find.Finder` – and in that particular case the patch could now be easily replaced with the new-ish `search_terms_prepared` hook.

Among `aqt`, the targeted classes were almost exclusively `Editor`, `Preferences`, `Browser`, `EditCurrent`, and `AddCards`. In most of these cases the patches served the purpose of introducing user-visible changes to the UI or its behavior in some way.

I think this points to another issue of having a generalized init hook: Depending on the type of class modified, it can make sense to call add-ons at an earlier stage in the initialization process. For instance, with a Qt widget you'd ideally have add-on code be executed before `widget.show()` (and necessarily before `widget.exec()`). A general init hook would not be able to account for special needs like that.

So I guess my general point is this: I think it would be best if we went with explicit hooks, rather than implicit ones, specifically targeted "high add-on traffic" areas, and tailored the hooks to the common needs of add-ons interfacing with those areas.

glutanimate:
BTW, just to not sound like a negative Nancy: I really appreciate your sparking off this discussion, as there definitely is a need for init hooks in some key areas!

Arthur-Milchior:
I could also add manually each hooks in the most changed __init__ methods. I must admit that the point executing hook before "show" makes a lot of sens.
I actually thought about allowing the class HookOnInit to call the hooks before "show" or "exec". But I guess that you would also object since it means hooks would not be seen.

I should note however that, as far as I'm concerned, the problem you're mentioning: hooks are not clearly shown, already exists with web pages. I never knew that there was a hook to catch js->py command before you wrote it a few days ago. And I think I can state that I know anki code correctly.
It will also be a problem which occurs with your recent change which allows to add code in webpage. Since we don't see those hooks in editor/reviewer/stats page

glutanimate:
Yes, webview hook discoverability is not great, but I think there still is a difference between hooks that are at least listed in the genhooks files / appear verbatim in the code, and hooks that are constructed completely dynamically.

Personally I think the manual way is the way forward, with the aforementioned aqt classes as the first candidates for init hooks. But let's wait for Damien to chime in.

Arthur-Milchior:
A comment in each call to webview and bridge could be nice to let people know there is a hook

dae:
I'm afraid I'm not a big fan of this change either Arthur - it touches a lot of places, and relies on magic mixins instead of being explicit about its behaviour.  The two hooks in webview.py may not be obvious to people who are already familiar with the codebase, as they were only recently added, but I have now added them to the documentation, which should help new users find them:

https://ankitects.github.io/addon-docs/#/hooks-and-filters?id=notable-hooks

dae:
(as a general rule, I think it would be best to add more hooks as needed, rather than pre-emptively trying to add them everywhere "just in case")

PR #458, opened 2020-02-24
Add deck options hooks

glutanimate:
Adds a number of hooks for modifying the deck options dialog, which paves the way for add-ons like Speed Focus Mode to move away from patching `DeckConf` for their own deck-specific settings.

**A few notes**:

- Personally, `deck_conf_did_setup_ui_form` is sufficient for all my current use cases, but I could see `deck_conf_will_show` coming in handy for things like:
    - adjusting the combo menu entries
    - tweaking signal/slot assignments and window geometry

    More generally, I think it would make sense to keep `deck_conf_will_show` given that we are likely to see more hooks of this type to cover the use cases we discussed in #455
- I went with `Any` as the type for `config` and `deck` because the corresponding methods in `anki.decks` don't specify more specific types at the moment. I wasn't sure whether I should tweak that because a lot of code paths depend on `anki.decks` and I assume `Any` is used there for a reason

**Demo add-on**:

```python
from typing import Any

from PyQt5.QtWidgets import QLineEdit

from aqt import gui_hooks
from aqt.deckconf import DeckConf


def on_deck_conf_did_setup_ui_form(deck_conf: DeckConf):
    form = deck_conf.form
    form.myLineEdit = QLineEdit(form.tab_5)
    form.verticalLayout_6.addWidget(form.myLineEdit)


def on_deck_conf_did_load_config(deck_conf: DeckConf, deck: Any, config: Any):
    my_line_edit: QLineEdit = deck_conf.form.myLineEdit
    my_line_edit.setText(config.get("myLineEdit", ""))


def on_deck_conf_will_save_config(deck_conf: DeckConf, deck: Any, config: Any):
    my_line_edit: QLineEdit = deck_conf.form.myLineEdit
    config["myLineEdit"] = my_line_edit.text()


gui_hooks.deck_conf_did_setup_ui_form.append(on_deck_conf_did_setup_ui_form)
gui_hooks.deck_conf_did_load_config.append(on_deck_conf_did_load_config)
gui_hooks.deck_conf_will_save_config.append(on_deck_conf_will_save_config)
```

### DIFF
--- a/qt/aqt/deckconf.py
+++ b/qt/aqt/deckconf.py
@@ -221,7 +221,7 @@ class DeckConf(QDialog):
         f.replayQuestion.setChecked(c.get("replayq", True))
         # description
         f.desc.setPlainText(self.deck["desc"])
-        gui_hooks.deck_conf_did_load_config(self, self.conf)
+        gui_hooks.deck_conf_did_load_config(self, self.deck, self.conf)
 
     def onRestore(self):
         self.mw.progress.start()
--- a/qt/aqt/deckconf.py
+++ b/qt/aqt/deckconf.py
@@ -29,6 +29,7 @@ class DeckConf(QDialog):
         self._origNewOrder = None
         self.form = aqt.forms.dconf.Ui_Dialog()
         self.form.setupUi(self)
+        gui_hooks.deck_conf_did_setup_ui_form(self)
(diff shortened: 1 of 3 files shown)


dae:
Looks good, thanks!

PR #461, opened 2020-02-24
Add new will_init_bottom and did_init_bottom hooks

jorgebucaran:
This is my first attempt at adding two new hooks that might prove useful (and I would benefit from). Please let me know if I need to fix or change anything, etc. 👋 

- ~~`will_init_bottom`~~ `toolbar_bottom_will_init` allows you to hide the bottom web (status bar) by simply returning `False`.

- ~~`did_init_bottom`~~ `toolbar_bottom_did_init` is called after bottom web is created, allowing you to easily extend it, e.g. add more card stats, show how many times you have failed this card, and so on.

### Background

I'd like to hide the toolbar bottom while reviewing. 

I hope that these hooks allow me to do this as easily as:

```python
from aqt import mw
from aqt import gui_hooks as hooks

hooks.toolbar_bottom_will_init.append(lambda _: False)
```

Compare to my current solution monkey-patching `_initWeb`:

```python
Reviewer._initWeb = (
    lambda init:
    lambda self: init(self) or mw.reviewer.bottom.web.eval(
        "outer.style.display='none'"))(Reviewer._initWeb)
```

### DIFF
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -156,7 +156,7 @@ class Reviewer:
             context=self,
         )
 
-        if gui_hooks.will_init_bottom():
+        if not gui_hooks.toolbar_bottom_will_init(True):
             return
 
         # show answer / ease buttons
@@ -168,7 +168,7 @@ class Reviewer:
             context=ReviewerBottomBar(self),
         )
 
-        gui_hooks.did_init_bottom()
+        gui_hooks.toolbar_bottom_did_init()
 
     # Showing the question
     ##########################################################################
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -155,6 +155,10 @@ class Reviewer:
             ],
             context=self,
         )
+
+        if gui_hooks.will_init_bottom():
+            return
+
         # show answer / ease buttons
         self.bottom.web.show()
         self.bottom.web.stdHtml(
@@ -163,6 +167,8 @@ class Reviewer:
             js=["jquery.js", "reviewer-bottom.js"],
             context=ReviewerBottomBar(self),
         )
+
+        gui_hooks.did_init_bottom()
 
     # Showing the question
     ##########################################################################
--- a/qt/tools/genhooks_gui.py
+++ b/qt/tools/genhooks_gui.py
@@ -392,15 +392,15 @@ hooks = [
         args=["player: aqt.sound.Player", "tag: anki.sound.AVTag"],
     ),
     Hook(name="av_player_did_end_playing", args=["player: aqt.sound.Player"]),
-    # Web
+    # Toolbar
     ###################
     Hook(
-        name="will_init_bottom", 
+        name="toolbar_bottom_will_init", 
         return_type="bool",
-        doc="Return True to not display the bottom web.",),
+        doc="Return True to not display the toolbar while reviewing.",),
     Hook(
-        name="did_init_bottom", 
-        doc="Used to extend the bottom web. E.g. add extra stats to the status bar",),
+        name="toolbar_bottom_did_init", 
+        doc="Used to extend the toolbar while reviewing E.g. add extra stats to the toolbar",),
     # Other
     ###################
     Hook(
--- a/qt/tools/genhooks_gui.py
+++ b/qt/tools/genhooks_gui.py
@@ -392,6 +392,15 @@ hooks = [
         args=["player: aqt.sound.Player", "tag: anki.sound.AVTag"],
     ),
     Hook(name="av_player_did_end_playing", args=["player: aqt.sound.Player"]),
+    # Web
+    ###################
+    Hook(
+        name="will_init_bottom", 
+        return_type="bool",
+        doc="Return True to not display the bottom web.",),
+    Hook(
+        name="did_init_bottom", 
+        doc="Used to extend the bottom web. E.g. add extra stats to the status bar",),
     # Other
     ###################
     Hook(
--- a/qt/tools/genhooks_gui.py
+++ b/qt/tools/genhooks_gui.py
@@ -397,7 +397,7 @@ hooks = [
     Hook(
         name="toolbar_bottom_will_init", 
         return_type="bool",
-        doc="Return True to not display the toolbar while reviewing.",),
+        doc="Return True to hide the toolbar while reviewing.",),
     Hook(
         name="toolbar_bottom_did_init", 
         doc="Used to extend the toolbar while reviewing E.g. add extra stats to the toolbar",),
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -26,6 +26,7 @@ Arthur Milchior <arthur@milchior.fr>
 Yngve Hoiseth <yngve@hoiseth.net>
 Ijgnd
 Yoonchae Lee
+Jorge Bucaran <https://jorgebucaran.com>
 
 ********************
 


qt/tools/genhooks_gui.py
```diff
@@ -392,6 +392,15 @@ def on_top_toolbar_did_init_links(links, toolbar):
         args=["player: aqt.sound.Player", "tag: anki.sound.AVTag"],
     ),
     Hook(name="av_player_did_end_playing", args=["player: aqt.sound.Player"]),
+    # Web
```
dae: Please pop these in a section called "toolbar" instead

qt/tools/genhooks_gui.py
```diff
@@ -392,6 +392,15 @@ def on_top_toolbar_did_init_links(links, toolbar):
...
+    ###################
+    Hook(
+        name="will_init_bottom",
```
dae: These would be clearer with the module at the start, eg

toolbar_bottom_will_init
toolbar_bottom_did_init

qt/aqt/reviewer.py
```diff
@@ -155,6 +155,10 @@ def _initWeb(self):
             ],
             context=self,
         )
+
+        if gui_hooks.will_init_bottom():
```
dae: Filters need to pass in the first argument so they can be chained, so the hook needs to take an argument like 'proceed', and be called with

```
if not gui_hooks.toolbar_bottom_will_init(True):
  return
```

jorgebucaran:
@dae Thanks for your review. Please let me know what else do you need! 👋

Arthur-Milchior:
@dae: do you mean related to toolbar ? Or related to initialization.
In the first case, no; I actually never gave a serious thought to tool bar.
In other case, yes, there are other initialization I want to change. that's why I did a general PR recently. I'll make a PR to add the exact hooks I want later, but there are certainly things I want to change while they are created.

dae:
jorge: the tests are failing - did you run Anki and check they work?

Arthur: I was talking specifically about the toolbar

glutanimate:
Hmm, a few suggestions:

- I think we might be able to make do without `toolbar_bottom_did_init`. At least looking through existing monkey patches to `Reviewer._initWeb` that target the bottom toolbar, I couldn't find any that wouldn't be covered by `gui_hooks.webview_will_set_content`. All existing add-ons pretty much  just modify the web content in some way. If, in the future, additional setup steps followed after `self.bottom.web.stdHtml()`, then going with an additional hook would make sense of course. ¹
- I would recommend renaming the hooks so that it's easy to tell that they originate in the "reviewer version" of the bottom toolbar. I feel like that would be important since the bottom toolbar widget is reused across all `mw` states.

   Maybe something like: `reviewer_bottom_bar_will_show` and (if we keep it) `reviewer_bottom_bar_did_show`. (I would avoid "`init`" since strictly speaking the toolbar is initialized much earlier, i.e. in `Reviewer.__init__` for the `BottomBar` instance, and – going back even further – in `AnkiQt.setupMainWindow` for the actual web view of the toolbar). 
- Also, I think it would be important to pass the caller on to the hooks in both cases, so that add-on code can just dependency-inject it rather than relying on accessing global state via `mw.reviewer.*`. For consistency's sake, I would recommend passing on the same `ReviewerBottomBar` instance as with `gui_hooks.webview_will_set_content`. This would also allow us to remain fairly consistent if the bottom toolbar undergoes refactoring at some point (my refactoring senses are tingling a bit with all the toolbar classes we now have, in no small part thanks to `webview_will_set_content`... )

---

1: Strictly speaking you could also use `gui_hooks.webview_will_set_content` to hide the toolbar, e.g.:

```python
from typing import Any, Union

from aqt import gui_hooks
from aqt.webview import WebContent
from aqt.reviewer import ReviewerBottomBar

def on_webview_will_set_content(
    content: WebContent, context: Union[Any, ReviewerBottomBar]
):
    if not isinstance(context, ReviewerBottomBar):
        return

    content.head += "<style>#outer{display: none}</style>"


gui_hooks.webview_will_set_content.append(on_webview_will_set_content)
```

...but it's certainly a lot more boilerplate code and less stable than having a clear API for hiding the toolbar.

glutanimate:
Actually, giving this a bit more thought, I'm not sure how I feel about preventing the bottom toolbar from running its web content. There's at least one add-on with [my "Speed Focus Mode"](https://github.com/glutanimate/speed-focus-mode/blob/master/src/speed_focus_mode/reviewer.py#L77) that uses the bottom web view to execute JS that interacts with the reviewer. Hiding the bottom bar by prematurely terminating `Reviewer._initHtml` rather than using CSS as shown above would prevent Speed Focus Mode from working. Now, there's a good point to be made that code like this should be run in `reviewer.web`, but the issue here is two-fold:

1. The reviewer [recycles its web view every 100 cards](https://github.com/Glutanimate/anki/blob/master/qt/aqt/reviewer.py#L103) which breaks JS code that needs to keep track of persistent state across longer reviewing sessions (without relying on the JS <-> Py bridge or something else). `reviewer.bottom.web` offers a solution to that.
2. Speed Focus Mode needs to synchronize UI elements in the bottom bar (e.g. a countdown timer and "stop" button) with effects in the reviewer, so it makes sense to have the corresponding code in the bottom bar. It would be possible to rewrite the add-on to handle more of this on the Python side of things, but it would be a chore compared to the simplicity of JS timers.

Basically, my point is this: For add-on authors who would like to have JS run reliably with persistent state across reviews, the bottom web element offers a fairly good solution. Introducing the `toolbar_bottom_will_init` / `reviewer_bottom_bar_will_show` hook would uproot that. Essentially it would make `mw.bottomWeb` the only web view in Anki that add-on authors could not reliably depend on being executed (monkey-patches to other web views by third-party add-ons aside).

jorgebucaran:
> @glutanimate Essentially, it would make mw.bottomWeb the only web view in Anki that add-on authors could not reliably depend on being executed (monkey-patches to other web views by third-party add-ons aside).

Let me first clarify that an Anki option to hide the UI while reviewing would be the ideal solution for me.

A second-best would be the ability to hide both the top and bottom toolbar via an add-on (without monkey-patching or injecting CSS). 

Before I opened this PR, I just wanted to run this somewhere:

```js
mw.toolbar.web.hide()
mw.reviewer.bottom.web.hide()
```

...but there wasn't a hook that allowed me to do all that in one place. 

I'm currently using `reviewer_did_show_question` and `reviewer_will_end` hooks  to hide/show the top `toolbar`. And to hide the Reviewer's bottom toolbar, I'm monkey-patching Anki, but that's far from ideal, hence this PR.

> Speed Focus Mode

Conflicts among add-ons can't be helped sometimes. It would be a shame to miss the opportunity to add at least one of these hooks because it might conflict with another add-on.

glutanimate:
> Before I opened this PR, I just wanted to run this somewhere:
> 
> ...but there wasn't a hook that allowed me to do all that in one place.

Hmm, how about this:

```python
from aqt import gui_hooks
from aqt import mw

def on_state_change(new_state: str, old_state: str):
    if not mw:
        return
    if new_state == "review":
        mw.bottomWeb.hide()
        mw.toolbar.web.hide()
    elif new_state != "resetRequired":
        mw.bottomWeb.show()
        mw.toolbar.web.show()

gui_hooks.state_did_change.append(on_state_change)
```

> Conflicts among add-ons can't be helped sometimes. It would be a shame to miss the opportunity to add at least one of these hooks because it might conflict with another add-on.

For sure, but if we can sidestep a conflict by switching to a different implementation, then we should do that. Especially in case of an add-on like Speed Focus Mode that has a fairly large userbase that would be potentially affected.

jorgebucaran:
Oh, that works perfectly! Thank you, @glutanimate. 💯🎉

I created this PR because I wanted to hide the UI during review without monkey-patching Anki and it turns out I can do it without these hooks. I don't have a need for them now. Looks like @Arthur-Milchior is not very interested in these either "I actually never gave a serious thought to tool bar", so I'm going to close here. 

<h2></h2>

@dae Having this ability out of the box would be fantastic, though. Are you open to a new PR to add an option to hide the UI during review? 

e.g.

EN
EN-2
JP

dae:
@glutanimate thanks for looking at this more carefully than I did!

@jorgebucaran maybe in the future, but not right now I'm afraid

PR #462, opened 2020-02-25
Add a hook to change empty cards to delete

Arthur-Milchior:
A hook which allow to use the general "onEmpty" command, but limit the set of cards actually deleted.
For my add-on https://ankiweb.net/shared/info/1402327111

### DIFF
--- a/qt/aqt/gui_hooks.py
+++ b/qt/aqt/gui_hooks.py
@@ -862,6 +862,51 @@ class _EditorWillUseFontForFieldFilter:
 editor_will_use_font_for_field = _EditorWillUseFontForFieldFilter()
 
 
+class _EmptyCardsWillBeDeletedFilter:
+    """Allow to change the list of cards to delete.
+
+        For example, an add-on creating a method to delete only empty
+        new cards would be done as follow:
+```
+from anki.consts import CARD_TYPE_NEW
+from anki.utils import ids2str
+from aqt import mw
+from aqt import gui_hooks
+
+def filter(cids, col):
+    return col.db.list(
+            f"select id from cards where (type={CARD_TYPE_NEW} and (id in {ids2str(cids)))")
+
+def emptyNewCard():
+    gui_hooks.append(filter)
+    mw.onEmptyCards()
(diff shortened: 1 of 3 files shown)


dae:
How about empty_cards_will_be_deleted instead? It could go either way, but I'd also lean towards adding this as a GUI hook in onEmptyCards(), as the pylib function is not called anywhere else

dae:
(Strange that the contributor check is failing, as it should be case insensitive - did you make this change relative to an older branch?)

Arthur-Milchior:
This branch is up to date. It contains commit you made 3 hours ago.
I did rebase it, because it used to be on top of commits of two days ago. Where anyway I was already in the list of contributors.

Arthur-Milchior:
As a side note, every amke operation fails when there is no internet connectivity. It seems some part of python is looking for update and don't accept to simply fail to reach the web. I did this PR in a plane, and realized I could not fix the code anymore nor test it. Not a real problem, given that the code is super simple, but still anoying.

Arthur-Milchior:
I prefer the hook to be present in the /anki part. Because my add-on call the /anki method and not the /qt method. That could be changed of course. But as far as I'm concerned, it's the back-end part which I want to change

Arthur-Milchior:
I have no opinion concerning the name.

dae:
I had a think about it Arthur, and I think the GUI may be a better place place for this hook - emptyCids() is simply gathering a list of empty cards; it is the GUI side the does the actual deletion, so I think the hook is better placed there. Would you like to make that change, or would you like me to?

Arthur-Milchior:
I'll do it

dae:
Thanks Arthur!

dae:
Re the build-fails-when-net-offline issue, is that still a problem with the latest HEAD? If so, what exactly is failing/hanging?

Arthur-Milchior:
As far as I can tell, the problem is that my domain name was registered with "Milchior" and now it's "milchior". I don't really know why, I don't remember changing this value in my git setting. But I guess that having a non-sort-sensitive comparison would be nice here

PR #464, opened 2020-02-26
Improve mypy config

alanhdu:
- Make mypy stricter
- Warn for unnecessary workarounds

### DIFF
--- a/pylib/mypy.ini
+++ b/pylib/mypy.ini
@@ -4,6 +4,10 @@ pretty = false
 no_strict_optional = true
 show_error_codes = true
 check_untyped_defs = true
+disallow_untyped_decorators = True
+warn_redundant_casts = True
+warn_unused_configs = True
+warn_unused_ignores = True
 
 [mypy-win32file]
 ignore_missing_imports = True
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -26,6 +26,7 @@ Arthur Milchior <arthur@milchior.fr>
 Yngve Hoiseth <yngve@hoiseth.net>
 Ijgnd
 Yoonchae Lee
+Alan Du <alanhdu@gmail.com>
 
 ********************
 
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -183,7 +183,7 @@ class _Collection:
 select crt, mod, scm, dty, usn, ls,
 conf, models, decks, dconf, tags from col"""
         )
-        self.conf = json.loads(conf)  # type: ignore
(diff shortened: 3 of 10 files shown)


alanhdu:
The `make check` failure should be fixed by https://github.com/ankitects/anki-desktop-i18n

.github/scripts/contrib.sh
```diff
@@ -9,7 +9,7 @@ git log --pretty=format:' - %ae' CONTRIBUTORS |sort |uniq |sort -f | sed "s/@/$a
 
 headAuthor=$(git log -1 --pretty=format:'%ae')
 authorAt=$(echo "$headAuthor" | sed "s/@/$antispam/")
-if git log --pretty=format:'%ae' CONTRIBUTORS | grep -qi "$headAuthor"; then
+if git log --pretty=format:'%ae' CONTRIBUTORS | grep -i "$headAuthor" > /dev/null; then
```
dae: why did you change this?

pylib/mypy.ini
```diff
@@ -3,7 +3,12 @@ python_version = 3.7
 pretty = false
 no_strict_optional = true
 show_error_codes = true
+
```
dae: Why the space?

.github/scripts/contrib.sh
alanhdu: The CONTRIBUTOR check was failing otherwise:

https://github.com/ankitects/anki/runs/468917834

(I'll admit I don't really understand why -- I feel like it should work. Something about `pipefail` made the old version fail).

pylib/mypy.ini
alanhdu: I thought it'd be easier to group the settings into "display settings" and "typecheck setings" (although I see now that I missed `no_strict_optional`). Is this something you'd want, or would you like me to revert the new line here?

.github/scripts/contrib.sh
dae: Ah, that makes sense thanks - grep will be closing the input as soon as it finds a line, which causes git log to lose its pipe.

pylib/mypy.ini
dae: python_version is not a display setting either - probably simplest to remove the gap.

dae:
The build is currently failing due to these changes. Are they working in your local environment? Maybe the hints have been improved in the latest PyQt release?

alanhdu:
@dae: I think I just messed up and didn't push the right things. I rebased on top of master to incorporate your change and removed the extra line, so this should be ready for another look!

dae:
Great, thanks!

PR #465, opened 2020-02-27
Upgrade some rust deps

alanhdu:
This also reduces slightly reduces compile-time (especially the `prost-build` update).


dae:
Thanks!

PR #466, opened 2020-02-27
Fixed qt/i18n/po submodule invalid configuration

evandrocoan:
The submodule qt/i18n/po was had missing its declaration inside `.gitmodules` file, there fore, it would never work without it.


dae:
Anki's codebase does not use git modules. The i18n repos are pulled in on the first build, and then can be updated with 'make pull-i18n'.

evandrocoan:
The qt/i18n/po as added as a submodule today on this commit: 

image

1. https://github.com/ankitects/anki/commit/a63c9e5a9a1fce5c404cd8b4851d1cba4f08728e#diff-1858871bafd3e61810b7aad41a68d077

If it is not a submodule, the directory qt/i18n/po should be on the .gitignore

evandrocoan:
It is removed now and qt/i18n/po is on .gitignore

dae:
Thanks Evandro, I overlooked that change in Alan's patch. I'm not sure it makes sense to add it to .gitignore if the makefile is no longer generating that folder however, as it should be a one time cleanup unless I've missed something in the makefile.

dae:
@alanhdu heads up

evandrocoan:
You are right. It was removed recently and I did not had noticed it.

I removed that entry from the .gitignore, but you still need to merge this to unregister the submodule @alanhdu had regisred on the other commit.

evandrocoan:
I just saw you removed it on master. Thanks!

PR #467, opened 2020-02-27
Add some type annotations [1/n]

alanhdu:
This is the first of (hopefully) a series of PRs, where I use [MonkeyType](https://github.com/Instagram/MonkeyType) to automatically generate some type-hints for Anki (I use the test suite for `pylib` types).

This PR represents the relatively straightforward types so it can be reviewed easier: I'll follow-up with a PR that has more involved typing separately.

### DIFF
--- a/pylib/anki/db.py
+++ b/pylib/anki/db.py
@@ -5,7 +5,7 @@ import os
 import time
 from sqlite3 import Cursor
 from sqlite3 import dbapi2 as sqlite
-from typing import Any, List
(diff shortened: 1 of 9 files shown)


dae:
Thanks!

I think some of the more complicated types may be a case of diminishing returns - if the signatures start to get unwieldy, it may be worth just sticking with a summarised form.

dae:
(until the code can be updated to use beta data structures)

PR #468, opened 2020-02-27
fix comment on setWebExports usage

BlueGreenMagick:
The comment regarding usage on setWebExports is very misleading as it would not work on an OS that does not use `/` but `\` as the directory seperator.

### DIFF
--- a/qt/aqt/mediasrv.py
+++ b/qt/aqt/mediasrv.py
@@ -161,11 +161,10 @@ class RequestHandler(http.server.SimpleHTTPRequestHandler):
             if not pattern:
                 return path
 
-            if not re.fullmatch(pattern, subPath):
-                return path
-
-            newPath = os.path.join(addMgr.addonsFolder(), addonPath)
-            return newPath
+            subPath2 = subPath.replace(os.sep, "/")
+            if re.fullmatch(pattern, subPath) or re.fullmatch(pattern, subPath2):
+                newPath = os.path.join(addMgr.addonsFolder(), addonPath)
+                return newPath
 
         return path
 
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -149,6 +149,8 @@ class WebContent:
                       f"/_addons/{addon_package}/web/my-addon.css")
                   web_content.js.append(
                       f"/_addons/{addon_package}/web/my-addon.js")
+          
+          Note that '/' will also match the os specific path separator.
     """
 
     body: str = ""


BlueGreenMagick:
I did not realize that there shouldn't be such backslash in the docstring. What would be the best way forward?

1. Make the docstring a raw string
2. Change the given example so the resources are in addon root directory
3. In setWebExports, detect `/` and replace it with `(\/|\\)` (or replace it with os.sep)
4. Leave it as it is

glutanimate:
3) sounds good, but I'm not sure if there isn't a scenario where it might end up breaking existing code (e.g. if that code already is using an extended sep match pattern).

Perhaps for the time being we could otherwise just use an example based on `os.path.sep`?

Another solution might be to have the mediasrv code check web export patterns both against the Windows and Posix path representations, and allow access if either matches.

But yeah, sorry for this mess. In retrospect it would have been better to account for different path separators from the get-go.

dae:
I think it would be best if add-on authors didn't have to worry about Windows separator. How about the following:

- converting any \ in the pattern returned by getWebExports to a /
- convert any \ in the path in _redirectWebExports to a / prior to comparing it with the pattern

BlueGreenMagick:
Would it be necessary to convert \ to / in getWebExports? I think if an addon author changes the example to use \, he should be trusted to know what he is doing. And a quick search on github reveals that no add-on supports only \ as the directory seperator.

dae:
There are a bunch of existing add-ons that call setWebExports() using os.sep, eg `mw.addonManager.setWebExports(__name__, f"web{sep}.*")`.  Since subpath has been canonicalized to /, the same must be done to the web exports to ensure the comparison doesn't fail on Windows

BlueGreenMagick:
Oh then would the following approach be alright? Matching once with os.sep and once with / ?
```
subPath2 = subPath.replace(os.sep, "/")
if re.fullmatch(pattern, subPath) or re.fullmatch(pattern, subPath2):
        newPath = os.path.join(addMgr.addonsFolder(), addonPath)
        return newPath
```

glutanimate:
@dae Yeah, just checking against both variations in mediasrv would be my preferred approach also, to be honest. It just feels like replacing backslashes in RegEx patterns could lead to all sorts of trouble.

Also, to encourage authors to move away from using os.sep or Windows backslashes, we could add a small comment to `setWebExports` mentioning that `/` is our canonical path separator, and that backslashes are deprecated.

dae:
You'd not be replacing a '\', but only '\\' which is a literal backslash. But perhaps I'm overlooking something? If so, doing two checks would be another way to solve the problem. And yeah, it would make sense to update the docs as well.

glutanimate:
> You'd not be replacing a '', but only '\' which is a literal backslash. But perhaps I'm overlooking something?

I'm not 100% sure, either, tbh. I'm a bit rusty on those things. But this is what I experimented with earlier:

```python
import re

import ntpath
import posixpath

pattern = r"web{sep}\d.(css|js)"

posix_pattern = pattern.format(sep=posixpath.sep)
windows_pattern = pattern.format(sep=ntpath.sep)
windows_pattern = windows_pattern.replace(r"\\", "/")


subpaths = (
    "web/2.js",
    "web/1.css"
)

for pattern in (posix_pattern, windows_pattern):
    for subpath in subpaths:
        match = re.match(pattern, subpath)
        if match:
            print(f"Match '{match[0]}' with pattern '{pattern}'")
        else:
            print(f"No match for '{subpath}' with pattern '{pattern}'")

```

I tried multiple combinations of using raw strings, literal backslashes, etc., but I wasn't able to find a way to replace the path-separator backslash without losing the `\d`. I'm sure I'm missing something, though.

BlueGreenMagick:
I don't think the example `fr"{sep}"` is valid. For backward slashes, shouldn't that be `fr"{sep}{sep}"`? So I'm not sure how one would successfully use os.sep in the pattern.

Try the below test:
```
import re
import ntpath

ps= (
    r"web{sep}\d.(css|js)",
    r"web{sep}{sep}\d.(css|js)"
)
subpath = "web\\2.js"

for p in ps:
    pattern = p.format(sep = ntpath.sep)
    match = re.match(pattern, subpath)
    if match:
        print(f"Match '{match[0]}' with pattern '{pattern}'")
    else:
        print(f"No match for '{subpath}' with pattern '{pattern}'")

```

glutanimate:
Right, good point. I was playing around in the REPL and I forgot about how `__repr()__` and `print()` handle backslashes differently. And it makes sense, because looking at the actual add-on code it reads:

```python
sep = re.escape(os.path.sep)
```

(I completely forgot that I'd gone with that solution... Don't you just love it when 'past you' beats 'future you' to the punch? :man_facepalming: )

In any case, either approach – modifying the patterns or checking against both path variants – would be fine by me.

BlueGreenMagick:
I think replacing with `replace("\\\\", "/")` should work, but there might be some edge cases. For example, macOS seems to allow backslashes in file name.

dae:
I doubt they'd come up in practice, but this'll do - thanks for working on it!

PR #469, opened 2020-02-28
Monkeytype the rest of pylib [2/n]

alanhdu:
Follow up to #467, adding type annotations for more parts of `pylib`.

### DIFF
--- a/pylib/anki/find.py
+++ b/pylib/anki/find.py
@@ -1,10 +1,11 @@
 # Copyright: Ankitects Pty Ltd and contributors
 # License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+from __future__ import annotations
 
 import re
 import sre_constants
 import unicodedata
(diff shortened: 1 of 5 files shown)


pylib/anki/find.py
```diff
@@ -18,12 +19,15 @@
...
+if TYPE_CHECKING:
+    from anki.collection import _Collection
+
 # Find
 ##########################################################################
 
 
 class Finder:
-    def __init__(self, col) -> None:
+    def __init__(self, col: Optional[_Collection]) -> None:
```
dae: I don't think it's optional?

pylib/anki/find.py
```diff
@@ -481,7 +484,7 @@ def _findField(self, field, val) -> Optional[str]:
 
     def _findDupes(self, args) -> Optional[str]:
         # caller must call stripHTMLMedia on passed val
-        (val, args) = args
+        (val, __) = args
```
dae: perhaps these should be _ instead of __?

PR #470, opened 2020-02-28
Hook to decide whether a note should be added.

Arthur-Milchior:
This allow to purpose:
* removing the rule forcing the first field to be non empty
* allowing to add extra rules, such as: adding a tag

I should also note that filters show limits here. Because I'm quite limited by the fact that the input and the output should have the same type, that the output is given as input.
I would really love to have a type of hooks where a hook would  aggregate the different returned value. The simplest version would be: creating a list of returned values by the different function added to the hook. A more general version would be to have an optional parameter "call" when you generate hooks, which would replace the "call" method of the developped hook

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -167,8 +167,12 @@ class AddCards(QDialog):
     def addNote(self, note) -> Optional[Note]:
         note.model()["did"] = self.deckChooser.selectedId()
         ret = note.dupeOrEmpty()
+        rejection = None
         if ret == 1:
-            showWarning(_("The first field is empty."), help="AddItems#AddError")
+            rejection = _("The first field is empty.")
+        rejection = gui_hooks.add_card_accepts(rejection, note)
+        if rejection is not None:
+            showWarning(rejection, help="AddItems#AddError")
             return None
         if "{{cloze:" in note.model()["tmpls"][0]["qfmt"]:
             if not self.mw.col.models._availClozeOrds(
--- a/qt/aqt/gui_hooks.py
+++ b/qt/aqt/gui_hooks.py
@@ -23,6 +23,45 @@ from aqt.qt import QDialog, QMenu
 # @@AUTOGEN@@
 
 
+class _AddCardAcceptsFilter:
+    """Decides whether the note should be added to the collection or
+        not. It is assumed to come from the addCards window.
+
+        reason_to_already_reject is the first reason to reject that
+        was found, or None. If your filter wants to reject, it should
+        replace return the reason to reject. Otherwise return the
+        input."""
+
+    _hooks: List[Callable[[Optional[str], "anki.notes.Note"], Optional[str]]] = []
+
+    def append(
+        self, cb: Callable[[Optional[str], "anki.notes.Note"], Optional[str]]
+    ) -> None:
+        """(reason_to_already_reject: Optional[str], note: anki.notes.Note)"""
+        self._hooks.append(cb)
+
+    def remove(
+        self, cb: Callable[[Optional[str], "anki.notes.Note"], Optional[str]]
+    ) -> None:
+        if cb in self._hooks:
+            self._hooks.remove(cb)
+
+    def __call__(
+        self, reason_to_already_reject: Optional[str], note: anki.notes.Note
+    ) -> Optional[str]:
+        for filter in self._hooks:
+            try:
+                reason_to_already_reject = filter(reason_to_already_reject, note)
+            except:
+                # if the hook fails, remove it
+                self._hooks.remove(filter)
+                raise
+        return reason_to_already_reject
+
+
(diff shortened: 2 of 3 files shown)


dae:
I don't see how you're limited. A system like you describe would just accumulate the results into a list, and that is less flexible than passing a list around that each hook can mutate as it likes. And since you're using a list in this case, the hooks don't need to return it at all, they can just modify the list and return nothing: https://ankitects.github.io/addon-docs/#/hooks-and-filters

I'm not even sure you need to accumulate the problems in this case - if a single problem is found you can notify the user, and if any other problems exist they can fix them when they try to add the card again.  Then it just becomes a filter with a first argument of 'problem: Optional[str]'.

Arthur-Milchior:
What I don't really like about being able to change the list is that the result will depends on the order in which hooks are loaded.
Either all add-on simply add things in the list; in which case "aggregating all returned values" would works as well.
Or some add-on removes things from the list. In this second case, the content of the list become important. The add-on could only remove things which are added before it is executed, and not things which are added later. Which means that if you want to create an add-on which state "accept unconditionally" for example, you could do it only if it executed last. So it may works in your development area, if it's the only add-on you have, or if the folder starts with a letter while all other add-ons are in folders with number. And when you upload the add-on, it starts to fails for users having other add-ons. 
I believe that it's far safer to be clear that the hook allows to reject and does not allow to accept if anything did reject.

Furthermore, by having the condition "reject if first field is empty" as a hook element, it can be removed by my add-on, and it's clear that my add-on only removes this rules, and does not remove any rule created by another add-on.

dae:
If you know you're order-dependent because you're eg removing items, you can add yourself to the hook after all add-ons have loaded.

Arthur-Milchior:
I thus edited the hook, so that it does not return anything. I edited the method reject_empty_first_field accordingly

dae:
It still seems more complicated than it needs to be, and the less you change the existing code, the easier a hook is for me to accept. Why don't you just use a single item instead of accumulating a list? eg 

```
problem = None
if dupeOrEmpty: problem = _("...")
problem = gui_hooks.addcards_will_report_problem(problem, ..)
if problem is not None:
   <normal code>
```

Arthur-Milchior:
Usually, putting all string in a list and then joining them is more efficient than adding multiple time strings to strings. But that's only a matter of habit here, since I don't see a case where the difference will be noticeable even if there were 10 add-ons using this hook (which would probably not be the case). So it was mostly the intuition of using what I've read to be best practice.

It's also because this way, the add-on can simply append a string without having the need to check anything.
For example if default value is None, it needs to check whether it's None to decide whether to create a new string or to append its message to an existing one.
And even if None were replaced by the empty string, it would still need to check for a previous message to decide whether to add a separator (new line) or not. And it would not be clear how to separate messages. E.g. should the new line be added before or after the message ?

dae:
I was suggesting that the filter simply returns error/no error through the chain - a single error being sufficient to terminate the process. An add-on that wants to override Anki's standard empty field can place itself at the top of the chain and change the initial problem to None, and an add-on that wants to add a new condition can add itself later in the chain and return the problem there. Since only a single problem can exist, there is no need to think about joining the strings together.

Arthur-Milchior:
Filter no takes and return optional string

dae:
- Is there any reason why this patch is still introducing a separate reject_empty_first_field instead of just passing the problem string into the filter?
- please change the name to addcards_will_add_note(problem: Optional[str], ...)

dae:
sorry, I mean add_cards not addcards

Arthur-Milchior:
It seems to make more sens to me, to consider that this is an element of the hook as any other, instead of being treated separately. It's mostly a matter of taste.

I also find it more elegant that my add-on to allow empty first field can remove this from the list of hooked function, instead of having to search for this string specifically as the parameter.

dae:
Your proposed solution is more consistent. If this was a common code path that a lot of add-ons relied on, then it would probably be worth the change. But in this case, it's just a single add-on that would benefit, right? In such cases, please try to keep the changes to a minimum, as it takes time to review the changes, and the more that is changed, the greater the risk of something breaking accidentally.

Arthur-Milchior:
Clearly I don't see any reason to expect other add-on to want to do what my add-on does. I can't even imagine how they could do it differently.
I'll be sure that my add-on is executed first, in order to remove your error message. And then let other add-on potentially give their error message. 
even if I must state that 
`_AddCardAcceptsFilter._hooks.insert(0, my_function)` seems less elegant than 
`_AddCardAcceptsFilter._hooks.append(my_function)`.

Arthur-Milchior:
Normally this PR can now be merged. I think I did all changed you asked for

dae:
Thanks Arthur. I will rename this hook after merge, but otherwise looks good.

PR #472, opened 2020-02-28
Hook init clayout

Arthur-Milchior:
Those are two kind of related change. The first hook is only here to correct something which is unclear in the hook list.

I intend to use this hook for https://ankiweb.net/shared/info/915063177 . Even if it will still lead to some monkey-patching. I don't see any way to create hook for all changes I make. 

If you are interested, I think it may be nice to merge this add-on in anki. Being able to preview easily how cards in different cloze may be nice. Especially since you can test for a specific card number in the template.

### DIFF
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -65,6 +65,7 @@ class CardLayout(QDialog):
         v1.addLayout(self.buttons)
         v1.setContentsMargins(12, 12, 12, 12)
         self.setLayout(v1)
+        gui_hooks.card_layout_will_show(self)
         self.redraw()
         restoreGeom(self, "CardLayout")
         self.setWindowModality(Qt.ApplicationModal)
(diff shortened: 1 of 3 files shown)


dae:
"card_layout_will_show" would be more consistent

This is something that could be included in Anki, though I think the UI could be improved a bit. Instead of a spinner, perhaps a better approach would be to have the card type drop down be filled out with all the available cloze numbers, so the different cards are selected the same way as in a regular note type.

Arthur-Milchior:
I certainly do agree that the UI is not beautiful. For my personal use I don't really pay attention, and I hope add-on users are fine with it. 
I agree that the list makes a lot of sens. Especially since a card number N won't be generated when you don't have a cloze N in one of the cloze field.

Arthur-Milchior:
Name changed

dae:
Cheers!

PR #473, opened 2020-02-29
Fixed anki website docs link on pylib/anki/consts.py

evandrocoan:
The URL on `http` is being redirected to this new one on `https`.


dae:
cheers

PR #475, opened 2020-03-01
hook addon_current_config_is_invalid

Arthur-Milchior:
The main reason to use that is that, when I create add-on, I sometime make typo when I create/change the configuration. And then I want to be able to read the json error message in order to debug it.

By the way, would you accept a PR which check whether an edited config satisfies a json-schema before accepting it ?
I think json-schema for config would be useful, first to help user not to write anything wrong. And in a second time because it would help create automatic preference GUI, where the name of the fields are prefilled, and it's clear whether it's a boolean (check box), an integer (numeric box) or a float, or a string (text field), or another dict (which would be created by openining a new config window), a tuple or a list (probably created as in clayout, with a menu to choose which element of the list to check). However, this feature would only make sens if there is already some add-on with schemas

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -26,6 +26,7 @@ import aqt
 import aqt.forms
 from anki.httpclient import HttpClient
 from anki.lang import _, ngettext
+from aqt import gui_hooks
 from aqt.qt import *
 from aqt.utils import (
     TR,
@@ -248,11 +249,13 @@ class AddonManager:
     # in new code, use self.addon_meta() instead
     def addonMeta(self, dir: str) -> Dict[str, Any]:
         path = self._addonMetaPath(dir)
+        t = None
         try:
             with open(path, encoding="utf8") as f:
-                return json.load(f)
-        except:
-            return dict()
+                t = f.read()
+            return json.loads(t)
+        except Exception as e:
+            return gui_hooks.addon_current_config_is_invalid(dict(), e, t)
 
     # in new code, use write_addon_meta() instead
     def writeAddonMeta(self, dir, meta):
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -573,11 +573,13 @@ and have been disabled: %(found)s"
 
     def addonConfigDefaults(self, dir):
         path = os.path.join(self.addonsFolder(dir), "config.json")
+        t = None
         try:
             with open(path, encoding="utf8") as f:
-                return json.load(f)
-        except:
-            return None
+                t = f.read
+            return json.loads(t)
+        except Exception as e:
+            return gui_hooks.addon_default_config_is_invalid(None, e, t)
 
     def addonConfigHelp(self, dir):
         path = os.path.join(self.addonsFolder(dir), "config.md")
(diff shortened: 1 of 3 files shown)


glutanimate:
:+1: for JSON schema support, and potentially utilizing that to generate a GUI form at some point in the future. VS Code does that for its extensions, and it works really well.

FWIW, I looked into this a few months ago, and I did find a few existing solutions that might be helpful:

Qt:

- [qt-jsonschema-form](https://github.com/agoose77/qt-jsonschema-form)

Web:

- [JSON form](https://github.com/jsonform/jsonform)
- [JSON editor](https://github.com/json-editor/json-editor)
- [react-jsonschema-form](https://github.com/rjsf-team/react-jsonschema-form)

Arthur-Milchior:
I hope not to recreate everything from scratch. But that's clearly not something I want to work on without an interest from Damien, even if I end up using some already existing library.
Thank you for the link, if I eventually do it, that's really appreciated.

I clearly expect codes wanting to have their own GUI to have their own. I hope that an automated GUI would be better than a plain text, not that it would be the definitive solution for everything

glutanimate:
@Arthur-Milchior makes sense!

Adding onto this, if we do go with a config schema, it would be nice to add a key that describes whether Anki needs to be restarted for a particular value to apply. If an applicable value was changed, we could then remind users to restart Anki after accepting the config window.

Arthur-Milchior:
Makes perfect sens. Thanks for the idea

dae:
Re the hook, if this is not actually happening in the wild and is just to facilitate your personal development, perhaps two hooks is overkill and you could just have the code print the json error to stdout instead?

No objections to a patch that adds jsonschema validation of the config

Arthur-Milchior:
I thought that the point of adding hooks is to avoid doing monkey patching in hooks which are actually used. So the fact that it is useful for an add-on was positive.
My add-on show error message when json loading fail. It does it for every add-on, since there can be errors in config.json and meta.json of every add-on.

I could create a similar hook, with an extra parameter to know where the error come from.

dae:
I was suggesting you submit a patch that adds a print statement with the json error. Text printed to stdout will be invisible to the average user, but developers tend to run Anki in a console window and can thus see it.

Arthur-Milchior:
Oh okay. I thought you suggested me to create an add-on doing that. Such add-on already exists.
I'll do such a patch

Arthur-Milchior:
qt-jsonschema-form is quite nice. Thanks for the pointer.
It may not yet be perfect for really config, such as enhanced main window. But I believe that it'll be a huge improvement for small config; e.g. most of the usual one with a few value to switch. In particular, the "enum" feature of schema means that it'll be far easier to constrain the user to the expected values.

I forked it and made some change which I think were missing. I'll wait to see if the author respond to decide whether I fork it or whether we'll work on it together. I really hope to be able to incorporate it into anki; either in an add-on or in core anki. Because it may really help all add-on users.

One thing which is still missing however is a button to edit directly the json. Because when an object may have optional values, not in the schema, for example, the form may constrain too much. 

I'll close this PR, since it became useless now that https://github.com/ankitects/anki/pull/492 is merged.
However, I don't know where to resume this discussion with you both (and other people who may be interested)

PR #476, opened 2020-03-01
Hook browser init

Arthur-Milchior:
I moved the deck browser in a separate section of the gui_hooks.

It's the same idea than a recent commit. Being able to change the content of the window after it was initialized and before it is shown.


dae:
Thanks!

PR #477, opened 2020-03-01
Reset can be cancelled

Arthur-Milchior:
I believe that reset is called far too often. When a card is not
modified, it can still be displayed. It's actually sometime useful
when the content of the current card made me know I need to change
another card.
I want to be able to cancel the reset. Even if it means that the
scheduler will not be rebuild and thus in some case this change which
card I see

### DIFF
--- a/qt/aqt/gui_hooks.py
+++ b/qt/aqt/gui_hooks.py
@@ -907,6 +907,31 @@ def emptyNewCard():
 empty_cards_will_be_deleted = _EmptyCardsWillBeDeletedFilter()
 
 
+class _MainWindowShouldResetFilter:
+    _hooks: List[Callable[[bool], bool]] = []
+
+    def append(self, cb: Callable[[bool], bool]) -> None:
+        """(should_reset: bool)"""
+        self._hooks.append(cb)
+
+    def remove(self, cb: Callable[[bool], bool]) -> None:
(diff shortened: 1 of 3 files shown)


dae:
I'm not sure this hook has enough context to be able to make the correct decision without inspecting the call stack or active windows, and if it's going to be doing that, perhaps it should just be placed in the relevant windows instead? I'm also not not keen on the possible bugs this could introduce if someone decides to skip the reset when it was actually required.

Arthur-Milchior:
I guess for my purpose I can place it in the editor instead

dae:
Copying the text you accidentally added to a different PR:

"By the way, whats's wrong with a cancelled reset ?
I understand you should not review a deleted card, because it would add it back to the collection. But if you check that the card and note still exists and that the note content didn't change and that its template didn't change, I don't see why it would be really bad for an add-on to keep the card displayed."

The deletion case was the one I had in mind. There may be other cases that need special attention as well - it has been years since I have touched that part of the code.

Arthur-Milchior:
My intuition would be to tell that reset is forced when the note is deleted or edited. Allow to cancel only in other cases

dae:
Closing this one - if you need more hooks in the other parts of the code to accomplish this instead, please let me know.

PR #479, opened 2020-03-01
Change sched values

Arthur-Milchior:
I created an add-on which add a limit on the sum of new and rev cards by day. This needed to change the number of cards returned by those methods. Hence this add-on

### DIFF
--- a/pylib/anki/hooks.py
+++ b/pylib/anki/hooks.py
@@ -387,6 +387,62 @@ class _NotesWillBeDeletedHook:
 notes_will_be_deleted = _NotesWillBeDeletedHook()
 
 
+class _SchedulerNewLimitForSingleDeckFilter:
+    """Allows changing the number of new card for this deck (without
+        considering descendants)."""
+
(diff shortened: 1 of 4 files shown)


dae:
How about something like scheduler_did_get_review_limit_for_single_deck or schedule_review_limit_for_single_deck?

Arthur-Milchior:
Done

dae:
My apologies Arthur, I missed the 'r' in scheduler in the second example - it should be scheduler as schedule is a verb rather than a noun. The same 'Allows changing' comment from the other PR applies here as well. Otherwise looks good.

dae:
Sorry again, today's changes have created a merge conflict

dae:
(also may not need to pass in the scheduler for the new limit case, since they now share the same function?)

Arthur-Milchior:
For consistency I entirely remove version number everywhere. Anyway, it can always be accessed by doing mw.col.scheds.name

dae:
Thanks!

PR #481, opened 2020-03-01
Fixed rspy/Makefile trying to use python3 on Windows

evandrocoan:
I was not failing on my installation because I also have a `python3` binary on the system path (which works), but when someone else has only `python`, it will fail.


dae:
Thanks!

PR #482, opened 2020-03-01
Hook renaming

Arthur-Milchior:
All of those changes are related to renaming.

First, there seems to have been some errors, I don't really understand how, with the typing of renameForDragAndDrop. Indeed, the parameters are given from a js command, and are a substring of a command. So they are string (potentially empty), and not int. So I corrected that.
It never created any typing problem, probably because the type inference is not good enough to detect it.

I created an add-on which allow to merge decks. Instead of rejecting because of duplicate name, it simply move all cards to the existing deck and destroy the moved deck. After asking user for confirmation. It uses monkey patch. To avoid monkey patch, I need a filter which will do the merge and tell the program that, instead of raising an exception/warning, it should just return.

My add-on also need to compute the new name. The method to do so is almost in anki, except that the ne
 name is not returned but is directly used. So I separated a method in two, so that I can access to the renaming part and no change is seen from the point of view of users and of other add-ons

### DIFF
--- a/pylib/anki/decks.py
+++ b/pylib/anki/decks.py
@@ -284,11 +284,13 @@ class DeckManager:
         "Rename deck prefix to NAME if not exists. Updates children."
         # make sure target node doesn't already exist
         if self.byName(newName):
-            raise DeckRenameError(_("That deck already exists."))
+            raise DeckRenameError(_("That deck already exists."), g, newName)
         # make sure we're not nesting under a filtered deck
         for p in self.parentsByName(newName):
             if p["dyn"]:
-                raise DeckRenameError(_("A filtered deck cannot have subdecks."))
+                raise DeckRenameError(
+                    _("A filtered deck cannot have subdecks."), g, newName
+                )
         # ensure we have parents
         newName = self._ensureParents(newName)
         # rename children
--- a/pylib/anki/decks.py
+++ b/pylib/anki/decks.py
@@ -304,19 +304,27 @@ class DeckManager:
         # renaming may have altered active did order
         self.maybeAddToActive()
 
-    def renameForDragAndDrop(self, draggedDeckDid: int, ontoDeckDid: Any) -> None:
+    def renameForDragAndDrop(
+        self, draggedDeckDid: str, ontoDeckDid: Optional[str]
+    ) -> None:
         draggedDeck = self.get(draggedDeckDid)
+        new_name = self.new_name_for_drag_and_drop(draggedDeck, ontoDeckDid)
+        if new_name is not None:
+            self.rename(draggedDeck, new_name)
+
+    def new_name_for_drag_and_drop(
+        self, draggedDeck: Dict[str, Any], ontoDeckDid: Optional[str]
+    ) -> Optional[str]:
         draggedDeckName = draggedDeck["name"]
         ontoDeckName = self.get(ontoDeckDid)["name"]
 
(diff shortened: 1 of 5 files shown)


dae:
Why don't you use a single hook for the GUI, eg deck_browser_will_show_deck_already_exists_warning?

The pylib code may change in the coming months, breaking this hook.

Arthur-Milchior:
I guessed it could make sens to let hooks be distinct depending on the reason the renaming occured. Drag/drop being distinct from direct renaming. I don't remembrer whether renaming due to sync also use the rename method

Arthur-Milchior:
By the way, whats's wrong with a cancelled reset ?
I understand you should not review a deleted card, because it would add it back to the collection. But if you check that the card and note still exists and that the note content didn't change and that its template didn't change, I don't see why it would be really bad for an add-on to keep the card displayed.

To be more precise, the trouble I see are that, when editions are made, maybe other cards in the queue will disappear. Since the queue contains cids and not actualy cards, it means that loading the card will fail, and it won't be reviewed.
It also means that the queue may miss some cards that it should now contains, such as a card whose deck is changed, or a new card which was just generated. But next time queue is reconstructed it will contains those cards (except if the daily limit is reached, in which case it just means that you'll have seen some cards one day earlier, and some one day later)

I understand that it is not something that you want in main anki. But I don't see the problem created if an add-on decided that it's an acceptable cost in exchange of ensuring that the card remain displayed as long as it's not changed.

dae:
Replied on https://github.com/ankitects/anki/pull/477

Arthur-Milchior:
Oops, sorry for the wrong PR for my question.
I removed the two last hook. Actually, having a hook simply in the back end renaming was enough for my purpose. It's not beautiful since my add-on use front end, but that's not really important

dae:
Sorry, I was not clear in my above reply. I recommend this hook be placed in the GUI code, as the pylib deck handling code is likely to change over the coming months. As a general rule, hooks in the GUI code are more likely to be stable moving forward, so when functionality could be implemented in either end, please prefer the GUI over pylib.

Arthur-Milchior:
Corrected. It's now done in GUI.
I added the informations (deck and new name) in the Exception, since I don't have access to the information where the exception is catched

dae:
Changing the error signature seems like a bit of a hack. For the normal rename case, you already have access to the old and new names. For the dropping case, what about removing renameForDragAndDrop from pylib, and placing its contents in deckbrowser instead? Then you'll have access to the old and new names, and won't need to stuff them into an error.

Arthur-Milchior:
Putting all relevant informations in the exceptions seems quite natural to me. Not a hack at all. And it does not even create any trouble for other add-ons using this exception (if any add-on does it)
This change was due to the fact that you wanted a single hook, while I didn't have the same informations in both cases.
I like the fact that renameForDragAndDrop is in pylib, because putting a deck as a child of another seems to be a natural operation, and not specifically related to this user interface

dae:
You'll need to take a different approach for this now - in master the code will just add a suffix to make the name unique if it's a duplicate, so there's no error to catch anymore. I suggest just adding a hook like deck_will_be_renamed to the rename method - you can use .id_for_name() to see if there's an existing deck with the same name cheaply, and take the desired action if a deck with that name already existed.

Arthur-Milchior:
Thank you for letting me know. I'll take a look at it when I've more time.

PR #484, opened 2020-03-02
hook prettify_json

Arthur-Milchior:
The reason I want this hooks is for my add-on 112201952
In this add-on, to be more readable, I prettify the json.
I also change the data received so that the incorrect json becomes correct before it gets loaded

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -26,6 +26,7 @@ import aqt
 import aqt.forms
 from anki.httpclient import HttpClient
 from anki.lang import _, ngettext
+from aqt import gui_hooks
 from aqt.qt import *
 from aqt.utils import (
     TR,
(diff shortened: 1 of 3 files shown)


dae:
How about addon_config_editor_will_display_json and _will_save_json

Also a minor English point (I'm not being critical - your English is much better than my French!) - "Allows adding / Allows changing" would be more natural than "Allow to add/change" in the docs.

Arthur-Milchior:
Done
Fun fact, if you put "\n" in a hook comment string, develop replace it by an actual new line. Which makes a lot of sens if develop consider strings as strings and not as raw text. Except that I didn't pay attention to it and so obtained strange result

dae:
Would you mind tweaking the docstring as mentioned above as well?

Arthur-Milchior:
Done

dae:
You only adjusted 1 of the 2 hooks. The gui_hooks.py looks like it's including a bunch of unrelated changes as well - please check it.

dae:
Thanks!

PR #485, opened 2020-03-02
Allow to define python bin

evandrocoan:
Merge this after: (Fixed rspy/Makefile trying to use python3 on Windows #481) because this depends on a commit from there. Or do not merge that and only merge this.

With this, the user can set the variables `PYTHON_BIN` and `ACTIVATE_SCRIPT` to change the default python and binary scripts.

I need this on my Debian 9 because there, the default Python is 3.5 (python3), but I have installed/compiled Python 3.7 (python3.7). Then, I used this to change the default Python3 interpreter before calling `sh run`:
```sh
export PYTHON_BIN=python3.7
sh run
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,21 @@
 SHELL := /bin/bash
 
 ifeq ($(OS),Windows_NT)
-	PYTHON_BIN := python
-	ACTIVATE_SCRIPT := pyenv/Scripts/activate
+	ifndef ACTIVATE_SCRIPT
+		ACTIVATE_SCRIPT := pyenv/Scripts/activate
+	endif
+
+	ifndef PYTHON_BIN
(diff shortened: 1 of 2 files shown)


dae:
Maybe conflicts with the previous change?

evandrocoan:
Yes, I am not sure why but it is fixed now after I merged the actual master branch back and solved the conflict.

dae:
Thanks, will merge. By the way, the recent Windows changes have caused the following to print on other platforms, including CI - it would be nice if that could be fixed.

"Makefile:4: warning: undefined variable 'OS'"

evandrocoan:
I fixed it on Fixed "Makefile:4: warning: undefined variable 'OS'" #491

Any other problem with Windows you can just ping @evandrocoan

PR #491, opened 2020-03-03
Fixed "Makefile:4: warning: undefined variable 'OS'"

evandrocoan:
All `Makefiles` on this project are using `MAKEFLAGS += --warn-undefined-variables` and the variable `OS` is only defined on Windows. So, to suppress that warning, define the variable `OS` if it is not defined.


dae:
Thanks!

PR #492, opened 2020-03-04
print json decode error message

Arthur-Milchior:
As discussed in another thread. The main point is to allow add-on dev' to debug their own json.

Ideally, I'd have preferred to print to stderr, since it's an error. But it would prompt a warning window. So I stayed on standard output


dae:
Thanks!

PR #493, opened 2020-03-04
Type-annotate editor.card

glutanimate:
Super-small change to pacify a mypy error I ran into while working on an add-on.


dae:
Cheers!

PR #494, opened 2020-03-04
Debug hook

Arthur-Milchior:
The goal of this hook is to create an add-on which may:
* put a default value in the debug system. Since I always start by ```from aqt import mw
col = mw.col```
adding this text may win myself some time
* put the last value. Indeed, it occurs that I need to restart anki and try something again. This will avoid me the need to copy paste debug code

### DIFF
--- a/qt/aqt/gui_hooks.py
+++ b/qt/aqt/gui_hooks.py
@@ -520,6 +520,34 @@ class _CurrentNoteTypeDidChangeHook:
 current_note_type_did_change = _CurrentNoteTypeDidChangeHook()
 
 
+class _DebugConsoleDidEvaluatePythonFilter:
+    """Allows processing the debug result. E.g. logging queries and
+        result, saving last query to display it later..."""
+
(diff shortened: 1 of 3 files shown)


dae:
- typo in "debugf"
- please use debug_console_will_show, debug_console_did_evaluate_python
- perhaps the latter hook should be able to modify the text before it is inserted as well?

Arthur-Milchior:
Corrected.
For my purpose, I don't expect to edit the text. But it certainly does not hurt.
Do you know when the exception may be raised ? When adding string to string ? Or when showing the text. This would allow to limit the part in the try/except

Arthur-Milchior:
In this case, the error does not seems to be related to my PR.
I did hook "make check" and it is run when I push. And check did succeed on my computer

dae:
Please rebase on master which should fix the failing check, and also 'debugf' is still there.

Arthur-Milchior:
done

dae:
cheers!

PR #495, opened 2020-03-04
add-on config check schema of user config

Arthur-Milchior:
As discussed in another PR:
If there is a file "schema.json" then configs are validated against it before being accepted.

I suggest that you also validate config.json against schema.json when an add-on is uploaded. I can't write the code given that ankiweb is private. 
The reason being that if the default config is not valid, it's going to be pretty hard for users to do anything about it.

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -594,6 +594,24 @@ and have been disabled: %(found)s"
     def configUpdatedAction(self, addon: str) -> Callable[[Any], None]:
         return self._configUpdatedActions.get(addon)
 
+    # Schema
+    ######################################################################
+
+    def _addonSchemaPath(self, dir):
+        return os.path.join(self.addonsFolder(dir), "config.schema.json")
+
+    def _addonSchema(self, dir):
+        path = self._addonSchemaPath(dir)
+        try:
+            if not os.path.exists(path):
+                # True is a schema accepting everything
+                return True
+            with open(path, encoding="utf-8") as f:
+                return json.load(f)
+        except json.decoder.JSONDecodeError as e:
+            print("The schema is not valid:")
+            print(e)
+
     # Add-on Config API
     ######################################################################
 
@@ -1304,6 +1322,12 @@ class ConfigEditor(QDialog):
         txt = gui_hooks.addon_config_editor_will_save_json(txt)
         try:
             new_conf = json.loads(txt)
+            jsonschema.validate(new_conf, self.parent().mgr._addonSchema(self.addon))
+        except ValidationError as e:
+            # The user did edit the configuration and entered a value
+            # which can not be interpreted.
+            showInfo(tr(TR.ADDONS_CONFIG_VALIDATION_ERROR, problem=e.message))
+            return
         except Exception as e:
             showInfo(_("Invalid configuration: ") + repr(e))
             return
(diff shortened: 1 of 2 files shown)


dae:
How about config.scema to make it clearer what the schema is for?

Please use a fluent string instead of _(), like in https://github.com/ankitects/anki/pull/454

dae:
Suggested key:

addons-config-validation-error = There was a problem with the provided configuration: {$problem}

glutanimate:
I would suggest going with 'config.schema.json'. From what I've seen so far that seems to be a common naming scheme, and VS Code for instance recognizes the extension and provides autocompletion for schema keys.

Arthur-Milchior:
I'm 100% with Glutanimate here. I don't know VS code. But I really want to end with .json, because otherwise my text editor won't format correctly the file, and won't do proper syntax coloring.

dae:
Thanks Arthur, no objections to the .json ending. In future patches, for new functions/variables, please use snake_case instead of trying to match the camelCaps of the legacy functions so we can gradually move towards a consistent styling.

Arthur-Milchior:
I am suddenly realizing one thing which should be noted.
In this case, error messages will be in English only. It seems impossible to translate them, which is a bad news. But it seems that jsonschema library has no localization at all.

This is particularly noticeable because in this case, that's error message that users are supposed to see. Most python errors message are never supposed to be seen, they mostly means an add-on broke, or that something very unexpected occurred. In this case, it is expected that users may enter incorrect configuration and need to see the message in order to correct themselves

Arthur-Milchior:
I think there is a thing that still need to be decided: how to warn about problem when the schema is not valid json.
Hopefully, this is a problem that most users won't have; only add-on developers. But currently, the message is totally uninformative, and no config can be saved when the schema is an invalid json file

glutanimate:
Yeah, that could be an issue for sure – both in terms of the language barrier and the programming knowledge barrier. But I really can't think of any simple and good solution that would improve the readability of validation errors to regular users. We will probably just have to accept that a semi-understandable validation error at the config stage still is better than unforeseen behavior and errors later down the line because of invalid config values.

Other than that, the only definitive solution I see is restricting user input proactively through a GUI (like you're working on, Arthur).

BTW, I know that you ended up going with `qt-jsonschema-form` – that would have been my first choice as well based on preference. However, from briefly looking at the web solutions, they did seem to be better maintained (and with a wider range of support for the jsonschema specs).

If you run into many limitations with `qt-jsonschema-form` that would force you to ship your own solutions, it might be worthwhile to explore the web options instead, I feel.

Though at this point we should probably move this discussion to tenderapp or something as it's getting hard to keep an overview across multiple PRs, etc.

glutanimate:
One solution we could with, though, would be to provide add-on authors with the ability to display a custom error message instead of the default validation error message. This SO answer shows an example of that in practice: https://stackoverflow.com/a/57360801/1708932

Arthur-Milchior:
I like the "error_msg" idea.
I don't like that it means I'll need to change all of my new schema.

I've a preference for the python solution, because it means that it can easily be hooked.
Furthermore, I also want to open new window (e.g. in some case it will be nicer to have objects in their own subwindow instead of having them in the main window), which seems quite easier with qt than with web. At least the way qt deals with web page.

Anyway, I'll share my add-on probably this week. And I'll see later whether user feedback makes me reconsider my choice

PR #496, opened 2020-03-06
Use anki.consts.SYNC_BASE instead of hardcoded url

TooYoungTooSimp:
Use `anki.consts.SYNC_BASE` instead of hardcoded url in function `MediaSyncer._endpoint`


dae:
cheers

TooYoungTooSimp:
Thank you very much!

PR #497, opened 2020-03-06
hook addons_dialog_will_show

Arthur-Milchior:
I am creating an add-on which, reading the schema, does a GUI automatically. I thus want to add a button to open this GUI. And so want to add a hook to edit the window before it is shown

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -706,6 +706,7 @@ class AddonsDialog(QDialog):
         self.setAcceptDrops(True)
         self.redrawAddons()
         restoreGeom(self, "addons")
+        gui_hooks.addons_dialog_will_show(self)
         self.show()
 
     def dragEnterEvent(self, event):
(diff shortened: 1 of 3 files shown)


dae:
Hook looks good, but please use "Allows changing" like in your other PRs

Arthur-Milchior:
Corrected.
Sorry. I try to remember it. It sounds quite strange to my hear.

Note that I did more change than just the syntactical one. For some reason I now had a typing problem I didn't have before. I stated that the parameter "addon" was a string while actually it's a AddonMeta

dae:
Starting the new hook name you introduced with addons_ will make it easier for people to find. Maybe addons_dialog_did_change_selected_addon?

Arthur-Milchior:
Oh okay.
I rreally was not able to find a way to form a sentence starting by add-on, it seemed to me that add-on would appear later naturally. But Iguess that epeating add-on is okay too

dae:
cheers

PR #498, opened 2020-03-06
Fix media files being marked as unused

evandrocoan:
Fix media files being marked as unused when media files are used in fields only with the media name: https://anki.tenderapp.com/discussions/ankidesktop/39466-create-a-special-tag-to-not-auto-play-the-audio

For example, one can create flashcard fields only with the media name as `AudioEnglish=mysound.mp3` and on the HTML template, use it as `[sound:{{AudioEnglish}}]`. This also allows to create JavaScript play buttons as:

1. image
1. BackSide
    ```html
    {{FrontSide}}
    <hr id=answer>

    <br>
    [sound:{{AudioEnglish}}]
    <input id="clickMe" type="button" value="Greek" onclick="playaudio('{{AudioGreek}}');" />
    <input id="clickMe" type="button" value="Portuguese" onclick="playaudio('{{AudioPortuguese}}');" />

    function playaudio(filename) {
        var audio = new Audio(filename);
        audio.play();
    }
    </script>
    ```
1. image

### DIFF
--- a/rslib/src/media/check.rs
+++ b/rslib/src/media/check.rs
@@ -17,6 +17,7 @@ use crate::text::{normalize_to_nfc, MediaRef};
 use crate::{media::MediaManager, text::extract_media_refs};
 use coarsetime::Instant;
 use std::collections::{HashMap, HashSet};
+use std::iter::FromIterator;
 use std::path::Path;
 use std::{borrow::Cow, fs, time};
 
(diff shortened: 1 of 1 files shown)


dae:
I'm afraid I don't think comparing every field against the media files is the best solution - text that is supposed to be treated as a media file should be marked as such. I'd like to better support this use case in the future (and your original goal of sound tags that don't automatically play), just don't have time to think about it at the moment I'm afraid.

evandrocoan:
This should not be a problem because we can mark which fields are media and which ones are not, by using the `Fields...` editor:
image

I just did not figure out yet how to read the `This fields is a media name` on the rust code.

What do you think about this? I was adding it to allow anki to mark these files as missing and to make those fields be plain text only, i.e., without any HTML markup as in: https://anki.tenderapp.com/discussions/beta-testing/1838-why-anki-is-adding-an-extra-div-around-my-sound-tag

dae:
That's a better solution than checking every field, but I'd like to think about it more when I'm not so busy with the recent changes :-)

PR #499, opened 2020-03-07
Allow add-on authors to register their dialogs with the dialog manager

glutanimate:
Extends `aqt.DialogManager` with type hints and a new public API, `register_dialog`.

`register_dialog` covers two common design patterns in add-on development:

    - Creating a custom dialog and having it be managed by Anki's dialog manager
    - Replacing a default dialog with your own implementation

One concrete use case would be in "External Note Editor for the Browser": https://github.com/glutanimate/anki-addons-misc/blob/master/src/browser_external_editor/browser_external_editor.py#L113

But there are numerous other examples of add-ons extending `aqt.dialogs._dialogs`. This new method will hopefully provide a more stable way of doing things.

### DIFF
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -130,7 +130,18 @@ class DialogManager:
         self, name: str, creator: Union[Callable, type], instance: Optional[Any] = None
     ):
         """Allows add-ons to register a custom dialog to be managed by Anki's dialog
-        manager
+        manager, which ensures that only one copy of the window is open at once,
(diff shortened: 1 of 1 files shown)


dae:
Cheers!

Arthur-Milchior:
I do wonder how it will interact with https://ankiweb.net/shared/info/354407385
I don't have enough intuition about the closing part to know what will occur exactly.
However, it seems that it will indeed allow to open a window twice if the "default" is set to `true`. Or if the name of the window is entered explicitely.

PR #500, opened 2020-03-09
More precise schema failure message

Arthur-Milchior:
While creating schema for the add-on I used, I realized that my error
message was not precise enough.

For a big configuration such as in "Enhance main window", the message

> None is not of type 'string'

is almost useless.
So I added a path in the json config, to indicate where the error
is. And also which subschema is currently being used.


dae:
Thanks! Please note your original patch was missing a newline at the end of the line, which is what was causing https://anki.tenderapp.com/discussions/beta-testing/1836-problem-with-new-translations

Arthur-Milchior:
Thanks. I had no idea it was mandatory.

PR #501, opened 2020-03-10
Renamed BeautifulSoup import name for consistency with other usages of it

evandrocoan:
This was initially a fix for https://anki.tenderapp.com/discussions/ankidesktop/39543-anki-is-replacing-the-character-by-when-i-exit-the-html-edit-mode-ctrlshiftx

But now, I am just renaming a variable to keep it consistent with the rest of the code which uses BeautifulSoup.


dae:
Thank you, but probably not worth making such minor changes in the future if there's no clear benefit to it.

PR #502, opened 2020-03-10
normalize name in _checkDeckTree

Arthur-Milchior:
Currently, if there is two decks "A" and "a", they won't be considered
as duplicate. So they won't be corrected.
That's hard to test, because you're not supposed to be able to create
those two decks. However, it seems that it may still be possible on
AnkiDroid (correction is in master, and probabla alpha)

Thus, I suggest to compare normalized names instead of names
themselves.
Actually, I'm even surprised that I didn't do this PR while doing dae/anki@e44b0492781e8248318e87a47eb179561db94b7b

### DIFF
--- a/pylib/anki/decks.py
+++ b/pylib/anki/decks.py
@@ -491,26 +491,27 @@ class DeckManager:
 
         for deck in decks:
             # two decks with the same name?
-            if deck["name"] in names:
+            deck_name = self.normalizeName(deck["name"])
+            if deck_name in names:
                 self.col.log("fix duplicate deck name", deck["name"])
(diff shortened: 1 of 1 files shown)


dae:
You did implement it in the past, and it broke things :-)

5418af00f733ca62b0c087d1422feae01d6571b0

Arthur-Milchior:
I don't recall this. I mean, I do remember breaking things, but not with this particular change.

dae:
I don't know if I reported it to you at the time as I assumed you were following the commits, but I rolled it back when I found it was breaking things!

PR #503, opened 2020-03-11
Error msg in schema

Arthur-Milchior:
This PR follows @glutanimate's advice https://github.com/ankitects/anki/pull/495#issuecomment-596685227
The add-on dev can put a error message to explain more precisely the exact error and what is expected. This is a formatted string which can take a few parameters (but does not have to)

I also corrected my caml case to snake case


dae:
cheers

PR #504, opened 2020-03-13
Fixed BeautifulSoup breaking string paths

evandrocoan:
https://anki.tenderapp.com/discussions/ankidesktop/39543-anki-is-replacing-the-character-by-when-i-exit-the-html-edit-mode-ctrlshiftx


dae:
thanks!

for the future, 

if "<" in html

is easier to read than .find()

evandrocoan:
You are right, I totally forgot I could use `if "<" in html`.

PR #505, opened 2020-03-16
Give the 'typearrow' id to the type answer arrow

evandrocoan:
So, it can be styled by the user with something like:
```css
#typearrow {
  color: #e56727;
  position: absolute;
  margin-top: -7px;
}
```

Making it centered and with a different color:

image

Instead of:

image


dae:
looks good, thanks!

PR #506, opened 2020-03-16
Hook editor_web_view_did_init

Arthur-Milchior:
I hope to use this in oder to had features to the web view. In particular, I expect the "spell checking" feature to use it hopefully. Because currently it's incompatible with other add-on dealing with the editor.

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -954,6 +954,7 @@ class EditorWebView(AnkiWebView):
         self._markInternal = False
         clip = self.editor.mw.app.clipboard()
         clip.dataChanged.connect(self._onClipboardChange)
+        gui_hooks.editor_web_view_did_init(self)
 
     def _onClipboardChange(self):
         if self._markInternal:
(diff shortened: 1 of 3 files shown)


glutanimate:
@Arthur-Milchior Which spell checking add-on were you seeing conflicts with (and which add-on of yours if affected)? If it's my spell checker I'll be happy to update it accordingly.

and +1 for the hook, I could see it being useful

Arthur-Milchior:
It was yours. It was having a conflict with at least two add-ons. 
https://ankiweb.net/shared/info/2064123047 (this is still the case) and with the resize image add-on (this conflict has been resolved on my side since.)
Actually, I only installed the add-on because the anKing told me resizer and spell checker did conflict. I don't use it myself. In particular I never subscribed to your patreon, so I would not have had access to this add-on if the anKing didn't send it to me.
I actually believe I could use this hook for some of my own add-ons, but since Damien ask for a single example of use, I took the first example I was thinking about.

glutanimate:
Thanks for letting me know! I'll see if we can fix things with the hook, or opt for a different solution if that doesn't work.

dae:
Thanks!

PR #507, opened 2020-03-16
reminder: special fields when importing

Arthur-Milchior:
I add three hooks which may be used for the add-on "Special fields". 
There is a real problem here, which has been hard to solve.
I want the hooks to have access to the type Anki2Importer. But if I don't import anki.importing.anki2 (or at least anki.importing ?) mypy tells 
mypy error: Name "anki.importing.anki2.Anki2Importer" is not defined [name-defined]
However, if I decide to import anki.importing.anki2, it executes anki/importing/__init__.py which creates a loops of import, since __init__.py imports anki2, which imports anki.collection, and so on...

So "if false" before the import seems to be required for mypy to accept. Which leads to pylint error 
anki/hooks.py:23:0: W0125: Using a conditional statement with a constant value (using-constant-test)
So I need to tell pylint to disable this warning

### DIFF
--- a/pylib/anki/hooks.py
+++ b/pylib/anki/hooks.py
@@ -323,6 +323,48 @@ class _ImporterDoesItUpdateNoteFilter:
 importer_does_it_update_note = _ImporterDoesItUpdateNoteFilter()
 
 
+class _ImporterDoesItUpdateNoteTypeFilter:
+    """Decides whether a note type should be replaced when importing a
+        collection. By default, a note type is replaced iff its last
+        modification date is more recent in the imported deck than in
+        the current collection. It is assumed that the schemas are the
+        same, otherwise update is technically impossible anyway."""
+
+    _hooks: List[
+        Callable[[bool, "anki.models.NoteType", "anki.models.NoteType"], bool]
+    ] = []
+
+    def append(
+        self, cb: Callable[[bool, "anki.models.NoteType", "anki.models.NoteType"], bool]
+    ) -> None:
+        """(importing: bool, srcModel: anki.models.NoteType, dstModel: anki.models.NoteType)"""
+        self._hooks.append(cb)
+
+    def remove(
+        self, cb: Callable[[bool, "anki.models.NoteType", "anki.models.NoteType"], bool]
+    ) -> None:
+        if cb in self._hooks:
+            self._hooks.remove(cb)
+
+    def __call__(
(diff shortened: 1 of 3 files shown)


dae:
Regarding the issue you were experiencing, a cleaner way to work around it may be to use 'if TYPE_CHECKING:'.

But I will need to hold off on these hooks for now, as the importing code is likely to change in the coming months, and it may make sense to incorporate some or all of this functionality directly into Anki. Will leave this open as a reminder

Arthur-Milchior:
Thanks for the advice. I didn't know this trick.
I won't correct my PR if you don't think you'll merge it.
If it's not already the case, I strongly suggest you discuss importing functionnality with @ankingmed. I'm not really sure why the user of his deck needs so many feature to decide how to import decks, but if he is to be believed, he knows what will be useful to thousands of users at once

AnKingMed:
@dae I'll just comment to let you know how I'm using this.  I've got thousands of medical students using a deck for medical school that I constantly update (correct errors, add new info, etc).  The special fields add-on allows users to "protect" certain fields while getting updates in other fields when they import a deck. It does a bunch of other things as well to control what gets updated and what doesn't. It's absolutely vital to the updating process (couldn't be done without it).  Arthur and another user (John Pincock) have been really kind to help me maintain this add-on.  I just wanted to let you know so you understand its popularity and how it is being used.

dae:
Thanks for elaborating; I'll keep this is mind when I have a chance to rework the code, and may ask you to elaborate on the workflow at that point

AnKingMed:
No problem! More than happy to explain anything you need. I checked with the addon author and he said there's been 12,000 downloads so my estimate was correct in how many are using it.  I made a video on the update process here if you want to watch that, but I'm also more than happy to detail out any specifics you need:  https://youtu.be/cg-tQ6Ut0IQ

Arthur-Milchior:
I'd like to note that, unless I'm getting something extremely wrong, updates are also counted as downloads. Not really ideals (especially since in my job interviews I gave this number to prove that I create code with a real user base. Hopefully, this was not used to much to decide to hire me), but that means that the actual number of people who downloaded the add-on may be less than 12 000. (And of course, actual users is less than the number of downloaders)

dae:
Closing this in favour of a separate issue: https://github.com/ankitects/anki/issues/1018

PR #508, opened 2020-03-17
Template filter all field

Arthur-Milchior:
I love the add-on which allow to edit fields directly in reviewer.
I'd appreciate to let it be applied to each field. Without having to manually enter a filter name in each field. This this hook

### DIFF
--- a/pylib/anki/hooks.py
+++ b/pylib/anki/hooks.py
@@ -594,6 +594,40 @@ class _TagAddedHook:
 
 
 tag_added = _TagAddedHook()
+
+
+class _TemplateFilterAllFieldsFilter:
+    """Apply hooks to each field content, during rendering, once all
+        explicitly called filters are applied."""
+
+    _hooks: List[Callable[[str, "anki.rsbackend.TemplateReplacement"], str]] = []
+
+    def append(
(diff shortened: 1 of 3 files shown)


dae:
That won't work - only fields with custom filters are passed back to Python, so that hook would only receive the ones that already have custom filters on them.

Arthur-Milchior:
Oh okay. I didn't realize that the back end just returned a single string if all fields are known.
Is there any easy way to have this kind of effect with a hook ? I still would find it quite practical. However, I don't know rust at all right now, so I won't try to change it immediately

dae:
I'm afraid not

Arthur-Milchior:
Would you accept to reopen this PR ? 
I just pushed a new commit, which ensure that if the hook template_filter_all_field is non-empty, then each fields are returned separately. This allow the hook to works and does not slow anki unless an add-on uses such a hook.
I assume that at some point, you may want to use rust to quickly check whether a card is "empty" or not instead of relying on 'req'. If it is ever the case, more change will need to occur in `render_card_from_context` in order to ensure that no `return_all_fields` remains true (and probably that no filter is applied, in order to win some time, and that only question is computed...)
But I'll leave this to later commits if and when it's necessary

PR #512, opened 2020-03-18
Add pylib hook sync_chunk_will_be_applied

hgiesel:
This adds a new pylib hook for review chunks. 

This allows the user to inspect the chunks, before they are uploaded, however it does not allow them to alter it (as it shouldn't).

My use case is the add-on [Straight Reward](https://ankiweb.net/shared/info/957961234). It inspects the changes made to the revlog to update the ease of certain cards. This hook additionally let's you inspect all the other updates in the chunk. My thinking is, that one hook to inspect all three would be better than 3 to inspect each by itself, because it's easier to maintain and a little less overhead.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -29,6 +29,7 @@ Yoonchae Lee
 Evandro Coan <github.com/evandrocoan>
 Alan Du <alanhdu@gmail.com>
 Yuchen Lei <lyc@xuming.studio>
+Henrik Giesel <hengiesel@gmail.com>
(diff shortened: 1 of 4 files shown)


dae:
I'm afraid the syncing code is likely to be rewritten in Rust in the future, which will prevent this hook from working. Instead of intercepting the sync, your add-on could accomplish this functionality by noting down seen revlog ids in the last few days in its own data structure, and then when loading the profile, checking which revlog ids are new.

PR #513, opened 2020-03-18
Pass the flag --autoplay-policy=no-user-gesture-required

evandrocoan:
To the QtWebEngine:
1. https://anki.tenderapp.com/discussions/beta-testing/1858-can-you-pass-autoplay-policyno-user-gesture-required-to-chrome-engine

### DIFF
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -234,8 +234,6 @@ class AnkiApp(QApplication):
     TMOUT = 30000
 
     def __init__(self, argv):
-        # https://anki.tenderapp.com/discussions/beta-testing/1858-can-you-pass-autoplay-policyno-user-gesture-required-to-chrome-engine
-        argv.append("--autoplay-policy=no-user-gesture-required")
         QApplication.__init__(self, argv)
         self._argv = argv
 
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -234,6 +234,8 @@ class AnkiApp(QApplication):
     TMOUT = 30000
 
     def __init__(self, argv):
+        # https://anki.tenderapp.com/discussions/beta-testing/1858-can-you-pass-autoplay-policyno-user-gesture-required-to-chrome-engine
+        argv.append("--autoplay-policy=no-user-gesture-required")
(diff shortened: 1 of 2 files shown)


dae:
Did you see my reply on tenderapp?

evandrocoan:
Not yet. I will look now.

dae:
I'm not sure it makes sense to apply this as the default across all webviews - consider a user using an add-on that displays arbitrary webpages for example. You could probably do this in a little add-on, and only apply it to the reviewing-related webviews?

evandrocoan:
In my experience, it is not usual to have websites trying to play sounds without user interaction. Just some quite bad news/advertising websites would do that. So it should not be a problem because who uses anki to browse the web?

It does not seem nice to have users installing an addon just to use a deck that plays sounds with JavaScript. Addons should be only for custom behaviors which affect the way anki works for a small selected group of people. I could restrict this applying it to the classes:
```java
F:\anki\qt\aqt\editor.py:
  993:         AnkiWebView.__init__(self, title="editor")

F:\anki\qt\aqt\main.py:
  732:         self.web = aqt.webview.AnkiWebView(title="main webview")

F:\anki\qt\aqt\browser.py:
 1591:         self._previewWeb = AnkiWebView(title="previewer")
```

Even if restricting it to these classes does not help, that addons could revert this before opening web pages.

dae:
Anki does not advertise any support for playing audio with Javascript, and most users will just use the [sound:....] tags that Anki provides, so I'd argue this is actually a niche use case. I will think about it and get back to you.

evandrocoan:
I think I will not be using Javascript because when you open the review edit note type window, it creates two qtwebengines and each one of them is a full browser. So, if the sounds are played by Javascript, they would play simultaneous on the front and on the back.

But it would be nice if anki did not limited Javascript from being fully used if someone feels like using it.

dae:
As this is not an officially supported use case at the moment, I think it's probably better to leave things as they are for now - it should be easy to override in an add-on for users who want to do so. Thanks for the contribution in any case, and for all of the hard work you have put in recently - Anki's build system is considerably nicer thanks to your efforts!

PR #515, opened 2020-03-19
reminder: use card template names in empty cards screen

Arthur-Milchior:
I expect to use this with my add-on 25425599. This add-on explain which are the name of the cards that are going to be deleted, instead of only their ord. I find it quite clearer when I just changed a card type and want to make sure that the correct cards will be removed/try to understand a bug.

I believe this add-on should be incorporated in anki by the way, it's simple and quite more helpful

### DIFF
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -583,19 +583,19 @@ select id from notes where id in %s and id not in (select nid from cards)"""
         return rem
 
     def emptyCardReport(self, cids) -> str:
-        rep = ""
+        rep = []
         for ords, cnt, flds in self.db.all(
             """
 select group_concat(ord+1), count(), flds from cards c, notes n
 where c.nid = n.id and c.id in %s group by nid"""
             % ids2str(cids)
         ):
-            rep += self.tr(
+            line = self.tr(
                 TR.EMPTY_CARDS_CARD_LINE,
                 **{"card-numbers": ords, "fields": flds.replace("\x1f", " / ")},
-            )
-            rep += "\n\n"
-        return rep
+            ) + "\n\n"
+            rep.append(line)
+        return "".join(rep)
 
     # Field checksums and sorting fields
     ##########################################################################
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -584,16 +584,20 @@ select id from notes where id in %s and id not in (select nid from cards)"""
 
     def emptyCardReport(self, cids) -> str:
         rep = []
-        for ords, cnt, flds in self.db.all(
+        for ords, cnt, flds, nid in self.db.all(
             """
-select group_concat(ord+1), count(), flds from cards c, notes n
+select group_concat(ord+1), count(), flds, n.id from cards c, notes n
 where c.nid = n.id and c.id in %s group by nid"""
             % ids2str(cids)
         ):
-            line = self.tr(
-                TR.EMPTY_CARDS_CARD_LINE,
-                **{"card-numbers": ords, "fields": flds.replace("\x1f", " / ")},
-            ) + "\n\n"
+            line = (
+                self.tr(
+                    TR.EMPTY_CARDS_CARD_LINE,
+                    **{"card-numbers": ords, "fields": flds.replace("\x1f", " / ")},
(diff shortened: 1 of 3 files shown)


dae:
This code is likely to be refactored in the coming months, and I agree with you that the numbers should be changed to template names, so rather than merge this I'll leave it open as reminder to implement this functionality

Arthur-Milchior:
Is there by any luck a list of things you expect to change. That might be quite interesting. Not so much for PR, since this does not even take 10 minutes to create. But so that I know to be wary when I create add-ons dealing with those parts of the code.

Arthur-Milchior:
As an advance notice, related to open PR:
I expect that sometime in April or May, I'll need to stop contributing to Anki's code. So if there is any PR still open, I won't be able to participate to them anymore. I expect to still be active on AnkiDroid, add-ons, subreddit, etc... but may not be able to release any code under AGPL, even for personal projects.

dae:
I presume that means you landed a job - congratulations! Bit surprised they prohibit you from contributing to an AGPL project in your free time but allow GPL. I've heard of companies not allowing the use of AGPL software when at work, but haven't heard of ones that go so far as to ban contributions to AGPL software at home - might be worth double-checking the contract just to be sure.

My primary focus at the moment is the code in pylib. Once the changes in .24 settle, I'd like to start work on improving the deck manager/note type manager/etc, and slowly move away from the current untyped json code.

Arthur-Milchior:
Thank you.
I indeed received a job offer yesterday.
One of the manager I spoke with thought that ankidroid/addons may qualify for the 20% time. Anki does not, that at least is sure: https://opensource.google/docs/using/agpl-policy/
It seems that, unless otherwise specified (https://opensource.google/docs/iarc/), the copyright of all code written even on personal time will belong to them. As long as it's release under a free license, it does not change anything in practice, appart that I'll need to add a remark about it in my commits.
I may ask for a permission, by declaring officially that my contributions are personal and not professional, but that may be denied. https://opensource.google/docs/patching/#forbidden

Anyway, to be honest, I guess that now that I'll have a job again, I'll have far less free time. Especially since it means moving to another continent (I mean, after the end of the coronavirus.) So this remarks is mostly theoretic right now. While I still expect to review a lot of things will anki, I will probably not have time to do all features I was looking for

dae:
Congrats! I applied for a job at Google back when I was living in Japan, but they didn't want to have me :-)

I think @agentydragon is a Googler and they have submitted PRs in the past, so I'm not sure what the exact situation is there, but I certainly imagine you will be busier in the future in any case!

Arthur-Milchior:
@glutanimate already told me about @agentydragon. We realized that we actually met twice without knowing that we were two contributors to anki. Maybe I wast just being too paranoid and everything will be okay. I just didn't want to have to leave a lot of things open without advance warning.

dae:
That's quite a coincidence!

Thanks for the heads up :-)

Arthur-Milchior:
Not exactly to be honest. The rationalist community (less wrong) is where I've met him and where I discovered SRS and Anki.

dae:
Card template names are in the empty cards screen in master; please let me know how it goes for you.

Arthur-Milchior:
I've just taken a look at qt/aqt/emptycards.py
I believe I'd still appreciate to have hooks in it. I do like having links. However, there are two changes I'd like to be able to do: changing the list of card to remove (I've an add-on which allow to remove only new cards; this ensure I don't remove cards I've reviewed, and which became empty by accident), and a hook which allow to change the message, so that I can decide to display more informations.

I fail to start anki from source right now; for some reason it now tells that protoc is not installed; so I can't test directly.

dae:
Happy to accept a hook that applies changes to EmptyCardsReport. If your add-on wants to modify .notes (removing particular cards, etc), it will need to modify or re-create .report as well

Arthur-Milchior:
I'll do it as soon as I figure out how to have protoc3 working on my machine, so I can make develop.

You may be interested in knowing that the hook empty_cards_will_be_deleted is not called anywhere anymore.

dae:
I've updated the hook - you should be able to access .report on the dialog if you want to modify it

PR #517, opened 2020-03-19
hook backup_is_done

Arthur-Milchior:
I expect to use this hook in add-on 529955533. It'll serve me to create long term backup at the same time than your backup


dae:
backup_did_complete would be keeping more in line with the other hooks; will rename this after merge

Arthur-Milchior:
Thank you for taking care of renaming things. I fear that letting you do this work is going to make us both be quicker, given that I can't get my head around the exact name rules

PR #521, opened 2020-03-22
Template filter all field

Arthur-Milchior:
This is a reopening of https://github.com/ankitects/anki/pull/508 , with more changes to ensure that it is indeed possible to apply a hook to all fields. The main reason to re-open is that it allows me to put a link to a PR which list the correct list of changes.

### DIFF
--- a/proto/backend.proto
+++ b/proto/backend.proto
@@ -194,6 +194,7 @@ message RenderCardIn {
     string answer_template = 2;
     map<string,string> fields = 3;
     int32 card_ordinal = 4;
+    bool return_all_fields = 5;
 }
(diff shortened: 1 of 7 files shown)


dae:
Currently the Python code grabs the note data and template data from the database via Rust, then sends it back to Rust to render, and Rust sends back the result. I'd like to avoid this roundabout step in the future by having Rust just send the (partially rendered) template directly, so I'm afraid I can not accept this change. Instead of trying to dynamically alter the template content, you could do a find&replace on your templates with a few lines of code, and then there would be no need to modify them at run time.

PR #523, opened 2020-03-23
Hook editor_note_will_load

Arthur-Milchior:
A current problem I have is that there is nothing similar to hook inside of javascript. It seems that it would be easier to be able to add other methods in javascript and call them in loadNote. Currently I simply redefined loadNote, which is far from perfect

I am not sure at all that's the best way to achieve this feature. But I am sure that I want to execute some function when a note is loaded. In particular because add-ons changing the editor need to pass more informations to the code

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -456,6 +456,7 @@ class Editor:
             json.dumps(focusTo),
             json.dumps(self.note.id),
         )
+        js = gui_hooks.editor_will_load_note(js, self.note, self)
         self.web.evalWithCallback(js, oncallback)
 
     def fonts(self) -> List[Tuple[str, int, bool]]:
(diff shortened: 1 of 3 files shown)


dae:
Please use 'editor_will_load_note' to match the existing 'editior_did_load_note', and "allows changing"

Arthur-Milchior:
Changed
Merge conflict corrected

On master, I can't run "make check" anymore, it directly return "make: *** [Makefile:123: check] Error 127
"

dae:
it still says "allow to change", and it looks like you've accidentally included package-lock.json changes. While you're fixing those issues, would you mind switching the editor and js arguments so editor comes first?

dae:
(can't reproduce the makefile issue - could it be related to trailing whitespace?)

dae:
(of course it's a filter - nevermind about the switching args comment :-)

PR #524, opened 2020-03-23
Ident html in editor

Arthur-Milchior:
Sometime I like to print html to understand why it bugs. And I really prefer to have html indented if it's okay with you. There is no change appart from the identation and quotes


Arthur-Milchior:
The second commit add an id to the name part of the table. I believe that it would be quite simpler to add button (such as "frozen", "single line" etc..) if the name could be accessible more easily

dae:
cheers

PR #525, opened 2020-03-23
Change not not to bool

Hk-tang:
Replace `not not` with `bool` as it is clearer to understand the logic and is more pythonic.


dae:
thanks

PR #526, opened 2020-03-24
strip to test whether field is empty

Arthur-Milchior:
There are two reason for this PR.
The first one is that a field with only spaces and new line should probably be considered as empty anyway.
The second is that my last PR created a bug. The field value contained far more spaces than it should have. I didn't realize that because of this, even if those spaces were not taken into account in the GUI, there were taken into account to decide that the field is not empty
https://github.com/ankitects/anki/pull/524

Of course, I could also correct the change done in the javascript file. But in this case, I believe that this correction is more appropriate

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -496,9 +496,10 @@ class Editor:
             return True
         m = self.note.model()
         for c, f in enumerate(self.note.fields):
+            f = f.replace("<br>", "").strip()
             notChangedvalues = {"", "<br>"}
             if previousNote and m["flds"][c]["sticky"]:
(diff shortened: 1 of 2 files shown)


dae:
I think the HTML changes are causing problems in other areas like the HTML editor (ctrl+shift+x) as well, and there may be other areas as well - I think it's probably safest to remove the whitespace from the .ts file as well.

dae:
(not all of it, just inside the field div)

Arthur-Milchior:
One more commit pushed, to crerect it

dae:
Thanks!

PR #527, opened 2020-03-24
Explode on bridge cmd

Arthur-Milchior:
My ultimate goal here is to do the following:
* have my add-on see that a blur action was sent
* execute the blur action (i.e. save the new field value)
* do some more actions. 

Currently, I can't really do that, appart by copying the code of blur in my add-on.
By putting the code of the blur action in another method, it'll let me simply call the method from another method.

As usual, I tried to put multiple commit so that each change are easy to check.

The last commit feels better to me, it ensure that commands with and without arguments are considered the same way. But my goal would be achieved even without this last commit

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -367,20 +367,16 @@ class Editor:
             # shutdown
             return
         # focus lost or key/button pressed?
-        type, remaining = cmd.split(":", 1)
-        if type == "blur":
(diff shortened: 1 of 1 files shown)


dae:
Happy to accept this once the merge is fixed

dae:
Formatting is failing

Arthur-Milchior:
Corrected

dae:
cheers

PR #530, opened 2020-03-27
Set to checks.yml install and import anki wheels

evandrocoan:
Just some basic testing to assure they are installing properly. Right away, the Anki wheels were not installing because the `setup.py` did not have `Qt` as a dependency. The Qt dependency was inside e `requirements.qt`, which was installed by `make pyenv`. As there was no need to use a `requirements.qt`, the `Qt` dependencies were moved into the `setup.py`. Now the Python Wheels are working fine installing all required dependencies.

Related to: https://anki.tenderapp.com/discussions/ankidesktop/40029-how-to-generate-anki-build-packages-for-linuxmac-oswindows

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -97,14 +97,21 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ${{ github.workspace }}${{ matrix.SEP }}pyenv
-          key: ${{ runner.os }}-pyenv-${{ hashFiles('**/requirements.*') }}-${{ hashFiles('**/setup.py') }}-${{ hashFiles('**/Makefile') }}-14-
+          key: |
+            ${{ runner.os }}-pyenv-
+            ${{ hashFiles('**/requirements.*') }}-
+            ${{ hashFiles('**/setup.py') }}-
+            ${{ hashFiles('**/Makefile') }}-14-
 
       - name: Cache pip wheels
         if: matrix.BUILD_TYPE == 'build'
         uses: actions/cache@v1
         with:
           path: ${{ matrix.PIP_WHEELS_DIR }}
-          key: ${{ runner.os }}-pip-wheels-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('**/setup.py') }}-14-
+          key: |
+            ${{ runner.os }}-pip-wheels-
+            ${{ hashFiles('**/requirements.txt') }}-
+            ${{ hashFiles('**/setup.py') }}-14-
 
       - name: Cache cargo index
         uses: actions/cache@v1
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -270,7 +270,7 @@ jobs:
           cd dist
           setlocal EnableDelayedExpansion
           cmd /C set "wheels=" && for /f "delims=" %%i in ('dir /b *.*') DO set "wheels="%%i" !wheels!"
-          python -m pip install %wheels%
+          python -m pip install pyqtwebengine %wheels%
           python -c "import aqt; # aqt.run()"
 
       - name: Check Linux/Mac OS wheels
@@ -278,7 +278,7 @@ jobs:
         run: |
           set -x
           cd dist
-          python -m pip install *.*
+          python -m pip install pyqtwebengine *.*
           python -c "import aqt; # aqt.run()"
 
       - run: make check
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -97,23 +97,14 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ${{ github.workspace }}${{ matrix.SEP }}pyenv
-          key: |
-            ${{ runner.os }}-pyenv-
-            ${{ hashFiles('**/requirements.*') }}-
-            ${{ hashFiles('**/setup.py') }}-
-            ${{ hashFiles('**/Makefile') }}-14-
+          key: ${{ runner.os }}-pyenv-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('**/setup.py') }}-${{ hashFiles('**/Makefile') }}-14-
 
-      # Disabling these caches for now because they do not seem to be used/help
-      # https://github.com/ankitects/anki/pull/528
-      # - name: Cache pip wheels
-      #   if: matrix.BUILD_TYPE == 'build'
-      #   uses: actions/cache@v1
-      #   with:
-      #     path: ${{ matrix.PIP_WHEELS_DIR }}
-      #     key: |
-      #       ${{ runner.os }}-pip-wheels-
-      #       ${{ hashFiles('**/requirements.txt') }}-
-      #       ${{ hashFiles('**/setup.py') }}-14-
+      - name: Cache pip wheels
+        if: matrix.BUILD_TYPE == 'build'
+        uses: actions/cache@v1
+        with:
+          path: ${{ matrix.PIP_WHEELS_DIR }}
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('**/setup.py') }}-14-
 
       - name: Cache cargo index
         uses: actions/cache@v1
@@ -237,6 +228,7 @@ jobs:
       - name: Set up Ubuntu ripgrep, pyaudio, gettext, rename
         if: matrix.os == 'ubuntu-latest'
         run: |
+          set -x
           sudo apt update
           sudo apt install portaudio19-dev gettext rename
           sudo snap install ripgrep --classic
@@ -244,6 +236,7 @@ jobs:
       - name: Set up brew ripgrep, pyaudio, gettext, rename
         if: matrix.os == 'macos-latest'
         run: |
+          set -x
           brew install portaudio protobuf gettext rename ripgrep make
           brew link gettext --force
 
@@ -263,16 +256,35 @@ jobs:
           node-version: 12
 
       - run: make develop
-        if: matrix.BUILD_TYPE != 'check'
+        if: matrix.BUILD_TYPE == 'build'
 
       - run: make build
-        if: matrix.BUILD_TYPE != 'check'
+        if: matrix.BUILD_TYPE == 'build'
+
+      - name: Check Windows wheels
+        if: matrix.BUILD_TYPE == 'build' && matrix.os == 'windows-latest'
+        shell: cmd
+        run: |
+          echo on
+          copy %pyaudio% dist
+          python -m pip install %pyaudio%
+          cd dist
+          cmd /C for /R %%i in (*.whl) DO python -m pip install "%%i"
+          python -c "import aqt; # aqt.run()"
+
+      - name: Check Linux/Mac OS wheels
+        if: matrix.BUILD_TYPE == 'build' && ( matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest' )
+        run: |
+          set -x
+          cd dist
+          python -m pip install *
+          python -c "import aqt; # aqt.run()"
 
       - run: make check
         if: matrix.BUILD_TYPE == 'check'
 
       - name: Upload python wheels
-        if: matrix.BUILD_TYPE != 'check'
+        if: matrix.BUILD_TYPE == 'build'
         uses: actions/upload-artifact@v1
         with:
           name: ${{ matrix.ANKI_PYTHON_WHEELS }}
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -97,7 +97,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ${{ github.workspace }}${{ matrix.SEP }}pyenv
-          key: ${{ runner.os }}-pyenv-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('**/setup.py') }}-${{ hashFiles('**/Makefile') }}-14-
+          key: ${{ runner.os }}-pyenv-${{ hashFiles('**/requirements.*') }}-${{ hashFiles('**/setup.py') }}-${{ hashFiles('**/Makefile') }}-14-
 
       - name: Cache pip wheels
         if: matrix.BUILD_TYPE == 'build'
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -104,7 +104,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ${{ matrix.PIP_WHEELS_DIR }}
-          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('**/setup.py') }}-14-
+          key: ${{ runner.os }}-pip-wheels-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('**/setup.py') }}-14-
 
       - name: Cache cargo index
         uses: actions/cache@v1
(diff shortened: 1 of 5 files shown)


dae:
aqt does not have pyqt dependencies listed because webengine was split into a separate package at one point. Depending on webengine will prevent installing with an earlier qt version, and not depending on webengine will prevent Anki from running on later Qt versions. It also doesn't make sense to hard-code a particular version when Anki can run on multiple different versions. So this code should be changed to install a particular version in advance for testing purposes, but not include the dependency in setup.py

evandrocoan:
If after reading this, you still would like for me to remove `pyqt/pyqtwebengine` from `setup.py`, of course, I will do it. Just say the word.

> Depending on webengine will prevent installing with an earlier qt version, and not depending on webengine will prevent Anki from running on later Qt versions.

If there are some restrictions about the version of webengine, `setup.py` should include the version range from which it requires instead of not installing a require dependency. What do you gain by letting the program crash when running due to missing dependency?

I did not know Anki could run in other versions of Qt, so I restored the original code to install specific version for development while letting `setup.py` to install any version of `webengine`/`PyQt` if no version is installed.

> So this code should be changed to install a particular version in advance for testing purposes, but not include the dependency in setup.py

If the user already has some version of `PyQt` or `webengine`, `setup.py` will not upgrade it (because we put no minimum version). But if the user has no version of `PyQt` at all, then `setup.py` will install the latest version it can find at the moment. This is no different from letting the user install the wheels, trying to run anki, and see it crashing because there is not `PyQt/pyqtwebengine` installed. The user will run `pip install PyQt pyqtwebengine` and it will install the latest version available at the moment. So why make trouble?

The problem with this is, how a nonprogrammer user will know it has to install PyQt? As programmers, we see some import error, we first curse the other developer who did not put the dependency requirement on `setup.py`, only then and we search about from where that dependency is coming from and install how to it. After that, we open a pull request adding the missing dependency to `setup.py`.

As a quoting from: https://stackoverflow.com/questions/43658870/requirements-txt-vs-setup-py
> The short answer is that `requirements.txt` is for listing package requirements only. `setup.py` on the other hand is more like an installation script. If you don't plan on installing the python code, typically you would only need `requirements.txt`. 
> 
> **requirements.txt**
> 
> This helps you to set up your development environment. Programs like `pip` can be used to install all packages listed in the file in one fell swoop. After that you can start developing your python script. Especially useful if you plan to have others contribute to the development or use virtual environments. This is how you use it:
> 
>     pip install -r requirements.txt
> 
> **setup.py**
> 
> This allows you to create packages that you can redistribute. This script is meant to install your package on the end user's system, not to prepare the development environment as `pip install -r requirements.txt` does. See [this answer][1] for more details on setup.py.
> 
> The dependencies of your project are listed in both files.
> 
>   [1]: https://stackoverflow.com/questions/1471994/what-is-setup-py
> 
> The file `setup.py` describes, in addition to the package dependencies, the set of files and modules that should be packaged (or compiled, in the case of native modules (i.e., written in C)), and metadata to add to the python package listings (e.g. package name, package version, package description, author, ...).
> 
> Because both files list dependencies, this can lead to a bit of duplication. Read below for details.
> 
> **requirements.txt**
> **********************
> 
> This file lists python package requirements. It is a plain text file (optionally with comments) that lists the package _dependencies_ of your python project (one per line). It _does not_ describe the way in which your python package is installed. You would generally consume the requirements file with `pip install -r requirements.txt`.
> 
> The filename of the text file is arbitrary, but is often `requirements.txt` by convention. When exploring source code repositories of other python packages, you might stumble on other names, such as `dev-dependencies.txt` or `dependencies-dev.txt`. Those serve the same purpose as `dependencies.txt` but generally list additional dependencies of interest to developers of the particular package, namely for testing the source code (e.g. pytest, pylint, etc.) before release. Users of the package generally wouldn't need the entire set of developer dependencies to run the package. 
> 
> If multiple`requirements-X.txt` variants are present, then usually one will list runtime dependencies, and the other build-time, or test dependencies. Some projects also cascade their requirements file, i.e. when one requirements file includes another file ([example](https://stackoverflow.com/a/20720019/5556676)). Doing so can reduce repetition.
> 
> **setup.py**
> *************
> 
> This is a python script which uses the `setuptools` module to define a python package (name, files included, package metadata, and installation). It will, like `requirements.txt`, also list runtime dependencies of the package. Setuptools is the de-facto way to build and install python packages, but it has its shortcomings, which over time have sprouted the development of new "meta-package managers", like pip. Example shortcomings of setuptools are its inability to install multiple versions of the same package, and lack of an uninstall command.
> 
> When a python user does `pip install ./pkgdir_my_module` (or `pip install my-module`), pip will run `setup.py` in the given directory (or module). Similarly, any module which has a `setup.py` can be `pip`-installed, e.g. by running `pip install .` from the same folder.
> 
> **Do I really need both?**
> ****************************
> 
> Short answer is no, but it's nice to have both. They achieve different purposes, but they can both be used to list your dependencies.
> 
> There is one trick you may consider to avoid duplicating your list of dependencies between `requirements.txt` and `setup.py`. If you have written a fully working `setup.py` for your package already, and your dependencies are mostly external, you could consider having a simple `requirements.txt` with only the following:
> 
>      # requirements.txt
>      #
>      # installs dependencies from ./setup.py, and the package itself,
>      # in editable mode
>      -e .
>      
>      # (the -e above is optional). you could also just install the package
>      # normally with just the line below (after uncommenting)
>      # .
> 
> The `-e` is a special `pip install` option which installs the given package in _editable_ mode. When `pip -r requirements.txt` is run on this file, pip will install your dependencies via the list in `./setup.py`. The editable option will place a symlink in your install directory (instead of an egg or archived copy). It allows developers to edit code in place from the repository without reinstalling.
> 
> You can also take advantage of what's called "setuptools extras" when you have both files in your package repository. You can define optional packages in setup.py under a custom category, and install those packages from just that category with pip:
> 
> ```python
> # setup.py
> from setuptools import setup
> setup(
>    name="FOO"
>    ...
>    extras_require = {
>        'dev': ['pylint'],
>        'build': ['requests']
>    }
>    ...
> )
> ```
> 
> and then, in the requirements file:
> ```
> # install packages in the [build] category, from setup.py
> # (path/to/mypkg is the directory where setup.py is)
> -e path/to/mypkg[build]
> ```
> 
> This would keep all your dependency lists inside setup.py. 
> 
> _Note_: You would normally execute pip and setup.py from a sandbox, such as those created with the program `virtualenv`. This will avoid installing python packages outside the context of your project's development environment.
>

agentydragon:
FYI, I'm interested in this PR, because I have a use-case where I want to use Anki Python modules outside Anki.

https://anki.tenderapp.com/discussions/ankidesktop/40025-anki-in-python-package-index

dae:
pyqtwebengine depends on the equivalent pyqt version. For older installs that don't have that package, it will overwrite the old Qt version (that may have taken hours to build) with the latest PyQt. Don't ask me how I know this. :-)

Had PyQt used pyqtwebengine from the start, it would have made sense to just list the supported versions in the requirements. But unless it is possible to say "pyqt < 5.12 || (pyqt >= 5.12 & qtwebengine >= 5.12), I think we'll need to leave the deps off.

evandrocoan:
> But unless it is possible to say "pyqt < 5.12 || (pyqt >= 5.12 & qtwebengine >= 5.12), I think we'll need to leave the deps off.

It should be possible. Just confirming before updating the code, you meant this scenario bellow?
```py
if pyqt is not installed:
    install pyqt and qtwebengine

else if pyqt < 5.12:

    if qtwebengine is not installed:
	    install qtwebengine

    else if qtwebengine < 12:
        you cannot install qtwebengine

else if pyqt >= 5.12:
    you can install pyqt and qtwebengine
```

dae:
if pyqt < 5.12, then qtwebengine must not be installed

dae:
- I think a wheel is more appropriate than an sdist, because what is being distributed is partially built (eg protobuf-generated .py files), not the original source code.
- The version checks need to be listed in the requirements so they are checked at install time instead of build time

evandrocoan:
> I think a wheel is more appropriate than an sdist, because what is being distributed is partially built (eg protobuf-generated .py files), not the original source code.

1. If you inspect the contents of both `sdist` and `bdist_wheel`, you will see they both have the same built files (including the protobuf-generated .py files).
    1. In our case, there is no difference when a user is installing a `sdist` of `aqt` or a `bdist_wheel` of `aqt`, except that `bdist_wheel` will not run `setup.py` while `sdist` will. Meaning that we can use a `sdist` to create the dynamic dependencies of `pyqt` and `webengine`.
    1. The files on `aqt` and `pylib` are transpired (https://stackoverflow.com/questions/44931479/compiling-vs-transpiling). Meaning they are not dependent on a particular processor (intel IA32, x86_x64, arm, etc). They are just transformed from one high level language to another high level language. As already said, both `sdist` and `bdist_wheel` have the same files, which includes the protobuf-generated .py files.
    1. So, they can be used as `sdist`. What cannot be used as a `sdist` are the `rslib` or `rspy`, because they are built by cargo into a architecture dependent code (binary for intel IA32, x86_x64, arm, etc). So, if a `sdist` of `rslib` or `rspy` were created, the user which ran `pip install rslib` would be required to have the `cargo/rust` compilers installed on their system.
1. This changes only affects `aqt` which do not use `protofuf` (pylib uses it).
    1. Even if it is used, it would not be problem because although the files are "built", they are not binary while both `sdist` and `bdist_wheel` have the same generated files.


> The version checks need to be listed in the requirements so they are checked at install time instead of build time

If pyqt is listed in the requirements, we cannot enforce pyqt to dependent on webengine because `bdist_wheel` is just a dumb zip file with a list of static dependencies. As already said, on this particular case, the only difference of using `bdist_wheel` over `sdist` is that:
1. The `bdist_wheel` is going to read a static list of dependencies and install them.
1. Unzip the contents of the wheel inside the user site packages.

While using `sdist`:
1. The `sdist` is doing to run `setup.py` and install the list of dependencies dynamically created.
1. Unzip the contents of the file inside the user site packages (the contents unzipped by the `sdist` build or the `bdist_wheel` are the same).

A reason to use `sdist` over `bdist_wheel` is that we could allow it to be installed in any processor architecture anki supports, while we have create a `bdist_wheel` for each operating system, processor architecture.

A reason to not use `sdist` over `bdist_wheel` is if the `sdist` would require the user to have some third part tool installed as `C++` compiler, which is not the case. Our `sdist` is only going to install missing dependencies and unzip the contents of the file. The same thing the `bdist_wheel` is going to do, except that the `bdist_wheel` can only have a dumb static list of dependencies, while `sdist` wheel can dynamically select the dependencies (because is runs `setup.py`).

Another example of `sdist` versus `bdist_wheel` is the anki dependency `pyaudio`. The default distribution of it on `pypi` uses a `sdist` file. This is why it requires the `pylibaudio.h` header and a `C++` compiler to be installed on the user. Although we can also download `bdist_wheel` from https://www.lfd.uci.edu/~gohlke/pythonlibs/#pyaudio so, we do not have to have a `C++` compiler in order to use it.

The same does not applies here for `aqt`. Either using a `sdist` or a `bdist_wheel`, we are not required to have a `C++` or `Rust` compiler, a protobuf compiler or any third part tool installed in order to use any of them. So, when the `sdist setup.py` file runs, it is only going to install dependencies and unzip the files (with the bonus of dynamically calculating the list of dependencies).

In order to implement the https://anki.tenderapp.com/discussions/ankidesktop/40025-anki-in-python-package-index, I am going to setup `aqt` and `pylib` be uploaded to `pypi` as `sdist`, while `rslib` and `rspy` are going to (have to) be uploaded as `bdist_wheel`. But as an optimization, I am going to see if I can also upload the `rspy` and `rslib` as `sdist` too, just in case someone tries to install them for a system/architecture not yet supported by the `bdist_wheel` distributions of `rslib` and `rspy`.

So, both `rslib` and `rspy` are going to have `sdist` and `bdist_wheel`, where the `sdist` is only a fall back in case there is not a `bdist_wheel` available for your processor architecture/operating system. Then, if the user have a cargo/rust compiler installed, everything will be built automatically, instead of failing the installation. If the user does not had the compiler, the installation would be failing anyways because there was no `bdist_wheel` available for this platform. Now, even after failing the installation, the user could install the cargo compiler instead of asking us for a special `bdist_wheel` for his processor architecture/operating system.

This puts us on the same problem `pyaudio` dependency create for us. The `pypi` repository for `pyaudio` does not have any `bdist_wheel` for us. It only has `sdist` files, forcing us to have an `C++` compiler. Then, if anyone tries to install anki from `pypi`, they are already forced to either have a `C++` compiler to build `pyaudio`, or manually download some `bdist_wheel` from https://www.lfd.uci.edu/~gohlke/pythonlibs/#pyaudio.

The only solution for this `pyaudio` problem would be creating anki wheels which embed all its dependencies (including pyqt, webengine, pyaudio, beautifulsoup4, send2trash, etc). On this case, these wheels must to be `bdist_wheel` because they are now having binary/compiled files for specific processor architectures/operating systems. But this is what you already do when you create anki releases (https://anki.tenderapp.com/discussions/ankidesktop/40029-how-to-generate-anki-build-packages-for-linuxmac-oswindows). So, there is no need to redo this work for `pypi` releases, just let the dependencies to be downloaded as the are usually done. Anyone installing anki from `pypi` will have to deal this `pyaudio` problem of requiring a `C++` compiler. And `aqt` being a `sdist` will not interfere with their installation by any means as `pyaudio` does.

dae:
Evandro, I understand the problem you are trying to solve, I'm just not keen on introducing source dists at this time. If you look at the existing generated wheels (eg aqt-2.1.24+d3964603-py3-none-any.whl), you'll see they are architecture and platform independent (https://packaging.python.org/guides/distributing-packages-using-setuptools/#pure-python-wheels). Distributing the files as a source dist implies that Anki can be built from scratch using those files, which is not the case, and I don't want to deal with testing and possible regressions of packaging changes at the moment. I think you can achieve 90% of the same result with minimal changes by just requiring pyqt>=5.9, and adding a helpful print() message into qt.py if qtwebengine fails to import.

dae:
Any ideas why the Mac builds are failing at the moment?

dae:
Nevermind - looks like the recent cache changes you made have fixed it, thanks.

evandrocoan:
Okay. Now it is time for that old bug to hunt us:
```
        assert c.left % 1000 == 3
>       assert c.left // 1000 == 1
E       AssertionError: assert (2003 // 1000) == 1
E        +  where 2003 = {'data': '', 'did': 1, 'due': 1585873047, 'factor': 0, 
            'flags': 0, 'id': 1585879647548, 'ivl': 0, 'lapses': 0, 'left':...447,
            'nid': 1585872447547, 'odid': 0, 'odue': 0, 'ord': 0, 'queue': 1, 'reps': 1,
            'type': 1, 'usn': -1, 'wasNew': True}.left
```

This only happens between 02:00 and 04:00 UTC.

The time seems to be mostly assigned by these functions:
```python
anki/pylib/anki/sched.py
403:     def _startingLeft(self, card: Card) -> int:
404:         if card.type == CARD_TYPE_REV:
405:             conf = self._lapseConf(card)
406:         else:
407:             conf = self._lrnConf(card)
408:         tot = len(conf["delays"])
409:         tod = self._leftToday(conf["delays"], tot)
410:         return tot + tod * 1000

anki/pylib/anki/schedv2.py ( called by _startingLeft() )
755:     def _leftToday(
756:         self, delays: List[int], left: int, now: Optional[int] = None,
757:     ) -> int:
758:         "The number of steps that can be completed by the day cutoff."
759:         if not now:
760:             now = intTime()
761:         delays = delays[-left:]
762:         ok = 0
763:         for i in range(len(delays)):
764:             now += int(delays[i] * 60)
765:             if now > self.dayCutoff:
766:                 break
767:             ok = i
768:         return ok + 1
	```

dae:
Thanks Evandro!

PR #531, opened 2020-03-27
hook browser_header_will_show_context_menu

Arthur-Milchior:
Advanced browser adds elements in the browser's header's context menu. I believe that hooks would be better here. And I would add it to one of my own add-on

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -970,6 +970,7 @@ QTableView {{ gridline-color: {grid} }}
             a.setCheckable(True)
             a.setChecked(type in self.model.activeCols)
             a.toggled.connect(lambda b, t=type: self.toggleField(t))
+        gui_hooks.browser_header_will_show_context_menu(self, m)
         m.exec_(gpos)
 
     def toggleField(self, type):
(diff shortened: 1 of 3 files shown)


dae:
Cheers

Also a heads up I had to roll back another recent PR due to a bug

evandrocoan:
What do you think about adding Integration Tests to anki? i.e., A script which installs and opens anki, and uses it like it was a user. It would do pre-determined actions as:
1. Opening the add screen, clicking on Type, and choosing a note type. Catch bug (https://github.com/ankitects/anki/commit/47bd6264bdd03b835eba2adf4c52ffa66735d7d0)
1. Create a decks and add a few notes.
1. Create a new note type and use it.
1. etc.

The only problem of those tests is that they take much more time as we have to wait for anki open, to wait the UI to respond to our actions, etc. Today a full build and test run of the GitHub Actions takes about 10 minutes to complete everything. This integration test could take much more, depending on how many tests there are.

Arthur-Milchior:
You mean, something as http://www.milchior.fr/blog_en/index.php/post/2019/07/29/How-I-automated-the-test-of-anki-s-GUI?

Arthur-Milchior:
It seems unrelated to this PR by the way

evandrocoan:
Something like that, but automated, i.e., another program opens and uses anki as it was a real person. I just commented this here because of the bug caught while running anki.

dae:
Evandro: I'm not sure it would be practical to deploy to CI, but might be useful as a tool that could be run locally before a release. But there are so many other things that need doing as well... :-)

PR #532, opened 2020-03-29
Previewer

Arthur-Milchior:
The previewer is quite practical. However, it can hardly be used in add-ons right now. So add-ons such as https://ankiweb.net/shared/info/1423933177 by @ijgnd basically needs to extract the previewer code and recreate them in the add-on.
I feel it makes things uselessly complex and it would be far better to have access to the previewer directly from add-on.

I'm not stating that this should be THE way to implement this feature, if you accept it. But I prefer to show some code as a starting point, which can always be changed if anyone has improvements for it.

The first commit simply create a class for the previewer, and ensures the browser uses it. Since the previewer has a lot of features which are related specifically to the browser, I then split the class into two classes. The parent class is the previewer which can be used anywhere, and the subclass deals with the features which are related to the browser. Finally, I added a third class, which is a previewer specially created to display a single card; i.e. exactly what the above-mentioned add-on needs (and what an add-on I'm working on needs too)

### DIFF
--- a/qt/aqt/previewer.py
+++ b/qt/aqt/previewer.py
@@ -2,7 +2,7 @@ import json
 import re
 import time
 from dataclasses import dataclass
-from typing import Any, Optional, Union
+from typing import Any, List, Optional, Union
 
 from anki.cards import Card
 from anki.lang import _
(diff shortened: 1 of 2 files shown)


dae:
It would be nice to move that code out into a separate module, though I imagine it will break some add-ons. 

@glutanimate any thoughts?

Arthur-Milchior:
The last commit also create a class which allows to show a fixed list of card (this list can always be changed by changing the variable if an add-on want to).

Arthur-Milchior:
Sorry if it spammed you (I don't know which message you get). I wasn't able to run make check until someone helped me on tenderapp. It know should works finally. And the typer didn't like abstract function

glutanimate:
Looks good to me, but let me take a closer look later today. Currently a bit short on time.

Arthur-Milchior:
I can answer most of your question by the same answer. I know that minimal changes are preferred. I know that's strange to write in a PR adding +397 lines. But still, I tried to be semantically as minimal as possible.
I tried to ensure that, when you click on "preview" in the browser, exactly the same path in the code will be used (except for the call to super().) So I didn't inherit from a Qt class because originally there was no inheritance. I Kept `PreviewDialog` because that's the way it was in original code.
I kept method's name to be consistent. Even if I agree that making them public would have sens.

If @dae you agree with Glutanimate's request I'll be happy to implement them

lovac42:
Dozens of addons are already broken, what's a few more. You might as well clean it up properly now and be done with it rather than taking shortcuts and having to rewriting this a year or two down the line.

Arthur-Milchior:
I made all changes suggested by Glutanimate in the most recent commits

qt/aqt/previewer.py
```diff
@@ -0,0 +1,371 @@
...
+
+    def _on_other_side(self):
+        if self._state == "question":
+            self._state = "answer"
+        else:
+            self._state = "question"
+        self.render()
+
+
+class PreviewerSingleCard(Previewer):
```
dae, line 347: SingleCardPreviewer, BrowserPreviewer & MultipleCardPreviewer would read a bit more naturally.

dae:
Thanks Arthur, I gave it a quick look over and nothing jumped out at me. Since it's a breaking change it might be nice to move the methods and instance vars to snake_case at the same time, if you feel like doing that.

Arthur-Milchior:
I'll do it

qt/aqt/previewer.py
Arthur-Milchior: Changed

dae:
Thanks for your hard work Arthur!

glutanimate:
Thanks, Arthur! Really nicely done! I went ahead and notified the authors of the potentially affected add-ons (wanted to avoid what happened with MorphMan with my toolbar commit recently).

Arthur-Milchior:
No idea what you mean about morphman. Thanks for doing it

PR #533, opened 2020-03-30
Use double quotes in taglimit queries

zjosua:
Is it even necessary to quote tags?
I think it's not possible to create tags containing spaces. Unless maybe through an add-on or the debug console.


dae:
Quotes are required, for example if the tag has a ) in it. Thanks for the fix!

PR #537, opened 2020-04-03
hook add_card_did_init

Arthur-Milchior:
I'm creating an add-on which set some default value for add card. I can make it using editor_did_init, but add_card_did_init seems more natural


Arthur-Milchior:
Actually, editor_did_init does not work. Because the note is not yet set when editor_did_init is called, while it is set when add_card_did_init is called. But this difference was quite unexpected when I created the PR

dae:
cheers!

PR #538, opened 2020-04-03
Deprecated maturin alternative version fix as version 0.8.0 was

evandrocoan:
Version 0.8.0 was just released within it: 
1. https://github.com/pyo3/maturin/releases
1. https://github.com/PyO3/maturin/issues/289 - Stop passing author.email to python
1. https://github.com/PyO3/maturin/pull/250 - 32 bit Python modules use i386, not x86


dae:
Thanks Evandro, I've made this change and added a version requirement as well.

PR #539, opened 2020-04-05
Remove useless recursion in DeckManager.rem

Arthur-Milchior:
For any deck the children of it's children are its children. So applying rem to children of children is useless and actually slightly costly for deck with deep subdecks. 

I should note also that "def count" counts the default deck even if it is not listed. This is not a problem actually since this method is never called.


dae:
cheers

PR #540, opened 2020-04-06
Added arguments as [sound:filename|argument1 argument2...]

evandrocoan:
Replaces: https://github.com/ankitects/anki/pull/498 Fix media files being marked as unused

Now, we can use the `[sound:file.mp3]` passing arguments to it as `[sound:file.mp3|arg1 arg2]`. For example:
1. I could pass an argument as `[sound:file.mp3|skipautoplay]` to anki does not play it automatically.
1. I can call `mplayer -af scaletempo -speed 0.5 filename.mp3` to play a file with `0.5` speed. The `-af scaletempo` is to `mplayer` do not distort the [audio pitch](https://unix.stackexchange.com/questions/387803/programmatically-change-the-playback-speed-not-pitch-in-real-time), while playing it.

I just do not know how I could use/access the arguments and pass them to `mplayer`. Any suggestions?

Created the argument 'fileonly' to sound, stopping anki from inserting and interpreting the sound tags. If the argument `fileonly` is used as `[sound:file.mp3|fileonly]`, anki will skip processing that file while rendering the card and will replace the `[sound:file.mp3|fileonly]` directly by the sound name `file.mp3`. This allows to completely render the audio by JavaScript, while keeping anki Unused Files checking fully working.
```
// Front
<input type="button" value="♬" onclick="playsound('{{Audio}}', 1.5);" />

    var audio;
    function playsound(filename, speed=1.5) {
        if( audio !== undefined) { audio.pause(); }
        audio = new Audio(filename);
        audio.playbackRate = speed;
        audio.play();
    }
    if( !document.getElementById("answer") ) {
         playsound('{{Audio}}', 1.5);
    }
</script>

// Back
    playsound('{{Audio}}', 1.5);
</script>
```

### DIFF
--- a/pylib/anki/media.py
+++ b/pylib/anki/media.py
@@ -36,7 +36,7 @@ CheckMediaResponse = media_pb2.CheckMediaResponse
 
 class MediaManager(DeprecatedNamesMixin):
 
-    sound_regexps = [r"(?i)(\[sound:(?P<fname>[^]]+)\])"]
+    sound_regexps = [r"(?i)(\[sound:(?P<fname>[^\]|]+)(?:\|(?P<args>[^\]]+))?\])"]
     html_media_regexps = [
         # src element quoted case
         r"(?i)(<[img|audio][^>]* src=(?P<str>[\"'])(?P<fname>[^>]+?)(?P=str)[^>]*>)",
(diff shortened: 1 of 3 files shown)


dae:
It may take me a while to get back to you about this one, as I don't have time to think it through at the moment. FWIW, the code currently avoids named match groups, as they're not as efficient as number-based captures.

aleksejrs:
How about the "run" command?

aleksejrs:
Why not "linkonly" or "nameonly"? "fileonly" makes no sense to me: maybe it's about not playing directories, or about avoiding deletion by Check Media.

evandrocoan:
Replaced by: https://github.com/ankitects/anki/pull/2323 - Create field options to store media file path

PR #541, opened 2020-04-06
detect renamed cloze fields in note type preview

BlueGreenMagick:
Detect cloze fields such as
{{cloze:Text2}}
{{cloze:hint:Text}}
{{edit:cloze:Text}}

I assumed that only '}' and ':' were not allowed as field name. Would that be alright?
Another thing I'm not sure about: should the existence of cloze fields be checked in the answer template as well?


dae:
It's probably fine for now, thanks. In the future the previewing code is likely to be updated to leverage the template parsing code, but some more work is required to get to that point.

PR #542, opened 2020-04-06
detect cloze cards with other filters

BlueGreenMagick:
for prompting user if should create card without cloze deletion.

Such as {{hint:cloze:Field}} or a custom filter like {{edit:cloze:Field}} (which is used in edit fields during review addon)


dae:
This code should probably just be checking model["type"] instead of searching in the template

BlueGreenMagick:
Done. Is it possible to view or change model type of an existing note type?

dae:
Thanks for that. 

Anki doesn't support changing the type after creation, so we should be able to rely on it

PR #543, opened 2020-04-06
Sort according to path

Arthur-Milchior:
Here is a bug and it's correction.
2020-04-06-231008_130x100_scrot
On the left, the deck browser. On the right, the list of deck (in wrong order)
2020-04-06-224729_131x111_scrot
On the right, the list of deck in correct order.

The error was that the keys used for sorting was the string, and not the path. :: was sorted between 9 and A.

The first commit allows _path to easily be used outside of this class. The second uses the correct key everywhere. You may want to merge those two and ignore the remainings, if you prefer short commits.
The other ones simply uses methods everywhere where it is possible, in order to avoid code redundancy and to increase clarity. It makes methods static, so that they can be used easily outside of the deck manager.

One reason why it would help me to have the first two commits merged is that ankidroid is trying to be as close as possible to Anki. I'm trying to increase ankidroid speed and codebase and it would be so much simple for me if I just could ensure that a single sorting function were used everywhere (because, yes, currently, ankidroid has two sorting functions for decks and at first, I didn't understand why)

### DIFF
--- a/pylib/anki/decks.py
+++ b/pylib/anki/decks.py
@@ -126,7 +126,7 @@ class DeckManager:
             # child of an existing deck then it needs to be renamed
             deck = self.get(did)
             if "::" in deck["name"]:
-                base = deck["name"].split("::")[-1]
+                base = self._path(deck["name"])[-1]
                 suffix = ""
                 while True:
                     # find an unused name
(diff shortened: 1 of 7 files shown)


pylib/anki/decks.py
```diff
@@ -283,24 +282,44 @@ def _canDragAndDrop(self, draggedDeckName: str, ontoDeckName: str) -> bool:
...
+        return ancestorPath == self.path(descendantDeckName)[0 : len(ancestorPath)]
 
-    def _path(self, name: str) -> Any:
+    @staticmethod
+    def path(name: str) -> Any:
         return name.split("::")
 
-    def _basename(self, name: str) -> Any:
-        return self._path(name)[-1]
+    _path = path
```
c-okelly, line 297: What is the rationality behind having both `path` and `_path`

dae: preserving compatibility with code using the old style

dae:
Looks good, thanks.

PR #544, opened 2020-04-07
strip HTML when displaying deck names in non-web

BlueGreenMagick:
This PR will let users have html tags in deck names, and not get those html tags to show up in deck chooser and in browser.

Potential use cases are to bold/italicize deck names in the overview, or to put a d-day countdown script.


dae:
I'm afraid the only reason HTML tags in deck names currently render in the deck browser is due to oversight - in the future they will likely be shown as text instead.

AnKingMed:
If it means anything, I personally like being able to have HTML in the deck names. I can provide an example of why if there's a possibility to keep it moving forward

dae:
Yep, I've seen you taking advantage of it :-) But I'm afraid this was unintended, and is likely to be removed, so I recommend you avoid using it in future editions of your decks.

PR #545, opened 2020-04-07
allow dragging fields to change its position

BlueGreenMagick:
It's much faster than using position button and manually typing in numbers.


dae:
Thanks!

dae:
Mind looking into this?

https://anki.tenderapp.com/discussions/ankidesktop/41822-drag-dropping-fields-in-the-field-list-dialog-off-by-one

PR #548, opened 2020-04-09
Documented ANKIDEV environment variable on README.development

evandrocoan:


dae:
It also changes syncing to a server running on localhost, so it's not useful for most people in its current state.

dae:
That seems a bit more complicated that necessary. I wasn't using it anyway, so I've just stripped it out. Please note that ANKIDEV also disables automatic backups: https://github.com/ankitects/anki/commit/330848abe36e1fe423d3ae3c03e95d3147aefd2f

evandrocoan:
I saw it, but I backup my whole anki directory every day with this: https://github.com/evandrocoan/dotfiles/blob/18b2db70c34f6f5a5cce6f598d8d63f768284d93/scripts/backupanki.sh, and if it fails somehow I should get a big error pop up. It automatically runs with a Windows Task Scheduler task.

dae:
So you're all set, but we still need to make others aware of the consequences of using that option :-)

PR #549, opened 2020-04-10
Allow changing height of debug console text edit boxes 

BlueGreenMagick:
Two textedits were put inside a splitter so users can change their height. (Especially the one for writing code)


dae:
cheers

PR #550, opened 2020-04-10
add clear actions to context menu in debug console

BlueGreenMagick:
Adds "Clear Log" and "Clear Code" context menu to the debug console, because it's really not obvious that the hotkeys exist.


dae:
New strings should use fluent rather than _(). This is sufficiently niche that's it's probably easier to just drop the _() and not bother to make it translatable.

PR #551, opened 2020-04-10
Include self

Arthur-Milchior:
The last commit is related to https://github.com/ankitects/anki/pull/543 ; there was one place where I missed that sorting should not use alphabetical order: when you select in which order you review cards of subdecks.

The other two commits does not change any feature. As an add-on dev', it's a simple change that would sometime be easier to use and leads to more readable code according to my taste. Often, when we compute the list of parents/children of a deck D, we add D to the result. This adds an option to automatically adds it in the returned result. 
As a mathematician, I would simply states that I am allowing replacing "less than" by "less than or equal to" in the "parent" order

### DIFF
--- a/pylib/anki/decks.py
+++ b/pylib/anki/decks.py
@@ -601,11 +601,18 @@ class DeckManager:
 
         return childMap
 
-    def parents(self, did: int, nameMap: Optional[Any] = None) -> List:
+    def parents(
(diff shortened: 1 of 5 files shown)


dae:
I'm afraid I think it's confusing to have parents() and children() return anything except for parents and children. Please hold off on any further changes to the deck handling code, as this module will be getting a refactor soon.

Arthur-Milchior:
Well noted.

PR #552, opened 2020-04-11
Fix caches always missing

evandrocoan:
GitHub Actions does not seem to like/support multiline caches, then, some caches seem to be always missing.


dae:
Thanks! I will look into that day rollover bug at one point, just haven't been able to find them time.

evandrocoan:
No problem. I also should look into it, but first I am looking into other things I think are more important.

PR #553, opened 2020-04-11
Improve fill lrn

Arthur-Milchior:
Either I am missing something fundamental, or there is a really simple speed improvement to Anki. Sorting in queries instead of sorting in python.

I believe I may miss something, because I have no clue why the comment states that cards are sorted according to their did, which does not seems to be required at all by the current code.

To be clear, the reason why I care about this PR is that if you accept it, then I can do the same PR in AnkiDroid; and AnkiDroid loose a lot ot time sorting this list; so this would be a noticeable improvement for me. (It spends actually more than one second on my phone and collection just to sort the _lrnQueue)

Note that there is a difference when the query returns more than self.reportLimit cards (i.e. 1000 cards in learning mode in a deck), since the set of cards won't be the same. But it means that the in this case, my PR will gives back cards with the smallest due instead of random cards in learning mode and already due. In practice, I hope noone has 1000 cards in learning mode, so that it won't change anything for anyone

### DIFF
--- a/pylib/anki/sched.py
+++ b/pylib/anki/sched.py
@@ -281,14 +281,13 @@ and due <= ? limit %d"""
             f"""
 select due, id from cards where
 did in %s and queue = {QUEUE_TYPE_LRN} and due < ?
+order by due ASC, id ASC
 limit %d"""
             % (self._deckLimit(), self.reportLimit),
             self.dayCutoff,
         )
         for i in range(len(self._lrnQueue)):
             self._lrnQueue[i] = (self._lrnQueue[i][0], self._lrnQueue[i][1])
-        # as it arrives sorted by did first, we need to sort it
-        self._lrnQueue.sort()
         return self._lrnQueue
 
     def _getLrnCard(self, collapse: bool = False) -> Optional[Card]:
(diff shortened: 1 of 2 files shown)


dae:
They arrive in did order because sqlite uses the did index. Your change requires all results to be collected prior to sorting, so it could potentially be worse than before if there are a lot of due learning cards. Did you benchmark the before and after of this change in Python, and test when there are thousands of due learning cards?

Arthur-Milchior:
I do recognize I was entirely wrong about the comment. I'm surprised and need to understand why it's sorted by did and not by the standard order of the table. 

I just created a benchmark with 1000 cards in learning in 1000 decks, thanks to debugger and
```python
from aqt import mw
from anki.consts import *

for i in range(1000):
  did = mw.col.decks.id(f"top::{i}")
  mw.col.decks.select(did)
  f = mw.col.newNote()
  f["Front"] = str(i)
  f["Back"] = str(i)
  mw.col.addNote(f)
  c = f.cards()[0]
  c.did = did
  c.startTimer()
  mw.col.sched.answerCard(c, 2)
  c.flush()
```
Then I select the deck "top"
Using both the previous and the current code, it takes between 6 and 8 miliseconds to run both sorts.

PR #554, opened 2020-04-12
Fixed HTML5 media tags

evandrocoan:
I was trying to use HTML5 audio tag to display audios like:
```html
<audio id="elem_audio" src="myfile.mp3" controls></audio>
```
image

But the progress bar seek was not working. After researching, I found the problem was the HTML server not properly responding to the HTML5 header requests. The HTML server should respond to quite complicated things as 206 partial, properly handle keep-alive, provide media ranges and other HTTP headers: https://stackoverflow.com/questions/37044064/html-audio-cant-set-currenttime

To implement all these on the Simple HTTP server would be quite complicated. Then, instead, I imported the `flask` web server, which is quite simple and straight forward to use. Now, the back-end is using a secure complaint HTTP back-end:
1. https://palletsprojects.com/p/flask/
    > Flask is a lightweight WSGI web application framework. It is designed to make getting started quick and easy, with the ability to scale up to complex applications. It began as a simple wrapper around Werkzeug and Jinja and has become one of the most popular Python web application frameworks.
    > 
    > Flask offers suggestions, but doesn't enforce any dependencies or project layout. It is up to the developer to choose the tools and libraries they want to use. There are many extensions provided by the community that make adding new functionality easy.

1. https://docs.pylonsproject.org/projects/waitress/en/latest/
    > Waitress is meant to be a production-quality pure-Python WSGI server with very acceptable performance. It has no dependencies except ones which live in the Python standard library. It runs on CPython on Unix and Windows under Python 2.7+ and Python 3.5+. It is also known to run on PyPy 1.6.0 on UNIX. It supports HTTP/1.0 and HTTP/1.1.

Right now, anki does not support fields passing file names directly to HTML audio tags, but this can be easily done with (https://github.com/ankitects/anki/pull/540 - Added arguments to the sound tag) plus the commit https://github.com/evandroforks/anki/commit/826a97df61b99814041c41c0f2c84268280ed8ad, the HTML5 audio tag can be used like this:
```html
// Audio = [sound:myfile.mp3|onlyfilename]
<audio id="elem_audio" src="{{Audio}}" controls controlsList="nodownload"></audio>
```
image

### DIFF
--- a/qt/aqt/mediasrv.py
+++ b/qt/aqt/mediasrv.py
@@ -93,6 +93,18 @@ def allroutes(path):
             mimetype="text/plain",
         )
 
+    directory = os.path.realpath(directory)
+    path = os.path.normpath(path)
+    fullpath = os.path.realpath(os.path.join(directory, path))
+
+    # protect against directory transversal: https://security.openstack.org/guidelines/dg_using-file-paths.html
+    if not fullpath.startswith(directory):
+        return flask.Response(
+            "Path for '%s - %s' is a security leak!" % (directory, path),
+            status=HTTPStatus.FORBIDDEN,
+            mimetype="text/plain",
+        )
+
     if isdir:
         return flask.Response(
             "Path for '%s - %s' is a directory (not supported)!" % (directory, path),
@@ -103,6 +115,9 @@ def allroutes(path):
     try:
         if devMode:
             print("Sending file '%s - %s'" % (directory, path))
+
+        path = os.path.basename(fullpath)
+        directory = os.path.dirname(fullpath)
         return flask.send_from_directory(directory, path)
 
     except Exception as error:
--- a/qt/aqt/mediasrv.py
+++ b/qt/aqt/mediasrv.py
@@ -116,9 +116,7 @@ def allroutes(path):
         if devMode:
             print("Sending file '%s - %s'" % (directory, path))
 
-        path = os.path.basename(fullpath)
-        directory = os.path.dirname(fullpath)
-        return flask.send_from_directory(directory, path)
+        return flask.send_file(fullpath, conditional=True)
 
     except Exception as error:
         if devMode:
--- a/qt/aqt/mediasrv.py
+++ b/qt/aqt/mediasrv.py
@@ -126,13 +126,6 @@ def allroutes(pathin):
 
     try:
         if flask.request.method == "POST":
-            if not pathin.startswith("_anki/"):
-                return flask.Response(
-                    "Path for '%s - %s' is a security leak!" % (directory, path),
-                    status=HTTPStatus.FORBIDDEN,
-                    mimetype="text/plain",
-                )
-
             if path == "graphData":
                 body = request.data
                 data = graph_data(allroutes.mw.col, **from_json_bytes(body))
@@ -140,7 +133,7 @@ def allroutes(pathin):
                 data = allroutes.mw.col.backend.i18n_resources()
             else:
                 return flask.Response(
-                    "Path for '%s - %s' is a security leak!" % (directory, path),
+                    "Post request to '%s - %s' is a security leak!" % (directory, path),
                     status=HTTPStatus.FORBIDDEN,
                     mimetype="text/plain",
                 )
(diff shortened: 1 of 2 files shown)


dae:
Thank you for the contribution, the code itself looks good. It took quite a while for various issues in the current HTTP server code to be ironed out, and by switching it to something else there is a risk that new bugs will be introduced. I currently have my hands full with refactoring work and don't want to have to deal with extra issues, so I'm afraid I do not think I will be able to merge this in the near future, but I will leave it open to return to at a later date.

evandrocoan:
No problem. If you have/find some problem with this after/before you merge this, just ping and I will look into it. When you merge this I do not think you are going to have any problems because both `flask` and `waitress` are production frameworks. Meaning they are used by countless servers/websites. Then, if they had some problem, it should have been reported and fixed a long time ago. Until you look into this, I will be daily using it here.

qt/aqt/mediasrv.py
```diff
@@ -28,146 +31,127 @@ def _getExportFolder():
...
-            try:
-                addMgr = self.mw.addonManager
-            except AttributeError:
-                return path
 
-            addonPath = path[len(targetPath) :]
+def _redirectWebExports(path):
+    # catch /_anki references and rewrite them to web export folder
+    targetPath = "_anki/"
+    if path.startswith(targetPath):
```
dae, line 128: Is path already normalized? I'm wondering what happens in the case of /_anki/../../some/filesystem/path

qt/aqt/mediasrv.py
```diff
@@ -28,146 +31,127 @@ def _getExportFolder():
...
-        return path
+        pattern = addMgr.getWebExports(addon)
+        if not pattern:
+            return addMgr.addonsFolder(), path
 
+        if re.fullmatch(pattern, subPath):
+            return addMgr.addonsFolder(), addonPath
 
-# work around Windows machines with incorrect mime type
-RequestHandler.extensions_map[".css"] = "text/css"
```
dae, line 173: Does flask use a different mime type map? If it's using the same system settings then it's likely to suffer from the same issue.

evandrocoan: I just checked with `QTWEBENGINE_REMOTE_DEBUGGING=8080` and the response has the correct type:

image

qt/aqt/mediasrv.py
evandrocoan: I tried doing this, but paths as `/../something` entries are converted to `/something`. I run this code on the `QTWEBENGINE_REMOTE_DEBUGGING=8080` console:
```
var request = new XMLHttpRequest();
request.open("GET", "/../deleted.txt", true);
request.send();
```

The result on the server was:
```
Sending file 'D:\User\Documents\Anki2\Evandro\collection.media - deleted.txt'
Caught HTTP server exception,
  File "F:\anki\qt\aqt\mediasrv.py", line 106, in allroutes
    return flask.send_from_directory(directory, path)
  File "F:\anki\pyenv\lib\site-packages\flask\helpers.py", line 767, in send_from_directory
    raise NotFound()
werkzeug.exceptions.NotFound: 404 Not Found: The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.
```

Also, the function `flask.send_from_directory` used, states that it is a safe function as long the base directory is correct. In our case, it will send anything inside `addMgr.addonsFolder()`, `_redirectWebExports.mw.col.media.dir()` and `_exportFolder`.

evandrocoan:
After thinking about how flask `send_from_directory` function works, I noticed I was not correctly using it, as it looks like the function expects only to server files in a directory (but not files in subdirectories). Then, I added an update ensure directory transversal is not allowed outside the base directory the `_redirectWebExports` function gives back.

qt/aqt/mediasrv.py
dae: The issue is for machines where the configured mime type is wrong:

https://anki.tenderapp.com/discussions/beta-testing/725-anki-210-beta-13-14-15

dae:
Thanks Evandro, let's give it a go. It looks like flask use mimetypes to get the mime type, so a fix will still be required for CSS.

evandrocoan:
You are welcome. Any problem with this just ping and I will look into it.

PR #555, opened 2020-04-13
apply color to reviewer buttons

BlueGreenMagick:
All other Anki clients have colorful review buttons, It's time desktop had them too. I reused the colors used for displaying due card count in deckbrowser.

Here is how it looks like:
colorful-anki-night

colorful-anki-day

As for addon compatibility, it works with atleast 4 of the more popular addons that modify answer buttons: 'The KING of Button Add-ons', '[Large and Colorful Buttons'](https://ankiweb.net/shared/info/1829090218), '[RememorizeButtons](https://github.com/lovac42/ReMemorizeButtons/releases)', '[Button Colours (Good, Again)](https://ankiweb.net/shared/info/2494384865)'. (Although for some addons the buttons gets class names such as `<font color='red'>again</font>`)

### DIFF
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -621,14 +621,14 @@ time = %(time)d;
             return 2
 
     def _answerButtonList(self) -> Sequence[Tuple[int, str]]:
-        l = ((1, _("Again")),)
+        l = ((1, "Again"),)
         cnt = self.mw.col.sched.answerButtons(self.card)
         if cnt == 2:
-            return l + ((2, _("Good")),)
+            return l + ((2, "Good"),)
         elif cnt == 3:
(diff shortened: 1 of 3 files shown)


evandrocoan:
I would rather keep my buttons all black. Couldn't this be added in by an addon?

dae:
The approach you've taken will break the translations, as they are extracted from the source code, and can't be extracted when you pass a variable into _().

I think it may be better to hold off on making this change for a bit longer - existing users can already use an add-on to get colours, so this can probably wait until more breaking changes are required to the review screen.

BlueGreenMagick:
Sure, sounds reasonable. I hadn't realized that some people actually prefer black button text over colored ones!

PR #556, opened 2020-04-14
Render deck tree table more efficiently

BlueGreenMagick:
Current implementation of _renderDeckTree is not optimal, and it's almost impossible for an addon modify a part of the deck browser without replacing the whole function.

The proposed change separates the method so addons can only modify parts they need to. Even if monkey patching is discouraged, I don't see a clean way to add in versatile hooks with current code. With this change it will make writing addons modifying deckbrowser easier. If this gets accepted, I'll be doing a follow up PR with a hook. 

The PR also improves speed quite a bit. Here's a result of a speed test done on my laptop with approximately 20\*10\*10 decks, averaged over 5 times. (* means subdecks, total 2000 decks)

```
expanded original: 0.314s
expanded new: 0.099s
collapsed original: 0.022s
collapsed new: 0.001s
```

The only drawback is that it's probably going to break most addons monkey patching _renderDeckTree or _deckRow method.

### DIFF
--- a/qt/aqt/deckbrowser.py
+++ b/qt/aqt/deckbrowser.py
@@ -146,12 +146,10 @@ where id > ?""",
     def _render_deck_tree(self, nodes):
         if not nodes:
             return ""
-        buf = """
-<tr><th colspan=5 align=left>%s</th><th class=count>%s</th>
-<th class=count>%s</th><th class=optscol></th></tr>""" % (
-            _("Deck"),
-            tr(TR.STATISTICS_DUE_COUNT),
-            _("New"),
+        buf = """\
+<tr><th colspan=5 align=left>{}</th><th class=count>{}</th>
+<th class=count>{}</th><th class=optscol></th></tr>""".format(
+            _("Deck"), tr(TR.STATISTICS_DUE_COUNT), _("New"),
         )
         depth = 0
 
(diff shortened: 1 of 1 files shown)


BlueGreenMagick:
Not sure why the checks are failing. They are successful on my machine. And the changes shouldn't have any effect on the scheduler.

dae:
From a brief look, the code looks good, and the performance improvements are nice. My only concern is that I want to move away from building HTML in Python in the future. As you've pointed out on the support site, sending large amounts of HTML from Python can butt up against the 2MB limit, and it's not very efficient. Even greater performance improvements (not to mention ergonomic ones) could be achieved by sending just the data to the webview, and building the HTML on the Javascript side instead. Since that's likely what Anki will end up doing in the future, I'm not sure it's worth breaking add-ons at the moment for a change that (while an improvement) may end up being short-lived.

BlueGreenMagick:
One of the reasons for this PR was actually because I was trying to collapse and expand decks without refreshing the page. I imagine sending data to the web in the first place would be more efficient in that regard.

However, a concern I have as an add-on developer is, will it still be possible (and easy enough) for add-ons to interact with the code if many functions in Python are moved to JS side?

dae:
JS is a dynamic language like Python, and it should be possible to implement a similar hook mechanism. If anything it's likely to be easier as the DOM could be adjusted instead of adjusting/replacing HTML strings

BlueGreenMagick:
Sounds good.

PR #557, opened 2020-04-14
Fix editor buttons not highlighted

simgunz:
As it was, the selector 'button.linkb' takes precedence on '.highlighted' because
it is more specific, so that '.highlighted' is never applied.


dae:
Thanks!

PR #560, opened 2020-04-15
Fixed javaScriptConsoleMessage showing 'error on line'

evandrocoan:
Fixed javaScriptConsoleMessage showing 'error on line' for all messages. 

The new output format is:
```
JS info /_anki/reviewer.js:166 autoadd
JS error /_anki/reviewer.js:240 Uncaught (in promise) NotAllowedError: play() can only be initiated by a user gesture.
...
```

### DIFF
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -67,15 +67,24 @@ class AnkiWebPage(QWebEnginePage):  # type: ignore
         script.setRunsOnSubFrames(False)
         self.profile().scripts().insert(script)
 
-    def javaScriptConsoleMessage(self, lvl, msg, line, srcID):
+    def javaScriptConsoleMessage(self, level, msg, line, srcID):
         # not translated because console usually not visible,
         # and may only accept ascii text
-        buf = "JS error on line %(a)d: %(b)s" % dict(a=line, b=msg + "\n")
-        # ensure we don't try to write characters the terminal can't handle
-        buf = buf.encode(sys.stdout.encoding, "backslashreplace").decode(
-            sys.stdout.encoding
+        srcID = srcID.replace(AnkiWebView.webBundlePath("/"), "")
+        if level == QWebEnginePage.InfoMessageLevel:
+            level = "info"
+        elif level == QWebEnginePage.WarningMessageLevel:
+            level = "warning"
+        elif level == QWebEnginePage.ErrorMessageLevel:
+            level = "error"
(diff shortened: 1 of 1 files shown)


qt/aqt/webview.py
```diff
@@ -67,15 +67,24 @@ def cmd(self, str):
...
+            level = "warning"
+        elif level == QWebEnginePage.ErrorMessageLevel:
+            level = "error"
+        buf = "JS %(t)s /%(f)s:%(a)d %(b)s" % dict(
+            t=level, a=line, f=srcID, b=msg + "\n"
+        )
         # ensure we don't try to write characters the terminal can't handle
-        buf = buf.encode(sys.stdout.encoding, "backslashreplace").decode(
-            sys.stdout.encoding
+        buf = buf.encode(sys.stderr.encoding, "backslashreplace").decode(
```
dae, line 84: I believe switching to stderr will cause an error popup. There are shared decks floating around with bad Javascript, and changing to this behaviour is likely to cause an increase in posts to the support site

qt/aqt/webview.py
```diff
@@ -412,7 +421,8 @@ def stdHtml(
         # print(html)
         self.setHtml(html)
 
-    def webBundlePath(self, path: str) -> str:
+    @classmethod
```
dae, line 424: This looks like an unrelated change?

qt/aqt/webview.py
evandrocoan: I do not know anki deep enough to know this will cause problems, so I am reverting it back. I had set it to stderr because by convention, we should send to stdout only program results, computations, etc. While stderr is for everything else.

qt/aqt/webview.py
evandrocoan: I am using this function upstairs here:
```
File: anki/qt/aqt/webview.py
70:     def javaScriptConsoleMessage(self, level, msg, line, srcID):
71:         # not translated because console usually not visible,
72:         # and may only accept ascii text
73:         srcID = srcID.replace(AnkiWebView.webBundlePath("/"), "")
74:         if level == QWebEnginePage.InfoMessageLevel:
```

evandrocoan: Now I am not using it anymore because I set to use a regex as it should be a more reliable solution in case of future updates. Then, I kept the function as a classmethod because it can be a classmethod.

dae: Sorry, I missed that you were using it above.

PR #562, opened 2020-04-15
make help text selectable

ijgnd:
This allows the user to copy e.g. values so that they don't have to retype them which might introduce typos.


dae:
Thanks!

PR #563, opened 2020-04-15
Attempt to fix macos cache breaking for no reason

evandrocoan:
https://github.com/ankitects/anki/pull/560/checks?check_run_id=590022813#step:27:128 (https://github.com/ankitects/anki/pull/560), if the macos build starts breaking again in the future, after merging this, just ping and I will look into it.


dae:
Thanks!

PR #567, opened 2020-04-17
Fix regex performance

evandrocoan:
Sorry for the bug https://anki.tenderapp.com/discussions/beta-testing/1918-reliably-reproduced-bug-in-2124beta6

With this, I stopped compiling the regex everything, which should be a great improvement as regex object creations are quite expensive.


dae:
I suspect the wins will not be that great unless something is spamming console.log, but there's no reason not to. Thanks!

PR #569, opened 2020-04-17
Also limit the size of the error message

evandrocoan:
Also limit the size of the error message in just case it is built by a concatenation of stack traces, i.e.,
```js
throw new Error(err + err.stack)
```


dae:
cheers

PR #571, opened 2020-04-19
add .idea/ to .gitignore

zenogantner:


dae:
Anki doesn't require users to use a specific editor, so this is better placed in your ~/.gitignore instead

PR #572, opened 2020-04-19
mimic Makefile logic in diagnostic Python version output

zenogantner:


dae:
I think Evandro put the --version call in there while debugging, and rather than complicate that file, I think it's easier to just remove that line instead. Thank you in any case :-)

zenogantner:
:+1: makes sense to simplify!

PR #573, opened 2020-04-20
Set .prettierrc printWidth to 88 because the default is too short

evandrocoan:
Some codes were getting too much vertical, for example:
```js
pycmd(
    type +  
        ":" +   
        currentFieldOrdinal() + 
        ":" +   
        currentNoteId + 
        ":" +   
        currentField.innerHTML  
);
// Instead of  
pycmd(type + ":" + currentFieldOrdinal() + ":" + currentNoteId + ":" + currentField.innerHTML);
```

### DIFF
--- a/qt/ts/.prettierrc
+++ b/qt/ts/.prettierrc
@@ -1,5 +1,6 @@
 {
   "trailingComma": "es5",
+  "printWidth": 88,
   "tabWidth": 4,
   "semi": true
 }
--- a/qt/ts/src/editor.ts
+++ b/qt/ts/src/editor.ts
@@ -274,10 +274,7 @@ function enableButtons() {
 
 // disable the buttons if a field is not currently focused
 function maybeDisableButtons() {
-    if (
-        !document.activeElement ||
-        document.activeElement.className !== "field"
-    ) {
+    if (!document.activeElement || document.activeElement.className !== "field") {
         disableButtons();
     } else {
         enableButtons();
--- a/qt/ts/src/reviewer.ts
+++ b/qt/ts/src/reviewer.ts
@@ -41,10 +41,8 @@ function _updateQA(html, fadeTime, onupdate, onshown) {
         } catch (err) {
             qa.html(
                 (
-                    `Invalid HTML on card: ${String(err).substring(
-                        0,
-                        2000
-                    )}\n` + String(err.stack).substring(0, 2000)
+                    `Invalid HTML on card: ${String(err).substring(0, 2000)}\n` +
+                    String(err.stack).substring(0, 2000)
                 ).replace(/\n/g, "<br />")
             );
         }


dae:
Prettier advises against changing the defaults: https://prettier.io/docs/en/options.html

Could I trouble you to submit the rg changes in a separate PR?

evandrocoan:
They explicitly advise against 120, but they do not say about 100:
> Prettier, on the other hand, strives to fit the most code into every line. With the print width set to 120, prettier may produce overly compact, or otherwise undesirable code.

If you look at the changes this causes, they are quite minimal. I am writing some JavaScript in which the default limit of 80 characters is quite distressing.

dae:
The very first line in bold says "For readability we recommend against using more than 80 characters" :-)

evandrocoan:
The line wrap width is also relative to the code base characteristics. What criteria they are using when they recommend against using more than 80 characters? The only thing they said was:
> Developers often use whitespace to break up long lines for readability. In practice, the average line length often ends up well below the maximum.

They are also telling a lie here. When wrapping this code at 80 characters they are not doing what they say:
> Prettier, on the other hand, strives to fit the most code into every line.
```js
// This should not be wrapped
pycmd(type + ":" + currentFieldOrdinal() + ":" + currentNoteId + ":" + currentField.innerHTML);

// as:
pycmd(
    type +  
        ":" +   
        currentFieldOrdinal() + 
        ":" +   
        currentNoteId + 
        ":" +   
        currentField.innerHTML  
);

// If they were "strives to fit the most code into every line". 
// If so, it should be wrapped either as:
pycmd(
    type + ":" + currentFieldOrdinal() + ":" + currentNoteId + ":" + 
    currentField.innerHTML  
);

// or:
pycmd(
    type + ":" + currentFieldOrdinal() + ":" + 
    currentNoteId + ":" + currentField.innerHTML  
);

// But they just stick with the opposite and started "strives to fit the less code into
// every line" wrapping as much as it can. 
// They also go against the recommendation to put/wrap operators in new lines, then,
// instead they should wrap like this: (which looks much more readable)
// https://github.com/prettier/prettier/issues/3806
pycmd(
    type 
        + ":" 
        + currentFieldOrdinal() 
        + ":" 
        + currentNoteId 
        + ":" 
        + currentField.innerHTML
);

// instead of they wrap like:
pycmd(
    type +  
        ":" +   
        currentFieldOrdinal() + 
        ":" +   
        currentNoteId + 
        ":" +   
        currentField.innerHTML  
);
```

Readability is a very subjective term neither me, they or you can definitely say what its good readability. Just compare the codes against 80 characters versus 100 characters and draw out your conclusions. In my opinion, wrapping at 80 is too short, wrapping at 120 is too long, and wrapping at 100 is neither too short or too long. Unless you are planning something like have anki code ready to copy-paste into PowerPoint presentions, you could increase the line width:
1. https://dev.to/buinauskas/maximum-line-length-in-your-code-2ia4
1. https://softwareengineering.stackexchange.com/questions/312889/how-to-follow-the-80-character-limit-best-practice-while-writing-source-code

dae:
The goal of automatic formatting is not perfection, since no two people can exactly agree on what is perfect. The goal is to avoid time wasted manually formatting things, and sticking with the default settings avoids wasting time bikeshedding about what settings are best.

For what it is worth, In your "this should not be wrapped" example, the text is truncated in GitHub and needs to be scrolled to the right to read. Longer lines also make it harder to view multiple files side by side. I am not arguing that prettier creates the prettiest code, and it could certainly be made more intelligent, but I'll gladly live with imperfectly formatted code if it means not having to worry about manually formatting and not having to debate the relative merits of different line lengths.

evandrocoan:
> The goal of automatic formatting is not perfection, since no two people can exactly agree on what is perfect. The goal is to avoid time wasted manually formatting things, and sticking with the default settings avoids wasting time bikeshedding about what settings are best.

I never thought about their use with this purpose. Thanks for sharing!

> For what it is worth, In your "this should not be wrapped" example, the text is truncated in GitHub and needs to be scrolled to the right to read. Longer lines also make it harder to view multiple files side by side.

At least consider increasing the limit to 88 characters:
1. The example I showed is how wrapped again.
   ```js
    pycmd(
        type +  
            ":" +   
            currentFieldOrdinal() + 
            ":" +   
            currentNoteId + 
            ":" +   
            currentField.innerHTML  
    );
   ```
1. The `black` formatter (the one from python we use) also uses 88 characters as line wrap (with 4 spaces for each indentation level).
1. While `prettier` recommends 80 characters as line limit, we are using 4 spaces for each level of indentation, then, increasing it to 88 should make it even when compared for `JavaScript` projects which use `2` spaces for each indentation level.

I pushed a new commit changing it to 88 characters. Now there are only 2 changes: (and none of them are wrapping here on GitHub)
```js
    if (
        !document.activeElement ||
        document.activeElement.className !== "field"
    ) {
        disableButtons();
    } else {
        enableButtons();
    }
// ->
    if (!document.activeElement || document.activeElement.className !== "field") {
        disableButtons();
    } else {
        enableButtons();
    }

...

        } catch (err) {
            qa.html(
                (
                    `Invalid HTML on card: ${String(err).substring(
                        0,
                        2000
                    )}\n` + String(err.stack).substring(0, 2000)
                ).replace(/\n/g, "<br />")
            );
        }
// -->
        } catch (err) {
            qa.html(
                (
                    `Invalid HTML on card: ${String(err).substring(0, 2000)}\n` +
                    String(err.stack).substring(0, 2000)
                ).replace(/\n/g, "<br />")
            );
        }
```

While here, changing to 88 characters wrap limit is not making much changes, on this code I have on another branch, there are more changes impacted by this: (as there are more indentation levels of 4 spaces each): https://github.com/evandroforks/anki/commit/3b6b22ebe1ca0456bfefa0ad657051e888342e99

dae:
Thanks, I think that's a reasonable compromise, and IIRC it matches what black uses to format the Python code.

PR #574, opened 2020-04-20
Fixed ripgrep not installing on ubuntu anymore

evandrocoan:
https://github.com/BurntSushi/ripgrep#installation
https://github.com/BurntSushi/ripgrep/issues/1232

N.B. Various snaps for ripgrep on Ubuntu are also available, but none of them seem to work right and generate a number of very strange bug reports that I don't know how to fix and don't have the time to fix. Therefore, it is no longer a recommended installation option.)


dae:
thanks

PR #575, opened 2020-04-21
Keep window open during normal sync

Arthur-Milchior:
In case it is of interest to you, here is a modification that ensure that during normal sync, windows stay open. If a full sync is needed, it'll ask whether windows should be closed and still allow to cancel syncing if the user reject.

Otherwise, if you would accept to tell me why the database is closed each time, I'd be interested in learning it, seeing what I ovrelooked

I of course tested full upload/download (sorry for your server) and failing sync by cutting internet access during syncs.

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -861,16 +861,24 @@ title="%s" %s>%s</button>""" % (
 
     # expects a current profile and a loaded collection; reloads
     # collection after sync completes
+    def onSyncForce(self, action: str):
+        self.unloadCollection(lambda: self._onSync(action=action))
+
     def onSync(self):
         if self.media_syncer.is_syncing():
             self.media_syncer.show_sync_log()
         else:
-            self.unloadCollection(self._onSync)
+            self.col.save(trx=False)
+            self._onSync()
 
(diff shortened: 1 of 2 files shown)


dae:
I don't think this is safe at the moment:

- Backups are taken when the collection is closed, so if you avoid closing the collection, no backup will be taken before the sync.
- Syncing does not expect any changes to be made to the database during the sync, and your changes may allow timers or add-ons that expose a web interface to modify the collection in the middle of a sync.

Another reason why the database is closed is because until .24, the database object could not be accessed from a background thread.

Arthur-Milchior:
Thank you for your feedback. I'll put warning in the add-on description about the absence of backup.

I'll take a look also about the other problem

Arthur-Milchior:
By the way, I am realizing it did not really answer https://anki.tenderapp.com/discussions/ankidesktop/40723-rewriting-code-to-allow-avoiding-closing-of-window
Having the windows access to col through mw would allow to close the database and collection and still ensuring that those windows still work correctly. Of course, windows should be frozen during sync, but that's something that mw can already do, and frozen is far more usable than closed

PR #577, opened 2020-04-22
Add gcc-multilib to README.development

evandrocoan:
The rust compilation was failing due to this error:
1. https://stackoverflow.com/questions/6329887/compiling-problems-cannot-find-crt1-o
1. https://reposcope.com/package/gcc-multilib

### DIFF
--- a/README.development
+++ b/README.development
@@ -24,6 +24,7 @@ To start, make sure you have the following installed:
  - ripgrep (cargo install ripgrep)
  - git
  - curl
+ - gcc-multilib (sudo apt install gcc-multilib)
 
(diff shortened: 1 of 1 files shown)


dae:
Linux-specific instructions don't really belong at the top, and I wonder how many users actually need to explicitly install multilib? It's not something we do for CI AFAIK, and not something I've had to do when building locally.

evandrocoan:
I just tested and `multilib` was not required to be installed on Debian 9 and Ubuntu 18.04. I only had to install `multilib` on Linux Ming 19.2. 

Just for the record, what Linux distribution do you use? You can check whether multilib is installed by running `dpkg -l | grep multilib`

PR #578, opened 2020-04-25
Fixed Makefile comments showing up on the console

evandrocoan:
Makefile comments inside rules have to be put in the beginning
of the line, otherwise they are passed to shell when running the
recipe.


dae:
Thanks Evandro!

PR #580, opened 2020-04-26
mw.debug_diag is the dialog and not the window

Arthur-Milchior:
This will allow, for example, for add-ons to do self.debugDiag.text
and access the debugger text. I see very little reason to have a
direct access to the dialog window.

I should note that self.debugDiag is never used anywhere in the code,
so I think it should be deleted; unless you've heard of some add-on
using it.

This PR is based on a discussion with @ijgnd , which stated that he had trouble accessing the debugger text; and I realized there were good reason for it

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -1369,7 +1369,7 @@ will be lost. Continue?"""
     def onDebug(self):
         d = self.debugDiag = QDialog()
         d.silentlyClose = True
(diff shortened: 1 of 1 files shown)


dae:
The various parts of Anki call that the "form", so something like debug_dialog_form would be clearer

Arthur-Milchior:
Pushed with another name

dae:
Thanks

PR #583, opened 2020-04-26
Simplify the code slightly

Arthur-Milchior:
While updating an add-on which broke on beta, I found a really simple way to slightly simplify the code. I thought it might interest you.

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -126,8 +126,6 @@ class AddCards(QDialog):
             for n in range(len(min(note.fields, oldNote.fields))):
                 if not keep or flds[n]["sticky"]:
                     note.fields[n] = oldNote.fields[n]
-                else:
-                    note.fields[n] = ""
         self.setAndFocusNote(note)
 
     def removeTempNote(self, note: Note) -> None:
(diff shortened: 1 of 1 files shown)


dae:
cheers

dae:
This change was incorrect and has been reverted.

Arthur-Milchior:
I must admit that I do not have a clear idea about how to test more than what I did. Unit test worked, test done by myself worked. I would be happy to have any advice about how to test this kind of change which I did not already do.
The only thing that I usually have that could have helped here is a strong type system, which would have told me that "min" of array is probably not what I wanted; but I don't have it here.

I did test the branch before pushing. And it still worked. If I open the add card window and then change a note type, it did work perfectly. It probably only worked by accident, because the smallest array according to min was also the smallest one according to the size. 

Of course it was supposed to be `min(len(note.fields), (oldNote.fields))`, and if you accept this change it's already on https://github.com/Arthur-Milchior/anki/tree/remove_a_try

I also spent a lot of time double checking

dae:
Let's call it a string of bad luck. If you submit the fixed version as a new PR, I'll merge it.

PR #585, opened 2020-04-26
addButton always save function

Arthur-Milchior:
I had a strange problem while trying to make an add-on. I then realized that the trouble is that each time I called addButton I was using a slightly different method. I was expecting the button to call the function I passed. Instead, it called the function which was passed in the first call to addButton and it ignores next function. 

This change does just that: it always register the function if it is given.
I should note that addButton is never called in anki codebase. This change won't break any add-on which give multiple time the same method to addButton.

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -221,7 +221,7 @@ class Editor:
         disables: bool = True,
     ):
         """Assign func to bridge cmd, register shortcut, return button"""
-        if cmd not in self._links:
+        if func:
             self._links[cmd] = func
         if keys:
             QShortcut(  # type: ignore


dae:
@glutanimate I think this check was from one of your PRs - do you remember what the check is for?

glutanimate:
To be honest, I don't fully recall, and the [PR](https://github.com/ankitects/anki/pull/201) doesn't yield any additional info.

I think it was meant to protect against overwriting existing `cmd` values and thus against collisions with Anki's buttons and other add-ons, but it doesn't seem particularly well-suited for that purpose. In that case it would have made sense to fail early and loud.

I can't think of any reason to keep the check in. We don't check for similar collisions anywhere else, so I don't think it makes sense to do it here.

Generally speaking though, it would be nice to properly fortify the add-on API against add-on conflicts in the future (at least in places like this where it's somewhat feasible).

Arthur-Milchior:
We could request the add on name, and so ensure that the actual command is prefixed by the name/number which is assumed to be unique. I could do it easily. That would still create a problem of compatibility with add one who do not give their names

PR #586, opened 2020-04-26
Add rename method to ModelManager

BlueGreenMagick:
Currently, note types can only be renamed in note type manager. But I'd like to add a context menu to the browser sidebar treeview to allow renaming note types. It will be easier to maintain a single rename function in ModelManager.

### DIFF
--- a/pylib/anki/errors.py
+++ b/pylib/anki/errors.py
@@ -24,3 +24,12 @@ class DeckRenameError(Exception):
 
     def __str__(self):
         return "Couldn't rename deck: " + self.description
+
+
+class ModelRenameError(Exception):
+    def __init__(self, description: str) -> None:
+        super().__init__()
+        self.description = description
+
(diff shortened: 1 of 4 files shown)


BlueGreenMagick:
Is there a reason TR is imported from anki.rsbackend instead of anki.fluent_pb2? 

I'm asking because adding `from anki.rsbackend import TR` in models.py causes circular import. In this case, it can be solved by replacing `from anki.models import AllTemplateReqs` with `from anki import models`, and changing `AllTemplateReqs` type hint to `models.AllTemplateReqs` in rsbackend.py. But I was wondering if it would be better to import from `anki.fluent_pb2` instead.

dae:
I'm currently refactoring the notetype and deck handling, so please hold off on changes to those areas for now - will return to this.

BlueGreenMagick:
Sure, will do.

PR #588, opened 2020-04-27
Add rename method to TagManager

BlueGreenMagick:
As noted in #586, I'm planning to add a context menu to the deck browser sidebar tree view. It'd be great if tags could be renamed, instead of having to bulk add, then bulk delete the tags.

If you plan to accept my PR, please scrutinize over my changes more than usual. I have close to zero knowledge in db, and pretty much just copied over the code from bulkAdd.

### DIFF
--- a/pylib/anki/tags.py
+++ b/pylib/anki/tags.py
@@ -119,6 +119,9 @@ class TagManager:
 
     def rename(self, old_tag: str, new_tag: str) -> None:
         "old_tag, new_tag must not contain whitespace. old_tag may contain wildcard *."
+        if " " in old_tag or " " in new_tag:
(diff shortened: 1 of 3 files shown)


dae:
Might be a while before I can return to this one

dae:
The bulkAdd code is likely to be moved into Rust in the future, and it would make sense to do this code there as well. So while I'll hold off on merging the Python code, I'll leave this open as a reminder.

dae:
Please give bulk_update() in master a test

BlueGreenMagick:
As far as I can tell, it works great! It is going to be incredibly useful. (When it gets added to the gui)

PR #590, opened 2020-04-28
Sidebar context menu

BlueGreenMagick:
This PR adds a custom context menu to the browser sidebar.

The context menu will likely have different actions per SidebarItem type. For example, a filter should have an action that can edit the filter. And according to its type, an action that looks the same ("Rename" for example) will likely need different function. So when calling SidebarItem, I passed its type to the constructor, which can later be accessed by SidebarTreeView when context menu is triggered.

I thought I should get a green light beforehand, so I only added one action for renaming deck as an example.

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -589,6 +589,10 @@ class SidebarTreeView(QTreeView):
         self.expanded.connect(self.onExpansion)
         self.collapsed.connect(self.onCollapse)
 
+        self.setContextMenuPolicy(Qt.CustomContextMenu)
+        self.customContextMenuRequested.connect(self.onContextMenu)
+        self.context_menus = {SidebarItemType.DECK: ()}
+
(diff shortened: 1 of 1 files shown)


BlueGreenMagick:
If you accept the PR, I think you should also consider seperating Browser and Sidebar into different files. The file is going to get quite a bit longer when actions get added :)

dae:
Will get back to you after the new release has settled down a bit

dae:
It looks like an improvement, and moving it to another file would make the code easier to maintain as well. But it also means breaking a bunch of add-ons that interact with that area :-( I think it's best to hold off on this for a little bit, accumulate some more changes, and then make them all at once to try to minimize churn. SidebarItem would be nicer as a dataclass, the deck tree is awkward to work with as a list, tags could be made into a tree, etc.

dae:
Added to https://github.com/ankitects/help-wanted/issues/6

PR #591, opened 2020-04-28
make addon text selectable and link accessible

BlueGreenMagick:
Related: #562, ce589fb97592f88129c981ad37bd21aa748acf6d, https://anki.tenderapp.com/discussions/beta-testing/1920-config-can-copy-paste-but-cant-click-links-now


dae:
Is this actually necessary? Both are set in the designer file now, so it should not be required.

BlueGreenMagick:
You are right! I didn't think to look at the ui file

PR #592, opened 2020-04-29
Set tab width to 4 spaces

BlueGreenMagick:
Instead of which I presume is an arbitrary number?


dae:
Thanks!

evandrocoan:
@BlueGreenMagick, this is causing this bug: https://anki.tenderapp.com/discussions/ankidesktop/41140-duplication-on-back-of-card

PR #536, opened 2020-04-03
Fix time bug between 02:00 and 04:00 UTC

evandrocoan:
Continuing from https://github.com/ankitects/anki/pull/530, Now, this has nothing but some prints. Tomorrow at 02:00 I will rerun the tests on GitHub and compare the results with my local machine (where the tests are passing).

Logs:
1. [notworking_local.txt](https://github.com/ankitects/anki/files/4432692/notworking_local.txt)
1. [notworking_remote.txt](https://github.com/ankitects/anki/files/4432693/notworking_remote.txt)
1. [working_local.txt](https://github.com/ankitects/anki/files/4432694/working_local.txt)
1. [working_remote.txt](https://github.com/ankitects/anki/files/4432695/working_remote.txt)

PR #594, opened 2020-04-29
Fixed timestamp.rs not shifting the time when running python tests

evandrocoan:
By using the debug code from https://github.com/ankitects/anki/pull/536 (Fix time bug between 02:00 and 04:00 UTC) I finally manage to figure out the problem and fix it. The problem is that when you run the Unit Tests on Python, the rust code also needs to shift the time as the Python code is doing:
```python
File: anki/pylib/tests/shared.py
 8: # Between 2-4AM, shift the time back so test assumptions hold.
 9: lt = time.localtime()
10: if lt.tm_hour >= 2 and lt.tm_hour < 4:
11:     orig_time = time.time
12: 
13:     def adjusted_time():
14:         return orig_time() - 60 * 60 * 2
15: 
16:     time.time = adjusted_time
17: else:
18:     orig_time = None
```
But the `#[cfg(test)] fn elapsed()` use cannot be seen when running the unit tests from Python code:
```rust
File: anki/rslib/src/timestamp.rs
29: // when running in CI, shift the current time away from the cutoff point
30: // to accomodate unit tests that depend on the current time
31: #[cfg(test)]
32: fn elapsed() -> time::Duration {
33:     use chrono::{Local, Timelike};
34: 
35:     let now = Local::now();
36: 
37:     let mut elap = time::SystemTime::now()
38:         .duration_since(time::SystemTime::UNIX_EPOCH)
39:         .unwrap();
40: 
41:     if now.hour() >= 2 && now.hour() < 4 {
42:         elap -= time::Duration::from_secs(60 * 60 * 2);
43:     }
44: 
45:     elap
46: }
```

This way, only the times used on the Python code are correctly shifted, while the rust code is not, thus, causing that weird time bug.

This pull request uses the trick from https://github.com/rust-lang/cargo/issues/4669 (cfg test is not set during doctests) to build rust code using the correct test code when running `make check`.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -264,6 +264,27 @@ jobs:
       - run: make check
         if: matrix.BUILD_TYPE == 'check'
 
+      - name: Upload mypy
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v1
+        with:
+          name: mypy ${{ matrix.os }}
+          path: qt/.mypy_cache
+
+      - name: Upload pylib
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v1
+        with:
+          name: pylib ${{ matrix.os }}
+          path: pylib
+
+      - name: Upload rslib\src
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v1
+        with:
+          name: rslib\src ${{ matrix.os }}
+          path: rslib\src
+
       - name: Upload python wheels
         if: matrix.BUILD_TYPE == 'build'
         uses: actions/upload-artifact@v1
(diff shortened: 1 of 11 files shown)


dae:
Thank for looking into this Evandro!

This is probably the most "correct" approach, though I'm guessing it's going to ~double the modify/compile/test time when making changes to the Rust code and then running the Python tests, since 'make check' now has to compile rspy twice. Will give it some thought.

dae:
Unless I'm mistaken, this workaround was only required for the Python tests, so using the 'test' configuration seems like overkill - I've pushed an update that uses an envvar instead.

PR #597, opened 2020-04-30
Fixed rspy check not being called

evandrocoan:
Only the develop rule was being called when running make check.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ BUILDFLAGS := --release --strip
 DEVFLAGS := $(BUILDFLAGS)
 RUNFLAGS :=
 CHECKABLE_PY := pylib qt
-CHECKABLE_RS := rslib
+CHECKABLE_RS := rslib rspy
 DEVEL := rslib rspy pylib qt
 
 .PHONY: all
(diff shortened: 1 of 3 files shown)


dae:
Unfortunately this makes development really painful on macOS. Python modules need to be compiled with separate build options, so each time maturin is run, it invalidates the build products, requiring cargo clippy to recompile everything again. And vice versa - cargo clippy means maturin has to recompile everything as well. That's why I removed the check, as apart from missing out on some clippy lints (which I've just pushed a separate fix for), the rspy check doesn't gain us much.

evandrocoan:
I see. Thanks for explaining. With my latest commit:
1. I removed `clippy` from `make check` on `rspy/Makefile` and I created an exclusive rule with `make clippy` on `rspy/Makefile` in case someone would like to call it.
1. I made rspy/Makefile `make check` call `develop` instead of build, meaning it is not required to exclicity call for `make develop` on the main make file (as just done in 43978b72570ea0bbebcee283e2c153dc2647ce8e)

With this, the main make file rule can be simplified from:
```make
	.github/scripts/trailing-newlines.sh; \
	$(SUBMAKE) -C rslib develop; \
	for dir in $(CHECKABLE_RS); do \
		$(SUBMAKE) -C $$dir check; \
	done; \
	. "${ACTIVATE_SCRIPT}"; \
	$(SUBMAKE) -C rspy develop; \
	$(SUBMAKE) -C pylib develop; \
	for dir in $(CHECKABLE_PY); do \
		$(SUBMAKE) -C $$dir check; \
	done;
```

To:
```make
	.github/scripts/trailing-newlines.sh; \
	. "${ACTIVATE_SCRIPT}"; \
	for dir in $(CHECKABLE_RS) $(CHECKABLE_PY); do \
		$(SUBMAKE) -C $$dir check; \
	done;
```

evandrocoan:
> Python modules need to be compiled with separate build options, so each time maturin is run, it invalidates the build products

Could `maturin` build things into a directory X while `clippy` build things into another one? i.e., using some flag as `CARGO_TARGET_DIR`?

dae:
cheers

PR #535, opened 2020-04-02
Created the GitHub Actions step Upload to PyPi

evandrocoan:
Runs only on tagged releases of anki:
1. https://anki.tenderapp.com/discussions/ankidesktop/40025-anki-in-python-package-index

Requires a GitHub Actions Secret to be set with the name `TWINE_PASSWORD` where the contents are a pypi token, i.e., `TWINE_PASSWORD: pypi-12k3j12lj3hj1kj2bç31...`. Notice the token contents have to start with `pypi-`, e.g., `pypi-12k3j12lj3hj1kj2bç31...`.
1. http://github.com/ankitects/anki/settings/secrets
1. http://pyfound.blogspot.com/2019/07/pypi-now-supports-uploading-via-api.html

As there are 3 different projects:
1. https://pypi.org/project/aqt/
1. https://pypi.org/project/anki/
1. https://pypi.org/project/ankirspy/

The token `TWINE_PASSWORD` will have to have access to all repositories on the account. An alternative to this would be to duplicate the code 3 times, each one of them, uploading one repository at a time with a different `TWINE_PASSWORD` token. On this case, 3 pypi tokens would have to be created, one for each repository.
1. `TWINE_PASSWORD_AQT`
1. `TWINE_PASSWORD_ANKI`
1. `TWINE_PASSWORD_ANKIRSPY`

```yml
      - name: Upload aqt to PyPi Linux/Mac OS
        if: matrix.BUILD_TYPE == 'build' && startsWith(github.ref, 'refs/tags/')
        env:
          TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD_AQT }}
          TWINE_USERNAME: __token__
        run: |
          python -m pip install twine
          twine upload --non-interactive --skip-existing --verbose dist/aqt*

      - name: Upload Anki to PyPi Linux/Mac OS
        if: matrix.BUILD_TYPE == 'build' && startsWith(github.ref, 'refs/tags/')
        env:
          TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD_ANKI }}
          TWINE_USERNAME: __token__
        run: |
          python -m pip install twine
          twine upload --non-interactive --skip-existing --verbose dist/anki-*

      - name: Upload ankirspy to PyPi Linux/Mac OS
        if: matrix.BUILD_TYPE == 'build' && startsWith(github.ref, 'refs/tags/')
        env:
          TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD_ANKIRSPY }}
[...]

PR #600, opened 2020-04-30
Add missing aqt and anki modules dependency requirements

evandrocoan:
1. https://github.com/ankitects/anki/pull/535#issuecomment-620268597 - Created the GitHub Actions step Upload to PyPi
1. https://github.com/PyO3/maturin/issues/304 - `maturin develop` hides the package from pip


dae:
Thanks!

PR #602, opened 2020-05-01
Fixed GitHub Actions tagged wheels

evandrocoan:


dae:
Doesn't the --skip-existing arg to twine protect against the uploads being overwritten? Perhaps I'm misunderstanding what you're trying to fix here?

evandrocoan:
This is not about PyPi, but the artifacts that are uploaded to Github Actions: 
image

dae:
Gotcha - thanks :-)

PR #528, opened 2020-03-25
Create actions for Windows and Mac OS

evandrocoan:
I saw you recently created the `uses: ankitects/setup-protoc@master`, because `setup-protoc` was not working. I also had the same problem when I was creating the GitHub actions for Mac OS. The solution come after opening a issue on (https://github.com/actions/virtual-environments/issues/602 - API rate limit exceeded for Mac OS hosted by GitHub itself). The problem was that the README.md of the `setup-protoc` project was not updated to use the correct version of `setup-protoc`. I opened a issue there, asking for their README.md to be updated (https://github.com/arduino/actions/issues/23 - Mark the repository as deprecated). Meanwhile, on this pull request, I am already using the correct version of the `setup-protoc` actions as suggested by https://github.com/actions/virtual-environments/issues/602.

1. I created an Mac OS tests, after fixing (https://github.com/actions/virtual-environments/issues/602 - API rate limit exceeded for Mac OS hosted by GitHub itself).
1. I added to also run `build develop` because while I was building this I got a case where `make check build` was working/passing, but `make develop` not. I put the develop rule first because it should be the most important, i.e., a user can run the develop build without errors. Then, I check whether Anki can pack itself into python wheels and finally I run all unit tests. Also, by putting `develop` before `check` allows to each one of them build half of things. If `check` is the first rule, it build all rust libraries at once instead of just part of them.
1. I separated `make develop`, `make check` and `make build` in different rules just to:
    1. Know more easily in which step the build failed
    1. Know how much time each one of them is taking to complete
1. I created caches for everything I could, to speed up the build process. GitHub automatically deletes the caches every week and the caches are also regenerated on certain conditions:
    1. For the cache of `pyenv` directory, they are rebuild when you edit a `Makefile` or some `requirements.txt` file.
    1. For the cache of package manager `pacman`, they are rebuild every time you edit the `windows_checks.yml` file.
    1. To build the project using all caches should take:
        1. About 7 minutes on Linux and Mac (without cache about 12 to 20 minutes)
        1. About 10 minutes on Windows (without cache about 22 minutes)
1. Additionally to all these, you can always for the cache to be rebuilt by editing the line `key:` and increment its the number at the end of the line or just by changing anything else on this composition.
1. I created the special environment variable `SHELLFLAGS` to allow the user to easily debug installation errors by defining the environment variable `SHELLFLAGS=-x` before running anything. The flag `-x` will cause bash to print out every command it is running. This is vital to figure out how to make the GitHub Actions to work.
[...]

PR #603, opened 2020-05-02
Also run rspy clippy on macos

evandrocoan:
Almost the same thing as https://github.com/ankitects/anki/pull/597 (Fixed rspy check not being called), but also runs `clippy` on Mac OS by changing/using a different `CARGO_TARGET_DIR` just for `clippy`.

Detect we are on Mac OS and pass different locations for `clippy` and `maturin`. It would just use more disk/file system space for Mac OS, but keep the compilation times short.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ BUILDFLAGS := --release --strip
 DEVFLAGS := $(BUILDFLAGS)
 RUNFLAGS :=
 CHECKABLE_PY := pylib qt
-CHECKABLE_RS := rslib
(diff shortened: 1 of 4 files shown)


evandrocoan:
Closing it because:
1. This would make the Mac OS CI take longer to also build `clippy` things.
1. Even if we cached the new `targetclippy` directory, it would require `500MB` of space and we already use enough space of the 5GB quota (https://github.com/ankitects/anki/pull/528 - Create actions for Windows and Mac OS).

PR #606, opened 2020-05-04
Remove pyaudio as mandatory dependency

evandrocoan:
Pyaudio on Windows is difficult to install and complicates things
when someone is using the wheels just for autocomletion on IDEs as
PyCharm:
1. https://anki.tenderapp.com/discussions/add-ons/44009-problems-with-code-completion

Related to:
1. https://github.com/ankitects/anki/pull/530 - Set to checks.yml install and import anki wheels


dae:
Sounds good, thanks

PR #607, opened 2020-05-04
Simplify the copying of field contents when resetting

Arthur-Milchior:
This replace 8 lines by 3 and allow to remove a try/except. Otherwise, it's not supposed to change anything in code's behavior.


dae:
thanks

PR #608, opened 2020-05-04
Fix missing pyaudio blocking Anki from running

evandrocoan:
Continuing from https://github.com/ankitects/anki/pull/606 - Remove pyaudio as mandatory dependency

### DIFF
--- a/qt/aqt/sound.py
+++ b/qt/aqt/sound.py
@@ -13,8 +13,6 @@ from concurrent.futures import Future
 from operator import itemgetter
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
-import pyaudio
-
 import aqt
 from anki.cards import Card
 from anki.lang import _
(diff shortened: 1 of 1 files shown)


dae:
Normal users don't see anything printed to stdout, so this warning will likely be missed. It would be better to use showWarning() when the user actually tries to record something, telling them they need to install pyaudio. Instructing them to use python/pip is also not a great idea, as that may not be appropriate for users running using a system-installed Python

evandrocoan:
I used show warning and suggested to install pyaudio. This warning should only be seen by someone who installed the wheel from pypi or from the source as the packed version of Anki should have pyaudio embed.

dae:
Thanks!

dae:
Re your previous comment, some Linux distributions package up Anki themselves and don't always list the dependencies as required, so this warning will be seen more frequently than you might expect :-(

PR #609, opened 2020-05-07
Update pylib/anki/consts.py HELP_SITE link to the new location

evandrocoan:
1. https://apps.ankiweb.net/docs/manual.html
1. https://docs.ankiweb.net/#/


dae:
The old page redirects a bunch of the old anchors to the new ones, so this change will break all those old links if the links in the code are not updated at the same time.

evandrocoan:
I did not know the anchos were being redirected. Then, I am closing it until something forces an update.

PR #610, opened 2020-05-07
Ask user confirmation before moving the Anki directory

evandrocoan:
https://anki.tenderapp.com/discussions/ankidesktop/41245-unwanted-file-movingdeleting-behavior

1. image
1. image
1. image

### DIFF
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -33,7 +33,7 @@ appUpdate = "https://ankiweb.net/update/desktop"
 appHelpSite = HELP_SITE
 
 from aqt.main import AnkiQt  # isort:skip
-from aqt.profiles import ProfileManager  # isort:skip
+from aqt.profiles import ProfileManager, AnkiRestart  # isort:skip
 
 profiler = None
 mw: Optional[AnkiQt] = None  # set on init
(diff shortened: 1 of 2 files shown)


dae:
I feel a bit uneasy about this approach - I'm not sure constructing multiple QApplications is safe, and worry that having the GL fixes applied after the first QApplication is created may cause issues. I don't see an easy way this could be done differently in a way that requires minimal changes to the code however - any other ideas?

evandrocoan:
It should be safe because the first application object would be destroyed by going out of scope. What could be concerning is a Qt bug in some platform due QApplication being a global class. But this can be fixed by restarting the Anki Application.

References:
1. https://stackoverflow.com/questions/46304070/multiple-qapplication-instances
1. https://stackoverflow.com/questions/43111713/totally-deleting-the-qapplication-instance-and-recreating-it-in-another-thread
1. https://stackoverflow.com/questions/13048371/pyqt4-initialize-and-delete-qapplication-instances-many-times

dae:
I think simplest is probably best - this is a rare code path, and I don't want to sink a lot of time into testing it on all three platforms and dealing with any accidental regressions. My only critique on the code is that using an exception for a success case is a bit confusing - I'll tweak the code after merge to use a normal retval instead. Thanks for the contribution!

dae:
I take that back -  an exception looks like the least-bad option here.

PR #611, opened 2020-05-09
Automatically install pyaudio when running make check

evandrocoan:
Continuing from https://github.com/ankitects/anki/pull/606 - Remove pyaudio as mandatory dependency

As the Linux/Mac OS instructions on `README.development` do not require to set `ANKI_EXTRA_PIP`, they should be updated asking it to be set to `ANKI_EXTRA_PIP="python -m pip install pyaudio`, but once `portaudio` is installed on Linux Mac OS (as asked on `README.development`), it is better to automatically install `pyaudio` on `make check` instead of asking Linux/Mac OS users to redundantly set `ANKI_EXTRA_PIP="python -m pip install pyaudio"`.


dae:
cheers

PR #612, opened 2020-05-09
avoid playing entire media collection with mpv when audio path is empty

siikamiika:
There's a bug on platforms that use mpv to play sound that if the `[sound:]` is empty, Anki will play every single file in the media directory. We noticed that when fixing another bug in https://github.com/FooSoft/yomichan/pull/524. It didn't reproduce on Windows, so I decided to investigate mpv related code in Anki and found the issue.

In this case it's caused by mpv's `loadfile` playing directory contents by default, but other players used by Anki could have similar issues as well, so I'm not sure if this is the correct place to fix it.

Also, I wasn't sure if `gui_hooks.av_player_did_begin_playing(self, tag)` is meant to be triggered even when the player doesn't start playing anything, so I didn't make the code return on invalid paths. It could also be done this way:
```python
# avoid playing the entire directory
if not os.path.isfile(path):
    return
self.command("loadfile", path, "append-play")
```

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -189,7 +189,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ${{ github.workspace }}${{ matrix.SEP }}pyenv
-          key: ${{ runner.os }}-pyenv-${{ hashFiles('**/requirements.*') }}-${{ hashFiles('**/setup.py') }}-${{ hashFiles('**/Makefile') }}-15-
+          key: ${{ runner.os }}-pyenv-${{ hashFiles('**/requirements.*') }}-${{ hashFiles('**/setup.py') }}-${{ hashFiles('**/Makefile') }}-16-
 
       # # Disable it in attempt to reduce the overall cache size (https://github.com/ankitects/anki/pull/528)
       # - name: Cache pip wheels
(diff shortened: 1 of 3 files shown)


dae:
While not officially supported, some users have http links in their sound tags, and I think this PR will break that workflow. 

Checking for a blank filename would be a safer option, though I do wonder whether this is an issue that crops up frequently?  I don't recall anyone reporting it before.

siikamiika:
> While not officially supported, some users have http links in their sound tags, and I think this PR will break that workflow.

Ok. I don't see that working with `MpvManager` currently though; what it does is joining `os.getcwd()` and the filename in the tag together before playing anything. Maybe other sound player components have a regex check for URLs?

> Checking for a blank filename would be a safer option, though I do wonder whether this is an issue that crops up frequently? I don't recall anyone reporting it before.

Sure, I can switch to that approach instead because that's what the issue is mainly about. I thought that it would be the most exhaustive option to check if the joined path is a file because then it would also avoid trying to play non-existent files.

I'm not sure how common this issue is, but when it happens it can be very disruptive. We are generating the cards automatically using [a plugin](https://github.com/FooSoft/anki-connect) that's also used by some other projects. What happened was that we returned an empty string in order to skip unwanted audio, but it was interpreted by mpv as a command to play everything.

Side note, I don't know how relevant security issues are for Anki because users are expected to mostly create the content themselves, but there's also a risk of path traversal and requesting resources on the local network.

siikamiika:
Now it ignores paths that are the same as cwd. The error in the CI fail seems to have happened before: https://github.com/PyO3/maturin/issues/283

evandrocoan:
It looks like a problem in the cache of pyenv or maturin is going nuts again. You can force reset the pyenv cache by bumping `-15-` on `checks.yml - Cache pyenv` section

siikamiika:
@evandrocoan Thanks, that did the trick :+1:

dae:
Good point about getcwd(), I must have broken this when updating the audio code a few months back. I've made a change to the audio file extraction code instead - could you confirm the latest master solves the problem for you?

dae:
I'll assume that's a yes - if not, please feel free to post back.

siikamiika:
Sorry, I was a bit busy and couldn't test until now. It's indeed fixed for the case that we found when developing Yomichan.

It's still possible to trigger this behavior if you do something like `[sound:./]`, but as I understood it Anki is trying to give the users as much freedom as possible with their use cases and not strictly enforce loading a single file from the media collection directory.

PR #613, opened 2020-05-10
Add missing close file descriptors using context managers

evandrocoan:


dae:
Is this trying to fix a problem, or are you just trying to make the code "neater"? cpython is reference counted, so open("foo").write(...) should immediately close the file after writing AFAIAA

evandrocoan:
After researching I found this: https://stackoverflow.com/questions/41375713/can-reference-counting-be-relied-on-to-close-a-file-in-python

> Someone pointed to the [Data Model section of the Python Language Reference][1] where it says very explicitly "_Do not depend on immediate finalization of objects when they become unreachable (so you should always close files explicitly)._". So, this makes it clear that the code is relying on behavior that's not guaranteed.

Then, this should be to guarantee correctness. But also:
1. There is a bug fix on the class `csvfile.py`. At the end of the function it is called `self.fileobj.close()` but the `self.fileobj` is not set to `None` meaning the next time some function calls `open()`, it will check that `self.fileobj` is not `None` and it will try to access a closed file descriptor.
1. On the class `webview.py` I put retrieve the error message from Qt instead of just failing an assertion, so we get a better context/reason about the failure.

  [1]: https://docs.python.org/3/reference/datamodel.html#index-2

dae:
I suspect CPython will never change this behaviour, as it would break a lot of existing code. But fair point in any case. Most of the changes look fine, but if we're going to be explicit about closing, I think the caller of csvfile should explicitly call .close(), instead of relying on a __del__ method

evandrocoan:
According to [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) (Resource Acquisition Is Initialization), the caller to `csvfile.py` should not call close because it is not him to calls `open()`. It is our internal `cacheFile()` function which does this.

The caller to `cvsfile.py` should call `close` if the caller had opened the file by himself, i.e., he had passed the file descriptor as a parameter to the `csvfile.py` constructor. For example:
```python
class TextImporter(NoteImporter):
    def __init__(self, col, fileobj):
		self.fileobj = fileobj

fileobj = open("something.txt")
importer = TextImporter(col, fileobj)
fileobj.close()
```

But this is not what we do. `TextImporter` is the one which opened the file in his constructor (by lazy initialization on `self.cacheFile`), so `TextImporter` must close the file on its destructor:
```python
class TextImporter(NoteImporter):
    def __init__(self, _Collection, filestr):
	NoteImporter.__init__(self, col, filestr)

    def __del__(self):
        self.close()

importer = TextImporter(col, "something.txt")
```

But, of course, beyond having the destructor closing the file, we can also make an explicit call to `importer.close()` when/every time we are done with the object. Then, for compliance with the rest of the code, I pushed a commit adding an explicit call to `importer.close()` once we are done with its usage.

pylib/anki/importing/csvfile.py
```diff
@@ -134,6 +134,20 @@ def fields(self) -> int:
...
+
+    def __del__(self):
+        self.close()
+        try:
+            super().__del__  # type: ignore
+        except AttributeError:
+            pass
+        else:
+            super().__del__(self)  # type: ignore
+
```
dae, line 150: How about the following instead, which IMHO is clearer about what it's doing:

```
s = super()
if hasattr(s, "__del__"): 
    s.__del__(self)
```

dae:
__del__ was a bit of a footgun in earlier Python versions, so seeing it sets of alarm bells :-) But I gather it's better behaved these days

dae:
Thanks :-)

PR #614, opened 2020-05-10
Automatically relaunches anki on migration

evandrocoan:
This is the same thing as https://github.com/ankitects/anki/pull/610 - Ask user confirmation before moving the Anki directory

With the exception, this has an extra commit which automatically reopens Anki instead of just exiting:
```python
except AnkiRestart as error:
    import subprocess

    if error.exitcode:
        sys.exit(error.exitcode)

    # first argument will be `qt/runanki` when calling it directly from a python interpreter
    if os.path.abspath(argv[0]) == os.path.abspath(sys.executable):
        arguments = argv
    else:
        arguments = [sys.executable] + argv

    print("Relaunching Anki", arguments)
    newanki = subprocess.Popen(arguments, env=os.environ)
    newanki.communicate()
    return
```

### DIFF
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -406,8 +406,20 @@ def _run(argv=None, exec=True):
         pm = ProfileManager(opts.base)
         pmLoadResult = pm.setupMeta()
     except AnkiRestart as error:
+        import subprocess
+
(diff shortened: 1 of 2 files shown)


dae:
Closing in favour of #610

PR #615, opened 2020-05-11
Create the opengl setup class

evandrocoan:
Similar to https://github.com/ankitects/anki/pull/610 - Ask user confirmation before moving the Anki directory

But instead of asking the user to restart Anki, it removes the base directory logic from `ProfileManager` and creates a separate class called `OpenGlSetup` to handle the OpenGl and base directory logic.

This way, the only thing which cannot be set before the first Qt application is launched is the `QT_SCALE_FACTOR` which requires the profile/database to be loaded before access to `pm.uiScale()`. But after testing, the `QT_SCALE_FACTOR` is correctly set for the second Qt application when it is created (while the first application has the standard UI scale, the second application has it correctly set).

To be more conservative, I could still ask for the Anki application to be restarted after the directory migration has completed. The advantage of this pull request against the others is that:
1. The code seems better placed (after removing the base directory logic from `ProfileManager`)
1. The first Qt application will have all OpenGl fixes (with the exception of the UI scale)

### DIFF
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -406,8 +406,20 @@ def _run(argv=None, exec=True):
         pm = ProfileManager(opts.base)
         pmLoadResult = pm.setupMeta()
     except AnkiRestart as error:
+        import subprocess
+
         if error.exitcode:
             sys.exit(error.exitcode)
+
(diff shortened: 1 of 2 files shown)


dae:
Thanks, will come back to this.

dae:
Closing in favour of #610

PR #618, opened 2020-05-11
Also reset checks.yml pyenv cache on Cargo.toml file changes

evandrocoan:
Attempt to fix pyenv failing with `` Expected `python` to be a python interpreter inside a virtualenv `` - https://github.com/ankitects/anki/pull/612, by also resetting checks.yml pyenv cache on Cargo.toml file changes


dae:
Thanks as always for your work on this Evandro - it's nice to have one less thing to worry about :-)

PR #621, opened 2020-05-13
Use mpv on Windows if available

ANH25:
Starting from version 2.1.20, Anki starts a new copy of mplayer for each audio file on Windows. This can cause noticeable slowness when playing audio.

Using mpv if available could solve that problem (unless there are caveats I'm not aware of).

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -33,6 +33,7 @@ Henry Tang <hktang@ualberta.ca>
 Simone Gaiarin <simgunz@gmail.com>
 Rai (Michal Pokorny) <agentydragon@gmail.com>
 Zeno Gantner <zeno.gantner@gmail.com>
+ANH <github.com/ANH25>
 
 ********************
 
--- a/qt/aqt/sound.py
+++ b/qt/aqt/sound.py
@@ -630,13 +630,12 @@ def setup_audio(taskman: TaskManager, base_folder: str) -> None:
     # legacy global var
     global mpvManager
 
-    if not isWin:
-        try:
-            mpvManager = MpvManager(base_folder)
-        except FileNotFoundError:
-            print("mpv not found, reverting to mplayer")
-        except aqt.mpv.MPVProcessError:
-            print("mpv too old, reverting to mplayer")
+    try:
+        mpvManager = MpvManager(base_folder)
+    except FileNotFoundError:
+        print("mpv not found, reverting to mplayer")
+    except aqt.mpv.MPVProcessError:
+        print("mpv too old, reverting to mplayer")
 
     if mpvManager is not None:
         av_player.players.append(mpvManager)


evandrocoan:
The Mac OS build failure does not make much sense. I would rerun it to confirm something is actually broken.

dae:
MPV was disabled on Windows because it was buggy and tended to get stuck - probably due to an issue in the Windows pipes handling in mpv.py rather than an issue with mpv itself.

kelciour:
Can confirm, it happened to me too a few months ago, but outside of Anki, when I used python to read a list of 250-500 audio files and parse subtitles generated and aligned by https://github.com/readbeyond/aeneas, then for each audio file tell mpv to load the audio file and skip to the last subtitle to quickly check if it was synced more or less correctly or not, and wait for Enter. And after pressing Enter ~7-10 times mpv just stopped loading new audio files and pressing Enter did nothing. When I removed skipping and only left loading, it changed nothing and mpv still failed to load new audio files.

In this particular case it was "fixed" by creating new `_sock = win32file.CreateFile` every time before loading new audio file.

For some unknown reason, I can't reproduce it in Anki after creating a deck with the same audio files and temporarily switching from mplayer to mpv using the following code in Debug Console in Anki 2.1.20+.

	import anki.sound
	print("Before:", anki.sound.av_player.players)
	mpvManager = anki.sound.MpvManager(mw.pm.base)
	anki.sound.av_player.players[0] = mpvManager
	print("After:", anki.sound.av_player.players)

If it can be somehow useful, I've attached a zip archive with a few mp3 audio files in `media` subfolder (~100MB) from YouTube channel Every Frame a Painting - https://www.youtube.com/user/everyframeapainting - and two simple python scripts to illustrate the problem and possible workaround.

How to reproduce:

1. Start mpv with input-ipc-sever on Windows

`start "" mpv --idle --osd-status-msg="${filename} - ${time-pos}" --osd-level=3 --input-ipc-server=\\.\pipe\ankimpv`

2. `python mpv-ipc-test.py`

3. press Enter to tell mpv to load a new audio file

4. after pressing Enter ~7-10 times mpv will stop responding and loading any new audio file

5. press Ctrl+C to interrupt the python script

Strangely, at that moment the last file from the console output message will be loaded and played by mpv

6. rerun `python mpv-ipc-test.py` and after ~7-10 times it'll happen again

7. run `python mpv-ipc-workaround.py` and audio will be loaded every time when Enter is pressed

[#621-mpv-ipc-media.zip](https://drive.google.com/open?id=1twYhthou0WXn9eMOdBHMibblftFZEAIn)

[#621-mpv-ipc-scripts.zip](https://github.com/ankitects/anki/files/4647052/621-mpv-ipc-scripts.zip) or https://gist.github.com/kelciour/501a280a63d26e793ed5bfa7a46596d4

ANH25:
@kelciour
I got the same result. Also, reading messages from the pipe after sending the `loadfile` command seems to solve the issue without having to call `win32file.CreateFile` again.

```python
result = 0
buf = b""
while result == 0:
    try:
        (result, b) = win32file.ReadFile(_sock, 4096)
        buf += b
    except pywintypes.error as err:
        if err.args[0] == winerror.ERROR_NO_DATA:
            break
```

dae:
If someone can submit a patch that fixes the deadlock issue, I'd be happy to trial switching back to mpv on Windows

evandrocoan:
Why switch back? Unless `mplayer` is not working, but I rather have `mpv` just as a fallback in case `mplayer` fails.

dae:
mpv is used on Mac+Linux, is more actively maintained than mplayer, and doesn't require a write-a-temp-file hack to feed unicode filenames to it when in slave mode. Anki can fall back on mplayer if mpv is unavailable, as it does on other platforms.

kelciour:
Thanks @ANH25, just tested and it's working great, the pipe is no longer blocked, mpv is no longer unresponsive and can play any number of new audio files without any issue.

I have another possible problem with mpv and mpv.py at least on Windows.

  Exception in thread Thread-1791:
    File "threading.py", line 932, in _bootstrap_inner
    File "threading.py", line 870, in run
    File "aqt\mpv.py", line 492, in _event_reader
    File "aqt\mpv.py", line 502, in _handle_event
  AttributeError: 'MpvManager' object has no attribute '_callbacks'

It doesn't seem to happen very often. I got it once when I used Debug Console to create mpv process and replace mplayer with mpv.

It seems to be already fixed in python-mpv with this commit from a few months ago - https://github.com/gustaebel/python-mpv/commit/c0d954481223ce22e658ec2f9b0aa610a0392984 ("Queue events until the callbacks are installed")

kelciour:
After googling about blocking pipes, it looks like I did a mistake and my python test script was incorrect. It's no wonder that a pipe got blocked since I only wrote and never read from the pipe.

And mpv.py shouldn't have the same problem since it uses a separate thread to read from the pipe  if there's some data or sleep for 0.1 if there's no data, and only throws MPVTimeoutError if there were no response for 1 second after sending a new message.

I found an old thread on the support forum at the time when Anki 2.1.0 beta switched back from mpv to mplayer on Windows. It's been related to `<class 'anki.mpv.MPVTimeoutError'>: unable to get response`.

https://anki.tenderapp.com/discussions/ankidesktop/26415-error-after-starting

I'm not sure how to reproduce it and so far mpv works fine for me after testing it with Anki on Windows.

ANH25:
I also have been using mpv with Anki on Windows 7 since opening this PR with no problems except getting this AttributeError exception one or two times, especially when running from source.

dae:
Your PR was conflicting so I've applied the change manually - will leave this open as a reminder that I need to include mpv in the bundle again. Fingers crossed it works better this time!

https://github.com/ankitects/anki/commit/d04699ad06d7489be5d5f184129475f652a0f7d6

dae:
2.1.28 Alpha 2 bundles mpv - please let me know if you encounter any issues.

evandrocoan:
`mpv` seems a better option as the user on https://forums.ankiweb.net/t/how-to-embed-mp4-files/264/16 could only play videos after switching to `mpv`.

There is also the problems `mplayer` causes as in https://anki.tenderapp.com/discussions/ankidesktop/42018-sound-playing-after-10sec-or-more

PR #622, opened 2020-05-14
Allow addons to disable the progress bar when the have their own

evandrocoan:
https://github.com/kelciour/awesometts-anki-addon/pull/24 - Disables the Anki progress bar before using our progress bar

### DIFF
--- a/qt/aqt/progress.py
+++ b/qt/aqt/progress.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import time
+import weakref
 from typing import Optional
 
 import aqt.forms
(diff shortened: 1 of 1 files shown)


dae:
What Anki code is the add-on calling that is displaying the progress bar?

evandrocoan:
`browser.model.beginReset()/endReset()`: https://github.com/kelciour/awesometts-anki-addon/pull/24/files

dae:
Rather than add a hack to disable the progress bar of routines that create a progress bar, wouldn't it be clearer to modify the add-on to not call beginReset/endReset? It can just call .reset() at the end to trigger a browser refresh

evandrocoan:
I do not know why it was calling these things, or that it could just call `.reset()` at the end. I will update the addons accordingly.

PR #623, opened 2020-05-18
Remove unrequired qtminor assertion == 10

evandrocoan:


dae:
The current code is correct - we support 5.9 (an LTS release), but 5.10 broke Anki.

PR #624, opened 2020-05-19
Deprecated pacman usage because its is failing out of nowhere,

evandrocoan:
Deprecated pacman usage because its is failing out of nowhere, probably a server is down. Anyways, directly unziping rsync is faster. And soon, we will have builtin support for msys2:
https://github.com/actions/virtual-environments/pull/632 - Install MSYS2 to Windows


dae:
Thanks!

Just a minor English note: deprecate means to disapprove of  - eg adding a warning that something is bad and will be removed in the future.

PR #625, opened 2020-05-19
Fixed clayout.py rendering the contents twice when clicked

evandrocoan:
> The `toggle()` signal is emitted every time *any* radio button's state changes.  As a result, the `toggle()` signal is emitted when you click on a radio button and that radio button's state changes from unchecked to checked, and if clicking on the radio button automatically unchecks another radio button, then the `toggle()` signal is emitted again because the other radio button's state changes from checked to unchecked. 
> 
> https://stackoverflow.com/questions/36808257/pyqt-radiobutton-ischecked-is-executed-twice


dae:
Thanks!

PR #627, opened 2020-05-19
Added missing Anki icon to qt/aqt/previewer.py

evandrocoan:


dae:
Thanks!

For future reference, Anki prefers importing from aqt.qt instead, as that will make things easier if we ever need to move to Qt for Python.

PR #631, opened 2020-05-20
Fixed hooks.py:361: error: Name 'anki.storage._Collection' is not

evandrocoan:
Fixed hooks.py:361: error: Name 'anki.storage._Collection' is not defined [name-defined]
https://github.com/ankitects/anki/runs/691972681#step:25:925


dae:
Fixed in https://github.com/ankitects/anki/commit/e2bb5cd14a0979cc999535f3f68518a090d718f1

PR #632, opened 2020-05-20
It should close the HttpClient (which holds the session object)

evandrocoan:
I confused the name of the objects on my old pull request (https://github.com/ankitects/anki/pull/630).

But after looking, both objects have a close method.


dae:
I think https://github.com/ankitects/anki/commit/a3ac96443bce9beaaf31e8f955cf334f893686ee is a better approach than manually calling close, given you added enter/exit to HttpClient.

PR #633, opened 2020-05-22
Add three deck config hooks

hgiesel:
These PR will add three `gui_hooks` for working inside of the the deck config window. These are:

- `gui_hooks.deck_conf_did_add_config`
- `gui_hooks.deck_conf_will_remove_config`
- `gui_hooks.deck_conf_will_rename_config`

These are helpful, when you want to offer your add-on settings inside of the deck_conf window, and want them to behave correctly when the user utilizes these functionalities.

`did_add_config` is helpful because you have the new name and id, but you also have access to the previous settings. This allows for easily copying the previous settings to the new name or id (and thus behaving like Anki does when adding a new config group).

I specifically would use them in my addon [Straight Rewards](https://ankiweb.net/shared/info/957961234).

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -33,6 +33,7 @@ Henry Tang <hktang@ualberta.ca>
 Simone Gaiarin <simgunz@gmail.com>
 Rai (Michal Pokorny) <agentydragon@gmail.com>
 Zeno Gantner <zeno.gantner@gmail.com>
+Henrik Giesel <hengiesel@gmail.com>
 
 ********************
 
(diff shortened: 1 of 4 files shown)


evandrocoan:
@hgiesel, Please run `make fix` and commit the formatting fixes: https://github.com/ankitects/anki/pull/633/checks?check_run_id=700637499#step:25:507

hgiesel:
@evandrocoan I think this will be solved in this commit (788a3ee), I was just a bit confused first by the error messages. I also added a little more documentation for the hooks.

dae:
Thanks!

PR #634, opened 2020-05-24
don't install virtualenv

embe:
virtualenv is not needed, as the built-in venv is actually used.


evandrocoan:
If I recall correctly, it was required to install it on my Python (Windows). What system are you using?

embe:
I use Gentoo Linux, where the pip install fails (as it could interfere with system packages).

I don't think it's likely that installation of `virtualenv` would change anything, even on Windows – nothing in the code uses `virtualenv` (only `venv`, the built-in version). Maybe something changed in the rest of the code in the meantime?
The CI checks on Windows also seem to pass with my change.

evandrocoan:
If it is causing Gentoo to fail running anki, then, it should be removed because if windows users miss the package, it can be installed before running `make` with `pip install virtualenv`.

dae:
Thanks, let's give it a go.

PR #636, opened 2020-05-26
add evt type for contextMenuEvent method

BlueGreenMagick:
https://doc.qt.io/qt-5/qwidget.html#contextMenuEvent


evandrocoan:
@BlueGreenMagick, can you check the errors on Mac OS? https://github.com/ankitects/anki/pull/636/checks?check_run_id=708581630#step:25:493

evandrocoan:
It seems a bug on Anki. I just did a new build from the actual master the still the same: https://github.com/evandroforks/anki/runs/710302038?check_suite_focus=true#step:25:925

PR #637, opened 2020-05-26
fix drag drop field repositioning

BlueGreenMagick:
https://anki.tenderapp.com/discussions/ankidesktop/41822-drag-dropping-fields-in-the-field-list-dialog-off-by-one


dae:
Thanks for that :-)

PR #641, opened 2020-05-28
Add main_window_did_init hook

glutanimate:
Adds a single-shot hook that's executed once upon full initialization of Anki's main window and clarifies its use cases compared to `profile_did_open`.

Sample add-on:

```python
from aqt import gui_hooks

def on_main_window_did_init():
    print("main_window_did_init")

gui_hooks.main_window_did_init.append(on_main_window_did_init)
```


dae:
Cheers, I've updated the docs as well

glutanimate:
Thanks, Damien!

PR #642, opened 2020-05-28
Say which card failed with an exception

evandrocoan:


dae:
Probably the card id instead of the full repr is sufficient

evandrocoan:
Just showing the id is difficult to know right away about which card. I do not see a reason to not show the full card on an exception stack trace.

dae:
Can't say it's a frequent error, and don't want to waste time bikeshedding further

PR #643, opened 2020-05-28
Added __repr__ functions to common objects

evandrocoan:
It helps to see and understand anki inner workings when debugging and printing things.

### DIFF
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -193,7 +193,7 @@ class Card:
         del d["_render_output"]
         del d["col"]
         del d["timerStarted"]
-        return pprint.pformat(d, width=300)
+        return f"{super().__repr__()} {pprint.pformat(d, width=300)}"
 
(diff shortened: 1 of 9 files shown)


dae:
IMHO most of these are not really necessary - printing things like tags and the media manager is rare in my experience, and with models.py you're going to presumably get swamped by _cache. Notes/cards I think are definitely useful (and there's some code in the debug console that does something similar already, though I'm still not entirely happy with it) - maybe we can focus on those to start and add more as the need arises? It's easy to print obj.__dict__ directly for rare stuff.

evandrocoan:
While they do not seem necessary, I added them all when trying to understand how Anki object model is composed. Meaning they are quite useful to anyone starting to contribute to Anki understand what Anki objects are doing. I do not see a reason for not having them around and it is quite annoying to add them everywhere.

dae:
Will add them for now and see how they work out in practice

PR #644, opened 2020-05-29
Added support to syntax highlighting on clayout.py text editor

evandrocoan:
There is some time I was thinking about adding this, then, after seeing the request on https://anki.tenderapp.com/discussions/ankidesktop/41711-resize-windows-in-template-editor-syntax-highlighting, I decided to implement it.

In case someone has problems with it, it can be disabled in settings. Also, if the code fails to import the dependency or there is an error somehow it disables itself, i.e., it just does nothing, and the original code is run instead.

1. image
1. image
1. image

Example code used:
<details>
<p>

```python
import sys

# https://stackoverflow.com/questions/22411463/how-to-use-the-autocomplete-feature-with-qscintilla
# http://getfr.org/pub/dragonfly-release/usr-local-share/doc/qscintilla2/html/hierarchy.html
# https://www.scintilla.org/ScintillaDoc.html#Autocompletion
# https://stackoverflow.com/questions/42329818/how-to-use-autocomplete-using-qscintilla-and-c
# https://stackoverflow.com/questions/40002373/qscintilla-based-text-editor-in-pyqt5-with-clickable-functions-and-variables
from PyQt5.QtWidgets import *
from PyQt5.QtCore import *
from PyQt5.QtGui import *
try:
    from PyQt5 import Qsci
except:
    Qsci = None
class SimplePythonEditor(Qsci.QsciScintilla):

    def setAcceptRichText(self, *args, **kwargs):
        pass

    def setTabStopDistance(self, *args, **kwargs):
        pass

    def setPlainText(self, text):
        self.setText(text)

    def toPlainText(self):
        return self.text()

    def __init__(self, parent=None):
        super().__init__(parent)
        font = QFont()
        font.setFamily('Courier')
        font.setFixedPitch(True)
        font.setPointSize(10)
        self.setFont(font)
        self.setMarginsFont(font)

        fontmetrics = QFontMetrics(font)
        self.setMarginsFont(font)

        # Margin 0 is used for line numbers
        self.setMarginWidth(0, fontmetrics.width("0000"))
        self.setMarginWidth(1, 0)
        self.setMarginLineNumbers(0, True)
        self.setMarginsBackgroundColor(QColor("#cccccc"))

        # Clickable margin 1 for showing markers
        self.setMarginSensitivity(1, True)

        # Brace matching: enable for a brace immediately before or after
        # the current position
        self.setBraceMatching(Qsci.QsciScintilla.SloppyBraceMatch)

        # Current line visible with special background color
        self.setCaretLineVisible(True)
        self.setCaretLineBackgroundColor(QColor("#ffe4e4"))

        lexer = Qsci.QsciLexerHTML()
        lexer.setDefaultFont(font)
        self.setLexer(lexer)

        text = bytearray(str.encode("Courier"))
        self.SendScintilla(Qsci.QsciScintilla.SCI_STYLESETFONT, 1, text)
        self.setWrapMode(Qsci.QsciScintilla.WrapWord)
        self.setEolMode(Qsci.QsciScintilla.EolUnix)
        self.setWrapIndentMode(Qsci.QsciScintilla.WrapIndentSame)

        self.setAutoCompletionCaseSensitivity(False)
        self.setAutoCompletionReplaceWord(False)
        self.setAutoCompletionSource(Qsci.QsciScintilla.AcsDocument)
        self.setAutoCompletionThreshold(1)

if __name__ == "__main__":
    app = QApplication(sys.argv)
    editor = SimplePythonEditor()
    editor.show()
    # editor.setText(open(sys.argv[0]).read())
    editor.setText(
"""
<div class="fixBorders">
What is the <b>past simple</b> of the verb to <b>{{Verb}}</b>?</span><br>

<span style="display: flex; align-items: center;">
    <input type="button" value="x1.2" onclick="setAnkiMedia( media => { media.playbackRate = 1.2; } )" />
    <input type="button" value="x1.0" onclick="setAnkiMedia( media => { media.playbackRate = 1.0; } )" />
    <input type="button" value="x0.6" onclick="setAnkiMedia( media => { media.playbackRate = 0.6; } )" />
    <audio data-file="{{VerbAudio}}" controlsList="nodownload" controls></audio>
</span>

{{type:PastSimple}}
</div>

ankimedia.setup();
ankimedia.add( 'front', '{{VerbAudio}}', 1.2 );
</script>
""" )
    app.exec_()
```
</p>
</details>

### DIFF
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -283,11 +283,15 @@ class CardLayout(QDialog):
 
         self.fill_fields_from_template()
 
-    def on_search_changed(self, text: str):
+    def on_search_changed(self, text: str, selectNext=False):
         editor = self.tform.edit_area
         from aqt import utils as aqtUtils
 
         if aqtUtils.Qsci and aqtUtils.QsciEnabled:
+            if not selectNext:
+                line_from, index_from, line_to, index_to = editor.getSelection()
+                editor.setSelection(line_from, index_from, line_from, index_from)
+
             if not editor.findFirst(text, True, False, False, True):
                 tooltip("No matches found.")
         else:
@@ -301,7 +305,7 @@ class CardLayout(QDialog):
 
     def on_search_next(self):
         text = self.tform.search_edit.text()
-        self.on_search_changed(text)
+        self.on_search_changed(text, selectNext=True)
 
     def setup_preview(self):
         pform = self.pform
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -194,6 +194,10 @@ class CardLayout(QDialog):
         split.addWidget(left)
         split.setCollapsible(0, False)
 
+        from aqt.utils import setupSyntaxHighlighter
+
+        setupSyntaxHighlighter(tform, "edit_area", "verticalLayout")
+
         right = QWidget()
         self.pform = aqt.forms.preview.Ui_Form()
         pform = self.pform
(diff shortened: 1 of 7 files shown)


dae:
Hi Evandro,

I think syntax highlighting would be a nice usability win, but I'd like to get a better idea of the options available before committing to QScintilla. Have you looked into available options in the Javascript world?

scanf:
A cool project in the JavaScript ecosystem worth taking a look at is the monaco editor used in Visual Studio Code, https://microsoft.github.io/monaco-editor/

evandrocoan:
I did not think about using Javascript, but it seems certainly better than using any QT library. I will look into how it would be using a Javascript library.

ijgnd:
There's also the js code editor [codemirror](https://codemirror.net/). It was pretty simple to put it into an [add-on of mine](https://ankiweb.net/shared/info/1043915942) that creates a custom extended html editor for fields.

dae:
Sorry Evandro, didn't notice you'd edited your comment. Monaco does look too heavyweight, and doesn't appear to support mobile - something we can use across all the clients would be a better choice. Codemirror looks good - it seems to be quite popular, and does a better job of handling touch events on mobile than Ace does.

evandrocoan:
I will put this on hold because QScintilla is working fine for my day-to-day usage. Despite it not being a very nice editor, it's enough to have some colors while QScintilla is written in C++ which give high performance (which may not be always required).

@ijgnd if you already have addons doing this (https://github.com/ijgnd/anki__editor_edit_html_extended/blob/master/src/codemirror.py), you could use the same code or part of it to open a pull-request with codemirror replacing this one (with QScintilla).

evandrocoan:
This QScintilla is actually quite bad. A javascript solution should be used instead.

dae:
Thanks for the update :-)

PR #647, opened 2020-05-31
Adding hooks to TypeScript

Arthur-Milchior:
While looking at reviewer code, I saw there were hooks in it. I thought that adding hooks in editor might be nice too. I've many add-ons dealing with editor, and hooks would be really helpful to simplify the add-on code and improve compatibility.

I am not expecting this PR to be accepted as is, I would expect other devs to give feedback if they are interested in this feature.

### DIFF
--- a/qt/.gitignore
+++ b/qt/.gitignore
@@ -19,6 +19,7 @@ aqt_data/web/overview.js
 aqt_data/web/reviewer-bottom.js
 aqt_data/web/reviewer.js
 aqt_data/web/webview.js
+aqt_data/web/hooks.js
 dist
 aqt.egg-info
 build
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -192,7 +192,7 @@ class Editor:
         self.web.stdHtml(
             _html % (bgcol, bgcol, topbuts, _("Show Duplicates")),
             css=["editor.css"],
-            js=["jquery.js", "editor.js"],
+            js=["jquery.js", "hooks.js", "editor.js"],
             context=self,
         )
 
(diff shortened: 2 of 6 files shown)


evandrocoan:
@Arthur-Milchior, I think `qt/aqt_data/web/hooks.js` should be added to `qt/.gitignore` as it is created from `qt/ts/src/editor.ts` when typescript builds things.

Arthur-Milchior:
Thanks. I tried to state that js is to be ignored in this folder more generally. Seems more general

evandrocoan:
@Arthur-Milchior, `aqt_data` ships by default a bunch of `*.js` files (260 .js files) while only those 7 files should be ignored. It will be very bad to ignore all `*.js` files because when some of those default shipped libraries are updated, their files are not going to be seen (additions/etc).

As we need to ignore only 6 files, we should ignore each of the explicitly as it was being done before: (**qt\.gitignore**)
```
aqt_data/web/deckbrowser.js
aqt_data/web/editor.js
aqt_data/web/overview.js
aqt_data/web/reviewer-bottom.js
aqt_data/web/reviewer.js
aqt_data/web/webview.js
aqt_data/web/hooks.js
```

Instead of ignoring all `*.js` files, you should pay more attention before committing things. I suggest using some `git` client interface and do not use things as `git add --all` to add your changes. I use this git interface if you are interested:
image

Arthur-Milchior:
Thanks for the feedback. I didn't realize it was a recursive ignore here. change made

dae:
I probably won't be able to return to this until the backend refactoring is done, but just a code style comment for now - while I appreciate you trying to follow the Python code style, camelCaps is the standard in JavaScript

Arthur-Milchior:
Variable name changed.

dae:

PR #648, opened 2020-05-31
Print qt context if it exists

evandrocoan:
Qt won't give us the context unless it was compiled with `QT_MESSAGELOGCONTEXT`, but in case we got a qt compiled with `QT_MESSAGELOGCONTEXT`, print the function context: 
1. https://anki.tenderapp.com/discussions/ankidesktop/42070-anki-closes-without-warning-when-importing-conflicting-shared-deck
1. [How to debug an installed Qt5 library with GDB?](https://unix.stackexchange.com/questions/202374/how-to-debug-an-installed-qt5-library-with-gdb)
1. https://stackoverflow.com/questions/35894171/redirect-qdebug-output-to-file-with-pyqt5
1. https://doc.qt.io/qt-5/qtglobal.html#qInstallMessageHandler

### DIFF
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -325,7 +325,23 @@ def setupGL(pm):
         ctypes.CDLL("libGL.so.1", ctypes.RTLD_GLOBAL)
 
     # catch opengl errors
-    def msgHandler(type, ctx, msg):
+    def msgHandler(category, ctx, msg):
+        if category == QtDebugMsg:
+            category = "debug"
+        elif category == QtInfoMsg:
+            category = "info"
+        elif category == QtWarningMsg:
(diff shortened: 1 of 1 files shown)


dae:
Compiling Qt is a real pain, and if this is only giving us some extra debugging info in the case that the video driver fails to load, I'm not sure what it's actually buying us?

evandrocoan:
This runs every time any QT message is showed, not only for card drivers. The problem with the old code is that even if we run a QT version with debug mode, it will not show us the context. With this, if QT is compiled with `QT_MESSAGE_PATTERN` flag, then, it will show us the context/from where the messages are coming.

dae:
I'm still doubtful it will actually be useful in practice :-)

PR #649, opened 2020-05-31
Remove code duplication _toBury

Arthur-Milchior:
The methods _toBury have a lot of code duplicated. I would greatly appreciate if you'd accept to consider removing some of it. Not necessarily this exact way.

The reason behind this PR is that, for speed efficiency, I prefetch the next card that will be sent to the reviewer. This mean that I need to change the code that does the actual burying. This change is quite simpler if there is a single method than if the change need to be done four time. Furthermore, since ankidroid tries to stay very close to Anki, having the change accepted here would help keeping AnkiDroid close while avoiding having more code duplication.

### DIFF
--- a/pylib/anki/sched.py
+++ b/pylib/anki/sched.py
@@ -864,7 +864,7 @@ update cards set queue={QUEUE_TYPE_SIBLING_BURIED},mod=?,usn=? where id in """
     ##########################################################################
 
     def _burySiblings(self, card: Card) -> None:
-        toBury = []
+        toBury: List[int] = []
         nconf = self._newConf(card)
         buryNew = nconf.get("bury", True)
         rconf = self._revConf(card)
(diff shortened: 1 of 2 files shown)


Arthur-Milchior:
I actually don't understand why _burySiblings is distincts in sched and schedv2. It seems they do the same thing. Except that in sched, the burying and logging is done directly while in schedv2 it is delegated to `buryCards`.
In https://github.com/Arthur-Milchior/anki/tree/remove_burying_sched I removed this duplication.
I'm not entirely sure of myself, because the logging function uses the backstack, and it may be important to you to keep the logging exactly as it is right now. But if you would consider this branch, I believe it would improve code to incorporate it into anki.

dae:
Just an oversight. I didn't find actualBury to be clearer, but I've pushed something similar.

Arthur-Milchior:
Thanks

Arthur-Milchior:
I do admit that your version is quite clearer. I do prefer your choice

dae:
This introduced a regression: https://github.com/ankitects/anki/commit/5729e9e33645821798dd804e1b173ceab083cbbd

Arthur-Milchior:
Thank you very much for the notification and for not simply reverting it. This commit was ported to AnkiDroid's PR so we'll avoid it

PR #650, opened 2020-05-31
Remember last inputs for Find and Replace and Duplicates Find

evandrocoan:
https://anki.tenderapp.com/discussions/ankidesktop/39468-find-and-replace-does-not-remember-the-input-from-last-time

> I did a find and replace, in a single card (just for testing to see if things go right), but later when I was going to do it with all cards, it did not remember my last regular expression the input and replacement value.

.png)

### DIFF
--- a/qt/aqt/utils.py
+++ b/qt/aqt/utils.py
@@ -470,6 +470,28 @@ def restoreState(widget, key: str):
         widget.restoreState(aqt.mw.pm.profile[key])
 
 
+def saveSplitter(widget, key):
+    key += "Splitter"
+    aqt.mw.pm.profile[key] = widget.saveState()
+
+
+def restoreSplitter(widget, key):
+    key += "Splitter"
+    if aqt.mw.pm.profile.get(key):
+        widget.restoreState(aqt.mw.pm.profile[key])
+
+
+def saveHeader(widget, key):
+    key += "Header"
+    aqt.mw.pm.profile[key] = widget.saveState()
+
+
+def restoreHeader(widget, key):
+    key += "Header"
+    if aqt.mw.pm.profile.get(key):
+        widget.restoreState(aqt.mw.pm.profile[key])
+
+
 def save_is_checked(widget, key: str):
     key += "IsChecked"
     aqt.mw.pm.profile[key] = widget.isChecked()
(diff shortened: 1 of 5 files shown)


dae:
The case and regex options probably make sense to persist across restarts, but the other items might make more sense to forget when the program is closed - that way you still get the benefits of keeping the text in situations like your "test on one card" example, but without text from potentially months back appearing when the user opens find&replace for the first time in a while. Thoughts? Anyone else want to chime in?

dae:
(I foresee users forgetting their old input, then reporting a "bug" where strange text or a specific field is selected by default)

evandrocoan:
> (I foresee users forgetting their old input, then reporting a "bug" where a strange text is selected by default)

I just created an attribute on `profile.py`  called `session` for things that should not last between restarts and put the fields indexes remembering in there. So after restarting Anki, it will forget the field indexes.

> (I foresee users forgetting their old input, then reporting a "bug" where a specific field is selected by default)

I just set to it not to fill any fields by default. So after the user restarts Anki, it will see not text filling automatically. But it will automatically fill the text if the use reopens the `find and replace` dialogs on the same session (without an Anki restart).

qt/aqt/utils.py
```diff
@@ -459,17 +459,71 @@ def ensureWidgetInScreenBoundaries(widget):
...
 
 
-def restoreState(widget, key):
+def restoreState(widget, key: str):
     key += "State"
     if aqt.mw.pm.profile.get(key):
         widget.restoreState(aqt.mw.pm.profile[key])
 
 
+def saveIsChecked(widget, key: str):
```
dae, line 473: As these are new functions, please use snake_case

qt/aqt/utils.py
```diff
@@ -459,17 +459,71 @@ def ensureWidgetInScreenBoundaries(widget):
...
+    indexKey = key + "ComboActiveIndex"
+    text = aqt.mw.pm.session.get(textKey)
+    index = aqt.mw.pm.session.get(indexKey)
+    if text is not None and index is not None:
+        if index < len(history) and history[index] == text:
+            widget.setCurrentIndex(index)
+
+
+def saveComboHistory(comboBox: QComboBox, history: List[str], name: str):
+    name += "BoxHistory"
```
dae, line 502: - Since we now have a mix of routines that save permanently and save for the session, perhaps it would be better to encode that in the naming, eg "save_combo_index_for_session"
- Since these free-functions are just modifying state in the pm object, perhaps it makes more sense for them to be placed inside that class, or in a separate SessionState class accessible from mw that gets recreated on profile load.

evandrocoan: Only `save_combo_index` saves on the `session` object, `save_combo_history` uses the `session` object as a hint to not fill the results when the user has restarted Anki. Unless we have a specific problem to solve (requiring [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns)), a dictionary attribute on ProfileManager should be more than enough.

We could move these functions into the `Profile` object as class members, but only if we also move the existing functions as `saveState` and `saveSplitter` with it because they do the same things so it will be very nice for new developers to find all of them in a single place.

dae: Please at least add for_session for the one routine that doesn't use the profile, as it's confusing otherwise.

dae:
Thanks Evandro.

PR #651, opened 2020-06-01
Allowed to switch with the last view with Ctrl+Tab

evandrocoan:
I did not implement a full MRU (Most recently used) thing because it would be [more complex](https://github.com/spyder-ide/spyder/pull/4302) and we just have 3 views to switch so just selecting the next on this hardcoded MRU should be more than enough.

### DIFF
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -159,10 +159,40 @@ class CardLayout(QDialog):
             s += "+..."
         return s
 
+    def nextTab(self):
+        """ Fake implementation for MRU which just switches between the last 2 views
+        For a full implementation example, see: https://github.com/spyder-ide/spyder/pull/4302/files
+        """
+        if self.current_editor_index == 0:
+            if self.last_editor_index == 0:
+                self.tform.back_button.setChecked(True)
+            elif self.last_editor_index == 1:
+                self.tform.back_button.setChecked(True)
+            elif self.last_editor_index == 2:
+                self.tform.style_button.setChecked(True)
+        elif self.current_editor_index == 1:
+            if self.last_editor_index == 0:
+                self.tform.front_button.setChecked(True)
+            elif self.last_editor_index == 1:
+                self.tform.style_button.setChecked(True)
+            elif self.last_editor_index == 2:
+                self.tform.style_button.setChecked(True)
+        elif self.current_editor_index == 2:
+            if self.last_editor_index == 0:
+                self.tform.front_button.setChecked(True)
+            elif self.last_editor_index == 1:
+                self.tform.back_button.setChecked(True)
+            elif self.last_editor_index == 2:
+                self.tform.front_button.setChecked(True)
+        self.on_editor_toggled()
+
     def setupShortcuts(self):
         self.tform.front_button.setToolTip(shortcut("Ctrl+1"))
         self.tform.back_button.setToolTip(shortcut("Ctrl+2"))
         self.tform.style_button.setToolTip(shortcut("Ctrl+3"))
+        QShortcut(  # type: ignore
+            QKeySequence("Ctrl+Tab"), self, activated=self.nextTab,
+        )
         QShortcut(  # type: ignore
             QKeySequence("Ctrl+1"), self, activated=self.tform.front_button.click,
         )
@@ -218,6 +248,7 @@ class CardLayout(QDialog):
         qconnect(tform.style_button.clicked, self.on_editor_toggled)
 
         self.current_editor_index = 0
+        self.last_editor_index = 0
         self.tform.edit_area.setAcceptRichText(False)
         self.tform.edit_area.setFont(QFont("Courier"))
         if qtminor < 10:
@@ -250,6 +281,7 @@ class CardLayout(QDialog):
         self._renderPreview()
 
     def on_editor_toggled(self):
+        self.last_editor_index = self.current_editor_index
         if self.tform.front_button.isChecked():
             self.current_editor_index = 0
             self.pform.preview_front.setChecked(True)


dae:
Hi Evandro,

Any ideas about the current build failure? Looks like it's pulling in an older PyQt instead of grabbing the latest on pip.

Regarding this PR, why not just use ctrl+1/2/3? That way you always know which tab will appear

evandrocoan:
> Regarding this PR, why not just use ctrl+1/2/3? That way you always know which tab will appear

When pressing `Ctrl+1/2/3` keys I have my `littler finger` and my `middle finger`, which is not much comfortable when I have to change several times between the same 2 tabs.

When pressing `Ctrl+Tab`, I can just use my `little finger` and my `ring finger`, which is much easier to use/access then the `middle finger`. Also, is it much simpler to keep using `Ctrl+Tab` to select between the last tab and my actual tab.

As I do not need to keep remembering whether I should keep using `Ctrl+1 & Ctrl+2` or `Ctrl+2 & Ctrl+3` or `Ctrl+1 & Ctrl+3`, because I can just keep pressing `Ctrl+Tab` (single key combination other than 3 keys combinations which keeps varying accordingly to what I am doing). 

Meaning, `Ctrl+Tab` keeps a "memory" history of my last activity I am doing, freeing me from having to remember it myself (of the keys combination `Ctrl+1 & Ctrl+2` or `Ctrl+2 & Ctrl+3` or `Ctrl+1 & Ctrl+3`).

> Any ideas about the current build failure? Looks like it's pulling in an older PyQt instead of grabbing the latest on PyPI.

It is not instantly pyqt 5.15 because we have forced pyqt 5.13.2 on:
```
File: anki/qt/requirements.qt
1: pyqt5==5.13.2
2: pyqtwebengine==5.13.2
```

If the checks are passing on your computer, it means you probably have the old pyqt stubs laying around somewhere. Try running `git clean -xfd .` on the root of your Anki repository. It will clean 100% of all files not versioned by git.

I checkout on the new master and run `git clean -xfd .` and I have the same errors as the `CI` machines. even after changing things to:
```
File: anki/qt/requirements.qt
1: pyqt5==5.15.0
2: pyqtwebengine==5.15.0
```

I managed to reduce the errors by installing the package `python -m pip install pyqt5-stubs`. Now I got only these errors:
```java
python -m mypy  aqt
aqt\qt.py:16: error: Skipping analyzing 'PyQt5.QtWebEngineWidgets': found module but no type hints
or library stubs  [import]
    from PyQt5.QtWebEngineWidgets import *
    ^
aqt\pinnedmodules.py:23: error: Skipping analyzing 'PyQt5.QtSvg': found module but no type hints
or library stubs  [import]
    import PyQt5.QtSvg
    ^
aqt\pinnedmodules.py:23: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
aqt\tagedit.py:13: error: Incompatible types in assignment (expression has type
"Union[QCompleter, TagCompleter]", base class "QLineEdit" defined the type as
"Callable[[QLineEdit], QCompleter]")  [assignment]
        completer: Union[QCompleter, TagCompleter]
        ^
aqt\main.py:101: error: Unsupported operand types for & ("KeyboardModifiers" and
"KeyboardModifier")  [operator]
            self.safeMode = self.app.queryKeyboardModifiers() & Qt.ShiftModifier
                            ^
aqt\main.py:964: error: Incompatible types in assignment (expression has type "List[QShortcut]",
variable has type "Sequence[Tuple[str, Callable[..., Any]]]")  [assignment]
            self.stateShortcuts = self.applyShortcuts(shortcuts)
                                  ^
aqt\main.py:968: error: Argument 1 to "delete" has incompatible type
"Tuple[str, Callable[..., Any]]"; expected "simplewrapper"  [arg-type]
                sip.delete(qs)
                           ^
aqt\webview.py:24: error: Name 'QWebEnginePage' is not defined  [name-defined]
    class AnkiWebPage(QWebEnginePage):
                      ^
aqt\webview.py:186: error: Name 'QWebEngineView' is not defined  [name-defined]
    class AnkiWebView(QWebEngineView):
                      ^
aqt\webview.py:190: error: Name 'QWebEngineView' is not defined  [name-defined]
            QWebEngineView.__init__(self, parent=parent)
            ^
aqt\webview.py:203: error: Name 'QWebEngineProfile' is not defined  [name-defined]
            self._page.profile().setHttpCacheType(QWebEngineProfile.NoCache)
                                                  ^
aqt\webview.py:235: error: "QEvent" has no attribute "button"  [attr-defined]
                if evt.button() == Qt.MidButton and isLin:
                   ^
aqt\deckchooser.py:21: error: "Callable[[], QWidget]" has no attribute "setLayout"  [attr-defined]
            self.widget.setLayout(self)
            ^
aqt\browser.py:1052: error: "SidebarTreeView" has no attribute "mw"  [attr-defined]
            self.sidebarTree.mw = self.mw
            ^
aqt\studydeck.py:80: error: "QEvent" has no attribute "key"  [attr-defined]
                if evt.key() == Qt.Key_Up:
                   ^
aqt\studydeck.py:87: error: "QEvent" has no attribute "key"  [attr-defined]
                elif evt.key() == Qt.Key_Down:
                     ^
Found 15 errors in 8 files (checked 96 source files)
make[1]: *** [Makefile:107: .build/mypy] Error 1
make[1]: Leaving directory '/cygdrive/f/anki/qt'
make: *** [Makefile:154: check] Error 2

f:\anki>
```

dae:
Thanks Evandro - I should have looked into it further before assuming it was a caching issue. I'd forgotten the Qt version was pinned!

I've pushed some other tweaks that should mean we don't need the separate -stubs package.

qt/aqt/clayout.py
```diff
@@ -159,10 +159,40 @@ def _fieldsOnTemplate(self, fmt):
             s += "+..."
         return s
 
+    def nextTab(self):
+        """ Fake implementation for MRU which just switches between the last 2 views
+        For a full implementation example, see: https://github.com/spyder-ide/spyder/pull/4302/files
+        """
+        if self.current_editor_index == 0:
+            if self.last_editor_index == 0:
+                self.tform.back_button.setChecked(True)
```
dae, line 168: Is this complexity necessary? Won't current_editor_index == last_editor_index only be the case before the user has switched between any tabs?

evandrocoan: No, if the user pressed `Ctrl+1` or clicked in the first tag twice in a row, the last index will be equal to the current index. As we are hardcoding this small "memory", it is simple to just list all cases because if someday we actually replace it by a full MRU algorithm, it should be simpler.

evandrocoan:
I am going to rebase this branch with the master to rerun the tests.

qt/aqt/clayout.py
dae: Couldn't that be addressed with less code by not changing last_editor_index when the user clicks on the same tab again?

evandrocoan: I would not call these lines code, but a combination matrix. If we added that if on `last_editor_index`, we could remove one `inner` combination from each `outter` combination, but if somehow our `last_editor_index if` fail someday or we start changing the `last_editor_index` somewhere else and forget to add an if `last_editor_index` the code would break. As it is now, with the full matrix combination set, the code is more robust and less prone to fail (less fragile). Keeping the codebase more robust/prepared to failures should be better than removing a few lines and by adding an if somewhere else.

dae: IMHO listing out all the possibilities as you've done is more prone to accidental mistakes, and it makes it harder to maintain the code moving forward. I'm also still a bit skeptical as to the demand for this feature - I think the best option is to leave this PR open for a while and see if other people want to add their 2c.

dae: Closing for now; let's see if demand presents itself in the future

PR #652, opened 2020-06-02
Tell the user what was error on his search query

evandrocoan:
When there is an error on the query syntax, the user has no clue what he did wrong:
1. https://anki.tenderapp.com/discussions/ankidesktop/42026-regular-expression-search-does-not-work-as-expected

With this, it shows the exception causing the error: 

image

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -180,12 +180,15 @@ class DataModel(QAbstractTableModel):
                 ctx.card_ids = self.col.find_cards(ctx.search, order=ctx.order)
             gui_hooks.browser_did_search(ctx)
             self.cards = ctx.card_ids
+        except anki.rsbackend.DBError as e:
+            error = str(e).replace("\\n", "\n")
+            invalid = f"\n\n{error}"
+            print(f"search failed: {traceback.format_exc()}")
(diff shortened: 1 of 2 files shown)


dae:
A traceback doesn't buy us much, since the error is likely originating from the backend, and it's better not to dump a wall of text at the user.

To fix this nicely, localized_description() in err.rs also needs to be updated to catch this specific case and return just the inner message - the outer DBError / SqliteFailure is just noise.

evandrocoan:
I swallowed the stacktrace when the `SqliteFailure` exception is thrown.

image

But kept one showing up on the console to orientate newer developers on the project.

The best option would also have the Rust Backend giving us back a stacktrace. As of now, I have not idea how the Rust Backend communities and gets generated the regex error as not backtrace whatsoever is shown anywhere. If someday I have to work with the Rust Backend, I will look into adding a backtrace to the messages.

dae:
I've pushed something similar to this - it avoids the stacktrace and won't catch sqlite errors that aren't caused by the regex failing: https://github.com/ankitects/anki/commit/c77d017135ff3e504c19970d5caeff4f0afeeb3f

PR #653, opened 2020-06-03
Add missing quotes to typecheck-setup.sh for paths with spaces on

evandrocoan:
Add missing quotes to `typecheck-setup.sh` for paths with spaces on the name.


dae:
Thanks Evandro :-)

PR #654, opened 2020-06-03
Now the pyqt5.15.0 are required for its stubs

evandrocoan:
https://github.com/ankitects/anki/pull/651#issuecomment-637709601


evandrocoan:
Actually, this still working as expected. I just need to check out in a branch that uses `pyqt 5.15.0` and run `git clean -xfd .`

PR #655, opened 2020-06-03
History line

Arthur-Milchior:
I expect to use this hook to create an add-on which will simply remove the ",,". By ignoring empty fields I expect to make this easier to use. I may also try to show only some field content.


dae:
No problem with the hook itself, but the docs should use "Allows changing" as we've covered before, and the name of the hook should match other hooks - something like "addcards_will_add_history_entry"

Arthur-Milchior:
Rebased.

dae:
Thanks Arthur.

PR #657, opened 2020-06-05
don't allow addons21 as profile name

BlueGreenMagick:
because addons21 and profiles directory are stored in same directory


dae:
cheers

PR #660, opened 2020-06-07
Update tooltip() in utils.py

brownbat:
Allow x,y offsets to be specified for tooltip(), useful for accomodating longer tooltips.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -36,6 +36,7 @@ Zeno Gantner <zeno.gantner@gmail.com>
 Henrik Giesel <hengiesel@gmail.com>
 Michał Bartoszkiewicz <mbartoszkiewicz@gmail.com>
 Sander Santema <github.com/sandersantema/>
+Thomas Brownback <https://github.com/brownbat/>
 
 ********************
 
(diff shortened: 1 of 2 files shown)


evandrocoan:
As we are allowing any position to be passed on, it would be better to ensure the passed position is inside the screen/visible. If the passed position falls outside the user screen, fall back into using the original position `x=0` and `y=100`.

brownbat:
> 
> 
> As we are allowing any position to be passed on, it would be better to ensure the passed position is inside the screen/visible. If the passed position falls outside the user screen, fall back into using the original position `x=0` and `y=100`.

Great point. Should we revert to default, or would it be better if we just went as far as we could manage, set some limit based on window height/width and stopped at that?

I'm thinking, if you can get the tooltip to move, but just not as far as you thought, it might lower surprise and be easier to debug relative to it suddenly snapping back and not moving at all in some conditions. Like a resized window suddenly making something not work as expected...

I'll try out a few options, will see what I can do, thanks.

dae:
Happy to accept this as is - your add-on can call ensureWidgetInScreenBoundaries on the tooltip if necessary.

PR #661, opened 2020-06-07
Fix fixme "ensure nulls and x1f not in field contents"

bogglez:
Reused "invalid input" error, added unit tests for "field separator" and "null character" each.
Also added test case for invalid idx in set_field() while I was at it.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -36,6 +36,7 @@ Zeno Gantner <zeno.gantner@gmail.com>
 Henrik Giesel <hengiesel@gmail.com>
 Michał Bartoszkiewicz <mbartoszkiewicz@gmail.com>
 Sander Santema <github.com/sandersantema/>
+bogglez <bogglez@protonmail.ch>
 
 ********************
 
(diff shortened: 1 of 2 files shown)


dae:
Thanks for contributing! 

There are a couple of issues with this solution:

- set_field() is not used anywhere outside the unit tests at the moment, so these checks will not actually run in most cases
- throwing an error isn't ideal - it means the calling code needs to do extra work to explain it to the user, and is not something we want to happen during a sync.

A better approach would be to just remove the unwanted characters from the string prior to them being saved into the DB, similar to the approach in fix_name(), where extra strings are only created in the case that one of those characters is detected in the field. Somewhere near the start of prepare_for_update() is probably best, as it should happen prior to the checksum generation.

bogglez:
Thank you for the feedback!

I'll take a look at it during the weekend.
I assume the failing Windows-specific checks can be ignored by me?

dae:
Thanks!

Rebasing onto the latest master should fix the Windows issue.

PR #662, opened 2020-06-09
Handle None from QApplication.desktop().screen()

gaul:
This prevents a crash when Anki is open for multiple days.  Reference:
https://anki.tenderapp.com/discussions/ankidesktop/41879-qt-cannot-create-window-no-screens-available


dae:
Thanks!

PR #663, opened 2020-06-09
Update message to user to match gui

kenden:
The "Change Note Type"... command is actually called "Change Note Type..."
and is under the "Notes" menu, not the "Edit" menu.


BlueGreenMagick:
You should probably use Fluent translation style instead of gettext, since it's a new string.

dae:
Thanks Kenden.  Probably easier on translators to keep it in gettext for now, as the old string should appear as a fuzzy match.

PR #668, opened 2020-06-12
Choose which addon to update

BlueGreenMagick:
This PR allows users to choose which addon to update.

It was something that came up when discussing about addon security. Some people (including me) hold off updating addons for some time to avoid installing malicious addons. This PR would make it much easier by allowing users to choose only those addons that was available for update for some time, addons from trusted devs, or addons with major bug fix.

Here is how it looks like:
image

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -31,6 +31,7 @@ from aqt.qt import *
 from aqt.utils import (
     TR,
     askUser,
+    checkable_list,
     getFile,
     isWin,
     openFolder,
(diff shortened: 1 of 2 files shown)


dae:
This looks like it would be a nice addition, thank you! A couple of thoughts:

- I am guessing this will pop up daily at the moment, requiring the user to ignore the same update each day? It may be worth recording the most recent modification time, and avoiding prompting the user again unless another update comes. You'd probably need some way to force it to show when the user clicks 'check for updates' manually as well.
- Showing the update date in the add-on's title might be nice, and double-clicking or a separate info button to view the add-on's page (to see what changed) would be nice too.
- Remembering the checked/unchecked state for each add-on would be a nice convenience as well. You could add a new attribute to AddonMeta to record this, and default it to on.
- As no other parts of the code use checkable_list, and keeping things generic may make it harder to implement some of the above, would it make more sense putting it in addons.py or a new file, and giving it a less generic name?

I think the first point needs addressing; the other mentioned features are more nice to haves if you felt like implementing them.

ijgnd:
I have an add-on [postpone/delay addon update for selected addons](https://ankiweb.net/shared/info/1655214124). In case you have missed it: My code is not that nice but if there's anything useful it use it.

BlueGreenMagick:
Those are all good suggestions. However, I can't find the time needed to implement those. If anyone else wants to tackle this, feel free to use this code. Else I'll open a new PR when I find the time

PR #670, opened 2020-06-19
Create the command line argument -v/--version to print version

evandrocoan:
I only figured out `anki --help` would print the version after looking at the code.

A direct way to check for Anki version, instead of using `anki -h` (which is not much obvious).

https://forums.ankiweb.net/t/i-couldnt-open-anki-resetting-corrupt-global/190


dae:
cheers

PR #672, opened 2020-06-21
Update stats.py

noobj2:
Change background color for pie chart in night mode so its background isn't white in night mode


dae:
I'm afraid this won't work, as libanki is intended to work even without the qt module. Changes are coming to the graphs soon, so I recommend just waiting for that, instead of trying to modify this patch. Thanks for taking the time to contribute in any case :-)

PR #673, opened 2020-06-25
Add type hints to clayout.py

neitrinoweb:
Hello! This is my first PR (ever), got a few low-hanging fruits, any feedback would be appreciated!

Also, I'm really looking forward into contributing to Anki project and I would like to know if there's any developers chat where I can seek help? (except forum)


dae:
Thanks for contributing!

There are no official real time chat channels for development, but there may be community run ones out there - I think David mentioned an AnkiDroid one on the forums.

PR #674, opened 2020-06-26
Descriptive browser columns

sudomain:
This changes the "edited" and "changed"  browser columns to (in my opinion) the more descriptive "Note Modified" and "Card Modified".


dae:
I might move these into Fluent prior to changing them so we don't break the existing translations - will leave this PR open as a reminder.

dae:
Fixed in https://github.com/ankitects/anki/commit/f3febea4b059116abf30006853974091a2998890

PR #675, opened 2020-06-27
A few mpv changes 2

kelciour:
The first commit is a fixed commit from the previous pull request to replace default mpv "quit" keybindings if mpv version is >= 0.30.

The second commit replaces mpv "idle" property with "end-file" event and the third commit tries to fix no sound if the video window was closed manually and mpv version is less than 0.30. I'm not sure how to do it better and maybe it'd be okay since it'll be only used if mpv is less than 0.30 and the video window was closed manually. I need to register callbacks again and reused _event_thread instead of creating a new one in an attempt to fix RuntimeError("cannot join current thread") and used Exception to be able to catch pywintypes.error on Windows and BrokenPipeError on Linux.

### DIFF
--- a/qt/aqt/mpv.py
+++ b/qt/aqt/mpv.py
@@ -245,10 +245,13 @@ class MPVBase:
             else:
                 r, w, e = select.select([self._sock], [], [], 1)
                 if r:
-                    b = self._sock.recv(1024)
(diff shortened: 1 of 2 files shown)


dae:
Thanks kelciour, let's try it out

PR #676, opened 2020-06-29
Fixed pbjs not working on Windows because package.json runs by

evandrocoan:
cmd.exe which does not support pipe | and ; semicolons
https://github.com/ankitects/anki/commit/6fd444b9587918834bf5f641bd7a44cd064ff495#r40232987

If this message: https://github.com/ankitects/anki/runs/817774681#step:25:884
```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received an instance of Array
```

Was correctly implemented by the **incompetent developer** of that module, it should show us the `"path"` string is invalid. In our case, it would show that the invalid path would be `src/backend/proto.js; pbjs -t static-module ../proto/backend.proto | pbts...` (as ; and | are not interpreted by the Windows cmd.exe shell)

### DIFF
--- a/ts/Makefile
+++ b/ts/Makefile
@@ -24,7 +24,8 @@ PROTODEPS := ../proto/backend.proto ../proto/fluent.proto
 BUILDDEPS := .build/npm webpack.config.js
 
 .build/proto: $(BUILDDEPS) $(PROTODEPS)
-	npm run proto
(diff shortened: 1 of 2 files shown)


dae:
Thanks for looking into that Evandro, I appreciate it.

PR #678, opened 2020-07-05
Always use standard QT color dialog on Linux

rathsky:
On Linux only Gnome uses its own color dialog, and it's inferior to QT's own: See the same issue @ OpenShot/openshot-qt#2425

(This is/was visible only on some distributions' packaged versions of Anki, not via ./run.)

isLin check since QT on Windows already uses the default, whereas Mac's own picker seems superior (from what I can judge by screenshots).


dae:
cheers

PR #679, opened 2020-07-06
Fix _addons/ URL prefix not being removed on Windows systems

evandrocoan:
Anki 2.1.28 Beta
https://forums.ankiweb.net/t/anki-2-1-28-beta/629/26

> Did the requirements on how to expose resources on the internal server change? Because scripts, css files and the like, registered with setWebExports, do not work anymore with beta2.
> A look at the console shows the following for all resources:
> Failed to load resource: the server responded with a status of 500 (INTERNAL SERVER ERROR)

**\Anki2\addons21\someaddon\__init__.py**
```py
from aqt import mw
import aqt

# assuming add-on folder is named "test"
# and assuming a 'test.js' is in the same folder
addon_id = "someaddon"
port = mw.mediaServer.getPort()
mw.addonManager.setWebExports(addon_id, ".*\\.js$")

print("Hello")

aqt.editor._html += f"""
<script>
    var script = document.createElement('script');
    script.type = 'text/javascript';
    script.src = 'http://127.0.0.1:{port}/_addons/{addon_id}/test.js';
    document.body.appendChild(script);
</script>"""
```

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -181,7 +181,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ${{ github.workspace }}${{ matrix.SEP }}pyenv
-          key: ${{ runner.os }}-pyenv-${{ hashFiles('**/requirements.*') }}-${{ hashFiles('**/setup.py') }}-${{ hashFiles('**/Makefile') }}-${{ hashFiles('**/Cargo.toml') }}-17-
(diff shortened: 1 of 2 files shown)


evandrocoan:
The windows builds are failing because someone started installing Python `3.7.8` while only the old version `3.7.7` is added on path/being used.

evandrocoan:
After thinking, the build was just failing because of the pyenv caching, On windows, pyenv copies the python executable, instead of symlinking (probably because on windows hard links require administrator privileges).

dae:
Thanks for the quick fix!

PR #680, opened 2020-07-07
Add {{type:nc:…}} – ignores combining characters when comparing

twwn:
This allows to add :nc to type fields, causing Anki to ignore all combining characters so as to not fault their absence in the user input as errors.

It improves one's learning experience with languages such as Hebrew whose use of diacritics (nikud) is almost entirely limited to reading them (in class/references), with nobody writing them (very cumbersome anyway) in everyday use for ordinary words.

Example before / after:

wo-nc
with-nc

The approach basically consists of:
* Strip _given_ & _correct_ of their combining characters
* Use that for tokenizeComparison
* "Map" the stripped _correct_ to its original
* Restore the combining characters in the correction line with that "mapping"

It's either-or, meaning any combining characters in the input/_given_ are discarded as well (so no partial comparison).

Notable otherwise: one functional change affecting the non-:nc path and two clean-ups
* If no user input (_given_) was entered, simply return the field/_correct_ straight away with minimal formatting.
* Removed an unused correct() parameter.
* Skip the redundant tokenizeComparison for the non-:nc path when _given_ already matches _correct_.

### DIFF
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -51,6 +51,7 @@ class Reviewer:
         self.hadCardQueue = False
         self._answeredIds: List[int] = []
         self._recordedAudio: Optional[str] = None
+        self.combining: bool = True
         self.typeCorrect: str = None  # web init happens before this is set
         self.state: Optional[str] = None
         self.bottom = BottomBar(mw, mw.bottomWeb)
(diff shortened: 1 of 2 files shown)


dae:
Hi rathsky,

No objections to the idea itself, but the type-in-the-answer code is likely going to move into TypeScript in the future, so this would probably be better done after that change is made - will leave this open until then.

Regarding the specifics of this implementation, perhaps simply stripping all combining characters if the user has used nc: would suffice? It would make for a simpler implementation if you don't need to preserve the original text verbatim.

rathsky:
It would make it simpler, but while the diacritics are an obstacle to typing the words, they are still essential to their learning. So removing them entirely would require duplicating the field, which is unsightly and a waste of space.

dae:

PR #682, opened 2020-07-10
Add top_toolbar_did_redraw hook

glutanimate:
Just a minor hook that notifies add-ons of the top toolbar being redrawn. Can act as a substitute for add-ons which hooked into `gui_hooks.toolbar_did_init_links` in the past to subsequently update their content at runtime. (after the recent changes, `gui_hooks.toolbar_did_init_links` is only called once now at toolbar initialization time)

Example usage:

```python
from aqt import gui_hooks

def on_top_toolbar_did_redraw(toolbar):
    print("top toolbar redrawn")
    print(toolbar)

gui_hooks.top_toolbar_did_redraw.append(on_top_toolbar_did_redraw)
```

### DIFF
--- a/qt/aqt/gui_hooks.py
+++ b/qt/aqt/gui_hooks.py
@@ -2437,6 +2437,35 @@ class _TopToolbarDidInitLinksHook:
 top_toolbar_did_init_links = _TopToolbarDidInitLinksHook()
 
 
+class _TopToolbarDidRedrawHook:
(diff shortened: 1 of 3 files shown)


glutanimate:
Seems like there is an issue with the test environments on Ubuntu:

> 💥 maturin failed
>  Caused by: Expected `python` to be a python interpreter inside a virtualenv ಠ_ಠ

dae:
Thanks!

PR #683, opened 2020-07-12
Allow to choose note type from list by pressing ctrl + row number

flagist0:
This PR adds the ability to choose note type without selecting it with arrow keys or typing in the type name -- it can be done by pressing ctrl + row number. 
As the number of most used note types is limited, it's easy to remember one digit per type. So with the shortcut this operation becomes automatical and very fast (like "ctrl-n ctrl-5 enter" for choosing cloze type).
Multi-digit row indices are not supported for simplicity.


flagist0:
Sorry, committed it with wrong git config, it wouldn't pass contributors check, will remake PR

PR #684, opened 2020-07-12
Allow to choose note type from list by pressing ctrl + row number

flagist0:
This PR adds the ability to choose note type without selecting it with arrow keys or typing in the type name -- it can be done by pressing ctrl + row number.
As the number of most used note types is limited, it's easy to remember one digit per type. So with the shortcut this operation becomes automatical and very fast (like "ctrl-n ctrl-5 enter" for choosing cloze type).
Multi-digit row indices are not supported for simplicity.

### DIFF
--- a/qt/aqt/studydeck.py
+++ b/qt/aqt/studydeck.py
@@ -92,8 +92,9 @@ class StudyDeck(QDialog):
 
             if rows_count:
                 new_row %= rows_count  # don't let row index overflow/underflow
-            self.form.list.setCurrentRow(new_row)
-            return new_row != current_row
+            if new_row != current_row:
+                self.form.list.setCurrentRow(new_row)
+                return True
         return False
 
     def redraw(self, filt, focus=None):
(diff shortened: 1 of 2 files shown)


qt/aqt/studydeck.py
```diff
@@ -77,20 +77,23 @@ def __init__(
...
+            elif key == Qt.Key_Down:
+                new_row = current_row + 1
+            elif evt.modifiers() & Qt.ControlModifier and Qt.Key_1 <= key <= Qt.Key_9:
+                row_index = key - Qt.Key_1
+                if row_index < rows_count:
+                    new_row = row_index
+
+            if rows_count:
+                new_row %= rows_count  # don't let row index overflow/underflow
+            self.form.list.setCurrentRow(new_row)
```
dae, line 95: Would be nicer if setCurrentRow() is skipped when nothing has changed, but otherwise looks good, thanks!

flagist0: Yes, you are right, thanks!

dae:
Thanks!

PR #685, opened 2020-07-13
Scroll media sync log to the last line

kelciour:
If Sync button was clicked and media sync log is about 20 lines or longer, the last line or new lines will be hidden until the window is manually scrolled down.

With the change in the pull request, the last line with the most recent media sync status message will be visible


dae:
Thanks!

PR #686, opened 2020-07-13
Added parameter/return types, fixed possible logic error and typo

zerrs:
I added some simple return and parameter types for a few functions in qt/aqt/customstudy.py. There was also a possible logic error in the "plus" function.
```python
def plus(num: int) -> str:
    if num == 1000:
        num = "1000+"
    return "<b>" + str(num) + "</b>"
```

The "==" should be a ">"


dae:
Thanks! I removed the string change, since that will need to wait until it's moved to Fluent or it will break the existing translations. Also the same issue was in another string a little further down :-)

PR #689, opened 2020-07-16
Add IDs to stats sections

glutanimate:
Allows more targeted manipulation by add-ons.

Please let me know if the naming is OK with you. Will be happy to switch to a different naming scheme if you prefer a change.


dae:
LGTM, thanks

PR #691, opened 2020-07-17
Improve variable names in tests

Arthur-Milchior:
The commit states which sed comments make the most of the change. This is not entirely exact, I then had to check everything and decide which change to accept or to reject. 

I hope it'll make tests easier to understand for any new programmer. I also prefer to do this before porting them to AnkiDroid, where they are really overdue.


dae:
The tests are passing, so I'm going to assume it's ok. Should be a nice improvement, thanks.

PR #692, opened 2020-07-17
Remove useless variables

Arthur-Milchior:
Some minor change I made when commenting anki's code. In case you're interested in removing some variables used only once which does not had much readability

### DIFF
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -71,8 +71,7 @@ class Collection:
         return f"{super().__repr__()} {pprint.pformat(d, width=300)}"
 
     def name(self) -> Any:
-        n = os.path.splitext(os.path.basename(self.path))[0]
-        return n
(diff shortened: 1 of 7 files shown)


qt/aqt/mpv.py
```diff
@@ -232,7 +232,7 @@ def _reader(self):
         while not self._stop_event.is_set():
             if isWin:
                 try:
-                    (n, b) = win32file.ReadFile(self._sock, 4096)
+                    (n, b) = win32file.ReadFile(self._sock, 4096)[1]
```
dae, line 235: This looks incorrect

Arthur-Milchior: I first removed the n, then while re reading I thought that it's more clear to leave it this way, even if `n` is never used, so I added it back. Not correctly, sorry

dae:
Will merge after 2.1.28 release

Arthur-Milchior:
Ok. Thanks

PR #698, opened 2020-07-20
remove _clearOverdue

Arthur-Milchior:
This member is never read. If it should be kept, I'd like to know the purpose of the member. Even looking at git history I can't find anything


dae:
Probably from Anki 1.2 code, or from early 2.0 prototyping. Cheers.

PR #699, opened 2020-07-20
Refresh deck stats on reopen

Vorlent:
When I review my cards I usually click on the Stats button in the toolbar. Once the "DeckStats" dialog is open it refuses to refresh itself when I click on the "Stats" button again. I have to manually close the dialog and then click on the button again.

This PR just refreshes the dialog every time it is reopened.


dae:
I'm not sure this is what everyone would want - I imagine some users just want to bring the dialog back into view and retain their current settings/display. Perhaps it's worth creating a discussion on the forum to see what others think?

Vorlent:
No, I don't think so. There is no compromise that would make everyone happy.

PR #700, opened 2020-07-23
Update actions/upload-artifact@v1 to v2 as v1 is deprecated

evandrocoan:
Random error on Mac OS: [error]Too Many Requests
https://github.com/actions/toolkit/issues/503


dae:
Thanks as always Evandro!

PR #702, opened 2020-07-23
Add type hints for apt.dyndeckconf

mkrump:
Wanted start contributing to Anki and https://github.com/ankitects/help-wanted/issues/4 looked a like a good place to start. Was just planning on doing one file at a time, so this PR adds type hints for `dyndeckconf.py` Haven't been using Python as much recently, and not super familiar with `mypy`, so assume no prior knowledge when reviewing :). 

Saw the following errors after turning on `check_untyped_defs` for `dyndeckconf.py`. PR fixes errors, adds type signature to the function, and turns on type checking.

```
aqt/dyndeckconf.py:149: error: Incompatible types in assignment (expression has type "float", variable has type "str")  [assignment]
                    i = float(i)
                        ^
aqt/dyndeckconf.py:150: error: Unsupported operand types for > ("str" and "int")  [operator]
                    assert i > 0
                               ^
aqt/dyndeckconf.py:152: error: Incompatible types in assignment (expression has type "int", variable has type "str")  [assignment]
                        i = int(i)
                            ^
```

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -43,6 +43,7 @@ Nickolay Yudin <kelciour@gmail.com>
 neitrinoweb <github.com/neitrinoweb/>
 Andreas Reis <github.com/rathsky>
 Alexander Presnyakov <flagist0@gmail.com>
+Matt Krump <github.com/mkrump>
 ********************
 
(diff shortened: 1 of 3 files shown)


dae:
Thanks Matt!

PR #703, opened 2020-07-24
 fix help-wanted#13 

ANH25:
This fixes ankitects/help-wanted#13

the culprit is the `_processText` function in editor.py, which wraps the whole pasted text in anchor tags after checking if the **beginning** of the text looks like a URL (using `isURL`) and that it's not a media link.

I fixed that by processing the text line by line, checking for links and HTML-escaping it.

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -1091,34 +1091,35 @@ class EditorWebView(AnkiWebView):
             return None
 
         txt = mime.text()
+        processed = ''
+        lines = txt.split("\n")
 
-        # inlined data in base64?
-        if txt.startswith("data:image/"):
-            return self.editor.inlinedImageToLink(txt)
+        for line in lines:
+            for token in re.split('(\S+)', line):
+                # inlined data in base64?
+                if token.startswith("data:image/"):
+                    processed += self.editor.inlinedImageToLink(token)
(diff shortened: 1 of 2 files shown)


qt/aqt/editor.py
```diff
@@ -1091,34 +1091,35 @@ def _processText(self, mime: QMimeData) -> Optional[str]:
...
-
-        # if there's more than one consecutive space,
-        # use non-breaking spaces for the second one on
-        def repl(match):
-            return match.group(1).replace(" ", "&nbsp;") + " "
-
-        txt = re.sub(" ( +)", repl, txt)
-
-        return txt
+        processed = ''
```
evandrocoan, line 1094: Using an array of strings `processed = [] & "".join(processed)` is exponentially faster than string concatenation: https://waymoot.org/home/python_string/ (Efficient String Concatenation in Python)

ANH25: done. Thanks!

evandrocoan:
@ANH, you need to run `make fix` to fix the formatting and commit the change: https://github.com/ankitects/anki/runs/905447871#step:25:526

dae:
Thanks ANH! I'll merge this once 2.1.29 is released.

PR #704, opened 2020-07-24
Add type hints for apt.deckconf

mkrump:

Saw the following errors after turning on `check_untyped_defs` for `deckconf.py`. PR fixes errors, adds type signature to the function, and turns on type checking.

```
aqt/deckconf.py:267: error: Incompatible types in assignment (expression has type "float", variable has type "str")  [assignment]
                    i = float(i)
                        ^
aqt/deckconf.py:268: error: Unsupported operand types for > ("str" and "int")  [operator]
                    assert i > 0
                               ^
aqt/deckconf.py:270: error: Incompatible types in assignment (expression has type "int", variable has type "str")  [assignment]
                        i = int(i)
                            ^
```

Also, I had an issue where mypy was throwing random errors in unrelated files. I tried re-running `make check` after a `make clean`, but they were still there. I ended up deleting the `pyenv` and reinstalling again and it seemed to work after. Wasn't sure if I was missing something that I could have used in Makefile

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -42,8 +42,8 @@ kenden
 Nickolay Yudin <kelciour@gmail.com>
 neitrinoweb <github.com/neitrinoweb/>
 Andreas Reis <github.com/rathsky>
-Alexander Presnyakov <flagist0@gmail.com>
 Matt Krump <github.com/mkrump>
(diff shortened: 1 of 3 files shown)


dae:
Not sure what might have caused that - might be worth trying to remove the .mypy_cache to see if it helps if it crops up again. Thanks for the PR!

PR #707, opened 2020-07-26
Add a filter to control sibling burying via add-ons

glutanimate:
I thought it would be neat for add-ons to have the ability to influence the scheduler's sibling burying behavior, allowing for uses like conditional burying depending on the note type and other properties.

The immediate use case me for me would be in Cloze Overlapper, which currently monkey-patches `Scheduler._burySiblings` to optionally exclude sequential siblings from burying. Outside of that, I could also see this hook being useful for the various add-ons out there that also modify sibling burying to some degree.

### DIFF
--- a/pylib/anki/hooks.py
+++ b/pylib/anki/hooks.py
@@ -12,7 +12,7 @@ modifying it.
 
 from __future__ import annotations
 
-from typing import Any, Callable, Dict, List, Sequence, Tuple
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
 import decorator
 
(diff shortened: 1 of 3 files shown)


glutanimate:
On a related note (no pun intended): Do you think that a change [like this](https://anki.tenderapp.com/discussions/ankidesktop/15781-feature-request-set-sibling-bury-options-per-deck) to have per-note-type control over burying is feasible for this development cycle? Or is this more of an Anki 2.2 feature? Asymmetrical deck designs like [this](https://ankiweb.net/shared/info/76882632) seem to be gaining traction over here, so it would be really neat to have that granular level of control across both desktop and mobile.

dae:
The scheduling code is due to be refactored soon, so I'll need to hold off on scheduling-related hooks for now, but will keep the note type burying in mind - I can definitely see how it might make sense for certain note types.

dae:
(keeping this open as a reminder)

dae:
Good news and bad news. The v3 scheduler lets you change the sort order of siblings so they're either separated or come together. That setting only applies to the selected deck and can't be adjusted in a hook, but whether cards are buried at the start happens on a per-deck basis - so I think it would be possible to have some cards buried and some not buried if they were placed in different subdecks.

PR #708, opened 2020-07-27
Attempt to fix TypeError: allroutes() got an unexpected keyword

evandrocoan:
Attempt to fix TypeError: allroutes() got an unexpected keyword argument ‘path’
https://forums.ankiweb.net/t/anki-2-1-28-error-message-when-reviewing-cards/1439


dae:
I fixed this in https://github.com/ankitects/anki/commit/f45627dbd3dfd3450871330e222c98c3a4cd5152 - if there's something that change missed than yours didn't, please let me know.

PR #709, opened 2020-07-28
Set to to checks.yml not run check rules on tagged releases

evandrocoan:
https://forums.ankiweb.net/t/tagged-checks/1465


dae:
Thanks a lot Evandro :-)

PR #710, opened 2020-07-28
Set to to checks.yml not run check rules on tagged releases

evandrocoan:
https://forums.ankiweb.net/t/tagged-checks/1465


evandrocoan:
GitHub did it again. It created two pull requests at once (https://github.com/ankitects/anki/pull/709).

PR #713, opened 2020-07-29
updated type hints in models.py

phwoo:
Small clarifications to type hints


phwoo:
`make check` didn't throw errors when I ran before the PR but it is now. Will revise.

PR #715, opened 2020-07-30
Added typehints to qt/aqt/modelchooser.py, updated qt/mypy.ini

phwoo:
Modelled typehints after `deckchooser.py`


dae:
Hmm, that author check is failing - would you mind making a minor edit to the contributors file again, preferably using the GitHub web editor so it uses your noreply email for the commit?

PR #717, opened 2020-07-30
Add night mode and mobile class toggles in card layout screen

ANH25:
This addresses ankitects/help-wanted#12.

It toggles night mode globally and restores it to the user preference when the screen is closed (I couldn't figure a way to apply it just to the preview section).

I grouped the the new two options, along with the `Fill Empty Fields` option, under an options button.

Note that I labeled the button `Options` which could be confusing since we have another options button above, so maybe we should call it something else like `Preview Options`?


glutanimate:
Nice work! Maybe "Preview Settings" to have an even clearer separation between the two?

dae:
Thanks ANH, it looks good! I had to merge a conflict here - in future PRs you may want to create a separate branch instead of using master, so your master branch stays clean.

dae:
The custom styling+long name looked bad on macOS; I've pushed some tweaks. A user has already reported issues with the screen being too wide in another language, so if the button can be kept shorter that would be better.

Screen Shot 2020-07-31 at 2 17 05 pm

PR #718, opened 2020-07-30
Added typehints for qt profiles

phwoo:
* `Any` used for pickle methods, this could probably be improved
with some kind of Callable
* `str` used for self.base, though this may be a problem for
different OSes. Some type of os.PathLike might be good.
See https://docs.python.org/3/library/os.html#os.PathLike
* Line 75, type ignored: mypy was complaining about no. of args,
and kwargs there didn't seem to be needed. Separate issue to test,
though.


dae:
Thanks for your continued work on this phwoo!

PR #720, opened 2020-07-31
Turn on check_untyped_defs for aqt.webview, aqt.addons and aqt.emptycards.

mkrump:
For ankitects/help-wanted#4

Saw the following errors after turning on check_untyped_defs for aqt.webview, aqt.addons and aqt.emptycards:

10f2f9c
```
aqt/emptycards.py:91: error: Name 'note' already defined on line 90  [no-redef]
                note: NoteWithEmptyCards = note
                ^
```

7d8f856
```                
aqt/addons.py:191: error: Incompatible types in assignment (expression has type "Iterator[Any]", variable has type "List[Any]")  [assignment]
                l = reversed(l)
```

a56690b
* when i tried to use Optional[str] for `result` mypy would complain `"pyqtSlot" has incompatible type "object"; expected "Optional[str]"`, so ignored 

* Seemed unlikely `QWebChannel` wasn't actually being imported, but mypy was complaining that it couldn't find it, so i added import to \_\_init\_\_ with rest of `PyQt5` imports.
```
aqt/webview.py:35: error: Argument "result" to "pyqtSlot" has incompatible type "Type[str]"; expected "Optional[str]"  [arg-type]
                @pyqtSlot(str, result=str)
                                      ^
aqt/webview.py:42: error: Name 'QWebChannel' is not defined  [name-defined]
            self._channel = QWebChannel(self)
```

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -179,7 +179,7 @@ class AddonManager:
         sys.path.insert(0, self.addonsFolder())
 
     # in new code, you may want all_addon_meta() instead
-    def allAddons(self):
+    def allAddons(self) -> List[str]:
         l = []
         for d in os.listdir(self.addonsFolder()):
             path = self.addonsFolder(d)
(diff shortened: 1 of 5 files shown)


dae:
Thanks Matt!

PR #722, opened 2020-08-01
Updated qt addons with more extensive typehints

phwoo:

Update for bcc35c1822e5422c8aa50a50c18da51a2c4cbbcf with as many typehints as I could manage, and one main change. This has
```
def allAddons(self) -> Iterable[str]:
    ...
    if os.getenv("ANKIREVADDONS", ""):
        it = reversed(l)
        return it
```
returning an Iterable, rather than a List, to match `all_addon_meta(...)`. Though you could probably take it or leave it.

The only function missing a return type is `def _addon_schema(...)`, as I wasn't sure what the best practice for returning after throwing an expection is, and mypy isn't too happy about not having a return statement along all conditional branches.

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -171,7 +171,7 @@ class AddonManager:
         "required": ["package", "name"],
     }
 
-    def __init__(self, mw: aqt.main.AnkiQt):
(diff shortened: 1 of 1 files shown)


phwoo:
Getting a runtime error in addons.py, will update

PR #723, opened 2020-08-01
Typehints for qt addons

phwoo:

Update for bcc35c1822e5422c8aa50a50c18da51a2c4cbbcf with as many typehints as I could manage.

The only function missing a return type is `def _addon_schema(...)`, as I wasn't sure what the best practice for returning after throwing an expection is, and mypy isn't too happy about not having a return statement along all conditional branches.

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -171,7 +171,7 @@ class AddonManager:
         "required": ["package", "name"],
     }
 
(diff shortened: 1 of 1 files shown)


dae:
Thanks for your continued work on this phwoo!

Could _addon_schema use Optional[Union[bool, Dict]], and return None in the exception caught case?

PR #724, opened 2020-08-05
NF: Factorize schedulers

Arthur-Milchior:
I must first state that I expect at least some of those change to be useless as they may be ported to rust. While this part of the scheduler is still in Python, I still have a preference for factorizing both schedulers if possible.

When I want to look at the schedulers, I like to know what is the difference between them. In some cases, the difference is very small and hard to find. Moving it to a member ensure that it can be found easily and that I know exactly what is the difference between them. Furthermore, in some cases, using snippes SQL also helps readability, since I can read a single word instead of reading a condition stating that queue is in a pair of elements.

 In an ideal world, I'd love a class, let's say `abstractSched`, which contains all the code in common, and then have sched.py and schedv2 containing only the code which differs, it would help a lot understanding the actual code. If you'd accept such a change I'd be happy to implement it. Right now I tried to limit myself to a bunch of small change trivial to check.

The real reason behind this change is that I expect to improve getCard in ankidroid. It may be quite slow when a deck has many subdecks. In order to do that as safely as possible, one of the step that I wish to take is to factorize the existing code as much as possible so that I don't have to do duplicate work and changes gets simpler to check.

### DIFF
--- a/pylib/anki/sched.py
+++ b/pylib/anki/sched.py
@@ -699,16 +699,6 @@ did = ?, queue = %s, due = ?, usn = ? where id = ?"""
         self.removeLrn(ids)
         super().suspendCards(ids)
 
-    def unsuspendCards(self, ids: List[int]) -> None:
-        "Unsuspend cards."
-        self.col.log(ids)
-        self.col.db.execute(
-            f"update cards set {self._restoreQueueSnippet},mod=?,usn=? "
-            f"where queue = {QUEUE_TYPE_SUSPENDED} and id in " + ids2str(ids),
-            intTime(),
-            self.col.usn(),
-        )
-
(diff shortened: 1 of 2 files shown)


dae:
I plan to move more of the scheduler into Rust in the near future, which will likely render these changes obsolete I'm afraid.

Again, please feel free to run an idea by me before starting work on it to avoid wasted effort.

Arthur-Milchior:
I feel free to run idea by you, as I did for changing variables names in unit test not a month ago.
In this case, I felt like trying to convey my idea in english would be more difficult than just showing what I expected to do, and since I was just copying code change I made in java, I expected this to be quick anyway. The only real time lost I didn't expect was ensuring that make check runs on my computer.

PR #725, opened 2020-08-06
add card layout save shortcut

ANH25:
This just sets `Ctrl+Return` as a shortcut to save modifications to card layout.


dae:
Thanks!

PR #726, opened 2020-08-08
Add main_window_will_require_reset gui_hook

hgiesel:
This will add a gui_hook for when dialogs require the main window to require a reset and display an overlay.

The use case is my add-on [Persistent Editor](https://ankiweb.net/shared/info/1686259334), which allows the EditCurrent window to stay open during review and "move along" with the cards (or rather their notes).

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -177,8 +177,8 @@ class AddCards(QDialog):
         self.mw.col.add_note(note, self.deckChooser.selectedId())
         self.mw.col.clearUndo()
         self.addHistory(note)
-        self.mw.requireReset()
         self.previousNote = note
+        self.mw.requireReset('addCardsAddNote', self)
         gui_hooks.add_cards_did_add_note(note)
         return note
 
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -178,7 +178,7 @@ class AddCards(QDialog):
         self.mw.col.clearUndo()
         self.addHistory(note)
         self.previousNote = note
-        self.mw.requireReset("addCardsAddNote", self)
+        self.mw.requireReset(reason="addCardsAddNote", context=self)
         gui_hooks.add_cards_did_add_note(note)
         return note
 
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -178,7 +178,7 @@ class AddCards(QDialog):
         self.mw.col.clearUndo()
         self.addHistory(note)
         self.previousNote = note
-        self.mw.requireReset('addCardsAddNote', self)
+        self.mw.requireReset("addCardsAddNote", self)
         gui_hooks.add_cards_did_add_note(note)
         return note
 
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -12,6 +12,7 @@ from anki.lang import _
 from anki.notes import Note
 from anki.utils import htmlToTextLine, isMac
 from aqt import AnkiQt, gui_hooks
+from aqt.main import ResetReason
 from aqt.qt import *
 from aqt.sound import av_player
 from aqt.utils import (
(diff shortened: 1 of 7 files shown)


dae:
This looks similar to @Arthur-Milchior 's https://github.com/ankitects/anki/pull/477 where I was concerned about possible bugs if resets are skipped. How are you planning to handle things like a card being deleted in the browse screen?

hgiesel:
Okay, I see the problem.

My suggestion:
1. `requireReset`s parameter `modal` is not used at the moment. It could be removed.
2. Rather than modal, we could have `reason`, and `args`
3. I have enumerated the different uses of `requireReset` in the code:

In browser.py it is used in the following cases:
- deletion
- deckChange
- tagsChange
- suspend
- reposition
- reschedule
- findReplace
- tagDuplicates

In addcards.py it is used in the following cases:
- addNote

In editcurrent.py it is used in the following cases:
- initEditCurrent

In editor.py it is used in the following cases:
- bridgeCmd

These "cases" could be passed as a `reason` parameter to `requireReset`. Additionally to the reason parameter, the can also be additional args. All of these `args` can also be passed to the gui_hook call, e.g. `requireReset('bridgeCmd', editor, cmd)`. `requireReset` will then look somewhat like this:

```
def requireReset(self, reason, *args):
    "Signal queue needs to be rebuilt when edits are finished or by user."
    self.autosave()
    if gui_hooks.main_window_will_require_reset(self.interactiveState(), reason, args):
        self.moveToState("resetRequired")
```

This allows the developer to easily pick out the cases, in which he wants to alter the reset behavior. The `deletion` case would require more attention, than let's say "initEditCurrent", but at this point, the user can easily distinguish between these cases.

dae:
There are many add-ons that call requireReset() at the moment that this will break, and I'm concerned this general approach will be error-prone. What situations are you wanting to catch in your add-on, and what would you do instead?

hgiesel:
I want to catch the following two `requireReset` scenarios:
* `EditCurrent` is opened
* `Editor` within `EditCurrent` has input (bridgeCmd)

So this any scenario where cards are deleted, or otherwise _big changes_.

I explored multiple approaches, but I think the least invasive one is using `activeWindow()` on `AnkiQt.interactiveState`:

```
def do_not_require_from_editcurrent(self, _old):
    active_window = mw.app.activeWindow()

    if ((isinstance(active_window, AnkiQt) and mw.state == 'review') or
            isinstance(active_window, EditCurrent)):
        return False

    return _old(self)

AnkiQt.interactiveState = wrap(AnkiQt.interactiveState, do_not_require_from_editcurrent, pos='around')
```

I _think_ this catches all `requireReset` calls that happen when interacting with `EditCurrent`, at least in vanilla Anki.

But I think this is code requires too many assumptions. For example, some hook in the browser could `editcurrent.editor.web.eval()`, which would not be picked up by this approach.

dae:
Sorry, been a bit busy recently - will try to get back to you soon.

hgiesel:
No, problem.
Since the last time, I just changed the signature of the changed `requireReset()` to be 100% backwards compatible.

---

Another thing is considered is that since v3.8, Python supports [Positional-Only Parameters](https://www.python.org/dev/peps/pep-0570/). Using that feature, we could improve future compatability.

```
def requireReset(self, modal=False, *, reason='unknown', context=None):
```

This signature would prevent users to call it like `mw.requireReset(False, 'foobar', ctx)`, and forces them to use `mw.requireReset(False, reason='foobar', context=ctx)`.

dae:
Ok, it looks pretty good. A couple of points:

- it might be nicer if we could use an enum instead of string reasons. It doesn't look like enums can easily be extended dynamically, so the hook would need to take Union[AnkiReason, str] so that add-ons could provide a string reason. "unknown" could remain as a string
- please add a warning to the doc string that care is required, and the reason should definitely be checked
- should_require_reset would be better than will_require_reset, since the latter implies no control

We still support Python 3.7, so can't use positional-only yet.But that's the '/' character; * means keyword only, not positional only. You'll see Anki's already making use of the latter in rsbackend_gen.py

dae:
Thanks for making those changes, and sorry about the conflicts.

PR #727, opened 2020-08-08
Add editor_will_munge_html hook

hgiesel:
This hook generalizes the functionality of Editor.mungeHTML, so that you can easily extend by using the hook.

This used in my add-on [Persistent Editor](https://ankiweb.net/shared/info/1686259334). The add-on places a div inside of the text field, so that it can adopt the size of the field, but also obscure its content.


dae:
cheers

PR #728, opened 2020-08-09
Move "remove null bytes" and "reverse url formatting" functionality to editor_will_munge_html hook

hgiesel:
This will clean up the `Editor.onBridgeCmd` method a little bit.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -218,7 +218,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ${{ github.workspace }}${{ matrix.SEP }}rslib${{ matrix.SEP }}target
-          key: ${{ runner.os }}-cargo-rslib-${{ hashFiles('**/requirements.*') }}-${{ hashFiles('**/setup.py') }}-${{ hashFiles('**/Makefile') }}-${{ hashFiles('**/Cargo.toml') }}-${{ matrix.BUILD_TYPE }}-22-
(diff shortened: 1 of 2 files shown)


hgiesel:
I don't see how my changes could make this check fail, especially it has to do with Rust. Maybe somebody could restart the check?

evandrocoan:
You can try resetting the cardo index cache by bumping its number from 22 to 23 
https://github.com/ankitects/anki/blob/b380f304bf4a1b02b980be7c5d65835ffd2ea78e/.github/workflows/checks.yml#L200

If it still not working, try resetting the cargo registry
https://github.com/ankitects/anki/blob/b380f304bf4a1b02b980be7c5d65835ffd2ea78e/.github/workflows/checks.yml#L207

And if it still not working, reset the cargo rslib
https://github.com/ankitects/anki/blob/b380f304bf4a1b02b980be7c5d65835ffd2ea78e/.github/workflows/checks.yml#L221

And if it still not working, reset the cargo rspy
https://github.com/ankitects/anki/blob/b380f304bf4a1b02b980be7c5d65835ffd2ea78e/.github/workflows/checks.yml#L228

It is important to not reset all of them at the same time (do one per commit) because this will tell us where the problem is. And when we know where the problem is on, we can try to fix it so it does not happen again in the future.

dae:
Thanks, this looks a bit cleaner.

@evandrocoan interestingly, bumping just rspy did not fix the problem https://github.com/ankitects/anki/runs/965477230

evandrocoan:
Next time this ``can't find crate for `prost_derive` which `prost_build` `` problem happens, we can try just bumping pyenv (https://github.com/ankitects/anki/pull/681) and see whether it also fixes it as this looks directly related to maturin.

PR #729, opened 2020-08-10
Add debug info for CI failure

evandrocoan:
``can't find crate for `prost_derive` which `prost_build` ``

Set an `upload artifacts` to upload so we can download it and compare a good one with a bad one: https://github.com/ankitects/anki/pull/728#issuecomment-671187035


dae:
Cheers

PR #730, opened 2020-08-11
fix Ctrl+right/left handling in RTL fields

ANH25:
As I've reported in the [forum](https://forums.ankiweb.net/t/some-caret-navigation-shortcuts-in-the-editor-are-reversed-for-rtl-languages/1228), the Ctrl+right/left shortcuts are reversed for right-to-left fields.

This is the same workaround shown in the forum post, except limiting it to RTL fields.


dae:
Thanks!

PR #731, opened 2020-08-11
Search_model has context

Arthur-Milchior:
have an add-on which allow to open multiple windows. So would like to know which browser has done the search. This
would allow me to have browser with various setting. In particular, I've an add-on which allow to show a single line by
note, and I would like for some browser to have this feature on and for some other to have this feature off.

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -70,6 +70,7 @@ class FindDupesDialog:
 @dataclass
 class SearchContext:
     search: str
+    browser: Browser
     order: Union[bool, str] = True
     # if set, provided card ids will be used instead of the regular search
     card_ids: Optional[Sequence[int]] = None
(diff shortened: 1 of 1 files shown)


Arthur-Milchior:
Related to "a single line by note in the browser", I would love for such a query to be created directly in rust if you can. 
Before moving to rust, I think I was making the query directly, which allowed it to be quick. With rust, today, I get a list of card id, and I need to make a lot of query to get the note id and then be able to find out which cards to keep and which cards to remove. That made this add-on quite slow. (Almost half a minute for my whole collection)

dae:
There is search_notes(), but it wasn't intended for use in the browse screen, and does not support ordering. Your best bet is probably to query all the nids in one go using id2str

dae:
Re the PR, how about using the browser instance instead? The data model can be retrieved from it, and it seems to make more sense to pass the top levels object.

Arthur-Milchior:
My thought was that the Context is created in the Model and I expect it to change only informations regarding the model and not the remaining of the browser. Anyway, there is one browser by model and one model by browser, so anything is fine

dae:
Either option could work, since they both link to each other - I was just thinking of how other browser hooks tend to pass in the browser, so this would be a bit more consistent. Not a big issue in any case :-)

PR #732, opened 2020-08-11
Typing

Arthur-Milchior:
Please let me know if you would prefer different name. 

As usual, I would expect some of those lines to be deleted by moving to rust, I'm fine with it


dae:
Looks good, thanks!

PR #733, opened 2020-08-13
Add unit test for negation and "and/or" templates

Arthur-Milchior:
I love that anki change its generation rules. I wanted to test them, in order to port unit test to ankidroid and being sure what changes to port.

The last assertion fails, I would love to get help to understand why. This is especially strange because doing the same test manually in anki does not fail.

### DIFF
--- a/pylib/tests/test_cards.py
+++ b/pylib/tests/test_cards.py
@@ -1,5 +1,7 @@
 # coding: utf-8
 
+from copy import deepcopy
+
 from tests.shared import getEmptyCol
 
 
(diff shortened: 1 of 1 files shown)


dae:
These tests would be simpler to implement in Rust, and would then not need to be ported to the other apps either. Take a look at nonempty() in template.rs - it should be easy to adapt to cover any cases that you think need covering.

dae:
(You can ignore renders_with_fields_for_reqs - that's the old behaviour based on "reqs")

Arthur-Milchior:
I don’t know when we’ll have rust and love the new behavior. So i’d prefer to port it so user have a consistent experience across device (i’m the user here)

dae:
That's fine, but these tests would still be simpler and easier to follow if they were in template.rs - please take a look at nonempty() if you have not already.

Arthur-Milchior:
I will. I am fine with both kind of tests.

However, as an add-on dev', I still would like to understand why this test fail, because if I use this method and don't get the expected result (or even a result consistent with what I see using the software) that's going to be hard to use.

dae:
If you print the model, you'll see why it's failing. By cloning existing fields instead of creating new ones, you're including the old ordinal from the field. Anki uses the old ordinals to detect when a field has been added vs renamed, and adjusts the templates for renamed fields. Either adding a fresh field or nulling ord should fix it.

dae:
..same with the templates. To avoid any field renames being processed, you can either restructure your code to preserve the existing Front field, or save the fields first.

```patch
diff --git a/pylib/tests/test_cards.py b/pylib/tests/test_cards.py
index 618b3dd18..1c1458c4a 100644
--- a/pylib/tests/test_cards.py
+++ b/pylib/tests/test_cards.py
@@ -160,17 +160,21 @@ def test_gen_not():
     model = models.byName("Basic")
     flds = model["flds"]
     fld0 = flds[0]
+    fld0["ord"] = None
     fld0["name"] = "First"
     fld1 = deepcopy(fld0)
     fld1["name"] = "Front"
     fld2 = deepcopy(fld0)
     fld2["name"] = "AddIfEmpty"
     model["flds"] = [fld0, fld1, fld2]
+    models.save(model)
+
     tmpls = model["tmpls"]
     # ensure first card is always generated,
     # because at last one card is generated
     tmpls[0]["qfmt"] = "{{AddIfEmpty}}{{Front}}{{First}}"
     tmpl = deepcopy(tmpls[0])
+    tmpl["ord"] = None
     tmpl["qfmt"] = """
 {{^AddIfEmpty}}
     {{Front}}
```

Arthur-Milchior:
thanks !

Arthur-Milchior:
> These tests would be simpler to implement in Rust,

I'll take time to learn Rust one day I guess. I assume I want to test rslib/src/notetype/cardgen.rs but it's not yet clear to me how to create a model and test it. So I guess if you want Rust it'll wait. Here is a version which works if you are interested

dae:
Since the tests did not reveal any issues and partially overlap the existing tests, I'm afraid it's probably not worth the maintenance burden to merge them in as is.

cardgen.rs is probably the file you want if you want to test the entire generation infrastructure, but template.rs is a better choice if you're more interested in whether a given template+fields will generate a card or not.

PR #734, opened 2020-08-14
Update to use the new version of actions/cache

evandrocoan:
Networking issues impacting cache save/restore
https://github.com/actions/cache/issues/267

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -179,7 +179,7 @@ jobs:
       # Disable it for macos because it pyenv cache is bugged (https://github.com/ankitects/anki/pull/563)
       - name: Cache pyenv
         if: matrix.os != 'macos-latest' && matrix.python == '3.7'
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         with:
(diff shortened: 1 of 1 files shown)


dae:
Thanks for keeping on top of these Evandro. Is the debugging code from yesterday still required, or should I revert it?

evandrocoan:
Before reverting we need to wait until the error `` can't find crate for `prost_derive` which `prost_build` depends on `` which happened on the pull request (https://github.com/ankitects/anki/pull/728) happens again.

If in some months it does not happens anymore, then, we could revert it because it could be that something random just went wrong on the cache unpacking.

But if you would like to not have that long Mac OS build running, we can try reverting it and let that code over a pull request. Once it happens again, we can try to replicate it over that pull request.

dae:
Error's back: https://github.com/ankitects/anki/runs/1010643070

evandrocoan:
I tried to run diff with a good cache and bad cache, but there were too my differences (due to bumped hashes being used as directories names by Rust). 

Let's hope that using GNU tar fixes it.

PR #735, opened 2020-08-16
Add hook for initializing answer buttons

brownbat:
A few add ons (MIA's, my ease factor add-on, More Buttons) change the number of answer buttons either up or down. So far all the ones I've found just use monkey patching, so I think a filter would be useful.

Bad news -- My testing environment is still not up and running,[0] and this is my first shot at this, so it definitely needs another pass by somebody who has implemented a hook before before merge.

Sorry it's not ready to go out of the box. Hopefully this draft inspires somebody to carry it over the finish line. Shouldn't be too much more than this skeleton.

[0] Mentioned in the Discord. Anki fails to import QtWebEngineWidgets with ModuleNotFound, but if I copy and paste that import into the interpreter or any other file, the import works fine somehow. I'm guessing some sort of circular dependency or something with the bash run / make scripts that I don't understand.

### DIFF
--- a/qt/tools/genhooks_gui.py
+++ b/qt/tools/genhooks_gui.py
@@ -59,6 +59,17 @@ hooks = [
         legacy_no_args=True,
     ),
     Hook(
+        name="reviewer_will_init_answer_buttons",
+        args=["reviewer: aqt.reviewer.Reviewer", "card: Card"],
+        doc="""Used to modify list of answer buttons
+
+        Return a tuple of the form ((1, "Label1"), (2, "Label2"))
+
+        import _ from anki.lang to support translation, as
+           ((1, _("Label1)), ...)
+        """,
+    ),
+    Hook(
         name="reviewer_will_answer_card",
         args=[
             "ease_tuple: Tuple[bool, int]",
--- a/qt/tools/genhooks_gui.py
+++ b/qt/tools/genhooks_gui.py
@@ -68,14 +68,14 @@ hooks = [
         return_type="Tuple[Tuple[int, str], ...]",
         doc="""Used to modify list of answer buttons
 
-        buttons_tuple is a sequence of buttons, with each button represented
-        by a tuple containing an int for the button's number and a string for
-        the button's label.
+        buttons_tuple is a tuple of buttons, with each button represented by a 
+        tuple containing an int for the button's ease and a string for the 
+        button's label.
 
-        Return a tuple of the form ((1, "Label1"), (2, "Label2"), ...)
+        Return a tuple of the form ((int, str), ...), e.g.:
+            ((1, "Label1"), (2, "Label2"), ...)
 
-        Note: import _ from anki.lang to support automatic translation, using,
-        e.g.,
+        Note: import _ from anki.lang to support translation, using, e.g.,
             ((1, _("Label1")), ...)
         """,
     ),
--- a/qt/tools/genhooks_gui.py
+++ b/qt/tools/genhooks_gui.py
@@ -61,19 +61,21 @@ hooks = [
     Hook(
         name="reviewer_will_init_answer_buttons",
         args=[
-            "buttons_tuple: Optional[Tuple[Tuple[int, str], ...]]",
+            "buttons_tuple: Tuple[Tuple[int, str], ...]",
             "reviewer: aqt.reviewer.Reviewer",
             "card: Card",
         ],
         return_type="Tuple[Tuple[int, str], ...]",
         doc="""Used to modify list of answer buttons
 
-        buttons_tuple is a tuple of buttons, with each button represented by a tuple
-        containing an int for the button's number and a string for the button's label.
+        buttons_tuple is a sequence of buttons, with each button represented
+        by a tuple containing an int for the button's number and a string for
+        the button's label.
 
         Return a tuple of the form ((1, "Label1"), (2, "Label2"), ...)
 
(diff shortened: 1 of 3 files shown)


dae:
Anki uses the python environment in the pyenv folder of your checkout. Perhaps you're adding QtWebEngineWidgets to your global python instance, or a separate venv?

brownbat:
> 
> 
> Anki uses the python environment in the pyenv folder of your checkout. Perhaps you're adding QtWebEngineWidgets to your global python instance, or a separate venv?

Oh, thanks, this sounds like exactly it. Never really developed anything outside of a short self-contained script, so I'm reading up on virtual environments now. Will let you know how it goes.

brownbat:
Ok, tested and fixed with those latest commits. 

This let me radically simplify YesOrNo.py, which is used in my add-on and MIA's, should help with other button add-ons too.

Demo code (using a just barely abbreviated version of that):
```
from aqt import mw
from aqt import gui_hooks

BUTTON_LABEL = ["button 0", "button 1"]

def two_button_mode(button_tuple, reviewer, card):
    button_count = mw.col.sched.answerButtons(card)
    if button_count in [2, 3]:
        button_tuple = ((1, BUTTON_LABEL[0]), (2, BUTTON_LABEL[1]))
    else:
        button_tuple = ((1, BUTTON_LABEL[0]), (3, BUTTON_LABEL[1]))
    return button_tuple

def remap_answers(ease_tuple, reviewer, card):
    button_count = mw.col.sched.answerButtons(card)
    new_ease = remap[button_count][ease_tuple[2]]
    return (ease_tuple[0], new_ease)

gui_hooks.reviewer_will_init_answer_buttons.append(two_button_mode)
gui_hooks.reviewer_will_answer_card.append(remap_answers)
```

dae:
Please resolve the typecheck errors and then I'll get this merged. Perhaps you were looking for Tuple[type, ...] rather than Sequence, or you could change the top level object to a list instead of a tuple.

brownbat:
And... learned a ton about type checking and mypy. 

1. I removed all the Optional[] wrappers I added, there's not a real way for an add-on to return None for the list of buttons and have it make any kind of sense.

2. The original _answerButtonList(self) was typed to Sequence[foo, ...], but when I tried to match that, mypy would error: Sequence should only have one argument. ie, ```Sequence[Tuple[foo, ...]]```.

I flipped everything to: 
`Tuple[Tuple[int, str], ...]`

I could build a type from the ground up though, if you think that would be more readable and the flexibility would be useful:

`ButtonType = Tuple[int, str]
ButtonListType = Iterable[Button]`

(It seems like the variable length for other iterables and sequences is implied, so the `...` is only used explicitly with Tuple[] as far as I can tell.)

qt/tools/genhooks_gui.py
```diff
@@ -58,6 +58,27 @@ def on_overview_will_render_content(overview, content):
...
+        args=[
+            "buttons_tuple: Tuple[Tuple[int, str], ...]",
+            "reviewer: aqt.reviewer.Reviewer",
+            "card: Card",
+        ],
+        return_type="Tuple[Tuple[int, str], ...]",
+        doc="""Used to modify list of answer buttons
+
+        buttons_tuple is a sequence of buttons, with each button represented
+        by a tuple containing an int for the button's number and a string for
```
glutanimate, line 72: Super minor, but I would suggest calling this value `ease' for consistency.

brownbat: > I would suggest calling this value `ease' for consistency.

Oh, fair point. How about ```ease_buttons_tuple``` though?

I just worry `ease` gets a little overloaded -- it's 

- the multiplier for the interval (usually called 'factor' in the code), 
- the numeric answer (1-4), 
- here I'm talking about the actual buttons with labels that point to that second meaning, so about half a step removed?

brownbat: Hmm, took a look at renaming, we have _answerButtonList() in the existing code before this, realized that's what I was matching. 

Changing this to `will_init_ease_buttons` or `ease_buttons_tuple` starts pulling on a thread... how far do we want to go? Refactor `_answerButtonList` to `_easeButtonList`?

Might be safer to leave it as submitted, I don't know.

glutanimate: Hmm, sorry, I think my comment might have been a bit too vague. I was only thinking about changing the naming in the docstring from "number" to "ease", since other hooks like `reviewer_will_answer_card` also use `ease` to describe the button index.

You're definitely right in that `ease` is a bit ambiguous in Anki's code base. I guess both terms have their merits here: We could go with `ease` and be ambigiously consistent with other hooks and Anki's variable naming, or go with `button number` or something similar and be more clear, but less consistent.

@dae, super minor point, so don't want to bother you with this too much, but any thoughts?

brownbat: Oh, yeah, docstring is the easy fix, definitely agree, updated.

dae: The new graphs code and Rust code are using 'button' in preference to 'ease'
https://github.com/ankitects/anki/blob/master/rslib/src/revlog.rs#L20

dae:
'make fix' should fix the formatting issue. Would you mind rebasing over master and force pushing as well, so we get a clean merge?

dae:
Actually, please scratch that rebase comment - it makes it harder to review.

brownbat:
Will remember to make fix next time, sorry about that.

dae:
I'll just fix it for you this time

PR #736, opened 2020-08-16
Create the hook scheduler_did_bury_notes

evandrocoan:
Create the hook `scheduler_did_bury_notes` to allow changing siblings buried cards of 'card'. 

Addon example: https://github.com/evandrocoan/AnkiAddons/blob/master/bury_source_siblings/__init__.py


dae:
Such an approach will not work with the v3 scheduler I'm afraid. It can tell you all cards that are due up front however, so I'd recommend you make your add-on bury at the start of the review session instead.

PR #740, opened 2020-08-20
Fix no sound after restarting mpv

kelciour:
Steps to reproduce:

1. Play some long audio file
2. Emulate crash by force closing mpv during the audio playback via the task manager
3. There's no sound anymore after clicking the audio play button


dae:
Thanks!

PR #741, opened 2020-08-21
Fix error reading scss;..\aqt_data\web: no such file or directory

evandrocoan:
Recently on my Windows, out of nowhere, it started throwing:
```
../../ts/node_modules/.bin/tsc --build
../../ts/node_modules/.bin/sass --no-source-map scss:../aqt_data/web/
Error reading scss;..\aqt_data\web: no such file or directory.
make: *** [Makefile:26: build] Error 66
```

### DIFF
--- a/qt/aqt/customstudy.py
+++ b/qt/aqt/customstudy.py
@@ -6,7 +6,7 @@ import aqt
 from anki.consts import *
 from anki.lang import _
 from aqt.qt import *
-from aqt.utils import showInfo, showWarning
+from aqt.utils import TR, showInfo, showWarning, tr
 
 RADIO_NEW = 1
 RADIO_REV = 2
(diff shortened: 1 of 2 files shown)


qt/ts/Makefile
```diff
@@ -23,7 +23,8 @@ all: check
 PHONY: build
 build:
 	$(BIN)/tsc --build
-	$(BIN)/sass --no-source-map scss:../aqt_data/web
+	$(BIN)/sass --no-source-map scss:css
+	rsync -a css ../aqt_data/web
```
dae, line 27: Is that correct? IIRC that will create a css folder inside the web folder.

Do you have a backup of an earlier lockfile to see which package change broke the behaviour?

evandrocoan: You are right. It broke. I have no lock file as I just did a clean install with `git clean -xfd .` (to ensure it did not have broke). I accidentally closed this.

PR #746, opened 2020-08-23
Create a gui_hook for initializing buttons in Models

hgiesel:
I have two add-ons which create an additional button in the "Models" dialog, one of them being [Asset Manager](https://ankiweb.net/shared/info/656021484).
I found it much better putting an add-ons configuration here over putting it into the Add-ons dialog, when the add-ons configurations are per note type.

### DIFF
--- a/qt/aqt/models.py
+++ b/qt/aqt/models.py
@@ -51,22 +51,21 @@ class Models(QDialog):
         box = f.buttonBox
 
         default_buttons = [
-            ("Add", self.onAdd),
-            ("Rename", self.onRename),
-            ("Delete", self.onDelete),
-            ("Options...", self.onAdvanced),
+            (_("Add"), self.onAdd),
+            (_("Rename"), self.onRename),
+            (_("Delete"), self.onDelete),
         ]
 
         if self.fromMain:
-            from_main_buttons = [
-                ("Fields...", self.onFields),
-                ("Cards...", self.onCards),
-            ]
+            default_buttons.extend([
+                (_("Fields..."), self.onFields),
+                (_("Cards..."), self.onCards),
+            ])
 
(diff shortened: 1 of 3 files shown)


qt/aqt/models.py
```diff
@@ -49,20 +49,26 @@ def setupModels(self) -> None:
...
-        b = box.addButton(_("Rename"), t)
-        qconnect(b.clicked, self.onRename)
-        b = box.addButton(_("Delete"), t)
-        qconnect(b.clicked, self.onDelete)
+
+        default_buttons = [
+            ("Add", self.onAdd),
+            ("Rename", self.onRename),
+            ("Delete", self.onDelete),
+            ("Options...", self.onAdvanced),
```
dae, line 57: You've broken the translations

qt/aqt/models.py
```diff
@@ -49,20 +49,26 @@ def setupModels(self) -> None:
...
-            b = box.addButton(_("Cards..."), t)
-            qconnect(b.clicked, self.onCards)
-        b = box.addButton(_("Options..."), t)
-        qconnect(b.clicked, self.onAdvanced)
+            from_main_buttons = [
+                ("Fields...", self.onFields),
+                ("Cards...", self.onCards),
+            ]
+
+            default_buttons[-1:-1] = from_main_buttons
```
dae, line 66: This is difficult to read; I'd prefer you just append or extend() these, and then append Options at the end

dae (changes requested):
No issues with the hook itself, but some comments below

hgiesel:
I don't quite see how I could fix the Ubuntu builds failing? I assume it is a build issue.

dae:
Looks like an old caching issue, should be good to go. Thanks!

PR #410, opened 2020-01-03
Make .ankiaddon files installable via the CLI, registering Anki as their default file type handler

glutanimate:
Extends Anki's CLI with the ability to install `.ankiaddon` files and registers Anki as their file-type handler. This will allow users to conveniently install downloaded add-on packages right from their file manager without walking through a more complex installation procedure.

In order to prevent accidental installations and address potential security concerns, users are prompted for confirmation before Anki parses the `.ankiaddon` file in any way. The messaging is the same as on AnkiWeb:

In their current state, these changes only assign Anki to `.ankiaddon` files on Linux. Further adjustments on the packaging side of things will be necessary to do the same on Windows and macOS.

Full disclosure: These changes were written as part of my work with AMBOSS and their rights of use lie with AMBOSS MD Inc.

PR #747, opened 2020-08-23
remove Optional when Optional is already implied

AlexRiina:
In a parameter assignment, if the default value is None, then `Optional[]` is implied

```python
# optional_test.py

def test(x: int = None):
    reveal_type(x)
```

```
$ mypy optional_test.py
optional_test.py:4: note: Revealed type is 'Union[builtins.int, None]'
```

I noticed this come up in https://github.com/ankitects/anki/pull/410#discussion_r362962695 and confirmed that Optional there is not required. This pull request removes all extra Optiona[] annotations.

Note: Optional is not implied in variable assignment like `x: int = None` though I'm not sure why not.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -47,6 +47,7 @@ Alexander Presnyakov <flagist0@gmail.com>
 abdo <github.com/ANH25>
 aplaice <plaice.adam+github@gmail.com>
 phwoo <github.com/phwoo> 
+Alex Riina <alex.riina@gmail.com>
 
 ********************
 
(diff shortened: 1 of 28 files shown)


dae:
I'm afraid this is relying on behaviour that is likely to change:

https://stackoverflow.com/questions/49724085/should-arguments-that-default-to-none-always-be-type-hinted-as-optional

AlexRiina:
Thanks! Didn't realize that was changing

dae:
No problem, thanks for contributing in any case :-)

PR #748, opened 2020-08-25
fix drag & drop issue when dropping things over existing content

ANH25:
See https://forums.ankiweb.net/t/drag-drop-images-wont-work/2630


dae:
Thanks for looking into this and the exporting issue ANH - much appreciated.

PR #749, opened 2020-08-26
Only autoplay added media when recording in editor

ANH25:
With the recent change in drag & drop code, dropping media files onto the editor screen outside fields still play these files even though they're not added to the note. This is because the dropped data is processed before detecting if the drop falls in the fields area.
Doing `doDrop`'s job in the drag & drop event handlers (and before calling `_processMime`) instead seems to be a bit involved since accepting or rejecting the events would depend on the result of the asynchronous evalWithCallback.

I think it's better to only autoplay media when recording. Autoplay in this case is useful to check how the recorded audio sounds, but in other cases users can simply open the media file they're adding if they want to check it.

A bit unrelated, but audio recording has a delay issue where actual recording starts several seconds after the recording message box appears (at least in my case).

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -741,6 +741,7 @@ to a cloze type first, via 'Notes>Change Note Type'"""
             )
             return
         if file:
+            av_player.play_file(file)
             self.addMedia(file)
 
     # Media downloads
@@ -758,7 +759,6 @@ to a cloze type first, via 'Notes>Change Note Type'"""
             name = urllib.parse.quote(fname.encode("utf8"))
             return '<img src="%s">' % name
         else:
-            av_player.play_file(fname)
             return "[sound:%s]" % fname
 
     def urlToFile(self, url: str) -> Optional[str]:


dae:
Thanks!

The recording delay is likely the time it takes for the audio interface to initialize on your system - we launch the recorder in a separate thread so don't know when it actually receives its first read. pyaudio can be a bit hard to install, so it may be worth investigating if we can use the Qt audio tools instead. We'll still only get a pcm file so will still require lame, but it may make the recording process a bit smoother.

https://doc.qt.io/qt-5/qaudiorecorder.html

macborow:
FWIW, I actually see value in playing the audio when a file is added to a card as it allows me to make sure that the recording matches the text and catch any typos (missing articles/classifiers/etc.). I'd prefer if there was a setting allowing me to decide whether to play the sound or not in this situation.

dae:
I think I saw someone else saying they preferred the old behaviour as well. ANH: maybe we should just roll this back?

ANH25:
I proposed this mainly to fix the aforementioned drag & drop issue, but it's pretty a minor one.
So I'm fine with rolling it back if people prefer the old behavior.

PR #751, opened 2020-08-30
Remove unicode isolation markers from add-on config editor title

ANH25:
They show as squares in some systems (e.g. Windows 7).

anki-demo-1


dae:
Thanks!

PR #753, opened 2020-09-02
add Dockerfile

kaczmarj:
[Docker](https://www.docker.com/) provides a standard way of installing software. This is particularly useful with software that requires fairly complex dependencies, like Anki. The Dockerfile added in this commit builds a Docker image with Anki. The Docker image is based on `python:3.8`, which is based on a Debian Buster image. This Docker image can be useful to end users, because it can be installed on any system that has Docker (e.g., Windows, macOS, Linux), and it is also useful to developers who might not have -- or want to install -- the dependencies that Anki requires at compile time.

To build the image, run `docker build --tag anki .` in the project root directory.

To run the image with a GUI (on Unix-like systems), run

```
docker run --rm -it --env "DISPLAY=$DISPLAY" --volume /tmp/.X11-unix:/tmp/.X11-unix anki
```

One may also mount a user's Anki directory into the container and run the container as the current user.

One may also run with [Singularity](https://sylabs.io/singularity/) (which typically has more transparent support for GUI applications and behaving as the current user) by first building the Docker image as above and then converting to Singularity Image Format (SIF) with

```
sudo singularity build anki.sif docker-daemon://anki:latest
```

Run the Singularity image with

```
singularity run anki.sif
```

I have uploaded an Anki Docker image (build using the Dockerfile in this pull request) to https://hub.docker.com/r/kaczmarj/anki. Here are instructions to download and use it:

```
docker pull kaczmarj/anki
xhost +local:root
docker run --rm -it --env "DISPLAY=$DISPLAY" --volume /tmp/.X11-unix:/tmp/.X11-unix kaczmarj/anki
xhost -local:root
```


dae:
Thanks Jakub!

I don't have much experience with Docker - will you be happy to handle any issues that crop up with it in the future if I merge this in?

You've used 'make build' here, which takes extra time to bundle up the wheel files and install them each time the source is change. Have you considered using 'make develop' and running directly from the source tree instead?

kaczmarj:
hi @dae 

>will you be happy to handle any issues that crop up with it in the future if I merge this in?

Absolutely. I have added my name and email in the `maintainer` label in the image to make it easier to find me. The labels can be found with

```
docker inspect --format='{{json .Config.Labels}}' anki:latest
```

and it is present in the Dockerfile.

>You've used 'make build' here, which takes extra time to bundle up the wheel files and install them each time the source is change. Have you considered using 'make develop' and running directly from the source tree instead?

Yes, I did consider this. I opted instead to create the Python wheel files, and then install them in a fresh Python Docker image. This dramatically reduces the size of the final Docker image (from several gigabytes to under one gigabyte).

One can also build only the first stage of the Dockerfile (the stage that installs compile-time dependencies and compiles Anki) by running this in the project root directory

```
docker build --tag anki --target builder .
```

Then, the image can be run with

```
docker run --rm -it anki /opt/anki/run
```

What do you think? Is that OK?

dae:
Would you mind adding a README.docker file with the instructions you've posted above? I think the edit+run+repeat cycle is going to be the more common use case (eg add-on authors that want to add and test a hook), so it might make sense to cover that use case higher up in the file.

kaczmarj:
@dae - I added a README.docker file that describes how to use Docker in the edit+run+repeat cycle. That document is geared towards Linux because that is all I have access to at the moment for testing.

dae:
Thanks for that Jakub, I'm sure some users will find this useful.

evandrocoan:
@kaczmarj, its great addition to Anki!

By the way, do you know about the usage of docker-composer? With it, I could run things just by typing `docker-compose up` instead of directly running a long docker command. As an example, I have this [testing repository](https://github.com/evandroforks/GithubRepositoryResearcher#deployment-docker) where I used `docker-compose`.

kaczmarj:
@evandrocoan - yes you could use `docker-compose` for this, but I would not. `docker-compose` is typically used to "for defining and running multi-container Docker applications" ([source](https://docs.docker.com/compose/)), and Anki is a single-container application. If you don't want to use the long `docker run` command, maybe set a shell alias or function, or write out the command once and for every other time, you can quickly search for the command with <kbd>Ctrl</kbd> + <kbd>R</kbd> and then type `docker run`.

kaczmarj:
For any readers here who would like to use anki on their computer through docker -- I included the following function in my `~/.bashrc` so I can call `anki` from the command line.

```
anki() {
    docker run --rm -it \
        --name anki \
        --volume $HOME/.local/share:$HOME/.local/share:rw \
        --volume /etc/passwd:/etc/passwd:ro \
        --user $(id -u):$(id -g) \
        --volume /tmp/.X11-unix:/tmp/.X11-unix:rw \
        --env DISPLAY=$DISPLAY \
        kaczmarj/anki "$@"
}
```

PR #755, opened 2020-09-03
Add hooks for next review interval override

robwe:
Two new hooks to allow overriding the next review interval


dae:
I'm afraid this will need to wait for a little while, as I'm currently in the process of refactoring the scheduling code.

What did you add the error message for?

robwe:
I created the add-on [Override next review interval for Easy answer button](https://ankiweb.net/shared/info/1499032797) which currently uses monkey patching. A user can enter the next review interval in days. The error message is to warn about an invalid entry.

dae:
This has been exposed via JS instead of Python: https://faqs.ankiweb.net/the-2021-scheduler.html#add-ons-and-custom-scheduling

PR #756, opened 2020-09-06
Added required Ubuntu packages to install instructions

robwe:
Starting from a fresh Ubuntu 20.04 image I had to additionally install python3-dev, python3-venv and libxcb-xinerama0 to get ./run working without errors.
I also added some tips to the documentation that would have helped me.


dae:
Thanks Robert, I've committed a slight variant of this.

PR #757, opened 2020-09-08
Tweak stats range box items' padding

abdnh:
Just a small padding tweak to fix a cramped radio button in RTL layout.


dae:
Unfortunately this won't work on the alternate Anki build, which uses Chromium 57: https://caniuse.com/?search=padding-inline-end

I'll just add a padding to both sides instead, which should solve the problem.

PR #758, opened 2020-09-14
Add sync_will_start and sync_did_finish hook

hgiesel:
These are two hooks which should take place:
* just before syncing starts
* after the sync of the collection finished (not necessarily media_sync)


dae:
I'd prefer the hook goes before the reset, as discussed on the forum

hgiesel:
Updated. Pretty sure you're referring to [this thread](https://forums.ankiweb.net/t/post-sync-hook/2093/6).

dae:
Sorry, I assumed it was you who created that thread.

PR #759, opened 2020-09-14
Config classes for simplying working with Anki settings

hgiesel:
These are classes which I started writing for some of my add-ons, as they provide a unified wrapper to some configuration interfaces in Anki.
It also avoids having to carry around the setting keywords, and default values.

For example one could be initializing a settings UI dialog in the addon manager, which is initialized with the current settings, and concludes by setting these options.
Then in some other file, you might once again read those options, and then use them in some fashion.

Without these classes, you always have to be aware, whether it was a profile setting, or a collections setting and then use the correct method for accessing/setting the value.

I wonder what you think about an interface like this?


dae:
It looks clean enough, but generally speaking I'd prefer not to include in the base program code that is not used by the base program, as it increases the maintenance burden. I'm also not sure that there's much confusion between them in practice - generally speaking I'd suggest using collection config for anything you want synchonized that is not too big, and add-on config or files in the add-on's folder for everything else.

PR #761, opened 2020-09-16
Revert mpv back to listening for idle callback

kelciour:
https://forums.ankiweb.net/t/anki-2-1-33-linux-audio-stops-working/3023


dae:
Thanks Kelciour, let's see how this goes.

PR #762, opened 2020-09-16
Change sip import

zjosua:
As discussed in [the forums](https://forums.ankiweb.net/t/webviews-blank-on-linux/1433), I've changed the sip import.
To be honest, I don't really understand what's going on here, but it works for me.

Tested on Debian bullseye 32 and 64 bit.

### DIFF
--- a/qt/aqt/qt.py
+++ b/qt/aqt/qt.py
@@ -24,10 +24,7 @@ from anki.utils import isMac, isWin
 os.environ["LIBOVERLAY_SCROLLBAR"] = "0"
 
 
-try:
-    from PyQt5 import sip
-except ImportError:
-    import sip  # type: ignore
+import sip  # type: ignore
 
 
 def debug():


dae:
If I recall correctly, I previously had to add the extra import statement because when building Qt from source, it wasn't working with the direct sip import. The PyQt docs say it should work for compatibility, so perhaps they changed the behaviour over time.

I'd prefer to understand the problem before making a change. Perhaps you have a system-installed version of sip that is being picked up? If that's the case, perhaps creating the pyenv folder with --no-site-packages might help?

zjosua:
Yes I actually have system installed sip 4.19.24 on the 32bit laptop.
But I need the system packages because PyPI doesn't offer a 32bit wheel for PyQt5.
Maybe I'll find the time to build PyQt5 for 32bit someday.

Btw on the 64bit laptop, PyQt5 installed from PyPI depends on [PyQt5-sip](https://pypi.org/project/PyQt5-sip/), not [sip](https://pypi.org/project/sip/). Only PyQt5-sip 12.8.1 is installed on that device.

dae:
I'd still like to know what the problem is, but am not able to reproduce the problems with importing sip directly that I encountered in the past, so will give this a go, thanks.

dae:
Actually I just noticed that it has broken the tests. Could you please ensure 'make check' works?

zjosua:
Actually I agree that we should find out more about the problem before pursuing this any further.

I built PyQt5 and PyQtWebEngine from source yesterday but didn't quite get it all to work. Since I switched back to the PyQt packages from the debian repo, I can now run Anki as it is in the master branch without any errors.

dae:
Will close for now, please post back if you feel it needs reopening.

PR #763, opened 2020-09-19
Change 'Show Both Sides' to 'Answer Side Only'

abdnh:
See https://forums.ankiweb.net/t/misnamed-button-in-card-previewer/3485

Should the autoplay code be changed? maybe to only play audio from answer side, or additionally play question side only in the presence of FrontSide, or maybe we shoud just call `_on_replay_audio` (in this case, playback when this is toggled on will depend on the “Always include question side when replaying audio“ deck option I think)?

### DIFF
--- a/qt/aqt/previewer.py
+++ b/qt/aqt/previewer.py
@@ -35,7 +35,7 @@ class Previewer(QDialog):
     _card_changed = False
     _last_render: Union[int, float] = 0
     _timer = None
-    _show_both_sides = False
+    _show_only_answer = False
 
     def __init__(self, parent: QWidget, mw: AnkiQt, on_close: Callable[[], None]):
         super().__init__(None, Qt.Window)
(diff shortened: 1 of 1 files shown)


glutanimate:
I'm not sure we gain much from this, while esp. the hotkey reassignment could irritate users used to the old one. Do we have any data on users finding this confusing?  If I was a user coming from physical flashcards or more traditional apps, I would assume that "answer only" would quite literally only show the answer.

Changing the internals of the setting will also break any add-ons hooking up to it.

If we do decide to go with a different wording, then I would propose to change it to "back side only" and retain the B hotkey. Ideally we'd also not change the API and config value.

But even with that wording, I'm not convinced that it will serve newer users better than the current one.

Maybe a simple "(Also) Show Answer" would be best? But again, not a huge fan of hotkey changes and updating internals to fit user-visible strings.

abdnh:
I was reluctant to change the hotkey, but changed it anyway for consistency.
I think your point about this change breaking some add-ons is valid (never thought of that).
I don't personally find 'Show Both Sides' confusing, but it could be for users who don't use FrontSide (like the poster in the forum thread).
Changing the toggle label without the setting should be enough but it will introduce some discrepancy with naming, so I'm actually no longer sure if this change is worth it.

abdnh:
I think a better solution would be to show both sides regardless of FrontSide. I assume this is more involved, but I'll experiment with it.

glutanimate:
TBH, I think the number of users who use custom templates without FrontSide is fairly small (though I don't have any hard data on that and am of course a bit biased given that I mostly navigate medical decks). Still, I'd be hesitant to introduce too much added complexity to cover rarer use cases like this.

I honestly think that, if we do think changing this is worthwhile, going with "Show back side (hotkey b)" would be best. It's understandable enough where users on the default templates know what to expect, while also vague enough to cover our bases for power users with non-FrontSide back sides.

As for the internals, I think that the existing config key and method names are understandable even with the change. However, if you do think it's worthwhile to go for 100% consistency I would definitely suggest that you search through GitHub to see if any add-ons published here use the methods that you would rename. It's likely not going to be many, since the previewer isn't touched by a lot of add-ons, but it's always good to check first and ideally notify the authors.

It can be very difficult for add-on authors to keep up with every commit, so a quick scan to estimate the breakage potential of API changes is always good.

abdnh:
I actually searched through Github after your first comment; there are just about a dozen of addon-ons using `previewBothSides`.
I see your point about 'Show back side' being a better wording. I think changing the label should be enough.
No need to break things.

dae:
Sorry if I was not clear; I'd intended for this to be a change in wording only. Like Glutanimate, I suspect that back templates without FrontSide are fairly rare, and I don't think the existing behaviour should be altered. But if a clearer wording is available, changing the wording won't break things. 

I think "Show Back Side" would be better as "Back Side Only" however, as enabling it disables the the front side/template from being shown, and the former wording sounds more like the opposite.

PR #766, opened 2020-09-21
[regression] deckconf: fix broken default ease factor

cyphar:
While moving the deck configuration to protobuf, a small numerical error
was introduced when converting between schema11 and schema15. This
resulted in new profiles having a default deck configuration such that
the ease factor was 2.5% -- the Python scheduling code doesn't like this
and caps it at 130% silently.

The net result is that all new profiles created since Anki 2.1.28 have a
default deck configuration where all cards are immediately given the
worst possible ease factor.

Ref: <https://forums.ankiweb.net/t/3575>
Fixes: 6114836484ca ("move deck config to protobuf")

### DIFF
--- a/rslib/src/deckconf/schema11.rs
+++ b/rslib/src/deckconf/schema11.rs
@@ -237,7 +237,7 @@ impl From<DeckConfSchema11> for DeckConf {
                 reviews_per_day: c.rev.per_day,
                 bury_new: c.new.bury,
                 bury_reviews: c.rev.bury,
-                initial_ease: (c.new.initial_factor as f32) / 10.0,
+                initial_ease: (c.new.initial_factor as f32) / 1000.0,
                 easy_multiplier: c.rev.ease4,
                 hard_multiplier: c.rev.hard_factor,
                 lapse_multiplier: c.lapse.mult,
@@ -298,7 +298,7 @@ impl From<DeckConf> for DeckConfSchema11 {
             new: NewConfSchema11 {
                 bury: i.bury_new,
                 delays: i.learn_steps,
-                initial_factor: (i.initial_ease * 10.0) as u16,
+                initial_factor: (i.initial_ease * 1000.0) as u16,
                 ints: NewCardIntervals {
                     good: i.graduating_interval_good as u16,
                     easy: i.graduating_interval_easy as u16,


cyphar:
I'm not sure that this change is backwards-compatible (in fact it probably isn't) -- if anyone saved their configuration they'd end up with an ease factor of `13000%` (due to the off-by-100 factor). I don't know if this means we'd need a new schema, but I'll leave that up to your judgement.

cyphar:
I'm not sure that this change is backwards-compatible (in fact it probably isn't and my guess is that if anyone saved their configuration they'd end up with an ease factor of `13000%` due to the off-by-100 factor). I don't know if this means we'd need a new schema, but I'll leave that up to your judgement.

Also we will need to add some mitigation for existing decks to make sure that we haven't permanently made someone's deck much harder than necessary -- for my use-case I just set all of the card eases to `2500` after resetting the configuration but if someone has lots of learned cards the existing intervals could make reviews debilitating for the foreseeable future. In principle one solution would be to just emulate all card reviews to calculate what the correct ease factor should be (had this bug not been introduced) but if a user changed their deck options after doing reviews, emulation would not produce correct behaviour.

dae:
Ugh, thanks for catching this. Those lines are not the problem per se, it's https://github.com/ankitects/anki/blob/133fe8518b3b19408b693235aca0d55beaada6ac/rslib/src/deckconf/mod.rs#L50 using 2.5 instead of 250. All the other multipliers are factors rather than percentages though, making this a footgun, so I'll look into changing it into a factor too. A schema upgrade will definitely be required to avoid breaking all the existing decks that were upgraded from an older version.

cyphar:
Yeah, I wasn't sure whether you'd originally intended the on-disk format to be in terms of multiplicative factors (`2.5 * x`) or percentages (`250% * x`) so I just flipped a coin and went with the former since it seemed to match everything else. Did you want to me to continue working on this, or did you want to sort it out yourself (since you know better much better than me how to deal with schema upgrades in Anki)?

dae:
Probably easiest if I handle this, but if you notice any problems with the patch, feedback would be appreciated. :-)

dae:
@david-allison-1 A heads up that if AnkiDroid will be starting with 2.1.30, you'll want to patch line 50 above to 250 for now.

dae:
Patch is up

PR #768, opened 2020-09-24
Fix previewer not refreshing when selecting multiple cards in browser

abdnh:
See https://forums.ankiweb.net/t/previewer-nonetype-error-for-audio-if-additional-card-selected/3638

The call to Editor.setNote in Browser._onRowChanged eventually results in a call
to Browser._renderPreview only when `note` is not None (i.e., when a single card is selected).
I just added an explicit call to Browser._renderPreview.


dae:
Thanks abdo, your continued help is greatly appreciated.

PR #769, opened 2020-09-25
Escape browser sidebar search terms

abdnh:
Special characters (`*`, `%`, `_`) in the browser sidebar search terms are not escaped.

I noticed this after this [forum post](https://forums.ankiweb.net/t/filtering-symbols/3692).

Another issue not fixed here: escape sequences don't seem to work when searching in fields (e.g. the asterisk is not escaped in `front:\*`).

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import html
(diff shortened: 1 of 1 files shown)


dae:
Thanks!

PR #770, opened 2020-09-27
NF: uses consts in col

Arthur-Milchior:


dae:
Please run 'make fix' then push again

dae:
Thanks

PR #772, opened 2020-10-02
added *.sh eol=lf to .gitattributes 

lukkea:
git wouldn't pick up the .sh files with crlf endings but I suspect it was my environment that caused this rather than crlf existing in the codebase.
this change should ensure such changes don't make it into the codebase.


dae:
Thanks!

PR #773, opened 2020-10-02
Process local ico files like other image formats

abdnh:
Local icons are being pasted as normal links (through _processText) in the editor even though they can be displayed with img tags.


dae:
Thanks abdo :-)

PR #774, opened 2020-10-02
fix: sizing issue with scale UI

androidpcguy:
Fixes ankitects/help-wanted#18. Height of webview should be the height
of the contents inside it. Previously, scale factors increased fixed
height of the container webview, thus creating empty space (e.g.
at 200% scaling, height of webview was double the height of the inner
HTML page). Now height of webview is always the height of the contained
element.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -60,6 +60,7 @@ phwoo <github.com/phwoo>
 Soren Bjornstad <anki@sorenbjornstad.com>
 Aleksa Sarai <cyphar@cyphar.com>
 Jakub Kaczmarzyk <jakub.kaczmarzyk@gmail.com>
+Akshara Balachandra <akshara.bala.18@gmail.com>
 
 ********************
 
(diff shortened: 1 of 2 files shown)


dae:
@evandrocoan looks like http://repo.msys2.org/msys/x86_64/rsync-3.1.3-1-x86_64.pkg.tar.xz is timing out; I presume things will fix themselves when they get the site back up

@androidpcguy Thanks Akshara! I am actually surprised this works, as I seem to recall that simply setting the height failed in the past. But perhaps a tooltip upgrade or changes in other parts of the code have changed things - I've tested Qt 5.12 and 5.9 and it seems to be doing the right thing.

evandrocoan:
They are trying to reboot the server, but it not coming up. I think they will have to go to it physically: https://github.com/msys2/MSYS2-packages/issues/2171

evandrocoan:
You can download a rsync now from this mirror: https://mirror.yandex.ru/mirrors/msys2/msys/x86_64/ and put it on the repository https://github.com/ankitects/pyaudio-windows so these problems do not happen again.

PR #776, opened 2020-10-03
Paste links as plain text in basic mode

abdnh:
I think it's useful to be able to paste links as plain text without Anki processing them specially.


dae:
Looks good, thanks!

PR #777, opened 2020-10-03
Separate player for videos on Windows

kelciour:
Create a new mpv process for videos on Windows to prevent an occasional Anki freeze once the video window had been closed.

https://forums.ankiweb.net/t/anki-crashes-because-of-mpv/3359/13


dae:
Thanks kelciour!

PR #778, opened 2020-10-03
Always focus main web instead of bottom web in reviewer

hgiesel:
The reviewer by default usually focuses the bottom bar, unless there's a typing field in the card.

However there are many cases, where you'd still prefer focusing the main web instead of the bottom bar: That is, if there are key event listeners on the card, which some card types do (like Image Occlusion Enhanced, and my Closet card types).

There's this [add-on](https://ankiweb.net/shared/info/1642550423), which realizes this by using the gui hooks. I have used this add-on for a very long time now. And I don't really see the point of the Anki built-in functionality as opposed to what this add-on does.

Key events sent to the main web can still be received by the bottom bar (1, 2, 3, 4, Space, and E, etc.), so it seems like no functionality is lost.

### DIFF
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -201,9 +201,7 @@ class Reviewer:
         self._drawFlag()
         self._drawMark()
         self._showAnswerButton()
-        # if we have a type answer field, focus main web
-        if self.typeCorrect:
-            self.mw.web.setFocus()
+        self.mw.web.setFocus()
         # user hook
         gui_hooks.reviewer_did_show_question(c)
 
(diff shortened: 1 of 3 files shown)


dae:
Unfortunately this means the highlighting of the Show Answer button and Good button is lost, which is nice to have, as it directs the user's attention to the button they are most likely to use. Maybe it could be approximated with some CSS?

hgiesel:
I added a CSS class "focus" which mimics the Chromium focus highlight on focus.
This means that this kind of focus is not lost by clicking anywhere else in the GUI. I would actually consider that a benefit, because it keeps the UI a bit more consistent.

Additionally, if the focus actually falls into the button bar, the artificial focus is removed, and it will simply follow the normal focus.

dae:
Thanks hgiesel, this is great.

PR #779, opened 2020-10-04
Fix "Sign Up" link location

david-allison-1:
String states "Sign Up", but currently points to the "Sign In" screen: https://ankiweb.net/account/login

https://github.com/ankitects/anki/blob/aff28a38e56356582179c3330c3de00e02ae2ec4/rslib/ftl/sync.ftl#L38-L42

----

Untested as this is a trivial fix


dae:
Thanks David!

PR #781, opened 2020-10-04
Refactor Buttons in editor and add editor_did_init_left_buttons gui_hook

hgiesel:
This will adds a new parameter to the `Editor._addButton` and `Editor.addButton` methods, so that they can used for the "Fields..." and "Cards..." buttons in the editor as well.

It also adds the `editor_did_init_left_buttons` hook, which works analogously to `editor_did_init_buttons`.

My personal use case is for the [Asset Manager](https://ankiweb.net/shared/info/656021484) add-on, which allows you to manage assets per note type. Adding a button directly in the editor, which lets you access these assets would be a nice feature.

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -216,7 +216,7 @@ class Editor:
 
     def addButton(
         self,
-        icon: str,
+        icon: Optional[str],
         cmd: str,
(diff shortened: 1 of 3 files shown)


dae:
Looks good, thanks!

PR #783, opened 2020-10-07
Fix help-wanted#20

abdnh:
This fixes ankitects/help-wanted#20 in addition to an issue introduced in 0c340eba645b4babf9a6bda60160ee08b4e91854 which broke the 'Filter' menu item in deck filters.


dae:
Thanks for the quick fix!

PR #784, opened 2020-10-07
Rename night mode to dark mode

FrazerClews:
Night mode usually refers to the reducing the amount of blue light from your device, where as dark mode is making the UI a dark theme

Also, removed trailing spaces from changed files


abdnh:
Won't this break users' custom styling that rely on the presence of the `nightMode` class?

FrazerClews:
> Won't this break users' custom styling that relies on the presence of the `nightMode` class?

I feel at least `<string>Night mode</string>` in `qt/designer/preferences.ui` should be changed at least which shouldn't break anything I can think of, and fix the correct terminology to the user. Maybe the rest can be changed during a more major release so the code makes more sense down the road, with marking the `nightMode` class as deprecated in the meantime?

dae:
I'm afraid I don't think this is worth breaking translations or existing code for. If I could go back in time I'd pick dark mode instead, but it's been in use for years now, and I think the meaning is clear to most people. Thanks for taking the time to contribute in any case!

PR #786, opened 2020-10-09
added notes for running Anki from PyCharm with dev profile

lukkea:


dae:
While not necessarily bad advice, it's not really specific to the Windows version, and "upgraded" comes across as a bit snarky :-)

lukkea:
> While not necessarily bad advice, it's not really specific to the Windows version, and "upgraded" comes across as a bit snarky :-)

Thanks for the feedback; I've been known to come across in ways I would never want to - this is one of those times. **_Zero_** snark intended.

Great point about it not being Windows specific. I'm going to give it another go because I think it's helpful and may save others a few moments of concern and time reverting... Please feel free to tell me to stop if it's not helping though!

dae:
Thanks!

PR #787, opened 2020-10-10
Switch constants to enum

hgiesel:
This is by no way complete, but meant as a suggestion/idea.

This way it would allow parts of the code to use more descriptive types (e.g. `CardTypes` class rather than simply `int`)

Using `globals().update(CardTypes.__members__)` allows for backwards compatibility, by exporting the enum members to file scope (thus making them importable).

Using `IntEnum` we can make enum classes which to some degree mimic normal `int` values, e.g. `CardTypes.CARD_TYPE_NEW == 0` would equal `True`.

What breaks is any cases, where you would want to serialize the enum constant. Instead you'd have to cast it to `int` explicitly like so `CardTypes.CARD_TYPE_NEW.value`. Also other edge cases, like `type(CARD_TYPE_NEW) == int` would change, now being `False`.

What do you think about this?

### DIFF
--- a/pylib/anki/consts.py
+++ b/pylib/anki/consts.py
@@ -2,87 +2,115 @@
 # License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
 from typing import Any, Dict
+from enum import IntEnum
 
 from anki.lang import _
 
 # whether new cards should be mixed with reviews, or shown first or last
-NEW_CARDS_DISTRIBUTE = 0
-NEW_CARDS_LAST = 1
-NEW_CARDS_FIRST = 2
+class NewCardMixType(IntEnum):
+    NEW_CARDS_DISTRIBUTE = 0
+    NEW_CARDS_LAST = 1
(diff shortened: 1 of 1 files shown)


dae:
I think it's best to hold off on this for now - more of the scheduling code will be moving to Rust and/or Typescript, so some work spent updating the scheduling code at the moment may go to waste. If this does go ahead in the future, it'd probably be best to pick one of the less frequently used constants, and convert the existing references to it as well, so we can see how it plays out ergonomically in practice. Also, I like that it preserves backwards compatibility, but CardType.CARD_TYPE_xxx is redundant - CardType.xxx would be nicer.

hgiesel:
Of course, makes sense.

> Also, I like that it preserves backwards compatibility, but CardType.CARD_TYPE_xxx is redundant - CardType.xxx would be nicer.

In this case, we would have to make the legacy support more verbose. Basically `CARD_TYPE_xxx = CardType.xxx`.
We even could have this be `CARD_TYPE_xxx = CardType.xxx.value`, to preserve `int` as the type`.

PR #788, opened 2020-10-10
Add .opus as an audio extension

pkubowicz:
Wikimedia Commons has pronunciation recordings with .opus extension.
Anki is able to play such files, but until now adding them required
changing the extension to .ogg or some other supported one.

Also sort the audio extensions list.


dae:
Thanks!

PR #789, opened 2020-10-11
Decode pasted URLs for nicer display

abdnh:
URLs with escaped foreign characters are hard to identify and not nice to look at :)


dae:
Looks good, thanks!

PR #792, opened 2020-10-12
Add command-line argument to skip addons

nwwt:
Shift didn't work for me occasionally, for whatever reason. So just add an -s to skip that way.

(Also, update my github nick)


dae:
Please call it 'safemode' instead, and mention that it also prevents automatic syncing

nwwt:
right, done

dae:
Thanks!

PR #604, opened 2020-05-03
Created the AnkiMediaQueue for JavaScript media support

evandrocoan:
JavaScript media support should be a great addition to Anki in cases the standard player is somehow not working:
1. https://anki.tenderapp.com/discussions/ankidesktop/41235-how-to-add-video-saved-in-my-mac-to-my-deck-or-cards
1. https://anki.tenderapp.com/discussions/ankidesktop/41438-error-on-database
1. https://anki.tenderapp.com/discussions/ankidesktop/41106-card-audio-playback-changes-bug-in-212x
1. https://anki.tenderapp.com/discussions/ankidesktop/41344-o-audio-dos-meus-flash-cards-no-esta-funcionando

The pull-request https://github.com/ankitects/anki-manual/pull/128 Created the ankimedia documentation section on src/media.md, adds documentation for this feature. The `ankimedia` object implementation and automated integration tests are split in a second repository, https://github.com/evandrocoan/AnkiMediaQueue, and shipped with Anki only the build typescript file.

PR #793, opened 2020-10-14
Audio & object tag support

nwwt:
This remedies two issues to allow both tags — they're already fully supported (& observed) by Anki[Droid]'s WebView — to be used freely:

1. It makes Anki's media check recognize them in fields / not wipe their files.
2. It sets the WebView autoplay policy accordingly so `<audio>` can be made to play just like `[sound:…]`.

I do not suggest adding official support, this is just to let users / deck creators utilize the functionality. IOW, like Javascript, let it be _"provided without any support or warranty."_

Use case examples:
* `<audio>` allows playback to be dynamically/arbitrarily jscripted/controlled.
* `<object>` allows to load SVGs so that embedded CSS with user-interaction pseudo-classes is observed.

### DIFF
--- a/pylib/anki/media.py
+++ b/pylib/anki/media.py
@@ -33,8 +33,8 @@ def media_paths_from_col_path(col_path: str) -> Tuple[str, str]:
 
 class MediaManager:
 
-    soundRegexps = [r"(?i)(\[sound:(?P<fname>[^]]+)\])"]
-    htmlRegexps = [
+    sound_regexps = [r"(?i)(\[sound:(?P<fname>[^]]+)\])"]
+    html_media_regexps = [
         # src element quoted case
         r"(?i)(<[img|audio][^>]* src=(?P<str>[\"'])(?P<fname>[^>]+?)(?P=str)[^>]*>)",
         # unquoted case
@@ -44,7 +44,7 @@ class MediaManager:
         # unquoted case
         r"(?i)(<object[^>]* data=(?!['\"])(?P<fname>[^ >]+)[^>]*?>)",
     ]
-    regexps = soundRegexps + htmlRegexps
+    regexps = sound_regexps + html_media_regexps
 
     def __init__(self, col: anki.collection.Collection, server: bool) -> None:
         self.col = col.weakref()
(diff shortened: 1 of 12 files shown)


dae:
I'm happy to handle the tags in check media, as that's not easily extended by add-ons - but sound tags don't belong in the images section :-)

I'm not so sure about the calls to setPlaybackRequiresGesture - if the user has multiple audio tags set to autoplay, they'll all play at once. If you weren't aware of it, Evandro has a more comprehensive solution in #604

nwwt:
>but sound tags don't belong in the images section :-)

True ofc, I forgot to ask if it's alright to just rename the variable (at least in python) since I'm not sure how to best handle an extra one in rust.

>if the user has multiple audio tags set to autoplay,

Frankly, that's ⓐ working as the standard intends and ⓑ the user's own fault if they don't, just like if they would do something silly with javascript.

>Evandro has a more comprehensive solution in #604

IDK, it seems convoluted to me, but that might just be my bias from not being the author (and having my own solution in my AnkiWeb account's *_japanese.js* that's at ~150 LOC, dunno exactly since it mixes a few note features).

Though I believe their might be a real issue in that it's not possible to include multiple files in one field in his notation?

nwwt:
Any update regarding my last reply and the variable rename?

dae:
Sorry for the delay, I was busy with the Bazel work. 

The changes look pretty good, but would you mind making a few tweaks before this gets merged?

- htmlRegexps → html_media_regexps
- HTML_TAGS → HTML_MEDIA_TAGS
- escapeImages → escape_media_filenames + docstring
(+ an alias to the old escapeImages, as some add-ons use it)
strip_html_preserving_image_filenames → strip_html_preserving_media_filenames

nwwt:
Done, hope it's all good.

pylib/anki/media.py
```diff
@@ -159,7 +163,11 @@ def strip(self, txt: str) -> str:
         return txt
 
     def escapeImages(self, string: str, unescape: bool = False) -> str:
-        "Apply or remove percent encoding to image filenames."
+        "escape_media_filenames alias for compatibility with add-ons."
+        return self.escape_media_filenames(string, unescape)
```
dae, line 167: Thanks for making those changes. For future reference, you can do this more succinctly with a variable:

https://github.com/ankitects/anki/blob/e3eca1c5637b6b5e285e11c69d50623d4bb85d27/pylib/anki/decks.py#L368

PR #794, opened 2020-10-14
Don't apply suggestion when no suggestion list is shown in tagedit

abdnh:
Pressing Enter on a tagedit selects a suggestion even when the suggestion list is not shown in some cases. This could cause confusion especially in the deck override option, where clicking Enter (with no text entered) selects some deck and closes the widget without giving any indicator that the option was activated.


dae:
Thanks!

PR #795, opened 2020-10-14
Set card to None in reviewer cleanup

abdnh:
Is there a reason reviewer.card is kept after cleanup? This causes the browser to show the card whenever it's opened.
This came up in this [forum thread](https://forums.ankiweb.net/t/stop-anki-from-showing-the-next-card-when-opening-browser-from-the-study-screen/2651) previously.
I originally thought this was causing an importing problem I had but I was wrong, then I figured this could be a useful change anyway.


dae:
Probably just an oversight - thanks :-)

PR #797, opened 2020-10-15
Fix notes being imported to the wrong deck in some cases

abdnh:
See https://forums.ankiweb.net/t/bug-notes-are-imported-to-the-wrong-deck-in-certain-cases/4318

I also removed some unused code in modelChanged.


dae:
Thanks!

PR #799, opened 2020-10-19
Introduce Browser.search_for and Browser.update_search

hgiesel:
There are 4 methods on `Browser` which search while avoiding using `_onSearchActivated`:
- `setupSearch`
- `_selectNotes`
- `dupeLinkClicked`

I've generalized these, plus `_onSearchActivated` to use a method called `search_for` instead.
This method is also really useful for add-on developers to search something in the browser for the user, and give the appropriate visual feedback.

`update_search` is also a useful method for the add-on developer, to write to history without invoking `_onSearchActivated`.


dae:
Sounds good, thanks Henrik.

PR #800, opened 2020-10-19
Different first weekday support

hgiesel:
This is about the calendar graph in the Stats screen.
By default, weeks start there on Sundays, however switching it to another first week day would be quite trivial.

I think the only problem here is how should the first weekday be decided? Solely by locale, or make it settable in the preferences.

If the locale should decide over it, then I implemented a first draft of a method, however it would still need to account for many more languages.

### DIFF
--- a/qt/aqt/preferences.py
+++ b/qt/aqt/preferences.py
@@ -23,6 +23,7 @@ class Preferences(QDialog):
         self.silentlyClose = True
         self.prefs = self.mw.col.backend.get_preferences()
         self.setupLang()
+        self.setup_first_weekday()
         self.setupCollection()
         self.setupNetwork()
         self.setupBackup()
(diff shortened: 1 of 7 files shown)


dae:
A config setting would be the more flexible option, but is it something people care strongly about? GitHub have managed to put it off for 4 years: https://github.com/isaacs/github/issues/644

hgiesel:
> GitHub have managed to put it off for 4 years

Actually that's funny, because I never noticed that before, however when using the new Calendar view in Anki, it's almost the first thing I noticed. I think that's due to the different ways that I would use the Anki stats.

The GitHub calendar section to me, is more like showing me, and the other people, when and how active I was on GitHub. "How many contributions did I make last Monday?", is not really something I would ever look up.
However "How many reviews did I do last Monday? (or generally weekday X)" is a thing that I look up quite frequently on Anki.

hgiesel:
I've pushed a new version, which uses a config setting in `mw.pm.meta`, similar to the language.
It also adds a few strings which would need to be translated, however.

Concerning, how much people care about it: Obviously I cannot really say for sure, but I've set out to add some improvements to the stat screen, which I "tend to notice".
I also made [a poll](https://www.reddit.com/r/medicalschoolanki/comments/jizopp/poll_how_often_do_you_use_the_stats_screen_in_anki/) trying to figure out some stats on gripes people actually have with the stats screen (no pun intended).

dae:
Sorry for not getting back to you sooner. Rather than stashing this in the query string, it's probably better returned as part of the graph data by the Rust backend, as that means it will update each time the graph is refreshed, and automatically carry over to other platforms - the query string reading was a hack and I'd like to migrate away from it in the future.  While I take your point about it being different to GitHub,  this is not something that has been requested by anyone else so far, so I'm bit wary of adding another option in the preferences screen at this point - so I would lean towards not exposing it in the UI at the moment, and just having it as something that can be set using the debug console. Are you aware of cultures where something other than Sunday/Monday is the norm?

Please hold off on further changes for now, as the ts folder has some renames in the pipeline that will cause a merge conflict. Once they are pushed out, it should be safe to proceed.

hgiesel:
Ok, I totally get your point.

Concerning the weekday: I used [this Wikipedia section](https://en.wikipedia.org/wiki/Week#Days_of_the_week) as my basis, which states that most Muslim-majority countries regard Friday as their first weekday, and the Maldives (?) regard Saturday as their first weekday.

dae:
Thanks, learnt something new!

dae:
Should be safe to proceed on this now

PR #801, opened 2020-10-21
Strip double quotes from names

abdnh:
See https://github.com/ankitects/anki/pull/798

Is it on purpose that the backend allows creating note types, fields, etc. with empty names?

### DIFF
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -581,7 +581,7 @@ class CardLayout(QDialog):
 
     def onRename(self):
         template = self.current_template()
-        name = getOnlyText(_("New name:"), default=template["name"])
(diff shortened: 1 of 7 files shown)


dae:
Thanks abdo! We should probably be ensuring the names are non-empty as well, though the GUI code should be covering the common cases already.

abdnh:
I mistakenly used strip instead of replace here in the Python code. I was about to push a fix.

dae:
Not a problem, I missed it too. Would you like me to revert, or await a follow-up?

abdnh:
> We should probably be ensuring the names are non-empty as well

I will have a try at this and fix the strip issue along the way.

PR #802, opened 2020-10-22
Ensure non-empty names in the backend

abdnh:
See https://github.com/ankitects/anki/pull/801


dae:
Thanks for the quick fix!

PR #804, opened 2020-10-26
Set MIME type for javascript files

hgiesel:
Right now, when the reviewer requests a file ending in `.js`, it is not certain, whether the web view will actually recognize it as a application/javascript file, or text/plain.

The Anki on my Mac does it correctly, however [Windows doesn't seem to do it](https://github.com/hgiesel/closet/issues/33). This prevents the user from using dynamic importing via `import()`.


dae:
Thanks!

PR #805, opened 2020-10-27
align rslib get_subnode impl with anki pythonapi find_deck_in_tree

cecini:
Fix nestdeck due counts issue
Here ankidesktop will use get_subnode in sanity_check to get the schedduecounts, but if other client or server impl use the anki pythonapi find_deck_in_tree when using nestdeck, then will alway fail the comparison between server and client's  sanitycheck results and lead to the unexpected scene.

### DIFF
--- a/rslib/src/decks/tree.rs
+++ b/rslib/src/decks/tree.rs
@@ -175,9 +175,12 @@ fn hide_default_deck(node: &mut DeckTreeNode) {
 }
 
 fn get_subnode(top: DeckTreeNode, target: DeckID) -> Option<DeckTreeNode> {
+    if top.deck_id == target.0 {
+        return Some(top);
+    }
     for child in top.children {
-        if child.deck_id == target.0 {
-            return Some(child);
+        if let Some(node) = get_subnode(child, target) {
+            return Some(node);
         }
     }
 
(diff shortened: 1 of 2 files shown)


rslib/src/decks/tree.rs
```diff
@@ -175,10 +175,17 @@ fn hide_default_deck(node: &mut DeckTreeNode) {
 }
 
 fn get_subnode(top: DeckTreeNode, target: DeckID) -> Option<DeckTreeNode> {
+    if top.deck_id == target.0 {
+        return Some(top);
+    }
```
dae, line 180: This part should not be necessary; the provided deck id should never be 0.

rslib/src/decks/tree.rs
```diff
@@ -175,10 +175,17 @@ fn hide_default_deck(node: &mut DeckTreeNode) {
...
     for child in top.children {
         if child.deck_id == target.0 {
             return Some(child);
         }
+        let subnode: Option<DeckTreeNode> = get_subnode(child, target);
```
dae, line 185: This could be expressed more elegantly as:

```rust
if let Some(node) = get_subnode(child, target) {
  return Some(node)
}
```

dae:
Thank you, happy to merge once the above issues and your CONTRIBUTOR edit are fixed.

When submitting a pull request, GitHub Actions will check that the Git email you
are submitting from matches the one you used to edit this file. A common issue
is adding yourself to this file using the username on your computer, but then
using GitHub to rebase or edit a pull request online. This will result in your
Git email becoming something like user@noreply.github.com. To prevent the
automatic check from failing, you can edit this file again using GitHub's online
editor, making a trivial edit like adding a space after your name, and then pull
requests will work regardless of whether you create them using your computer or
GitHub's online interface.

cecini:
Sorry not familar with the github rebase,so need another commit.

dae:
Thanks cecini!

PR #806, opened 2020-10-28
Upgrade some Rust Dependencies

alanhdu:


dae:
Thanks as always Alan! Will return to this once I work through some changes to the build system.

dae:
I've manually merged these into master, and updated pyo3 to the latest git so that rspy no longer depends on a specific Python version.

PR #807, opened 2020-10-29
Add Ctrl+Shift combo to update filters in card search

RumovZ:
If a deck, note type or tag filter is clicked with the control and shift modifiers, corresponding filters in the current search expression will be replaced with the clicked filter.
E.g., ctrl+shift clicking a deck “new deck” transforms the current search expression `is:due deck:"old deck"` to `is:due deck:"new deck"`.

(Also fixes two minor bugs where clicking “Whole collection” with modifiers lead to invalid or undesired searches and the shift modifier didn't check for the default search prompt.)

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -1231,11 +1231,13 @@ QTableView {{ gridline-color: {grid} }}
                 # update instances of 'key':'anything' in the old search
                 if mods & Qt.ControlModifier and mods & Qt.ShiftModifier:
                     if len(args) == 2:
-                        quoted = "\"" + args[0] + r":(\\\\|\\\"|[^\"])*\""
+                        quoted = '"' + args[0] + r":(\\\\|\\\"|[^\"])*\""
                         partially = args[0] + r":\"(\\\\|\\\"|[^\"])*\""
                         unquoted = args[0] + r":[^\)\s\"]*(?=($|\)|\s))"
                         before = r"(?<![^\s\(])(?<!\\\()-?"
-                        pattern = re.compile(f"{before}({quoted}|{partially}|{unquoted})")
+                        pattern = re.compile(
+                            f"{before}({quoted}|{partially}|{unquoted})"
+                        )
                         txt = pattern.sub(txt, cur)
                 elif mods & Qt.ControlModifier:
                     txt = cur + " " + txt
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -1220,16 +1220,27 @@ QTableView {{ gridline-color: {grid} }}
                     items.append(txt)
                     txt = ""
             txt = " ".join(items)
-        if self.mw.app.keyboardModifiers() & Qt.AltModifier:
-            txt = "-" + txt
-        if self.mw.app.keyboardModifiers() & Qt.ControlModifier:
+        # is there something to replace or append with?
+        if txt:
+            mods = self.mw.app.keyboardModifiers()
+            if mods & Qt.AltModifier:
+                txt = "-" + txt
+            # is there something to replace or append to?
             cur = str(self.form.searchEdit.lineEdit().text())
             if cur and cur != self._searchPrompt:
-                txt = cur + " " + txt
-        elif self.mw.app.keyboardModifiers() & Qt.ShiftModifier:
-            cur = str(self.form.searchEdit.lineEdit().text())
-            if cur:
-                txt = cur + " or " + txt
+                # update instances of 'key':'anything' in the old search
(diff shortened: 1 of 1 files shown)


dae:
I'm not sure about this I'm afraid - what does that operation mean if there are multiple deck references for example? Regexs tend to be difficult to follow, so if we were to add this feature, I think it would be better done in the Rust code, by applying a transformation to the parsed AST, then serializing it back out to a string again. It's more work, but would be easier to maintain going forward.

RumovZ:
> I'm not sure about this I'm afraid - what does that operation mean if there are multiple deck references for example?

You mean like `deck:one or deck:two`? It would replace all instances, so in general it wouldn't be of much use in a case like this. You could of course use more input information like selection or the position of the cursor for more pinpoint replacement. But that would require much more modifications and my reasoning was that complicated searches are just a small fraction of the searches performed by the users. So my goal was to make the simple ones more user-friendly. From my own experiencing, it's quite often that you want to search for the same thing in different decks consecutively.

> Regexs tend to be difficult to follow

Would it help to annotate them with the verbose flag? Also, they could be broken down further and the parts be assigned to variables with meaningful names.

> I think it would be better done in the Rust code, by applying a transformation to the parsed AST, then serializing it back out to a string again.

I wouldn't mind writing a backwards parser, but it seems a bit overkill just for some string replacements and I don't have any other practical uses coming to mind. Do you? No doubt, in principle all sorts of cool transformations could be done with that.

RumovZ:
So one benefit I could come up with is the ability to normalise the user's input. That would help him understand how his search was interpreted in case he got unexpected results. E.g., for some time I was wondering what `--deck:Something` did. If Anki had transformed it into `-"-" and deck:"Something"`, that would have cleared things up.

However, I think it's vital to first rework the parser before building anything on top of it. As I had to learn when I worked with it, it has quite a few inconsistencies and is not always in line with the documentation.
So if you think this is the way to go, let me know, @dae. Then I'll make a small PR for the bugs mentioned above and open a topic on the forum to discuss how Anki's search syntax is supposed to work.

dae:
Would you mind elaborating on any other issues you've found?

Normalizing the user's input does sound like it could be useful, though it would require the parsing to succeed.

I was thinking if the backend exposed methods to convert a search string to a JSON representation and back again, then that would also be useful to add-ons that wish to transform the search as well. For example, the ignore accents add-on currently tries to convert bare text references to nc:xxx, and would benefit from not having to try to parse the search text itself.

RumovZ:
I've actually collected a few examples:

1. Single `\` matches nothing but doesn't throw an error, either.
1. `\\*` matches literal `\*` (instead of `\` and then anything).
1. `%` is treated like a wildcard.
1. `(` can be matched anywhere, but `)` only in quotes.
1. If unquoted, `\"text\"` throws an error.
1. `a"b"`, `"a""b"`, `"a"(b)`, `(a)(b)` are all valid, but `a(b)` is not.
1. `tag:"a c"` doesn't match anything, but `tag:a_c` and `tag:"a c*"` (apart from matching the expected) match cards with tags `a` and `c` unless they have a tag `b`.

Those are not all bugs, but they reflect the inconsistent handling of escape sequences and special characters which can give the user a hard time. E.g., even if you have read the manual, you will be surprised by the fact that your search for `99%` doesn't yield the expected results.

dae:
Thanks for that, and sorry for the delay. Fixes that address some of those issues but don't make the parser significantly more complicated would be appreciated, but it may be a case of diminishing returns for some of these.

% and _ as a wildcard have existed for a long time, so we may break saved searches for some users if we treat it differently. That said, it may still be worth treating them literally in the future - they're an implementation detail that leaked through, rather than something that Anki deliberately tried to support, and I don't think we document them.

RumovZ:
From what I've seen, I don't believe fixing these issues will make the parser more complicated. The changes needed may seem rather extensive, but they are mostly restricted to parsing text (quoted, unquoted, partially quoted) and unescaping it, so it comes down to switching some nom parsers, modifying their parameters and RegExes. The overall logic shouldn't need to be touched.

As for the returns, though most (but not all) of the mentioned issues are edge cases and can be worked around, having a consistent set of parsing rules should pay of big time in the long run. It will not only help future maintaining and extending but also allow documenting details which are currently too intricate for mentioning in the manual (like when to escape `\`).

You may remember that I wrote my own search keyword (`math`). It was a pain figuring out which characters could be passed to the math parser and whether they would have to be escaped because it was different for the three possible quotation styles. That shouldn't be necessary.

I've already given some thought to what would be necessary to harmonise the parser, but there are a few decisions to be made. That's why I wanted to create a topic in the forum and get your and the community's feedback.

PR #811, opened 2020-11-04
Add hooks for filtering media.

ianki:


ianki:
This change is to support the Media Sync Filter.

Let me know if these new hooks look reasonable, and if there are any other places that media paths should be filtered.  Thanks

dae:
Sorry, this will need rebasing to merge into master.

Paths are relative to the media folder already, so you should not need a full path - eg in mediasrv:

```
    if file_is_excluded(path):
        path = f"hidden/{path}"

    return aqt.mw.col.media.dir(), path
```

Please avoid adding the current working directory to the audio file - it can break things when paths contain non-Latin text on Windows for example.

qt/aqt/gui_hooks.py
```diff
@@ -11,3 +11,5 @@
 
 # You can find the definitions in ../tools/genhooks_gui.py
 from aqt.hooks_gen import *
+
+
```
dae, line 15: These could be removed; otherwise looks good thanks. Rebasing onto master again should fix the Windows tests not working

ianki: Done.

ianki: I'm still getting a failure on Windows it seems.

dae:
Cheers, will merge this once I sort out the Windows CI issue (which is an issue with the setup, not a problem with your changes)

PR #813, opened 2020-11-08
Fix clicking filters with keyboard modifiers

RumovZ:
Fix two minor bugs where 
1. Clicking "Whole collection" filter with modifiers leads to invalid or unintended search.
2. Shift clicking a filter doesn't remove the default prompt.

While we're at it, simplify regex, don't shadow built-in name chr and use new format syntax.


dae:
Thanks!

PR #814, opened 2020-11-08
Improve compatibility with graphics pad/touchscreen

k12ish:
When using Anki with a graphics pad, text is often selected on accident. This one line snippet of CSS will fix that issue.

I have not tested this code, but the other two instances of `-webkit-user-select: none;` in Anki prevent text from being selected and work as required.


k12ish:
To be honest, I didn't think I needed to add myself to the CONTRIBUTING file for something this trivial. Seeing as this seems to be a requirement, I'll get that sorted.

dae:
Thanks!

PR #816, opened 2020-11-11
Update only enabled addons when available

cecini:
Some bugging  addon even disabled, once anki desktop start, still need be update.

Workaround a bug when ankidesktop start, the update dialog and syncing dialog all show, and the syncing is still working on for some reason, never quit, then  we cannot to close any window.

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -1249,7 +1249,7 @@ def handle_update_info(
 
     updated_ids = mgr.updates_required(update_info)
 
-    if not updated_ids:
+    if not [id for id in updated_ids if mgr.addon_meta(str(updated_ids)).enabled]:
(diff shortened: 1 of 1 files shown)


dae:
I don't believe that is related to add-ons? There was a known issue with "Anki Updated" appearing and making the sync get stuck, but that should already have been fixed in recent Anki versions.

cecini:
Oh, presume your idea is correct. 
Now, do you think  the feature "update only enabled addons" is reasonable？ Even not related the sync stuck bug.
If no, we can close the pr.

dae:
If a user decides they don't want to receive updates for an add-on, they can delete it - disabling is intended to be a temporary action, so I think it makes sense to keep disabled add-ons up to date.

PR #817, opened 2020-11-11
Update mypy and black format check to python version 3.8

cecini:


dae:
Thanks! The files in qt/ will need updating too; I will fix this.

PR #818, opened 2020-11-11
Replace pip_import with pip_install

cecini:


dae:
What's your goal with this change? rules_python fetches all the dependencies in advance, whereas the current solution fetches them on demand.

cecini:
Not been aware the on demand fetch. 
Sorry.

PR #819, opened 2020-11-11
Specify full path to bazelisk and output root on Windows

abdnh:
Building fails if Anki's source directory is not in the same drive as bazelisk.

### DIFF
--- a/bazel.bat
+++ b/bazel.bat
@@ -1,2 +1,2 @@
 @set PATH=c:\msys64\usr\bin;c:\python;%PATH%
-\bazel\bazel  --output_user_root=\bazel\anki %*
+c:\bazel\bazel  --output_user_root=c:\bazel\anki %*


dae:
This forces the user to store the gigabytes of data files on their C drive - perhaps a better approach would be to tell users to put \bazel on the same drive as they plan to store Anki's source?

abdnh:
Shouldn't Bazelisk be in c:\bazel as documented in [docs/windows.md](https://github.com/ankitects/anki/blob/master/docs/windows.md) though? It seems this is not actually required, but I can't verify that now since the javascript errors got more persistent for some reason and I can no longer build Anki whether the source and Bazelisk are on the C drive or not :(

If it was actually required, we can change Bazelisk's path but leave output_user_root alone.

dae:
That's my point - if we update the docs to tell people to use \bazel on the drive with the source code instead of c:\bazel, then further adjustments should not be necessary.

If the compile has broken, try running bazel clean --expunge, then removing ts/node_modules, and building again.

dae:
All up and running now?

abdnh:
Yes. Thanks!

PR #821, opened 2020-11-13
fix

AbdelrahmanHamdy1996:


dae:
I appreciate you taking the time to contribute, but my preference is not to include auto-generated Sphinx docs in the source tree. Users can install the wheels and then run 'pydoc -p 8080' to achieve similar results.

PR #822, opened 2020-11-14
Fix handling of no-break spaces when showing duplicates

abdnh:
no-break spaces are replaced with normal spaces in showDupes, which causes browser search to fail to find duplicate notes containing no-break spaces.

I've imported lots of duplicate notes recently and started to hit this issue every day and I'm a bit surprised this has not been reported until now considering that notes usually have a lot of non-breaking spaces since Anki inserts them automatically.

By the way, I noticed that non-breaking spaces (the character `\xa0`, not `&nbsp;`) are also stored as normal spaces in the database, but I can't find in which file this is done. I thought this is due to unicode normalization but Anki does NFC which preserves non-breaking spaces (while NFKC actually converts them). Out of curiosity, where is this done?

### DIFF
--- a/pylib/anki/utils.py
+++ b/pylib/anki/utils.py
@@ -61,19 +61,19 @@ reEnts = re.compile(r"&#?\w+;")
 reMedia = re.compile("(?i)<img[^>]+src=[\"']?([^\"'>]+)[\"']?[^>]*>")
 
 
-def stripHTML(s: str) -> str:
+def stripHTML(s: str, nbsp_to_sp: bool = True) -> str:
     s = reComment.sub("", s)
     s = reStyle.sub("", s)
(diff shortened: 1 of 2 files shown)


dae:
I can't think of a place in the code that converts `\xa0` prior to writing the note.

It looks like strip_html_preserving_media_filenames() in the Rust code doesn't match the current Python implementation - it is not doing any entity decoding at all. And since it's called in write_dupes() on the provided string, it makes me wonder whether we need to be doing this at all - perhaps the Python code should just be passing in the field as-is?

abdnh:
That makes sense.
strip_html_preserving_media_filenames and stripHTMLMedia also use quite different regular expressions, which can cause more problems.

PR #823, opened 2020-11-15
Rework search parser

RumovZ:
Ready for a PR but not for a merge. Some questions:

1. I wonder if it is was best to have the parser only check for invalid escape sequences and let the sqlwriter call all the unescape functions. The main benefit would be better intelligibility because, as you have pointed out, it is often incomprehensible why the parser unescapes in a certain way. Also, it then might be possible to have a wrapper function for the unescape functions which also handles normalising, so `norm_note` doesn't have to called for every single SearchNode variant.
2. Regarding 57787368a1f0421885b96ca9fc952e1e4b5bcfbd, I did this because yet another handling of escape sequences would have to be implemented (Regex if wildcards, Sql otherwise). Does this have any side effects apart from maybe a minimal performance penalty? If I realise `1.`, restoring the old handling (using registered tags from the collection) would probably not be much of a problem.


dae:
I'm guessing it would be easiest to reason about if the parser handles any escapes in the search syntax itself, and leaves the handling of escapes in text fields to sqlwriter.

I seem to recall regex searches being approximately twice as slow in some earlier testing (not specifically of tags), so it may be worth doing a simple benchmark on a large collection before changing over to a regex, as searching by tag is a common use case.

RumovZ:
> if the parser handles any escapes in the search syntax itself

I don't quite understand. Which escapes do you mean? As far as I'm aware, escapes will only be used for user named structures (decks, notetypes, templates, fields, tags) or field searches (with or without field name prefix).

I did a quick benchmark with 1,000,000 notes, 1 card and 10 tags each: Searching for an existing tag was only 25% faster with the old method (~10s to ~7.5s). But of course, searching for a non-existing tags is considerably faster (~4s to ~1s). I didn't really have that case on my radar.

dae:
I mean for characters like ":() that are part of the parsing process itself, rather than part of the text passed to particular search terms.

Searches for non-existent tags aren't worth worrying about - users mainly click on the tags in the sidebar.

RumovZ:
Ah, now I understand. You mean compared to escaped wildcards which are only relevant to the writer.
That's a good point!
At first, I was deliberately avoiding unescaping multiple times. But I should have questioned this decision when the distinctions of cases started piling up.

RumovZ:
I've implemented your suggestion. Looks good!
Here are some more questions/todos:

- [ ] Testing for the correct behaviour when searching for tags with whitespace should be implemented in the Python part.
- [ ] Deck names are the only Unicode user input that gets normalized regardless of `col.normalize_note_text()`. Is that intentional?
- [ ] I'd like to retire `text_to_re` from `text.rs` in favor of the new `to_re`. It's only used once in `tags.rs`. Should I make that change in this PR or in a spearate one? Of course, this will only make sense once this PR is merged.
- [ ] What is the reasoning behind using `ParseResult` instead of `IResult` in the low-level parser functions? As I understand it, this has two downsides:
    1. These functions can't return a `Failure` (in nom nomenclature), so even if it is clear that parsing won't succeed, the higher-level functions will still iterate over all the `alt` branches.
    1. The cause of the error can't be passed through to the user.

    I feel that the latter would be a real boon to Anki's user friendliness, especially when new syntax rules are introduced as with this PR (even if it is only to a very small degree).

dae:
>  Deck names are the only Unicode user input that gets normalized regardless of col.normalize_note_text(). Is that intentional?

Do you mean in the search code? I think there are a bunch of norm() calls in sqlwriter.rs in master? If you mean when content is added, there are calls like ensure_string_in_nfc() in notes.rs, but it's possible something's been overlooked and needs to be updated.

>  I'd like to retire text_to_re from text.rs in favor of the new to_re. It's only used once in tags.rs. Should I make that change in this PR or in a spearate one? Of course, this will only make sense once this PR is merged.

Probably best to get this done first.

>  What is the reasoning behind using ParseResult instead of IResult in the low-level parser functions? 

I think that's just what was easiest at the time - no objections to a change to IResult if it doesn't make things unwieldy.

dae:
Did you merge with the GitHub UI? That may have resulted in a different email address to the one you usually use.

RumovZ:
> Do you mean in the search code? I think there are a bunch of norm() calls in sqlwriter.rs in master?

I don't know how I've overlooked that, sorry for the confusion. It makes sense then: Names (decks, templates etc.) get always normalised but for field content, it's config dependent.

> I think that's just what was easiest at the time - no objections to a change to IResult if it doesn't make things unwieldy.

I'll look into it, but do you reckon conveying a more detailed error message to the user is feasible in general? I don't know anything about the Protobuf stuff.
If this is not something that is realisable in the near future, it's probably not worth the effort. (Though it could also make the parsing code more straightforward.)

> Did you merge with the GitHub UI? That may have resulted in a different email address to the one you usually use.

Embarrassingly, I just added the wrong e-mail alias to my GitHub account...

dae:
I don't recall the raw parsing errors being particular intelligible back when I was working on the code, but providing the remaining string would at least help the user know the approximate location where things are failing. Perhaps something that doesn't need to be in the initial changes though?

Errors get mapped into the AnkiError enum in err.rs. When they're passed back by the backend, anki_error_to_proto_error() converts them into the relevant protobuf message - currently AnkiError::SearchError is mapped to InvalidInput, but this can be extended in backend.proto if necessary.

I'm afraid I'm not sure I see much need for an additional literal syntax. While it doesn't require a great deal of code at the moment, it still needs documenting and maintaining, and is currently inconsistent with the handling of globs in say deck searches. It seems easier to just teach the user "use \* and \_ when you want those characters" than to provide a separate search syntax for such cases.

RumovZ:
> I don't recall the raw parsing errors being particular intelligible back when I was working on the code, but providing the remaining string would at least help the user know the approximate location where things are failing.

I didn't have returning the raw nom erros in mind but rather a custom enum whose variants translate to messages like “Unmatched `"` at position xy!”, “Invalid escape sequence `\x` at position xy!”, “`is:dew` at position xy is not a valid state!” etc.
Would be great if this funcionality was available with the new parser. But I see that this is probably too big of change to be just implemented with this PR along the way .
Thanks for elaborating on the protobuf part.

> I'm afraid I'm not sure I see much need for an additional literal syntax.

I was thinking about decks with heavy use of special characters where escaping could become tedious. E.g., there is a deck for Morse code using _ for the long signal.
But I can easily revert this commit. Just so I understand, what do you mean by

> and is currently inconsistent with the handling of globs in say deck searches.

dae:
For example, users can search for 'deck:parent::*', but they'd still need to use a backslash there, as this literal syntax only works in the field/note searching case. Making it consistent would require supporting the prefix everywhere a wildcard is supported, which makes the change more unwieldy.

RumovZ:
I wouldn't call that incosistent.
Just as `re:`, it's an aid to handle the arbitrarily long and complex content of fields. Names (of decks and so forth) on the other hand, have a limited length and character set and there's only a manageable number of them in every collection. So simple wildcards and single character escapes should be enough to deal with them.

Apart from that, do you want me to remove/add anything else? I don't have anything left on my list.

dae:
I gave it a quick lookover but didn't spot any problems - thanks for all your work on this. Let's see how it goes in the next beta!

PR #824, opened 2020-11-15
Improvements to MathJax and ts hooks

hgiesel:
1. I added the default MathJax packages for tex-chtml, which include:
    1. require: using \require{package-name} to load a package
    1. autoload: using a command from a different package automatically loads it
    1. configmacros: allows for definition of predefined macros

This would allow users to use dynamically load Mathjax extensions. For example, [in this case, where a user wanted "boldsymbol"](https://forums.ankiweb.net/t/bold-math-does-not-work-in-mathjax-on-the-linux-desktop-client/5073), he could now simply use `\require{boldsymbol}`
However there's a big catch here: [MathJax 3 internally relies on the "onload" event of a script tag being fired](https://github.com/mathjax/MathJax-src/blob/fbd71f3f7ca06dc71304e8a2c03f13be233ccac5/ts/components/package.ts#L326), which does not work in the AnkiWebView (possibly, again, because it is a data URL), and which atm makes the rendering simply stop within Anki. To actually take advantage of this feature, the user has to handily call `MathJax.typesetPromise()` again.

2. I changed the way ts hooks work: Rather than just executing a function returning `void`, the function can now optionally return a Promise, which is then waited upon using [`Promise.all`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all). This way, add-on developers can register asynchronous actions on ts hooks.

I also added a `typesetClear` call, which is recommended to be executed before you start a new typeset.

### DIFF
--- a/qt/aqt/data/web/js/mathjax.js
+++ b/qt/aqt/data/web/js/mathjax.js
@@ -3,17 +3,19 @@ window.MathJax = {
     displayMath: [["\\[", "\\]"]],
     processRefs: false,
     processEnvironments: false,
-    packages: [
-      'base',
-      'ams',
-      'noerrors',
-      'noundefined',
-      'mhchem',
-      'require',
-    ]
+    packages: {
+      '[+]': [
+        'noerrors',
+        'mhchem',
+      ],
+    }
   },
   startup: {
-    typeset: false
+    typeset: false,
+    pageReady: () => {
+      console.log('page is ready');
+      return MathJax.startup.defaultPageReady();
+    },
   },
   options: {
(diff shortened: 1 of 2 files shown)


dae:
Thanks Henrik!

dae:
It looks like https://github.com/ankitects/anki/pull/824/commits/a7b5a7efe1b4b1aa01a2a7cc0d39f54010074bcd has broken scrolling to the answer. If you add more than a screen's worth of text/images to a card's question, 'show answer' no longer scrolls down. Any ideas?

PR #825, opened 2020-11-16
Switch from jQuery pre 3.0.0 pseudo promises to ES6 promises for ts hooks

hgiesel:
I slipped up a little bit. While the previous PR was not broken, it didn't achieve what I intended it to.

Turns out that [pre-3.0.0 jQuery had a different kind of promise/deferred](https://jquery.com/upgrade-guide/3.0/#breaking-change-and-feature-jquery-deferred-is-now-promises-a-compatible), which didn't behave like the ES6 promises I'm used to, especially when it comes to returning promises in the callback function.

More specifically, the `.then(_runHook(onUpdateHook/onShownHook))` lines did not wait on the promises returned, but rather continued immediately with the next `.then(f)`

### DIFF
--- a/qt/aqt/data/web/js/reviewer.ts
+++ b/qt/aqt/data/web/js/reviewer.ts
@@ -39,8 +39,7 @@ function _updateQA(html, fadeTime, onupdate, onshown) {
     var qa = $("#qa");
 
     // fade out current text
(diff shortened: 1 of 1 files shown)


dae:
Thanks!

PR #827, opened 2020-11-17
Fix TypeError in Card.timeTaken

abdnh:
This works around an issue reported a couple of days ago in [this forum post](https://forums.ankiweb.net/t/mw-col-sched-answercard-add-on-coding-problem/4999).

### DIFF
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -181,7 +181,10 @@ class Card:
 
     def timeTaken(self) -> int:
         "Time taken to answer card, in integer MS."
-        total = int((time.time() - self.timerStarted) * 1000)
+        if self.timerStarted is not None:
+            total = int((time.time() - self.timerStarted) * 1000)
+        else:
+            total = 0
         return min(total, self.timeLimit())
 
     # legacy


dae:
This routine assumes it will be called during review. I'm not sure changing this is ideal, as it will result in an unusually short review time (0s) being logged, pulling down the averages.

The rescheduling code was recently updated to use an ease of '0' to denote a manual reschedule, making it possible for the graphs code to ignore the answer times for such reschedules. The issue is that it won't automatically pick up the user's configured good interval, and is only currently usable for 1+ day delays. I'm hoping to make the scheduling code easier to use in a piecemeal fasion in the future, so such things are easier to implement.

abdnh:
> This routine assumes it will be called during review. I'm not sure changing this is ideal, as it will result in an unusually short review time (0s) being logged, pulling down the averages.

I overlooked this point.

Actually, I now agree this is not a good change; the exception can serve as a reminder for the caller to provide a reasonable value for timerStarted as p4nix did afterwards in his add-on instead of us "helpfully" taking account of that.

PR #830, opened 2020-11-18
Strip isolation characters from more strings

abdnh:
By the way, does the string `QT_MISC_DEFAULT` used a couple of times in [deckchooser.py](https://github.com/ankitects/anki/blob/98a4a1927a8e781a1c0b4a297caa8e86134ac027/qt/aqt/deckchooser.py#L50) need stripping?
It seems the string `DeckConfigDefaultName` is used [here](https://github.com/ankitects/anki/blob/ccfd6244de9802a10ddf11687a8c86d0c1309cb4/rslib/src/storage/deck/mod.rs#L269) to create the default deck without any stripping so I'm not sure.

Update: I forgot that argumentless strings don't have isolation characters so we don't need to strip them.

### DIFF
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -9,6 +9,7 @@ from typing import Any, Dict, List, Optional
 import aqt
 from anki.cards import Card
 from anki.consts import *
+from anki.lang import without_unicode_isolation
 from anki.notes import Note
 from anki.rsbackend import TemplateError
 from anki.template import TemplateRenderContext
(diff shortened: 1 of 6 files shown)


dae:
Thanks Abdo! Yeah, it's only a concern if we're replacing variables, and either printing it in a place that the font may struggle, or embedding it into the DB. I thought about just stripping all strings all the time, but that may lead to issues for RTL text that the characters are trying to prevent

PR #831, opened 2020-11-18
Fix the subdeck add with missing parent deck sync issue

cecini:
Once add subdeck with missing parent deck, such as  add deckname  "deck::subdeck" in the anki desktop. Previously the created parent deck have usn = 0, so never find it need to be synced to server. 
So just need add usn = -1 when create ,additionally normalize the deck name

### DIFF
--- a/rslib/src/decks/mod.rs
+++ b/rslib/src/decks/mod.rs
@@ -290,9 +290,12 @@ impl Collection {
 
     /// Add a single, normal deck with the provided name for a child deck.
     /// Caller must have done necessarily validation on name.
-    fn add_parent_deck(&self, machine_name: &str) -> Result<()> {
+    fn add_parent_deck(&mut self, machine_name: &str) -> Result<()> {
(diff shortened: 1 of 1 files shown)


dae:
Thanks cecini! Your basic approach was good, but we should pass in the usn from the parent code, and prepare_deck_for_update() calls ensure_deck_name_unique() which is not useful on that path.

cecini:
Oh, we should normalize the deck name ,but  not to call ensure_deck_name_unique(). What do you think about this? Some deck format maybe is "nonnormalizedeckname::subdeck".

dae:
You are right, update_renamed_deck() does not look correct, as it is creating the parents before normalizing in add_or_update_single_deck(). It's also calling ensure_deck_name_unique() which is not necessary, so this routine may need some rethinking.

dae:
I think ensure_deck_name_unique() is necessary, but I've made some other tweaks - please let me know if you find any further issues.

PR #832, opened 2020-11-19
Fix description of exporters

abdnh:
Description broke yet again with the full migration to Fluent.

drawing


dae:
Thanks for catching this, I should have tested it.

PR #833, opened 2020-11-20
Initialize translation system to default language on first run

abdnh:
The setlang screen is trying to use the translation system before we initialize it, causing an AttributeError.

There are some useful but untranslated error messages in `__init__.py` and `profiles.py`. Could not we initialize the translation system to the system language early on in _run (instead of setDefaultLang) and make those strings translatable? I can look into this if there are no caveats I'm not aware of.


dae:
Thanks!

PR #835, opened 2020-11-21
Set up default language before loading profile

abdnh:
We can make more strings translatable this way.


dae:
Thanks Abdo. The original approach was based on the fact that some Anki users in Japan struggled to understand startup error messages appearing in Japanese if they were not a native speaker, but it's easier to change the system language these days.

PR #836, opened 2020-11-25
Fix UnboundLocalError in profiler code

abdnh:
I noticed that the profiling code is broken while working on my previous PR.


dae:
Thanks as always abdo :-)

PR #837, opened 2020-11-27
Revert implementation of `r:` prefix

RumovZ:


dae:
Sorry for the delayed reply to your docs PR, it's been a busy few days. The door is not shut on introducing this functionality in the future, but it might be worth waiting for a bit more demand to present itself first, as it's not something I recall coming up frequently.

PR #838, opened 2020-11-28
Fix write_dupes checking sort field instead of first field

abdnh:
This fixes the issue reported here: https://forums.ankiweb.net/t/show-duplicates-not-working/2883/16


dae:
Hi Abdo,

Your solution was perfectly fine, but I went in a slightly different direction with this because sqlite functions become an API surface, and we then have to worry about not changing them when some add-on starts relying on them. There was an existing function that only needed minor modifications to grab the data we needed, so I've used that instead.

abdnh:
Yeah, I've anticipated a problem with add-ons but thought no one would use the function anyway.

PR #839, opened 2020-12-01
Fix fields_did_rename_field gui hook

hgiesel:
Current implementation didn't act like I specified. It should be newly renamed field, and the old name as a string, however you just had the new field, and the new name.


dae:
No problem - thanks for the fix!

PR #845, opened 2020-12-09
Switch back to fadeTo rather than fadeIn/fadeOut

hgiesel:
This is the fix to https://github.com/ankitects/anki/pull/824#issuecomment-741725158.

It's a weird one: It seems `$.fadeTo` and `$.fadeIn` / `$.fadeOut` work differently in this aspect. I'm sorry for (unnecessarily) breaking it.

It seems like, `$.fadeIn` / `$.fadeOut` first alter the opacity, just like `$.fadeTo`, but at the end, they set `display: none/block`. With `display: none`, the text won't occupy any space, which is why scrolling to it doesn't move the screen.


dae:
No problem, thank you for the quick fix!

PR #846, opened 2020-12-14
Fix duplicate check not decoding entities

abdnh:
See https://forums.ankiweb.net/t/bug-duplicates-not-detecting-on-paste/5753


dae:
Thanks Abdo :-)

PR #848, opened 2020-12-15
Update rslib dep nom

cecini:
-----

Maybe we should not commit the change introduce by the python update.py to the repo?   
For ci， we first run the python update.py.
For user,  maybe we can add wheel pkg,
For developer, maybe use the rust nightly version, or another version, or some another rust dep, so no need the change from the upstream, just run the pyhont update.py.


dae:
Thanks!

Running update.py at build time would mean users need to have cargo-raze installed, which makes it more complicated for people to run from source. Currently update.py also updates the lock file to the latest versions, and we don't want to do that at build time - it's important to have specific versions specified in the repo for reproducible builds, and to avoid the occasional breakages that can happen when a minor version changes.

PR #850, opened 2020-12-15
fix float to int conversion DeprecationWarning

cecini:


dae:
Where is the warning coming from?

cecini:
Use Python with debug to run anki.
Maybe use python3 -W default also get this warning.

dae:
Thanks, I've added this to the ./run command so we catch such errors in the future.

cecini:
Nice that you ignore the pickle.dumps warning which need fix the pyqt.

PR #851, opened 2020-12-16
Remove unicode isolation from card type names

hgiesel:
Right now, Anki saves unicode isolate characters in card type names.

This breaks browser search `card:"Card 2"`, will not match the card type name `Card ⁨2⁩`, because of invisible characters.


dae:
Thank!

PR #852, opened 2020-12-16
Resolved some mypy errors in qt

lifanchu:
- Added two missing type imports in `browser.py`
- Added `mypy_path` to `qt/mypy.ini`. This resolves the "Cannot find implementation or library stub" import errors. Currently requires user to be in the same directory as `mypy.ini`, but this will be fixed in a future version of mypy (see [here](https://github.com/python/mypy/pull/9414)).


dae:
Hi Lifan,

Thank you for contributing!

How are you running mypy? Rather than running it separately, if you use 'bazel test ...' from the project root, it should run mypy with the correct environment set up, and a change to qt/mypy.ini should not be required.

lifanchu:
Okay, my mistake! Thanks 👍

PR #853, opened 2020-12-18
Strip isolation characters from app link

abdnh:
Unicode isolation characters causing problems yet again.
Reported in https://forums.ankiweb.net/t/bug-link-in-about-window/5894


dae:
Thanks!

PR #854, opened 2020-12-19
Add method `card_count_from_did()`

k12ish:
Finding the number of cards in a given deck is a common operation, add-on developers shouldn't require database queries for this.

The code for `card_count_from_did` is taken from the `_delete` method from deckbrowser.py and seems to work desirably.


dae:
Thanks! I'll push a slight change to the naming after this, but otherwise looks good.

dae:
Update here: https://github.com/ankitects/anki/commit/e99a7c0f90cebc4b40215501fed5627d0350bed1

PR #855, opened 2020-12-19
WinRT TTS API support

rye761:
Hello! This is WinRT TTS API support as I have mentioned in help wanted number 16. This is almost ready to go, but I could use some guidance on a couple of things.

1. How do I add the necessary library as a dependency? I tried the README.md in dependencies but found myself a bit confused. This feature depends on winrt==1.0.20239.1.
2. How to ensure graceful failure on previous windows platforms? How to test that?
3. Is there a cleaner way to overcome the need for a separate thread in the get_available_voices function?
4. How to deal with voices that are offered by both APIs? I get a couple of duplicate voices for this reason on my machine.

### DIFF
--- a/qt/aqt/sound.py
+++ b/qt/aqt/sound.py
@@ -916,9 +916,10 @@ def setup_audio(taskman: TaskManager, base_folder: str) -> None:
 
         av_player.players.append(MacTTSPlayer(taskman))
     elif isWin:
-        from aqt.tts import WindowsTTSPlayer
+        from aqt.tts import WindowsTTSPlayer, WindowsRTTTSFilePlayer
 
         av_player.players.append(WindowsTTSPlayer(taskman))
+        av_player.players.append(WindowsRTTTSFilePlayer(taskman))
 
     # cleanup at shutdown
     atexit.register(av_player.shutdown)
--- a/qt/aqt/sound.py
+++ b/qt/aqt/sound.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import atexit
 import os
+import platform
 import re
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


qt/aqt/sound.py
```diff
@@ -711,9 +711,10 @@ def setup_audio(taskman: TaskManager, base_folder: str) -> None:
 
         av_player.players.append(MacTTSPlayer(taskman))
     elif isWin:
-        from aqt.tts import WindowsTTSPlayer
+        from aqt.tts import WindowsTTSPlayer, WindowsRTTTSFilePlayer
```
dae, line 714: Please add a check for platform.release() == "10" - if it's not 10, we should avoid importing and adding the new code, which should avoid breaking older clients.

qt/aqt/tts.py
```diff
@@ -516,3 +518,63 @@ def _rate_for_speed(self, speed: float) -> int:
...
+        id: Any
+
+    class WindowsRTTTSFilePlayer(TTSProcessPlayer):
+        voice_list = None
+        tmppath = os.path.join(tmpdir(), "tts.wav")
+
+        def import_voices(self) -> None:
+            import winrt.windows.media.speechsynthesis as speechsynthesis
+            self.voice_list = speechsynthesis.SpeechSynthesizer.get_all_voices()
+        def get_available_voices(self) -> List[TTSVoice]:
```
dae, line 533: Spaces between method names would be good - you can use 'bazel run qt:format' to automatically format the code.

dae (commented):
Thanks for your continued work on this!

- The automatic checks are failing because you branched off an older commit - if you 'git add upstream https://github.com/ankitects/anki.git; git fetch upstream master; git rebase upstream/master', it should bring you up to date. Please run the tests locally with Bazel first if you can: https://github.com/ankitects/anki/blob/master/docs/development.md
- I've added the winrt requirement in latest master, so you don't need to worry about that
- get_available_voices() should probably in hindsight have been implemented as a function with a callback. Changing it will break the existing implementations, so lets defer that for now
- Since RT is presumably the blessed API moving forward, I guess the best option is to change the default_rank of the old TTS implementation to -1, so that the RT voices take precedence.

dae:
By the way, I pinned winrt to the version you specified, but that's not the latest. Does the latest not work?

rye761:
Thanks for the comments! I'll get those changes made. I had issues with the latest version complaining of a missing DLL when used with Python 3.8. The latest works on 3.9, so when anki moves to 3.9 then the package can be upgraded.

dae:
Thanks Ryan, let's give this a go!

PR #856, opened 2020-12-20
Update jQuery from 1.12.4 to 3.5.1, include in Bazel build process

hgiesel:
I noticed, that the different jQuery distributions are on quite different versions:

* Anki is on 1.12.4
* AnkiWeb is on 3.5.1
* AnkiMobile is on 2.2.2

This PR updates Anki jQuery to 3.5.1 (newest, non-slim version, just like AnkiWeb), and also includes moving from comitting jQuery to the repo, to making it part of the Bazel build process, similar to MathJax.

I made clicked through all the menus, and everything seems to behave well.
jQuery UI supports 3.5.1 as well.

Note: I moved the `_copy_files` function from mathjax.bzl to its own file copy.bzl. I intend to make a follow-up PR, which moves other JS parts (css-browser-selector, and protobuf) to use this function as well.


dae:
Thanks Henrik, this looks good, and it makes sense to move the others like css-browser-selector to the same setup. A couple of points:

- For single-file libraries, perhaps we should put them in js/vendor's BUILD file rather than in separate subfolders for each of them?
- The formatting changes in mathjax.js look unrelated? The current formatting is the default prettier formatting from VS Code.
- redirectWebExports() needs to be updated to automatically map the old vendor .js paths to the new ones, so we don't break existing add-ons

hgiesel:
1. I kind of like the uniform treatment of all libraries. If there was ever a case, where the library was a single file, but suddenly we want to include a second file, like a `file.map.js` file, it would only require a single line change. But I also agree that `vendor/jquery/jquery.min.js` look very redundant if there's only one file in that directory. I'll leave it up to you.
2. It seems that reformat was done by running `//ts:format`, which is based on `ts/.prettierrc`, which specifies `"tabWidth": 4`. Does this mean you want to have different formatting rules for JavaScript vs TypeScript? But I agree it doesn't belong in this PR.
3. I will implement this once we decided on 1.

dae:
1. I slightly lean towards vendor/ - even if we need to add an extra file in the future, that doesn't necessarily mean we need to split it off into a separate folder at that point. But valid arguments can be made for both approaches, so if you feel strongly that separate folders would be better I won't object.
2. Ah, this looks like a bug in the rules - ts:format should not be making changes if format_check is not reporting any problems. Will look into it.

dae:
Is this still in progress? Currently we have jquery in a subfolder, and jquery-ui in vendor. The formatting issue should be fixed.

hgiesel:
1. I changed jquery to only insert the file, rather than create a new directory.
2. I already removed the formatting of the MathJax file.
3. I added the remap for the redirectWebExport method

BTW. looking at `redirectWebExports`, [this line](https://github.com/ankitects/anki/blob/aa816177d0daaaf86a91a62763105510b93459f5/qt/aqt/mediasrv.py#L189) seems awfully wrong to me... shouldn't it be mapped to `js/vendor/`?

dae:
Thanks, you're right - I've pushed a fix. Thanks for the work on this PR as well :-)

PR #858, opened 2020-12-21
Remove placeholder `<br>`s in empty fields in editor

hgiesel:
This replaces the insertion of `<br>` elements in the editor view, and replaces it with a neat small CSS specifier.

I left the filtering of `<br>` and `<div><br></div>` sequences in editor.py, as otherwise users might end up with a lot of those text sequences in their notes.

One way users can notice this change is when they select the whole content of a field (Ctrl-A), delete it, and undo it using `Ctrl-Z`, it won't "insert" a extra newline on the start.

### DIFF
--- a/qt/aqt/data/web/css/editor.scss
+++ b/qt/aqt/data/web/css/editor.scss
@@ -8,9 +8,14 @@
     padding: 5px;
     overflow-wrap: break-word;
     overflow: auto;
+
+    &:empty::after {
+        content: "\A";
+        white-space: pre;
+    }
 }
 
-.clearfix:after {
+.clearfix::after {
     content: "";
     display: table;
     clear: both;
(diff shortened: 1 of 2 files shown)


dae:
Thanks Henrik. I've tried to drop the BR in the past but ran into problems, so it'll be great if this solves it. I'd like to get .38 out soon, so might hold off on this until the start of the next beta round just in case.

hgiesel:
Understandable, bug fixes have priority.

hgiesel:
I assume the issues in the last attempt were mostly caused by removing `mungeHTML`. I left that functionality in tact, so `<br>` and `<div><br></div>` are still filtered out.

dae:
There was also an issue where the fields were slightly different height when empty IIRC, which your solution doesn't seem to suffer from.

PR #859, opened 2020-12-21
Consider removing special handling of Shift+Tab in Editor on macOS

hgiesel:
I was confused by some weird Tab behavior, when I noticed this piece of code.

I'm running BigSur, and I can use the `Shift+Tab` shortcut without this piece of code. In fact, it runs better without, because it respects HTML semantics, like `tab-index`.

According to 53fb3b0e8c86131914eb7e4679b59c4347788c26, it seemed to be necessary at some point, but I cannot confirm this necessity anymore.

### DIFF
--- a/qt/aqt/data/web/js/editor.ts
+++ b/qt/aqt/data/web/js/editor.ts
@@ -54,12 +54,6 @@ function onKey(evt: KeyboardEvent) {
         currentField.blur();
         return;
     }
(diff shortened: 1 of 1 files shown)


dae:
Tested on Qt 5.14 and an older macOS and it seems to work, so assuming the -alternate build doesn't come back, it'll be nice to have a bit less code to maintain. Thanks!

PR #860, opened 2020-12-22
Normalise search

RumovZ:
When implementing what we've discussed in #807, I realised that those are actually three quite independent features:

1. Normalising the user's search input: This can best be done on the backend, which can use the strongly typed AST from the parser.
2. Provide a parsed AST for add-ons: This must be done via JSON. A bit messy for the add-on developers, but I think in Anki we can simply use the default serde derive macros and expose the according functions.
3. Search string transformations: If we parse and transform on the backend, we can pass back the new search string with the writer function this PR introduces.

------
This PR implements 1. I can add the implementation for 3 here or open another PR.
I still need to add tests but before I do, I'd like to ask for your opinion regarding the search syntax. Here are the deliberate choices I've made trying to guess Anki's preferred style:

1. The `and` operator is represented by a space.
2. The `or` operator is lowercase.
3. Quoted terms are separated by a space (`a "b"`, not `a"b"`).
4. Groups are separated from their surroundings but there is no space between the parentheses and the content (`a (b c)`.
1. Parentheses are escaped by quoting.
2. A preceding `-` is escaped by `\`.
3. Keyword searches are quoted partially (if necessary).
4. Field searches are quoted entirely (if necessary).

Do you disagree with anything or is there something I've missed?


dae:
Hi Rumo,

I had originally envisaged this functionality only being used when applying transforms, rather than for also canonicalizing the user's input. Doing the latter could be useful, but if we're going to go that route, perhaps it would make sense to always favour quoting and being more explicit? eg If a user types in:

ten bottles

then they get

"ten" "bottles"

which should hopefully make it apparent that they've actually searched for two different terms.

We could extend it to making AND explicit as well, eg 'ten bottles' → '"ten" AND "bottles"'.

I personally prefer to use the lowercase equivalent because it's easier to type, but for making it clearer to readers, the 
uppercase version is probably a better choice.

For things like is:due, quoting the entire term rather than the bit after the colon might be better?

What do you think?

Also a minor nit - I try to be more verbose/clear with the backend method names, so normalize_search() would be preferable for the backend, and probably the frontend too.

RumovZ:
Thanks for the feedback!
I was trying to hew to the typical input style which users should be familiar with. But I also like your suggestion since it is more explicit. In that case, I agree that we should only quote entirely, `is:"due"` doesn't look quite right.
I'm not good with naming things (do you think the name `writer` is fine for the module?), but I chose `norm` deliberately to avoid the s/z struggle. 😁 I'll revise the names and add some doc strings.

dae:
I'm not great at naming either; the only alternative that springs to mind is "serialize(r)". I've resigned myself to using Americanisms in the code for consistency, so would favour "normalize" even though that's not my native dialect.

dae:
Look good, just needs a reformat.

RumovZ:
Yeah, I use to forget that.
At the moment I'm still working on implementing the functions for replacing, concatenating and negating searches. I thought it would be a good idea to replace the existing frontend code which only adds text with more robust backend functions as well while I'm at it. 
I'll push these changes in the new year and let you know when they're ready to be merged from my side.

dae:
Righto - I'll mark this as a draft; please switch it back when you're ready to review. Thanks for your work on this :-)

RumovZ:
That should be it. Referring to the bullet points above:

1. Normalising is implemented on the backend.
2. A JSON AST for add-ons is not available, but a normalised search string should be nearly just as good.
3. - Negating and concatenating searches is now performed on the backend and will show an error for invalid input.
    - A single search node can now be replaced in an arbitrary search string by ctrl-shift-clicking a filter. Replacing with a more complex search string, like a group or a negated node, is not supported (will do nothing). It could be implemented but I think the interpretation of such an operation would be ambiguous and it's probably not worth the additional complexity in the Rust code.

qt/aqt/browser.py
```diff
@@ -1210,27 +1217,34 @@ def setFilter(self, *args):
...
                     txt = ""
-            txt = " ".join(items)
-        # is there something to replace or append with?
-        if txt:
+            txt = " AND ".join(items)
+        try:
             if self.mw.app.keyboardModifiers() & Qt.AltModifier:
-                txt = "-" + txt
-            # is there something to replace or append to?
+                txt = self.col.backend.negate_search(txt)
```
dae, line 1227: Ideally rsbackend.py will import (and thus re-export) required messages like ConcatenateSearchesIn, and pylib will provide simple wrappers so that users/GUI code doesn't need to call the backend directly - which gives us a bit more flexibility to change things around in the future. Looking through the codebase, I can see I haven't always managed to follow that rule myself however :-(

rslib/src/backend/mod.rs
```diff
@@ -426,6 +432,18 @@ impl BackendService for Backend {
         })
     }
 
+    fn negate_search(&self, input: pb::String) -> Result<pb::String> {
```
dae, line 435: Since these don't require the collection and should complete quickly, we can add them to want_release_gil() in rsbridge

rslib/src/search/writer.rs
```diff
@@ -0,0 +1,272 @@
...
+    } else {
+        write_node(&Not(Box::new(Group(nodes))))
+    })
+}
+
+/// Take arbitrary Anki-style search strings and return their concatenation where they
+/// are separated by the provided boolean operator.
+/// Empty searches (whole collection) are left out.
+pub fn concatenate_searches(sep: i32, searches: &[String]) -> Result<String> {
+    let bool_node = vec![match Separator::from_i32(sep) {
```
dae, line 42: It would be cleaner to define a native enum for sep, and convert from the Protobuf enum to the native one in the backend methods. That way if other parts of the Rust code want to call this routine, they don't need to  bother with Protobuf. You'll find a bunch of 'impl From' lines in backend/mod.rs so we can take a protobuf value and '.into()' it to get a native equivalent. I started using the from/into stuff part-way through - an older example of enum conversion is sort_kind_from_pb()

dae (commented):
Thanks Rumo, this looks great. I'll merge this in as-is, but I've left a few comments below - if you could at least address the Rust enum issue and rsbridge, that'd be great.

PR #862, opened 2020-12-26
Hide stats buttons in reviewer bottom bar, if window width is too small

hgiesel:
Specifically the "Edit" and "More..." buttons.

Before:
Screenshot 2020-12-26 at 15 53 29

After:
Screenshot 2020-12-26 at 15 49 56

Two caveats:
1. The specific pixel size prevents a scroll bar from ever appearing. I don't whether this is applicable for every platform though. 
2. The hiding is not really necessary on the front side, but I would argue hiding it there as well makes for a cleaner UI (otherwise buttons appear on the front, disappear on the back, and only reappear, if you make it bigger, which I would consider odd behavior).

### DIFF
--- a/qt/aqt/data/web/css/reviewer-bottom.scss
+++ b/qt/aqt/data/web/css/reviewer-bottom.scss
@@ -20,6 +20,10 @@ button {
 
 .stat {
     padding-top: 5px;
+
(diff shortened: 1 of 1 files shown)


dae:
Cheers

PR #863, opened 2020-12-26
Prevent stats screen from overscrolling

hgiesel:
If you're on the stats screen, and you scroll all the way to the bottom, or from the bottom all the way to the top, you _overscroll_ the page, which causes a "bouncing" effect of the scroll bar, which is annoying for the user, as it takes half a second too finish, for no good use.

This PR gets rid of this animation.


dae:
The slow bounce is a Qt bug unfortunately - it's been on their tracker for ages.

PR #864, opened 2020-12-26
Decouple GraphsPage from individual graphs and RangeBox

hgiesel:
This PR aims to decouple some svelte components for the stats, specifically:

1. Decouple GraphsPage from individual graphs:

Graphs are now passed as a list argument, and the GraphsPage has no explicit knowledge about them. They are all passed the same arguments.

This allows to easily create a `GraphsPage` with a different list of graphs, without having to modify it.

2. Decouple GraphsPage from RangeBox:

RangeBox is also passed in as a `controller` argument. It emits the `update` event, which GraphsPage is listening to.

There is some duplicated code here, when transforming `days` into `revlogRange`, which happens both in `GraphsPage` and `RangeBox`.
However I'd argue the duplication here is actually in the individual graphs requiring `revlogRange`, even though that information should already be present in `sourceData`.

This allows to easily create a `GraphsPage` with a different controller unit, or even none.

---
I specifically aimed to not modify any visible behavior. This should be a pure refactor.

### DIFF
--- a/ts/graphs/BUILD.bazel
+++ b/ts/graphs/BUILD.bazel
@@ -30,6 +30,7 @@ ts_library(
         "GraphsPage",
         "lib",
         "//ts/lib",
+        "@npm//svelte",
         "@npm//svelte2tsx",
     ],
 )
--- a/ts/graphs/GraphsPage.svelte
+++ b/ts/graphs/GraphsPage.svelte
@@ -13,6 +13,7 @@
 
     export let search: string;
     export let days: number;
+    export let withRangeBox: boolean;
 
     let sourceData: pb.BackendProto.GraphsOut | null = null;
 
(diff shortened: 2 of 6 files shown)


dae:
Hi Henrik,

The changes look good, thanks.

In terms of future changes, currently we have all graphs shown on one page, which means changes to the range selector need to redraw all the graphs at once, which can be slow. This could be fixed by splitting the graphs up into separate tabs, and/or allowing the user to customize which graphs are enabled and disabled - and the latter may allow for some extra graphs being distributed defaulting to off, sort of like blessed add-ons. Bearing that in mind, is there anything you'd want to change at this point, or are you happy to have the PR merged in as-is?

hgiesel:
I'm happy with the PR as is.

Concerning the future changes:
How would you like to design this "tabulation" function? I could imagine using some widget provided by jquery-ui? Or maybe try bootstrap?

dae:
Yeah, I was thinking something like bootstrap tabs. Not urgent, just something we'll need to address if we keep adding graphs in the future.

PR #866, opened 2020-12-27
Update consts.py

guillem-palau-salva:
I noticed that type 4 is added in the revlog when rescheduling. Reviews are logged as 0 time and type 4. I am not sure when such feature was introduced but I see it is not documented.


dae:
It's listed in revlog.rs as "Manual", as is "manually rescheduled". No objections to adding it to consts.py, but you'll need to add yourself to CONTRIBUTORS.

guillem-palau-salva:
done. I think it may be useful to have it here as well. We may save some time for anybody who wants to check that and doesn't look at the Rust code.

PR #867, opened 2020-12-27
split repo deps  as pylib_dep and desktop_dep

cecini:
For use pylib_deps.bzl and pylib_defs.bzl in the other project such as ankisyncserver

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -9,11 +9,11 @@ config_setting(
 
 genrule(
     name = "buildinfo",
-    srcs = ["//:defs.bzl"],
+    srcs = ["//:pylib_defs.bzl"],
     outs = ["buildinfo.txt"],
     cmd = select({
-        "release": "$(location //scripts:buildinfo) $(location //:defs.bzl) bazel-out/stable-status.txt release > $@",
-        "//conditions:default": "$(location //scripts:buildinfo) $(location //:defs.bzl) bazel-out/stable-status.txt devel > $@",
+        "release": "$(location //scripts:buildinfo) $(location //:pylib_defs.bzl) bazel-out/stable-status.txt release > $@",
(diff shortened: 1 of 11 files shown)


dae:
Hi Cecini,

I'm not sure there's much value in this:

- if you want to consume pylib without changes, you can use the pre-built wheels on pypi
- if you want to consume a modified version of pylib, you need to fork it anyway, and you can strip out any unwanted dependencies in the fork
- many of the dependencies are not fetched unless they are actually used
- the current divisions are wrong - ts/ is not specific to the desktop code, and can be used by the mobile clients as well

cecini:
> Hi Cecini,
> 
> I'm not sure there's much value in this:
> 
>     * if you want to consume pylib without changes, you can use the pre-built wheels on pypi

Maybe this is for the regular user.  For the develop build,  specifally for the bazel reproducially build , we maybe should use the pylib in the way.
> 
>     * if you want to consume a modified version of pylib, you need to fork it anyway, and you can strip out any unwanted dependencies in the fork

Since the pylib is  the base dependency of the syncserver, We should use the depend natively, not to fork or clone into the syncserver code. So best to leave it for the third party software use.
> 
>     * many of the dependencies are not fetched unless they are actually used
> 
I ack. But actually  i use the pip_install, not the pip_import. because i found pip_import 's  features smaller than the pip_install.  Such as pip_import cannot support local wheel in requirement.txt. 
>     * the current divisions are wrong - ts/ is not specific to the desktop code, and can be used by the mobile clients as well
I also admint that. This is pr just a first step. The further split still need to do.

dae:
I'm sorry, I'm still not convinced there's a need for this. You can override individual @repositories by defining them before calling setup_deps(). If your goal is to convert ankisyncserver to build with Bazel, have you checked with the other developers that it is even a direction they wish to go in? They appear to be using poetry at the moment.

PR #869, opened 2020-12-27
New field name checks are more specific

Arthur-Milchior:
There are two small bugs.
If you rename "Back" to " Front", it's accepted. Stripping should occur sooner.

If you create a field "#Test", it's accepted, but this field can't be displayed (I mean, it probably can using filters, but that's not really ideal), so I would suggest rejecting it

### DIFF
--- a/ftl/core/fields.ftl
+++ b/ftl/core/fields.ftl
@@ -13,3 +13,4 @@ fields-size = Size:
 fields-sort-by-this-field-in-the = Sort by this field in the browser
 fields-that-field-name-is-already-used = That field name is already used.
 fields-name-first-letter-not-valid = The field name should not start with #, ^ or /.
+fields-name-invalid-letter = The field name should not contain :, ", { or }.
--- a/ftl/core/fields.ftl
+++ b/ftl/core/fields.ftl
@@ -12,3 +12,4 @@ fields-reverse-text-direction-rtl = Reverse text direction (RTL)
 fields-size = Size:
 fields-sort-by-this-field-in-the = Sort by this field in the browser
 fields-that-field-name-is-already-used = That field name is already used.
+fields-name-first-letter-not-valid = The field name should not start with #, ^ or /.
--- a/qt/aqt/fields.py
+++ b/qt/aqt/fields.py
@@ -85,6 +85,9 @@ class FieldDialog(QDialog):
         txt = getOnlyText(prompt, default=old).replace('"', "").strip()
         if not txt:
             return
+        if txt[0] in "#^/":
+            showWarning(tr(TR.FIELDS_NAME_FIRST_LETTER_NOT_VALID))
+            return
         for f in self.model["flds"]:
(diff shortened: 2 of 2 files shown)


ftl/core/fields.ftl
```diff
@@ -12,3 +12,5 @@ fields-reverse-text-direction-rtl = Reverse text direction (RTL)
 fields-size = Size:
 fields-sort-by-this-field-in-the = Sort by this field in the browser
 fields-that-field-name-is-already-used = That field name is already used.
+field-name-first-letter-not-valid = The field name should not start with #, ^ nor /.
+field-name-invalid-letter = The field name should not contain :, { nor }.
```
dae, line 16: Note for next time - the prefix should match the filename, even if it's not grammatical - eg fields-first-letter-not-valid

dae (changes requested):
Hi Arthur,

These checks are best placed in the Rust code, so we don't need to reimplement them on each platform. fix_name() in fields.rs already exists, and could be extended to cover any cases that are not currently covered.

In principal warning the user is the best course of action, but the checking code runs not only on user input, but also on upgrade - we need to deal with bad data coming from previous Anki versions. We can't throw an error in that case, as it will prevent people from upgrading, so fix_name() opts for just silently removing the characters instead.

Arthur-Milchior:
I just taught me rust to understand the back-end, so I feel like I can do the change in the codebase.

I fear that applying the method which rename field to a field beginning with "{", "#", "/" or "^", ending with "}" or containing "}}" (and potentially ":", "{" and "}") will lead to erroneous template. More specifically, it will lead to them being eroneous in a potentially different way than the previous way the user knew. 
The advantage of doing the string transformation during the renaming/adding field is that it will ensure that the user know the template may have changed and will be less surprised.

I understand this should be a rare case anyway, so probably an acceptable risk, and I'm fine doing the same PR in rust if you want. Just wanted to be sure we considered this before I do it

dae:
:{} are already stripped, and I'm ok with removing a ^#/ prefix - any existing fields with that prefix will likely not have worked anyway.

Warning the user is the more user friendly option, but ensuring bad data doesn't get in is the priority, and that gets us most of the way there :-)

Arthur-Milchior:
Rust is changed in https://github.com/ankitects/anki/pull/870 

I just tested, no add-on, last version, renaming to "Ba{ck" lead to 
2020-12-28-031948_413x353_scrot

Indeed, it was stripped later, which I didn't test. That is confusing in my opinion and indeed would require warning, or at least showing in the list of field the final name that will be used. Are you willing to consider this PR ? Appart from changing the name of the string, is there any other change I must do to it?

dae:
I don't have any objections to adding a check in the GUI now that it's already covered in the backend. When you change the TR key, please change 'nor' to 'or' - 'nor' looks odd without 'neither' in front of it and with more than two choices

dae:
Cheers

PR #870, opened 2020-12-28
Trim the start of field name if it is #, /, ^ or a whitespace

Arthur-Milchior:
I tested and the initial # of "#back" was correctly removed.

I actually need to trim whitespace again to deal with a field name of the form "# foo"

This is a second try at https://github.com/ankitects/anki/pull/869
I leave the other PR open because I believe warnings are important too, even if the PR needs to be corrected

### DIFF
--- a/rslib/src/notetype/fields.rs
+++ b/rslib/src/notetype/fields.rs
@@ -42,7 +42,8 @@ impl NoteField {
             self.name = self.name.replace(bad_chars, "");
         }
         // and leading/trailing whitespace
-        let trimmed = self.name.trim();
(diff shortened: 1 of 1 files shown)


dae:
Looks good, thanks Arthur.

dae:
For reference, here is how you'd add a simple unit test:

https://github.com/ankitects/anki/commit/1d15a207bb1e8168daf51904a25da6f07bae9c90
https://github.com/ankitects/anki/commit/14c0066a534ebdd72ed68b2723bfece58f383e64

PR #872, opened 2020-12-28
Fix deck check and thus blue sync arrow bug

RumovZ:
[Your pointer](https://forums.ankiweb.net/t/why-is-my-sync-button-blue/2078/34?u=rumo) was spot on. A string was compared to an integer in `col.decks.select`.

Side note: During troubleshooting I noticed that `did:...` doesn't included cards of child decks (compared to `deck:...`). Is that intentional?

### DIFF
--- a/pylib/anki/decks.py
+++ b/pylib/anki/decks.py
@@ -430,7 +430,7 @@ class DeckManager:
 
     def selected(self) -> int:
         "The currently selected did."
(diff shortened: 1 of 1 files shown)


dae:
Thanks Rumo. In the long run we need to fix the other clients that are setting this to a string instead of an int.

PR #873, opened 2020-12-29
Update plotting library to 4.2.1 and include in Bazel build

hgiesel:
This also includes updating it from version 0.8.3 to 4.2.1.
Honestly, I thought that would break the stats page, but everything seems fine to me. If anything it might have gotten quicker (?).

EDIT: also installing flot caused the yarn.lock to reflow, creating this huge diff. I don't know if there's a way to prevent this, I often have this on my own projects as well, where I usually just suck it up.


hgiesel:
Never mind, I thought that plotting library was used for the new graphs...

However, it still stands that, surprisingly, the old stats still works. However updating it makes less sense, as it's already considered obsolete. I'll leave it up to you, [here are the versions that are available on npm](https://www.npmjs.com/package/flot).

dae:
Yeah, this was not a high priority as the old graphs are on the way out, but since you've done the work already, we may as well use it :-)

dae:
I had to revert this:

https://github.com/ankitects/anki/commit/9e1eaacc0695248d2f72b8b41cda4e0e9280c282

hgiesel:
Testing it again now, I also get the same issue. I don't know what I did the first time, I might have tested in the wrong branch 🤦 I'm sorry for the trouble.

dae:
No problem, appreciate your work on moving these JS deps over to Bazel.

PR #874, opened 2020-12-29
Update _updateQa

hgiesel:
1. Use `jQuery`s `.promise()`, now that we update jQuery to 3.x
2. Fix a typing error I've overseen (`() => Promise<any>[]` is a function that returns an array of promises, not an array of functions returning promises)
3. Prefer `declare var` over `@ts-ignore`
4. Use async/await syntax, making it easier on the eyes


dae:
Thanks Henrik - I'm relatively inexperienced with JavaScript, so the help is appreciated.

PR #876, opened 2020-12-29
Remove duplicated "in" in Python 3.8 instructions

gukgurgel:


dae:
Thanks Gustavo, I've pushed a similar change.

PR #877, opened 2020-12-30
Update Jquery UI and include in Bazel deps

hgiesel:
Same treatment for jquery-ui. Updates jquery-ui from 1.11.4 to 1.12.1.

Small caveat: I've installed from an npm repo called [jquery-ui-dist](https://www.npmjs.com/package/jquery-ui-dist), which I'm not sure is affiliated with the official [jquery-ui](https://www.npmjs.com/package/jquery-ui) repo, however the official one does not include the minified dist file.

This time I made sure to observe the broken behavior first, without the library, and that reinserting it fixed it again. :)
Also included again unfortunately is that giant yarn.lock reflow


dae:
Thanks!

dae:
The yarn.lock churn is likely due to you using a different yarn version on your machine. Please use the newly added ./add.sh in the future - it will use the pinned yarn version, and ensure the licenses are updated after adding.

PR #878, opened 2020-12-30
Replace `_updatingQA` with promise-based `_updatingQueue`

hgiesel:
This PR revises the synchronization mechanism in `reviewer.ts`.
Before this, you used a `_updatingQa` boolean flag, to check whether an update was already running.
This PR replaces this with a promise-based synchronization mechanism.
All updates are pushed to the global `_updatingQueue`. This ensures (unlike before) that queued actions are executed in order.

A fun way to test it, is executing this inside the webview:
```js
_queueAction(() => new Promise((resolve) => setTimeout(() => { console.log('First'); resolve(); }, 5000)));
_queueAction(() => new Promise((resolve) => setTimeout(() => { console.log('Second'); resolve(); }, 5000)));
_showAnswer('<i>Third</i>')
```
After 5s, "First" is logged. After another 5s, "Second" is logged. Almost immediately afterwards the web view will be replaced with an italic "Third". 

Additionally, I added types for the remaining functions in reviewer.ts.

### DIFF
--- a/qt/aqt/data/web/js/reviewer.ts
+++ b/qt/aqt/data/web/js/reviewer.ts
@@ -27,7 +27,12 @@ function _queueAction(action: () => Promise<void>): void {
     _updatingQueue = _updatingQueue.then(action);
 }
 
-async function _updateQA(html, fadeTime, onupdate, onshown): Promise<void> {
+async function _updateQA(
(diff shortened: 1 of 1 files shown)


dae:
Thanks!

PR #879, opened 2020-12-30
Add Alt text to latex image

Arthur-Milchior:
I'm trying to go over accessibilities issues in AnkiDroid. Since we'll use rust backend, I thought I might as well go
over it in rust directly. The only side effect in anki is that, if you copy a text with the image, you'll get the LaTeX
copied instead of nothing. Alas, it seems qt does not show alt text.


dae:
Looks good, thanks!

dae:
(Qt does appear to show alt text, but the desktop will always try to compile LaTeX and show errors if the compile failed)

PR #880, opened 2020-12-31
Add css-browser-selector and protobufjs as bazel deps

hgiesel:
I did the rest in one sweep.

protobufjs was updated from 6.10.1 to 6.10.2.
css-browser-selector was updated from 0.4.0 to 0.6.5.

It seems the newer version of css-browser-selector includes a few more utility classes:
On my reviewer webview for example, it went from:

```
night-mode webkit chrome mac js
```

to

```
webkit chrome chrome83 mac mac10 mac10_16 js night-mode retina orientation_landscape maxw_1440
```


dae:
Thanks as always Henrik :-) Looks like we already had 6.10.2 in the lock file, so this should only change css-browser-selector.

PR #881, opened 2021-01-01
Support prop:order to search for new cards by order

abdnh:
This can be useful for quickly finding new cards with high due numbers.

### DIFF
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -97,7 +97,7 @@ pub(super) enum PropertyKind {
     Reps(u32),
     Lapses(u32),
     Ease(f32),
-    Order(u32),
+    Position(u32),
 }
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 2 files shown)


dae:
You can already accomplish this with 'is:new prop:due>xxx', can't you? If so, I'm not sure it's worth adding a separate search term just to save 5 characters here - if it's a frequently used search, the user can already save the search in the browse screen.

dae:
(Happy new year by the way :-)

abdnh:
AFAIK, `is:new` can't be combined with `prop:due` since the latter [assumes review and learning cards](https://github.com/ankitects/anki/blob/eecdc07c686f4a6f6ede2cf23e3c90b64380a7d4/rslib/src/search/sqlwriter.rs#L243).
I guess it's more ideal to make `prop:due` work for new cards too (by removing the queue check altogether) but I thought this is a change you may not want to introduce.

Happy new year!

dae:
You're quite right, I'd forgotten about the existing limit on review cards.Your approach looks good, but could we rename 'order' to 'pos' or 'position'?

abdnh:
I initially named it 'pos' actually but changed to 'order' because I thought it's already short and easily understood without being abbreviated. I'll change it anyway.

dae:
Perhaps a subtle distinction, but I do prefer pos/position - each card has a position, and those positions determine the order the cards will be shown in. Thanks for the feature and for making the change :-)

PR #882, opened 2021-01-01
Implement copy_select_files for removing duplication in js dep Bazel recipes

hgiesel:
This avoids the copy-paste of the "selection" algorithm for each js dependency.
I also added another argument, called `base`, which is supposed to be the entry point in the repo, where the search for files to be copied starts. This allowed me to make the search for files a bit stricter, by using `str.startswith()` instead of `in`. This might prevent some unexplainable behavior in the future, where it copies files we did not intend it to.

I also fixed a small typing error for _drawFlag, which otherwise wouldn't be worth a PR.

EDIT: Also, happy new year :)

### DIFF
--- a/qt/aqt/data/web/js/reviewer.ts
+++ b/qt/aqt/data/web/js/reviewer.ts
@@ -120,7 +120,7 @@ const _flagColours = {
     4: "#77aaff",
 };
 
-function _drawFlag(flag: 0 | 1 | 2 | 3): void {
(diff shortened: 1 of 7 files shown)


dae:
Happy new year Henrik :-)

This looks good, and I think we can even take it a little further - I'll merge this and push a follow-up.

dae:
Follow-up commit: https://github.com/ankitects/anki/commit/5ec5a4770885e0d77f709ee0d3fe1649610a3e5f

PR #884, opened 2021-01-02
refresh deck browser after renaming deck in browser sidebar

BlueGreenMagick:
So the deckbrowser updates its deck name


dae:
This will force the screen to change if the user is not in the deck browser, but perhaps that's not a bad thing, as it means the other screens don't need to have their own refreshing behaviour. Thanks!

PR #885, opened 2021-01-02
NF: Add a hook to change the default search from reviewer

Arthur-Milchior:
The goal here is to simplify my add-on 1555020859 which show a single card or the whole deck when the browser open


dae:
Looks good, thanks.

PR #886, opened 2021-01-03
add rename_tag to browser sidebar contextmenu

BlueGreenMagick:
This PR adds 'rename' context menu to browser sidebar tags.

Notes

1. To remove old_name tag from the sidebar the code calls `clearUnusedTags()`, which has the side effect of clearing all other unused tags. Even though`clearUnusedTags` is quite heavy, it may be good to consider calling it when initializing Browser.
2. when new_name is blank, the code cancels rename instead of deleting the tag, as I assume a 'delete tag' context menu will be added in the future.
3. Would it be better to add a rename function to tags.py instead of implementing it in sidebar.py?

Related: ankitects/help-wanted#6

### DIFF
--- a/ftl/core/actions.ftl
+++ b/ftl/core/actions.ftl
@@ -24,6 +24,7 @@ actions-rebuild = Rebuild
 actions-red-flag = Red Flag
 actions-rename = Rename
 actions-rename-deck = Rename Deck
+actions-rename-tag = Rename Tag
 actions-replay-audio = Replay Audio
 actions-reposition = Reposition
 actions-save = Save
--- a/ftl/core/browsing.ftl
+++ b/ftl/core/browsing.ftl
@@ -94,6 +94,7 @@ browsing-step = Step:
 browsing-studied-today = Studied Today
 browsing-suspended = Suspended
 browsing-tag-duplicates = Tag Duplicates
+browsing-tag-rename-warning-empty = You can't rename a tag that has no notes.
 browsing-target-field = Target field:
 browsing-today = Today
 browsing-toggle-mark = Toggle Mark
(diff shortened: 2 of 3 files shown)


abdnh:
By the way, I have done some work a while ago on hierarchical tags support along with tag rename/remove but it needs more testing.
Now that there is more activity in this area, I may open a PR, maybe after this gets merged and after integrating my changes with the latest master.

dae:
Thanks for contributing!

A fair amount of effort has been put into making browser load quickly, so I'm afraid I don't think clearing unused tags on load is a good idea.

This will need to be run on a background thread to avoid freezing the UI; please see the follow-up I push to see how you can do it for next time.

The search at the start is not necessary, as you can check the number of changes returned from the bulk_update call. The remaining code is basically just GUI glue, so in this case there's probably no need/benefit from moving it into libanki.

dae:
Sorry, thinko on my part - we do of course need to know which notes need updating. Ideally this would be exposed as a backend method from Rust, but it's fairly trivial so I'll pop it in tags.py for now.

dae:
Here's the background change:
https://github.com/ankitects/anki/commit/f3fa9daae252e2be56ee98f3bb9a747349517919

Looks like the commit removing quotes was useless - black is adding them back.

BlueGreenMagick:
I think the current code still freezes the UI as `self.browser.model.beginReset()` calls `mw.progress.start()`.

But just removing mw.progress.start would create several bugs. Consider the following:
1. Tag 'a' is being renamed in the background
2. User loads a note with tag 'a' to the browser editor
3. Tag 'a' of that note is renamed to 'b' in the background
4. User finishes editing the note -the note is saved with tag 'a'

dae:
The goal is not to allow the user to make concurrent changes, it's simply to ensure the UI repaints while the operation is running. If you run the long-running op on the main thread, the GUI will freeze until the task completes. 

https://github.com/ankitects/anki/commit/0f38514ad705aa42cbe02276cd5929e0be6cf139

BlueGreenMagick:
Thank you for clarifyng!

PR #889, opened 2021-01-03
Escape literal curly braces in Fluent

abdnh:
Pontoon checks fail because `{ or }` is interpreted as a message reference.


dae:
Thanks Abdo, I'll add a lint for this.

PR #890, opened 2021-01-03
Add closeWithCallback method to Addons dialog

hgiesel:
It seems like, when Anki get's a SIGINT, it tries to `closeWithCallback` on the current dialog.

However, `AddonsDialog` does not have this method, so sending SIGINT, while the addons window is open, evokes this error message:

    AttributeError: 'AddonsDialog' object has no attribute 'closeWithCallback'

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -735,6 +735,8 @@ class AddonsDialog(QDialog):
 
         return QDialog.reject(self)
 
+    silentlyClose = True
+
     def name_for_addon_list(self, addon: AddonMeta) -> str:
         name = addon.human_name()
 


dae:
Since the add-ons screen doesn't have any prompt on close, perhaps silentlyClose would be a better option?

```
 Dialog manager
# ensures only one copy of the window is open at once, and provides
# a way for dialogs to clean up asynchronously when collection closes

# to integrate a new window:
# - add it to _dialogs
# - define close behaviour, by either:
# -- setting silentlyClose=True to have it close immediately
# -- define a closeWithCallback() method
```

dae:
Also might be worth testing if things work if a child window like the edit config screen is open at the time.

hgiesel:
I switched to `silentlyClose`.

Config windows print this:

> Window should have been closed: <aqt.addons.ConfigEditor object at 0x7f92cae84040>

My custom modal windows behave similar:

> Window should have been closed: <MyAddonDev.gui.settings.Settings object at 0x7fd91fb7f040>
> 2021-01-04 11:51:42.379 Python[65254:7074120] modalSession has been exited prematurely - check for a reentrant call to endModalSession:

dae:
In the long term we should be closing them properly. The add-ons screen could take responsibility for that in the closeWithCallback you originally had, but it's not really a high priority, as this is a pretty niche case - let's just fix the error for now and move on :-)

hgiesel:
I assume that would mean it calls `reject()` on all of its children dialogs?
In most of my cases, `reject` is essentially "not saving / not doing anything", so it works for me, but I can imagine cases where that wouldn't be the case.

Maybe I'll have a look at it at some point.

dae:
I presume it would deallocate the children when the parent is rejected, but would need to test to be sure :-)

PR #891, opened 2021-01-04
Sidebar delete deck

BlueGreenMagick:
Add 'delete' to sidebar deck context menu.

### DIFF
--- a/qt/aqt/deckbrowser.py
+++ b/qt/aqt/deckbrowser.py
@@ -283,17 +283,16 @@ class DeckBrowser:
 
     def ask_delete_deck(self, did: int) -> bool:
         deck = self.mw.col.decks.get(did)
-        extra = None
-        if not deck["dyn"]:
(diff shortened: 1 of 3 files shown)


dae:
I like the feature, just a couple of points:

- can we break the code into a function that is shared between deckbrowser.py and this file, instead of copying & pasting?
- a fut.result() call is required in on_done to check for any errors that happened in the background - you can discard the result; it will raise on error

BlueGreenMagick:
Regarding having a shared function that asks for deck rename, do you want me to put it in utils.py, or for sidebar to call it from deckbrowser.py? Or is there a better option?

BlueGreenMagick:
Done. Is this alright?

dae:
Looks good, thanks!

Some minor points:
- the count checking should ideally also be in the background, which means that ask_deck_delete would need to be converted to use a callback. It's not a dealbreaker, as there are still lots of other places in the code that need similar updating, and the counts check is usually not too slow.
- future.result() can be used without assigning to anything

PR #892, opened 2021-01-04
Add "By card type" and "By scheduling queue" variants of Card Counts stats graph

hgiesel:
This adds a radio button group to the Card Counts Graph.

This deals with one of my issues I had with this graph. Imagine the following scenario:
1. I have a deck with five new cards all belonging to the same note, with sibling burying enabled.
2. My reviews finish after one card.

Now I ask myself "How many new cards do I have in this deck? Do I need to make new cards?". But what I see on the graph is this:

Screenshot 2021-01-04 at 14 00 47

Of course in this simple case it's easy to infer the amount of new cards. But what if I also have buried review cards, user-buried cards, etc. What I'd rather see would be this:

Screenshot 2021-01-04 at 15 30 51

So, none of the old information is lost: The only change to the queue pie, is that instead of Young/Mature in the scheduling graph, we have Learn/Review. However, the Young/Mature distinction can be better made in the card type pie, where it is not influenced by Buried/Suspended cards.

Note, that this PR is not quite complete:
1. There are some translations that would have to be added (see other PR)
2. The colours are a point of contention


dae:
I get the problem you're trying to solve, but this feels a bit complicated at the moment. What if instead of two radios, you had a single "separate buried/suspended" checkbox? When disabled, the total would stay the same, but the buried/suspended cards would move into the appropriate categories (eg new, mature, etc). No need to think about different colours, or understand the distinction between 'type' and 'queue'. What do you think?

hgiesel:
Okay, I think that might actually be a better idea, toggling those two categories on and off.

However, what do you think about the distinction between "Learning" and "Relearning" in the graph? Having them summed up under "Young" seems very confusing to me. Even the [Anki manual](https://docs.ankiweb.net/#/stats?id=types-of-cards) states "A young card is one that has an interval of less than 21 days, but is not in learning.". This graph does not adhere to this, and makes learning cards young cards.

dae:
Oh, I'd forgotten the graphs don't already have Learning. No objections to adding learning/relearning - it was just the radio buttons showing different things that I found confusing. For the colours, maybe red for learning and darker red for relearning, similar to the light green/dark green young/mature?

hgiesel:
Sounds good, I'll work on it.

Another question, somewhat related to this PR: I think a really cool feature would be to connect the stats graph with the card browser. E.g. clicking on "New" in the table data of Card Count and get all new cards in the browser. Similar to the tags in the browser, this would facilitate the handling of cards for users, who are not that comfortable with writing queries (esp. complicated ones, like "how do you search for relearning cards").

However I know you use this code on mobile. I noticed that in the congrats screen you use `bridgeLink` which makes a call to `bridgeCommand` (which on Desktop is an alias for `pycmd`) for both "unbury" and "customStudy". You think it would be feasible to extend this to a "browserSearch:argument" call, which opens the card browser and searches for the argument?

The individual graph would emit a message "search requested", with a string of the subquery, e.g. `is:new`, communicate this up to the `GraphsPage` container, which then concatenates it with the base search, e.g. `deck:current`, and then calls `bridgeCommand("browserSearch:<searchQuery>")`.

However I also noticed that the congrats page has this `info.bridgeCommandsSupported`.
I assume the `GraphsOut` would also need this attribute then?

hgiesel:
I redesigned it, and now it looks like this:

Screenshot 2021-01-05 at 16 49 54
Screenshot 2021-01-05 at 16 49 58

In the code, I had to add another `bool` to the `Count` type, because otherwise, the pie graph didn't correctly update. Instead of growing (tweening) into its size, the buried and suspended cards simply popped up.

With more knowledge of the d3 libraries, that boolean might be optimized away.

dae:
Looks good, thanks for your work on this Henrik. For a single checkbox in the future it may be worth just inlining it instead of creating a separate file, but it's not something I feel strongly about.

I don't think we could use the bridgeCommandsSupported trick in this case - GraphsOut is potentially tens of megabytes, and we don't really want the client code to have to deserialize it just to adjust a property, then re-serialize it before passing it on. It could potentially be handled in some other way, but I do wonder whether enough people would find it useful to justify the effort. If this is about discoverability for new users, what about just giving each of the sections a tooltip of the form "To find these cards in the browse screen, search for is:suspended". But we do already have the Filter button→Card State menu item as well.

dae:
Just in terms of the default, I think it might be better to default to including suspended cards - that's what we had previously, and hiding them by default may lead some users to wonder why some card types have non-zero numbers but are not showing up. What do you think?

hgiesel:
I think my reasoning for defaulting to off is that I typically expect check boxes to default to off. Because the label of a checkbox kinda sounds like a command, so rather than => 

[x] Separate suspended / buried cards

it could be something like:

[ ] Merge in suspended / buried cards

I could follow that up in another PR.

dae:
Hmm, I'm not sure that's necessary. The backlog checkbox on another graph already defaults to on, and there are a whole bunch of options in the preferences that default to on. IMHO the priority should be on natural wording, rather than having things all default to off. From https://ux.stackexchange.com/questions/115263/should-checkboxes-be-checked-or-unchecked-by-default-in-forms:

"No general rule about UI widget states can trump the fundamental rule that your users must understand what they are saying. Always choose the simpler, more intellegible phrasing, then decide about the state of the checkbox based on that."

PR #894, opened 2021-01-04
Add note_did_load hook

hgiesel:
This hook is the opposite of `note_will_flush`, it is invoked every time a card is loaded from the database, or newly created.

It can be used to set some custom data on the `Note` object.


dae:
Mind elaborating on the use case? There's an existing hook in the editor, so I presume you want to be able to do this during review?

hgiesel:
Hm, after sleeping another night on it, I can't say it provides me with any functionality I wouldn't have without it.
Even though I like the symmetry to `note_will_flush`, I will close this one.

PR #895, opened 2021-01-05
Change meaning of "is:learn" to filter on type, not queue

hgiesel:
Currently "is:learn" filters on the queue of a card, not the type.
This makes it impossible to compose with the "is:buried", and "is:suspended" filters. Some examples

1. I'd like to find all my suspended learning cards:
My first instinct would be `is:learn is:suspended`, which would not yield anything currently.
2.  I'd like to find all relearning cards:
The Anki manual suggests: `is:review is:learn`, however this only includes cards which are not also buried a/o suspended. There is currently no way to include those.

I don't think this behaviour makes sense to the uninitiated user. It requires knowledge of the scheduling queues.

After this change (`is:new` | `is:review is:learn` | `is:review -is:learn` | `-is:review is:learn`) would all behave nicely when paired with (`is:buried` | `is:suspended` | negated versions), and all examples in the [manual](https://docs.ankiweb.net/#/searching?id=card-state) still work as described as well.

### DIFF
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -279,9 +279,9 @@ impl SqlWriter<'_> {
             ),
             StateKind::Learning => write!(
                 self.sql,
-                "c.queue in ({},{})",
-                CardQueue::Learn as i8,
-                CardQueue::DayLearn as i8
+                "c.type in ({}, {})",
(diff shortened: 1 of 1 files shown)


dae:
Sounds good - in the v1 scheduler these are synonymous since learning cards can't be suspended, but that's no longer the case with v2.

PR #896, opened 2021-01-06
Default CardCounts separate suspended/buried to on

hgiesel:
Continuation of https://github.com/ankitects/anki/pull/892

Okay, you're right.
In that case, "separating the cards" is more precise in the wording.


dae:
Oh, sorry, I could have done that :-) I was more bringing it up to discuss it than asking you to make the change. But thank you!

PR #897, opened 2021-01-06
Triggering searches from the stats screen.

hgiesel:
This conversation started in https://github.com/ankitects/anki/pull/892

> I don't think we could use the bridgeCommandsSupported trick in this case - GraphsOut is potentially tens of megabytes, and we don't really want the client code to have to deserialize it just to adjust a property, then re-serialize it before passing it on. It could potentially be handled in some other way, but I do wonder whether enough people would find it useful to justify the effort. If this is about discoverability for new users, what about just giving each of the sections a tooltip of the form "To find these cards in the browse screen, search for is:suspended". But we do already have the Filter button→Card State menu item as well.

1. I don't exactly understand how `bridgeCommandsSupported` works, tbh. I think the stats are used on two platforms at the moment, right? Anki and AnkiMobile. And I do see the activated bridgeCommands on both platforms.

2. Also I don't exactly know why the data needs re-serialization? So if I understand correctly, you intercept the `CongratsOut` data, only to set that property and then pass it on?

I would prefer a "show, don't tell" approach in that case. Opening the browser with the search typed in, delivers an obvious message of "Let me google that for you", but can also be used by those, who can't be bothered to remember the queries, or feel overwhelmed by the idea of entering "special commands" to search.

### DIFF
--- a/qt/aqt/deckbrowser.py
+++ b/qt/aqt/deckbrowser.py
@@ -65,7 +65,7 @@ class DeckBrowser:
 
     def _linkHandler(self, url):
         if ":" in url:
-            (cmd, arg) = url.split(":")
+            (cmd, arg) = url.split(":", 1)
         else:
             cmd = url
         if cmd == "open":
--- a/qt/aqt/stats.py
+++ b/qt/aqt/stats.py
@@ -51,6 +51,7 @@ class NewDeckStats(QDialog):
         gui_hooks.stats_dialog_will_show(self)
         self.show()
         self.refresh()
+        self.form.web.set_bridge_command(self._on_bridge_cmd, self)
         self.activateWindow()
 
     def reject(self):
@@ -88,6 +89,14 @@ class NewDeckStats(QDialog):
 
     def changeScope(self, type):
         pass
+
+    def _on_bridge_cmd(self, cmd: str) -> bool:
+        if cmd.startswith("browserSearch"):
+            _, query = cmd.split(':', 1)
+            browser = aqt.dialogs.open("Browser", self.mw)
+            browser.search_for(query)
+
+        return False
 
     def refresh(self):
         self.form.web.load_ts_page("graphs")
--- a/qt/aqt/stats.py
+++ b/qt/aqt/stats.py
@@ -92,7 +92,7 @@ class NewDeckStats(QDialog):
 
     def _on_bridge_cmd(self, cmd: str) -> bool:
         if cmd.startswith("browserSearch"):
-            _, query = cmd.split(':', 1)
+            _, query = cmd.split(":", 1)
             browser = aqt.dialogs.open("Browser", self.mw)
             browser.search_for(query)
 
--- a/rslib/backend.proto
+++ b/rslib/backend.proto
@@ -1092,6 +1092,7 @@ message GraphsOut {
   uint32 scheduler_version = 5;
   /// Seconds to add to UTC timestamps to get local time.
   int32 local_offset_secs = 7;
+  bool bridge_commands_supported = 8;
 }
 
 message GraphPreferences {
--- a/rslib/backend.proto
+++ b/rslib/backend.proto
@@ -1092,7 +1092,6 @@ message GraphsOut {
   uint32 scheduler_version = 5;
   /// Seconds to add to UTC timestamps to get local time.
   int32 local_offset_secs = 7;
-  bool bridge_commands_supported = 8;
 }
 
 message GraphPreferences {
@@ -1104,6 +1103,7 @@ message GraphPreferences {
   }
   Weekday calendar_first_day_of_week = 1;
   bool card_counts_separate_inactive = 2;
+  bool browser_links_supported = 3;
 }
 
 message RevlogEntry {
--- a/rslib/src/stats/graphs.rs
+++ b/rslib/src/stats/graphs.rs
@@ -44,7 +44,6 @@ impl Collection {
             next_day_at_secs: timing.next_day_at as u32,
             scheduler_version: self.sched_ver() as u32,
             local_offset_secs: local_offset_secs as i32,
-            bridge_commands_supported: true,
         })
     }
 
@@ -52,6 +51,7 @@ impl Collection {
         Ok(pb::GraphPreferences {
             calendar_first_day_of_week: self.get_first_day_of_week() as i32,
             card_counts_separate_inactive: self.get_card_counts_separate_inactive(),
+            browser_links_supported: true,
         })
     }
 
--- a/rslib/src/stats/graphs.rs
+++ b/rslib/src/stats/graphs.rs
@@ -44,6 +44,7 @@ impl Collection {
             next_day_at_secs: timing.next_day_at as u32,
             scheduler_version: self.sched_ver() as u32,
             local_offset_secs: local_offset_secs as i32,
+            bridge_commands_supported: true,
         })
     }
 
--- a/ts/graphs/AddedGraph.svelte
+++ b/ts/graphs/AddedGraph.svelte
@@ -39,7 +39,7 @@
         <GraphRangeRadios bind:graphRange {i18n} revlogRange={RevlogRange.All} />
     </div>
 
-    <HistogramGraph data={histogramData} {i18n} />
+    <HistogramGraph data={histogramData} {i18n} on:search />
 
     <TableData {i18n} {tableData} />
 </div>
--- a/ts/graphs/AddedGraph.svelte
+++ b/ts/graphs/AddedGraph.svelte
@@ -9,6 +9,7 @@
     import HistogramGraph from "./HistogramGraph.svelte";
     import GraphRangeRadios from "./GraphRangeRadios.svelte";
     import TableData from "./TableData.svelte";
+    import { createEventDispatcher } from "svelte";
 
     export let sourceData: pb.BackendProto.GraphsOut | null = null;
     export let i18n: I18n;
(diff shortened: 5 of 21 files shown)


hgiesel:
And concerning demand. I agree that for the "Card Counts" section, it's quite easy to search for. But hopefully enabling it for one graph, means that we could use it for all of them. "Which cards did I answer hard on", "Which cards did I review September 17th?". I know that these are all technically searchable.

I remember from one conversation you saying, that most people don't enter tag searches, but just click on the tag in the sidebar. I think that will also be true for a lot of these other cases. A lot of people will probably never use more complicated queries, like `rated`, but they would click on the graph to search for it.

hgiesel:
After investigating the Rust code a little bit more, I think I know understand the issue:
1. Both the rustlib and the stats screens are supposed to be reused on other platforms
1. However those other platforms might not support the notion of `bridgeCommand` in the webview where they are gonna render the stats
1. Which is why the rustlib exports the `bridgeCommandsSupported` property, which defaults to true, but could be intercepted by that other platform and set to false, thus rendering the stats slightly differently, without scenarios, where it would normally use bridge commands.

However: doesn't the webview only truly support bridge commands, if it implements the bridgeCommand function? So rather than using:

```javascript
{#if data.bridgeCommandsSupported}
    // use bridge commands
{/if}
``` 

just use:

```javascript
{#if bridgeCommand}
    // use bridge commands
{/if}
/* or */
{#if bridge}
    bridge.bridgeCommand("cmd")
{/if}
```

This would also make the process of disabling bridge commands easier, as the webview could just do nothing or alternatively explicitly set `bridgeCommand = null`, and would allow us to remove `bridgeCommandsSupported` from the rust payload.

This suggestion would be independent of this PR.

dae:
Perhaps it's worth making a short video of this feature, and posting it over on the forum with a "would you use this?" poll, and perhaps asking people to elaborate on their use case if they would. Without any data, we're left just guessing as to the usefulness/demand.

Not all of the graphs could be converted into browser queries - rated searches are limited to 1 year for performance reasons for example.

Currently we have different HTML pages for different screens, but in the long run it probably makes sense to migrate to something closer to a SPA, and a check on the presence of bridgeCommand might not work in that case. While the feature can't work without bridgeCommand, a client may be able to support bridgeCommands but not want to implement this -AnkiMobile doesn't support accessing the browse screen from within the graphs screen for example. The current check in the congrats screen is necessary because AnkiWeb doesn't support custom study or unbury actions at the moment, so it would not be able to move to the  new congrats screen if those links/messages were hard-coded - even if it could emulate bridgeCommand() with a http request.

I do think it's a neat feature. It's not particularly invasive to add, and some other solution for disabling this selectively shouldn't be too hard. But before we proceed, I'd just like to see what sort of demand there is for this.

hgiesel:
I created a [poll in the Anki forum].

dae:
Thanks Henrik. Let's run it for a week, but initial indications seem quite positively in favour of it, and your recent changes to searches have expanded the scope of what can be searched for.

Coming back to the bridgeCommandsSupported topic, if the backend had something like GetGraphPreferences and SetGraphPreferences, it would make it easy to add extra settings in the future, and we could throw bridgeCommandsSupported in there as well (perhaps giving it a more specific name like browserLinksSupported). As it would be a fairly small structure, it would be cheap for the frontend code to adjust before passing on to typescript. The weekday field added in your recent PR coud be moved to the graph preferences, and the graphs code could just fetch the graph data and preferences separately. The recent 'buried/suspended separately' checkbox might be another useful thing to persist to a preference. What do you think?

hgiesel:
@dae I think this sounds like a good idea, and I think a first implementation can be within the scope of this PR. What I'm not quite sure is how we should deal with graph-specific preferences. `browserLinksSupported` would be valid for all graphs, whereas something like the card counts checkbox is very graph-specific.

I would probably implement it as one controller config object and for each graph a graph-specific config object, where the saving is triggered by a message from within the individual `Graph.svelte` files, similar to how the controller (`RangeBox`) currently invokes the `update` message.

dae:
I was thinking of just having a flat namespace in a single proto message, with a prefix on keys where necessary, eg

```proto
message GraphPreferences {
  enum Weekday {
    SUNDAY = 0;
    MONDAY = 1;
    FRIDAY = 5;
    SATURDAY = 6;
  }

  bool browser_links_supported = 1;
  bool card_counts_separate_suspended = 2;
  int64 card_counts_some_other_future_var = 3;
  Weekday calendar_start_day = 4;
}
```

On the TS side, separate controllers and a message sounds like it could work. I haven't played with them before, but perhaps [Svelte stores](https://svelte.dev/tutorial/writable-stores) might be worth considering as well - by the looks of it, we could have a single preferences controller that exports each preference as a separate observable value. I think graphs could then import the values they were interested in, and the controller could take care of updating the backend when a value is changed. But having not used it in practice, I do not know if it would be a win over your message approach or not.

hgiesel:
Having looked at Svelte stores, I'd say they are a perfect match.

Two points:
1. Would a simple encompassing `SetGraphPreferences` suffice? I'd imagine that each individual change of settings (e.g. the user clicks the "Separate Suspended / Buried" checkbox) would have to be immediately synced, because we cannot know when the webview will be closed. Maybe it could work on the `onAccept`/`onRefuse` method of the stats QDialog, but I don't think what's the direction the stats graph is heading. The other option is sending the whole preferences, as soon as one preference changes, which I also don't think is ideal.

2. I cannot use imports from svelte packages from `.ts` files other than `bootstrap.ts`. For example, putting `import { writable } from 'svelte/internal';` into `calendar.ts` will cause the following error when trying to `./run`:

```
ERROR: /Users/hgiesel/Developer/repos/anki/ts/graphs/BUILD.bazel:43:11: Compiling TypeScript (devmode) //ts/graphs:lib failed (Exit 1): tsc_wrapped.sh failed: error executing command bazel-out/host/bin/external/npm/@bazel/typescript/bin/tsc_wrapped.sh @@bazel-out/darwin-fastbuild/bin/ts/graphs/lib_es5_tsconfig.json
ts/graphs/calendar.ts:9:26 - error TS2307: Cannot find module 'svelte/internal' or its corresponding type declarations.

import { writable } from 'svelte/internal';
                         ~~~~~~~~~~~~~~~~~
```

dae:
Bazel requires the dependency to be explicitly listed:

```patch
diff --git a/ts/graphs/BUILD.bazel b/ts/graphs/BUILD.bazel
index 0db44c94d..34d8083bc 100644
--- a/ts/graphs/BUILD.bazel
+++ b/ts/graphs/BUILD.bazel
@@ -71,6 +71,7 @@ ts_library(
         "@npm//d3-transition",
         "@npm//lodash.debounce",
         "@npm//lodash.throttle",
+        "@npm//svelte",
     ],
 )

diff --git a/ts/graphs/calendar.ts b/ts/graphs/calendar.ts
index a5b1deb23..5ba89cbfc 100644
--- a/ts/graphs/calendar.ts
+++ b/ts/graphs/calendar.ts
@@ -23,6 +23,7 @@ import {
 } from "d3-time";
 import type { CountableTimeInterval } from "d3-time";
 import type { I18n } from "anki/i18n";
+import { writable } from "svelte/store";

 export interface GraphData {
     // indexed by day, where day is relative to today
```

I'd be tempted to take the YAGNI option for now, writing out all preferences every time one is changed. I don't think it will grow past 10-20 settings in the near future, and that is cheap to read and write. The store will abstract the storage from the rest of the code, so if we need/want to switch to multiple methods that persist subsections of the total preferences, only the store's internal code and the backend methods would need to change, as we'd already be exposing per-setting getters/setters to the individual graphs.

One alternative would be to expose a method that allows writing a single preference via a enum. I have no objections to doing it that way if that's what you prefer, and am open to other ideas if you had some other plan.

hgiesel:
I have issues making the code type check. I saw you included a `d3_missing.d.ts`. However I don't think it actually utilizes this file?

dae:
Maybe there's a better way, but this seems to compile:

```patch
diff --git a/ts/graphs/GraphsPage.svelte b/ts/graphs/GraphsPage.svelte
index f26aae9fa..2597a09f0 100644
--- a/ts/graphs/GraphsPage.svelte
+++ b/ts/graphs/GraphsPage.svelte
@@ -8,6 +8,7 @@
     import type pb from "anki/backend_proto";
     import { getGraphData, RevlogRange, daysToRevlogRange } from "./graph-helpers";
     import { getPreferences } from "./preferences";
+    import { bridgeCommand } from "anki/bridgecommand";

     export let i18n: I18n;
     export let nightMode: boolean;
diff --git a/ts/lib/bridgecommand.ts b/ts/lib/bridgecommand.ts
index b7a0f721b..ff506fabf 100644
--- a/ts/lib/bridgecommand.ts
+++ b/ts/lib/bridgecommand.ts
@@ -5,3 +5,13 @@
 export function bridgeLink(command: string, label: string): string {
     return `<a href="javascript:bridgeCommand('${command}')">${label}</a>`;
 }
+
+declare global {
+    interface Window {
+        bridgeCommand: (string) => void;
+    }
+}
+
+export function bridgeCommand(command: string): void {
+    window.bridgeCommand(command);
+}
```

Svelte's TypeScript support is a bit of a hack unfortunately, so I try to keep as much of the TypeScript in separate .ts files, and just call into it from the Svelte.

dae:
Thanks Henrik, I had a play with this and it does work well - maybe it will help users discover some of the searching options available in the future as well.

The only slightly negative thing I noticed was the lack of typing for 'dispatch'. What do you think about wrapping the dispatch functionality in a simple class like "ControllerHandle", which has a .search() method on it, which calls the underlying dispatch object? We could then pass around a type-safe object, and if we need to pass further messages to the controller in the future, it would be easy to extend.

PR #898, opened 2021-01-06
Activate items of the new sidebar only when left-clicked

abdnh:
I think this makes more sense now that we have a context menu.
The behavior of the old sidebar is unchanged.


dae:
I think this can be done more simply - please let me know if the pushed change doesn't work for you.

PR #899, opened 2021-01-06
remove context help button

BlueGreenMagick:
Screenshot_20210106_221752

This PR removes 'context help' button from all dialogs, which doesn't do anything when clicked.

For some reason, mypy thinks the type of `dialog.setWindowFlags(self.windowFlags() & ~Qt.WindowContextHelpButtonHint` is an `int`, even though it's actually `Qt.WindowFlags`

### DIFF
--- a/qt/aqt/about.py
+++ b/qt/aqt/about.py
@@ -28,6 +28,7 @@ class ClosableQDialog(QDialog):
 
 def show(mw):
     dialog = ClosableQDialog(mw)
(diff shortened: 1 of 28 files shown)


dae:
Thanks!

Just a note for next time, this would have been a good candidate to create a helper def in utils.py, then each line could be replaced with something like disable_help_button(self)

dae:
This introduced some bugs, so I went ahead and added the helper.

https://github.com/ankitects/anki/commit/94064b8230adbd956ba1c68cee7306fbd3f3f623

PR #900, opened 2021-01-06
Hierarchical tags

abdnh:
Related: https://github.com/ankitects/help-wanted/issues/6

A few notes:
- Some parts are adopted from the deck tree code with the necessary changes.
- I don't know if tag IDs are a good idea. I felt the need to add them just because tags are getting heavy now but I guess we can do without them.
- I also added remove_tag to the sidebar along with the backend method clear_tag so we don't need to use clearUnusedTags.

Everything works according to my tests, but I'm pretty new to Rust and SQL, so please take your time to review this.


RumovZ:
This would be such a great feature to have in vanilla Anki. Thank you for this!  
I'm not fit to review the overall PR, but if I can do anything to help (especially with the Rust part), let me know.  

One thing I've noticed is the use of regexp in sqlwriter to search for tags. I suppose the faster SQL comparison wasn't possible anymore. But this makes me wonder whether it's still ideal to save note tags as a single string? After all, they've become quite complex by now.

dae:
Thanks for your hard work on this abdo! It's going to take me a while to review it properly, but just some issues I noticed in initial testing:

- if you run scripts/runopt and import https://ankiweb.net/shared/info/1939287975, then run check database and check the browser list, there are some new, wierd names that look like a recursion issue, like "tag:#FirstAid::basic::biochem::metabolism::lipds::lysosomalstorage::sphingolipidmetabolismbasic::biochem::metabolism::lipds::lysosomalstorage::sphingolipidmetabolismbasic::biochem::metabolism::lipds::lysosomalstorage::sphingolipidmetabolismbasic::biochem::metabolism::lipds::lysosomalstorage::sphingolipidmetabolismbasic::biochem::metabolism::lipds::lysosomalstorage::sphingolipidmetabolismbasic::biochem::metabolism::lipds::lysosomalstorage::sphingolipidmetabolism"

- unlike decks, we can't assume parents will exist when upgrading - it's normal for users to have foo::bar::baz without foo::bar and maybe even foo. Missing decks seem to cause weird issues - tags go missing from the list, or end up under different parents until a DB check is done, eg

[collection-2021-01-07@12-37-32.colpkg.zip](https://github.com/ankitects/anki/files/5779214/collection-2021-01-07%4012-37-32.colpkg.zip)

Not sure if the two issues are related.

It may also be worth thinking about whether we want to enforce the existence of parents in the DB, or just fill them in at tree building time. I would lean towards the latter, because bulk note operations are potentially registering many tags in bulk, and we don't want to check for their parents each time.

@RumovZ that's a change we could potentially make in the future, but it will make maintaining compatibility with older clients more of a headache & slow down the upgrade/downgrade process, so I think it's best to keep the notes table untouched for now.

abdnh:
>  if you run scripts/runopt and import https://ankiweb.net/shared/info/1939287975, then run check database and check the browser list, there are some new, wierd names that look like a recursion issue

Funnily enough, this seems to be an actual tag in the deck 🙄 (I bet the deck author made a mistake.)
I imported the deck into Anki 2.1.38 and searched for the tag and got one match. I also double-checked by opening the collection.anki2 file of the deck package in SQLite browser and got the same result...



About the second issue:

> It may also be worth thinking about whether we want to enforce the existence of parents in the DB, or just fill them in at tree building time. I would lean towards the latter, because bulk note operations are potentially registering many tags in bulk, and we don't want to check for their parents each time.

In this PR, col.register_tag does actually check for parents but their existence is not properly enforced as your .colpkg shows.
The problem is in upgrade_tags_to_schema17 which uses storage.register_tag which doesn't check for parents. I think it's easy to fix it, but I see your point about the alternative approach so I'll look into that.

dae:
> Funnily enough, this seems to be an actual tag in the deck

Sorry, yeah, you're quite right. The initial issue I spotted was a tag being re-grandparented on upgrade due to a missing parent. This was happening in a private deck, so I went searching for a public example I could share, but appear to have just ended up finding bad data :-)

abdnh:
I just realized that filling missing parents at tree building time and not checking for them at registeration time has the drawback that we can't save the browser collapse state of these tags.

dae:
That's not a problem, the backend API that toggles a given node on or off can just make sure to create the node if it's not yet available

rslib/src/storage/tag/mod.rs
```diff
@@ -85,11 +139,49 @@ impl SqliteStorage {
...
+            .query_and_then(NO_PARAMS, |r| {
+                Ok(Tag {
+                    name: r.get(0)?,
+                    usn: r.get(1)?,
+                    ..Default::default()
+                })
+            })?
+            .collect::<Result<Vec<Tag>>>()?;
+        self.db.execute_batch(
+            "
```
dae, line 163: When the sql exceeds a single line, please move it into a separate .sql file, and include_str!() it. I've just added a formatter for .sql files, so we'll get consistent formatting that way, and syntax highlighting in code editors.

qt/aqt/browser.py
```diff
@@ -1132,15 +1142,44 @@ def _favTree(self, root) -> None:
...
+                    full_name = head + node.name  # pylint: disable=cell-var-from-loop
+
+                    def toggle(_):
+                        try:
+                            self.mw.col.tags.toggle_browser_collapse(full_name)
+                        except NotFoundError:
+                            # tag is missing, register it first
+                            self.mw.col.tags.register([full_name])
+                            self.mw.col.tags.toggle_browser_collapse(full_name)
+
```
dae, line 1165: Instead of the backend exposing get() and update() methods and requiring the frontend to check for the missing case, perhaps it would be nicer to just expose a set_tag_collapsed(tag: str, collapsed: bool) routine, where the backend takes care of fetching, creating if necessary, and setting the value? For collapsed we should be able to pass in 'not node.collapsed' to get toggling behaviour.

rslib/backend.proto
```diff
@@ -66,6 +66,10 @@ message DeckConfigID {
     int64 dcid = 1;
 }
 
+message TagID {
+    int64 tid = 1;
+}
+
```
dae, line 72: Perhaps can be removed?

dae (commented):
I've given it a quick look over, and it's looking good - just a few comments below for now; I should be able to give it another review within a few days.

rslib/backend.proto
abdnh: I was going to go ahead and remove tag IDs. I initially thought this will be useful for some reason.

dae: Please hold off on that for now, I'll give it some thought. If they don't end up getting dropped, we'll need to remember to add an index on the name column, as currently it'll be doing a table scan.

dae:
Ok, I've given it another review and some more thought. I didn't notice earlier that you were rewriting the notes table - I assumed that it was only in the tags table that you were storing the machine name. I'm concerned this is going to make the upgrade/downgrade process considerably slower on large collections - some users are bouncing back and forth between newer and older Anki versions and thus need to frequently upgrade/downgrade, and operations like full syncs and deck exports also need to upgrade/downgrade. Even just rebuilding the tags table can be slow on a large collection, and if we have to rewrite the notes as well, that's going to make things worse.

We may eventually want to update the structure of the notes and/or tags tables, but I'm wondering if we can delay doing that for now. Tags are thankfully a different case to decks. In the decks case, we have older Anki clients that expect all parents of a deck to exist, and they can throw errors when that's not the case. The use of \x1f as a separator takes advantage of the fact that it is a lower ascii value than any character users will use in their tags, which means when the deck names are sorted, they're guaranteed to arrive parents-first, even if the user has used a different uppercase/lowercase combination in the parent or child's name. But in the tags case, we are creating any missing parents as we go, so we shouldn't gain much from using a different separator.

What do you think about reworking this to skip the upgrade/downgrade, and make use of the current tags table instead?

abdnh:
>  I didn't notice earlier that you were rewriting the notes table

I should have written a bit more detailed description for such a change.

Thanks for clarifying the purpose of \x1f in detail. I've not given it much thought actually.
I had the idea of keeping tables as they are, but thought that since we need to update the tags table to store the browser collapse state, we might as well use \x1f as a separator and update the notes table. But I understand that this may not be worth an expensive schema upgrade.

I'll rework this later on (maybe in a new PR).

dae:
Good point about collapsing, we will still need an upgrade unless we go the hacky way of packing these into collection config. But like schema 14,  we should be able to read and write to the tags table and just ignore the notes table.

As an aside - the DB check drops the tags list and rebuilds it, so it means the collapsed state will be lost each time users do a DB check unless we modify it to preserve the state.

abdnh:
The collection config solution seems simpler to me if we want to avoid an upgrade now and in terms of modifying the DB check code.

dae:
I fear that replaces one problem with another - if tag info is stored in the collection config, it will either continue to take up space even when the associated tags are removed, or the DB check will need to be updated to clear out unused tag entries from the config.

Can you think of any other tag config info we might need to store in the future? If it's just the collapsed state, perhaps that's worth storing in a separate column of the tags table (eg collapsed=0/1) rather than in protobuf, as it makes it easier to query/update. If we also add 'config' as a nullable blob column and leave it null for now, that will allow us to pack new config in there in the future if necessary without another schema change, and for now it could just go unused.

abdnh:
> the DB check will need to be updated to clear out unused tag entries from the config.

I thought of doing this and this is what I came up with:

```diff
diff --git a/rslib/src/dbcheck.rs b/rslib/src/dbcheck.rs
index dc78e2bc2..5cc119440 100644
--- a/rslib/src/dbcheck.rs
+++ b/rslib/src/dbcheck.rs
@@ -294,6 +294,20 @@ impl Collection {
             }
         }

+        let tags_collapse = self.get_tags_collapse_config();
+        let tags = self.storage.all_tags()?;
+        let mut tags_collapse_new = HashSet::<String>::new();
+        for collapsed_tag in tags_collapse.into_iter() {
+            for tag in tags.iter() {
+                if tag.name.starts_with(&collapsed_tag) {
+                    tags_collapse_new.insert(collapsed_tag);
+                    break;
+                }
+            }
+        }
+        self.set_config("tagsCollapse", &tags_collapse_new)?;
+
```
tagsCollapse is a set storing collapsed tag names.
The backend method set_tag_collapsed can also be modified easily to work on this config (I've already done that).


> Can you think of any other tag config info we might need to store in the future?

I can't think of anything very useful (but maybe fancy colors or something?), but as the idea of the config column is more future-proof than the collection config hack, I prefer we go this way.
It seems preserving collapse state during a DB check is more involved in this case, but I'll look into this and into making separate collapse and config columns.

abdnh:
I pushed a commit addressing your suggestions.
The change to the DB check code is admittedly not ideal and there may be a better way to do it.


By the way, I think there is a bug in the tag bulk update code. I thought this was introduced by me, but I tested in ⁨2.1.38 and got the same result: e.g. for a note with the tags `*` and `foo`, deleting `*` will also wipe `foo`. Seems like a regex escaping issue.

The issue can be reproduced with this code:
```python
from aqt import mw

mw.col.tags.bulk_update(mw.col.find_notes("tag:\*"), "*", "", False)
```

qt/aqt/sidebar.py
```diff
@@ -111,16 +114,37 @@ def rename_deck(self, item: "aqt.browser.SidebarItem") -> None:
...
+    def remove_tag(self, item: "aqt.browser.SidebarItem") -> None:
+        self.browser.editor.saveNow(lambda: self._remove_tag(item))
+
+    def _remove_tag(self, item: "aqt.browser.SidebarItem") -> None:
+        old_name = item.full_name
+
+        def do_remove():
+            self.mw.col.backend.clear_tag(old_name)
+            self.col.tags.rename_tag(old_name, "")
+
```
dae, line 126: In the long run it might be nicer to have a backend method which does both these things in one go, as there's not much use in clearing a tag without altering any notes. But it will do for now :-)

rslib/src/backend/mod.rs
```diff
@@ -1341,14 +1349,26 @@ impl BackendService for Backend {
...
+                .collect())
+        })?;
         Ok(pb::AllTagsOut { tags })
     }
 
+    fn set_tag_collapsed(&self, input: pb::SetTagCollapsedIn) -> BackendResult<pb::Bool> {
+        self.with_col(|col| {
+            col.transact(None, |col| {
+                col.set_tag_collapsed(&input.name, input.collapsed)?;
+                Ok(pb::Bool { val: true })
```
dae, line 1367: If we don't have a return value:

```patch
diff --git a/rslib/backend.proto b/rslib/backend.proto
index 974e9d759..179e210c3 100644
--- a/rslib/backend.proto
+++ b/rslib/backend.proto
@@ -210,7 +210,7 @@ service BackendService {

   rpc RegisterTags(RegisterTagsIn) returns (Bool);
   rpc AllTags(Empty) returns (AllTagsOut);
-  rpc SetTagCollapsed(SetTagCollapsedIn) returns (Bool);
+  rpc SetTagCollapsed(SetTagCollapsedIn) returns (Empty);
   rpc ClearTag(String) returns (Bool);
   rpc TagTree(Empty) returns (TagTreeNode);

diff --git a/rslib/src/backend/mod.rs b/rslib/src/backend/mod.rs
index 170f978f9..1bf987fd1 100644
--- a/rslib/src/backend/mod.rs
+++ b/rslib/src/backend/mod.rs
@@ -1360,11 +1360,11 @@ impl BackendService for Backend {
         Ok(pb::AllTagsOut { tags })
     }

-    fn set_tag_collapsed(&self, input: pb::SetTagCollapsedIn) -> BackendResult<pb::Bool> {
+    fn set_tag_collapsed(&self, input: pb::SetTagCollapsedIn) -> BackendResult<pb::Empty> {
         self.with_col(|col| {
             col.transact(None, |col| {
                 col.set_tag_collapsed(&input.name, input.collapsed)?;
-                Ok(pb::Bool { val: true })
+                Ok(().into())
             })
         })
     }
```

rslib/src/dbcheck.rs
```diff
@@ -294,6 +295,18 @@ impl Collection {
...
+            for new in new_tags.iter() {
+                if new.name == old.name {
+                    self.storage.set_tag_collapsed(&new.name, new.collapsed)?;
+                    break;
+                } else if new.name.starts_with(&old.name) {
+                    self.set_tag_collapsed(&old.name, old.collapsed)?;
+                }
+            }
+        }
+
```
dae, line 309: This code would be better split up into a separate function implemented in tags.rs.

rslib/src/tags.rs
```diff
@@ -2,16 +2,80 @@
...
+    }
+}
+
+impl Eq for Tag {}
+
+impl Default for Tag {
+    fn default() -> Self {
+        Tag {
+            name: "".to_string(),
+            usn: Usn(-1),
```
dae, line 53: The server should never have a usn of -1, and it's a bit dangerous to be providing a default value. Is this default implementation just to make creating tags easier? If so, an alternative is to create a new() function that takes required functions and fills out other defaults, eg

```rust
impl Tag {
  fn new(name: String, usn: Usn) → Self {
    Tag { name: name.into(), usn, collapsed: false }
  }
}
```

The Into part is not required, it just makes it easier to pass in string

rslib/src/tags.rs
```diff
@@ -2,16 +2,80 @@
...
+pub struct Tag {
+    pub name: String,
+    pub usn: Usn,
+    pub collapsed: bool,
+}
+
+impl Ord for Tag {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.name.cmp(&other.name)
+    }
```
dae, line 32: Because the DB stores items in a unicase collation, the tags will arrive in unicode sorted order. But here you're using String::cmp to order, so sorting a vec of tags will change the unicode sort order to simple ascii order. It looks like the missing tag function does a simple sort, so the unicase order would be lost. To prevent that, you'd need to create temporary UniCase objects like in canonify_tags().

rslib/src/tags.rs
```diff
@@ -67,20 +221,26 @@ impl Collection {
...
-            .collect();
+        let tags: Vec<_> = tags.into_iter().map(|s| s.into_inner()).collect();
 
         Ok((tags, added))
     }
 
-    pub(crate) fn register_tag<'a>(&self, tag: &'a str, usn: Usn) -> Result<Cow<'a, str>> {
-        if let Some(preferred) = self.storage.preferred_tag_case(tag)? {
-            Ok(preferred.into())
+    pub(crate) fn register_tag(&self, tag: Tag) -> Result<(Tag, bool)> {
```
dae, line 229: Please document the return value

dae (commented):
Looking good Abdo, I don't think this is too far off. Thanks for all your work on this; some more comments below.

dae:
@abdnh re the bulk_update option, "*" will replace any tag - perhaps you wanted "\*"?

abdnh:
> 
> 
> @abdnh re the bulk_update option, `"*"` will replace any tag - perhaps you wanted `"\*"`?

I had the wrong assumption that tags should be treated literally when the regex flag is off (even when this is not actually regex).

abdnh:
Is it on purpose that [rename_tag](https://github.com/ankitects/anki/blob/5f02be4943f0834cdf4302d72ae4448483c6661a/pylib/anki/tags.py#L103) passes the unescaped old name to bulk_update?

dae:
Looks like a bug; when renaming tags from the tags list we should be treating the content literally.

abdnh:
I pushed some changes.

But there is the issue of renaming a tag to use a different letter case when it has siblings or parents in the DB.

For two notes `a` and `b` with the tags `foo::a` and `foo::b`.
image


Rename `foo::b` to `FOO:b` to get this:
image

With this in the DB:
```python
>>> from aqt import mw
... print(mw.col.tags.all())
['foo::a', 'FOO::b']
```

abdnh:
A potential fix may be to lowercase tag names in the sidebar tree, but use the actual names when renaming to allow changing letter case.

dae:
We should be able to avoid that by using UniCase - will look into it and get back to you.

pylib/anki/tags.py
```diff
@@ -103,11 +87,11 @@ def bulk_update(
     def rename_tag(self, old: str, new: str) -> int:
         "Rename provided tag, returning number of changed notes."
         escaped_name = re.sub(r"[*_\\]", r"\\\g<0>", old)
-        escaped_name = '"{}"'.format(escaped_name.replace('"', '\\"'))
-        nids = self.col.find_notes("tag:" + escaped_name)
+        quote_escaped = escaped_name.replace('"', '\\"')
+        nids = self.col.find_notes(f'tag:"{quote_escaped}"')
```
dae, line 91: Thanks to Rumo's recent change, we can now skip the manual escaping:

```
search = self.col.backend.filter_to_search(FilterToSearchIn(tag=old))
nids = self.col.find_notes(search)
```

qt/aqt/browser.py
```diff
@@ -1156,15 +1161,35 @@ def _favTree(self, root) -> None:
...
-            )
-            root.addChild(item)
+        tree = self.col.backend.tag_tree()
+
+        def fillGroups(root, nodes: Sequence[TagTreeNode], head=""):
+            for node in nodes:
+
+                def set_filter():
+                    full_name = head + node.name  # pylint: disable=cell-var-from-loop
+                    return lambda: self.setFilter("tag", full_name)
```
dae, line 1171: Looks unused?

dae (commented):
I've pushed a change which should fix the parents issue, please let me know if you spot any problems with it. Looks like we're almost there!

pylib/anki/tags.py
abdnh: Fantastic!
Though we also need to escape the name manually for bulk_update.

abdnh:
Looks great!

There is an issue with write_tag not taking account of case folding because of my initial change.
Bringing preferred_tag_case back should fix this in the non-glob case as it used to be.

I'll push a fix later on.

Thanks @RumovZ for your related note!

pylib/anki/tags.py
dae: Ah, that's a shame. Maybe we can change bulk_update to an enum in the future and allow passing in a literal text. Not something we need to do for this PR though.

dae:
Is this for tag renaming? I suspect simple unicode folding will be sufficient for the vast majority of users, so it probably makes sense to keep things simple. If it proves to be a problem in the future, we can revisit the issue.

abdnh:
Yes, this is a problem for tag renaming and only in complex and rare cases (e.g. Maße -> MASS). So, not a big issue.
if unicase provided access to the casefolded form, maybe this can solved by casefolding before the regex search (through a sql function or something).

dae:
Happy for a final review + merge at this point, or was there anything you wanted to do first?

abdnh:
I gave this one more test, found a little issue, and fixed it.

Nothing to add other than that.

dae:
Thanks again for your work on this Abdo!

One minor issue I noticed that we might want to fix in the future - preferred_tag_case could be extended to consider existing parents, like the way the decks code will reuse an existing parent case. Currently if you add 'Foo' and then 'foo::bar', it uses a lowercase foo instead of Foo. This doesn't break the tree list, it's just a bit inconsistent.

dae:
(match_or_create_parents() is the relevant code in decks/mod.rs)

PR #782, opened 2020-10-07
fix outdated help site

johan456789:

PR #901, opened 2021-01-06
replace broken help link

wallgrenen:
The previous link gave a 404 error.

I chose to use the same link as the one in helpRequested() in qt/aqt/addcards.py that was chosen in  #782


dae:
Thanks!

PR #903, opened 2021-01-07
Writer cleanup

RumovZ:
I've tried to implement your notes from #860 but am not quite sure I got it right. Please let me know if anything's still off.
Concerning rsbackend.py: Did you mean there should be a wrapper for `self.col.backend.negate_search` etc. or was it only about the pb enum? Since we don't use `ConcatenateSearchesIn` explicitly, I didn't see the need to import it.

### DIFF
--- a/pylib/anki/rsbackend.py
+++ b/pylib/anki/rsbackend.py
@@ -46,6 +46,7 @@ TagUsnTuple = pb.TagUsnTuple
 NoteType = pb.NoteType
 DeckTreeNode = pb.DeckTreeNode
 StockNoteType = pb.StockNoteType
+ConcatSeparator = pb.ConcatenateSearchesIn.Separator
 SyncAuth = pb.SyncAuth
 SyncOutput = pb.SyncCollectionOut
 SyncStatus = pb.SyncStatusOut
--- a/pylib/rsbridge/lib.rs
+++ b/pylib/rsbridge/lib.rs
@@ -53,6 +53,10 @@ fn want_release_gil(method: u32) -> bool {
                 | BackendMethod::LatestProgress
                 | BackendMethod::SetWantsAbort
                 | BackendMethod::I18nResources
+                | BackendMethod::NormalizeSearch
+                | BackendMethod::NegateSearch
+                | BackendMethod::ConcatenateSearches
+                | BackendMethod::ReplaceSearchTerm
         )
     } else {
         false
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -21,7 +21,7 @@ from anki.consts import *
 from anki.lang import without_unicode_isolation
 from anki.models import NoteType
 from anki.notes import Note
-from anki.rsbackend import DeckTreeNode, InvalidInput, pb
+from anki.rsbackend import DeckTreeNode, InvalidInput, ConcatSeparator
 from anki.stats import CardStats
 from anki.utils import htmlToTextLine, ids2str, isMac, isWin
 from aqt import AnkiQt, gui_hooks
@@ -1236,13 +1236,13 @@ QTableView {{ gridline-color: {grid} }}
                 elif mods & Qt.ControlModifier:
                     txt = self.col.backend.concatenate_searches(
                         # pylint: disable=no-member
-                        sep=pb.ConcatenateSearchesIn.Separator.AND,
+                        sep=ConcatSeparator.AND,
                         searches=[cur, txt],
                     )
                 elif mods & Qt.ShiftModifier:
                     txt = self.col.backend.concatenate_searches(
                         # pylint: disable=no-member
-                        sep=pb.ConcatenateSearchesIn.Separator.OR,
+                        sep=ConcatSeparator.OR,
                         searches=[cur, txt],
                     )
         except InvalidInput as e:
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -21,7 +21,7 @@ from anki.consts import *
 from anki.lang import without_unicode_isolation
 from anki.models import NoteType
 from anki.notes import Note
-from anki.rsbackend import DeckTreeNode, InvalidInput, ConcatSeparator
+from anki.rsbackend import ConcatSeparator, DeckTreeNode, InvalidInput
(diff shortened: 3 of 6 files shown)


RumovZ:
Weird. `.\bazel run //qt:format` wouldn't order browser.py's imports, so I had to do it manually.

RumovZ:
I've found and fixed an issue with saved filters. Since they get normalised when clicked on in the browser, the GUI wouldn't recognise them making it very hard to delete them.
Now, we can even recognise unequal but equivalent searches. I also thought it would be a good idea to reject saving invalid searches.

dae:
Great, thanks Rumo.

dae:
Sorry, sort_kind_from_pb was not a great example - I've changed it and the bool separator to accept the enum directly instead, and relied on the default value:

https://github.com/ankitects/anki/commit/1a5b5f09cc43fa08ba596a9231d54432c6426087
https://github.com/ankitects/anki/commit/cbeb714062082ac149ca6af111d50b796ffc3658

The format worked for me by the way:
```
INFO: From Testing //qt:format_check:
==================== Test output for //qt:format_check:
All done! ✨ 🍰 ✨
62 files would be left unchanged.
--- /Users/dae/Work/code/dtop/qt/aqt/browser.py:before	2021-01-07 23:10:35.731304
+++ /Users/dae/Work/code/dtop/qt/aqt/browser.py:after	2021-01-07 23:10:43.738662
@@ -21,7 +21,7 @@
 from anki.lang import without_unicode_isolation
 from anki.models import NoteType
 from anki.notes import Note
-from anki.rsbackend import DeckTreeNode, InvalidInput, ConcatSeparator
+from anki.rsbackend import ConcatSeparator, DeckTreeNode, InvalidInput
 from anki.stats import CardStats
 from anki.utils import htmlToTextLine, ids2str, isMac, isWin
 from aqt import AnkiQt, gui_hooks
Skipped 3 files
ERROR: /Users/dae/Work/code/dtop/qt/aqt/browser.py Imports are incorrectly sorted and/or formatted.
================================================================================
INFO: Elapsed time: 57.781s, Critical Path: 56.89s
INFO: 60 processes: 51 remote cache hit, 10 darwin-sandbox, 2 worker.
INFO: Build completed, 1 test FAILED, 60 total actions
//qt/aqt/data/web/js:format_check                               (cached) PASSED in 0.0s
//qt:mypy                                                                PASSED in 15.2s
//qt:pylint                                                              PASSED in 30.7s
//qt:pytest                                                              PASSED in 3.2s
//qt:format_check                                                        FAILED in 3.6s
  /private/var/tmp/_bazel_dae/4928556feb6f3e9c4d256de8caa46cdd/execroot/net_ankiweb_anki/bazel-out/darwin-fastbuild/testlogs/qt/format_check/test.log

Executed 4 out of 5 tests: 4 tests pass and 1 fails locally.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line op
INFO: Build completed, 1 test FAILED, 60 total actions
zsh: exit 3     bazel test qt/...
dae@dip:Work/code/dtop% bazel run qt:format
INFO: Invocation ID: f5067fd0-d803-4306-a9ca-f2a5c393d879
INFO: Analyzed target //qt:format (0 packages loaded, 1 target configured).
INFO: Found 1 target...
Target //qt:format up-to-date:
  bazel-bin/qt/format
INFO: Elapsed time: 0.453s, Critical Path: 0.03s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
All done! ✨ 🍰 ✨
61 files left unchanged.
Fixing /Users/dae/Work/code/dtop/qt/aqt/browser.py
Skipped 3 files
dae@dip:Work/code/dtop% bazel test qt/...
INFO: Invocation ID: 1f42444f-482a-4f7a-a620-15d199863bd0
INFO: Analyzed 92 targets (2 packages loaded, 118 targets configured).
INFO: Found 87 targets and 5 test targets...
INFO: Elapsed time: 25.838s, Critical Path: 25.49s
INFO: 5 processes: 2 remote cache hit, 6 darwin-sandbox.
INFO: Build completed successfully, 5 total actions
//qt/aqt/data/web/js:format_check                               (cached) PASSED in 0.0s
//qt:format_check                                                        PASSED in 3.4s
//qt:mypy                                                                PASSED in 12.3s
//qt:pylint                                                              PASSED in 24.2s
//qt:pytest                                                              PASSED in 3.2s
```

dae:
Turns out you can call an enum field's getter instead of having to deal with from_i32(): https://github.com/ankitects/anki/commit/10630ab72361967e9d1c26bb94d9328e3b0f1246

RumovZ:
> Turns out you can call an enum field's getter instead of having to deal with from_i32(): [10630ab](https://github.com/ankitects/anki/commit/10630ab72361967e9d1c26bb94d9328e3b0f1246)

Ah, so that's how it's done. I was starting to wonder what the point of a protobuf enum was in the first place. 
Thanks for sharing.

I was considering whether the backend shouldn't provide functions for writing any of the filters used by the GUI. Then we could strictly confine the definition of the search syntax to the backend's search module and could especially ditch the wacky code in `setFilter()` (see below).
Would that make sense or would it introduce more complexity than would be worth it?

```
def setFilter(self, *args):
        if len(args) == 1:
            txt = args[0]
        else:
            txt = ""
            items = []
            for i, a in enumerate(args):
                if i % 2 == 0:
                    txt += a + ":"
                else:
                    txt += re.sub(r'["*_\\]', r"\\\g<0>", a)
                    txt = '"{}"'.format(txt.replace('"', '\\"'))
                    items.append(txt)
                    txt = ""
            txt = " AND ".join(items)
        ...
```

dae:
I think a function that made it easier to construct or add a search clause without having to do any regexs/quoting would be good. I experimented this morning with exposing json (de)serializing routines, but I'm not sure they're worth the extra complexity. Currently some validation occurs at the parsing stage (such as checking the position of AND/OR), and that would need to be reworked to happen later if we were accepting input from json. Perhaps we'd be better off with a more focused routine that just accepts a string and optional qualifier, and returns the relevant quoted string, which could then be passed to concatenate_searches?

PR #904, opened 2021-01-07
Streamlining `is:due`, `prop:due`, and the FutureDue graph

hgiesel:
These are all issues that rose up, when I was working on #897, specifically on the `FutureDue` graph.

1. This one is definitely a bug:
The Histogram graph for the stats had a one-off error, when displaying past ranges. The easiest way is to look at the `FutureDue` graph with Backlog enabled.
The first "future" label reads: "In 0-4 days".
The first past label reads: "0-4 days ago". This should actually be "1-5" days ago.

1. This rest of the changes streamline how `is:due`, `prop:due`, and the `FutureDue` graph deal with due dates:
* The FutureDue showed learning and preview repeat cards, which were scheduled for the past as being scheduled today (In fact it displayed all learning cards as scheduled today). Now it correctly displays them as backlog.
* The `is:due` selector did not consider PreviewRepeat cards. Now it does.
* The `prop:due` selector did not search for Learning or PreviewRepeat cards. Now it does. This requires having SQL do some math. Something I stumbled upon is that in SQL, dividing two integer values automatically does integer division.

Now these three features allow for a few sanity tests:
1. `-prop:due<=0 is:due` should only ever show learning / preview repeat cards, which will be due some time later today, but are not yet.
2. When the FutureDue graph shows: "In x-y days: n cards due", searching for `prop:due>=x prop:due<=y` should yield n cards. Additionally, when it shows "x-y days ago: n cards due", searching for `prop:due<=-x prop:due>=-y` should yield n cards.

### DIFF
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -235,16 +235,24 @@ impl SqlWriter<'_> {
 
     fn write_prop(&mut self, op: &str, kind: &PropertyKind) -> Result<()> {
         let timing = self.col.timing_today()?;
+
         match kind {
             PropertyKind::Due(days) => {
                 let day = days + (timing.days_elapsed as i32);
                 write!(
                     self.sql,
-                    "(c.queue in ({rev},{daylrn}) and due {op} {day})",
+                    "(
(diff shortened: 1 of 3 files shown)


dae:
Looks good, thanks!

PR #906, opened 2021-01-07
Omit zero interval, and don't nice the values in Review intervals graph

hgiesel:
This is a follow up to #905, but independent of it.

1. The graph starts at cards with 0 day intervals atm, which are cards in the new or learning cards, however those are by the definition of the graph never going to be used for filling in the graph.

2. Right now, the interval days are "niced", which causes 34 days to be shown for a month.
The call to `.nice()` could technically be kept for percentile options, however I don't notice a great improvement in display there, because the review interval tends to converge to zero anyway.

After this PR, the interval graph starts at 1, and is capped at 31 days, if you click on 1 month.

### DIFF
--- a/ts/graphs/card-counts.ts
+++ b/ts/graphs/card-counts.ts
@@ -197,7 +197,7 @@ export function renderCards(
 
     x.range([bounds.marginLeft, bounds.width - bounds.marginRight]);
 
-    const tableData = (data as any).flatMap((d: SummedDatum, idx: number) => {
+    const tableData = data.flatMap((d: SummedDatum, idx: number) => {
         const percent = ((d.count / xMax) * 100).toFixed(1);
         return d.show
             ? ({
--- a/ts/graphs/intervals.ts
+++ b/ts/graphs/intervals.ts
@@ -108,9 +108,7 @@ export function prepareIntervalData(
 
     const bins = histogram()
         .domain(scale.domain() as [number, number])
-        .thresholds((scale.ticks(desiredBars) as any).flatMap(adjustTicks))(
-        allIntervals
-    );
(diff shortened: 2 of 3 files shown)


dae:
Without the nicing, it can cause a bar to be truncated - note the rightmost one here for example:
Screen Shot 2021-01-08 at 4 20 41 pm

dae:
.nice() will also try to pick a y axis that looks good - without it I seem to recall the y axis max and associated ticks could be suboptimal sometimes, so if you work around the x axis issue, it may still be worth nicing the y axis

hgiesel:
I fixed the truncation issue.

The x axis is now only `.nice()`d when you're not using the month option. It even makes sense, because `nice` is supposed to be used when we're dealing with uncertain domain values, however with months, we always want to display exactly 31 bins.

Regarding the y-axis: [I think that already happens](https://github.com/ankitects/anki/blob/fe4da25e15d3a07e7c1dc266c044cf291dc95fc8/ts/graphs/histogram-graph.ts#L67)

In order for the 1-x range to display, I have to manually adjust the output values of `.nice()`.
I increment the domain values, and I increment + append a new value to the ticks, e.g.:

```
adjustedDomain = [1, 31]
// automatically generated ticks will now look like this:
generatedTicks = [10, 20, 30]
// I modify it like this:
newTicks = [1, 11, 21, 31]
// same when it's in the hundreds:
domain = [1, 401]
generatedTicks = [100, 200, 300, 400]
newTicks = [1, 101, 201, 301, 401]
```

I also included `es2019.array` in the typescript libs, so we don't have to cast arrays to `any` before being able to use `.flatten()`, or `.flatMap()`.

dae:
Great, thanks for that - the improved typing is an extra bonus!

PR #910, opened 2021-01-09
Exclude aqt/hooks_gen.py from formatting

abdnh:
This was causing qt:format to fail.

### DIFF
--- a/qt/tests/run_format.py
+++ b/qt/tests/run_format.py
@@ -22,7 +22,7 @@ if __name__ == "__main__":
             "black",
             "-t",
             "py38",
-            "--exclude=aqt/forms|colors",
+            "--exclude=aqt/forms|colors|_gen",
             "aqt",
             "tests",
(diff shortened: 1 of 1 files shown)


dae:
Hmm, seems to work here - what was the error you were getting out of curiosity? Excluding the file shouldn't hurt in any case, thanks.

abdnh:
I got this error (on Windows 10 if that matters):
```
error: cannot format C:\anki\qt\aqt\hooks_gen.py: Cannot parse: 1:2: ../../bazel-bin/qt/aqt/hooks_gen.py
```

pylib:format failed in the same way for anki/hooks_gen.py when I removed _gen from pylib's run_format.py.

dae:
Ah, I suspect I'm not hitting this as I'm rsyncing files over to a Windows machine from a non-Windows machine rather than checking out from git. Thanks for the fix :-)

PR #911, opened 2021-01-09
Backend filters

RumovZ:
https://github.com/ankitects/anki/pull/903#issuecomment-757053009
> I think a function that made it easier to construct or add a search clause without having to do any regexs/quoting would be good. I experimented this morning with exposing json (de)serializing routines, but I'm not sure they're worth the extra complexity. Currently some validation occurs at the parsing stage (such as checking the position of AND/OR), and that would need to be reworked to happen later if we were accepting input from json. Perhaps we'd be better off with a more focused routine that just accepts a string and optional qualifier, and returns the relevant quoted string, which could then be passed to concatenate_searches?

I came to a very similar conclusion. In this PR, the frontend can get filters from the backend which take either one (like deck, tag etc.) or zero (fixed filters stored in an enum) arguments. This should be easy enough to extend if the GUI needs other filters in the future.
Custom filters are still saved and passed as search strings. I think that's the only way to keep them compatible and simple.

I marked this as a draft because I think the frontend part could still be prettified a bit.
The backend part should be rather finished, but again I couldn't get it to work with the pb enums and had to resort to i32s and defaults. Is there really no manual for protobuf? 😕


rslib/backend.proto
```diff
@@ -756,6 +757,35 @@ message BuiltinSearchOrder {
...
+        NO_FLAG = 15;
+        ANY_FLAG = 16;
+    }
+    oneof filter {
+        NamedFilter name = 1;
+        string tag = 2;
+        string deck = 3;
+        string note = 4;
+        uint32 template = 5;
+    }
```
dae, line 786: It took me a while to figure out how _filterFunc() was using this - the use of kwargs obscures what is valid input, and prevents us from knowing when we've passed in some incorrect value. Instead of extending setFilter() to expect kwargs, what about defining various helper functions that return a string, and using its existing args interface which just takes a list of strings? eg~

```
    def _named_filter(self, filter: NamedFilter) -> str:
        return self.col.backend.filter_to_search(FilterToSearchIn(name=filter))

    def _deck_filter(self, deck: str) -> str:
        return self.col.backend.filter_to_search(FilterToSearchIn(deck=deck))

    def _commonFilters(self):
        return self._simpleFilters(
            (
                (tr(TR.BROWSING_WHOLE_COLLECTION), self._named_filter(NamedFilter.WHOLE_COLLECTION),
```

But to be honest, what I originally had in mind was just keeping the existing strings, and simply having a function in the backend like:

make_string_term(key="deck", val='deck_with_"quotes') → '"deck:deck_with_\"quotes'
make_string_term(key=None, val="is:new") → "is:new"
make_string_term(key="is", val="new") → "is:new"

The former approach buys us some extra static typing, but is more work to implement. Either would be fine :-)

dae:
You can explore available methods by running the following in the rslib folder (assuming you have rustup/rust installed):

cargo doc --document-private-items --open

It looks like there is no accessor here, presumably because the enum is wrapped in a oneof - I suspect there would be one if the field was stored directly in the message instead. Protobuf and prost can be a bit awkward at times :-(

rslib/backend.proto
dae: Hmm, with the exception of _named_filter(), perhaps we don't need the other helpers and could just call the code directly in the calling function, eg _deckFilters() would call the filter_to_search routine itself

RumovZ: With the second approach, I don't see much improvement to the current situation, so I'll go with the first one.

Of course, you're right about the kwargs. I only wonder, wouldn't it make more sense to implement these functions directly in rsbackend.py? Or do you want to keep that module concise, whereas all hope is already lost for browser.py? 😅

Also, I think it's neat to have `_commonFilters()` etc. only pass a name like `NamedFilter.WHOLE_COLLECTION` to `_simpleFilters()`, so we only have to call `_named_filter()` once there. (Maybe we can even rename `_simpleFilters()` to `_namedFilters()` since it coincides so well with `NamedFilter`.)

RumovZ: > Hmm, with the exception of _named_filter(), perhaps we don't need the other helpers and could just call the code directly in the calling function, eg _deckFilters() would call the filter_to_search routine itself

Since we seem to have to call `filter_to_search` with `pb.FilterToSearchIn`, we can better just use a wrapper like `backend._deck_filter(deck: str)` instead of importing `pb.FilterToSearchIn` through rsbackend and then calling `backend.filter_to_search(FilterToSearchIN(deck="..."))` in my opinion.

RumovZ:
Thanks, at least that should be better than what I was doing before.

RumovZ:
I ended up pretty much implementing your suggestion except that I cut out `_filtFunc()`.
By the way, the `rslib:proto_format` test is failing on my machine and `.\bazel run //rslib:proto_format` is doing weird stuff like adding a blank line after every line.

dae:
Thanks Rumo, much easier to follow now. I might follow up with a tweak that makes the enum matching in Rust a bit more verbose for readability's sake, and see if the Any typing can be made clearer. Protobuf formatting issue should be fixed now.

rslib/backend.proto
dae: Re the appropriate file, rsbackend.py (and _gen.py) expose the various backend methods largely as is, and it's the job of the rest of pylib to wrap them and provide a more Pythonic API where necessary. Code in the GUI calling methods on .backend directly is a bit of a code smell - ideally every function would have a wrapper in pylib that we'd call instead. I haven't bothered with that in cases where the code is not intended to be user facing - which is partly laziness, and partly because Python doesn't let us enforce public/private methods - so even if we went to the trouble of wrapping backend methods that are not intended for public consumption, we can't stop users from ignoring the methods and calling the backend directly, making it feel a bit pointless.

RumovZ: Thank you for taking the time to elaborate. It's good to know about this consideration.

PR #912, opened 2021-01-09
Add explicit errors when entering added/edited/rated:0 or rated:>365

hgiesel:
This will throw errors when users search for e.g. `added:0`. By definition `added:1` searches for cards added today, so `added:0` is a nonsensical query. Same for `edited`, and `rated`.
Especially because other queries like `prop:due=0` carry the meaning of 0 = today, it would be nice, if a user typed `rated:0`, expecting cards reviewed today and doesn't see what he expected, he would instead get an error message, making it less obscure. 

Additionally it will throw an error if users search `rated:366`. Right now the value is silently coerced to 365 later down the line, which is not obvious to the user. This would also allow for making the query creation function more "pure" (follow-up PR).

Custom error messages would be very nice here, telling the user what exactly he did wrong, and I'd be happy for any pointers on that.

### DIFF
--- a/pylib/tests/test_find.py
+++ b/pylib/tests/test_find.py
@@ -204,13 +204,21 @@ def test_findCards():
         assert len(col.findCards("rated:1:1")) == 1
         assert len(col.findCards("rated:1:2")) == 1
         assert len(col.findCards("rated:1")) == 2
-        assert len(col.findCards("rated:0:2")) == 0
         assert len(col.findCards("rated:2:2")) == 1
+        with pytest.raises(Exception):
+            col.findCards("rated:0")
+
+        with pytest.raises(Exception):
+            col.findCards("rated:366")
         # added
-        assert len(col.findCards("added:0")) == 0
         col.db.execute("update cards set id = id - 86400*1000 where id = ?", id)
         assert len(col.findCards("added:1")) == col.cardCount() - 1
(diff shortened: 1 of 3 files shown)


RumovZ:
> Custom error messages would be very nice here, telling the user what exactly he did wrong, and I'd be happy for any pointers on that.

I thought so, too, but I found that all the error information is being lost when the higher-level parsers call the lower-level parsers through `map_res`.
Backtracing would require a different approach with a consistent error type, but I think that would require an extensive revision of parser.rs. I planned to look into this deeper since I've reworked the parser some months ago.

hgiesel:
@RumovZ Thanks for the heads up! I'm still very new to Rust, so it would have probably taken me a long time to figure that out.

RumovZ:
Just now when writing #913, I realised that there may be a better way than throwing errors: Starting with the next release, the user's search input will be normalised, see #860. So if the parser implicitly shifts input values into a valid range, the user will be able to see the actual interpretation of his search in the search bar. E.g. `rated:500` will appear as `rated:365`.

This actually already happens for `card:0`. On any past release, this will show the results for `card:1` instead. On master as well, but there the user input is also modified to reflect the results.

hgiesel:
@RumovZ Actually, that sounds like the better idea to me, too. I will probably open a new PR for it.

I already saw the new auto-normalisation functionality, while writing my past PRs, and it works really well.

PR #914, opened 2021-01-10
Preview Button in Editor instead of Browser

hgiesel:
Trying out the new auto-normalize functionality in the Browser, I noticed that the search queries now take up more horizontal space (especially because of the `AND`s). Here's an example:

```
// A common search of mine:
(changes:_* or tag:leech or tag:marked or (is:suspended -tag:ok)) tag:* deck:*jp
// After normalization:
("changes:_*" OR "tag:leech" OR "tag:marked" OR ("is:suspended" AND -"tag:ok")) AND "tag:*" AND "deck:*jp"
```

This PR tries to remedy this a little bit, by moving the Preview button from the browser to the Editor in the Browser window, so it turns from a QT button to an HTML button.

I also fixed some minor issues along the way:
* Don't show the focus ring around buttons in the Editor. They were usually triggered on the left top buttons, and just looked ugly.
* Make `Editor.addButton` usable if you want both `toggleable=True` and have a keys shortcut. The shortcut invocation will now trigger a toggle action (for which it needs a transient id).
* Make the Previewer window closable with "Ctrl+Shift+P", because that's also how you open it, and can close if the browser is the active window.

A change in behavior is that now the Preview window closes, if the search yielded no result, as otherwise, the user would not have access to the preview button anymore.

After looking at it for a while, I think even thematically the button makes more sense in the Editor, so that the top row is now fully occupied with items regarding "Search", and not things regarding individual cards.

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -868,18 +868,24 @@ QTableView {{ gridline-color: {grid} }}
     def setupEditor(self):
         def add_preview_button(leftbuttons, editor):
             preview_shortcut = "Ctrl+Shift+P"
-            leftbuttons.insert(0, editor.addButton(
-                None,
-                "preview",
-                lambda _editor: self.onTogglePreview(),
-                tr(TR.BROWSING_PREVIEW_SELECTED_CARD, val=preview_shortcut),
-                tr(TR.ACTIONS_PREVIEW),
-                id="previewButton",
-                keys=preview_shortcut,
-                disables=False,
-                rightside=False,
-                toggleable=True,
-            ))
+            leftbuttons.insert(
+                0,
+                editor.addButton(
+                    None,
+                    "preview",
+                    lambda _editor: self.onTogglePreview(),
+                    tr(
+                        TR.BROWSING_PREVIEW_SELECTED_CARD,
+                        val=shortcut(preview_shortcut),
+                    ),
+                    tr(TR.ACTIONS_PREVIEW),
+                    id="previewButton",
+                    keys=preview_shortcut,
+                    disables=False,
+                    rightside=False,
+                    toggleable=True,
+                ),
+            )
 
         gui_hooks.editor_did_init_left_buttons.append(add_preview_button)
         self.editor = aqt.editor.Editor(self.mw, self.form.fieldsArea, self)
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -867,13 +867,15 @@ QTableView {{ gridline-color: {grid} }}
 
     def setupEditor(self):
         def add_preview_button(leftbuttons, editor):
+            preview_shortcut = "Ctrl+Shift+P"
             leftbuttons.insert(0, editor.addButton(
                 None,
                 "preview",
                 lambda _editor: self.onTogglePreview(),
-                tr(TR.BROWSING_PREVIEW_SELECTED_CARD, val=shortcut("Ctrl+Shift+P")),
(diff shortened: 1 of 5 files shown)


kleinerpirat:
Since a preview within the Add-mode is a regularly requested feature, I'd like to know if there is anything that speaks against providing a stripped down version of the previewer from `clayout.py` for the Add-mode.

Tried this out the other day and it seems practical (e.g. when dealing with notes with multiple cards):
<hr>

image

<hr>

Some of the requests:
> - [Add-on to preview while adding?](https://www.reddit.com/r/Anki/comments/6z3kua/addon_to_preview_while_adding/)
> - [How to Preview cards while Adding them rather than going in to the browser to preview them .](https://www.reddit.com/r/Anki/comments/9j5g4i/how_to_preview_cards_while_adding_them_rather/)
> - [Is there any way to have a live preview whilst adding cards like the one offered when browsing?](https://www.reddit.com/r/Anki/comments/gyxolc/is_there_any_way_to_have_a_live_preview_whilst/)
> - [Addon Request: Preview while creating cards.](https://www.reddit.com/r/Anki/comments/fguokb/addon_request_preview_while_creating_cards/)

hgiesel:
@kleinerpirat Actually, while writing this PR I also had this on my mind. Maybe taking this functionality of `clayout.py`, modularizing it, and combining it with a new subclass of the `Previewer` class from `preview.py`, could be turned into something like a Previewer for the Add Cards dialog. I'm certainly gonna have a look at it after this PR.

dae:
Thanks Henrik, LGTM. No objections to extending this to the add screen in terms of functionality, though it's a bit more involved. Are you thinking a separate preview window for the add screen and browse screen?

hgiesel:
@dae I only just now realised that `qt/aqt/forms/preview.ui` is not used for the Browser preview, but just for the Card layout...

What I'd like to do is rewrite `qt/aqt/preview.py`, and turn it to the basis to the basis for all three previewers (clayout, browser, add cards). This would start with a base class for "`QWidget` that displays card, can flip between front/back, and has the options combo box, for:
1. Fill Empty fields checkbox (from current clayout)
2. Night Mode checkbox (from current clayout)
3. Add mobile class checkbox (from current clayout)
4. Replay audio (from current browser previewer)

This Previewer will be subclassed for an `EphemeralPreviewer`, which instead of using an existing card, will create one out of thin for the sake of previewing, which would be used by the addcards, and clayout.

I will also add a gui_hook to the previewer for initializing buttons, which each user of the class can utilize to implement the remaining functionality. Those buttons will be inserted left of the Options combo box, to generate something like this:

previewers

This would require a complete rewrite of `preview.py`, and ripping out a lot of functionality of `clayout.py` for the new `preview.py`. What do you think?

dae:
If this is going to require non-trivial changes, before investing time in it, I think it's worth looking into whether it would be practical to move the surrounding GUI elements into a TypeScript/Svelte page, with the card content in an inner iframe, as that's a direction I think Anki needs to be moving in. Qt's Chromium port seems to choke on every second video driver on Windows, and they recently dropped LTS releases for open source users, and don't have a new release with WebEngine support planned until something like September. The LTS changes were likely motivated by money issues, and it makes me worry about their long term viability.

While we likely won't be able to move away from Qt any time soon, there may come a day that we need/want to, and if at that point the bulk of the UI code is implemented in TS/HTML, we'll have options like switching to Electron or a native webview.

Aside from Qt's future, I think in the long term it will just make things easier as well - driving the UI with Python and having to send asynchronous messages back and forth to the various webviews is needlessly complicated. :-)

PR #915, opened 2021-01-10
Fix _template_filter and update want_release_gil()

RumovZ:
Don't add 1 when calling _template_filter() (see #913) and add FilterToSearch to want_release_gil().


dae:
Thanks!

PR #917, opened 2021-01-10
Coerce added/edited:0 to 1, Constrain rated:n to 1-365

hgiesel:
This is a revision #912, with coercion rather than throwing explicit errors.


dae:
Looks good, thanks!

PR #918, opened 2021-01-11
Introduce `resched:n` instead of `rated:n:0` 

hgiesel:
Currently, searching for `rated:1` will give you cards studied today, OR rescheduled today.
However the Anki manual and Anki itself tries to skim over this, and simply says "Studied Today", which however would technically be `rated:1:1 or rated:1:2 or rated:1:3 or rated:1:4`, but does make sense, because this is what the user probably wants to search for. I'd say this is an implementation detail leaking to the user.

That's what this PR is trying to solve: `rated:n` will now be restricted to revlog entries with `ease between 1 and 4`. The old  behavior of `rated:n` can now be recreated with `rated:n or resched:n`.

### DIFF
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -122,7 +122,7 @@ pub enum TemplateKind<'a> {
 pub(super) enum EaseKind {
     Rated(u8),
     Reviewed,
-    All,
+    Manually,
 }
 
 impl std::fmt::Display for EaseKind {
@@ -132,7 +132,7 @@ impl std::fmt::Display for EaseKind {
         match self {
             Rated(u) => write!(f, " and ease = {}", u),
             Reviewed => write!(f, " and ease in (1, 2, 3, 4)"),
-            All => write!(f, ""),
+            Manually => write!(f, " and ease = 0"),
         }
     }
 }
@@ -305,6 +305,7 @@ fn search_node_for_text_with_argument<'a>(
         "is" => parse_state(val)?,
         "flag" => parse_flag(val)?,
         "rated" => parse_rated(val)?,
+        "resched" => parse_resched(val)?,
         "dupe" => parse_dupes(val)?,
         "prop" => parse_prop(val)?,
         "re" => SearchNode::Regex(unescape_quotes(val)),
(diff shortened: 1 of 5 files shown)


rslib/src/search/parser.rs
```diff
@@ -359,19 +367,31 @@ fn parse_rated(val: &str) -> ParseResult<SearchNode<'static>> {
 
     let ease = match it.next() {
         Some(v) => {
-            let n: u8 = v.parse()?;
-            if n < 5 {
-                Some(n)
+            let u: u8 = v.parse().unwrap();
```
dae, line 370: Unwrap should only ever be used when we're sure the value exists. Because of this change, rated:1:x will throw an exception, and would crash the program if pyo3 didn't catch panics.

rslib/src/search/parser.rs
```diff
@@ -118,6 +118,13 @@ pub enum TemplateKind<'a> {
     Name(Cow<'a, str>),
 }
 
+#[derive(Debug, PartialEq, Clone)]
+pub enum EaseKind {
+    Rated(u8),
+    Reviewed,
+    Manually,
```
dae, line 125: What do you think about changing the wording to the following?

AnswerButton(u8)
AnyAnswerButton
ManualReschedule

rslib/src/search/sqlwriter.rs
```diff
@@ -717,19 +719,28 @@ mod test {
         assert_eq!(
             s(ctx, "rated:2").0,
             format!(
-                "(c.id in (select cid from revlog where id>{}))",
+                "(c.id in (select cid from revlog where id>{} and ease between 1 and 4))",
```
dae, line 722: I'd be inclined to simplify this to ease > 0

dae (commented):
Hi Henrik,

I think treating the 0 case differently is not a bad idea, but wonder if we need to introduce a separate search term. What do you think about making ease default to 1-4 if not provided, but allowing 0 to be passed in if the user wants it? Eg resched:3 would become rated:3:0.

Some other comments below.

rslib/src/search/parser.rs
hgiesel: That's probably clearer, will fix it.

hgiesel:
I don't really like the idea of keeping `:0` around, as it doesn't make any sense to the user.
`rated:n:1` makes sense to the user, because  the user has to click "1" to rate a card again, and also it is the first button from the left. The same for the other answers. However "0" only makes sense if you know the database implementation.

Having `rated:n:0` match anything other than a subset of `rated:n` is also I think a suboptimal design decision.

I originally had the idea of introducing new flags, like `rated:n:r`, however which will keep the issue mentioned above. I think introducing a new search term, makes the most sense, because for the users rescheduling cards is something inherently different than reviewing cards. I even doubt users would every want to search for them together.

dae:
Yes, on reflection I think you're right - thanks Henrik.

rslib/src/search/parser.rs
```diff
@@ -359,19 +367,31 @@ fn parse_rated(val: &str) -> ParseResult<SearchNode<'static>> {
...
-        None => None,
+        None => EaseKind::AnyAnswerButton,
     };
 
     Ok(SearchNode::Rated { days, ease })
 }
 
+/// eg resched:3
+fn parse_resched(val: &str) -> ParseResult<SearchNode<'static>> {
+    let mut it = val.splitn(1, ':');
```
RumovZ, line 385: I'm currently resolving the merge conflicts in #922 and was wondering: Is this line intentional?
I think when a user inputs `resched:1:1` he must have very flawed assumptions about this kew word and should get an error message rather than silently stripping away the suffix. Or am I the one with the flawed assumptions?

hgiesel: When I currently enter `resched:1:1`, I do get an error however?

RumovZ: Ah well, you pass 1, so it's a redundant operation anyway. I have omitted it in my PR.
```
assert_eq!("any string", "any string".splitn(1, ' ').next().unwrap());
```

hgiesel: Oh, now I see it! Yes, thanks for omitting it.

PR #921, opened 2021-01-12
Allow to enter values in m/h/d for steps in Deck conf

hgiesel:
Related: https://github.com/ankitects/faqs/pull/3

This is a PR after answering [this question](https://forums.ankiweb.net/t/hard-interval-longer-than-easy-help/6449/2) on the forum. "Why is the hard interval longer than the easy interval?" is such a common question, and this user basically had the answer right in front of this eyes.

This PR would change for example "15 60 1440" to "15m 1h 1d", for displaying the values to the user, but does not touch, what is saved in the `conf` object.

If a user entered a value without a unit, it would be interpreted as minutes (like current behavior). If a user entered `1440m`, next time the value is loaded, it will be shown to him as `1d`.

An issue here would be missing internalization, even though `m`, `h`, `d` are also SI units. An alternative would be to take a `TR` values.

What do you think about this idea?

### DIFF
--- a/qt/aqt/deckconf.py
+++ b/qt/aqt/deckconf.py
@@ -3,6 +3,7 @@
 # License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 from operator import itemgetter
 from typing import Any, Dict
+from enum import Enum
 
 from PyQt5.QtWidgets import QLineEdit
 
@@ -24,6 +25,12 @@ from aqt.utils import (
     tooltip,
     tr,
 )
+
+
+class StepKind(Enum):
+    MINUTES = 1
+    HOURS = 60
+    DAYS = 1440
 
 
 class DeckConf(QDialog):
@@ -172,10 +179,22 @@ class DeckConf(QDialog):
 
     def listToUser(self, l):
         def num_to_user(n: Union[int, float]):
-            if n == round(n):
-                return str(int(n))
-            else:
-                return str(n)
+            step_kind = StepKind.MINUTES
+            suffix = 'm'
+
+            if n % 144 == 0:
+                step_kind = StepKind.DAYS
+                suffix = 'd'
+
+            elif n >= 60 and n % 6 == 0:
+                step_kind = StepKind.HOURS
+                suffix = 'h'
+
+            num = n / step_kind.value
+            if num == int(num):
+                num = int(num)
+
+            return f"{num}{suffix}"
 
         return " ".join(map(num_to_user, l))
 
@@ -255,7 +274,6 @@ class DeckConf(QDialog):
 
     # Saving
     ##################################################
-
     def updateList(self, conf: Any, key: str, w: QLineEdit, minSize: int = 1) -> None:
         items = str(w.text()).split(" ")
         ret = []
@@ -263,11 +281,24 @@ class DeckConf(QDialog):
             if not item:
                 continue
             try:
+                step_kind = StepKind.MINUTES
+
+                if item.endswith('m'):
+                    item = item[:-1]
+
+                elif item.endswith('h'):
+                    item = item[:-1]
+                    step_kind = StepKind.HOURS
+
+                elif item.endswith('d'):
+                    item = item[:-1]
+                    step_kind = StepKind.DAYS
+
                 i = float(item)
                 assert i > 0
                 if i == int(i):
                     i = int(i)
-                ret.append(i)
+                ret.append(i * step_kind.value)
             except:
                 # invalid, don't update
                 showWarning(tr(TR.SCHEDULING_STEPS_MUST_BE_NUMBERS))


dae:
I seem to recall translators mistakenly mapping multiple periods to the same abbreviation in the past, so if we went down the translatable route, we'd need to sanity check the values were distinct, and make sure they do not contain spaces. Might be easier to start with the SI units and localize them later if there's demand.

No objections to the idea, though it would be nicer if this were implemented as backend methods, so AnkiMobile and AnkiDroid could take advantage of it as well. There are some helpers in timespan.rs that could be useful - Timespan::from_secs(...).natural_span() for example.

dae:
Also, while I think the labels would be handy, I'm not sure it would really help with the question you linked to. To solve that, we may need to either warn or refuse to allow learning steps that are higher than the configured graduating interval.

hgiesel:
I think this will require at least some parsing, so I will wait on #922.

Regarding the question I linked to:
My thinking was: If he would have seen the Learning interval explicitly say "6d", it might have clicked for him, or at least he would have inferred, where that interval of 6 days comes from. But of course, it doesn't solve the root cause of his confusion. 

Refusing setting the learning steps too high seems problematic to me, as we don't know whether people might want this behaviour.  Maybe we could a small red info box, in the options menu, that activates when the graduating interval or easy interval is lower than one of the learning steps, and links to the new FAQ page. This would constitute a new PR though. And would also require some more backend calls, I assume.

RumovZ:
What about a GUI solution for this? That would be more user-friendly and solve the localisation problem. Here's a quick mockup:
mock_learning_steps_widget

I've used push buttons for the steps to illustrate that they should be removable when clicked on. A new learning step would be automatically inserted at the right position.
Of course, this would have to be reimplemented for other platforms, but they all have similar widgets.

I've always found that the use of line edits when there are more input-appropriate options contributes to that quirky open-source feeling – which is adorable, but usually not the intention. 😉

dae:
Learning steps were intended to be used during the initial acquisition phase, so 30 day or 1 year delays is really pushing their intended design goal. Having explicit buttons at that range gives an implicit signal to users that this is something they should/might want to do.

Turning the textbox into a bunch of clickable areas that allowed editing like the bottom half of your screenshot could work, but I wonder if it's really worth the effort at the moment - it doesn't help users actually understand how learning steps work, and advanced users who are used to them may find it easier just to type things in.

RumovZ:
So the inaccessibility is kind of intentional. I haven't thought about it that way! A quote of my favourite author comes to mind:

>  “Some humans would do anything to see if it was possible to do it. If you put a large switch in some cave somewhere, with a sign on it saying 'End-of-the-World Switch. PLEASE DO NOT TOUCH', the paint wouldn't even have time to dry.”
>― Terry Pratchett, Thief of Time 

You have mentioned at some other point that the options already feel crowded. On the other hand, power users constantly ask for new ways to modify Anki. Was it ever a consideration to move the more advanced options to an external XML oder JSON file?

dae:
That would make extra work for us in the short term, as we'd need to make some way to keep that config file synced across platforms, and if it's just exposing the same settings as we currently have, I don't think it would gain us much. One possible option in the future would be to expose some JavaScript hooks to alter key parts of the scheduling - it would offer maximum flexibility, but it gives users and even more powerful gun with which to shoot themselves in the foot with, and could be a pain to support.

RumovZ:
I don't think that file would need to be synced. The settings would get stored internally as they are now. Only when Anki starts would it try to find and parse such a file. If that is done successfully, the internal settings are updated. Then when closing Anki, the internal settings would get serialized into the config file if it doesn't already exist.
The advantages I see are the simpler GUI and that inexperienced users could hardly access settings they don't understand. (If they tampered with the config file it would simply get unreadable).

dae:
A number of deck and add-on authors have taken to telling their users to adjust the default options to something else, and hiding the options away in a file will likely just result in these third parties updating their instructions unfortunately ^_^; It might help a bit with users who discover the deck options screen and feel like they're missing out if they don't twiddle all the knobs, but it's also a change that could be perceived as hostile by some users, and I'm not sure the gains would be worth the effort and churn - at least for now.

RumovZ:
> A number of deck and add-on authors have taken to telling their users to adjust the default options to something else, and hiding the options away in a file will likely just result in these third parties updating their instructions unfortunately

It could even make it easier for them as they only would have to provide a config file then. 😉 But maybe replacing a technical looking file would feel more like the drastic interference with Anki's internals that it really is.

> but it's also a change that could be perceived as hostile by some users

Good point, existing options probably shouldn't be moved.

I just wanted to put the idea out there. Maybe it's something worth considering in the future.

hgiesel:
Seeing as the new deckoptions now feature these suffixes, I'll close this

PR #922, opened 2021-01-12
Custom search errors

RumovZ:
This is a revison of the search parser to allow for custom search errors.
Currently, the search errors are not yet integrated but should be fully testable as they are printed to the console.

Before I continue, I'd like to know whether this approach is agreeable to you, @dae. As you can see, it adds some complexity, mainly to the search node parsers (`partially_quoted_term`, `unquoted_term`, `quoted_term_str`) and `parse_prop`.
Maybe the search node parsers could be written a bit cleaner with a different parse approach (despite what it looks like in Git, I've basically kept the structure of the old parser). Of course, `parse_prop`'s complexity could be reduced by reducing the accuracy of the errors.

@hgiesel, maybe this is of interest for you, too.


dae:
Your code looks solid, and I don't think it really harms readability/maintainability, so LGTM. Perhaps we could make it a bit more succinct with a helper, such as the following?

```patch
diff --git a/rslib/src/search/parser.rs b/rslib/src/search/parser.rs
index d540e078d..f3eadfb82 100644
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -53,6 +53,10 @@ enum ErrorKind {
     InvalidMid,
 }

+fn anki_err(input: &str, kind: ErrorKind) -> nom::Err<ParseError<'_>> {
+    nom::Err::Failure(ParseError::Anki(input, kind))
+}
+
 type IResult<'a, O> = std::result::Result<(&'a str, O), nom::Err<ParseError<'a>>>;
 type ParseResult<'a, O> = std::result::Result<O, nom::Err<ParseError<'a>>>;

@@ -193,7 +197,7 @@ fn group_inner(input: &str) -> IResult<Vec<Node>> {
     }

     if nodes.is_empty() {
-        Err(Failure(ParseError::Anki(input, ErrorKind::EmptyGroup)))
+        Err(anki_err(input, ErrorKind::EmptyGroup))
     } else if nodes.last().unwrap() == &Node::And {
         Err(Failure(ParseError::Anki(input, ErrorKind::MisplacedAnd)))
     } else if nodes.last().unwrap() == &Node::Or {
```

RumovZ:
Good idea.

I find nom's escape parser a bit hard to follow. It's very useful, but the result is unintuitive when called with `is_not()` in the first argument. Also, it's annoying when I have to annotate the error type like this: `escaped::<_, ParseError, _, _, _, _>`.
I was considering reimplementing it, but the current option is probably the lesser evil.

I'll have a go at the TR strings then and give notice when I have something to work with so you can say if the general direction is all right. The final result will require a native speaker anyway.

dae:
Happy to defer to your judgement on the escape handling - you're probably more familiar with the code than I am at this point :-)

I think glob importing an enum into a file's top-level namespace may be frowned upon, and it has the disadvantage that when calling the function you don't easily get a list of available options. If ParseErrorKind is too wordy, how about a shorter (but still readable) alias? eg

```patch
diff --git a/rslib/src/search/parser.rs b/rslib/src/search/parser.rs
index feab52532..66c7d2c8c 100644
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -18,12 +18,12 @@ use nom::{
 };
 use regex::{Captures, Regex};
 use std::borrow::Cow;
+use ParseErrorKind as FailKind;
 use ParseErrorKind::*;
-
 type IResult<'a, O> = std::result::Result<(&'a str, O), nom::Err<ParseError<'a>>>;
 type ParseResult<'a, O> = std::result::Result<O, nom::Err<ParseError<'a>>>;

-fn parse_failure(input: &str, kind: ParseErrorKind) -> nom::Err<ParseError<'_>> {
+fn parse_failure(input: &str, kind: FailKind) -> nom::Err<ParseError<'_>> {
     nom::Err::Failure(ParseError::Anki(input, kind))
 }

@@ -132,9 +132,9 @@ fn group_inner(input: &str) -> IResult<Vec<Node>> {
                     // before adding the node, if the length is even then the node
                     // must not be a boolean
                     if node == Node::And {
-                        return Err(parse_failure(input, MisplacedAnd));
+                        return Err(parse_failure(input, FailKind::MisplacedAnd));
                     } else if node == Node::Or {
-                        return Err(parse_failure(input, MisplacedOr));
+                        return Err(parse_failure(input, FailKind::MisplacedOr));
                     }
                 } else {
                     // if the length is odd, the next item must be a boolean. if it's
```

The change of parse_failure() sig is not necessary, but means the shorter version is shown in the signature when you use code complete.

I do wish Rust supported eliding the name like in Swift (eg parse_failure(input, .misplacedOr))

RumovZ:
I don't know Swift but that looks like a great feature. I'll use an alias then.

As for the escaped parser, turns out I got it wrong...
For future reference:
```
============= TESTING PARSER: =============
escaped(is_not("ab\\"), '\\', none_of("Bb"))
================ variables ================
'\': control character
'A': accepted, escapable
'B': accepted, unescapable
'a': unaccepted, escapable
'b': unaccepted, unescapable
===========================================
Empty string:  Ok(("", ""))
"\":           Err(Error(("\\", Escaped)))
"A":           Ok(("", "A"))
"B":           Ok(("", "B"))
"a":           Err(Error(("a", Escaped)))
"b":           Err(Error(("b", Escaped)))
"\A":          Ok(("", "\\A"))
"\B":          Err(Error(("B", NoneOf)))
"\a":          Ok(("", "\\a"))
"\b":          Err(Error(("b", NoneOf)))
"A\A":         Ok(("", "\\A"))
"A\B":         Err(Error(("B", NoneOf)))
"A\a":         Ok(("", "\\a"))
"A\b":         Err(Error(("b", NoneOf)))
"A\":          Err(Error(("b", NoneOf)))
```
This actually makes more sense than my assumption.

dae:
Thanks for all your work on this Rumo, and for the extensive tests. I'll review the wording of the FTL files before pushing them to the translation site, but at first glance they look fine, and I don't think I'd have even known you weren't a native speaker if you hadn't said so :-)

The only thing I might tweak is add a separate string for the 'invalid search: ...' prefix so we and the translators don't need to repeat ourselves endlessly - I know the translation docs talk about erring on the side of verbosity, but I'll update them to add this as an example of where a separate string is probably the better option.

dae:
Some more thoughts now that I've reviewed the strings:

Each time we modify a template string, we either have the option of keeping the same ID, which will mean translators are not notified of the change, or we need to pick a new ID, which will switch all of the existing translations for that string into English until the translators have a chance to translate the new string. Contributions are sporadic, and it can be months or sometimes even years until a string gets (re)translated, so we need to try our best to avoid strings that are likely to change in the future. While we can be fairly confident the number of answer buttons won't change in the short term, strings that list out options exhaustively like the following are more likely to go out of date:

```
search-invalid-prop-property = <code>prop:</code> must be followed by one of the predefined card properties: <code>ivl</code> (interval), <code>due</code>, <code>reps</code> (repetitions), <code>lapses</code>, <code>ease</code> or <code>pos</code> (position).
```

How about just listing the invalid text instead? Eg

search-invalid-prop-property = Found <code>prop:{ $provided }</code>, but "{ $provided }" is not a valid card property.

This means we don't need to remember to keep the string up to date, and won't be periodically breaking the translation. We do lose the at-a-glance list of available options and the short descriptions of what they do, but if you want to keep that, I think we'll need to split them into a bunch of different translatable strings we can join together, and I'm not sure it's worth it - simply pointing the user at the part that is wrong is probably getting them 90% there.

Another point is that translators are often not programmers, and they can struggle to preserve things like HTML in the original string. Pontoon does a decent job of showing the semantic parts of the text, but it's still another thing that can go wrong, so it's best to minimize its use. And explicit line breaks are best avoided, as they can make things odd when displayed on narrow screens. Some other strings already take advantage of markdown, so I've switched us over to that here.

I'll push out some changes I made based on the above now - please take a look and let me know if there's anything you disagree with / want to change; I'll hold off on pushing these out to Pontoon until I hear from you.

RumovZ:
Makes all sense to me. Thanks for the improvements. 🙂 
Apart from the greater value to the user of listing all valid options, I was also avoiding printing uncertain user input. An invalid prop number is preceded by a valid property and a valid operator, an unknown escape are exactly one or two chars, but an invalid property itself could be any string of any length.
However, you make a good point about the fast aging of translations. Many users wouldn't benefit from untranslated search errors anyway.

I noticed one thing, though. The new arguments also need to be escaped:

grafik

dae:
It looks like the only way to do that is to use a larger number of outer backticks in the Fluent string. I think we can probably call it "good enough" already - developer time is limited, and users providing a backtick to prop: is likely to be rare if it happens at all.

PR #923, opened 2021-01-13
Fix writing dupe node and escape issues

RumovZ:
1. Fix writing dupe nodes and rename comments and dupe functions on the backend accordingly to avoid further confusion.
2. Add a backend filter for searching for dupes. This enforces the escaping of backslashes which wasn't handled properly before. Unfortunately, this requires a third escaping variant because unlike with regex, we can't expect the dupe text not to end in an odd number of backslashes.


dae:
Looks good, thanks!

dae:
(rebased to fix a merge conflict)

PR #924, opened 2021-01-13
Gitignore root target folder

RumovZ:


dae:
What's creating it in your workflow? If you're using Rust Analyzer in VS Code that should be building in a separate location, and bazel build/test shouldn't be creating the target folder either. It is possible I have missed something though, as I have CARGO_TARGET_DIR=~/Local/rust defined in my environment, which avoids littering target folders through the filesystem when building other projects.

RumovZ:
It seems to be created by rust analyzer. When you say it should be building in a separate location, what location would that be? rslib?

dae:
Ok, I think it must just be the env var I had set - I'd wrongly assumed RA was storing them in some other temporary folder on its own accord.

Looks like RA have finally fixed that missing json macro message on save that was going on for a while 🎉

PR #925, opened 2021-01-14
Prefer <br> over <div></div> in Editor

hgiesel:
In the Anki Editor, pressing Enter will create a mix of `<div></div>` and `<br>`.

Most WYSIWYG editors out there either use `<div></div>` or even `<p></p>`. But I think the Anki Editor is somewhat special here, because it allows you to use templating tools, like MathJax or Anki clozes.

This behavior of occasionally using `<div>` makes using these template languages significantly harder.

1. You have to fit MathJax into one line. MathJax can get quite out of hand sometimes, but in Anki, you always have to fit it in a single line, making it very hard to read. MathJax offers an [option to allow specific self-closing tags within itself](http://docs.mathjax.org/en/latest/options/document.html), which Anki would greatly benefit from.

**EDIT**: I noticed that actually Anki strips html inside MathJax tags now, and at the time of writing I wasn't aware of "Shift+Enter". So this problem is somewhat alleviated now.

2. Anki Clozes break when used with multiple lines. This is an example of a `Cloze 1` with broken Anki Clozes:
Screenshot 2021-01-14 at 14 01 48

There's three things wrong here:
1. The clozes are missing the blue highlight entirely (sometimes only the first line has it), because the Anki cloze wants to wrap it in a span, but the `div`s, break the layout.
2. The newlines are inconsistent. `}}` on a new line might or might not create a newline.
3. When clicking three times on a line, and activating the cloze shortcut, you might get one of the following results:

```
{{c1::Another line}}
// or
{{c1::Another line
}}
// or
{{c1::
Another line
}}
```

What I would expect is the first behavior, always.

I introduced this small change into `editor.ts`, which would change this behavior to always insert `<br>` instead. After that, you get consistent behavior in that regard:

Screenshot 2021-01-14 at 13 30 45

Wondering about this, I dug a little bit in Anki history, and found [this old post](https://anki.tenderapp.com/discussions/ankidesktop/4363-div-instead-of-br-are-now-inserted-when-pressing-enter), and this old commit 077d6b8187db4e5b2cfc44fe58acad2bd6614c53 from 2013, which seem to imply that this used to be the behavior, but was changed because of unspecified issues (at least I couldn't find an exact description).

Considerations:
* `execCommand("insertLineBreak")` [only works in WebKit](https://w3c.github.io/editing/docs/execCommand/#the-insertlinebreak-command), which means it won't work on AnkiMobile (even though I don't know whether the editor there is a webview).

### DIFF
--- a/qt/aqt/data/web/js/editor.ts
+++ b/qt/aqt/data/web/js/editor.ts
@@ -55,6 +55,13 @@ function onKey(evt: KeyboardEvent) {
         return;
     }
 
+    // prefer <br> instead of <div></div>
+    if (evt.which === 13) {
+        evt.preventDefault();
+        document.execCommand("insertLineBreak");
+        return;
+    }
+
     // fix Ctrl+right/left handling in RTL fields
     if (currentField.dir === "rtl") {
         const selection = window.getSelection();
--- a/qt/aqt/data/web/js/editor.ts
+++ b/qt/aqt/data/web/js/editor.ts
@@ -56,7 +56,8 @@ function onKey(evt: KeyboardEvent) {
     }
 
     // prefer <br> instead of <div></div>
-    if (evt.which === 13) {
+    if (evt.which === 13 && !inListItem()) {
+        console.log("Enter");
         evt.preventDefault();
         document.execCommand("insertLineBreak");
         return;
@@ -87,6 +88,25 @@ function onKey(evt: KeyboardEvent) {
     triggerKeyTimer();
 }
 
+function nodeIsElement(node: Node): node is Element {
+    return node.nodeType == Node.ELEMENT_NODE;
+}
+
+function inListItem(): boolean {
+    const anchor = window.getSelection().anchorNode;
+
+    let n = nodeIsElement(anchor) ? anchor : anchor.parentElement;
+
+    let inList = false;
+
+    while (n) {
+        inList = inList || window.getComputedStyle(n).display == "list-item";
+        n = n.parentElement;
+    }
+
+    return inList;
+}
+
 function insertNewline() {
     if (!inPreEnvironment()) {
         setFormat("insertText", "\n");
@@ -113,11 +133,10 @@ function insertNewline() {
 }
 
 // is the cursor in an environment that respects whitespace?
-function inPreEnvironment() {
-    let n = window.getSelection().anchorNode as Element;
-    if (n.nodeType === 3) {
-        n = n.parentNode as Element;
-    }
+function inPreEnvironment(): boolean {
+    const anchor = window.getSelection().anchorNode;
+    const n = nodeIsElement(anchor) ? anchor : anchor.parentElement;
+
     return window.getComputedStyle(n).whiteSpace.startsWith("pre");
 }
 
--- a/qt/aqt/data/web/js/editor.ts
+++ b/qt/aqt/data/web/js/editor.ts
@@ -57,7 +57,6 @@ function onKey(evt: KeyboardEvent) {
 
     // prefer <br> instead of <div></div>
     if (evt.which === 13 && !inListItem()) {
-        console.log("Enter");
         evt.preventDefault();
         document.execCommand("insertLineBreak");
         return;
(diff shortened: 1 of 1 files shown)


abdnh:
> You have to fit MathJax into one line. MathJax can get quite out of hand sometimes, but in Anki, you always have to fit it in a single line, making it very hard to read.

Shift+Enter inserts `<br>` instead, by the way.
It's documented in the manual: https://docs.ankiweb.net/#/math?id=mathjax

(But of course, it will be nicer if Enter was enough)

hgiesel:
@RumovZ 
I think that's actually true for all `contenteditable` fields, if I'm not mistaken?
I also edited my opening post, because I noticed, that Anki strips away HTML in MathJax equations by now (and which is not mentioned in the manual).

However I think the clozes already warrant having a look at this. (It also makes Anki fields much easier to reason about, IMO).

kleinerpirat:
I feel like there still needs to be a difference between hard and soft line-breaks, otherwise semantic value would be lost.

If `Enter` would insert `<br>` just like `Shift+Enter` (soft line-break), how would one distinguish between creating a new `<li>` and adding a new line within the current `<li>`? In WSIWYG editors like Word, soft line-breaks are used for that purpose.

RumovZ:
@hgiesel, I think you wanted to mention @abdnh.

hgiesel:
Yes, excuse me 🙇
@kleinerpirat I added a check for that. Basically to keep the old behavior, if you're in a list.

I also noticed it's not perfect, there's still certainly ways, `div`s can get in the fields. For example, if you have a two item `ul`, and delete the first one, it will also create a list. However I also noticed that these remaining `div`s typically don't change the formatting overall.

abdnh:
>  I think that's actually true for all `contenteditable` fields, if I'm not mistaken?

Yes, it seems so. I was not aware of that.

notnono:
As a note in the discussion, making contenteditable display: inline-block and then custom handling for enter when exiting lists seems to solve the issue. https://forums.ankiweb.net/t/where-in-source-to-replace-newline-div-behaviour/6443/9

dae:
This is a change I've wanted to make in the past, but ran into problems each time I tried.If you've figured out a way to solve it that doesn't cause regressions, that will be great.

Some of the issues I seem to recall encountering in the past:

- hitting enter at the end of a field not expanding it (perhaps due to stripping trailing BRs)
- the inline-block css change caused Chrome to use BRs when typing text manually, but as soon as the user pastes content with divs in, it switched backed to using divs. Your change doesn't seem to suffer from this as it's catching the enter key, though I'm not sure we'll be able to use the same trick for the mobile clients.
- in Anki 2.0, switching to BRs caused WebKit to crash (!). Presumably not an issue with Chromium.

I gave it a quick play and didn't spot problems, so will merge this in for now. Let's hope it doesn't introduce any regressions🤞

noinkling:
I see this is already merged and in the latest betas, which is great. I just wanted to give another example of how divs break expectations in case you're thinking about reverting this change for whatever reason.

If part of your template looks like:

```html
<p style="color: red;">{{SomeField}}</p>
```

which seems like a very reasonable thing to do for a field containing text content, it mostly works fine until `SomeField` contains a `div`. This is because the `p` element can only contain [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content), and encountering a `div` automatically closes the `p` tag, so you end up with the DOM looking like:

```html
<p style="color: red;">First line</p>
<div>Second Line</div>
<p></p>
```

When you're editing such a field and all you want to do is have some text start on a newline, you're just pressing <kbd>Enter</kbd> and regarding it like a text editor or word processor, you're not thinking about the actual elements being inserted and how they interact to produce quirks like this. That's why using `br` by default is really the only sensible option.

heinfl:
Is it possible to create a configuration entry in which the user can decide for himself whether a <br> should be inserted on Enter or whether the old behavior should be kept?
I have various JavaScript rules that require a <div>. At the moment I have no way of deciding whether a <div> or a <br> should be inserted.

heinfl:
Is it possible to create a configuration entry in which the user can decide for himself whether a \<br\> should be inserted on Enter or whether the old behavior should be kept?
I have various JavaScript rules that require a \<div\>. At the moment I have no way of deciding whether a \<div\> or a \<br\> should be inserted.

hgiesel:
@heinfl 
Could you explain your use case?

heinfl:
@hgiesel 
I write cards like this:
image
In the old behavior i could determine a hard break line and a soft break line to handle styles and add links to show the hints for a line:
image
Now this is only possible by using the html editor to insert divs

whitemuu:
I though now I can input <div> with `shift+enter`, but in fact I cannot?

How to input <div> now?

PR #926, opened 2021-01-14
Introduce "prop:rated" and "prop:resched"

hgiesel:
There is still an issue with the `rated` and `resched` queries.

Let's say I want to search which cards I reviewed 20 days ago from today.

You would think, it would be as simple as `rated:21 -rated:20`, however that's not the case. It would be the case for queries like `added`, and `edited`, but not for rated.

`rated:21 -rated:20` actually gives you the cards, which you studied 20 days ago,  but will also exclude those, which you happened to study starting today till 19 days ago. So cards you reviewed that day, but haven't since.

This PR is trying to solve this, by introducing equivalent `prop` searches, which allow for more specificity when searching for past reviews. So searching for 20 days ago would be as simple as `prop:rated=-20`.

This also means There is some overlay with `rated:n`. In fact `rated:n:m` would be fully equivalent to `prop:rated>=(-n + 1):m`

There are some considerations, and also some open issues:
* I decided to keep negative indices for `prop:rated`, because I modelled it to be similiar to `prop:due`. So `prop:due=-1 or prop:rated=-1` would give you cards which you reviewed yesterday, or are still due from yesterday.
* This also means that `prop:rated` only accepts integers `<= 0`.
* At the moment there is no functionality to cap rated searches, which go the other way, like `prop:rated<0`
* While those would be easy to cap to 365 days, `prop:rated!=n` is not quite as straightforward to put a cap on. I could cap it 182 days in each direction. Another idea would be to throw a new UnsupportedOperator error in this case.

And one question out of curiosity:
* I don't really have that many cards/reviews in my collection, so I don't really know how out-of-hand such a query can get, but I wonder what would be the consequences of lifting that one year cap?

### DIFF
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -87,7 +87,7 @@ pub enum PropertyKind {
     Lapses(u32),
     Ease(f32),
     Position(u32),
-    Rated(u32, Option<u8>),
+    Rated(u32, EaseKind),
 }
 
 #[derive(Debug, PartialEq, Clone)]
@@ -433,7 +433,16 @@ fn parse_prop(s: &str) -> ParseResult<SearchNode> {
                 }
             }
             None => EaseKind::AnyAnswerButton,
-        }
+        };
+
+        PropertyKind::Rated(days, ease)
+    } else if key == "resched" {
+        let mut it = val.splitn(2, ':');
+
+        let n: u32 = it.next().unwrap().parse()?;
+        let days = n.max(1).min(365);
+
+        let ease = EaseKind::ManualReschedule;
 
         PropertyKind::Rated(days, ease)
     } else if key == "resched" {
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -461,7 +461,10 @@ fn parse_prop(s: &str) -> ParseResult<SearchNode> {
         ));
     };
 
-    Ok(SearchNode::Property { operator, kind })
+    Ok(SearchNode::Property {
+        operator: operator.to_string(),
+        kind,
+    })
 }
 
 /// eg added:1
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -87,7 +87,7 @@ pub enum PropertyKind {
     Lapses(u32),
     Ease(f32),
     Position(u32),
-    Rated(u32, EaseKind),
+    Rated(i32, EaseKind),
 }
 
 #[derive(Debug, PartialEq, Clone)]
@@ -439,8 +439,8 @@ fn parse_prop(s: &str) -> ParseResult<SearchNode> {
     } else if key == "resched" {
         let mut it = val.splitn(2, ':');
 
-        let n: u32 = it.next().unwrap().parse()?;
-        let days = n.max(1).min(365);
+        let n: i32 = it.next().unwrap().parse()?;
+        let days = n.max(-365).min(0);
 
         let ease = EaseKind::ManualReschedule;
 
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -406,7 +406,7 @@ fn parse_prop(s: &str) -> ParseResult<SearchNode> {
         let mut it = num.splitn(2, ':');
 
         let days: i32 = if let Ok(i) = it.next().unwrap().parse::<i32>() {
-            i
+            i.min(0)
         } else {
             return Err(parse_failure(
                 s,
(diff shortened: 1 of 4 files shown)


dae:
Introducing a separate prop query sounds good to me. Re requiring negative values, due numbers can be both in the past and the future, but rated dates will always be in the past or present unless the user sets their clock backwards. Using a positive value instead would make things slightly easier, and while it would lose consistency with other due values, it would be consistent with the existing rated search. I wonder which is the better option?

I am ok with dropping the limit completely - it can be slow with millions of reviews to sort through, but there are plenty of other queries that can be slow too.

hgiesel:
First of all, right now, the `rated` searches now correspond like this:

```
rated:1 == prop:rated>-1
rated:2 == prop:rated>-2
...
```

When flipping the prop indices, they would correspond like this:
```
rated:1 == prop:rated<1
rated:2 == prop:rated<2
...
```

In fact my first implementation used positive values. But I decided against it after my own initial usage, as I found it much harder to wrap my head around about flipped values when dealing with "less" / "greater".

Let's say we want to search for due and reviewed cards from yesterday and the day before using flipped `prop:rated`. That would be:
```
(prop:due>=-2 and prop:due<=-1) or (prop:rated>=1 and prop:rated<=2)
```

1. Having them both exist parallel to each other, means you will always trip over your brain, when using one of them, because you have to think about what exactly "less" / "greater" meant again in terms of time in this particular case.
2. I also think it's more easy to think of the past to be smaller in relation to the future, which is why I find `prop:due` to be more intuitive.

I found this to be more striking than the negated values, because they would somewhat fall into a pattern, with all of `added:n`/`edited:n`/`rated:n` working one way, and `prop:rated`/`prop:due` working in another.

hgiesel:
I've now amended it in such a way that `prop:rated/resched` searches are coerced in a way, which allow for both negative and positive days, e.g.

```
"prop:rated>=5" is coerced to "prop:rated<=-5"
"prop:resched<2" is coerced to "prop:resched>-2"
"prop:rated=7" is coerced to "prop:rated=-7"
```

We could technically flip these around, and coerce from negative to positive, but I'm still somewhat in favour of keeping the negative days, because of the confusion with `prop:due` that I mentioned.

rslib/src/search/sqlwriter.rs
```diff
@@ -214,14 +214,21 @@ impl SqlWriter<'_> {
...
+
+        write!(self.sql, "c.id in (select cid from revlog where id").unwrap();
+
+        match op {
+            ">" => write!(self.sql, " >= {}", target_cutoff_ms),
+            ">=" => write!(self.sql, " >= {}", day_before_cutoff_ms),
+            "<" => write!(self.sql, " < {}", day_before_cutoff_ms),
+            "<=" => write!(self.sql, " < {}", target_cutoff_ms),
+            "=" => write!(self.sql, " between {} and {}", day_before_cutoff_ms, target_cutoff_ms - 1),
+            _ /* "!=" */ => write!(self.sql, " not between {} and {}", day_before_cutoff_ms, target_cutoff_ms - 1),
```
dae, line 230: Minor nit: matching '!='explicitly, and mapping _ => unreachable!("unexpected op") would ensure we don't accidentally fall through if another comparator is ever added in the future.

dae (commented):
I don't feel strongly about it, and am happy to go with negative values if you feel that's a better way to work with them. Not so sure about the automatic coercion from positive numbers - it's more code to maintain, and I fear it may just confuse things. Would you object to rolling the last two commits back?

Otherwise looks good to go, save the minor nit posted below.

rslib/src/search/parser.rs
```diff
@@ -400,6 +415,63 @@ fn parse_prop(s: &str) -> ParseResult<SearchNode> {
...
+
+        let ease = match it.next() {
+            Some(v) => {
+                if let Ok(u) = v.parse::<u8>() {
+                    if (1..5).contains(&u) {
+                        EaseKind::AnswerButton(u)
+                    } else {
+                        return Err(parse_failure(
+                            s,
+                            FailKind::InvalidPropInteger(format!("{}{}", prop, operator)),
```
RumovZ, line 445: `InvalidPropInteger` does not really apply here. The generic `search-invalid-argument` is available, but since this will be a new feature I think it would be preferrable to provide some information about the expected value.
`search-invalid-rated-ease` could be easily generalised to work for both `rated:` and `prop:rated`.
@dae, seeing this I realise that I could and probably should have made the error strings more generic. There is no need to bind `search-invalid-prop-float` to a property search, for example. You haven't pushed them to Pontoon, yet, have you? Then I'll make a quick PR as soon as #926 is merged.

hgiesel: I added `search-invalid-rated-ease` within this PR now.

dae:
Thanks Henrik!

PR #927, opened 2021-01-18
simplify parse_prop() and associated translations

dae:
@RumovZ @hgiesel 

~~write_props()~~ parse_prop() had grown a bit too large and difficult to follow, so I've split it up into some helper functions.

Regarding Rumo's comment on the previous PR about non-generic messages, I do think the pendulum has swung a bit too far, as each separate error message we add consumes developer and translator time.

In this PR I've added an InvalidNumber error that is used for parsing floats, and both signed and unsigned ints. So for example, instead of "prop:ease>5x" yielding 

```
Invalid search: ⁨prop:⁨ease&gt;⁩ must be followed by a decimal number.⁩
```

(whoops, we're not escaping correctly, and the actual number portion is not shown)

we instead get:

```
Invalid search: ⁨⁨ease>5x⁩: invalid number provided⁩
```

I think this pinpoints the error with enough specificity for the user to not be left wondering where in their search the problem lies, while making life easier for us and the translators. What do you think?

I've left the old translations in while we discuss this, and won't push them out to Pontoon until these changes settle down. There are currently separate functions for parsing floats and ints added in this patch; they could be merged into one if we decide not to have separate messages for the different number types.

### DIFF
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -17,7 +17,11 @@ use nom::{
     sequence::{preceded, separated_pair},
 };
 use regex::{Captures, Regex};
-use std::borrow::Cow;
+use std::{
+    borrow::Cow,
+    num::{ParseFloatError, ParseIntError},
+    str::FromStr,
+};
 
 type IResult<'a, O> = std::result::Result<(&'a str, O), nom::Err<ParseError<'a>>>;
 type ParseResult<'a, O> = std::result::Result<O, nom::Err<ParseError<'a>>>;
@@ -362,7 +366,7 @@ fn parse_resched(s: &str) -> ParseResult<SearchNode> {
 }
 
 /// eg prop:ivl>3, prop:ease!=2.5
-fn parse_prop(s: &str) -> ParseResult<SearchNode> {
+fn parse_prop(prop_clause: &str) -> ParseResult<SearchNode> {
     let (tail, prop) = alt::<_, _, ParseError, _>((
         tag("ivl"),
         tag("due"),
@@ -372,8 +376,13 @@ fn parse_prop(s: &str) -> ParseResult<SearchNode> {
         tag("pos"),
         tag("rated"),
         tag("resched"),
-    ))(s)
-    .map_err(|_| parse_failure(s, FailKind::InvalidPropProperty(s.into())))?;
+    ))(prop_clause)
+    .map_err(|_| {
+        parse_failure(
+            prop_clause,
+            FailKind::InvalidPropProperty(prop_clause.into()),
+        )
+    })?;
 
     let (num, operator) = alt::<_, _, ParseError, _>((
         tag("<="),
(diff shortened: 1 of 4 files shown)


rslib/src/search/parser.rs
```diff
@@ -472,6 +415,43 @@ fn parse_prop(s: &str) -> ParseResult<SearchNode> {
...
+fn parse_prop_rated<'a>(num: &str, prop_clause: &'a str) -> ParseResult<'a, PropertyKind> {
+    let mut it = num.splitn(2, ':');
+    let days = parse_prop_integer::<i32>(it.next().unwrap(), prop_clause)?.min(0);
+    let ease = match it.next() {
+        Some(v) => match parse_prop_integer(v, prop_clause)? {
+            u @ 1..=5 => EaseKind::AnswerButton(u),
+            _ => {
+                return Err(parse_failure(
+                    prop_clause,
+                    FailKind::InvalidRatedEase(prop_clause.to_owned()),
```
dae, line 445: If we wanted to go one step further, even this case could be covered by "invalid number provided".

RumovZ:
I would prefer keeping the individual errors for float, integer, unsigned and ease.
For me, it's not only about telling the user if and where he made a mistake, but also providing some elementary information about the available filters because this is the only internal source for it at all. Even the external manual is rather sparse with details and might be outdated or not even available at all in the user's native language.
I think it would be easier for both developers and users to keep track of what is valid input for the various filters and properties via specific search errors than via the manual.

RumovZ:
`search-invalid-rated-days` can be replaced by `search-invalid-followed-by-positive-days`, though.
And should we bother at all with different error kinds when they aren't treated differently? `InvalidAdded`, `InvalidEdited`, `InvalidResched` and `InvalidRatedDays` could then be merged into a new `InvalidPositiveDays`, `InvalidDupeMid`, `InvalidDupeText` and `InvalidPropProperty` into `InvalidArgument`.

hgiesel:
I'm gonna agree with @RumovZ here. If the error message can educate the user on what to do instead, that should be preferred, because that was the whole point of error messages in the first place.

The thing that should be avoided however is coupling error messages to specific queries. This would also include `InvalidRatedEase` to simply `InvalidEase`, and possible dropping the semicolon after `{ $val }`, and explicitly passing it.

dae:
The majority of the term-specific errors boil down to the user providing the wrong number type or an invalid number. I've expanded the float/int functions in the previous change into separate ones for positive and negative ints, answer button # and so on, and I think it makes the code easier to follow - and hopefully strikes a balance between being informative enough to the user, without a burden on us or the translators. It yields messages like:

Invalid search: ⁨expected a positive whole number in "⁨reps>5x⁩", but found "⁨5x⁩".⁩
Invalid search: ⁨expected an answer button between 1-4 in "⁨5:1x⁩", but found "⁨1x⁩"⁩

I've kept the answer button and flag messages specific, but terms that deal with days are just covered by a generic number class now (like "positive whole number"). I suspect this is sufficient, as the user will not be typing in random queries and hoping it matches a search term - they will have had to either consult the manual to find out the term exists in the first place, or will have located one of the pre-defined ones in the Filter button. We could theoretically add more helper functions dealing with day-specific error messages, but I feel like it's getting into diminishing returns territory, and am not sure it's worth the extra code and work for translators. Is the approach in the latest push something you guys feel you could live with, or do you think something should be changed?

RumovZ:
I assume you don't pass the key word as context because it was stripped from the str slice and would have to be explicitly written to a new string. However, I feel like we can't do without it. E.g., `rated:` now gives the error message `expected a whole number in "", but found "".`.
Apart from that, your solution looks great to me. 👍

dae:
Oh, looks like I was passing the wrong value there - in other queries like added: the context should appear already. We lose the prop: prefix on prop searches, but figure there's probably enough context there to identify the issue still. If you're keen to fix that, some way that doesn't require an allocation in the non-error case would be good (extra &str arg or closure could do it - but diminishing returns again :-)

RumovZ:
No, I agree that it's rather redundant in the `prop` case.
Can't wait for the next release!

dae:
Actually, the prop: prefix is not a hot path, and while it feels a bit dirty, a few extra allocations are likely not even going to be measurable. So I take back my objection on those grounds. But we run into lifetime issues if we try to create the string on the fly, as the created string is deallocated before the ParseResult is returned. It'd be a bit of a pain to switch the error over to a Cow, so I'm glad you think we can live without the prop: prefix. :-)

Thank you and Henrik for all your hard work! Assuming Henrik doesn't object, I plan to push the strings to Pontoon tomorrow, and beta1 shouldn't be too far behind.

PR #928, opened 2021-01-18
Different first weekday support for Calendar View, Take 2

hgiesel:
Another take on https://github.com/ankitects/anki/pull/800.
After having worked with the backend for a bit, this was now quite easy to do.

Right now, you can only set it after opening the console via, and executing `mw.col.config.set("firstWeekday", n)`, where

    n = 0 is for Sunday
    n = 1 is for Monday
    n = 5 is for Friday
    n = 6 is for Saturday

### DIFF
--- a/rslib/backend.proto
+++ b/rslib/backend.proto
@@ -1082,6 +1082,12 @@ message GraphsIn {
 }
 
 message GraphsOut {
+  enum Weekday {
+    SUNDAY = 0;
+    MONDAY = 1;
+    FRIDAY = 5;
+    SATURDAY = 6;
+  }
   repeated Card cards = 1;
   repeated RevlogEntry revlog = 2;
   uint32 days_elapsed = 3;
@@ -1090,6 +1096,7 @@ message GraphsOut {
   uint32 scheduler_version = 5;
   /// Seconds to add to UTC timestamps to get local time.
   int32 local_offset_secs = 7;
+  Weekday first_weekday = 8;
 }
 
 message RevlogEntry {
--- a/rslib/src/config.rs
+++ b/rslib/src/config.rs
@@ -18,6 +18,7 @@ pub(crate) fn schema11_config_as_string() -> String {
         "curDeck": 1,
         "newSpread": 0,
         "collapseTime": 1200,
+        "firstWeekday": 0,
         "timeLim": 0,
         "estTimes": true,
         "dueCounts": true,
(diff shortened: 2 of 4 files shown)


dae:
Is 'bazel test ...' not working for you locally? It's a faster feedback loop than having to push to CI after each change, and can be added to your local git push hook to run automatically.

hgiesel:
@dae It does work for me, it's just some sort of laziness with setting up precommit /  prepush + forgetting the exact bazel commands on my part. I will sit down to it once and for all and set it up, also to avoid unnecessary notifications on your part, sorry for any inconvenience.

dae:
Doesn't really bother me, just thought it might save you some time :-) One thing to be aware of is that your PATH may be different in the pre-commit hook, which will cause Bazel to retest/recompile things unnecessarily - explicitly setting the path to your normal path should avoid that.

Change looks good, and based on your comments about the day being different for different cultures, I'm no longer against adding this to the UI as well. "Throw it in the Preferences screen" is the easy default option. A little options button in the graph area might be nicer (but more work).

hgiesel:
After trying out just now, I realized that `bazel test ...` is the shortcut for running all tests, where `...` is not just meant as an ellipsis for me to fill in the test name. I was just about to paste every single test into my pre-push... That will things much easier in the future.

By "graph area", do you mean into the stats screen? Wouldn't the option then be a little bit too present? I think this would be more similar to the language setting, where you set it once, and never touch it again. Close to the language setting is also where I would expect it, (and actually did the first time) as part of the locale setting.

dae:
You're probably right that it wouldn't be something people frequently change, but having it in the graphs area would make it more discoverable. I don't have a strong preference at the moment, as we can always move things around later. But I'll post a related comment in #897

mackuba:
Is there a UI for this, or is it still only accessible through the console?

mackuba:
Ooh, I've looked around that screen and clicked things, but I haven't found that! Thanks.

Does this also work on iOS? (I haven't updated the app there yet)

Andrew15-5:
> It can be configured by clicking on the weekday abbreviations in the calendar view inside the stats screen.

Is this documented somewhere? Searching "change first day of the week anki" doesn't show anything usefull, but through this comment I found this PR: https://forums.ankiweb.net/t/first-day-of-week-in-calendar-heatmap-in-statistics/5334/2.

Although it's kinda cool this way (clicking right where the change is needed), but it's not user-friendly. There's should be a setting somewhere (in the preferences).

dae:
I've updated the manual.

PR #929, opened 2021-01-19
Make tags match their parents case

abdnh:
A follow-up to your comment in https://github.com/ankitects/anki/pull/900/#issuecomment-762018745


dae:
Thanks for the quick fix! I've pushed some tweaks to this code - please let me know if you notice any problems.

PR #930, opened 2021-01-19
Refactor editor.ts

hgiesel:
This is mostly a refactor of `data/web/js/editor.ts`

It covers multiple areas:
* Correctly type the functions
* Avoid old/deprecated features:
  1. Prefer `element.addEventListener()` over `element.on~`
  2. Prefer `evt.code` over `evt.which`
  3. Prefer template strings over template concatenation (generally)
  4. Avoid `arguments`
* Avoid iterating through same array twice
* Remove functionality around mouseDown: `mousedown`/`mouseup` is only triggered, if the action is performed with the mouse on the window, so it was really easy to break anyway. I don't really see a difference without tbh, which is why I removed it completely.
* Remove most top functions out of jquery load to speed up things

### DIFF
--- a/qt/aqt/data/web/js/editor.ts
+++ b/qt/aqt/data/web/js/editor.ts
@@ -167,11 +167,8 @@ function onInput(): void {
 function updateButtonState(): void {
     const buts = ["bold", "italic", "underline", "superscript", "subscript"];
     for (const name of buts) {
-        if (document.queryCommandState(name)) {
-            $(`#${name}`).addClass("highlighted");
-        } else {
-            $(`#${name}`).removeClass("highlighted");
-        }
+        const elem = document.querySelector(`#${name}`) as HTMLElement;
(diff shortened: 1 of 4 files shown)


dae:
The code improvements are great, thanks. There currently seems to be an issue clicking on the editor buttons though - it looks like they're stealing focus from the field, causing them to become disabled.

hgiesel:
Okay I fixed that issue.

What was happening is that the code inside of `$(function () { ... })` is waiting for jQuery to finish loading. And the events happen in this order:

1. Code waits for jQuery to finish
2. Buttons are inserted
3. jQuery finishes loading
4. Code executes

Removing the dependency on jQuery makes them execute before the buttons exist in the document. This is still true for `adjustFieldsTopMargin()`, but I intend to replace that logic with a small CSS adjustment in a follow-up PR.

dae:
Thanks Henrik!

PR #931, opened 2021-01-19
Retouch the reviews graph a little bit

hgiesel:
* Put the card types in a similiar order to the card counts:
  1. Learning
  2. Relearning
  3. Young
  4. Mature
  5. Early

* Adjust the colors of the card counts
  1. Learning cards right now are blue, however this color is typically associated with new cards. I changed it to orange to match the Card counts.
  2. Now that orange is taken up by learning again, I changed the color of Early to purple, to have a new color, because early reviews don't really fit in with the existing card type colors.
* Remove (by turning transparent) the grey box in front of Running total (because there's nothing grey in the graph itself to associate it with)

### DIFF
--- a/ts/graphs/reviews.ts
+++ b/ts/graphs/reviews.ts
@@ -8,10 +8,10 @@
 
 import pb from "anki/backend_proto";
 import {
-    interpolateBlues,
     interpolateGreens,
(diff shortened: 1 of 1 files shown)


dae:
Thanks!

PR #933, opened 2021-01-20
Add labels to calendar days

hgiesel:
This adds labels on the left of the calendar graph to show the weekday. Right now it shows "S" for Sunday, "M" for Monday, etc.

I added support for localization, and made sure that it supports weekday labels longer than one character, because I know at least a few languages where it is uncommon to have a single character as a day label (e.g. German, or Spanish)

image

My main motivation here in fact is to have an inconspicuous target for setting the first weekday in https://github.com/ankitects/anki/pull/897.

### DIFF
--- a/ftl/core/statistics.ftl
+++ b/ftl/core/statistics.ftl
@@ -156,6 +156,13 @@ statistics-hours-subtitle = Review success rate for each hour of the day.
 # shown when graph is empty
 statistics-no-data = NO DATA
 statistics-calendar-title = Calendar
+statistics-calendar-label-sunday = S
+statistics-calendar-label-monday = M
+statistics-calendar-label-tuesday = T
+statistics-calendar-label-wednesday = W
+statistics-calendar-label-thursday = T
+statistics-calendar-label-friday = F
+statistics-calendar-label-saturday = S
 
 ## An amount of elapsed time, used in the graphs to show the amount of
 ## time spent studying. For example, English would show "5s" for 5 seconds,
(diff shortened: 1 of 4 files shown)


dae:
Maybe it's worth taking advantage of the existing translations in the browser?

new Date().toLocaleString("ja-JP", {weekday: "short"});
"木"
new Date().toLocaleString("de", {weekday: "short"});
"Do"
new Date().toLocaleString("en-US", {weekday: "short"});
"Thu"

We could do this by adding a routine to I18n, which uses the stored language, converting underscores to hyphens. It would mean more than one letter for English, but we could make it seem less dense by showing every second or more day, like on GitHub - what do you think?

hgiesel:
Using the built-in translations is the better idea, however like I've written, one motivation for me was for have a non-invasive but still obvious way to set the first weekday, e.g. by clicking on "M" to have Monday be the first weekday. If that label is skipped, then there's no way to set it.

dae:
Oh, I assumed you meant clicking on one of them would pull up a menu where the day could be selected. Maybe we could shrink the font instead?

hgiesel:
1. I switched to `Date.prototype.toLocaleString` and removed the fluent translations. It also offers a `{ weekday: "narrow" }` option, which might be optimal for our case, because it mostly resolves to a single character (even for German or Spanish), which I think is fine. They are mostly gonna be for the initial orientation (and the setting of the first weekday).

2. Making a menu in Javascript is somewhat more involved. I've done it before, but I'm also unsure how a context menu would translate to mobile platforms. That's why I thought something you can simply click on is more generally applicable.

dae:
Ah, I didn't know about narrow, and agreed on point 2 - thanks!

PR #934, opened 2021-01-21
Add GraphsPreferences API to graphs for setting persistent preferences

hgiesel:
This is still a draft. I have some issues right now, which I cannot solve:

1. When the backend creates the `GraphsPreferencesOut` message, it only contains config options, which are set within `aqt.mw.col.conf`.  I would like there to be a default value, which is provided if there is no value set in `aqt.mw.col.conf`. Is this possible? I thought adding it to `schema11_config_as_string()` in `rslib/src/config.rs`, would add a default, but that is not what is happening, as e.g. "firstWeekday" is already set in there, which is the option that gives me that issue.

2. I intended to reuse `GraphsPreferencesOut` also for setting the graphs preferences (I will probably rename it to GraphsPreferencesInOut or something). However I don't manage to send the data from TS to rust. At first I did the following:

```js
export async function setGraphPreferences(prefs: pb.BackendProto.GraphsPreferencesOut): Promise<void> {
    return (async () => {
        await postRequest("/_anki/setGraphPreferences", pb.BackendProto.GraphsPreferencesOut.encode(prefs).finish())
    })()
}
```

In `mediasrv.py`, upon inspecting `request.data`, I got `b'\x10\x00'`, which I think is correct (?). However I don't see any way to just pass on the raw bytes to Rust. I guess I could also just convert to JSON within TS, however I'd like to avoid that if possible.


dae:
Haven't had a chance to review the code yet, but re your comments:

schema11_config_as_string() exists because Anki traditionally required certain config variables to always be set, and if they're not set, older clients will throw an error. It's only run when the collection is first created. With newer code we can just avoid setting defaults, and apply them on the fly, which also covers the case when a key is set to an invalid value. We can also tell serde how it can convert the enum to a number directly, which makes the code quite simple to read: https://github.com/ankitects/anki/commit/99d732799cd1669eb80c7951062e8f901b0058ff

If you add the setter method to SKIP_UNROLL_INPUT in genbackend.py, it will change the method signature to accepting a Protobuf instance instead of the individual arguments. The Python code will still need to decode the bytes into the Protobuf object before passing it on with `g = pb.GraphPreferences(); g.ParseFromString(bytes)`. Decoding only to encode again is silly, but should not be a performance issue in this case. Modifying genbackend.py to allow flagging a method as taking bytes directly might be an option in the future, but we can probably punt on it until we need to transfer something non-trivial in size where decoding/encoding would be an issue.

dae:
Re the naming, how about using GetGraphPreferences() for the getter, then the message that get and set pass around can just be called GraphPreferences - like Get/SetPreferences()

dae (commented):
Sorry, SKIP_UNROLL_INPUT may not be strictly necessary, as render_method() also looks at the name and number of the fields to decide whether to list each entry separately. But we probably are still best sending the bytes directly from TS, and decoding on the Python side, rather than using JSON.

Code looks good, just one minor point below.

hgiesel:
I think it's mostly done, however the issue with the missing option was still not resolved. The `firstDayOfTheWeek` option will still be empty, if there is no value set in `mw.col.conf`. It is now more obvious than before, because the day labels will bug out, if the option is not set.

ts/graphs/CardCounts.svelte
```diff
@@ -56,7 +58,7 @@
     <h1>{graphData.title}</h1>
 
     <div class="range-box-inner">
-        <SeparateInactiveCheckbox bind:separateInactive {i18n} />
+        <SeparateInactiveCheckbox {i18n} {cardCountsSeparateInactive} />
```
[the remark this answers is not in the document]
hgiesel: I agree, it was now even more ridiculous than before. I inlined it.

hgiesel:
Nevermind, fixing the types introduced an issue...

dae:
Thanks for your work on this Henrik. I understand what you mean about the missing values after trying it out, and have pushed a fix.

I noticed while testing this that 4826369d7de2a5c43bb7c9cbe6359f8c918b8504 has broken the range box - it now defaults to having 'all time' selected instead of the current year - would you mind taking a look?

ts/graphs/preferences.ts
```diff
@@ -58,7 +58,11 @@ function preparePreferences(
         setGraphPreferences(constructPreferences());
     }
 
-    for (const [key, value] of Object.entries(graphsPreferences)) {
+    for (const [key, value] of Object.entries(
+        pb.BackendProto.GraphsPreferences.toObject(graphsPreferences, {
+            defaults: true,
```
hgiesel, line 63: I have to say this confuses me quite a bit. I thought the defaults logic was implemented in Rust. From [the language guide] I can gather, that it can fill up default values in a very generic way, but how can it know that when Rust prepared the messaged, that "cardCountsSeparateInactive" would default to true?

dae: Protobuf can make messages shorter by skipping keys when they are set to zero/empty, as that's the default value in protobuf. If we look at the message in Python, note how it only shows the non-zero key, but if we explicitly request another key, we do get a value:

```text
>>> from anki.backend_pb2 import GraphPreferences as G
... g = G()
... g.ParseFromString(mw.col.backend.get_graph_preferences())
... print(g)
... print("**")
... print(g.calendar_first_day_of_week)
card_counts_separate_inactive: true

**
0
```

the defaults arg in toObject() is necessary to create an object that has all keys set, even if they are zero, so we can iterate over them. So there are two separate "defaults" here - the logical value we think each key should start with, which is controlled by config.rs - and the zero/empty value that protobuf uses on the wire.

PR #935, opened 2021-01-21
Add manage note types menu item to the sidebar

abdnh:
This adds a "Manage..." menu item for note types in the sidebar as discussed in https://github.com/ankitects/help-wanted/issues/6#issuecomment-764184198

For preselecting the note type, I preferred the general solution of adding a new hook, `models_dialog_will_show`, instead of changing the signature of the Models class. Not sure if there is a simpler solution.

By by the way, what is the reason behind the fromMain parameter of Models's init function that is used for hiding the Fields and Cards buttons when the dialog is opened from the model chooser?

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -1230,6 +1230,7 @@ QTableView {{ gridline-color: {grid} }}
                 ":/icons/notetype.svg",
                 self._note_filter(m.name),
                 item_type=SidebarItemType.NOTETYPE,
+                id=m.id,
             )
             root.addChild(item)
 
--- a/qt/aqt/models.py
+++ b/qt/aqt/models.py
@@ -28,10 +28,17 @@ from aqt.utils import (
 
 
 class Models(QDialog):
-    def __init__(self, mw: AnkiQt, parent=None, fromMain=False):
+    def __init__(
+        self,
+        mw: AnkiQt,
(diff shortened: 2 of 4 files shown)


dae:
It's 8 years ago now so I don't remember specifically :-) But my guess is that the non-main actions may not expect an existing note type to change structure/trigger a one way sync. Since nobody's asking for the fields/cards functionality, I'd be tempted to leave it in rather than risking a regression.

The hooks approach is not my favourite I'm afraid - while there are some parts of the Anki code that use hooks, they are a bit of a hack when used internally. The select() function you've implemented in sidebar.py requires knowledge of models.py's internals, and we get no warning if something like 'dialog.models' is mistyped. I think if Models.__init__() accepted a selected_notetype_id=Optional[int] it would be clearer - the logic that handles that dialog stays within that file, and sidebar.py doesn't have to think about it.

dae:
(__init__() could store the provided value in a local variable, and then setupModels() could call something like .maybe_select_provided_notetype() which has your select() logic in it)

abdnh:
Done.

abdnh:
setCurrentRow(0) in maybe_select_provided_notetype is unnecessary since updateModelsList does that already. I'll change that.

dae:
Thanks Abdo, that's nice functionality to have.

PR #936, opened 2021-01-21
Leave resizing of editor top buttons to CSS

hgiesel:
Right now some `adjustFieldsTopMargin`, makes sure that the editor toolbar doesn't overlap with the first fieldname label. This can easier be solved by moving `#fields` and `#topbutsOuter` into a common div, and making `#topbutsOuter` sticky.

The event listener for clicking images was never triggered, because the `.className` property of fields will yield "field clearfix".


dae:
Thanks Henrik, it's great to see this code getting cleaned up.

PR #937, opened 2021-01-21
Clarify comments in search.ftl

RumovZ:
I felt this was necessary after seeing the first transaltions on Pontoon.


dae:
Thanks - it can be difficult for translators without context, and I might add a few more examples to these as well.

dae:
Well, one at least. :-)

PR #938, opened 2021-01-22
Small refactor of topbar buttons

hgiesel:
1. `flex-wrap` + `justify-content: space-between` solves the issue of wanting both left-floating and right-floating items. There's a small change in behavior here: once the buttons split up to two rows, the right buttons will also be left aligned. The old behavior could be recreated with `margin-left: auto`, but I think this behavior is somewhat more natural.

2. I changed the spacing individual buttons to be done using margins, instead of spaces between the `button` elements. A much better way to do that would have been `row-gap`, but it is not yet supported in the underlying chromium version, but I still think a more verbose CSS solution is better than the implicit HTML solution.

3. When e.g. the bold button is highlighted, it will shift the topbar ever so slightly. I added a negative margin to avoid this behavior.

### DIFF
--- a/qt/aqt/data/web/css/editor.scss
+++ b/qt/aqt/data/web/css/editor.scss
@@ -33,7 +33,11 @@ body {
     margin: 0;
 }
 
-#topbutsOuter {
+#topbuts {
+    display: flex;
+    flex-wrap: wrap;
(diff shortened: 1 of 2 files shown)


dae:
Looks good, thanks!

PR #939, opened 2021-01-22
Render card templates as trees in the sidebar

abdnh:
See https://github.com/ankitects/help-wanted/issues/6#issuecomment-765170983

Reusing the note type icon and not caring about saving collapse state for now.

I hacked together an icon with my horrible Inkscape skills. See https://gist.github.com/abdnh/0c9b19e8b78358e1ea87156d13e61174

I can add it If you think this is palatable, or maybe someone can contribute a nicer one.
I think a new icon will be more useful when the search bar gets implemented, as that will make distinguishing note types and templates easier if we will flatten all entries (even though template names are usually predictable enough to make them distinguishable).

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -1224,14 +1224,25 @@ QTableView {{ gridline-color: {grid} }}
 
     def _modelTree(self, root) -> None:
         assert self.col
-        for m in self.col.models.all_names_and_ids():
+
+        for nt in sorted(self.col.models.all(), key=lambda nt: nt["name"].lower()):
             item = SidebarItem(
-                m.name,
+                nt["name"],
                 ":/icons/notetype.svg",
-                self._note_filter(m.name),
+                self._note_filter(nt["name"]),
                 item_type=SidebarItemType.NOTETYPE,
-                id=m.id,
+                id=nt["id"],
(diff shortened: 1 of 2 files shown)


RumovZ:
I like your icon. Though I think it would look cleaner if you avoided those extremely narrow areas by either shortening the inner lines or connecting them to the outer ones. Another consideration would be to choose a wide format which flashcards are generally more associated with.
By the way, does anybody know what the icon for note types is supposed to depict? To me, it always looked like a bar chart.

abdnh:
I tweaked the icon a bit. Thanks for the suggestion!
I also wonder what the note type icon means.

hgiesel:
I actually think the note type icons are the fields of the editor. The black parts are name labels, the big white parts are the text area fo the field.

RumovZ:
@hgiesel, I don't really see it but it would make sense. 😅

dae:
Yeah, they were intended to represent fields ^_^;  If someone wants to contribute a better icon it would be welcome.

Thanks for this Abdo! Your icon looks pretty good to me. The idea I had was to have "{{ }}" or "{ }" for templates, but I don't know if that would look any good, or whether it would be clear to many users.

PR #940, opened 2021-01-22
Use scaleSqrt for calendar review values

hgiesel:
Here is before:

Screenshot 2021-01-22 at 22 36 51

and here is after:

Screenshot 2021-01-22 at 22 32 14

When reviewing, it is more likely to have some outlier days, where you do a lot of reviews, usually catching up after having let loose for a while. At the moment, these days are a bit overrepresented in the calendar graph. In the first picture, you can basically only make out three days as outstanding, and the others kind of blend into each other. You wouldn't think that some of those have 20 reviews, whereas others have 200, because they have the same color.

Using `scaleSqrt` devalues these higher outlier days, which I think makes sense for the graph.


dae:
Sounds reasonable - thanks!

PR #941, opened 2021-01-23
Fix RangeBox RevlogRange defaulting to All instead of Year

hgiesel:
That addresses what you mentioned in #934.
It was just a flipped ternary, but I moved it into a helper function avoid the duplicated logic.


dae:
Thanks Henrik :-)

PR #942, opened 2021-01-23
Correct browse help link

Arthur-Milchior:
I found other broken link in 2.1.38 but it seems they were repaired since. I went over all call to openHelp, recursively, and did not find other broken link.

In my opinion, one thing that could help avoid loosing time to find all help link could be to replace `help: str` to `help: HelpSuffix`, a new class which would just encapsulate a string but and would allow to find and test all links easily if needed. As often, I'm available to implement the suggestion I give


dae:
Thanks Arthur! Switching to an enum would be a positive change if you are interested in doing that.

PR #943, opened 2021-01-25
Add comment explaining why sfld column has type integer

kerrickstaley:
The `notes.sfld` column in the `collection.anki2` database has type `integer`, even though it contains text. This is [something SQLite allows](https://dba.stackexchange.com/questions/106364/text-string-stored-in-sqlite-integer-column) and was done to allow integer values to sort numerically. Add a comment explaining this.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -73,6 +73,8 @@ rye761 <ryebread761@gmail.com>
 Guillem Palau Salvà <guillempalausalva@gmail.com>
 Meredith Derecho <meredithderecho@gmail.com>
 Daniel Wallgren <github.com/wallgrenen>
+Kerrick Staley <kerrick@kerrickstaley.com>
+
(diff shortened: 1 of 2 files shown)


dae:
It was actually a deliberate choice for the reason you mentioned - when fields contain only numbers, they sort in the order users expect. It's been like that for about 10 years without causing issues. :-)

kerrickstaley:
@dae gotcha. I updated the PR to undo the code change and instead add a comment explaining this design decision.

dae:
Happy to accept that thanks, but would you mind force pushing to remove the first commit first?

kerrickstaley:
@dae done

dae:
Thanks Kerrick!

dae:
Whoops, looks like there's a format test failing?

dae:
Thanks :-)

PR #944, opened 2021-01-25
NF: HelpPage in an enum

Arthur-Milchior:
Hopefully, this can help with updating on next manual update and maybe even linking to manual translation


dae:
Hi Arthur,

There are probably some add-ons that are passing a string into the help argument, so this change will break them if we try to call .value on a provided string. The easiest way to solve it is probably to have these routines take a Union[str, HelpPage], and then only call value if isinstance(help, HelpPage)

Arthur-Milchior:
Changed

dae:
Thanks Arthur!

PR #945, opened 2021-01-25
Normalize dyndeck search and handle exception

RumovZ:
I was aware that the filtered deck search needed some attention to profit from the newer search features, but I only just learned from the forum that search exceptions aren't handled at all here currently.
That's why I wanted to push this out quickly, even though I might soon make another PR to switch the code to the new backend functions.


dae:
Thanks Rumo!

PR #946, opened 2021-01-26
Correctly type createEventDispatch and dispatch

hgiesel:
In #897 you suggested to add a new class, however `dispatch` and `createEventDispatcher` [actually are typed](https://github.com/sveltejs/svelte/blob/55b11b0664ba176cc7af3f7fdeb9dac00a2bc6e6/src/runtime/internal/lifecycle.ts#L30), however I forgot about adding them in the end. In this PR, you can see how that would look like.

An alternative would be to use a class, which would look something like this:

```javascript
class Handler {
    searchAvailable(): boolean {
        return $browserLinksSupported;
    }

    search(innerQuery: string) {
        const query = `${search} ${innerQuery}`;
        bridgeCommand(`browserSearch:${query}`);
    }
}
```

I refrained for now, to show you, I guess, the "svelte-way".
If we resorted to using a class, there wouldn't even be a need for `dispatch`. The event is dispatched from the component, where `dispatch` was initiated. That's why for example the Interval graph does not need to bubble the search event from the inner Histogram graph, because the event dispatcher is initiated inside of the Intervals graph anyway.

If we just pass in the object, which does the search, then there's no need for dispatch either. I don't have a strong inclination for either solution, but I thought I'd show you at least.

### DIFF
--- a/ts/graphs/AddedGraph.svelte
+++ b/ts/graphs/AddedGraph.svelte
@@ -1,6 +1,6 @@
 <script lang="typescript">
     import { RevlogRange, GraphRange } from "./graph-helpers";
-    import type { TableDatum } from "./graph-helpers";
+    import type { TableDatum, SearchEventMap } from "./graph-helpers";
     import type { I18n } from "anki/i18n";
     import type { HistogramData } from "./histogram-graph";
     import { gatherData, buildHistogram } from "./added";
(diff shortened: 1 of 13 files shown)


dae:
Thanks, I didn't know about that. This change gets us code completion/checking which was the main thing I was interested in. If we were doing it from scratch I might opt for the handler approach, but that's probably just my lack of experience with event dispatch talking, and this code works and already exists, so happy to run with it for now.

PR #947, opened 2021-01-26
No clicking on empty bins in the Histogram

hgiesel:
This prevents triggering searches from empty bins in the several histograms. I already had this for the Calendar, but I think it also makes sense for the histogram.


dae:
Sounds good - thanks!

PR #948, opened 2021-01-26
Replace individual d3 libraries with encompassing d3

hgiesel:
What do you think about replacing the individual d3 libraries with just `d3` and `@types/d3`? I think the biggest advantage would be that the types are more likely to harmonise, which isn't the case right now. It might also not have that much of an impact on the code size, as rollup has pretty good treeshaking.

I would use this PR to:

1. Make graphs type check against the `d3` lib
2. Update graphs if there are any incompatibilities against the newer versions.


hgiesel:
I just noticed that you published a Beta for 2.39, in which case I'm happy to delay this until 2.40, or even later if there are emergent bug fixes.

dae:
They moved to making things more modular in version 4, so I got the impression that was the preferred way - but maybe that's not the case? If it improves typechecking that would be a strong argument in favour of it, though I would like to know how much of a hit it would be to the bundle size. On current master:

```
% bazel build -c opt ts/graphs && du -hs bazel-bin/ts/graphs/graphs.js
INFO: Invocation ID: e5bc90cb-4d6a-4e8a-8dc1-c769259d20f6
INFO: Analyzed target //ts/graphs:graphs (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //ts/graphs:graphs up-to-date:
  bazel-bin/ts/graphs/graphs.js
INFO: Elapsed time: 0.279s, Critical Path: 0.02s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
240K	bazel-bin/ts/graphs/graphs.js
```

The graphs all show on one screen, so (famous last words) I think regressions should be fairly easy to spot, and thus we can probably slip this in before a release if it doesn't take too long.

hgiesel:
Just made it compile for the first time with the `d3` package, and it seems like rollup is doing its job:

```
232K    bazel-bin/ts/graphs/graphs.js
```

hgiesel:
Okay I think there are no bugs left.
I only added very few types, this PR is mostly just supposed to switch the library.

There were a few changes to the d3 library that I had to adapt:
* `mouse` function was replaced with `pointer` which takes different arguments
* `selection.on` has a different function signature, without `this` and `index`.

I also changed the `add.sh` to a more general `yarn.sh` as I used for uninstalling the packages.

dae:
Thanks Henrik, great to see we can make things simpler and actually save a few Ks!

PR #951, opened 2021-01-26
Add search bar to the sidebar

abdnh:
See https://github.com/ankitects/help-wanted/issues/6

My first idea was to use [QCompleter](https://doc.qt.io/qt-5/qcompleter.html) by setting the sidebar as its pop-up, but it seems the stored SidebarItem objects get lost this way (internalPointer returns None). So I just wrote the matching code manually.

The tree is flattened and the full names of the items are shown for clarity.

I also added the shortcut <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>B</kbd> to focus the search bar.

image
image
image

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -35,7 +35,7 @@ from aqt.main import ResetReason
 from aqt.previewer import BrowserPreviewer as PreviewDialog
 from aqt.previewer import Previewer
 from aqt.qt import *
-from aqt.sidebar import SidebarTreeView
+from aqt.sidebar import SidebarSearchBar, SidebarTreeView
 from aqt.theme import theme_manager
 from aqt.utils import (
     TR,
@@ -912,6 +912,17 @@ QTableView {{ gridline-color: {grid} }}
         self.sidebar = SidebarTreeView(self)
         self.sidebarTree = self.sidebar  # legacy alias
         dw.setWidget(self.sidebar)
+        self.sidebar.searchBar = searchBar = SidebarSearchBar(self.sidebar)
+        qconnect(
+            QShortcut(QKeySequence("Ctrl+Shift+B"), self).activated,
+            self.focusSidebarSearchBar,
+        )
+        l = QVBoxLayout()
+        l.addWidget(searchBar)
+        l.addWidget(self.sidebar)
+        w = QWidget()
+        w.setLayout(l)
+        dw.setWidget(w)
         self.sidebarDockWidget.setFloating(False)
 
         self.sidebarDockWidget.setTitleBarWidget(QWidget())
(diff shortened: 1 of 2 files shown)


dae:
Hi Abdo,

Looks like this works well on most content!

Unfortunately with the med school decks, it's hard to see what is being matched unless the user expands their sidebar quite wide:

Screen Shot 2021-01-27 at 8 41 49 pm

AnkiMobile has a similar search bar, but it's displayed full screen, so I didn't think about the width of the sidebar being as much of an issue.

I've run out of time today, but I might poke about a bit tomorrow and see what I come up with - will get back to you.

abdnh:
> Unfortunately with the med school decks, it's hard to see what is being matched unless the user expands their sidebar quite wide

True. As an extreme example, I tested a bit with that monstrous Anking deck from the other day after submitting this.
Expanding the sidebar automatically and highlighting the relevant text will be nice. I'll look into it too.

dae:
I've pushed some experimental changes using a filter proxy - what do you think? It still needs some polish and I've listed a few things that still need doing in https://github.com/ankitects/anki/pull/951/commits/8b08687b0ca8eb8ecbcb7fc20fcd6d3a994ed31b - but the highlighting should be working now.

abdnh:
Looks great!
I spent hours skimming through the Qt docs trying different classes and didn't notice QSortFilterProxyModel :(
I'll push a change to make the matching work with `full_name` instead of `name`, as this will allow us to search for `a::` and get all tags under `a` for example.

dae:
Thanks Abdo! This already looks useful so I'll merge it in, but it would still be nice to handle the "change while searching" case, eg deleting a card while a search is performed currently refreshes the list but keeps the search text at the top - we should either repeat the search with the new root, or at least clear the search bar. There's no actual use in refreshing the list after deleting cards/notes, but that's a separate issue that needs fixing :-)

dae:
... whoops, looks like you beat me to it!

dae:
Looks good - thanks for your work on this!

abdnh:
> ... whoops, looks like you beat me to it!

Yeah, I was about to write a comment :)

Thank you too for your time. (My work on this was mostly unproductive probably)

dae:
Ah, just noticed the change to full_name has caused a problem with the highlighting - if you search for something that appears only in a parent tag, then all the children get highlighted as well. Changing the highlighting to use name instead of full_name fixes that case, but stops it working if the search includes ::

I wonder how much the need to search for '::' will be needed in practice?

hgiesel:
Something I just noticed while testing just now: In night mode, the highlighted items are almost impossible to read, as its white text on yellow background.

abdnh:
Oops.


> I wonder how much the need to search for '::' will be needed in practice?

Maybe not much useful for sufficiently unique names, which I think is usually the case for decks and tags at least.
I'm with reverting this change if you want.

dae:
Ok, I've reverted it for now - if it turns out people do need this, then we can come back to it :-)

PR #952, opened 2021-01-26
Deal with inline content vs block content and <br> in editor.ts

hgiesel:
I tried out the 2.1.39Beta1 released, and unfortunately noticed a bug in the editor, with the behavior of `<br>`. That occurs, when you have an empty field, you add some content `Content`, now using "Ctrl-Shift-X" will show `Content`. If you create a newline and remove it again, and use "Ctrl-Shift-X" again, you will see that it now reads `Content<br>`. The only way to get rid of this `<br>` will be now to remove it in the HTML editor.

This to some degree already existed in previous versions, in a different form: The field would read `Content`, but clicking "Ctrl-Shift-X" would reveal `<div>Content</div>`, causing the same problem.

I had to think a bit about how to solve this once and for all, and also what it even means to solve it. I came to the following conclusion: Sometimes fields have a role in the template, which is more akin to inline elements, whereas sometimes you want them to fulfil a role more akin to block elements.

For example, in the picture below, the fields "left" and "left_back" are more inline in nature, whereas "bottom_back" is more like a block element.

Screenshot 2021-01-26 at 23 13 50

The user doesn't have to make a conscious decision here, but he will make it unconsciously by what he fills into the field. If he puts in a block element like a list, it can only be used as a block. If he only puts in text, or inline elements like `span`, or `i`/`a`/`u` it can be used as both inline, or as a block.

The contenteditable element treats its content more like it's a block by default, and especially in our case because the editor fields are `div`s. To make it suitable to being an inline element, we have to modify it, and this modification is stripping away *one* trailing `<br>` if there is one. For example, if the user text field shows:

```
This is the first line of content
<empty line>
```

What we would expect from it is `This is the first line of content<br>`, however what's actually in the div is `This is the first line of content<br><br>`. In the template, this will not generate what the user expects, as he might use it in an inline fashion, and expects it to have one trailing newline. So we have to modify it, before we send it to Python via `pycmd("key/blur:...")`.

However in block scenarios, we don't want to strip them away. E.g. `<ul><li>Item 1</li><li>Item2</li></ul><br><br>` renders two full newlines after the unordered list. Stripping away one `<br>` would change the content. However in this case we can detect that `ul` is a block element, and leave it intact.

The list of inline elements is taken [from here](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements).

This also contains some other changes like:
1. Don't use table for editor fields anymore, but use a flatter layout.
2. Don't create fields as one big chunk of HTML, but prefer `createElement`. This avoids losing type safety.
3. Avoid using jQuery in some cases.
4. Instead of tables and clearfix, use flex (just like topbars). Here we also have the caveat of not being able to use the `gap` property again, so we resort to margins.

### DIFF
--- a/qt/aqt/data/web/css/editor.scss
+++ b/qt/aqt/data/web/css/editor.scss
@@ -1,6 +1,12 @@
 /* Copyright: Ankitects Pty Ltd and contributors
  * License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html */
 
+#fields {
+    display: flex;
+    flex-direction: column;
+    margin: 5px;
+}
+
 .field {
     border: 1px solid var(--border);
     background: var(--frame-bg);
--- a/qt/aqt/data/web/css/editor.scss
+++ b/qt/aqt/data/web/css/editor.scss
@@ -5,6 +5,18 @@
     display: flex;
     flex-direction: column;
     margin: 5px;
+
+    & > * {
+        margin: 1px 0;
+
+        &:first-child {
+            margin-top: 0;
+        }
+
+        &:last-child {
+            margin-bottom: 0;
+        }
+    }
 }
 
 .field {
--- a/qt/aqt/data/web/css/editor.scss
+++ b/qt/aqt/data/web/css/editor.scss
@@ -14,12 +14,6 @@
     }
 }
 
-.clearfix::after {
-    content: "";
-    display: table;
-    clear: both;
-}
-
 .fname {
     vertical-align: middle;
     padding: 0;
--- a/qt/aqt/data/web/js/editor.ts
+++ b/qt/aqt/data/web/js/editor.ts
@@ -109,6 +109,10 @@ function nodeIsElement(node: Node): node is Element {
     return node.nodeType === Node.ELEMENT_NODE;
 }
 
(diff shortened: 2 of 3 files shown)


hgiesel:
I wanted to do another example to show the different behaviours:

image

`Field1` and `Field2` show what I mean by "using it in an inline fashion (first example)", or "using it in a block fashion (second example)".

Example 3 shows what an additional `<br>` does with the users content, if he uses it in an inline fashion: it adds another newline (unsurprisingly), which might break the layout.

Example 4 shows content which only has inline elements (text nodes and `<u>` element) used in a block fashion: having that additional `<br>` doesn't change the way it is displayed, so truncating it when saving the editor field to the note wouldn't cause no change.

Example 5 shows content with block elements (`<ul>`) used in a block fashion. Truncating that last `<br>` DOES have a layout-changing effect, so we couldn't do that in this case.

This covers all the cases, as you cannot put block elements into inline elements.

dae:
My initial impression was "this sounds like a hack", but then there's not much about contentEditable that isn't. Your logic seems like it makes sense, and I hadn't had any luck solving this issue in the past, so happy to have another pair of eyes on it. Thanks Henrik.

PR #954, opened 2021-01-27
Sample deck

RumovZ:
The goal is to provide a sample deck to new users so they can start discovering and learning about Anki immediately. As of yet, new users and those upgrading from a prior release will be shown a brief welcome message with the option to create the sample deck. The deck itself will then be created on the backend using fluent translations.

There is still a lot of work to be done (definitely nothing for 2.1.39) but I'd like to know whether you approve of this feature and the general direction of its implementation.

### DIFF
--- a/ftl/core/sample-deck.ftl
+++ b/ftl/core/sample-deck.ftl
@@ -1,0 +1,28 @@
+### Text used in the sample deck.
+
+shall-create-sample-deck = Welcome to Anki, a flashcard program that uses spaced repetition to let you study in the most efficient way! Do you want Anki to create a sample deck with some introductory information?
+created-sample-deck = The sample deck has been created. Click its name “Anki Basics” to begin learning about Anki.
+could-not-create-sample-deck = A deck of the name “Anki Basics” already exists. You must delete or rename it before you can create the Anki sample deck.
+# Must be a valid deck name.
+sample-deck-name = Anki Basics
+sample-notetype-name = Anki Basics
+sample-template-name = Card 1
+sample-field-question = Question
+sample-field-answer = Answer
+sample-field-details = Details
+sample-field-manual = Manual section
+
+## Flashcard text for the sample deck.
+
+sample-question-card = What is a <b>card</b> in Anki?
+sample-answer-card = A question and answer pair.
+sample-details-card = Just like a paper flashcard with a question on one side and the answer on the back. You're looking straight at one!
+sample-question-deck = What is a <b>deck</b> in Anki?
+sample-answer-deck = A group of cards sharing certain settings and usually studied together.
+sample-details-deck = For example right now, you're studying the “Anki Basics” deck. Decks can also contain other decks.
+sample-question-note = What is a <b>note</b> in Anki?
+sample-answer-note = A collection of related pieces of information from which Anki creates your flashcards.
+sample-details-note = With paper flashcards, you would write each card individually, even if some of them share certain content. In Anki, you only have to enter information once. You can then tell Anki how many and what kind of cards you want to be created from it.
+sample-question-field = What is a <b>field</b> in Anki?
+sample-answer-field = A piece of information belonging to a note.
+sample-details-field = For example, the note from which this card was created has four fields: a question, an answer, some additional information and a link to the manual.
--- a/ftl/qt/qt-misc.ftl
+++ b/ftl/qt/qt-misc.ftl
@@ -14,6 +14,7 @@ qt-misc-closing = Closing...
 qt-misc-configure-interface-language-and-options = Configure interface language and options
 qt-misc-copy-to-clipboard = Copy to Clipboard
 qt-misc-create-filtered-deck = Create Filtered Deck...
+qt-misc-create-sample-deck = Create Anki Basics Deck
 qt-misc-debug-console = Debug Console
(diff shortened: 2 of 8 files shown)


dae:
While you've nicely integrated this, and I know it's been requested more than once in the past, I'm afraid I'm not a huge fan of the idea of using a deck to teach people how to use the program.

My primary concern is that I think different approaches are needed for teaching new material vs remembering already-learnt concepts. Teaching may require multiple paragraphs to get a new point across, and perhaps additional media such as videos or screenshots. But once a concept has been learnt, subsequent reviews are most effective if that information is condensed into bite-sized pieces - cards that contain multiple paragraphs of text or videos are not ideal. Anki currently is focused on the memorization/retention stage, and assumes that the initial learning will either be done externally, or the material being entered will be simple associations that do not require verbose explanations.

If we create a sample deck with long-form text explaining things to the user, I fear we would be sending them a signal that such long cards are the way they should be using Anki when adding their own material, when that is not the case. And if the sample deck kept things brief to show how cards are intended to be constructed, then it would be less useful as a learning aid.

Another concern is that making good documentation is hard already, and trying to word it in a question/answer format and then embed it in the ftl/code files makes it harder still.

One middle-ground I can think of between a deck-based guide and the existing manual might be to have an intro wizard that displays some information on first Anki startup. That would get you translations in different languages, and would allow for long-form text without encouraging its use on cards. It would still be a fair amount of work to produce to a good standard however, and it partially duplicates the role of the existing manual and intro videos. To make it more user friendly, we'd need the ability to overlay pointers to relevant interface elements and guide the user through steps, and I suspect getting that all going is a bigger project than just creating some updated intro videos and showing them on first startup. I've been meaning to update a number of the intro videos for quite a while, it's just a matter of finding the time :-(

RumovZ:
Thank you for the detailed reponse.

Maybe I should start with the advantages I see in this approach. My idea of this feature is to:

1. Welcome the user and indicate a way to go forward. My experience is that new users can feel lost and quickly get frustrated because Anki is so overwhelming when you see the interface the first time.
2. Show where to get help. Every card has a link to the according manual section and I was going to create separate cards regarding the manual and the forum.
3. Give something to toy around with. Users can experiment with the sample cards without having to create something first or worrying about messing up their personal decks.
4. Make aware of Anki's complexity. It seems, some users manage to capitalise a lot on Anki's most basic features. But at some point, they want to do more and suddenly everything stops making sense (most prominent: the concept of cards vs. notes). Then, they hopefully remember seeing something in the sample deck and have an entry point to find out more.
4. Showcase some features to keep the user engaged and make him aware of options he might later have need for.
5. Provide some basic internal documentation that is available offline and (hopefully at some point) in many more different languages than the manual is.
<hr>

> My primary concern is that I think different approaches are needed for teaching new material vs remembering already-learnt concepts.

I agree with this assessment and I don't want the sample deck to replace any existing documentation. On the contrary, my hope is to make more people engage with the manual. Comprehensive as it is, very few people have the patience to read it first, but I think if we present selected portions of it, it would be far less intimidating.

> Another concern is that making good documentation is hard already, and trying to word it in a question/answer format and then embed it in the ftl/code files makes it harder still.

I agree with this one, too. I actually tried pasting directly from the manual but it didn't always work and I also wanted to make use of the interactive element.

> One middle-ground I can think of between a deck-based guide and the existing manual might be to have an intro wizard that displays some information on first Anki startup.

I thought about that, too. But it doesn't give all of the advantages listed above and would be a lot more work I assume.

<hr>

Since most of your concerns address the content of the sample deck, another way could be to give new users the option to automatically import one of a few preselected decks. As an example, the Ultimate Geography deck could be of interest to a lot of users. However, that would mean being dependent on external decks and less options to localise.

<hr>

In conclusion, I agree with all of your concerns but I think with the right implementation, the advantages would outweigh them by a lot.

dae:
Really old versions of Anki actually included some sample decks, but that went away when shared decks became available on AnkiWeb. We had a sort of wizard back then too.

Screen Shot 2021-01-30 at 3 25 00 pm
Screen Shot 2021-01-30 at 3 26 02 pm
Screen Shot 2021-01-30 at 3 26 50 pm
Screen Shot 2021-01-30 at 3 27 12 pm

I moved away from that partly because more and more things needed to fit on the screen and partly because it was extra work to implement as the interface evolved. The interface tended to be wordier as a whole, but I tried to move away from that in favour of more succinct wording, as users need to learn once, but use many times, and being able to quickly skim the screen makes it easier to navigate.

I agree that Anki doesn't do a great job of welcoming new users, but wouldn't a popup on first startup that linked to or embedded a few getting started intro videos go a long way to solving that? The subtitles are translatable, so it wouldn't mean only covering an English-speaking audience. I think the average user will find it easier to follow along with a video showing exactly what needs to be done, compared to reading text and extrapolating.

RumovZ:
A welcome popup would indeed solve many problems, escpecially if it provides an explicit tutorial (or several) to get started. It could offer users the choice to learn from the manual or from videos. We could even embed dubbed videos if they come in good quality, because the problem with subbed tutorials is that the interface is still in English.

RumovZ:
I realised I wrote "dubbed" above, but of course the videos would need to be re-enacted. Anyway, shouldn't be too much trouble for willing translators.
I also like the wizard. Only, it's probably be really hard to implement without complicating the codebase.

PR #955, opened 2021-01-28
More backend search

RumovZ:
A bunch of minor adjustments regarding various instances of search code:

- Add new backend filters.
- Switch literal searches to use filters from the backend.
- Add helpers to Collection so aqt can use col instead of backend when handling search strings.
- Prevent search prompt from being saved as a filter.
- Add helper in mw to perform a browser search with col.search_string(). (Hope that's the right place for it?)
- Handle search exception in dupes dialogue.

Regarding literal searches in the Python code, that should be pretty much it. Next would be the Typescript part, but I haven't looked into it yet and don't know how feasible that is. But it would be nice to have pretty search errors in the stats screen, too.

As always, please let me know if anything's off.
One more thing: How would you feel about removing the search prompt? It permanently needs special treatment in the browser code and in my opinion, it's doing more harm than good GUI-wise.


pylib/anki/collection.py
```diff
@@ -467,6 +481,72 @@ def findDupes(self, fieldName: str, search: str = "") -> List[Tuple[Any, list]]:
...
+        name: Optional["FilterToSearchIn.NamedFilterValue"] = None,
+        tag: Optional[str] = None,
+        deck: Optional[str] = None,
+        note: Optional[str] = None,
+        template: Optional[int] = None,
+        dupe: Optional[Tuple[int, str]] = None,
+        forgot_in: Optional[int] = None,
+        added_in: Optional[int] = None,
+        due_in: Optional[int] = None,
+        nids: Optional[List[int]] = None,
```
dae, line 507: These _in feel a bit "half-way" - how about either using just "forgot", "added" etc (matches the other short terms), or the full "forgot_in_days", "added_in_days", etc?

dae (commented):
Code looks good, just some comments on wording below.

I'd be ok with removing that prompt, it complicates the code, and we may as well just search for deck:current automatically when the user has is not studying. It might be nice to preserve the nid:xxx search when they're studying a particular card - the primary reason for that is so that users don't accidentally reveal the answer text to other cards they may study later in the session.

dae:
Re showing an error in the graphs, that's now done. Normalizing the text and/or using helpers to build search queries in the graphs is probably not worth the effort IMHO, though if you wanted to do the former it could perhaps be done by sending it back in GraphsOut.

pylib/anki/collection.py
```diff
@@ -467,6 +481,72 @@ def findDupes(self, fieldName: str, search: str = "") -> List[Tuple[Any, list]]:
...
+    # Pass multiple to concatenate (defaults to 'and').
+    # Pass 'negate=True' to negate the end result.
+    # May raise InvalidInput.
+    def search_string(
+        self,
+        *,
+        negate: bool = False,
+        concat_by_or: bool = False,
+        searches: Optional[List[str]] = None,
+        name: Optional["FilterToSearchIn.NamedFilterValue"] = None,
```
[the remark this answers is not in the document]
RumovZ: I realised, too, that that the name is not that meaningful because in the end, they are all "named" filters. The unique trait of the ones in the enum is that they don't take arguments, but I couldn't come up with a concise name conveying that.
We could actually use them directly in FilterToSearchIn's oneof, I think. But that would mean a whole lot of kwargs more to handle for the frontend. What do you think?

pylib/anki/collection.py
```diff
@@ -467,6 +481,72 @@ def findDupes(self, fieldName: str, search: str = "") -> List[Tuple[Any, list]]:
...
+    # Helper function for the backend's search string operations.
+    # Pass search strings as 'searches' to normalize.
+    # Pass multiple to concatenate (defaults to 'and').
+    # Pass 'negate=True' to negate the end result.
+    # May raise InvalidInput.
+    def search_string(
+        self,
+        *,
+        negate: bool = False,
+        concat_by_or: bool = False,
```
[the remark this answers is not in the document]
RumovZ: Sounds better, but both are builtin names and I try to avoid those. I'll try to come up with something similar but unique.

pylib/anki/collection.py
```diff
@@ -467,6 +481,72 @@ def findDupes(self, fieldName: str, search: str = "") -> List[Tuple[Any, list]]:
...
+        searches: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
RumovZ: What about `search_strings`? A bit lengthy, but it conveys that these arguments are often output of `search_string()` themselves.

pylib/anki/collection.py
RumovZ: I'll go with the full names then to make it clear how the arguments are interpreted.

RumovZ:
> It might be nice to preserve the nid:xxx search when they're studying a particular card

You're talking about the `is:current`, correct? I'm wondering if it has to be implemented using a literal. Do users actually type it in? Else, we could just pass a nid search directly.

> Normalizing the text and/or using helpers to build search queries in the graphs is probably not worth the effort IMHO, though if you wanted to do the former it could perhaps be done by sending it back in GraphsOut.

Thanks for the pointer. Maybe a more useful feature would be to open the stats screen using the current browser search. But I already tried doing that for filtered decks and ran into problems. The filtered deck dialogue would need to be handled differently. I'll make a PR when I'm sure I can do that without any side effects.

dae:
Re some of those comments, just throwing another idea out there - what about moving all the named filters into the oneof as bools, renaming FilterToSearchIn to SearchTerm, and change the signature of search_string to:

```
def build_search_string(self, *terms: Union[str, SearchTerm], *, negate=False, match_any=False)
```

Then something like

```
search = self.mw.col.search_string(searches=[search], name=NamedFilter.DUE)
```

could become
```
search = self.mw.col.build_search_string(search, SearchTerm(is_due=True))
```

We'd avoid the need to manually map each case or have to distinguish between arg/no arg cases. What do you think?

dae:
I'm ok with getting rid of is:current if it makes things simpler - it started out more as an implementation detail than a deliberate way of hiding the underlying search from the user.

pylib/anki/collection.py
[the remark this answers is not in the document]
dae: That works for me, though perhaps we should just make this argument implicit as mentioned below.

RumovZ:
Great idea, I'll do that!

dae:
Some of those named filters could probably go too - if you've got added_in_days, added_today is a bit redundant. And maybe the various flag options could use a separate oneof flag=FlagKind referencing an embedded enum, instead of being listed out individually? Sorry, I'm bikeshedding now. :-)

RumovZ:
The refactoring should be done. Additional changes:

1. Search helpers such as browser_search() and find_notes() can take SearchTerms directly and pass them on to build_search_string().
2. I encountered a circular import issue in `anki/find.py` when trying to import `SearchTerm`. I solved it by making `findeDupes()` a method of `Collection`. Is there any special reason for this module to exist in the first palce?

dae (commented):
I'm conscious SearchTerm is going to be a public API, making it difficult to modify once people start using it, so I'm keen to get it polished before the first release. I'll push some changes which I think make it a bit more consistent - what do you think?

dae:
re find.py, that's where the code used to live, and it was left around to try to avoid breaking some add-ons that import from there. findDupes() probably wasn't used frequently compared to Finder however.

Should have said above, but the code looks good - will be a nice improvement. :-)

ftl/core/browsing.ftl
```diff
@@ -83,6 +83,7 @@ browsing-reposition = Reposition...
 browsing-reposition-new-cards = Reposition New Cards
 browsing-reschedule = Reschedule
 browsing-save-current-filter = Save Current Filter...
+browsing-search-bar-hint = Type here and press Enter to search. Leave empty to show whole collection.
```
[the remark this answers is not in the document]
RumovZ: Oh right, that was the plan, just forgot about it. 😄 
I think in most cases, the additional startup time is worth it. It also provides a nice reminder of what the current deck is, which might be useful for following searches, and it's consistent with the implicit search for the current card.

ftl/core/browsing.ftl
```diff
@@ -83,6 +83,7 @@ browsing-reposition = Reposition...
 browsing-reposition-new-cards = Reposition New Cards
 browsing-reschedule = Reschedule
 browsing-save-current-filter = Save Current Filter...
+browsing-search-bar-hint = Search cards/notes
```
RumovZ, line 86: I was actually trying to be verbose because the search bar is so long and I thought less empty space looked better. That might not be a good enough reason, but I also think it we could provide more helpful text. For example: "Type here and press Enter to search for cards/notes."

qt/aqt/addcards.py
```diff
@@ -7,6 +7,7 @@
 import aqt.editor
 import aqt.forms
 import aqt.modelchooser
+from anki.backend_pb2 import SearchTerm
```
RumovZ, line 10: On a previous PR, I got the impression that aqt should not call (and import from?) the backend directly. Is that accurate?
Either way, I'd have all modules import from one place: rsbackend, backend_pb2 or collection.

rslib/backend.proto
```diff
@@ -50,6 +50,10 @@ message NoteID {
   int64 nid = 1;
 }
 
+message NoteIDs {
```
RumovZ, line 53: Looks like this is not needed then? Same for rsbackend.py.

qt/aqt/addcards.py
dae: That should have been from collection - the IDE added it and I didn't notice when reviewing the patch. Good catch :-)

I wish there were a way to mark rsbackend and backend_pb2 as unimportable from outside pylib. A leading _ might help with automatic completion, but we originally had code importing TR from rsbackend, which we'd break by doing that. :-(

ftl/core/browsing.ftl
dae: Yeah, I think I went too far. I was also thinking it might be nice to remove the Search button to add some more room, and to do that we really do need to mention pressing enter.

rslib/backend.proto
dae: Yep, sorry, missed that.

rslib/backend.proto
```diff
@@ -760,37 +764,58 @@ message BuiltinSearchOrder {
...
+    WITHOUT = 0;
+    ANY = 1;
+    RED = 2;
+    ORANGE = 3;
+    GREEN = 4;
+    BLUE = 5;
+  }
+  enum Rating {
+    ANY_ANSWER_BUTTON = 0;
+    ANSWER_BUTTON_1 = 1;
```
RumovZ, line 782: Would it make sense to use the names Again, Hard, Good and Easy? I find the term answer button a bit misleading (on the backend it's called ease, which isn't better).
It would be consistent with flags and help to decouple the internally used indices from the frontend representation. It would be escpecially helpful if we get customisable buttons at some point and Good isn't necessarily the third button anymore.

Of course, it's shorter, too, and unambiguous within its namespace.

dae: Either way works for me - feel free to change it if you'd like. Ease is a holdover from the DB layer and old code; if you want to rename it AnswerButtonKind that would be an improvement.

dae: (or RatingKind, or some other suggestion)

qt/aqt/browser.py
```diff
@@ -194,23 +186,18 @@ def flags(self, index):
...
         self.beginReset()
         self.cards = []
-        exception: Optional[Exception] = None
         try:
             ctx = SearchContext(search=txt, browser=self.browser)
             gui_hooks.browser_will_search(ctx)
             if ctx.card_ids is None:
-                ctx.search = self.browser.normalize_search(ctx.search)
-                ctx.card_ids = self.col.find_cards(ctx.search, order=ctx.order)
+                ctx.card_ids = list(self.col.find_cards(ctx.search, order=ctx.order))
```
RumovZ, line 193: Quick note: I had to explicitly make card_ids a list because it would be passed as a `protobuf.internal.containers.RepeatedScalarFieldContainer` which doesn't support all features lists do.
I have no idea why this suddenly became necessary and if it indicates I did anything wrong.

dae: What was the traceback?

dae: If it's just a typechecking issue, self.cards could probably be declared as a Sequence instead of a List

ftl/core/browsing.ftl
RumovZ: That's a cool idea. I leave the hint as it is then and it can be updated accordingly with the GUI change.

qt/aqt/browser.py
RumovZ: The problem was in browser.py/focusCid() because it doesn't have an index method. But somehow it's working on the master branch.
I didn't get an error because of the universal exception handling.

RumovZ: It might have something to do with calling `model.search` multiple times. On the main branch, current card search is only supported with a closed browser, so `self.model.cards.index(cid)` is only called after the very first search.

dae: Ah, makes sense. That routine will only need to run with a single note shown, so perhaps we could just use list() inside focusCid, and avoid having to build a separate list in other cases.

dae: (the catch-all exception there is also nasty)

RumovZ:
I'm trying to wrap it up, now, but I'm a bit confused about the correct way to handle imports from the backend.
We have `SearchTerm` which is used by collection itself, and `InvalidInput` which isn't.

1. Should collection import from `backend_pb2` or from `rsbackend` where we would have to import it first?
2. Should other modules import `SearchTerm` from collection? If so, what about `InvalidInput`? If we want to import it from the collection, we have to import it there only for the sake of exporting and suppress the pylint warnings for all backend imports. If we just import from the backend, why not do this for `SearchTerm` as well?

dae:
Sorry for the confusion, I haven't been very consistent here, and am still figuring things out.

The original plan was to have rsbackend insulate the rest of the codebase from everything backend related, including the protobuf files, which meant re-exporting any that pylib or other libraries needed to use. But then I realised we don't want external libraries importing from rsbackend, as that's intended to be internal to pylib, so they need to re-export any messages intended for public consumption. But if the items are imported into rsbackend as well, the IDE will offer that as a code completion when working on the Qt code, which is not good.

Before we go any further, I might see if we can make rsbackend and backend_pb2 semi-private - will get back to you.

dae:
Sigh, underscore prefixes don't really help, and editors will still offer to auto-complete the private members.

Take DeckTreeNode, which decks.py should be publicly exporting. Currently it autocompletes to this when there's a parallel _backend module:

Screen Shot 2021-01-31 at 11 40 36 am

We can change the import in decks.py to an assignment, which gets PyCharm to notice the re-export, though now it declares it's a "variable", and the other completions are still there:

```python
DeckTreeNode = anki.rsbackend.DeckTreeNode
```

Screen Shot 2021-01-31 at 11 42 08 am

Changing the assignment in rsbackend.py to an import makes it go away, but the backend_pb2.py match remains:

```
#DeckTreeNode = pb.DeckTreeNode
from anki.backend_pb2 import DeckTreeNode
```

Screen Shot 2021-01-31 at 11 46 43 am

Given how Python and the IDE is not helping us enforce privacy here, even from ourselves, I'm tempted to drop the extra import+rexport step in rsbackend, and have the pylib code import from _pb2 directly, re-exporting messages intended for public consumption. This would make it consistent with the Rust side, which does not try to import+re-export items from the generated protobuf code.

dae:
This _backend refactor looks like it's going to be a bit of a pain. To minimize merge conflicts I think we should sort out the conflicts on this and merge it in, and work from there - will do this now.

dae:
Righto, all merged - please let me know if something got missed, there were a few conflicts to deal with. Might take me a while to work through a move of things into _backend + public exports.

PR #957, opened 2021-01-29
Implement editor as Web Component

hgiesel:
Implements the editor fields as web components, putting the main editing area inside of a shadow root.
This isolates the HTML in the editing area from the rest of the editor. For this new editing area, I created a new file `editing-area.scss`. What this also means, is that the CSS from the editor cannot affect the editing area anymore, which allows us for more possibilities going forward.

This allows for a cool feature, I already implemented: Python now instantiates the editor with the CSS from the note type. So if the user defines some general CSS, for example `b { color: red }`, then this CSS will also translate to the editor. This will make the editor fields look more like how they will end up looking in the reviewer.
This is somewhat implemented in several add-ons, like [this one](https://ankiweb.net/shared/info/1215991469), or [this one](https://ankiweb.net/shared/info/1899278645), which I currently maintain.

The editing area is now also wrapped in a custom-named `<editing-area>`. This means that rogue `</div>`s won't close the editing area anymore, like [what happened here](https://forums.ankiweb.net/t/words-bugging-when-i-search-and-replace/6916/5).

Other features:
1.  I implemented is avoiding rebuilding the editor fields, but adjust the amount of editor fields via `adjustAmountOfFields`, and then update their contents, which will avoid blinking when `saveNow` is executed.
2. The `#dupes` section is now sticky, and centered, making it stand out a bit more. Having worked with note types with a lot of fields for some time now, I somewhat forgot it even existed.
3. The highlight for buttons in dark mode is now white (which was black for, being very hard to see).

### DIFF
--- a/qt/aqt/data/web/css/editable.scss
+++ b/qt/aqt/data/web/css/editable.scss
@@ -1,4 +1,4 @@
-editing-area {
+anki-editable {
     display: block;
     overflow-wrap: break-word;
     overflow: auto;
--- a/qt/aqt/data/web/css/editing-area.scss
+++ b/qt/aqt/data/web/css/editing-area.scss
@@ -9,3 +9,11 @@ editing-area {
         white-space: pre;
     }
 }
+
+img.drawing {
+    zoom: 50%;
+
+    .nightMode & {
+        filter: unquote("invert() hue-rotate(180deg)");
+    }
+}
--- a/qt/aqt/data/web/css/editing-area.scss
+++ b/qt/aqt/data/web/css/editing-area.scss
@@ -1,0 +1,11 @@
+editing-area {
+    display: block;
+    overflow-wrap: break-word;
+    overflow: auto;
+    padding: 5px;
+
+    &:empty::after {
+        content: "\a";
+        white-space: pre;
+    }
+}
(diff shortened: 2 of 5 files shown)


dae:
It seems like a positive improvement, though it's probably best to hold off on this until the changes in 2.1.39 have settled.

Inheriting the styling from the card template is neat, but can have unintended consequences - if you give divs a blue foreground in the styling, then copy the blue text from the editor into another editing area with a different note type selected for example, the colouring gets baked into the resulting HTML.

hgiesel:
I would ask you to reconsider this for .39 for practical reasons.
There are add-ons which affect the editor which have not been broken by any of the previous changes to the editor, but also won't be broken by this one like Mini Format Pack. However there are add-ons which are broken by changes so-far, but also might be broken by this one, because it makes changes to the way the HTML is laid out, and also how the fields are initialized.

If we could get out these structural changes in one go, that would be nice.

I'm more than happy to rollback some changes, and I'll start with the insertion of the note styling. It was more a proof of concept to me, but certainly could use some more thorough testing, how it works out in practice.

dae:
I've split the pending bugfixes off into a separate release so that we can have more time to work through any issues with this and the recent sidebar changes.

A minor issue I noticed is that the first field is not focused when opening the add screen anymore - would you mind looking into it? Otherwise looks like a nice move towards more modern TS, and isolating broken HTML from the parent document is great. Thanks!

PR #958, opened 2021-01-29
Make sticky topbar have a positive z-index

hgiesel:
A typical way to style an icon would be like this:

```
i.icon {
    display: inline-block;
    width: 14px;
    height: 14px;
    margin-right: 3px;
    background-image: url(../icons/open.png);
    background-size: cover;
    background-repeat: no-repeat;
    background-origin: content-box;
}
```

However if you use an element with this styling now inside the editor fields, it will show through the topbar, because it is defined later than the topbar. Giving the topbar a higher z-index resolves this.

### DIFF
--- a/qt/aqt/data/web/css/editor.scss
+++ b/qt/aqt/data/web/css/editor.scss
@@ -53,6 +53,7 @@ body {
     position: sticky;
     top: 0;
     left: 0;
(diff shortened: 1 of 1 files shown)


dae:
Thanks!

PR #959, opened 2021-01-30
NF: add AGPL licence missing in some file

Arthur-Milchior:
I noticed it when I looked at some files now used in AnkiDroid, wanting to be sure we clearly indicate that we have
AGPLv3 code linked in the app

### DIFF
--- a/rslib/BUILD.bazel
+++ b/rslib/BUILD.bazel
@@ -1,3 +1,6 @@
+# Copyright: Ankitects Pty Ltd and contributors
+# License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+
 load("@rules_proto//proto:defs.bzl", "proto_library")
 load("@io_bazel_rules_rust//rust:rust.bzl", "rust_binary", "rust_library", "rust_test")
 load("@io_bazel_rules_rust//cargo:cargo_build_script.bzl", "cargo_build_script")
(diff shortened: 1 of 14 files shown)


rslib/Cargo.toml
```diff
@@ -1,3 +1,6 @@
+# Copyright: Ankitects Pty Ltd and contributors
+# License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
```
dae, line 2: It's mentioned in the package section already.

dae (commented):
Hi Arthur,

I personally would not have bothered with these in short files, as the headers are just an extra level of protection over the LICENSE file. The .sql files are typically single statements for example, and the header is a bit noisy.

You've also put a copyright in the LaTeX header and css header, which will be embedded in newly created decks. That needs to be removed, as we do not claim copyright over the decks people create.

Happy to merge the changes to the Rust and Bazel files if you feel like changing this.

Arthur-Milchior:
I have removed the indication from css and .tex and cargo.toml. Thank you very much for ensuring that deck's licencing is and remain clear.
You wrote you would not have bothered with some change in sql file. Do you want me to also remove them, or do I leave them here since they are here?

The licence is indicated in

dae:
Sorry for not being clear, if you could revert the .sql changes, that would be appreciated.

Arthur-Milchior:
done

dae:
Thanks Arthur!

PR #962, opened 2021-01-30
Make editor a rollup package within data/web/js

hgiesel:
This would turn editor into a rollup package, with its entry point in `data/web/js/editor/index.js`. Rollup will package it as an `iife` extending the `globalThis` object, i.e. `window`. This would allow us several things:

1. Make functions private, and thus have more control what to publish as a public API.
2. Being able to import from other directiories, means that we can modularize the code more. This could be done for all files in `data/web/js`. I already did a rough draft, splitting `editor.ts` into three files.
3. `eslint` expects functions to be either referenced somewhere in the file, or exported. Functions we use via `pycmd` are currently neither, which means, we could get around this warning/error without turning it off. `eslint` is currently turned off for files in `data/web/js`, but it would be great if we could turn it on.
4. In https://github.com/ankitects/anki/pull/914 you mentioned how it would be good to move more UI parts to typescript in the future. I think this could be a good basis for this, as the step from this kind of module, to a full-blown page, wouldn't be so hard anymore.

At the same time, we don't wave to rewrite the import path in mediasrv, which means to the python code, this change won't be noticed.


dae:
Sounds great to me. The goal of the ts/ folder is to store TS code that can be used across the different platforms, so eventually I'd like to refactor all the desktop-specific code in aqt/ into something that the mobile clients can take advantage of as well - ideally as a single component designed to work across the platforms like the graphs, but if that's not practical, at least with parts available as a shared library which each platform can use. While the actual refactoring doesn't need to happen now, might it be worth moving these files into ts/ already? The resulting bundle can be placed in the old location with copy_file(), like we're doing for the the various pages, and ts/editor is a lot easier to navigate to than qt/aqt/data/web/js :-)

hgiesel:
I moved the JS to `ts/editor` now, along with the css files. I've resorted to using `copy_page` instead of `copy_file`, because it's easier to use, and in its functionality not really exclusive to pages.
Maybe we could rename it and put it in a more neutral place like `qt/aqt/data`?

ts/rollup.aqt.config.js
```diff
@@ -0,0 +1,15 @@
...
+import process from "process";
+const production = process.env["COMPILATION_MODE"] === "opt";
+
+export default {
+    output: {
+        name: "globalThis",
+        extend: true,
+        format: "iife",
+    },
+    plugins: [resolve({ browser: true }), commonjs(), production && terser()],
```
dae, line 14: I wonder if it might be better to re-use the same rollup config, and just override the part we need, so we don't need to maintain an extra file?

```patch
diff --git a/ts/editor/BUILD.bazel b/ts/editor/BUILD.bazel
index 59658204a..56e4b1b9e 100644
--- a/ts/editor/BUILD.bazel
+++ b/ts/editor/BUILD.bazel
@@ -27,7 +27,13 @@ ts_library(

 rollup_bundle(
     name = "editor",
-    config_file = "//ts:rollup.aqt.config.js",
+    args = [
+        "--output.name",
+        "globalThis",
+        "--output.extend",
+        "true",
+    ],
+    config_file = "//ts:rollup.config.js",
     entry_point = "index.ts",
     format = "iife",
     link_workspace_root = True,
```

The protobuf references are superfluous in this case though, and it's not a big win. What do you think?

dae:
Looks good, thanks! I've made a comment above.

dae:
Re moving copy_page, sounds good. It's not page-specific, so we should probably call it something like copy_files_into_group. Feel free to put it in /ts/copy.bzl - in the future we might want a separate folder for the various .bzl files, but that will do for now.

ts/rollup.aqt.config.js
hgiesel: I have not forgotten about this, I will revisit this once I split off the card displaying functionality from the reviewer for use in the previewer, and see how that works out with different rollup setups.

PR #963, opened 2021-01-30
Coerce ease graph min to 130 if applicable

hgiesel:
Right now the ease graph often starts at 120, once again because of using `.nice()`.

This PR will coerce the graph to start at 130, if applicable, by which I mean:

* it won't change the graph, if it wants to start at 100, with ticks at every 50.
* it would change the graph, if it starts at 120, with ticks at every 10.


dae:
Sounds good

PR #964, opened 2021-01-30
Make graphs typecheck without disabling warnings

hgiesel:
I started with `added.ts`, `intervals.ts`, and `buttons.ts`.
I'd say most of these changes are rather silly (`bin.x0!` => `bin.x0 ?? 0`), or extra verbose (`.transition(trans)` => `.transition().duration(transitionTime)`), but maybe a bit more explicit? I'll let you decide if enabling the warning is worth the extra verbosity, or we'll just forget about it.

### DIFF
--- a/ts/graphs/added.ts
+++ b/ts/graphs/added.ts
@@ -1,11 +1,6 @@
 // Copyright: Ankitects Pty Ltd and contributors
 // License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
-/* eslint
-@typescript-eslint/no-non-null-assertion: "off",
-@typescript-eslint/no-explicit-any: "off",
- */
-
 import type pb from "anki/backend_proto";
 
 import {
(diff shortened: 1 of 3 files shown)


dae:
Thanks for giving it a go. I don't find the changes a significant drop in readability (apart from `<T extends BaseType>(    const updateBar = (sel: any): any => {	        sel: Selection<T, Datum, BaseType, unknown>    return sel	    )` - I don't have enough experience with TS types to be able to parse that :-). But my primary concerns with typing are a) will it help us catch mistakes and b) will it help us read the code. I know some programmers set strict lint rules and refuse to have any explicit unwraps in the codebase, but in this case I don't feel like the unwrap/any changes are really helping much with either point a or point b, so I'd be fine with leaving those lints disabled in those files and focusing on other areas, unless it's something you are particularly keen to work on.

hgiesel:
The only thing is that these general disables can mask potential problems, but I agree for now the effort is better spent somewhere else.

PR #965, opened 2021-01-31
Cleanup after #955

RumovZ:
Here are two things I found.

Also, in the light of the newfound protobuf rules (at least to me it was news), it might be worth to reconsider the enum for argumentless SearchTerms. At the moment, we call the filter with `SearchTerm(current_deck=True)`. If we put them in an enum (Other, Preset, Named, NoArgs...) again, we could use `SearchTerm(other=SearchTerm.CURRENT_DECK)` which is longer, but more intuitive than assigning an inconsequential variable.
I hate to bring it up again, but since you said it's important to get the public interface right, I wanted to have at least pointed it out.
If you think that would be an improvement, I would see to it.

### DIFF
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -29,7 +29,6 @@ from anki.models import ModelManager
 from anki.notes import Note
 from anki.rsbackend import (  # pylint: disable=unused-import
     TR,
-    BackendNoteTypeID,
     ConcatSeparator,
     DBError,
     FormatTimeSpanContext,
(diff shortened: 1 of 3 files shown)


dae:
The embedded messages do make things more ergonomic, and an enum is more intuitive than a oneof bool. I wish protobuf did not export the individual enum cases into the scope they're defined in, as it means they all need to be prefixed to avoid conflicts, but I realise that's probably a byproduct of targeting C++. At least the Rust libraries are nice enough to strip off the common prefix for us.

I'm fine with moving to an enum, but those enum names all feel a bit awkward. It doesn't look like we have any other no-arg searches at the moment.  Maybe we could rename CardState to State and pack the current_deck/collection options in there? Or just cheat and pass in "current" or "*" to the existing deck arg. ;-)

PR #966, opened 2021-01-31
Remove protobuf filters whole_col and current_deck

RumovZ:
I opted to cheat. Feels a bit like surrender. 🏳️ 
Anyway, we can use that `build_search_string` already defaults to whole collection when called with no args which should be intuitive enough.
We have to handle the unwrapping different, too, because we lack the default option now, but that's actually an improvement in my opinion.


dae:
It's ok, nothing wrong with a pragmatic approach, and the chances of a mistyped "current" in calling code are exceedingly small. Thanks for all your work on this in any case, I think it's a nice improvement over the old string building we were doing before. 🎉

PR #967, opened 2021-01-31
Dyn deckconf

RumovZ:
Some modifications to make building a filtered deck more ergonomic:

1. Enable creating and updating filtered decks from the browser search.
2. If the current deck is a filtered deck, copy its settings.
3. Make dynDeckConf a registered dialogue. (More of an internal necessity.)

Further plans:

1. [x] Move renaming feature inside dynDeckConf dialogue. At the moment, you create a filtered deck and can only afterwards rename it to something meaningul. Or you want to rename an existing deck, but have to open the deck options first to see what it does, and then have to close them and open the rename dialogue.
2. [x] Add button to view filter in the browser. So if you're creating a filtered deck, you can quickly switch to the browser and see, which cards would be included. If you adjust the search in the browser, you can then use the added browser feature to update the dyndeck dialogue with it.
3. [x] Enable filter switching between options dialogue and browser for second filter as well. Like implementation 1. and plan 2., but for the second filter in the dyndeck dialogue (requires v2 scheduler).

Please let me know if you have any reservations.


ftl/core/browsing.ftl
```diff
@@ -36,6 +36,7 @@ browsing-enter-tags-to-add = Enter tags to add:
 browsing-enter-tags-to-delete = Enter tags to delete:
 browsing-filter = Filter...
 browsing-filtered = (filtered)
+browsing-filter-to-deck = Current Filter to Filtered Deck...
```
dae, line 39: Could we just call it "Create Filtered Deck"?

qt/aqt/forms/browser.ui
```diff
@@ -583,6 +585,14 @@
     <string notr="true">Ctrl+Shift+E</string>
    </property>
   </action>
+  <action name="actionFilterToDeck">
+   <property name="text">
+    <string>BROWSING_FILTER_TO_DECK</string>
+   </property>
+   <property name="shortcut">
+    <string notr="true">Ctrl+Q</string>
```
dae, line 593: This is a special keystroke on Macs, so will need to be changed to something else.

dae (commented):
Looks good so far. I'm having trouble following the future plans though - would you mind elaborating a bit?

RumovZ:
I've updated the original post, hope that it's clearer now.
I'll rename the action and change the shortcut. Any other reserved keystrokes I should know about? I could not find a useful list for this.

qt/aqt/main.py
```diff
@@ -1055,9 +1054,9 @@ def onDeckConf(self, deck=None):
         if not deck:
             deck = self.col.decks.current()
         if deck["dyn"]:
-            import aqt.dyndeckconf
+            import aqt
```
RumovZ, line 1057: This makes the pylint check fail. But without reimporting, I get an UnboundLocalError at runtime. Do you know why that is and how to solve it?

dae: Moving the import statement in the else clause to before the if should fix it - not exactly clear!

dae:
Gotcha, thanks for elaborating. Sounds good to me. Minor issue I noticed - the search link doesn't look great in night mode. There's probably a suitable link colour in theme.py

dae:
(sorry about the conflicts)

RumovZ:
Thanks for the colour hint, I hope I managed the style the correct way.
The actual work is done, there are just a few minor things I'm not perfectly happy with, yet:

- The browser doesn't handle exceptions for passed searches, because there is no good way to communicate it to the user: We can display the error message in the browser, but not the passed search causing it because we failed to build its search string. Maybe it's enough to mention in browser's docstring that the caller must check validity? Of course, I made sure that that's already the case for existing code.
- "Create Filtered Deck (2nd Filter)..." is not really catchy...
- Would be cool to have similar stretching behaviour for all the group boxes in the dialogue.

dae:
Sorry, it looks like I did not understand what you meant about the second filter wording. Having an extra menu item for the second filter feels a bit overboard - could we just check for Alt being held down at processing time and switch to the second search in that case? Otherwise looks good thanks :-)

dae:
Sorry, re shortcuts: I'm not sure if there's a cross-platform guide unfortunately. Qt will remap ctrl to cmd on Macs, so standard cmd shortcuts will conflict, such as ones mentioned on https://www.techrepublic.com/article/20-apple-keyboard-shortcuts-business-users-need-to-know/

RumovZ:
I made do with https://en.wikipedia.org/wiki/Table_of_keyboard_shortcuts. You have to search for ctrl and cmd separately and then check for every result if it's relevant/system-wide. But should be enough for the occaional look-up.

PR #968, opened 2021-02-01
Expand sidebar match trees one level

abdnh:
See comments under https://github.com/ankitects/anki/commit/132bb5ff365ccac00b9d4ac8ba7690bddd49c9b6 for context.

This expands match trees one level while including all items under them in collapsed state:

image

Regarding apps with a similar interface, I can think of some examples:

1. There is a thing in Visual Studio called Object Browser that has a hierarchical interface.
visual-studio-object-browser-1
Searching for "Program" shows a collapsed, partially flattened tree.
visual-studio-object-browser-2

2. Another example is an app called [TreeSize](https://www.jam-software.com/treesize_free) for visualizing disk usage.
Searching for "bar" here shows the match trees collaped, with all files/folders under matches included, obviously.
treesize-1

Personally, I now think it makes more sense to include all items under matches, even if that means making the search a bit slower.
The approach taken by Visual Studio may result in a less cluttered interface but I guess it will be more complex to implement.
This PR takes a similar approach to TreeSize's, with the difference that match trees are expanded one level.

What do you think about only expanding matched section roots, while keeping other matches collapsed?

if anyone has got any suggestions, please share them here.

### DIFF
--- a/pylib/anki/tags.py
+++ b/pylib/anki/tags.py
@@ -68,9 +68,9 @@ class TagManager:
         res = self.col.db.list(query)
         return list(set(self.split(" ".join(res))))
 
-    def set_collapsed(self, tag: str, collapsed: bool) -> None:
-        "Set browser collapse state for tag, registering the tag if missing."
-        self.col._backend.set_tag_collapsed(name=tag, collapsed=collapsed)
+    def set_expanded(self, tag: str, expanded: bool) -> None:
+        "Set browser expansion state for tag, registering the tag if missing."
+        self.col._backend.set_tag_expanded(name=tag, expanded=expanded)
 
     # Bulk addition/removal from notes
     #############################################################
--- a/pylib/anki/tags.py
+++ b/pylib/anki/tags.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 
 import pprint
 import re
-from typing import Collection, List, Match, Optional, Sequence, Tuple
+from typing import Collection, List, Match, Optional, Sequence
 
 import anki  # pylint: disable=unused-import
 import anki._backend.backend_pb2 as _pb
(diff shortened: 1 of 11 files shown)


dae:
Yes, perhaps only expanding the roots one level is a good middle ground.

I like the functionality here, but the current implementation will end up doing many more string comparisons than necessary, as for each item it needs to re-check parent items as well. We would probably be better off doing our own matching in a single pass, and writing state out which the filtering/highlighting code can read. I'll try mock something up for comparison.

dae:
Ok, I've pushed a change - what do you think?

abdnh:
Looks great and simpler than I thought.

All items are now included as we got rid of QSortFilterProxyModel; I guess changing this is not worth the extra complexity.

I pushed a commit to add more section roots.

abdnh:
By the way, non-matching searches results in the root tree being shown:
image

I'll look into changing this.

dae:
Thanks for that. I thought about it while, and think collapsing rather than excluding unmatched items might be the better option. If users want to drag a deck from one location to another for example, they can search for the deck to drag, manually expand to where the target is, and then perform the operation. If items are hidden, such a workflow is not possible.

item.search() returns a bool indicating whether anything was matched on that node or children, so the return value of the root node should make it easy to special-case the 'nothing matched' case. What do you think we should do in that case?

abdnh:
I pushed a change to hide the tree when there are no matches.

dae:
Looks good! Was there anything else you wanted to change on this before merging it in?

dae:
On a somewhat-related note, I've noticed that displaying the full tags tree on large collections slows down browser opening and refreshing a perceptible amount. With our fancy new search feature and the ability to expand and collapse all immediate children of a given node, maybe it's worth thinking again about whether expanding newly added tags by default is the best option. Maybe tags could start off collapsed, and the user could expand the ones they wanted instead?

@ankingmed want to weigh in here?

AnKingMed:
100% would recommend having tags closed by default. I'd say the only ones that should be open when opening the browser are the ones the user had open the last session, but no need to have them all open. I think I have around 100,000 tags at the moment in my collection

dae:
Thanks @AnKingMed, we'll give that a go

@abdnh might merge this is now to minimize the chances of conflicts with related changes, but please feel free to continue the discussion/make changes if you think something's not right at the moment.

PR #969, opened 2021-02-01
Svelte Previewer for Clayout, Browser, Editor

hgiesel:
This idea started in #914.

I've implemented the visuals of a new Preview window as a TS/Svelte/Bootstrap page.
Like I've described in #914, it has three versions, depending on where it would be opened:
1. If opened in Browser, it would use the `BrowserSelect` component as `extraControls`
2. If used in the AddCards editor, or Clayout, it would use either `ClozeSelect`, or `CardTypeSelect` with an "ephemeral card".

This is e.g. the `CardTypeSelect` version:

image

I've created a `qt/aqt/previewer2.py`, which could be renamed `previewer.py` once the Previewer is fully implemented.

However now when it comes to connecting the page to Anki, I've got some implementation questions. `Congrats` and `Graphs`, only interact with Python via HTTP requests, and otherwise work mostly autonomous. Because of the nature of the previewer, I don't think that's possible?

My approach now would be to wrap it in a QDialog, which then interacts with it via `eval`, like other Anki components. Svelte has an [client-side API](https://svelte.dev/docs#Client-side_component_API) which would allow for this.

What do you think about this?

Some other comments:
1. At the first stage, it might not be worth to touch the previewer in Clayout, as it might involve changing some other parts as well, whereas the Editor, and Browser is more straightforward to implement.
2. While developing this, I tried to install `svelte_preprocess` and `sass` for `<script lang="sass">` support, but failed fiddling with the svelte rules. At the same I remembered faintly texting with you about this. I guess it's not possible with the current Svelte support for Bazel yet?


dae:
We're currently used a [hacked version](https://github.com/ankitects/rules_svelte/tree/ts-support) of rules_svelte to get TypeScript support. It should be possible to patch it to add sass support as well - TS is implemented via svelte-preprocess as well, so it's hopefully not too complicated. You can test with a local repo by uncommenting the following in repos.bzl:

```
    # native.local_repository(
    #     name = "build_bazel_rules_svelte",
    #     path = "../rules_svelte",
    # )
```

Skipping clayout for now sounds good too.

How to pass data around is a good question. The add screen holds the current note content in memory, and generates the front and back without ever committing the note or card to the DB, and the card layout screen uses in-memory copies of the templates and css, so we can't just pull data from the DB.

There are two options I can think of:

1) Have the svelte code call a HTTP endpoint with arguments (optional note_id, card_idx, optional cloze_number). The Python side would need to receive the request, and figure out the correct question/answer/css to return - whether that's fetching from the DB, extracting the ephemeral note from the add screen, etc. With this approach we could limit .eval() to a simple prop update/function call that tells the JS side to refresh.
2) Pass the question/answer/css in via a JSON message with .eval(), like the current preview code does. If something like the cloze number is changed on the JS end, it would need to send a signal to the Python side requesting new data be pushed.

While packing JSON into a function call with string manipulation feels dirty, I'm slightly leaning towards it, as it seems to more closely map to how single page might handle it, with the parent component passing the necessary data into the child. But I can see pros and cons of either approach - what are your thoughts?

If we did go the pack-into-function-call route, maybe one slightly less nasty way would be to pass a single JSON dict in as an argument instead of unrolling each argument separately.

dae:

PR #970, opened 2021-02-02
Cleanup 967

RumovZ:


RumovZ:
Regarding the last commit, I realised that those translations should actually not be necessary. However, on my machine, Qt translations don't seem to work because all standard buttons have English labels. Is there something wrong in aqt/__init__.py?

dae:
Please try with latest main

RumovZ:
Thanks, it works now. 👍

dae:
Thanks Rumo, looks good, and appreciate the typecheck change.

PR #971, opened 2021-02-02
Add checkpoints for sidebar drag & drop routines

abdnh:
Also save note and reset browser model.

I noticed that editor.saveNow is not called for sidebar.rename_deck and the browser model is not reset too. Is this a mistake or maybe this not necessary in this case?

### DIFF
--- a/qt/aqt/sidebar.py
+++ b/qt/aqt/sidebar.py
@@ -399,12 +399,19 @@ class SidebarTreeView(QTreeView):
             return False
 
         def on_done(fut: Future) -> None:
+            self.browser.model.endReset()
             fut.result()
             self.refresh()
+            self.mw.deckBrowser.refresh()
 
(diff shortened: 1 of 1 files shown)


dae:
Thanks Abdo!

IIRC, saveNow is required any time the user may have a field focused - it saves the current text and removes focus, so that the blur handling in onBridgeCmd() in editor.py doesn't end up firing after the current note has changed or been invalidated. The rename action requires clicking elsewhere, so I think we'd probably be safe there as the blur should already have fired?

I think we should probably be resetting after any operation that may change the contents of the card list, though interestingly it does appear to get updated when renaming currently, at least when I tested it.

abdnh:
> though interestingly it does appear to get updated when renaming currently, at least when I tested it.

Yes, I also didn't find a problem here. But the reset was required for _handle_drag_drop_tags when the editor is opened at least.

PR #973, opened 2021-02-02
Render error if MathJax raises error

hgiesel:
This addresses what was observed in [this thread](https://forums.ankiweb.net/t/amscd-rendering-in-mathjax/7035/4).


dae:
Thanks for the quick fix, Henrik.

PR #974, opened 2021-02-02
Move copy_page to ts/copy.bzl and rename to copy_files_into_group

hgiesel:
Like talked about in https://github.com/ankitects/anki/pull/962.


dae:
Thanks!

PR #1006, opened 2021-02-05
Remove some dead editor code

hgiesel:
This code is currently not used anywhere.
The previous editor changes already probably broke almost all add-ons out there, so it might be worth to do some cleanup.

The functionality of `String.prototype.format` is nowadays better achieved using template strings.

`insertNewline` is currently not used anywhere, I'm also not sure what it used to achieve. Using `pre` tags in the editor fields seem to work mostly fine right now.

The occurence of `window.event` is something I overlooked last time.

### DIFF
--- a/ts/editor/index.ts
+++ b/ts/editor/index.ts
@@ -10,10 +10,6 @@ let changeTimer: number | null = null;
 let currentNoteId: number | null = null;
 
 declare global {
-    interface String {
-        format(...args: string[]): string;
-    }
-
     interface Selection {
         modify(s: string, t: string, u: string): void;
         addRange(r: Range): void;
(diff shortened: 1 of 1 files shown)


dae:
These changes don't look like they'll cause too much impact. insertNewline is a relic from one of my attempts to avoid divs in the past, and it looks like it was missed when rolling it back

17bb179d0685a6146bdddfa9c2a176c9e91f9be7
554ff3d8d2ddd8f3e3f84b63b342cfac731712e5

"broke almost all add-ons out there" sounds a bit alarming, and I'm hoping it's slight hyperbole? :-) A brief test seems to indicate things like 'mini format', 'add table' and 'image occlusion' are still working, but 'frozen fields' and 'searching/pdf reading' are not.

hgiesel:
I guess what I should have said was "broke almost _all of my editor_ add-ons" ^^;
Add-ons that were depending on the structure of the editor HTML, and the way that fields were updated.

dae:
Ah :-)

PR #1008, opened 2021-02-05
Move reviewer to ts/reviewer

hgiesel:
This will move `data/web/js/reviewer.ts` and `data/web/css/reviewer.scss` to `/ts/reviewer`.

The motivation here is the following:
Previewer will also display the card to the user, so it would be nice, if it could reuse the code that Reviewer uses. So the first step is moving reviewer to `ts`. In the Previewer PR, I will split off the "displaying" functionality of Reviewer into another directory, probably `ts/cardDisplayer`, which can then be imported by both Reviewer and Previewer.

### DIFF
--- a/qt/aqt/data/web/css/BUILD.bazel
+++ b/qt/aqt/data/web/css/BUILD.bazel
@@ -26,12 +26,21 @@ copy_files_into_group(
     package = "//ts/editor",
 )
 
+copy_files_into_group(
+    name = "reviewer",
+    srcs = [
+        "reviewer.css",
+    ],
+    package = "//ts/reviewer",
+)
+
 filegroup(
     name = "css",
     srcs = [
         "core.css",
         "css_local",
         "editor",
+        "reviewer",
     ],
     visibility = ["//qt:__subpackages__"],
 )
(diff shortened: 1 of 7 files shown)


dae:
This is showing a JS error when reviewing, as _getTypedAnswer() expects typeans to be in the global scope. A brief grep through add-ons seems to indicate it is not being accessed directly, so we should be able to provide an accessor func instead of having to export it directly. onUpdateHook/onShownHook and ankiPlatform were also public API, so we'll break things if we don't export them.

In terms of release planning, I'd like to start holding off on further breaking changes until the next stable release, as we've accumulated quite a few changes at this point, and the editor and sidebar changes will already cause disruptions.
While we could potentially squeeze this file rename in before the first beta if the missing globals were added, if you're also planning to split the code into separate files, perhaps it's worth either holding off on merging this, or copying rather than moving the existing reviewer.ts - that way you'd be free to make changes as the current code could be kept pointed at the old file until we're ready to switch over.

dae:
(re ts/cardDisplayer: very minor, and perhaps irrational, but I'd lean towards a lowercase name)

hgiesel:
Copying sounds good to me. I can't really estimate how well my plans will play out at his point, because there are still a few unknowns, and I'm also wary about the amount of changes for the next version.

PR #1014, opened 2021-02-09
Refactor "currentField" concept for editor, export for add-on developers

hgiesel:
I started porting my editor add-ons for Anki 2.1.41 beta, and I realized, that there was one big omission in the new visible editor API, which is an easy way to get the current field.

However I didn't want to simple export a way to access `currentField`, because I always thought it's kind of an anti-pattern to have this kind of global variable available, that all functions access. The event handlers wouldn't even need it, as they have access to `event.currentTarget`, which is guaranteed to be of type `EditingArea`, as well.

So this turned out to be a bigger refactoring PR, with the main goal, to remove the global `currentField`.

Other changes include:
* For cases, where there is no easily accessible "currentField" via an `Event` interface e.g. `wrap`, or `pasteHTML`, there is a new exported function `getCurrentField`, which utilizes `document.activeElement`, which will also return `EditingArea`, if the focus is inside the contenteditable, as anything else would be hidden within the shadow root.
* I tried to put global variables into their own files, so that only functions that absolute require them have them in their scope, this includes:
  1. The old `currentField`, is now `previousActiveElement`, and is in `focusHandlers.ts`
  2. `currentNoteId` is accessible for add-ons via `getNoteId`, and is in file `noteId.ts`
  3. `changeTimer` is in the file `changeTimer.ts`
* I rewrote the logic for scrolling the currentField into view, utilizing `scrollIntoView`, and `scrollBy`
* I put a 2px margin between the the button rows, which is noticeable when they wrap. Before this, they would touch each other, now there's a small gap.
* When clicking on text, which has bold, italic, or underline applied to them, it would not highlight the button. Now it does, by calling `updateButtonState` upon `mouseup`.

### DIFF
--- a/ts/editor/changeTimer.ts
+++ b/ts/editor/changeTimer.ts
@@ -1,0 +1,47 @@
+import type { EditingArea } from ".";
+
+import { getCurrentField } from ".";
+import { bridgeCommand } from "./lib";
+import { getNoteId } from "./noteId";
+import { updateButtonState } from "./toolbar";
+
+let changeTimer: number | null = null;
+
+export function triggerChangeTimer(currentField: EditingArea): void {
+    clearChangeTimer();
+    changeTimer = setTimeout(function () {
+        updateButtonState();
+        saveField(currentField, "key");
+    }, 600);
+}
+
+function clearChangeTimer(): void {
+    if (changeTimer) {
+        clearTimeout(changeTimer);
+        changeTimer = null;
+    }
+}
+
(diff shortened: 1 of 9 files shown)


dae:
Thanks Henrik, this code is looking a lot nicer than it was a few weeks ago :-). I might create a forum wiki post later with some hints for add-on authors on required changes - perhaps you could add anything you think worth mentioning about the editor related changes there?

PR #1015, opened 2021-02-11
Search API bikeshedding

dae:
@RumovZ This started out as just making some minor tweaks, but it turned into something bigger than I was originally intending. I think it makes the API a bit nicer, but thought you might want to look over it before it gets merged in - what do you think? I've left some comments in the individual commits.

### DIFF
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -11,7 +11,7 @@ import sys
 import time
 import traceback
 import weakref
-from typing import Any, List, Optional, Sequence, Tuple, Union
+from typing import Any, List, Literal, Optional, Sequence, Tuple, Union
 
 import anki._backend.backend_pb2 as _pb
 import anki.find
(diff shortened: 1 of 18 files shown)


dae:
Oh, and very minor thing I noticed while working on this - I think Rust convention is assert_eq!(actual, expected) instead of vice versa. I also gather it's different from the convention in some other languages.

RumovZ:
Looks good to me! I appreciate the heads-up.
I think the sidebar context menu could make use of this to help with the construction of more advanced searches.
When I started looking into that part of the code, I was mainly thinking about building the preset filters. It's much more powerful than that by now. Is the goal to provide the full functionality available by literal search input? In that case, there would still be some keyword searches missing. And I imagine accepting wildcard searches would complicate things as we would need some sort of escaping.

Writing JSON to a string sounds very feasible. I'm not sure what's the demand for that is though as working with the normalised strings should already be pretty straightforward.

dae:
Thanks for looking over the changes!

We've covered the common cases where automatic escaping is useful, so I'd be inclined to leave it as it is for now, as we can always add more later if the need arises. Same with JSON - let's wait for a strong demand to present itself first.

PR #1019, opened 2021-02-12
Note.ephemeral_card() fix

hgiesel:
This is a fix for what you mentioned [here](https://github.com/ankitects/anki/commit/6e9dfefb24e42d8f4367e0a7fb98e6af0030de64#commitcomment-47036672).

Now you can pass in both a custom note type, and a custom template for rendering. Really should have noticed it the first time around.


dae:
Thanks for the quick fix :-)

PR #1020, opened 2021-02-14
Add missing preconditions for browser search in filtered deck options

abdnh:
I believe browser search should match the parameters used to create the filtered deck, as in [move_cards_matching_term](https://github.com/ankitects/anki/blob/d2c580033b9b59c3a8f5e0e25649ba90fc0b5784/rslib/src/filtered.rs#L226).

### DIFF
--- a/qt/aqt/dyndeckconf.py
+++ b/qt/aqt/dyndeckconf.py
@@ -152,7 +152,26 @@ class DeckConf(QDialog):
 
     def _on_search_button(self, line: QLineEdit) -> None:
         try:
-            search = self.mw.col.build_search_string(line.text())
+            node = self.mw.col.group_searches(
+                line.text(),
+                SearchNode(
+                    negated=SearchNode(card_state=SearchNode.CARD_STATE_SUSPENDED)
+                ),
+                SearchNode(negated=SearchNode(card_state=SearchNode.CARD_STATE_BURIED)),
+                SearchNode(negated=SearchNode(deck="filtered")),
+            )
+
+            if self.mw.col.schedVer() == 1:
+                search = self.mw.col.join_searches(
+                    node,
+                    SearchNode(
+                        negated=SearchNode(card_state=SearchNode.CARD_STATE_LEARN)
+                    ),
+                    "AND",
+                )
+            else:
+                search = self.mw.col.build_search_string(node)
+
         except InvalidInput as err:
             line.setFocus()
             line.selectAll()


qt/aqt/dyndeckconf.py
```diff
@@ -152,7 +152,26 @@ def on_search_button_2(self) -> None:
...
+                ),
+                SearchNode(negated=SearchNode(card_state=SearchNode.CARD_STATE_BURIED)),
+                SearchNode(negated=SearchNode(deck="filtered")),
+            )
+
+            if self.mw.col.schedVer() == 1:
+                search = self.mw.col.join_searches(
+                    node,
+                    SearchNode(
+                        negated=SearchNode(card_state=SearchNode.CARD_STATE_LEARN)
```
dae, line 168: This could be slightly simplified by listing the search nodes above in a list called `extra_filters`, optionally adding this one in v1, and then just calling .group_searches(line.text(), *extra_filters) once

dae (commented):
Thanks Abdo, LGTM - any objections @RumovZ?

qt/aqt/dyndeckconf.py
dae: Actually this might need a bit more thought - skipping join_searches will wrap multiple terms in (), and if you move back and forth from the browse screen the extra filters and brackets may end up being added multiple times

abdnh: > if you move back and forth from the browse screen the extra filters and brackets may end up being added multiple times

I thought of this as I read the discussion in Rumo's PR, but guessed moving back and forth multiple times will be a rare event anyway.

RumovZ:
On the one hand, there are often questions about cards missing in filtered decks and this would be great way to let the user discover the restrictions. On the other hand, I see a few problems:

- Even if you only switch once it will significantly complicate the deck filter making later adjustments harder (or you will have to manually delete the added filters before switching back).
- For longer existing filtered decks, cards that can't be moved to a filtered deck currently may still be of interest to the user as they may be included in a future rebuild.
- I don't find it very intuitive, because you click on "Search" and then expect that search to open. If it was about previewing the deck, you would also expect the second filter and other options to be applied.

So what about adding an extra button for previewing the deck? The preview wouldn't be able to reflect the card limits and order, but it might still be useful to see the pool of eligible cards.

dae:
Throwing out some ideas is order of complexity:

One:

- avoid rewriting the search, and add a label somewhere in the filtered search dialog: "Searches implicitly have '-is:learn -is:suspended -is:buried -is:filtered' added to them." The exact queries would need to be passed in based on the scheduler version.

Two:

- Under the "search" row, have "Implicit", with the extra -is:learn etc shown on the right. Implicit or the terms could be clickable, and could act like this PR, adding the extra terms to the query before searching.

Three:

- we add another backend method add_filtered_deck_limits(node: SearchNode)->str
- If the top level node is an AND group, it looks at the top-level search terms, and adds the learning/filtered/etc restrictions if they are missing
- if it's a single node or an OR group, the existing search is wrapped in an AND group first
- if the existing learning/filtered/etc restrictions are already there, no changes are made
- the rendered search text is returned
- it will need to know the scheduler version - the backend method can look that up and pass it into the function in writer.rs

We could then use that instead of build_search_string(), and use it not only when sending text to the browse screen, but when receiving text from the browse screen as well. There would be no change when moving between the browse screen and filtered deck screen if those limits were already listed, but if users wanted to see what was available that's currently suspended/filtered, they'd need to manually edit the search.

RumovZ:
I like option two best. It makes the dialogue longer, but crystal clear which cards can be added and gives the user full control of what to show in the browser. I'd use an uneditable LineEdit and possibly indent the row.

grafik

dae:
Any reason for the LineEdit instead of a Label?

I wonder whether if it would work if we got rid of "implicit" and just showed a label "AND -is:learn AND -is:suspended"... underneath in a label?  Not strongly pushing for that, just figured it might be worth trying - it might be a little more subtle/less prominent.

RumovZ:
I thought text and search text in one label might look odd, so a label with only search text is fine by me. We could hardcode `... AND ` and then add a properly built search string.

abdnh:
Does this look fine?
image

abdnh:
I just realized that adding the extra filters unconditionally will also make the search buttons less useful when rebuilding an existing filtered deck because not all cards will be included (due to `-deck:filtered`)

dae:
Looks pretty good. The only slight negative is that the blue text draws the attention of the eye away from the actual search box a bit - which is good if users are not aware of the extra filters, but might be a bit annoying for regular users? I wonder how it would look if that text was black or grey, and there were a clickable 🔍 at the end?

But you're right that -deck:filtered may introduce more confusion for users who are wondering why cards that are already in the filtered deck are not being shown in the browser.

Maybe we shouldn't try to automatically link to the full query, and just leave the text selectable for users who want to copy and paste it?

Not sure what the best path forward is :-)

RumovZ:
Right-aligning could help to reduce the prevalence. (In any case, the buttons should span the three rightmost columns to avoid the stretching.)
I like the idea with the magnifying glass. If used consistently through Anki, it could be a recognisable indicator to the user that something is searchable in the browser and we already discussed using it in connection to the sidebar.
Nevertheless, two other possibilities: making the clickable pseudo-label gray but keeping the hover effect, or only making the first `AND` a button and the rest a plain label.

Good point about rebuilding. I mean, the search could be easily corrected by replacing `-deck:filtered` with `-deck:filtered OR "deck:Filtered Deck 1"` where the last part is generated dynamically if necessary. But the string is starting to get quite long and using a deck name introduces uncertainty about the actual length.
So maybe it would be better to show a message (clickable/with clickable icon) mentioning the limitations. A click would reveal the exact preconditions in the browser anyway.

dae:
Looks like we can change the existing setStyleSheet call to use colors.DISABLED instead of colors.LINK to get grey text that will underline when hovered, and that's probably the easiest option.

We could possibly sidestep the 'missing items in this deck' issue by showing a popup "Note, the browser search will be missing cards that are already in this filtered deck" if the filtered deck is non-empty, or by asking the user something like "This will empty the filtered deck then search in the browse screen. Proceed?"

dae:
Any objections to closing this one in favour of #1039 abdo?

dae:
Just noticed your comment in the other thread, so I assume that's a yes. :-)

PR #1028, opened 2021-02-16
Add max-width: 100% to editable.scss

hgiesel:
This was mentioned [here](https://forums.ankiweb.net/t/anki-2-1-41-beta/7305/46?u=hengiesel).
I removed it, because I didn't know of its effect in the editor fields.

Instead of `max-width: 90%`, I set it to `max-width: 100%`. For a sufficiently vertical pictures, both 90% and 100% will make the picture occupy too much space.

It would be nice to add a dynamic `max-width` [range bar](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) to Anki for users to adjust the image width, however that would most likely mean, that we would have to overwrite the context menu.

### DIFF
--- a/ts/editor/editable.scss
+++ b/ts/editor/editable.scss
@@ -10,10 +10,14 @@ anki-editable {
     }
 }
 
-img.drawing {
-    zoom: 50%;
+img {
+    max-width: 100%;
 
-    .nightMode & {
-        filter: unquote("invert() hue-rotate(180deg)");
+    &.drawing {
+        zoom: 50%;
+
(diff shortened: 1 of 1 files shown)


dae:
Thanks Henrik!

Could you elaborate on your last point? Are you suggesting the range appear on a right click? Perhaps we could avoid altering that behaviour by having an options button drop-down/popover like in the preview screen?  But if the main goal would be to shrink down vertical images, maybe a separate non-configurable max-height with something like 90vh might make sense?

Issue #1027, opened 2021-02-16
Deleting tags also removes prefixes


RumovZ:
If a note has a tag `foobar`, using the tool for deleting tags with `foo` will change the tag to `bar`.
This is reproducable on the main branch and the current beta.

abdnh:
I think I'm to blame for this 😅

Looks like the fix is as simple as:

```diff
diff --git a/rslib/src/tags.rs b/rslib/src/tags.rs
index 2a91013b9..02f337fa5 100644
--- a/rslib/src/tags.rs
+++ b/rslib/src/tags.rs
@@ -334,7 +334,7 @@ impl Collection {
         let tags = split_tags(tags)
             .map(|tag| {
                 let tag = if regex { tag.into() } else { to_re(tag) };
-                Regex::new(&format!("(?i)^{}(::.*)?", tag))
+                Regex::new(&format!("(?i)^{}(::.*)?$", tag))
                     .map_err(|_| AnkiError::invalid_input("invalid regex"))
             })
             .collect::<Result<Vec<Regex>>>()?;
```

Maybe it's worth adding a test too.

PR #1029, opened 2021-02-17
Fix tag replacement matching substrings

abdnh:
Reported by @RumovZ in https://github.com/ankitects/anki/issues/1027

A similar issue was fixed in https://github.com/ankitects/anki/commit/8a9a9591cb9cfeb9eeaef14d08ea70fc15dff3c5 but I reintroduced it in my tag tree PR.


dae:
Thanks for the quick fix @abdnh, and thanks for the report @RumovZ!

Closes #1027

PR #1033, opened 2021-02-18
Note view: Display notes in the browser table

RumovZ:
mIi8e038kO

Preface
---
This PR introduces a new view to the browser where the table rows correspond to notes instead of cards. To achieve a clean and understandable implementation, the browser class had to be refactored largely, splitting all of the table management code into a new class.
I'm aware that other things are currently going on, but I thought it was time to show the work that has already been done. It will take some more time to finish it, anyway.
I appreciate anyone taking the time to look into it and give feedback.

Merge prospect
---
Naturally, I hope that this feature will be merged into the main branch at some point. It has been requested before and its usefulness is hopefully self-evident.
Whatever the future plans are for the browser, I'm confident that the new abstractions and modularisations here will make them easier. Also, while this PR would definitely break some add-ons working with the browser (see below), it should reduce future breakages overall and make it so much easier to work with the browser for both core and add-on developers.

Add-on Compatability
---
Since a lot of funcionality has been moved to a new class, many method calls by add-ons will no longer work. A good portion can probably be reinstalled by aliasing and rolling back some minor refactorings. For code working with the table itself (e.g. code hooking into the search process), it might be possible to support it as long as the browser is in Cards View. But I'm sceptical whether it's worth it and the right thing to do.

Architecture
---
* `Table` is a class for handling table data, a table view, the table state (old card view or new note view) and selections. It exposes methods for every need of the Browser which holds an instance of it.
* `DataModel` is a subclass of `QAbstractItemModel` managing the actual table items as before, but more strictly confined and only accessed through table. The `items` are card or note ids, depending on the current state.
* `ItemState` is an abstract base class. It's instantiated as `NoteState` or `CardState` where all the information about the current state is stored. The state instance takes items from the model and returns `Card` and `Note` objects of which it also maintains a cache.

Other Changes
---
* In the context menu, actions not related to the current state have been moved to a submenu (see the demo) making it simpler and communicating the target of the action more clearly to the user. This is not necessarily linked to this PR, but makes more sense in connection with it.
* (ongoing) `Browser` exposes more public methods instead of relying on other modules to access its fields directly.
* (ongoing) The table reset api is more consistent, differentiating better between changes to the collection, the table items or the cache.
* (ongoing) More documentation.
* ...

To-Dos
---
* [x] Implement note sorting on backend.
* [ ] (See below.)

Discussion
---
Following are some points I'm not sure about:

* Add new columns that make sense in note state, like number of cards. Other metadata about the note's cards could be also interesting (total lapses, cards per queue...) but potentially more difficult to implement and less performant (i.e., slower per row compared to the card state, not slower in total).
* Enable previewing other cards than the first of the note in note state. Seemed as a good idea at first, but maybe only showing one preview per note is more useful, especially as it wouldn't be obvious to the user which card is presented in the previwer (and also in the card info box and flag menu).
* Enable card columns in note state. The backend would need to know which template to use for ordering and adjust the SQL accordingly.
* Related to the previous two points: Add a way to select the template (per notetype or note?) that is used in note state for card-level data.
* Behaviour after state toggle: Either reset completely, ensuring order, or calculate the new items from the old ones, somewhat keeping the order from the previous state (also faster). More generally, that also applies to actions in the browser modifying the collection where the current implementation has no clear preference.


dae:
I'm going to be tied up for a while as I work on updates to the scheduler, but will try return to this within a week or so.

qt/aqt/browser.py
```diff
@@ -786,155 +381,70 @@ def add_preview_button(leftbuttons: List[str], editor: Editor) -> None:
...
+        self.editor.saveNow(callback)
+
+    def begin_reset(self) -> None:
+        """Must be called before modifying the current note from outside the editor to prevent
+        overwriting the changes with the cache. Unsaved changes in the editor will be lost.
+        After the modification, _onRowChanged() must be triggered, e.g. by end_reset() or search(),
+        to move the editor to a consistent state.
+        """
+        self.editor.setNote(None, hide=False)
+
```
dae, line 394: This is different to the old beginReset() and endReset() behaviour, and I'm not 100% sure it's safe at the moment. If changes are made such as deleting a note without beginReset() being called on the model, Qt may queue up a request that tries to redraw a row that references that note, and if that request is processed prior to the model being reset, we could end up getting an error. This was a big issue in the past when the DB layer would periodically call back into Python and drain the UI event queue, but I'm not sure how much of an issue it is now.

dae (commented):
Still busy with scheduling work, so just some brief comments for now.

This is a change that I haven't been particularly enthusiastic about in the past. One of the reasons was that I was worried it would make the code even harder to maintain. By refactoring the code at the same time and cleaning up the API, you've reduced my concerns about that a fair bit. I still think it's a big project - even with the core functionality implemented, there's a lot of peripheral things that need thinking about, such as some of the things you listed in the Discussions section.

I had also thought that a note mode would not make much sense for card-based operations. But I assume the approach you've taken here is to apply them to all cards of the selected note? That doesn't seem like a bad compromise.

One thing that might be nice is to explore moving columnData() and related functionality into the backend. If we teach the backend how to derive all the string values for the configured columns, we could have an API like:

def browser_rows_for_cards(self, card_ids: List[int]) → Sequence[Sequence[str]]

Then data() could just ensure the given row data was available, and index into it. We could keep things simple by only grabbing a row at a time, but this would allow for fetching ahead if we wanted to do that later. Any future Javascript-based implementation would also be able to take advantage of the code, as could the two mobile clients. It would just need to return an empty string for any unrecognized columns, that are added by add-ons. And a similar browser_rows_for_notes() could be added too.

This might be particularly useful for columns where we call stripHTML() - that can be quite slow on columns with larger amounts of text content currently.

pylib/anki/notes.py
```diff
@@ -75,6 +76,14 @@ def __repr__(self) -> str:
...
         return joinFields(self.fields)
 
+    def card(self, card: int, reload: bool = False) -> Optional[anki.cards.Card]:
+        if reload or card not in self._cards:
+            try:
+                self._cards[card] = self.col.getCard(self.card_ids()[card])
+            except IndexError:
+                return None
+        return self._cards[card]
+
```
dae, line 86: I fear this may lead to accidental cache coherency issues down the road. If the browser needs to cache the lookup, it might be better to explicitly do it at the call site, instead of implicitly in a method here. But perhaps we would not even need this if the .data() was calculated by the backend?

pylib/anki/collection.py
```diff
@@ -465,27 +465,38 @@ def genCards(self, nids: List[int]) -> List[int]:
     # col.find_cards("", order=BuiltinSortKind.CARD_DUE)
     # the reverse argument only applies when a BuiltinSortKind is provided;
     # otherwise the collection config defines whether reverse is set or not
+    @staticmethod
+    def build_sort_mode(
+        order: Union[bool, str, BuiltinSortKindValue],
+        reverse: bool,
+    ) -> _pb.SortOrder:
+        if isinstance(order, str):
```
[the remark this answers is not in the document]
RumovZ: It's probably best to make it private. There's no point in calling it outside `find_cards` and `find_notes`.

qt/aqt/browser.py
RumovZ: You are right, I realised that `delete_notes` could be a potential problem after changing the resetting process. Maybe that method could be handled separately as there isn't much else that could go wrong.
I'm not sure if it's possible to trigger that error at all, though, as the notes should already be loaded into the cache.

RumovZ:
> But I assume the approach you've taken here is to apply them to all cards of the selected note?

Right, that's actually what's already being done for note actions in the current browser. I tried to make that more obvious to the user by moving not state-related actions to a submenu in the context menu. I always thought it was confusing that you'd rightclick a card, select `Delete` and thereby actually delete the note.

> One thing that might be nice is to explore moving columnData() and related functionality into the backend.

That sounds promising! I'll look into it.

dae:
@hssm with your work on the advanced browser add-on, thought you might want to be involved in this discussion

RumovZ:
Closed since it will be easier to make a new PR by now.
<hr>

> One thing that might be nice is to explore moving columnData() and related functionality into the backend. If we teach the backend how to derive all the string values for the configured columns, we could have an API like:
def browser_rows_for_cards(self, card_ids: List[int]) → Sequence[Sequence[str]]

I have thought about that and a few question have arisen. Before I commit to a time-consuming and futile approach, I'd like to ask for your opinion.
1. `question` and `answer` can't be rendered on the backend. Do we just leave those cells empty and fill the gaps in Python or should we try to return the necessary data so the frontend at least doesn't have to query the exact same data again in a different form? Or may it even be possible to do more of the template rending on the backend? I don't see how that could work with custom filters, though.
2. Instead of fetching all the cards and notes from the backend we could build a pinpoint SQL query and mapper for generating the strings. Those would only have to be rebuilt when the active columns change. Would that be efficient or unnecessarily complex? 🤔

dae:
I'd be inclined to fully render the templates into a string, ignoring any unrecognized filters. Performance is important for the card list, and I think users will be able to live without custom filters there.

For card queries, I think it would be easiest just to fetch the card for each row. We're not rendering the entire search query; we only need to render the lines displayed on screen, so we're only talking about 20-100 at max. While each extra field we fetch from the db does add some delay, I think for that number of cards, it's going to be fairly negligible, and it will simplify the code. The logic for displaying the correct due dates, interval and so on depends on the card type and queue, and that will be easier to implement in Rust than trying to implement it all in SQL.

For notes, skipping the fields would be cheaper if we only need to show the sort field. We potentially need to show other data like tags, mtime and so on though, and it would be easier to work with in Rust. What about adding a new routine to the storage layer get_note_without_fields(), which leaves fields as an empty vec? We could then decide on the function to use based on the presence of the question or answer columns, and it would keep the code simple. I'm not too worried about other unused fields in notes, it's really just the fields we need to worry about - some users have fields with hundreds or thousands of kilobytes of HTML in them. The sort field will have the HTML stripped, so it tends to be more manageable.

Re this PR, the undo and resetting work looks like it's going to take a while, and need an extending testing period afterwards, so it might be a good time for this PR as well. columnData() in Rust won't conflict with any of the changes I'm making at the moment, so might be a good place to start, and once the reset/undo stuff has settled down a bit, the Python end of things should be safer to work on (and would give me time to return to the scheduler changes)

RumovZ:
> What about adding a new routine to the storage layer get_note_without_fields(), which leaves fields as an empty vec?

Don't we need all the fields to render question and answer?
And maybe instead of adding a routine for loading a partial note, we could add a query that fetches all (potentially) needed fields from the Cards and Notes tables at once and stores them in a `RowData` struct.

dae:
If we need question or answer, we'll need fields. I was suggesting we could skip them if neither question nor answer is needed as an optimization.

Apart from skipping fields, my gut feeling is that the performance gains we might get from trying to only fetch the required note/card columns are not worth the extra code complexity/extra maintenance burden.

RumovZ:
All right then. I guess, optimisations could always be made afterwards.

One thing I don't quite understand is how we decide which rows to fetch. The item view asks for data cell-wise, so we can of course get the cell index row from the backend. But you were saying 20 to 100, so how do we know which other 19 to 99 rows should be fetched?

As a sidenote, we don't only have to provide the table cell strings, but also row colours and rtl info. Just to keep in mind.

dae:
Yep, you're right about needing colours/rtl - we could just pack it into a struct or protobuf message and return one for each column of the row.

Currently we use .cardObjs to cache a card object for each row we process. My thinking was that we could do something similar, but cache the row object instead.

Qt calls data() for each displayed cell. eg qt calls data() for row 0, column 0
We check the row cache, and as we don't have an entry for row 0, call a backend method to fetch it.
With the returned value, we extract the requested column data and return it

then qt calls data() for row 0, column 1
we already have the row cached, so just return the desired column text/colour/etc

same for column 2, 3, etc.

then qt calls data() for row 1, column 0, and that results in a new backend call

We could potentially return more than one row at once. Eg when Qt requests row 0, we fetch self.cards[0..20], and persist the other rows in the cache. It would probably buy us a bit of performance in the initial display + scrolling down case, but it would cost us performance when scrolling up, unless we were more intelligent about the range we were requesting. So I'd be inclined to ignore that for now, and just focus on fetching a single row at once.

RumovZ:
Thank you for the detailed explanation, it was only the last paragraph I was unclear on, though. I thought you might have an idea of how we can kind of predict which rows will be needed next, apart from manually checking which rows are in the vicinity and haven't been fetched, yet.
I'll put it on the list of possible future optimisations and focus on the single-row routine, then.

dae:
Sorry, I didn't mean to imply you didn't know the first part, just wanted to be clear :-)

RumovZ:
No worries, I appreciate your help! Only my ego made me clarify that I got that part. ;)

PR #1034, opened 2021-02-19
Shortcuts to preview clozes in card layout screen

abdnh:
This assigns the Alt+{number} shortcuts to to select cloze cards.

I remember reading something similar discussed in the forum (along with changing the Ctrl+{n} shortcuts to Ctrl+tab), but I can't find the thread now...

### DIFF
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -187,6 +187,12 @@ class CardLayout(QDialog):
             self,
             activated=self.tform.style_button.click,
         )
+        for i in range(min(len(self.cloze_numbers), 9)):
+            QShortcut(  # type: ignore
+                QKeySequence(f"Alt+{i+1}"),
+                self,
+                activated=lambda n=i: self.pform.cloze_number_combo.setCurrentIndex(n),
+            )
 
     # Main area setup
     ##########################################################################


abdnh:
Oh, this seems to conflict with access keys to select the card type. (They don't work reliably for me though)

dae:
Perhaps you could only enable these shortcut keys when the note type is a cloze type,  and only enable the existing alt shortcuts when it's a normal note type?

abdnh:
They don't conflict actually, because the card type combo box is disabled for cloze types (I forgot about that when writing my previous comment).

(Sorry for force-pushing too much uselessly)

qt/aqt/clayout.py
```diff
@@ -187,6 +187,12 @@ def setupShortcuts(self) -> None:
             self,
             activated=self.tform.style_button.click,
         )
+        for i in range(min(len(self.cloze_numbers), 9)):
```
dae, line 190: Perhaps this needs to be 8, not 9?

dae: ...nevermind, it's being passed to range()

dae:
Thanks Abdo!

PR #1035, opened 2021-02-20
change _on_click_current to _on_click_index

BlueGreenMagick:
This PR fixes a bug where clicking on the blank space below sidebar items would still trigger currently selected item's click event. 

This is most noticeable when you have `Saved Searches` selected(clicked). clicking anywhere not an item would trigger the click event, which in turn triggers the context menu event.

As `_on_click_current` is supposed to be an internal method and it's a brand new method, removing it shouldn't affect any add-ons.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -35,7 +35,7 @@ zjosua <zjosua@hotmail.com>
 Arthur Milchior <arthur@milchior.fr>
 Yngve Hoiseth <yngve@hoiseth.net>
 Ijgnd
-Yoonchae Lee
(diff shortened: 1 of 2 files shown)


BlueGreenMagick:
This PR introduces a minor bug though. While currentIndex is changed upon mousePress, the click event is fired upon mouseRelease. That makes it possible to change currentIndex with the keyboard keys(up down arrows) after mousePress but before mouseRelease. In that situation, the item click event was fired upon will be different from selected item.

dae:
Thanks BGM! That bug doesn't sound too serious, and we can store it in a variable in the future if need be.

PR #1036, opened 2021-02-22
Source graphs.js as library in the HTML head

hgiesel:
This is a small change that was required, to make `graphs.js` as a library that can be sourced in the HTML head.

Functionally, there will be no difference, but it allows for more flexibility with the library.
I gotta admit, my motivation here is also personal. I wrote an add-on which allows you to include the graphs section into other web views, and it works quite well:

image

I also have a question concerning your view on the future of Anki. Svelte does allow for both client-side, and server-side rendering. Right now, Anki only utilises client-side rendering. If it were to stay that way, we could consider using the [Custom Elements API](https://svelte.dev/docs#Custom_element_API) from Svelte to truly encapsulate graphs component, and make them reusable. Right now, global CSS will affect the svelte components, and the graphs page even specifically uses that feature. I think moving to encapsulated components would also be beneficial for the future, once more components move to Svelte.

### DIFF
--- a/ts/graphs/graphs.html
+++ b/ts/graphs/graphs.html
@@ -4,30 +4,31 @@
         <meta charset="utf-8" />
         <meta name="viewport" id="viewport" content="width=device-width" />
         <link href="graphs.css" rel="stylesheet" />
+        <script src="../js/vendor/protobuf.min.js"></script>
+        <script src="graphs.js"></script>
     </head>
     <body>
         <div id="main"></div>
+
+        <script>
+            anki.graphs(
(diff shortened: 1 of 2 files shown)


dae:
Change looks good, thanks.

I don't see SSR happening any time soon, so that's not a blocker. I don't have any experience with Custom Elements though, so don't know how they would work in practice/what if any issue they would cause.

hgiesel:
Well, I'm also not sure how it's going to work out either, but at least that makes it worth exploring :)

PR #1037, opened 2021-02-23
Remove logic for pulling editor field into view

hgiesel:
There's a bug in current beta concerning this.
In the browser, if you select a note, and select its field, select another note, then select the first field again, it will not enable the toolbar buttons.
This bug is not exclusive to the browser though, it's just easier to reproduce.

The whole issue here is the refocusing. When clicking away from a window, and clicking on a editor field again, we get two scenarios:
1. User clicks on the same editor field, in which case we get one `FocusEvent`, from `null` to previous/new focus.
2. User clicks on a different editor field, in which case we get two `FocusEvent`, one focusing from `null` to the previous focus, and one from previous focus to new focus.

There is no way to distinguish the `FocusEvent` of scenario one, with the first `FocusEvent` from scenario two. So we always have to err on one side. Previously I erred on not focusing. This new version errs on the side of focusing, which will make sure the toolbar buttons activate, but it makes it impossible to use the "scroll field into view" code, because it will be activated for the "refocus" first event of scenario two as well. Even though it was nice, I also don't think it's that valuable. A focus change by clicking tab will still move the window, so it was only used, when the user clicked an editor field barely in view.

I also researched a bit what explains this weird behavior, and surprisingly it's not Qt specific, it's just how browser engines work. When clicking on an inactive browser window, it will also refocus the previous focus first.

Closes hgiesel/anki_collapsible_fields#6

### DIFF
--- a/ts/editor/focusHandlers.ts
+++ b/ts/editor/focusHandlers.ts
@@ -1,30 +1,11 @@
 /* Copyright: Ankitects Pty Ltd and contributors
  * License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html */
 
-import type { EditingArea, EditorField } from ".";
+import type { EditingArea } from ".";
 
 import { bridgeCommand } from "./lib";
 import { enableButtons, disableButtons } from "./toolbar";
 import { saveField } from "./changeTimer";
-
-enum ViewportRelativePosition {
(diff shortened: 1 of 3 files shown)


dae:
Thanks Henrik, sounds like the best path forward.

PR #1038, opened 2021-02-23
Fix scrollbar issue in anki-editable component in two ways

hgiesel:
This addresses what you mentioned [here](https://forums.ankiweb.net/t/anki-2-1-41-beta/7305/49?u=hengiesel)

- Using `:host-context(.nightMode)` allows for applying the night mode
  scroll bar inside the component. `:host-context` seems to be particular suited for these theming issues, even [the examples on Mozilla mention it](https://developer.mozilla.org/en-US/docs/Web/CSS/:host-context()). For this I moved the scrollbar CSS to its own file.

- Apply `max-width: 100%` to all element within editable, not just images. You can still set a div to `max-width: 20000px; width: 20000px;`, in which case it will be bigger than the field, but the scrollbar will then be in night mode.

I've done some tests with tables, and it works very well. The table will be bigger than the field, only if necessary, and the scroll bars show up in the correct color.


dae:
Thanks Henrik!

It looks like Safari does not support host-context, so this code won't have any effect if it gets used in AnkiMobile in the future: https://caniuse.com/?search=host-context. That might not be a problem in this case, as the default scrollbar colours automatically follow the system theme on iOS. But something to bear in mind if you were planning to use host-context for other uses in the future.

hgiesel:
It also seems like `:host-context` will [never be supported in Safari](https://github.com/w3c/csswg-drafts/issues/1914). It seems like "[styling hooks](https://developers.google.com/web/fundamentals/web-components/shadowdom#stylehooks)" are an adequate replacement.

PR #1039, opened 2021-02-24
Clickable hint in dyndeck dialogue for unmovable cards

RumovZ:
This is a slighty different approach than #1020 as it shows unmovable cards matching the filter(s) instead of the eligible cards. I tried to take into account all of our reasoning and catch every case where a card is not movable. What do you reckon, @dae and @abdnh?

On a related note, we need to handle the case where the browser modifies the deck that is currently opened in a dyndeck dialogue. In the worst case, the deck is deleted and Anki has to be closed via the OS because the modal dyndeck dialogue cannot be exited.
Also, it's a bit weird to see the new filtered deck in the browser (or in the main window if it somehow gets resetted) while the dialogue is still open. When I reworked dyndeckconf, I kept the workaround of creating the new deck on setup because it's convenient for loading the default dyndeck settings. But now that concurrent usage of this dialogue and other parts of Anki is getting more common, it would be much nicer if the new deck wasn't created until the user had really hit `Build`.

### DIFF
--- a/ftl/core/decks.ftl
+++ b/ftl/core/decks.ftl
@@ -31,6 +31,7 @@ decks-reschedule-cards-based-on-my-answers = Reschedule cards based on my answer
 decks-study = Study
 decks-study-deck = Study Deck
 decks-the-provided-search-did-not-match = The provided search did not match any cards. Would you like to revise it?
+decks-unmovable-cards = Some cards may be excluded despite matching the search.
 decks-it-has-card =
     { $count ->
         [one] It has { $count } card.
--- a/qt/aqt/dyndeckconf.py
+++ b/qt/aqt/dyndeckconf.py
@@ -185,12 +185,12 @@ class DeckConf(QDialog):
         else:
             aqt.dialogs.open("Browser", self.mw, search=(search,))
 
-    def _second_filter(self) -> Tuple[str]:
+    def _second_filter(self) -> Tuple[str, ...]:
         if self.form.secondFilter.isChecked():
             return (self.form.search_2.text(),)
         return ()
 
-    def _learning_search_node(self) -> Tuple[SearchNode]:
+    def _learning_search_node(self) -> Tuple[SearchNode, ...]:
         """Return a search node that matches learning cards if the old scheduler is enabled.
         If it's a rebuild, exclude cards from this filtered deck as those will be reset.
         """
--- a/qt/aqt/dyndeckconf.py
+++ b/qt/aqt/dyndeckconf.py
@@ -40,12 +40,13 @@ class DeckConf(QDialog):
 
         QDialog.__init__(self, mw)
         self.mw = mw
+        self.col = self.mw.col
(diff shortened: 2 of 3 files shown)


abdnh:
I think this looks like a better approach.
A really minor nitpick: should not the period in the "Some cards may be excluded despite matching the search." message be removed? Not sure if there is a convention about such a thing, but it looks out of place to my eyes. Sorry for bikeshedding :)

RumovZ:
I assume that a complete sentence requires a full stop, but I'm not emotional about it. 😉

abdnh:
Yes, but I feel there is a tendency to leave full stops out in labels, especially clickable ones. Anyway, I'm not a UI designer :)

dae:
I probably am not always consistent with this myself, but I would lean towards a period here - similar examples are in the preferences and add-ons screen about changes not applying until a restart. 

I like this approach, though I fear it's still somewhat confusing when the filtered deck already has some cards in it. Maybe the search for is:filtered could be adjusted to exclude cards in the current filtered deck?

col.decks.new_deck_legacy() returns an empty filtered deck not committed to the DB which you could use. We'd probably still want to add the deck prior to building it (to avoid it becoming a bigger change, and so we don't need to search twice), but in the no-matches case it could just delete the deck again immediately.

RumovZ:
>  I like this approach, though I fear it's still somewhat confusing when the filtered deck already has some cards in it. Maybe the search for is:filtered could be adjusted to exclude cards in the current filtered deck?

Not sure I am understanding correctly, doesn't it do just that? So for a rebuild, the search string would look like that:

    ("deck:Standard" AND "is:due") AND ("is:suspended" OR "is:buried" OR ("is:learn" AND -"deck:Filtered Deck 1") OR ("deck:filtered" AND -"deck:Filtered Deck 1"))

dae:
Ah, I only tested in the newly created deck case, where those were not added. Sounds good to me then - thanks!

dae:
I ended up tweaking the wording to something shorter, and it allows us to omit the period - any objections to the current version?

RumovZ:
Hm, wasn't the "despite matching your search" part the important bit?

dae:
The user's search is the first part of what's placed into the browser, so my thinking was that it should be fairly clear what is being shown?

RumovZ:
Once you click on the link it's certainly clear. But is the shortened text enough to make the many users asking why their filtered deck is missing cards click it? I'd prefer a word like "unmovable" to emphasise that these cards will be excluded regardless of the search.
But I'm probably overthinking this. 🙂

dae:
I'm assuming that when cards are not included that the user expects, they'll return to that dialog and look at the available options, and see that message. We may want to consider changing the message that is shown when no cards match in the future, as that presents an opportunity to explain things in more detail. But I lean towards "starting small" - it's easy enough to make the message more verbose in the future if needs be, and IMHO, the wordier messages are, the harder it is to see everything at a glance.

PR #1040, opened 2021-02-24
Catch deck rename errors and localise warnings

### LINKED ISSUE #1024: Clean up filtered deck error messages
https://forums.ankiweb.net/t/error-message-for-children-of-of-filtered-deck/7488

RumovZ:
Fixes #1024.

### DIFF
--- a/pylib/anki/errors.py
+++ b/pylib/anki/errors.py
@@ -47,7 +47,15 @@ class ExistsError(Exception):
     pass
 
 
-class DeckIsFilteredError(Exception):
+class DeckRenameError(Exception):
+    """Legacy error, use DeckIsFilteredError instead."""
+
+    def __init__(self, description: str, *args: object) -> None:
+        super().__init__(description, *args)
+        self.description = description
+
+
+class DeckIsFilteredError(StringError, DeckRenameError):
     pass
 
 
@@ -78,7 +86,7 @@ def backend_exception_to_pylib(err: _pb.BackendError) -> Exception:
     elif val == "exists":
         return ExistsError()
     elif val == "deck_is_filtered":
-        return DeckIsFilteredError()
+        return DeckIsFilteredError(err.localized)
     elif val == "proto_error":
         return StringError(err.localized)
     else:
@@ -95,12 +103,3 @@ class AnkiError(Exception):
 
     def __str__(self) -> str:
         return self.type
-
-
-class DeckRenameError(Exception):
-    def __init__(self, description: str) -> None:
-        super().__init__()
-        self.description = description
-
-    def __str__(self) -> str:
-        return f"Couldn't rename deck: {self.description}"
--- a/ftl/core/errors.ftl
+++ b/ftl/core/errors.ftl
@@ -1,3 +1,5 @@
+errors-invalid-deck-name = Invalid deck name: { $reason }
 errors-invalid-input-empty = Invalid input.
 errors-invalid-input-details = Invalid input: { $details }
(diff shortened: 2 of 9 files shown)


dae:
Appreciate you taking a stab at this. There are two things I'm not so fond of in the current approach:

- it pushes the complexity of providing the correct i18n'd error message to the frontend, so each time the error can occur, the frontend needs to check for it and provide the appropriate message, which is error-prone, and more work. If we write an appropriate i18n string to the 'localized' property of BackendError, then the frontend does not need any special-casing, and can just catch the error and call showWarning(str(the_exception)). See handle_sync_error() for an example of how we are catching error types that need special-casing, and then just printing the localized message in other cases. DeckIsFilteredError will need to subclass from StringError, and have localized passed into it.
- it's doing comparisons against a string error, which is error-prone. That's not your fault - DeckRenameError() is being called with a string argument there. It used to cover other cases like a duplicate name being provided, but the backend code just tolerates that now, so the filtered deck case is the only time it will be instantiated. Some add-ons import that name and rely on the 'description' property, so we probably shouldn't get rid of it for now - but we could change its __str__ to behave like a string error, and pass localized to it in decks.py:update() so that it's functionality equivalent to raising DeckIsFilteredError directly.

RumovZ:
I guess the reason I chose a kind of hacky solution is the following `decks.py` method that I didn't want to touch because I didn't fully understand its intention:

``` Python
def update(self, g: Deck, preserve_usn: bool = True) -> None:
    "Add or update an existing deck. Used for syncing and merging."
    try:
        g["id"] = self.col._backend.add_or_update_deck_legacy(
            deck=to_json_bytes(g), preserve_usn_and_mtime=preserve_usn
        )
    except DeckIsFilteredError as exc:
        raise DeckRenameError("deck was filtered") from exc
```

Is there a reason it forces `DeckIsFilteredError` to the more general `DeckRenameError` or can I just remove the `try` clause?

dae:
DeckRenameError() was added 9 years ago, and at the time there were multiple errors that could happen. Existing code and add-ons reference it, so that catch is to map the new, single backend error to the legacy error. The legacy code references the DeckRenameError name, and the .description property on it. So we could just change its __str__ to be the same as StringError, and create it with DeckRenameError(str(exc)) to effectively make it behave like a normal string error, without breaking existing code.

pylib/anki/errors.py
```diff
@@ -47,7 +47,15 @@ class ExistsError(Exception):
...
-class DeckIsFilteredError(Exception):
+class DeckRenameError(Exception):
+    """Legacy error, use DeckIsFilteredError instead."""
+
+    def __init__(self, description: str, *args: object) -> None:
+        super().__init__(description, *args)
+        self.description = description
+
+
+class DeckIsFilteredError(StringError, DeckRenameError):
```
RumovZ, line 58: What do you think about this solution? It means we can use `DeckIsFilteredError` just like any other `StringError`, but legacy code can still catch it as `DeckRenameError` and access its `description` field.

RumovZ:
By the way, somewhere between 2.1.40 and 2.1.41b2 it became possible to make filtered decks parent decks by drag and drop.

dae:
Looks good, thanks for that. In most cases where we're calling a backend method, stack traces are not very useful, so we probably don't need to check for a specific error first, and could just catch any exception - the specific-error checks are mainly useful for when we need to take specific action for a specific error.

Thanks for catching the filtered deck issue; will fix.

Issue #1011, opened 2021-02-08
Multiple selection in the browser sidebar


dae:
Multiple selection is currently disabled.

https://github.com/ankitects/anki/commit/126b64211566b473d1c2d25fbef9f79ddd062f1d

There are two things we'll need to fix before we can turn it on:

- [ ] shift+clicking to OR items leads to unwanted selections. Maybe we need to way to toggle multiple selection on and off? Other ideas on how this could be solved welcome.
- [ ] most of the right click context menu actions in the sidebar assume they should operate only on the right-clicked item. In some cases that's the only thing that makes sense (eg rename), but other actions like "delete deck" would ideally function on the whole selection. 
  - [ ] This will require updating the Rust end to support a list of IDs instead of a single ID, eg remove_deck_and_child_decks() will need to change. While calling it in a loop from the Python end would work for now, that will make life harder when we move undoing to the backend in the future.
  - [ ] Will need to handle cases where the user has selected a range of different kinds of items, and ignore ones not appropriate for the action

dae:
I think the first one needs solving before we can enable this; the second one could potentially be done later.

dae:
Sorry, fat fingers. :-)

kleinerpirat:
Toggling multiple selection on and off seems cumbersome.
Maybe I'm missing something, but

- Ctrl+click for AND
- Ctrl+Shift+click for OR

seems like the simpler option? One would have to prevent multiple selection when Ctrl+Shift are pressed simultaneously. If more than one node is selected, they should be chained with OR in search to match the visual clue of the highlight.

People who would use logical operators in search are probably power users, so I reckon they are already used to more complex keyboard commands.

RumovZ:
I would prefer not breaking the current filter clicking because it's an established and arguably more useful feature.
Multiselection in the sidebar could be implemented for Ctrl+*Right*click. We would even have Shift+Rightclick available for the probably less useful but expected range selection. The context menu would only show for Rightclick without modifiers then.

dae:
Feedback on the current approach (or better ideas) welcome :-)

kleinerpirat:
Using right click for AND/OR/NOT is a neat way of preventing the conflict. I like the simplicity of this!

I'm not going to be a frequent user of the logical operators, but I imagine that some color coding for the node background (like green for AND, blue or yellow for OR and red for NOT) would be helpful for long queries where the search bar is filled up.

dae:
[...]

PR #1044, opened 2021-02-25
Add sidebar modes for different click behaviour

RumovZ:
2Cqeylgwh9

See #1011.
This PR introduces different sidebar modes, an icon toolbar to switch between them and various context actions for selections.

The Modes
---
1. Searching: Click to search, single selection, no drag and drop. (Same behaviour as on previous stable releases.)
2. Selecting: Click to select, multi selection via dragging or keyboard modifiers, no drag and drop.
3. Editing: Click to select, but no multi selection, drag and drop enabled.

I restricted the drag and drop feature to a separate mode because it is conflicting with drag selecting and, more importantly, I think that it can be triggered pretty easily by accident. It already happened to me, and while I could simply undo it, a user not aware of that feature might run into trouble later. He might not even make the connection between the deck tree on the start page and the one in the browser.

The Toolbar
---
I'm not perfectly happy with its position as it leaves a lot of horizontal space empty. Maybe someone has a better idea? Here are some things I've tried that didn't work: left or right of the searchbar and tree (vertically), at the bottom, between searchbar and tree.
I'll try to make the icons a bit prettier when it's decided how they will be used. Or maybe there's even someone who can do it less amateurishly?

Context Menus
---
Searches can now be triggered from the context menu, supporting groups and keyboard modifiers. Other group actions are on the todo list.
We could have mode-specific context menus, but currently, I don't see any benefit in it.


dae:
Looks like a good start!

The tag and deck drag&drop code was written to work with multiple items being dragged at once - imagine the user selecting a bunch of sibling decks, and then dragging them onto a different parent for example. With the three current modes, this would no longer be possible.

I appreciate the problem you're trying to solve, and I too found it a bit confusing when dragging sometimes selects and sometimes drags. I just wonder if there might be another way to solve this doesn't require choosing one or the other. If we could disable the 'dragging selects multiple items' behaviour (always requiring a modifier key), then the behaviour would at least be consistent between items that are draggable and ones that aren't. But I don't know how easy that would be. 

I also wonder whether we really need three different modes or whether we could somehow get away with two, though I guess that sort of depends on the earlier issue as well. Agreed that users can accidentally drag & drop things, so some sort of protection - be it via non-standard mode, modifier key or confirmation dialog/popup might be good.

The buttons are too prominent IMHO. My immediate thought would be to make them smaller and put them to the right of the sidebar search bar - was that one of the things you were saying you tried and didn't work?

RumovZ:
> The tag and deck drag&drop code was written to work with multiple items being dragged at once - imagine the user selecting a bunch of sibling decks, and then dragging them onto a different parent for example. With the three current modes, this would no longer be possible.

If we prevent clearing the selection on click in Edit mode or even enable multi selection via modifier, that would not be a problem.

I, too, feel like the third mode doesn't fully justify its existence. The d&d conflict was the only reason I added it (and maybe because the toolbar looked a bit empty). But maybe more editing features would make it worthwhile? Doubleclick-to-rename would be a neat feature, e.g.

I really think mouse selection is the more intuitive and useful feature here. It enables users to do a lot without using the keyboard and can't do any harm.

> The buttons are too prominent IMHO. My immediate thought would be to make them smaller and put them to the right of the sidebar search bar - was that one of the things you were saying you tried and didn't work?

toolbar_right

Reducing icon size and margins could be an improvement, but it still wouldn't look like the toolbar is part of the sidebar in my opinion. Also, isn't width more valuable than height in the browser?
I would love to see the toolbar floating at the right top corner of the tree view as there is usually a lot of space right of the standard filters, but I don't think Qt can do that?

dae:
I was suggesting replacing the existing sidebar filter input area with an hbox(sidebarfilter, buttons), so the only change would be a reduced area to type in a sidebar search

RumovZ:
Ah, gotcha! I'll try that.

kleinerpirat:
> If we prevent clearing the selection on click in Edit mode or even enable multi selection via modifier, that would not be a problem.

Requiring the use of two different modes (Select + Edit) to achieve multiple drag and drop seems off. Like dae said, you could remove one of the three modes and add a keyboard modifier instead of "drag-to-select-multiple" to prevent confusion.

It felt like you guys were so close to the goal of getting every feature in the sidebar without different modes, so it surely is possible with two:

- Advanced search
- Editing tag/deck hiearchies

_Love how you went and implemented the feature you talked about in the forum. Feeling bad because I can only critique here but not bring forth a solution myself. I'm not familiar enough with the search features to be able to consider every requirement here._

RumovZ:
> Love how you went and implemented the feature you talked about in the forum. Feeling bad because I can only critique here but not bring forth a solution myself. I'm not familiar enough with the search features to be able to consider every requirement here.

Don't feel bad! I think feedback from people who are not deep into the code is very valuable, that's why I try to include GIFs.
Of course, that doesn't mean I agree with you. 😉 

Binding drag-selection to a modifier doesn't make sense because you can already Shift-click to select ranges. So we could disable it altogether, but that doesn't make drag&drop any safer. As a safeguard, I prefer a separate mode for two reasons: 1. It leaves room for future features. 2. While preventing accidental triggering, users can deliberately enable it and then reorder items to their heart's content without further obstacles, like a confirmation request for every operation.

RumovZ:
Here's an update on what I've added in the meantime:

- Multiple decks can be deleted. The backend takes a vector of ids, now. Backwards compatibility is ensured on the frontend.
- Multiple saved searches can be delted. As before, that's done on the frontend and not reversible.
- Multiple tags can be deleted. I've created new backend functions for this so it's all performed in one transaction, we only need to pass the tags and it's faster than using bulk_update (but still pretty slow).
- Items can now be edited directly in the sidebar without affecting the parents. For now, editing can be triggered via keyboard and in edit mode via double-click as well.

Questions
---
- If you are still not convinced of the edit mode, what could be an alternative? It seems to be impractical to differentiate between dragging selection and dragging items by a modifier because we would need to handle mouse movement or selecting manually as far as I can see.
- How should we handle deleting if items of different types are selected? I can think of four ways:
  1. Have one delete action for deleting everything that is selected and deletable.
  2. One action per selected and deletable item type (`Delete Decks`, `Delete Tags`...).
  3. Only deleting the current (right-clicked) item type is available. That's the current implementation.
  4. Deleting is only available if all selected items are of the same type.
- The other context actions are bit more straightforward. Searching is currently always available as long as there is at least one searchable item selected. I think collapsing/expanding could be handled the same way. Any other suggestions?
- Should we restore the selection or at least the current item after renaming?
- Should cards and notetypes also be deletable and editable in the sidebar?
- Previously, when deleting a deck, the user was asked something like "Delete deck X? It has N cards." I've kept that and added a new string for when multiple decks are being deleted: "Delete M decks with N cards?". I see two problems, though:

  1. It has two count arguments. Provided that's valid at all, translators would have to nest cases.
  2. Translators will be asked to translate the `deck_count=1` case even though it's never used.

    What's the best way to deal with this?

kleinerpirat:
I just wanted to clarify that with this

> you could remove one of the three modes and add a keyboard modifier instead of "drag-to-select-multiple" to prevent confusion.

I didn't mean you should bind _dragging selection_ to a modifier, but _remove_ it and leave only Shift-click (which is what I meant by that keyboard modifier). 

So basically the same thing that Damien said:

> If we could disable the 'dragging selects multiple items' behaviour (always requiring a modifier key), then the behaviour would at least be consistent between items that are draggable and ones that aren't. But I don't know how easy that would be.

Since that is default behaviour in all file browsers, anyone should know it. That's the way I would make room for multi-drag.

dae (commented):
I've only briefly skimmed through the code, but have left one small comment.

> If you are still not convinced of the edit mode, what could be an alternative? It seems to be impractical to differentiate between dragging selection and dragging items by a modifier because we would need to handle mouse movement or selecting manually as far as I can see.

My thinking was not to try support 'drag to select' at all. Click on start, shift+click on end to select a range doesn't conflict with dragging ops, so if we can disable the 'drag to select' functionality somehow, we should be safe.

I find the edit/select distinction a bit confusing, since you can select while in editing mode, and edit while in selecting mode.

> How should we handle deleting if items of different types are selected? I can think of four ways:

I suspect "I want to delete multiple types of things at once" would be a very niche use case, so maybe it makes sense to hide the delete option completely?

Maybe rename should also be hidden when there is more than one item selected?

> Should cards and notetypes also be deletable and editable in the sidebar?

I'd rather we funnel users into the card layout screen, as those actions will usually force a full sync.

> Previously, when deleting a deck, the user was asked something like "Delete deck X? It has N cards." I've kept that and added a new string for when multiple decks are being deleted: "Delete M decks with N cards?". I see two problems, though:

Yeah, this can be tricky for translators, as while Fluent supports this fine, the translating website requires dropping down into the raw syntax. 

How about a different i18n string for the multiple selection case - something like "Delete the selected decks and the xxx cards they contain?", where only xxx cards is singular/plural? The number of selected decks should be fairly obvious as the user just selected them.

RumovZ:
> My thinking was not to try support 'drag to select' at all. Click on start, shift+click on end to select a range doesn't conflict with dragging ops, so if we can disable the 'drag to select' functionality somehow, we should be safe.

That doesn't solve our problems, though: Qt seems to only know ExtendedSelection or no ExtendedSelection, without separate settings for mouse and keyboard input. If we enable ExtendedSelection we have to reimplement the mouse movement slot to catch drag events and suppress selection. If we disable it we have to reimplement clicking with modifiers and set the selection manually. I could investigate how feasible that is, but we still wouldn't have a safeguard for accidental drag&drop.

> I'd rather we funnel users into the card layout screen, as those actions will usually force a full sync.

Maybe we can at least support renaming? Apart from syncing, that should be a pretty innocuous, but currently not very accessible action.

dae:
The importing code currently treats renamed fields or templates as an unmergable change, so a bit more friction there may not be a bad thing at the moment.

We'd still have two separate modes for searching + editing/selecting, so drag&drop wouldn't be active by default. Re: how we could accomplish it, the two approaches I can think of that might be practical are:

1) make all items draggable, but cancel the drag event if it doesn't make sense - either on drag start or by specifying no valid targets. Any draggable item can't be multi-selected without a modifier, so this should effectively disable the drag-to-select functionality. A bit of a hack, but might be the easiest way to solve it.
2) Implement our own selection model (https://doc.qt.io/qt-5/qitemselectionmodel.html), or completely ignore Qt's selection handling and do our own highlighting/tracking of selected items

RumovZ:
> The importing code currently treats renamed fields or templates as an unmergable change, so a bit more friction there may not be a bad thing at the moment.

What exactly do you mean by friction? I would propose adding a warning before accepting the new name, however, when I rename models or templates Anki doesn't enforce a full sync. Does that also depend on something else?

> make all items draggable, but cancel the drag event if it doesn't make sense - either on drag start or by specifying no valid targets. Any draggable item can't be multi-selected without a modifier, so this should effectively disable the drag-to-select functionality. A bit of a hack, but might be the easiest way to solve it.

Not sure if that was exactly what you had in mind, but I opted to set the valid targets depending on the selection giving us full control over which items can be dropped on which. That's also an improvement to 2.1.41 where you can seemlingly drop decks onto tags and the other way around.
Now, all items are behave the same under mouse events and indicate that drag&drop is enabled, hopefully making accidents more unlikely.
I'm happy with that solution. Hope you are, too.

dae:
Seems to work well!

The icons look a bit out of place on macOS:
Screen Shot 2021-03-09 at 2 51 07 pm

The following seems to help:

```patch
diff --git a/qt/aqt/sidebar.py b/qt/aqt/sidebar.py
index 9d4f512a7..ede513fc0 100644
--- a/qt/aqt/sidebar.py
+++ b/qt/aqt/sidebar.py
@@ -287,7 +287,8 @@ class SidebarToolbar(QToolBar):
         self._action_group = QActionGroup(self)
         qconnect(self._action_group.triggered, self._on_action_group_triggered)
         self._setup_tools()
-        self.setIconSize(QSize(18, 18))
+        self.setIconSize(QSize(16, 16))
+        self.setStyle(QStyleFactory.create("fusion"))

     def _setup_tools(self) -> None:
         for row in self._tools:
```

Screen Shot 2021-03-09 at 3 15 30 pm

By friction, I mean requiring the user to jump through extra steps (eg, go into the manage note types screen) in order to rename the items. The renames don't force a full sync, but will prevent users from being able to import updates if reimporting a deck.

RumovZ:
> By friction, I mean requiring the user to jump through extra steps (eg, go into the manage note types screen) in order to rename the items. The renames don't force a full sync, but will prevent users from being able to import updates if reimporting a deck.

Wouldn't the best solution be to inform the user whenever he triggers that? Could be a bit tricky to decide when to show the hint, mind. Maybe just once per opened browser/model manager?

dae:
Well, the best solution would probably be to better handled the renames in the importing code :-) All I'm advocating for here is keeping the status quo (ie, not making it easier to rename those items than it currently is), and hopefully we can address the underlying issues when the importing code is updated in the future.

qt/aqt/sidebar.py
```diff
@@ -1078,47 +1246,110 @@ def _get_saved_searches(self) -> Dict[str, str]:
...
+        ]
+
+    def ask_remove_tags(self, tags: List[str]) -> bool:
+        count = len(
+            self.col.find_notes(
+                self.col.build_search_string(
+                    *(SearchNode(tag=tag) for tag in tags), joiner="OR"
+                )
+            )
+        )
```
dae, line 1348: Counting the number of notes matching the tag can be fairly slow on a large collection, especially if the user has a spinning disk. In the deck deletion case, we display the number of cards up front, because it may not be clear to the user that deleting the deck will result in cards (and also notes) being deleted, and also because they're comparatively cheap to count. But in this case there's no risk of anything but the tag being deleted - so I wonder if we need to be presenting the count to the user up front?

If you do think we should keep it, it would be better run in the background (eg via mw.taskman.run_in_background()), so that the UI doesn't freeze until the data becomes available. 

I pushed the initial work on multi-step undo support earlier today, and as that gets enabled for more of the codebase, it will allow us to be less conservative with confirmation prompts: https://goodui.org/#8

ftl/core/decks.ftl
```diff
@@ -2,6 +2,7 @@ decks-add-new-deck-ctrlandn = Add New Deck (Ctrl+N)
 decks-are-you-sure-you-wish-to = Are you sure you wish to delete { $val }?
 decks-build = Build
 decks-cards-selected-by = cards selected by
+decks-confirm-deletion = Are you sure you want to delete all selected decks including { $count } cards?
 decks-create-deck = Create Deck
```
dae, line 6: Wording suggestion:

... all selected decks, and the { $count →
  [one] { $count } card
  *[other] { $count } cards
} they contain?

Note we also need to list out the plurals in this and the other strings, as otherwise we'll get "1 cards" in English, and it will make life harder for the translators. "1 card" would be clearer than "{ $count } card", but that confuses some translators when languages have different plural rules.

qt/aqt/sidebar.py
```diff
@@ -918,95 +1030,132 @@ def _notetype_tree(self, root: SidebarItem) -> None:
...
+    def rename_deck(self, item: SidebarItem, new_name: str) -> None:
         deck = self.mw.col.decks.get(item.id)
         old_name = deck["name"]
-        new_name = getOnlyText(tr(TR.DECKS_NEW_DECK_NAME), default=old_name)
-        new_name = new_name.replace('"', "")
-        if not new_name or new_name == old_name:
-            return
+        new_name = re.sub(
+            re.escape(item.name) + "$", new_name.replace("\\", r"\\"), old_name
+        )
```
dae, line 1136: I had trouble following this - could we split it up and comment - something like the following if I'm understanding correctly?

```python
# escape any backslashes the user has provided
new_name = new_name.replace("\\", r"\\")
# modify the full old name, replacing the last component
new_name = re.sub(
            re.escape(item.name) + "$", new_name, old_name
        )
```

dae: And same for _rename_tag(). Maybe we could move it into a function?

dae (commented):
Thanks for all your work on this Rumo, it's looking good! Just a few points to clear up and then I think this should be good to go.

rslib/src/decks/mod.rs
```diff
@@ -440,22 +440,23 @@ impl Collection {
         self.storage.get_deck_id(&machine_name)
     }
 
-    pub fn remove_deck_and_child_decks(&mut self, did: DeckID) -> Result<()> {
+    pub fn remove_decks_and_child_decks(&mut self, dids: Vec<DeckID>) -> Result<()> {
```
dae, line 443: Grr, GitHub ate my comment for some reason.

This would be better as `dids: &[DeckID]`, and passing the vec in as a reference. Then code like the following further down:

```rust
        col.remove_decks_and_child_decks(vec![top.id])?;
```

can be written more easily as

```
        col.remove_decks_and_child_decks(&[top.id])?;
```

qt/aqt/sidebar.py
RumovZ: Apart from asking for confirmation if any notes were to be affected, my intention was to indicate to the user that this operation may take a while if a lot of notes needed to be updated. But I see now that it might do more harm than good.

Favouring reversability over confirmation sounds very reasonable, although undoing will only help you if you realise your mistake and if you delete a deck thinking it didn't contain any notes or a tag thinking it wasn't used, you might not find out until much later.
The logical consequence seems to be to remove the confirmation prompt but show the same information afterwards as in the form of an uninstrusive tooltip. The counts can then be retrieved from the backend without further cost.
What do you reckon?

ftl/core/decks.ftl
RumovZ: Thanks, I missed that. If you agree above, I will just delete these prompt strings and replace them with info strings as mentioned.

qt/aqt/sidebar.py
RumovZ: That was a bit hacky. I made the item constructor store the name prefix separately so we don't need re.

qt/aqt/sidebar.py
dae: Yeah, I think reporting the affected count at the end of the operation is best, as it avoids all the duplicate work. Some of the other backend methods like tags.bulk_add() and tags.rename() already do this.

For the operations that we think might take a while, we should probably be using taskman.with_progress() instead of run_in_background(), as that'll take care of showing a progress dialog as well.

ftl/core/decks.ftl
dae: Sounds good

qt/aqt/sidebar.py
RumovZ: I switched to `with_progress()`, but for the record: I see a progress dialogue either way and for some operations it freezes together with the  GUI.

ftl/core/browsing.ftl
```diff
@@ -70,14 +76,11 @@ browsing-question = Question
...
 browsing-queue-top = Queue top: { $val }
 browsing-randomize-order = Randomize order
-browsing-remove-current-filter = Remove Current Filter...
-browsing-remove-from-your-saved-searches = Remove { $val } from your saved searches?
 browsing-remove-tags = Remove Tags...
 browsing-replace-with = <b>Replace With</b>:
 browsing-reposition = Reposition...
 browsing-reposition-new-cards = Reposition New Cards
 browsing-reschedule = Reschedule
-browsing-save-current-filter = Save Current Filter...
```
RumovZ, line 80: I think this and `browsing-remove-current-filter` are remnants from the filter button, so I removed them. Correct me if I'm wrong.

dae: Tests should fail if you accidentally delete a key that's used, so we should be good :-)

qt/aqt/sidebar.py
dae: Which op is freezing? The progress window usually should pop up after a few seconds IIRC.

dae:
I feel bad about causing merge conflicts here. Would you mind resolving these conflicts and I'll merge this in? Any further tweaks that might be needed can come in a follow-up PR

RumovZ:
I'll do that right away. Had only one last commit in the pipeline. I assumed the arguments passed to `decks.rem()` were only there for legacy reason? I added a deprecation warning using a new decorator.

dae:
Yep, the extra arg was used for syncing in the past IIRC

RumovZ:
Tests pass and the conflicting actions seem to still work but I'd be surprised if that merge didn't break anything. So sorry in advance! 😉

qt/aqt/sidebar.py
RumovZ: It pops up but then everything freezes altogether. It doesn't seem to depend on the operation. Sometimes it works, but more often than not, the progress animation is frozen and Windows marks the application as unresponsive.

2PgqFliBWA

dae:
Looks good, thanks! I'll just need to update remove_deck_and_child_decks to add the undo support back

qt/aqt/sidebar.py
dae: Thanks for the repro, I've posted an update on https://github.com/ankitects/anki/issues/1065

PR #1046, opened 2021-02-28
Sticky icons in the editor window

hgiesel:
This PR will add sticky / frozen / pinned buttons directly into the Editor window.

 The biggest point of contention imo are the icons. I've included the [Bootstrap icons](https://icons.getbootstrap.com/?), because I've already used them in Previewer PR. Looking at the icons in `/qt/aqt/data/web/imgs`, Bootstrap icons contain equivalents for all of them except the superscript and subscript icons.

Alternatives would be:
1. Create icons ourselves.
2. [Font Awesome](https://fontawesome.com/icons?d=gallery&p=2&m=free) offers a wide selection of free icons as well, certainly more than the Bootstrap icons. It would contain equivalents for all icons in `/qt/aqt/data/web/imgs`. Font Awesome has both a free and a paid pro tier.

Other small changes this PR includes:
1. Split up `/ts/editor/index.ts` into four files
2. I had to make some changes to `editor.scss`, as well as the button generation code in `editor.py` to make it behave well with Bootstraps CSS
3. I reworked the code for updating the buttons states a little bit. Now, all buttons will be unhighlighted on blur.
4. I removed all last uses of jQuery from `/ts/editor`. It could now technically removed from `editor.py` without breaking it. Interestingly, on the initial load of the editor, loading jQuery takes 3-4 times as long as Bootstrap. However I've left it in for addon backwards compatibility.

This PR does not entirely cover the functionality of my add-on, as there are no shortcuts associated with toggling the sticky/frozen status of fields. Anki has so far no configuration settings for shortcuts, so I wouldn't want to add this. At the same time, I don't think the default shortcuts of Ze Frozen Fields (F9 and Shift+F9) are that good, that I would want to have them directly in Anki.

My suggestion here would be that we don't add any built-in shortcuts, and I can transform my "Ze Frozen Fields" add-on to something like "Sticky Fields Shortcuts", to offer the functionality for those who want it.

### DIFF
--- a/qt/aqt/data/web/css/BUILD.bazel
+++ b/qt/aqt/data/web/css/BUILD.bazel
@@ -32,6 +32,7 @@ filegroup(
         "core.css",
         "css_local",
         "editor",
+        "//qt/aqt/data/web/css/vendor",
     ],
     visibility = ["//qt:__subpackages__"],
 )
--- a/qt/aqt/data/web/css/vendor/BUILD.bazel
+++ b/qt/aqt/data/web/css/vendor/BUILD.bazel
@@ -1,0 +1,20 @@
+load("//ts:vendor.bzl", "copy_bootstrap_css", "copy_bootstrap_icons")
+
+copy_bootstrap_css(name = "bootstrap")
+
+copy_bootstrap_icons(name = "bootstrap-icons")
+
+files = [
+    "bootstrap",
+    "bootstrap-icons",
+]
+
+directories = []
+
+filegroup(
+    name = "vendor",
+    srcs = glob(["*.css"]) +
+           ["//qt/aqt/data/web/css/vendor:{}".format(file) for file in files] +
+           ["//qt/aqt/data/web/css/vendor/{}".format(dir) for dir in directories],
+    visibility = ["//qt:__subpackages__"],
+)
--- a/qt/aqt/data/web/js/vendor/BUILD.bazel
+++ b/qt/aqt/data/web/js/vendor/BUILD.bazel
@@ -1,4 +1,12 @@
-load("//ts:vendor.bzl", "copy_css_browser_selector", "copy_jquery", "copy_jquery_ui", "copy_protobufjs")
+load(
+    "//ts:vendor.bzl",
+    "copy_css_browser_selector",
+    "copy_jquery",
+    "copy_jquery_ui",
+    "copy_protobufjs",
+    "copy_bootstrap_js",
+    "copy_popperjs",
+)
 
 copy_jquery(name = "jquery")
 
@@ -8,11 +16,17 @@ copy_protobufjs(name = "protobufjs")
 
 copy_css_browser_selector(name = "css-browser-selector")
 
+copy_bootstrap_js(name = "bootstrap")
+
+copy_popperjs(name = "popperjs")
+
 files = [
     "jquery",
     "jquery-ui",
     "protobufjs",
     "css-browser-selector",
+    "bootstrap",
+    "popperjs",
 ]
 
 directories = [
--- a/qt/aqt/data/web/js/vendor/BUILD.bazel
+++ b/qt/aqt/data/web/js/vendor/BUILD.bazel
@@ -5,7 +5,6 @@ load(
     "copy_jquery_ui",
     "copy_protobufjs",
     "copy_bootstrap_js",
-    "copy_popperjs",
 )
 
 copy_jquery(name = "jquery")
@@ -18,15 +17,12 @@ copy_css_browser_selector(name = "css-browser-selector")
 
 copy_bootstrap_js(name = "bootstrap")
 
-copy_popperjs(name = "popperjs")
-
 files = [
     "jquery",
     "jquery-ui",
     "protobufjs",
     "css-browser-selector",
     "bootstrap",
-    "popperjs",
 ]
 
 directories = [
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -179,7 +179,7 @@ class Editor:
                 "colour",
                 tr(TR.EDITING_SET_FOREGROUND_COLOUR_F7),
                 """
-<span id="forecolor" class="topbut topbut--rounded" style="background: #000"></span>
+<span id="forecolor" class="topbut rounded" style="background: #000"></span>
 """,
             ),
             self._addButton(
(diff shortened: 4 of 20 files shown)


hgiesel:
After @kleinerpirat 's [suggestion](https://github.com/hgiesel/anki_frozen_fields/issues/3#issuecomment-787204819), I changed from a pin icon to a lock icon.

kleinerpirat:
Thanks for the quick reaction!

I think it would be for the best to combine your two approaches. Color differentiation is useful for knowing what's up at a quick glance, hence the different colors in my [suggestion](https://github.com/hgiesel/anki_frozen_fields/issues/3#issuecomment-787204819). 

image

With my design, it would not be good to go without striking color differences (locked and unlocked too similar), with the Bootstrap one you could go colorless of course - but just because you can, doesn't mean it's best right? :)

I'm asking you to consider giving the inactive (open) locks a gray color and the locked ones a highlight color. People are used to this from Glutanimate's Frozen Fields, where the inactive flake is also grayed out.

hgiesel:
@kleinerpirat I had given this some thought. I think in the end it wouldn't be too big of a difference, and either one could easily be implemented via add-ons from the other one (you don't even need to provide another icon, you just need to set the color via CSS).

Two things made me stick with these plain icons for now:
* Looking at the editor window right now, you only have two elements, which are not in one of the two base colors: the color button, and the color picker, both of which actually have to do with the concept of color. 
* You said it draws more attention to it, when it's colorful, but I wondered whether that is even a good thing. The main attention should still be on the field content. And if that's black/white text on white/black, then the icon might actually draw more attention than the content.

dae:
Thanks for your work on this Henrik, I'll look into merging this after 2.1.41 is out the door.

Re the icon, to me, a lock denotes "you cannot change this value", not "this value will stay", so I think a pin might be slightly better there. Were you drawn to the lock because it's clear when it's on or off, or was there some other reason?

The icons in the screenshots feel a little prominent, especially because they appear before the field text. What do you think about either moving them to the right of the text, or to the far right?

Not sure if it's a good idea or not, but one other idea I thought of is to hide the pins by default, and make clicking on the field name itself toggle it on or off. That way no more clutter than we currently have in the default case, and it becomes more obvious when a field is pinned. Or alternatively perhaps we could make the off state be quite greyed out, so it's reasonably subtle.

The fields/cards buttons also seem to have lost their rounding on macOS:

Screen Shot 2021-03-01 at 6 06 00 pm
Screen Shot 2021-03-01 at 6 05 16 pm

kleinerpirat:
From [the issue](https://github.com/hgiesel/anki_frozen_fields/issues/3#) on @hgiesel's repo:
> ... it's great that you don't even need to adjust colors to differentiate states.

I think that was the argument. I suggested the locks to him without knowing that you already had pins in mind, which are a great fit aswell! Both are superior to _freezing_ in my opinion.

> Re the icon, to me, a lock denotes "you cannot change this value", not "this value will stay"

This is a very good point. Also, there's a psychological aspect to this: Seeing a bunch of locks on your add-cards window could induce an uncanny feeling, even if it's just subconcious.

Only showing pins on stickied fields is a nice idea. The less clutter, the better.
But if you're going to hide them by default (undoubtedly the cleanest solution), you might want to show some sort of hint on hover, like "Pin this field". Otherwise, a lot of people could miss out on this great feature.

hgiesel:
> Were you drawn to the lock because it's clear when it's on or off, or was there some other reason?

I felt like pins are _slightly_ off. E.g. a pinned post in the forum stands out among the rest, as in _it's more important_, or _you need to pay attention to it_. Let's say we wanted to add the functionality for users to mark their favorite tags, so that they can easier find them; in that case I'd say pins are a great match.

I've looked through the bootstrap icon list again, and found that they actually have an icon literally called `sticky`, which displays a sticky note. I think that might be the best match, as it even implies the terminology which is used in the code.

I've implemented your suggestion solution with opacity, as I think it provides the best discoverability. Inactive icons are displayed with 10% opacity. Both active and inactive icons get 50% opacity on mouseover.

sticky-icons

Regarding the regression, that was caused by [Bootstrap's reboot](https://getbootstrap.com/docs/4.0/content/reboot/), fyi.

I think a tooltip would be a great idea, and we could just reuse the label from the fields dialog ("Remember last input when adding"), however I don't see a good way to include it. Maybe that has to wait till we refactor the buttons (and utilize `/ts/lib/i18n`?)

dae:
To me that icon denotes some sort of attached note, not persistence :-) I like that it's subtle when off though. Might post about this briefly on the forums so we can get more feedback.

dae:
Posted on https://forums.ankiweb.net/t/poll-icon-for-sticky-pinned-frozen-fields/7910

hgiesel:
The masses have spoken, and pin icon it is :)

dae:
Thanks Henrik :-) Will let the dust settle on 2.1.41 for a little bit, and then merge this in.

hgiesel:
After #1055, I noticed that upon opening the editor, the caret will be at the beginning of the first field, not at the end.

I added the fix for this to this PR. It now contains four changes to how focus is handled:

1. disabling buttons will clear button highlight
2. enabling button will set button highlight
3. `move caret to end` executed _before_ enabling buttons (so button highlight will be for actual position of caret)
4. move caret to end will also be executed if `previousActiveElement` is `null`, which will only be the case before the first `onBlur` was executed.

dae:
The pins seem to be inverted to me - I'd expect them to be almost hidden when fields are not frozen/sticky, and shown in bold when sticky is enabled. Was that your intention?

I expect we'll need to release .42 within a week or two to fix some bugs. Because this PR has changed a fair bit of code, I'm a bit nervous about including it without giving it a bit of time in beta testing. Is it much work to backport the caret-at-start fix into a separate PR?

hgiesel:
I've fixed the inverted icons.

The focus logic in this PR is now outdated, and it will need a rebase once #1060 is merged.

dae:
Thanks Henrik! 🚀

oakkitten:
unless i'm missing something, it seems that all fields have the sticky button—except for the tags field. the field seems to be also sticky. i understand the rationale, but i'm only tagging a few cards while adding them, and as the field is sticky i inadvertedly end up tagging cards that shouldn't be tagged. can the tags field also have the sticky button perhaps?

hgiesel:
@oakkitten
I totally agree, a sticky tag function would be helpful. It has already been suggested [here](https://forums.ankiweb.net/t/can-anki-add-sticky-tags-function/12005/5). And I can only repeat what I said there. After the new Tag Editor is settled, this will be one thing I tackle next.

PR #1047, opened 2021-03-01
Make clickable dyndeck labels fixed in size

RumovZ:
I had used text alignment instead of aligning the buttons themselves which led to the clickable areas extending much further than the actual labels.
This also gets rid of some horizontal spacing. I could add some spacers if you think that's necessary.


dae:
Thanks Rumo, looks good to me.

PR #1048, opened 2021-03-01
Fix focus handling when coming from top left buttons

hgiesel:
This fixes the bug reported [in the forum](https://forums.ankiweb.net/t/anki-2-1-41-beta/7305/52).


dae:
Thanks Henrik!

PR #1049, opened 2021-03-02
Include user styles in note editor

hgiesel:
This is a standalone PR of a feature I already suggested in #957.

It would automatically apply the user styles in the individual editor fields, making it more likely that the fields will match how the HTML will look in the reviewer.
Perhaps we could give it another look after 2.1.41. This PR is so we/I don't forget about it.


dae:

PR #1050, opened 2021-03-03
Choose addon to update

BlueGreenMagick:
This is a continuation of #668 

I've addressed the 4 points you suggested:
* Addon update specific widget instead of generic
* The checkbox state is remembered using `AddonMeta.update_enabled`
* For daily update checks, whether the window is showed is determined by whether `update_enabled` is `False` for all update-able addons. Users will have to deselect all addons for this window to not show up every day. I'm not quite sure if it is the best solution - if you have a better idea, I'd be happy to edit.
* You can open the AnkiWeb addon page via context menu. 
* Show update date. I experimented with QTreeWidget or QTableWidget to render the update date in seperate columns, but they all seem to add a lot more complexity to create a checkable list-like widget. As long as only update date is showed, I think it should be fine to just seperate them with whitespaces.

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -105,6 +105,7 @@ class AddonMeta:
     max_point_version: int
     branch_index: int
     human_version: Optional[str]
+    update_enabled: bool
 
     def human_name(self) -> str:
         return self.provided_name or self.dir_name
@@ -140,6 +141,7 @@ class AddonMeta:
             max_point_version=json_meta.get("max_point_version", 0) or 0,
             branch_index=json_meta.get("branch_index", 0) or 0,
             human_version=json_meta.get("human_version"),
+            update_enabled=json_meta.get("update_enabled", True),
         )
 
 
@@ -243,6 +245,7 @@ class AddonManager:
         json_obj["branch_index"] = addon.branch_index
         if addon.human_version is not None:
             json_obj["human_version"] = addon.human_version
+        json_obj["update_enabled"] = addon.update_enabled
 
         self.writeAddonMeta(addon.dir_name, json_obj)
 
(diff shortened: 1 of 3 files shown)


qt/aqt/addons.py
```diff
@@ -1132,6 +1136,164 @@ def download_addons(
...
+            update_enabled = addon_meta.update_enabled
+            addon_name = addon_meta.human_name()
+            update_timestamp = update_info.suitable_branch_last_modified
+            update_time = datetime.fromtimestamp(update_timestamp)
+
+            addon_label = f"{update_time:%Y-%m-%d}   {addon_name}"
+            item = QListWidgetItem(addon_label, self)
+            # Not user checkable because it overlaps with itemClicked signal
+            item.setFlags(Qt.ItemFlag(Qt.ItemIsEnabled))
+            print(update_enabled)
```
dae, line 1178: This may have been missed

qt/aqt/main.py
```diff
@@ -848,7 +848,7 @@ def maybe_check_for_addon_updates(self) -> None:
 
         if elap > 86_400:
             check_and_prompt_for_updates(
-                self, self.addonManager, self.on_updates_installed
+                self, self.addonManager, self.on_updates_installed, routine=True
             )
```
dae, line 852: I presume 'routine' is intended to distinguish automatic checks from the user clicking the 'check for updates' button? If so, perhaps a name like 'requested_by_user' might be clearer?

qt/aqt/addons.py
```diff
@@ -104,6 +105,7 @@ class AddonMeta:
     max_point_version: int
     branch_index: int
     human_version: Optional[str]
+    update_enabled: bool
```
dae, line 108: Perhaps something like 'show_in_automatic_update_checks' might be clearer?

dae (commented):
Looks good, just a few minor suggestions

qt/aqt/addons.py
BlueGreenMagick: Sorry about that! Thought I checked for prints before pushing.

qt/aqt/addons.py
dae: Ah, I didn't realise it was also used there. In that case, it sounds like it's ok as is - thanks!

PR #1052, opened 2021-03-04
Import Dialog: Do not change delimiter if cancel button is pressed

tqml:
## What?

In the current version of Anki (v2.1.40) the default delimiter is removed and replaced by TAB, even if the "cancel" button is clicked. Usually canceling an operation should not make any changes therefore I assume this is not intended behavior. 

Video showing the bug

## How to reproduce

Select `Import File`, select e.g. a CSV file and click the `Fields separated  by` button. A modal dialog should open. Close the modal dialog using `Cancel`. The preselected delimiter (Semicolon) is replaced by TAB.

## Fix

I made a minor change where the result of the modal dialog is checked and only if it the OK button pressed the delimiter is updated.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -75,6 +75,7 @@ Meredith Derecho <meredithderecho@gmail.com>
 Daniel Wallgren <github.com/wallgrenen>
 Kerrick Staley <kerrick@kerrickstaley.com>
 Maksim Abramchuk <maximabramchuck@gmail.com>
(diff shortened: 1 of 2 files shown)


dae:
The change seems fine, but please make sure the tests are passing

tqml:
Aah, I'm sorry, missed that, everything should be fine now :)

dae:
Thanks Benjamin!

PR #1055, opened 2021-03-06
Do not execute caretToEnd when Alt-tabbing back to application

### LINKED ISSUE #1054: Cursor position is lost and set at the end of line after text input field loses focus in editor and browser
This bug occurs using the latest `main` branch commit `3ba08d1189999c38f983288e980f39494c1c0057` and the latest beta version.

Cursor position before losing focus, underlined in red for clarity:
image

Cursor position after add window has lost focus (same bug occurs in browser editor too) i.e. after switching to another window:
image

Some more information on my system given that the WM and QT version might be of importance here:
KDE Plasma Version: 5.21.1
KDE Frameworks Version: 5.79.0
Qt Version: 5.15.2
Graphics Platform: X11

hgiesel:
Closes #1054.

This is very much related to #1037, and might even be a regression from it, even though I did not test it.
It also works with the fact, that when you refocus a window, the first `FocusEvent` has `relatedTarget === null`.

The fix solves the problem when you Alt-tab back to the application. I wouldn't know any other way to trigger this behavior, because otherwise you'd click on the window to reactivate it, in which case the clicking sets the new focus anyway.

### DIFF
--- a/ts/editor/focusHandlers.ts
+++ b/ts/editor/focusHandlers.ts
@@ -20,7 +20,6 @@ function focusField(field: EditingArea) {
     field.focusEditable();
     bridgeCommand(`focus:${field.ord}`);
     enableButtons();
-    caretToEnd(field);
 }
 
 // For distinguishing focus by refocusing window from deliberate focus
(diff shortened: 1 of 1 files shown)


sandersantema:
I've just tried the fix out and can confirm it fixes the bug :)

dae:
Thanks for the quick fix Henrik, and for testing sander - I'll slip this into the 2.1.41 release

PR #1057, opened 2021-03-07
Fix editor RTL check

abdnh:
isRightToLeft is checking inline styles instead of the base stylesheet.
This apparently broke with the editor refactoring.


dae:
Thanks Abdo!

PR #1060, opened 2021-03-08
Fix focus handling by moving caretToEnd logic

hgiesel:
Ok, let's hope this will be the last focus handling PR...

I think one way we made it unnecessarily complicated for us, was that we executed `caretToEnd` from the focus handler, even though it's not really connected to focusing. As I see it, we have two situations where we want to move the caret to the end of the field:

1. When we programmatically focus a field (e.g. on opening the editor)
2. When changing focus via Tab

I've moved `caretToEnd` out of `onFocus`, and moved it to these two places. This also allows us to vastly simplify `onFocus`.

### DIFF
--- a/ts/editor/changeTimer.ts
+++ b/ts/editor/changeTimer.ts
@@ -6,16 +6,12 @@ import type { EditingArea } from ".";
 import { getCurrentField } from ".";
 import { bridgeCommand } from "./lib";
 import { getNoteId } from "./noteId";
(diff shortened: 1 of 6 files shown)


dae:
Thanks Henrik. I've been frustrated with this area of the code too  - I can empathise!

dae:
(re the first tab starting at the start, that's not ideal, but at least we're moving in the right direction - let's see if bothers people in practice)

PR #1062, opened 2021-03-09
Strip HTML comments from external pastes

abdnh:
Fixes a regression caused by https://github.com/ankitects/anki/commit/150de7a683348198183166ddda5791eb5ff6bc28

This is especially annoying on Windows with the `<!--StartFragment --><!--EndFragment -->` thingy getting inserted in every HTML paste.

I suspect the code previously used to also strip many non-text nodes like CDATA sections, but I didn't bother with those here.


dae:
Thanks Abdo!

PR #1063, opened 2021-03-10
Fix bug with initially enabled toolbar in browser

hgiesel:
This fixes a regression from #1037, reported [here](https://forums.ankiweb.net/t/error-in-the-html-editor/8194/8).


dae:
Thanks Henrik!

PR #1064, opened 2021-03-11
Clicking the expander icon should not trigger click event

### LINKED ISSUE #1059: In browser, the deck tree's arrow open the deck
2.1.41 beta 7
Previously, in the deck browser, in the left pannel, when you click on an arrow near a deck name, it opened/closed the list of subdeck and that's it.
Now it also show the deck content. I believe it is a bug and should be reversed. Indeed:
* It loses current selection and search result and search bar content while I did not expected it too
* Loading a big deck takes 5 seconds on my computer. This totally stop the interaction and forbids to close everything/open every subdeck until I find the one I'm looking for.

BlueGreenMagick:
Fixes #1059

Uses the fact that `currentIndex` does not change when expander triangle icon is clicked. I *think* 1) clicking the expander icon and 2) clicking outside the sidebar item, are the only cases where `currentIndex` is different from sidebar item on mouse position.

I've also considered some other solutions:
* Use `currentChanged` for the searches. The search is done only when currentIndex is changed, not when the items are clicked. A drawback is that you can't re-click the selected item to re-trigger the search and update the notes list.
* Check if expander icon was clicked by mouse position. Qt does this by calling `itemDecorationAt`, but the method is not exposed. It is non-trivial to calculate if mouse position is on the icon.

### DIFF
--- a/qt/aqt/sidebar.py
+++ b/qt/aqt/sidebar.py
@@ -443,7 +443,8 @@ class SidebarTreeView(QTreeView):
         super().mouseReleaseEvent(event)
         if event.button() == Qt.LeftButton:
             idx = self.indexAt(event.pos())
-            self._on_click_index(idx)
+            if idx == self.currentIndex():
+                self._on_click_index(idx)
 
     def keyPressEvent(self, event: QKeyEvent) -> None:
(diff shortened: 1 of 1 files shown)


dae:
Thanks for looking into it BGM!

PR #1066, opened 2021-03-11
Add decorators for calling editor.saveNow()

RumovZ:
Here's a small cosmetic change I intended to make for a while.
I've only updated the trivial cases so far but in my opinion, even a few unnecessary saves would be acceptable if that means less lambda and proxy functions which decrease readability.

### DIFF
--- a/qt/aqt/browser.py
+++ b/qt/aqt/browser.py
@@ -47,6 +47,7 @@ from aqt.utils import (
     restoreState,
     save_combo_history,
     save_combo_index_for_session,
+    save_editor,
     save_is_checked,
     saveGeom,
     saveHeader,
@@ -581,12 +582,10 @@ class Browser(QMainWindow):
         self.mw.deferred_delete_and_garbage_collect(self)
         self.close()
 
+    @save_editor
     def closeWithCallback(self, onsuccess: Callable) -> None:
-        def callback() -> None:
-            self._closeWindow()
-            onsuccess()
-
-        self.editor.saveNow(callback)
+        self._closeWindow()
+        onsuccess()
 
     def keyPressEvent(self, evt: QKeyEvent) -> None:
         if evt.key() == Qt.Key_Escape:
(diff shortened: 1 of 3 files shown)


dae:
Looks like a nice improvement - thanks!

In the 2.0 → 2.1 upgrade process when we moved to webengine and async JS callbacks, there were lots of bugs caused by the blur event of a field firing after another operation had been initiated. If you can be sure that a field won't be focused when the action is initiated, then a decorator shouldn't be required. I've erred on the side of caution and perhaps used them in places they're not strictly required out of a desire not to relive that experience :-)

RumovZ:
My apologies for the troubles. Always a bad idea to quickly push out some changes before going to bed...
I'll look into it and make a new PR if I can find a way to fix this and the decorator is still a facilitation then.

dae:
Happens to the best of us :-)

RumovZ:
It's annoying that Python has tools to set a function's signature but there are no real consequences:

```Python
from functools import wraps
from inspect import signature

def decorator(func):
    @wraps(func)
    def decorated(*args, **kwargs):
        print("but no TypeError here despite unfitting signature")
        func(*args, **kwargs)
        print("but this ^ will raise a TypeError")
    print(f"decorated's signature is {signature(decorated)}")
    return decorated

@decorator
def f():
    pass

f("unexpected argument")
```

Also, a curious design choice by PyQt to send different arguments depending on the slot's signature. I couldn't find out what the purpose of that additional argument is.

I think the best solution would be to accept `*args: Any` in every slot (maybe even regardless of the decorator). The decorator could look at the decorated function's signature and strip any unaccepted arguments but that's messy and hard to follow. `*args: Any` on the other hand hopefully communicates that PyQt may decide to pass extraneous arguments and we explicitly ignore them.

I'll wait how your refactorings go and make the PR if it isn't redundant by then.

dae:
I'm still thinking through perform_op(). I think we need a general solution that's not specific to the browse screen,  and requireReset() also needs a rethink

RumovZ:
Maybe it would indeed be best if all writing to the collection was funnelled through the main window. It would then be responsible for setting checkpoints, showing progress and notifying all windows. The windows would in turn expose resetting methods.

dae:
Found a workaround: https://github.com/ankitects/anki/commit/6bbd1e615b0e53ec1cb8abf31bcbd4b5587b636d

RumovZ:
That's great that you've got it to work. Mystery solved!

PR #1067, opened 2021-03-12
Fix regex error formatting and search error escaping

RumovZ:
1. Reported here: https://forums.ankiweb.net/t/potential-bug-with-escape-sequences-on-regex/8260
9686cd99 assimilated the regex error into the Markdown-formatted search error breaking the plaintext error message from the regex crate. (As far as I recall, it wasn't properly displayed before, either, because no monospace font was used.)
2. Either `htmlescape::encode_minimal()` was used or Markdown wasn't escaped at all. Since there is no clean way to escape Markdown, I used a hack. But I think it's pretty good compared with the alternatives.


dae:
Thanks Rumo!

PR #1069, opened 2021-03-13
Emptying field lead to proper message

Arthur-Milchior:
This solve the following bug:

I have a note type with a field named "/" in a note type called Algebra
Today, I wanted to change a card type of algebra. However, when I tried to save, I got an error message stating that there was an empty field name. This made no sens, A quick search show that no field name is actually empty. Reading the code, I understood that "/" was stripped, leading to an empty field name, but the error message made no sens. So this correct that.

Similar correction is made to the template name

### DIFF
--- a/rslib/src/notetype/fields.rs
+++ b/rslib/src/notetype/fields.rs
@@ -1,7 +1,10 @@
 // Copyright: Ankitects Pty Ltd and contributors
 // License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
-use crate::backend_proto::{NoteField as NoteFieldProto, NoteFieldConfig, OptionalUInt32};
+use crate::{
+    backend_proto::{NoteField as NoteFieldProto, NoteFieldConfig, OptionalUInt32},
+    err::{AnkiError, Result},
+};
 
 #[derive(Debug, PartialEq)]
 pub struct NoteField {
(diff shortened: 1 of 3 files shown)


rslib/src/notetype/fields.rs
```diff
@@ -35,18 +35,22 @@ impl NoteField {
         }
     }
 
-    pub(crate) fn fix_name(&mut self) {
-        // remove special characters
+    pub(crate) fn fix_name(&mut self) -> bool {
+        // Return whether the name is valid. Remove special characters if it leads to a valid name.
```
dae, line 39: To make this a docstring, it should start with three slashes (///), and appear above the function name.

rslib/src/notetype/fields.rs
```diff
@@ -35,18 +35,22 @@ impl NoteField {
...
         let bad_chars = |c| c == ':' || c == '{' || c == '}' || c == '"';
         if self.name.contains(bad_chars) {
             self.name = self.name.replace(bad_chars, "");
         }
```
dae, line 43: The string is being replaced here, before the check.

rslib/src/notetype/fields.rs
```diff
@@ -35,18 +35,22 @@ impl NoteField {
...
         // and leading/trailing whitespace and special chars
         let bad_start_chars = |c: char| c == '#' || c == '/' || c == '^' || c.is_whitespace();
         let trimmed = self.name.trim().trim_start_matches(bad_start_chars);
+        if trimmed.is_empty() {
+            return false;
+        }
```
dae, line 49: It would be nicer to have fix_name return a Result<()> instead of a bool, as it's less surprising that the mutation doesn't happen when an error is returned.

dae (commented):
Thanks Arthur, just some small notes.

rslib/src/notetype/fields.rs
Arthur-Milchior: Thanks.

I'm really new to Rust, learned it for Anki. Nice to get feedback about the proper way. I'll correct it soon.

Arthur-Milchior: Done. And that also allow to use try_for_each, which is nice here

dae:
Thanks Arthur!

Arthur-Milchior:
You're welcome. Interesting to learn this language

PR #1071, opened 2021-03-13
Only delete by key when selected items have same type

RumovZ:
This ensures that deleting by key press only works if the same conditions are met we established for showing the delete action in the context menu. (Second commit does some trivial ordering.)
Renaming is still available by key press when multiple items are selected (the context action is not in this case). I think it's harmless enough.


dae:
Thanks Rumo!

Issue #986, opened 2020-12-01
Track latest tag list change time


dae:
If additions to the tag list were timestamped (perhaps in CollectionState), then register_tag() could update it. Combined with a backend method to poll the tag modification time, it would allow the UI to check for tag changes fairly cheaply.

https://forums.ankiweb.net/t/propagating-tag-changes-to-the-sidebar-and-to-add-ons/3075

https://forums.ankiweb.net/t/renaming-clear-unused-tags/5481/5

dae:
The use case is solved in the new undo/state handling code in 2.1.45

PR #1072, opened 2021-03-13
Experimental changes to resetting

dae:
Main points are https://github.com/ankitects/anki/commit/26d2907bc26e492def405e059a101c3cd70d8e89 and https://github.com/ankitects/anki/commit/3e716b724e8514c149ac374233a991a5f20af0fc - thoughts?


dae:
Also related to #986

RumovZ:
> The disadvantage of this approach is that the displayed content may
not reflect the specified search, which has the potential to be confusing.

Personally, I prefer it this way. It makes it easier to see the effects and if you want to perform multiple changes on the same notes, it's annoying if the notes disappear in between.
Also, the disadvantages you mention already exist in the current implementation: The search text can be modified independently of the table as long as you don't confirm and if you alter a note in the editor, the table data will update but the row won't ever be removed or reordered.

I sincerely hope this PR works out!

RumovZ:
I noticed 71a7bb0 is pretty much the same as what I'm already implementing to support `columnData` from the backend. I'll try to make my PR until tomorrow so we don't have to do work twice.
Do you want me to open it on main or on refresh? Or will you merge refresh in soon anway?

dae:
There's still more to do, but I've merged it into main now. I wrote that caching code with your work in mind - we should be able to replace the calls to columnData() and so on with a single backend call

RumovZ:
My API is a bit different, though (see below). The differences probably look more striking than they actually are, but maybe you can let me know if I can keep the structure when rewriting it for the new browser. Especially with regard to the following points:
* Storing font information on the row level
* Retrieving color data from the backend as an enum variant

<hr>

`collection.py`
```Python
browser_row = _pb.BrowserRow
...
    def browser_row_for_card(
        self, cid: int
    ) -> Tuple[Generator[str, bool], browser_row.ColorValue, str, int]:
        row = self._backend.browser_row_for_card(cid)
        return (
            ((cell.text, cell.is_rtl) for cell in row.cells),
            row.color,
            row.font_name,
            row.font_size,
        )
```

<hr>

`browser.py`
```Python
class BrowserCell:
    def __init__(self, text: str, is_rtl: bool) -> None:
        self.text: str = text
        self.is_rtl: bool = is_rtl


class BrowserRow:
    def __init__(
        self,
        cells: Tuple[Generator[str, bool]],
        color: browser_row.ColorValue,
        font_name: str,
        font_size: int,
    ) -> None:
        self.cells: Tuple[BrowserCell] = tuple(BrowserCell(*cell) for cell in cells)
        self.color: Optional[str] = backend_color_to_str(color)
        self.font_name: str = font_name or "arial"
        self.font_size: int = font_size if font_size > 0 else 12


def backend_color_to_str(color: browser_row.ColorValue) -> Optional[str]:
    if color == browser_row.COLOR_MARKED:
        return colors.MARKED_BG
    if color == browser_row.COLOR_SUSPENDED:
        return colors.SUSPENDED_BG
    if color == browser_row.COLOR_FLAG_RED:
        return colors.FLAG1_BG
    if color == browser_row.COLOR_FLAG_ORANGE:
        return colors.FLAG2_BG
    if color == browser_row.COLOR_FLAG_GREEN:
        return colors.FLAG3_BG
    if color == browser_row.COLOR_FLAG_BLUE:
        return colors.FLAG4_BG
    return None


# Data model


class DataModel(QAbstractTableModel):
    def __init__(self, browser: Browser) -> None:
        QAbstractTableModel.__init__(self)
        self.browser = browser
        self.col = browser.col
        self.sortKey = None
        self.activeCols = self.col.get_config(
            "activeCols", ["noteFld", "template", "cardDue", "deck"]
        )
        self.cards: Sequence[int] = []
        self._rows: Dict[int, BrowserRow] = {}

    def get_row(self, row: int) -> BrowserRow:
        cid = self.cards[row]
        if not cid in self._rows:
            self._rows[cid] = BrowserRow(*self.col.browser_row_for_card(cid))
        return self._rows[cid]
```

dae:
My code was just a quick fix until your code lands, and I didn't spend much time thinking it through. Row-based fonts and enums indeed look like a better solution, so please feel free to replace it as you see fit :-)

PR #1074, opened 2021-03-17
Restructure data/web/pages and /ts

hgiesel:
This is entirely a refactoring PR.

Right now `graphs.{ts,html}` and `congrats.{ts,html}` are strongly coupled in the same respective directory. Going forward, it won't necessarily be "one svelte lib <=> one html page" anymore. That is what this PR is dealing with.
* I moved the `graphs.html` and `congrats.html` directly into `data/web/pages`
* I renamed the svelte lib entrance files from `bootstrap.ts` to `index.ts`, which is more idiomatic
* I removed `graphs.scss` in an effort to avoid global CSS. I did so in three ways:
  * A lot of styling could just be moved directly into svelte components
  * Some of the styling was used across unrelated svelte components, or for `d3` in the `ts` files. For those I've installed the CSS-in-JS library [@emotion/css](https://emotion.sh/docs/@emotion/css). Emotion is mostly a React library, but @emotion/css is a framework-agnostic with a very small API, that was mentioned [on the Svelte blog](https://svelte.dev/blog/svelte-css-in-js). The new `graph-styles.ts` is where those styles are mostly located.
  * a tiny amount was moved into `core.css`
* Now both graphs and congrats are compiled to `qt/aqt/data/js` to signal that they are truly _components_ not bound to one page, but can be used on multiple pages, or combined with other components.


dae:
I like the direction you're going with decoupling these things, and emotion looks convenient. I notice that's also brought in Babel though, and the rollup step for the graphs has gone from ~2s on my computer to about ~7s, which is unfortunate. Can we accomplish this without Babel, perhaps using https://github.com/LeetCode-OpenSource/emotion-ts-plugin or the library without a plugin?

dae:
(FWIW, the Bazel nodejs rules introduced esbuild support recently. I'd love the extra speed esbuild supposedly offers, but haven't looked into whether we could use it to do what we're currently doing)

dae:
esbuild PR is here: https://github.com/ankitects/anki/pull/1081

On second thought, could we get away without adding @emotion/css? It looks like you can define "global" css in a top level component that will be inherited by children of the component - would that solve our use case?

https://svelte.dev/docs#style

hgiesel:
Using `:global` we'd need to put the CSS into unrelated components, e.g. putting the CSS for `.tick` (tickCustom) into possibly `Graph.svelte` or `GraphsPage.svelte`, similiar to `graphs.scss`. This makes it hard to track from where the styling was applied. With emotion, we need to import it from the source, which makes this easier to track.

It also breaks encapsulation, because now, the individual graphs would depend on being used within `Graph.svelte`, just like before they depended on being included alongside `graphs.scss`.

So altogether, we could skip it, but we'd also reintroduce two potential issues I tried to solve by removing `graphs.scss`.

dae:
Thanks for elaborating Henrik. I like most of this PR, and from a clarity/maintenance perspective, I can see the appeal of @emotion/css's functionality. I'm just not sure the gains we'd be getting from that part would be worth the performance tradeoff, and the extra libraries we're pulling in. Your PR has already moved the bulk of the styling into the relevant components which is great, and we only have a few styles left over that are shared among components. I'll do a bit more poking about to see if we can accomplish similar behaviour without emotion, but if it proves not to be practical, I'd lean towards just doing what we can for now, and hoping Svelte improves handling of this use case in the future.

dae:
I couldn't find an easy way to achieve the @emotion/css behaviour, but the bulk of the styling that was using it could just be placed in a component instead, eg https://github.com/ankitects/anki/commit/9dc6e059f232ebe3bb68abbc3da0d7f869d21a29

It would be nice if the page exporting the (scoped) global could export a constant so we don't have to identify the selector with a string, but unfortunately this seems to trip the tooling up - the svelte2tsx typings don't expect anything except a default export on *.svelte.

Hopefully it's not too far off what you were aiming for?

dae:
> It also breaks encapsulation, because now, the individual graphs would depend on being used within Graph.svelte, just like before they depended on being included alongside graphs.scss.

Ah, I just realised that your add-on is probably the motivating reason for this change? :-) Using the graphs outside of the Graph context is indeed something I hadn't really considered. I think it's a good idea in principal, just wish there was a lighter-weight way to solve it.

hgiesel:
Looks pretty good to me.
During my refactoring, `.tick`, `.tick-odd`, and `.clickable` was also the point where I realized that we couldn't entirely achieve it using the means Svelte provides us with.

However I wonder why if we couldn't get rid of `graphs.scss` after all, and move that to `Graph.svelte` as well.

>  I just realised that your add-on is probably the motivating reason for this change?

Not, entirely. It would be nice to include the graphs on their own, but there would still be a few steps left to realize that. :-) I'm mostly just exploring best or at least better practices, also in the interest of the other Svelte components I'm in the process of writing

PR #204, opened 2017-08-30
Improve keyboard navigation in TagEdit

glutanimate:
This pull request includes a number of changes that target the tag entry widget. As a whole, they are meant to improve the tag editing workflow when using the keyboard. Most of these modifications build upon features already found in the [Tag Entry Enhancements](https://ankiweb.net/shared/info/1348430474) add-on which has seen a lot of positive feedback and downloads over the past few months.

Specifically I have performed the following changes:

- With the tag entry widget focused, **Ctrl+Tab** will now allow users to **rotate** through the list of auto-suggestions. If the popup is not visible it will automatically be invoked. Up to this point users had to move their hands away from the home row in order to switch between entries using the arrow keys. With this new key assignment that will no longer be necessary.
- Also in line with this change is the assignment of the **Enter/Return** key to **applying the first available auto-suggestion**. This should allow users to enter their tags lightning-fast, especially when coupled with the next addition:
- Inserted tags are now **automatically separated by spaces**. Users will no longer have to move to the next tag manually, but can rather keep typing their tags fluently.
- Hitting the **space**-bar again will now also update the auto-suggestion popup and provide users with a **selection of all available tags**, just like when focusing the tag entry initially. I think this is much more desirable than the old behaviour where the popup would only show the previous completion until users typed in a character other than space
- Last, for users that are used to the **arrow keys**, this PR also includes a change that allows these keys to **invoke the popup** if it isn't visible, yet.

Taken together, these changes – albeit small – should have a major impact on the usability of the tag entry widget for keyboard-centric users.

PR #1079, opened 2021-03-19
Svelte Tag Editor

hgiesel:
This is a PR for a new Tag Editor implemented in the Editor in TypeScript. It's still far from complete. The idea behind it is, that the new tag editor could be shared across platforms.

- [x] Basic visuals + functionality
- [ ] Autocompletion menu
- [ ] Tag prefix search in Backend
- [ ] Connect to Python to save tags
- [ ] Remove old tag editor

### DIFF
--- a/qt/aqt/data/web/css/BUILD.bazel
+++ b/qt/aqt/data/web/css/BUILD.bazel
@@ -26,11 +26,20 @@ copy_files_into_group(
     package = "//ts/editor",
 )
 
+copy_files_into_group(
+    name = "tag-editor",
+    srcs = [
+        "tag-editor.css",
+    ],
+    package = "//ts/tag-editor",
+)
+
 filegroup(
     name = "css",
     srcs = [
         "css_local",
         "editor",
+        "tag-editor",
     ],
     visibility = ["//qt:__subpackages__"],
 )
--- a/qt/aqt/data/web/js/BUILD.bazel
+++ b/qt/aqt/data/web/js/BUILD.bazel
@@ -37,11 +37,20 @@ copy_files_into_group(
     package = "//ts/editor",
 )
 
+copy_files_into_group(
+    name = "tag-editor",
+    srcs = [
+        "tag-editor.js",
+    ],
+    package = "//ts/tag-editor",
+)
+
 filegroup(
     name = "js",
     srcs = [
         "aqt_es5",
         "editor",
+        "tag-editor",
         "mathjax.js",
         "//qt/aqt/data/web/js/vendor",
     ],
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -88,7 +88,7 @@ _html = """
     <div id="dupes" class="is-inactive">
         <a href="#" onclick="pycmd('dupes');return false;">%s</a>
     </div>
-    <div id="tagEditor"></div>
+    <anki-tageditor></anki-tageditor>
 </div>
 """
 
@@ -465,11 +465,12 @@ $editorToolbar.addButtonGroup({{
                 self.web.setFocus()
             gui_hooks.editor_did_load_note(self)
 
-        js = "setFields(%s); setFonts(%s); focusField(%s); setNoteId(%s); setTagEditor(); " % (
+        js = "setFields(%s); setFonts(%s); focusField(%s); setNoteId(%s); setTags(%s); " % (
             json.dumps(data),
             json.dumps(self.fonts()),
             json.dumps(focusTo),
             json.dumps(self.note.id),
+            json.dumps(self.note.tags),
         )
 
         if self.addMode:
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -88,6 +88,7 @@ _html = """
     <div id="dupes" class="is-inactive">
         <a href="#" onclick="pycmd('dupes');return false;">%s</a>
     </div>
+    <div id="tagEditor"></div>
 </div>
 """
 
@@ -145,6 +146,7 @@ class Editor:
             js=[
                 "js/vendor/jquery.min.js",
                 "js/vendor/protobuf.min.js",
+                "js/tag-editor.js",
                 "js/editor.js",
             ],
             context=self,
@@ -463,7 +465,7 @@ $editorToolbar.addButtonGroup({{
                 self.web.setFocus()
             gui_hooks.editor_did_load_note(self)
 
-        js = "setFields(%s); setFonts(%s); focusField(%s); setNoteId(%s);" % (
+        js = "setFields(%s); setFonts(%s); focusField(%s); setNoteId(%s); setTagEditor(); " % (
             json.dumps(data),
             json.dumps(self.fonts()),
             json.dumps(focusTo),
--- a/ts/editor/index.ts
+++ b/ts/editor/index.ts
@@ -33,7 +33,7 @@ declare global {
     interface Global {
         anki: {
             tagEditor(target: HTMLDivElement): void;
-        }
+        };
     }
 }
 
--- a/ts/editor/index.ts
+++ b/ts/editor/index.ts
@@ -160,15 +160,9 @@ export function setFonts(fonts: [string, number, boolean][]): void {
     );
 }
 
-export function setTagEditor(): void {
-    const target = document.getElementById("tagEditor") as HTMLDivElement;
-
-    if (target.hasAttribute("is-init")) {
-    }
-    else {
-        globalThis.anki.tagEditor(target);
-        target.setAttribute("is-init", "");
-    }
+export function setTags(tags: string[]): void {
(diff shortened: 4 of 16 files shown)


dae:
I've updated repos.bzl so I could try this out. Looking good!

A couple of initial thoughts:

- I'd recommend not doing any sorting, as it can be confusing when what you just typed is reordered. The backend will take care of sorting the tags when the note is saved, so they'll display as ordered once we reload the note.
- The '::' on space is surprising to me - I expected it to behave how enter does. But maybe I'm just too used to how Anki currently behaves!

dae:
> Tag prefix search in Backend

2.1.41 switched to searching anywhere in tags, instead of requiring a prefix match - is that what you mean?

hgiesel:
> The '::' on space is surprising to me.

That was a spontaneous idea I had. I thought, now that Anki supports hierarchical tags natively, it might be worth considering an easier way of entering them. Entering double-colon can be quite straining on some keyboards, especially Asian keyboards that enter full-width symbols by default, and we wouldn't loose any functionality, because you can still use Enter. Opinions welcome :-)

> Tag prefix search in Backend

That is also related to hierarchical tags: Let's say I have the tags `english` and `examples`, but I also a lot of subtags to `english` : `english::verbs`, `english::nouns`, etc.. If you now enter `e` into the tag editor, the suggestions would be:

```
english
english::nouns
english::verbs
...
examples
```

You might not even be able to see `examples`, because there are so many subtags to `englis`. Maybe it would make sense to have a hierarchy-aware / breadth-first search that would either _only_ show `english`, and `examples`, or something like:

```
english
examples
english::nouns
english::verbs
...
examples::good
```

and making it easier to navigate the tag suggestions (e.g. using the up/down key to navigate the list, and using the right key to "accept" a suggestion, updating the tag list)

dae:
I get the logic of completing at the current level - it's nice if your goal is to specify the tag from the start. But unfortunately it's not so useful if you know the tail you want - to get "this::is::a::deeply::nested::tree", if you know you want "tree", it's much faster to type "tree" than to go through all those different layers first

dae:
Ordering by ascending depth like in your second example sounds like it might be a good compromise, though it would mean we'd either need to potentially fetch thousands of matches, or do the sorting prior to list truncation

hgiesel:
However, ordering by ascending depth, when searching for `nested`,  would yield:

```
nested
nested::more
nested::even::more
...
this::is::a::deeply::nested
```

Which is also not ideal, and would place long tags at an unjust disadvantage, just like the current algorithm puts lexicographically bigger words at a disadvantage. I guess the *ideal* sorting would be firstly ordered by the __amount of delimiters to the right side of the target search__, and possibly secondarily by length / depth, so searching for `nested` would yield:

```
// no delimiter to the right
nested
en::nested
foo::nested_extra
foo::extra_nested
this::is::very::deeply::nested
// one delimiter to the right
nested::bar1
nested::bar2
en::nested::test
// etc.
```

However I fear that is beyond the realm of possibility.

dae:
Good points; let's keep things simple for now. People seem to be pretty happy with the substring search change, and that probably gets us close enough

glutanimate:
While this looks cool, please keep in mind that the suggested changes / todo list will basically break all add-ons that interact with `TagEdit` in any way. That includes add-ons like:

- Image Occlusion
- Searching, PDF Reading & Note-Taking in Add Dialog
- Migaku Add-ons
- Search and Replace Tags
- BetterTags
- Field History
- Incremental Reading
- Quick Tagging
- Search in Add Card Dialog
- Extended tag add/edit dialog
- ...

The list goes on because quite a few add-ons use `TagEdit` as a generic entry field with auto-complete support.

The bottom-line: We're likely looking at hundreds of thousands of affected users.

IMHO, if we want to go ahead with this, let's please:

1. Keep `TagEdit` and its backend around and functioning until most add-ons have had a chance to switch to a potential Svelte-based widget. This should at least keep Image Occlusion and most of the add-ons in the list above working (the exceptions being *Searching, PDF Reading*, *Migaku add-ons*, *BetterTags*, for instance)

2. Create a Python/Qt-based wrapper with a small web-view, so that the completer may also be used for:
      - various Anki dialogs that currently use `TagEdit` outside the editor, e.g. the add tags dialog in the browser
      - add-ons that may want to embed a tag widget in their own dialogs
      - (optional): add-ons that may want to embed a generic text completer widget

3. For add-on developers that need to interact with the actual tag completer / modify its behavior (rather than just embedding it in their own views): Let's figure out which parts of the Qt API the add-ons in question most commonly depend on, and make sure that the new component offers corresponding APIs, so that add-on authors do not have to each ship their own solution. These APIs should also be exposed via the Qt wrapper. Some of the API members could look like this:
      - setting the tag/string collection to look through
      - getting/setting the current text
      - getting/setting the current suggestion
      - opening/closing the completer
      - a way to subscribe to certain events, e.g. key presses as exemplified in @fonol's current use of `TagEdit` in [in *Searching, PDF Reading & Note-Taking in Add Dialog*](https://github.com/fonol/anki-search-inside-add-card/blob/8d6cab3e68f60612344c38294fd6f801ac1e7efc/src/__init__.py#L572)

dae:
I suspect Henrik meant "remove it from the editor" rather than "remove it from Anki". I'm not sure this completer even makes sense as a standalone widget, as the completion pop-up would not be able to extend outside the webview bounds. So I suspect Anki will continue to be using tagedit.py in some capacity for for some time still. I appreciate you letting me know that it's widely used by add-ons in any case, and if we do stop using it in the future, then it can be kept around for a while to give people time to transition away from it.

glutanimate:
Ah, right, forgot about the webview bounds. Thanks for the reassurance on the transition period, really appreciate that.

My only remaining requests would then be to please:

- Keep extensibility in mind when designing the new tag widget. One of the advantages of using Qt-based widgets is that they come with a ready-made mature API, and thus are fairly malleable with a low barrier of entry.

   It might just be me, but that doesn't seem to be the case for a lot of the new Svelte-based components – or at least if they are malleable, it's not through well-defined APIs, but requires DOM manipulation and other hacky & tightly coupled approaches that are prone to breaking.

    But again, perhaps I'm wrong on this, in which case I would really appreciate if someone could contribute a proof-of-concept to the add-on development docs, illustrating best practice approaches to modifying Svelte-based views (to take the graphs page as an example: how to a add a new graph, how to modify an existing graph in place, all while being as maintainable and stable as possible)

    In any case, I'll try chiming in on this again as this PR evolves and try my best to point out areas where I think adding a hook or some other form of add-on API might be sensible.

- Speaking as a user: Please preserve the current UX as best as you can / reasonably build on top of it. Over the past few years, I've gotten used to using the tag entry in a particular way, and if that behavior were to suddenly change fundamentally, it would be a bit disrupting. What would be important to me personally would be to - best as possible - preserve the behavior introduced in this PR, for instance: https://github.com/ankitects/anki/pull/204.

    I'm sure the UX can also be improved of course, and there's no reason to assume that the current behavior is the best behavior in all respects. I just want us to be careful not change things for the sake of change.

hgiesel:
@glutanimate 
Thanks for your feedback. Especially the UX PR will prove useful. I am already trying to mimic the old tag editor as far as it makes sense.

> It might just be me, but that doesn't seem to be the case for a lot of the new Svelte-based components – or at least if they are malleable, it's not through well-defined APIs, but requires DOM manipulation and other hacky & tightly coupled approaches that are prone to breaking.

The big difference is that Svelte components cannot be monkey-patched. The extensibility has to be provided through decoupling, moving internals to arguments, providing hooks, emitting events, etc..

glutanimate:
Glad the link is helpful!

> The big difference is that Svelte components cannot be monkey-patched. The extensibility has to be provided through decoupling, moving internals to arguments, providing hooks, emitting events, etc..

Yeah, I feared that would be the case. The good thing is that it does encourage us to build proper APIs rather than rely on crutches like monkey-patching. But it does increase our workload and take some of the pace out of add-on development as you can no longer monkey-patch your way to a functioning add-on, only to then later solidify it by adding hooks.

I feel like this also turns on its head the way I think about add-on development, as I have to figure out which interfaces I will need ahead of time and build them from the ground – which can be rather tricky in the early stages of writing an add-on.

Also, what compounds the situation to me is that, with custom web components, you don't have the same luxury as with Qt, where you're basically always guaranteed to have a rich API to fall back on – even if the adjacent code doesn't have any APIs or hooks specific to add-ons. (This is one of the reasons I why think that – past a small barrier of entry for people more familiar with web dev – historically Qt has really been a boon to the add-on ecosystem).

In any case, I didn't want to derail the discussion too far from the actual PR, but I do think it's important that we talk about these things before we just continue down this new tech stack. I think it would be a shame to end up in a situation where large parts of Anki are no longer easily accessible to add-ons, and perhaps this PR or your previewer PR, being some of the first major Svelte-based contributions, could be a good trial ground for building such an add-on API.

PR #1082, opened 2021-03-20
Backend rows

RumovZ:
Prior discussion on #1033.

This can be considered a prototype. Apart from what's listed below, some general polishing of the backend code may be required (appreciating pointers anyway).

### Todos
* [x] The code is currently in a module of the search crate, but it's not essentially related to searching. Should it get its own crate? And there's probably a better name than `browser.rs`.
* [x] In a handful of instances, we could save a string allocation if we instead referenced the previously fetched data. Do you think that's worth the additional (and significant I think) complexity?
* [x] Tests are failing because of `pb.BrowserRow has no member ...`, but I don't see what we're doing differently e.g. for SearchNode where there aren't any errors.
* [x] I'm matching column names as string literals because that's how they are currently stored, but I think now that we are using them directly on the backend, they should be deserialisable as a Vec of enum variants. How should we proceed?
    1. As is.
    2. Write custom serde code for deserialising the stored list of literals.
    3. Update collections so we can (de)serialise with the default implementation.
* [x] Currently, I think we render fields until a custom filter is encountered and then return the current text. As I don't use custom filters myself, I'm not sure if that's the most useful behviour. Other options would be to skip custom filters or to return the unresolved reference.

### Related
* Are there plans to update only single rows/ids after ops in the browser?
* Is there a reason why Anki doesn't localise dates?


dae:
Thanks Rumo, it's looking good!

> The code is currently in a module of the search crate, but it's not essentially related to searching. Should it get its own crate? And there's probably a better name than browser.rs.

A different folder would typically be called a module - the crate is the entire rslib/.  At ~300 lines at the moment we could just put it in a separate file like src/browser_row.rs or similar for now - don't have a strong preference on the name.

> In a handful of instances, we could save a string allocation if we instead referenced the previously fetched data. Do you think that's worth the additional (and significant I think) complexity?

I'd opt for the simpler solution; I don't think Rust will be the bottleneck here. You can run Anki with the following to get a profile:

ANKI_PROFILE_CODE=1 ./scripts/runopt

I did so with the browser open to a bunch of columns full screen and scrolled down for ten seconds. This is what I got:

Screen Shot 2021-03-20 at 10 08 53 pm

My reading of that is we're spending the majority of our time in the Qt paint events already :-( 

> Tests are failing because of pb.BrowserRow has no member 

pylint doesn't understand the generated protobuf, so we need to exclude the class in qt/.pylintrc

> I'm matching column names as string literals because that's how they are currently stored, but I think now that we are using them directly on the backend, they should be deserialisable as a Vec of enum variants. How should we proceed?

While the thought of doing string matches on every row leaves a bad taste in my mouth, we have add-ons packing other strings into the config array that we'd break by changing to an enum - and given we don't seem to be bottlenecked, it's probably best to keep things simple for now.

> Are there plans to update only single rows/ids after ops in the browser?

Again I'd lean towards keeping things simple. It's scrolling where the redraw speed is a real issue; one redraw every couple of seconds as the user types is fairly cheap in comparison.

> Is there a reason why Anki doesn't localise dates?

Not sure if we could do that on the backend without pulling in a 10MB unicode crate - but haven't looked into it

rslib/src/notetype/render.rs
```diff
@@ -82,7 +82,7 @@ impl Collection {
         })
     }
 
-    fn render_card_inner(
+    pub fn render_card_inner(
```
dae, line 85: _inner routines are not intended to be public - I'd either make it pub(crate) or rename it

rslib/src/search/browser.rs
```diff
@@ -0,0 +1,335 @@
...
+                ((context.card.due - context.timing.days_elapsed as i32) * 86400) as i64,
+            ),
+            _ => return Ok("".into()),
+        };
+        date.date_string(context.offset)
+    })
+}
+
+fn card_ease_str(context: &RowContext) -> String {
+    match context.card.ctype {
```
dae, line 226: All of these routines have a RowContext in common. Rather than having them as free-standing functions that take that argument, it might be better to just wrap them all in an impl RowContext and have them take self instead.

RumovZ:
> A different folder would typically be called a module - the crate is the entire rslib/.

Right, I keep messing up the terminology, same with Python modules/packages.

> You can run Anki with the following to get a profile: ANKI_PROFILE_CODE=1 ./scripts/runopt

That tool is great! However, when I run `.\bazel run qt:profile` it looks for a non-existing `c:\anki\anki.prof`. I can only get it to work when I grab an `anki.prof` from somewhere deep in the build folders.

> Not sure if we could do that on the backend without pulling in a 10MB unicode crate - but haven't looked into it

A quick search seems to indicate that there aren't any usable crates for this, yet. :/
But maybe we can hardcode some format strings for the more common languages? Or we could even use a fluent string. It should be possible to comment it in such a way that the translators will know what to do.

dae:
> it looks for a non-existing c:\anki\anki.prof

I've only tested it on Mac, which uses a different running strategy. Maybe joining anki.prof to os.environ["BUILD_WORKSPACE_DIRECTORY"] in aqt/__init__.py would help?

I don't recall anyone every requesting a native date format - maybe what we have is good enough for now?

> Currently, I think we render fields until a custom filter is encountered and then return the current text. As I don't use custom filters myself, I'm not sure if that's the most useful behviour. Other options would be to skip custom filters or to return the unresolved reference.

I suspect terminating will be fine for the vast majority of users. There are plenty of other things that need doing too, so I wouldn't worry about it. If many people request the feature, we can revisit the issue then - and if they don't, we just saved ourselves some time. :-)

RumovZ:
Comparing the old with the new output raised some more questions:

* I copied the timestamp part from card/stats.rs but it doesn't work out. It uses the current local offset to calculate dates in the past which is often off, mostly due to daylight saving time. I assume you need fixed offsets for the compley scheduling code, but as for creating date strings from timestamps, can't we just use chrono's `Local.timestamp()`?
* The old columnData uses the raw content of the sort field and strips the HTML like we do for question/answer, preserving some whitespace. The new implementation uses the actual sort field stored in the note where HTML is stripped without any replacement. I think that's an improvement, just wanted to document that here.
* I have some cards which show something like `DecodeErr { position: 873, kind: UnknownEntity }` or `DecodeErr { position: 604, kind: PrematureEnd }` in the answer or question columns. As soon as I open the problematic field in the HTML Editor, the correct strings appear in the table. My assumption is that there are broken entities which Anki's HTML Editor silently repairs and that `htmlescape::decode_html()` is less forgiving than the Python function that was used before and apparently just stopped at the broken entity without an error message. If that's correct, I don't see what we can do about it, but maybe Anki could fix entities in the future either from the regular field editor or at least when checking the database.

RumovZ:
> I've only tested it on Mac, which uses a different running strategy. Maybe joining anki.prof to os.environ["BUILD_WORKSPACE_DIRECTORY"] in aqt/init.py would help?

Worked. 🙂 Can we put it on the main branch?

```diff
diff --git a/qt/aqt/__init__.py b/qt/aqt/__init__.py
index 95fd829c..ee4d46be 100644
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -411,7 +411,8 @@ PROFILE_CODE = os.environ.get("ANKI_PROFILE_CODE")
 def write_profile_results() -> None:

     profiler.disable()
-    profiler.dump_stats("anki.prof")
+    profile = os.path.join(os.environ["BUILD_WORKSPACE_DIRECTORY"], "anki.prof")
+    profiler.dump_stats(profile)
     print("profile stats written to anki.prof")
     print("use 'bazel run qt:profile' to explore")```

**Comment by dae** [comment]
> I copied the timestamp part from card/stats.rs but it doesn't work out. It uses the current local offset to calculate dates in the past which is often off, mostly due to daylight saving time. I assume you need fixed offsets for the compley scheduling code, but as for creating date strings from timestamps, can't we just use chrono's Local.timestamp()?

You're right, we should fix the card stats screen to use Local.timestamp() as well

> The old columnData uses the raw content of the sort field and strips the HTML like we do for question/answer, preserving some whitespace. The new implementation uses the actual sort field stored in the note where HTML is stripped without any replacement. I think that's an improvement, just wanted to document that here.

Unfortunately they aren't identical. htmlToTextLine() converts br and div tags to spaces first - if we just use the sort field, `hello<br>world<div>this is</div>a test` will become `helloworldthis isa test`

I don't like that we're having to strip the HTML on the fly, and htmlToTextLine() has its own issues as well - it can't handle the `<br/>` the editor is outputing because it expects there to be a space there. We could potentially change the way the sort field is produced to fix this, and users would get the updated display after a DB check - though we'd need to make sure we don't reuse the produced value when calculating a note's checksum. 

In the future it might be nice to pack all fields into the sort field with HTML stripped, as it makes for less surprising searches when the fields have HTML in them. But I'd like to think that through more, so for now I'd lean towards replicating the existing Python code (but perhaps with a more intelligent regex instead of multiple find&replaces)

> I have some cards which show something like DecodeErr { position: 873, kind: UnknownEntity } or DecodeErr { position: 604, kind: PrematureEnd } in the answer or question columns. As soon as I open the problematic field in the HTML Editor, the correct strings appear in the table. My assumption is that there are broken entities which Anki's HTML Editor silently repairs and that htmlescape::decode_html() is less forgiving than the Python function that was used before and apparently just stopped at the broken entity without an error message. If that's correct, I don't see what we can do about it, but maybe Anki could fix entities in the future either from the regular field editor or at least when checking the database.

Maybe we should just return the unchanged string in that case, instead of an error?

**Comment by RumovZ** [comment]
> Unfortunately they aren't identical. htmlToTextLine() converts br and div tags to spaces first

I was aware of that, but I saw it as a good thing. Given that it's called "Sort Field", it can be confusing if sorting by it doesn't reflect the displayed content:

grafik

I've already ported `html_to_text_line()` because it's needed for question and answer. However, that means we can basically ditch the optimisation of not loading the note fields, because Sort Field is a default column and I can't imagine many users use the browser without it.

> Maybe we should just return the unchanged string in that case, instead of an error?

Can we do that in general or is it important that we keep the current behaviour for calculating checksums and so on?

**Comment by dae** [comment]
I think the typical user, who may not even be aware that their fields contain HTML, would prefer their words are separated and not joined together. :-) But good point about us having to load the fields, that is unfortunate. :-( How about we just comment out the optimization for now (or mark it with #[dead_code]), and we can look at updating the sort field in the not too distant future.

Changing decode_entities() will break duplicate detection for notes with invalid entities. We could either use a separate version when calculating the checksum, or just rely on the user running Check Database - it probably won't be a big issue in practice.

**Comment by RumovZ** [comment]
> Changing decode_entities() will break duplicate detection for notes with invalid entities.

I don't understand how it can work with the current implementation to be honest. Wouldn't two different inputs with an invalid entity at the same position give the same output and thus be falsely marked as duplicates?

> We could either use a separate version when calculating the checksum, or just rely on the user running Check Database - it probably won't be a big issue in practice.

I'm happy with latter, though my point above was that Check Datababse doesn't fix invalid entities.

**Comment by dae** [comment]
You're right - I'd forgotten that the comparison after the checksum is also done based on the stripped content. But the fact that nobody's reported spurious duplicates seems to indicate it's not so common for identical errors to happen in practice.

I'd be worried that an attempt to fix the entities might modify fields in a way users might not want - I wonder if it can be done safely?

**Comment by RumovZ** [comment]
So to confirm, I can just change the existing `decode_entities()`, yes?

> I'd be worried that an attempt to fix the entities might modify fields in a way users might not want - I wonder if it can be done safely?

The invalid entities probably got there in the first place because HTML fields were imported that were not really HTML or at least not properly encoded. In that case, encoding the &s that are not part of a valid entity should be what the user wants.
If the user really wanted some entity (say &, but the case didn't match: `&Amp;`), no information will be lost by encoding the & and the user can easily spot and fix the broken entity that now will be displayed as `&Amp;`.
So I think it should be possible without risk.

**Comment by dae** [comment]
> So to confirm, I can just change the existing decode_entities(), yes?

Yeah, I guess that's the best option.

> So I think it should be possible without risk.

That makes sense, thanks for thinking it through. Please feel free to make such a change to the db check, or throw it on the issue list.

**Comment by RumovZ** [comment]
Apparently, `if true { ... } else { ... }` doesn't even count as dead code.

File: rslib/backend.proto [hunk 3]
```diff
@@ -276,6 +277,10 @@ service CardsService {
   rpc SetFlag(SetFlagIn) returns (OpChanges);
 }
 
+service BrowserRowsService {
+  rpc BrowserRowForCard(CardID) returns (BrowserRow);
+}
```
dae, line 282: It's a bit odd to have a service for a single method - I might merge this into the search service, and maybe rename it to browser.

rslib/src/browser_rows.rs
```diff
@@ -0,0 +1,356 @@
...
+        )?;
+        let qnodes_text = render
+            .qnodes
+            .iter()
+            .map(|node| match node {
+                RenderedNode::Text { text } => text,
+                RenderedNode::Replacement {
+                    field_name: _,
+                    current_text,
+                    filters: _,
```
dae, line 119: Note for the future: you can do this more succinctly with Replacement { current_text, .. }

dae (commented):
Looks good Rumo, thanks for all your work on this.

rslib/src/browser_rows.rs
```diff
@@ -0,0 +1,356 @@
...
+fn card_render_required(columns: &[String]) -> bool {
+    columns
+        .iter()
+        .any(|c| CARD_RENDER_COLUMNS.contains(&c.as_str()))
+}
+
+impl Collection {
+    pub fn browser_row_for_card(&mut self, id: CardID) -> Result<Row> {
+        let columns: Vec<String> = self.get_config_optional("activeCols").unwrap_or_else(|| {
+            vec![
```
dae, line 82: It's best to define defaults in a file in the config/ folder, so that the rest of the codebase only has to call a function to ensure it'll get a valid results with a valid default. For bools/strings we have dedicated enums; other types of content currently need their own getter/setter funcs

dae: + the config key should be listed in the enum so we don't need to identify it by a string - will fix this up.

dae:
@hssm I've added a hook that can be used to replace the row content - will that meet your needs?

https://github.com/ankitects/anki/commit/3199dcb54daeac8025cc4ec6b60fb9cce1308154

rslib/src/browser_rows.rs
RumovZ: Thanks, I thought there must be a neat way to write this. 🙂

rslib/src/browser_rows.rs
RumovZ: Right, I noticed that we currently define the defaults independently in the frontend and backend which will be solved when the frontend can access a proper getter.
Initially I thought, we would make the columns deserialisable later anyway.

dae: Currently only the strings and bools are exposed to the frontend - maybe we could have a different enum for arbitrary json values.

It bugs me that we're allocating strings for every row, even though it's probably lost in the noise floor of Qt's rendering :-) We could potentially get serde to deserialize a list of int|str into an enum, and just preserve unrecognized strings for the benefit of add-ons, though we might want to use a new config key if doing that so downgrading doesn't cause issues.

RumovZ: If serialise into and deserialise from the old string list, why would that be a problem to downgrading? I thought the benefit of this approach would be that it's fully backwards compatible.
However, I assumed we would have to write a lot of custom serde code for that and that updating the databse when up- or downgrading would be easier.
Now then you mention a separate key: Can we maybe use a different key with enum columns from now on, but if it's absent, look for the old key and transform the strs into enum variants? As far as I can see, neither downgrading steps nor custom deserialisation would be necessary.

dae: The advantage of serializing into a list of numbers (for known keys) is that it's cheaper to read back later, but we'd still need to serialize Column::Other(String) into a string value. Serde can handle a mix of strings and integers with "untagged", but I don't think it could handle a value like Other(String) - we'd probably need another enum, or need to parse into a number|str first, and convert that number to the enum variant. I'm starting to gravitate back towards my initial impression of "let's just leave it for now"  ^_^;

RumovZ: Gotcha. One last suggestion before I shut up about it ;) :
Maintain two config entries, one for the backend, a list of numbers deserialisable into an enum where all unrecognised names are mapped to the same number, and one for the frontend, a list of strings. We only a need a special setter that saves both entries from a list of strings.

dae: Add-ons and previous versions could update one without changing the other though, meaning the backend would need to do extra validation to make sure everything matched

RumovZ: Yeah, the config setter on front- or backend would have to check for the "activeCols" key and call the special setter to avoid that. Alternatively, the browser model could set the backend entry whenever it gets the frontend entry.

PR #1083, opened 2021-03-21
Move ticks.scss directly into Graph.svelte

hgiesel:
See https://github.com/ankitects/anki/pull/1074#issuecomment-803593800.


dae:
My thinking with making it a separate file was that it allowed us to avoid all the :global() wrappings, but I don't have strong feelings about it. 

One outstanding question - when setting "global" css, we can either have it set globally (ie, the css will affect other parts of the page even if they are at a higher level), or we ensure it only applies to children of our component by using a local selector first (which I used inconsistently on https://github.com/ankitects/anki/pull/1083/files#diff-4ec261feed5a00dbbf510f5d7d4de830c29d866dbaf521d06a85dbf2b62477cbL31). I wonder which one is preferable?

hgiesel:
> avoid all the :global() wrappings
I like to think of the `:global()` as an implicit TODO, maybe that's why I preferred it in the Svelte component :)

> I wonder which one is preferable?
If we make it _entirely_ global, it can even affect other Svelte components on the page. If we at least scope it to to children of the component, we can limit the side effects it can cause by quite a lot.

dae:
Ah, I didn't notice you'd nested all those globals under an outer .graph - so we're all set already. :-)

PR #1084, opened 2021-03-22
Svelte "hooks" - Streamline method of getting async data

hgiesel:
This PR introduces a new directory in `/ts` called `sveltelib`, which is supposed to contain helpers to streamline the svelte components in their respective directories.

Right now, it contains two functions, which both deal with fetching asynchronous data, which I assume is a thing most svelte components will have to do.

They are both inspired by React hooks, which is why I adopted the same naming scheme of  `use*`. `useAsync` and `useAsyncReactive` are both inspired by [this hook](https://usehooks.com/useAsync/).

They are both used in `graphs/WithGraphData.svelte`, which now encapsulates the "fetch data logic" which before that was in `GraphsPage.svelte`.

Minor changes:
- Move css for `.no-focus-outline:focus` directly to `Graph.svelte`.
- ~In RangeBox, I changed the reactive statements (`$:`) to functions, which are called `on:input`. I think everything that resembles events should always be behind event handlers, because otherwise they will also be called once upon component creation.~


dae:
No problem with the general approach, but a couple of issues I noticed in testing:

- The spinner only appears to show on initial load
- Toggling between decks and collection, the displayed results seem inverted

hgiesel:
The issues should be solved now.

dae:
Thanks Henrik! Still think you might want to look into a git pre-push hook ;-)

PR #1088, opened 2021-03-23
Add context action 'Update with Current Search'

RumovZ:
As suggested on the forum. I think it's a natural addition.


dae:
Thanks Rumo!

PR #1090, opened 2021-03-25
Add hooks to the filtered deck dialog

glutanimate:
Inspired by #458, adds a number of hooks that allow add-ons to conveniently extend the filtered deck dialog with their own widgets and filtered-deck-specific state.

My use case for this is to allow users to configure add-ons like Speed Focus Mode or Exam Notifier on a filtered deck basis, which is a frequent feature request I've seen.

**A few notes**:

- I wasn't  too keen on the length of the hook names, but couldn't think of a good way to make them shorter without making them less semantic
- I tried to limit the number of hooks as best possible and adapt them to the needs of the filtered deck dialog, which is why they aren't fully in sync with the deck config hooks.
- Ideally we wouldn't have both `filtered_deck_dialog_will_add_or_update_deck` and `filtered_deck_dialog_did_add_or_update_deck`, but this seemed like the only way to me to allow add-ons to both modify config options passed to the backend via `FilteredDeckForUpdate` objects, and the classic config options that still reside in the deck's dictionary (and are commonly used by add-ons to store deck-specific add-on data)

**Proof of concept**

```python
from PyQt5.QtWidgets import QLineEdit

from anki.scheduler import FilteredDeckForUpdate
from aqt import gui_hooks, mw
from aqt.filtered_deck import FilteredDeckConfigDialog

CONFIG_KEY = "my_line_edit"


def on_filtered_deck_dialog_did_load_deck(
    filtered_deck_dialog: FilteredDeckConfigDialog, filtered_deck: FilteredDeckForUpdate
):

    form = filtered_deck_dialog.form
    layout = form.verticalLayout

    my_line_edit = QLineEdit(filtered_deck_dialog)
    form.my_line_edit = my_line_edit  # type: ignore
    layout.insertWidget(layout.count() - 1, my_line_edit)

    if filtered_deck.id:
        deck_config_dict = mw.col.decks.get(filtered_deck.id)
        text = deck_config_dict.get(CONFIG_KEY, "")
    else:
        text = ""

    my_line_edit.setText(text)


def on_filtered_deck_dialog_did_add_or_update_deck(
    filtered_deck_dialog: FilteredDeckConfigDialog,
    filtered_deck: FilteredDeckForUpdate,
    deck_id: int,
):
    my_line_edit: QLineEdit = filtered_deck_dialog.form.my_line_edit  # type: ignore

    deck_config_dict = mw.col.decks.get(deck_id)
    deck_config_dict[CONFIG_KEY] = my_line_edit.text()

    mw.col.decks.save(deck_config_dict)


gui_hooks.filtered_deck_dialog_did_load_deck.append(
    on_filtered_deck_dialog_did_load_deck
)
gui_hooks.filtered_deck_dialog_did_add_or_update_deck.append(
    on_filtered_deck_dialog_did_add_or_update_deck
)

```

----

BTW, I don't want to derail this PR too much, but I assume you're planning on switching to a custom DTO for regular decks as well, right? (dropping `DeckDict`, etc.). If so, it would be really cool if that object had a field for add-on data, as using the deck config dict for add-on specific data has become a common pattern in add-on design over the years (although probably never really sanctioned or recommended). That's the approach I went with in the PoC above for instance and what I would be planning to do for Speed Focus Mode, etc.

And in that case, perhaps it would also make sense to add a custom field to `FilteredDeckForUpdate`, which would eliminate the need for the slightly awkward `filtered_deck_dialog_did_add_or_update_deck` hook.

### DIFF
--- a/qt/aqt/filtered_deck.py
+++ b/qt/aqt/filtered_deck.py
@@ -9,8 +9,7 @@ from anki.decks import DeckDict, DeckID, FilteredDeckConfig
 from anki.errors import SearchError
 from anki.lang import without_unicode_isolation
 from anki.scheduler import FilteredDeckForUpdate
-from aqt import AnkiQt, colors
-from aqt import gui_hooks
+from aqt import AnkiQt, colors, gui_hooks
 from aqt.qt import *
 from aqt.scheduling_ops import add_or_update_filtered_deck
 from aqt.theme import theme_manager
--- a/qt/aqt/filtered_deck.py
+++ b/qt/aqt/filtered_deck.py
@@ -303,7 +303,9 @@ class FilteredDeckConfigDialog(QDialog):
             return
 
         def success(out: OpChangesWithID) -> None:
-            gui_hooks.filtered_deck_dialog_did_add_or_update_deck(self, self.deck)
+            gui_hooks.filtered_deck_dialog_did_add_or_update_deck(
+                self, self.deck, out.id
+            )
             saveGeom(self, self.GEOMETRY_KEY)
             aqt.dialogs.markClosed(self.DIALOG_KEY)
             QDialog.accept(self)
--- a/qt/aqt/filtered_deck.py
+++ b/qt/aqt/filtered_deck.py
@@ -10,6 +10,7 @@ from anki.errors import SearchError
 from anki.lang import without_unicode_isolation
 from anki.scheduler import FilteredDeckForUpdate
 from aqt import AnkiQt, colors
+from aqt import gui_hooks
 from aqt.qt import *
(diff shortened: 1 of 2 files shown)


glutanimate:
Apologies about the failing tests before. I thought I still had a pre-push hook set up, but that was in my pre-bazel project.

dae:
Looks good, thanks, and I lean more towards verbosity these days anyway.

In the future I'd like to expose an API to set auxiliary data on decks and notetypes that is stored in the config table, but it'll be some time until I can get around to that, and until the sync protocol is updated to sync individual config entries, packing data into the deck objects is probably preferable.

PR #1091, opened 2021-03-25
Making html-filter its own ts package

hgiesel:
This PR moves /ts/editor/html-filter to its own package in /ts/html-filter and does a lot of refactoring. There should be no functional differences.

There are currently multiple issues which deal with pasting. Another idea would be to move some processing which is currently done by BeautifulSoup into JS.

In order to do this correctly, I would say a testing framework would be a good bet. Filtering pasted seems like a very good case for unit tests for me.

Originally I also wanted to include a very simple first unit test, but I couldn't quite get it to run. I tried off the example offered by rules_nodejs [here](https://github.com/bazelbuild/rules_nodejs/tree/stable/examples/jest/ts). What do you think about this idea?


ts/html-filter/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
+load("@npm//@bazel/typescript:index.bzl", "ts_library")
+load("//ts:prettier.bzl", "prettier_test")
+load("//ts:eslint.bzl", "eslint_test")
+
+ts_library(
+    name = "html-filter",
+    srcs = glob(["*.ts"]),
+    module_name = "html-filter",
+    tsconfig = "//qt/aqt/data/web/js:tsconfig.json",
+    visibility = ["//visibility:public"],
```
dae, line 10: ts/ should not depend on qt/, please either move or duplicate the tsconfig.json into ts/

dae (commented):
100% agree with you about having unit tests for this code, it's something I've been wanting for ages. 

The filtering code makes use of the DOM, which I gather we won't have access to when running under Node? Wouldn't we need  either a separate DOM implementation, or to run tests via a browser (slow and flaky from what I gather?)?

What was the problem you had with Jest? Was there a specific reason you picked Jest over another testing library? I have almost no experience with JS testing libraries, so do not have any preference, I am just curious.

hgiesel:
Jest uses a library called [jsdom](https://jestjs.io/docs/configuration#testenvironment-string) by default, which is a node lib that merely mimics the browser. However there is also [playwright](https://github.com/microsoft/playwright), which are headless browsers for testing, and it [seems like you can use it with jest](https://github.com/playwright-community/jest-playwright).

I would suggest we give jsdom a try for now, as `ts/html-filter` does not use that complicated Web APIs.

I picked jest, as it seems like it has the most backing atm. For example [here](https://timdeschryver.dev/blog/how-to-test-svelte-components)'s a tutorial for using Svelte with Jest.

I also tried to use `load("@npm//jest-cli:index.bzl", "jest_test")`. In the end, I could inspect the `bazel_out` directory, and see that the files `*.test.ts` were compiled to `*.test.js` and everything, but `jest.config.js` could not find those tests.

dae:
I've rebased this code on main & added a fix for Jest - thanks for your work on this Henrik!

PR #1092, opened 2021-03-25
Adding some types

Arthur-Milchior:


dae:
Well, my plan had been to run with DeckID for a while before expanding to the other IDs - you only gave me three days :-) But the changes for the other IDs don't look too invasive, and you've already done the work, so I figured it's better to merge this in now before conflicts develop. Thanks for your work on this Arthur.

I had to fix a few issues I found - please see the following changes on https://github.com/ankitects/anki/commits/main

PR #1094, opened 2021-03-25
Fix extraneous whitespaces from strip_html_for_tts

shaunren:
This PR fixes a bug in `strip_html_for_tts`.

The function `strip_html_for_tts` currently adds extraneous whitespaces for tags such as `<b>` and `<i>`, causing the generated TTS audio to be incorrect.


dae:
Thanks Shaun!

PR #1095, opened 2021-03-26
Use svg-loader to load svgs directly into ts packages

hgiesel:
Currently we load the bootstrap icons as a library over `<head>`.

That seems kinda wasteful, as we certainly won't need _all_ of the icons. It would be better if we could more easily _pick and choose_ from icons / iconssets.

Esbuild offers support for [importing files as text](https://esbuild.github.io/content-types/#text). This way we could use inline svgs instead of the icon fonts provided by bootstrap-icons. Using inline svgs also [seems to be the better way to use icons](https://www.lambdatest.com/blog/its-2019-lets-end-the-debate-on-icon-fonts-vs-svg-icons/).

I've somewhat tried, but I already have many hours of unfruitful Bazel tinkering behind me, so I want to get straight to you with this idea, and if you think it's a good one, any pointers.

### DIFF
--- a/ts/esbuild/upstream.bzl
+++ b/ts/esbuild/upstream.bzl
@@ -66,6 +66,7 @@ def _esbuild_impl(ctx):
     args.add_joined(["--platform", ctx.attr.platform], join_with = "=")
     args.add_joined(["--target", ctx.attr.target], join_with = "=")
     args.add_joined(["--log-level", "info"], join_with = "=")
+    args.add_joined(["--loader", ".svg=text"], join_with = ":")
     args.add_joined(["--metafile", metafile.path], join_with = "=")
     args.add_all(ctx.attr.define, format_each = "--define:%s")
     args.add_all(ctx.attr.external, format_each = "--external:%s")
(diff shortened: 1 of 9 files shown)


dae:
Please give that a go

hgiesel:
Thanks for your pointer. SVG inclusion works now.

dae:
Thanks Henrik!

ts/esbuild/upstream.bzl
```diff
@@ -66,6 +66,7 @@ def _esbuild_impl(ctx):
     args.add_joined(["--platform", ctx.attr.platform], join_with = "=")
     args.add_joined(["--target", ctx.attr.target], join_with = "=")
     args.add_joined(["--log-level", "info"], join_with = "=")
+    args.add_joined(["--loader", ".svg=text"], join_with = ":")
```
dae, line 69: Just a note for the future: https://github.com/ankitects/anki/commit/f1780e6e24c66335ca9cb2ceb0d730d383d1a3a3

PR #1098, opened 2021-03-27
NF: add an extra option to search card to return one by note

Arthur-Milchior:
This PR's goal is to allow add-on 797076357 to be quick again. It became impossible when search moved to rust. 

I believe it would be a nice feature to get natively, so that you can check all notes, fields, tags, without having to go through all cards. But this would require also a UI change, and the PR do not contain it (It would be far harder work and I would not do this kind of thing without discussion first. I'm okay with loosing this work, not UI work.) However, right now, this PR do not do any change visible to the user, it just add an optional boolean to some function and a protobuf input.

In my opinion, it would also be useful for AnkiDroid. Browser is limited by the size of the screen, so avoiding duplicate entry for a single note may help the UX. And since ankidroid will move to rust, we can't just add it on our java

### DIFF
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -495,6 +495,7 @@ class Collection:
         query: str,
         order: Union[bool, str, BuiltinSort.Kind.V] = False,
         reverse: bool = False,
+        one_by_note: bool = False,
     ) -> Sequence[CardId]:
         """Return card ids matching the provided search.
 
@@ -526,7 +527,10 @@ class Collection:
                 builtin=_pb.SortOrder.Builtin(kind=order, reverse=reverse)
             )
         return [
-            CardId(id) for id in self._backend.search_cards(search=query, order=mode)
+            CardId(id)
+            for id in self._backend.search_cards(
+                search=query, order=mode, one_by_note=one_by_note
+            )
         ]
 
     def find_notes(self, *terms: Union[str, SearchNode]) -> Sequence[NoteId]:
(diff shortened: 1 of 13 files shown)


dae:
Please see #1100

Arthur-Milchior:
What am I supposed to see ? That someone is already working on it and do a back-end AND UI change, so that this PR can be closed as it's redundant ?
If so, I'm really happy, I entirely agree with @RumovZ and believe this is a really great feature to have. And I'll be happy to deprecate my add-on when it's merged... 

If there is still something relevant that I can do, please let me know. Always eager to help

dae:
Yes, I think #1100 removes the need for this - though please note it means returning note ids, not card ids. Please give the code in main a try, as it's still new and could use some testing.

Arthur-Milchior:
If it changes ui and stop the user from doing any card based action then it makes perfect sens and i’m more happy with this change than with the one i suggested

RumovZ:
It does change the UI but it doesn't prevent any actions. Instead, it provides access to cards and notes regardless of the state by these rules for the notes state: The current card is the first card of the current note and the selected cards are all cards of the selected notes.
The data displayed in the table is now pretty decoupled from the card and note objections used in any actions.

Arthur-Milchior:
Thanks very much for your work. I can't wait to see it.

If I understand correctly, when you select a line on the browser, and click "suspend" it'll suspend all cards of the note, or unsuspend them all, right ? May I ask how it decides whether it is displayed as suspended/unsuspended and which action it'll do ?

RumovZ:
It'll be decided by the current card, i.e. the card with the index 0 of the current note. Of course, that's a bit opaque to the user, so actions corresponding to the other state (like suspending in notes state) have been moved to a submenu of the context menu to encourage the use of the suitable mode.

Maybe all toggle actions should be checkable, like the flag actions already are. Then the user can see beforehand what will happen.

PR #1099, opened 2021-03-27
Adding some other types

Arthur-Milchior:
It made me realize that some constants are not used anymore. Anyway, still seems useful if it's used to have proper typing. Even if it's not ideal since dictionnary of configuration can't (yet?) have detailled type annotation the way json can have


dae:
I'm sorry Arthur, I don't really see the value here. Half the constants in that file are no longer used, and newtypes don't really help us when we're only accessing these items in one location, and getting back untyped ints from json. Only the deck conf changes are really adding anything here, and they'll be obsoleted once I can update the deck conf screen to behave like filtered_deck.py.

PR #1100, opened 2021-03-27
Notes View II

RumovZ:
Pretty much the same as #1033 after the new resetting and backend rows have been integrated.
It hasn't been tested thoroughly and I have yet to implement the suggestions made by the community, but a timely merge seems advisable to reduce the inevitable merge conflicts.

I'll leave some comments, just one question in general: Do you intentionally avoid folders in qt/aqt/? I find these huge files with 1000+ lines pretty unwieldy. First and foremost browser.py, of course, with its almost 2000 lines and still over 1200 lines after this PR. If you have no objections I would make browser a folder and sidebar and table subfolders.


qt/aqt/table.py
```diff
@@ -0,0 +1,1135 @@
...
+
+
+@dataclass
+class SearchContext:
+    search: str
+    browser: aqt.browser.Browser
+    order: Union[bool, str] = True
+    # if set, provided card ids will be used instead of the regular search
+    # fixme: legacy support for card_ids?
+    ids: Optional[Sequence[Item]] = None
```
RumovZ, line 54: Should we try to make this backwards compatible in card state? I guess we could maintain card_ids as an alias for ids and ignore it note state, but we would still need a new hook to prevent old add-ons from messing up order.

qt/aqt/table.py
```diff
@@ -0,0 +1,1135 @@
...
+    def _fetch_row_from_backend(self, item: int) -> CellRow:
+        try:
+            row = CellRow(*self.col.browser_row_for_id(item))
+        except NotFoundError:
+            return CellRow.deleted(self.len_columns())
+        except Exception as e:
+            return CellRow.generic(self.len_columns(), str(e))
+
+        # fixme: hook needs state
+        gui_hooks.browser_did_fetch_row(item, row, self._state.active_columns)
```
RumovZ, line 905: The hook should probably be passed the state here?

qt/aqt/table.py
```diff
@@ -0,0 +1,1135 @@
...
+        try:
+            gui_hooks.browser_will_search(context)
+            if context.ids is None:
+                context.ids = self._state.find_items(context.search, context.order)
+            gui_hooks.browser_did_search(context)
+            self._items = context.ids
+        finally:
+            self.end_reset()
+
+    def reverse(self) -> None:
```
RumovZ, line 1026: We've been emptying the cache here before, but I don't see why that should be necessary.

qt/aqt/table.py
```diff
@@ -0,0 +1,1136 @@
...
+        self._load_active_columns()
+        self._sort_column = self.col.get_config("noteSortType")
+        self._sort_backwards = self.col.get_config_bool(
+            Config.Bool.BROWSER_NOTE_SORT_BACKWARDS
+        )
+
+    def _load_columns(self) -> None:
+        self._columns = [
+            ("note", tr.browsing_note()),
+            ("noteCards", tr.qt_accel_cards().replace("&", "")),
```
RumovZ, line 726: Should we add a string without accelerator or is this good enough?

dae:
Sorry about the merge conflicts there. Any objections to squashing these commits into a single one, then rebasing over main so we can get rid of the giant merge commit?

Will get back to you tomorrow with a proper review

dae:
(perhaps consider hiding the card/note selector button unless an env var is set for now - allows us to merge in the code and avoid conflicts without introducing the new feature before it's ready)

qt/aqt/table.py
dae: In Japanese they put the accelerators after the text (eg  `ファイル (&F)`), so it would look odd if we re-used accelerators for other things

RumovZ:
> Any objections to squashing these commits into a single one, then rebasing over main so we can get rid of the giant merge commit?

Of course not.

> (perhaps consider hiding the card/note selector button unless an env var is set for now - allows us to merge in the code and avoid conflicts without introducing the new feature before it's ready)

Might be a good idea. Though I'm fairly confident by now that the PR doesn't break anything as long as no add-ons are used. The testing has improved a lot since my last PR, worth the merge conflicts. ;)

dae (commented):
Lovely to see such neat code - thank you for this! I've squashed this into one commit and made a few follow-up commits to tweak things - please give them a look. A few more comments below as well.

qt/aqt/table.py
dae: A grep of the add-ons seems to indicate none are using the .card_ids property. I vote for just removing it entirely until demand presents itself

qt/aqt/table.py
dae: Even just an is_note or similar boolean would be nice, and card_id will need renaming (card_or_note_id might be clearer than item_id?)

qt/aqt/table.py
dae: I think this was to ensure that when a card was changed, we didn't serve up stale results based on the old cached version we had. But you're right, we probably don't need it now

pylib/anki/collection.py
```diff
@@ -495,55 +495,26 @@ def find_cards(
...
-        You must add ' asc' or ' desc' to the order, as Anki will replace asc with
-        desc and vice versa when reverse is set in the collection config, eg
-        order="c.ivl asc, c.due desc".
-
-        If order is a BuiltinSort.Kind value, sort using that builtin sort, eg
-        col.find_cards("", order=BuiltinSort.Kind.CARD_DUE)
-
-        The reverse argument only applies when a BuiltinSort.Kind is provided;
-        otherwise the collection config defines whether reverse is set or not.
+        To define a sort order, see _build_sort_mode().
```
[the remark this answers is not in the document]
RumovZ: That's tricky. On the one hand, it should be private since it returns pb stuff and is only called from inside other pylib methods. On the other hand, its documentation needs to be public.
I think I'll move the docstring back into find_cards() and just reference it in find_notes().

qt/aqt/table.py
```diff
@@ -0,0 +1,1136 @@
...
+        return self.col.db.list(
+            f"select distinct nid from cards where id in {ids2str(items)}"
+        )
+
+    def card_ids_from_note_ids(self, items: Sequence[Item]) -> List[CardId]:
+        return self.col.db.list(f"select id from cards where nid in {ids2str(items)}")
+
+    # Columns and sorting
+
+    @abstractproperty
```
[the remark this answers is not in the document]
RumovZ: This is intentional, I should have added a comment. mypy can't handle decorated properties: https://github.com/python/mypy/issues/1362
If we used the new syntax we would have to add `type: ignore` to every line where such a property is used. Alternatively, we could remove all the properties and use methods instead.
I think the deprecated decorator is the least bad option. What do you say?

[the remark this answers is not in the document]
dae: Ugh, that's unfortunate. Looks like it's still available in 3.10, so I think you're right

dae:
Sorry, I missed this:

> Do you intentionally avoid folders in qt/aqt/? I find these huge files with 1000+ lines pretty unwieldy

Me too. There are two contributing factors:

- lots of stuff being implemented as instance methods, with no way to define them across multiple files and keep typing
- not wanting to break add-ons by moving things around

For new functionality, please feel free to create files as you need them, and if there's more than ~2 files in some theme, submodules are fine too. The new *_ops.py files take mw as an arg so that they don't all have to sit in main.py, and you've prompted me to move them into a separate subfolder as well.

PR #1101, opened 2021-03-28
NF: add type ts

Arthur-Milchior:
It's obvious when you know what `forEditorField` is, but it would have helped otherwise while I was discovering this code

### DIFF
--- a/ts/editor/index.ts
+++ b/ts/editor/index.ts
@@ -111,8 +111,10 @@ export function setFields(fields: [string, string][]): void {
         .getPropertyValue("--text-fg");
 
     adjustFieldAmount(fields.length);
-    forEditorField(fields, (field, [name, fieldContent]) =>
-        field.initialize(name, color, fieldContent)
+    forEditorField(
+        fields,
+        (field: EditorField, [name, fieldContent]: [string, string]): void =>
+            field.initialize(name, color, fieldContent)
     );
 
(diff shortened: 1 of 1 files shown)


dae:
Thanks Arthur. I might add a comment on the function as well.

Arthur-Milchior:
You may not remember but the first time I contacted you was to know whether you'd accept PR adding comments to some methods. You rejected, it seems I did get some comment wrong the first time. During a year, I kept a commented branch. I stopped when you move to rust as I didn't read rust at the time, and some add-on devs found it helpful.

If you'd accept PR adding comments, I'd be happy to contribute to this. It's the ideal kind of works for time when I want to do something but can't concentrate on complex work

dae:
That's kind of you to offer, but I'm a bit hesitant. Good comments can definitely improve things, but it can be a fine line between a comment that's helpful, and one that adds negative value (for example, by being misleading, or just restating the function signature without any useful insight, or by explaining the inner workings of a function instead of just what the caller needs to know).  I am a little OCD about them, and fear I would end up spending non-trivial amounts of time reviewing your contributions and making minor corrections to them. ^_^; Maybe in the future when things are less busy?

Arthur-Milchior:
Please let me know then. And if I have rules, (such as: don't restate what the name/type state) I'll try to follow it

PR #1102, opened 2021-03-29
More browser fixes and features

RumovZ:
Feel free to merge this at any time to avoid merge conflicts and such. Until then I might keep pushing smaller commits.
Interested to hear what you think about the new GUI element. Is there a better folder to place it in?


dae:
It's a nicer use of screen real-estate, and better than what we had before :-) Two minor downsides are that it feels slightly out of place compared to the other widgets (maybe mainly the color?), and that it animates a bit clunkily, presumably because the table list is being redrawn at the same time - at least when I tested in a non-optimized build. Things like aggregate fail count for a note look neat!

RumovZ:
I think it looks better on night mode. In both modes I would have preferred a darker blue but the Anki palette is very limited. I will try to add a suitable colour and maybe a contour in day mode. I also thought about making the sidebar toolbar a switch, too, to reduce the number of different widgets.
It looks choppy to me, too, when there are a lot of cells. Not sure if anything can be done about it apart from removing the animation entirely, which is worse in my opinion.

PR #1104, opened 2021-03-30
Column enum for backend

RumovZ:
I got fed up with matching literals so I went ahead and implemented the suggestion I had made previously. It should be perfectly safe with add-ons and downgrading but allows for ergonomic and efficient matching on the backend.


rslib/src/browser_table.rs
```diff
@@ -53,13 +77,13 @@ pub struct Font {
 }
 
 trait RowContext {
-    fn get_cell_text(&mut self, column: &str) -> Result<String>;
+    fn get_cell_text(&mut self, column: &Column) -> Result<String>;
```
dae, line 80: Minor thing: as Column is a Copy type and smaller than a pointer, it's more ergonomic and faster to just pass it directly instead of passing a reference

PR #1106, opened 2021-03-30
Fix deprecated implicit conversion from float to int using __int__()

abdnh:
This DeprecationWarning has been showing up for a while after Python warnings were enabled.


dae:
Thanks as always Abdo.

There are some issues in the deck config screen as well - I'd recommend avoiding them for now, as that screen will likely be refactored in the future anyway.

PR #1108, opened 2021-03-31
Even more browser fixes and features

RumovZ:


ftl/core/browsing.ftl
```diff
@@ -8,6 +8,7 @@ browsing-answer = Answer
 browsing-any-cards-mapped-to-nothing-will = Any cards mapped to nothing will be deleted. If a note has no remaining cards, it will be lost. Are you sure you want to continue?
 browsing-any-flag = Any Flag
 browsing-average-ease = Average Ease
+browsing-average-interval = Average Interval
```
RumovZ, line 11: Those are a bit long compared to the strings in their columns. Should we use "Avg." or "⌀"?

qt/aqt/switch.py
```diff
@@ -91,7 +91,7 @@ def _current_knob_rectangle(self) -> QRectF:
         )
 
     def _paint_knob(self, painter: QPainter) -> None:
-        painter.setBrush(QBrush(theme_manager.qcolor(colors.HIGHLIGHT_BG)))
+        painter.setBrush(QBrush(theme_manager.qcolor(colors.LINK)))
```
RumovZ, line 94: Does this work for you in both modes? It's the same colour as before in night mode.
Unfortunately, outlining the rounded rectangle looks really bleary.

qt/aqt/switch.py
```diff
@@ -108,7 +108,8 @@ def mouseReleaseEvent(self, event: QMouseEvent) -> None:
             animation.setDuration(100)
             animation.setStartValue(self.start_position)
             animation.setEndValue(self.end_position)
-            animation.start()
+            # make triggered events execute first so the animation runs smoothly afterwards
+            QTimer.singleShot(50, animation.start)
```
RumovZ, line 112: Turns out there *was* something that could be done about it. 🙂

ftl/core/browsing.ftl
```diff
@@ -100,6 +101,7 @@ browsing-suspended = Suspended
 browsing-tag-duplicates = Tag Duplicates
 browsing-tag-rename-warning-empty = You can't rename a tag that has no notes.
 browsing-target-field = Target field:
+browsing-toggle-cards-notes-mode = Toggle Cards/Notes Mode
```
dae, line 104: It's fine as-is, but maybe "Toggle Showing Cards/Notes" would be slightly clearer?

qt/aqt/table.py
```diff
@@ -717,8 +717,10 @@ def _load_columns(self) -> None:
             ("note", tr.browsing_note()),
             ("noteCards", tr.editing_cards()),
             ("noteCrt", tr.browsing_created()),
+            ("noteDue", tr.statistics_due_date()),
```
dae, line 720: This is a bit tricky to follow at the moment, because we have the enum defined in rslib, the column labels defined in Python, and the mapping of names to the enum defined in rslib/backend. I would lean to moving this all into browser_table.rs, and exposing backend methods for the frontend to get the data it needs.

dae (commented):
Looks great!

qt/aqt/forms/browser.ui
```diff
@@ -597,6 +605,14 @@
     <string>qt_accel_forget</string>
    </property>
   </action>
+  <action name="action_toggle_mode">
+   <property name="text">
+    <string>browsing_toggle_cards_notes_mode</string>
+   </property>
+   <property name="shortcut">
+    <string notr="true">Ctrl+M</string>
+   </property>
```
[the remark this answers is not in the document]
dae: (the 'n' being "show Notes")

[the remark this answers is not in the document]
RumovZ: Those are already taken for Change Notetype and Go to Note. Is there any conflict with Ctrl+T?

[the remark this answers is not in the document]
dae: Ah, whoops - I was using ripgrep and didn't escape the + when searching. Ctrl+T is used by the editor unfortunately

RumovZ:
I don't know how to proceed with the palette issue. Maybe we have to fallback on a black and white design in day mode. Please let me know if you have any ideas.
Good point about the columns. The whole column logic can be moved to the backend just like the row logic. Maybe we can even get a performance gain out of this since we currently do a lot of string matching for the font and alignment roles.
I'll come back to this next week. Until then, happy holidays! 🐰

dae:
No ideas off the top of my head, let's just leave it as-is for now :-) Hope you have a lovely break!

qt/aqt/forms/browser.ui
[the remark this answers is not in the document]
dae: I just realised we're already violating this - the editor defines ctrl+m as the MathJax shortcut 😅 . Nobody has complained so far (thankfully), but it means the shortcut is taken already anyway.

PR #1110, opened 2021-04-02
Fix graph labels display in RTL layout

abdnh:
This fixes the display of some labels in stats screen for RTL layout by setting LTR direction unconditionally for these labels.

### Before:
calendar-rtl-before
future-due-rtl-before

### After:
calendar-rtl-after
future-due-rtl-after

### DIFF
--- a/ts/graphs/buttons.ts
+++ b/ts/graphs/buttons.ts
@@ -153,28 +153,30 @@ export function renderButtons(
     const xGroup = scaleBand()
         .domain(["learning", "young", "mature"])
         .range([bounds.marginLeft, bounds.width - bounds.marginRight]);
-    svg.select<SVGGElement>(".x-ticks").call((selection) =>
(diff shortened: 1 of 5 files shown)


dae:
Thanks Abdo. I seem to recall testing this back in the early days of the new graphs - maybe this was a regression at one point? Or maybe my memory is just wrong :-)

abdnh:
I found these discussions in the forums:
https://forums.ankiweb.net/t/short-strings-in-graph-axis/1306/7
https://forums.ankiweb.net/t/anki-2-1-29-beta/1220/7

It seems you were aware of this issue in the first thread, but [switched most webviews to rtl](https://github.com/ankitects/anki/commit/01ff6ab55d4c87ad0622300c3c0146ce23d60248) to solve similar issues affecting multiple screens.

Hopefully this is the last of it :)

PR #1113, opened 2021-04-06
Backend columns

RumovZ:
I'm pleased with the frontend part which should be an improvement especially for add-ons which can now simply add their columns to the dictionary. Also, users can now remove leftover add-on columns using the GUI.
But I'm not really happy with the backend part. I saw e73359510d15c5c92bc9eb806c02f87d5e0667b9 and used strum to avoid some boilerplate code but the main issues I see remain:

1. The relationship between Python and Rust columns is fragile because we're stuck with the string literals. (So I don't see any value in an additional pb enum.)
2. We have to store the lists of state columns in a separate array (can be easily overlooked).
3. There will be no warning if a column isn't matched in `get_cell_text()`.

1\. is just listed for completeness, I don't see what we could do about it. The other two issues have the common cause that the Column enum doesn't contain the information to which state(s) a column belongs. (3. could be resolved by listing all non-applicable columns instead of using a wildacard, though.)
I thought of using a nested enum like this:

```rust
enum Column {
    Custom,
    CardAndNoteColumn(CardAndNoteColumn)
    CardColumn(CardColumn)
    NoteColumn(NoteColumn)
}

enum CardAndNoteColumn {
    NoteTags,
    Notetype,
    NoteCreation,
...
```

Then we could iterate over the appropriate sub enums to generate the state columns and match them exhaustively, i.e. we would get a compiler error if a new column was added but not matched. But we wouldn't be able to serialise easily.

Do you know how to resolve this?
<hr>

Unrelated sidenote: I noticed the formatting of the regex search error is broken again.


qt/aqt/forms/browser.ui
```diff
@@ -607,10 +607,10 @@
   </action>
   <action name="action_toggle_mode">
    <property name="text">
-    <string>browsing_toggle_cards_notes_mode</string>
+    <string>browsing_toggle_showing_cards_notes</string>
    </property>
    <property name="shortcut">
-    <string notr="true">Ctrl+M</string>
+    <string notr="true">Alt+T</string>
```
RumovZ, line 613: How about this? 😅

qt/aqt/table.py
```diff
@@ -149,11 +148,8 @@ def invert_selection(self) -> None:
     def select_single_card(self, card_id: CardId) -> None:
         """Try to set the selection to the item corresponding to the given card."""
         self.clear_selection()
-        if self.is_notes_mode():
-            self._view.selectRow(0)
-        else:
-            if (row := self._model.get_card_row(card_id)) is not None:
-                self._view.selectRow(row)
+        if (row := self._model.get_card_row(card_id)) is not None:
+            self._view.selectRow(row)
```
RumovZ, line 153: Thanks for the fixes in my code. This one I reverted and fixed `get_item_from_card_id()` instead so the method does as it says it does, even though it's not actually useful with the current implementation.

dae:
Some thoughts in no particular order:

- backend/ is intended to be simple mapping from native objects to protobuf and vice versa, so logic and business rules are better placed in the main source. in backend/search/browser_tree.rs, the From impls and to_pb_columns should stay, but the rest of the code would be better placed in src/browser_table.rs IMHO
- since we're building the enums from strings at browser startup/column change, I gather there's no value in persisting the data to disk? It might be worth storing the list of columns in col.state instead, as it would save us having to fetch it from the db on each row.

I think you're right about those arrays being icky, and it would be nicer if we could avoid the need for a catchall in the enum matches. For the former issue, one way to solve it would be to add is_note_column() and is_card_column() methods to the enum. Then the entire enum could be iterated over, and that method could be used as a predicate to build a list of columns of one type. For the matches, instead of a _ => catchall, we could list out all the unhandled cases, eg:

```rust
            Column::Custom
            | Column::NoteCards
            | Column::NoteDue
            | Column::NoteEase
            | Column::NoteInterval
            | Column::NoteLapses
            | Column::NoteReps => "".to_string()
```

that way we're explicit about our intentions, and conscious thought will be required when a new column is added.

But before we head down that path, perhaps it's worth considering whether we could potentially show more of the columns in both modes? You've already got noteDue and noteIvl aggregating card values, and we could potentially take a similar approach with some of the other card properties as well. What if instead of defining a separate CardDue and NoteDue, we switched the behaviour based on the selected mode instead? For a number of columns, it could be something like

```rust
let due = if showing_cards { 
  self.cards[0].due
} else {
  self.cards.iter().map(|c| c.due).min()
}
```

or even just always aggregate, being slightly less efficient in the cards mode case:

```rust
let due = self.cards.iter().map(|c| c.due).min()
```

By storing the single card in a vec in the cards-mode case, we could potentially merge the two different context objects and just specialise individual columns as appropriate. question/answer could always use the first available card for example, and columns where aggregate functions don't make sense (such as 'deck') could do something more fancy, like show the deck name in the case of a single deck, or just a count of decks if there's more than one.

What do you think?

qt/aqt/forms/browser.ui
dae: Seems not to be taken, let's give it a try

pylib/anki/collection.py
```diff
@@ -696,6 +696,16 @@ def _pb_search_separator(self, operator: SearchJoiner) -> SearchNode.Group.Joine
     # Browser Table
     ##########################################################################
 
+    def all_browser_card_columns(
+        self,
+    ) -> Generator[Tuple[str, str, bool, bool, bool, bool], None, None]:
```
dae, line 701: The unlabeled tuple is not a very nice signature for a public API. I'm guessing the motivation for this was that the dataclass makes reference to add-ons, and you thought that didn't belong in pylib? Since this is just a "data transfer object" that doesn't need any special logic associated with it, what I'd recommend you do instead is make the protobuf object public by adding it to the top of collection.py, and then return that protobuf message instead. We could then either convert directly from that protobuf to the dataclass, or just use it as-is. We'll need to run it through a hook for add-ons to inject the correct label for a given key anyway, and at display time we could just do "column.label or tr.browsing_addon()" - so I'd lean towards dropping the dataclass unless you think there's some other reason we'd need it.

rslib/backend.proto
```diff
@@ -1052,6 +1054,19 @@ message FindAndReplaceIn {
   string field_name = 6;
 }
 
+message BrowserColumns {
+  repeated BrowserColumn columns = 1;
+}
+
+message BrowserColumn {
```
dae, line 1061: If this message is only going to be used in the BrowserColumns message, I'd embed it there and call it "Column"

rslib/backend.proto
```diff
@@ -1052,6 +1054,19 @@ message FindAndReplaceIn {
...
+  string key = 1;
+  string label = 2;
+  bool is_sortable = 3;
+  bool sorts_reversed = 4;
+  bool uses_cell_font = 5;
+  bool aligns_centered = 6;
```
dae, line 1067: "centered" maybe? Or we could make it an "alignment" enum of { START, CENTER } to allow for us possibly including end alignment in the future

rslib/backend.proto
```diff
@@ -1052,6 +1054,19 @@ message FindAndReplaceIn {
...
 
+message BrowserColumns {
+  repeated BrowserColumn columns = 1;
+}
+
+message BrowserColumn {
+  string key = 1;
+  string label = 2;
+  bool is_sortable = 3;
+  bool sorts_reversed = 4;
```
dae, line 1065: Not sure if it's a good idea or not, but one other possibility would be to merge the two into an enum of { NO, NORMAL, REVERSED }

pylib/anki/collection.py
RumovZ: I rather thought we would add a hook after the state was set up. Add-ons could then modify the column dict and the active column list directly.
My main reason to avoid the pb class was it's opaqueness. We don't have auto completion or type hints and even finding its definition is tough for people unfamiliar with the source code. It's the same approach I used for CellRow but I see that Column is a pure dataclass in contrast. Also, if we use enums instead of bools as you suggested below, it makes more sense to use the pb structures directly.

rslib/backend.proto
RumovZ: That's what I started with but I didn't like the sound of `BrowserColumns.Column`. Would you mind if I added a `BrowserTable` message (just as a namespace) and made Column, Row etc. members?

RumovZ:
> since we're building the enums from strings at browser startup/column change, I gather there's no value in persisting the data to disk? It might be worth storing the list of columns in col.state instead, as it would save us having to fetch it from the db on each row.

Are you referring to the active columns here?

> perhaps it's worth considering whether we could potentially show more of the columns in both modes?

Here are the columns that currently don't have equivalents in both states:
* Question and Answer (can be shown for the first card in notes mode)
* Deck and Template (can be shown if there is only per note in notes mode and replaced by a count otherwise in notes mode)
* Card Modified (can be shown for the last modified card in notes mode)
* Number of cards per note: Maybe we can make this the equivalent of the template column?

So making the "same" columns available in both states should be doable. (But to anticipate a potential future idea: I think it still makes a lot of sense to let the user save one set of active columns per state.)
As for the labels, should we just use the same ones and rely on the user understanding which aggregation method has been used? As for the keys, if we merge them together we will have to change the way we set the sort key which is currently using the same string literals.

rslib/backend.proto
dae: Sounds even better

pylib/anki/collection.py
dae: The symlinks in the source tree are to help editors like PyCharm/intelliJ pick up the generated types:

```
dae@dip:pylib/anki/_backend% ls -l
total 56
-rw-r--r--  1 dae  staff  2019 Apr  3 10:25 BUILD.bazel
-rw-r--r--  1 dae  staff  6819 Apr  5 11:11 __init__.py
lrwxr-xr-x  1 dae  staff    54 Feb  7 13:54 backend_pb2.pyi -> ../../../bazel-bin/pylib/anki/_backend/backend_pb2.pyi
lrwxr-xr-x  1 dae  staff    48 Apr  3 10:25 fluent.py -> ../../../bazel-bin/pylib/anki/_backend/fluent.py
-rwxr-xr-x  1 dae  staff  5382 Apr  3 10:25 genbackend.py
lrwxr-xr-x  1 dae  staff    51 Feb  7 13:54 generated.py -> ../../../bazel-bin/pylib/anki/_backend/generated.py
-rw-r--r--  1 dae  staff  2533 Apr  3 10:25 genfluent.py
-rw-r--r--  1 dae  staff   235 Apr  3 10:25 rsbridge.pyi
```

On my machine, I run ./run to ensure the build products are in bazel-bin, and then get completions/help in the editor:
Screen Shot 2021-04-08 at 9 42 16 am

The editor sometimes doesn't seem to notice updates, but I've found opening the .pyi file or generated.py causes it to rebuild the index.

The .pyi files are included in the Python wheels, so the typings should be available to end users as well, with the ability to click on a definition to jump to it.

If the symlinks are not working on Windows, you could write a little script to copy the files into the tree instead of symlinking them in perhaps?

Re the placement of hooks, I haven't been following that part closely - we'll need a hook somewhere, but it doesn't have to be here. :-)

dae:
> Are you referring to the active columns here?

Yes, sorry for not being clear.

>  I think it still makes a lot of sense to let the user save one set of active columns per state

Me too, I wasn't suggesting we change that. :-)

> As for the labels, should we just use the same ones a

Just as the cell content could change depending on mode, so too could the column labels - ease vs avg. ease, etc.

> As for the keys, if we merge them together we will have to change the way we set the sort key 

The searching code knows whether it's operating in cards or notes mode, so that seems doable? It is a bit of a shame that we're currently repeating logic here - using sql to aggregate first so we can sort, then doing the aggregation again in Rust for each cell. But I can't think of a better way to handle it, as the code needs to work for column that are not currently being sorted on as well.

pylib/anki/collection.py
RumovZ: Thanks for the instructions. Now it's only a matter of me bucking up and setting up my Python environment correctly. 😅

RumovZ:
Adding the active columns to the collection state makes a lot of sense. I wasn't really aware of that option. So we don't need serde for the Column enum.

> The searching code knows whether it's operating in cards or notes mode, so that seems doable?

Well, the search code knows if it's searching for cards or notes, which doesn't necessarily correspond to the active browser mode. I thought that was a problem because we might (outside the browser) want to search for cards but use a note ordering and wouldn't be able to specify that because the sort key would be the same. But I guess that use case shouldn't occur if each two respective sort orders do basically the same thing, only in a way that makes the most sense for the items being searched.

rslib/backend.proto
RumovZ: I've tried it, but apparently I can only use top-level messages in an rpc definiton.
`rpc AllBrowserCardColumns(Empty) returns (BrowserTable.Columns);` results in `AttributeError: module 'anki._backend.backend_pb2' has no attribute 'Columns'`. 😕

dae: Ah,  I guess that makes sense. BrowserColumns.Column works for me - it's not much different to having a browser_columns.rs that has a Column struct in it.

RumovZ:
I realised it could be nice to have cell-based alignment, e.g. in the Deck column where we could then center counts like "(2)" but left-align names. Do you think that's worth the trouble?

Also, I think we can drop the SortKind enum and use Column instead, can't we?

dae:
Perhaps it would be a bit difficult to read? Maybe if we used a translation string to get something like "(2 decks)", it would look less out of place when left aligned? No strong preference here though.

I'm not aware of any add-ons using the enum, and if they are identical, let's merge them

rslib/src/backend/search/browser_table.rs
```diff
@@ -3,114 +3,88 @@
...
+            sorting: self.sorting() as i32,
             uses_cell_font: self.uses_cell_font(),
-            aligns_centered: self.aligns_centered(),
+            alignment: self.alignment() as i32,
         }
     }
 
-    fn is_sortable(self) -> bool {
-        !matches!(self, Self::Question | Self::Answer | Self::Custom)
+    fn sorting(self) -> pb::browser_columns::Sorting {
```
dae, line 33: Still think the methods here apart from to_pb_column() would be better placed in src/browser_table.rs - otherwise looking good!

RumovZ: No objections, I only wanted to finish the logic first. 🙂

RumovZ: Apropos, should I define an internal struct then or keep using pb::BrowserColumns::Column?

dae: For simple enums, especially longer ones, I think it's fine to just reuse it. I'd use a separate Rust struct in cases like:

- one or more fields would be better represented with newtypes
- one or more fields have an enum (which gets turned into an int in protobuf)
- there are any embedded messages, since in protobuf they'll be marked optional, and that's inconvenient to deal with when the data is expected to exist

If multiple .rs files will be referencing the struct, you can "pub(crate) use pb::Something", to allow other files to import it from the file that exports it instead of from the proto file

RumovZ:
> Perhaps it would be a bit difficult to read? Maybe if we used a translation string to get something like "(2 decks)", it would look less out of place when left aligned? No strong preference here though.

The other scenario is the template column where there's always a name in cards mode and aways a count in notes mode (I've decided against using the name in the n=1 case). So here it would be enough to have mode-dependent alignment.
I'll leave it as is for now and it can be revisited in the future when someone has developed a stronger opinion.

> I'm not aware of any add-ons using the enum, and if they are identical, let's merge them

For now I've kept the pb enum. But maybe we can get rid of it, too, and have the caller of `col.find_...()` pass a `pb::BrowserColumn::Column` instead? Then we can check if `column.sorting` is valid on the frontend and if so only pass the `column.key` into the backend.

rslib/src/backend/search/browser_table.rs
RumovZ: 2 actually applies, but I don't find it that inconvenient because the pb crate lets us use the Rust enums and cast them into ints at the end.
3 also applies, but I guess it's fine because the field is `repeated`, so we have a Vec instead of an Option?

Put like that, `browser_table::{Row, Cell, Color, Font}` are all unnecessary.

dae: Yep, repeated fields are not a worry. 

This module is building up data that's only really useful to consume by frontend GUI code as far as I can tell? If other Rust methods needed to operate on the Column struct, then I'd suggest we at least keep a Rust version of that, building directly to the pb struct as you're currently doing LGTM

rslib/src/browser_table.rs
```diff
@@ -21,31 +23,54 @@ use crate::{
     timestamp::{TimestampMillis, TimestampSecs},
 };
 
-#[derive(Serialize_repr, Deserialize_repr, Debug, PartialEq, Clone, Copy)]
-#[repr(u8)]
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone, Copy, Display, EnumIter, EnumString)]
+#[serde(rename_all = "camelCase")]
+#[strum(serialize_all = "camelCase")]
 pub enum Column {
```
RumovZ, line 29: This is bugging me. We do the exact same things with serde to serialise the columns as sort keys as with strum to (de)serialise them as column keys. But I couldn't get it to work with the old string literals elsewise.
I think the only difference is that the json strings are wrapped in a pair of double quotes.

dae: Does this issue go away if we remove the ability to sort via an enum from find_cards()? That functionality is not actually used by Anki (outside of the unit tests) or any add-ons - I added it at one point because I thought it might be nice to have, but I imagine people will just continue to pass the order in as a string instead of going to the trouble of fetching all the columns first, so it's probably not adding any value.

rslib/src/search/mod.rs
```diff
@@ -192,11 +174,11 @@ impl Collection {
         if mode == &SortMode::FromConfig {
             *mode = match items {
                 SearchItems::Cards => SortMode::Builtin {
-                    kind: self.get_browser_sort_kind(),
+                    column: self.get_browser_sort_column(),
```
dae, line 177: Perhaps "SearchContext", "SearchScope", "SearchMode" or similar might be slightly better than "SearchItems"?

rslib/src/backend/search/browser_table.rs
```diff
@@ -29,65 +16,6 @@ impl browser_table::Column {
             alignment: self.alignment() as i32,
```
dae, line 16: For consistency, I think we should use cards_label, and maybe even be a bit more verbose, eg cards_mode_label, notes_mode_label

rslib/src/browser_table.rs
RumovZ: No, that's not a problem.
We use strum to
* deserialise the list of active columns received via pb
* deserialise the name of the sort column when a builtin sort order is received via pb
* serialise columns to string keys when all available browser columns are requested via pb

We use serde to
* deserialise the sort column from the config json

So dropping serde is the most realistic. I believe we could achieve that by letting the frontend get the sort column from the config (as it already does) and passing it to the backend explicitly instead of the empty `from_config` variant.

rslib/src/search/mod.rs
RumovZ: I don't mind changing the name but I have trouble with these particular suggestions. SearchMode sounds good, but `mode` is already used as a paramater name for SortMode. SearchContext sounds a bit generic given the importance of the enum to the search behaviour whereas SearchScope sounds as if only Cards *or* Notes where searched (when we actually search both but only return one type). I currently don't have a better idea, though.
By the way, the name SearchItems was inspired by the ItemId on the frontend.

dae: ReturnType? ReturnItemType? ItemType?

rslib/src/browser_table.rs
dae: Sorry, it can be hard to keep track of all the specifics :-) Passing in the desired value instead of relying on config is probably a better API in any case, but we'd need to be able to distinguish a provided string key from an arbitrary sql order clause that we also accept

RumovZ: No problem, I make these pretty lists as much for myself as for you. I certainly didn't have a clear picture of it myself before writing it down. 😅 

If we pass `reverse` as well, we can just use the Builtin overload of find_cards(). Also, I would keep the `order=True` overload because col can simply fetch the config sort itself if it's not provided.

dae:
Thanks Rumo, looks good. I pushed a quick fix for an issue I ran into; feel free to modify it if you think a different approach should be taken.

dae:
And thank you for all of your work this :-)

RumovZ:
Ah, I knew I had forgotten something! I got the same error, so thanks for the fix, looks like the best way to handle it. 🙂

PR #1117, opened 2021-04-12
Refactor browser and table into folders

RumovZ:
1. I didn't move `sidebar.py` as it would break the current path. But since the file hasn't been around for that long, maybe it's now or never?
2. Not sure if it's a good idea to export every single class, but it was hard to define what is strictly internal and what would be potentially used by add-ons.
3. Not sure about the smaller classes, either. They don't seem to belong to one specific submodule (except for StatusDelegate) but I also didn't want to create separate files for them.


dae:
Much nicer, thanks Rumo!

You could potentially move sidebar.py into the correct location, and then inject the old location into sys.modules to keep compatibility, like in anki/scheduler/__init__.py

The exports look fine - I don't think anything is using the FindDupes dialog, but we're exporting the others, so we might as well export that one as well.

PR #1118, opened 2021-04-12
Remove webview.css and prefer SASS over including files

hgiesel:
This will make it easier to work with Svelte Components in the future.

Instead of including "webview.css" or "bootstrap.min.css" as a "base CSS", the individual CSS files now `@include`/`@use` directly from `ts/sass` or `ts/node_modules` directly, as the see fit.

This means that less unnecessary CSS is loaded from webviews, and it's easier to trace back from where stuff was included. Almost all webviews now have a single source of truth, except for views including toolbar-bottom.css.

This also means that the CSS is duplicated in the individual CSS files. This change is similar to the change of Bootstrap icons. We get more flexibility with including files / SASS, but pay in the form of bigger build products.

After this PR all webviews will use [Bootstrap Reboot and Bootstrap utilities](https://github.com/twbs/bootstrap/tree/main/scss) as their "base".


qt/aqt/data/web/css/webview.scss
```diff
@@ -1,26 +0,0 @@
...
-@use 'ts/sass/core';
-@use 'ts/sass/scrollbar';
-@use 'ts/sass/buttons';
-
-// core.scss sets border-box, but we need to
-// keep the old behaviour for now to avoid breaking
-// add-ons/card templates
-* {
-  box-sizing: content-box;
-}
```
dae, line 13: The screens that show card templates like the reviewer and clayout seem to be using border-box now, which I suspect will break some card templates

dae: (aside from that, looks good :-)

hgiesel: Hm, honestly, I might have not quite thought over the consequences of changing the base CSS for reviewer.css as well.
Including Reboot into the reviewer css will have more effects than just box-sizing, like removing top margins from `<p>`, giving lists a left padding, etc..

I think most users upon this PR would just notice that "everything looks a little bit better now", but users with intricate card templates which are not bootstrap-compatible will certainly complain.

I could revert the change for reviewer.css, and set up a new PR, making this choice opt-in through the Preferences, and we give users a few versions to adapt their templates.

What do you think?

dae: Ugh, good point. Unfortunately users may be using card templates from a shared deck published years ago and no longer maintained by the deck author, so I fear we'll need to support the old styling for a long time. Can we start by unblocking this PR, either by changing reviewer.scss to not include reset, or by rolling back the changes to that file?

dae: Re a flag, the review area is largely the domain of the user rather than Anki's UI, so I wonder whether changing the default styling there is really worth it? But assuming it is, maybe we could do so with a flag on the template - that way new templates could default to the new styling, with old templates using the older styling. It would probably mean we'd need to inline the reset css into the .card div, and not sure if that would have any effect on performance or not.

dae:
This appears to have caused a few regressions:

- the gear/cog next to decks is now misaligned
- buttons no longer have rounded corners, and are unnaturally thin on macOS (border-radius of 5px and min-width of 75px seems close to what we had before)
- the congrats page looks bad

dae:
This change will likely cause issues with add-ons that touch those screens as well. I think this change is moving things in the right direction, but perhaps we might want to be more conservative with where it's applied? Newly changed areas like the congrats screen, editor and graphs are fine since they've received changes already, but screens like the decklist, overview, and top and bottom toolbars are things we'll likely want to rewrite at one point, and it may be better to hold off on making other changes to them until we're ready to do that - otherwise we end up breaking things for not much visible gain.

Also one more issue:
- the headings in the graphs have changed font, and I think the old font may have looked better - what do you think?

hgiesel:
Okay, that sounds like a good compromise, I don't wanna cause too much confusion and brekaing, I basically just wanted a safe base for working with SCSS across multiple views. I will rewrite the PR from scratch though.

Regarding the headings: At some point something starts looking interesting, just because it's new, so I kinda like the new font (it's just Helvetica btw :) ). But no strong opinion. The 

However one thing I noticed during this PR though is that if you edit a SASS file, files that `@import` / `@use` this file will not be updated by Bazel (e.g. changes in `ts/sass/_buttons.scss` will not update `qt/aqt/data/web/css/toolbar`). It seems like SASS is avoiding explicit dependencies.

dae:
I think that's because you're referencing the files directly, which is invisible to Bazel. Each partial that will be included in other files needs to be placed in a sass_library() declaration somewhere, and then the referencing files need to include that sass_library in their deps - see how I did it with core_lib for example.

dae:
It looks like the font was Helvetica previously too, but the font-weight has been adjusted from 700 to 500, and it feels too thin to me (though I am aware I may be influenced by what I'm used to :-)

PR #1119, opened 2021-04-12
Remove last jQuery from reviewer.ts / Remove fadeTime

hgiesel:
This removes the last uses of jQuery from reviewer.

Now that we don't utilizes `fadeTime` anymore, I think it's a good bet to remove it altogether. So far I've only seen it being used for setting it to zero anyway :)

### DIFF
--- a/qt/aqt/data/web/css/reviewer.scss
+++ b/qt/aqt/data/web/css/reviewer.scss
@@ -27,7 +27,6 @@ img {
     right: 10px;
     top: 0;
     font-size: 30px;
-    display: none;
     -webkit-text-stroke-width: 1px;
     -webkit-text-stroke-color: black;
 }
@@ -38,7 +37,6 @@ img {
     top: 0;
     font-size: 30px;
     color: yellow;
-    display: none;
     -webkit-text-stroke-width: 1px;
     -webkit-text-stroke-color: black;
 }
--- a/qt/aqt/data/web/js/reviewer.ts
+++ b/qt/aqt/data/web/js/reviewer.ts
@@ -6,7 +6,7 @@ declare var MathJax: any;
 type Callback = () => void | Promise<void>;
 
 var ankiPlatform = "desktop";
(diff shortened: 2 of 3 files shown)


dae:
I think we should just remove the opacity changes completely - they cause an ugly flash as the question momentarily disappears before it reappears with the answer, and MathJax renders considerably faster than it did previously. Any objections?

You've changed the signature of _updateQA(), which is going to break anything that was calling it. It's not marked as public, but that doesn't stop people :-( The whole reviewing area could do with a cleanup, but if we don't have time to be doing that right now, it's best to avoid making breaking changes.

I don't think you can just place .html() with .innerHTML, as a .html() call will also run any embedded scripts, which doesn't happen when setting innerHTML directly. This is how AnkiMobile does it:

```typescript
var setInnerHTML = function (elm, html) {
  for (const oldVideo of elm.getElementsByTagName("video") as [
    HTMLVideoElement
  ]) {
    oldVideo.pause();
    while (oldVideo.lastChild) {
      oldVideo.removeChild(oldVideo.lastChild);
    }
    oldVideo.load();
  }

  elm.innerHTML = html;

  for (const oldScript of elm.querySelectorAll("script") as [
    HTMLScriptElement
  ]) {
    const newScript = document.createElement("script");
    const oattrs = oldScript.attributes;
    for (let i = 0; i < oattrs.length; i++) {
      const attr = oattrs[i];
      newScript.setAttribute(attr.name, attr.value);
    }
    newScript.appendChild(document.createTextNode(oldScript.innerHTML));
    oldScript.parentNode.replaceChild(newScript, oldScript);
  }
};
```

It's been a while since I wrote that code, so whether it's the best way to handle it or not, I don't know. :-)

hgiesel:
I've changed the signature of _updateQA to include fadeTime again. There should be no API breaking changes now.

dae:
Thanks for that :-)

PR #1121, opened 2021-04-13
More for the browser folder

RumovZ:
Thanks for the tip with the pathname aliases. Browser should contain all appropriate files, now.


dae:
Much cleaner - thank you!

PR #1122, opened 2021-04-13
Prefer SASS over including files

hgiesel:
Another take on #1118

Additionally I moved `compile_sass` from `data/web/css` to `ts`. It is now used in editor, graphs, and congrats.


hgiesel:
Cannot quite explain why mypy is failing

> aqt/tagedit.py:102: error: Argument 1 to "isdeleted" has incompatible type "QCompleter"; expected "simplewrapper"  [arg-type]
aqt/main.py:1077: error: Argument 1 to "delete" has incompatible type "QShortcut"; expected "simplewrapper"  [arg-type]
aqt/main.py:1371: error: Argument 1 to "isdeleted" has incompatible type "QWidget"; expected "simplewrapper"  [arg-type]
aqt/webview.py:452: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
aqt/webview.py:568: error: Argument 1 to "isdeleted" has incompatible type "AnkiWebView"; expected "simplewrapper"  [arg-type]
aqt/progress.py:184: error: Argument 1 to "isdeleted" has incompatible type "Optional[ProgressDialog]"; expected "simplewrapper"  [arg-type]
aqt/sidebar.py:459: error: Argument 1 to "isdeleted" has incompatible type "SidebarTreeView"; expected "simplewrapper"  [arg-type]

It seems unrelated to the PR?

qt/aqt/data/web/css/BUILD.bazel
```diff
@@ -1,22 +1,22 @@
...
         ["*.scss"],
         exclude = ["_*.scss"],
     ),
     group = "css_local",
+    deps = [
+        "//ts/sass:core_lib",
+        "//ts/sass:buttons_lib",
+        "//ts/sass:scrollbar_lib",
+        "//ts/sass:card_counts_lib",
+    ],
```
dae, line 16: The way you have done it is fine too, but just in case you weren't aware, I think we can also bundle multiple sass files into a single library, which may be a bit more convenient (but does mean changing one file will mean rebuilding anything that uses the library)

qt/aqt/data/web/pages/BUILD.bazel
```diff
@@ -13,6 +14,7 @@ copy_files_into_group(
 copy_files_into_group(
     name = "congrats_page",
     srcs = [
+        "congrats-base.css",
```
dae, line 17: Maybe it would be simpler to import the base file in index.ts on the svelte page, so it gets bundled into congrats.css?

```patch
diff --git a/ts/congrats/index.ts b/ts/congrats/index.ts
index 547b04c37..83d934114 100644
--- a/ts/congrats/index.ts
+++ b/ts/congrats/index.ts
@@ -4,7 +4,7 @@
 import { getCongratsInfo } from "./lib";
 import { setupI18n, ModuleName } from "anki/i18n";
 import { checkNightMode } from "anki/nightmode";
-
+import "./congrats-base.css";
 import CongratsPage from "./CongratsPage.svelte";

 export async function congrats(target: HTMLDivElement): Promise<void> {
```

ts/sass/BUILD.bazel
```diff
@@ -18,8 +35,16 @@ sass_library(
...
+        "scrollbar.scss",
+    ],
+    visibility = ["//visibility:public"],
+)
+
 # qt package extracts colours from source file
 exports_files(
-    ["_vars.scss"],
-    visibility = ["//qt:__subpackages__"],
+    ["_vars.scss"] + glob(["*.scss"], exclude = ["_*.scss"]),
```
dae, line 48: Do we need this? The pages should be listing the _libs as deps instead of listing the raw .scss files

dae (commented):
Some comments below, but nothing to hold up merging. Thanks Henrik!

dae:
Oh, re the failure, I could only see a single error in the Buildkite logs, and that was just a missing annotation

qt/aqt/data/web/pages/BUILD.bazel
hgiesel: Hm, I think we have to differentiate between the component on the one hand, and the view (page) where it's used on the other (Even though the components are called CongratsPage and StatsPage :-) ). With congrats it's maybe not that dire, but for graphs it would probably affect my add-on including it in other views.

dae: I find the distinction a bit tricky to follow at the moment, but that may be my inexperience talking. :-)

hgiesel: Let's say we wanted to turn the `toolbar` and `toolbar-bottom` into Svelte components, and include them into a new view including the congrats page. Now if all bundled the base css into their respective CSS files, we would have included (most-likely) the identical CSS three times by that time, because the components pack the CSS that is intended for the page.

dae: Thanks, that makes it clearer

PR #1124, opened 2021-04-13
Remove jQuery from remaining qt/aqt files

hgiesel:
I've tested each individual change to make sure everything behaves as before.


dae:
Thanks :-)

PR #1127, opened 2021-04-14
NF: factorize browser.current_search

Arthur-Milchior:
I was looking at search code and find out that's already slightly more clear/short even if NF


dae:
Thanks Arthur!

PR #1129, opened 2021-04-14
Fix index out of range error when changing cloze model

RumovZ:
https://forums.ankiweb.net/t/list-index-out-of-range-when-updating-note-type-of-cloze-deletion-notes-in-browse/9120

Phew, found the issue. Since this code is hopefully gone soon, I didn't investigate further why it broke and why only in this special case. 🙂


dae:
Great, thanks for that - I had added those lines for type checking, but put them in the wrong spot 😅

PR #1130, opened 2021-04-15
Add AddCards.deckChooser property for legacy add-on support

glutanimate:
Reintroduces `AddCards.deckChooser` through a property, and adds a warning that discourages its use.

Not a huge fan of adding back legacy stuff, but Image Occlusion, Searching PDF Reading, and a bunch of other add-ons still depend on accessing the deck chooser via `deckChooser`.

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -3,7 +3,6 @@
 
 from typing import Callable, List, Optional
 
-import aqt.deckchooser
 import aqt.editor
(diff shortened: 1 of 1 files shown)


glutanimate:
(considered doing the same for `modelChooser`, but it didn't seem like it would help much as `NotetypeChooser` differs too much from `ModelChooser`)

dae:
Looks good, I must have missed this one by mistake.

PR #1131, opened 2021-04-15
Fix minor typo

glutanimate:
Just a super minor typo I stumbled on while browsing through things. Not sure if worth filing a PR for, but :man_shrugging:


dae:
Thanks!

PR #1132, opened 2021-04-15
Refactor decks/mod.rs

RumovZ:
I noticed a lot of comments and names describe if the name is in human or machine form. Wouldn't newtypes be the better solution? I imagine the same could be used for tags.

### DIFF
--- a/rslib/src/decks/add.rs
+++ b/rslib/src/decks/add.rs
@@ -93,7 +93,9 @@ impl Collection {
         deck.set_modified(usn);
         self.add_or_update_single_deck_with_existing_id(&mut deck, usn)
     }
+}
 
+impl Collection {
     /// Add a single, normal deck with the provided name for a child deck.
     /// Caller must have done necessarily validation on name.
     fn add_parent_deck(&mut self, machine_name: &str, usn: Usn) -> Result<()> {
--- a/rslib/src/decks/add.rs
+++ b/rslib/src/decks/add.rs
@@ -1,0 +1,163 @@
+// Copyright: Ankitects Pty Ltd and contributors
+// License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+use super::name::{immediate_parent_name, normalize_native_name};
+use crate::{error::FilteredDeckError, prelude::*};
+use std::borrow::Cow;
+
+impl Collection {
+    /// Normalize deck name and rename if not unique. Bumps mtime and usn if
+    /// name was changed, but otherwise leaves it the same.
+    pub(super) fn prepare_deck_for_update(&mut self, deck: &mut Deck, usn: Usn) -> Result<()> {
+        if let Cow::Owned(name) = normalize_native_name(&deck.name) {
+            deck.name = name;
+            deck.set_modified(usn);
+        }
+        self.ensure_deck_name_unique(deck, usn)
+    }
+
+    /// Add or update an existing deck modified by the user. May add parents,
+    /// or rename children as required. Prefer add_deck() or update_deck() to
+    /// be explicit about your intentions; this function mainly exists so we
+    /// can integrate with older Python code that behaved this way.
+    pub(crate) fn add_or_update_deck(&mut self, deck: &mut Deck) -> Result<OpOutput<()>> {
+        if deck.id.0 == 0 {
+            self.add_deck(deck)
+        } else {
+            self.update_deck(deck)
+        }
+    }
+
+    /// Add a new deck. The id must be 0, as it will be automatically assigned.
+    pub fn add_deck(&mut self, deck: &mut Deck) -> Result<OpOutput<()>> {
+        if deck.id.0 != 0 {
+            return Err(AnkiError::invalid_input("deck to add must have id 0"));
+        }
+
+        self.transact(Op::AddDeck, |col| col.add_deck_inner(deck, col.usn()?))
+    }
+
+    pub(crate) fn add_deck_inner(&mut self, deck: &mut Deck, usn: Usn) -> Result<()> {
+        self.prepare_deck_for_update(deck, usn)?;
+        deck.set_modified(usn);
+        self.match_or_create_parents(deck, usn)?;
+        self.add_deck_undoable(deck)
+    }
+
+    pub fn update_deck(&mut self, deck: &mut Deck) -> Result<OpOutput<()>> {
+        self.transact(Op::UpdateDeck, |col| {
+            let existing_deck = col.storage.get_deck(deck.id)?.ok_or(AnkiError::NotFound)?;
+            col.update_deck_inner(deck, existing_deck, col.usn()?)
+        })
+    }
+
+    pub(crate) fn update_deck_inner(
+        &mut self,
+        deck: &mut Deck,
+        original: Deck,
+        usn: Usn,
+    ) -> Result<()> {
+        self.prepare_deck_for_update(deck, usn)?;
+        deck.set_modified(usn);
+        let name_changed = original.name != deck.name;
+        if name_changed {
+            // match closest parent name
+            self.match_or_create_parents(deck, usn)?;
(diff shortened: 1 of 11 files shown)


rslib/src/decks/add.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    pub(crate) fn add_deck_inner(&mut self, deck: &mut Deck, usn: Usn) -> Result<()> {
+        self.prepare_deck_for_update(deck, usn)?;
+        deck.set_modified(usn);
+        self.match_or_create_parents(deck, usn)?;
+        self.add_deck_undoable(deck)
+    }
+
+    pub fn update_deck(&mut self, deck: &mut Deck) -> Result<OpOutput<()>> {
+        self.transact(Op::UpdateDeck, |col| {
```
dae, line 47: Recently I've been trying to have an impl Collection {} at the top of the file with any public methods, and then another one for any crate/private methods, just to keep things a bit neater (eg tags/findreplace.rs)

dae (commented):
Looks much neater, thank you! It's currently marked as a draft - did you want me to hold off on merging for now?

dae:
Newtypes for decks sound like they could be handy, would you like to give it a go?

For tags I think the only time we use the machine form is when sorting for the tree, and the conversion is done by mutating the tag object, so I'm not sure it would work there.

RumovZ:
I mostly marked it as a draft to address the type issue in this PR as well. I'll add them for the decks then.

rslib/src/decks/add.rs
RumovZ: I'll do that then. 🙂

RumovZ:
> Newtypes for decks sound like they could be handy, would you like to give it a go?

The current macro derives Copy. So I think we need a new one for a string newtype?

dae:
Might be easier to just define it manually until we have more than one instance where we want to store a non-copy type. We can skip half the implementations there - the sql code could be skipped if we pass the inner value / parse it into a temporary on load

dae:
(though if there are multiple places where we need to load from sql, at least implementing that half would probably make sense)

RumovZ:
Well, I mean NativeName and HumanName are already two instances. And I haven't tried it, yet, but I assumed it would mostly be copying the other macro and removing the non-applicable parts.

dae:
I'm just going off memory and have not dug into the code, but what about just using a newtype for the machine name? We only convert to human name when sending something to the frontend or generating strings for a user IIRC. And if it's just the one newtype, seems easier to copy&paste the parts required from the macro, as we can always add one later if the code ends up being repeated: https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

But again that's going off memory, so after digging through the code, if you think having a newtype for the human name makes sense too, please let me know. I just figured it might be a bit overkill given we don't tend to manipulate the human names.

RumovZ:
After looking into it further, I think you're right about the HumanName struct being superfluous.
Also, I realised a simple String tuple wouldn't cut it because we pass around slices of one deck name as parent deck names and that would of course not be possible with a `NativeName(String)` struct. I guess a `NativeName(Cow<str>)` would do to the trick.
It definitely turns out to be more complicated than I had anticipated. However, I'm also even more convinced of the need for this struct because there's a lot of splitting and joining based on the hierarchy char going on which would be great for DRYness and readability to have implemented as methods (like `NativeName::from_components(comps)`, `native_name.components()`, ...).
So if you also think it's still worthwhile I'll go on without it.

dae:
Sounds good

RumovZ:
Only, we probably don't want to give Deck a lifetime, either...
I think it would be best to merge this PR and I'll make another one for the name issue.

dae:
Can't you just wrap a string, and pass around a &NativeName?

dae:
(thanks!)

dae:
Ah, and re the splitting, I was more talking about having unqualified pub methods at the top, and everything else (pub(crate/super), regular fn) in a separate section

RumovZ:
> Can't you just wrap a string, and pass around a &NativeName?

But we want to create NativeNames from slices:
```Rust
pub(crate) fn immediate_parent_name(machine_name: &str) -> Option<&str> {
    machine_name.rsplitn(2, '\x1f').nth(1)
}
```
If I return a NativeName here, I either have to allocate a string from the slice or have NativeName take a reference, e.g. by wrapping a Cow.
I'm back to NativeName(String), now, but there's a lot of additional allocating. I think it could be reduced by moving the slicing logic to storage, but it doesn't feel like it belongs there.
Alternatively, maybe two structs, NativeName(String) and NativeName(&str), and the owned one implements AsRef or Deref to be passed to the storage?
Really not sure what's the best way to proceed, here.

> Ah, and re the splitting, I was more talking about having unqualified pub methods at the top, and everything else (pub(crate/super), regular fn) in a separate section

Oh, okay. What is Anki's convention for pub vs. pub(crate)? In practice, it doesn't make a dfferrence, does it?

dae:
I'd prefer to avoid the extra allocations if possible. A separate struct ala String/str would presumably work, but it feels like we may be over-engineering. Are there many cases apart from get_deck_id(immediate_parent_name(machine_name))? If not, maybe the easier option would be to just return a &str of the inner value, and have get_deck_id take the &str instead. I don't find myself making many mistakes when calling the methods in eg. storage, because the machine_name/human_name argument that the IDE shows is a useful reminder - it's when accessing deck.name directly that it's easiest to forget. And we could renaem it get_deck_id_by_machine_name() if it wasn't clear enough.

while Anki currently only goes via backend/, the Rust crate is usable directly as well, and external consumers will only have access to the public methods. The methods marked public also skip the dead code check, so will sit around even if they're not used.

RumovZ:
Yeah, that sounds right. The deck storage is almost exclusively called from the decks module where slices are generally derived from deck.name. And if that's guaranteed to be a NativeName there shouldn't be many human name slices to be used accidentally.
I think we can ensure normalisation similarly, by the way. If we normalise in NativeName::from_human_name(), we shouldn't have to check for normalisation later like we currently do in `prepare_deck_for_update()`.

> while Anki currently only goes via backend/, the Rust crate is usable directly as well, and external consumers will only have access to the public methods. The methods marked public also skip the dead code check, so will sit around even if they're not used.

Ah, all right. I'll be more conscious about using pub vs. pub(crate) in the future.

RumovZ:
I'm having trouble with the rslib tests, they aren't printed to the console (or log file) anymore:

grafik

As another minor thing, this quote from the dev doc doesn't work for me:

> If you're in a subfolder, ... will run the tests in that folder.

dae:
Seems to work for me, maybe a 'bazel shutdown' or a computer restart would help?

Screen Shot 2021-04-18 at 9 00 15 am

Another fallback option if you have rustup installed is to run 'cargo check' inside the rslib folder - the crate can be built outside Bazel.

Maybe bazel.bat is not preserving the current working directory? What does it print when you try?

RumovZ:
Compiler errors like in your screenshot do appear. What's missing is the output for failed tests. Restarting and `bazel clean --expunge` don't help (or did you mean something else with shutdown?). `cargo test` does as expected in the subfolder.

I ran `.\bazel test ...` in rslib, but now that I think about it, it's obvious why it doesn't work on Windows, where I have to use the path.  `..\bazel test ...` works fine. 🙂 (Still no test output, though.)

dae:
Ah, gotcha. Yep, I can reproduce that - likely either caused by the move to Bazel 4.0, or an upgrade to rules_rust. You could try switching .bazelversion to 3.7; if that doesn't help, it may require bisecting and be a bit of a pain to figure out.

dae:
(sadly the bulk of the issues I've run into with Bazel have been with Windows :-( )

RumovZ:
Changing the version did not help, unfortunately. I'll stick with cargo test and running individual tests from VS Code for now.

PR #1133, opened 2021-04-16
Toolbar PR Cleanup

hgiesel:


ts/editor-toolbar/ButtonGroup.svelte
```diff
@@ -7,7 +7,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import { getContext } from "svelte";
     import { nightModeKey } from "./contextKeys";
 
-    export let id: string;
+    export let id: string | undefined = undefined;
```
hgiesel, line 10: This is technically how optional arguments are supposed to be declared in Svelte components, which is a quite verbose unfortunately. I could translate this to all the downstream components, which also have several optional arguments, however I'd argue it doesn't improve readability.

dae: It's not actually optional in this case though, is it? If id is not provided, I think the html will end up with id="undefined". Or did you mean all props?

dae: Nevermind - in raw JS, node.id = undefined seems to set it to a string, but Svelte seems to just ignore the property instead.

dae: And re adding the signatures, if it's passing the typechecks, and forgetting to include a prop will not lead to subtle issues down the road, then I'm happy to leave things as is

hgiesel: I previously had it set to `export let id = ""`, which made the buttons have empty id attributes: `<button id></button>`, however, like you said, if it is undefined, Svelte will ignore it.

PR #1134, opened 2021-04-16
Add block formatting options to Editor

hgiesel:
This adds the following buttons:
- insert unordered list
- insert ordered list
- align left
- center
- align right
- justify
- indent
- outdent

Providing many more formatting options for the editor. Most of them are in a `ButtonDropdown` component, as not to overwhelm the main toolbar too much.

There are no shortcuts provided: I'd like to tackle them in another PR, setting them from JS, like discussed.

There are no shortcuts in the Fluent strings: I don't think the shortcuts really belong into the Fluent strings, as they need to adapted according to OS. Also, for decoupling, I think it would make sense to detach them from the Fluent strings. Which means I would also plead for changing the Fluent strings for e.g. `bold-text`, `bold-italic`, which include their shortcuts.

Maybe we could just add the modifier shortcuts `Ctrl`, `Alt` etc. as individual items as Fluent strings, to get the translated version?

This PR also includes some style adjustments, mostly for `ButtonDropdown`, and `DropdownMenu`, mostly so they match their respective color scheme a bit more.
Screenshot 2021-04-16 at 18 49 44

Screenshot 2021-04-16 at 18 38 50


dae:
Looks good! 

Agreed about splitting the shortcuts out; I'd just resisted doing so until now because I didn't want to break the existing translations. But if you're going to rework the rest of them at the same time, now seems like a good time. Please leave the old labels in the .ftl for now (moving them into one place and adding a "## deprecated, do not use" block comment)

Qt couldn't cope with translated shortcuts,  so they had been set to English. But with some helper functions, I imagine we could write the shortcut once, and then automatically convert that into something the JS shortcut listener can use, and something separate that can be displayed to the user.

I am a bit concerned about the behaviour when using indent outside of list mode - it adds inline styling to control the indent, and I'm not sure that's going to work correctly when switching across platforms. Maybe it would make sense to only apply an indent when in a list?

One minor issue I noticed - when you hit enter to go to the next line of a list, then click on the list button to try to return to normal adding, keyboard input gets lost and you have to click on the field again.

hgiesel:
> One minor issue I noticed - when you hit enter to go to the next line of a list, then click on the list button to try to return to normal adding, keyboard input gets lost and you have to click on the field again.

I also noticed that. As all of this is working of `execCommand`, so it's almost impossible to just adjust behavior, but rather we'd have to reimplement `execCommand` for those commands where we don't like the behavior.

dae:
In the indent case, perhaps we could switch indent/outdent to a normal icon button (I gather queryCommandState is not useful in that case anyway?) , and it could then avoid calling execCommand if a parent node is not a block. The list focus issue is an inconvenience, but I'm more concerned about the indenting.

Two other issues I noticed:

- after opening the editor, clicking on the alignment drop down, then clicking away from it to hide it, it's now at the bottom of the document, visible when you scroll down
- things like text alignment are not being copied when copying&pasting between fields, which is probably not what users would expect - perhaps we need to relax our internal paste filter to allow these properties?

hgiesel:
> things like text alignment are not being copied when copying&pasting between fields, which is probably not what users would expect - perhaps we need to relax our internal paste filter to allow these properties?

I cannot seem to recreate this on macOS.

> it adds inline styling to control the indent, and I'm not sure that's going to work correctly when switching across platforms. 

I already adressed this in #1140, but still wonder what are your worries here? The text-alignment is also handled via inline styling (`text-align`)?

dae:
Steps to reproduce:

- open the editor
- click on right align (or center align)
- type in "hello"
- cmd+a, cmd+c
- tab
- cmd+v

In the target field, the styling seems to be applied to a span instead of a div:

```html
<span style="text-align: right;">right</span>
```

> I already adressed this in #1140, but still wonder what are your worries here? The text-alignment is also handled via inline styling (text-align)?

I think it's a reasonably safe bet that the different browsers will all be able to cope with text-align. My concern was that the indenting uses a hard-coded pixel offset (`<blockquote style="margin: 0 0 0 40px`), and I'm not sure whether different browsers out there will all be able to do the right thing when asked to adjust the indent on such styling, or whether they'll all use the same level of indent.

hgiesel:
> My concern was that the indenting uses a hard-coded pixel offset.

Okay, I agree. There seems to be [text-indent](https://developer.mozilla.org/en-US/docs/Web/CSS/text-indent), which might be more suitable and worth further investigating, if we choose to pursue this further. (Even though we'd probably want the `each-line` attribute, which is not supported on any Browser currently).

dae:
Sounds like one to put on the backburner for now :-)

PR #1135, opened 2021-04-17
Add newtype NativeDeckName

RumovZ:


rslib/src/decks/add.rs
```diff
@@ -1,17 +1,12 @@
...
 impl Collection {
-    /// Normalize deck name and rename if not unique. Bumps mtime and usn if
+    /// Rename deck if not unique. Bumps mtime and usn if
     /// name was changed, but otherwise leaves it the same.
     pub(super) fn prepare_deck_for_update(&mut self, deck: &mut Deck, usn: Usn) -> Result<()> {
-        if let Cow::Owned(name) = normalize_native_name(&deck.name) {
-            deck.name = name;
-            deck.set_modified(usn);
-        }
         self.ensure_deck_name_unique(deck, usn)
```
RumovZ, line 10: I believe we can drop the check for normalisation because it should already be ensure when a name is constructed.
Should this routine also go then and we call `ensure_deck_name_unique()` directly?

rslib/src/decks/remove.rs
```diff
@@ -37,7 +37,7 @@ impl Collection {
             // if the default deck is included, just ensure it's reset to the default
             // name, as we've already removed its cards
             let mut modified_default = deck.clone();
-            modified_default.name = self.tr.deck_config_default_name().into();
+            modified_default.name = NativeDeckName(self.tr.deck_config_default_name().into());
```
RumovZ, line 40: Can we assume fluent strings are normalised?

rslib/src/storage/deck/mod.rs
```diff
@@ -23,7 +24,7 @@ fn row_to_deck(row: &Row) -> Result<Deck> {
     let id = row.get(0)?;
     Ok(Deck {
         id,
-        name: row.get(1)?,
+        name: NativeDeckName(row.get(1)?),
```
RumovZ, line 27: Similarly, are names from the database guaranteed to be normalised?

rslib/src/decks/add.rs
dae: I'm a bit wary of removing this, because it's covering not only the :: → x1f conversion, but also unicode normalization, and checks for blank components, and I'm a bit worried parts of the Rust code may accidentally cause those problems now or in the future. I'm not sure we can rely on the names coming from the DB to be valid for example - a previous Anki version or some external app may have written bad data into them, so a check on writing is probably safest.

rslib/src/decks/remove.rs
dae: Would be safest not to assume they're normalized

rslib/src/storage/deck/mod.rs
dae: Even if Anki never had bugs, we can't prevent external apps or add-ons from modifying the DB

dae:
I've pushed some tweaks to make things more explicit; what do you think?

rslib/src/decks/add.rs
RumovZ: Since we already normalise when converting from human names, I would think that normalising in the other constructors would make the most sense. Now that you've hidden the string field, NativeDeckName can guarantee that it always contains a normalised machine name.
We only need to split from_native_str() into two different constructors: `from_root()` (normalises, especially stripping \x1f) and `from_native_str()` (normalises components, preserving hierarchy).
What do you think?

dae: We read decks a lot more than we write them - displaying the deck list requires reading them all in at once for example, and some users have thousands or even tens of thousands of them. If the from_native() constructor performed normalization then, we'd be doing extra work on every read operation, instead of only doing it when writing a change to the DB, so I'm not sure that would be the best approach?

RumovZ: Ah, that makes sense. I'll add a new normalise method then.

dae:
Thanks Rumo, looks good!

PR #1138, opened 2021-04-18
Add search keyword for "first review in x days"

### LINKED ISSUE #1093: add a firstreview:x search that matches cards introduced for the first time in x days?
The information is not directly available to us in the revlog unfortunately, so it would mean looking for cards that have their first revlog entry within the provided timeframe.

Better suggestions on naming welcome too.

https://forums.ankiweb.net/t/browser-show-how-many-cards-i-suspended-today/8578/6

RumovZ:
Resolves #1093.
Testing on different collections would be welcome.

### DIFF
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -667,6 +667,16 @@ mod test {
         );
         assert_eq!(s(ctx, "added:0").0, s(ctx, "added:1").0,);
 
+        // introduced
+        assert_eq!(
+            s(ctx, "introduced:3").0,
+            format!(
+                "((select min(id) > {} from revlog where cid = c.id))",
+                (timing.next_day_at.0 - (86_400 * 3)) * 1_000
+            )
+        );
+        assert_eq!(s(ctx, "introduced:0").0, s(ctx, "introduced:1").0,);
+
         // deck
         assert_eq!(
             s(ctx, "deck:default"),
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -479,23 +479,25 @@ impl SqlWriter<'_> {
         Ok(())
     }
 
-    fn write_added(&mut self, days: u32) -> Result<()> {
+    fn cutoff_in_secs_from_days(&mut self, days: u32) -> Result<i64> {
         let timing = self.col.timing_today()?;
-        let cutoff = (timing.next_day_at.0 - (86_400 * (days as i64))) * 1_000;
+        Ok(timing.next_day_at.0 - (86_400 * (days as i64)))
+    }
+
+    fn write_added(&mut self, days: u32) -> Result<()> {
+        let cutoff = self.cutoff_in_secs_from_days(days)? * 1_000;
         write!(self.sql, "c.id > {}", cutoff).unwrap();
         Ok(())
     }
 
     fn write_edited(&mut self, days: u32) -> Result<()> {
-        let timing = self.col.timing_today()?;
-        let cutoff = timing.next_day_at.0 - (86_400 * (days as i64));
+        let cutoff = self.cutoff_in_secs_from_days(days)?;
         write!(self.sql, "n.mod > {}", cutoff).unwrap();
         Ok(())
     }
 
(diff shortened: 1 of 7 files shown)


rslib/src/search/sqlwriter.rs
```diff
@@ -478,20 +479,34 @@ impl SqlWriter<'_> {
         Ok(())
     }
 
-    fn write_added(&mut self, days: u32) -> Result<()> {
+    fn cutoff_in_secs_from_days(&mut self, days: u32) -> Result<i64> {
         let timing = self.col.timing_today()?;
-        let cutoff = (timing.next_day_at.0 - (86_400 * (days as i64))) * 1_000;
+        Ok(timing.next_day_at.0 - (86_400 * (days as i64)))
```
dae, line 484: It's not adding a lot of value, but there are some helpers on the newtype that you can use:

```patch
diff --git a/rslib/src/search/sqlwriter.rs b/rslib/src/search/sqlwriter.rs
index 749a23595..7a9d8fa10 100644
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -479,25 +479,25 @@ impl SqlWriter<'_> {
         Ok(())
     }

-    fn cutoff_in_secs_from_days(&mut self, days: u32) -> Result<i64> {
+    fn previous_day_cutoff(&mut self, days_back: u32) -> Result<TimestampSecs> {
         let timing = self.col.timing_today()?;
-        Ok(timing.next_day_at.0 - (86_400 * (days as i64)))
+        Ok(timing.next_day_at.adding_secs(-86_400 * days_back as i64))
     }

     fn write_added(&mut self, days: u32) -> Result<()> {
-        let cutoff = self.cutoff_in_secs_from_days(days)? * 1_000;
+        let cutoff = self.previous_day_cutoff(days)?.as_millis();
         write!(self.sql, "c.id > {}", cutoff).unwrap();
         Ok(())
     }

     fn write_edited(&mut self, days: u32) -> Result<()> {
-        let cutoff = self.cutoff_in_secs_from_days(days)?;
+        let cutoff = self.previous_day_cutoff(days)?;
         write!(self.sql, "n.mod > {}", cutoff).unwrap();
         Ok(())
     }

     fn write_introduced(&mut self, days: u32) -> Result<()> {
-        let cutoff = self.cutoff_in_secs_from_days(days)? * 1_000;
+        let cutoff = self.previous_day_cutoff(days)?.as_millis();
         write!(
             self.sql,
             "(select min(id) > {} from revlog where cid = c.id)",
```

rslib/src/search/sqlwriter.rs
```diff
@@ -478,20 +479,34 @@ impl SqlWriter<'_> {
...
     }
 
+    fn write_introduced(&mut self, days: u32) -> Result<()> {
+        let cutoff = self.cutoff_in_secs_from_days(days)? * 1_000;
+        write!(
+            self.sql,
+            "(select min(id) > {} from revlog where cid = c.id)",
+            cutoff
+        )
+        .unwrap();
```
dae, line 506: Adding an extra restriction seems to speed things up:

```
sqlite> select count(*) from cards c where (select min(id) > ((strftime('%s','now')-86400*30)*1000)  from revlog where cid = c.id);
count(*) = 94
Run Time: real 0.027 user 0.026584 sys 0.000219
sqlite> select count(*) from cards c where ((select min(id) > ((strftime('%s','now')-86400*30)*1000)  from revlog where cid = c.id) and c.id in (select cid from revlog where id > ((strftime('%s','now')-86400*30)*1000)));
count(*) = 94
Run Time: real 0.005 user 0.003979 sys 0.001189
```

RumovZ: That's a tremendous speedup, but I wonder why. 🤔 
Is it that SQL can cache the list of cards in the second statement while it would have to filter the whole revlog table for `cid = c.id` for every row otherwise?

dae: Yeah, I believe it means instead of having to do a separate subquery for each card in the collection, it can do one index scan of revlog to fetch candidate card ids, and then do a subquery for only those cards instead.

Thanks as always!

PR #1139, opened 2021-04-19
Fix ButtonDropdown not hiding when closed

hgiesel:
Otherwise ul.{svelte-class} in ButtonGroup has higher priority then .show from Bootstrap


dae:
Thanks Henrik!

PR #1140, opened 2021-04-20
Restructure editor-toolbar and editor + Change Enter behavior within paragraphs + Restrict indent / outdent to <li>

hgiesel:
1. I introduced a _paragraph_ button, which wraps text into &lt;p> elements. I think this is the replacement for div we should aim for, because it has better behavior:
- You cannot endlessly wrap &lt;p> elements: `<p>This<p>is<p>nested</p>deeply</p>.</p>.` will actually evaluate to: `<p>This</p><p>is</p><p>nested</p><p>deeply</p><p>.</p>`. This is the defined behavior. You can try by pasting into the field HTML editor.
- &lt;p> has a bottom-margin, which makes it easier to communicate that cloze deletions should not span multiple paragraphs, whereas &lt;div> look exactly like &lt;br> to the uninitiated.

2. I fixed the indent / outdent buttons to only work in list items. I think they could also be adjusted to work in paragraphs, but they still create a blockquote at the moment.

3. I generalized `CommandIconButton` to easier work with functionality beyond `document.execCommand`. It's is now basically an IconButton that can updates its active state depending on `onUpdate` when appropriate.  

4. That's the big one: I changed editor-toolbar, so it can be included in the build of editor. This means that all editor / editor-toolbar JS is wrapped into one file and is included in the webview. This is how I wanted to do it from the beginning, but I thought we could wait until the tag editor is done as well. However when adjusting the indent / outdent button and paragraph button I realized that I should do it now. 
- `editor-toolbar` is now basically just a visual container, whereas all the functionality is defined in `editor` and injected into the toolbar, initialized in `editor/toolbar.ts`.
- This makes sense because the buttons need logic similar to `inputHandlers` in editor, which I didn't wanna duplicate. Also, we can get rid of all the `@ts-ignore` in editor I had to introduce before.

This PR is a bit of a mess, but I left in the paragraph and indentation bit to illustrate why this restructuring is necessary / how it can be useful.


ts/editor/BUILD.bazel
```diff
@@ -25,28 +36,92 @@ ts_library(
...
     visibility = ["//visibility:public"],
     deps = [
         "base_css",
-        ":bootstrap-icons",
-        ":editor_ts",
+        "bootstrap-icons",
+        "mdi-icons",
+        "editor_ts",
+        "//ts/editor-toolbar:local_css",
+        "//ts/editor-toolbar:ButtonDropdown",
```
dae, line 111: Any reason why these are all exported individually? Perhaps they could be placed into a filegroup in editor-toolbar, and then this file could just require the single filegroup?

ts/tsconfig.json
```diff
@@ -3,12 +3,20 @@
     "compilerOptions": {
         "target": "es6",
         "module": "es6",
-        "lib": ["es2017", "es2019.array", "es2018.promise", "dom", "dom.iterable"],
+        "lib": [
+            "es2017",
+            "es2020.string",
```
dae, line 8: matchAll() is not supported on iOS 12: https://caniuse.com/mdn-javascript_builtins_string_matchall

dae (commented):
Hi Henrik,

Moving the code into the editor sounds good, and the bulk of this PR looks good 👍 

Regarding the paragraph button, am I understanding correctly that it's trying to solve the following two issues?

- the guy who wanted an indent at the start of each paragraph
- giving the user a hint that multi-line cloze deletions may not work

Are there others as well?

For the first one, we've only heard from one person so far, so my thinking was we might add an option that controls the behaviour, rather than exposing a button directly.  I know TinyMCE uses and recommends `<p>` by default, but I believe they put their `<br>` toggle behind a config option, and I don't think it's something they let you toggle in the middle of the document? The button seemed to be a bit confusing in the brief time I tried it out - half the time it didn't do what I'd expect when used on a selection, and it only produced predictable results for me when pressing it and then typed something. Perhaps with some tweaks it could be improved somewhat, but maybe putting it in an option either in the preferences screen or in a dropdown might be a better approach?

My objection there just lies with the button though, and if `<p>` is a better solution `<div>`, I'm happy to go with the former.

For the second point, I'm not sure it will help us so much, as I gather the bulk of cloze deletions are done on content that's pasted in from external sources, where we won't have control of the line separator?

will the onUpdate() change allow us to fix bold/italic/etc not updating when triggered via the shortcut keys?

One other thing I noticed, I'm getting the following each time I hit enter:

JS error /_anki/js/editor.js:6985 Uncaught RangeError: Maximum call stack size exceeded

ts/tsconfig.json
dae: (globalThis is also iOS 12.2+, but is at least supported by the later patch releases)

hgiesel:
> so my thinking was we might add an option that controls the behaviour

However for me personally I could say that both of these might come in handy, depending on the situation, and I'm for avoiding feature toggles, whenever possible. I kinda liked the idea of toggling &lt;p> via a button, as it behaves similiar to toggling the lists, and it's also able to show its active state, but that could also be reserved for an add-on.

But for now I removed the button. With the PR the Enter behavior will still be adapted, to break paragraphs, rather than inserting new &lt;br>, which could be helpful for the user in question. But other than that, I'll leave it at that. I think the paragraph button could also be part of an extended formatting add-on.

> I believe they put their <br> toggle behind a config option, and I don't think it's something they let you toggle in the middle of the document?

I think we talked about this before, when we made the switch to &lt;br>. I think Anki is special because of the templating why &lt;br> should be preferred. However &lt;p> could still make sense because of easier formatting. I think both have their use case, and it's not easy how to guide users to the correct one.

> I gather the bulk of cloze deletions are done on content that's pasted in from external sources, where we won't have control of the line separator?

I thought to adress this by adjusting html filter. I already added testing, and the next step would be to convert divs to &lt;br> (or possibly to &lt;p>).

> will the onUpdate() change allow us to fix bold/italic/etc not updating when triggered via the shortcut keys?

That is not what I had in mind. `onUpdate` is triggered by `updateActiveButtons`. `onUpdate` lets us customize how to infer whether a button should be considered active, it was necessary for implementing the paragraph button, which used `getParagraph(currentField)`, to detect whether the caret is situated within a paragraph element currently (similar to how `document.queryCommandState("unorderedList")` checks whether the caret is inside a &lt;ul> tag).

Shortcuts is what I intended to tackle after this PR.

dae:
> However for me personally I could say that both of these might come in handy, depending on the situation

Every extra button we add makes the interface slightly more intimidating, so I think it might be better to be conservative here, adding it in a less prominent area, and then seeing how much demand there is for being able to change it directly via a button.

Could I trouble you to elaborate on what you meant by "easier formatting" with P? Maybe there are other advantages that I have not understood.

> I think we talked about this before, when we made the switch to <br>. I think Anki is special because of the templating, which is why <br> should be preferred.

I wasn't suggesting we change the default `<br>` behaviour - was just pointing out that switching between BR and P modes is a config option, not a button exposed on the UI.

Thanks for making those changes!

PR #1141, opened 2021-04-20
Ctrl-Shift-C creates smallest unused cloze

k12ish:
Currently, hitting Ctrl-Shift-C produces a larger cloze in all cases, so hitting the shortcut here:
```
{{c2::some text}}
```
Produces:
```
{{c2::some text}} {{c3::}}
```
This pull request fixes that, so that the smallest possible cloze is created in all instances:
```
{{c2::some text}} {{c1::}}
```

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -711,18 +711,19 @@ $editorToolbar = document.getElementById("editorToolbar");
             else:
                 showInfo(tr.editing_to_make_a_cloze_deletion_on())
                 return
-        # find the highest existing cloze
-        highest = 0
-        for name, val in list(self.note.items()):
-            m = re.findall(r"\{\{c(\d+)::", val)
-            if m:
-                highest = max(highest, sorted([int(x) for x in m])[-1])
-        # reuse last?
+        used_nums = set()
+        for _, val in list(self.note.items()):
+            matches = re.findall(r"\{\{c(\d+)::", val)
+            matches = map(int, matches)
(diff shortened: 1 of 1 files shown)


RumovZ:
I think the current implementation is deliberate. If you don't assign numbers in a monotone fashion, you can't know the last used number to reuse it when the Alt modifier is used. Just reusing the highest existing number isn't intrisically useful.

k12ish:
> Just reusing the highest existing number isn't intrinsically useful.

You're right, but I'd argue that this is a corner case. Personally, I have "invalid" clozes only when I refactor cards. This is rarely a linear process, so the alt-shortcut isn't useful in these cases.

Ultimately, this pull request tries to produce valid notes every time whilst still keeping the fast workflow when adding cards. Perfection is not possible without keeping track of keypresses, but we can get fairly close.

RumovZ:
What do you mean by invalid clozes/notes?

dae:
I presume you meant to write c1 in your last example? 

A disadvantage of this change is that it would make it easier to accidentally reuse an existing card that had not yet been cleared up by Empty Cards. Anki doesn't care if you have gaps in your cloze numbers, and given that it's a little safer, I'm afraid I'm not sure I see what advantages this change would bring?

k12ish:
Thanks for spotting the typo!

> Anki doesn't care if you have gaps in your cloze numbers

Wow, that's a surprise to me. In which case, the advantages aren't particularly clear cut and I'd be happy to reject this PR.

PR #1142, opened 2021-04-20
Editor Preferences and Editor Data

hgiesel:
This is a WIP for having `EditorPreferences` and `EditorData` for the editor webview, similar to `GraphPreferences`, and `GraphData`.

I'm not entirely sure, whether `EditorData` makes sense, maybe it should be renamed to `NoteData`, which could at some point contain all the field data, and could also be both get/set. At the first level (for this PR), it could contain the information whether a note supports clozes.

Regarding `EditorPreferences`, I've noticed that there's already `EditingPreferences` which is used in editor.py. I've gathered, that it that is more useful for notetype/note independent editing settings however. `EditorPreferences` could contain the colors for the new ColorPicker, however I'd do this in another PR.

What do you think about the suggestions?

### DIFF
--- a/rslib/src/editor.rs
+++ b/rslib/src/editor.rs
@@ -1,0 +1,18 @@
+// Copyright: Ankitects Pty Ltd and contributors
+// License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+
+use crate::{backend_proto as pb, prelude::*};
+
+impl Collection {
+    pub(crate) fn editor_data(&mut self, _input: pb::EditorIn) -> Result<pb::EditorOut> {
+        Ok(pb::EditorOut {})
+    }
+
+    pub(crate) fn get_editor_preferences(&self) -> pb::EditorPreferences {
+        pb::EditorPreferences {}
+    }
+
+    pub(crate) fn set_editor_preferences(&mut self, _prefs: pb::EditorPreferences) -> Result<()> {
+        Ok(())
+    }
+}
--- a/rslib/src/editor.rs
+++ b/rslib/src/editor.rs
@@ -2,6 +2,7 @@
 // License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
 use crate::{backend_proto as pb, prelude::*};
+use std::convert::*;
 
 impl Collection {
     pub(crate) fn editor_data(&mut self, _input: pb::EditorIn) -> Result<pb::EditorOut> {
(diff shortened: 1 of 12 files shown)


dae (commented):
I couldn't find any reference to EditingPreferences in editor.py - did you mean Preferences.Editing in backend.proto, which is used in the preferences screen?

Pulling settings and note data over a web request is likely the architecture we'll want to move to in the long run, as it's more flexible and efficient than passing things with JSON over a web channel. I am guessing you're thinking of getting the Python code to do something like web.evalJS(f"loadNote({note_id}), which in turn triggers a request on the JS end to pull the data?

My only two concerns are:

- the possible race conditions this may introduce (eg, if the Python code loads a note and then immediately evaluates some JS to query the note data, before the web request to fetch the data has completed, what happens?). Another example is loading the browse screen in a large collection - each database request will block any others from happening until the previous one completes, so the order those requests come in determines which parts of the browse screen are shown first. This "tell JS to call us back" approach could cause the note data request to arrive after the sidebar had started loading, so it would mean the editor would hang until the sidebar had loaded. Not necessarily a deal-breaker, but worth thinking about.
- breaking add-ons that rely on the blur and key events - if changing the behaviour here, it would be good if we could still call the old hooks somehow

It might be worth starting with an experiment that loads one preference that way - if you get a large collection and run with ./run instead of ./scripts/runopt to ensure things run slowly, it would be interesting to see how that affects browser startup?

hgiesel:
> did you mean Preferences.Editing in backend.proto, which is used in the preferences screen?

Yes.

---

I added an interface for getting/setting `favoriteColors`, which I would use for an upcoming PR regarding an improved `ColorPicker`

However, trying to set the preferences will trigger this issue:

> Uncaught (in promise) Error: 500: set_editor_preferences() got an unexpected keyword argument 'input'

which is thrown from `pylib/anki/collection.py`:

```py
def set_editor_preferences(self, prefs: EditorPreferences) -> None:
    self._backend.set_editor_preferences(input=prefs)
```

I was just mimicing `set_graph_preferences`, so I feel like I missed something here?

---

If we're only going to use EditorPreferences for now, I would remove the EditorData endpoint for now. Does that sound right?

rslib/src/editor.rs
```diff
@@ -2,17 +2,30 @@
...
+        self.set_favorite_colors(colors)?;
         Ok(())
     }
 }
+
+struct FavoriteColors(Vec<u32>);
+
+impl From<FavoriteColors> for [u32; 8] {
+    fn from(v: FavoriteColors) -> Self {
+        v.0.try_into().unwrap_or_else(|_| [0xffffffff; 8])
```
dae, line 29: Perhaps it would be beter to just use a vec directly, instead of using a fixed size array?

- specifying a fixed size will make moving to older/newer versions difficult if we ever wish to change the number of available favourites
- using a vec would allow us to dynamically extend the number of favourite colours as users add them - instead of starting with 8 identical white colours, we could start with an empty vec.
- we need a vec for the protobuf anyway, and the JS end will presumably represent the data with a dynamic JS array

dae (commented):
> Uncaught (in promise) Error: 500: set_editor_preferences() got an unexpected keyword argument 'input'

adding SetEditorPreferences to SKIP_UNROLL_INPUT in genbackend.py will change the method signature to take the input message instead of its individual properties

> If we're only going to use EditorPreferences for now, I would remove the EditorData endpoint for now. Does that sound right?

If you don't need any per-note preferences at the moment, that sounds good.

dae:

PR #1145, opened 2021-04-22
solve missing escape of ' (fixes #1144)

### LINKED ISSUE #1144: Plain text import bug with single quotes: '
```
a ''
```
is incorrectly imported with an empty body. Note that there is supposed to be a tab between a and ''.
It's possible that this is actually an export bug; perhaps some escaping is needed?

rgreenblatt:
fixes #1144


dae:
Thanks Ryan!

PR #1146, opened 2021-04-22
Shortcut API for Editor

hgiesel:
This will mostly move shortcut handling into the editor with help of `lib/shortcuts`.

Shortcuts are connected to their respective toolbar buttons via `WithShortcuts.svelte`, which means that using a keyboard shortcut will act as if you had clicked the button.
This is also helpful in other ways:
- Hiding a button will unmount the shortcut (useful for clozes)
- Shortcuts for disabled buttons are automatically disabled as well

Shortcuts have to provided with their names according to `event.code`, for which [this website](https://keycode.info/) can be helpful.

I also added new fluent strings, without included shortcuts. I opted for the americanisms (i.e. color).


hgiesel:
After updating Typescript, `lib/shortcuts` could have some more typing via [Template Literal Types](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#template-literal-types).

ts/lib/shortcuts.ts
```diff
@@ -0,0 +1,149 @@
...
+// License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+import * as tr from "./i18n";
+
+export type Modifier = "Control" | "Alt" | "Shift" | "Meta";
+
+const modifiers: Modifier[] = ["Control", "Alt", "Shift", "Meta"];
+
+// how modifiers are mapped
+const platformModifiers =
+    navigator.platform === "MacIntel"
```
dae, line 11: When Qt gains native Arm support, I wonder whether this will still be MacIntel?

ts/lib/shortcuts.ts
```diff
@@ -0,0 +1,149 @@
...
+        ? ["Meta", "Alt", "Shift", "Control"]
+        : ["Control", "Alt", "Shift", "OS"];
+
+function splitKeyCombinationString(keyCombinationString: string): string[][] {
+    return keyCombinationString.split(", ").map((segment) => segment.split("+"));
+}
+
+function modifiersToPlatformString(modifiers: string[]): string {
+    const displayModifiers =
+        navigator.platform === "MacIntel"
```
dae, line 21: Maybe we can split these checks into a separate function, or move isApplePlatform() https://github.com/ankitects/anki/blob/a20606530ed181b75046c3c4d2b2bae06382c022/ts/sveltelib/shortcuts.ts#L4 into this module?

ts/lib/shortcuts.ts
```diff
@@ -0,0 +1,149 @@
...
+};
+
+function keyToPlatformString(key: string): string {
+    return key.startsWith(alphabeticPrefix)
+        ? key.slice(alphabeticPrefix.length)
+        : key.startsWith(numericPrefix)
+        ? key.slice(numericPrefix.length)
+        : Object.prototype.hasOwnProperty.call(keyToCharacterMap, key)
+        ? keyToCharacterMap[key]
+        : key;
```
dae, line 57: I'm surprised these are not being caught by https://eslint.org/docs/rules/no-nested-ternary - I think once there's more than one of them, the code would be more readable as an if/elseif, even if it's more verbose.

ts/lib/shortcuts.ts
```diff
@@ -0,0 +1,149 @@
...
+    optionalModifiers: Modifier[],
+    modifiersAndKey: string[]
+): boolean {
+    return (
+        checkKey(event, modifiersAndKey[modifiersAndKey.length - 1]) &&
+        checkModifiers(event, optionalModifiers, modifiersAndKey.slice(0, -1))
+    );
+}
+
+const shortcutTimeoutMs = 400;
```
dae, line 102: I initially thought multi-key shortcuts were not working, because I was not pressing them fast enough. The old Qt code will accept a shortcut without any timeout - is there a reason why you added a timeout here? If we do need one, maybe it needs to be a few seconds long?

dae (commented):
Thanks Henrik! Some minor nits, but nothing to hold up a merge. 

Am I understanding correctly that registerShortcut could also replace the handler in  https://github.com/ankitects/anki/blob/main/ts/deckconfig/DeckConfigPage.svelte#L13?

ts/lib/shortcuts.ts
hgiesel: Hm, Qt is the only platform that I know that allows multi key shortcuts, so I don't really have anything to compare to. But making it a few seconds should be fine.

ts/lib/shortcuts.ts
hgiesel: Honestly, I wasn't aware of this file. May I ask you why you put it into sveltelib? When I created it, I had in mind functionality that depended on Svelte in some way (e.g. Svelte stores). For that reason I'd be for moving `isApplePlatform`. Just wonder what your reasoning was for putting them under sveltelib.

ts/lib/shortcuts.ts
hgiesel: My main reason why I prefer ternary expressions, is that you can be sure that you don't miss a return statement, or a pathway by accident. But I'm fine with turning into an if statement. Too bad we don't have syntax like in Rust here :(

ts/lib/shortcuts.ts
hgiesel: AFAIK a lot of 64-bit Windows still have `Win32` as their platform, so I don't think it's unlikely. :)

ts/lib/shortcuts.ts
dae: I only added it yesterday, so not surprised you missed it :-) You're right that it probably belongs in lib/ instead; my thinking at the time was just that it's more GUI related than logic-related, but lib/ already contains other GUI stuff.

ts/lib/shortcuts.ts
dae: I do really miss implicit returns, and wasted half an hour the other day tracking down a bug that was caused by me not returning a value in a .map() call, that TypeScript didn't warn about because I had not specified the return type. :-( But in this case it would presumably be caught by the return type on the function, and be considerably more readable :-)

ts/lib/shortcuts.ts
dae: Emacs is another example - it doesn't have a timeout either FWIW :-)

hgiesel:
> Am I understanding correctly that registerShortcut could also replace the handler in https://github.com/ankitects/anki/blob/main/ts/deckconfig/DeckConfigPage.svelte#L13?

Yes. That should be `registerShortcut(() => state.save(false), "Enter")`

dae:
That's nicer - thanks!

PR #1148, opened 2021-04-22
Update Typescript to 4.2.4

hgiesel:


ts/editor-toolbar/index.ts
```diff
@@ -186,7 +186,7 @@ export class EditorToolbar extends HTMLElement {
 customElements.define("anki-editor-toolbar", EditorToolbar);
 
 /* Exports for editor */
-// @ts-expect-error
+// @ts-expect-error insufficient typing of svelte modules
```
dae, line 189: Ah, that's nicer

hgiesel: It's actually required now ^^;

dae: Ah :-)

PR #1149, opened 2021-04-22
Rename anki/ to lib/ for export

hgiesel:
`import _ from "anki/x";`

will become

`import _ from "lib/x";`

to fit the directory name.


dae:
I'll sort out the conflicts

PR #1150, opened 2021-04-23
Editor Toolbar cleanup

hgiesel:
Originally I wanted to add the note type chooser in this PR. This is what it would have looked like:
Screenshot 2021-04-24 at 00 10 01

But at the last stage I realized that I couldn't remove the Qt chooser buttons without disabling the modals ^^;.  1f05ebb shows how it would have been defined.

Instead, the PR contains all the preparation work for making this change, and some refactors:
- Instead of the raw Svelte classes, I know export the dynamic component functions `rawButton`, `iconButton`, etc. This way add-ons (and our Python code) and our TS can use the same interface.
- I've generalized the methods in editor-toolbar/index.ts, so they operate recursively. This allows for easier typing via `IterableToolbarItem`. Now we can also make `menusPromise` and `buttonsPromise` private.

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -68,21 +68,31 @@ class AddCards(QDialog):
 
             editor.web.eval(
                 f"""
-const notetypeChooser = editorToolbar.labelButton({{
-    label: `Choose note type`,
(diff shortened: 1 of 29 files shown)


dae:
Looks good! Thanks for fixing that regression I introduced too.

PR #1151, opened 2021-04-25
Fix some problems with browser actions while selection is empty

RumovZ:
Curiously, the problem we had with `@ensure_editor_saved` did not occur. I would have thought that adding another decorator on top would cause the issue with the additional argument for triggers again?

### DIFF
--- a/ftl/core/browsing.ftl
+++ b/ftl/core/browsing.ftl
@@ -65,6 +65,7 @@ browsing-nd-names = { $num }: { $name }
 browsing-new = (new)
 browsing-new-note-type = New note type:
 browsing-no-flag = No Flag
+browsing-no-selection = No cards or notes selected.
 browsing-note = Note
 # Exactly one character representing 'Notes'; should differ from browsing-card-initial.
 browsing-note-initial = N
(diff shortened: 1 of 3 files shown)


dae:
Not sure exactly how Qt is inspecting the signature - glad it didn't break things in any case :-)

abdnh:
Well, this seems to have broken delete_selected_notes at least :)
```
Caught exception:
  File "\\?\D:\bazel\anki\d63ysmdv\execroot\net_ankiweb_anki\bazel-out\x64_windows-fastbuild\bin\qt\runanki.exe.runfiles\net_ankiweb_anki\qt\aqt\utils.py", line 1018, in decorated
    func(self, *args, **kwargs)
TypeError: delete_selected_notes() takes 1 positional argument but 2 were given
```

RumovZ:
I had tested `_on_export_notes()` and thought I could interpolate for other methods without another decorator. Now I see that unlike `delete_selected_notes()`, `_on_export_notes()` isn't connected directly, but via a lambda. I guess that's progress. 🙃 

Thanks for the report at any rate!

PR #1153, opened 2021-04-25
Make EditorToolbar more similar to DeckOptions

hgiesel:
This ditches the web component approach in `editor-toolbar/index.ts`, and instead exports its API over `EditorToolbar.svelte`, just like DeckOptions. This also allows us to ditch both the Promise and Store around `buttons` and `menus`, and even the hack with `globalThis`, which I think is great :)

However I think exporting the wrappers from `dynamicComponents` is still better than exporting them directly, as it allows us more control going forward.
If we wanted to change an implementation detail, or remove a component entirely, we can basically shim them or mark them as deprecated (log a warning e.g.) individually.


dae:
Thanks Henrik, will follow up on the wrappers on the commit page.

PR #1154, opened 2021-04-25
Fix skip_if_selection_is_empty() for triggers

RumovZ:
Fix @abdnh's report in #1151.


dae:
Thank you both!

dae:
Merged this for now, but now that we've got multiple decorators, maybe it would make more sense to have a separate decorator like @no_arg_trigger that we put first? Presumably then the other triggers could then just be used as is, without a special _trigger version?

PR #1156, opened 2021-04-26
Add separate no_arg_trigger decorator

RumovZ:
Actually, we could decorate with `@pyqtSlot()` directly. But I guess it's better to be explicit and have a place for documentation.


dae:
That would work too, but being explicit is a bit nicer :-) Thanks Rumo!

PR #1159, opened 2021-04-28
WIP: Prefer idiomatic Svelte features over dynamic components

hgiesel:
With this PR, the editor toolbar buttons would be defined within `.svelte` files in ts/editor.

---

I am not able to get rid of warnings in the `.svelte` files in ts/editor.

> [ts 2607] [E] Element does not support attributes because type definitions are missing for this Svelte Component or element cannot be used as such.
Underlying error: JSX element class does not support attributes because it does not have a  '$prop_def' property.

I must have missed something in `BUILD.bazel`, but I don't know what. It still compiles, so it doesn't block me, but it shows me errors all over ^^;


ts/editor/ColorButtons.svelte
```diff
@@ -0,0 +1,50 @@
...
+        color = (currentTarget as HTMLInputElement).value;
+    }
+</script>
+
+<ButtonGroup id="color">
+    <WithShortcut shortcut="F7" let:createShortcut let:shortcutLabel>
+        <IconButton
+            class="forecolor"
+            tooltip={`${tr.editingSetForegroundColor} (${shortcutLabel})`}
+            on:click={wrapWithForecolor}
```
dae, line 38: I like how this makes the wrapper's functionality more explicit, and we could potentially make the tooltip less cumbersome by returning a function that appends the label instead, eg:

```html
    <WithShortcut shortcut="F7" let:createShortcut let:withShortcutLabel>
        <IconButton
            tooltip={withShortcutLabel(tr.editingSetForegroundColor()}
```

That said, would it make sense to teach each button component to take a shortcut prop directly instead? That allows us to push more of the complexity into the components, and simplify the code that consumes them. Eg something like:

```diff
diff --git a/ts/components/IconButton.svelte b/ts/components/IconButton.svelte
index 56269538e..ab417beb6 100644
--- a/ts/components/IconButton.svelte
+++ b/ts/components/IconButton.svelte
@@ -4,19 +4,19 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 -->
 <script lang="typescript">
     import type { Readable } from "svelte/store";
-    import { hasContext, getContext, onMount, createEventDispatcher } from "svelte";
+    import { getContext, onMount } from "svelte";
     import { disabledKey, nightModeKey, dropdownKey } from "./contextKeys";
+    import { useShortcut } from "lib/shortcuts";
 
     export let id: string;
     let className = "";
     export { className as class };
 
     export let tooltip: string | undefined;
+    export let shortcut = "";
     export let active = false;
     export let disables = true;
 
-    let buttonRef: HTMLButtonElement;
-
     const disabled = getContext<Readable<boolean>>(disabledKey);
     $: _disabled = disables && $disabled;
 
@@ -25,8 +25,14 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     const dropdown = getContext(dropdownKey);
     const dropdownProps = dropdown?.getDropdownTriggerProps() ?? {};
 
-    const dispatch = createEventDispatcher();
-    onMount(() => dispatch("mount", { button: buttonRef }));
+    let buttonRef: HTMLButtonElement;
+    let withShortcutSuffix: (string) => string;
+    $: tooltipWithShortcut = withShortcutSuffix?.(tooltip);
+    onMount(() => {
+        const handle = useShortcut(buttonRef, shortcut);
+        withShortcutSuffix = handle.withShortcutSuffix;
+        return handle.onDestroy;
+    });
 </script>
 
 <style lang="scss">
@@ -69,7 +75,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     class:btn-day={!nightMode}
     class:btn-night={nightMode}
     tabindex="-1"
-    title={tooltip}
+    title={tooltipWithShortcut}
     disabled={_disabled}
     class:dropdown-toggle={Boolean(dropdown)}
     {...dropdownProps}
diff --git a/ts/editor/ColorButtons.svelte b/ts/editor/ColorButtons.svelte
index db3c2afd0..cbe2dc0a6 100644
--- a/ts/editor/ColorButtons.svelte
+++ b/ts/editor/ColorButtons.svelte
@@ -31,15 +31,13 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 </script>
 
 <ButtonGroup id="color">
-    <WithShortcut shortcut="F7" let:createShortcut let:shortcutLabel>
-        <IconButton
-            class="forecolor"
-            tooltip={`${tr.editingSetForegroundColor} (${shortcutLabel})`}
-            on:click={wrapWithForecolor}
-            on:mount={createShortcut}>
-            {@html squareFillIcon}
-        </IconButton>
-    </WithShortcut>
+    <IconButton
+        class="forecolor"
+        tooltip={tr.editingSetForegroundColor()}
+        shortcut="F7"
+        on:click={wrapWithForecolor}>
+        {@html squareFillIcon}
+    </IconButton>
 
     <WithShortcut shortcut="F8" let:createShortcut let:shortcutLabel>
         <ColorPicker
diff --git a/ts/lib/shortcuts.ts b/ts/lib/shortcuts.ts
index 00ee2df7a..9e80b8fc3 100644
--- a/ts/lib/shortcuts.ts
+++ b/ts/lib/shortcuts.ts
@@ -154,3 +154,42 @@ export function registerShortcut(
     document.addEventListener("keydown", handler);
     return (): void => document.removeEventListener("keydown", handler);
 }
+
+export interface ShortcutHandle {
+    onDestroy: () => void;
+    withShortcutSuffix: (string) => string;
+}
+
+/// Call within a component's onMount to assign the provided shortcut.
+export function useShortcut(
+    button: HTMLButtonElement,
+    shortcut = "",
+    optionalModifiers: Modifier[] | undefined = []
+): ShortcutHandle {
+    if (!shortcut) {
+        return {
+            onDestroy: () => {
+                return;
+            },
+            withShortcutSuffix: (s) => s,
+        };
+    }
+
+    const onDestroy = registerShortcut(
+        (event: KeyboardEvent) => {
+            button.dispatchEvent(new KeyboardEvent("click", event));
+            event.preventDefault();
+        },
+        shortcut,
+        optionalModifiers
+    );
+
+    const shortcutLabel = getPlatformString(shortcut);
+
+    return {
+        onDestroy,
+        withShortcutSuffix: (label: string | undefined) => {
+            return `${label ?? ""} (${shortcutLabel})`;
+        },
+    };
+}
```

dae (commented):
Are you still getting those errors? They don't seem to be happening for me when building or running bazel test. If nothing else helps, you could try a 'bazel clean', or a 'baze clean --expunge', though it'll mean a big redownload of dependencies+build.

I've only looked at it briefly so far, and probably won't have a good feel for it until I've had a chance to use it myself (maybe I could port deckoptions after this settles a bit), but the initial impressions are positive in terms of readability - it feels a lot more Sveltey. :-) Would be interested to hear your thoughts on how you're finding it.

How are you planning to allow add-ons to modify the buttons?

ts/editor/ColorButtons.svelte
dae: (or another possible alternative, if it made sense, would be to keep the existing WithShortcut, but use that in each component, instead of in the calling code?)

hgiesel:
> Are you still getting those errors? They don't seem to be happening for me when building or running bazel test. If nothing else helps, you could try a 'bazel clean', or a 'baze clean --expunge', though it'll mean a big redownload of dependencies+build.

I don't get this during build time, but during editing. I did make some progress, but I'm not quite there yet. While running VSCode, I get the following error message:

> Type '{ tooltip: string; active: any; }' is not assignable to type 'IntrinsicAttributes'.
  Property 'tooltip' does not exist on type 'IntrinsicAttributes'.ts(2322)

I get those in `ts/editor/Format{Inline,Block}Buttons.svelte`, e.g. `DropdownItem` uses it differently then buttons with tooltips. At some point I started using `appendInParentheses()`, you think that is sufficient?

ts/editor/ColorButtons.svelte
hgiesel: I originally had each button take a `shortcutLabel`, which made for awkward functions like this:

```js
export function mergeTooltipAndShortcut(
    tooltip: string | undefined,
    shortcutLabel: string | undefined
): string | undefined {
    if (!tooltip && !shortcutLabel) {
        return undefined;
    }

    let buf = tooltip ?? "";
    if (shortcutLabel) {
        buf = `${buf} (${shortcutLabel})`;
    }
    return buf;
}
```

It's also not quite clear cut, how an item should deal with the shortcut label, e.g. `DropdownItem` uses it differently then buttons with tooltips. At some point I started using `appendInParentheses()`, you think that is sufficient?

dae: I'm coming from the Qt world where things like shortcuts, hiding and so on is all implemented on QWidget, and inherited by other components like push buttons, so at first glance a ShortcutIconButton feels a bit strange to me. But that may be trying to map what I'm used to with Qt into a different domain :-)

ShortcutIconButton keeps the implementation of IconButton cleaner, but makes things a bit less ergonomic if the majority of our IconButtons have shortcuts. I also wonder how this will combine with other features - for the sake of argument for example, imagine we decide we want to add a hiding ability as well. If the logic is inside IconButton with a separate prop, users could use the same widget with hiding on/off, and shortcuts on/off. If we expose separate components for this functionality, I fear it may become awkward if we need to have separate IconButton, ShortcutIconButton, HidableIconButton, and ShortcutAndHidableIconButtons. What do you think?

dae:
Sorry for the delay; I can reproduce this locally as well, and I'll look into why the props are not getting picked up in the d.ts files in bazel-bin. In the mean time, you may be able to work around it by commenting out the sveltelib and component entries in 'paths' in tsconfig.json

dae:
I've pushed a fix to master for the missing props, and will see if I can get proper types for them going as well.

ts/editor/ColorButtons.svelte
hgiesel: I agree that is terrible:

>  IconButton, ShortcutIconButton, HidableIconButton, and ShortcutAndHidableIconButtons.

However that is also not a pretty option:

>  If the logic is inside IconButton with a separate prop, users could use the same widget with hiding on/off, and shortcuts on/off. 

Because it would lead to code that looks like this:

```javascript

{#if shortcuts && state}
    <WithShortcut>
        <WithState>
        <IconButton>
            ...
        </IconButton>
        </WithState>
    </WithShortcut>
{#else if shortcuts}
    <WithShortcut>
        <IconButton>
            ...
        </IconButton>
    </WithShortcut>
{#else if state}
    ...
{#else}
    ...
{/if}
```

Now, we could write the shortcut / state handling logic into each component directly, but then we'd have to duplicate the logic for each button, which I'd argue is the worst option to go with.

I've also intended to implement the hiding logic, which would make for 8 permutations altogether (if I remember my combinatorics correctly). That's make me think that the current implementation, with some helpers like `tooltip={appendInParentheses(tr.editingSetForegroundColor(), shortcutLabel)}` is not that bad after all. What do you think?

dae:
Proper typing for the .d.ts files is a pain at the moment, as it will require reworking the Svelte rules to process multiple Svelte files all at once, and we'll need to pass in all dependencies as well - given that currently the editor and svelte-check are still catching the errors, I've put off further work on this for now

ts/editor/ColorButtons.svelte
dae: We could avoid all the extra if statements if each "mixin" like WithShortcut just did a sensible thing when provided with a default value. useShortcut() above avoids adding a shortcut if an empty string is passed in, and a hideable mixin would behave as if it wasn't used if hidden=false - we'd just need to expose the prop in each component, and give it a default value.

> Now, we could write the shortcut / state handling logic into each component directly, but then we'd have to duplicate the logic for each button, which I'd argue is the worst option to go with.

But that's not really what the diff I posted above is doing, is it? The useShortcut() function behaves similarly to a mixin, with the logic in shortcuts.ts being reusable in other components as well. There are about 8 lines of boilerplate that each component would need to include, but that's mainly because Svelte doesn't give us an easier way to get a reference to the HTML element, and we can't have a helper function that calls onMount() itself. That's the appeal of using <WithShortcut ...> instead of useShortcut() - it would let us avoid some of that boilerplate.

dae: I think the crux of the issue is: should WithShortcut et al be placed inside each component, or should each component by wrapped by them in the calling code instead? I think the former option might be better, at least in cases like shortcuts and hiding, as:
- we'll be writing these components once, and using them many times in calling code. Having to wrap components in multiple mixins each time we use them is quite inconvenient and error-prone
- with default values for the props, each mixin can basically be a no-op if its functionality is not requested
- while it adds some boilerplate to each component implementation, the bulk of the logic is still contained in the mixin itself

hgiesel:
I've implemented the functionality that was available when using dynamic components, by which I mean:
1. You can hide/show buttons from arbitrary positions in the button group
2. You can dynamically add new buttons
3. You can dynamically add new button groups

For this there are the pairs of `ButtonGroup` <> `ButtonGroupItem` and `ButtonToolbar` <> `ButtonToolbarItem`.

They have to be used together, so they can communicate with each other. `Button{Group,Toolbar}` then exports the "extension API" as the `api` prop, which has to be bubbled up to the top component, so that it can be used from outside Svelte.

ts/editor/ColorButtons.svelte
hgiesel: Looking at the current code, you can see that it includes yet another `mixin`, called `ButtonGroupItem`, which could not be included in the individual button themselves (as they have to be interacted with from `ButtonGroup`.

I didn't tackle this yet, but I have to say, I'd really dislike, making the mixins no-ops on demand. It would mean that we have to make the arguments optional, and we'd muddy the signature of the mixin. Also the idea of "we're invoking the Svelte component, but we don't want it to have any effect whatsoever" seems just wrong to me. I'll give the initial `useShortcut` suggestion a try.

dae (commented):
Hi Henrik,

Thanks for all your work on this. I'd be interested to hear how you found this approach compared to the previous JS helper approach - do you feel it is a good path forward, or do you have concerns/think a different approach would be better?

The API for add-ons looks quite flexible and easy to use.

VS Code is currently showing me lots of errors in eg. FormatInlineButtons.svelte. Commenting out components and sveltelib from tsconfig.json's paths seems to fix it, without breaking the Bazel tests - do we need those lines?

There's a bunch of other errors of the form:

```
Type '{ key: string; update: () => boolean; }' is not assignable to type 'IntrinsicAttributes & { key: string; update: (event: Event) => unknown; state: unknown; }'.
  Property 'state' is missing in type '{ key: string; update: () => boolean; }' but required in type '{ key: string; update: (event: Event) => unknown; state: unknown; }'.
```

This seems to be because of the 'export let state: T' line in WithState - when passing props in slots, it looks like we should not be using 'export', and changing it to 'let state: T' seems to remove the majority of the errors. 

There are a few other errors that scripts/svelte-check picks up, like id and tooltip in DropdownItem not having default values.

> To implement the PreviewButton in Svelte, I need to include two i18n modules (ACTIONS, and BROWSING) for one label each, maybe this could be handled better

If you could leave comments in the file noting the strings that need moving/copying, I can run a script in the future to clone the keys and translations. Not a super high priority though.

A few other minor things:

hgiesel:
`WithState` could technically used for any kind of value, which is why I used the `T`. It seems pretty impossible to type correctly with Svelte though, so I just changed it to default to a `boolean` for now.

For `WithShortcut`, I noticed, that we can't inline them quite yet, because `PreviewButton` looks like this `tooltip={tr.browsingPreviewSelectedCard({ val: shortcutLabel })}`, whereas the others look like this `tooltip={appendInParentheses(tr.editingCustomizeCardTemplates(), shortcutLabel)`. That's kind of what I meant by flexibility.

> I'd be interested to hear how you found this approach compared to the previous JS helper approach - do you feel it is a good path forward, or do you have concerns/think a different approach would be better?

My motivation was not just to replace the old solution, but also have one approach, that can work across all views. When I started I wasn't so sure, but now I'm quite happy with how it turned out. I'd certainly want to try writing an add-on which uses the JS API directly, and see how well the ergonomics are. One functionality we lost is to directly modify the attributes on existing components (that are inserted directly with slots). How far these could be mimicked by just using JS (e.g. modifying event listeners with `getEventListeners`) is also something that would have to be tested / experimented with.

dae:
Shame about the lack of generics - hopefully that's something they address at one point, as I can see issues about it on their tracker.

browsingPreviewSelected seems to be a bit of a corner case, and if that's the only string holding us up, changing it is also an option :-) I do appreciate the desire to remain flexible though, and don't have super-strong feelings about this - we can always change it later if necessary.

> My motivation was not just to replace the old solution, but also have one approach, that can work across all views. When I started I wasn't so sure, but now I'm quite happy with how it turned out.

I'm happy to merge this if you think this is a good path - did you want me to merge it now, or hold off until you do the other experiments you mentioned?

hgiesel:
I'd say this is ready for merging (unless you want to tackle the issue with shortcuts mentioned).
I think it's more important be uniformly extensible, and in that sense, this PR has much more potential than the previous approach.

dae:
Let's get this merged 🚀

Issue #1068, opened 2021-03-13
move "type in your answer" comparison code into Javascript


dae:
Currently each Anki client differs in the way typing comparisons are shown, and it might be nice to unify these implementations into a single shared one. AnkiMobile uses the objc port of diff match patch, and AnkiDroid is using the Java one IIRC. The computer version is currently using python's difflib. 

The JS implementation of diff match patch would probably the best thing to try first. I suspect the issues described [here](https://forums.ankiweb.net/t/confusing-answer-comparison-on-ankimobile/8222) are due to the way AnkiMobile is using the objc library rather than a fundamental problem with diff match patch, but this would require some testing.

While working on this, it would also be nice to support the changes in https://github.com/ankitects/anki/pull/680 if it is practical.

teymour-aldridge:
dtolnay has written a Rust implementation of the "diff" part of diff match patch which might be useful? https://crates.io/crates/dissimilar

dae:
That man is a machine! I was aware of https://github.com/distill-io/diff-match-patch.rs, but dtolnay's one looks to be much more popular. Rust code works for me too, though the actual wrapping of each section with spans/css is still probably better done in typescript rather than building a HTML string in Rust.

mikehardy:
typescript for all the things ;-) (IMHO, but especially if there is still post-processing to do with spans and such, why split an implementation into "a new implementation" (Rust) plus you still have to do per-client implementations"? That's just rearranging deck chairs I think

But mostly noting that AnkiDroid is using "a" java port of diff-match-patch vs "the" java port, there have been 3 or 4 over time and we're stuck on an old one because of minSdkVersion issues as you saw in the issue you commented in the ankidroid repo. Apparently it's important for the world to have 30 diff match patches for each language

dae:
The ts/ folder in the desktop repo contains code intended to be used across the Anki clients. Whether we fully implement it in TS, or use TS to post-process some output from Rust, the code would be shareable in either case

snoyes:
[...]

PR #1167, opened 2021-05-16
Store given input of type field in the frontside into data attribute of tag

wathan:
Store the originally given answer in a data attribute of `typeans` in addition to the formatted answer comparison.

'Type in answer comparison' in the backside of the card scatters the originally given input from the frontside over many span tags. To get the original input from the backside template, template developers have to parse the built-in answer comparison output and stitch the spanned answer together.  This makes it hard to write custom answer comparisons.

In `correct` method of `Reviewer`, `data-given` custom attribute is added to `<code id="typeans">` in the backside to store the input information from the frontside. Card template developers can easily access this attribute via javascript.

Screen Shot 2021-05-16 at 14 27 40


dae:
Thanks for contributing! This seems like a reasonable approach, but I'd like to wait until #1068 is implemented first. That will save us from having to reimplement it on each platform, and it may be that a hook makes more sense at the time.

PR #1170, opened 2021-05-17
Fix column issues

RumovZ:
Fixes the browser bugs reported [here](https://forums.ankiweb.net/t/anki-2-1-45-alpha/10061/2) and two minor issues. @kleinerpirat, please let me know if there is there is still any issue with state changing/sorting on the next release. 🙂


dae:
Thanks for the quick fix!

PR #1172, opened 2021-05-18
Deck Options Top Bar

hgiesel:
Have deckoptions use the same components as Editor for the top bar.

It also introduces a theming mechanism for label buttons.
I also renamed `OptionsDropdown.svelte` to `SaveButton.svelte` as it seems more descriptive, as it is the most prominent component of that section.

### DIFF
--- a/ts/deckoptions/index.ts
+++ b/ts/deckoptions/index.ts
@@ -1,6 +1,10 @@
 // Copyright: Ankitects Pty Ltd and contributors
 // License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
+/* eslint
+@typescript-eslint/no-explicit-any: "off",
+ */
+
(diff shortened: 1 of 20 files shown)


ts/deckoptions/index.ts
```diff
@@ -10,20 +14,33 @@ import SpinBoxFloat from "./SpinBoxFloat.svelte";
...
     deckId: number
 ): Promise<DeckOptionsPage> {
-    checkNightMode();
-    await setupI18n({
-        modules: [ModuleName.SCHEDULING, ModuleName.ACTIONS, ModuleName.DECK_CONFIG],
-    });
-    const info = await getDeckOptionsInfo(deckId);
+    const [info] = await Promise.all([
+        getDeckOptionsInfo(deckId),
+        setupI18n({
```
hgiesel, line 25: This will make the fetches execute in parallel rather than sequentially.

dae:
Looks good - thanks Henrik!

PR #1173, opened 2021-05-18
Fix black workaround for format calls

RumovZ:
Apparently another environment is used for format calls like  `.\bazel run //pylib:format`, so we must not try to access "HOME" in this case.


dae:
Thanks!

PR #1177, opened 2021-05-19
Strip Unicode isolation characters from title

abdnh:
Unrelated: I'm getting this error when trying to open the new deck options screen running from source:
```
JS error /_anki/pages/deckoptions.js:22447 Uncaught (in promise) TypeError: Cannot read property 'decode' of undefined
```

It's coming from this line apparently: https://github.com/ankitects/anki/blob/581480191acc512b64b7af67f4660e3a49355f9b/ts/deckoptions/lib.ts#L18

### DIFF
--- a/qt/aqt/deckoptions.py
+++ b/qt/aqt/deckoptions.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import aqt
+from anki.lang import without_unicode_isolation
 from aqt import gui_hooks
 from aqt.qt import *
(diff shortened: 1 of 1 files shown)


dae:
Are you getting any errors when you build that don't clear up after using ./run a few times? I just tried running from source on Windows here, but it seems to work for me.

Thanks for the fix; I'll eventually remember that this is an issue on window titles on Windows :-)

abdnh:
The error mysteriously went away after pulling the latest changes from today. Previously, even `bazel clean --expunge` didn't help.

dae:
JS is the flakiest part of Anki's built system unfortunately :-(

PR #1178, opened 2021-05-19
Custom labels for flags

RumovZ:
This makes flags renamable via the browser sidebar. The custom labels are used in the browser and in the reviewer. I've also tried to make the future addition of new flags less painful by replacing exhaustive listings of flags with loops.
What do you think of adding a flag module? As flags are sometimes named after their index and sometimes after their colour, I think it would at least be nice to have all these related variables in one place. But I'm not sure it should be in utils.py.

### DIFF
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -42,6 +42,7 @@ from aqt.utils import (
     current_top_level_widget,
     ensure_editor_saved,
     getTag,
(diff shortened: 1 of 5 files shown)


dae:
Looks good from both a code and functionality perspective - thanks Rumo! A separate file would probably be better, as a "utils" file is a bit of an anti-pattern

PR #1180, opened 2021-05-20
Pronounce "[...]" as "blank" with TTS

### LINKED ISSUE #1171: customizable replacements in TTS filters
https://forums.ankiweb.net/t/ankimobile-tts-reading-cloze-as-blank/4241/5

I can think of two possibilities:

- storing a dict of replacement in the collection config of the form {"...": "blank"}. Would case-insensitive matches be sufficient, or do we need to support regexs?
- requiring the user to specify the replacements in the tts tag

The former is probably more user-friendly.

RumovZ:
Resolves #1171 (in my opinion).

Sidenote: When I run ` rslib:format`, I am drowned in warnings:

> Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
    Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.

### DIFF
--- a/ftl/core/card-templates.ftl
+++ b/ftl/core/card-templates.ftl
@@ -1,3 +1,5 @@
+# This word is used by TTS voices instead of the elided part of a cloze.
+card-templates-blank = blank
 card-templates-changes-will-affect-notes =
     { $count ->
         [one] Changes below will affect the { $count } note that uses this card type.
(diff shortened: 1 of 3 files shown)


dae:
Thanks Rumo!

Currently rustfmt.toml includes some nightly-only directives. Rust analyzer is built with nightly, so having the file lying around allows the imports to be reordered in a nice way each time a file is saved. The warnings on an explicit format are annoying, but can be safely ignored.

PR #1182, opened 2021-05-20
Change switch color depending on state

RumovZ:
On the forum, I said the existing green wouldn't work in day mode. After a bit of experimenting and changing the blue to a lighter hue as well, I take that back.
@dae, we had the colour discussion before. Do you think this is an improvement? @kleinerpirat, what do you think? The switch is also a bit less obtrusive with the new colours.

grafik grafik
grafik grafik

### DIFF
--- a/qt/aqt/switch.py
+++ b/qt/aqt/switch.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright: Ankitects Pty Ltd and contributors
 # License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+from typing import Tuple
 
 from aqt import colors
 from aqt.qt import *
@@ -21,6 +22,8 @@ class Switch(QAbstractButton):
         radius: int = 10,
         left_label: str = "",
         right_label: str = "",
+        left_color: Tuple[str, str] = colors.FLAG4_BG,
+        right_color: Tuple[str, str] = colors.FLAG3_BG,
         parent: QWidget = None,
     ) -> None:
         super().__init__(parent=parent)
@@ -29,6 +32,8 @@ class Switch(QAbstractButton):
         self.setSizePolicy(QSizePolicy.Fixed, QSizePolicy.Fixed)
         self._left_label = left_label
         self._right_label = right_label
+        self._left_color = left_color
+        self._right_color = right_color
         self._path_radius = radius
         self._knob_radius = radius - self._margin
         self._left_position = self._position = self._path_radius + self._margin
@@ -54,6 +59,11 @@ class Switch(QAbstractButton):
     @property
     def label(self) -> str:
         return self._right_label if self.isChecked() else self._left_label
+
+    @property
+    def knob_color(self) -> QColor:
+        color = self._right_color if self.isChecked() else self._left_color
+        return theme_manager.qcolor(color)
 
     def sizeHint(self) -> QSize:
         return QSize(
@@ -93,7 +103,7 @@ class Switch(QAbstractButton):
         )
 
     def _paint_knob(self, painter: QPainter) -> None:
-        painter.setBrush(QBrush(theme_manager.qcolor(colors.LINK)))
+        painter.setBrush(QBrush(self.knob_color))
         painter.drawEllipse(self._current_knob_rectangle())
 
     def _paint_label(self, painter: QPainter) -> None:


kleinerpirat:
Copying bits from [the forum post I just made](https://forums.ankiweb.net/t/anki-2-1-45-alpha/10061/45?u=kleinerpirat) here:

## [Codepen](https://codepen.io/kleinerpirat/pen/bGqgWYG) to experiment
Instead of changing the knob color, you could also change the background of the switch. By default, it is using var(--flag4-fg) for the card state and var(--flag3-fg) for the note state. The advantage would be that the knob's label is readable with any color.
1. card-view
2. note-view

image image

For these images I used `--slightly-grey-text` for the text, but personally, I prefer `--medium-border`. In day mode it would probably be too light, though. You can test the two variants by commenting out either one of them in the stylesheet.

RumovZ:
Thanks for the demo. I think that you can only really tell if it works if it's embedded in the actual GUI, though.

kleinerpirat:
Sure, you're right! I didn't get around to building Anki from source yet, but since there are some UX-Design challenges ahead where I could potentially help, I guess it's time that I do :)

dae:
The letter does feel a bit difficult to read at the moment. I have no strong feelings about using a separate colour for notes mode if that's what people would prefer :-) Would be interesting to see how @kleinerpirat's variants look like in the program too

RumovZ:
grafik grafik
grafik grafik

Of course, the contrast can be increased independent of the variant by using the foreground colours as @kleinerpirat has done in the meantime. I personally prefer the unobtrusive low-contrast version. Now that the mode can be identified by position, label _and_ colour, legibility should be less important.

RumovZ:
The narrow bright area in night mode looks a bit dirty in my opinion, but removing the margin doesn't work in Qt (and arguably in svelte neither):
grafik

So I would fo for a different style in that case:
grafik

kleinerpirat:
I've experimented a bit with that look as well. Here I gave the knob the same radius as the path:
### Night Mode
image image
### Day Mode
image image

I'm getting really switchy vibes from this design. It's quite satisfying to toggle.

- knob color: `--frame-bg`
- label color: `--text-fg`
- width of path reduced by 6.25% (really arbitrary, but I found the original a bit too wide)

To get rid of the narrow bright area, I changed this:
https://github.com/ankitects/anki/blob/25685f7083420adf5578ee8b90f3f97ec2d264d8/qt/aqt/switch.py#L37-L40

to this:
```
        self._path_radius = radius
        self._knob_radius = radius
        self._left_position = self._position = self._path_radius
        self._right_position = 2.8125 * self._path_radius + self._margin
```
(That 2.8125 is meant to correct for my 6.25% path-width-reduction.)

And this is the knob_rectangle:
```
    def _current_knob_rectangle(self) -> QRectF:
        return QRectF(
            self.position - self._knob_radius,  # type: ignore
            self._margin,
            2 * self._knob_radius,
            2 * self._knob_radius,
        )
```

dae:
- I've just tried out the original PR change, and it's already looking pretty good - night mode is easy to read, and the blue C seems fairly easy to read in day mode (green is a little harder)
- Increasing the circle size to remove the tiny margins does seem like an improvement
- I had trouble replicating the changes in your code @kleinerpirat - I'd recommend you use 'git diff > file.txt' and then copy&paste the text here, so we can apply the patch directly.

kleinerpirat:
Sure, here's the diff:
```
diff --git a/qt/aqt/switch.py b/qt/aqt/switch.py
index b478fbf4e..61c05a8c7 100644
--- a/qt/aqt/switch.py
+++ b/qt/aqt/switch.py
@@ -35,9 +35,9 @@ class Switch(QAbstractButton):
         self._left_color = left_color
         self._right_color = right_color
         self._path_radius = radius
-        self._knob_radius = radius - self._margin
-        self._left_position = self._position = self._path_radius + self._margin
-        self._right_position = 3 * self._path_radius + self._margin
+        self._knob_radius = radius
+        self._left_position = self._position = self._path_radius
+        self._right_position = 2.8125 * self._path_radius + self._margin
 
     @pyqtProperty(int)  # type: ignore
     def position(self) -> int:
@@ -61,13 +61,13 @@ class Switch(QAbstractButton):
         return self._right_label if self.isChecked() else self._left_label
 
     @property
-    def knob_color(self) -> QColor:
+    def path_color(self) -> QColor:
         color = self._right_color if self.isChecked() else self._left_color
         return theme_manager.qcolor(color)
 
     def sizeHint(self) -> QSize:
         return QSize(
-            4 * self._path_radius + 2 * self._margin,
+            3.75 * self._path_radius + 2 * self._margin,
             2 * self._path_radius + 2 * self._margin,
         )
 
@@ -85,7 +85,7 @@ class Switch(QAbstractButton):
         self._paint_label(painter)
 
     def _paint_path(self, painter: QPainter) -> None:
-        painter.setBrush(QBrush(theme_manager.qcolor(colors.FRAME_BG)))
+        painter.setBrush(QBrush(self.path_color))
         rectangle = QRectF(
             self._margin,
             self._margin,
@@ -97,17 +97,17 @@ class Switch(QAbstractButton):
     def _current_knob_rectangle(self) -> QRectF:
         return QRectF(
             self.position - self._knob_radius,  # type: ignore
-            2 * self._margin,
+            self._margin,
             2 * self._knob_radius,
             2 * self._knob_radius,
         )
 
-        painter.setBrush(QBrush(self.knob_color))
+        painter.setBrush(QBrush(theme_manager.qcolor(colors.FRAME_BG)))
         painter.drawEllipse(self._current_knob_rectangle())
 
     def _paint_label(self, painter: QPainter) -> None:
-        painter.setPen(QColor("white"))
+        painter.setPen(QColor(theme_manager.qcolor(colors.TEXT_FG)))
         font = painter.font()
         font.setPixelSize(int(1.5 * self._knob_radius))
         painter.setFont(font)
```
This is with the width change. The only thing that still bugs me here is that the label seems to be a bit off-center (lower). Perhaps @RumovZ knows how to fix this.

RumovZ:
I feel that the knob should either be smaller or larger than the path. That's personal preference, of course, but it's also practical to avoid the ugly outline seen in my screenshot above.
Changing the constants to arbitrary values isn't a viable solution, because it makes figuring out the exact positions intractable. And without comprehensible positioning, I'm not confident that it will work with different parameters and environments.
As for the label, I think that's just caused by the size being too large.
I'll make a new PR with the other style I mentioned.

dae:
Thank you both, I've just merged in #1188

Issue #1162, opened 2021-05-08
Shortcuts based on raw keycodes need a rethink


dae:
@hgiesel Unfortunately I'm not sure the current approach with raw keys is going to work. Consider a user who uses a Dvorak keyboard layout for example, where most physical keys represent a different letter. When the user sees a shortcut like 'cmd+shift+c', they expect to press the key that presents a c in Dvorak, and not the physical c key. I would have caught this myself (I'm a Dvorak user), but my main machine uses a hardware Dvorak keyboard that remaps the physical keycodes.

I suspect it will adversely affect users with some language keymaps as well.

Original report from @nwwt:
https://forums.ankiweb.net/t/card-browser-hotkeys-semi-broken/9815/3

hgiesel:
Yes, you're right. We should use `event.key` for these cases. I wonder if we can use `event.code` for some cases though, e.g.  "Control+Shift+2". We could only check for "Control+@", which in turn could be in inconvenient positions

hgiesel:
1. What I meant was that displaying: "@", we could display "Shift+2", and activate it on clicking `Shift+Digit2` based on `event.code`. The digits are more consistent on keyboards, and if they not on they keyboard top row, they are usually completely missing from that locale (e.g. Thai or Vietnamese keyboard layouts). This is not that relevant yet, but could be interesting once we move the reviewer shortcuts to JS.

2. There is one possible complication I noticed. We do have some which involve the Alt modfier, like the cloze shortcuts. On Macs, the Alt modifier may also change the output of `event.code`. [For example](https://keycode.info/) `Ctrl+Alt+Shift+C` will give me `¸`. Weirdly enough, that does not happen on Chromium inside Anki, but it does happen in any up-to-date browser on my machine.

dae:
@hgiesel is this safe to close now, or are there still issues remaining?

hgiesel:
I think this is solved.

PR #1183, opened 2021-05-20
Switch to event.key for keyboard sequences

hgiesel:
This adresses #1162 by moving to `event.key` instead of `event.code`. I still think there might be use cases for `event.code`, as I mentioned in the issue, for when we port the reviewert shortcuts.

This also removes the keyboard sequence timeout, and disables a started keyboard sequence if you press a "general key" (characterized by `event.location === 0`).

### DIFF
--- a/ts/components/WithShortcut.svelte
+++ b/ts/components/WithShortcut.svelte
@@ -8,7 +8,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import { onDestroy } from "svelte";
     import { registerShortcut, getPlatformString } from "lib/shortcuts";
 
-    export let shortcut: string;
+    export let shortcut: string[][];
     export let optionalModifiers: Modifier[] | undefined = [];
 
     const shortcutLabel = getPlatformString(shortcut);
--- a/ts/components/WithShortcut.svelte
+++ b/ts/components/WithShortcut.svelte
@@ -6,8 +6,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import { onDestroy } from "svelte";
     import { registerShortcut, getPlatformString } from "lib/shortcuts";
 
-    export let shortcut: string[][];
-    export let useCode = false;
+    export let shortcut: string;
 
     const shortcutLabel = getPlatformString(shortcut);
 
@@ -15,14 +14,10 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     function createShortcut({ detail }: CustomEvent): void {
         const mounted: HTMLButtonElement = detail.button;
-        deregister = registerShortcut(
-            (event: KeyboardEvent) => {
-                mounted.dispatchEvent(new MouseEvent("click", event));
-                event.preventDefault();
-            },
-            shortcut,
-            useCode
-        );
+        deregister = registerShortcut((event: KeyboardEvent) => {
+            mounted.dispatchEvent(new MouseEvent("click", event));
+            event.preventDefault();
+        }, shortcut);
     }
 
     onDestroy(() => deregister());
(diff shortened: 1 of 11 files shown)


hgiesel:
There is one more complication:
Let's take the subscript shortcut, which is `Control+=`. On an US standard keyboard, you gotta press `Control+Equals`, whereas e.g. on a German keyboard you have to press `Control+Shift+0`.
Currently Anki will reject this, as we've programmed it only take `Control` as a modifier, and all others have to be not-pressed for the shortcut to fire.

Not quite sure how to solve this.
As we'll never know which `event.key` values are hidden behind modifiers, this basically means we can only use shortcuts in the form of `Control+X`, because both Shift and Alt can change the output of `event.key`. Another approach would be to use optional modifiers more generously to cover most cases...

dae:
Thanks for working on this. I'm not sure what the best path forward here is either, and Qt had some issues with certain keystrokes as well. If there aren't many shortcuts that are affected by this, maybe we should consider changing them to something more compatible? I thought about making some keystrokes translatable, but we can't be sure the user's keyboard will match their interface language, and presumably that would be a pain with keystrokes listed as an array.

Will event.location === 0 work with a numeric keyboard, which appears to specify a different location?

hgiesel:
Another complication: Control+Shift+Equals and Control+Equals will both produce `=` as event.key. This means, we can no longer distinguish between `=` and `+` even on US ANSI keyboards.

I had a look into Summernote and TinyMCE, and investigated how they deal with shortcuts. Both use `event.key` for shortcuts which involve letters, and use `event.code` for shortcuts with the digit keys `Digit{0-1}`. Summernote also uses `event.code` for symbols (Remove formatting is `Ctrl+\`, where `\`, will always be the `Backslash` key), whereas TinyMCE completely avoids keys beyond the digit keys, and event.key with letters. Also, both avoid using the Alt/Option modifier with letters.

This new version uses `event.key` for shortcuts based on letters, and `event.code` for the Superscript and Subscript shortcuts, Enter, and Tab. I also left in the Cloze shortcut, as it still works, although we need to keep an eye once the underlying Chromium is updated, and then might have to change the shortcut.

dae:
Qt 6.2 is planned for the end of Sep, so a new Chromium is hopefully not too far off.

Another issue I noticed is that the latest Chrome and Safari use the lowercase version in 'key' when shift is held down, but Qt uses the uppercase version, so I'm guessing we'll need to lowercase or uppercase the codes before comparison. What a mess ^_^;

I feel a bit uneasy about this current approach, but our options seem to be limited here. :-( Maybe one alternative would be to revert to the keycode approach and have some translation layer to map physical keys to their virtual equivalents. It would require the user to specify the layout they're using in the preferences though, and we'd need to include maps for the layouts in our code.

Are you using a qwertz keyboard? If so, when we were using the original keycode approach, was that working fine on your keyboard? I'm wondering whether it's just dvorak/colemak we need to worry about, or whether there are a bunch of other languages we'd need to provide translations for as well.

hgiesel:
> Qt 6.2 is planned for the end of Sep, so a new Chromium is hopefully not too far off.

But AFAIK web engine is still not ported?

> Maybe one alternative would be to revert to the keycode approach and have some translation layer to map physical keys to their virtual equivalents. It would require the user to specify the layout they're using in the preferences though, and we'd need to include maps for the layouts in our code.

From my experience, people who use multiple keyboard layouts typically switch between keyboard layouts, whenever they switch the language. For example, they will switch to a German layout to type German, and switch back to English to type English. Having to set this in the preferences

hgiesel:
The only small issue remaining is with symbol keys, which you can see with the example of subscript and superscript again. The shortcut will show `Ctrl+=`, which on a German keyboard e.g. would be `Ctrl+´`, however the shortcut still works correctly. On a Dvorak layout it is actually mapped correctly to `BracketRight` key, because it's a perfect match for it.

However I think that's as close as we can get to the perfect solution.

dae:
Thanks Henrik, let's give this a go.

PR #1185, opened 2021-05-21
Scroll pixels instead of columns horizontally

RumovZ:
https://forums.ankiweb.net/t/unintuitive-horizontal-scrolling-in-the-card-browser/10161


dae:
Thanks Rumo!

PR #1188, opened 2021-05-22
Change switch color depending on state

RumovZ:
Same as #1182 with the knob overlapping instead of embedded in the path.

### DIFF
--- a/qt/aqt/switch.py
+++ b/qt/aqt/switch.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright: Ankitects Pty Ltd and contributors
 # License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+from typing import Tuple
 
 from aqt import colors
 from aqt.qt import *
@@ -21,6 +22,8 @@ class Switch(QAbstractButton):
         radius: int = 10,
         left_label: str = "",
         right_label: str = "",
+        left_color: Tuple[str, str] = colors.FLAG4_BG,
+        right_color: Tuple[str, str] = colors.FLAG3_BG,
         parent: QWidget = None,
     ) -> None:
         super().__init__(parent=parent)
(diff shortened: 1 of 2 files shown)


RumovZ:
The size set in browser should probably be smaller in this case.

kleinerpirat:
Looking good!

Hope I'm not frustrating you yet. I saw this button discussion as a chance to get used to the Anki git workflow, but please don't let your development progress be negatively impacted by such a trivial issue.

<details>
  <summary>Two suggested tweaks, if you're interested.</summary>

```
diff --git a/qt/aqt/switch.py b/qt/aqt/switch.py
index e8d835d8a..0891a54b5 100644
--- a/qt/aqt/switch.py
+++ b/qt/aqt/switch.py
@@ -103,12 +103,13 @@ class Switch(QAbstractButton):
 
         painter.setBrush(QBrush(theme_manager.qcolor(colors.FRAME_BG)))
+        painter.setPen(theme_manager.qcolor(colors.FAINT_BORDER))
         painter.drawEllipse(self._current_knob_rectangle())
 
     def _paint_label(self, painter: QPainter) -> None:
         painter.setPen(theme_manager.qcolor(colors.SLIGHTLY_GREY_TEXT))
         font = painter.font()
-        font.setPixelSize(int(1.2 * self._knob_radius))
+        font.setPixelSize(int(1.33 * self._knob_radius))
         painter.setFont(font)
         painter.drawText(self._current_knob_rectangle(), Qt.AlignCenter, self.label)
```
### This PR
image image
### Tweaks
image image

This increases the label font size a little bit and uses a faint border for the knob. I find that this helps discern the button from the window background a bit better in light mode.
</details>

RumovZ:
Not frustrated, yet. 😉 
I'm impartial about the border.
The font size is problematic because the font rendering is so inconsistent. What looks fine on your machine is already too large on mine (note the offset again):
grafik
So I would err on the side of caution and maybe even decrease the factor to 1.1.
I also note that the whole widget is smaller on your machine, at least compared to the other GUI elements.

kleinerpirat:
Ah I see, no big deal.

I think I'll withhold from interfering in Qt related design processes in the future, since I lack the experience you, Damien & the others got with that framework.

If I'm not mistaken, Anki is shifting towards a web-UI anyway, so there will be plenty of opportunities for me to contribute in the future.

RumovZ:
Oh, I pick it up as I go along as well. 🙂 But indeed, for better or worse, everything's heading towards web-based technologies nowadays.

dae:
Thanks to you both! It looks good, and this variant seems to be a bit more popular in the forum poll.

PR #1189, opened 2021-05-22
Fix browser.onChangeModel()

RumovZ:
https://forums.ankiweb.net/t/error-message-when-trying-to-change-note-type-2-1-45a2/10183

I deleted the old helper method because I think it behaved a bit awkward. Or should it be kept around for legacy reasons?


dae:
There appears to be only one fairly unpopular add-on using it, so probably not worth keeping around a legacy version just for that. Thanks for the fix!

PR #1190, opened 2021-05-23
Make legacy buttons look and feel like Svelte buttons

kleinerpirat:
Credit to forum user [Ab_Bc](https://forums.ankiweb.net/t/anki-2-1-45-alpha/10061/4?u=kleinerpirat) for the inspiration.

## Showcase
The buttons should behave exactly like the Svelte buttons, except for the transition (explained in comment below):
Screenshot from 2021-05-23 01-42-49
Screenshot from 2021-05-23 01-44-55

## TODO: Fix wrapping

Currently, the button groups wrap in a very wasteful manner:

image

Perhaps we can find a solution to split the legacy group dynamically on resize. That would fix the issue of these slightly overlapping buttons of a single group.

A CSS-only way would be nice, but I don't think it's possible.


kleinerpirat:
I had to remove the transition imitation for the legacy buttons, because it caused a dark-colored blink. The transition seems to conflict with the linear gradient here:

https://github.com/ankitects/anki/blob/195c41cba3b107db80e0df9ede26dbe90effdfbb/ts/sass/buttons.scss#L31-L36

Dropdown items are currently affected by the special night mode treatment. This results in a weird plastic look, so I excluded the styling, but kept the hover effect:

### Before PR
Screenshot from 2021-05-23 16-06-23
### After PR
Screenshot from 2021-05-23 16-16-43

dae:
The original version posted on the forums messed up the alignment of the image occlusion add-on's icon; this doesn't, so that's good. :-) But the inversion of colours in night mode makes things look odd on coloured buttons:

Screen Shot 2021-05-24 at 11 48 03 am

And it seems to differ between the add screen and the browse screen:

Screen Shot 2021-05-24 at 11 56 22 am

Making the legacy buttons look better is a good goal, but I think any changes you make should be self-contained, and only apply when they match one of the legacy selectors. If there's a good reason for changing the standard styling as well, that's best done as a separate PR. Your changes to buttons.scss has changed the appearance of buttons across the app - see the opening deck screen for example:

Screen Shot 2021-05-24 at 11 59 16 am
Screen Shot 2021-05-24 at 11 58 58 am

I got the same impression about hovering being laggy, and I think I prefer the faster response in this PR to the status quo

@hgiesel you're most familiar with this code; would appreciate your input :-)

kleinerpirat:
Thank you for your feedback!

I reverted the changes I did to `buttons.scss`, but made some styling (padding, border) only apply to buttons that are not in a `btn-group`.

Regarding the inversion of colors: That was the default behaviour pre-2.1.45 as well (looking at the Closet image occlusion button, which is light blue by default):
IMG-20210504-WA0006__01
I think the benefits far outweigh the disadvantages here, but perhaps one could implement an option to ignore the inversion (e.g. with a `no-inversion` class for these special cases).

kleinerpirat:
I am happy with the current state of this PR. Please let me know if the solution with `!important` (commit https://github.com/ankitects/anki/pull/1190/commits/eedc8274da571145f41a7a7041b85b602954e021) is valid.

`buttons.scss` remains untouched and the legacy buttons behave exactly like the new ones, except for the transition, which I would like to adress in a new PR, together with the dropdown items.

hgiesel:
One issue I noticed (and why I didn't go down that route myself) is that inverting is not always a clearcut operation for the buttons:
Here you can see the "Mini Format Pack" button for background color, which shows green but which is actually inverted magenta:
Screenshot 2021-05-24 at 19 17 54

kleinerpirat:
Thanks @hgiesel for the feedback!

> Here you can see the "Mini Format Pack" button for background color, which shows green but which is actually inverted magenta.

This is easily fixed by restricting the inversion to images:

Screenshot from 2021-05-24 19-49-13

dae:
The risk of !important is that it can sometimes override explicit styling that add-ons have provided, but hopefully that's not an issue here. Since Henrik didn't raise any other issues, I assume that's a tacit approval - if I've got that wrong Henrik, please let me know. :-) In any case, thank you for your work on this Matthias!

PR #1192, opened 2021-05-24
Tweak button styling for more consistency

kleinerpirat:
This addresses two issues from https://github.com/ankitects/anki/pull/1190:
### status quo
Screenshot from 2021-05-23 16-06-23
### PR
Screenshot from 2021-05-23 16-16-43
## Transition
By default, Svelte buttons have a .15s background transition on hover. This results in a slow, perhaps even laggy feeling.
### Changes
- remove button styling from dropdown items
- override the default background transition in favour of instant hover response

---

### Extra
@dae After you showed me the three buttons on the bottom of the main window, I noticed they don't cause the cursor to switch to pointer mode. Since I couldn't think of a button that wouldn't benefit from this, I made it the default behaviour with https://github.com/ankitects/anki/commit/f2c286792af8b1e0b9c4af71f4813eaf4d0cd10c.

I would love to hear everyone's feedback on these changes and potential other improvements.

### DIFF
--- a/ts/sass/buttons.scss
+++ b/ts/sass/buttons.scss
@@ -23,6 +23,10 @@
     }
 }
 
+button {
+    cursor: pointer;
+}
+
 .nightMode {
     button {
         -webkit-appearance: none;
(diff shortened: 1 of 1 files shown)


kleinerpirat:
Closed with https://github.com/ankitects/anki/pull/1193.

PR #1193, opened 2021-05-24
Tweak button styling for more consistency 

hgiesel:
@kleinerpirat
Closes #1192. This puts the styling directly into the components. I'm somewhat trying to move away from `buttons.scss` (and `bootstrap-dark.scss`).

Regarding the cursor change: The other Qt buttons and other input widgets don't change the cursor either, so for consistency for now I'd still leave it out. (Looking at an electron app like Slack however, I can see that they make all buttons/inputs change the cursor).


dae:
Thank you both!

PR #1194, opened 2021-05-25
Fix height misalignment of legacy buttons

kleinerpirat:
Follow-up on https://github.com/ankitects/anki/pull/1190 due to crippling perfectionism.
Sorry for not noticing earlier, but the legacy button height didn't match the default button height.

### Before
image

**The mismatch causes this atrocity** (when all buttons are in one line):
image

The color picker seems to be the only button with a fixed height of 28px, all the others adjust to the greatest height (in that case defined by the legacy buttons) automatically.

---

### After
image

**Misalignment fixed:**
image

## Changes:
- add min-height 28px
- set padding to 0px (class btn-group has enough padding on its own)

### DIFF
--- a/ts/editor/legacy.scss
+++ b/ts/editor/legacy.scss
@@ -12,8 +12,9 @@
     border-bottom-right-radius: var(--border-right-radius) !important;
     
     min-width: 28px;
+    min-height: 28px;
     margin-left: -1px;
-    padding: 3.5px !important;
+    padding: 0px !important;
 }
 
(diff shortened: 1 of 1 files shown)


hgiesel:
The height/width should be based on `--toolbar-size`, not static. I'll have a look into it.

kleinerpirat:
Continued with https://github.com/ankitects/anki/pull/1195.
Thanks Henrik!

PR #1195, opened 2021-05-25
Fix height misalignment of legacy buttons

hgiesel:
Another approach to #1194.

@kleinerpirat Could you have a look this covers all your cases? Specifically, I removed this line: ` filter: invert(1) grayscale(1) brightness(100)`. I couldn't quite see what was its purpose, and it removed the color of the color selector from "Mini format pack" upon `:active` state.

### DIFF
--- a/ts/editor/legacy.scss
+++ b/ts/editor/legacy.scss
@@ -10,10 +10,11 @@
     border-bottom-left-radius: var(--border-left-radius) !important;
     border-top-right-radius: var(--border-right-radius) !important;
     border-bottom-right-radius: var(--border-right-radius) !important;
-    
-    min-width: 28px;
+
+    min-width: calc(var(--toolbar-size) - 2px);
+    min-height: calc(var(--toolbar-size) - 2px);
     margin-left: -1px;
-    padding: 3.5px !important;
+    padding: 0 !important;
 }
 
 .linkb:hover,
@@ -45,7 +46,7 @@
     .linkb {
         margin-left: 1px;
     }
-    
+
     .linkb:hover {
         background-color: #7a7a7a;
         border-color: #7a7a7a;
@@ -59,7 +60,7 @@
             rgba(0, 0, 0, 0.35);
         border-color: #525252;
     }
-    
+
     img.topbut,
     .linkb:active:hover img.topbut {
         filter: invert(1);
--- a/ts/editor/legacy.scss
+++ b/ts/editor/legacy.scss
@@ -1,9 +1,11 @@
-@use "ts/sass/button_mixins" as button;
-
 .linkb {
     display: inline-block;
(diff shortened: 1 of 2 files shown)


kleinerpirat:
Thank you for the elegant restructuring, Henrik!

The following misalignment is caused by your usage of `height` instead of `min-height`:
image

---

The purpose of `filter: invert(1) grayscale(1) brightness(100)` was to imitate the effect of the Svelte buttons in day mode as best as I could.
Compare clicking and holding a default button and an add-on button in day mode in this PR and you will see that they differ. The default buttons invert their color while the legacy buttons stay the same.
`grayscale(1) brightness (100)` was meant to ensure that it is white.

Screenshot from 2021-05-25 15-49-20

With my first PR I wanted to create a perfect replica of the Svelte buttons, but perhaps that's not necessary.

hgiesel:
> Compare clicking and holding a default button and an add-on button in day mode in this PR and you will see that they differ. The default buttons invert their color while the legacy buttons stay the same.

Are you sure, that's not the effect of some add-on? I cannot recreate it on my side.

Screenshot 2021-05-25 at 20 32 42

kleinerpirat:
Yes, I'm positive. I disabled all the add-ons and it looks like this on my Linux system:

buttons

hgiesel:
@kleinerpirat 
Hm, I cannot recreate this on my Mac. I also cannot recreate if I change the `isMac` `isLin` `isWin` selectors in the HTML. Very mysterious...

kleinerpirat:
Not a big deal i guess :D

If it shouldn't be there by default, it was certainly for the best that you removed the filter 👍🏻

I didn't notice any other issues with this PR.

Thanks again for the help with this! I can't wait for the other users to see the shiny new buttons ☺️

hgiesel:
I guess you didn't see my Edit:

Looking at it through the Web Inspector, can you tell the HTML file it comes from. The selectors should have the file of origin next to them. Could you have a look there?

Screenshot 2021-05-25 at 22 23 27

I'd still be interested in where the color comes from...

kleinerpirat:
Ah yes, I didn't see the edit because GitHub on the phone doesn't refresh automatically.

As you can see, I got a mysterious button styling coming out of nowhere:

image

With all add-ons except WebView inspector disabled, this is very odd indeed...

hgiesel:
Then this must be part of the base CSS. I'm not even sure who exactly provides it: the OS, Chromium, or Qt?

Also not sure whether to consider this something to fix: what would you say, @dae?

dae:
It's probably part of the Qt or Chromium theme, and I see it on an Ubuntu 18 system here as well, so I'm guessing it affects all/many Linux users, so it would be nice to fix if we can do so in a clean way. Could we just specify the color for hover/active explicitly instead, with a comment that it's required to override the default on Linux?

PR #1196, opened 2021-05-25
Use TextInputModal directly

hgiesel:
A `Map` containing all modals is placed in a `context` object, from where the modal can than be referenced. I think this makes the dependency much clearer, also it avoids building the modal on every invocation.


ts/deckoptions/DeckOptionsPage.svelte
```diff
@@ -41,21 +41,5 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     onDestroy(() => registerCleanup?.());
 </script>
 
-<style lang="scss">
-    .editor {
-        // without this, the initial viewport can be wrong
-        overflow-x: hidden;
```
dae, line 47: I think moving to a toolbar for the top area removed the need for this. One downside to the change though is that the top when the window is very narrow, the save button wraps onto a separate line, which doesn't look great. You can reproduce it in the desktop by adding a config with a long name, then making the window narrow. Is there a way we could limit the left dropdown to say 70-80% of the total width?

dae (commented):
Thanks Henrik!

PR #1197, opened 2021-05-25
Escape HTML in template error message

RumovZ:
This was leading to hard to decipher error messages in broken templates like `{{Field 1} <hr> {{Field 2}}` (no such field).


dae:
Thanks!

PR #1200, opened 2021-05-27
Template checks

RumovZ:
What do you think of this approach? The wording could probably be improved.

Here are two more possible checks:
- Currently, non-existing field names generate an error, but the notetype can be saved anyway. Should the user be required to fix the names first?
- Should the user be required to have at least one field on the front side?


qt/aqt/clayout.py
```diff
@@ -791,7 +792,7 @@ def on_done(changes: OpChanges) -> None:
 
         update_notetype_legacy(parent=self.mw, notetype=self.model).success(
             on_done
-        ).run_in_background()
+        ).failure(lambda e: showWarning(str(e))).run_in_background()
```
dae, line 795: That's the default behaviour if you don't specify it, and the default uses the provided parent as well. :-)

ftl/core/card-templates.ftl
```diff
@@ -21,6 +21,10 @@ card-templates-night-mode = Night Mode
 card-templates-add-mobile-class = Add Mobile Class
 card-templates-preview-settings = Options
 card-templates-invalid-template-number = Card template { $number } in notetype '{ $notetype }' has a problem.
+card-templates-identical-front = Its front side is identical to the one of card template { $number }.
```
dae, line 24:
```suggestion
card-templates-identical-front = The front side is identical to card template { $number }.
```

ftl/core/card-templates.ftl
```diff
@@ -21,6 +21,10 @@ card-templates-night-mode = Night Mode
...
+card-templates-missing-cloze = The 'cloze' filter must be used on both sides of a cloze template.
```
dae, line 25:
```suggestion
card-templates-missing-cloze = Expected to find '{{cloze:Text}}' or similar on the front and back of the card template.

File: ftl/core/card-templates.ftl [hunk 4]
```diff
@@ -21,6 +21,10 @@ card-templates-night-mode = Night Mode
...
+card-templates-extraneous-cloze = The 'cloze' filter can only be used on cloze templates.
```
dae, line 26:
```suggestion
card-templates-extraneous-cloze = 'cloze:' can only be used on cloze notetypes.
```

ftl/core/card-templates.ftl
```diff
@@ -21,6 +21,10 @@ card-templates-night-mode = Night Mode
...
+card-templates-see-preview = See the render preview for more information.
```
dae, line 27:
```suggestion
card-templates-see-preview = See the preview for more information.
```

ftl/core/card-templates.ftl
dae: {{FrontSide}} is valid on the back of a cloze template IIRC, and the current check doesn't handle it.

rslib/src/notetype/mod.rs
```diff
@@ -330,8 +371,10 @@ impl Notetype {
             return Err(AnkiError::TemplateSaveError(TemplateSaveError {
                 notetype: self.name.clone(),
                 ordinal: idx,
+                details: TemplateSaveErrorDetails::TemplateError,
             }));
         }
+        self.ensure_cloze_if_and_only_if_cloze_notetype(&parsed_templates)?;
```
dae, line 377: For issues like this that can be checked with a single template, maybe we should just show them in the preview area immediately, instead of waiting until save time to report them?

dae (commented):
> Currently, non-existing field names generate an error, but the notetype can be saved anyway. Should the user be required to fix the names first?
> Should the user be required to have at least one field on the front side?

Both of those seem sensible to me, WDYT?

rslib/src/notetype/mod.rs
dae: On the other hand, it would mean a bit more processing in the common case of displaying a card. hmm.

qt/aqt/clayout.py
RumovZ: Right, I intended to do something more originally, like switching to the faulty template. But I don't see how that would be possible with the current implementation.

ftl/core/card-templates.ftl
RumovZ: I just checked and the implicit `{{cloze:text}}` in `{{FrontSide}}` isn't expanded, i.e. it doesn't reveal the answer when used on the back side.

dae: Ah, sorry, the brackets needed to be escaped

rslib/src/notetype/mod.rs
RumovZ: I think there may also be cases where the template becomes temporarily invalid while the user is working on it, and it would be disruptive to obscure the preview unnecessarily.

dae: That happens when the user is typing in something like `{{#Field}}` as well though, and I'm not sure it's such a big problem? :-) But I don't feel strongly either way

dae: Let's just keep it the way it is for now then :-)

pylib/tests/test_collection.py
```diff
@@ -143,10 +143,6 @@ def test_furigana():
     n["Front"] = "foo[sound:abc.mp3]"
     n.flush()
     assert "anki:play" in c.q(reload=True)
-    # it shouldn't throw an error while people are editing
-    m["tmpls"][0]["qfmt"] = "{{kana:}}"
-    mm.save(m)
-    c.q(reload=True)
```
RumovZ, line 149: I've already asked if you are okay with enforcing valid field names, but just to be sure: It's okay to remove this test and reject something like `{{kana:}}`?

rslib/src/notetype/mod.rs
```diff
@@ -272,6 +272,103 @@ impl Notetype {
...
+        }
+    }
+
+    /// True if any name in names does not belong to a field of this notetype.
+    fn unknown_field_name<T, I>(&self, names: T) -> bool
+    where
+        T: IntoIterator<Item = I>,
+        I: AsRef<str>,
+    {
+        let special_fields = [
```
RumovZ, line 331: The only other place I could find where we list out special fields is `notetype/render.rs/add_special_fields()`. It would probably make sense to exhaustively iterate over a list of all special fields like this array here.
Should I maybe move the array into `render.rs` and make it public?

rslib/src/notetype/mod.rs
```diff
@@ -272,6 +272,103 @@ impl Notetype {
...
+            "Subdeck",
+            "Tags",
+            "Type",
+        ];
+        names.into_iter().any(|name| {
+            !special_fields.contains(&name.as_ref())
+                && self
+                    .fields
+                    .iter()
+                    .map(|field| &field.name)
```
RumovZ, line 345: (Just noted that this map is redundant.)

pylib/tests/test_collection.py
dae: the empty field name is a special case used by add-ons; we need to keep supporting that
https://github.com/ankitects/anki-addons/blob/5f775143c2cf906f689bac6d83f47493494e5e0d/demos/field_filter/__init__.py#L22

rslib/src/notetype/mod.rs
dae: If we'll use it in multiple locations, by all means make it pub(crate). If you meant completely public, I'd prefer to keep things limited to the crate unless there's a clear use case for outside consumers, as it's extra API surface we need to support

RumovZ: I can't get it to work, unfortunately. If I do it like this:

```rust
SPECIAL_FIELDS.iter().map(|&field| match field {
            "Tags" => map.entry(field).or_insert_with( || tags.into()),
            "Type" => map.entry(field).or_insert_with(|| nt.name.clone().into()),
            "Subdeck" => map.entry(field).or_insert_with(|| subdeck_name.to_string().into()),
            "Deck" => map.entry(field).or_insert_with(|| deck_name.to_string().into()),
            "CardFlag" => map.entry(field).or_insert_with(|| flag_name(card.flags).into()),
            "Card" => map.entry(field).or_insert_with(|| template.name.clone().into()),
            // handled by Python
            "FrontSide" => "",
        });
```

I get ```captured variable cannot escape `FnMut` closure body```, which makes sense because the compiler can't know that every closure will only be called once.
Do you know of an idiomatic way to ensure we have handled (almost) all values in the array?
I could write a test, but this wouldn't work at compile time, of course.

dae: You could perhaps accomplish it with strum, but if it proves too complicated, feel free to duplicate the list and just leave comments pointing to the other location that needs to be changed when updates are made - it's not an area that gets updated frequently.

rslib/src/notetype/mod.rs
```diff
@@ -322,25 +332,17 @@ impl Notetype {
...
-            "Deck",
-            "Subdeck",
-            "Tags",
-            "Type",
-        ];
         names.into_iter().any(|name| {
             // The empty field name is allowed as it may be used by add-ons.
             !name.as_ref().is_empty()
-                && !special_fields.contains(&name.as_ref())
+                && !SPECIAL_FIELDS.contains(&name.as_ref())
```
dae, line 345: Maybe this would be better as a lazy_static HashSet?

dae: (I've found myself wondering on multiple occasions where exactly the inflection point is where a set becomes cheaper, but never bothered to benchmark it before)

RumovZ: Yes, that would be interesting to know. I can't imagine it makes a difference in performance here, but I guess it's more idiomatic. 🙂

dae:
Seems to work well, thanks Rumo. Hopefully this reduces the "my cloze template doesn't work" questions a bit!

PR #1202, opened 2021-05-27
Uniformly use properties to Button{Toolbar,Group} for setting button properties

hgiesel:
Instead of using `WithTheming` around `ButtonToolbar`/`ButtonGroup`.

Also fixes Deck options toolbar to pop into two lines, if the preset name is too long.

### DIFF
--- a/ts/components/ButtonDropdown.svelte
+++ b/ts/components/ButtonDropdown.svelte
@@ -6,7 +6,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import { setContext } from "svelte";
     import { dropdownKey } from "./contextKeys";
 
-    import WithTheming from "./WithTheming.svelte";
     import ButtonToolbar from "./ButtonToolbar.svelte";
 
     export let id: string;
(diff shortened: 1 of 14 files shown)


dae:
Not sure of the best place to put it, but we'll need an overflow-x somewhere, eg

```patch
diff --git a/ts/deckoptions/ConfigSelector.svelte b/ts/deckoptions/ConfigSelector.svelte
index 524717d49..2c10e994b 100644
--- a/ts/deckoptions/ConfigSelector.svelte
+++ b/ts/deckoptions/ConfigSelector.svelte
@@ -116,3 +116,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         </ButtonToolbarItem>
     </ButtonToolbar>
 </StickyBar>
+
+<style>
+    :global(.btn-toolbar) {
+        overflow-x: hidden;
+    }
+</style>
```

Without it, the starting viewport is too big and it wraps anyway:

Screen Shot 2021-05-28 at 5 25 16 pm
Screen Shot 2021-05-28 at 5 25 06 pm

dae:
Hmm, the dropdown next to Save no longer seems to work for me either

hgiesel:
> Hmm, the dropdown next to Save no longer seems to work for me either

Cannot recreate this.

dae:
Hmm, seems to work fine now. Thanks :-)

PR #1206, opened 2021-05-28
Fix some sidebar issues

RumovZ:
Is it intentional that the `rename_tag()` op returns success in the case of an unused tag which can't be renamed?
I ended up copying the success code to catch this case.


dae:
Please feel free to turn it into an error if you think that'd be preferable

PR #1207, opened 2021-05-29
Deck Options refactoring

hgiesel:
This is a WIP about refactoring the deckoptions.

Some points which should be improved by this:
- (?) Should look better on wider viewports
- [x] Have components dynamically adapt to wider viewports
- [x] Extension API

Some points of interest:
- I rewrote the layout using [Bootstrap Grid system](https://getbootstrap.com/docs/5.0/layout/grid/) (which uses CSS Flexbox, not CSS Grid, despite its name) with its `.container`, `.row`, and `.col` classes.
- I factored out `HelpPopup`, the label handling, and `RevertButton` from the individual components. Making it obvious, that these are individual components, and can stand on their own. Of course this creates a lot of verbose code like this:

```jsx
<Row>
    <Col size={7}>
        {tr.schedulingMaximumInterval()}
        <HelpPopup html={marked(tr.deckConfigMaximumIntervalTooltip())} />
    </Col>
    <Col size={5}>
        <SpinBox
            min={1}
            max={365 * 100}
            bind:value={$config.maximumReviewInterval}
        />
        <RevertButton
            defaultValue={defaults.maximumReviewInterval}
            bind:value={$config.maximumReviewInterval}
        />
    </Col>
</Row>
```

I could wrap these up into their own respective components, let's say `SpinBoxRow`, `CheckBoxRow`, etc., as to avoid boilerplate.

- I'm not sure yet about the CSS I removed from `deckoptions-base.scss`, will have to check after #1202 is merged. (EDIT: was tested on mobile)


dae:
Re "look better", I'm not sure that's the case at the moment? Compared to what we had before, the label and the input box are spaced further apart, making it harder to see which label corresponds to which input box. Maybe instead of a 1fr 1fr layout after a certain size, it would look better as 1fr 2fr or similar? The maximum width is probably too large as well.

Screen Shot 2021-05-29 at 10 54 02 am
Screen Shot 2021-05-29 at 10 53 35 am

I think we'll want components like SpinBoxRow as well, as it's both easier to add new options that way, and will ensure consistency.

dae:
I also feel like the large text boxes feel a bit ugly - what I was going for with the original design was boxes that aren't too much bigger than the text they'll contain. But I'm not a designer :-) Maybe it would be worth getting feedback on the forums as well.

hgiesel:
Hm, I agree that being spaced too far apart is ugly, but I'd argue that the dead space on the left and right are worse :)
The dead space makes it feel like, the options window should actually not be resizable, because it doesn't adapt to the window size. 

There are some options to better accomodate the wider space. One would be to use have a [scrollspy](https://getbootstrap.com/docs/5.0/components/scrollspy/), which appears on the left as a navigation element at a certain width, and as a positive side-effect restricts the width of `ConfigEditor`.

We could also frame the options in a way, which makes it easier to associate them, below's an example of a Chrome preferences, which do this

Screenshot 2021-05-29 at 11 59 20

dae:
I've given the code changes a quick review, and they look good. Regarding the design changes being discussed on the forums, we'll presumably want to merge this first in either case, due to all the other changes as well. If the consensus is for a design like #1212, will it be easy to adjust this to that design?

Some other issues I noticed:

- the revert button is difficult to see in day mode
- the left and right margins are too big on mobile now. Before:

IMG_1347

after:


IMG_1346

hgiesel:
> the left and right margins are too big on mobile now.

Which margins are you referring to exactly? The left margin of the main content seems to be the same to me. The right margin is bigger, because it leaves space for the revert button. I made the horizontal padding of the Revert Icon a bit smaller:

This is what it looks like on my phone now:

image

Which pointed me to another issue, when labels wrap. :)

>  If the consensus is for a design like #1212, will it be easy to adjust this to that design?

Actually no. With Boostrap Grid being built upon Flexbox, it cannot behave in such a way. But maybe it's for the better to move away from Bootstrap Grid. It was written before CSS Grid existed, and I read multiple blog posts about people moving away from it in favor of CSS Grid. But we might still be able to use some features of it, like the breakpoints. I'll give it a shot.

dae:
Hmm, @kleinerpirat seemed to think it would be a simple change?

The remaining margin issue is in the selector at the top - the former version extended the selector and buttons closer to the edges.

dae:
And the right margins are still different - the revert button previously closer to the edges (sort of like being in the "gutter")

hgiesel:
> And the right margins are still different - the revert button previously closer to the edges (sort of like being in the "gutter")

This is something I intentionally avoided, as the screen edges are also used for system functions on iOS.

One Suggestion: Reverse input and revert position
image

dae:
I tried that in the past and found it looked a bit crowded, especially when the description text comes close to the input box. It can make it harder to hit the right button in that case as well.

Not sure what you mean by "system functions"? Swipes yes, but this is just a single tap, and the location is not far off the right arrow disclosure indicator shown in things like Settings.app.

dae:
(I do like how the warning is correctly balanced in that case though)

hgiesel:
What do you think about putting the tooltip on the label instead, and possibly indicating with an underline. A dashed underline indicating a tooltip is also a common way to indicate a tooltip.

image

dae:
It's kind of ugly isn't it? :-) Especially when the descriptions are longer. I thought about doing something only on hover, but then it's not discoverable on mobile devices. IIRC I also ran into difficulty with popper putting the popup arrow in the wrong place when the entire text was the trigger.

I do like how it makes things neater (the info icons are a bit ugly too), but the conclusion I came to when originally implementing this was that they were the least bad option.

dae:
(Maybe if the indicator were more subtle it wouldn't be as bad?)

hgiesel:
I also agree not a piece of art, but thought it was on the same level as having two icons per row :). Regarding making it more beautiful, we could decrease the opacity:

image

I don't see any other way right now.  (Ignore the revert icons having switched the position again :) )

dae:
That's not too bad, and am happy to give it a try in an alpha to see what people think. I still think it would be nice if the top area collapsed the margins when the screen size is very small so that more of the deck config can fit on the screen - it's a navigation area so I don't think it needs to follow the same indent as the rest of the page.

dae:
Hmm, one other option would be to collapse the margins in the main area as well when the screen is too narrow. On small mobile devices we don't want much of a margin as horizontal space is at a premium, and the left margin was only there to match the right margin, which was only there to allow for the revert icon. So if the revert icon is being moved to the left, we should be able to get away with just a very small margin on left and right, that expands out once a reasonable width has been reached.

Presumably having the main area behave the same as the top area would make things easier too?

kleinerpirat:
> Hmm, @kleinerpirat seemed to think it would be a simple change?

I agree with Henrik that a grid would be more complicated, but I believe you are actually referring to the two-column-layout I [showed on the forum](https://forums.ankiweb.net/t/anki-2-1-45-alpha/10061/74?u=kleinerpirat), which would indeed be super easy to implement, as it doesn't use grid at all:

image

I actually built that demo on this PR.
All I did for that was to wrap half of the `Items` in a `<div class="left">`, the other half in `<div class="right">` and added three CSS rules beneath:
```
    @media (min-width: 992px) {
        .left {
            width: 48%;
            float: left;
            position: relative;
        }
        .right {
            width: 48%;
            float: right;
        }
        .left::after {
            content: "";
            position: absolute;
            border-left: 1px solid var(--medium-border);
            height: 95%;
            top: calc(24px + 1.5em);
            right: -4%;
        }
    }
```

---

Regarding the revert buttons: You could hide most of their ugliness by only showing the one whose corresponding input is focused. Or you delete them alltogether (which is what I would do) and show a clickable tooltip "Revert to default?" when the input is empty || all of its text is marked.

hgiesel:
> All I did for that was to wrap half of the Items in a <div class="left">, the other half in <div class="right"> and added three CSS rules beneath:

I'd want to avoid having to split the sections into left/right, as it is supposed to be extensible (and the API for it already exists, it's the same as for extending the buttons in the editor).

Moving the revert button into the input sounds fine with me, even though we need to make sure they are still easy to hit, when on mobile. The incrementor/step buttons are just a browser default, and they [could be hidden](https://www.w3schools.com/howto/howto_css_hide_arrow_number.asp).

kleinerpirat:
> I'd want to avoid having to split the sections into left/right, as it is supposed to be extensible

Ah I understand.

Perhaps there is a dynamic way to do that splitting.

With grid, one would need to set a max-height to tell it when to wrap to the next column.
That's certainly a bit problematic. Hm..

> Moving the revert button into the input sounds fine with me, even though we need to make sure they are still easy to hit, when on mobile.

I think this would be a smart move. I didn't notice any issues on mobile with my codepen.


With that change, there would be enough room for the info buttons again. Although the underlines are intuitive, they make the text harder to read and look a bit off, because usually this style is only used on single-word definitions.

hgiesel:
Hm, however where to put the revert button for the checkboxes then... The checkbox does not need to focused / active, to change its value...

kleinerpirat:
Oh yes. Didn't think about that...
Also, for `<select>` elements this isn't that straightforward either.

> I'll do some experiments...

Curious what you come up with. But really, reveal on hover is not that essential, right?
As long as the icons are within the `<input>` elements, we save a lot of space already - and it looks tidier. Checkboxes don't take a lot of room anyway. It's just the `<select>`'s that bother me a bit... Where would you put the revert button on those? To the left of the arrow? Or to the right of the whole thing?

---
Sadly I got a big exam coming up in three weeks, so I don't have much time to experiment on this one.

hgiesel:
So, the gutters cannot be entirely avoided, as they're built into Bootstrap Grid. If we're going to abandon Bootstrap Grid in favor of CSS Grid, this will be fixed anyway. However I'd like to do that in a new PR. Things I want to tackle:

- Use CSS Grid in favor of Bootstrap Grid. I might need one or two attempts to get it right, and I do not want to muddy the PR.
- Make "Revert tooltip" use Bootstrap Tooltip
- Maybe we could expose the Revert button as a an `IconButton` on the left side of the number input, or make it absolute positioned on the left edge of the number input. Not sure which approach is better here.

dae:
Some more thoughts now that I've tried the latest code:

One of the goals of the revert button was to show which options have been changed from the default, so that it's immediately obvious which options have been changed. This worked best when they were shown on the right, as you can scroll down the page and the indicators are all in the one column. Now we've got the revert button in the far right for checkboxes, 2/3rd for input boxes, and very left for dropdowns that take up the full screen width.

Embedding the revert icon in the right side of the input box would look nice for text inputs, but the dropdowns and selects are going to be problematic as you both have mentioned. I can't help but feel keeping them on the right might be the least-bad option.

When viewed as a whole screen, the underlining feels a bit harder to read than I was expecting, and I find myself preferring the old info icons :-(. For some reason the underline is not appearing on mobile, which looks better, but we lose the discoverability.

Re being unable to remove the outer margins, something like https://getbootstrap.com/docs/5.0/layout/gutters/#no-gutters doesn't help?

Re two columns, I had assumed they were implemented using an outer grid. I wonder if outer grid + each section wrapped in div + special border styling might be able to achieve a similar result?

kleinerpirat:
>  I wonder if outer grid + each section wrapped in div + special border styling might be able to achieve a similar result?

I think I found a solution. See [this CodePen](https://codepen.io/kleinerpirat/pen/zYZjyKX?editors=1111) I just made.

---
At first I thought it would be impossible because of the following issue: How would Grid now when to wrap to the next column without a fixed height? And with a fixed height, there would be a limit to the amount of `Items` that can be displayed before the grid overflows to the right.

**But while experimenting I stumbled upon the following:**

When the rows are explicitly stated, the two-column-layout will work as expected:
```
.grid {
    grid-template-rows: auto auto auto auto auto;
}
```
That definition can be more elegantly expressed with repeat():
```
.grid {
    grid-template-rows: repeat(5, auto);
}
```

Now we just need to calculate how many rows we need (see CodePen) and use a CSS variable:
```
.grid {
    grid-template-rows: repeat(var(--rows), auto);
}
```
---
There's **only one requirement** that needs to be added to reduce empty spaces:
- the amount of rows each `Item` will span (1, 2, 3, ...) via CSS variable

dae:
@hgiesel In terms of moving ahead with this PR, if you think the multiple columns are best left until bootstrap→native grid conversion, that's fine - don't want wasted effort if possible.

hgiesel:
Unfortunately I won't have my development machine for the next 1-3 days. I'm trying to setup an development environment on a spare PC, but it proved more complicated than anticipated...

dae:
No rush, a stable release is still some time away

hgiesel:
I just realized, that Bootstrap 5 actually supports [switches](https://getbootstrap.com/docs/5.0/forms/checks-radios/#switches). 

If we layout them like the Chrome preferences, we could also have a more balanced design, by having all labels left, and input elements right:

Screenshot 2021-06-12 at 01 25 15

What do you think about using them? Either for both platforms, or just on mobile?

dae:
Happy to give it a try - it's nice that it moves the interaction to the right like the other elements. Presumably clicking on the text will still select/deselect the switch?

hgiesel:
> Presumably clicking on the text will still select/deselect the switch?

Actually preferably not. Both the Chrome preferences, and the iOS system settings have switches, but they don't show this behavior. Also it would allow us to activate the tooltip upon clicking the label, not just the small info icon, which I find more natural. Before, putting the tooltip on the label, would mean that on mobile, you couldn't activate the tooltip without triggering the checkbox.

ts/deckoptions/DisplayOrder.svelte
```diff
@@ -35,44 +39,59 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+        <EnumSelectorRow
+            bind:value={$config.reviewOrder}
+            defaultValue={defaults.reviewOrder}
+            choices={reviewOrderChoices}
+            markdownTooltip={tr.deckConfigReviewSortOrderTooltip()}
+        >
+            {tr.deckConfigReviewSortOrder()}
+        </EnumSelectorRow>
+    </Item>
+</TitledContainer>>
```
kleinerpirat, line 97: Duplicate > sign after TitledContainer.

ts/deckoptions/deckoptions-base.scss
```diff
@@ -28,16 +28,18 @@
...
-    max-width: 300px;
     text-align: left;
+    max-width: 300px;
 
-    p {
-        margin: 0.8em 0.4em 0.8em 0.8em;
+    // marked transpiles tooltips into multiple paragraphs
+    // where trailing <p>s cause a bottom margin
+    > p:last-child {
+        display: inline;
```
kleinerpirat, line 37: That's better, but the tooltips as a whole still need some padding in my opinion.
Looks a bit packed currently:

Screenshot from 2021-06-12 10-21-29

hgiesel: There's also another issue with no padding applied I noticed:

image

ts/deckoptions/DisplayOrder.svelte
hgiesel: I wonder why that is not detected by the formatters? Or what it is parsed as?

kleinerpirat: Seems to be treated like the input of Anki's HTML editor or card templates, just as a literal character:

image

Found it because I had 2021 scheduler enabled.

dae:
I don't mind the switches themselves - they're not quite as space-efficient as the checkboxes, but they are more consistently positioned, so I guess either could work.

The undo buttons are too close to the input elements, so the user can end up accidentally tapping the wrong thing - a little bit of padding on the left of them would improve things. It doesn't matter so much in the switch case, as a tap on either will result in the same change, but for the input boxes the spacing is more important.

The gutters are still too big, and cause the text to wrap more than necessary. Something like the following makes it
look better on the right hand side when a revert icon is shown, but the left side ends up looking unbalanced.

```patch
diff --git a/ts/components/Container.svelte b/ts/components/Container.svelte
index e27da6103..4e00e518b 100644
--- a/ts/components/Container.svelte
+++ b/ts/components/Container.svelte
@@ -12,7 +12,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     export let api: Record<string, never> | undefined = undefined;
 </script>
 
-<div {id} class={`container mb-1 ${className}`}>
+<div {id} class={`container mb-1 g-1 ${className}`}>
     <Section {api}>
         <slot />
     </Section>
```

> Actually preferably not. Both the Chrome preferences, and the iOS system settings have switches, but they don't show this behavior. Also it would allow us to activate the tooltip upon clicking the label, not just the small info icon, which I find more natural

If I open Chrome's preferences and click on the text of "show home button", it does toggle the checkbox for me.  iOS switches do not activate on single tap, but when the user enables VoiceOver, a double-tap on the text toggles the switch. A lot of docs on the web stress the importance of having labels associated with controls for accessibility reasons, so I'm not sure this is something we should be deviating from.

hgiesel:
> I don't mind the switches themselves - they're not quite as space-efficient as the checkboxes.
This is somehwhat intended. I tried to make them as high as the other inputs, so the rows are more consistently spaced. This is what they look like at their default size:

image

They'd also have that size on mobile, which is another reason, why I changed the base size.

hgiesel:
I'm sorry, but I still can't justify it to myself to put the revert button on the right edge.
1. The rightmost action is the one closest to the thumb on a mobile device. The risk of mistapping it is very high imo. There's not even currently a warning currently, it's just reset. Having the settings for a longer amount of time makes me question myself: Might I have accidently tapped a Reset button and just not noticed it .... Especially because the right edge of the screen is where I typically position my thumb to scroll down, while still being able the content of the page.
2. This one I can only base on my own experience: Is it really that important to always have at first sight, which settings are set at default, and which one's deviate?
3. I don't think to put the important information in the tooltip is a good strategy. Tooltips are not very mobile friendly, as I've read it in many blogs. I wanted to avoid having the tooltip only on the small info icon on mobile, so instead I put it on the entire label. But like you pointed out, it's also a good idea to associate the label with the input. And on mobile, we only can do one of those things, as we don't have `hover`.

The one settings I've found, which I think resembles our scenario the best is the Chrome settings (again) under chrome://settings/cookies:

Screenshot 2021-06-17 at 00 18 13

This collapsible section could include what is currently the tooltip, and could also include the Reset button. We could slide it in similar to the warnings with Sveltes `in:slide`.

dae:
Yes, the back and forth can be a bit tiring :-) I realise there is still room for improvement on this screen, but there are a lot of other things that need doing too :-)

The revert icons are partly modeled after the color-based ones in VS Code:

Screen Shot 2021-06-17 at 10 29 01 pm

Reverting there is a two-step process - tap on the gear then tap on reset setting. Would such a popup or a confirmation dialog alleviate your concerns? I'm not sure how likely an accidental tap will be, and without tapping on Save nothing will be changed, but an extra confirmation step wouldn't do much harm.

I experimented with tips directly next to each option when first reimplementing this screen, like in the screenshot above. The disadvantage of doing it that way is that the page becomes long, and it can be slower to find the option you're looking to change, which more advanced users may find annoying.

dae:
Re mergeable state, the only two blockers for me are:

- labels associated with the controls instead of tooltips (and restoring the old info icon handling should also improve the source arrow, which is a bit all over the place at the moment)
- a revert button layout/handling we can both agree on :-)

hgiesel:
I hopefully addressed the issues. The revert button now has a gear icon and is positioned on the left.

Screenshot 2021-06-21 at 22 43 04

However I also seem to have discovered a bug in svelte-check. It says `createTooltip is not used / is not defined`, which is not the case.

dae:
Thanks Henrik. I still have some niggles about the revert button, but I don't want to hold this up any more, so I'll merge it in as-is and open up a new PR for discussion.

PR #1208, opened 2021-05-29
Detect MathJax and Wikipedia equations when pasting text

Upabjojr:
This PR adds an idea to transform equations from Wikipedia and MathJax HTML elements into Anki's `\[ ... \]` and `\( ... \)` formats.

- Wikipedia equations are straightforward, I lookup the class name for equations and extract the LaTeX string from the image attribute.
- MathJax contains the MathML expression, which needs to be transformed into LaTeX. For this purpose, I have added a downloader of [these files](https://github.com/oerpub/mathconverter/tree/master/xsl_yarosh), (under [MIT license](https://github.com/oerpub/mathconverter/blob/master/LICENSE)), which are XSL transformation rules to convert MathML XML into LaTeX.

### DIFF
--- a/pip/requirements.txt
+++ b/pip/requirements.txt
@@ -156,3 +156,5 @@ pip==21.1.1
     # via pip-tools
 setuptools==56.1.0
     # via jsonschema
+
+lxml
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -892,6 +892,53 @@ $editorToolbar.then(({{ toolbar }}) => toolbar.appendGroup({{
 
     removeTags = ["script", "iframe", "object", "style"]
 
+    def _parse_math_from_mathjax(self, doc):
+        from lxml import etree
+        mathjax_elements = doc.find_all(None, {"class": "MathJax"})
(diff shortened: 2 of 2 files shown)


dae:
This might be better implemented as an add-on - it's serving a fairly niche audience, and the fetching of files at runtime feels like it could be fragile. Appreciate you taking the time to contribute though. :-)

Upabjojr:
I can understand that parsing MathML back to LaTeX is a bit of a complicated operation. But what about Wikipedia equations? Wikipedia renders LaTeX equations as images but provides the original LaTeX expression as a tag of the `` elements. That's a fairly simple extraction.

dae:
It's still extra code that needs to be maintained, and every change carries with it the risk of an unexpected regression. Given that I don't recall anyone ever asking for this feature before, I'm afraid I think it would probably be better done in an add-on for now.

dae:
Native modules are more complicated; you would need to bundle a different version for each platform you want to support: https://addon-docs.ankiweb.net/#/more?id=standard-modules

PR #1209, opened 2021-05-30
Drag & drop for saved search

RumovZ:
This allows dragging a searchable item onto the saved searches area to make it a new saved search (basically pinning the item to the top).

We lost the ability to focus not yet existing items after refreshing with 1ece868d029bb7ed360b07df5434d915957e63b4. I've worked around that by passing a mock item that can be used for finding the real new item.


dae:
I found myself wanting that the other day - thank you!

PR #1211, opened 2021-05-30
Implement new text color / highlight color buttons

hgiesel:
I've made some CSS changes to `IconButton`, which allows for better control:
1. `iconSize` allows how much of the available width/height of a button an icon uses.
2. `widthMultiplier` allows for changing the icon button width from its original square design

Screenshot 2021-05-31 at 00 34 38

What do you think about instead of using the terms "foreground color" and "background color" for those buttons, using "text color" and "highlight color"? I think those terms might better describe the behavior for users, and better suit their icons as well.


dae:
Thanks Henrik! Word calls it "text highlight color", maybe that would be clearer? It also uses "font color" instead of "text color", though I'm not sure if that's preferable

PR #1212, opened 2021-05-31
Apply grid layout to deck options 2

kleinerpirat:
Closes https://github.com/ankitects/anki/pull/1210.

This is a less intrusive version of my previous PR. I have split `GeneralOptions.svelte` into
- `TimerOptions.svelte`
- `AudioOptions.svelte`

for more consistency.

The grid layout is now better encapsulated in `ConfigEditor.svelte`.

---

### Other changes
- reduce `<h2>` font size to 1.5em
- adjust padding of `#main` and `.outer`

### Showcase

image

---
*I would highly appreciate if someone could tell me how to run Prettier on checkout via VS-Code.*

### DIFF
--- a/ts/deckoptions/AudioOptions.svelte
+++ b/ts/deckoptions/AudioOptions.svelte
@@ -4,7 +4,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 -->
 <script lang="ts">
     import * as tr from "lib/i18n";
-    import SpinBox from "./SpinBox.svelte";
     import CheckBox from "./CheckBox.svelte";
     import type { DeckOptionsState } from "./lib";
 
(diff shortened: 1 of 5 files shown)


hgiesel:
> I would highly appreciate if someone could tell me how to run Prettier on checkout via VS-Code.

`bazel run //ts:format`

kleinerpirat:
Thanks, Henrik!

I thought about using a Svelte prop `rows` instead of the two classes `small` and `large`, but that would require changes within the individual option components (similar to the previous PR).

Is this https://github.com/ankitects/anki/commit/eb03d8ec74170b4b32f1f1d1d9def1c6d3aeb85e a viable option or would you prefer a more dynamic or completely different approach? I don't mind if anyone (looking at you @hgiesel) creates an alternative since I'm just starting out here.

Just curious what you think about the grid layout in general.

hgiesel:
First of all, I think there's some duplicated effort here with #1207, e.g. splitting up `TimerOptions` and `AudioOptions`.

In this PR, I also made another suggestion regarding the issue of wider windows, which was to have a scrollspy / navigation element to the left of the main content, taking inspiration from other configuration menus, like the Chrome preferences.

Having the content collapse into columns is a good way to prevent overly wide input elements, however you still have to scan the entire page to find the appropriate section.

In the end, it will be up to @dae to make a judgement :)

kleinerpirat:
Ah yes, glad you pointed that out. I wasn't aware of your refactoring efforts when I worked on these changes. I'm happy to postpone design contributions such as this one until the foundations are more solid / reworked.

The last thing I want is for you guys to have more work because of me - that would be the opposite of what's intended ;)

dae:
Thanks Matthias :-)

PR #1213, opened 2021-05-31
Add pink, turquoise and purple flags

RumovZ:
I think the hues I've picked are the most distinct from the existing flag hues available, though the colours could certainly be improved. This can always be done later, but the names should be definite before a merge and I'm not quite sure about them. Is what I've called violet maybe more of a pink? And then maybe purple should be named violet instead?

Did you have any reference points when you added the other flag colours? I've tried to pick one hue and only adjust saturation and value for the different modes and roles, but that does not seem to have been the approach for the existing flag colours.


dae:
Pink and purple are a bit more obvious than violet IMHO. Turquoise is also a little less obvious to me, but I'm struggling to come up with a better alternative that hasn't already been used (eg yellow).

If I recall correctly, the existing color shades were just picked via experimentation rather than with a specific method.

RumovZ:
> If I recall correctly, the existing color shades were just picked via experimentation rather than with a specific method.

Then I'll try to tweak them as well to improve differentiation. A pink flag should be possible if flag 1 is shifted a bit.

RumovZ:
In reviewer.ts, I've just used the day mode foreground colours. Since the background these flags appear on is completely up to the user, I wouldn't know what to adjust them to.

ftl/core/search.ftl
```diff
@@ -16,7 +16,7 @@ search-unclosed-quote = an opening double quote `"` was found, but there was no
 search-missing-key = a colon `:` was found, but there was no keyword preceding it. If you want to search for a literal `:`, prepend a backslash: `\:`.
 search-unknown-escape = the escape sequence `{ $val }` is not defined. If you want to search for a literal backslash `\`, prepend another one: `\\`.
 search-invalid-argument = `{ $term }` was given an invalid argument '`{ $argument }`'.
-search-invalid-flag = `flag:` must be followed by a valid flag number: `1` (red), `2` (orange), `3` (green), `4` (blue) or `0` (no flag).
+search-invalid-flag = `flag:` must be followed by a valid flag number: `1` (red), `2` (orange), `3` (green), `4` (blue), `5` (pink), `6` (turquoise), `7` (purple) or `0` (no flag).
```
dae, line 19: we'll need to create a new string key for this, as AFAIK simply changing the text will not notify translators that the string needs changing. search-invalid-flag-2 or similar would be fine

dae (commented):
Looks good, thanks Rumo! Just one comment below.

ftl/core/search.ftl
RumovZ: Isn't there a way to mark a string as updated? I think this is a good example where it would make more sense to fall back on an outdated translation than on another language.

RumovZ: Pity, but thanks for explaining.

PR #1218, opened 2021-06-03
fix documentation links

cherryblossom000:
Fix help pages and other links so that they use URL fragments (`#foo`) instead of `?id=foo`. This also removes the `#/` before the path of these URLs.

For example:

- before: https://docs.ankiweb.net/#/getting-started?id=note-types (goes to home introduction page)
- after: https://docs.ankiweb.net/getting-started#note-types (correctly goes to getting started page and the note types section)


dae:
Thank you for the fixes!

PR #1221, opened 2021-06-07
Audio pause fix

### LINKED ISSUE #1164: Audio is paused unusual
When I am playing a audio file with anki-2.1.37 (windows), I press pause key almost at the end of the file, then it is paused.
But it can't be replayed when I press the replay key, then I have to press the pause key next and it starts to replay, It 's really weird！

I downgrade to version 2.0.52, using a Adjust_Audio_Speed plugin, it works well.  The code of new anki version which control the Mpv Player must have some bugs. Is anybody can fix that in a new version?

kelciour:
An attempt to fix #1164 and update toggle pause with 'cycle' command.

The issue could be reproduced by repeatedly pressing `5` to pause/unpause the player until the end of the audio to try to set the player to be in paused state when idle and before playing the next audio file.

It was fixed by adding 'pause=no' option to `loadfile` command instead of using `--reset-on-next-file=pause` command line option.

https://mpv.io/manual/master/#command-interface-[%3Coptions%3E]]

It appears both of these options might not work properly on recent mpv builds (20200830-git-bb1f821 seems to be good and 20200906-git-f57b90b seems to be bad), the audio file might be cut short and the last second of the audio won't be played, if the player is in paused state, but adding `--gapless-audio=no` seems to help and everything seems to work fine.

### DIFF
--- a/qt/aqt/mpv.py
+++ b/qt/aqt/mpv.py
@@ -89,7 +89,6 @@ class MPVBase:
         "--ontop",
         "--audio-display=no",
         "--keep-open=no",
-        "--reset-on-next-file=pause",
         "--autoload-files=no",
     ]
 
--- a/qt/aqt/mpv.py
+++ b/qt/aqt/mpv.py
@@ -90,6 +90,7 @@ class MPVBase:
         "--audio-display=no",
         "--keep-open=no",
         "--autoload-files=no",
+        "--gapless-audio=no",
     ]
 
     def __init__(self, window_id=None, debug=False):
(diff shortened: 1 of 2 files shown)


dae:
Thanks as always for your work in this area, kelciour. Do you know if those new options will work ok on Linux distros with an older mpv?

kelciour:
Yes, it should be fine.

`cycle` and `loadfile with [options]` appear to be present at least since v0.7.0 (https://github.com/mpv-player/mpv/commit/0f70d0e927e7361120a4f2b6d612c53380cd911f) and `gapless-audio` since v0.4.0 (https://github.com/mpv-player/mpv/commit/3e2f16a89e625d9798217f39cc5be6211059acf9).

dae:
Great, thanks again :-)

PR #1222, opened 2021-06-07
Change handling of browser's card parameter

RumovZ:
Should I preserve `show_single_card()` for legacy reasons?

### DIFF
--- a/qt/aqt/browser/table/table.py
+++ b/qt/aqt/browser/table/table.py
@@ -129,6 +129,8 @@ class Table:
         self.clear_selection()
         if (row := self._model.get_card_row(card_id)) is not None:
             self._view.selectRow(row)
+            # editor may pop up and hide the row later on
+            QTimer.singleShot(100, lambda: self._scroll_to_row(row))
 
     # Reset
 
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -80,8 +80,9 @@ class Browser(QMainWindow):
         search: Optional[Tuple[Union[str, SearchNode]]] = None,
     ) -> None:
         """
-        card  : try to search for its note and select it
-        search: set and perform search; caller must ensure validity
+        card -- try to select the provided card after executing "search" or
+                "deck:current" (if "search" was None)
+        search -- set and perform search; caller must ensure validity
         """
 
         QMainWindow.__init__(self, None, Qt.Window)
@@ -241,9 +242,12 @@ class Browser(QMainWindow):
         if search is not None:
             self.search_for_terms(*search)
             self.form.searchEdit.setFocus()
-        elif card:
-            self.show_single_card(card)
-            self.form.searchEdit.setFocus()
+        if card is not None:
+            if search is None:
+                # implicitly assume 'card' is in the current deck
+                self._default_search(card)
+                self.form.searchEdit.setFocus()
+            self.table.select_single_card(card.id)
 
     # Searching
     ######################################################################
@@ -261,13 +265,11 @@ class Browser(QMainWindow):
         self.form.searchEdit.addItems(self.mw.pm.profile["searchHistory"])
         if search is not None:
             self.search_for_terms(*search)
-        elif card:
-            self.show_single_card(card)
         else:
-            self.search_for(
-                self.col.build_search_string(SearchNode(deck="current")), ""
-            )
+            self._default_search(card)
         self.form.searchEdit.setFocus()
+        if card:
+            self.table.select_single_card(card.id)
 
     # search triggered by user
     @ensure_editor_saved
@@ -330,17 +332,11 @@ class Browser(QMainWindow):
         self.form.searchEdit.setEditText(search)
         self.onSearchActivated()
 
-    def show_single_card(self, card: Card) -> None:
-        if card.nid:
-
-            def on_show_single_card() -> None:
-                self.card = card
-                search = self.col.build_search_string(SearchNode(nid=card.nid))
-                search = gui_hooks.default_search(search, card)
-                self.search_for(search, "")
-                self.table.select_single_card(card.id)
-
-            self.editor.call_after_note_saved(on_show_single_card)
+    def _default_search(self, card: Optional[Card] = None) -> None:
+        search = self.col.build_search_string(SearchNode(deck="current"))
+        if card is not None:
+            search = gui_hooks.default_search(search, card)
+        self.search_for(search, "")
 
     def onReset(self) -> None:
         self.sidebar.refresh()


dae:
Thanks for the quick fix! The show_single_card() functionality was only added recently, so hopefully removing it won't affect any add-ons.

qt/aqt/browser/table/table.py
```diff
@@ -129,6 +129,8 @@ def select_single_card(self, card_id: CardId) -> None:
         self.clear_selection()
         if (row := self._model.get_card_row(card_id)) is not None:
             self._view.selectRow(row)
+            # editor may pop up and hide the row later on
+            QTimer.singleShot(100, lambda: self._scroll_to_row(row))
```
dae, line 134: The interface would look a bit more polished if we could open to the correct scroll location the first time. Sometimes these delay hacks are unavoidable, but I wonder if that's the case here? The following changes prevent the redraw for me - would you mind having a play with them to see if they cause any regressions?

```diff
diff --git a/qt/aqt/browser/browser.py b/qt/aqt/browser/browser.py
index 194c60def..52cdc7b07 100644
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -106,9 +106,9 @@ class Browser(QMainWindow):
         self.setupEditor()
         # disable undo/redo
         self.on_undo_state_change(mw.undo_actions_info())
-        self.setupSearch(card, search)
         gui_hooks.browser_will_show(self)
         self.show()
+        self.setupSearch(card, search)
 
     def on_operation_did_execute(
         self, changes: OpChanges, handler: Optional[object]
diff --git a/qt/aqt/browser/table/table.py b/qt/aqt/browser/table/table.py
index 1cecd8086..67f2a0451 100644
--- a/qt/aqt/browser/table/table.py
+++ b/qt/aqt/browser/table/table.py
@@ -129,8 +129,7 @@ class Table:
         self.clear_selection()
         if (row := self._model.get_card_row(card_id)) is not None:
             self._view.selectRow(row)
-            # editor may pop up and hide the row later on
-            QTimer.singleShot(100, lambda: self._scroll_to_row(row))
+            self._scroll_to_row(row)
 
     # Reset
 
@@ -408,8 +407,7 @@ class Table:
             current = current or rows[0]
             self._select_rows(rows)
             self._set_current(current)
-            # editor may pop up and hide the row later on
-            QTimer.singleShot(100, lambda: self._scroll_to_row(current))
+            self._scroll_to_row(current)
         if self.len_selection() == 0:
             # no row change will fire
             self.browser.onRowChanged(QItemSelection(), QItemSelection())
```

dae: I'll merge this in for now as it's about time to get another build out, but if you spot any problems, please let me know or feel free to revert it in a follow-up PR.

RumovZ: The problem the timer is supposed to solve only seems to occur when the browser is already open, so this commit doesn't help I'm afraid. To reproduce:
1. Open the browser while reviewing.
2. Scroll up until the selected row is hidden.
3. Select multiple rows to hide the editor.
4. Repeat step 1.

The editor should now conceal the selected row.

But I'm in favour of removing the timer. It can't be trusted to work in all situations, it makes debugging much harder (as I had to learn the hard way), and it gets worse exponentially if there's more than one.
I'll investigate if there's another way.

dae: One indirect way to "solve" it might be to change to PositionAtTop instead of PositionAtCenter :-)

RumovZ: Does this work on your machine? I had tried it and it didn't make any difference. The current `PositionAtCenter` hint doesn't seem to be obeyed, either.

dae: This seems to work for me when applied to latest HEAD:

```difff
diff --git a/qt/aqt/browser/table/table.py b/qt/aqt/browser/table/table.py
index 437277ed2..157bdc0e6 100644
--- a/qt/aqt/browser/table/table.py
+++ b/qt/aqt/browser/table/table.py
@@ -458,7 +458,7 @@ class Table:
         visible = top_border >= 0 and bottom_border < self._view.viewport().height()
         if not visible or scroll_even_if_visible:
             horizontal = self._view.horizontalScrollBar().value()
-            self._view.scrollTo(self._model.index(row, 0), self._view.PositionAtCenter)
+            self._view.scrollTo(self._model.index(row, 0), self._view.PositionAtTop)
             self._view.horizontalScrollBar().setValue(horizontal)
 
     def _scroll_to_column(self, column: int) -> None:
``` 

Following the 4 steps you listed above, it always scrolls to the top

RumovZ: Must be a Windows issue then. Ugh.

dae: Works for me on Windows - maybe I'm doing something different? Before and after pressing b in the main window:

Screen Shot 2021-06-09 at 8 31 51 am
Screen Shot 2021-06-09 at 8 31 59 am

RumovZ: Looks like what I'm doing. Here's a GIF:

3qTRB0GWrO

As always, no add-ons involved. This is with `PositionAtCenter`, but as can be seen, it scrolls until the row appears at the bottom. If I had scrolled down on the other hand, it would scroll until the row appeared at the top. So it appears that the behaviour is always `EnsureVisible`, regardless of the actual scrollHint set.

RumovZ: Okay, I know what happened. `selectRow()` already scrolls the row into view, then `_scroll_to_row()` isn't executed because the row is already visible (before the editor pops up). I hadn't pulled c5e56a5fe81d1c56ea1c415d086cee96c7f75b8b which fixes the issue by scrolling independently of visibility.
Another fix would be to scroll before selecting:

```diff
diff --git a/qt/aqt/browser/table/table.py b/qt/aqt/browser/table/table.py
index 67f2a045..f9fb9e4c 100644
--- a/qt/aqt/browser/table/table.py
+++ b/qt/aqt/browser/table/table.py
@@ -128,8 +128,8 @@ class Table:
         """Try to set the selection to the item corresponding to the given card."""
         self.clear_selection()
         if (row := self._model.get_card_row(card_id)) is not None:
-            self._view.selectRow(row)
             self._scroll_to_row(row)
+            self._view.selectRow(row)

     # Reset
```

However, the issue remains that we scroll before the editor shows. So the row isn't centered in the reduced view, or may even be hidden if the editor covers more than half of the vertical space. But like you said in the beginning, this won't be a problem if we `PositionAtTop` as in your screenshots.

PR #1224, opened 2021-06-10
Keep the internal mpv playlist from growing on

kelciour:
Just a small change, just in case.

Probably, since Anki 2.1.20, `append-file` is no longer necessary to play multiple files sequentially and it started to misbehave a bit. If the audio playback was never interruped, the internal mpv playlist might keep growing on and on.

It could be reproduced by reviewing a few cards with audio by listening to the whole audio file each time and running this code in the debug console a few times.

```
from aqt.sound import mpvManager
length = mpvManager.get_property("playlist-count")
for i in range(length):
  print("Filename:", mpvManager.get_property(f"playlist/{i}/filename"));
print("==================")
print("Length:", length)
print("==================")
```


dae:
Thanks as always :-)

PR #1225, opened 2021-06-10
Fix  Segoe UI got replaced with Arial on Windows

kelciour:
I'm not sure if this is the best fix, but it looks like this line was never used in recent Anki versions, but with https://github.com/ankitects/anki/commit/fdd162a7b9d187cb4a88a372d87a8b08e2162ca0, to move the default font size and family from body to html, it started to take priority and Segoe UI font got replaced with Arial on Windows, and simply deleting it seems to fix the issue.

изображение

### DIFF
--- a/ts/sass/core.scss
+++ b/ts/sass/core.scss
@@ -8,7 +8,6 @@
 }
 
 body {
-    font-family: Helvetica, Arial;
     color: var(--text-fg);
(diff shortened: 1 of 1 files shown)


dae:
That definition probably predates the platform-specific styling we were injecting, so we probably don't need it - let's give it a try without.

PR #1228, opened 2021-06-11
Disable color picker buttons when a field is not focused

### LINKED ISSUE #1220: color selection buttons should probably be disabled
It looks a bit silly when a field is not focused:
Screen Shot 2021-06-07 at 1 00 35 pm

hikaru-y:
Fixes #1220.


dae:
Thanks Hikaru!

PR #1230, opened 2021-06-12
Check for misplaced or missing clozes when adding and in the editor

RumovZ:
I couldn't get mypy to shut up. Do I have to ignore the pb class somewhere else than in .pylintrc?
Some more notes:
- I've made adding a cloze mandatory. (Before you could skip the warning.) It seems more in line with the new strict checks.
- I've decided against reporting the field with the misplaced cloze to keep the pb message simple and preserve backwards compatibility.

### DIFF
--- a/ftl/core/adding.ftl
+++ b/ftl/core/adding.ftl
@@ -6,4 +6,6 @@ adding-history = History
 adding-note-deleted = (Note deleted)
 adding-shortcut = Shortcut: { $val }
 adding-the-first-field-is-empty = The first field is empty.
-adding-you-have-a-cloze-deletion-note = You have a cloze deletion note type but have not made any cloze deletions. Proceed?
+adding-missing-cloze = You have a cloze notetype but have not made any cloze deletions.
+adding-cloze-outside-cloze-notetype = Cloze deletion can only be used on cloze notetypes.
+adding-cloze-outside-cloze-field = Cloze deletion can only be used in fields which use the 'cloze:' filter (see 'Cards...').
--- a/ftl/core/adding.ftl
+++ b/ftl/core/adding.ftl
@@ -6,6 +6,6 @@ adding-history = History
 adding-note-deleted = (Note deleted)
 adding-shortcut = Shortcut: { $val }
 adding-the-first-field-is-empty = The first field is empty.
-adding-missing-cloze = You have a cloze notetype but have not made any cloze deletions.
+adding-you-have-a-cloze-deletion-note = You have a cloze notetype but have not made any cloze deletions. Proceed?
 adding-cloze-outside-cloze-notetype = Cloze deletion can only be used on cloze notetypes.
 adding-cloze-outside-cloze-field = Cloze deletion can only be used in fields which use the 'cloze:' filter. This is typically the first field.
--- a/ftl/core/adding.ftl
+++ b/ftl/core/adding.ftl
@@ -8,4 +8,4 @@ adding-shortcut = Shortcut: { $val }
 adding-the-first-field-is-empty = The first field is empty.
 adding-missing-cloze = You have a cloze notetype but have not made any cloze deletions.
 adding-cloze-outside-cloze-notetype = Cloze deletion can only be used on cloze notetypes.
-adding-cloze-outside-cloze-field = Cloze deletion can only be used in fields which use the 'cloze:' filter (see 'Cards...').
+adding-cloze-outside-cloze-field = Cloze deletion can only be used in fields which use the 'cloze:' filter. This is typically the first field.
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -81,8 +81,7 @@ _html = """
 <div id="dupes" class="is-inactive">
     <a href="#" onclick="pycmd('dupes');return false;">{}</a>
 </div>
-<div id="not-a-cloze-notetype" class="is-inactive">{}</div>
-<div id="not-a-cloze-field" class="is-inactive">{}</div>
+<div id="cloze-hint"></div>
 """
 
 
@@ -131,11 +130,7 @@ class Editor:
 
         # then load page
         self.web.stdHtml(
-            _html.format(
-                tr.editing_show_duplicates(),
-                tr.adding_cloze_outside_cloze_notetype(),
-                tr.adding_cloze_outside_cloze_field(),
-            ),
+            _html.format(tr.editing_show_duplicates()),
             css=[
                 "css/editor.css",
             ],
@@ -515,16 +510,16 @@ $editorToolbar.then(({{ toolbar }}) => toolbar.appendGroup({{
 
     def _update_duplicate_display(self, result: NoteFieldsCheckResult.V) -> None:
         cols = [""] * len(self.note.fields)
-        wrong_notetype = wrong_field = "false"
+        cloze_hint = ""
         if result == NoteFieldsCheckResult.DUPLICATE:
             cols[0] = "dupe"
         elif result == NoteFieldsCheckResult.NOTETYPE_NOT_CLOZE:
-            wrong_notetype = "true"
+            cloze_hint = tr.adding_cloze_outside_cloze_notetype()
         elif result == NoteFieldsCheckResult.FIELD_NOT_CLOZE:
-            wrong_field = "true"
+            cloze_hint = tr.adding_cloze_outside_cloze_field()
 
         self.web.eval(f"setBackgrounds({json.dumps(cols)});")
-        self.web.eval(f"setClozeHints({wrong_notetype}, {wrong_field});")
+        self.web.eval(f"setClozeHint({json.dumps(cloze_hint)});")
 
     def showDupes(self) -> None:
         aqt.dialogs.open(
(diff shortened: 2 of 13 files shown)


hgiesel:
Is it still necessary to warn people when they use `{{c1::...}}` notation in non-Cloze notetypes? The cloze button as well as the cloze shortcuts won't work anymore on non-Cloze notetypes, so  aren't the chances of a user doing it by accident rather slim? If it did happen, I would rather expect that the user did it very much on purpose.

RumovZ:
Thanks for the info, I wasn't aware of that. I still think this is the way to go, though:
1. Is there any scenario where it's necessary to put clozes in a non-cloze notetype?
2. We warn users who already have invalid clozes.
3. Users may add clozes manually because they don't realise why the button is hidden. (That's me testing this PR. 🙃)
4. We also warn about the wrong field, so it's consistent.
5. Since we reject notes with invalid clozes, it's best to warn the user as early as possible.

ftl/core/adding.ftl
```diff
@@ -6,4 +6,6 @@ adding-history = History
 adding-note-deleted = (Note deleted)
 adding-shortcut = Shortcut: { $val }
 adding-the-first-field-is-empty = The first field is empty.
-adding-you-have-a-cloze-deletion-note = You have a cloze deletion note type but have not made any cloze deletions. Proceed?
+adding-missing-cloze = You have a cloze notetype but have not made any cloze deletions.
```
dae, line 9: This should probably be a warning, not an error. Some users want to quickly dump some text into Anki and then go back and edit it later, and they now how no way to proceed.

ftl/core/adding.ftl
```diff
@@ -6,4 +6,6 @@ adding-history = History
...
+adding-cloze-outside-cloze-notetype = Cloze deletion can only be used on cloze notetypes.
+adding-cloze-outside-cloze-field = Cloze deletion can only be used in fields which use the 'cloze:' filter (see 'Cards...').
```
dae, line 11: This is nice, it will hopefully cut down on users adding clozes to the Extra field accidentally. Since the strings in core/ are intended to be used across the different clients though, it would be better if the reference to the desktop "Cards..." could be made more generic, eg something like:

Cloze deletion can only be used in fields which use the 'cloze:' filter. This is typically the first field.

or

Cloze deletion can only be used in fields which use the 'cloze:' filter in the card template.

The former is probably more user-friendly.

qt/aqt/editor.py
```diff
@@ -79,8 +79,10 @@
 _html = """
 <div id="fields"></div>
 <div id="dupes" class="is-inactive">
-    <a href="#" onclick="pycmd('dupes');return false;">%s</a>
+    <a href="#" onclick="pycmd('dupes');return false;">{}</a>
 </div>
+<div id="not-a-cloze-notetype" class="is-inactive">{}</div>
```
dae, line 84: Since the error is an enum and can only represent one value at once, do we gain anything from having a separate div for each warning? Are you intending them to be styled differently?

dae (commented):
.pylintrc won't help with mypy issues :-) Looks like it was just a mypy bug.

ftl/core/adding.ftl
dae: (I seem to recall Anki originally had this as a hard error, and some people pushed back on it, resulting in it being turned into a warning)

dae:
Re warning about clozes,  you make a good point Henrik about it being harder to do accidentally. But is it really something we need to support? Cloze deletion was only ever intended to be used with a cloze notetype, and the code that checks it has already been written.

RumovZ:
Thanks for fixing mypy!
One thing that bugs me a bit, is that the less severe dupe error is concealing the other more critical errors. So if you add a dupe with a misplaced cloze, there will only be the dupe hint. Do you think that's fine or should I do something about it?
Sorry about the redundant merge commit. Do you want me to clean it up?

dae:
Thanks for the fixes!

Presumably we could just change the order of the if/else to put the more critical errors first?

If you could rebase instead, that'll keep the history cleaner. I could squash these commits, but you've gone to the trouble of making nice atomic commits, and it'd be good if we could preserve them.

In case you weren't aware, you can use `git config --global pull.ff only` to make sure a pull doesn't accidentally add a merge commit.

RumovZ:
Thanks, I will use this config!

I could update the `note_fields_check()` to perform the cloze check before the duplicate check, but I'm not sure that's the best way forward. For one thing, no warnings are put out while the first field is empty.
`EMPTY` and `DUPLICATE`, and the cloze error states are each mutually exclusive, so I guess it would be the most natural to return two states instead, like `FirstFieldState` and `ClozeState`. What do you think?

By the way, do add-ons really use this method?

hgiesel:
@dae
> But is it really something we need to support?

I haven't really thought of it as something "we should / need to support", but rather code we could drop from the code base in the vein of the less code, the better. :) Similar to how moving the Preview Button in the Browser window to the Editor meant, that you cannot click Preview anymore without having a card selected, so the Previewer doesn't need to deal with "no card" anymore.

However @RumovZ I don't quite know what you mean by 4. and 5., so it might well be that the warnings make sense, if they could lead to an error later on. 2. is also a good point.

RumovZ:
There are 5 checks:

1. Empty first field
2. Duplicate
3. Extraneous cloze (for non-cloze notetypes)
4. Missing cloze (for cloze notetypes)
5. Cloze in a non-cloze field (for cloze notetypes)

2, 3 and 5 are used to display a warning in the editor. When trying to add a new note, 1, 3 and 5 are errors, 4 is only a warning.

I had gathered that you were only objecting to the *editor warning* regarding *3*. So I argued for consistency. Maybe that was the misunderstanding?

hgiesel:
@RumovZ Thanks for the list. Yes, I would argue to remove both the warning and the following error, for the reasons I mentioned above.

dae:
I agree with you that this probably isn't going to add a lot of value, and reducing the amount of code is a good thing. On the other hand, it's not zero value either - there are users with existing cloze deletions in basic notetypes that will get warned when they try to edit the cards, and they could potentially be importing them from a text file as well. Maybe on balance it's not worth keeping though - what do you think Rumo?

Re add-ons, the hook was added in #470, and there's some discussion there.

> I could update the note_fields_check() to perform the cloze check before the duplicate check, but I'm not sure that's the best way forward. For one thing, no warnings are put out while the first field is empty. EMPTY and DUPLICATE, and the cloze error states are each mutually exclusive, so I guess it would be the most natural to return two states instead, like FirstFieldState and ClozeState. What do you think?

The advantage of the current approach is that it's simpler :-) The frontend can just handle the different enum cases in a single if/switch statement. In practice I don't think we need to worry too much about other warnings not appearing when the front field is empty, as that's typically the first thing that gets filled in Separate enums may model the relationships between the warnings better, but I'm not sure it would really buy us much in this case. If we wanted to show multiple warnings at the same time then it would probably make more sense to return a list of enum values, but that feels a bit like overkill in this case, where typically there would be only one problem.

RumovZ:
This check requires one ftl string, one pb enum variant, one `else if ` clause in Rust and one in Python. In my opinion, it adds enough value to justify this handful of lines of code. 🙂 

> The advantage of the current approach is that it's simpler :-)

It would feel more idiomatic if the enum variants were truly exclusive, but you're right. I'll only switch the precedence of errors then.

andrewsanchez:
Hi, @RumovZ! If I'm reading correctly, these two points:

> * I've made adding a cloze mandatory. (Before you could skip the warning.) It seems more in line with the new strict checks.

and:

> There are 5 checks:
> 
> 1. Empty first field
> 2. Duplicate
> 3. Extraneous cloze (for non-cloze notetypes)
> 4. Missing cloze (for cloze notetypes)
> 5. Cloze in a non-cloze field (for cloze notetypes)
> 
> 2, 3 and 5 are used to display a warning in the editor. When trying to add a new note, 1, 3 and 5 are errors, 4 is only a warning.

seem contradictory...  Or were you only describing the current behavior in the second quote?  Sorry if I'm misreading something!  I'm wondering if you changed your mind about an error vs a warning?  Just wondering because I am very much in favor of _allowing_ the creation of cards _without_ any cloze deletions and not making cloze deletions mandatory as it sounds like it would severely disrupt my workflow.  My workflow is to add notes on the fly (text only, no cloze deletions) and then create the cloze deletions later when I have time.  One thing I like about this workflow is that it allows me to use Anki to _passively_ review information until I have the time to create cloze deletions for active recall practice.  Here is some related discussion on an AnkiDroid issue:  https://github.com/ankidroid/Anki-Android/issues/6121

RumovZ:
Hi, @andrewsanchez. Thanks for sharing your workflow. I always find it enlightening to read how other people use Anki.
I reverted that change in 5574cb1 after Damien had made a similar point. So the list you've quoted correctly describes the status quo of this PR.

andrewsanchez:
Awesome, thanks for clarifying, @RumovZ.  I think that the concept of passively reviewing "empty cards," i.e., cards that have no support for active recall practice by way of a cloze deletion or a back side to the card, will be integral to supporting some sort of incremental reading workflow, which I've sort of been working on.

PR #1231, opened 2021-06-13
Fix filtered deck from browser

RumovZ:
https://forums.ankiweb.net/t/anki-2-1-45-beta/10664/14


dae:
Whoops, I didn't notice that typo before either. Thanks :-)

PR #1232, opened 2021-06-13
Add hook to onDelete in AddonsDialog

ThoreBor:
This would be useful for showing a message before deleting an add-on. I would use it for either informing the user that deleting the add-on will only delete the local files, not the data on my server, or for automatically deleting the data before deleting the add-on.
I'm not sure if `"selected: List[str]"` is correct, so please let me know if I should change it.

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -842,11 +842,11 @@ class AddonsDialog(QDialog):
 
     def onDelete(self) -> None:
         selected = self.selectedAddons()
-        gui_hooks.addons_dialog_will_delete_addon(self, selected)
         if not selected:
             return
         if not askUser(tr.addons_delete_the_numd_selected_addon(count=len(selected))):
             return
+        gui_hooks.addons_dialog_will_delete_addons(self, selected)
         for dir in selected:
             if not self.mgr.deleteAddon(dir):
                 break
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -842,6 +842,7 @@ class AddonsDialog(QDialog):
 
     def onDelete(self) -> None:
         selected = self.selectedAddons()
+        gui_hooks.addons_dialog_will_delete_addon(self, selected)
         if not selected:
             return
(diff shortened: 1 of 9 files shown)


qt/tools/genhooks_gui.py
```diff
@@ -876,6 +876,11 @@ def on_top_toolbar_did_init_links(links, toolbar):
         args=["dialog: aqt.addons.AddonsDialog", "add_on: aqt.addons.AddonMeta"],
         doc="""Allows doing an action when a single add-on is selected.""",
     ),
+    Hook(
+        name="addons_dialog_will_delete_addon",
```
dae, line 880:
```suggestion
        name="addons_dialog_will_delete_addons",
```

qt/tools/genhooks_gui.py
```diff
@@ -876,6 +876,11 @@ def on_top_toolbar_did_init_links(links, toolbar):
...
+        args=["dialog: aqt.addons.AddonsDialog", "selected: List[str]"],
```
dae, line 881:
```suggestion
        args=["dialog: aqt.addons.AddonsDialog", "ids: List[str]"],
```

qt/aqt/addons.py
```diff
@@ -842,6 +842,7 @@ def onViewFiles(self) -> None:
 
     def onDelete(self) -> None:
         selected = self.selectedAddons()
+        gui_hooks.addons_dialog_will_delete_addon(self, selected)
```
dae, line 845:
```suggestion
        gui_hooks.addons_dialog_will_delete_addons(self, selected)
```

dae: The hook should probably be run after the 'not selected' and 'askUser' lines, not before it.

dae (commented):
Thank you for contributing! Some comments below.

ThoreBor:
Okay, you're right. It makes more sense like this. Thanks! 
Sorry, I had some issues resolving local conflicts, but I think it should be fine now.

dae:
Thanks!

PR #1233, opened 2021-06-14
Fix missing hyphen escape when normalizing search

RumovZ:
https://forums.ankiweb.net/t/escaping-hyphens-in-search-no-longer-working/10746


dae:
Thanks Rumo :-)

PR #1234, opened 2021-06-15
Remove `tag:none`, add new icon for `-tag:_*`

hgiesel:
This PR features two major changes:

1. It removes the special treatment of `tag:none`, and prefers `-tag:_*` instead. This is an attempt to streamline `field-name:` and `tag:`.

2. I adds new tag icons, giving "Untagged" its own icon. Both are from the mdi icon set:

image


dae:
The separate icon looks good, and I like how manual edits to the qrc file will no longer be required!

I'm not as sure about removing tag:none. It was a documented search, so we'd be breaking backwards compatibility if we remove it. It does not represent much of a maintenance burden, and I don't think I've ever heard someone complain that they wanted to use 'none' as a tag. I do appreciate the desire to unify handling across the different searches though.

You've made escape_anki_wildcards() no longer always do what it says it'll do, so if we do decide to go down this path, it will need an updated docstring or a rename. 

@RumovZ thoughts?

RumovZ:
I completely agree with you, @dae. It's great to remove unnecessary special cases, but I'm not sure if it's worth the disruption it causes to users accustomed to the long existing search syntax.

The icon changes are wonderful!

dae:
Thanks Henrik!

hgiesel:
For some reason I thought `tag:none` was a very recent addition. The icons will still put out `[-]tag:_*`, but you can still use `tag:none`.

I will then also remove `tag:none` from the documentation, letting it exist for purely backwards compatibility.

PR #1235, opened 2021-06-15
Make fieldname:test act like fieldname:%test%

hgiesel:
This would make it act more like unqualified searches or `nc:text` searches. The old behavior could be recreated with `fieldname:re:^test$`.

I think users would benefit from this more "fuzzy" search, because you can do without resorting to special symbols (`*`, `^`, `$`).

### DIFF
--- a/pylib/tests/test_find.py
+++ b/pylib/tests/test_find.py
@@ -109,7 +109,7 @@ def test_findCards():
     assert len(col.findCards("front:sheep")) == 0
     assert len(col.findCards("back:sheep")) == 2
     assert len(col.findCards("-back:sheep")) == 3
-    assert len(col.findCards("front:do")) == 0
+    assert len(col.findCards("front:do")) == 1
     assert len(col.findCards("front:*")) == 5
     # ordering
     col.conf["sortType"] = "noteCrt"
--- a/pylib/tests/test_find.py
+++ b/pylib/tests/test_find.py
@@ -219,12 +219,12 @@ def test_findCards():
     else:
         print("some find tests disabled near cutoff")
     # empty field
-    assert len(col.findCards("front:")) == 0
+    assert len(col.findCards("-front:_*")) == len(col.findCards("front:re:^$")) == 0
     note = col.newNote()
     note["Front"] = ""
     note["Back"] = "abc2"
     assert col.addNote(note) == 1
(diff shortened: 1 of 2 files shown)


dae:
It would make things more consistent with some of the other searches, but as a counterpoint, the current behaviour is similar to the w: search, in that leading/trailing *s can be added to change it.

It's also a backwards-incompatible change, and likely to affect more users than #1234. The manual mentions the example of having a 'page' field to list a page number, and such cases are going to go from page:123 to page:re:^123$, which is more awkward than the current prefix/suffix arrangement.

So not so sure about this - maybe worth discussing further on the forums to see what the rest of the community thinks?

hgiesel:
My thought was that this change would only widen the search, not restrict it.
In my experience it's more convenient do have the wider search by default, as you can always just scan the output with your eyes and get what you want.

Another point, why I think this is a good change is HTML. I myself was often tripped up by some HTML invalidating my search. Let's say the user applied some styling to the page number, or has a rogue `<br>` behind it (which was a common occurrence until recently), or maybe just typed a space by accident, it would invalidate the search and require some background knowledge on why that was the case.

Maybe the use case `re:^...$` even warrants another "Regex shortcut", like `w:`. Something like `page:m:123` for "exact *m*atch".

dae:
IIRC regexs are 2-4x speed penalty over the sql 'like' statements. Treating everything as regexes would make the code simpler, but would be slower.

Introducing another prefix for an exact like match is a possible solution, and the unprefixed field case could be mapped to it to retain backwards compatibility. We're not restricted to colons though - perhaps it's also worth considering other options like f:page=52?

hgiesel:
> We're not restricted to colons though - perhaps it's also worth considering other options like f:page=52?

Although we should probably wait with something like this after we drop the implicit field names. Otherwise queries like `src="my_image_name.png"`, e.g., when searching for HTML attributes will also be interpreted as a field query.

dae:
To be clear, I'm not pushing this syntax specifically, just throwing out ideas. But my thought was to have two syntaxes for full/partial matches, eg

f:page:52 → partial match
f:page=52 → full match

Then existing unqualified searches would be mapped to the latter, eg

page:52 → f:page=52

The = wouldn't be special outside of an f: prefix, so the src= search would act like it usually does.

dae:

PR #1236, opened 2021-06-15
Make field queries default to field:Fieldname:Text

hgiesel:
My original intention was to generalize `re`, `nc`, and `w`, to a new `TextNode`, and make it available to more search queries. The biggest hurdle to this realization is the catch-all `SingleField`. If we had field queries prefixed with `field:`, we could be a bit more creative with queries going forward. This PR will not turn them off yet, just "auto-correct" them to be prefixed with `field:` instead.

Another issue with non-prefixed field queries, that users cannot search for field content anymore, if the field is named `Introduced` or other already existing queries.

Together with #1235, this would mean that we can have a query `field:Fieldname`, which will search for notes/cards of note types which have the field "Fieldname", which I think makes sense.

### DIFF
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -382,6 +382,7 @@ fn search_node_for_text_with_argument<'a>(
     val: &'a str,
 ) -> ParseResult<'a, SearchNode> {
     Ok(match key.to_ascii_lowercase().as_str() {
+        "field" => parse_field(val)?,
         "deck" => SearchNode::Deck(unescape(val)?),
         "note" => SearchNode::Notetype(unescape(val)?),
         "tag" => SearchNode::Tag(unescape(val)?),
@@ -659,6 +660,18 @@ fn parse_dupe(s: &str) -> ParseResult<SearchNode> {
             s,
             FailKind::Other(Some("invalid 'dupe:' search".into())),
         ))
+    }
+}
+
+fn parse_field<'a>(s: &'a str) -> ParseResult<'a, SearchNode> {
+    let mut it = s.splitn(2, ':');
+    let key = it.next().unwrap();
+
+    if let Some(val) = it.next() {
+        parse_single_field(key, val)
+    } else {
+        // Searches for existence of field
+        parse_single_field(key, "")
     }
 }
 
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -663,7 +663,7 @@ fn parse_dupe(s: &str) -> ParseResult<SearchNode> {
     }
 }
 
-fn parse_field<'a>(s: &'a str) -> ParseResult<'a, SearchNode> {
+fn parse_field(s: &str) -> ParseResult<SearchNode> {
     let mut it = s.splitn(2, ':');
     let key = it.next().unwrap();
 
@@ -834,9 +834,9 @@ mod test {
 
         // escaping is independent of quotation
         assert_eq!(
-            parse(r#""field:va\"lue""#)?,
+            parse(r#""fieldname:va\"lue""#)?,
             vec![Search(SingleField {
-                field: "field".into(),
+                field: "fieldname".into(),
                 text: "va\"lue".into(),
                 is_re: false
             })]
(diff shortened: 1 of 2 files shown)


dae:
Could you elaborate more on your end goal here? Which queries were you thinking would benefit from re/nc/w?

Am I understanding correctly that you're intending field:xxx: to be an alias for xxx:, or are you suggesting that the latter be removed eventually?

hgiesel:
> Am I understanding correctly that you're intending field:xxx: to be an alias for xxx:, or are you suggesting that the latter be removed eventually?

I would suggest to remove it eventually, but that doesn't have to happen anytime soon, and I recognize that it would be a big change, so I'd opt for deprecating it for now.

> Which queries were you thinking would benefit from re/nc/w?

My thought was that re/nc/w basically represent string functions. So any `SearchNode` taking a string as input could benefit from it, very generically.
So instead of:

```rust
pub enum SearchNode {
    UnqualifiedText(String),
    SingleField { field: String, text: String, is_re: bool },
    Regex(String),
    NoCombining(String),
    WordBoundary(String),
    ...
}
```
We could have:

```rust
pub enum SearchNode {
    UnqualifiedText(TextNode),
    SingleField { field: TextNode, text: TextNode }
    ...
}
pub enum TextNode {
    Raw(String),
    Regex(TextNode),
    NoCombining(TextNode),
    WordBoundary(TextNode),
}
```

There's already some duplicated logic I think with `Regex(String)` and `SingleField` with `is_re == true`, and extending it further (`is_nc`, `is_w`) would lead to more duplicate code. But with this structural change, we could do easily do something like `field:fname:re:nc:mytext`. `deck`, `tag` could also all benefit from regular expression search to some degree, and with this change it will be quite natural to implement it for them.

To parse this, it would take some change to the following function:
```rust
fn  search_node_for_text_with_argument<'a>(
    key: &'a str,
    val: &'a str,
) -> ParseResult<'a, SearchNode> {
    Ok(match key.to_ascii_lowercase().as_str() {
        ...
        // anything else is a field search
        _ => parse_single_field(key, val)?,
    })
}
```

Where we'd basically have to bubble up this catch all one level up and instead whenever we encounter a key:
1. Search for a `SearchNode` key, including `field`, etc.
2. Search for an `UnqualifiedText` and its `TextNode`, like `re`, `nc`, `w`
3. And catch it as an "unqualified" field search

However there's still one issue here, while catch-all fieldname search still exists, instead of `   SingleField { field: TextNode, text: TextNode }`, we can only possibly have `SingleField { field: String, text: TextNode }`, otherwise we couldn't distinguish between an `UnqualifiedText` using `re`, or a unqualified field search using `re` for its fieldname.

dae:
The search code uses direct SQL for some queries. An unqualified search uses a couple of SQL 'like' queries for example, and re:/nc: use SQL helper functions. That is going to complicate an attempt to unify the different searches, as it's not just a case of applying a list of functions to a string in Rust like in the template filter case. And things may get even more complicated if we start using SQLite's FTS capabilities in the future.

combining w and re does not make much sense, as w is just a convenient shortcut for a regex.

What you propose would be more flexible, but I think before we get too focused on answering the "can we" question, we should consider the "should we" question - is this extra flexibility going to be worth the extra engineering cost, and the possible regressions caused by changing known-working code? That is why I asked what queries you thought this may be useful for. From your example we've got "limiting to specific fields by regex", and "combining nc and re in a search". Are these things you've found yourself needing in the past, and if so, how frequently? Would these changes be solving real pain points for users, or are they more about nice-to-haves?

hgiesel:
Probably what set me off to this idea was when I added a `_` in a deck search. At first I thought it was a straight-out bug, but after some experiments I noticed there was a pattern to it, adding an implicit `*` to the beginning and end of the search (governed by `is_glob` in the code). Then I thought, I could maybe do something like `deck:re:^deckn_me$`, but well it also didn't work (it doesn't give an error either), which is when I had a deep dive.

This is certainly not a real pain point for the average user (it would be interesting to know how many users are even entering queries by hand beyond unqualified text searches), but it would certainly be nice if the user could take the experience from one query and apply it to another. I think for any potential users who would want to try to understand the queries a bit more, an experience of "this does not intrinsically make sense", would be a big turn-off.

Speaking of the very specific change this PR does, I would say it does solve a pain point. I do remember at one point having to rename a field from "nid" to "noteid", so you could actually search for it in the browser for my add-on "Ze Add Note Id" 886709118. I also think a user having a field by the name "Introduced" is not a far-fetch.

RumovZ:
I agree that the current behaviour of interpreting all unrecognised keywords as field names is a bit awkward, both code- and interfacewise, and I would be in favour of softly deprecating it via search normalisation. For completeness' sake, though, it's already possible to perform field searches for reserved keywords: `introdued*:text` (or `introduce_:text *d:text` if you really need an equivalent search).

dae:
Your arguments for changing the default make sense, I just worry this is going to be a bit cumbersome for people to type/edit, especially if the intention is to perhaps one day remove the implicit search. Is it a common enough case to justify a single character prefix like f:introduced:text I wonder? field: is more obvious and user friendly, but f: may be preferred by the power user. Or maybe it's less of an issue if we one day offer autocomplete for search terms.

hgiesel:
Once we remove the implicit search, we could generally be more creative with query aliases. Another idea would be to add fields in the Sidebar under note type.

dae:

PR #1237, opened 2021-06-16
Add escape_anki_wildcards_for_search_node

hgiesel:
This addresses a concern you mentioned in #1234


dae:
Thanks for the quick fix!

PR #1238, opened 2021-06-16
Update icons of the browser sidebar

hgiesel:
All icons of `SidebarItem`s are updated, except for flags.

Images:
Screenshot 2021-06-16 at 16 23 07

Screenshot 2021-06-16 at 16 24 59

Highlights:
* I've added Notetype fields as `SidebarItem`s. These items will search for cards of this notetype, which are non-empty in this specific field. They also have a special menu item, to go to the Fields dialog. @RumovZ Could you possibly review d918a195856ca93c866af16bb55726c4e5b64de6?
* Collapse decks and tags have special icons.
* The current deck has a special icon
* Filtered decks have special icons
* The saved search icon now resembles a bookmark. There would also a heart and multiple heart icon available in the mdi iconset. 
* Flags and Card State have a neutral colour on their root icon, rather than grey.

Personally I like how the icons look more uniform now. 
I didn't change the flag icon, because the mdi version was a tad smaller, which made it hard to distinguish the colour, and ours doesn't stick out.
I'm not super happy with how the notetype icon and template/card type icon look a bit too similiar...

### DIFF
--- a/qt/aqt/browser/sidebar/item.py
+++ b/qt/aqt/browser/sidebar/item.py
@@ -25,6 +25,7 @@ class SidebarItemType(Enum):
     NOTETYPE_ROOT = auto()
     NOTETYPE = auto()
     NOTETYPE_TEMPLATE = auto()
+    NOTETYPE_FIELD = auto()
     TAG_ROOT = auto()
     TAG_NONE = auto()
     TAG = auto()
@@ -144,7 +145,10 @@ class SidebarItem:
                 SidebarItemType.CARD_STATE,
             ):
                 return self.name == other.name
-            elif self.item_type == SidebarItemType.NOTETYPE_TEMPLATE:
+            elif self.item_type in [
+                SidebarItemType.NOTETYPE_TEMPLATE,
+                SidebarItemType.NOTETYPE_FIELD,
+            ]:
                 return (
                     other.id == self.id
                     and other._parent_item.id == self._parent_item.id
--- a/qt/aqt/browser/sidebar/tree.py
+++ b/qt/aqt/browser/sidebar/tree.py
@@ -666,15 +666,16 @@ class SidebarTreeView(QTreeView):
 
     def _flags_tree(self, root: SidebarItem) -> None:
         icon = ":/icons/flag.svg"
+        icon_outline = ":/icons/flag-outline.svg"
+
         root = self._section_root(
             root=root,
             name=tr.browsing_sidebar_flags(),
-            icon=icon,
+            icon=icon_outline,
             collapse_key=Config.Bool.COLLAPSE_FLAGS,
             type=SidebarItemType.FLAG_ROOT,
         )
         root.search_node = SearchNode(flag=SearchNode.FLAG_ANY)
-        colored_icon = ColoredIcon(path=icon, color=colors.DISABLED)
 
         for flag in load_flags(self.col):
             root.add_child(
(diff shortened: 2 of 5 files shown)


RumovZ:
That looks great! I've added one little comment.
I wouldn't worry too much about the template icon as it's hidden most of the time and hierarchy and identation also help to differntiate between notetypes and fields.

dae:
I like it as a whole; the previous tag icon stood out a bit compared to the existing other icons, and these mdi icons look better.

A couple of thoughts:

- Separate icons for decks with subdecks makes it feel a bit busy. There's already a disclosure icon denoting items with subitems - maybe that's enough?
- Same with the top level items - if we're showing multiples for the top level items then we probably should be doing it for all top-level items, and multiples for things like flags and today might look odd?
- Saved searches feels a bit similar to the deck icon - the advantage of the heart was it was more visually distinct

kleinerpirat:
After trying this out, I have to say this works really well. Bravo, Henrik!

> Saved searches feels a bit similar to the deck icon - the advantage of the heart was it was more visually distinct

@dae I would actually prefer the new bookmark icon because it represents the saving-functionality in a more literal way. Since it is all the way on top, I don't think the similarity to the deck icons will matter too much.

The resemblance between the deck and note icons (in comparison to the old icons) might be a bit more problematic. As they are quite small and rectangular in shape, they don't differ a lot at a quick glance:

image image


 Perhaps an icon that denotes HTML in some way would be more fitting:
`import { mdiCodeTags } from '@mdi/js';` code-tags

But I guess I'm biased and most users won't instantly connect note types with HTML.

dae:
I don't feel strongly about the saved searches icon, happy to go with whatever the majority likes

hgiesel:
I've removed the multiple-icons, and reverted to hearts for saved searches for now. I also couldn't unsee the resemblance to the bookmark in the deck icon. I think a star could also make sense, however that's also associated with marked cards. This could be a forum poll, but for now I'll stick to the heart.

I've adopted @kleinerpirat suggestion. After all, an "alternative" (old?) term for card types was "templates", so I've taken one that symbolizes this to some degree.

Screenshot 2021-07-01 at 00 39 44

The `{}` here stands for `{{field references}}`.
At the same type I switched the note type icon. Being a black-on-white icon, and its frilly top edge makes it stand out a bit more.

To explain why I didn't choose other maybe _simpler_ icons, I already considered icons for other "objects" within Anki, and saved them for future use. I think the [card-bulleted-outline](https://materialdesignicons.com/icon/card-bulleted-outline), which I first used for card types, could symbolize cards, [note-outline](https://materialdesignicons.com/icon/note-outline) could symbolize notes, and [beaker-outline](https://materialdesignicons.com/icon/beaker-outline) could resemble deck options.

hgiesel:
I've reconsidered my previous judgement on the flag icons:
- By using the md icons, we also have access to the modifier versions, with little pluses or minuses, etc.
- For "no flag", I used the unfilled version. I think that's a nice way to symbolize "no color". (there's no striked through version, like the one for tags, which might have been even better).

Screenshot 2021-07-01 at 01 22 05

dae:
The flag icon feels a bit pixelated to me, but the modifiers will presumably be useful, and I don't want to hold up the PR. Thanks!

hgiesel:
> The flag icon feels a bit pixelated to me.

That was exactly my thought as well. There's also `flag-alternative`, which is less pixelated, but has the above-mentioned issue of being smaller, so that the colours are harder to see (additionally there's no modifier versions).

PR #1240, opened 2021-06-17
Add darwin arm 64 to platforms (Apple silicon / M1)

sergioisidoro:
Hey there,

This is a long shot because I'm still trying to understand the build process

I hope I got this right, that build-kite just uses Bazel with dynamic arguments picked up from the platforms.
But I'm not sure how this will work on build-kite, and if there needs to be some configuration to run this build in an ARM machine.

Thought of opening an PR instead of just opening an issue.

Have a great day.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -83,6 +83,7 @@ qubist-pixel-ux <github.com/qubist-pixel-ux>
 cherryblossom <github.com/cherryblossom000>
 Hikaru Yoshiga <github.com/hikaru-y>
 Thore Tyborski <github.com/ThoreBor>
+Sergio Isidoro <smaisidoro@gmail.com>
 
 ********************
 
--- a/platforms/BUILD.bazel
+++ b/platforms/BUILD.bazel
@@ -20,7 +20,7 @@ config_setting(
     name = "macos_darwin_arm64",
     constraint_values = [
         "@platforms//os:macos",
-        "@platforms//cpu:darwin_arm64",
+        "@platforms//cpu:arm64",
     ],
 )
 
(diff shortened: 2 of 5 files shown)


sergioisidoro:
https://github.com/protocolbuffers/protobuf/pull/8557 
This PR adds protobuff compatibility, but the binaries still are not uploaded. I think that was the idea:

> so I attempt here to upload the same file used in the osx-x86_64 classifier for the osx-aarch_64 classifier.

Anyway, in this release we can see that Apple Silicon support was added.
https://github.com/protocolbuffers/protobuf/releases/tag/v3.17.3

So for now I'm just pointing to the x86_64 as a placeholder.
Still needs the Clang binaries that are uploaded to this repository.

dae:
Unfortunately it'll likely be at least 4 months before we can take advantage of these changes, as Qt and PyQt do not currently provide binaries for arm64

sergioisidoro:
@dae Thank you for your prompt reply 

Well, it might be less than 4 months. I was astonished by the amount of things already available for arm64. Things are moving at incredible speed.

And there seems to be promising progress in QT: https://bugreports.qt.io/browse/QTBUG-85279
It's already possible to build QT from source for arm64, and pyqt is already available for apple silicon with Homebrew.

If you know of anything I can do to help push this forward, let me know :)

dae:
Anki requires WebEngine, which won't be available until Oct: https://wiki.qt.io/Qt_6.2_Release

mvanaltvorst:
It seems that all dependencies have been updated @dae, how can we help implement this? Would help a lot of Apple M1 users save battery power

dae:
An initial test build is now available.

https://forums.ankiweb.net/t/new-toolkit-and-packaging-test-round-2/14513

PR #1242, opened 2021-06-18
In-line HTML-Editing for Editor

hgiesel:
This is already pretty feature complete.
Instead of opening an entirely new window, the HTML button will now toggle the current field into HTML-edit mode. The HTML-editing is done in an element called `Codable` (as compared to `Editable` for the WYSISWYG-editing).
Changes made in the `Codable` element are saved exactly like changes in the `Editable`.
During HTML-editing most toolbar buttons are disabled. We have syntax highlighting, line numbers, and the matching tags will be highlighted when selected.
All of this is built upon CodeMirror.

One feature that would be nice to have would be pretty formatting, when activating the HTML-Editor, and minifying/uglifying when writing back. This could be done using the `prettyify` and `terser` libraries, although it might add a bit overhead. I would add this in another PR if this one checks out.

Screenshot 2021-06-18 at 02 55 26

The only question is, was some vital processing done by BeautifulSoup, that is not done by DOMParser?


dae:
The 200k+ increase to the bundle size is hefty, but this looks like a big enough quality of life improvement to justify it. I imagine there will be lots of people who appreciate this change - thank you!

Prettying could potentially be a breaking change if the user is using whitespace: pre-wrap or similar in their card template, so it may be better done as a separate command rather than automatically. 

beautifulsoup was mainly there to correct issues like the user inputing unbalanced HTML tags, which could lead to field content spilling out of the field before you implemented shadow root. If we can accomplish the same thing in JS, that'll be one less dependency we need to worry about it.

PR #1243, opened 2021-06-18
Add minimal tests to html-filter

hgiesel:
This is more of an issue, then a PR: When actually trying to create actual tests that use the DOM API, I get this error:

```
 FAIL  ts/html-filter/index.bundle.test.js
  filterHTML
    ✕ zero input creates zero output (1 ms)

  ● filterHTML › zero input creates zero output

    The error below may be caused by using the wrong test environment, see https://jestjs.io/docs/configuration#testenvironment-string.
    Consider using the "jsdom" test environment.

    ReferenceError: document is not defined

      at filterHTML (../../../../ts/html-filter/index.ts:41:22)
      at Object.<anonymous> (../../../../ts/html-filter/index.test.ts:8:16)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        0.754 s
Ran all test suites.
```

I tried adding the header 
```
/**
  * jest-environment jsdom
  */
```
to the header of `index.test.ts`, like it is suggested in the link. However this is filtered out, once it is compiled to `index.bundle.test.js`.

### DIFF
--- a/ts/html-filter/BUILD.bazel
+++ b/ts/html-filter/BUILD.bazel
@@ -5,17 +5,18 @@ load("//ts:eslint.bzl", "eslint_test")
 
 ts_library(
     name = "html-filter",
-    module_name = "html-filter",
     srcs = glob(
         ["*.ts"],
         exclude = ["*.test.ts"],
     ),
+    module_name = "html-filter",
     tsconfig = "//ts:tsconfig.json",
     visibility = ["//visibility:public"],
     deps = [],
 )
 
 jest_test(
+    env = "jsdom",
     deps = [
         "html-filter",
     ],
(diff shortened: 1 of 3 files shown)


dae:
Fix pushed

hgiesel:
In that case, would you merge this PR? I would like to try a few things on top of this change.

hgiesel:
[Here](https://github.com/ankitects/anki/compare/main...hgiesel:nonexported?expand=1) I tried to enable the testing of non-exported functions using `rewire`. However after reading their docs more carefully, I don't think `rewire` can work with ES syntax. There's also a [Babel plugin](https://github.com/speedskater/babel-plugin-rewire), which is inspired by rewire, but works with ES syntax, but I haven't found anything for esbulid.

Let me know if you have any other ideas.

dae:
Could we use an approach like this? 

https://stackoverflow.com/a/65422568

dae:
The alternative may to be switch to calling Jest outside of Bazel, which would presumably allow us to run the tests on Windows as well, and might make interactive debugging possible, but it would mean test results could no longer be cached, and a separate testing step would be required.

PR #1244, opened 2021-06-19
Make Color Pickers also format with colors

hgiesel:
Was mentioned [here](https://forums.ankiweb.net/t/anki-2-1-45-beta/10664/20).
It also gives a good reason why to deactivate them when no field is selected :)


dae:
Thanks for that!

PR #1247, opened 2021-06-21
Call update_notetype_legacy after toggling sticky

### LINKED ISSUE #1246: pin fields forgotten after opening cards screen
1. Have one or more fields pinned
2. Open Cards... and close it again (no need to save)
3. Add the card, and the pinned fields are not retained

@hgiesel mind looking into this when you have a chance?

While I have your attention, do you have any other changes in the pipeline that are important to get into 2.1.45? I'd like to start pushing larger changes back until a subsequent release soon, so we can get 2.1.45 out in the not too distant future.

hgiesel:
Closes #1246

When the Fields dialog is opened from the AddCards dialog is opened, the sticky value from before the AddCards was opened will be read, and then written back.


dae:
Thanks for the quick fix!

PR #1248, opened 2021-06-21
Move deck options revert button to right; hide when inactive

dae:
IMHO, showing an inactive button next to every element makes the screen feel a bit busy:

IMG_8F064466F6E0-1

If we hide the button when inactive, the margins then look odd. Note also that the second bury switch wraps awkwardly:

IMG_1363

While no masterpiece, having the revert button on the right seems to be slightly preferable to me. Now that it has a separate confirmation step, there shouldn't be a risk of accidental triggering, and it allows a bit more room for the text on the left. Thoughts?

IMG_1364

### DIFF
--- a/ts/deckoptions/EnumSelectorRow.svelte
+++ b/ts/deckoptions/EnumSelectorRow.svelte
@@ -20,10 +20,10 @@
 
 <Row>
     <Col size={7}>
-        <RevertButton bind:value {defaultValue} />
         <TooltipLabel {markdownTooltip}><slot /></TooltipLabel>
     </Col>
     <Col {breakpoint} size={5}>
         <EnumSelector bind:value {choices} />
+        <RevertButton bind:value {defaultValue} />
     </Col>
 </Row>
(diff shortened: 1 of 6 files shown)


dae:
On a wider screen, before:

Screen Shot 2021-06-22 at 9 55 16 am

After:

Screen Shot 2021-06-22 at 9 54 55 am

hgiesel:
I like it. Hiding them is better than keeping them inactive. My thought was they could work kinda like bullet points, but having the empty gap on the right looks better than on the left.

There are still several things I might want to touch on, including:

* why the badge isn't positioned on the second line here:
Screenshot 2021-06-22 at 13 17 26
* The inputs should be aligned with the first line of the label, not the entire label (?)
* two-column layout
* Having a different background-color for one each section

But now I'll keep it in small PRs again ^^;

dae:
Sounds good!

Do you mean mean alternating between two colours like the Browser grid? Are you thinking of doing that for each section, or for each option in a section?

PR #1249, opened 2021-06-22
Adapt use of createDropdown for editor buttons

hgiesel:
I introduced a regression in #1207 for the dropdowns in the editor when creating the dropdown on the gear icon badge.


dae:
Thanks!

Issue #1051, opened 2021-03-03
Convert divs when cloze deleting?


dae:
The issue mentioned on https://github.com/ijgnd/anki__editor__apply__font_color__background_color__custom_class__custom_style/issues/45 also affects cloze deletions made across multiple divs - they end up having extra lines inserted, and don't display properly during review, as the subsequent text manipulation creates invalid HTML.

While the new editor code should use `<br>`s when typing in, a lot of users paste in content from other sources. I wonder if we could make the cloze deletion action handle this? If it detected a selection across multiple nodes, maybe it could automatically convert the nodes covered by the selection into linebreaks instead? Or maybe this would cause more problems than it would fix?

hgiesel:
If the user wanted to cloze over both "Second line" and "Third line", in the following text:
```html
<div><div><div>First line</div>Second line</div>Third line</div>
```

Would the operation result in:
```html
<div>First line</div>Second line<br>Third line
```

or
```html
First line<br>Second line<br>Third line
```

?
It's not always quite clear, because the nested `div`s are so hard to predict.

The operation could be cleaner if we could act at least on one whole fragment,  like the whole pasted content (putting the `<div>` -> `<br>` logic into `editor/htmlFilter.js`).

Another concern is if the `div` elements are styled. `<div style="...">content</div>` is not equivalent to `<span style="...">content</span>` if it contains advanced styling, e.g. background-color, borders, etc..

dae:
Anki already has fairly strict filtering of content when pasting, so maybe rewriting the divs on paste into BRs is not outside the realm of possibility. I do fear it may lead to unwanted style changes though, so not sure how practical it would be. Suggestions/proof of concepts welcome :-)

dae:
The problem is not just limited to divs either - another example of invalid HTML after cloze deleting:

https://forums.ankiweb.net/t/random-bullet-point-added-when-using-cloze-deletions-after/8598

hgiesel:
Another occurrence with what seems like exclusively inline elements.:

https://forums.ankiweb.net/t/some-parts-of-clozed-text-not-colored/10936/3

PR #1250, opened 2021-06-23
Clozes as web components

hgiesel:
This is a draft which addresses https://github.com/ankitects/anki/issues/1051, but in a different way than suggested in the issue.

I have tried something similar before, but always failed. Two takeaways I can take from those failed attempts:
1. pseudo elements (`::before`, and `::after`) are terribly buggy in contenteditable. (Especially in WebKit browsers) (e.g. https://jsfiddle.net/bj0p3evr/).
2. Shadow roots are buggy in contenteditable (interestingly in the same way as pseudo elements)

This PR uses neither.
In the folllwing screenshot, `c1` is a web component, `c2` is just regular text.

Screenshot 2021-06-23 at 12 22 05

To the unknowing eye, the web component cloze does not look much different. But it has separate representations, depending on whether you ask `EditingArea`, `Editable`, or `Codable`.:

`EditingArea.fieldHTML` sees it as just text, just as the current implementation:
```
This is a {{c1::cloze}}.
```

`Editable.fieldHTML` sees it as a web component:
```html
This is a <anki-cloze card="1"><span contenteditable="false" data-cloze="open">{{c<input is="anki-cloze-number-input" type="number" tabindex="-1" min="1" max="499" class="anki-cloze-number-input" style="width: 25px;">::</span>cloze<span contenteditable="false" data-cloze="close">}}</span>.</anki-cloze>
```

`Codable.fieldHTML` sees it as a web component without the extra markup (allows for easier editing, and hiding of implementation details):
```html
This is a <anki-cloze card="1">cloze.</anki-cloze>
```

Treating the cloze as HTML in Editable and Codable will prevent invalid HTML being rendered upon review.

Other advantages could include:
1. The tags are just an editor phenomenon. To everything beyond the editor, nothing will have changed (we keep compatibility with other platforms).
2. Hitting delete when the caret is situated right after the opening element `{{c1::`, or the closing element `}}` will entirely unwrap the clozed text from the cloze markup.
3. The cloze number is an input element now, allowing only for valid values.
4. Clicking three times within the clozed text will select the entire clozed text.

Disadvantages might include (might be able to be addressed):
1. If the `<anki-cloze>` is the last element in the `Editable`, the caret will be invisible, if it is positioned behind it
2. `<anki-cloze>` will sometimes "swallow up" text positioned behind it. This is certainly a bug.

Things not implemented yet:
- [ ] Copying the anki-cloze should probably copy the cleaned up version.
- [ ] Using Control-Shift-C needs to use `Range.surroundContents` instead of `execCommand`. `Range.surroundContents` is quite low level, so we need some additional logic (currently only works, if you're clozing over plain text (no other inline / block tags))
- [ ] The undo history does not work for this. We currently just rely on the built-in undo history of contenteditable in the Editor. For this to work, we would have to make `Control+Z` trigger our own undo and make the Editor create snapshots regularly. These snapshots would be the set of editor fields (and the tag editor, see below...) and a list of offsets, indicating the caret position at the moment of the snapshot.

PRs which need to come first.
1. Tag Editor: #1079
2. A PR implementing our own undo history for the editor, which will probably be required after the Tag Editor anyway.

What do you think of the idea in general?
My hope is that we could also use a similar strategy to display rendered MathJax (and maybe Latex) as images in the editor and add some controls, to edit the underlying MathJax code. (`<anki-mathjax>`, `<anki-mathjax-chemistry>`, etc.)

### DIFF
--- a/ts/editor/BUILD.bazel
+++ b/ts/editor/BUILD.bazel
@@ -27,17 +27,6 @@ compile_svelte(
 
 compile_sass(
     srcs = [
-        "editable.scss",
-    ],
-    group = "editable_scss",
-    visibility = ["//visibility:public"],
-    deps = [
-        "//ts/sass:scrollbar_lib",
-    ],
-)
-
-compile_sass(
-    srcs = [
         "fields.scss",
     ],
     group = "base_css",
@@ -45,6 +34,17 @@ compile_sass(
     deps = [
         "//ts/sass:base_lib",
         "//ts/sass:buttons_lib",
+        "//ts/sass:scrollbar_lib",
+    ],
+)
+
+compile_sass(
+    srcs = [
+        "editable.scss",
+    ],
+    group = "editable_scss",
+    visibility = ["//visibility:public"],
+    deps = [
         "//ts/sass:scrollbar_lib",
     ],
 )
--- a/ts/editor/ClozeButton.svelte
+++ b/ts/editor/ClozeButton.svelte
@@ -39,7 +39,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     function onCloze(event: KeyboardEvent | MouseEvent): void {
         const highestCloze = getCurrentHighestCloze(!event.getModifierState("Alt"));
-        wrapCurrent(`{{c${highestCloze}::`, "}}");
+        wrapCurrent(`<anki-cloze card="${highestCloze}">`, "</anki-cloze>");
     }
 </script>
 
--- a/ts/editor/ClozeButton.svelte
+++ b/ts/editor/ClozeButton.svelte
@@ -10,9 +10,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import WithShortcut from "components/WithShortcut.svelte";
     import WithContext from "components/WithContext.svelte";
 
+    import type { Cloze } from "./cloze";
     import { bracketsIcon } from "./icons";
-    import { forEditorField } from ".";
-    import { wrapCurrent } from "./wrap";
+    import { forEditorField, getCurrentField } from ".";
 
     const clozePattern = /\{\{c(\d+)::/gu;
     function getCurrentHighestCloze(increment: boolean): number {
@@ -39,7 +39,15 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     function onCloze(event: KeyboardEvent | MouseEvent): void {
         const highestCloze = getCurrentHighestCloze(!event.getModifierState("Alt"));
-        wrapCurrent(`<anki-cloze card="${highestCloze}">`, "</anki-cloze>");
+        const clozeWrapper = document.createElement("anki-cloze") as Cloze;
+        clozeWrapper.setAttribute("card", String(highestCloze))
+
+        const currentField = getCurrentField()!;
+        const selection = currentField.getSelection()!;
+
+        const range = selection.getRangeAt(0);
+        range.surroundContents(clozeWrapper);
+        clozeWrapper.caretToEnd();
     }
 </script>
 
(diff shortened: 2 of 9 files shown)


dae:
Can we put this on the backburner for a bit? I'm busy with updates to AnkiWeb & AnkiMobile at the moment, and would like to get to the import/export update soon. The current cloze handling is not ideal, and it would be nice if it could be handled better, but I just don't feel like I have the bandwidth to think about this at the moment :-)

hgiesel:
I assumed this would be the case, so it's fine with me. Just regard this as a notification that I've been thinking about #1051, in case you wanted to tackle it yourself :)

dae:
Thanks for understanding :-)

kleinerpirat:
Perhaps you could implement a function to decrease all following indices when the last cloze of an in-between index is deleted.

### Example: 
```
The quick brown {{c1::fox}} jumps over the lazy {{c2::dog}}.

Pack my box with {{c3::five dozen}} {{c4::liquor jugs}}.

The {{c5::five}} {{c6::boxing wizards}} jump quickly.
```

Deleting the second sentence would yield the following output:

```
The quick brown {{c1::fox}} jumps over the lazy {{c2::dog}}.

The {{c3::five}} {{c4::boxing wizards}} jump quickly.
```
## Motivation
While experimenting with clozes for the forum, I realized how annoying it would be to create big cloze texts and then delete a paragraph in the middle. Sure, no empty cards would be created, but wrong indices are detrimental to the overview. Right now, you would need to manually decrease them one by one, afaik.

While I'm not yet sure about the validity of huge cloze texts, I can't think of any drawbacks to this function... can you?

hgiesel:
@kleinerpirat 
After finishing the edit, this would associate the scheduling of cards 3 & 4 with those of cards 5 & 6.

kleinerpirat:
Ah yes, my thinking was limited to the Add-window here.

I assume shifting the entries from top-down (scheduling of each decremented card is copied to the lower card, then delete the uppermost ones) is not as trivial as it sounds to a layman like me?

hgiesel:
But then we'd have to make assumptions about which cards stay _semantically_ the same:

Let's say the user changes:
```
The quick brown {{c1::fox}} jumps over the lazy {{c2::dog}}.

Pack my box with {{c3::five dozen}} {{c4::liquor jugs}}.

The {{c5::five}} {{c6::boxing wizards}} jump quickly.
```

into:
```
The quick brown {{c1::fox}} jumps over the lazy {{c2::dog}}.

30 + {{c3::5}} = {{c4::35}}.
```

How would you tell that `c3` and `c4` should not inherit the original schedulings of `c5` and `c6`? And what if he changes into:

```
The quick brown {{c1::fox}} jumps over the lazy {{c2::dog}}.

The {{c3::six}} {{c4::boxing sorcerers}} jump quickly.
```

This would be quite some complexity. And in the end, what's the damage if there are some empty clozes?

kleinerpirat:
Thanks for the rapid and wise responses, Henrik.

Sitting in front of a [medicine timeline](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4379645/) with 257 clozes (experiment for the [forum](https://forums.ankiweb.net/t/anki-as-a-knowledge-base-with-a-massive-cloze-note/11658?u=kleinerpirat)) still makes me think there are use cases for _manual_ batch index editing. I forgot to cloze the date `500 BC` in the second paragraph and would love to squeeze it in there, but I can't. 

```
- {{c1::2600}}: BC The Egyptian {{c2::Imhotep}} describes the diagnosis and treatment of 200 diseases

- 500 BC {{c3::Alcmaeon of Croton}} distinguished {{c4::veins}} from {{c4::arteries}}

- {{c5::460}} BC: Birth of {{c6::Hippocrates}},
  the Greek father of medicine begins the scientific study of medicine
  and prescribes a form of {{c7::aspirin}}
```

Each paragraph has its date clozed as a separate card. To cloze `500 BC` afterwards, there are three options:
- use `{{c::258}}`
- look downwards and steal the index of an unneccesary cloze
- add it to card 3 or 4 (breaks the system)

None of them are attractive. In that case, it is clear what I want to achieve (increment all following indices) - Anki would not need to make assumptions.

It seems like massive cloze texts like this are not feasible at the moment. But maybe it shouldn't stay like that forever.

hgiesel:
I'll close this in favor of #1324. This PR was only meant as a show-case anyway, while #1324 is also very functional.

@kleinerpirat 
Like you admitted on the forum, having cloze indices that do not strictly incrase over the course of the note do some trigger some OCD (cannot deny that), but are in practice quite harmless.
A feature like "normalize cloze indices" could be introduced, that streamlines indices, and also assigns the correct cards, but I won't put that much priority on it for now.

PR #1251, opened 2021-06-24
Fix toggle night mode in clayout

### LINKED ISSUE #1166: in day mode, enabling 'night mode' in the card layout screen's options dropdown does not turn the background dark

hikaru-y:
Fixes #1166.

### Regarding emulation of day mode (especially on non-Linux platforms):

Taking into account the possibility that the background color is not set in the `.card` styling, I think we need to set the background color of the `<body>` element to the `QPalette.Window` color value of the system default palette in order to emulate it correctly. However, on Windows (and probably on macOS as well), once the `fusion` QStyle is applied to enable night mode, the palette retrieved by `standardPalette()` will be different from that in day mode.

#### On Windows:
QPalette_Window_day_mode
QPalette_Window_night_mode

This is probably because the default QStyle on Windows and macOS is not `fusion`. So in this PR, the system default palette is stored in `theme_manager.default_palette` before the `fusion` style is applied, so that the `QPalette.Window` color value in day mode can be used in night mode on any platform.

### DIFF
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -344,6 +344,10 @@ class CardLayout(QDialog):
 
     def on_night_mode_action_toggled(self) -> None:
         self.night_mode_is_enabled = not self.night_mode_is_enabled
+        force = json.dumps(self.night_mode_is_enabled)
+        self.preview_web.eval(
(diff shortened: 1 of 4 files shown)


dae:
Sounds like a good approach - thanks Hikaru!

PR #1252, opened 2021-06-24
Fix expanding/collapsing tags triggering refresh

RumovZ:
This is the easy fix. Despite what I've written on the forum, I thought you might prefer this way.

https://forums.ankiweb.net/t/anki-2-1-45-beta/10664/42

### DIFF
--- a/qt/aqt/browser/sidebar/tree.py
+++ b/qt/aqt/browser/sidebar/tree.py
@@ -688,7 +688,7 @@ class SidebarTreeView(QTreeView):
                 full_name = head + node.name
                 return lambda expanded: set_tag_collapsed(
                     parent=self, tag=full_name, collapsed=not expanded
-                ).run_in_background()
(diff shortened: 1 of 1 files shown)


dae:
Thanks Rumo, I was tempted to take this path, but realised that having expand/collapse undoable was making other things more complicated. The deck list was not updating when an expand/collapse was updated, and fixing it would have been more trouble than it's worth, now that we can exclude certain operations from the undo queue.

RumovZ:
Sorry, I didn't pay enough attention before. [This](https://forums.ankiweb.net/t/anki-2-1-45-beta/10664/59) made me realise that 73d9391 doesn't actually fix the reported problem. Since expanding/collapsing is still handled as an Op, maybe this PR is yet necessary after all?

dae:
Silly me, I overlooked that. Thanks for catching it :-)

PR #1253, opened 2021-06-24
Fix Codable carrying on content from other notes in Browser

hgiesel:
Reported [here](https://forums.ankiweb.net/t/anki-2-1-45-beta/10664/46).


dae:
Thanks for the quick fix!

PR #1254, opened 2021-06-25
Add hook "browser_sidebar_will_show_context_menu"

RumovZ:
Probably a hook that will come in handy for a lot of post-2.1.45 add-ons.


dae:
Sounds good - thanks Rumo!

PR #1255, opened 2021-06-26
Extensible two-column layout for deck options

kleinerpirat:
I had another go at making the deck options responsive for wide screens.
(Previous attempts: https://github.com/ankitects/anki/pull/1210, https://github.com/ankitects/anki/pull/1212)

The task was to make [this idea of mine](https://github.com/ankitects/anki/pull/1207#issuecomment-855393388) work with the Svelte components. [Here is the CodePen](https://codepen.io/kleinerpirat/pen/zYZjyKX) I created as a proof-of-concept.
I tried my best to keep the necessary changes at a minimum and make it extensible.

image
The grid layout takes effect when viewport width is `>= 992px`.

### How the current approach works
1. approximate size of each `TitledContainer` at runtime via amount of contained `.row` elements
2. add up total amount of `TitledContainer` sizes with store `lineCount`
3. calculate amount of rows needed to fit all elements on two columns

#### Why the extra effort with individual row spans?
 
**Left**: all elements span 1 row -> Big elements like "Advanced" create a lot of white space
**Right**: individual row spans reduce that white space significantly

### Extra
- set upper limit (currently `2rem`) for h1 font-size with [clamp](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp()) function, because indefinite scaling with viewport width doesn't make much sense for the two-column layout.

I am not really happy with the way I implemented the `lineCount` store to get the total row span of all components. I'm sure @hgiesel knows of multiple better ways to allow child components to change a parent component's variable.
The addition to `Container.svelte` is a bit dirty, but I found it more elegant than to wrap the options in an extra div within `DeckOptionsPage.svelte`. Let me know if I should still go for that second option.

### Tasks
- [ ] Find more elegant way to add up total amount of `TitledContainer` sizes (`lineCount`) across components
- [ ] Use more accurate and universal way to approximate size (e.g. [offsetHeight](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetHeight) of containing `div`)
One issue is that the correct offsetHeight is not immediately available with [onMount](https://svelte.dev/docs#onMount), probably due to rendering delay.

---

I don't have much faith in this getting merged because it feels a bit hacky. But I'd invite you to try it out and perhaps create something along these lines yourself (or guide me in the right direction), if you like the look of it.

### DIFF
--- a/ts/deckoptions/DeckOptionsPage.svelte
+++ b/ts/deckoptions/DeckOptionsPage.svelte
@@ -2,6 +2,10 @@
 Copyright: Ankitects Pty Ltd and contributors
 License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 -->
+<script context="module">
+    import { writable } from "svelte/store";
+    export const lineCount = writable(0);
+</script>
 
 <script lang="ts">
     import ConfigSelector from "./ConfigSelector.svelte";
@@ -59,11 +63,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     $: {
         rows = Math.ceil($lineCount / 2) + 1;
     }
-</script>
-
-<script context="module">
-    import { writable } from "svelte/store";
-    export const lineCount = writable(0);
 </script>
 
 <ConfigSelector {state} />
--- a/ts/deckoptions/DeckOptionsPage.svelte
+++ b/ts/deckoptions/DeckOptionsPage.svelte
@@ -2,17 +2,8 @@
 Copyright: Ankitects Pty Ltd and contributors
 License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 -->
-<script context="module">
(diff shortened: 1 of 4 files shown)


ts/deckoptions/DeckOptionsPage.svelte
```diff
@@ -63,6 +54,16 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+    let rows: number | undefined = undefined;
+    $: {
+        rows = Math.ceil($lineCount / 2) + 1;
+    }
+</script>
+
+<script context="module">
+    import { writable } from "svelte/store";
+    export const lineCount = writable(0);
 </script>
```
kleinerpirat, line 67: Checks failing because Prettier wants this all the way on top, but it's a hacky solution anyway - bound to get replaced.

kleinerpirat:
Damien's approach is much simpler and also better, as it is guaranteed not to waste any space. Time to close this one.

PR #1257, opened 2021-06-27
Scroll to first sidebar search match

abdnh:
This saves from a bit of manual scrolling to the highlighted results especially for long tag trees.

### DIFF
--- a/qt/aqt/browser/sidebar/tree.py
+++ b/qt/aqt/browser/sidebar/tree.py
@@ -215,23 +215,35 @@ class SidebarTreeView(QTreeView):
         # start from a collapsed state, as it's faster
         self.collapseAll()
         self.setColumnHidden(0, not self.model().search(text))
-        self._expand_where_necessary(self.model(), searching=True)
+        self._expand_where_necessary(
+            self.model(), searching=True, scroll_to_first_match=True
+        )
 
     def _expand_where_necessary(
         self,
         model: SidebarModel,
         parent: Optional[QModelIndex] = None,
         searching: bool = False,
-    ) -> None:
+        scroll_to_first_match: bool = False,
+    ) -> bool:
         parent = parent or QModelIndex()
         for row in range(model.rowCount(parent)):
             idx = model.index(row, 0, parent)
             if not idx.isValid():
                 continue
-            self._expand_where_necessary(model, idx, searching)
+            scroll_to_first_match = self._expand_where_necessary(
+                model, idx, searching, scroll_to_first_match
+            )
             if item := model.item_for_index(idx):
                 if item.show_expanded(searching):
                     self.setExpanded(idx, True)
+                if item.is_highlighted() and scroll_to_first_match:
+                    self.selectionModel().setCurrentIndex(
+                        idx, QItemSelectionModel.SelectCurrent
+                    )
+                    self.scrollTo(idx)
+                    scroll_to_first_match = False
+        return scroll_to_first_match
 
     def update_search(
         self,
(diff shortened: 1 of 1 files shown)


dae:
Thanks, it sounds like a useful addition. The use of an instance variable makes the code a bit harder to reason about. Could this instead check for `searching` inside expand_where_necessary, and keep the state as a local variable instead?

dae:
@RumovZ please jump in if you have any thoughts

abdnh:
Maybe this will do it?

```diff
--- a/qt/aqt/browser/sidebar/tree.py
+++ b/qt/aqt/browser/sidebar/tree.py
@@ -222,16 +222,20 @@ class SidebarTreeView(QTreeView):
         model: SidebarModel,
         parent: Optional[QModelIndex] = None,
         searching: bool = False,
+        scrolled_to_first_search_match: bool = False,
     ) -> None:
         parent = parent or QModelIndex()
         for row in range(model.rowCount(parent)):
             idx = model.index(row, 0, parent)
             if not idx.isValid():
                 continue
-            self._expand_where_necessary(model, idx, searching)
+            self._expand_where_necessary(model, idx, searching, scrolled_to_first_search_match)
             if item := model.item_for_index(idx):
                 if item.show_expanded(searching):
                     self.setExpanded(idx, True)
+                if item.is_highlighted() and not scrolled_to_first_search_match:
+                    self.restore_current(item)
+                    scrolled_to_first_search_match = True

     def update_search(
         self,
```

dae:
Presumably we only want to do this when searching though, and not refreshing? Maybe one option would be to invert the bool, then search_for() could pass in eg scroll_to_first_match=True, and it could be set to False once scrolling is done

abdnh:
I guess the is_highlighted() call makes checking `searching` unnecessary since is_highlighted() will only return true when searching?
The code never runs when refreshing as far as I can tell.

I just noticed that my previous diff doesn't work as intended though: it results in excessive scrolling. We probably need to make _expand_where_necessary return the updated scroll_to_first_match so that subsequent checks don't result in more scrolling if we want to keep the state local (returning scroll state for a method called _expand_where_necessary is a bit inelegant, but maybe it's better than using an instance variable).

```diff
--- a/qt/aqt/browser/sidebar/tree.py
+++ b/qt/aqt/browser/sidebar/tree.py
@@ -215,23 +215,32 @@ class SidebarTreeView(QTreeView):
         # start from a collapsed state, as it's faster
         self.collapseAll()
         self.setColumnHidden(0, not self.model().search(text))
-        self._expand_where_necessary(self.model(), searching=True)
+        self._expand_where_necessary(
+            self.model(), searching=True, scroll_to_first_match=True
+        )

     def _expand_where_necessary(
         self,
         model: SidebarModel,
         parent: Optional[QModelIndex] = None,
         searching: bool = False,
-    ) -> None:
+        scroll_to_first_match: bool = False,
+    ) -> bool:
         parent = parent or QModelIndex()
         for row in range(model.rowCount(parent)):
             idx = model.index(row, 0, parent)
             if not idx.isValid():
                 continue
-            self._expand_where_necessary(model, idx, searching)
+            scroll_to_first_match = self._expand_where_necessary(
+                model, idx, searching, scroll_to_first_match
+            )
             if item := model.item_for_index(idx):
                 if item.show_expanded(searching):
                     self.setExpanded(idx, True)
+                if item.is_highlighted() and scroll_to_first_match:
+                    self.restore_current(item)
+                    scroll_to_first_match = False
+        return scroll_to_first_match

     def update_search(
         self,
```

RumovZ:
Good idea, this PR!

`restore_current()` performs an expensive recursive search, even though we already know the item and its index, and it changes the current element. So I'd just use `scrollTo(idx)`.

abdnh:
I think changing the current element too is desirable for quick keyboard navigation. But good point about the search in `restore_current()`.

dae:
I've made a minor tweak that lets us avoid having to return a bool. Thanks for this Abdo, and thanks for the suggestion Rumo.

PR #1260, opened 2021-06-28
use css columns instead of grid

dae:
Alternative approach to #1255

@kleinerpirat @hgiesel what do you think of this approach instead? The only issue I noticed is that the section on the top right doesn't have the same top margin/padding as the one on the left.

Screen Shot 2021-06-28 at 6 07 49 pm

### DIFF
--- a/ts/deckoptions/DeckOptionsPage.svelte
+++ b/ts/deckoptions/DeckOptionsPage.svelte
@@ -57,37 +57,57 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
 <ConfigSelector {state} />
 
-<Container api={options} class="g-1">
-    <Item>
-        <DailyLimits {state} api={dailyLimits} />
-    </Item>
-
-    <Item>
-        <NewOptions {state} api={newOptions} />
-    </Item>
-    <Item>
-        <LapseOptions {state} api={lapseOptions} />
-    </Item>
-    <Item>
-        <BuryOptions {state} api={buryOptions} />
-    </Item>
-
-    {#if state.v3Scheduler}
+<Container api={options} class="g-1 mb-3 mt-3">
+    <div class="multi-column">
         <Item>
-            <DisplayOrder {state} api={displayOrder} />
+            <DailyLimits {state} api={dailyLimits} />
         </Item>
-    {/if}
 
-    <Item>
-        <TimerOptions {state} api={timerOptions} />
-    </Item>
-    <Item>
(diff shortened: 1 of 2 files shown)


kleinerpirat:
Feeling pretty blockheaded right now. This approach is so much simpler and there is zero waste of space.
If we could fix the following two things, I would be very happy about your approach:

- For some odd reason, the padding of the first element on the right is not applied correctly:
image

- The increasing h1 scaling with window width still bugs me, because why would the titles need to get bigger, if the columns are essentially the same width as if the window was scaled to a mobile aspect ratio?

Thanks @dae for working on this! I know you got your hands full right now, so I really appreciate you taking the time to improve the visuals.

hikaru-y:
> The only issue I noticed is that the section on the top right doesn't have the same top margin/padding as the one on the left.

I think `display: inline-block;` would do the trick. This would increase the spacing between the sections a bit, so you might want to adjust it with margins or padding.

```diff
--- a/ts/deckoptions/TitledContainer.svelte
+++ b/ts/deckoptions/TitledContainer.svelte
@@ -24,5 +24,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html

     .container-fluid {
         break-inside: avoid;
+        display: inline-block;
     }
 </style>
```

kleinerpirat:
> This would increase the spacing between the sections a bit, so you might want to adjust it with margins or padding.

Nice one @hikaru-y . I feel like this kills two birds with one stone. I was already tempted to comment that the options look a little bit too tightly packed on a small window.

hgiesel:
`container-fluid` is a Bootstrap class, so I feel a bit queazy about redefining it as `inline-block`, when Bootstrap defines it as `flex`.
This is a alternative solution, that just adjusts the margins a bit:
https://github.com/ankitects/anki/commit/8475f19d09a03eb1ef028e7ec7f2b7764bf4bc39

Screenshot 2021-06-29 at 02 55 11

dae:
thanks @hikaru-y and @hgiesel! I had to add a separate mb-3 as well, as otherwise the bottom is squished up against the window edges. For some reason my-3 didn't work instead. Also increased the bottom margin of the titles so they're less squished out, like in hkr's approach.

@kleinerpirat the header resizing seems to be a standard feature of bootstrap 5: https://getbootstrap.com/docs/5.0/getting-started/rfs/

dae:
Will merge this in for now, anyone please feel free to open a follow-up PR if you think something needs tweaking

PR #1261, opened 2021-06-28
Fix id check for `today` and `card state` items

RumovZ:
Some more missing cases for the item comparison. But that should finally be it.


dae:
Thanks as always Rumo!

PR #1262, opened 2021-06-29
Make input in TextInputModal in dark mode visible

hgiesel:


dae:
Great, thanks for that :-)

PR #1265, opened 2021-06-30
Refactor WithDropdown to be more versatile and cause less bugs

hgiesel:
The JS component of Bootstrap dropdowns will dynamically set the class `show` on both the toggle and the dropdown menu itself.
Now if we set the classes on the toggle ourself via Svelte, like this:

```js
<IconButton class={`foo bar ${conditionalClass}`} on:mount={...createDropdown...}>
    ...
</IconButton>
```

A change in `conditionalClass`, might overwrite Bootstrap `show` class on the toggle, causing a dropdown to be unclosable, which is the bug I reported in #1264.

The solution is to use a common parent:
```js
<div class:conditionalClass>
    <IconButton class="foo bar" on:mount={...createDropdown...}>
         ...
    </IconButton>
</div>
```

I've looked at some Bootstrap Svelte component libraries, and they typically only use the CSS part, and recreate the Bootstrap JS with Svelte logic. We might consider doing that at some point as well.


dae:
Thanks Henrik, hopefully this is less brittle moving forward.

PR #1266, opened 2021-06-30
More uniform file naming

hgiesel:
kebap-case for .ts, .scss, .html, and directories.

I don't really care whether JS/TS files use camelCase, or kebap-case, I just care that it's uniform :)


dae:
Yeah, when I used ChangeNotetype I thought we'd need to make a change down the line to make things consistent ^_^; It would be nice if there was one clear direction here! A brief survey of some of our dependencies:

svelte: snake_case
d3: camelCase
typescript: camelCase
lodash: camelCase
eslint: kebab-case
yarn: kebab-case

One argument against kebab-case is that Bazel's Python support [can choke on the hyphens](https://github.com/bazelbuild/bazel/issues/6841#issuecomment-485883114), but we probably
won't need to place Python libraries inside /ts, and html-filter is working fine already, so happy to merge this once the conflicts are resolved.

hgiesel:
One thing I should also mention, that at first I tried to use camelCase for all JS/TS files, but setting
```
esbuild(
    name = "deck-options",
    ....
    output = "deckOptions.js",
    output_css = "deck-options.css",
    ...
)
```

failed, and could only be fixed by setting `output` and `output_css` to the same value.

dae:
Thanks Henrik!

PR #1267, opened 2021-07-01
Use better cloze button icon

hgiesel:
While looking through the icons yesterday, I also found a an icon, that includes brackets with an ellipsis.

image


dae:
Looks good!

PR #1270, opened 2021-07-02
Add flag manager and hook

RumovZ:
The former approach didn't allow for listening to flag label changes, which some add-ons like Advanced Browser probably want to do, and relied on reloading the flag object list all the time.

This approach caches the list and notifies about changes with a new hook. I've taken inspiration from collection objects like CardManager, only that this manager needs to be in aqt. Please let me know if there's a better way to do this.


dae:
LGTM - thanks Rumo!

PR #1271, opened 2021-07-02
Preload answer images

hikaru-y:
Related thread on the forum: [Pre-load the back side of the card](https://forums.ankiweb.net/t/pre-load-the-back-side-of-the-card/9865)

This PR is based on my posts in the above thread.

If you look at the `Network` tab in the *chrome developer tools* while reviewing a card, you'll see that while the question is being displayed, the images that will be displayed on the answer side are being loaded. I think it is easy to see the effect of preloading when scrolling to the `#answer` element occurs. Without preloading, there is sometimes a momentary lag before scrolling, but with preloading, there seems to be no lag at all.   If you don't see any effect, please make sure that the `Disable cache` option in the network tab is unchecked, otherwise you won't be able to see the effect.

I think that preloading images would be useful for users who use decks that have a lot of images on the back, such as some pre-made decks for med students. Ideally, it would be nice to be able to preload the images for the questions on the next card as well, but I think even just preloading the answer images would improve the UX a bit.

### DIFF
--- a/qt/aqt/browser/previewer.py
+++ b/qt/aqt/browser/previewer.py
@@ -171,6 +171,7 @@ class Previewer(QDialog):
             return
         c = self.card()
         func = "_showQuestion"
+        ans_txt = ""
         if not c:
             txt = tr.qt_misc_please_select_1_card()
             bodyclass = ""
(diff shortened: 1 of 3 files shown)


hikaru-y:
My apologies, I forgot to edit `browser/previewer.py`. I edited it and have force-pushed the PR branch to include the changes.

qt/aqt/browser/previewer.py
```diff
@@ -218,7 +220,14 @@ def _render_scheduled(self) -> None:
             txt = self.mw.prepare_card_text_for_display(txt)
             txt = gui_hooks.card_will_show(txt, c, f"preview{self._state.capitalize()}")
             self._last_state = self._state_and_mod()
-        self._web.eval(f"{func}({json.dumps(txt)},'{bodyclass}');")
+
+        js: str
```
dae, line 224: This doesn't do any harm, but for future reference, it's not necessary to pre-declare variables in Python.

dae (commented):
Thanks Hikaru, this should be a nice improvement!

PR #1272, opened 2021-07-02
Move selection into editable, if it's not there after focus

hgiesel:
I recently noticed that there is still a focus issue (our old favorite issue), when the Editable contains  mostly block elements (e.g. just a single `<div>`)), and you click the field.

The blue highlight will show, but the caret will not be inside the editor field. This seems to have been the case ever since 972993b42.


dae:
Thanks for staying on top of these!

PR #1273, opened 2021-07-02
Execute Ctrl+C/V/X/A via execCommand in editor

hgiesel:
Right now, Cmd+C/V/X/A are captured in AnkiWebViews on Macs so that we cannot modify it from within the AnkiWebView.

This is related to #1264: You can select tags via Ctrl/Shift - Click and then I would like them to be copyable via Control+C, and selectable via Control+A (and maybe overwritable via Control+V), to make the interaction with tags as natural and easy as possible.

I personally could not notice any difference, once I registered the shortcuts directly in JS via `execCommand`.


dae:
Thanks - some testing seems to indicate they're no longer required, so I've removed them completely.

PR #1275, opened 2021-07-03
Add `config_will_get` hook

RumovZ:
For my work on advanced browser, I had to monkey patch whole classes just to use a separate config key. For literal keys like "activeCols", that could be avoided if these were stored as variables in `consts.py`, e.g. (And this might be a good idea, anyway.) But it wouldn't help with enum keys (string and bool), because add-on authors can't overwrite those.
This PR would add an easy way to replace arbitrary config values. If you approve, I'll go ahead and add the counterpart for `set_config()`.

### DIFF
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -82,6 +82,15 @@ class LegacyCheckpoint:
 LegacyUndoResult = Union[None, LegacyCheckpoint, LegacyReviewUndo]
 
 
+@dataclass
+class ConfigContext:
+    key: Optional[str] = None
+    bool_key: Optional[Config.Bool.Key.V] = None
+    string_key: Optional[Config.String.Key.V] = None
+    value: Any = None
+    default: Any = None
+
+
 class Collection(DeprecatedNamesMixin):
     sched: Union[V1Scheduler, V2Scheduler, V3Scheduler]
 
@@ -703,11 +712,21 @@ class Collection(DeprecatedNamesMixin):
     # Config
     ##########################################################################
 
-    def get_config(self, key: str, default: Any = None) -> Any:
+    def _get_config(self, context: ConfigContext) -> Any:
+        hooks.config_will_get(context)
+        if context.value is not None:
+            return context.value
+        if context.bool_key is not None:
+            return self._backend.get_config_bool(context.bool_key)
+        if context.string_key is not None:
+            return self._backend.get_config_string(context.string_key)
         try:
-            return self.conf.get_immutable(key)
+            return self.conf.get_immutable(context.key)
         except KeyError:
-            return default
+            return context.default
+
+    def get_config(self, key: str, default: Any = None) -> Any:
+        return self._get_config(ConfigContext(key=key, default=default))
 
     def set_config(self, key: str, val: Any, *, undoable: bool = False) -> OpChanges:
         """Set a single config variable to any JSON-serializable value. The config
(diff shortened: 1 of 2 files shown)


dae:
Hi Rumo,

Generally speaking, hooks that can be used in a number of different cases are nice, as it allows us to avoid having to add separate hooks for lots of individual cases. But I'm not sure that's the best approach here: quite a few config entries are referenced directly in Rust, so altering the key that is read or written in Python won't alter their behaviour, so I fear this hook may lead to confusion. NORMALIZE_NOTE_TEXT that you used in your example is one such instance of this.

Maybe in this case a hook specific to the column rendering would be more appropriate? Is it just activeCols/activeNoteCols that you're trying to override, or are there other instances that need to be handled as well? Do we even need a hook? If "activeCols" was a class variable, the advanced browser add-on would be able to overwrite it with a different string when it initialises, and the code changes would be fairly minimal.

RumovZ:
> quite a few config entries are referenced directly in Rust

Oh, good point, I didn't think of that.

> If "activeCols" was a class variable, the advanced browser add-on would be able to overwrite it with a different string when it initialises, and the code changes would be fairly minimal.

That's what I meant above. In my case, I want to override "activeCols", "sortType" and `Config.Bool.BROWSER_SORT_BACKWARDS` (and the notes mode equivalents). The first two could be easily replaced if the keys weren't used literally, but defined as constants somewhere. However, that's not possible for the boolean key, because I can't add my own. I'd need a way to replace it with a literal key.

Maybe then I'll just replace these config values with my own on startup, and restore the original ones when shutting down.

dae:
IIRC we kept "activeCols" defined on the Python side instead of Rust so that we'd have flexibility here. I think we should probably just do the same for BROWSER_SORT_BACKWARDS - removing it from ConfigBool and referencing it as a string like the other two. I'm guessing that would solve the problem?

RumovZ:
> IIRC we kept "activeCols" defined on the Python side instead of Rust so that we'd have flexibility here.

I thought it was because Anki has only key enums for boolean and string values at the moment.

Replacing the ConfigBool with a string key would solve it for me, but I can make it work without making changes to Anki's codebase. My intention was to create an easy, universal way of using custom config values, so add-on authors aren't tempted to just use Anki's own keys, which can be problematic, if values are stored, that are not valid in vanilla Anki (like custom browser columns).
I'm not in favour of making changes for the benefit of a single add-on.

Another idea: What if I added a routine that can be passed any key and value, replaces the stored value and takes care of restoring the original values before Anki is shut down? Would that work?
If not, we'll just have to keep handling it on the add-on side.

dae:
I'd prefer to see some more examples of other add-ons that need to override config vars before we start thinking about a general solution: https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming). I feel like mutating config is approaching things from the wrong end, and areas that add-ons would like to customize are better handled by hooks or passing in the desired options, instead of having them read from config.

I don't mind the sort backwards config bool being removed, as it would make things more consistent. But either removing or keeping it works for me.

Modifying the DB is inadvisable - apart from it being a hack, an unclean shutdown would result in bad data remaining in the collection.

RumovZ:
Ok, I'll refrain from trying to find a general solution for now.

> Modifying the DB is inadvisable - apart from it being a hack, an unclean shutdown would result in bad data remaining in the collection.

That's how advanced browser has handled it before, by the way. But I understand that higher standards apply for code in vanilla Anki.

PR #1276, opened 2021-07-04
Follow-up to #1271 (Preload answer images)

hikaru-y:
After the commit 14110ad, the following message is shown in the console log:
`qt\aqt\reviewer.py:341:escape_media_filenames() is deprecated: no longer required`


dae:
Sorry, I overlooked that call. Thanks for this, and for the tidyups!

PR #1277, opened 2021-07-04
Add constants for literal config keys

RumovZ:
Not sure where the assignments should go. I've put them in the classes that use them, but now something like "sortBackwards" is redeclared in `Collection._build_sort_mode()` because pylib can't access `CardState`. It doesn't matter in this case, but looks bad.
Should I put them all in `consts.py`? That means we have to do more namespacing, though. Maybe something like below?

```py
class ConfigKey:
    class Browser:
        class Cards:
            SORT_BACKWARDS = "sortBackwards"
            ...
        class Notes:
            SORT_BACKWARDS = "browserNoteSortBackwards"
            ...
        ...
    ...
```

### DIFF
--- a/pylib/anki/browser.py
+++ b/pylib/anki/browser.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+# Copyright: Ankitects Pty Ltd and contributors
+# License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+
+
+class BrowserConfig:
+    ACTIVE_CARD_COLUMNS_KEY = "activeCols"
+    ACTIVE_NOTE_COLUMNS_KEY = "activeNoteCols"
+    CARDS_SORT_COLUMN_KEY = "sortType"
+    NOTES_SORT_COLUMN_KEY = "noteSortType"
+    CARDS_SORT_BACKWARDS_KEY = "sortBackwards"
+    NOTES_SORT_BACKWARDS_KEY = "browserNoteSortBackwards"
+
+    @staticmethod
+    def active_columns_key(is_notes_mode: bool) -> str:
+        if is_notes_mode:
+            return BrowserConfig.ACTIVE_NOTE_COLUMNS_KEY
+        return BrowserConfig.ACTIVE_CARD_COLUMNS_KEY
+
+    @staticmethod
+    def sort_column_key(is_notes_mode: bool) -> str:
+        if is_notes_mode:
+            return BrowserConfig.NOTES_SORT_COLUMN_KEY
+        return BrowserConfig.CARDS_SORT_COLUMN_KEY
+
+    @staticmethod
+    def sort_backwards_key(is_notes_mode: bool) -> str:
+        if is_notes_mode:
+            return BrowserConfig.NOTES_SORT_BACKWARDS_KEY
+        return BrowserConfig.CARDS_SORT_BACKWARDS_KEY
+
+
+class BrowserDefaults:
+    CARD_COLUMNS = ["noteFld", "template", "cardDue", "deck"]
+    NOTE_COLUMNS = ["noteFld", "note", "noteCards", "noteTags"]
--- a/qt/aqt/browser/table/state.py
+++ b/qt/aqt/browser/table/state.py
@@ -6,21 +6,24 @@ from __future__ import annotations
 from abc import ABC, abstractmethod, abstractproperty
 from typing import List, Sequence, Union, cast
 
+from anki.browser import BrowserConfig
 from anki.cards import Card, CardId
-from anki.collection import Collection, Config
+from anki.collection import Collection
 from anki.notes import Note, NoteId
 from anki.utils import ids2str
 from aqt.browser.table import Column, ItemId, ItemList
 
 
 class ItemState(ABC):
-    config_key_prefix: str
+    GEOMETRY_KEY_PREFIX: str
+    SORT_COLUMN_KEY: str
+    SORT_BACKWARDS_KEY: str
     _active_columns: List[str]
-    _sort_column: str
-    _sort_backwards: bool
 
     def __init__(self, col: Collection) -> None:
         self.col = col
+        self._sort_column = self.col.get_config(self.SORT_COLUMN_KEY)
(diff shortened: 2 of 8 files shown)


rslib/backend.proto
```diff
@@ -1447,8 +1447,6 @@ message Config {
   message Bool {
     enum Key {
       BROWSER_TABLE_SHOW_NOTES_MODE = 0;
-      BROWSER_SORT_BACKWARDS = 1;
-      BROWSER_NOTE_SORT_BACKWARDS = 2;
```
RumovZ, line 1451: I've wondered before: Do I have to fix the numbers when reordering or removing variants? As far as I understand it, the only benefit is that it looks neater, and in other applications, where pb messages are exchanged between clients with possibly different versions, it would even be vital *not* to change the numbers.

dae: Currently the backend proto is considered internal, so we're free to reorder things as we please, but you're right this is something we shouldn't be doing if it were used in a network service.

If you'd like to avoid gaps without renumbering everything, one option is to move the tail elements in their place (16/17 → 1/2, sort of like https://doc.rust-lang.org/std/vec/struct.Vec.html#method.swap_remove)

dae (commented):
Instead of catch-alls like collection or consts.py, how about using a separate file for the namespace? We could create a  pylib/anki/browser.py that contains a BrowserConfig class that can be imported, and then change

self.BROWSER_ACTIVE_CARD_COLUMNS_KEY

to eg

BrowserConfig.ACTIVE_CARD_COLUMNS_KEY

The constants at the top of CardState are ok I think, but perhaps instead of duplicate strings, they could reference BrowserConfig? Eg

```
class CardState(ItemState):
    GEOMETRY_KEY_PREFIX = "editor"
    SORT_COLUMN_KEY = BrowserConfig.SORT_CARDS_COLUMN_KEY
    SORT_BACKWARDS_KEY = BrowserConfig.SORT_CARDS_BACKWARDS_KEY
```

WDYT?

rslib/backend.proto
RumovZ: If you don't mind the gaps, I don't, either. 🙂

RumovZ:
I've added some helpers, which allow us to pull more code into ItemState. A dict with a `True` and a `False` key would have also been possible, but I dislike guessing which key means which mode.

qt/aqt/browser/table/state.py
```diff
@@ -6,21 +6,25 @@
...
-    config_key_prefix: str
     _active_columns: List[str]
-    _sort_column: str
-    _sort_backwards: bool
 
     def __init__(self, col: Collection) -> None:
+        notes_mode = self.is_notes_mode()
+        self.GEOMETRY_KEY_PREFIX = BrowserConfig.geometry_key_prefix(notes_mode)
+        self.SORT_COLUMN_KEY = BrowserConfig.sort_column_key(notes_mode)
+        self.SORT_BACKWARDS_KEY = BrowserConfig.sort_backwards_key(notes_mode)
```
dae, line 24: It feels a bit strange to be setting constants in an init method. If you made is_notes_mode() a  classmethod,  and moved the staticmethods from BrowserConfig into ItemState, making them also classmethods, then you could avoid having to pass in the bool each time, and we wouldn't need extra work in init. eg:

```
class ItemState:
    @classmethod
    def is_notes_mode() -> bool:
        # to be defined by subclasses
        assert False

    @classmethod
    def sort_key(cls) -> str:
        if cls.is_notes_mode():
            return "one"
        else:
            return "two"
```

Then we could call sort_key() on an instance of CardState or NoteState to get the correct key, and if add-ons modified BrowserConfig, those changes would be reflected in the returned values.

This approach of handling both cases in ItemState does feel a little atypical of OOP however - I think a more standard solution would be to override sort_key() in CardState and NoteState, instead of pushing that logic into the parent class.

pylib/anki/browser.py
```diff
@@ -0,0 +1,43 @@
...
+
+class BrowserConfig:
+    ACTIVE_CARD_COLUMNS_KEY = "activeCols"
+    ACTIVE_NOTE_COLUMNS_KEY = "activeNoteCols"
+    CARDS_SORT_COLUMN_KEY = "sortType"
+    NOTES_SORT_COLUMN_KEY = "noteSortType"
+    CARDS_SORT_BACKWARDS_KEY = "sortBackwards"
+    NOTES_SORT_BACKWARDS_KEY = "browserNoteSortBackwards"
+    CARDS_GEOMETRY_KEY_PREFIX = "editor"
+    NOTES_GEOMETRY_KEY_PREFIX = "editorNotesMode"
```
dae, line 14: The geom keys are Qt specific and unlikely to be accessed elsewhere, so you could store them directly on CardState/NoteState ... but not a big deal either way.

qt/aqt/browser/table/state.py
dae: My earlier example of assigning directly to the constants instead of using a classmethod would have been even more straightforward - was there a problem with doing it that way that I overlooked? You could declare the same constants in ItemState with a dummy value and then set them correctly in CardState/NoteState

RumovZ: My reasoning was, the more we handle in ItemState, the less duplicated code and the easier to read it. But I'll use your suggestion with the dummy variables. That means declaring the constants thrice, but at least we can keep the properties in ItemState.

> If you made is_notes_mode() a classmethod, and moved the staticmethods from BrowserConfig into ItemState, making them also classmethods

That wouldn't work, though, because the helpers are also used by `Collection._build_sort_mode()`, which is independent of any state.
I've left the helpers for now, but I can remove them if you want.

RumovZ: Just realised I don't even need dummies, if I'm not mistaken.

rslib/backend.proto
dae: They're a minor irritation, but we can always fill them when adding future entries. :-)

qt/aqt/browser/table/state.py
dae: Thanks, I hadn't noticed _build_sort_mode. And just declaring the variables without dummies is nicer. Thanks Rumo!

PR #1278, opened 2021-07-04
Prevent #dupe bgcolor from being carried over when copying text

hikaru-y:
This is a workaround for the issue reported in the forum:
[21](https://forums.ankiweb.net/t/anki-2-1-45-alpha/10061/21), [31](https://forums.ankiweb.net/t/anki-2-1-45-alpha/10061/31) @ Anki 2.1.45 Alpha

This issue only occurs in day mode, because `stylingLightMode` includes `background-color`, while `stylingNightMode` does not.
https://github.com/ankitects/anki/blob/6ffa7440967ca537d721924474466f527815c572/ts/html-filter/styling.ts#L12-L22

### DIFF
--- a/ts/editor/fields.scss
+++ b/ts/editor/fields.scss
@@ -19,7 +19,7 @@
     background: var(--frame-bg);
 
     &.dupe {
-        background: var(--flag1-bg);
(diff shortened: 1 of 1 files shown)


dae:
Thanks Hikaru!

PR #1279, opened 2021-07-05
Remove disabled / Export via Components

hgiesel:
These are some changes I accumulated while writing an Anki add-on for the editor toolbar.

First of all, I think we should probably expose components over a more general variable name like `components`. Right now we expose them as editorToolbar.component, however let's say we have two UI elements, like a toolbar, and a main element, which both have a component in common. We'd might end up exposing `editorToolbar.component` and `mainElement.component`. We expose them now over `Components.svelte`, which we could also use for the deck options.

Second change is renaming/removing `disabled`. When first implemented, it still made sense, but now it's a bit confusing, which is why I named it `fieldFocused`.

I've also made a small change to ButtonGroup. Sometimes the button group got its item mixed up, which is why I indexed them.

### DIFF
--- a/ts/components/ButtonGroup.svelte
+++ b/ts/components/ButtonGroup.svelte
@@ -41,15 +41,17 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         makeInterface(makeRegistration);
 
     $: for (const [index, item] of $items.entries()) {
-        if ($items.length === 1) {
-            item.position.set(ButtonPosition.Standalone);
-        } else if (index === 0) {
-            item.position.set(ButtonPosition.Leftmost);
-        } else if (index === $items.length - 1) {
-            item.position.set(ButtonPosition.Rightmost);
-        } else {
-            item.position.set(ButtonPosition.Center);
-        }
+        item.position.update(() => {
+            if ($items.length === 1) {
+                return ButtonPosition.Standalone;
+            } else if (index === 0) {
+                return ButtonPosition.Leftmost;
+            } else if (index === $items.length - 1) {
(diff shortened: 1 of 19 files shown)


hgiesel:
I also wanted to use `WithShortcut` in my demo add-on, but it seems like `on:mount` or generally `onMount` is not fired for our dynamically added `ButtonGroupItem`s.

[Here](https://github.com/hgiesel/anki_new_format_pack/blob/572ffca6f174fa1951121f15c3434bcf1cc3140f/web/Paragraph.svelte#L57)'s how I would have used it. Alternatively I wanted to use `bind:this` on `IconButton`, but that also did not work. It almost seems like we uncovered an Svelte bug... For now I left `WithShortcut` unexported in this PR.

dae:
Seem like sensible changes, thanks. 

onMount seems to be called on https://svelte.dev/repl/16b375da9b39417dae837b5006799cb4?version=3.25.0 when I replace the contents of Thing.svelte with the following, so it seems to work in the simple case, and perhaps is a bug:

```html
<script>
	import { onMount } from 'svelte';
	
	onMount(() => {
		console.log("onMount of thing called");
	});
	
	export let answer;
</script>

<p>the answer is {answer}</p>
<slot></slot>
```

hgiesel:
I've tracked down the issue that I mentioned. It seems to be specific to combing Svelte from two compilation units. I've made an example repo [here](https://github.com/hgiesel/svelte-bug-example), and opened a bug report [here](https://github.com/sveltejs/svelte/issues/6530). In the repo, I demonstrated the bug with bindings, but custom events do not work either.

dae:
Thanks for that, let's hope it's not something they consider out of scope :-)

PR #1280, opened 2021-07-06
Export via components for Deck Options

hgiesel:
I guess in the future, we will always want to export our globals under anki namespace, which is why I included  an assignment in `editor/index_wrapper.ts`.

I also wanted to export `components/Item.svelte` from `deck-options/index.ts` for `components`, but couldn't get it to work. Just trying to add an import from `ts/components` caused a bug (which is just like in editor), but when I tried a hack via `Components.svelte`, like I used in #1279, it wanted me to use a default import. Happy for any pointers, `Item.svelte` is technically not necessary to use within dynamic components, but for future guarantees it would be better.

### DIFF
--- a/ts/deck-options/DeckOptionsPage.svelte
+++ b/ts/deck-options/DeckOptionsPage.svelte
@@ -57,8 +57,8 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
 <ConfigSelector {state} />
 
-<Container api={options} class="g-1 mb-3 mt-3">
-    <div class="multi-column">
+<div class="multi-column">
+    <Container api={options} class="g-1 mb-3 mt-3">
         <Item>
             <DailyLimits {state} api={dailyLimits} />
         </Item>
@@ -96,17 +96,17 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         <Item>
             <AdvancedOptions {state} api={advancedOptions} />
         </Item>
-    </div>
-</Container>
+    </Container>
+</div>
 
 <style lang="scss">
-    .multi-column {
+    .multi-column :global(.container) {
         column-count: 2;
         column-gap: 5em;
     }
 
     @media (max-width: 1000px) {
-        .multi-column {
+        .multi-column :global(.container) {
             column-count: 1;
         }
     }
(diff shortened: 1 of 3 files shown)


dae:
This works:

```diff
diff --git a/ts/deck-options/BUILD.bazel b/ts/deck-options/BUILD.bazel
index 6703cb8b5..9cc2fa1cf 100644
--- a/ts/deck-options/BUILD.bazel
+++ b/ts/deck-options/BUILD.bazel
@@ -26,8 +26,8 @@ compile_svelte(
     name = "svelte",
     srcs = svelte_files,
     deps = [
-        "//ts/sveltelib",
         "//ts/components",
+        "//ts/sveltelib",
         "@npm//@types/bootstrap",
         "@npm//@types/marked",
     ],
@@ -48,8 +48,9 @@ ts_library(
     deps = [
         "DeckOptionsPage",
         "lib",
-        "//ts/lib",
         "//ts/components",
+        "//ts/components:Item",
+        "//ts/lib",
         "@npm//@popperjs",
         "@npm//svelte2tsx",
     ],
@@ -66,10 +67,10 @@ ts_library(
     module_name = "deckoptions",
     deps = [
         "//ts:image_module_support",
+        "//ts/components",
         "//ts/lib",
         "//ts/lib:backend_proto",
         "//ts/sveltelib",
-        "//ts/components",
         "@npm//lodash-es",
         "@npm//svelte",
     ],
diff --git a/ts/deck-options/index.ts b/ts/deck-options/index.ts
index 3d6794f89..a2af34dbb 100644
--- a/ts/deck-options/index.ts
+++ b/ts/deck-options/index.ts
@@ -49,6 +49,7 @@ import SpinBoxRow from "./SpinBoxRow.svelte";
 import SpinBoxFloatRow from "./SpinBoxFloatRow.svelte";
 import EnumSelectorRow from "./EnumSelectorRow.svelte";
 import SwitchRow from "./SwitchRow.svelte";
+import Item from "components/Item.svelte";
 
 export const components = {
     TitledContainer,
@@ -56,4 +57,5 @@ export const components = {
     SpinBoxFloatRow,
     EnumSelectorRow,
     SwitchRow,
+    Item,
 };
```

If we switched svelte_components from a filegroup to a ts_library presumably that would work too, but trying to do so at the moment throws up other errors:

```
bazel-out/darwin-fastbuild/bin/ts/components/Container.svelte.d.ts:2:76 - error TS2315: Type 'Section__SvelteComponent_' is not generic.

  2 declare const Container__SvelteComponent__base: SvelteComponentConstructor<Section<{
                                                                               ~~~~~~~~~
  3     id?: string;
    ~~~~~~~~~~~~~~~~
...
  9     default: {};
    ~~~~~~~~~~~~~~~~
 10 }>, Svelte2TsxComponentConstructorParameters<{
    ~~
```

Might be best to just hack in individual files for now, as presumably Svelte's typing will [improve over time](https://github.com/sveltejs/rfcs/pull/38#issuecomment-864577289)

PR #1281, opened 2021-07-07
Fix Ctrl-X behaviour in Codable

hikaru-y:
This fixes https://forums.ankiweb.net/t/anki-2-1-45-beta/10664/103 .


dae:
Thanks again hkr!

PR #1283, opened 2021-07-09
Add the CardOrdinal special field

abdnh:
I have a JavaScript script that pulls data from a service called [Youglish](https://youglish.com/) and need a way to get the current card ordinal to support kind of a custom implementation of the `cloze-only` filter that reveals the elided sections on both sides of the card. I initially wrote it as an add-on ([38866997](https://ankiweb.net/shared/info/38866997)) without problems, but then I wanted this to work on mobile, so I tried rewriting it in JavaScript.

This is how I would use the field in templates:
```html
<script src="_aglish.js"></script>
<div class="aglish">
	<div class="aglish-filter">lang=turkish clozeonly={{CardOrdinal}}</div>
	<div class="aglish-field">{{Text}}</div>
</div>
```

Hopefully, there is some way to get the card ordinal without adding a new special field that I overlooked (using the `c[n]` fields is not practical).

### DIFF
--- a/rslib/src/notetype/mod.rs
+++ b/rslib/src/notetype/mod.rs
@@ -64,6 +64,7 @@ lazy_static! {
         "Subdeck",
         "Tags",
         "Type",
+        "CardOrdinal",
     ]);
 }
 
--- a/rslib/src/notetype/render.rs
+++ b/rslib/src/notetype/render.rs
@@ -156,6 +156,8 @@ impl Collection {
             .or_insert_with(|| flag_name(card.flags).into());
         map.entry("Card")
             .or_insert_with(|| template.name.clone().into());
+        map.entry("CardOrdinal")
+            .or_insert_with(|| (card.template_idx + 1).to_string().into());
 
         Ok(())
     }


hgiesel:
On AnkiDesktop, you can already query the card ordinal, by looking at the class of the body tag, e.g. `card card3 isMac nightMode night_mode` would be card ordinal 3. Not sure whether this works on other platforms.

It would be great to have more more extra keywords to query card information, however I see here the same danger as in #1236, as in every new special field added will increase the risk of collisions with existing field names in templates.

abdnh:
> On AnkiDesktop, you can already query the card ordinal, by looking at the class of the body tag, e.g. `card card3 isMac nightMode night_mode` would be card ordinal 3. Not sure whether this works on other platforms.

Thanks! I was not aware of this. I checked AnkiDroid and it's supported there (not sure about AnkiMobile). This solves my problem, so I'm going to close this PR.

dae:
Thanks Henrik, that's a better workaround than the one I was thinking of with {{#c1}}/etc. I'd prefer not to add more special fields, but exposing extra details via a filter or some JS methods may come in the future.

PR #1284, opened 2021-07-09
Tweak deck options tooltips for RTL languages

abdnh:
Just some small tweaks.

The direction override on the code tag is because the `Deck` and `Position` labels in the Display Order section strangely run backward character-wise with RTL langs (maybe `unicode-bidi` is applied by default or something).

image


dae:
Thanks Abdo!

PR #1285, opened 2021-07-10
Persistent Editor / Reviewer as preview

hgiesel:
This could be seen as the first PR adressing the issue of "Previewing cards from the editor" (and thus is post v2.1.45 :) ), even though it does not add any Previewer component. It adds several features:

1. Flip cards during review with <kbd>f</kbd>. "f" previously was "Create filtered deck", which I put on "Ctrl+F". Other options would be "c", etc, but I thought the mnemonic was good.
2. Make <kbd>F5</kbd> update the current card, rather than just replay audio. "r" already replays audio, and F5 refreshing the card makes intuitive sense.
3. Make Edit Current update alongside the reviewer. Advancing in the reviewer will update Edit Current (this also makes it act more according to its name).
4. Make Reviewer live-update upon edits in Edit Current. I don't think NOTE_TEXT ops can lead the reviewer to an invalid state (?).
5. Add "obscure mode" to the Edit Current window. Obscure mode will:
    * Generally obscure fields, if edit current has no focus
    * While displaying the question/front, it will only unobscure the currently focused field.
6. Remove the Close button in EditCurrent, <kbd>Ctrl+Enter</kbd> will focus the reviewer, allow <kbd>Ctrl+W</kbd> to close the window.

There are two add-ons which currently implement something akin to this PR<sup>1</sup>:
1. My [Persistent Editor](https://ankiweb.net/shared/info/1686259334)
2. [Migaku Editor](https://ankiweb.net/shared/info/1051095155)

These changes combined enable a new way to use the Edit Current window: Rather than opening the editor, making the change, and closing, the user can alternatively leave the window open, focus it when he wants to edit the current card, and refocus the reviewer afterwards. This is really beneficial, if the user wants to make a lot of small changes during review.

I personally constantly edit during review, and [it seems I'm not alone with this](https://www.reddit.com/r/Anki/comments/ogxc4g/how_often_do_you_edit_cards_during_review/). Updates to the Undo mechanism, has made flipping cards back to the question side impossible (previously it was possible by opening and closing the editor, which I actually did a lot).

I think obscuring the fields upon opening the editor during question makes sense even, if he does not want to persist the window: Following scenario: The user sees an error in the question, wants to edit it immediately, however spoils himself because he sees the answer to the question in the Edit Current window.

<sup>1</sup>There are also several add-ons, which implement editing cards directly in the reviewer. The issue with "Editing in the reviewer" is, that it cannot work if the user uses any JavaScript that modifies field HTML, which is impossible to predict.

For future reference: there's actually a "blur" in mdicons.

### DIFF
--- a/qt/aqt/editcurrent.py
+++ b/qt/aqt/editcurrent.py
@@ -54,8 +54,7 @@ class EditCurrent(QDialog):
         QDialog.reject(self)
 
     def reopen(self, mw: aqt.AnkiQt) -> None:
-        if card := self.mw.reviewer.card:
-            self.editor.set_note(card.note())
+        self.editor.web.setFocus()
 
     def reject(self) -> None:
         self.saveAndClose()
--- a/qt/aqt/editcurrent.py
+++ b/qt/aqt/editcurrent.py
@@ -27,6 +27,11 @@ class EditCurrent(QDialog):
         self.editor = aqt.editor.Editor(self.mw, self.form.fieldsArea, self)
         self.editor.card = self.mw.reviewer.card
         self.editor.set_note(self.mw.reviewer.card.note(), focusTo=0)
+        self.set_card_format(self.mw.reviewer.state)
+        self.editor.web.eval(
+            '$editorToolbar.then(({ templateButtons }) => templateButtons.appendButton({ component: editorToolbar.ObscureButton })); '
+        )
+
         restoreGeom(self, "editcurrent")
         gui_hooks.operation_did_execute.append(self.on_operation_did_execute)
         gui_hooks.card_will_show.append(self.refresh_editor_on_review)
@@ -48,11 +53,15 @@ class EditCurrent(QDialog):
 
             self.editor.set_note(note)
 
+    def set_card_format(self, format: str) -> None:
+        self.editor.web.eval(f'setCardFormat("{format}"); ')
+
     def refresh_editor_on_review(self, qa: str, card, type: str) -> None:
-        if type.startswith("review") and (
-            not self.editor.card or self.editor.card.nid != card.nid
-        ):
-            self.editor.setNote(card.note())
+        if type.startswith("review"):
+            if not self.editor.card or self.editor.card.nid != card.nid:
+                self.editor.setNote(card.note())
+
+            self.set_card_format(type[len("review"):].lower())
 
         return qa
 
--- a/qt/aqt/editcurrent.py
+++ b/qt/aqt/editcurrent.py
@@ -49,8 +49,11 @@ class EditCurrent(QDialog):
             self.editor.set_note(note)
 
     def refresh_editor_on_review(self, qa: str, card, type: str) -> None:
-        if type.startswith("review"):
+        if type.startswith("review") and (
+            not self.editor.card or self.editor.card.nid != card.nid
+        ):
             self.editor.setNote(card.note())
+
         return qa
 
     def cleanup_and_close(self) -> None:
--- a/qt/aqt/editcurrent.py
+++ b/qt/aqt/editcurrent.py
@@ -3,6 +3,7 @@
 from typing import Optional
 
 import aqt.editor
+from anki.cards import Card
 from anki.collection import OpChanges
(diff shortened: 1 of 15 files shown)


dae:
Hi Henrik,

You can flip back to the front side with 's s' already, can't you? In the v3 scheduler you should be guaranteed to get the same card again IIRC. That's not necessarily an argument against adding a flip feature (people do ask for it from time to time), though we shouldn't be breaking people's muscle memory too lightly. Would you mind elaborating on the times you need to flip back to the front side?

I'm having trouble understanding the goal of 2 - if the card automatically redraws as you edit, why is a redraw command required?

I wouldn't be surprised if some prefer the current behaviour for 3, but you are right that it would better match its name, and with fields being hidden it is probably the better option.

The reason why I didn't do 4 as part of the undo changes is because:

- it causes the card's audio/video to replay every time you pause as you're typing
- cards with animations or heavy JS are not ideal to be reloaded after every half second break in typing

Do we really need it? What advantages does a live update bring?

5 is nice, and I don't think the automatic updating makes sense without it. It could lead some users to think Anki was "broken". How about instead of displaying the field as blank (as if it has no content), a subtle "click to show" message was shown overlaid on top?

It seems to have a race condition where the field contents briefly flashes when opening the window - only seems to happen occasionally.

6 sounds good

hgiesel:
1. I didn't know that "s s" had that behavior, but a standalone shortcut (without visual overhead) would be nice. I was maybe a bit too quick with changing the `f` shortcut, because I thought who'd wanna create filtered decks in the middle of review. Maybe we could make `f` still create filtered decks in deck browser / deck overview, but change in for the reviewer.

Some use cases for flipping:
- The front has fields that are not on the back.
- Clicking a shortcut to flip is easier than scrolling up to see the question.
- Some cards of a notetype have an identical backside, and can only be distinguished by their frontside. E.g. it contains multiple sections, and for each card, one of the fields is missing (e.g. Japanese / Pronunciation / Meaning). If you spend some time looking at the backside (maybe making a change in the EditCurrent) you might forget which card you were on, and how you wanted to rate. Taking a peek on the frontside can tell you again, which card you were dealing with.

2. The "refresh" functionality for me would be more important for the "previewer" aspect. Scenarios include:
- The card includes something randomly generated
- Something on the card template crashed, and you want to see if it happens every time
- In `clayout`, you can crash the previewer by messing with the DOM structure too much (or just `document.write`). A way to restart it here would be nice. And at that point, consistency across _card displayers_ would also be nice :)

4. Audio/Video being replayed should definitely be addressed, I have not considered this. The previewer does not show this behavior: Maybe this could be streamlined.

However for animations / heavy JS I'd argue that it's still better to show it. First of all, during the review process, those animations / heavy JS, will be replayed all the time anyway. If I take 1 ~ 2 seconds to review a card, it's not far from the level what an auto-update EditCurrent would do.

Also a lot of edits in the Edit Current for me end in the act of clicking the reviewer, making sure everything is ok, and then continue reviewing. This is especially true, if you anything to do with templates (clozes, furigana). If the reviewer auto-updates, on a sufficiently big screen, I would not need this extra action, until I can start "reviewing".

5. I agree. Maybe a crossed-out eye icon could also convey that message.

dae:
You've sold me on flipping, but I'm not sure what to do about the shortcut. 'f' would be simplest during review, but that means we either need to change the flip shortcut outside review for consistency, or have inconsistent shortcuts. Given that cmd+f on a Mac has a standard meaning, maybe we'd be better off with ctrl+shift+f for filtered decks? 

I seem to recall some users needing F5 for replaying audio, because they couldn't easy type 'r' with their keyboard layout, so I'm not sure we should be changing that shortcut unless we offer an alternative non-alpha key. I think people need to replay audio a lot more than they need to refresh cards (which seems like it would mainly be done when editing, not reviewing), so perhaps the priority should be on keeping the existing shortcut here. Do we even need to refresh during review?

Thanks for elaborating; if you can get audio/video replay handled like the preview screen so it doesn't repeat, then it sounds like we can give live updating a try.

dae:

PR #1288, opened 2021-07-12
Created shortcut to switch between card types on clayout.py

evandrocoan:
Allow selecting the next and previous fields from the keyboard. It helps quickly edit something in different cards.

### DIFF
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -185,6 +185,24 @@ class CardLayout(QDialog):
             self,
             activated=self.tform.style_button.click,
         )
+        QShortcut(  # type: ignore
+            QKeySequence("F3"),
+            self,
+            activated=lambda: (
+                self.update_current_ordinal_and_redraw(self.ord - 1)
+                if self.ord - 1 > -1
+                else None
+            ),
+        )
+        QShortcut(  # type: ignore
+            QKeySequence("F4"),
+            self,
+            activated=lambda: (
+                self.update_current_ordinal_and_redraw(self.ord + 1)
+                if self.ord + 1 < len(self.templates)
+                else None
+            ),
+        )
         for i in range(min(len(self.cloze_numbers), 9)):
             QShortcut(  # type: ignore
                 QKeySequence(f"Alt+{i+1}"),


dae:
Hi Evandro,

No objection to adding shortcuts for this, but I'm not sure ctrl+4/5 is the best choice. If we wanted to add a separate script tab in the future for example, we wouldn't be able to give it a logical shortcut that way. Can you think of some alternative keys? Maybe something like ctrl+shift+n/p?

dae:
I'm guessing you chose E and R because they fall together and can be pressed with one hand, rather than for their mnemonic value? That makes things a bit difficult for Dvorak users, as the keys are not in the same place, and not easy to remember. Dvorak users are in the minority though. Anyone want to chime in with thoughts?

evandrocoan:
Hi. I changed them to `Ctrl+E` and `Ctrl+R` because I can use them easier with a single hand they are quite close to `Ctrl+1`, `Ctrl+2`, and `Ctrl+3`. This way it is easier to batch change/switch between cards, front and back.

I was thinking about using `F3` and `F4` because I could easily use them with a single hand, but, doing combos with  `Ctrl+1`, `Ctrl+3`, and `Ctrl+3` would not be much nice, because I would have to leave and get the `Ctrl` key, and doing the combo `Ctrl+F3` or `Ctrl+F4` is not much nice because takes greater hand movement and `Ctrl+4`  is bound to close some applications like Google Chrome (the same as `Alt+F4` does). 

If you think `Ctrl+E`/`Ctrl+R` is not good, I prefer using `F3` and `F4` or even `Ctrl+F3`/`Ctrl+F4`. Other lower keys would not be much nice because they would be distant from `Ctrl+1`/`Ctrl+2`/`Ctrl+3`. Other keys as `Ctrl+R` and `Ctrl+T` could be used instead, but would not prefer `Ctrl+Q` and `Ctrl+W` because they are usually bound to close the application or window.

dae:
I'm not sure what the best choice is here - picking shortcuts that everyone likes is tricky. 😅 If you think F3/F4 will work, that's what I'd lean towards, as they're likely to be next to each other on almost all keymaps out there

evandrocoan:
After thinking, I updated them to F3 and F4 because on some keyboard layouts, E and R could be far from 1, 2, 3 numbers. But for good mnemonic, also having `Ctrl+N` and `Ctrl+P` could be good, so, we could have them both.

The F3 and F4 keys should be better than `Ctrl+F3/F4` because, on some keyboards, F3 and F4 can require an FN key combo (`FN+F3`, `FN+4`), and the combo `Ctrl+FN+F3/F4` would not be much nice to hit with the combos `Ctrl+1/2/3`.

dae:
Thanks Evandro!

PR #1290, opened 2021-07-12
Add mock browser model to support legacy resetting

RumovZ:
Should solve errors like this: https://forums.ankiweb.net/t/anki-2-1-45-beta/10664/123


dae:
Thanks Rumo!

PR #1291, opened 2021-07-12
Show key of unknown browser columns

RumovZ:
This should help users to understand where add-on columns are coming from. Also useful for debugging.


dae:
Makes sense, thanks.

Issue #1120, opened 2021-04-13
rework reviewer.ts


dae:
reviewer.ts could do with an update at one point, and it would be best if we can make any changes to it in one go, instead of trickling out breaking changes over time. This issue is for discussing and tracking what needs to be done.

Some things that should probably be addressed:

- [ ] as mentioned on #1119, we should probably remove the opacity changes and the promise race
- [ ] moving away from Python style _underscorePrefixed() functions, and instead explicitly exporting our public API. We could use an approach like editor/index_wrapper.ts, or maybe we should consider exporting them into a single global, like `anki` or `anki.reviewer`?
- [ ] _updateQA() and the hook system could use a rethink - we no longer need to fade in and out, and the current mix of promises and hooks is a bit tricky to follow - could it be simplified perhaps? onShownHook and onUpdateHook were exposed to allow templates to perform different actions before the content was revealed and after it was revealed, but without an opacity change, perhaps one of them is redundant?
- [ ] should we perhaps pass the MathJax and image promises to hooks in an object, so they can call .then() on them if they want to?
- [ ] @glutanimate has requested a permanent hook that doesn't get reset on each card load. The behaviour can be emulated by injecting the same hook each load via the Python API, but perhaps it's worth considering adding a separate long-lived one as well, which would get spread into the per-card one.

Thoughts? Any other things that could do with a change?

hgiesel:
1. If we just remove the opacity changes, simple MathJax like `\( \begin{array}{cc} a & b \\ c & d \end{array} \)` will already cause a flash of unrendered content for me.
I personally also have a horse in this game, as my add-on Closet also heavily builds on `onUpdatePromise`, and it operating on hidden content.

2. `onUpdatePromise` is useful for changing the content before it is shown, for anything template related. `onShownPromise` can be useful for showing any of animation on card shown, start a timer, etc.. I think both hooks still have their reason for existing.

hgiesel:
One way to get rid of the flash for most of the cards is, if we had some logic to *test* whether MathJax was used on the card, and only hide if they do. That's actually similar to how AnkiDroid does it.

dae:
Ah, scratch that idea then.

It would be great if we could do the rendering in the background, then atomically swap the old content with the new. That hasn't proved practical in the past due to concerns with the same id being used in the old and new content, and preloading too early resulting into html audio possibly autoplaying, but maybe it would be achievable if we used a shadow dom and preloaded right before updating the display?

[...]

PR #1292, opened 2021-07-12
Move reviewer to ts/reviewer v2

hgiesel:
This addresses #1120 by moving reviewer to ts/reviewer, just like #1008, but also tries to solve some issues in addition:

1. css_browser_selector, jquery and mathjax are imported directly via `ts/reviewer`.
2. This is especially a big change for MathJax, which we now use as a library. We directly parse the card html via MathJax, and insert the result on the page. This usage of MathJax is very sparsely documented, so I'm unsure I do everything correctly...
3. This means we can do it synchronously before even inserting the card HTML into the page, which means no need for opacity change, which means we can remove it :)

Recently there's been several reports, that cards do not show up during review (e.g. https://forums.ankiweb.net/t/space-bar-does-not-show-the-answer/11347). I still wasn't able to recreate it, but this might fix this, as the opacity change is gone.

This breaks setting MathJax configs via `onUpdate`. In fact, this breaks setting MathJax configs via the template JS altogether :( . IMO this PR is the way to go, but we'll need some way to execute custom JS either now, or in the future.


hgiesel:
Rebasing messed this one up - I'll update after I fixed it.

hgiesel:
Fixed ^^;

dae:
I presume you were intending this for after 2.1.45? If not, that MathJax change feels a bit too invasive to be slipping in at the moment, especially if we don't have a way for users to customize the MathJax anymore.

I seem to recall seeing users talking about resizing the window causing the answer side to appear, which seems to point to it being a redraw bug in QtWebEngine rather than a bug in our JS. Will follow up in #1120

hgiesel:
OK, I thought the reviewer being blank was a more pressing issue, which is why I considered breaking MathJax extensibility for the few in favour of the reviewing experience of the many.

dae:
I appreciate you trying to address the issue, but before we make that call, I think we had better a) make sure we understand why it's happening, and be sure such a change will fix it, and b) decide whether fixing an intermittent issue for the majority is worth completely breaking things for a minority. For a), maybe we can get some more data from users on what does/doesn't reveal the answer, or monkey-patch the handling in an add-on as a proof of concept.

hgiesel:
I've removed the MathJax changes from this PR in 272f2f7.
I think the remaining changes are less controversial.

dae:
Thanks Henrik. I'll mark _updateQA as exported, as unfortunately some add-ons are using it, and ideally we give add-on authors some time to migrate away from using it.

glutanimate:
> Thanks Henrik. I'll mark _updateQA as exported, as unfortunately some add-ons are using it, and ideally we give add-on authors some time to migrate away from using it.

Could we please do the same for `_queueAction`? At least until we've had a chance to revisit the reviewer API for add-ons (calling back to our original discussion that lead to some of the points in #1120)?

Otherwise this will introduce yet another regression to the AMBOSS add-on.

dae:
Done, please let us know if there's anything else that's been missed

glutanimate:
Thanks, appreciate it! Will keep a look-out.

PR #1295, opened 2021-07-15
scheduler: use deck config's initial ease in set_due_date

cyphar:
Previously we would just use 250% ease for any new card that had no
pre-configured ease, but this will result in decks that have
non-standard ease values to have "set due date" cards in them that don't
match. In order to make this somewhat more efficient, we cache
deckid->ease lookups during this operation.

Ref: <https://forums.ankiweb.net/t/set-due-date-doesnt-obey-default-ease-factor/9184>

### DIFF
--- a/rslib/src/scheduler/reviews.rs
+++ b/rslib/src/scheduler/reviews.rs
@@ -1,6 +1,8 @@
 // Copyright: Ankitects Pty Ltd and contributors
 // License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
+use std::collections::HashMap;
+
 use lazy_static::lazy_static;
 use rand::distributions::{Distribution, Uniform};
 use regex::Regex;
@@ -9,7 +11,6 @@ use crate::{
     card::{Card, CardId, CardQueue, CardType},
     collection::Collection,
     config::StringKey,
-    deckconfig::INITIAL_EASE_FACTOR_THOUSANDS,
     error::Result,
     prelude::*,
 };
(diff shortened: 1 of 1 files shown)


rslib/src/scheduler/reviews.rs
```diff
@@ -100,12 +109,33 @@ impl Collection {
...
         self.transact(Op::SetDueDate, |col| {
             col.storage.set_search_table_to_card_ids(cids, false)?;
             for mut card in col.storage.all_searched_cards()? {
+                let deck_id = card.original_deck_id.or(card.deck_id);
+                let ease_factor = match decks_initial_ease.get(&deck_id) {
+                    Some(ease) => *ease,
+                    None => {
+                        let config_id = col
+                            .get_deck(deck_id)?
+                            .ok_or(AnkiError::NotFound)? // TODO: use the right error
```
cyphar, line 122: I wasn't quite sure what errors would be reasonable to return here -- should I just go with `NotFound` for all of them, or should I return some kind of string-based error in all of these cases?

dae: For issues that the DB check will clean up, a NotFound error is fine. IIRC, it handles missing decks and incorrect original_deck_ids, but does not handle missing deck configurations. We should probably address that in the future, but for now I'd recommend you just enable the fallback option on get_deck_config()

cyphar: Will do, thanks!

dae:
LGTM - happy to merge when you're ready.

cyphar:
I've tested this with filtered decks and regular decks (selecting cards from multiple original decks) and things appear to work correctly. Thanks.

dae:
Thank you!

PR #1296, opened 2021-07-16
Revert "Follow-up to #1271 (Preload answer images)"

hikaru-y:
Reverts ankitects/anki#1276 in accordance with e97c381


dae:
Thanks Hikaru, I'd forgotten about that :-)

PR #1297, opened 2021-07-17
Allow Using Ctrl+Shift+P from Browser when focusing Qt components

hgiesel:


dae:
I think we can skip the lambda if we add @no_arg_trigger to onTogglePreview

dae:
Thanks Henrik!

PR #1298, opened 2021-07-17
Fix browser search completer

RumovZ:
Adding an empty entry at the top fixes
https://forums.ankiweb.net/t/anki-2-1-45-beta/10664/45
and allows for scrolling back to an empty line.


dae:
Thanks!

PR #1300, opened 2021-07-17
Refactor editor fields

hgiesel:
This started as an attempt to implement something like [Multi-column note editor](https://ankiweb.net/shared/info/3491767031), or [Grid fields](https://ankiweb.net/shared/info/796923308) directly into Anki.

I wanted to implement this using Drag'n'Drop, however QWebView does not allow for `dragover` / `dragenter` events. Instead this features some other changes / improvements:

- Use `:host` instead of `:host-context`, as `:host-context` does not have wide support.
- Make StickyBar fixed rather sticky. Fixes an issue when an editor field contained an element wider than the viewport. It should also be renamed as the name is kinda misleading, but I'd rather do that after #1264.
- Make `#fields` a grid rather than flexbox

My end goal with this PR was, and also #1285, is that I want to add some of the popular editor features / add-ons concerning fields, before we can make it the `Fields` Svelte Component, and as a result of this, make the whole Editor a svelte component, which will probably break several editor add-ons yet again.

How would you recommend me to continue with this? My optimal UI solution would have Drag'n'Drop, something like [this](https://github.com/SortableJS/Sortable), however like mentioned above, this won't work. Should I:
1. Drop the functionality altogether
2. Try to build a UI solution without drag'n'drop (e.g. sorting with key presses)
3. Postpone UI interface, and just build the backend support, and add access to the raw preferences, which would look kinda like this:
```
[[0, 0, 1, 1], [0, 1, 1, 1], /* one for each field */]
-- position-x, position-y, extent-x, extent-y
```

### DIFF
--- a/ts/components/Spacer.svelte
+++ b/ts/components/Spacer.svelte
@@ -1,0 +1,12 @@
+<!--
+Copyright: Ankitects Pty Ltd and contributors
+License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+-->
+<div />
+
+<style lang="scss">
+    div {
+        width: var(--width, auto);
+        height: var(--height, auto);
+    }
+</style>
--- a/ts/components/StickyBar.svelte
+++ b/ts/components/StickyBar.svelte
@@ -6,15 +6,16 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     export let id: string | undefined = undefined;
     let className: string = "";
     export { className as class };
+    export let height: number;
 </script>
 
-<nav {id} class={`container-fluid pb-1 pt-1 ${className}`}>
+<nav {id} bind:offsetHeight={height} class={`container-fluid pb-1 pt-1 ${className}`}>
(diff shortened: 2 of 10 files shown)


dae:
Sorry for the delay Henrik, I'll try to get back to you in the next few days.

dae:
The drag&drop issues may not necessarily be a Qt issue - it's possible it's being caused by our redefinition of dragEnterEvent()/dropEvent() in editor.py.

These add-ons are a bit niche, so I'm not sure it makes sense to integrate the drag and drop functionality at the moment. No objections to a hidden preference, if it can be done in a way that doesn't introduce new issues or require large changes.

No strong preference for sticky over fixed, but currently this PR has the top of the deck options flowing under the header.

hgiesel:
Thanks for the pointers. I'll investigate it a bit further.

I'll close this one in favor of #1312, as I think it's the only reasonable change for now.
After investigating again, there was a good reason why I made StickyBar sticky and not fixed, so that it can possibly work in a window with other components (like editor + previewer) combined ^^;

PR #1301, opened 2021-07-18
Prevent webview from freezing due to error in _runHook

hikaru-y:
Fixes https://forums.ankiweb.net/t/2-1-45-release-candidate/11362/20 .

I know that add-on authors and users should make sure that there are no errors in the hook functions, but I created this PR because the clayout's webview also freezes while editing a template.

### DIFF
--- a/ts/reviewer/index.ts
+++ b/ts/reviewer/index.ts
@@ -27,14 +27,12 @@ export function getTypedAnswer(): string | null {
     return typeans?.value ?? null;
 }
 
-function _runHook(arr: Array<Callback>): Promise<void[]> {
-    const promises: (Promise<void> | void)[] = [];
-
-    for (let i = 0; i < arr.length; i++) {
-        promises.push(arr[i]());
+async function _runHook(arr: Array<Callback>): Promise<void> {
+    for (const callback of arr) {
+        const promise = new Promise((resolve) => resolve(callback()));
+        // swallow error to prevent webview from freezing
+        await promise.catch((e) => console.error(e));
     }
-
-    return Promise.all(promises);
 }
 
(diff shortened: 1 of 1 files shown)


hgiesel:
This will start all promises sequentially, instead of running them in parallel. The `catch` should be on the `Promise.all()` call.

hikaru-y:
@hgiesel Thanks for pointing that out.

>The catch should be on the Promise.all() call.

This means something like the following?
```diff
--- a/ts/reviewer/index.ts
+++ b/ts/reviewer/index.ts
@@ -27,14 +27,14 @@ export function getTypedAnswer(): string | null {
     return typeans?.value ?? null;
 }

-function _runHook(arr: Array<Callback>): Promise<void[]> {
+function _runHook(arr: Array<Callback>): Promise<void[] | void> {
     const promises: (Promise<void> | void)[] = [];

     for (let i = 0; i < arr.length; i++) {
         promises.push(arr[i]());
     }

-    return Promise.all(promises);
+    return Promise.all(promises).catch((e) => console.error(e));
 }

 let _updatingQueue: Promise<void> = Promise.resolve();
```
I tried this, but it did not prevent the webview from freezing.

As you said, in this PR, hooks are executed sequentially. However, considering the case where there are multiple hooks that operate on the same DOM, I thought sequential execution would be more appropriate for this function than parallel execution, so that's what I did. I should have written the reason why I did it that way, though.

For example, if there are multiple hooks for manipulating images with `querySelectorAll("img")`, I think that parallel execution would not guarantee consistency of the results. Am I wrong?

hikaru-y:
Closed this PR. See https://github.com/ankitects/anki/pull/1302#issuecomment-882588177 .

PR #1303, opened 2021-07-19
Add a hook to execute after pasting in a field

Arthur-Milchior:
I expect to use it to ensure that images are resizable after they get pasted in a field


dae:
editor_did_paste would better match the existing hook names. Can't guarantee this code path will stick around for a long time, as the paste handling might get refactored in future editor changes, but happy to add the hook for now.

dae:
(to clarify, please fix the name first :-)

Arthur-Milchior:
Anyway, if you deal with resizing image directly in Anki, I may never use this hook. As I expected to use it to deal with ensuring new images are resizable. 
Name changed, because it might still be useful maybe

dae:
Thanks Arthur!

PR #1304, opened 2021-07-19
fix card flag clearance when exporting

alexgiorev:
The card tuple is inserted into `data` for insertion after the loop, but then a separate list is created and modified which has no effect because it is the tuple that will be used to insert data into the new collection.


dae:
Thanks Alex! Would you like to add yourself to contributors, or would you like me to make the change to the file instead?

dae:
Thank you!

PR #1306, opened 2021-07-21
ignore importing cards of existing notes

alexgiorev:
When `Anki2Importer` imports notes, it doesn't import notes which already exist and were modified more recently than the ones being imported. But I notice that it also doesn't "ignore" them by inserting them into `_ignoredGuids`. Is this a bug?


dae:
What if the user transfers the cards to a different profile, studies them, and wants to import them back? The user may not want them ignored just because the note has not been edited.

In any case, the importing code is due for a rewrite, so I'd prefer to hold off on any changes that are not strictly bugfixes, as it may result in wasted effort otherwise. Thank you for your contribution in any case :-)

PR #1307, opened 2021-07-21
Image Resizer + Maximum image size

hgiesel:
Inspired by the summernote Image handle and @Arthur-Milchior's [Resize images in editor](https://ankiweb.net/shared/info/1103084694).

1. This introduces `editable-container`, which contains the style sheets and editable. This is where the shadow root now starts, instead of in editing-area. The shadow root was necessary to isolate editor styles from user content styles, however if we keep including more of our styles into `editable.scss`, it kinda defeats its purpose :)

2. Clicking on an image activates the image handle overlay. The image handle overlay itself is written in Svelte. There are no modifications done to the HTML in the editable field, it all works through overlays. In the overlay you can change the float property of the image. You can also see its dimension. You can also (by clicking the bottom button, or dblclick) toggle the size between:
- **constrained size**: constrained to 250x125px. The image is constrained so that big pictures do not make the editor hard to navigate. 
- **display size**: the size that will be used when the image is shown in the (p)reviewer.

3. The image handle overlay adapts to RTL mode (otherwise the float icons would be terribly confusing).
4. The image can be resized in display mode by clicking on its bottom corners.

Screenshot 2021-07-21 at 18 37 49


Arthur-Milchior:
Thanks for the mention. I am SO happy that I will not have to maintain it anymore. 
For the context, this add-on was downloaded 58095 times. One of my biggest success in term of add-on. That is to say, I suspect this feature will be really loved.

I wish you good luck. I don't really understand what you wrote ( I don't know Svelte. I see that you seems to reimplement a lot of resizing yourself. Given how hard it was to use jquery-ui, I may understand why.). In any case, you have all my gratitude

dae:
This is quite slick Henrik! And as Arthur said, it looks to be something quite a few people want.

I noticed a few minor issues:
- Resizing when the size is constrained behaves a bit oddly, with the size numbers changing while the image doesn't alter size.
- While the aspect ratio is maintained while dragging, if the image is shown in a narrow window during review, it can end up looking distorted. Would it make sense to use max-width/height instead of width/height?
Screen Shot 2021-07-26 at 4 38 28 pm

- Because the aspect ratio is constrained, perhaps mouse movements up or left should also reduce the size? Currently only diagonal movements change the image size, which can feel a bit odd.

hgiesel:
@dae 
1. Resizing during constrained mode should not have possible, that was an oversight.

2. I've now changed it to only set the width of the image, because I guess we typically want it to adjust to the width of the window (or device) when reviewing the card. If the `height` is unset, it will default to `auto`, which will make it try to preserve the aspect ratio.
I mostly set the `height`, because that's what both Summernote and Resize Image in browser does, however I also think the preserving the aspect ratio during review makes sense.
@Arthur-Milchior Was that a deliberate decision on your part, that both `width` and `height` are set on the image?

3. I've implemented this in 5bad9a9, but decided to revert again. The current behavior is basically: The longer length (width or height) will decide the dimensions of the rectangle. This is also the behavior in a few graphical editors I've tried just now, which is why I decided to revert.
Some apps also show another behavior (e.g. Apple Pages), but none of them show a behaviour of "the smaller length decides the dimensions of the rectangle".

Arthur-Milchior:
I think I recall that in some case, people wanted resizes that are not proportional. So I had to set both width and height. I also added an option to make it proportional if I recall correctly.
Essentially, I just did what was asked. I don't use this add-on myself. 

Maybe @AnKingMed would have more context. I discussed a lot with him to try to understand what are some complex use case for image heavy deck.

AnKingMed:
I do have many thoughts on this and would be happy to help. 

- with arthur's add-on, I set it so all images resized would have the class .resizer and that class in the styling is listed below. We quickly found that users were having problems with images being resized for desktop, but then being too small on mobile or other issues. This was the simplest fix
`/*Compatibility with resize images in editor add-on */
    .resizer {min-width: 0%!important; }
    .mobile .resizer {min-width:100%!important;}`

- Users are going to want to be able to customize the constrained size. This is because a small picture on one screen is not necessarily a small picture on a different screen resolution. 
- Users will want to be able to resize directly from review (Currently Edit Field During Review Cloze allows this, but it broke with the latest beta for some reason)
- Arthur had an issue with users resizing images too small and then "disappearing" at first. Not sure if your solution has something to prevent this
- I don't think we ever found an elegant solution to having max-width,max-height set in the collapsible size and then adjusting the width/height of the card. It did leave some images looking weird when collapsed. 
- The reason arthur's add-on has height and width is because some users prefer to adjust width while others prefer to adjust height

Let me know if there are any questions with this. I've spent a lot of time researching how users want this feature back when I was working with Arthur

hgiesel:
@AnKingMed 
1. With c70b18f, the images now adapt if the viewport gets too small. Maybe that is sufficient.
2. That is certainly possible. My favorite solution would be to allow for this via #1049, which I intend to follow up on after this.
3. That we cannot do. Editing in the reviewer is a flawed concept as long as we allow arbitrary JS to be executed from the template.

4. I guess that could happen if the image is resized to 0 pixel. I'll work on it.
5. The same issue exists in this PR. The reason I didn't get it to work is because while we want bigger images to be constrained to a smaller size, we don't want smaller images to enlarged by the constrained size.
6. I would like to hear it from those users then what their use case is to have a specific height set. I cannot really think of a use case.

AnKingMed:
1. The issue I described is really that on mobile users will want almost every image at 100% due to small screen size
2. Sounds like you have this addressed 
3. I'm pretty sure edit field during review cloze is one of the most popular addons out there. It makes the most sense to actually adjust the image in the screen you are viewing it in (even I personally prefer to do it this way). Editing image size in the editor works, but often requires resizing the editor screen to be larger so the image can be adjusted large enough for the preview screen (does that make sense?)
4. Looks like you fixed this 
5. Is there a way to constrain with proportions? (Ie constrain the width but height constrain is "auto"? 
6. I believe #5 is why users cared about this. If you only edit the width and constrain by Max width, then cards may be super tall still in the editor. Some users preferred to constrain height and thus editing the height of the images worked better.  I remember running into this problem with Arthur's addon and that's why ultimately we made it so the addon could constrain max height and/ or max width and then users could edit height and/or width.

hgiesel:
**1**. However that is also not generally true. E.g. I have some cards, which depend on images being inline in the flowing text. If those were resized to 100%, it would break the layout. Or what about icons. If users can resize the images to be bigger, and on mobile those are resized to fit the device width via `max-width: 100%`, I think we can satisfy those users.


**3**. I won't deny it's a popular (and convenient) feature, I'm just saying we won't be able to implement this so it works for everybody. There was a reason why the add-on needed to be adapted to work with Clozes, because every kind of JS usage that modifies the field HTML will break it. E.g. you also cannot use it to edit MathJax, for exactly that reason. I'm taking some steps to make the EditCurrent more useful and dynamic to make editing while reviewing easier, see e.g. #1285.

**5** & **6**. I'll give this another look.

hgiesel:
One thing I haven't really considered is that the `image[src]` cannot be used uniquely identify an image within a field. This causes weird behavior, if one field has the same image (same src) multiple times. I'll need to find an alternative way...

dae:
Thank you both for weighing in. I don't feel particularly qualified to comment, as I don't have a good feel for when/why people would use this - I always thought it made more sense to make adjustments in the template, instead of micro-managing each image 😅 

> image[src] cannot be used uniquely identify an image within a field

I wonder how much of an issue that is in practice? How many users out there are including the same image twice in different sizes on the same card?

AnKingMed:
@dae us medical students rely on thousands of images of histology under the microscope, gross anatomy, etc.  When you're sitting in front of anki for 2-4 hrs/day and have to really understand all the images, it's nice to make them the size you want.  There's also often 3-10 images on one card and sometimes they need to be big to see things, but other times they can be smaller so you can fit more into one viewing panel and not have to scroll.    I know.. we're crazy for doing Anki 2-4 hrs/day and you probably never imagined that use case when Anki began :P

dae:
Indeed, it's not something I originally anticipated :-) Just to be clear, I'm not arguing against this feature, just saying it's something I don't have a good feel for. The programmer in me wants to identify patterns and treat them consistently - just as card templates define a common structure that cards share, do you think there is any scope to do similar with image sizing? For example, if images could be marked as small, medium or large, users could customize what each of those sizes meant, and it would be easy to make their meaning change when moving between devices with different screen sizes. Would that make sense? Or do users really need the ability to adjust sizes pixel by pixel in some cases?

AnKingMed:
I imagine 3 sizes would work, but based off my past experience I have a feeling people would prefer the pixel by pixel method.  It's a fine balance between simplification and allowing as much customizability as possible

hgiesel:
> I wonder how much of an issue that is in practice? How many users out there are including the same image twice in different sizes on the same card?

I actually had some cards, which had this issue. E.g. there are some characters which I included as inline pictures, like "image" (it's hard to see in dark theme GitHub), which makes it more likely to appear more than once.

It's already fixed in the latest version, images are now identified by their CSS path, which is certainly unique :)

> I always thought it made more sense to make adjustments in the template, instead of micro-managing each image.

I think the issue here is that we're serving two vastly different user groups here:
1. The tech-savvy, who can _quickly make some template adjustments_.
2. The tech-noob, who wants as much WYSIWYG as possible.

For me as a tech-savvy, I wouldn't even use the "bold" button, e.g.: I would set up a semantic CSS class in the template, and then use an add-on like [Custom Styles](https://ankiweb.net/shared/info/1899278645), to set up a custom shortcut for a `span` with this class applied.

When I developed this I was also trying to think of both groups. E.g. the image size being constrained, and being able to see the image dimensions would also be very helpful for me :)

I think if a non tech-savvy will encounter the situation of "this image is too big/small", his natural instinct will be to resize the image, similar to how he'd do it in Photoshop, or MS Word. Also I'm afraid we'll put more maintenance burden on us, if we introduce something like standard image sizes. There's probably going to be people wanting more standard sizes than just three, the ability to customize them, etc. and then everything will get more complicated again.

In the newest version, instead of inline CSS, I'm using the [img[width] attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-width), which sets an intrinsic width _preference_ at a very low priority.

It can easily be overwritten from the user CSS, even though it is an "inline" property. Two scenarios:
1. user wants to discard all his previous settings, and use CSS exclusively now: `img { width: 100% }`, which will ignore `img[width]` entirely.
2. user wants a default CSS for images, but he wants to keep those resizes he has previously done: `img:not([width]) { width: 100%; height: auto; }`, which will target images, he hasn't resized manually in the editor.

Maybe this can work?

P.S.
And we could still allow for "standard image sizes" through a backdoor by implementing #1049. Users could create their own size classes in their styling, and fill out the class in the HTML editor, and the constrained/display mode would still operate correctly.

hgiesel:
Related: https://www.reddit.com/r/Anki/comments/ouiggr/scaling_images/

Arthur-Milchior:
Related, from the meme channel: 
5ha62p

hgiesel:
@dae 
What is the CSS for `img` on AnkiMobile? That's my last worry here.
It seems like AnkiDroid is using [this](https://github.com/ankidroid/Anki-Android/blob/5efa59bf0de963a4ecc679cd022c3d8f794caacc/AnkiDroid/src/main/assets/flashcard.css#L80).

dae:
Sorry, I missed your reply. AnkiMobile uses

```
img {
  max-width: 95%;
}
```

Re the field styling changes:

- I wonder how it might look with a solid instead of dashed line, or something a little more subtly dashed?
- The field name is also part of the highlight now, but I find that a bit inconsistent, since if I try to click on a different field name, it does not select the field.
- The icon and name feel a little vertically off-center.

Screen Shot 2021-09-02 at 10 51 36 pm

Re the original feature :-)

- Looking quite polished, well done.
- Just an idea - what about putting the no float button between the left and right float buttons - would that be strange?
- I can make the outline behave strangely by resizing an image to fill the field, then resizing the dialog

hgiesel:
First, to explain why I suddenly changed how fields look:
1. At the state of the commit named "Remove comma" the highlight was a bit messed up, mostly because it was still on the editable element. This is already noticeable in the Anki 2.1.45-47 as when you change to the HTML-Editor, the field highlight disappears. Now it is consistent.
2. I've added the border around the label container to make the sticky icon more present - give it a more logical place to stay. A common complaint I heard was that people do not find it.

The container itself is inspired by the very field I'm typing in right now, the GitHub editor. You can see the dashed border at the bottom here.

> The icon and name feel a little vertically off-center.

The left padding was made to be the same as that of the field itself (6px):

Screenshot 2021-09-02 at 17 04 39

However the right edge still had the margin of the pin icon applied, which I removed, should look better now.

Alternatives to the dashed border:
Dotted border:
Screenshot 2021-09-02 at 22 23 00

Solid border:
Screenshot 2021-09-02 at 22 23 37

Those are all [border styles](https://developer.mozilla.org/en-US/docs/Web/CSS/border-style) that are different at 1px width. Everything else would require a bit more setup (through `background-image` most likely).

Personally, I like the dashed border and the dotted border. The dashed border makes it look a bit like a physical sticky note. The dotted border makes it very subtle which is also nice. The solid border makes the separation too strong IMO.

dae:
Thanks for fixing those issues; looks great. I think I slightly prefer the dots, but either works - let's see what people think when this arrives in a beta.

PR #1308, opened 2021-07-22
Build and check help links on the backend

RumovZ:
I'll leave any questions in the comments.


rslib/Cargo.toml
```diff
@@ -18,6 +18,8 @@ prost-build = "0.7.0"
 
 [dev-dependencies]
 env_logger = "0.8.4"
+linkcheck = { git = "https://github.com/ankitects/linkcheck.git", rev = "2f20798ce521cc594d510d4e417e76d5eac04d4b" }
+tokio = { version = "*", features = ["macros"] }
```
RumovZ, line 22: Is that the correct way to add a feature flag only needed for testing?

rslib/src/links.rs
```diff
@@ -0,0 +1,122 @@
...
+    DeckOptions,
+    #[strum(serialize = "editing#features")]
+    EditingFeatures,
+}
+
+pub fn help_page_link(page: HelpPage) -> String {
+    format!("{}{}", HELP_SITE, page)
+}
+
+pub fn help_page_link_from_str(page: &str) -> String {
```
RumovZ, line 50: We could try to parse the url here to report any errors. But since vanilla Anki should never use this, it seemed superfluous. Let me know if you think differently.

rslib/Cargo.toml
dae: Seems ok to me. Unfortunately our use of a custom reqwest & linkcheck using a standard reqwest is making the update.py script fall over - I'll look into that more tomorrow.

rslib/src/links.rs
```diff
@@ -0,0 +1,122 @@
+// Copyright: Ankitects Pty Ltd and contributors
+// License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+
+use strum::{Display, EnumIter, EnumString};
+
+static HELP_SITE: &'static str = "https://docs.ankiweb.net/";
+
+#[derive(Debug, PartialEq, Clone, Copy, Display, EnumIter, EnumString)]
+pub enum HelpPage {
+    #[strum(serialize = "getting-started#note-types")]
```
dae, line 10: I'd probably go with a separate impl like to_page() on the enum that returns the relevant help page, instead of relying on serialize here. It just feels a bit odd to be returning webpage fragments in a derive, and a match expr would be a bit easier to read, eg BrowsingFindAndReplace => "browsing#find-and-replace".

rslib/src/backend/links.rs
```diff
@@ -0,0 +1,46 @@
...
+            ProtoHelpPage::Latex => HelpPage::Latex,
+            ProtoHelpPage::Preferences => HelpPage::Preferences,
+            ProtoHelpPage::Index => HelpPage::Index,
+            ProtoHelpPage::Templates => HelpPage::Templates,
+            ProtoHelpPage::FilteredDeck => HelpPage::FilteredDeck,
+            ProtoHelpPage::Importing => HelpPage::Importing,
+            ProtoHelpPage::CustomizingFields => HelpPage::CustomizingFields,
+            ProtoHelpPage::DeckOptions => HelpPage::DeckOptions,
+            ProtoHelpPage::EditingFeatures => HelpPage::EditingFeatures,
+        }
```
dae, line 44: Since this enum is really only for the frontend code, and we're not deriving any extra value from having a native Rust enum for it, I'd probably drop it, and use the protobuf enum directly. We can get an EnumIter on it by modifying the bottom of protobuf.rs.

proto/anki/links.proto
```diff
@@ -0,0 +1,38 @@
...
+    FILTERED_DECK = 12;
+    IMPORTING = 13;
+    CUSTOMIZING_FIELDS = 14;
+    DECK_OPTIONS = 15;
+    EDITING_FEATURES = 16;
+  }
+  oneof page {
+    HelpPage variant = 1;
+    string literal = 2;
+  }
```
dae, line 37: This is a totally valid approach, and fine as is, but I'd be tempted to drop the string case, allowing us to use the enum directly instead of having to deal with a oneof. The use of string arguments in the Python end is us supporting legacy code rather than something we actually want to encourage, and we could keep handling that legacy path in the Python end. Not a big deal though.

dae (commented):
Looking good! A few comments below.

proto/anki/links.proto
RumovZ: No, I agree. If we don't do anything special with the literal, it's a lot cleaner to drop the oneof and handle it on the frontend.

dae:
I think the link check is probably best only run in CI - otherwise when we go to bisect an old version in the future to chase down some bug, we may end up with failing tests if the website has been updated.

Unfortunately reqwest is a bit tricky when used in an upstream crate - we need to control the SSL implementation for use in Bazel, but need to jump through hoops to do that. Hopefully it should be ok now.

RumovZ:
Great! Thanks for also fixing the lints.
Checks failed again, but it looks like that's an unrelated issue? Ready to be merged from my side then.

dae:
Looks great! Thanks Rumo. The build failure looks like JS tooling flakiness :-(

PR #1309, opened 2021-07-25
Fix importing when current deck is filtered

RumovZ:
The first commit prevents the current deck to be set in the DeckChooser (viz the Import Dialogue) if it's filtered. The second one prevents a KeyError if the current deck is filtered when importing.
Not sure if this is the correct way of fixing it.


dae:
Seems reasonably, and this just needs to tide us over until the importing code can be rewritten. Thanks!

PR #1310, opened 2021-07-26
Fixes regarding text color buttons

hgiesel:
- "Set foreground color" becomes "Set text color"
- Introduces a label for "Set text highlight color"
- Changes `WithShortcut`, so it can work with ColorPicker (F8 shortcut didn't work before)


dae:
Thanks Henrik!

PR #1314, opened 2021-07-29
Export jquery correctly

hgiesel:


dae:
Thanks Henrik :-)

BlueGreenMagick:
@hgiesel Can you also export jquery as `jQuery` as well? My addon imports jquery-ui into the reviewer, but jquery-ui uses `jQuery` instead of `$`.

dae:
Done

PR #1315, opened 2021-07-30
Remove registerShortcut for Ctrl+C/V/A/X from editor/index.ts

hgiesel:
Should fix https://forums.ankiweb.net/t/shift-ctrl-v-to-strip-preserve-formatting-of-pasted-text-results-in-text-being-pasted-twice-in-2-1-45/11792.

I swore I remember that I wasn't able to copy the field names on macOS without these, but I guess I messed up ^^;

### DIFF
--- a/ts/editor/index.ts
+++ b/ts/editor/index.ts
@@ -8,8 +8,6 @@
 import { filterHTML } from "html-filter";
 import { updateActiveButtons } from "./toolbar";
 import { setupI18n, ModuleName } from "lib/i18n";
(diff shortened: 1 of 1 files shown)


dae:
It seems like it's the shift shortcut that's the issue - with this PR, cmd+shift+v no longer works. I guess we'll need to conditionally apply the bridgeCommand line for macOS?

dae:
Thanks Henrik, that should do it :-)

PR #1320, opened 2021-08-02
Remember last text and highlight color again

hgiesel:


dae:
Thanks Henrik!

dae:
If you've got time for https://forums.ankiweb.net/t/hotkey-for-sticky-fields/11891/2, I think that's the last thing we should cover before I can bake a 2.1.46rc

PR #1321, opened 2021-08-03
Sticky shortcuts

hgiesel:
I added F9 and Shift+F9 as shortcuts for toggling sticky for current field or all fields respectively.

Toggling all fields will turn it off for all fields, if any fields are sticky. It will turn it on for all fields, if no fields are sticky.

I also increased the inactive opacity of the sticky icons a little (`.1` -> `.2`), and added a tooltip which shows the F9 shortcut.


dae:
Thanks for the quick fix Henrik. Hope it didn't keep you up - it looks rather late (early?) over there!

PR #1324, opened 2021-08-07
MathJax Preview + In-line editor

hgiesel:
This PR builds upon #1307 literally (I rebase upon it, the first commit in this PR only relevant to the Mathjax editor is `Prevent overflow of sizeDimensions or image handle buttons`). This PR also builds upon the idea I mentioned in #1250.

Here's a quick video:

https://user-images.githubusercontent.com/7188844/128614043-41ec2863-fc30-48c4-9954-c31fe1056649.mov

It basically introduces two subsystems:
1. A MathJax handle, which is an overlay which hovers over MathJax images, and allows the editing of them (and switching between inline/block).
2. Inside the editor field is, what I called "decorated component", which is a web component, which injects a Svelte component. There is some documentation to this idea in https://github.com/ankitects/anki/commit/09eec3dd4deadc6daecedc28f89e42190948c401.

Advantages:
- Preview Mathjax in the editor
- You can structure your Mathjax with linefeeds
- You can see if you do an error in the Mathjax, the error message will be shown on hover.
- Prevents usage of HTML within Mathjax
- Hides some implementation details of Mathjax from user
- The Mathjax editing process is entirely an artefact of the editor, every component outside of the Editor won't know anything changed.

Disadvantages:
- Mathjax editing is one more click away.
- Sometimes left/right arrows a/o deleting is a bit wonky. E.g. when you have a Mathjax block, put your caret on the line below, and keep hitting left.

TODOs
- [x] Shortcuts don't behave in Mathjax editor


dae:
Just gave this a quick test and it's looking really good Henrik. Haven't checked out the code yet, just had a quick play around with it. Some minor things I noticed:

- We'll probably want to ensure the editor webview is extended to 100vh, or the bottom of the editing area will get truncated if no fields follow it:
Screen Shot 2021-08-09 at 4 04 38 pm

- To me the corner markers look like drag handles. Would it make sense to use a dashed line around the entire rectangle instead of the corners? I found myself thinking a similar thing in the image resizing PR as well.

hgiesel:
1. This seems to be the same issue I noticed (and fixed) [in the Tag Editor PR](https://github.com/ankitects/anki/pull/1264#issuecomment-870675185). I'll defer it for now, depending on which PRs are going to be merged first.
2. I've just been doing what summernote does when it comes to the image resizing, however that also sounds good. I'll give it a try.

hgiesel:
Some remarks:

* After this PR, the scrollbar for the fields will be in #fields, no longer on the whole page.  ​
This was just a consequence of making TagEditor `sticky` instead of `fixed`, but was something I wanted at some point anyway.
* By fixing the behavior upon entering colon in the [TagEditor PR](#1264), I broke the behavior for space. I fixed it here.

The only thing which still needs work is Copy/Paste of [decorated elements](https://github.com/ankitects/anki/commit/09eec3dd4deadc6daecedc28f89e42190948c401). I imagined that if you copy Mathjax out of Anki, you would get the old syntax, e.g. `\( 1 + 2 = 3 \)`, and if you paste this syntax into Anki (or maybe only if it comes from within Anki (prefixed with `<!-- anki -->`), you will get the image immediately.

dae:
Looks great Henrik, will merge this in and the remaining issues can be sorted out in a separate PR.

A couple of JS errors in the console I noticed:

repeatedly:
JS error /_addons/dist/web/editor.js:599 Uncaught TypeError: Cannot read property 'getSelection' of null 

when creating first mathjax entry:
JS error /_anki/js/editor.js:38708 Uncaught IndexSizeError: Failed to execute 'setStart' on 'Range': The offset 4294967285 is larger than the node's length (3).

when typing text and clicking outside the box before the screen has refreshed:
JS error /_anki/js/editor.js:44410 Uncaught (in promise) TypeError: Cannot read property 'closest' of null

I also ran into a case where the entry got stuck - it would not update and I could not click away from it to hide it. I have not been able to reproduce it. I think this error appeared at the same time:
JS error /_anki/js/editor.js:54268 Uncaught (in promise) TypeError: t4.slice is not a function

sandersantema:
Sorry to chime in on this closed PR but as a heavy mathjax user I have a slight concern, can I still disable this and edit the plain mathjax? The preview looks really nice and is sure to be a improvement for a large part of the userbase but when editing a lot of mathjax I think having to click the formula every time could be a big nuisance. Especially when already using a plugin such as [editor live preview](https://ankiweb.net/shared/info/1960039667) which serves the preview purpose for me, no inline editing needed. 

Although it might be nice to have a option of displaying the previews of mathjax formula's in a separate preview window such as in the editor live preview plugin, but better since you can edit the preview inline. Now that this editor preview is implemented a built in preview might not be that far fetched.

hgiesel:
@sandersantema 
Disabling it is certainly possible, even though there's no built-in option.
But I ask you to give it an honest try in the next feature release. Mathjax being one click is a disadvantage (I've mentioned it in my opening post), but there are also big advantages, the biggest one probably being able to use linebreaks in Mathjax (!).

For example this is the equation from the link you provided with spacing:
Screenshot 2021-09-21 at 14 43 06

I think it's such a boon to readability.

Depending on user feedback, we could also give a try to showing the editor at a lower opacity on mouse hover events.

sandersantema:
I really hope you could add an option toggling this. I can agree that readability of mathjax formulas is bad, but as I write my formulas as short as possible (in line with the minimum information per note principle) and have become quite good at reading them I've got no real problem with readability. 

I also wonder how this would work with the combination of closet and mathjax, which I do heavily rely upon. All in all I really think this a wonderful function for the usability of Anki but I do really hope a toggle would be included for me and (and I'd imagine) other users who don't have so much need for the inline preview but who are quite hampered by having to click the formula and for instance use their mouse. Something which I don't do a lot in my normal card creation workflow.

hgiesel:
> you can not add a button telling users how to produce a normal newline

Why not? We could add buttons for both cases in any case.

Screenshot 2021-10-06 at 17 37 14

I think consistency among _input elements_ is a very strong argument here.

hgiesel:
Another option would be a placeholder text on empty editor:

Screenshot 2021-10-06 at 18 51 25

I-Al-Istannen:
Yea, that is a somewhat significant drawback. If it isn't any concern though, I quite like the placeholder text idea. The buttons would be seldomly used and mainly act as documentation, so putting that in a placeholder text sounds nice to me :)

PR #1335, opened 2021-08-19
Quote font family names

hikaru-y:
Fixes https://forums.ankiweb.net/t/font-script-is-not-changing-in-the-anki-windows/12465 .

According to [the w3c specification](https://www.w3.org/TR/2018/REC-css-fonts-3-20180920/#generic-family-value), generic families must not be quoted, so in this PR, font names will not be quoted if the string consists only of lowercase alphabetic characters and hyphens. This is just in case an add-on sets the font family to a generic family using `editor_will_use_font_for_field` hook.

### DIFF
--- a/ts/editor/editing-area.ts
+++ b/ts/editor/editing-area.ts
@@ -105,10 +105,18 @@ export class EditingArea extends HTMLDivElement {
         firstRule.style.color = color;
     }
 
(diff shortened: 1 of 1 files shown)


dae:
I didn't know that about the quoting! Thanks for looking into this issue Hikaru.

PR #1337, opened 2021-08-21
Mention .anki can be imported with add-on 175027074

Arthur-Milchior:
I believe this message would be more helpful.
Maybe not ideal as a user may not know how to install add-ons, but I don't  how much details can be given in such a warning. A link would be ideal, but that would lead to a problem of translating its content.


dae:
How confident are you that the add-on will be kept up to date in the future? I seem to recall it broke in the past. The advantage of installing an old Anki version is that it's fairly likely to work.

If you are committed to maintaining the add-on and would like to proceed with this, I'd recommend mentioning the add-on by  name instead of id, as  users may struggle to look it up by id.

Arthur-Milchior:
The main advantage of id is that you can just copy paste it into the ID field of the add-on manager. Admittedly, that requires to know how to use an add-on. But I'm fine with name, or link.

The advantage of the add-on is that you can just install it, and import as usual. Far easier than dealing with two version of anki, importing, exporting and reimporting.

You're right, the add-on broke at least twice when Anki updated, and I updated it twice. Last update, this week-end, was because the file rsbackend was renamed to _backend.
I'm fine with keeping a .anki file and testing it when you release new version of anki. 
Up to now, keeping this add-on up to date was really easy, as the only thing that broke were a few imports. More drastic change may potentially break the add-on in a way that is not easy to repair. However, since the add-on is mostly a port of anki 2.0 import method, breaking everything would require a huge change in Anki, I don't think it'll happen soon.

"in the future" is too vague for me to answer. In short term, very confident. If for some reason I decide to leave the anki community, I expect to first try to find someone to give all my add-ons to; I spent too much time here and owe too much to Anki to decide to leave without ensuring that things go smoothly. In 10 years, no idea. I would not have predicted you'd move to rust today when I started being active here, I won't try to predict what you'll do of anki in 3 years.

dae:
I don't doubt your intentions Arthur :-) If you'll do your best to keep the add-on working, that works for me. I had been thinking that users might try to search for the add-on by id on AnkiWeb, but you're right that the code will make adding it into Anki easier, so let's just leave it as-is.

PR #1340, opened 2021-08-23
Export Svelte internals

hgiesel:
This is an implementation of what you suggested in #1327 and it works very well.
This increases the amount of setup for add-ons a little bit, but it's the solution we want. Maybe  the `svelte-shim.js` file could also be packaged with npm. I'll look into that (probably in a week though, as I won't be online much next week).

I've setup New Format Pack with this version [here](https://github.com/hgiesel/anki_new_format_pack/tree/svelteshim/web).


dae:
Thanks Henrik, feeling much more positive about this now.

By moving the require definition into Anki, we'll gain more flexibility, and it will mean the only thing required on the add-on end is to mark svelte as external: https://github.com/hgiesel/anki_new_format_pack/commit/ff41a9c1225368b87a2e22f2f7d70cb307f77d63

PR #1343, opened 2021-08-27
Fix bug when add-on package string is ""

BlueGreenMagick:
Trying to install an add-on whose package string in manifest is "" deletes the entire add-on folder and installs the addon inside the base addon folder. (Caused by using `os.path.join(base, addon_package_string)` as add-on directory path)

I forgot to set the package string by mistake and lost all my add-ons. Luckily it wasn't my main Anki profile :relieved:

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -157,7 +157,7 @@ class AddonManager:
         "type": "object",
         "properties": {
             # the name of the folder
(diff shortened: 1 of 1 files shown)


dae:
This doesn't fully solve the problem - a '/' or '..' as the package name is still accepted. I'll follow this up with another patch.

PR #1344, opened 2021-08-28
Add a hook for subscribing to dialog creation / opening

glutanimate:
Super-simple hook that allows add-ons to subscribe to dialog creation / raise events.

One use case would be in Speed Focus Mode, which currently [monkey-patches](https://github.com/glutanimate/speed-focus-mode/blob/59165bd3bd057b7e9b2f613bcede24557d7a30d5/src/speed_focus_mode/reviewer.py#L317) `aqt.DialogManager.open` to time suspending its timers upon e.g. editing the current card.

Example snippet to test / debug:

```python
from aqt import gui_hooks

def on_dialog_manager_did_open_dialog(dialog_manager, name, instance):
    print(dialog_manager, name, instance)

gui_hooks.dialog_manager_did_open_dialog.append(on_dialog_manager_did_open_dialog)
```

### DIFF
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -18,6 +18,7 @@ from anki.buildinfo import version as _version
 from anki.collection import Collection
 from anki.consts import HELP_SITE
 from anki.utils import checksum, isLin, isMac
+from aqt import gui_hooks
 from aqt.qt import *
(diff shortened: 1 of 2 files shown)


dae:
LGTM - thanks!

PR #1345, opened 2021-08-28
Use different artificial reviewer button highlight for Windows

hgiesel:
This should fix https://forums.ankiweb.net/t/good-button-outline-gone/12636/4, however my Windows machine still cannot compile from source, so this needs to be tested.


dae:
Thanks Henrik, looking good on Windows and Linux. What was the problem you're having building on Windows? Does a 'bazel clean --expunge' and removal of ts/node_modules not help?

PR #1348, opened 2021-08-30
Fix flags not updating when changing profile

RumovZ:
As pointed out here: https://forums.ankiweb.net/t/flag-names-and-profiles/12757


dae:
Thanks Rumo, and welcome back!

RumovZ:
Thank you. 🙂

PR #1349, opened 2021-08-30
Disable 'Rename with Parents' for top-level items

RumovZ:
As suggested here: https://forums.ankiweb.net/t/in-sidebar-dont-show-rename-with-parents-when-irrelevant/12755


dae:
Thanks!

Issue #1341, opened 2021-08-25
Avoid JS globals where possible?


dae:
(for after the .47 release)

@hgiesel is it practical to move almost everything we export under an 'anki' namespace? eg components→anki.components, $editorToolbar→anki.editorToolbarPromise, etc?

- Do you agree it's a good idea?
- We'll of course need to leave around various globals that people are already relying on. Maybe with a deprecation message to direct people to the correct location?
- Maybe we could add a routine in ts/lib that adds an entry to the anki namespace, creating it if it doesn't already exist?
- Presumably in cases where we're using --global-name, we'll need to switch to an approach like editor/index_wrapper.ts (with it perhaps using the routine in the previous point)?

hgiesel:
I agree it's a good idea.

After 12ea482c8768e7e6f33e4e531682968d6ef077b6, I thought it would also be nice to able to use the `import` syntax for ankis functionality.

For example

```typescript
const {
    IconButton,
    contextKeys,
} = anki.components;
// becomes
import { IconButton, contextKeys } from "anki/components";
// or without compiler
const { IconButton, contextKeys } = require("anki/components");
```

This might also prevent issues, where addon devs might be tempted to modify the `anki` global.

Another danger of just extending the `anki` object I see is with ease of understanding. I imagine the following situation: We expose some utility function `do_thing()` in the editor. An add-on dev now wants to write an add-on for the deck-options screen, and wants to use this `do_thing()` utility, so he goes:

```typescript
// my-deck-options-addon.js
const { do_thing } = anki;

do_thing();
```
And gets an error "do_thing is not a function". Even though in a different context (webview) this would work. However if we exported those libraries under different names, this would make it a bit clearer:

```typescript
// editor-addon.js
const { do_thing } = require("anki/editor");
/* works */

// deck-options-addon.js
const { do_thing } = require("anki/editor");
/* Error: Cannot require(anki/editor) at runtime. */
```

And once we start having bigger webviews doing more things (let's say the browser) we'll have multiple "packages" exposed in the same webview:
`require("anki/editor"); require("anki/browser"); require("anki/sidebar");`

[...]

PR #1350, opened 2021-08-30
Implement register-package and export some packages

hgiesel:
This is the idea I menioned in https://github.com/ankitects/anki/issues/1341#issuecomment-907808073.

Exported packages are:
* `anki/packages`
* `anki/shortcuts`
* `anki/bridgecommand`

We can define the exported package directly in the file, and have the "package registration" as a side effect of using the functions in our own code. This way we don't need to pay that much attention to what is exported, but rather add-on devs can always _generally do what Anki does_.

As a test, I've set up my add-on "Asset Manager" to use this approach. You basically just need to mark "anki" as external, and you can use the `import` syntax for anki: https://github.com/hgiesel/anki_asset_manager/compare/ankilibs.

Trying to import a nonexistent package results in:

> Uncaught Error: Cannot require(anki/nonexistent) at runtime.
>     at globalThis.require (editor.js:19900)
>     at __require (editor-compiled.js:11)
>     at editor-compiled.js:41
>     at editor-compiled.js:123

The package `anki/packages` contains some inspection utils. `listPackages` simply lists all available packages, and `hasPackages` checks for their existence. This combined with the fact, that you can still use `require` syntax, allows you to be quite flexible when importing (similiar to python):

```typescript
import { hasPackages } from "anki/packages";

if (hasPackages("anki/nonexistent")) {
    /* do something */
} else {
    console.log("not there");
}

const { registerShortcut } = hasPackages("anki/shortcuts")
    ? require("anki/shortcuts")
    : {
        registerShortcut: function() { console.log("shim") },
    }
```

### DIFF
--- a/ts/editor/index.ts
+++ b/ts/editor/index.ts
@@ -7,6 +7,7 @@
  */
 
 import "sveltelib/export-runtime";
+import "lib/register-package";
 
 import { filterHTML } from "html-filter";
 import { updateActiveButtons } from "./toolbar";
--- a/ts/lib/bridgecommand.ts
+++ b/ts/lib/bridgecommand.ts
@@ -1,6 +1,8 @@
 // Copyright: Ankitects Pty Ltd and contributors
 // License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
+import { registerPackage } from "./register-package";
+
 declare global {
     interface Window {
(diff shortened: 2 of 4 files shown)


dae:
Thanks Henrik, this looks nice. I've got my hands full with some AnkiWeb stuff at the moment, but hope to be able to return to this and your other PRs within the next few days - sorry for the wait.

PR #1351, opened 2021-08-31
Add FONT as an inline tag

hgiesel:
Reported [here](https://forums.ankiweb.net/t/coloring-text-sometimes-adds-br/12754/9).

### DIFF
--- a/ts/editor/helpers.ts
+++ b/ts/editor/helpers.ts
@@ -29,6 +29,7 @@ const INLINE_TAGS = [
     "DFN",
     "EM",
     "EMBED",
+    "FONT",
     "I",
     "IFRAME",
     "IMG",


hgiesel:
Hm, maybe I should have looked for non-existence of block elements when I implemented this instead, rather than checking that all are inline tags, as [block tags are better defined](https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements), and also undefined tags (e.g. `<xyz></xyz>`) are inline by default.

Other ways to trick the algorithm:

1. If you include block elements in inline elements (which is _technically_ invalid HTML, but is accepted by the `DOMParser`), e.g.
```html
<span style="color: red"><ul><li>Item</li></ul></span>
```
In that case an extraneous line break will be added in the editor.

2. Another way is to use custom HTML elements, which should behave like inline elements, but of course are not included in the list (just like `font` is not included currently), e.g.
```html
<xyz>asdf</xyz><br>
```
The `<br>` will not be visible in the `Editable`.

PR #1353, opened 2021-08-31
Check for non-existence of block tags instead of exclusive existence of inline tags in editable

hgiesel:
This is a revision of #952 and deals with the issues I mentioned in #1351.

We could merge #1351 as a quick fix for another patch release, and consider this PR for the next minor release.

Something I noticed along the way is that we don't need the special treatment of empty `anki-editable`s anymore, even though I'm not entirely sure why.


dae:
Thanks Henrik, probably best to give this one a bit more testing, so will merge the other one in first.

PR #1354, opened 2021-09-02
Disallow setting, deleting or defining properties on packages

hgiesel:
I've noticed that you can actually set/overwrite/delete properties on our packages.
So you could do:

```js
require("anki/packages").listPackages = 5;
const n = require("anki/packages").listPackages;
// n is 5
```

I don't think we'll want this. It's not like it would affect Ankis code, but it could be confusing to add-on devs if somebody were to do this.

### DIFF
--- a/ts/lib/register-package.ts
+++ b/ts/lib/register-package.ts
@@ -3,6 +3,8 @@
 
 import { runtimeLibraries } from "./runtime-require";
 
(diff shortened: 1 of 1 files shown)


dae:
I presume this will still allow us to do things like mutate an exported array? We may want that for something like a future anki/hooks/... package as a solution for #647

PR #1356, opened 2021-09-05
Add a hook for modifiying pasted MIME data

tatsumoto-ren:
See https://forums.ankiweb.net/t/investigating-an-ambigous-add-on-error/12846

After running into a bug that prevented me from wrapping the right-click-paste functionality I decided to add a new hook. I hope to use it in [Paste Images As WebP](https://ankiweb.net/shared/info/1151815987).

I had to move around some existing code, but only when necessary. Namely, cursor position has to be calculated before the hook is called because its position might move when the user interacts with the popup window.

screenshot

An additional variable `drop_event` is passed to the hook in order to let the user take different actions depending on where the event comes from (drop or paste).


dae:
Thank you! This will probably not be in the next bugfix update, but should be in the release after that.

PR #1357, opened 2021-09-07
Prevent sticky bubble

hgiesel:
* Clicking on the sticky does not focus the field - I think toggling the sticky state does not necessarily mean you want to focus/edit the field.
* Save the field before you toggle the sticky state:
  * Before, typing something and quickly toggling the sticky state before the change timer resolved would undo what you just typed.
  * It still does reset the caret position, because toggling the sticky state will actually update all fields. I think this is best to be fixed, when we change how sticky works, as discussed [here](https://forums.ankiweb.net/t/default-editor-settings-instead-of-persistent/12761/7).
* Put the tooltip on the sticky icon. It was on the label container before, which was probably an oversight

### DIFF
--- a/ts/editor/editor-field.ts
+++ b/ts/editor/editor-field.ts
@@ -41,11 +41,11 @@ export class EditorField extends HTMLDivElement {
     }
 
     connectedCallback(): void {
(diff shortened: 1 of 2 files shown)


dae:
Thanks!

PR #1364, opened 2021-09-08
Fix ResourceWarning: unclosed file <_io.BufferedWriter name=5>

evandrocoan:
https://forums.ankiweb.net/t/aqt-sound-py-resourcewarning-unclosed-file-io-bufferedwriter-name-5/13070


evandrocoan:
Closing it for now. I just see this popping up when the warning showed up:

image

evandrocoan:
The fix is actually working. The problem I saw today was because I was missing to clone the stdin from another _process variable.

dae:
Thanks Evandro!

PR #1369, opened 2021-09-14
Wait for HTTP server startup

### LINKED ISSUE #1362: Wait until mediasrv.py to completely start
Some rare times when I open anki, I see the top toolbar with unloaded/not found image (sync button maybe?). I think this is because we do not wait until the waitress/wsi server to completely start up as done in: https://github.com/Pylons/webtest/blob/4b8a3ebf984185ff4fefb31b4d0cf82682e1fcf7/webtest/http.py#L123-L132
```python
    def wait(self, retries=30):
        """Wait until the server is started"""
        running = check_server(self.adj.host, self.adj.port,
                               '/__application__', retries=retries)
        if running:
            return True
        try:
            self.shutdown()
        finally:
            return False

def check_server(host, port, path_info='/', timeout=3, retries=30):
    """Perform a request until the server reply"""
    if retries < 0:
        return 0
    time.sleep(.3)
    for i in range(retries):
        try:
            conn = http_client.HTTPConnection(host, int(port), timeout=timeout)
            conn.request('GET', path_info)
            res = conn.getresponse()
            return res.status
        except (socket.error, http_client.HTTPException):
            time.sleep(.3)
    return 0
```

May be related to https://github.com/ankitects/anki/issues/1137 - Investigate page-loading race conditions

evandrocoan:
Fix #1362 and add a default favicon.ico

Related to: https://github.com/ankitects/anki/pull/554 - Fixed HTML5 media tags


dae:
Thanks Evandro. Thinking about this a bit more, I'm wondering if this will actually help - by the time _ready.set() is called, we already have the effective port, so presumably the server is already online at that point. Before I merge this in with the risk of causing a regression, I'd appreciate it if you could give it a go in your branch for a while and see if it does make the loading issue go away or not.

evandrocoan:
Sure, I can try it out for a few months and see if the problem happens again or not. But I think is will be the solution because I tested by adding a `time.sleep(9)` before `self.server.run()`: 
```python
            self._ready.set()
            import time
            time.sleep(9)
            self.server.run()
```

Then, new code I added to `MediaServer.wait_start_up()` kept waiting until `self.server.run()` was called. That `MediaServer._ready` flag is not actually much useful because its use now, is to not now allow `MediaServer.getPort()` to be called before `self.server = create_server(...)` has run (setting the port `getPort()` needs).
```
INFO: Build completed successfully, 240 total actions
\\?\F:\bazel\anki\rybdqld5\execroot\ankidesktop\bazel-out\x64_windows-fastbuild\bin\qt\runanki.exe.runfiles\pypi__38__stringcase_1_2_0\stringcase.py:247: DeprecationWarning: invalid escape sequence \W
  return re.sub("\W+", "", string)
2021-09-15 20:25:39.708933 waiting media server on 127.0.0.1:52819...
2021-09-15 20:25:43.021312 waiting media server on 127.0.0.1:52819...
mpv not found, reverting to mplayer

DevTools listening on ws://127.0.0.1:8087/devtools/browser/526c74aa-fb57-4b43-ba68-3bf1a96f1934
```
I think the problem happens because when we call https://github.com/ankitects/anki/blob/9daf037c0ba765a8834f6abfe80e16cff85b2ed7/qt/aqt/main.py#L1687 we immediately start a Python thread on background, which does not wait until `self.server.run()` is called (ignoring also our `self._ready.set()`). This way the QT interface starts parallelly with our `MediaServer.run()` call and if QT starts up first than our `MediaServer` thread,  some UI elements are not going to get a meaningful response from our waitress server. 

I would say this problem is random (and rare) because the Python GIL (Global Interpreter Lock) does not have any criteria about which python thread is going to run, for how long it is going to run, and it only allows a single thread to run at a time: https://stackoverflow.com/questions/52007923/python-threading-inconsistent-execution-time

I would say it happens on my computer because it is not much fast (a little old too, from 2009) and on top of that, and I underclocked a little my processor too (I am trying to keep my processor from getting too much hot and last longer). I also live with a bunch of SWAP going and I never turn my computer off (I just hibernate, so no daily memory leak cleaning for programs who like to leak memory, except for Google Chrome I restart some times because my swap is getting over 20 GB). But doing SWAP is fast over my SSD "disk".
image

dae:
Thanks Evandro, I presume you haven't encountered any issues during this time, so will merge this in now so it can make it into 2.1.50.

evandrocoan:
Hi. I rewrote this part today, because of the conflicts on the rebase, so it is not much tested, but I just tested your master and looks likes it is working fine:
```python
@app.route("/favicon.ico")
def favicon() -> Response:
    request = BundledFileRequest(os.path.join("imgs", "favicon.ico"))
    return _handle_builtin_file_request(request)
```

Today I got my top bar loading empty, that is why I updated my fork with your main branch and run the branches rebase. I was about 200 commits behind when this happened:
image

As I always enable the Chrome Debugger, I could see the page:
image

And inside it:
image

There was nothing on my console:
image

This is my bazel.log [bazel-.log](https://github.com/ankitects/anki/files/7585654/bazel-.log), my bazel generates this file because I added this setting: (https://github.com/evandroforks/anki/commit/10fb1dc8621fc557694eea8bfd458447224531ce#diff-ce61097c4ce2e1ce5bc8c7030457491c7376abe7600c5c38b06890c69d1f4a17)
```bash
call .\bazel.bat run --explain=bazel-.log --subcommands --show_progress --worker_verbose --verbose_failures --verbose_explanations=true --test_output=streamed %BUILDARGS% //qt:runanki -k -- %*
```

Well, my bazel was doing some random invalid permissions access sometimes since last week. I think is related to bazel messing something up in my installation. The problem would fix itself if I just restarted the bazel build.

dae:
Let's keep an eye on things and see how it goes when 2.1.50 hits beta

dae:
I think I finally got to the bottom of this: https://github.com/ankitects/anki/commit/db5117ce1ab3b1393419d4fcc9c03678094f2f30

PR #1371, opened 2021-09-14
Experiments with Server-Sent Events (SSE)

hgiesel:
I've been thinking about the architecture of the Editor a bit.
For the sticky icon, e.g., we currently set it via `web.eval` from Python, making the NoteEditor _aware_, that it is run in different _contexts_ (like from the `AddCards` window for example).

In the end, I'd like to remove this _context-awareness_ and rather configure the NoteEditor top-down, e.g. from an `AddCards.svelte` file with the NoteEditor having multiple slots:

```svelte
// AddCards.svelte
<AddCardsButtons />
<NoteEditor
    on:fieldsupdate={...}
    on:tagsupdate={...}
/>
    <ButtonsAddMode slot="toolbar" />
    <FieldAddMode slot="field" />
    <TagAddMode slot="tag" />
</NoteEditor>
``` 

This way they
* pass slots to the NoteEditor to provide their individualed functionality

* can be defined in their own directory ( `ts/addcards`, or `ts/editcurrent`) for better code separation

* can also dispense with `bridgeCommand`, and prefer bubbling up events, like `fieldsupdate`, `tagsupdate`, which then can be handled differently in the different contexts (and make `postRequest` calls directly to Rust if necessary).

My ultimate goal would be that that these views are completely decoupled from Python, so no `web.eval`, and instead they use `postRequest` to contact the rslib APIs directly (just like the Graphs screen), and maybe minimal use of `bridgeCommand` (for OS things like closing the window).

However... there's one thing I don't thing we can currently implement just in Svelte, and that is the functionality of the `operation_did_execute` hook, and how it updates the Editor if the note is updated upstream in another view.

For that, I think, we'd need either SSE (server-sent events) using `EventSource`, or `WebSocket`s.
This way, we could basically _listen_ to an EventQueue in Python from JavaScript.

---

In this PR I implemented SSE. With the AnkiWebView Inspector you can connect to the `EventQueue` in the EditCurrent window by executing:

```javascript
q = new EventSource(`_anki/_stream/update_note/${getNoteId()}`);
q.addEventListener("message", console.log);
```

This way, whenever the note changes in a different view (e.g. the Browser), the EditCurrent will receive an event, informing it of the change (however it will also inform if itself updates the note, i.e. any update to the note). You can also listen to removal events:

```javascript
q2 = new EventSource(`_anki/_stream/remove_notes/${getNoteId()}`);
q2.addEventListener("message", console.log);
```

However I also noticed several issues along the way:
- For the update event, we'd need to figure out if the event originated from a change in the current view or not, or maybe just a good way to save and restore the current position 
- The server (?) only allows for 4 `EventSource` connections to be opened. Not sure whether this is because the media server cannot accept more connections, or it's a [HTTP/1.1 restriction](https://developer.mozilla.org/en-US/docs/Web/API/EventSource).
- Freeing the resources on the server works only suboptimal (it kinda has to be forced by emitting a few `update_note` events). Trying to release the resources with `q.close()` also yields an error message. It looks I'm implementing the protocol wrong.
- While I was working with the server in `qt/aqt/mediasrv.py`, I wondered whether this is a make-shift solution, and the media server / Backend server is eventually going to be moved to Rust?

### DIFF
--- a/qt/aqt/editcurrent.py
+++ b/qt/aqt/editcurrent.py
@@ -28,7 +28,6 @@ class EditCurrent(QDialog):
         self.editor.card = self.mw.reviewer.card
         self.editor.set_note(self.mw.reviewer.card.note(), focusTo=0)
         restoreGeom(self, "editcurrent")
-        gui_hooks.operation_did_execute.append(self.on_operation_did_execute)
         self.show()
 
     def on_operation_did_execute(
@@ -47,7 +46,6 @@ class EditCurrent(QDialog):
             self.editor.set_note(note)
 
     def cleanup_and_close(self) -> None:
-        gui_hooks.operation_did_execute.remove(self.on_operation_did_execute)
         self.editor.cleanup()
         saveGeom(self, "editcurrent")
         aqt.dialogs.markClosed("EditCurrent")
--- a/qt/aqt/mediasrv.py
+++ b/qt/aqt/mediasrv.py
@@ -4,9 +4,9 @@
 from __future__ import annotations
 
 import json
-from functools import wraps
 import logging
 import os
+import queue
 import re
 import sys
 import threading
@@ -148,8 +148,7 @@ def allroutes(pathin: str) -> Response:
         if flask.request.method == "POST":
             return handle_post(path)
         elif flask.request.method == "GET" and path.startswith("_stream/"):
-            shortened = path[len("_stream/"):]
-            print("eyo lets start stream", shortened, request.args)
+            shortened = path[len("_stream/") :]
             return handle_stream(shortened)
 
         if fullpath.endswith(".css"):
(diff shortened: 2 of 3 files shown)


dae:
Yep, this is something I had in the back of my mind when implementing the current operation_did_execute() hook. Agreed it's probably the path we'll want to go down, though it's probably going to need to wait until Q1 2022  - we have the Qt/arm/Python upgrade coming up next month, and that plus updates to import/export are probably going to take us through to the end of the year at the least.

I'm not very familiar with either SSE or WebSockets, but the latter seems like it doesn't suffer from the connection limit SSE has, and as it's on a different port, we could potentially have that served up by the backend code while retaining mediasrv for (some) regular HTTP requests / add-on support.

I think this would need to be integrated into the undo/redo system like the current operation system is, as otherwise the hooks would not be called when the user undoes/redoes an operation. If we implemented something akin to operation_did_execute() + CollectionOp() in the JS end, we could presumably implement something similar to the Python code's initiator= argument, where the calling instance is able to ignore operations it initiated, and any events that arrived via the websocket would just have a null initiator.

dae:

PR #1374, opened 2021-09-17
Fix NotFoundError when deleting notes

RumovZ:
Solving https://forums.ankiweb.net/t/error-when-deleted-card-within-browser/13392.
What do you think about the `append` flag? It's not 100% clean, but I think it can be very useful.
Another problem is that no row change is fired after undoing a deletion. So you may end up with a single selected row, but the editor remains hidden. I don't know enough about the undo-redo-mechanism, yet, to solve it, though.

### DIFF
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -595,7 +595,7 @@ class Browser(QMainWindow):
         nids = self.table.get_selected_note_ids()
         # select the next card if there is one
         self.focusTo = self.editor.currentField
-        self.table.to_next_row()
+        self.table.to_unselected_note_row()
 
         remove_notes(parent=self, note_ids=nids).run_in_background()
 
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -597,6 +597,7 @@ class Browser(QMainWindow):
         # if possible, select a row that will not be deleted
         self.focusTo = self.editor.currentField
         if not self.table.to_unselected_note_row():
+            self.table.clear_current()
             # no row change will fire
             op.success(lambda _: self.onRowChanged(None, None), append=True)
 
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -592,12 +592,15 @@ class Browser(QMainWindow):
         if focus != self.form.tableView:
             return
 
-        nids = self.table.get_selected_note_ids()
-        # select the next card if there is one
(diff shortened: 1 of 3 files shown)


dae:
So that we don't hold up the 2.1.48 release (much), I wonder if this could be solved more simply for now? For example, do we actually need the "advance to next card on delete" behaviour? Or could we just get away with clearing the selection, or leaving it alone (but clearing the current card)?

Will get back to you on append :-)

RumovZ:
I think dropping this feature would be justifiable. However, `onRowChanged()` needs to be called either way, so we only save the new `to_unselected_note_row()` routine. Does that really speed things up?
If you only want to fix the NotFoundError and don't care about the empty editor in the upcoming release, the simplest way should be to catch and discard the error.

RumovZ:
The third commit fixes another niche error that could occur if the info dialog got requested while focus was on a deleted row.

After a little bit of experimentation, it doesn't seem advisable to just remove the current element after deleting notes, because then, if the first row is disabled, the card list can't be accessed via keyboard. Only after clicking in the card list do the jump keys work again.

dae:
Swallowing the errors feels dirty, but you're right that it may be the simplest short-term solution. Could you let me know the steps you're using to reproduce this? I can't seem to trigger it in my own testing.

RumovZ:
1. Go to card mode.
2. Search for a note with multiple cards. (I use the Select Notes action.)
3. Select a single row, but not the last one.
4. Delete it.

As for the other bug:
1. Delete the last row.
2. Rightclick the deleted row, which will still be focused.
3. Click "Info...". (Ctrl+I won't do it.)

dae:
Thanks for the repro steps - turns out my issue was that I wasn't using a shortcut key to trigger the delete. Does https://github.com/ankitects/anki/commit/972059251993c87c3395dc5d09cd96c3bc53c7ef look ok to you as a temporary fix for the first issue? If so, I'll include this in a stable build tomorrow.

append seems like a reasonable solution - we could also do it as a list of callbacks instead, but wrapping is probably fine - we're unlikely to have more than a few at max.

RumovZ:
> Thanks for the repro steps - turns out my issue was that I wasn't using a shortcut key to trigger the delete.

Hm, I can reproduce it either way. As it's a race condition, it's probably affected by a number of factors and might just happen to be easier to reproduce on my system than on yours.
The fix looks good as far as I can tell.

dae:
Might be worth thinking about #1381 at the same time?

PR #1375, opened 2021-09-17
Disable unusable browser actions on row change

RumovZ:
This goes well with #1374, but I assume you want to defer it until the next release, so I made a separate PR.

I've deliberately kept the `@skip_if_selection_is_empty` decorator as a safeguard for invalid API calls. But maybe it should throw an exception now instead of showing a tooltip?


dae:
Thanks Rumo!

dae:
Re throwing an exception, since we've already got a translation in place, I guess it might make more sense to keep the current user-friendly message. But I don't have strong feelings about it.

PR #1377, opened 2021-09-17
Reverse-engineer surrounding with execCommand

hgiesel:
Ever since [I broke the Custom styles add-on](https://github.com/ijgnd/anki__editor__apply__font_color__background_color__custom_class__custom_style/issues/45), I've been meaning to add some better support for surrounding code with HTML or text wrappers (surrounders).

In this PR, I already did the first step, and reverse-engineered what `execCommand` does when you want to wrap with simple inline styles (`<b>`, `<i>`) in `surroundNoSplitting` (and it works very well).

Other things I'd like to implement:
- [ ] surround text in a way that ensures that you use only of the surrounding element will exist afterwards (useful for text wrappers (including `{{c1::`,`}}`), `<anki-mathjax>`, wrapping with `<ruby><rb>`, basically every situation where you only want to have one tag of the kind afterwards) (in the code below I called it `surroundSplitting`)
- [x] unsurround `surroundNoSplitting`, and `surroundSplitting`
- [x] trigger the _surrounding effect_ on text input (like what happens when you press "Bold" without text having selected, and then you start type)
- [ ] Have a third kind of surrounding that works for block surrounders (like unordered / ordered list), and behaves more like the summernote list buttons.

We could use these functions ourselves for our existing buttons, but could also expose them as API. This would free us from `document.execCommand`, and allows us to customize the behavior, but it also adds a lot of logic, as you can see below.

In any case, I'd like to defer this until I ported some Web components to Svelte components.


dae:
Could you help me understand the problems you're trying to solve with these changes?

hgiesel:
This is mostly adressing things that would be _nice to have_:

1. The surrounding code mostly works by using `document.execComand` and `document.queryCommandState`, which works fine, but..
    1. we're limited to built-in commands.
    2. in some cases, it uses deprecated elements like `<font>`
    3. we can't do anything about their quirks. For example, `document.queryCommandState("underline")` will return `false` (button is inactive), if you place your caret in a `<span style="text-decoration: underline">{}</span>`, however you can still undo it. On the other hand, `document.queryCommandState("bold")` will return `true`, if you place your caret in a `<span style="font-weight: bold">{}</span>`, and you can also undo it.
    4. Some more oddities for lists: If your caret is in an unordered list, and you press ordered list, the intent should be obvious: you want to turn it into an ordered list, however what happens is that the current list item is turned into an independent ordered list, while the rest remains unordered (he Summernote editor does that correctly)
        * Except if the caret is currently on the very beginning of the `li`, in which case the list item is undone, and the caret vanishes (very strange).
2. The problem here is that all points mentioned above by themselves are _small things_, but you can't do anything about any of them, because the responsible code is out of our hands. If we moved this code into our code base instead, we could make it behave like we want.
2. Surrounding HTML with other HTML (or text) is basically synonymous to formatting. If we can provide a nice API to define formatting commands, and use it ourselves as well, we can benefit from it, but also provide it to add-on devs as an API (similiar to `registerShortcut`)

dae:
Thanks Henrik, I appreciate you taking the time to explain in more detail, and you make a good case for the change. It will certainly by nice to get away from some of the quirks we've had to deal with up until now.

dae:
Looks like deprecation of execCommand is going to make this necessary anyway :-) And require us to come up with alternatives to the other commands as well unfortunately.

hgiesel:
So...
The three buttons bold, italic, and underline now use our code instead of `document.execCommand`. Some notes:

- unsurrounding a section of text is actually quite the complex operation, if you think about it, and was by far the hardest to implement.
- There were some bugs with jsdom which mislead me: There are still parts commented out in the tests, regarding the ranges. I've actually fixed the bugs in upstream jsdom, however we'll have to way until Jest 28, until we can actually profit from my fixes (and uncomment the tests)
- So far I've only made the bold, underline, and italic buttons use the new code. I could extend this to the superscript, subscript, removeFormat, textColor, and highlightColor buttons. I've left it like this, for now, so we can compare the how those three buttons behave compared to the ones which still use document.execCommand.
- There are certainly some differences in behavior from my implementation compared to document.execCommand, for example, my implementation normalizes the text more aggressively
- I'd like to leave the remaining surround operations (like the ones for lists, or `anki-mathjax`) to a later time, to first focus on some other things and come back to it eventually ^^;

dae:
This looks like it was a lot of work! I've only briefly skimmed the code, but happy to see quite a few comments and tests.  Let's give this a go - if any serious issues pop up, we can presumably temporarily switch back to the old command buttons while keeping the rest of the code in place. Thanks as always Henrik :-)

dae:
(as an aside, I was a bit surprised the other day to see `<anki-mathjax>` tags in the HTML editor - could we be showing the user the original `\(` etc instead?)

hgiesel:
Concerning the mathjax, IMO that would bring several drawbacks:
- Right now, actually the syntax highlight changes inside `anki-mathjax` tags. It recognizes sequences like `\begin{matrix}` e.g., but does not react to `<b>`. It's not perfect, but I think it could be further fine-tuned. The mathjax editor has support for this, it could not be done when using `\(`.
- **(Upon further investigation, I think the following point broke at some point, I'm looking into it)** We would require another processing step for this: [Here](https://github.com/ankitects/anki/blob/8b3d964866af772396b826822d217226e15a39f2/ts/editor/PlainTextInput.svelte#L48), we parse the text to transfer it from PlainText back to RichText, and for that we need to tell the parser, that anki-mathjax is `white-space: pre`, otherwise `&` will become `&amp`, etc.. So if the field contained the delimiters `\(`, we would first need to transfer that to `<anki-mathjax>`, and then parse it again. 
- I wanted to use a similar method for clozes at some point. A visual method to edit clozes in the RichText, and HTML for the PlainText. Here it would be even more important to make the cloze delimiters HTML tags in the PlainText.

dae:
Ok, thanks for clarifying. If it's impractical, we'll just need to document it instead, like with the percent-escaping of spaces in media filenames.

noinkling:
I just want to say that if you go this route, make sure the undo functionality is robust. I've been developing an addon where I've similarly found `execCommand()` unfit for purpose due to quirks/bugs (so I've implemented some similar stuff to what has been done here), but a big advantage with it is that you get undo/redo for "free". In my case I've been using hacky workarounds (involving calling `execCommand('insertHTML')` on the whole field at the last step) to make it work "good enough", but to do it properly you really need to implement your own undo stack.

> Looks like deprecation of execCommand is going to make this necessary anyway

It's deprecated, but realistically probably won't be going anywhere for a long time due to web compatibility reasons, especially since there is no replacement API (that can actually trigger such commands) in sight.

Instead of taking the in-house approach, it might be worth considering using an off-the-shelf WYSIWYG editor that's already robust (CKEditor springs to mind, but I'm sure there are others) and has the desired functionality in its APIs.

hgiesel:
Right now, we provide no undo functionality of our own.
Some of the functionality is there (like saving the caret position), but now work has been done on an editing undo queue.

> CKEditor [...]

Of course I have considered this before doing any major work.
I've looked at several, and discovered issues with each of them:
- Some of them do not support _multi root editors_. These are editors where there are multiple editing fields (_roots_) with a single toolbar.
- Some have incompatible licenses or have pricy licenses which would not make sense for Anki
- Some of them don't have HTML editing mode, or are not suited for this kind of editing
- Some of them are semantic editors, which are editors where the user input is translated to editing operations, which are executed on an underlying logic layer, which is then rendered back as visible HTML to the user. That is a really cool idea in theory, but it doesn't quite fit with Anki, as the semantics of the fields are mostly set by how the notetype, and the cardtypes are set up. I'm trying to make up for it by having some "Semantic elements", which you can see in the newest Anki version in the form of Mathjax elements. 

> especially since there is no replacement API (that can actually trigger such commands) in sight.

I doubt there will ever be a replacement API for this, as these kind of editing commands are just too complex to define like this. Pointer Events, Selection/Range API is as close as we get to a replacement.
The only thing that there is a discussion around is an more close to metal alternative to `contentEditable=true` (typically `contentEditable=events`, or `contentEditable=typing`).

PR #1379, opened 2021-09-20
set CardLayout fill_empty=False for add cards editor

BlueGreenMagick:
In Add Cards dialog, I think the Card layout button is almost always used as a previewer. So displaying accurately may be more important than having placeholders.

The render is especially bad for card layouts using conditional fields like the following:
```
{{#Front-opt}}{{Front-opt}}{{/Front-opt}}
{{^Front-opt}}{{Vocab}}{{/Front-opt}}
```


dae:
Seems reasonable; thanks BGM.

PR #1381, opened 2021-09-21
In browser.table.model.flags, don't fetch rows

BlueGreenMagick:
In the browser table, when `Ctrl + A` selecting all the cards, `flags()` is called for all rows . And because `flags` fetches its row, that makes selecting all rows an expensive operation.

This PR changes it so `flags` only uses cached rows. Because `table.StatusDelegate.paint` calls `model.get_cell()` which fetches row data, I don't think there will be any difference except when selecting items that are not painted.

It decreases the time taken by `flags` while selecting ~10000 notes from 4.7s to 0.51s.

### DIFF
--- a/qt/aqt/browser/table/model.py
+++ b/qt/aqt/browser/table/model.py
@@ -291,7 +291,9 @@ class DataModel(QAbstractTableModel):
         return None
 
     def flags(self, index: QModelIndex) -> Qt.ItemFlags:
-        if self.get_row(index).is_deleted:
+        item = self.get_item(index)
+        row = self._rows.get(item)
+        if row and row.is_deleted:
             return Qt.ItemFlags(Qt.NoItemFlags)
         return cast(Qt.ItemFlags, Qt.ItemIsEnabled | Qt.ItemIsSelectable)
 


dae:
Good catch! But I wonder if this may cause issues by allowing the user to select rows that correspond to a deleted element. @RumovZ, what do you think?

RumovZ:
Yes, that's what I would be afraid of, too. On the other hand, I didn't get any errors during a short trial. The backend seems to handle non-existing passed ids robustly — and has to regardless of this PR, I guess.
However, there are issues with the GUI:
1. The reported numbers in tooltips and the window title is off when there are deleted rows.
2. When deleting or restoring rows, the flags for the first cell are inverted at first:

1dOHUKHmaH

Deleted rows cause a lot of headeache... Maybe if we hooked into `notes_will_be_deleted` or similar, we could inspect the browser table's id list and manually render all affected rows deleted. That would resolve 1. and maybe 2., as all deleted rows will immediately be added to the cache.
It's also a more wholistic approach than what I've done in #1374 where I've only fixed the delete issues when triggered via the browser action.

dae:
Hooks aren't triggered when changes are undone and redone, so we can't rely on them to update the UI. For issue 1, what sort of ops are triggering that? A fair few ops return exact counts from the backend now.

RumovZ:
> For issue 1, what sort of ops are triggering that?

I was mistakenly thinking about the "... of ... cards updated" message. But apparently it's only used in Find&Replace for notes. So the incorrect number only appears in the window title.

> Hooks aren't triggered when changes are undone and redone, so we can't rely on them to update the UI.

Right, but as mentioned in #1374, that's already a problem, because we might end up with an empty editor and other invalid states. Would it be feasible to add hooks for undoing and redoing, so the GUI can react appropriately?

RumovZ:
I just had a look. If only `OpChanges` had a property like `deleted_nids`, we could handle everything properly in `browser.on_operation_did_execute()`.

dae:
That feels like a bit of a slippery slope - I suspect if we did that, it would be followed by other additional fields in the future, and OpChanges would grow into a bit of a mess. We also need to consider the case where an add-on is mutating the database directly - when that happens, we just synthesize an OpChanges with every category listed, and we would not have access to the note ids there.

I have not spent much time thinking about this yet, and you know the current browser table code better than I do, so I may be missing something. But couldn't we solve these issues just by using the existing OpChanges? For example, couldn't we fix the empty editor by teaching the browser code to check for a possible deletion when notes have changed, and hide the editor or advance the current row if necessary?

RumovZ:
I can see the flaws, but regrettably, I don't have any other ideas.

> teaching the browser code to check for a possible deletion when notes have changed

That would mean querying the backend for _all_ the ids in the table after every such operation, wouldn't it? That seems quite wasteful.
I guess we could fix the issues addressed in #1374 like this, because we would only need to check the focused and selected rows. But the issues with this PR would remain.

dae:
Yeah, I was talking about the editor state, not refreshing all rows up front - I agree that would be wasteful.

I'm not too concerned about the row count issue for find&replace, "1 of 3 notes updates" meaning "1 of 3 (some of which may have been deleted) notes" doesn't seem to be a big deal? 

Regarding the incorrect flags, that seems to be a Qt refresh issue - the row.is_deleted property is being correctly updated, but the Qt interface is not immediately refreshing, because it appears that dataChanged.emit() does not actually trigger an update of flags. If you resize the window, you should see the flags update correctly. So it seems the solution there is to force Qt to refresh the flags somehow - maybe we could accomplish it by adjusting the selection or using some other signal after dataChanged?

RumovZ:
So we've established that it's not possible to efficiently and immediately detect deleted rows. Then we have to accept that the selection may contain deleted items, meaning the window title and editor state may be technically inaccurate. At the very least, a correction to these states should be made when a missing row is detected later on, e.g. because it's scrolled into view. It follows that we have to check during the fetching routine whether the fetched row has been deleted (or restored).

I'm on it, but it's quite tricky to interfere at this stage and I'm not yet sure how to solve the flag issue.

dae:
Assuming this is closed by #1392; please feel free to reopen/reply if I've got it wrong

PR #1382, opened 2021-09-21
Tags from sidebar

RumovZ:
https://forums.ankiweb.net/t/creating-tags-from-sidebar/13462

I felt it was more natural to make the availability of sidebar actions independent of the table selection, so in fact, I depend on the forgiving `@skip_if_selection_is_empty` check (#1375).

### DIFF
--- a/ftl/core/browsing.ftl
+++ b/ftl/core/browsing.ftl
@@ -1,8 +1,8 @@
 browsing-add-notes = Add Notes...
 browsing-add-tags = Add Tags
 browsing-add-tags2 = Add Tags...
-browsing-add-to-selected-notes = Add to selected Notes
-browsing-remove-from-selected-notes = Remove from selected Notes
+browsing-add-to-selected-notes = Add to Selected Notes
+browsing-remove-from-selected-notes = Remove from Selected Notes
 browsing-addon = Add-on
 browsing-all-card-types = All Card Types
 browsing-all-fields = All Fields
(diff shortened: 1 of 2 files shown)


ftl/core/browsing.ftl
```diff
@@ -1,6 +1,8 @@
 browsing-add-notes = Add Notes...
 browsing-add-tags = Add Tags
 browsing-add-tags2 = Add Tags...
+browsing-add-to-selected-notes = Add to selected Notes
```
dae, line 4:
```suggestion
browsing-add-to-selected-notes = Add to Selected Notes
```

ftl/core/browsing.ftl
```diff
@@ -1,6 +1,8 @@
...
+browsing-remove-from-selected-notes = Remove from selected Notes
```
dae, line 5:
```suggestion
browsing-remove-from-selected-notes = Remove from Selected Notes
```

ftl/core/browsing.ftl
dae: I think a capital there is probably more natural. We could also potentially use "Add to Selected" to make it a bit less verbose, but no strong feelings there either way.

dae (commented):
Thanks for the quick implementation Rumo! Looks good, just a minor comment on wording:

ftl/core/browsing.ftl
RumovZ: I actually wrote "Add to Selected" at first, because that's already a comparatively long label. But since, at the moment, it's the only sidebar action that interacts with the table selection (instead of the sidebar selection), I thought it might be necessary to be a little more detailed.

dae:
Thanks!

PR #1384, opened 2021-09-22
Fix memory leak in previewer

hikaru-y:
This fixes the issue reported here:
https://forums.ankiweb.net/t/high-memory-usage-over-time-on-2-1-44/13544/


dae:
Thanks for the quick fix Hikaru - much appreciated!

PR #1387, opened 2021-09-22
Translate Editor entirely to Svelte

hgiesel:
This will make the editor a Svelte component hosted at `document.body`.

It's not yet complete (Codable (HTML editor), Image Resizer, and Mathjax Editor do not work yet), but I thought it's good enough to make a PR.

Some noteworthy things:
1. I tried to bubble up everything that isn't inherent to the process of note editing out of the component. These are crammed into `OldEditorAdapter`, which will be a holdover until we have components specific to the different editor use cases (AddCards, etc.).

2.
I think I finally found a good way to built up a nice API for add-ons and ourselves that allows access to inside. The trick is to set contexts with JS objects, that access stores, which are filled in by nested components.  You can have a look at how it can be used in `onFieldUpdate` in `OldEditorAdapter.svelte`, where I access `noteEditor.multiRootEditor.fields[index].editingArea.activeInput.fieldHTML`, which is a property is originally defined in `Editable.svelte` or `Codable.svelte` (`activeInput` abstracts over both of them).

If we go with this approach, I'll also update the tag editor and toolbar to use this approach. If we go this route all the way, code could access a Svelte object all the way up the context it's hosted in, and access something completely different in the tree.


dae:
Might take me a while to wrap my head around this :-) Is my understanding correct that the parent components are exporting a property to get at the child APIs, and also making their own APIs available via a context, so children can reach up the tree to call a parent API?

It's great how the API is type safe :-)

hgiesel:
If you try out the PR in its current state, you'll see that the Editable and Codable section can live-update each other. This even works for the mathjax element `<anki-mathjax>`:

https://user-images.githubusercontent.com/7188844/135465010-b272631d-c25c-48cd-bcca-aa8b2127fc56.mov

Having the `Editable` and `Codable` section both open at the same time is a nice touch imo:
* it gives you a field preview, if you more common work in the Codable
* if you're not that solid in HTML, you can instantly check the output of your HTML

However I think we should also preserve the ability to only have one of them open. I'd suggest that we put them in the area where currently the sticky icon is. In the code I already named it "FieldState":

Screenshot 2021-09-30 at 17 11 11

Regarding extensibility:
I also haven't wrapped my head around it entirely either, yet :)
I want to expose something top-down, where components, or even outside code can go down the tree to achieve some effect, and also something bottom-up (via contexts) which inserted add-on components can access. But I also don't want to couple our components too tight too each (make them asssume where they will be slotted, or what kind of slots they'll have).

dae:
It looks cool, and I can see how showing them both at once could be useful. It presumably solves https://github.com/ankitects/anki/pull/1324#issuecomment-924819405 as well, which is nice. A couple of thoughts:

- the default for new users would stay as the WYSIWYG editor only I presume?
- it may be a bit difficult to use if the field is long, as one of the two windows may not be visible on the screen - a side by side view might work better?
- the icons are too small to tap on touch devices - we'd need some workaround like a different way to trigger them (tapping could produce a dropdown with toggle items for example)

Re extensibility -

On the one hand, some sort of generic solution similar to https://doc.qt.io/qt-5/qobject.html#findChildren is appealing - it doesn't require assumptions about the way objects are composed, and should not require a lot of boilerplate. We presumably couldn't do it via DOM walking, but perhaps with a top-level wrapper exposing a context that allows children to register weakrefs of themselves could work.

On the other hand, we might be opening ourselves up to performance issues like that, and I'm not sure whether it would work for things like when components are inside an {#if ...} clause. Components manually exposing the API they wish to make public is more boilerplate, but it means we have control over what we're willing to support and what we aren't, and it offers better type checking than dynamic casts.

hgiesel:
> the default for new users would stay as the WYSIWYG editor only I presume?

Yes. 

> a side by side view might work better?

Hm, I'd say this depends on the window width. I'll come back to this.

> the icons are too small to tap on touch devices

How about a tap for focus, and a long tap for contextmenu (which can also expose sticky).

dae:
Sorry Henrik, I fear I might not have explained things well :-) Instead of squashing the changes from the main branch, I was suggesting a squash of the changes made in this branch, turning all of the commits you've made here into a single one.

There are two times it's useful:

- Just before merge into main, after all tests are passing. This ensures that the commits going into main will pass when we git bisect
- For longer-running branches where you need to periodically rebase your changes over the latest main branch, squashing your existing changes down before rebasing them tends to produce less merge conflicts on rebase. For changes that don't conflict with changes in the main branch, this is often not required, but in this case I think it would make the rebase job simpler.

You've removed the draft tag on this PR, but the tests are not passing yet, and it seems to be broken in my local testing: the icons are not appearing at the top right of a field, and the editor opens to both views at once. I'm also getting errors like this in the console:

JS error :1 Uncaught ReferenceError: saveNow is not defined
JS error :1 Uncaught ReferenceError: activateStickyShortcuts is not defined

So I presume this still needs a bit more work, or something went wrong during the merging process. Was there something in particular you wanted me to take a look at now?

hgiesel:
I'm sorry to have created so much noise. That was basically me learning how to squash-merge. I've removed the draft tag, because otherwise I don't get notified whether I have conflicts with the base branch (I'll make it a draft again). I've taken the advice you've made #1398, and had a look at some other Git strategies. I think I've been abusing rebase in the past in my bigger PRs, creating a lot of broken commits, because the underlying dependencies changed. I'll try to do that better in the future.

Regarding this mess here now: Would you suggest that I remove the squash merge of main (the last two commits) entirely and instead squash all the other commits into one big commit, and rebase that one on main?

dae:
No harm done, just wanted to make sure we were on the same page :-) I can use the "squash and merge" option when merging the PR later, so no need to manually tidy things up for the sake of neatness - my main concern at the moment was making it easier for you to rebase your changes on the latest main. The instructions I provided before won't work now that main has changed, and ts_project was rebased before merging into main. Git is complicated 😅 

ts_project_merge was created by taking your PR branch, and merging ts_project into it with this commit:

https://github.com/ankitects/anki/commit/5916f09297d43132e10958f42850540199502086

Unfortunately I edited the ts_project commit message after that merge, which gave it a new id. In retrospect, I should have merged ts_project into main before doing ts_project_merge - main's history is never rewritten, so had the merge been made from a commit on main, it would have made things easier. Anyway, we should be able to recover things. 

Start by checking out the merged branch I made for you:

```diff
dae@dip:Work/code/dtop% git co ts_project_merge
Switched to branch 'ts_project_merge'
```

If we look back at your PR on https://github.com/ankitects/anki/commits/ts_project_merge?after=173fad51f78e97083a06241c79317a7939cbcdaa+34&branch=ts_project_merge we see that the first commit before your changes started was https://github.com/ankitects/anki/commit/19bde7b52d4f89966c6ac21330bb2ee8a7393f41, so that will be our starting point. Clear out all the commits that were made, leaving the changes in the working dir:

```diff
dae@dip:Work/code/dtop% git reset --soft 19bde7b52d4f89966c6ac21330bb2ee8a7393f41
dae@dip:Work/code/dtop% git add .
```

Our working directory now contains all the changes made during your PR and the merge I did in ts_project_merge. If we were to commit now, it would be based on the old starting point, but we want to commit relative to the point ts_project was merged into main (https://github.com/ankitects/anki/commit/3ab544e4876b282f476143f5702ef8de7a7c96b6):
 
```diff
dae@dip:Work/code/dtop% git reset --soft 3ab544e4876b282f476143f5702ef8de7a7c96b6
dae@dip:Work/code/dtop% git co -b new-pr
Switched to a new branch 'new-pr'
dae@dip:Work/code/dtop% git commit -m 'squashed changes'
```

We now have a single commit relative to the point in the main branch. Let's catch up with the latest changes on main:

```diff
dae@dip:Work/code/dtop% git rebase main
Removing ts/editor/saving.ts
Removing ts/editor/editor-field.ts
Removing ts/editor/codable.ts
Auto-merging ts/editor/BUILD.bazel
CONFLICT (content): Merge conflict in ts/editor/BUILD.bazel
Removing ts/editable/editable-container.ts
error: could not apply 83ac34d0b... squashed changes
Resolve all conflicts manually, mark them as resolved with
...
```
... just a single conflict. Yay! Fix it manually, and then we're up to date. Finally, let's apply the two changes you made since the "satisfy linter" commit:

```diff
dae@dip:Work/code/dtop% git cherry-pick b30ab7864166a2cec6ddb990111cfcb879b3de9b
Auto-merging ts/editor/MultiRootEditor.svelte
Auto-merging ts/editor/EditorField.svelte
[new-pr 5d3745b20] Make editor fields grid rather than flexbox
 Author: Henrik Giesel <hengiesel@gmail.com>
 Date: Sat Sep 25 12:07:46 2021 +0200
 2 files changed, 10 insertions(+), 7 deletions(-)
dae@dip:Work/code/dtop% git cherry-pick 087902cdc32943386251f05101491a45289dec72
Auto-merging ts/editor/TagEditor.svelte
Removing ts/components/WithTheming.svelte
Auto-merging ts/components/ButtonToolbar.svelte
[new-pr c73c2003c] Refactor ButtonToolbar margins
 Author: Henrik Giesel <hengiesel@gmail.com>
 Date: Sat Sep 25 15:11:03 2021 +0200
 6 files changed, 15 insertions(+), 49 deletions(-)
 delete mode 100644 ts/components/WithTheming.svelte
```

Then we build and find there's a build error due to the symlinks lying around. Let's clean them up:

```diff
dae@dip:Work/code/dtop% find ts -type l | grep -v node_modules | xargs rm
dae@dip:Work/code/dtop% git commit -a -m 'remove remaining svelte.d.ts symlinks'
```

Then we're all set! You could optionally squash the changes again down to a single commit, but I'll stop here so you can see the current state instead. Example is in https://github.com/ankitects/anki/commits/new-pr

PR #1391, opened 2021-09-25
Add hook for selecting a new note type in the add window

sobjornstad:
Use case: I have an add-on that uses a couple of note types that the user shouldn't normally add notes to directly (when they sync using my add-on, notes of those note types added directly within Anki get deleted), so I'd like to display a reminder when one of these note types is selected. I could imagine other add-ons might find it helpful to set up custom functionality in the editor when an appropriate note type was selected.

Most of the add-cards hook names appear to start with `add_cards_`, while one starts with `addcards_`, so I went with the most frequent one. Let me know if I guessed wrong.

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -148,6 +148,7 @@ class AddCards(QDialog):
         self.editor.loadNote(
             focusTo=min(self.editor.last_field_index or 0, len(new.fields) - 1)
         )
+        gui_hooks.add_cards_did_change_note_type(old.note_type(), new.note_type())
 
     def _load_new_note(self, sticky_fields_from: Optional[Note] = None) -> None:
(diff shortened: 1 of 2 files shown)


sobjornstad:
Is it the list of people who have made changes to the CONTRIBUTORS file who are whitelisted, rather than the people actually physically listed in the file? My commit was authored the latter first time and it failed, then when I rewrote it with the former it worked. If that's the case, maybe the error message could be improved.

dae:
The check works via the git history - it checks the identities used to modify the file, not the actual file contents: https://github.com/ankitects/anki/blob/e660e0a3e56e2db160a1bb538edeb0f1d6653148/CONTRIBUTORS#L16

Hook looks good, thanks!

PR #1392, opened 2021-09-25
Handle missing rows consistently and speed up selections

RumovZ:
Phew, that took some pains, but it should be working robustly now, even with deletions from outside the browser. I would appreciate some testing though.
My only gripe is with the introduced complexity. I've made an effort to add some comments. Please let me know where things are still unclear.

The changes from #1374 and #1381 are incorporated.

Do I have to add a legacy alias for `browser.onRowChanged()`?

### DIFF
--- a/qt/aqt/browser/table/model.py
+++ b/qt/aqt/browser/table/model.py
@@ -119,6 +119,10 @@ class DataModel(QAbstractTableModel):
         )
         return row
 
+    def get_cached_row(self, index: QModelIndex) -> Optional[CellRow]:
+        """Get row if it is cached, regardless of staleness."""
+        return self._rows.get(self.get_item(index))
+
     # Reset
 
     def mark_cache_stale(self) -> None:
@@ -326,8 +330,10 @@ class DataModel(QAbstractTableModel):
         return None
 
     def flags(self, index: QModelIndex) -> Qt.ItemFlags:
-        if self.get_row(index).is_deleted:
-            return Qt.ItemFlags(Qt.NoItemFlags)
+        # shortcut for large selections (Ctrl+A) to avoid fetching large numbers of rows at once
+        if row := self.get_cached_row(index):
+            if row.is_deleted:
+                return Qt.ItemFlags(Qt.NoItemFlags)
         return cast(Qt.ItemFlags, Qt.ItemIsEnabled | Qt.ItemIsSelectable)
 
 
--- a/qt/aqt/browser/table/model.py
+++ b/qt/aqt/browser/table/model.py
@@ -4,7 +4,7 @@
 from __future__ import annotations
 
 import time
-from typing import Any, Dict, List, Optional, Sequence, Union, cast
+from typing import Any, Callable, Dict, List, Optional, Sequence, Union, cast
 
 import aqt
 from anki.cards import Card, CardId
@@ -32,7 +32,13 @@ class DataModel(QAbstractTableModel):
     _stale_cutoff -- A threshold to decide whether a cached row has gone stale.
     """
 
-    def __init__(self, col: Collection, state: ItemState) -> None:
+    def __init__(
+        self,
+        col: Collection,
+        state: ItemState,
+        row_state_will_change_callback: Callable,
+        row_state_changed_callback: Callable,
+    ) -> None:
         QAbstractTableModel.__init__(self)
         self.col: Collection = col
         self.columns: Dict[str, Column] = dict(
@@ -44,6 +50,8 @@ class DataModel(QAbstractTableModel):
         self._rows: Dict[int, CellRow] = {}
         self._block_updates = False
         self._stale_cutoff = 0.0
+        self._on_row_state_will_change = row_state_will_change_callback
+        self._on_row_state_changed = row_state_changed_callback
         self._want_tooltips = aqt.mw.pm.show_browser_table_tooltips()
 
     # Row Object Interface
@@ -59,15 +67,34 @@ class DataModel(QAbstractTableModel):
         if row := self._rows.get(item):
             if not self._block_updates and row.is_stale(self._stale_cutoff):
                 # need to refresh
-                self._rows[item] = self._fetch_row_from_backend(item)
-                return self._rows[item]
+                return self._fetch_row_and_update_cache(index, item, row)
             # return row, even if it's stale
             return row
         if self._block_updates:
             # blank row until we unblock
             return CellRow.placeholder(self.len_columns())
         # missing row, need to build
-        self._rows[item] = self._fetch_row_from_backend(item)
+        return self._fetch_row_and_update_cache(index, item, None)
+
+    def _fetch_row_and_update_cache(
+        self, index: QModelIndex, item: ItemId, old_row: Optional[CellRow]
+    ) -> CellRow:
+        """Fetch a row from the backend, add it to the cache and return it.
+        Thereby, handle callbacks if the row is being deleted or restored.
+        """
+        new_row = self._fetch_row_from_backend(item)
+        # row state has changed if existence of cached and fetched counterparts differ
+        # if the row was previously uncached, it is assumed to have existed
+        state_change = (
+            new_row.is_deleted
+            if old_row is None
+            else old_row.is_deleted != new_row.is_deleted
+        )
+        if state_change:
+            self._on_row_state_will_change(index, not new_row.is_deleted)
+        self._rows[item] = new_row
+        if state_change:
+            self._on_row_state_changed(index, not new_row.is_deleted)
         return self._rows[item]
 
     def _fetch_row_from_backend(self, item: ItemId) -> CellRow:
--- a/qt/aqt/browser/table/model.py
+++ b/qt/aqt/browser/table/model.py
@@ -80,7 +80,7 @@ class DataModel(QAbstractTableModel):
         self, index: QModelIndex, item: ItemId, old_row: Optional[CellRow]
     ) -> CellRow:
         """Fetch a row from the backend, add it to the cache and return it.
-        Thereby, handle callbacks if the row is being deleted or restored.
+        Then fire callbacks if the row is being deleted or restored.
         """
         new_row = self._fetch_row_from_backend(item)
         # row state has changed if existence of cached and fetched counterparts differ
--- a/qt/aqt/browser/table/model.py
+++ b/qt/aqt/browser/table/model.py
@@ -175,6 +175,8 @@ class DataModel(QAbstractTableModel):
 
     def get_card(self, index: QModelIndex) -> Optional[Card]:
         """Try to return the indicated, possibly deleted card."""
+        if not index.isValid():
+            return None
         try:
             return self._state.get_card(self.get_item(index))
         except NotFoundError:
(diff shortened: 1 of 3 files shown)


qt/aqt/browser/table/model.py
```diff
@@ -59,15 +67,34 @@ def get_row(self, index: QModelIndex) -> CellRow:
...
             return CellRow.placeholder(self.len_columns())
         # missing row, need to build
-        self._rows[item] = self._fetch_row_from_backend(item)
+        return self._fetch_row_and_update_cache(index, item, None)
+
+    def _fetch_row_and_update_cache(
+        self, index: QModelIndex, item: ItemId, old_row: Optional[CellRow]
+    ) -> CellRow:
+        """Fetch a row from the backend, add it to the cache and return it.
+        Thereby, handle callbacks if the row is being deleted or restored.
```
dae, line 83:
```suggestion
        Then fire callbacks if the row is being deleted or restored.
```

qt/aqt/browser/table/table.py
```diff
@@ -201,6 +207,30 @@ def to_first_row(self) -> None:
...
         self._move_current_to_row(self._model.len_rows() - 1)
 
+    def to_row_of_unselected_note(self) -> None:
+        """Select and set focus to a row whose note is not selected,
+        starting with the nearest row below, then above the focused row.
+        If that's not possible, clear selection.
+        """
+        nids = self.get_selected_note_ids()
+        for row in range(self._current().row(), self.len()):
+            nid = self._model.get_note_id(self._model.index(row, 0))
```
dae, line 217: This results in a 'select distinct nid' query for each row it tries, and could potentially be a bit expensive if the user has selected a large number of notes with the current row being at the top. It's a shame we can't use selectionModel().selectedRows() - it querying flags is a pain. I imagine we could work around it by implementing our own selection model instead of relying on Qt's, but probably not worth it at the moment.

dae:
Looks good - thanks for looking into this Rumo!

qt/aqt/browser/table/model.py
dae: Just a minor tweak to the comment and I think it's good to go :-)

qt/aqt/browser/table/table.py
dae: Hmm, one idea would be place an upper bound on the number of rows we try - or maybe we could even just try one above and one below the current row, and give up otherwise?

RumovZ: What if we fetch the corresponding card ids first (in cards mode), so we can compare these instead? Of course, that would be some unnecessary work too, if the next available row is just around the corner, but maybe it's fast enough?

RumovZ:
My last commit improves performance significantly when removing rows from a large selection. However, it's still slow when rows are added to a large collection, and deselecting many rows even got worse (Ctrl+A, then clicking a single row). I'll see if I can improve that.

May it now be worth considering to teach the backend fetching multiple rows at once?
When deleting rows, we can't move to the next row *afterwards*, because that's very slow if a lot of rows got deleted, leading to a lot of rows being fetched before a suitable candidate is found. However, if we could query multiple rows at once, I assume that would be a speed-up especially in the case of deleted rows. Do you think such an optimisation would be noticeable enough to be worth it?

By the way, do you plan to replace Qt completely in the long run?

qt/aqt/browser/table/table.py
dae: Hmm, having looked at the code, we're already incurring the flags() cost when we call get_selected_note_ids() at the start of the deletion. If we fetched the rows once and used them both to get the selected note ids and also passed them into to_row_of_unselected_note(), we could potentially use the min-1/max+1 method, which should be considerably faster in the reversed selection case. I just did a quick benchmark - selecting 180k cards from bottom to top of the list, the call to get_selected_note_ids() takes about 8 seconds, and the call to to_row_of_unselected_note() takes about 80 seconds.

The 8 seconds to get selected cards is ugly in itself unfortunately - Qt's selection model is not particularly efficient, and Qt's flags()/data() callback mechanism was designed for C++ where functions are much cheaper to invoke.

dae:
It might help, but I fear it would not solve our selection issues by itself. Pre-fetching all the rows first and doing a brief benchmark would give you an idea of how much could be saved vs how much is not the fetching code but something else.

We won't be able to move away from it any time soon, but I did do a prototype of the browser table in a web view about 2 years back, and it performed considerably better - one day I'd like to return to that and look into it more carefully. The downside is having to implement a lot of things we get for free with Qt.

The  Qt 6.2 upgrade is around the corner, so it will be interesting to see how easy or difficult that proves to be.

qt/aqt/browser/table/table.py
RumovZ: grafik

Looks like it's rather my fault than Qt's in this case. The loops are a bad choice, if there is no real upper limit.
Maybe we should only call `to_row_of_unselected_note()` when `len(nids) == 1`. That's when this feature is the most useful anyway, and we can guarantee that the loops won't run more than a couple of times.

> Qt's selection model is not particularly efficient, and Qt's flags()/data() callback mechanism was designed for C++ where functions are much cheaper to invoke.

It especially pains me to see how Qt is checking whether a row is selected: One call to `columnCount()` and `columnCount()` many calls to `flags()`...

dae: n=1 would cover the common case, though if the user is navigating with the keyboard, selecting a few rows, deleting them, and then navigating again, the PR's current solution is nice in that it doesn't force the user to start scrolling again from the top. I think we could preserve that behaviour with the min/max solution for a similar performance, but don't have strong feelings about it - your call. :-)

RumovZ: Maybe I've misunderstood, but wouldn't your suggestion only improve things if there are undeleted rows close above?
For the SnakeViz screenshot above I deleted all of 90,000 visible rows. It took 86 seconds, of which at least 60 can be attributed to the overhead in `to_row_of_unselected_note()`, i.e. excluding the time it took to get note ids.

We can also use a more generous boundary, like 10. Since most users don't have notes with more than a couple of cards, the additional computation time will probably hardly ever show.

dae: Rough proof of concept:

```
get selected row 7.878262996673584
get selected nid 0.39577388763427734
update rows      7.630562782287598
```

```diff
diff --git a/qt/aqt/browser/browser.py b/qt/aqt/browser/browser.py
index 2cf2c1b19..4b5bf3b0e 100644
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -619,8 +619,14 @@ class Browser(QMainWindow):
         if focus != self.form.tableView:
             return
 
-        nids = self.table.get_selected_note_ids()
-        self.table.to_row_of_unselected_note()
+        import time
+        t = time.time()
+        rows = self.table.get_selected_row_indices()
+        print("get selected row", time.time()-t); t = time.time()
+        nids = self.table.get_note_ids_from_rows(rows)
+        print("get selected nid", time.time()-t); t = time.time()
+        self.table.to_row_of_unselected_note(rows)
+        print("update rows     ", time.time()-t)
         remove_notes(parent=self, note_ids=nids).run_in_background()
 
     # legacy
diff --git a/qt/aqt/browser/table/table.py b/qt/aqt/browser/table/table.py
index 8e7585c40..2a410041a 100644
--- a/qt/aqt/browser/table/table.py
+++ b/qt/aqt/browser/table/table.py
@@ -107,6 +107,12 @@ class Table:
     def get_selected_note_ids(self) -> Sequence[NoteId]:
         return self._model.get_note_ids(self._selected())
 
+    def get_selected_row_indices(self) -> List[QModelIndex]:
+        return self._selected()
+
+    def get_note_ids_from_rows(self, rows: List[QModelIndex]) -> Sequence[NoteId]:
+        return self._model.get_note_ids(rows)
+
     def get_card_ids_from_selected_note_ids(self) -> Sequence[CardId]:
         return self._state.card_ids_from_note_ids(self.get_selected_note_ids())
 
@@ -207,24 +213,20 @@ class Table:
     def to_last_row(self) -> None:
         self._move_current_to_row(self._model.len_rows() - 1)
 
-    def to_row_of_unselected_note(self) -> None:
+    def to_row_of_unselected_note(self, selected_rows: List[QModelIndex]) -> None:
         """Select and set focus to a row whose note is not selected,
         starting with the nearest row below, then above the focused row.
         If that's not possible, clear selection.
         """
-        nids = self.get_selected_note_ids()
-        for row in range(self._current().row(), self.len()):
-            nid = self._model.get_note_id(self._model.index(row, 0))
-            if nid is not None and nid not in nids:
-                self._move_current_to_row(row)
-                return
-        for row in range(self._current().row() - 1, -1, -1):
-            nid = self._model.get_note_id(self._model.index(row, 0))
-            if nid is not None and nid not in nids:
-                self._move_current_to_row(row)
-                return
-        self.clear_selection()
-        self.clear_current()
+        bottom = max(*(r.row() for r in selected_rows)) + 1
+        if bottom < self._model.len_rows():
+            self._move_current_to_row(bottom)
+        else:
+            # todo: try top, clear otherwise
+            pass
+
+        # self.clear_selection()
+        # self.clear_current()
 
     def clear_current(self) -> None:
         self._view.selectionModel().setCurrentIndex(
```

dae: That should let you jump over the selected cards and start at the last/first one. It could be combined with your current approach to avoid siblings, and instead of having to potentially check the entire selection, you only need to check starting from the top and bottom of it.

RumovZ: Thanks, now I get it. I'll implement that.

RumovZ:
Adding rows to an already large selection is still slow. Something is triggering `flags()` for a large, arbitrary seeming subset of all visible rows. There doesn't seem to be anything we can do about it.

>Do I have to add a legacy alias for browser.onRowChanged()?

If not, I should be done here. 🙂

dae:
Thanks Rumo, have been fighting with JavaScript issues over the last few days, and missed that question last time. Grepping through add-ons, there do appear to be some that use onRowChanged, so an alias would be good. They appear to ignore the argument values, but will still expect two arguments, so we could probably get away with passing None/None. Will merge this in for now; if you could follow up with the alias that would be great.

RumovZ:
> but will still expect two arguments, so we could probably get away with passing None/None

Not sure I understand. I had something like this in mind:
```py
onRowChanged(self, _arg1, _arg2):
    self.on_row_changed()
```
If you say "passing" it sounds like you want to call the deprecated method. Is this about add-ons monkeypatching `onRowChanged()`?

While we're at it, `on_row_changed()` should actually be split in two, and the new half connected to `currentChanged()`. It usually doesn't matter, because it's rare to change the current element without triggering `selectionChanged()`. But if you Ctrl-rightclick a new row, you end up with the wrong context menu.
I guess now would be a good time to do it properly.

dae:
There're a couple of add-ons that are wrapping it and using it like a hook, so we could have the new func call the old one with (None, None) in order to trigger them. But that said, that's no guarantee they'll actually work correctly, and a hook for row changing has already been around for 3 years - so maybe it's not worth it in this case.

But the other change sounds good - better to fix this things now instead of having another breakage when we get around to fixing it later on :-)

RumovZ:
Also, accomodating both use cases, calling and wrapping the old method, further complicates things, so I'd vote to only add a simple alias. Maybe some of these add-ons are broken since 2.1.45 anyway.

PR #1394, opened 2021-09-27
Strip isolation characters from filenames in Check Media screen

abdnh:
Unicode isolation characters here can prevent searching for the files in the browser if copied, which is a very easy thing to do accidentally.
May be related: https://forums.ankiweb.net/t/how-to-detect-which-card-the-following-files-are-referenced-by-cards-but-were-not-found-in-the-media-folder-refers-to/11964


dae:
Good catch, thanks Abdo!

PR #1398, opened 2021-09-30
update to latest rules_nodejs & switch to ts_project

dae:
ts_library() is deprecated and will presumably be dropped from a
future rules_nodejs, and it wasn't working with the jest tests
after updating, so we switch over to ts_project().

There are some downsides:

- It's a bit slower, as the worker mode doesn't appear to function
at the moment.
- Getting it working with a mix of source files and generated files
was quite tricky, especially as things behave differently on Windows,
and differently when editing with VS Code. Solved with a small patch
to the rules, and a wrapper script that copies everything into the
bin folder first. To keep VS Code working correctly as well, the built
files are symlinked into the source folder.
- TS libraries are not implicitly linked to node_modules, so they
can't be imported with an absolute name like "lib/proto" - we need
to use relative paths like "../lib/proto" instead. Adjusting "paths"
in tsconfig.json makes it work for TS compilation, but then it fails
at the esbuild stage. We could resolve it by wrapping the TS
libraries in a subsequent js_library() call, but that has the downside
of losing the transient dependencies, meaning they need to be listed
again.  Alternatively we might be able to solve it in the future by
adjusting esbuild, but for now the paths have been made relative to
keep things simple.

Upsides:

- Along with updates to the Svelte tooling, Svelte typing has improved.
All exports made in a Svelte file are now visible to other files that
import them, and we no longer rebuild the Svelte files when TS files
are updated, as the Svelte files do no type checking themselves, and
are just a simple transpilation. Svelte-check now works on Windows again,
and there should be no errors when editing in VS Code after you've
built the project. The only downside seems to be that cmd+clicking
on a Svelte imports jumps to the .d.ts file instead of the original now;
presumably they'll fix that in a future plugin update.
- Each subfolder now has its own tsconfig.json, and tsc can be called
directly for testing purposes (but beware it will place build products
in the source tree): ts/node_modules/.bin/tsc -b ts
- We can drop the custom esbuild_toolchain, as it's included in the
latest rules_nodejs.

Other changes:

- "image_module_support" is moved into lib/, and imported with
`<reference types=...>`
- Images are now imported directly from their npm package; the
extra copy step has been removed.

@hgiesel I wanted to give you a heads up so we don't end up with too many conflicts over the work in #1387. I've manually merged most of the conflicts in https://github.com/ankitects/anki/tree/ts_project_merge, but the use of subfolders will need a bit of thought - will look into it tomorrow


hgiesel:
The subfolder thing seems very unfortunate. Going forward I would have liked to use more directories rather than less. Especially if now `.svelte` files will have their `.d.ts` right next to them.

I'm also slightly  confused how I can apply these resolved merge conflicts on top of my PR as their somewhat mixed in with merge conflicts unrelated to the PR - maybe my git expertise is a bit lacking here ^^;

dae:
I wasn't suggesting we remove the folders, was just wondering whether they should be built as part of the editor project, or as separate subprojects. Since they're somewhat small, I think we'll probably save a bit of time merging them into the parent editor project, since changes to lib tends to cause all other modules to need a recompile anyway, and each extra typescript project adds some invocation overhead. I've done so in https://github.com/ankitects/anki/commit/173fad51f78e97083a06241c79317a7939cbcdaa

Squashing the existing changes down into a single commit before merging them into the main branch will make merging a lot easier, so that's what I'd recommend we do in this case. We should probably squash more frequently, as currently the larger/more experimental changes have lots of commits that don't build, or don't pass the tests. This is not ideal, as when we need to use 'git bisect' to identify an issue, it's inconvenient when some commits won't build/pass.

dae:
Will get back to you with rebase steps soon

dae:
This should get you a new single commit relative to the current main branch that is easy to merge, and we can squash it again later if necessary:

```
# starting from commit 173fad5
dae@dip:Work/code/dtop% git co ts_project_merge
Switched to branch 'ts_project_merge'
dae@dip:Work/code/dtop% git log main..ts_project_merge  > ~/log.txt
dae@dip:Work/code/dtop% git co -b squashed
Switched to a new branch 'squashed'
dae@dip:Work/code/dtop% git reset main
[...]
dae@dip:Work/code/dtop% git add .
dae@dip:Work/code/dtop% git commit -F ~/log.txt

# then edit the commit message to include what you want to keep:
dae@dip:Work/code/dtop% git commit --amend
```

dae:
Re the extra symlinked files in the folders, I've gotten rid of them in https://github.com/ankitects/anki/commit/046c6c55d3b44f90f5638a188ffa108de29e7205

PR #1399, opened 2021-09-30
Add the homepage property to manifest.json

abdnh:
This adds the `homepage` property to add-ons' manifest.json file to allow providing an add-on page as an alternative to an AnkiWeb page for add-ons installed locally. I saw the `homepage` property used in some of glutanimate's add-ons' manifests and thought that it's a good idea to use it for the View Add-on Page button when no AnkiWeb ID is available.
I also removed the `addons-addon-was-not-downloaded-from-ankiweb` string since it looks redundant because the View Add-on Page button is disabled when no add-on page is available.

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -110,6 +110,7 @@ class AddonMeta:
     branch_index: int
     human_version: Optional[str]
     update_enabled: bool
+    homepage: Optional[str]
 
     def human_name(self) -> str:
         return self.provided_name or self.dir_name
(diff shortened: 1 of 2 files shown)


qt/aqt/addons.py
```diff
@@ -816,30 +825,28 @@ def _onAddonItemSelected(self, row_int: int) -> None:
         gui_hooks.addons_dialog_did_change_selected_addon(self, addon)
         return
 
-    def selectedAddons(self) -> List[str]:
+    def selectedAddons(self) -> List[AddonMeta]:
```
dae, line 828: There are some add-ons out there that call this function and will break if the signature changes. I recommend you make a new method called selected_addon_meta() instead, as that should avoid breaking things, and you won't have to change as many things in this PR.

dae (commented):
Thanks Abdo, change sounds good. One issue I spotted:

qt/aqt/addons.py
abdnh: Done.

dae: Nice and clean, thank you :-)

PR #803, opened 2020-10-24
Work around Qt's focus bug (fix help-wanted#22)

RumovZ:
Manually hide the editor's completer popup before accepting a short cut which changes the focus widget. Thus, work around the bug in PyQt 5.15 where a QLineEdit with open QCompleter popup won't get notified about focus changes handled by the Main Window, leading to inconsistent focus.

PR #1400, opened 2021-09-30
Remove workaround for Qt completer bug

RumovZ:
Reverts  #803.
This was causing errors when trying to use jump actions in the browser.


dae:
Thanks Rumo!

@hgiesel if editor.hideCompleters() is no longer safe to call, maybe we should turn it into a no-op? Would be good to tidy up / mark deprecated any other tag-related calls too, similar to 886ce15f2759fb2978ac574b9b22c63aa4657ef7

PR #1403, opened 2021-10-02
Translate Editor entirely to Svelte 2

hgiesel:
Continuation of #1387 

This seems about right. I'll try to turn this into something mergeable ASAP :)


kleinerpirat:
image

Love that move of the HTML editor button to the field header. It's something I was tempted to suggest too. Also, that live preview - Svelte paying off it seems!
It's all coming together :blush:

The field header has also become a great place for UI elements of add-ons that add functionality to the editor. Any plans for an API to add custom field state buttons?

hgiesel:
This PR does really feature much in terms of extensibility. I do have a vision for it however. I assume most of the Qt hooks for the editor will be deprecated, or can no longer guarantee what their name implies.
What we can offer instead is a hook system in TS.

One complication with hooks in TS is a egg - hen issue in terms of loading of scripts in the web view:
1. If add-on scripts load first, they will not have access to Anki APIs, like something like hooks
2. If we load Anki's code first, by the time add-ons scripts execute, the hooks might have already executed.

This will require some exploring on my side: e.g., if Anki's script executes, and schedules an asynchronous action (like waiting for i18n), will it only resumes after all other scripts have executed their synchronous actions? That's one thing I need to investigate.

Once this PR is merged, I will test this out, probably by porting New Format Pack, but also bigger add-ons (I intend to rewrite my occlusion editor for Closet).

ts/components/ButtonToolbar.svelte
```diff
@@ -101,7 +101,13 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         @include scrollbar.night-mode;
     }
 
-    .wrap-variable {
+    .button-toolbar {
         flex-wrap: var(--buttons-wrap);
+
+        > :global(*) {
+            /* TODO replace with gap once available */
+            margin-inline-end: 0.15rem;
```
dae, line 109: Just a quick comment for now - please note that there's no guarantee we can move to Qt6 across the platforms yet; it may be the case that a serious issue keeps one or more of the platforms back on 5.14 or .15, so we aren't free to adopt new Chromium features yet

hgiesel: That's unfortunate. The `gap` one is the most problematic, because margins won't work if the child has `display: contents`.

ts/sveltelib/store-subscribe.ts
```diff
@@ -0,0 +1,40 @@
...
+// License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+
+import type { Readable, Unsubscriber } from "svelte/store";
+
+interface StoreAccessors {
+    subscribe: () => void;
+    unsubscribe: () => void;
+}
+
+// Prevent double subscriptions / unsubscriptions
```
dae, line 11: How do they happen?

ts/components/ButtonToolbar.svelte
hgiesel: Oh, I see. I thought the original comment was about the `gap` property, but looking at caniuse, I guess it was about the logical properties. Swapping out with margin-left/bottom is no problem.

dae:
> > It would be nice if [focus and blur python hooks] could continue to be supported if practical.
> 
> The problem is that the underlying editor dramatically changed. I thought this might be better covered now through some hook in the JS, which maybe allows subscriptions to the editor code, etc..

Presumably it would still be possible to emulate the old API though, right? The blur bridgeCommand is still being sent - couldn't editor.py  check if any blur hooks are registered, run them if so, and then send a message back to the js to refresh the note? Are there complicating factors I've missed?

> > but since you're ignoring events while unfocused, doesn't that break the cycle?
> 
> Events are ignored while _focused_. This makes the blur event "correct" the field HTML to what it's "supposed" to look like, according to any filters etc.. Best way to illustrate this behavior is to enter invalid HTML (e.g. missing closing tag) into the HTML editor, and then blur.

So if the components disagree, it is possible for them to get stuck updating each other? I don't know the code well enough to know how likely that will be - if there's a non-negligible chance of that happening, it may be worth adding in some sort of circuit breaker?

qt/aqt/addcards.py
```diff
@@ -265,10 +265,6 @@ def _note_can_be_added(self, note: Note) -> bool:
         return True
 
     def keyPressEvent(self, evt: QKeyEvent) -> None:
-        "Show answer on RET or register answer."
-        if evt.key() in (Qt.Key_Enter, Qt.Key_Return) and self.editor.tags.hasFocus():
-            evt.accept()
-            return
```
dae, line 271: Probably better to remove the whole function, as it's not doing anything anymore.

ts/components/ButtonToolbar.svelte
dae: It was about gap - I presumed you were referring to https://caniuse.com/flexbox-gap

But it looks like margin-inline etc suffer from the same problem :-)

ts/editor/ClozeButton.svelte
```diff
@@ -5,20 +5,29 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+    import { get } from "svelte/store";
+    import {
+        getContext,
+        noteEditorKey,
+        activeInputKey,
+        focusInEditableKey,
+    } from "./context";
+
+    const noteEditor = getContext(noteEditorKey);
+    const activeInput = getContext(activeInputKey);
```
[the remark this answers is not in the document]
dae: "scopedProperty" could work too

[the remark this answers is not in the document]
hgiesel: It sounds good. What is nice about `context.ts`, that it would have been very straightforward to export for add-ons using `registerPackage`. But I guess also register those individual "context getters"

ts/editor/ClozeButton.svelte
```diff
@@ -5,20 +5,29 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
 
     const clozePattern = /\{\{c(\d+)::/gu;
     function getCurrentHighestCloze(increment: boolean): number {
         let highest = 0;
 
-        forEditorField([], (field) => {
-            const fieldHTML = field.editingArea.fieldHTML;
+        for (const field of noteEditor.fields) {
+            const content = field.editingArea?.content;
+            const fieldHTML = content ? get(content) : "";
```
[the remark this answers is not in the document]
hgiesel: Exporting the actual value of the store, we would have to export a function, like so:

```
Object.assign(api, {
    $content: () => $content,
})
```

Which would turn `get(component.subcomponent.value).subvalue` into `component.subcomponent.$value().subvalue`. I wasn't sure whether this is better, as it's somewhat obfuscated. It also increases the API surface by a bit.

Maybe we could write an extended version of the svelte built-in `writable`, and create a `gettable`, or I guess a `readGettable` and a `writeGettable`, which expose a `.get()` method, rather than having to use (and import) a function.

Actually another issue intertwined here is that of `.editingArea`. I exposed it as an optional property, but it would probably _more technically correctly_ but also maybe better to put expose it as a promise, because at the time `EditorField` sets the context, it is not available once it mounts. Especially if you access the component across other components, this might be an issue, as it's not always clear in which order components mount. Using a promise would turn the above code into:
```js
Promise.all(noteEditor.fields.map((field) => field.editingArea)).then(editingAreas => editingAreas.map(earea => get(earea.content)))
```
... Phew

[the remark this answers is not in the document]
dae: > Which would turn get(component.subcomponent.value).subvalue into component.subcomponent.$value().subvalue.

Ah, that seems worse.

> .. Phew

Indeed :-) If it's something that needs to be done regularly, maybe the top level editor could provide some convenience accessors so callers can do something like `await noteEditor.editingAreaForField(x)`? Not a recommendation, just brainstorming.

ts/editor/Codable.svelte
```diff
@@ -0,0 +1,156 @@
+<!--
+Copyright: Ankitects Pty Ltd and contributors
+License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+-->
+<script context="module" lang="ts">
+    import type { EditingInputAPI } from "./EditingArea.svelte";
+
+    export interface CodableAPI extends EditingInputAPI {
```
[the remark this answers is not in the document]
dae: I found the having Editable, Editor and Editing a bit confusing, so changing both sounds good to me :-) CodeInput/EditInput are not bad; some other possible ideas:

- RawInput
- PlainTextInput
- SourceCodeInput

- RenderedInput
- WYSIWYGInput
- VisualInput

ts/lib/promise.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright: Ankitects Pty Ltd and contributors
+// License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+
+/* eslint
+@typescript-eslint/no-non-null-assertion: "off",
+ */
+
+export function promiseResolve<T>(): [Promise<T>, (value: T) => void] {
```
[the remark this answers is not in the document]
dae: How about promiseWithResolver()? Resolve is a verb/action, which is why it trips me up.

ts/sveltelib/store-subscribe.ts
dae: I don't know the code well enough to fully follow that, but I'll take your word for it :-) Thanks for elaborating.

dae:
> So blur is sent again and the update mechanism should work. I just think it would be nice if add-ons could make use of live-update functionality that is now possible with Svelte stores. 

I'm not against exposing a better API, I just think it's best to avoid breaking existing use cases when it's practical (doesn't take a long time to implement, or make the code difficult to maintain/more fragile)..

> So at that point it might make sense to think about a circuit breaker

Thanks, that makes things clearer - sounds like implementing this is not so urgent.

ts/editor/ClozeButton.svelte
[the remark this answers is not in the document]
dae: Ok, #1434 should fix these. After rebasing over main, you'll need to add the deps to editable/editor (https://github.com/ankitects/anki/pull/1434/commits/bb5053990b86e0499afe7eebf358de9ce0dabb4d), and then build ts/...

You may also need to cmd+shift+p, Svelte: Restart Language Server after doing a build

If you spot further problems, please let me know.

ts/editor/ClozeButton.svelte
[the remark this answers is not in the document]
hgiesel: I agree some convenience accessors might be nice. But I'd prefer to implement these while also exploring writing add-ons with it myself. Otherwise it's hard to make judgements on what _might_ be useful.

hgiesel:
Okay from side, I think everything is good to go. Let me know if there are other show-stoppers.

ts/editor/ClozeButton.svelte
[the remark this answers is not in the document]
dae: Yep, totally agree

ts/editor/EditorField.svelte
```diff
@@ -0,0 +1,104 @@
...
+
+    export interface EditorFieldAPI {
+        element: HTMLElement;
+        index: number;
+        direction: "ltr" | "rtl";
+        editingArea?: EditingAreaAPI;
+    }
+
+    const key = Symbol("editorField");
+    const [set, getEditorField, hasEditorField] = contextProperty<EditorFieldAPI>(key);
```
dae, line 24: Not a change needed for this PR (if at all), but would it make sense to have contextProperty() take a string and create the symbol itself?

ts/editor/EditorField.svelte
```diff
@@ -0,0 +1,104 @@
...
+            {},
+            {
+                element: {
+                    get: () => editorField,
+                },
+                direction: {
+                    get: () => $directionStore,
+                },
+            }
+        ) as EditorFieldAPI
```
dae, line 62: Just out of curiosity, why is Object.create required here?

ts/lib/promise.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright: Ankitects Pty Ltd and contributors
+// License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+
+/* eslint
+@typescript-eslint/no-non-null-assertion: "off",
+ */
+
+export function promiseWithResolver<T>(): [Promise<T>, (value: T) => void] {
```
dae, line 8: Not a request for the current PR, but just a general comment - it would be nice to see a few more docstrings in future changes, especially on the more abstract/complicated parts. I'm guilty of not documenting my code as much as I should as well, so consider it advice to the both of us :-)

ts/editor/EditorField.svelte
hgiesel: So this:
```js
        element: {
            get: () => editorField,
        },
```

is necessary, because at the time of component instantiation, we don't have access to the DOM element `editorField` yet. Using `Object.create` (or `Object.defineProperties`) basically allows lazy loading these. This means you can use `editorField.element`, instead of `editorField.getElement()`. Same for `.direction` instead of `.direction()`. 

Both of these are up for refactoring very soon though. For any field preferences, I probably want to publish them as one object on editor field, and for `element, I already mentioned above that I probably should use promises for this.

dae:
Thanks Henrik, away we go! 🚀

hgiesel:
> so maybe they're Windows specific?

Unfortunately no ^^;
The textcolor thing was just an oversight, and I'm investigating the browser issue. It's not just searching, I think it has to with setting the editor with a new note type. I'm looking into it...

Issue #982, opened 2020-07-24
Qt 5.15 issues


dae:
This one is a note rather than a request for help.

- [ ] Slow scrolling performance on Windows - still a problem on 5.15.1:
https://forums.ankiweb.net/t/2-1-28-scroll-lags/1288

- [ ] Large images (eg 6500x4000) load more slowly on Windows 5.15.x: https://forums.ankiweb.net/t/large-images-load-slowly-on-2-1-34/3752/6

- [ ] Windows in tabs are no longer working on macOS - still an issue on 5.15.1: https://bugreports.qt.io/browse/QTBUG-85907

- [ ] On latest macOS, spaces are missing after commas and periods (eg, "...latest MacOS,spaces are missing..."). Fixed in 5.15.3... which is only available to commercial customers.

dae:
Closing this one in favour of #1427

PR #1405, opened 2021-10-03
Fix svelte imports

hgiesel:
Moving to `ts_project` has broken types for me on both VS Code and Neovim. This is an error message:

> [ts 2607] [E] Element does not support attributes because type definitions
 ⇒  are missing for this Svelte Component or element cannot be used as such.
>
> Underlying error:
JSX element class does not support attributes because it does not have a
⇒  '$prop_def' property.
─────────────────────────────────
[ts 2786] [E] Type definitions are missing for this Svelte Component. It
⇒  needs a class definition with at least the property '$prop_def' which
⇒  should contain a map of input property definitions.
Example:
  class ComponentName { $prop_def: { propertyName: string; } }
If you are using Svelte 3.31+, use SvelteComponentTyped:
  import type { SvelteComponentTyped } from "svelte";
  class ComponentName extends SvelteComponentTyped<{propertyName: string;}> {}
> 
> Underlying error:
'ButtonGroup' cannot be used as a JSX component.
  Its instance type 'ButtonGroup__SvelteComponent_' is not a valid JSX
⇒  element.
    Property '$prop_def' is missing in type 'ButtonGroup__SvelteComponent_'
⇒  but required in type 'ElementClass'.

This PR fixes types for me. Obviously we should find a solution that fixes it for everybody :)

To fix the typings for svelte files, I had to move the `rootDirs` parameter inside the sub projects.
To fix the typings for `tr` and `backend_proto` I had to reinstate the symlinks you originally had added.

I also used the opportunity to rewrite `i18n` quite a bit. The old `i18n_helpers` is now the main `i18n` file. As the generated files are more difficult to handle, I thought it would make more sense, if the other subprojects depended on the one actual file, not on generated files. I also removed the singleton class and replaced it with ordinary functions.

### DIFF
--- a/ts/deck-options/SaveButton.svelte
+++ b/ts/deck-options/SaveButton.svelte
@@ -3,7 +3,7 @@ Copyright: Ankitects Pty Ltd and contributors
 License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 -->
 <script lang="ts">
-    import * as tr from "../lib/i18n";
+    import { tr } from "../lib/i18n";
     import { createEventDispatcher } from "svelte";
     import type { DeckOptionsState } from "./lib";
     import type Dropdown from "bootstrap/js/dist/dropdown";
--- a/ts/deck-options/SaveButton.svelte
+++ b/ts/deck-options/SaveButton.svelte
@@ -29,6 +29,12 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         }
     }
 
+    /// Treat text like HTML, merging multiple spaces and converting
+    /// newlines to spaces.
+    export function collapseWhitespace(s: string): string {
+        return s.replace(/\s+/g, " ");
+    }
+
     function removeConfig(): void {
         // show pop-up after dropdown has gone away
         setTimeout(() => {
@@ -41,7 +47,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                     ? tr.deckConfigWillRequireFullSync() + " "
                     : "") +
                 tr.deckConfigConfirmRemoveName({ name: state.getCurrentName() });
-            if (confirm(tr.i18n.withCollapsedWhitespace(msg))) {
+            if (confirm(collapseWhitespace(msg))) {
                 try {
                     state.removeCurrentConfig();
                 } catch (err) {
--- a/ts/lib/BUILD.bazel
+++ b/ts/lib/BUILD.bazel
@@ -6,6 +6,10 @@ load("@rules_python//python:defs.bzl", "py_binary")
 load("@py_deps//:requirements.bzl", "requirement")
 load("//ts:jest.bzl", "jest_test")
 
+_generated = [
+    "i18n-generated.ts",
+]
+
 protobufjs_library(
     name = "backend_proto",
     proto = "//proto:backend_proto_lib",
@@ -35,9 +39,11 @@ genrule(
 
 typescript(
     name = "lib",
-    generated = [
-        ":i18n-generated.ts",
-    ],
+    srcs = glob(
+        ["*.ts"],
+        exclude = _generated,
+    ),
+    generated = [":" + file for file in _generated],
     deps = [
         ":backend_proto",
         "@npm//@fluent/bundle",
@@ -52,9 +58,9 @@ typescript(
 # Tests
 ################
 
-prettier_test()
+prettier_test(exclude = _generated)
 
-eslint_test()
+eslint_test(exclude = _generated)
 
 jest_test(
     deps = [
--- a/ts/lib/BUILD.bazel
+++ b/ts/lib/BUILD.bazel
@@ -25,7 +25,7 @@ py_binary(
 
 genrule(
     name = "fluent_gen",
-    outs = ["i18n.ts"],
+    outs = ["i18n-generated.ts"],
     cmd = "$(location genfluent) $(location //rslib/i18n:strings.json) $@",
     tools = [
         "genfluent",
@@ -36,7 +36,7 @@ genrule(
 typescript(
     name = "lib",
     generated = [
-        ":i18n.ts",
+        ":i18n-generated.ts",
     ],
     deps = [
         ":backend_proto",
--- a/ts/lib/BUILD.bazel
+++ b/ts/lib/BUILD.bazel
@@ -6,8 +6,13 @@ load("@rules_python//python:defs.bzl", "py_binary")
 load("@py_deps//:requirements.bzl", "requirement")
 load("//ts:jest.bzl", "jest_test")
 
-_generated = [
-    "i18n-generated.ts",
+_i18n_files = [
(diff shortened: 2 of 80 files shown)


dae:
Strange, the typings are working fine for me. I have the latest "Svelte for VS Code" installed ( v105.4.1 ) and start it by opening the root Anki folder after building with 'bazel build //...' to ensure all the Svelte files have been generated. Might be worth trying a 'bazel clean' to clear out old files that are in the wrong location, and making sure you're opening the base folder rather than just a single subfolder? I don't mind the rootDir change if it doesn't break anything, but would like to understand what's wrong first.

The reason for the current i18n layout is because rexporting symbols breaks tree shaking:

dae@dip:Work/code/dtop% bazel build -c opt ts/graphs && du -hs bazel-bin/ts/graphs/graphs.js
304K	bazel-bin/ts/graphs/graphs.js
dae@dip:Work/code/dtop% gh pr checkout 1405
dae@dip:Work/code/dtop% bazel build -c opt ts/graphs && du -hs bazel-bin/ts/graphs/graphs.js
408K	bazel-bin/ts/graphs/graphs.js

I agree the current layout is a bit awkward, and no objections from me on a change if you can preserve the tree shaking aspect.

hgiesel:
[Here](https://github.com/evanw/esbuild/issues/1420#issuecomment-873786196) the author of esbuild himself says it:

> Tree shaking is supported with import * as x but not when x is then re-exported and re-imported.

I've done several `bazel clean` and `git clean` but to no avail, I can't get typing for `tr`. I still refactored i18n quite a lot. `tr` is now outside of the `i18n` directory.

I also changes tsconfig's `lib` and `module` to `es2020`.
[es2018](https://caniuse.com/?feats=mdn-javascript_builtins_regexp_dotall,mdn-javascript_builtins_regexp_lookbehind_assertion,mdn-javascript_builtins_regexp_named_capture_groups,mdn-javascript_builtins_regexp_property_escapes,mdn-javascript_builtins_symbol_asynciterator,mdn-javascript_functions_method_definitions_async_generator_methods,mdn-javascript_grammar_template_literals_template_literal_revision,mdn-javascript_operators_destructuring_rest_in_objects,mdn-javascript_operators_spread_spread_in_destructuring,promise-finally), [es2019](https://caniuse.com/?feats=array-flat,mdn-javascript_builtins_object_fromentries,mdn-javascript_builtins_string_trimstart,mdn-javascript_builtins_symbol_description,mdn-javascript_statements_try_catch_optional_catch_binding,mdn-javascript_builtins_json_stringify_well_formed_stringify,mdn-javascript_builtins_array_sort_stable,mdn-javascript_builtins_function_tostring_tostring_revision), and [es2020](https://caniuse.com/?feats=mdn-javascript_operators_optional_chaining,mdn-javascript_operators_nullish_coalescing,mdn-javascript_builtins_globalthis,es6-module-dynamic-import,bigint,mdn-javascript_builtins_promise_allsettled,mdn-javascript_builtins_string_matchall,mdn-javascript_statements_export_namespace,mdn-javascript_statements_import_meta) all seem to have enough support.

ts/deck-options/SaveButton.svelte
```diff
@@ -29,6 +29,12 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         }
     }
 
+    /// Treat text like HTML, merging multiple spaces and converting
+    /// newlines to spaces.
+    export function collapseWhitespace(s: string): string {
+        return s.replace(/\s+/g, " ");
```
dae, line 35: The reason this was in i18n is because it's useful when you have a multi-line ftl string that you want to show in a non-HTML context. In HTML the whitespace is collapsed, but in a text context, you end up with unwanted newlines. There's a with_collapsed_whitespace() in the Python code as well.

ts/tsconfig.json
```diff
@@ -17,23 +17,13 @@
...
-            "es2018.intl",
-            "es2019.array",
-            "es2018.promise",
-            "es2020.promise",
-            "es2019.string",
-            "dom",
-            "dom.iterable"
-        ],
+        "module": "es2020",
+        "lib": ["es2020", "dom", "dom.iterable"],
```
dae, line 21: Is this safe? The current minimum Qt target is 5.14, which is Chromium 77 - which does not support things like the nullish coalescing operator. Neither does iOS 12 / early iOS 13.

Does this change enable something important that we were missing, or was it just about making things neater?

ts/lib/BUILD.bazel
```diff
@@ -22,21 +16,38 @@ py_binary(
         requirement("stringcase"),
     ],
 )
+_i18n_generated = [
+    "translate.ts",
```
dae, line 20: No strong feelings here, but 'translate' feels a bit odd to me as it's a verb. Would you be open to a noun like 'strings' or 'messages' instead?

dae (commented):
Refactor looks good; some comments below.

I'm still puzzled by this symlink requirement though. Will test on another machine and get back to you.

dae:
Ok:

- installed VS code on other machine
- installed svelte add-on
- checked out latest main branch from this repo
- opened it in VS Code
- went to CongratsPage.svelte. Was prompted to install TS tools; did so and restarted. Was told extension modified, restarted VS code again.
- I can see the "../lib/i18n" line underlined in red because the sources haven't been built yet.
- I go to lib/proto.ts and see ./backend_proto is also underlined red.
- I go to editor/index.ts and see that the i18n and components imports are highlighted red
- I go to a terminal and run bazel build ... in the checkout folder
- I restart vs code
- I see that the lib/i18n and backend_proto references are now working, but I'm getting errors in DeckOptionsPage about dynamic svelte component.

- I dug around to find out why my other machine wasn't showing the DeckOptionsPage errors, and it turns out I had a ~/node_modules folder that was affecting things - when I renamed that and restarted VS Code, my main machine now shows the same error with SveltedTypedComponent. Is it possible your i18n/proto issues are due to a node_modules folder?

Any ideas about the DynamicSvelteComponent issues in DeckOptionsPage?

dae:
I suspect this is caused by our node_modules being in ts/ instead of the root folder. With a symlink all files stay in ts, but when we reference ../../bazel-bin it presumably hunts further up the tree, and results in it not being able to find libraries like 'svelte', breaking the definitions. Will look into moving the location tomorrow and see if it helps

ts/tsconfig.json
hgiesel: In that case no, I'll revert it. I only needed to add support for [`.fromEntries`](https://caniuse.com/?search=fromEntries).
Is there some reference list of platforms Anki aims to support. I basically went through the lists on `icanuse` and looked out for:
- support for Chromium 83
- support for iOS 14/15
in which case we could go with es2020.

dae: (it might be worth setting up eslint to warn us when we use something outside those versions - feel free to look into that if you're interested :-)

dae: #982 tracks the 5.15 issues

dae:
Want to squash and rebase this over main so we get a nice clean commit? :-)

PR #1406, opened 2021-10-05
PyQt6 support

dae:
By happy coincidence, today is Anki's 15th birthday 🎉 

- Builds, runs and passes automated tests
- Preserves compatibility with Qt 5.14+ for now

Todo before merging into main:

- [ ] more manual testing (please report any issues you discover)
- [ ] build and test packaged builds
- [ ] black text on black background in night mode

Other issues:

- [ ] graphics driver selection no longer relevant
- [ ] need to rework stdHtml() approach to get rid of security warnings
- [ ] new qt recording implementation needed
- [ ] esc key being swallowed by browser search area


alt-romes:
Happy birthday! 🎈

glutanimate:
Exciting stuff! Also, Happy Anki-birthday :partying_face: - 15 amazing years !

Regarding add-on support:

Do you think it would be feasible (or even sensible) to implement an import hook / other workaround that would redirect all `PyQt5` imports to `PyQt6` when running on a `PyQt6` Anki build? A lot of breakages would still remain, especially with scoped enums (if there's no way to create a shim around that), but it should help keep a big chunk of add-ons working.

Unfortunately a huge number of add-ons depend on hard-coded `PyQt5` imports. Even add-ons that import from `aqt.qt` in their own codebases can be affected due to `pyuic5`/`pyrcc5` generated code and, in some cases, third-party dependencies.

dae:
@RumovZ profile importing should be fixed now

@glutanimate The forms are likely going to need to be regenerated regardless, as they tend to include a bunch of enums in them. You may be able to inject PyQt5 aliases into sys.modules to get some add-ons working, and if it doesn't cause any speed/reliability issues, I'd be happy to accept a PR that does so

dae:
Ok, I've tidied this up and split off some of the packaging changes into a separate branch. Will merge this in now, as the Qt5-compatible scoped enums will only pass typechecking when done under Qt6. The produced wheels no longer depend on a particular Qt version, and we can use ./run-qt5 for testing with qt5, so even if a showstopper issue is discovered in Qt6 that prevents us from using it for releases, we shouldn't need to roll this back.

PR #1408, opened 2021-10-07
move node_modules into root folder [action required]

dae:
Recommend removing ts/node_modules folder before attempting to
build after this update.

This moves ts/node_modules into the root of the project to work around
https://github.com/ankitects/anki/pull/1405#issuecomment-936213861

Also fixes the sass errors shown when running scripts/svelte-check

### DIFF
--- a/ts/svelte/svelte.bzl
+++ b/ts/svelte/svelte.bzl
@@ -3,17 +3,6 @@ load("@build_bazel_rules_nodejs//:providers.bzl", "DeclarationInfo", "declaratio
 load("@io_bazel_rules_sass//:defs.bzl", "SassInfo")
 load("@build_bazel_rules_nodejs//:index.bzl", "js_library")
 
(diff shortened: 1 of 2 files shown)


hgiesel:
I think we're getting closer, but this still does not work for me.

With the PR as is, VS Code can now find definitions. E.g. when opening `ts/editor/TemplateButtons.svelte`, right-clicking on `ButtonGroup` and "Go to definition", will lead me to `bazel-bin/ts/editor/ButtonGroup.svelte.d.ts`. This means no more red squiggles, however the component's props/slots/events are still not typechecked.

In comparison to that, when removing `"."` from `rootDirs`, like I mentioned in #1405, the `Go to definition` will lead me to `ts/editor/ButtonGroup.svelte`, which enables typechecking on props/slots/events for me (however it breaks typechecking for `tr`).

PR #1409, opened 2021-10-07
Put sass into repo directory

hgiesel:
The only thing, which I does not complete for me now is SCSS. When using `@use "button-mixins";`, it just shows that it cannot find the file.

This PR moves `ts/sass/` into `sass`. With this, it Svelte can find the SASS files. It also reintroduced absolute paths in Svelte components. In [this PR / comment](https://github.com/ankitects/anki/pull/1332#issuecomment-912101364), I mentioned that it would be nice if add-on devs could include our SCSS files, and they can only do that if we use relative paths, however that does not include Svelte files, only files in `sass/`. This means:

- We should use relative paths in `sass/`, so they can included by add-on devs
- We should use absolute paths in any svelte components, so we can see errors.


dae:
Thanks Henrik, looks good. Might explore the npm thing a bit more later, but it's not a big deal if we need to stick to copying.

PR #1410, opened 2021-10-08
replace dockerfile

kaczmarj:
This commit updates the Dockerfile to work with anki's bazel-based build system.

The anki Dockerfile was originally added in #753 back in September 2020. The file was moved to `docs/Dockerfile` in 0d354da, with a note that the file had to be updated to work with anki's updated build system. The file `docs/Dockerfile` was removed in 7cd2e96.


kaczmarj:
@dae - thanks for the suggestions. i implemented your comments. final docker image size is 735 MB uncompressed by the way.

PR #1414, opened 2021-10-08
Switch CardInfoDialog to ts page

RumovZ:
As always, I appreciate any feedback, especially to the TS code, which I'm new to. Some questions:

1. Would you have a look at `card-info/BUILD.bazel`? I've copy-pasted it and don't really now what I'm doing there. I assume some deps can be removed.
2. Should more logic be moved from Rust to TS? At the moment, string formatting is completely done on the backend, so styling parts of a string differently is not possible. (Formerly, the date part of the datetime string in the revlog has been bold.) I guess that also depends on the next point.
3. Should the new `card_stats()` routine be usable by add-ons? I've added the response message to the `SKIP_DECODE` set, but I guess that's not intended for parts of the public API? 
4. Also related: I've removed the old `CardsStats` code, since the goal was to get rid of a dependency (askama), and the old frontend code is useless without the old backend routine. Calls to `col.card_stats()` will break as the signature has changed. I don't see any reason/way for legacy support here, but correct me if I'm wrong.
5. `cargo/update.py` fails due to an OS error ("file already exists"). I seem to recall that I had this problem the last time, but I can't remember how I've solved it... Do you have any ideas? `bazel clean` did not do the trick.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -86,6 +86,8 @@ Thore Tyborski <github.com/ThoreBor>
 Alexander Giorev <alex.giorev@gmail.com>
 Ren Tatsumoto <tatsu@autistici.org>
 lolilolicon <lolilolicon@gmail.com>
+Gesa Stupperich <gesa.stupperich@gmail.com>
+git9527 <github.com/git9527>
 
 ********************
 
--- a/docs/mac.md
+++ b/docs/mac.md
@@ -38,6 +38,17 @@ $ ln -sf /usr/local/bin/{python3.9,python}
 This linking will not work if you're using the system Python from Big Sur,
 which is one of the reasons why we recommend using Python from python.org.
 
+But even so, if you still have a Python 2 binary on your system linked to 
+/usr/bin/python, it can happen that Bazel uses it and the build process 
+errors. In that case you might want to try forcing Bazel to use the 
+Python 3 binary by putting the following into a file called user.bazelrc at 
+the top of this repo (assuming /usr/local/bin/python links to your Python 
+3.9 binary).
+
+```
+build --action_env=PYO3_PYTHON=/usr/local/bin/python
+```
+
 ## Running Anki during development
 
 From the top level of Anki's source folder:
--- a/pylib/anki/browser.py
+++ b/pylib/anki/browser.py
@@ -31,4 +31,4 @@ class BrowserConfig:
 
 class BrowserDefaults:
     CARD_COLUMNS = ["noteFld", "template", "cardDue", "deck"]
-    NOTE_COLUMNS = ["noteFld", "note", "noteCards", "noteTags"]
+    NOTE_COLUMNS = ["noteFld", "note", "template", "noteTags"]
--- a/pylib/anki/utils.py
+++ b/pylib/anki/utils.py
@@ -290,8 +290,9 @@ def platDesc() -> str:
             elif system == "Linux":
                 import distro  # pytype: disable=import-error # pylint: disable=import-error
 
-                r = distro.linux_distribution(full_distribution_name=False)
-                theos = f"lin:{r[0]}:{r[1]}"
+                dist_id = distro.id()
+                dist_version = distro.version()
+                theos = f"lin:{dist_id}:{dist_version}"
             else:
                 theos = system
             break
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -74,7 +74,10 @@ class AddCards(QDialog):
             on_notetype_changed=self.on_notetype_change,
         )
         self.deck_chooser = DeckChooser(
-            self.mw, self.form.deckArea, starting_deck_id=DeckId(defaults.deck_id)
+            self.mw,
+            self.form.deckArea,
+            starting_deck_id=DeckId(defaults.deck_id),
+            on_deck_changed=self.on_deck_changed,
         )
 
     def helpRequested(self) -> None:
@@ -116,6 +119,9 @@ class AddCards(QDialog):
 
     def show_notetype_selector(self) -> None:
         self.editor.call_after_note_saved(self.notetype_chooser.choose_notetype)
+
+    def on_deck_changed(self, deck_id: int) -> None:
+        gui_hooks.add_cards_did_change_deck(deck_id)
 
     def on_notetype_change(self, notetype_id: NotetypeId) -> None:
         # need to adjust current deck?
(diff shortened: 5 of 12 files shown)


dae:
Thanks Rumo, this is a great start.

> Should more logic be moved from Rust to TS? At the moment, string formatting is completely done on the backend

Yep, doing the formatting on the frontend would be good - something like CongratsInfoResponse where the backend just delivers the raw data, and the frontend is responsible for converting that into something to be shown to the user. It's faster to iterate on the presentation that way, and also means that the data can be used by other consumers. lib/ contains some helpers for time processing, but stuff specific to the card info screen is best kept in card-info/lib.ts (looking back over the congrats code, buildNextLearnMsg and getCongratsInfo() probably should have been placed in the congrats folder instead of lib/)

> Should the new card_stats() routine be usable by add-ons? I've added the response message to the SKIP_DECODE set, but I guess that's not intended for parts of the public API?

I've partially split that off into https://github.com/ankitects/anki/issues/1417

I recall at least one person asking how they could get card stats data as data instead of strings in the past, so it could potentially be useful to expose a routine to get CardStats (which you're already exporting publicly)

> Also related: I've removed the old CardsStats code, since the goal was to get rid of a dependency (askama), and the old frontend code is useless without the old backend routine. Calls to col.card_stats() will break as the signature has changed. I don't see any reason/way for legacy support here, but correct me if I'm wrong.

There are some add-ons that use the current code to show current and previous cards at once (including my https://github.com/ankitects/anki-addons/blob/main/code/cardstats/__init__.py, and  https://ankiweb.net/shared/info/1008566916), so it would be nice if we could still support those use cases:

- showing more than one card on a page
- adding extra info to the page

The former should be easy, and in fact all we really need to do is rename CardStatsPage to CardStats (since it makes no assumptions about being the only thing on the screen)

For the latter, DOM injection should already work, and we could potentially add an empty div or an API endpoint if that ends up being needed.

> cargo/update.py fails due to an OS error ("file already exists"). I seem to recall that I had this problem the last time, but I can't remember how I've solved it... Do you have any ideas? bazel clean did not do the trick.

I've never tried it on Windows, so it's possible it's always been broken. I can either make the change for you, or feel free to dig into it a bit to find out what command (eg cargo-raze) or instruction in update.py is printing that message

RumovZ:
> A small nitpick: Using start and end instead of left and right for text-align can improve right-to-left display a bit.

As a matter of fact, I had tried that, but it didn't work for `<th>` tags for some reason, so I chose the lazy solution of reverting to `left`. I'll investigate how to do it properly. 🙂
By the way, shouldn't the position of the vertical header also be mirrored for rtl languages? But I guess that would require JS.

> There are some add-ons that use the current code to show current and previous cards at once (including my https://github.com/ankitects/anki-addons/blob/main/code/cardstats/__init__.py, and https://ankiweb.net/shared/info/1008566916), so it would be nice if we could still support those use cases:

To be clear, it is okay to break these add-ons? Of course, providing a new way to implement their functionality is not a problem.

> I've never tried it on Windows, so it's possible it's always been broken.

I'm listed as a contributor to Cargo.lock, so it must have worked at least once. Quite vexing that I don't remember what I did back then... I'll dig into it, and should I make it work, I'll definitely add a comment to cargo/README.md this time.

dae (commented):
Wrote these the other day, but forgot to submit them

pylib/anki/stats.py
```diff
@@ -13,45 +13,8 @@
...
-        txt += f"<b>{k}</b></td><td>{v}</td></tr>"
-        return txt
-
-    def date(self, tm: float) -> str:
-        return time.strftime("%Y-%m-%d", time.localtime(tm))
-
-    def time(self, tm: float) -> str:
-        return self.col.format_timespan(tm, context=FormatTimeSpan.PRECISE)
-
-
```
dae, line 54: There are quite a few add-ons that make use of this, and it's small and isolated, so I'd recommend we keep this class and its methods in for now. 

Emulating the old API probably won't be too hard - will give it a try and get back to you.

dae:
Ok, please start by rebasing this over main, as switching back and forth is slow, due to the move of node_modules. It will also pull in https://github.com/ankitects/anki/commit/d33c66e19510f011d2643dc54484eed02414ae26, which we want so a HTML string will have the same origin as the resources.

I've posted a proof of concept for preserving the old API here: https://github.com/ankitects/anki/commits/jstest2

RumovZ:
Thanks for the help.

I had to remove the night mode part in your proof of concept. How do I import from aqt in pylib? I didn't know this was possible.
`col.cardStats()` had already been a legacy alias, by the way, but `col.card_stats()`, `col.cardStats()` and `CardStats.report()` should all work now.

I'd appreciate it if you removed the askama dependency for me now.

When this PR is done, what do you think of making the card info dialog non-modal? In the browser, it would be much more convenient for showing multiple cards consecutively, and it looks like there is demand for keeping it open during review as well.

ts/lib/proto.ts
```diff
@@ -5,9 +5,28 @@ import { anki } from "./backend_proto";
...
+import Generic = anki.generic;
 import Notetypes = anki.notetypes;
 import Scheduler = anki.scheduler;
 import Stats = anki.stats;
 import Tags = anki.tags;
 
 export { Stats, Cards, DeckConfig, Notetypes, Scheduler, Tags };
+
+export function unwrapOptionalNumber(
+    msg:
```
dae, line 17: Nice, that should be handy.

ts/card-info/lib.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright: Ankitects Pty Ltd and contributors
+// License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+
+import { Stats } from "../lib/proto";
+import { postRequest } from "../lib/postrequest";
+
+export async function getCardStats(cardId: number): Promise<Stats.CardStatsResponse> {
+    return Stats.CardStatsResponse.decode(
+        await postRequest("/_anki/cardStats", JSON.stringify(cardId))
```
dae, line 9: I'll push a follow-up patch that converts this to a dict instead of passing a bare argument, as it's easier to extend in the future.

ts/card-info/Revlog.svelte
```diff
@@ -0,0 +1,141 @@
...
+    .right {
+        text-align: end;
+    }
+
+    .center {
+        text-align: center;
+    }
+
+    .revlog-container {
+        margin: 4em -2em 0 -2em;
```
dae, line 121: A couple of issues with the styling on smaller displays - will post a follow-up patch with some more details.

PR #1415, opened 2021-10-08
Fix deck change after toggling sticky fields

abdnh:
`AddCards.on_notetype_change` was being called by `AddCards.on_operation_did_execute` because `initiator` was not specified, causing the current deck to change if the "Change deck depending on note type" preferences option is enabled.


dae:
Great, thanks abdo!

PR #1419, opened 2021-10-10
Add deck id into hook: add_cards_did_add_note

git9527:
main changes:
1. add target_deck_id into add_cards_did_add_note hook args
2. add legacy_arg_names in Hooks generator in case legacy hook takes only part of the new hook arguments


dae:
This will break any existing users of the non-legacy hook. To avoid breaking existing callers, you'd need to create a new hook. But do you really need it? You have access to the note, and can locate the deck of the first card fairly easily.

git9527:
how to locate the deck from the newly added note?

git9527:
thanks, you're right, I missed that part, close the PR now

PR #1420, opened 2021-10-10
add new hook: add_cards_did_change_deck

git9527:
will be executed after the user selects a new different deck when adding cards.


qt/aqt/deckchooser.py
```diff
@@ -27,6 +28,8 @@ def __init__(
         if starting_deck_id is None:
             starting_deck_id = DeckId(self.mw.col.get_config("curDeck", default=1) or 1)
         self.selected_deck_id = starting_deck_id
+        self.on_deck_changed = None
```
dae, line 31:
```suggestion
```

PR #1422, opened 2021-10-12
Fix invalid 'noteCards' browser column being used

abdnh:
I guess `noteCards` is a leftover from Rumo's work on the browser.

The column shows as added by an add-on.
image


dae:
Thanks Abdo!

PR #1431, opened 2021-10-16
Non modal card info

RumovZ:
Hope I've done this right.


ts/card-info/index.ts
```diff
@@ -31,3 +33,19 @@ export async function cardInfo(
         props: { stats },
     });
 }
+
+export async function updateCardInfo(
```
dae, line 37: Instead of adding another entry to the "global" anki namespace, I think you could probably export a method in the Svelte instance instead. For example:

```diff
diff --git a/qt/aqt/browser/card_info.py b/qt/aqt/browser/card_info.py
index 26f99426c..e8e0fffd8 100644
--- a/qt/aqt/browser/card_info.py
+++ b/qt/aqt/browser/card_info.py
@@ -64,11 +64,11 @@ class CardInfoDialog(QDialog):
         self.setLayout(layout)
 
         self.web.eval(
-            f"let cardInfo = anki.cardInfo(document.getElementById('main'), {card_id}, true);"
+            f"const cardInfo = anki.cardInfo(document.getElementById('main'), {card_id}, true);"
         )
 
     def update_card(self, card_id: CardId) -> None:
-        self.web.eval(f"anki.updateCardInfo(cardInfo, {card_id}, true);")
+        self.web.eval(f"cardInfo.then((c) => c.updateCard({card_id}, true));")
 
     def reject(self) -> None:
         if self._on_close:
diff --git a/ts/card-info/CardInfo.svelte b/ts/card-info/CardInfo.svelte
index e1f8e1e97..90d72483c 100644
--- a/ts/card-info/CardInfo.svelte
+++ b/ts/card-info/CardInfo.svelte
@@ -7,6 +7,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import { Stats, unwrapOptionalNumber } from "../lib/proto";
     import { Timestamp, timeSpan, DAY } from "../lib/time";
     import Revlog from "./Revlog.svelte";
+    import { getCardStats } from "./lib";
 
     export let stats: Stats.CardStatsResponse;
 
@@ -93,6 +94,13 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     let statsRows: StatsRow[];
     $: statsRows = rowsFromStats(stats);
+
+    export function updateCard(cardId: number, includeRevlog: boolean) {
+        stats = await getCardStats(cardId);
+        if (!includeRevlog) {
+            stats.revlog = [];
+        }
+    }
 </script>
 
 <div class="container">
diff --git a/ts/card-info/index.ts b/ts/card-info/index.ts
index db972562c..61cd7d9ac 100644
--- a/ts/card-info/index.ts
+++ b/ts/card-info/index.ts
@@ -7,8 +7,6 @@ import { checkNightMode } from "../lib/nightmode";
 
 import CardInfo from "./CardInfo.svelte";
 
-const _updatingQueue: Promise<void> = Promise.resolve();
-
 export async function cardInfo(
     target: HTMLDivElement,
     cardId: number,
@@ -33,19 +31,3 @@ export async function cardInfo(
         props: { stats },
     });
 }
-
-export async function updateCardInfo(
-    cardInfo: Promise<CardInfo>,
-    cardId: number,
-    includeRevlog: boolean
-): Promise<void> {
-    _updatingQueue.then(async () => {
-        cardInfo.then(async (cardInfo) => {
-            const stats = await getCardStats(cardId);
-            if (!includeRevlog) {
-                stats.revlog = [];
-            }
-            cardInfo.$set({ stats });
-        });
-    });
-}
```

- the stats.revlog = [] inside the Svelte component shouldn't cause flicker, as rendering does not happen until the next render loop
- was the goal of updatingQueue to serialize access? I'm not sure we need it, but if we do, I think you'd probably need to return a new promise (there's no implicit return for => { ... }).
- You have the most experience here @hgiesel - any thoughts?

dae (commented):
Looks great! Just some minor comments:

ts/card-info/index.ts
dae: What about storing a promise inside the Svelte component, and having the component doing its own fetching? The initial load case could then share the logic, creating the component and then calling updateCard() on it. Untested, but maybe something like the following?

```typescript
    let lastRequest: Promise<void> = Promise.resolve();

    export async function updateCard(cardId: number, includeRevlog: boolean) {
        lastRequest = lastRequest
            .then(() => getCardStats(cardId))
            .then((data) => {
                if (!includeRevlog) {
                    data.revlog = [];
                }
                stats = data;
            });
    }
```

RumovZ: I was thinking something similar. And given your comment above, `cardInfo()` should probably be callable without a card id as well. So we could just as well skip the call to `updateCall()` and leave it to the consumer to set a card.
For clarity, I imagine an API like this:

```ts
/* build the page with some placeholder string */
const cardInfo = anki.cardInfo(document.getElementById('main'));
/* set a card */
cardInfo.then((c) => c.updateCard(123456789, true));")
```

dae: Yep, that's what I was thinking. An alternative approach would be to have cardId and the bool passed in as props and have the component call its internal update function each time the props change. It would be more ergonomic if this component was going to be embedded in a larger component, but since we're currently just calling it programmatically, I'm not sure we'd gain much from doing that at the moment.

pylib/anki/stats.py
```diff
@@ -38,7 +39,8 @@ def _legacy_card_stats(
     if ({1 if _legacy_nightmode else 0}) {{
         document.documentElement.className = "night-mode";
     }}
-    anki.cardInfo(document.getElementById('{random_id}'), {card_id}, {include_revlog});
+    const {varName} = anki.cardInfo(document.getElementById('{random_id}'));
+    {varName}.then((c) => c.$set({{ cardId: {card_id}, includeRevlog: {str(include_revlog).lower()} }}));
 </script>
```
dae, line 44: Good catch, I'd forgotten about that.

qt/aqt/browser/card_info.py
```diff
@@ -47,10 +64,87 @@ def _setup_ui(self, card_id: CardId) -> None:
         self.setLayout(layout)
 
         self.web.eval(
-            f"anki.cardInfo(document.getElementById('main'), {card_id}, true);"
+            "const cardInfo = anki.cardInfo(document.getElementById('main'));\n"
+            "cardInfo.then((c) => c.$set({ includeRevlog: true }));"
```
dae, line 68: I think it would be simpler to give includeRevlog a default value of true in the component, instead of defaulting to uninitialized, and requiring the caller to set it.

qt/aqt/browser/card_info.py
```diff
@@ -47,10 +64,87 @@ def _setup_ui(self, card_id: CardId) -> None:
...
         )
+        self.update_card(card_id)
+
+    def update_card(self, card_id: CardId | None) -> None:
+        val = "undefined" if card_id is None else card_id
+        self.web.eval(f"cardInfo.then((c) => c.$set({{ cardId: {val} }}));")
```
dae, line 74: This is fine as is, but the approach I'd take is:

- default to null instead of undefined in the component, which I personally think slightly better represents "deliberately set to nothing" instead of "omitted"
- then instead of manually making string values, you can just use `cardId: {json.dumps(card_id)}` in the eval line, and it will convert the None to null.

qt/aqt/browser/card_info.py
```diff
@@ -47,10 +64,87 @@ def _setup_ui(self, card_id: CardId) -> None:
...
+    def _on_close(self) -> None:
+        self._dialog = None
+
+
+class BrowserCardInfo(CardInfoManager):
+    def __init__(self, mw: aqt.AnkiQt):
+        super().__init__(
+            mw,
+            "revlog",
+            without_unicode_isolation(
```
dae, line 125: Another good catch! I always miss this, as it's a Windows-only issue.

ts/card-info/CardInfo.svelte
```diff
@@ -3,108 +3,53 @@ Copyright: Ankitects Pty Ltd and contributors
...
+    let stats: Stats.CardStatsResponse | undefined;
 
-    if (stats.interval) {
-        statsRows.push({
-            label: tr2.cardStatsInterval(),
-            value: timeSpan(stats.interval * DAY),
+    $: if (cardId === undefined) {
+        stats = undefined;
+    } else {
+        const sentCardId = cardId;
```
dae, line 20:
```suggestion
        const requestedCardId = cardId;
```

Bikeshedding, feel free to ignore.

dae (commented):
A few more minor points

RumovZ:
Okay, this should be it. 🙂

PR #1432, opened 2021-10-17
New cards: add reversed insertion order option

jdpopkin:
Hi! I want to be able to ask Anki to assign me my new cards in _reverse_ insertion order (that is, newest cards first).[0] This pull request makes that possible.

Potential problems:

1. I had to mess with the config a bit. I'm not familiar with Protobuf, so I don't really know what kinds of problems that could cause. I hope it should be fine (this doesn't change the structure or width of any field, it's just adding another accepted-value to some enum abstracting over an int). But maybe I'm wrong about that (maybe Protobuf does something clever where it automatically represents enums with the fewest number of bits necessary to cover all the options?).

2. The corresponding code isn't in AnkiDroid yet, and it would be bad if AnkiDroid broke for users who select this option on desktop. (I'd be happy to implement this on AnkiDroid too - I've [touched its deck-sorting code before](https://github.com/ankidroid/Anki-Android/pull/8173). But even if both support this feature, there's no way to guarantee that everyone updates both their desktop and mobile clients simultaneously.) Maybe the initial PR should just add sorting/insertion code, but shouldn't actually expose this option for selection in the deck config UI until both clients support it?

3. Maybe this use-case is too niche? Anki's a big program - maybe there's already some feature I don't know about that would solve my problem for me.

[0] The reason I want this is because sometimes I find myself adding huge amounts of material all at once, which builds up a backlog that would take weeks to work through. As I'm working through the backlog, I might find myself creating a handful of new cards each day. When that happens, I'd prefer to prioritize learning those new cards first, since they're relatively fresh in my mind. Right now, I _think_ the closest I can get to that is using random card order and relying on its gradually-increasing bias in favor of newly-added cards. But I'd prefer something more consistent/reliable.

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -60,6 +60,7 @@ deck-config-new-insertion-order-tooltip =
     this option will automatically update the existing position of new cards.
 deck-config-new-insertion-order-sequential = Sequential (oldest cards first)
 deck-config-new-insertion-order-random = Random
+deck-config-new-insertion-order-sequential-reversed = Sequential (newest cards first)
 
 ## Lapses section
 
--- a/proto/anki/deckconfig.proto
+++ b/proto/anki/deckconfig.proto
@@ -30,6 +30,7 @@ message DeckConfig {
     enum NewCardInsertOrder {
       NEW_CARD_INSERT_ORDER_DUE = 0;
       NEW_CARD_INSERT_ORDER_RANDOM = 1;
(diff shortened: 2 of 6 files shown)


dae:
This is not correct as-is, as it does not alter the numbers assigned when cards are added. Anki uses a config var to track the highest due number, so a separate one would be required, and it would make this a larger change. I suggest you just switch to the v3 scheduler and change the sorting options in the display order instead: https://faqs.ankiweb.net/the-2021-scheduler.html

PR #1435, opened 2021-10-18
Use trailingComma: all setting in .prettierrc

hgiesel:
What do you think about using trailing commas in the TypeScript more liberally?


hgiesel:
Actually I intended to make this PR ever since moving prettier up to the repo directory, but was just waiting for it until the Svelte editor was merged in.
A little more uniformity among the different languages is nice, and trailing commas also lead to nicer diffs with small changes; those were my motivating factors :)

Issue #1256, opened 2021-06-27
PEP8 pylib


dae:
0ddd31638815478c2abbb91c4a9ab0331e95545b introduced some helpers so we can get automatic conversion from camelCase methods to snake_case and mark methods as deprecated, and fee486aaa1a8fd4fc301bc6a68ebd7f7784a0b1c started migrating some files so that they pass pylint's invalid-name tests. While the mapping of camelCase to snake_case is automatic, it's still a semi-manual process, as some methods use different names, or take different arguments.

It would be nice to continue doing this with the rest of pylib's codebase, so we can move away from a mix of camelCase and snake_case methods, and so we can improve on the use of variable names like 'x'.

We'll probably want to avoid the following files, as they've been replaced with newer code, or will soon be:

- the importing/exporting code
- the v1 and v2 scheduler code
- the old stats code

The existing helper works with classes, so it doesn't help us in files like utils.py that contain a lot of free-standing
functions. We can use a module-level __getattr__ like at the bottom of decks.py, but we'll probably want to write a helper
to make things easier, and to give us automatic camelCase lookups.

Given that the goal is to get a stable release of 2.1.45 out soon, it probably wasn't the best timing to be starting on these changes 😓 . I'm going to stop here for now; hopefully they will not cause any problems in the next beta. It might be best to hold off on further work on this until 2.1.45 has had a chance to stabilize.

dae:
Did #1451 make this closable?

RumovZ:
> Did #1451 make this closable?

Was this directed at me? I just saw it by chance.

dae:
@RumovZ sorry, I assumed you'd be notified like when you create a PR :-)

RumovZ:
Ah, okay. 🙂 Then to answer the question, I think this was finalised with #1454, provided it was only about naming and no other PEP8 conventions.

PR #1438, opened 2021-10-19
Enable registering of legacy attributes without exact substitutes

RumovZ:
#1256 discusses true aliases, but what if the legacy function and its replacement not only differ in name? E.g.:

```py
def oldFunc(args):
    return new_func(someExtraProcessing(args))

def someFunc(args):
    return some_func(someExtraProcessing(args))
```

This PR introduces a convention for the new snakecase legacy name and a mechanism for registering them with the correct replacements.
I'm not sure this is the best way to do this, but hope it at least illustrates the problem I'm trying to solve.


dae:
I fear I may not be understanding this correctly. Is this so that instead of

```
SomeClass.register_deprecated_aliases(oldFunc=SomeClass._legacy_old_func)
```
you can instead do

```
SomeClass.register_deprecated_attributes(oldFunc="new_func")
```
?

If I've not misunderstood - why not just use the former instead? It makes it clearer what is actually being done, and the build will fail if a method is referenced which is later deleted, which we don't get when using strings.

RumovZ:
If we used `SomeClass.register_deprecated_aliases(oldFunc=SomeClass._legacy_old_func)`, the warning would print "oldFunc is deprecated: please use _legacy_old_func instead", which would be wrong.
With the new function, an arbitrary substitute can be specified (`new_func`). The remapped name (`_legacy_old_func`) can be derived from the old name (`oldFunc`), so it doesn't need to be passed.

I don't see a reason why the replacement `new_func` should need to be passed as a string. It can be a callable, just as with the other method.
It's true though that a missing `_legacy_old_func` won't be detected. We could either add an assertion in `register_deprecated_attributes()` or, maybe more elegantly, pass `_legacy_old_func` instead of `oldFunc`. That is, we pass a tuple `(new_func, _legacy_old_func)` instead of a kwarg, and if either callable is missing, we will know at build time. `oldFunc` can be derived from `_legacy_old_func` and stored at build time, or we only store `_legacy_old_func` and derive this name from `oldFunc` when it is called at runtime.

I'll make a commit tomorrow. That's probably a lot easier to decipher than my last paragraph...

dae:
Good point about the message; sounds good.

RumovZ:
Okay, I've decided against deriving any names in this case. It's convoluted and error-prone, whereas passing an additional key word is no work at all and will work in any scenario.

PR #1440, opened 2021-10-20
Maintain compatibility with PyQt5 add-ons in PyQt6 builds

glutanimate:
(following up on #1406)

**Overview**

This PR combines a number of patches, aliases, and shims to address common add-on breakage scenarios on PyQt6. The goal is to provide a transitional period where support for existing add-on releases is maintained under PyQt6 builds of Anki, giving authors some extra time to update their add-ons to support PyQt6 natively.

Some of these solutions are not particularly pretty. However, by grepping for breakage patterns, I would estimate that >50% of all add-ons on AnkiWeb, >70% of all Qt add-ons, and >80% of the top 100 rated add-ons are currently broken to some degree on PyQt6. With that in mind, I would argue that introducing a compatibility layer like this is a reasonable way forward, at least transitionally.

**Breakdown of Workarounds and Potential Impact**

Taking a closer look at the changes, these are the main workarounds contained in this PR:

- Alias `PyQt5` modules to `PyQt6`. Should help fix ≈280 AnkiWeb add-ons.
- Alias a select subset of unscoped enums to PyQt6 scoped enums. Should help fix ≈390 AnkiWeb add-ons.
- Patch `exec_` methods back into various Qt classes. Should help fix ≈210 AnkiWeb add-ons.
- Various other aliases and patches that should cumulatively help fix ≈100 AnkiWeb add-ons.

These fixes overlap of course, as add-ons are typically affected by multiple breakages, but taken together the changes in this PR should restore compatibility for **>450 add-ons total**. From my tests so far, in most of these cases this means full compatibility with PyQt6 Anki builds.

However, there definitely are add-ons that were hit harder by more localized API changes, e.g. removal of the PyQt resource system (_Review Heatmap_, _AnJoy_, _Customize Background_, etc.) or changes to QtWebEngine (_Searching & PDF Reading_). Other add-ons like _Image Occlusion Enhanced_ are also affected by changes to Anki APIs and so cannot be fully restored to working condition by this PR alone.

Overall, though, I'm convinced that this PR will ease the transition to PyQt6 quite a bit.

**Additional Notes**

- As the number of aqt.qt* modules just kept increasing while I worked on the PR, I took the liberty to move all aqt.qt* related modules into a `qt` package. I hope that's OK and doesn't have an side-effects in terms of packaging!
- The PR currently aliases ≈1000 unscoped enums to their scoped counterparts which I selected by grepping for enums that are in use in add-ons at the moment. If we wanted to alias all enums, the total number would be ≈9000, but I don't see much benefit in that beyond supporting a few more add-ons that aren't on AnkiWeb and to potentially account for some blind spots in the grepping process.
- I briefly considered setting the enums dynamically, so that we can print out a deprecation warning. However, I felt that performance- and stability-wise, plain old assignments would be the best option for now.


qt/aqt/sound.py
```diff
@@ -671,7 +671,7 @@ def _start_recording(self) -> None:
                 namedtmp("rec.wav"), self.mw, self._parent
             )
         else:
-            from .qt5 import QtAudioInputRecorder as Qt5Recorder
+            from aqt.qt.qt5extra import QtAudioInputRecorder as Qt5Recorder
```
dae, line 674: Since this is a package now, maybe aqt.qt.qt5_audio or similar would be a better choice?

qt/aqt/qt/compat.py
```diff
@@ -0,0 +1,1265 @@
...
+QHostAddress.LocalHost = QHostAddress.SpecialAddress.LocalHost
+QPrinter.Millimeter = QPrinter.Unit.Millimeter
+
+# Globally alias removed PyQt5.Qt to PyQt6.QtCore.Qt
+##########################################################################
+
+from . import qt5qt
+
+sys.modules["PyQt5.Qt"] = qt5qt
+sys.modules["PyQt6.Qt"] = qt5qt
```
dae, line 1265: Could you explain what you're trying to accomplish here? If I've read it right, it will attempt to import PyQt5 after PyQt6 has loaded, which will cause crashes if both are available in the Python environment (at least on Macs)

PR #1443, opened 2021-10-21
PEP8 pylib

RumovZ:
I've made a start on #1256 and collected some questions:

- Is 5c7bf5e somewhat like what you had in mind? I guess the class name could be improved.
- If someone mistyped the name of an attribute of any class (or module) inheriting `DeprecatedNamesMixin`, only "AttributeError" and a traceback pointing to `_legacy.py` would be printed. 4f8a291 tries to reintroduce some error information. (Edit: All's fine now except for the traceback, but I think that's tolerable.)
- Some functions are never called, and I suspect a few of them aren't used by add-ons, either. Is there a way for me to corroborate that assumption without downloading the one thousand 2.1 add-ons from Ankiweb?
- Due to polyphormism, it's not possible to refactor v3.py without doing the same for v1 and v2, so should we avoid v3 as well for now? It looks like most methods there only exist to support the polymorphic `col.sched` and the old test code.

### DIFF
--- a/rslib/src/scheduler/reviews.rs
+++ b/rslib/src/scheduler/reviews.rs
@@ -124,7 +124,8 @@ impl Collection {
                             .ok_or(AnkiError::NotFound)?;
                         let ease = col
                             .get_deck_config(config_id, true)?
-                            .unwrap_or_else(Default::default) // cannot happen
+                            // just for compiler; get_deck_config() is guaranteed to return a value
+                            .unwrap_or_default()
                             .inner
                             .initial_ease;
                         decks_initial_ease.insert(deck_id, ease);


dae:
> Is 5c7bf5e somewhat like what you had in mind? I guess the class name could be improved.

Yep, looks good!

> If someone mistyped the name of an attribute of any class (or module) inheriting DeprecatedNamesMixin, only "AttributeError" and a traceback pointing to _legacy.py would be printed.
 4f8a291 tries to reintroduce some error information. (Edit: All's fine now except for the traceback, but I think that's tolerable.)

👍 

> Some functions are never called, and I suspect a few of them aren't used by add-ons, either. Is there a way for me to corroborate that assumption without downloading the one thousand 2.1 add-ons from Ankiweb?

I've built a snapshot of the latest add-on branches with all non-src files stripped out - you can grab it here: https://apps.ankiweb.net/downloads/archive/addons_src_only_22oct21.zip (68MB). I use ripgrep to search through it.  when I want to see if a function/variable/etc may be used or not.

If only a single add-on or two are using some otherwise unused code, I tend to look the add-ons up - if they're already not working on the latest Anki, then the code is safe to cut.

> Due to polyphormism, it's not possible to refactor v3.py without doing the same for v1 and v2, so should we avoid v3 as well for now? It looks like most methods there only exist to support the polymorphic col.sched and the old test code.

Yeah, probably best to just skip that area for now.

pylib/anki/_legacy.py
```diff
@@ -98,6 +107,37 @@ def register_deprecated_attributes(
         }
 
 
+class DeprecatedNamesMixinStandalone(DeprecatedNamesMixin):
```
dae, line 110: "DeprecatedModuleNamesMixin"? "DeprecatedNamesMixinForModule"? Open to other ideas :-)

dae:
I've made an updated snapshot: https://apps.ankiweb.net/downloads/addons_src_only_31aug22.zip

PR #1448, opened 2021-10-22
Fix PlainTextInput RTL

hgiesel:
Reported [here](https://github.com/ankitects/anki/pull/1403#issuecomment-948911784) by @abdnh


dae:
Thanks Henrik!

PR #1456, opened 2021-10-25
Update browsing.ftl

schmidhauser:
Englished a menu item.

In addition, there are several entries that still use sentence-case – e.g. “Ignore case”, “Randomize order”. Would it not be more consistent if one were to rewrite them in title-case (i.e. headline-style) – e.g. “Randomize Order”?


schmidhauser:
Yes for the latter point.

As for the former, that seems to be partly a question of style. It’s true that there exist words with the suffix *‑type* such as *archetype* and *prototype*; but these are not exactly parallel since they don’t have an independent first constituent (there’s no word *arche* or *proto*). More importantly, right now, in ordinary English, one generally writes *note type* – just as one writes *word type*, *sentence type*, etc.

Also, I’m not sure I understand your argument. Whether *note type* or *notetype*, the item referred to is the type of one or several notes, and as such can be considered a separate entity. Or in case I’ve missed something – what would be an example of a property of a notetype that is not a property of the corresponding note type?

In any event, if you use *notetype*, then I think you should use it throughout – e.g. in the menu item “Manage Notetypes”. Likewise, you should then probably also use *cardtype*.

dae:
> Or in case I’ve missed something – what would be an example of a property of a notetype that is not a property of the corresponding note type

A given note's type is a specific notetype. The former refers to a property of a note ("a note's type", "a note's modification time", etc), the latter to the separate entity/object which a bunch of notes reference. I find "a notetype's modification time" easier to parse than "a note type's modification time", as the latter makes me think of "a note's type". But maybe that's just me.

> In any event, if you use notetype, then I think you should use it throughout – e.g. in the menu item “Manage Notetypes”. Likewise, you should then probably also use cardtype.

These changes were difficult in the past, as any change to the English would mean things needed to be retranslated. That is no longer the case with the new translation system, so we could go through and make things consistent now. This might be worth discussing with the community first though.

dae:

PR #1466, opened 2021-10-28
Make field state badges more discernible

kleinerpirat:
https://forums.ankiweb.net/t/poll-icon-for-sticky-pinned-frozen-fields/7910/4

Adds class `highlighted` to badge containers when toggled to a presumably less frequently used state. This approach should avoid as much visual clutter as possible, while also making the different states more discernible.

Highlighting works via opacity changes:
- `default`: 0.4
- `highlighted`: 1
- `hover`: 0.8 (this makes hovering visible both on highlighted and non-highlighted badges)

image image

### DIFF
--- a/ts/editor/RichTextBadge.svelte
+++ b/ts/editor/RichTextBadge.svelte
@@ -15,16 +15,19 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     $: icon = off ? richTextOff : richTextOn;
 </script>
 
-<span on:click|stopPropagation={toggle}>
+<span class:highlighted={off} on:click|stopPropagation={toggle}>
     <Badge iconSize={80}>{@html icon}</Badge>
 </span>
 
 <style lang="scss">
     span {
-        opacity: 0.6;
+        opacity: 0.4;
 
-        &:hover {
+        &.highlighted {
             opacity: 1;
+        }
+        &:hover {
+            opacity: 0.8;
         }
     }
 </style>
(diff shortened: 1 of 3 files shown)


ts/editor/RichTextBadge.svelte
```diff
@@ -15,16 +15,19 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     $: icon = off ? richTextOff : richTextOn;
 </script>
 
-<span on:click|stopPropagation={toggle}>
+<span class:highlighted={off} on:click|stopPropagation={toggle}>
```
hgiesel, line 18: Why is this highlighted when off? Shouldn't a greyed out state show that the rich text input is not visible?

kleinerpirat: My thinking was that rich text input will be visible far more frequently than hidden. Hiding fields is an exception, not the rule. Therefore, I'd recommend graying it out when it's on to reduce visual clutter.

You can try it out the other way. I've found it to be unneccessarily distracting.

hgiesel: Okay I see - makes sense.
This also leads me back to [this suggestion of mine](https://forums.ankiweb.net/t/default-editor-settings-instead-of-persistent/12761). In that case, the icon would be highlighted if it does not match the default.

hgiesel:
👍

ts/editor/RichTextBadge.svelte
kleinerpirat: Yeah, I think that's for the best. Highlights only serve their purpose when they're in the minority.
I renamed the commit to reflect that idea better.

hgiesel:
I guess it might not have been obvious, but the PR is good to merge from my side ;)

dae:
Thanks guys! 🚀

PR #1468, opened 2021-10-29
Remove data/ prefix from data folder in mediasrv.py

hgiesel:
b5e9eba26f71a65c6e165 broke the media server for me.

When using `./scripts/ts-run`, [this](https://github.com/ankitects/anki/blob/16ad0137f7ee9c2796d356c643861d744b153324/scripts/ts-run#L9) already adds `/data` as a suffix, which is why all the media requests failed. I'm now entirely sure which location should be corrected, but I was guessing `mediasrv.py`.


dae:
I've changed ts-run instead, as otherwise the ./run and packaged build cases will fail

PR #1470, opened 2021-10-30
Several CSS fixes - Editor Cleanup

hgiesel:
- Addresses some editor CSS oddities after moving to Qt6 and new Svelte editor:
  - SVGs changing color on hover
  - Tags and their delete icons are bit smaller now
- Editor does not include `webview.scss` anymore. I've included the button css from buttons.scss into button_mixins.scss, which are included in the button components (`IconButton` and `LabelButton`) themselves
- I generally moved some css from `-base` files into the sass directory
- I inlined `sass/core.scss` into `webview.scss`. The idea here is to leave it behind, once all webviews are in `/ts`.


dae:
Thanks Henrik, looks good. I had to add back core.scss for now; have added a comment so its reason for existence is clear.

dae:
This introduced a few minor regressions:

- when viewing deck options in two column mode, the left bottom is lacking a margin:
Screen Shot 2021-10-31 at 8 38 15 am


- the dropdown indicator at the top of the deck options has disappeared - was that deliberate?

dae:
Card info screen now is indented at the top:
Screen Shot 2021-10-31 at 8 39 29 am

hgiesel:
Concerning the select button: I was actually thinking about rebuilding it without HTML `<select>` and `<option>`, but more like [this](https://tailwindui.com/components/application-ui/forms/select-menus), where it uses just buttons and lists under the hood.
The current version gives an ugly white flash when opening it, which could be avoided with that approach.

The other things I'll address shortly

Issue #1416, opened 2021-10-08
V2 Scheduler Fuzz not working for cards with short intervals (<= 4 days)?


kuroahna:
OS: Windows 10
Anki Version: 2.1.48 (fb07bad3)
Python: 3.8.6
Qt: 5.14.2
PyQt: 5.14.2

I'm using the [load-balanced-scheduler addon](https://github.com/xquercus/load-balanced-scheduler) which basically patches the `_fuzzedIvl` function in the v2 scheduler by instead of using a random interval from the fuzz range, we choose the interval from the fuzz range that has the least amount of cards.

However, we noticed there's an issue with the algorithm where the best interval chosen by the load balancer addon isn't being respected and seemed to be "off by 1" sometimes. I've done some analysis in the [issue thread](https://github.com/xquercus/load-balanced-scheduler/issues/11#issuecomment-937912581) but the issue seems to arise from the `_constrainedIvl` function:

```python
    def _constrainedIvl(
        self, ivl: float, conf: QueueConfig, prev: int, fuzz: bool
    ) -> int:
        ivl = int(ivl * conf.get("ivlFct", 1))
        if fuzz:
            ivl = self._fuzzedIvl(ivl)
        ivl = max(ivl, prev + 1, 1)
        ivl = min(ivl, conf["maxIvl"])
        return int(ivl)
```

Particularly, this line `ivl = max(ivl, prev + 1, 1)`

For my Anki settings, I have interval modifier set to 100%, hard interval set to 120%, easy bonus set to 130%, starting ease set to 250% and new interval set to 20%, so the calculations below will reflect accordingly.

Looking at the scheduler code in `2.1.48`: https://github.com/ankitects/anki/blob/2.1.48/pylib/anki/scheduler/v2.py

From what I understand (correct me if I'm wrong, as I don't fully understand anki's scheduler code as I've only tried reading and playing around with it for a few hours), but the Good button's interval is dependent on the Hard button's calculated interval. And likewise, the Easy button's interval is dependent on the Good button's calculated interval. This can be seen in the `_nextRevIvl` function:

```python
    def _nextRevIvl(self, card: Card, ease: int, fuzz: bool) -> int:
        "Next review interval for CARD, given EASE."
        delay = self._daysLate(card)
        conf = self._revConf(card)
        fct = card.factor / 1000
        hardFactor = conf.get("hardFactor", 1.2)
        if hardFactor > 1:
            hardMin = card.ivl
        else:
            hardMin = 0
[...]

PR #1474, opened 2021-11-03
Rework v3 fuzzing

RumovZ:
Details in https://github.com/ankitects/anki/issues/1416#issuecomment-958208149.

How should I proceed with the Python tests? Do we stop testing v3 there and add Rust unit tests instead? I'll add some more comments in the code.


rslib/src/scheduler/states/mod.rs
```diff
@@ -95,45 +95,62 @@ pub(crate) struct StateContext<'a> {
...
+            let (lower, upper) = constrained_fuzz_bounds(interval, minimum, maximum);
+            (lower as f32 + fuzz_factor * ((1 + upper - lower) as f32)).floor() as u32
         } else {
-            interval
+            (interval.round() as u32).max(minimum).min(maximum)
         }
-        .round() as u32
     }
 
     pub(crate) fn fuzzed_graduating_interval_good(&self) -> u32 {
```
RumovZ, line 118: Probably doesn't matter in practice, but these respect the configured limits now.

rslib/src/scheduler/states/mod.rs
```diff
@@ -95,45 +95,62 @@ pub(crate) struct StateContext<'a> {
...
+        fuzz_range(interval, 0.15, 2.0)
+    } else {
+        fuzz_range(interval, 0.05, 4.0)
     }
 }
 
-fn fuzz_range(interval: f32, factor: f32, minimum: f32) -> (f32, f32) {
+fn fuzz_range(interval: f32, factor: f32, minimum: f32) -> (u32, u32) {
     let delta = (interval * factor).max(minimum).max(1.0);
-    (interval - delta, interval + delta + 1.0)
```
RumovZ, line 136: I couldn't figure out the purpose of the `+ 1.0`. Was it fine to drop it?

rslib/src/scheduler/answering/mod.rs
```diff
@@ -59,7 +61,7 @@ impl CardStateUpdater {
     /// state handling code from the rest of the Anki codebase.
     pub(crate) fn state_context(&self) -> StateContext<'_> {
         StateContext {
-            fuzz_seed: self.fuzz_seed,
+            fuzz_factor: get_fuzz_factor(self.fuzz_seed),
```
RumovZ, line 64: I (ab?)used this to work around the different algorithm in `with_learning_fuzz()`.

rslib/src/scheduler/states/mod.rs
dae: I don't remember why I added that, and removing it from HEAD doesn't appear to break any of the tests, so it seems like it's safe to cut.

rslib/src/scheduler/states/mod.rs
```diff
@@ -69,7 +68,8 @@ impl CardState {
 
 /// Info required during state transitions.
 pub(crate) struct StateContext<'a> {
-    pub fuzz_seed: Option<u64>,
+    /// In `[0, 1)`. Used to pick the final interval from the fuzz range.
```
dae, line 71:
```suggestion
    /// Between 0.0-1.0. Used to pick the final interval from the fuzz range.
```

dae (commented):
Looking good! I've pushed the beginnings of a unit test - if you change scheduled/elapesd days from 2 to 1, it will trigger a panic which needs looking into.

dae:
The Python tests are run with fuzzing disabled, so theoretically this change shouldn't break the existing tests. Any ideas why reviewing_early_normal() is a day off now?

We can test the fuzzing code in Rust, and some tests there would be great.

RumovZ:
> Any ideas why reviewing_early_normal() is a day off now?

I've removed flooring in 93708c6, because I thought my subsequent commits would break these tests anyway. Should I try to revert it?

rslib/src/scheduler/states/mod.rs
RumovZ: I wanted to emphasise that the upper bound is not included, but I can use more intelligible phrasing.

RumovZ:
> I've pushed the beginnings of a unit test

Sorry, I couldn't find where.

dae:
Sorry, I have a git commit hook that runs checks before pushing, and it failed because of the failing Python test. Should be available now: https://github.com/ankitects/anki/pull/1474/commits/4a745877846d29ec9e722b155d6d07424f67f17d

If it's only that one Python test we're breaking, then removing the flooring sounds good. It would be good if we could get some coverage of the same area either in Rust, or by cloning the Python function and making the minor changes required (that may be the more expedient option)

RumovZ:
> if you change scheduled/elapesd days from 2 to 1, it will trigger a panic which needs looking into.

A fuzz factor of 1 is undefined behaviour as we sample with `gen_range(0.0..1.0)`. Not sure if we can do anything to prevent future mistakes apart from the doc string.

rslib/src/scheduler/states/review.rs
```diff
@@ -253,4 +253,20 @@ mod test {
         assert!(leech_threshold_met(2, 1));
         assert!(leech_threshold_met(3, 1));
     }
+
+    #[test]
+    fn fuzz() {
```
RumovZ, line 258: I've added tests in `scheduler/states/mod.rs`. Do we still need tests here? I guess we could assert that intervals grow both for increasing fuzz factor and button numbers.

pylib/tests/test_schedv2.py
```diff
@@ -715,6 +717,7 @@ def test_suspend():
 
 
 def test_filt_reviewing_early_normal():
+    to_int = round if is_2021() else int
```
dae, line 720: Nice simple solution :-)

rslib/src/scheduler/states/review.rs
dae: Your tests look good, and this was just intended as a starting point; feel free to cut it.

rslib/src/scheduler/states/mod.rs
dae: "In range 0.0..1.0"? Or we could use the more explicit Swift syntax:  Between 0.0..<1.0

dae:
> A fuzz factor of 1 is undefined behaviour as we sample with gen_range(0.0..1.0). Not sure if we can do anything to prevent future mistakes apart from the doc string. 

Ah, so it was user error on my part :-)

dae (commented):
Looks good Rumo, apart from tweaking that comment and stripping the test in review.rs, I can't see anything that might need changing

RumovZ:
Great, ready from my side as well. 🙂

dae:
Thanks for this Rumo :-)

PR #1475, opened 2021-11-03
Consider setting "no-non-null-assertion: off" by default

hgiesel:
Whenever I use a non-null assertion, I will typically turn off eslint for it. While it is not _clean_ when it comes to typing, I'd consider it on the same level as upcasting using `as Obj`, which we also don't restrict.

Comparing non-null-assertion with using `any`: `any` is a magic trick, that allows us to do anything, which is why we should restrict it, whereas non-null-assertion still has very clear semantics.

Altogether I'd vote for turning off the error altogether.


dae:
Yep, sounds good

PR #1476, opened 2021-11-03
Editor Field Descriptions

kleinerpirat:
Inspired by https://forums.ankiweb.net/t/field-descriptions/10686

Set optional field descriptions from the fields dialog and display them as tooltips next to the field's name.

Screenshot from 2021-11-03 22-52-45

 

### Changes
- Add `QLineEdit` with placeholder text to `FieldDialog`
  - description is stored in `fld["description"]`
- Add `setDescriptions` to note initialization scripts
- Add `LabelDescription.svelte` component into `LabelName.svelte`
  - only appears if a description is set

---
As this is an addition to the data stored in the notetype, I'm aware this might get rejected. Here are some considerations:
- The initial forum topic indicates there's some community interest in such a feature.
- It could make it easier for people to learn how to use complex note types.
- This could of course be done with an add-on instead, but less experienced users - the ones who would benefit the most from field descriptions - probably wouldn't download such a niche add-on.

@dae feel free to edit this PR - it's my largest yet and there are probably some things I could improve, especially in Python.


ftl/core/fields.ftl
```diff
@@ -3,6 +3,8 @@ fields-delete-field-from = Delete field from { $val }?
 fields-editing-font = Editing Font
 fields-field = Field:
 fields-field-name = Field name:
+fields-description = Description
+fields-description-placeholder = Add a tooltip next to the field's name
```
dae, line 7: Maybe something like the following?

```suggestion
fields-description-placeholder = Tooltip to show next to the field's name in the editing screen
```

qt/aqt/editor.py
```diff
@@ -460,6 +460,12 @@ def loadNote(self, focusTo: int | None = None) -> None:
             (fld, self.mw.col.media.escape_media_filenames(val))
             for fld, val in self.note.items()
         ]
+
+        flds = self.note.note_type()["flds"]
+        descriptions = [
+            fld["description"] if "description" in fld else "" for fld in flds
```
dae, line 466:
```suggestion
            fld.get("description", "") for fld in flds
```

qt/aqt/fields.py
```diff
@@ -234,6 +235,10 @@ def saveField(self) -> None:
         if fld["rtl"] != rtl:
             fld["rtl"] = rtl
             self.change_tracker.mark_basic()
+        desc = f.fieldDesc.text()
+        if not "description" in fld or fld["description"] != desc:
```
dae, line 239: These references would benefit from a .get() as well

qt/aqt/forms/fields.ui
```diff
@@ -104,31 +108,47 @@
...
+      <widget class="QLabel" name="label_desc">
+       <property name="sizePolicy">
+        <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+         <horstretch>0</horstretch>
+         <verstretch>0</verstretch>
+        </sizepolicy>
+       </property>
        <property name="text">
-        <string>fields_reverse_text_direction_rtl</string>
+        <string>fields_description</string>
```
dae, line 134: In Designer, please go to Edit>Edit Tab Order, and ensure the newly added field comes in the correct order. You can right click + use "start from here"

dae (commented):
The Python/TypeScript changes look fine to me (presumably @hgiesel will add something if he notices any issues), but to do this properly would involve updating the Protobuf and Rust as well:

- the new attribute should be added to Fields in notetype.proto
- the attribute would need to be added to NoteFieldSchema11 with a comment that it's a recent addition (like the new items in DeckConfSchema11)
- the schema 11 upgrade/downgrade code will need to handle migrating from the json to protobuf and vice versa
- this last one is a bit involved: in order to avoid breakages when syncing between older and newer clients that may upgrade and downgrade, we need an approach like clear_other_duplicates() ... which is currently only implemented for deck configs, so will need to be added for notetypes/fields. If you're not comfortable with this part @RumovZ may be able to help out

qt/aqt/editor.py
kleinerpirat: Beautiful, thanks! :)

qt/aqt/fields.py
kleinerpirat: Indeed. Thanks for that suggestion!

kleinerpirat:
The first two points you mentioned are clear to me and I'll make these changes in the evening once I got time. Perhaps I'll also be able to make the other two, but help would be greatly appreciated! :)

RumovZ:
I'm not yet familiar with this particular part of the codebase, but I am happy to help if you are not comfortable with the Rust part. 🙂

kleinerpirat:
Thanks @RumovZ, that's very kind of you! I think I'm at my witt's end. Anything else from my side would be just guessing.

What's the best way forward? Merge this and let you handle the Rust part in a separate PR? Or could you do it via suggestions like Damien did, so I can merge them into my branch?

RumovZ:
I'd say suggestions are only practical for trivial changes, because among other things, you have to make them in markdown.
I assume @dae wouldn't want to merge this into the main branch before it's finished, but maybe he'll create a separate branch to which I can then push.

kleinerpirat:
The Svelte work isn't done yet, sadly. I just noticed that the descriptions don't update when switching between notetypes or when setting a new description through the fields dialog.

@hgiesel I'm currently investigating your code in `OldEditorAdapter.svelte` to find a suitable solution. Am I right to assume that updating `fieldsData` with `setDescriptions` similarly to [setFields](https://github.com/kleinerpirat/anki/blob/e50de51e561f5df94435702afc664292d698ef9b/ts/editor/OldEditorAdapter.svelte#L72) will solve this - or is there more to it?

hgiesel:
It's the tooltip in `WithTooltip` that doesn't update. So far we never had tooltips which could change their content. I don't think Bootstrap tooltips support this, which is unfortunate.

kleinerpirat:
Thanks, Henrik! Wanted to write that I found that issue a few minutes ago :)

> I don't think Bootstrap tooltips support this, which is unfortunate.

So we need to destroy the tooltip and create a new one each time the description changes?

hgiesel:
That would be one option, yes. However that would be quite hard to communicate to the where the tooltip is mounted, to the badge in this case.

Another option would be to to create a tooltip component and place it via `WithDropdown`.

Generally I've been thinking for some time, that using the Bootstrap components might make things more complicated than necessary for us. IMO it might be worth considering using just the positioning engine, PopperJS, and use that directly rather than indirectly through Bootstrap.

dae:
I'm not sure if I can set permissions on particular branches. Probably the easiest option is to wait until @kleinerpirat has finished any parts he wants to finish, and then create a new local branch, pushing it into a new PR when you're ready:

```
% git remote add kleinerpirat https://github.com/kleinerpirat/anki.git
% git fetch kleinerpirat
% git co kleinerpirat/field-descriptions
Note: switching to 'kleinerpirat/field-descriptions'.
% git co -b field-descriptions-pt-2
Switched to a new branch 'field-descriptions-pt-2'
```

qt/aqt/fields.py
```diff
@@ -214,6 +214,7 @@ def loadField(self, idx: int) -> None:
         f.fontSize.setValue(fld["size"])
         f.sortField.setChecked(self.model["sortf"] == fld["ord"])
         f.rtl.setChecked(fld["rtl"])
+        f.fieldDesc.setText(fld["description"] if "description" in fld else "")
```
dae, line 217: This could use a .get() too

RumovZ:
> I'm not sure if I can set permissions on particular branches. Probably the easiest option is to wait until @kleinerpirat has finished any parts he wants to finish, and then create a new local branch, pushing it into a new PR when you're ready:

Sorry, I phrased that poorly. I didn't mean to say "push", but rather "open a PR".

But we can do it like you say as well, or I open a PR on @kleinerpirat's fork. I don't have a preference.

RumovZ:
@dae, I think this is all that's required?
If that's correct, you can just apply this diff, @kleinerpirat. Checks were failing because `#[serde(default, deserialize_with = "default_on_invalid")]` was missing.

```diff
diff --git a/rslib/src/notetype/schema11.rs b/rslib/src/notetype/schema11.rs
index 14a61ffac..347c69997 100644
--- a/rslib/src/notetype/schema11.rs
+++ b/rslib/src/notetype/schema11.rs
@@ -137,6 +137,8 @@ fn bytes_to_other(bytes: &[u8]) -> HashMap<String, Value> {
 impl From<Notetype> for NotetypeSchema11 {
     fn from(p: Notetype) -> Self {
         let c = p.config;
+        let mut other = bytes_to_other(&c.other);
+        clear_other_duplicates(&mut other);
         NotetypeSchema11 {
             id: p.id,
             name: p.name,
@@ -160,11 +162,18 @@ impl From<Notetype> for NotetypeSchema11 {
             latex_post: c.latex_post,
             latexsvg: c.latex_svg,
             req: CardRequirementsSchema11(c.reqs.into_iter().map(Into::into).collect()),
-            other: bytes_to_other(&c.other),
+            other,
         }
     }
 }
 
+fn clear_other_duplicates(other: &mut HashMap<String, Value>) {
+    // see `clear_other_duplicates()` in `deckconfig/schema11.rs`
+    for key in &["description"] {
+        other.remove(*key);
+    }
+}
+
 impl From<CardRequirementSchema11> for CardRequirement {
     fn from(r: CardRequirementSchema11) -> Self {
         CardRequirement {
@@ -204,7 +213,8 @@ pub struct NoteFieldSchema11 {
     pub(crate) font: String,
     pub(crate) size: u16,
 
-    // Nov. 2021 addition: editor field description
+    /// Nov. 2021 addition: editor field description
+    #[serde(default, deserialize_with = "default_on_invalid")]
     pub(crate) description: String,
 
     #[serde(flatten)]
```

kleinerpirat (commented):
@hgiesel What do you think of this? It doesn't negatively affect tooltips of other screens and fixes the issue.

ts/components/WithTooltip.svelte
```diff
@@ -26,19 +26,27 @@
...
     }
 
     onDestroy(() => tooltipObject?.dispose());
+
+    let previousTooltip: string = tooltip;
+    $: if (tooltip !== previousTooltip) {
+        let element: HTMLElement = tooltipObject["_element"];
+        tooltipObject.dispose();
+        createTooltip(element);
+    }
```
kleinerpirat, line 49: Botched that earlier review. Let's do that properly :)

hgiesel:
It's a clever hack, I like it. It also shows once more that it might be good idea to move away from the Bootstrap components at some point - `WithDropdown` is also full of hacks that rely on Bootstrap internals.

Here are two more suggestions from my side: https://github.com/hgiesel/anki/compare/field-descriptions...hgiesel:field-descriptionsx.

ts/components/WithTooltip.svelte
```diff
@@ -26,19 +26,27 @@
...
             delay: { show: showDelay, hide: hideDelay },
             trigger,
-        });
+        };
+        tooltipObject = new Tooltip(element, tooltipOptions);
     }
 
     onDestroy(() => tooltipObject?.dispose());
+
+    let previousTooltip: string = tooltip;
```
hgiesel, line 44: `previousTooltip` will always point to the _original_ tooltip. Is that on purpose? If yes, then the wording is a bit confusing (also it could be `const`).

kleinerpirat: Thanks for the heads-up, that was not intended. Like the name suggests, I wanted it to be the previous tooltip. If it always compares with the original tooltip, switching back to the original note type will prevent an update.

Therefore, I won't use a `const`, but rather update `previousTooltip` within the reactive function to make it do what the name suggests.

kleinerpirat:
From my side, this is all the changes I want to make. Thanks for the help everybody, that was a great joint effort! :hugs:

dae:
Thanks everyone!

rslib/src/notetype/schema11.rs
```diff
@@ -162,12 +160,12 @@ impl From<Notetype> for NotetypeSchema11 {
...
             latexsvg: c.latex_svg,
             req: CardRequirementsSchema11(c.reqs.into_iter().map(Into::into).collect()),
-            other,
+            other: bytes_to_other(&c.other),
         }
     }
 }
 
-fn clear_other_duplicates(other: &mut HashMap<String, Value>) {
+fn clear_other_field_duplicates(other: &mut HashMap<String, Value>) {
```
RumovZ, line 168: Sorry for the mix-up!

dae: No problem, it's a bit confusing how there are multiple different `other`s.  :-)

PR #1478, opened 2021-11-04
Several editor fixes

hgiesel:
1. Fix the add-on buttons
2. Fixes the caret position when dealing with PlainTextInput:
  - Clicking on PlaintextInput will not automatically move the caret to the end, but rather to the location you click
  - When tabbing out of Anki and back, the caret will stay. This does not work in the RichTextInput yet: the problem is that when any editing input looses focus, we will overwrite its content via store subscriptions (as I've explained in the Editor PR). After the overwrite, the original location is lost. This means we must store and restore the location when blurring/focusing the input. CodeMirror had that already built in, for the contenteditable I've in fact already written the logic, but I have it [in the wrong PR](#1377) ^^;
3. Simplifies the logic of accessing the component tree. One interesting thing I learnt is that the main code (within the script tags) of each component is guaranteed to execute synchronously upon component instantiation. This means that we can connect all components can connect synchronously, it's only the dom elements which we need to connect asynchronously: that's why I've wrapped the element in `EditingArea` into a promise.


qt/aqt/browser/browser.py
```diff
@@ -408,7 +408,7 @@ def setupEditor(self) -> None:
         def add_preview_button(editor: Editor) -> None:
             editor._links["preview"] = lambda _editor: self.onTogglePreview()
             editor.web.eval(
-                "noteEditorPromise.then(noteEditor => noteEditor.toolbar.then(toolbar => toolbar.notetypeButtons.appendButton({ component: editorToolbar.PreviewButton, id: 'preview' })));",
+                "noteEditorPromise.then(noteEditor => noteEditor.toolbar.notetypeButtons.appendButton({ component: editorToolbar.PreviewButton, id: 'preview' }));",
```
dae, line 411: Nice

ts/editor/destroyable.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright: Ankitects Pty Ltd and contributors
+// License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+
+export interface Destroyable {
+    destroy(): void;
+}
+
+export function clearableArray<T>(): (T & Destroyable)[] {
```
dae, line 8: A docstring on the function would be nice. 

Is there any risk of the cached destroy method pointing to the wrong place if something is manually inserted into or removed from the list?

dae (commented):
Thanks Henrik!

ts/editor/destroyable.ts
hgiesel: The index is only created inside the destroy function, so I'd say no.
Something that could be awkward is if somebody tried to check for the existence of a field:

```
if (field[index]) { /* do something */ }
// this will always be correct, instead they should do this:
if (index in field) { /* do something */ }
```

dae: Hmm, I'm not fully understanding the purpose of this function. Is my understanding correct that it tries to catch numeric references (eg list[5]), and if they are false-ish, it returns an empty object with a destroy method that will remove the entry from the list and shift other entries up? If so, that seems a bit complicated and magical - is it not at all practical to do this a different way that is more explicit?

dae:
This appears to have broken cloze deletion:

JS error /_anki/js/editor.js:75296 Uncaught TypeError: Cannot convert a Symbol value to a number

ts/editor/destroyable.ts
hgiesel: I will give another look over the coming days, and see if I can refactor in. If I cannot, I'll explain in more detail why not.

PR #1481, opened 2021-11-08
Save and restore location on ContentEditable

hgiesel:
While having focus in the content editable and then tabbing out and back into the window, the location will now be correctly reset.
Most of the code is cherry-picked from #1377.

If you make a selection in one field, select another, and then reselect the original field by clicking the label, you an actually see your original selection. Not sure if this is a bug or a feature.


hgiesel:
The selection behaves a bit weird if you're having a non-collapsed selection in one field, click another field, and then click back. It will always act as if it was a Control-click (modifying the selection). I'm looking into it.

dae:
Looks good! Thanks as always Henrik.

PR #1482, opened 2021-11-09
Garbage collect unused Fluent strings

### LINKED ISSUE #1472: Tracking used Fluent strings
The codegen we used for .ftl files (eg tr.browsing_all_fields()) is great for catching references to missing strings, but does not help us in knowing which strings are no longer used by the codebase.  Currently we tend to remove strings from the .ftl files at the same time they are removed from the codebase, but this means that minor bugfix releases can not update to the latest translations.

Another thing to consider is that AnkiMobile, and likely AnkiDroid in the future, also reference the translations, and they need to be considered too.

To address this, I think we need a few scripts to implement a garbage collection approach:

- a script that extracts tr references from our Python, Rust and Typescript files into a set, sorts them, and writes them out to a json file. The script should accept an argument to the source folder root, so it can be extended to catch references in Swift, Java and Kotlin source files in the future
- another script that takes two arguments: the ftl/core folder, and another folder that contains one or more .json files that store the extracted references. It might contain files like:

anki-stable.json
anki-HEAD.json
ankimobile-stable.json
ankidroid-stable.json

Where 'stable' corresponds to the last major update (eg 2.1.45), not the absolute latest point release. We'd probably want to check the -stable files into git; the -HEAD one we could generate on the fly.

The script would read each of those files, and build a set()/HashSet of all used strings. It would then iterate over the .ftl files, build them into an AST (eg transform-string.py or rslib/i18n/extract.rs), remove the unused keys, and then write the updated .ftl files back out.

We could then periodically update a -stable file when a milestone release is met, and old strings could be cleaned up over time.

RumovZ:
Closes #1472.
You can run the two ignored tests in `garbage_collection.rs` to ascertain:

1. an idempotent round-trip with `formatting()` and looking at the diff,
2. accurate garbage collection with `garbage_collection()` and building.

Todos
---

- [x] Check `BUILD.bazel` and run cargo-raze.
- [x] Maybe refactor some stuff (see comments).
- [x] [deferred] ~Maybe add external dependency instead of copying `serializer.rs`. You would want to fork https://github.com/RumovZ/fluent-rs/tree/serializer.~
- [x] ~Maybe print out modified files and deleted messages.~
- [x] Maybe add more docs, like the caveats (see below).

I'd appreciate if you could help with the first one. Please let me know which of the maybes you want to be done.

Caveats
---

- ~Terms are not supported. They will not be tracked and messages referenced inside them will not be considered in use. If Anki should use this Fluent feature in the future, this script would need an update.~
- Messages are considered in use if they are referenced in other messages, even if those messages themselves are not in use and going to be deleted. Theoretically, you would need to run the garbage collection until nothing is deleted anymore to be sure that all unused messages are gone, but in reality, Anki hardly uses nested messages, so this should not be a concern.
- If we should fail to recognise a nested message and garbage collect it, the build will not fail, but an error will be printed to the console at runtime.


rslib/i18n_helpers/src/garbage_collection.rs
```diff
@@ -0,0 +1,222 @@
...
+pub fn remove_unused_ftl_messages(ftl_root: &str, json_root: &str) {
+    let mut used_ftls = HashSet::new();
+    for entry in WalkDir::new(json_root) {
+        let entry = entry.expect("failed to visit dir");
+        if entry.file_type().is_file()
+            && entry
+                .file_name()
+                .to_str()
+                .expect("non-unicode filename")
+                .ends_with(".json")
```
RumovZ, line 45: This is done a lot for various file endings. I'm still thinking how to add a helper, without needing to implement our own  `IntoIter` with an endless type soup.

rslib/i18n_helpers/src/garbage_collection.rs
```diff
@@ -0,0 +1,222 @@
...
+            extract_nested_ftl_messages(&mut &ast, &mut used_ftls);
+            let num_entries = ast.body.len();
+            ast.body = ast
+                .body
+                .into_iter()
+                .filter(|entry| match entry {
+                    ast::Entry::Message(msg) => used_ftls.contains(msg.id.name),
+                    _ => true,
+                })
+                .collect();
```
RumovZ, line 72: Might better be a separate function.

rslib/i18n_helpers/src/garbage_collection.rs
```diff
@@ -0,0 +1,222 @@
...
+        }
+        _ => (),
+    }
+}
+
+fn extract_references_from_file(refs: &mut HashSet<String>, entry: &DirEntry) {
+    lazy_static! {
+        static ref RUST_STYLE_TR: Regex = Regex::new(r"\Wtr\s*\.([0-9a-z_]+)\W").unwrap();
+        static ref TS_STYLE_TR: Regex = Regex::new(r"\Wtr2?\.([0-9A-Za-z_]+)\W").unwrap();
+        static ref DESIGNER_STYLE_TR: Regex = Regex::new(r"<string>([0-9a-z_]+)</string>").unwrap();
```
RumovZ, line 136: Some things that could be improved here. I don't think `\W` is the best choice to find a name delimiter in any of these languages and maybe at least the last two regexes should be verbose.

rslib/i18n_helpers/src/garbage_collection.rs
```diff
@@ -0,0 +1,222 @@
...
+            if entry.file_type().is_file() {
+                extract_references_from_file(&mut refs, &entry);
+            }
+        }
+    }
+    serde_json::to_writer(
+        fs::File::create(target).expect("failed to create file"),
+        &refs,
+    )
+    .expect("failed to write file");
```
RumovZ, line 31: I didn't implement any error handling and the `expect`s will need to be replaced depending on how you plan to use these scripts.

rslib/i18n_helpers/src/garbage_collection.rs
RumovZ: Somewhat alleviated by f72a51a.

dae:
Have to dash off in a second, so I'll need to get back to you with a more detailed reply later, but for now:

- looks great!
- results of running cargo-raze (with update line disabled) pushed
- bazel tests are failing due to clippy lints
- is much more work required to get the serializer changes into a state where they could be submitted as a new PR to the fluent crate?
- presumably the current tests would be exposed as a binary in the future?
- if we use serde::to_string_pretty and sort first, it will minimize conflicts if we check the resulting file into git (for old stable/mobile clients)

> Terms are not supported. They will not be tracked and messages referenced inside them will not be considered in use. If Anki should use this Fluent feature in the future, this script would need an update.

I think we use just one at the moment - -deck-config-delay-hint.

> Messages are considered in use if they are referenced in other messages, even if those messages themselves are not in use and going to be deleted. Theoretically, you would need to run the garbage collection until nothing is deleted anymore to be sure that all unused messages are gone, but in reality, Anki hardly uses nested messages, so this should not be a concern.

Yep, sounds totally reasonable. IIRC, the main reason we have nested messages at the moment is because I was trying to preserve some of the older gettext translations when migrating to Fluent.

rslib/i18n_helpers/src/garbage_collection.rs
dae: For build scripts, tests, and tools like these, unwrap and expect are fine; I'd likely do the same

RumovZ:
> is much more work required to get the serializer changes into a state where they could be submitted as a new PR to the fluent crate?

I don't have much experience in this respect, but I assume it wouldn't be that much. Last, the maintainer only said that it would need to be generalised to work with `Slice<ToString>`.

> presumably the current tests would be exposed as a binary in the future?

I think you mean the two public functions? The tests were only supposed to help me debug.

> I think we use just one at the moment - -deck-config-delay-hint.

Oh, I thought I had checked... Now that you mention it, I can even find another one: -errors-support-site. I'll better add support for terms too then.
I've also realised that there is a flaw in my approach to detect nested messages. I'm doing it in parallel to deleting while traversing the files, but a seemingly unused message might well be referenced in a later ftl file.
In that case, it's probably best to regex the ftl root beforehand, like we do with files passed to `extract_ftl_references()`. Supporting terms is hardly extra work then, either.

Thanks for cargo-razing. I'll take care of the other points you've mentioned!

rslib/i18n_helpers/src/garbage_collection.rs
dae: I'm not too fussed about these - verbose would be a bit more readable, but as they're fairly short, it's not super-important. \W seems to be working - can you come up with an example of a case we're likely to hit in the future where it would not be appropriate?

dae:
Works great! I've added some little scripts and the AnkiMobile keys in, and am happy to merge this in whenever you're ready. Thanks for submitting that PR as well - if they end up deciding to merge it, we can switch over to it when they push out a new version.

dae:
Regarding the todos:

> Maybe print out modified files and deleted messages.

A git diff shows us this, so I don't think we need to bother

>  Maybe add more docs, like the caveats (see below).

Sure, let's drop any caveats in remove-unused.sh

rslib/i18n_helpers/src/garbage_collection.rs
RumovZ: I was thinking more about the opposite, because \W is a unicode class and will match all sorts of stuff that we don't need. But I'm probably just being pedantic. \W is much better to understand and maintain than any explicit set of characters we could come up with, so never mind. 🙂

RumovZ:
Great, then it's good to go!

dae:
Thanks for all your work on this Rumo :-)

dae:
Finally was able to use this after the 2.1.50 release, and it all went smoothly - thanks again!

PR #1485, opened 2021-11-12
Fix incorrect card count in timebox after undo

abdnh:
Minimal steps to reproduce:
1. Add a card
2. Do a single repetition, then return to the deck list (reps=1, startReps=0)
3. Go to the reviewer again and undo the last review before rating another card (reps=0, startReps=1)
4. Wait for the timebox time limit to pass, then rate the current card (reps=1, startReps=1). The timebox dialog will incorrectly report studying 0 cards.


dae:
Thanks Abdo! Black is probably confused by the symlinks on Windows.

PR #1486, opened 2021-11-12
Make add cards dialog a main window

### LINKED ISSUE #980: Investigate switching AddCards dialog to a main window
Currently the AddCards screen is a dialog instead of a main window, which means on macOS the top menubar shows the entries for the last active main window, such as the main screen or the browse screen. This causes browser-defined shortcuts to conflict, breaking the 'switch note type' shortcut key.

Moving AddCards to a main window would resolve the issue, but may introduce other issues.

https://anki.tenderapp.com/discussions/ankidesktop/42324-change-note-type-with-browser-open

RumovZ:
Closes #980.
The only issue I've noticed so far, is the error reported in https://bugreports.qt.io/browse/QTBUG-79140. I'm not able to reproduce it anymore though, maybe due to a restart. While I was getting the error, I didn't notice any strange behaviour (like the reported above) except for the message in the console. But I might just not have tested enough.

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -19,9 +19,7 @@ from aqt.qt import *
 from aqt.sound import av_player
 from aqt.utils import (
     HelpPage,
-    addCloseShortcut,
     askUser,
-    disable_help_button,
     downArrow,
     openHelp,
     restoreGeom,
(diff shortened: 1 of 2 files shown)


dae:
Thanks Rumo! Seems to work fine in a brief test; let's see how it fares during beta testing.

hgiesel:
I think this broke the Close button in the AddCards window?

RumovZ:
Thanks, I've pushed a fix.

PR #1489, opened 2021-11-13
Fix duplicate conflicting add-on names being shown

abdnh:
A conflicting add-on's name could be listed twice when installing another add-on if both add-ons list each other in their conflicts list.
image


dae:
Thanks Abdo!

PR #1490, opened 2021-11-14
Fix `constrained_fuzz_bounds()`

RumovZ:
See https://github.com/ankitects/anki/commit/044d25330642b4b13f0e51be19ba5d08782848d4.

### DIFF
--- a/rslib/src/scheduler/states/review.rs
+++ b/rslib/src/scheduler/states/review.rs
@@ -255,7 +255,7 @@ mod test {
     }
 
     #[test]
-    fn low_multiplier_fuzz() {
+    fn extreme_multiplier_fuzz() {
         let mut ctx = StateContext::defaults_for_testing();
         // our calculations should work correctly with a low ease or non-default multiplier
         let state = ReviewState {
@@ -272,6 +272,11 @@ mod test {
         ctx.interval_multiplier = 0.1;
         assert_eq!(state.passing_review_intervals(&ctx), (2, 3, 4));
         ctx.fuzz_factor = Some(0.99);
-        assert_eq!(state.passing_review_intervals(&ctx), (2, 3, 4));
+        assert_eq!(state.passing_review_intervals(&ctx), (3, 5, 7));
+
+        // maximum must be respected no matter what
+        ctx.interval_multiplier = 10.0;
+        ctx.maximum_review_interval = 5;
+        assert_eq!(state.passing_review_intervals(&ctx), (5, 5, 5));
     }
 }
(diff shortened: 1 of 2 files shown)


rslib/src/scheduler/states/review.rs
```diff
@@ -272,6 +272,11 @@ mod test {
         ctx.interval_multiplier = 0.1;
         assert_eq!(state.passing_review_intervals(&ctx), (2, 3, 4));
         ctx.fuzz_factor = Some(0.99);
-        assert_eq!(state.passing_review_intervals(&ctx), (2, 3, 4));
+        assert_eq!(state.passing_review_intervals(&ctx), (3, 5, 7));
```
RumovZ, line 275: This illustrates the implications of ensuring at least 1 day of fuzzing. If you find this too extreme, we could restrict that logic to the hard interval. The other intervals wouldn't get fuzzed then, but always be 1 above the hard interval, effectively getting fuzzed across 2 days as well.
So in this case, we would have (3, 4, 5).

dae: It does feel a bit aggressive - a card with an ease of 1.3x and a 1 day interval could end up with a 5x increase on Good. 3,4,5 sounds like it could work. One other idea would be to do the opposite - pin hard to 2 and allow the fuzz only in good/easy in this low interval case, so we'd end up with something like 2,4,5 in the 0.99 case. No strong feelings either way at this point - WDYT?

RumovZ: To be honest, I don't like either solution very much. Not sure what'd be best study-wise, but it makes the code harder to read and the scheduler less transparent, because it couples fuzzing with the answer button.
A very simple, but not as comprehensive solution would be to increase the threshold above which we enforce fuzzing from 1 to 2. That would lead to (2, 4, 6) for a high fuzz factor.

dae:
+1 for keeping things simple - sounds like that’s the best of the available
choices

dae:
Will merge this in and push the change in a follow-up commit - feel free to follow up if you see a better way of doing it.

I noticed while working on this that our ranges suffer from a similar problem to the early review topic you raised over on the forums previously - cards with a 30 day interval have a range of +/- 4 days, and that doesn't change again until the interval is about 90 days. Instead of the separate 0/2/4 minimum intervals and multiplier jumps, maybe we should be just gradually ramping down the multiplier as the interval grows?

RumovZ:
> Instead of the separate 0/2/4 minimum intervals and multiplier jumps, maybe we should be just gradually ramping down the multiplier as the interval grows?

How exactly? If we just dropped the minimums, the fuzz delta would drop after every multiplier decrease.

I've toyed a bit with different options: https://gist.github.com/RumovZ/ce6417ba03b20d34c2cff0d9ea3e7e55
Option 1 uses an arbitrary mathematical function to model the desired increase at short intervals. Then it slowly transitions into the current formula for large intervals.
Option 2 uses the current approach, except that multipliers are only applied to the share of an interval that lies in its range. Hard to describe in words, maybe it's easier to gather from the code. Basically, it works like marginal tax rates. Obviously, the amount of fuzz is far too much currently, so the parameters would need to be adjusted to the new formular and the desired behaviour.

What do you think?

dae:
Either of those look like they could work - maybe we should opt for whichever is simplest to implement/reason about?

RumovZ:
That's probably the tax rate model.

PR #1493, opened 2021-11-16
Smooth fuzz

RumovZ:
Continuing from #1490. Here is the updated graph:

fuzz_deltas

As you can see, intervals above 80 days get fuzzed by two additional days as a side effect.


dae:
Much nicer! Thanks Rumo :-)

PR #1495, opened 2021-11-16
Introduce our own Container, Row, and Col components

### LINKED ISSUE #1479: minor regressions in HTML layout
Just a reminder about https://github.com/ankitects/anki/pull/1470#issuecomment-955608657 so we don't forget about it

hgiesel:
This replaces the use of the Bootstrap classes `container`, `row`, and `col` classes and instead we use our own components, which are not based upon Bootstrap anymore (but rather inspired). Their _gutters_ are mostly governed by the CSS props `--gutter-inline` and `--gutter-block`.
It addresses all of the issues mentioned in #1479, but there are enough changes that it might introduce its own regressions.

Closes #1479.

There are also some minor cleanups / fixes:

- Inline badge icons in deck-options, when the description wraps
- Replace StickyFooter and StickyHeader with just `StickyContainer`
- There are some cleanups concerning the ChangeNotetype -> I've also added some gutters via the new components @kleinerpirat


kleinerpirat:
Thanks for the mention :) I was struggling with boostrap flex while working on `ChangeNotetype`.

It's nice to know you want to move away from it. My rework for that screen got a bit ambitious (with a TabWidget Svelte component, similar to the Qt counterpart), so perhaps it's better to split it into two PRs, where the first one doesn't add any new components.

dae:
Mostly looks fine, but I noticed a few issues with the deck options screen.

Left and right columns aren't aligned:
Screen Shot 2021-11-17 at 8 00 10 am

A single section wouldn't previously be split up, but that can now happen:
Screen Shot 2021-11-17 at 8 00 48 am

On a narrow display the left margin doesn't seem to match the right one in the sticky bar: 
Screen Shot 2021-11-17 at 8 06 46 am

hgiesel:
@dae These commits should fix the issues you mentioned. One thing I noticed: when using `column-count`, it does not like any flexbox containers as its parent, or it will simply not take effect.

Off-topic: I noticed the recent iOS update made the Select buttons much more visually pleasing.

hgiesel:
@kleinerpirat 
I remember that we discussed the possibility of using a `TabWidget` when implementing the Deck options, and one particular concern was that you couldn't tell if changes would cause issues one other tab pages, which you elegantly solved here with an icon in the tab name.

Personally speaking, I'm not the biggest fan of the tab widget. On mobile, imo clicking on the tab label would certainly harder than _just scrolling down_, so this means were favoring desktop ergonomics over mobile here.
Maybe something like a [Scrollspy](https://getbootstrap.com/docs/5.0/components/scrollspy) would work better here: a scrollspy is basically like a tab widget, only that instead all tabs are concatenated together, and the label _scrolls_ you to the tab, instead of opening it.

Same with the uncollapsing on window width increase - for one, it means we have to maintain both views. Also, the window makes the assumption that if I increase the window width, it must mean I want to see it in two columns, which is not necessarily true. Also, if I had one item focused with my eyes, it would suddenly jump somewhere completely different - that's annoying.

So with these last two paragraphs I basically just wanted to say: Imo a scrollspy is a better alternative for the tab widget ^^; The other details, like the collapsing warnings, and the arrow symbolising the change of note types I like very much.

dae:
Thanks for the quick fixes. I was about to merge this but then noticed one more problem - the switch rows are wrapping the left content too early: 
Screen Shot 2021-11-17 at 10 51 54 am

@kleinerpirat It looks great, but I'm inclined to agree with Henrik here - scrolling is easier, and having to deal with joined and separated tabs sounds like it will make things more complicated.

kleinerpirat:
Thanks for the detailed and honest feedback! I agree - and the scrollspy sounds a lot easier to implement than a tab widget. I'll try to get a PR up over the next couple of days 👉

dae:
Thanks as always Henrik :-)

dae:
Just noticed this has removed the left margin from the congrats screen

dae:
Another minor regression: text is now overlapping the dropdown indicators.
Screen Shot 2021-11-22 at 11 02 44 am

Issue #1471, opened 2021-11-01
Investigate dynamic theme adjustment


dae:
Partly related to https://github.com/ankitects/anki/issues/1116

Currently Anki can not cope with night mode being enabled or disabled while the app is open. If you start in day mode and run the following in the debug console, you can reproduce it:

```
mw.pm.set_night_mode(True)
mw.setupStyle()
```

Some parts of the UI will update automatically; others remain as before, or only partly update. Webviews don't update, and the browser sidebar doesn't update properly. The top toolbar never updates.

It's worse when moving from night→day mode, because night mode hard-codes specific colors, whereas in day mode we rely on the system defaults.  If we hard-coded colors in day mode we could solve this, but then Anki would no longer follow the system theme. Perhaps as a workaround we could try to pull the default colors out of a QPalette prior to switching to dark mode.

- we'll need a hook that is fired when the theme changes, that different parts of the UI can listen to, and refresh themselves with
- webviews could be updated by injecting some JS with .eval(). The more recent JS pages probably just require adding/removing the night-mode class to document.documentElement; older pages might work with .nightMode?. Ideally  we have all webviews add/remove that class automatically, instead of each page having to set up the hook themselves

hgiesel:
For the Svelte pages to update, we need to make nightMode a Svelte store, and adjust all components which use the night mode variable to correctly use it as a store.

When it comes to how to notify the pages to update, this might be relevant: https://github.com/ankitects/anki/pull/1371

dae:
Currently quite a few of our Svelte files are using a context with nightModeKey. Would it make sense to use a night-mode class in a parent element instead? Eg in DropdownItem, which if instead of separate btn-day/btn-night classes, we just had a btn class, and defined the night mode stuff with

:global(night-mode) .btn { ... }

That would presumably make things a bit simpler in the common case? I'm getting a sense of deja vu here though, so apologies if we've discussed this previously :-)

#1371 could be a solution for the notification, but I had something simpler in mind for now - each webview could listen to the hook, and when it's called, it could just do something like "web.eval('try { require("anki/theme").updateTheme({mode: 'dark'}) } catch { }". That would presumably allow any page that was set up for theme updates to respond, without breaking things for pages that don't provide the function. WDYT?

hgiesel:
I think that could work. I was thinking for a long time: Will there ever be theme-related changes that are not entirely CSS, but I'll now stand with no.

[...]

PR #1497, opened 2021-11-17
Live theme changes

RumovZ:
This takes care of theme changes inside Qt so far.
I'm uncertain how to proceed with the webviews. What exactly did you mean by "older pages might work with .nightMode?.", @dae? The two options I can think of are either reloading the entire standard HTML or switching to night mode sensitive CSS and only toggling the class of the doc element.
As for the new pages, just a little reminder about the dsiscussion in #1471, @hgiesel. 🙂


RumovZ:
Oh, and one other thing: The background of the main window's menu bar's menus remain white when switching to night mode and I don't know why. If anyone had an idea why that might be, I'd appreciate it.

dae:
Thanks Rumo! I've pushed some fixes and further updates, and it seems to be in a workable state now. We'll want to update the rest of the nightModeKey contexts and check to see if anything has been missed, but let's first wait to see how the discussion on #1471 pans out

The menu issue was because of the QMenuBar styling we were only applying in day mode. Similar issues may appear in other cases where we only set styling for night mode, such as the scrollbar styling - we may need to apply different colours in day mode to work around it.

dae:
Hmm, how about rootTheme? 

```
/// The current theme that applies to this document/shadow root. When
/// previewing cards in the card layout screen, this may not match the
/// theme Anki is using in its UI.
export const rootTheme = readable(getThemeFromRoot(), (set) => {
```

I started to rename things to pageTheme, but realised a nice property of documentTheme/rootTheme is that they sort of indicate where they are being set. If you're not so fond of rootTheme, happy for me to go with documentTheme instead?

Naming things is hard 😅

hgiesel:
My agenda with `pageTheme` is that if we used an iframe at some point, the CSS would still be isolated from variables set at `document.documentElement`, even though the iframe also have a `Document` at their root. So they would need their own `rootTheme` or `frameTheme`, etc. At the same time I read this on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document):

> The Document interface represents any web **page** loaded in the browser [...]

If we think of Anki as a _web app to be_, we could also name it `appTheme` (so many options ^^;)

Once we are going to use themes for any iframes or shadow roots, they will need their own individual stores. For example for shadow roots, they don't have a `documentElement`, but rather a `shadowRoot.host`, where we'd set the class or attribute. I think for those, we'd just export a constructor from `lib/theme`, and put the theme store on the component tree (think `editor.editorFields[2].editingArea.rootTheme`)

dae:
Ok, I'll go with pageTheme for now so we can move forward with this - we can always change it again before the stable release if necessary. Assuming @RumovZ is interested, are you happy for him to proceed with updating the remaining nightModeKey references to use pageTheme instead, like in https://github.com/ankitects/anki/pull/1497/commits/c2163537d0626a65fad141bbc30f5ee5d5e4194c? Once that's done, I presume this will be ok to merge - if there's something else we've overlooked, please let us know.

dae:
@hgiesel gentle ping in case you missed this

hgiesel:
I didn't catch the question for me, sorry ^^; Sure, he can go ahead. It think `.isDark` can fully fill the role of the old `nightMode`.

RumovZ:
I'll take care of it. Thanks, the two of you!

RumovZ:
I had to format manually. Am I doing anything wrong? That's how I understand the docs, but apparently it doesn't work:
grafik

dae:
Does the latest main commit fix it?

Thanks for your work on this Rumo, I imagine there'll be quite a few users happy to see this land!

RumovZ:
> Does the latest main commit fix it?

Yes, thank you. 🙂

Glad I could help!

Issue #1174, opened 2021-05-19
improve font size handling in deck options screen


dae:
The inputs and label text appear to be inconsistent on some Windows installs:

https://aws1.discourse-cdn.com/standard11/uploads/anki2/original/2X/b/b70522d57fa183d1cab6f31429b7310feb0074a1.png

We'll presumably want to override bootstrap's defaults to get something closer to the rest of Anki's interface, rather than increasing the size of the labels as well.

~~The reason bootstrap defaults to 16px for inputs is because that is the minimum size required to prevent Mobile Safari from zooming in when you tap on an element~~ (this was not exactly correct, the default was 1rem, and the default html font size is 16px) 
https://github.com/ankitects/anki/pull/1172 has caused a regression here - the selector at the top now zooms in, because you're deriving the font size from the toolbar size.

So we'll need a way to reduce the size on computers, but ensure it's at least 16px when running on a mobile device.

You can test the appearance in Mobile Safari by starting Anki with:

ANKI_API_HOST=0.0.0.0 ./scripts/ts-run

Then surfing to http://your_computer_ip:40000/_anki/pages/deckoptions.html#test on your phone. You might need to hit refresh once, and Safari will cache the page after the first load, so you need to turn on the Safari developer tools menu on your computer's Safari, and use the web inspector to refresh the mobile page each time you want it to update (shift+clicking on the refresh button; you may need to hook the phone up via USB).

@hgiesel I presume you're planning to update more of the deck options screen to use shared components - would you mind looking into this at the same time?

dae:
Removing the explicit 16px and some other small tweaks seems to have solved this

dae:
3f0fd548d8803bee9226f95e82358c7e945cd62f made the deck options look better, but now the fields/cards buttons are too big compared to the other buttons in the add screen

dae:
I believe this was resolved some time back.

PR #1499, opened 2021-11-19
Change Notetype UI Rework

kleinerpirat:
After a week of trial and error, this is what I came up with thanks to your feedback. It's a WIP, but now that it's stable enough, I thought you might want to try it out.

The goal is to make the screen feel more intuitive and polished.

Changes
---
- Old notetype name is shown with an arrow to the new one
- Alerts moved above mappers and made less obtrusive
  - unused cards and templates aren't alphabetically sorted anymore, meaning they now show up in the correct order
- Responsive layout
  - sticky headers
  - doubled max-width of body
  - two colums in landscape view
  - window has a minimum height
- Performance optimization for large notetypes
- Several small tweaks and fixes

Todo
---
- [x] test if there are any regressions caused by my performance optimization f42beee45c27dba9433d76217fb583b117fb5231
- [x] make the headers behave [like this](https://webthemez.com/demo/sticky-multi-header-scroll/index.html) to make more use of the exclamation mark icons

---

@hgiesel suggested a ScrollSpy, so I tried bootstrap's implementation. Altough it worked (kind of), I came to the conclusion that it really isn't necessary for that screen. We already have a `StickyContainer` component which can be used to create a layout like this one, which seems even more intuitive: https://webthemez.com/demo/sticky-multi-header-scroll/index.html

I didn't want to over-complicate this PR though, so I kept it basic for now.

My current implementation is using breakpoints for `Col` items. This works automatically in conjunction with the sticky headers without any extra maintenance cost.

> Also, the window makes the assumption that if I increase the window width, it must mean I want to see it in two columns, which is not necessarily true.

I'm struggling to see why someone would increase a window's width for any other reason than to see more of the content. The body width is maxed at `35em`, so any further increase in width would be a waste of screen space currently.

>Also, if I had one item focused with my eyes, it would suddenly jump somewhere completely different - that's annoying.

I tried to replicate this, but I couldn't make out any major jumps that would disturb the reading flow. Also, once a user has found an ideal window size, he probably won't change it often afterwards.


rslib/src/notetype/notetypechange.rs
```diff
@@ -77,18 +79,20 @@ impl Collection {
...
             input: ChangeNotetypeInput {
                 current_schema,
                 note_ids: vec![],
+                old_notetype_name: old_notetype_name.to_string(),
                 old_notetype_id,
                 new_notetype_id,
                 new_fields,
                 new_templates,
             },
+            old_notetype_name: old_notetype_name.to_string(),
```
kleinerpirat, line 95: This is pretty bad. I don't know Rust, so I had to make it work somehow :sweat_smile:

ts/change-notetype/MapperRow.svelte
```diff
@@ -21,29 +21,34 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-            {#each $info.getOldNamesIncludingNothing(ctx) as name, idx}
-                <option value={idx}>{name}</option>
-            {/each}
+            {#if active}
+                {#each oldNames as name, idx}
+                    <option value={idx}>{name}</option>
+                {/each}
+            {:else}
+                <option value={0}>{current}</option>
+            {/if}
```
kleinerpirat, line 51: By only loading the selected `option` and loading the rest on focus, we can significantly reduce load times and window resize lag.

Say we have 500 fields and 500 templates, instead of 2*500² = 500.000 options, we only load 1000.

kleinerpirat (commented):
I hope the overall approach is alright. Any help (e.g. with Rust) is highly appreciated :)

rslib/src/notetype/notetypechange.rs
dae: Slightly more efficient:

```diff
diff --git a/rslib/src/notetype/notetypechange.rs b/rslib/src/notetype/notetypechange.rs
index 0c341a608..a910c6ca0 100644
--- a/rslib/src/notetype/notetypechange.rs
+++ b/rslib/src/notetype/notetypechange.rs
@@ -79,20 +79,20 @@ impl Collection {
             .ok_or(AnkiError::NotFound)?;

         let current_schema = self.storage.get_collection_timestamps()?.schema_change;
-        let old_notetype_name = &old_notetype.name.to_string();
+        let old_notetype_name = &old_notetype.name;
         let new_fields = default_field_map(&old_notetype, &new_notetype);
         let new_templates = default_template_map(&old_notetype, &new_notetype);
         Ok(NotetypeChangeInfo {
             input: ChangeNotetypeInput {
                 current_schema,
                 note_ids: vec![],
-                old_notetype_name: old_notetype_name.to_string(),
+                old_notetype_name: old_notetype_name.clone(),
                 old_notetype_id,
                 new_notetype_id,
                 new_fields,
                 new_templates,
             },
-            old_notetype_name: old_notetype_name.to_string(),
+            old_notetype_name: old_notetype_name.clone(),
             old_field_names: old_notetype.fields.iter().map(|f| f.name.clone()).collect(),
             old_template_names: old_notetype
                 .templates
```

ts/change-notetype/MapperRow.svelte
dae: Nice!

dae: ... but unfortunately this does not appear to work on mobile - only a single option is shown when tapping on a dropdown, until it is tapped again. Maybe the focus event is not firing quick enough? You can test locally with a mobile device using ts-run - if you're an Android user, it's possible the same issue exists in mobile Chrome: https://github.com/ankitects/anki/issues/1174

dae: // use #testXXXX where XXXX is notetype ID to test

dae:
Some nice changes here. Not sure about the white background - it's a bit blinding, and inconsistent with other screens like the deck options, graphs, and Qt screens like preferences, etc.

ts/change-notetype/MapperRow.svelte
kleinerpirat: That's a bummer :/ My phone cable is a bit loose, so I couldn't get mobile debugging to reliably work right now. I gotta go to bed now, but I'll test that tomorrow.

Generally, writing a `Lazy` Svelte component might be a good idea, since the editor shows similar performance issues :thinking:

dae:
It varies by platform - on macOS it is darker. I'm not sure we necessarily want to be imitating the Qt design exactly in any case, but I'm not a designer. Maybe @hgiesel has some ideas.

Screen Shot 2021-11-19 at 12 13 13 pm

abdnh:
Looks nice.

Some notes:

1. Maybe #main could use some top padding?
```diff
diff --git a/ts/change-notetype/change-notetype-base.scss b/ts/change-notetype/change-notetype-base.scss
index 0cc2e5acb..846e31503 100644
--- a/ts/change-notetype/change-notetype-base.scss
+++ b/ts/change-notetype/change-notetype-base.scss
@@ -24,7 +24,7 @@ html {
 }

 #main {
-    padding: 0 1em 1em 1em;
+    padding: 0.5em 1em 1em 1em;
     height: 100vh;
 }
```

2. The arrow needs to be inverted for right-to-left languages. Maybe using something like:
```diff
diff --git a/ts/change-notetype/NotetypeSelector.svelte b/ts/change-notetype/NotetypeSelector.svelte
index dd3fa45ce..84d4fba4f 100644
--- a/ts/change-notetype/NotetypeSelector.svelte
+++ b/ts/change-notetype/NotetypeSelector.svelte
@@ -13,7 +13,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html

     import LabelButton from "../components/LabelButton.svelte";
     import Badge from "../components/Badge.svelte";
-    import { arrowRightIcon } from "./icons";
+    import { arrowRightIcon, arrowLeftIcon } from "./icons";
     import SelectButton from "../components/SelectButton.svelte";
     import SelectOption from "../components/SelectOption.svelte";
     import SaveButton from "./SaveButton.svelte";
@@ -40,7 +40,12 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         </Item>
         <Item>
             <Badge iconSize={70}>
-                {@html arrowRightIcon}
+                {#if window.getComputedStyle(document.body).direction == "rtl"}
+                    {@html arrowLeftIcon}
+                {:else}
+                    {@html arrowRightIcon}
+                {/if}
+
             </Badge>
         </Item>
         <Item>
diff --git a/ts/change-notetype/icons.ts b/ts/change-notetype/icons.ts
index 05ca4142e..3f3701cf5 100644
--- a/ts/change-notetype/icons.ts
+++ b/ts/change-notetype/icons.ts
@@ -4,6 +4,7 @@
 /// <reference types="../lib/image-import" />

 export { default as arrowRightIcon } from "bootstrap-icons/icons/arrow-right.svg";
+export { default as arrowLeftIcon } from "bootstrap-icons/icons/arrow-left.svg";
 export { default as exclamationIcon } from "bootstrap-icons/icons/exclamation-circle.svg";
 export { default as plusIcon } from "bootstrap-icons/icons/plus-lg.svg";
 export { default as minusIcon } from "bootstrap-icons/icons/dash-lg.svg";
```
Or maybe just using `transform: scaleX(-1);`.

kleinerpirat:
Thanks, Abdo!

My initial thought was to implement this with a `directionStore`, but that would probably be too much overhead for this screen, so I went for your suggestion.

I also like the slight padding on top of main. Looks cleaner :+1:

kleinerpirat:
I'm on vacation until Sunday, so I won't be able to work much on this PR until Monday. I left you with this alternative approach that ditches the two-column responsive layout for a pure vertical + sticky alternative. I also tweaked the colors.

Known issue: Alerts messing up header offsets. Trying to fix this currently.

cqg2:
Looks nice. I still think that since most of us (at least those using LTR languages) write (and think) left to right. selectable items in the right colums would work better.
142516796-23524ead-8acd-4607-8dd8-77659d9b1868

kleinerpirat:
Absolutely agree. It makes sense to put it where the arrow points: for ltr to the right and for rtl to the left. I was so focused on the technical details that this obvious issue didn't come to mind 😊

hgiesel:
> We'll need to allow multiple selection on these dropdows.

One agenda with the new ChangeNotetype was to avoid one issue of the old one: Some input selections could change the active values of other input elements.

In the following examples, I'll use capital letters for old field names `A, B, C, ...`, and minuscules for new field names `a, b, c, ...`.

If we had the mapping `{A -> a, B -> b}`, and then change the mapping `B -> b`, to `B -> a`, we had to use a replacement value for `A`, because we can't have two old fields map to the same new field. This problem would still exist with multiple selections to some degree.

With multiple selections being possible, the mapping would look more like: `{A -> {a}, B -> {b}}`, and making the selection as above would change it to: `{A -> {}, B -> {a}}`, unselecting the field for the first field.

That's much better, but it's still an awkward design. We'd have to couple all select buttons together. Also, the user might not notice an automatic change further up on the page, if he changes something further down. Also `select` with `multiple` turned on doesn't have a good default on AnkiMobile imo, leaving it up to us to make a good SelectMultipleButton. I think to really fit the semantics of the operation, we'd have to use Drag'n'Drop here, something like this:

AssignFields

It also makes it obvious, what elements are not included in the mapping, without the need of warnings.

I'm not saying we should do this btw (in fact, we can't: D'n'D doesn't currently work in Anki web views and has poor support in Mobile unfortunately). I think we should leave it like it is for now. I was done like this for a reason, even if it feels a bit akward: It's easier to deal with than the "easy" alternatives.

kleinerpirat:
The way I would have dealt with this is to either
- gray unavailable options out and give it a tooltip that links to the field responsible for the block
- not load unavailable options at all

so that wrong selections can never occur.

---
Here's a design that works like I imagined that: https://svelte-multiselect.netlify.app/
(On the single select sample, two elements are unselectable.)

dae:
I recommend you split this PR into multiple PRs instead of trying to do everything at once. The large notetype optimization is blocking merging at the moment, and attempting to switch the dropdowns to the right is a bigger project that will further delay this. Better to get the non-controversial parts merged in, then the other changes can be discussed/worked on in isolation.

kleinerpirat:
Yeah, I was considering putting my work on this on hold for the next two weeks due to my uni workload, so that's a good idea :)
Also sorry about misnaming the branch, only noticed that today :smile:

dae:
Thanks Matthias, it seems to work well. I noticed only one minor issue: the slide animation is a bit jumpy:

https://user-images.githubusercontent.com/180542/143158637-8b9586bf-ae66-42c0-b1f0-1ac9d30da1dd.mp4

PR #1500, opened 2021-11-20
docs: Install pip in venv

dayjaby:
Running the build instructions in a fresh environment failed, because pip was missing in pyenv/bin. With this instruction, pip can be installed to this folder.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -24,6 +24,7 @@ python.org or from your distro.
 
 ```
 $ python3.9 -m venv ~/pyenv
+$ curl https://bootstrap.pypa.io/get-pip.py -o - | ~/pyenv/bin/python -
 $ ~/pyenv/bin/pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


dae:
venv installs pip by default, and I have not found that step to be required on any of the systems I have tested. Maybe your distro has split pip out into a separate Python package and altered the default behaviour?

https://docs.python.org/3/library/venv.html

dayjaby:
Confirmed, did a run in a ubuntu 20.04 docker container (my host with this no pip3.9 in pyvenv/bin was also ubuntu 20.04). However, it worked there:

```
# docker run -it ubuntu:20.04 /bin/sh
# apt update
# apt install python3.9 python3.9-venv
# python3.9 -m venv pyvenv
# ls pyvenv/bin		
Activate.ps1  activate	activate.csh  activate.fish  easy_install  easy_install-3.9  pip  pip3	pip3.9	python	python3  python3.9
```

Thanks! Guess no need to investigate further.

PR #1501, opened 2021-11-22
Remove all_names in notetypechooser

Arthur-Milchior:
The only purpose is to stop having warning in my console because all_names is deprecated


dae:
Thanks Arthur!

PR #1502, opened 2021-11-22
Mathjax editor improvements

### LINKED ISSUE #1491: mathjax editor is leaking memory
https://forums.ankiweb.net/t/new-toolkit-and-packaging-test-round-2/14513/14

hgiesel:
Improvements:
- Supports pressing Enter to _accept_ Mathjax, Shift-Enter for newline
- Show this information as placeholder on empty Mathjax
- Give Mathjax a min-width for less resizing on small sizes (and no resize when placeholder disappears)
- Delete button
- Only use one ResizeObserver for all Mathjax images (hopefully closes #1491 ?)

Unrelated:
- Add bazel directories to .prettierignore and run from root once. This formatted json and css files in rslib and ftl. I've executed it by accident on the root sometimes, which formatted bazel files, which broke bazel for me. I saw no reason not to include those files in other directories in formatting.


dae:
Thanks Henrik! 

@hikaru-y does this resolve the mem leak for you?

dae:
This introduced a warning during build - maybe we could work around it with a temporary variable?

Warning - unrecognized UnaryExpression operator 40!
                This is an edge case unaccounted for in svelte2tsx, please file an issue:
                https://github.com/sveltejs/language-tools/issues/new/choose
                 -$imageHeight

hikaru-y:
@dae 
I tried the latest main branch(675155e025ee2971cc0835614796b1913cc49d61), and unfortunately the issue(#1491) remains. I looked into the cause a bit. It appears that the MathJax images (SVG data URI) are cached in memory and are not released until the end of the session. I think this is because the contents are dynamically rewritten in the editor, and a page refresh never happens.

<details>
 <summary>Screenshot (DevTools)</summary>

mathjax-editor
</details>

To be honest, I'm not sure if this is an issue that needs to be solved or not, because the amount of memory usage that increases each time a card is selected is so small that it shouldn't be a big problem for users unless they select hundreds of cards in total.

dae:
Thanks for checking. If the leak is small and does eventually get cleaned up, I think you're probably right that we don't need to worry about it.

PR #1505, opened 2021-11-23
Fix some issues with new surround buttons

hgiesel:
I've fixed the active button not being influenced by the input trigger. You can see the change in behavior, if you put the cursor into a bold section (=> bold button will be active), and then click the bold button (=> bold button will be inactive, indicating that the next input is non-bold).

I've also fixed two surround algorithm issues and added tests for them.


dae:
Seems to work well - thanks for the quick fix, Henrik. The merging/normalizing of styling you added previously is also nice by the way :-)

PR #1510, opened 2021-11-27
Fix memory leak in AnkiWebView

hikaru-y:
Since commit f2173fddb04a95a3eff17494c9845736997131e4, a memory leak occurs every time all sub-windows with `AnkiWebView` inside are opened and closed.
<details>
 <summary>Screenshot of Task Manager (627f910635893f8205f7004572ba2ac7f3d987b8 on Win10, after opening and closing sub-windows several times)</summary>

anki-taskmanager
</details>

Currently, `on_theme_did_change` is added to `gui_hooks.theme_did_change` when `AnkiWebView` is initialized, and it is removed from the list when the finalizer function (`__del__()`) is called. However, at the point when the parent widget of `AnkiWebView` is deleted, the reference to `AnkiWebView` still remains in `gui_hooks.theme_did_change`, so the finalizer is not called and the object continues to exist until Anki shuts down.


dae:
Thanks Hikaru!

PR #1511, opened 2021-11-28
Add missing editorToolbar export

kleinerpirat:
This fixes preview and add-on buttons not showing due to a deletion in https://github.com/ankitects/anki/commit/610ef8f04396db124909975b6842178f1116e77f#r60989499 which I believe was accidental @hgiesel?

(I put it below the noteEditorPromise export because I figured it makes more sense there.)


dae:
Thanks Matthias! I'll assume it was accidental and merge this in for now; if it was deliberate Henrik, please let us know.

PR #1514, opened 2021-11-30
Fix bug(s) caused by deleting a notetype currently selected in AddCards

kleinerpirat:
## Issue
The following TypeError is thrown when deleting a notetype that's currently being used in the `AddCards` editor: 
```
Error
...
Anki 2.1.50 (dev) Python 3.9.7 Qt 6.2.1 PyQt 6.2.1
Platform: Linux
Flags: frz=False ao=False sv=2
Add-ons, last update check: 2021-11-30 11:09:28

Caught exception:
  ...
  File "/home/kleinerpirat/.cache/bazel/_bazel_kleinerpirat/6dc00256f718f1f61130be7a17f037a7/execroot/ankidesktop/bazel-out/k8-fastbuild/bin/qt/runanki.runfiles/ankidesktop/qt/aqt/addcards.py", line 142, in on_notetype_change
    elif n < len(old.note_type()["flds"]):
TypeError: 'NoneType' object is not subscriptable
```
### Reproduction steps
- Open the add window.
- With that window open, delete the current notetype (of the editor) via `Manage Note Types`.

## Fix
I removed the requirement for `old.note_type()` to return a value and made the logic a bit less verbose.

Please let me know if there are any issues with this approach.


kleinerpirat:
I don't see how my changes could introduce this, but I spotted another exception:
```
Error
  ...
  File "/home/kleinerpirat/.cache/bazel/_bazel_kleinerpirat/6dc00256f718f1f61130be7a17f037a7/execroot/ankidesktop/bazel-out/k8-fastbuild/bin/qt/runanki.runfiles/ankidesktop/qt/aqt/editor.py", line 357, in onBridgeCmd
    self.note.fields[ord] = self.mungeHTML(txt)
IndexError: list assignment index out of range
```

### Reproduction steps
- select `Basic` notetype in the editor
- with the editor open, delete `Basic` notetype via `Choose Note Type -> Manage`
- close `Manage Note Types` dialog
- switch to another notetype from (still opened) `Choose Note Type` dialog

On main, only the initial TypeError is thrown in this scenario. My guess is you can't get far enough to cause the other error without this PR.

kleinerpirat:
Turned out to be less complex of an issue than expected.
Some console logs revealed that there's a short transition phase where `note.fields` is empty.

Similarly to how late blurs are ignored, I handled this expected IndexError with the message `ignored blur on notetype change`.

dae:
Thanks Matthias, the original fix seems reasonable. The latter fix feels more like we are papering over the issue instead of fixing it properly, but that's not your fault - the editor already had problems with late blurs. Hopefully if we move more of the add screen state into JS in the future, we can solve this better.

PR #1517, opened 2021-12-01
Run media trashing in background

### LINKED ISSUE #1389: media deletion needs to be run in the background
https://forums.ankiweb.net/t/enhance-delete-bulk-of-unused-media-files-seems-a-blocked-process/13588

RumovZ:
Closes #1389.

I tried using `QueryOp`, but it didn't work. Instead I got warnings like `Qt warning: QObject::setParent: Cannot set parent, new parent is in a different thread`. Maybe because we're already on a child thread due to how media checking is implemented?


dae:
My apologies, some oversights on my part made this harder than it should have been:

- QueryOp was mistakenly trying to start the progress window on a background thread
- I'd forgotten that we want to show tangible progress instead of an infinite progress bar, which made the QueryOp invocation a bit less trivial

Issue #1516, opened 2021-12-01
Webview regressions


RumovZ:
The webviews look mostly great now and there don't seem to be any WIPs right now, so I've collected some regressions (and nitpicks 🙂) here. I didn't try to make PRs, because I'd likely make things worse for other platforms and devices.
@hgiesel @kleinerpirat 

General
---
- [x] Nightmode scrollbars have gone. (Replaced by lightmode scrollbars.)

Deck Options
---
- [x] 1-px top margin (Change notetype window on the right for comparison)
![grafik](https://user-images.githubusercontent.com/48286681/144215825-867d7d71-f18a-49a2-a9ad-c559bfa975e0.png) ![grafik](https://user-images.githubusercontent.com/48286681/144215923-a4535be6-4535-464d-9553-b35775b4f064.png)

Change Notetype
---
- [x] (More of a question) What is the reason for the background box? Looks quite different than other dialogs.
- [x] Name of target notetype is cut off at the bottom. ("g"s, "j"s, etc.)
![grafik](https://user-images.githubusercontent.com/48286681/144217189-4fb03901-12b8-4b38-9d5a-355c2760adcf.png)

Card Info
---
- [x] Container max-width was removed in ab6a68ec4913e2205b25ab1aaffbb4e9c32efc5f, leading to excessive padding in wide windows.
- [x] Almost no margin on the left, making it cumbersome to read on crowded backgrounds. 
![grafik](https://user-images.githubusercontent.com/48286681/144220848-ca66d849-34d9-495c-a975-91213c957c06.png)
- [x] Differrently aligned columns lead to extremely uneven padding. I don't have any strong opinions on which column should have which alignment, but it doesn't look very pleasing with the current mix. Interval and Ease are hardly recognisable as separate columns without a closer look.
![grafik](https://user-images.githubusercontent.com/48286681/144220985-399ffd44-42a6-42ab-ad65-788ec47c6605.png)

Editor
---
- [ ] When removing a list tag without having text selected, field focus is lost. When I do it by shortcut, I also get several of these `JS error /_anki/js/editor.js:42866 Uncaught (in promise) IndexSizeError: Failed to execute 'getRangeAt' on 'Selection': 0 is not a valid index.`
- [x] Add-on button styles.

Congrats
---
[...]

PR #1521, opened 2021-12-02
Show downwards arrow on SelectButton

kleinerpirat:
image

The arrow wasn't showing because the linear-gradient background for `button` overrides `background-image` for the select arrow.

Select can't have pseudo-elements, so I had to use an absolutely positioned div.

---
Also, I replaced `overflow-x: hidden` with `text-overflow: ellipsis`:

image

@RumovZ (re: https://github.com/ankitects/anki/issues/1516) To find the reason for the cutoff text, I'll have to set up a dev build on Windows first. Maybe someone else already has an idea what might be causing it?


hikaru-y:
> Maybe someone else already has an idea what might be causing it?

It appears that the following line causes the bottom of the text to be cut off. The deck options screen also has the issue on Win10.
https://github.com/ankitects/anki/blob/d7f55556422ffdd6cd1a0c5fb4e43e00bef31293/ts/components/SelectButton.svelte#L43
<details>
 <summary>Screenshot of Inspector</summary>

text_cut_off
</details>

kleinerpirat:
Thanks Hikaru, that helped a lot!

That variable is used for multiple properties of elements inside a `ButtonToolbar` componen, but removing the line for `ButtonSelect` doesn't seem to cause any harm. It should be fixed for both Change Notetype and Deck Options.

hikaru-y:
I guess that `height: var(--buttons-size);` is set in order to make the height of `<select>` elements match the height of `<button>` elements exactly, so it might be better not to remove that line. Removing that line would cause the height of the elements to differ by a few pixels. Having just played around with CSS a bit, I found that setting `padding-top/bottom` to `0` would prevent the text from being cut off, as follows:
```diff
diff --git a/ts/components/SelectButton.svelte b/ts/components/SelectButton.svelte
index 29df4027c..23aa7865e 100644
--- a/ts/components/SelectButton.svelte
+++ b/ts/components/SelectButton.svelte
@@ -43,6 +43,10 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         height: var(--buttons-size);
         /* Long option name can create overflow */
         overflow-x: hidden;
+        padding: {
+            top: 0;
+            bottom: 0;
+        }
     }

     .visible-down-arrow {
```
select-padding-0

kleinerpirat:
I can't test this right now, but the text cropping issue should be fixed @RumovZ thanks to @hikaru-y.

Thanks Hikaru, once again, for the effort you put into this! :hugs: Saved me a lot of time.

kleinerpirat:
@abdnh I noticed some rtl issues with dropdowns.

The default `<select>` arrow doesn't respond to rtl, but the text does:

image
The space below the arrow seems to be reserved, so the options got this awkward padding on the right.

Also, the `SelectButton`' (`NotetypeSelector`) and its options are left-aligned.
image



I'd put that arrow to the left for rtl and fix that text alignment. I've seen select implementations on the web that keep the arrow on the right, so I'm wondering if it would be unusual for to have it on the left?

kleinerpirat:
The `SelectButton` component now adjusts to rtl. I consider it a temporary solution.

To do it generally for all select's, it's best to create our own `Select` component some day. Because `<select>` is notoriously difficult to style, we'd probably be better off not using that tag at all, but something like the Bootstrap [dropdowns](https://getbootstrap.com/docs/4.0/components/dropdowns/).

We could also enable multi-selection for such a component (like [Svelte Multiselect](https://svelte-multiselect.netlify.app/)), which would enable mapping fields in `ChangeNotetype` in the other direction.

abdnh:
> so I'm wondering if it would be unusual to have it on the left for rtl?

It won't be unusual. Windows shows the arrows on the left.

image

I tried to get the arrows to show on the left when testing https://github.com/ankitects/anki/pull/1499 and I concluded likewise that `select`'s arrow doesn't respond to RTL, but then I tested today (with [this page](https://interactive-examples.mdn.mozilla.net/pages/tabbed/select.html) for example) and found that the arrow does actually flip... Not sure why it doesn't flip automatically in Anki.

dae:
Thanks everyone!

Re switching not happening automatically, maybe that's bootstrap's styling interfering with the defaults (eg via form-select class)

PR #1522, opened 2021-12-02
Align design of Change Notetype screen with rest of UI

kleinerpirat:
@RumovZ re: https://github.com/ankitects/anki/issues/1516
Removes the background (+ its CSS variable) and border of the scroll-area and adjusts the margins a bit to make the screen look more like the current deck options screen.

It looks a bit cleaner/less cluttered and also saves some horizontal space for mobile use.

### Current vs. PR

Also fixes a 1px "clipping" issue that sometimes occurred for the templates header:

image

@dae If you prefer this design, I'd wait with merging until we see which margin @hgiesel uses for that bottom border of the `StickyContainer` in the deck options screen. Then I can adjust `NotetypeSelector` to that.


dae:
Thanks @kleinerpirat, I think that looks better too. Will get back to your other PRs tomorrow.

dae:
@kleinerpirat @hgiesel: I'd like to include this in the next build, so will merge this in for now - if anything needs changing, we can sort it out before beta2.

kleinerpirat:
Sure, it's nothing of importance really :+1:  I can always align that padding later.

PR #1526, opened 2021-12-04
Flip arrows of Bootstrap-styled `<select>`s for RTL langs

abdnh:
Bootstrap's `form-select` class was causing the arrows not to flip automatically as you said in https://github.com/ankitects/anki/pull/1521#issuecomment-985875363

Instead of duplicating styles across all places where `<select>` is used, I chose to add the `rtl` class to `body` and put the fix in `base.scss`.

### DIFF
--- a/qt/aqt/theme.py
+++ b/qt/aqt/theme.py
@@ -11,9 +11,7 @@ import sys
 from ctypes import CDLL
 from dataclasses import dataclass
 
-import anki
 import aqt
-from anki.lang import is_rtl
 from anki.utils import is_lin, is_mac, is_win
 from aqt import QApplication, colors, gui_hooks
 from aqt.qt import (
@@ -141,8 +139,6 @@ class ThemeManager:
             classes.extend(["nightMode", "night_mode"])
             if self.macos_dark_mode():
                 classes.append("macos-dark-mode")
-        if is_rtl(anki.lang.current_lang):
-            classes.append("rtl")
         return " ".join(classes)
 
     def body_classes_for_card_ord(
--- a/qt/aqt/theme.py
+++ b/qt/aqt/theme.py
@@ -11,7 +11,9 @@ import sys
 from ctypes import CDLL
 from dataclasses import dataclass
 
+import anki
 import aqt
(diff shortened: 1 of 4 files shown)


dae:
This appears to have the odd side-effect of the congrats screen shrinking:

Screen Shot 2021-12-05 at 12 39 27 pm

body_class is only used by our older webviews IIRC, so it is sort of legacy code at this point. Perhaps instead of reusing it, we could set an rtl class on the `<html>` element, like we do for .night-mode in the newer webviews.

One other alternative worth considering is setting `dir=rtl` on the HTML element in setupI18n(). This will likely cause problems on some screens, so we'd need to go through each of the webviews, look for problems, and add an explicit ltr in cases where reversing the direction doesn't make sense. Any thoughts @hgiesel @hikaru-y @kleinerpirat ?

abdnh:
I didn't notice any problems trying either alternative.

We already set `direction: ltr` or `dir=ltr` in some places, like the stats screen. I'm not sure how using `dir=rtl` instead of `direction: rtl` can cause new issues.

kleinerpirat:
`dir=rtl` for `<html>` sounds good to me.

Flexbox layouts can automatically reflow if we use direction-aware properties `flex-start and flex-end` instead of explicit ones, so the future maintenance effort largely depends on our CSS practices.

For changes that can't happen automatically, we can use CSS inside components like this:
```
[dir="ltr"] .class {
  ...
}
```
And for things that are better done on the TS/HTML side, we can store `document.dir` in a context so components have a unified means to access the direction of their screen.

---
Fyi @dae @abdnh @hgiesel I'm currently working on a `Select` component (with optional multi-select) that can replace any `<select>` within Anki, but it's too large of a change for 2.1.50 probably. Hope there are no duplicate efforts :crossed_fingers:

hgiesel:
Yes, we should be using `dir="rtl"` instead of CSS `direction`. [Source here](https://www.w3.org/International/questions/qa-bidi-css-markup).

dae:
Looks great! Thanks everyone.

@kleinerpirat provided it's mobile friendly, sounds good :-)

PR #1529, opened 2021-12-05
Block {,max-}{width,height} from being copied

### LINKED ISSUE #1515: Copy-pasting images includes inline styles
### Reproduction steps
- Insert an image into the editor.
  - It will be styled with the new image handle, but the styles do not show in the HTML editor.
- Select the image in the `RichTextInput` and copy-paste it below.
  - The pasted `` tag is now polluted with tons of inline styling.

### Sample
```html
<!-- Original image -->
<img src="image.jpg">
<!-- Pasted image -->
<img src="image.jpg" style="max-width: min(250px, 100%) !important; max-height: 125px !important; width: 195px !important; height: auto !important;">
```

This issue might not be isolated to images, but at the moment, it's all I could reproduce (which is great).

In the past, I've seen borders get converted into `<hr>` and other funky stuff that really surprised me, because one would not expect styles from external sources get included inline like that. Perhaps the new toolkit has ironed out most of these issues.

hgiesel:
Closes #1515.

Btw, I noticed that when pasting images on GitHub, they also allow you to edit the `width` property, similar to how it works in Anki now. Seems like we are on the right track with that feature.

### DIFF
--- a/ts/html-filter/styling.ts
+++ b/ts/html-filter/styling.ts
@@ -25,6 +25,10 @@ const stylingInternal: BlockProperties = [
     "background-color",
     "font-size",
     "font-family",
+    "width",
+    "height",
+    "max-width",
+    "max-height",
 ];
 
 const allowPropertiesBlockValues =


dae:
That's reassuring! Thanks as always Henrik.

PR #1532, opened 2021-12-05
Decrease button and tag size for Linux and Windows

kleinerpirat:
https://forums.ankiweb.net/t/anki-2-1-50-beta/15608/22

This solves the regressions I mentioned in https://github.com/ankitects/anki/issues/1516, which were probably caused by the toolkit update providing new size values. I adjusted the size factor for non-Apple platforms and refactored the legacy button styles a bit.

Factor 1.8 looks good on Linux, but I don't know whether it's enough for Windows (looks pretty big on [cqg's screenshot](https://aws1.discourse-cdn.com/standard11/uploads/anki2/original/2X/5/5ed9ca2bc1346d3c21433069648664ef79524c03.png)). @hikaru-y @RumovZ @abdnh what do you think?

### DIFF
--- a/ts/editor/OldEditorAdapter.svelte
+++ b/ts/editor/OldEditorAdapter.svelte
@@ -77,7 +77,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         return fontFamily;
     }
 
-    let size = isApplePlatform() ? 1.6 : 2.0;
+    let size = isApplePlatform() ? 1.6 : 1.8;
     let wrap = true;
 
     let fieldStores: Writable<string>[] = [];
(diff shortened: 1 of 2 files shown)


hikaru-y:
The sizes of the toolbar buttons and the tag icon look just right to me.

I noticed an issue regarding the modification of `legacy.scss`. Probably due to the removal of the width and height of `.topbut`, if an add-on uses a large image for the toolbar icon, the image will stick out from the button, as follows:

pr1532
.

kleinerpirat:
Oh wow, we could put that screenshot in a modern art museum! :art: Thanks for the quick report on this, Hikaru!

dae:
Thanks Matthias and Hikaru! The buttons still feel a bit big on a Windows machine here compared to other parts of the UI, but it's a step in the right direction, and I'm guessing that may change with different DPI/text scale settings.

Screen Shot 2021-12-06 at 6 57 18 pm

Will merge this in to keep things moving, but further feedback/changes are of course welcome :-)

PR #1533, opened 2021-12-05
Update about.py

kleinerpirat:
I noticed some important contributors (and I) can't show off with their name on the Anki About screen yet.
@hgiesel @BlueGreenMagick @hikaru-y I can't force anyone into this, so just react to this comment with
- :+1: if you **want** to get added with this PR (or don't have anything against it)
- :-1: if you **don't** want to get added with this PR

and then I mark this PR as ready.

(I used the names from `CONTRIBUTORS`)
@abdnh I think your initials are already on the list, correct me if I'm wrong ^^

### DIFF
--- a/qt/aqt/about.py
+++ b/qt/aqt/about.py
@@ -223,6 +223,7 @@ def show(mw: aqt.AnkiQt) -> QDialog:
             "Hikaru Yoshiga",
             "Matthias Metelka",
             "Sergio Quintero",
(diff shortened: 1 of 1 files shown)


kleinerpirat:
I've offered this to the new additions privately. I suppose @hgiesel isn't against it, so it's ready @dae ;-)

dae:
Thanks for organising this Matthias!

PR #1535, opened 2021-12-06
Add browser action to create note copy

### LINKED ISSUE #1512: copying cards/notes
Re https://forums.ankiweb.net/t/include-copy-duplicate-note-in-anki-desktop-or-as-official-plugin/15287

My concerns about such a feature are its potential for abuse, as it's not too uncommon to see users asking for this because they want to add reverse cards or (which is better done with a template adjustment), or want to place the same cards in multiple decks (which will make the card scheduling less effective).

Maybe one way of avoiding the downsides would be to have it apply to the add screen, instead of cloning notes directly? We could have an action that takes the selected note, and opens the add screen with its content and note type pre-filled, and then the user could edit as desired before adding the new note.

Thoughts?

RumovZ:
Closes #1512.
I reckoned the action doesn't need to be added to the reviewer as one can simply press <kbd>B</kbd> > <kbd>Ctrl+Alt+E</kbd>. But I'll add it if anyone disagrees.
I chose this shortcut because adding notes, which opens the same dialog, is <kbd>Ctrl+E</kbd> and <kbd>Ctrl+Shift+E</kbd> is already used for exporting. However, considering Qt's AltGr dilemma, this might be a terrible choice—<kbd>AltGr+E</kbd> is `€` on European keyboards. What else could the shortcut be?


ftl/core/browsing.ftl
```diff
@@ -28,6 +28,7 @@ browsing-change-note-type2 = Change Note Type...
 browsing-change-notetype = Change Notetype
 browsing-clear-unused-tags = Clear Unused Tags
 browsing-confirm-saved-search-overwrite = A saved search with the name { $name } already exists. Do you want to overwrite it?
+browsing-create-copy = Create Copy...
```
dae, line 31: Minor point: the '...' is convention on desktop apps, but seems to be uncommon on mobile, so by using it we're sort of creating desktop-only strings. Not sure what the best way is to handle this - we could put the string in desktop/ instead, or create one string that references the other - but I fear that may end up being more complicated for the translators then just asking them to translate the string twice.

(the instances above this one were imported from the old translations)

dae (commented):
Thanks Rumo, the copying part seems to work well. Subsequent adding fails though, because the id hasn't been reset to 0.

If the shortcut will prevent people from typing the euro and there's no solution for that on the horizon, I guess we're best off picking a different shortcut. Could ctrl+shift+c work?

ftl/core/browsing.ftl
RumovZ: What if we had a

```
# [verbose explanation]
new-window-action = { $action }...
...
browsing-create-copy = Create Copy
```

and a util that takes an action name and wraps it in `tr.new_window_action(action=action_name)` if appropriate?

dae: That's a good idea :-)

dae:
Thanks for looking into that, looks good. Looks like I had misused the ellipses in some of those labels. 😅  I do find myself wondering how many users these days actually understand/benefit from this convention though!

RumovZ:
> I do find myself wondering how many users these days actually understand/benefit from this convention though!

I tell myself it's subconcious. 😅

abdnh:
> Could ctrl+shift+c work?

This conflicts with the cloze shortcut in the browser.

dae:
Thanks for catching that Abdo!

PR #1538, opened 2021-12-07
Disable saving fullscreen geometry

RumovZ:
This preserves the last non-fullscreen state. Before on macs, geometry was reset when a window was closed while in fullscreen. Was there a reason for that?


dae:
Thanks Rumo! The original change was in 23a3413c23c2ee28956550ef204464c5f4176c8a (found via git blame). I'm guessing this is not a very well-exercised part of Qt.

PR #1539, opened 2021-12-07
webview autoplay in accordance to Anki's setting v2

nwwt:
Adds back setting Chromium's autoplay policy according to Anki's, this time without globals.


dae:
Thanks!

PR #1540, opened 2021-12-08
Add .aac files to editor

TheFeelTrain:
.aac audio files work as expected, but Anki doesn't recognize them by default when attaching files.


dae:
Thanks!

PR #1541, opened 2021-12-08
Replace Create Copy shortcut with accelerator

RumovZ:
Thanks for pointing out the collision, @abdnh. My regex failed to match something like "Control+Alt?+Shift+C".

This PR means the action is activated by <kbd>Alt+N, C</kbd>. Maybe we should use more of these? It reduces collisions and is easier to remember.


dae:
Silly me for forgetting about the cloze shortcut!

Accelerators are a Windows/Linux-only thing - they don't work on Macs (and even if they did, we'd want to put this in desktop/).

It's a bit of a monster, but another idea would be ctrl+alt+shift+n (to go with the existing ctrl+shift+n)?

dae:
(not sure it's the best solution - other suggestions welcome :-)

RumovZ:
> Accelerators are a Windows/Linux-only thing

Pity, that explains why they aren't used more.
But what if we start using key successions in general? That would have similar benefits in terms of conflict redcution and semantics. Unfortunately, something like <kbd>Ctrl+E, Ctrl+C</kbd> doesn't work currently, because <kbd>Ctrl+E</kbd> is triggered on pressing E. 

Don't we need a more general solution anyway? There are a lot of <kbd>Ctrl+Alt</kbd> shortcuts already and conflicts are bound to increase with the rollout of 2.1.50. I can think of two options:

1. Deactivate conflicting shortcuts on Windows, possibly with a switch to reactivate them manually.
2. Use the opportunity to redesign some shortcuts, avoiding <kbd>Ctrl+Alt</kbd> for now and utilising successions.

dae:
I've seen quite a few users struggle to figure out how to activate the LaTeX/MathJax shortcuts, so it might add some extra complexity. That said, less technical users are typically not using shortcuts at all.

With the AltGr fix though, perhaps this is less pressing?

RumovZ:
Yes, due to the limitation of only being able to use successions when no preceding combo is already assigned, this will only work well in the context of a major shortcut redesign. And hopefully, there is no reason for that now. 🙂

PR #1542, opened 2021-12-08
Add py3.9 to hooks

Arthur-Milchior:
This follows examples from efb1ce46d4b6982833b0ed324b9d105ba83ba394 I assume the
hooks were missed because those were not considered types but strings.

I did not even try to run pyupgrade and did the change manually, then used bazel format

I've honestly no idea what is the difference between `Any | None` and `Any`, but since the same strangeness occurred with `Optional[Any]` I decided to just keep it

### DIFF
--- a/pylib/anki/find.py
+++ b/pylib/anki/find.py
@@ -32,7 +32,7 @@ class Finder:
 
 def findReplace(
     col: Collection,
-    nids: List[NoteId],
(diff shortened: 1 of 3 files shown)


dae:
Thanks Arthur. 'object | None' better represents what we want here, so I've changed that too.

Issue #1018, opened 2021-02-12
refactor import/export code


dae:
I'd like to refactor the import/export code soon, moving some or most of it to Rust. This will probably come after the scheduling rework.

When working on this, we'll need to bear in mind use cases that the Special Fields add-on currently enables.

https://github.com/ankitects/anki/pull/507

hgiesel:
I did work on some SpecialFields PRs, and knowing that you intended to refactor it soon, I took some notes. While SpecialFields makes the importing algorithm much more powerful, I don't think the abstractions it provides are necessarily easy to understand. So the following are some ideas of how to better abstract the functionality SpecialFields provides:

### Changing Note Type upon import

* Currently, if the note type of a note changed, the note cannot be imported anymore. It would be nice, if the user could convert the note to the new note type similar to how he does it in the browser.
  * In fact, SpecialFields uses [that exact method](https://github.com/johnpincock/SpecialFields/blob/84f5a9dc2ad76f14f1f8915273467ad2d7fc96e6/Specialfields21/__init__.py#L114).
  * For that, first the note types of the imported deck would need to be imported, then all notes which follow the same note change pattern, like `convert from "MyNoteTypeV1" to "MyNoteTypeV3"` would have to be accumulated, so you don't have to make the card/field assignment per note.
  * For making this easier on the user, there could be also some improvements to the "Change Note Type" dialog:
    * Make the default assignment in that dialog based on equality of the field or card type names. I think it's currently based solely on the position of the field / card type.
    * Make it possible to assign multiple old fields to the same new field: The old fields could first be joined with `<br>` or possibly `<hr>`, and then written to the single new field
    * Disable changing the card types/fields by scrolling on them. When having a few more fields, you want to scroll down the overview and then accidentally change all the assignments.

### Private Fields:

* There could be some way to mark a field of a note type as private, probably in the Fields dialog.
* When exporting a private field -->:
  * field is left empty
* When importing a  private field <--:
  * field content of imported note is discarded in favor of the content that the field has in the local collection

### Private Tags:

* There could be some way to mark tags as private. This could be done via a prefix, or suffix. Or maybe an icon in the Browser sidebar, that can be activated. Or an item in the context menu of the sidebar, where after activating the tag has some visual indication to mark it private.
* Marking a hierarchical tag private, could also mark all of its descendants private (?) It's hard to say what would be the best behavior here.
* When exporting a private tag -->:
  * Tag is not exported
* When importing tags <--:
[...]

PR #1543, opened 2021-12-08
The importer list have a Hook

Arthur-Milchior:
Previously, add-on 175027074 simply edited the list once. It became impossible
since the list became a function. Hence I need a filter to add the list here.

@kelciour (nice to meet you by the way), you may be interested by it too (at
least if I believe efb1ce46d4b6982833b0ed324b9d105ba83ba394 )

I would have preferred to use `anki.importing.base.Importer` instead of
`Any`. However, this leads to
> Name "anki.importing.base.Importer" is not defined  [name-defined]
when I run test.

Helps to solve this would be welcomed. Otherwise, I believe keeping "Any" is acceptable

### DIFF
--- a/pylib/anki/importing/__init__.py
+++ b/pylib/anki/importing/__init__.py
@@ -3,6 +3,7 @@
 
 from typing import Any, Callable, Sequence, Type, Union
 
+from anki import hooks
 from anki.collection import Collection
 from anki.importing.anki2 import Anki2Importer
 from anki.importing.apkg import AnkiPackageImporter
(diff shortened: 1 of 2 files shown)


dae:
Thanks Arthur. I've added a note about https://github.com/ankitects/anki/issues/1018

Arthur-Milchior:
I beg your pardon but I don't understand what it implies. Does it mean the hook will not be added because you'll rewrite importing anyway? Does it mean that you'll note during the rewrite that I should be noticed?

dae:
"Here is a hook that makes it easier to add importers using the current system, but please be aware that the current system will likely change in the coming months, and this solution may stop working then."

PR #1544, opened 2021-12-09
Fix Alt-Gr and add shortcut to create copy again

RumovZ:
Closes #1541.


dae:
Thanks Rumo, glad we were able to solve this! I found https://github.com/qutebrowser/qutebrowser/issues/5399#issuecomment-658321796 - maybe it ended up getting undone in a later change. What matters is that it's fixed in any case :-)

PR #1545, opened 2021-12-09
add wrap function

git9527:


hgiesel:
I'm not quite happy about this:
1. The `PlainText.surround` is actually unused currently by Anki. It would probably make more sense to just export the whole `CodeMirror.Editor` instance. and remove `PlainText.surround` altogether.

2. Instead of exposing `RichTextInput.surround` we should also consider accessing `richTextInput.element` instead, in order to keep the API surface lower.

git9527:
This PR's purpose is the keep consistency with previous Anki version before Editor Svelte translation, there are some add-on s have dependency on the `wrap` function

dae:
@hgiesel it would be nice to avoid abruptly removing it if we don't have a good reason for doing so. Keeping the API surface low is a good thing, but when code is already in use, we should ideally start with a deprecation warning, pointing existing users to what they should be using instead. If the plain text case is a problem, maybe we could make the routine a no-op unless the rich text area is active?

hgiesel:
I agree it's a good not to just abruptly remove the function.
I'll quickly prepare a PR which effectively will do the same thing (+ a bit cleanup).

PR #1547, opened 2021-12-09
Fix some Mathjax issues

hgiesel:
One issue with the current Mathjax approach is that the <anki-mathjax> element is `contenteditable="true"`, which causes the bug of the Mathjax suddenly being uneditable in some editing situations (it basically becomes an image).

The idea of this PR is to make `anki-mathjax` `contenteditable="false"`. This brings a whole slew of other issues, the biggest one being the placement of the caret around the cE=false element. E.g. it's impossible to place the caret _after_ the cE=false element, if nothing follows it, i.e. the end of the field.

If one wants to see another editor deal with this issue, you can have a peak [how CKEditor4 does it](https://ckeditor.com/docs/ckeditor4/latest/examples/placeholder.html). They basically wrap the cE=false element dynamically in some zero width spaces.

My approach is to have a _frame_ for the element, which basically looks like this:

```html
<anki-frame>
    <frame-handle-start> </frame-handle-start>
    <anki-mathjax ... />
    <frame-handle-end> </frame-handle-start>
</anki-frame>
```

The PR is not quite there yet, but I'll hopefully be able to finish it tomorrow.

### DIFF
--- a/ts/editable/ContentEditable.svelte
+++ b/ts/editable/ContentEditable.svelte
@@ -3,115 +3,46 @@ Copyright: Ankitects Pty Ltd and contributors
 License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 -->
 <script context="module" lang="ts">
-    export interface ContentEditableAPI {
-        flushLocation(): void;
-    }
+    export type { ContentEditableAPI } from "./content-editable";
 </script>
 
 <script lang="ts">
     import type { Writable } from "svelte/store";
     import { updateAllState } from "../components/WithState.svelte";
-    import type { SelectionLocation } from "../domlib/location";
-    import { saveSelection, restoreSelection } from "../domlib/location";
-    import { on, preventDefault } from "../lib/events";
-    import { caretToEnd } from "../lib/dom";
-    import { registerShortcut } from "../lib/shortcuts";
     import iterateActions from "../sveltelib/iterate-actions";
+    import contentEditableAPI, {
+        saveLocation,
+        prepareFocusHandling,
+        preventBuiltinContentEditableShortcuts,
+    } from "./content-editable";
+    import type { ContentEditableAPI } from "./content-editable";
+    import type { MirrorAction } from "../sveltelib/mirror-dom";
+    import type { InputManagerAction } from "../sveltelib/input-manager";
 
-    export let nodes: Writable<DocumentFragment>;
     export let resolve: (editable: HTMLElement) => void;
 
-    export let mirrors: Array<
-        (editable: HTMLElement, params: { store: Writable<DocumentFragment> }) => void
-    >;
-    export let managers: Array<(editable: HTMLElement) => void>;
-    export let api: Partial<ContentEditableAPI>;
+    export let mirrors: MirrorAction[];
+    export let nodes: Writable<DocumentFragment>;
 
     const mirrorAction = iterateActions(mirrors);
+    const mirrorOptions = { store: nodes };
+
+    export let managers: InputManagerAction[];
+
     const managerAction = iterateActions(managers);
 
-    let latestLocation: SelectionLocation | null = null;
+    export let api: Partial<ContentEditableAPI>;
 
-    function onFocus(): void {
-        if (!latestLocation) {
-            caretToEnd(editable);
-            return;
-        }
-
-        try {
-            restoreSelection(editable, latestLocation);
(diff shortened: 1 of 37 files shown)


dae:
Not an area I'm particularly familiar with, but it sounds reasonable. :-)

(minor note: rebasing over the latest main branch prior to pushing will speed up the CI checks, as it won't need to switch to older dependencies to run the tests)

hgiesel:
Features:

- I fixed the issue with getRange error messages mentioned in [this issue](#1516) 
- Allows to move the caret in and out of inline Mathjax. The best way is to put the activate a Mathjax, and then use the arrow keys to navigate.
- When dealing with Block Mathjax, clicking before or after the Mathjax Element can insert a `<br>`. Otherwise it could be impossible to navigate before/after it.

There's still some minor issues I'm aware of:
- [ ] When clicking out of Mathjax, you can not directly move the caret inside of a `<li>` (or generally any block element)
- [ ] In some _reframing_ situations, the "move in" functionality might be broken.

ts/editable/ContentEditable.svelte
```diff
@@ -2,47 +2,71 @@
...
-    export let mirror: (
-        editable: HTMLElement,
-        params: { store: Writable<DocumentFragment> },
-    ) => void;
 
-    export let inputManager: (editable: HTMLElement) => void;
+    export let mirrors: Array<
+        (editable: HTMLElement, params: { store: Writable<DocumentFragment> }) => void
+    >;
+    export let managers: Array<(editable: HTMLElement) => void>;
```
dae, line 27: These could do with docstrings

ts/sveltelib/iterate-actions.ts
```diff
@@ -0,0 +1,37 @@
...
+    destroy?(): void;
+    update?(params: P): void;
+}
+
+type Action<E extends HTMLElement, P> = (
+    element: E,
+    params: P,
+) => ActionReturn<P> | void;
+
+function iterateActions<E extends HTMLElement, P>(
```
dae, line 16: docstring would be nice

dae:
Having the ability to cursor into/out of a MathJax section is great Henrik - it feels really polished! Automatically selecting the text when you cursor in is also a nice touch, that should speed up navigation once people get used to it.

A minor issue I noticed: the add screen starts up without a field focused again.

Being unable to directly click into a block element is not ideal, but doesn't seem to be a major issue. What did you mean by 'reframing situations'?

hgiesel:
> What did you mean by 'reframing situations'?

I mean what happens when you press Backspace in this situation:
Screenshot 2022-01-06 at 21 41 10

In this case, the content of the `div` is _lifted out_, which means that everything is first removed from the document, and then reinserted. This is also when other weird things with cE can happen, like e.g. classes being removed or inlined.

ts/sveltelib/iterate-actions.ts
dae: Thanks, that makes it clearer. The current name makes me think that the actions will be iterated when that function is called, instead of when its return value is used. Would something like `actionList` or `groupedActions` make sense instead?

dae:
Thanks, I get you now. You've done a great job getting this far already - editing is such a complicated area :-(

ts/editable/ContentEditable.svelte
hgiesel: Instead of putting docstrings in ContentEditable, I've put doc strings in all libs/interfaces it uses. 

One problem I have with Svelte is, that in `.svelte` files it's hard to quickly overview all the Svelte-y things (actions, listeners, exports) we use for one specific component, because everything is spaced out by all the logic.

That's why I've moved most of the logic into `content-editable.ts` as the "lib" of that very component.

Other candidates for this would be RichTextInput, and PlainTextInput. Which I would do in another PR, if you're okay with this.

hgiesel:
I've fixed the "reframing" issue. I'm over-all quite happy with the PR now. The "clicking into block elements" issue still persists, but as you said, it's a minor issue and I'll tackle that at another time.

dae:
Thanks for all your work on this Henrik!

ts/editable/ContentEditable.svelte
dae: If the code does not reference things in the .svelte file , then I think putting it in a .ts file is a good idea - it's easier to read/edit that way, and can be tested separately should we wish to.

PR #1549, opened 2021-12-10
Use `save` utils in main window

RumovZ:
Fixes https://forums.ankiweb.net/t/anki-2-1-50-beta/15608/63.


RumovZ:
This fixes it for me. I don't think `saveState` is involved, unless cqg is experiencing a different bug.

dae:
Bugfixes that make the code simpler are always nice :-) Thanks Rumo.

PR #1550, opened 2021-12-10
Reimplement "Show Duplicates" button in editor

kleinerpirat:
https://forums.ankiweb.net/t/anki-2-1-50-beta/15608/66

image

I figured the `LabelContainer` is the better spot for this element, considering it turns red when there are dupes.

### DIFF
--- a/.buildkite/mac/entrypoint
+++ b/.buildkite/mac/entrypoint
@@ -2,16 +2,18 @@
 
 set -e
 
-BAZEL="bazel --output_user_root=~/bazel --output_base=~/bazel/anki"
+STATE=$(pwd)/../state/anki-ci
+mkdir -p $STATE
+BAZEL="bazel --output_user_root=$STATE/bazel --output_base=$STATE/bazel/anki"
 BUILDARGS="--config=ci --experimental_convenience_symlinks=ignore"
 
 echo "+++ Building and testing"
 
 # move existing node_modules into tree
-test -e ~/node_modules && mv ~/node_modules .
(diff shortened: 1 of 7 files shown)


hgiesel:
I like the positioning.

I tried to avoid putting any logic other than "Edit the content of a single field" into `EditorField`. [Here](https://github.com/kleinerpirat/anki/compare/editor-show-dupes...hgiesel:editor-show-dupes)'s a suggestion how we can also avoid this for dupes. Basically I just put it next to the badges :)

kleinerpirat:
Thanks for looking over this! I'm gladly incorporating your approach into this PR.
Personally, I prefer the link to be centered though:

image
image

Would you rather keep it within field-state or got a technical argument that speaks against https://github.com/ankitects/anki/pull/1550/commits/3a4511042da7083a52d67b58550b13d873dcbea5 ?

hgiesel:
I'd rather keep it within field-state (for the same reason I mentioned above). You can achieve the same with some styling on `FieldState`:

```css
flex-grow: 1;
justify-content: flex-end;
```

(And remove the `justify-content-between` class)

And then on the `DuplicateLink`, you set:

```css
text-align: center;
flex-grow: 1;
```

One advantage I'm thinking of: It allows us a bit more freedom, when adapting the editor styling for mobile.

kleinerpirat:
Nice one, didn't think of that! (also had some trouble understanding your argument before :sweat_smile:).

dae:
Looks great, thank you both!

PR #1552, opened 2021-12-10
Parse Markdown inside HTML tags in config.md

abdnh:
My use case for this is to wrap a config.md written in Arabic with a RTL div (with [`markdown="1"`](https://python-markdown.github.io/extensions/md_in_html/#1) set).

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -613,7 +613,7 @@ class AddonManager:
         path = os.path.join(self.addonsFolder(dir), "config.md")
         if os.path.exists(path):
             with open(path, encoding="utf-8") as f:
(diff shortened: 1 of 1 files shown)


dae:
I was worried this might limit our options when moving to a different markdown parser in the future, but it looks like this behaviour (without the extra markdown="1") is part of CommonMark. Thanks Abdo!

https://spec.commonmark.org/0.29/#html-blocks

dae:
Note that you'll need to keep a newline between the markdown and tags, eg:

```
<DIV CLASS="foo" markdown="1">

*Markdown*

</DIV>
```

Without the extra newlines, it will work at the moment but break if we switch to something like mw.col.render_markdown() in the future.

PR #1553, opened 2021-12-10
Add setField to OldEditorAdapter

matthayes:
I developed a plugin called [Cloze Anything](https://ankiweb.net/shared/info/330680661) that supports an alternative JavaScript-based cloze implementation.  One of the requirements the plugin has is the ability to update another field in the editor (besides the one being edited) based on certain conditions.  Previously the plugin was able to use `web.eval` to set individual fields in the editor, ex: 

```
editor.web.eval("""$("#f" + %d).html(%s)""" % (f["ord"], json.dumps(field_content)))
```

This stopped working at some point in recent releases.  As an alternative, I've had to switch to using `setFields`.  However this isn't optimal.  There are two cases where I need to update other fields.  One is for a `blur` event and the other is for a `key` event.  For a `blur` event, using `setFields` works fine (although it's overkill because I only need to update one field).  But for the `key` event, `setFields` doesn't work well because it resets the focus of the field being edited.  What I really need is the ability to set the value of just one field.

Here I've added a `setField` function, based on the implementation I saw for `setFields`.  I tested this new function in my plugin [here](https://github.com/matthayes/anki_cloze_anything/commit/8841ca96a9f6a15dc009839c20706460fadd87aa).

Is there a better way to achieve this besides adding this new function?  I tried searching the Python code for `eval` calls doing something similar and I couldn't find anything.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -88,6 +88,7 @@ Ren Tatsumoto <tatsu@autistici.org>
 lolilolicon <lolilolicon@gmail.com>
 Gesa Stupperich <gesa.stupperich@gmail.com>
 git9527 <github.com/git9527>
+Matthew Hayes <matthew.terence.hayes@gmail.com>
 
 ********************
 
--- a/ts/editor/OldEditorAdapter.svelte
+++ b/ts/editor/OldEditorAdapter.svelte
@@ -115,6 +115,10 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         fieldNames = newFieldNames;
     }
 
+    export function setField(index: number, fieldContent: string): void {
+        fieldStores[index].set(fieldContent);
+    }
+
     let fieldDescriptions: string[] = [];
     export function setDescriptions(fs: string[]): void {
         fieldDescriptions = fs;
--- a/ts/editor/index.ts
+++ b/ts/editor/index.ts
@@ -74,6 +74,7 @@ async function setupNoteEditor(): Promise<NoteEditorAPI> {
     });
 
     Object.assign(globalThis, {
+        setField: noteEditor.setField,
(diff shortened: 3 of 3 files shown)


hgiesel:
You can just use:

```js
const noteEditor = await noteEditorPromise;
noteEditor.fields[0].editingArea.content.set("Test<br>")
```

You can do that inside of an async function.

matthayes:
This doesn't seem to work from `editor.eval`.  Is there an alternative way I should execute that for a plugin?

```
JS error /_anki/legacyPageData?id=5365418736:1 Uncaught SyntaxError: await is only valid in async functions and the top level bodies of modules
```

matthayes:
Ah, so this does seem to work though:

```
cmd = """const noteEditor = noteEditorPromise.then((noteEditor) => noteEditor.fields[5].editingArea.content.set("Test!!!<br>"));"""
editor.web.eval(cmd)
```

Is this how you'd recommend achieving this?  I did see this pattern used in the code, but it was always from a setup-like method so it wasn't clear I could use it for this case.

hgiesel:
Here's another way to do it:

```python
# Define this function in a util file:
# This is what is called an IIAFE: Immediately invoked asynchronous function expression

def async_js(js):
    return f"""(async () => {{
        {js}
    }})();
    """

...
# Use it like this for the editor:

editor.web.eval(async_js("""
   const noteEditor = await noteEditorPromise;
   noteEditor.fields[0].editingArea.content.set("Test<br>");
""")))
```

By waiting on `noteEditorPromise`, you can be sure it executes _after_ the note editor initialized.

matthayes:
Thanks @hgiesel !  This is working pretty great.  This would be a good tip to add to the anki add-on writing guide.

matthayes:
Closing this since there is a working alternative for setting a field.

PR #1556, opened 2021-12-13
Rework field copying when switching notetypes

abdnh:
This PR tweaks the way Anki copies fields when switching notetypes in the Add screen:
- Copy identical fields as before, but maintain a list of copied fields and only add non-empty ones to it.
- Instead of setting remaining fields by their indices, assign remaining non-empty old fields to new fields sequentially

This results in less fields being lost when switching notetypes, while admittedly making the behavior less obvious.

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -148,15 +148,27 @@ class AddCards(QMainWindow):
         if old:
             old_fields = list(old.keys())
             new_fields = list(new.keys())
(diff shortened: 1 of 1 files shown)


dae:
Can anyone see any problems with this approach for their use case? If I don't hear from anyone, I'll merge this in in a few days.

dae:
Thanks Abdo! 🚀

PR #1558, opened 2021-12-14
Preload external css files to prevent flash of unstyled content

hikaru-y:
This implements the functionality of preloading external css files, which is the approach mentioned in the commit message of 46b85d508a0575cacc8013e4a63e994ff40ff851.

This solves an issue that was not fixed by that commit, and a side effect of it:
- Flickering occurs when a css file is referenced for the first time after Anki is started.
- Changes to a css file are not reflected until Anki is restarted.

I tested this PR with the following card templates, and it worked as expected.
- `@import` a large external css file such as Bootstrap in the Styling section, or refer to it with a `<link rel="stylesheet">` in the front/back template.
- `@Import` another css file from within an `@import`ed css file.

### DIFF
--- a/qt/aqt/mediasrv.py
+++ b/qt/aqt/mediasrv.py
@@ -172,8 +172,13 @@ def _handle_local_file_request(request: LocalFileRequest) -> Response:
     try:
         mimetype = _mime_for_path(fullpath)
         if os.path.exists(fullpath):
+            if fullpath.endswith(".css"):
+                # make changes to css files immediately reflected in the webview
+                max_age = 2
+            else:
+                max_age = 60 * 60
             return flask.send_file(
-                fullpath, mimetype=mimetype, conditional=True, max_age=60 * 60  # type: ignore[call-arg]
+                fullpath, mimetype=mimetype, conditional=True, max_age=max_age  # type: ignore[call-arg]
             )
         else:
             print(f"Not found: {path}")
--- a/qt/aqt/mediasrv.py
+++ b/qt/aqt/mediasrv.py
@@ -174,7 +174,7 @@ def _handle_local_file_request(request: LocalFileRequest) -> Response:
         if os.path.exists(fullpath):
             if fullpath.endswith(".css"):
                 # make changes to css files immediately reflected in the webview
-                max_age = 2
+                max_age = 10
             else:
                 max_age = 60 * 60
             return flask.send_file(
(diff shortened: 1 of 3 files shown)


qt/aqt/mediasrv.py
```diff
@@ -172,8 +172,13 @@ def _handle_local_file_request(request: LocalFileRequest) -> Response:
     try:
         mimetype = _mime_for_path(fullpath)
         if os.path.exists(fullpath):
+            if fullpath.endswith(".css"):
+                # make changes to css files immediately reflected in the webview
+                max_age = 2
+            else:
+                max_age = 60 * 60
```
hikaru-y, line 179: #### value of `max_age`
- css
  - I think `1` second should theoretically be enough for the preload to work properly, but for some reason it fails very occasionally, so I set it to `2` seconds to make sure it works.
- non-css
  - I was wondering if it would be better to change it to `None`, but I wasn't sure, so I left it at `60 * 60`.

dae: Perhaps it's worth making it 5 or 10 seconds, just to avoid the chance that it will still happen on users with slow computers?

For non-css, my main concern is possible redrawing when the user has embedded fonts, but it's not something I tested.

ts/reviewer/css.ts
```diff
@@ -0,0 +1,49 @@
...
+        )
+    );
+}
+
+function injectAndLoadCss(css: CssElementType): Promise<void> {
+    return new Promise((resolve) => {
+        css.classList.add(preloadCssClassName);
+
+        // this prevents the css from affecting the page rendering
+        css.media = "print";
```
dae, line 43: Clever!

dae (commented):
Thanks for taking the time to prepare a better fix! LGTM, just a minor comment on the 2 seconds:

qt/aqt/mediasrv.py
hikaru-y: > Perhaps it's worth making it 5 or 10 seconds, just to avoid the chance that it will still happen on users with slow computers?

Done.

> For non-css, my main concern is possible redrawing when the user has embedded fonts, but it's not something I tested.

I see. It's true that if you set `max_age` to `None` and load a font with `@font-face`, the font will be loaded every time and only the characters using that font will be rendered with some delay, which is annoying. It looks like it's best to keep the caching enabled as is.

hikaru-y:
Sorry for commenting on the code in duplicate here. At first I commented using the `Start a review` button, but later I thought I should have used the `Add single comment` button. I thought I canceled the first comment and made a new comment using the `Add single comment` button, but I think I didn't do it right.

dae:
No need to apologise :-) I think it's a GitHub bug, and it happens to me too. Thanks for this Hikaru!

PR #1559, opened 2021-12-15
New TTS/AV tag handling

### LINKED ISSUE #1507: Decide on a syntax for tts outside of field references
Comments/suggestions would be appreciated.

https://forums.ankiweb.net/t/tts-for-multiple-fields-together/15197/2

### LINKED ISSUE #1537: customize 'blank' replacement in TTS
Following on from https://github.com/ankitects/anki/issues/1171

Some users don't like the `blank`, and want to be able to turn it off. It would be nice if we could solve this in a way that doesn't require extra DB lookups on each card render.

One option would be to revert to the old behaviour, then provide a more general solution with a new filter, eg:

```
{{tts en_US:replace [...]=>blank:Field}}
```

Or it could be a specific argument to tts:

```
{{tts en_US blank_cloze=blank:Field}}
```

Another option would be to stuff the option into the notetype (which is cached). We'd need to decide on whether it's a simple boolean, a string of the desired replacement text, or an arbitrary list of replacements.

Thoughts?

RumovZ:
Closes #1507. Closes #1537. All squashed in two commits, sorry.
I'll leave notes on some implementation details. In addition, here are a few remaining points from #1507.

1. > Should we implement error handling like for the search syntax? I'd normally say we can always add it later if necessary. However, without error messages, I wonder how we can reserve certain characters like ". If we just refuse to evaluate a tag containing such characters, the user might never figure out what the problem is.

2. > Or we use a convention, e.g. "only tags starting with a bang can be nested".

     (Does not need to be implemented yet, but probably should be documented in advance.)

3. > It would also be great to have some catchy name for these environments, because a tag is already a concept both in Anki and nom.

### DIFF
--- a/rslib/src/browser_table.rs
+++ b/rslib/src/browser_table.rs
@@ -9,7 +9,7 @@ use strum::{Display, EnumIter, EnumString, IntoEnumIterator};
 use crate::{
     backend_proto as pb,
     card::{CardQueue, CardType},
-    card_rendering::extract_av_tags,
+    card_rendering::prettify_av_tags,
     notetype::{CardTemplate, NotetypeKind},
     prelude::*,
     scheduler::{timespan::time_span, timing::SchedTimingToday},
@@ -271,7 +271,7 @@ impl RenderContext {
                 } => current_text,
             })
             .join("");
-        let question = extract_av_tags(&qnodes_text, true, &col.tr).0;
+        let question = prettify_av_tags(&qnodes_text);
 
         Ok(RenderContext {
             question,
@@ -411,7 +411,7 @@ impl RowContext {
                 } => current_text,
             })
             .join("");
-        let answer = extract_av_tags(&answer, false, &self.tr).0;
+        let answer = prettify_av_tags(&answer);
         html_to_text_line(
             if let Some(stripped) = answer.strip_prefix(&render_context.question) {
                 stripped
--- a/rslib/src/browser_table.rs
+++ b/rslib/src/browser_table.rs
@@ -9,11 +9,12 @@ use strum::{Display, EnumIter, EnumString, IntoEnumIterator};
 use crate::{
     backend_proto as pb,
     card::{CardQueue, CardType},
+    card_rendering::extract_av_tags,
     notetype::{CardTemplate, NotetypeKind},
     prelude::*,
     scheduler::{timespan::time_span, timing::SchedTimingToday},
     template::RenderedNode,
-    text::{extract_av_tags, html_to_text_line},
+    text::html_to_text_line,
 };
 
 #[derive(Debug, PartialEq, Clone, Copy, Display, EnumIter, EnumString)]
@@ -270,7 +271,7 @@ impl RenderContext {
                 } => current_text,
             })
             .join("");
-        let question = extract_av_tags(&qnodes_text, true).0.to_string();
+        let question = extract_av_tags(&qnodes_text, true, &col.tr).0;
 
         Ok(RenderContext {
             question,
@@ -410,7 +411,7 @@ impl RowContext {
                 } => current_text,
             })
             .join("");
-        let answer = extract_av_tags(&answer, false).0;
+        let answer = extract_av_tags(&answer, false, &self.tr).0;
         html_to_text_line(
             if let Some(stripped) = answer.strip_prefix(&render_context.question) {
                 stripped
--- a/rslib/src/browser_table.rs
+++ b/rslib/src/browser_table.rs
@@ -271,7 +271,7 @@ impl RenderContext {
                 } => current_text,
             })
             .join("");
-        let question = prettify_av_tags(&qnodes_text);
+        let question = prettify_av_tags(qnodes_text);
 
         Ok(RenderContext {
             question,
@@ -411,7 +411,7 @@ impl RowContext {
                 } => current_text,
             })
             .join("");
-        let answer = prettify_av_tags(&answer);
+        let answer = prettify_av_tags(answer);
         html_to_text_line(
             if let Some(stripped) = answer.strip_prefix(&render_context.question) {
                 stripped
--- a/rslib/src/card_rendering/parser.rs
+++ b/rslib/src/card_rendering/parser.rs
@@ -12,7 +12,7 @@ use nom::{
     sequence::{delimited, pair, preceded, separated_pair, terminated, tuple},
 };
 
-use super::{CardNodes, Node, OtherTag, Tag, TtsTag};
+use super::{CardNodes, Directive, Node, OtherDirective, TtsDirective};
 
 type IResult<'a, O> = nom::IResult<&'a str, O>;
 
@@ -28,7 +28,7 @@ impl<'a> CardNodes<'a> {
     }
 }
 
-impl<'a> Tag<'a> {
+impl<'a> Directive<'a> {
     fn new(name: &'a str, options: Vec<(&'a str, &'a str)>, content: &'a str) -> Self {
         match name {
             "tts" => {
@@ -50,7 +50,7 @@ impl<'a> Tag<'a> {
                     }
                 }
 
-                Self::Tts(TtsTag {
+                Self::Tts(TtsDirective {
                     content,
                     lang,
                     voices,
@@ -59,7 +59,7 @@ impl<'a> Tag<'a> {
                     options: other_options,
                 })
             }
-            _ => Self::Other(OtherTag {
+            _ => Self::Other(OtherDirective {
                 name,
                 content,
                 options: options.into_iter().collect(),
@@ -149,7 +149,7 @@ fn tag_node(s: &str) -> IResult<Node> {
             pair(opening_parser(tag_name), content_parser(tag_name)),
             closing_parser(tag_name),
         ),
-        |(options, content)| Node::Tag(Tag::new(tag_name, options, content)),
+        |(options, content)| Node::Directive(Directive::new(tag_name, options, content)),
     )(s)
 }
 
@@ -201,7 +201,7 @@ mod test {
         // tags
         assert_parsed_nodes!(
             "[anki:foo]bar[/anki:foo]",
-            Tag(super::Tag::Other(OtherTag {
+            Directive(super::Directive::Other(OtherDirective {
                 name: "foo",
                 content: "bar",
                 options: HashMap::new()
@@ -209,7 +209,7 @@ mod test {
         );
         assert_parsed_nodes!(
             "[anki:foo bar=baz][/anki:foo]",
-            Tag(super::Tag::Other(OtherTag {
+            Directive(super::Directive::Other(OtherDirective {
                 name: "foo",
                 content: "",
                 options: [("bar", "baz")].into_iter().collect(),
@@ -218,7 +218,7 @@ mod test {
         // unquoted white space separates options, "]" terminates
         assert_parsed_nodes!(
             "[anki:foo\na=b\tc=d e=f][/anki:foo]",
-            Tag(super::Tag::Other(OtherTag {
+            Directive(super::Directive::Other(OtherDirective {
                 name: "foo",
                 content: "",
                 options: [("a", "b"), ("c", "d"), ("e", "f")].into_iter().collect(),
@@ -226,7 +226,7 @@ mod test {
         );
         assert_parsed_nodes!(
             "[anki:foo a=\"b \t\n c ]\"][/anki:foo]",
-            Tag(super::Tag::Other(OtherTag {
+            Directive(super::Directive::Other(OtherDirective {
                 name: "foo",
                 content: "",
                 options: [("a", "b \t\n c ]")].into_iter().collect(),
@@ -236,7 +236,7 @@ mod test {
         // tts tags
         assert_parsed_nodes!(
             "[anki:tts lang=jp_JP voices=Alice,Bob speed=0.5 cloze_blank= bar=baz][/anki:tts]",
-            Tag(super::Tag::Tts(TtsTag {
+            Directive(super::Directive::Tts(TtsDirective {
                 content: "",
                 lang: "jp_JP",
                 voices: vec!["Alice", "Bob"],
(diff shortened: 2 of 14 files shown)


rslib/src/browser_table.rs
```diff
@@ -270,7 +271,7 @@ impl RenderContext {
                 } => current_text,
             })
             .join("");
-        let question = extract_av_tags(&qnodes_text, true).0.to_string();
+        let question = extract_av_tags(&qnodes_text, true, &col.tr).0;
```
RumovZ, line 274: We can now easily implement a separate extract method for the browser, that skips the allocations for AvTags and replaces the tags with something more meaningful instead of "[anki:play:q:0]". But what would that look like?

rslib/src/card_rendering/parser.rs
```diff
@@ -0,0 +1,261 @@
...
+        }
+
+        Self(nodes)
+    }
+}
+
+impl<'a> Tag<'a> {
+    fn new(name: &'a str, options: Vec<(&'a str, &'a str)>, content: &'a str) -> Self {
+        match name {
+            "tts" => {
```
RumovZ, line 35: Currently using case-sensitive names.

rslib/src/card_rendering/writer.rs
```diff
@@ -0,0 +1,210 @@
...
+        (extractor.write(self), extractor.tags)
+    }
+}
+
+trait Write {
+    fn write<'iter, 'nodes: 'iter, T>(&mut self, nodes: T) -> String
+    where
+        T: IntoIterator<Item = &'iter Node<'nodes>>,
+    {
+        let mut buf = String::new();
```
RumovZ, line 33: This is probably not ideal. Should we iterate the nodes in advance to find an approximate capacity first?

rslib/src/card_rendering/parser.rs
```diff
@@ -0,0 +1,261 @@
...
+                let mut lang = "";
```
RumovZ, line 36: Currently, lang can be omitted and the default value is "". Should it be mandatory or should we add a default like "en_US"?

rslib/src/card_rendering/writer.rs
```diff
@@ -0,0 +1,210 @@
...
+            value: Some(pb::av_tag::Value::SoundOrVideo(
+                decode_entities(ressource).into(),
+            )),
+        });
+    }
+
+    fn write_tts_tag(&mut self, buf: &mut String, tag: &TtsTag) {
+        self.write_play_tag(buf);
+        self.tags.push(pb::AvTag {
+            value: Some(pb::av_tag::Value::Tts(pb::TtsTag {
```
RumovZ, line 156: "blank_cloze" is currently not reported to the frontend. Should it be inserted into `other_args`?

rslib/src/browser_table.rs
dae: Hmm, that's a bit tricky - if the user has only tts on their front, they'd want to see the text. But if they're using tts to read out a field that's already displayed, it would mean it's shown twice. The former seems a bigger problem than the latter, so unless we want to add some way to customize this, maybe the latter?

rslib/src/card_rendering/parser.rs
dae: Most things Anki exposes are case sensitive, so that's the consistent approach :-)

rslib/src/card_rendering/parser.rs
dae: I can see arguments for both approaches, but think I'm leaning slightly towards making it required, as that way it's consistent for every language. WDYT?

rslib/src/card_rendering/parser.rs
```diff
@@ -0,0 +1,261 @@
...
+    arr: &'arr str,
+) -> impl FnMut(&'s str) -> IResult<&'s str> + 'parser {
+    alt((is_not(arr), success("")))
+}
+
+fn node(s: &str) -> IResult<Node> {
+    alt((text_node, sound_node, tag_node))(s)
+}
+
+/// A sound tag `[sound:ressource]`, where `ressource` is pointing to a sound or video file.
```
dae, line 91: Very minor: there are few instances that should read 'resource' in this and another file

rslib/src/card_rendering/parser.rs
```diff
@@ -0,0 +1,261 @@
...
+        recognize(many1(pair(not(alt((sound_node, tag_node))), anychar))),
+        Node::Text,
+    )(s)
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    macro_rules! assert_parsed_nodes {
```
dae, line 168: Nice :-)

rslib/src/card_rendering/writer.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    fn write_play_tag(&self, buf: &mut String) {
+        write!(buf, "[anki:play:{}:{}]", self.side, self.tags.len()).unwrap();
+    }
+
+    fn transform_tts_content(&self, tag: &TtsTag) -> String {
+        strip_html_for_tts(tag.content).replace(
+            "[...]",
+            tag.blank.unwrap_or(&self.tr.card_templates_blank()),
```
dae, line 138: I'm surprised clippy didn't suggest using unwrap_or_else() here - maybe the reference made it think the call was cheap(er).

rslib/src/card_rendering/writer.rs
dae: Since the replacement is already done at that point, it's probably not too useful. If an add-on wanted to do more elaborate transforms, they'd be better off having the user set the replacement to [...], and then doing the replacement themselves.

dae:
Thanks Rumo, looking great!

> Should we implement error handling like for the search syntax?

If it can be done without too much extra complexity, some basic error(s) that get inlined into the output would be good. What I'd definitely like to avoid is lax rules that are later tightened - we don't want to accept slightly-malformed content now, and then end up breaking it later when checks are added. So better to be too strict than not strict enough.

> I'd normally say we can always add it later if necessary. However, without error messages, I wonder how we can reserve certain characters like ". If we just refuse to evaluate a tag containing such characters, the user might never figure out what the problem is.

Agreed

> Or we use a convention, e.g. "only tags starting with a bang can be nested".

> (Does not need to be implemented yet, but probably should be documented in advance.)

Might be simpler to just focus on the built-in tag case for now, and recommend add-on authors not try to use this for their custom tags just yet (but they can of course use custom options for the built-in tags). That gives us more time to polish this, and allows us to defer the decision on nesting to a later date.

> It would also be great to have some catchy name for these environments, because a tag is already a concept both in Anki and nom.

I can't think of anything better than your "Anki tag" off the top of my head, but suggestions welcome :-)

rslib/src/browser_table.rs
RumovZ: I agree, and content can always be stripped with a browser appearance template.
But how exactly do we represent a sound or tts tag? Maybe like so?


    [sound:my sound.mp3] [anki:tts lang=en_US]A Sentence.[/anki:tts]
->

    🔉my sound.mp3🔉 💬A Sentence.💬

Or is this distracting and we should stick to ascii?
Either way, we could add an option to customise this for all tags:

    [anki:caution lang=en_US browser_wrapper=⚠]A Sentence.[/anki:caution]
->

    ⚠A Sentence.⚠

rslib/src/card_rendering/parser.rs
RumovZ: Not sure. Codewise, mandatory options makes things a bit trickier. Currently, `Tag::new()` is infallible.
Maybe we can come back to this after deciding on an error management approach.

rslib/src/card_rendering/writer.rs
RumovZ: Hmm, `.unwrap_or_else(|| &self.tr.card_templates_blank()),` wouldn't work though due to lifetimes.

RumovZ:
> some basic error(s) that get inlined into the output

Could you provide an example of what that would look like? Here are some possible (overlapping) errors: missing opening/closing tag, unmachting closing tag, nested tag, empty sound tag, missing tts lang option, malformed options.
With most of them, it's not possible to parse as tags, but if just the options are invalid, maybe it would make sense to ignore the malformed options, parse the tag regardless and report the error in some other way, e.g. inlined?

> Might be simpler to just focus on the built-in tag case for now, and recommend add-on authors not try to use this for their custom tags just yet (but they can of course use custom options for the built-in tags). That gives us more time to polish this, and allows us to defer the decision on nesting to a later date.

Definitely. I was just thinking of a note like "A '!' can be prepended to the tag name to indicate that it possibly may be nested in other tags in the future." But that probably doesn't make a lot of sense before we know what nesting might look like and how (if) people would like to use it.
Also, if there's demand in the future, we could expose the parser, so frontend users could call the backend to recursively parse the content of tags.

> I can't think of anything better than your "Anki tag" off the top of my head, but suggestions welcome :-)

All right, here are some random suggestions: mark/marker/marking/predicate/object/item/artifact/block/bloc/target.

rslib/src/browser_table.rs
dae: It's nice and compact, and I don't mind it. There may be some users who don't like it, and may be some who expect to be able to click on the icons. The alternatives I can think of are short ascii (like [S] or wrapping the text in [...]), or just including the content without any special markup like is done with image filenames - I don't have a strong preference for any of the above. I'd resist making it customizable unless there was strong demand for it - we have our hands full already :-)

RumovZ: Then I'll go with the emojis for now and see how it's received.

rslib/src/card_rendering/parser.rs
dae: Sounds good, though the checking doesn't necessarily need to be handled at the parsing stage. With the current code, we require lang already - if the user inputs {{tts :Field}}, they get a popup that shows no voice is available (with a not particularly friendly message)

rslib/src/card_rendering/writer.rs
dae: Ah yep. We could presumably work around it by converting to a string and dropping the reference, but if None is the common case, it's probably better left as is.

dae:
If we can catch unbalanced closing/opening tags, that might cover the nesting-unsupported case for free? I'm not so concerned about the individual options and defaulting seems a valid choice too, but if we decide we want to check the language at that point of the code, then adding checks for the other args isn't much extra work.

Have run out of time, will get back to you on the remaining points tomorrow :-)

rslib/src/card_rendering/parser.rs
RumovZ: I completely forgot about that, so when I said "mandatory", I was thinking of refusing to parse that tag (into a tts tag) if no language was specified. The pop-up makes much more sense!
So the Rust code is fine here if I understood correctly.

RumovZ:
Okay, after contemplating the other discussions, my perspective on error handling has shifted a bit.

1. Mismatched or nested tags will appear on the card, so the problem should be obvious to the user.
2. Missing required options can be reported on the frontend as you pointed out above. (At least `lang`, which is the only one currently.) The same goes for invalid option values.

The comparison with the search syntax is not entirely accurate, because the tag syntax isn't discoverable without consulting any external sources. Nevertheless the option syntax might be a bit tricky, because the user may often copy paste the other parts, but create the option part themself.
So what if we collect any unparsable options in a separate field on the backend, and on the frontend report them to the user in a tooltip like a missing language.

 - `[anki:tts lang=jp_JP garbage]` -> Tooltip: "Unparsable options found: `garbage`. See https://docs.ankiweb.net/tags.html#syntax"

If we want to check invalid options on the backend, we can collect them in yet another field while also falling back to the default value.

 - `[anki:tts lang=jp_JP speed=0]` -> Tooltip: "Invalid options found: `speed=0`" (but using `speed=1`)

dae:
Sorry, I'd planned to review this more carefully today, but ran out of time again :-( Will get back first thing tomorrow

rslib/src/card_rendering/writer.rs
dae: I wouldn't bother - the rewriting case is not a particularly hot path, as IIRC, the most we're going to deal with is a few hundred requests at a time when showing browser rows.

The way a trait has been used here surprised me a bit at first, as each time a new tag type is added, the trait would need to be extended. But I see you've done that to be able to reuse the default methods, and it seems fine.

dae:
Ok, sorry for the delay. I've got benchmarking set up now, which will hopefully be useful in the future when we find ourselves wondering whether a particular approach is worth it or not - will push that in a sec. Collecting directly into a HashMap before converting to a TtsTag felt a little more straightforward, and I thought a performance hit might be worth a slightly simpler approach, but realised we're probably better off leaving it as it is.

Re naming: of the suggestions above, block was the closest. Some more ideas: "directive" or "Anki directive"? "block directive"? "statement"?  

Re tooltips: the current TTS tooltip is ephemeral because the user's machine/environment may alter whether the tts statement works or not, and a pop-up that requires confirmation could be quite disruptive. The issues we're talking about detecting shouldn't depend on the environment, and as users can struggle to read/process popups in time, perhaps that might not be the best approach here.

Maybe inlining the error text would be the simplest approach? For example, instead of rendering the [anki:play:...], the text would instead read something like:

(error in anki:tts directive: lang not set)

We could do a similar thing with speed, but I suspect that's going to be a problem a lot less frequently than a required argument like lang, so maybe defaulting is fine? Not a big deal either way.

RumovZ:
> Collecting directly into a HashMap before converting to a TtsTag felt a little more straightforward, and I thought a performance hit might be worth a slightly simpler approach, but realised we're probably better off leaving it as it is.

I don't mind either way.
Benchmarking is great!

> Re naming: of the suggestions above, block was the closest. Some more ideas: "directive" or "Anki directive"? "block directive"? "statement"?

I like directive and wouldn't mind statement either. I wouldn't use more than one word though, because it's used in compounds itself and "TtsAnkiDirective" sounds clunky in my ears. That's why I originally moved away from "AnkiTag".

> (error in anki:tts directive: lang not set)

Will implement that.

dae:
Ah, I was thinking of the user-facing name - two words would definitely be unwieldy in the code.

RumovZ:
I left the pb names as is for now. Maybe we'll have to modify these messages anyway in the future? There's currently no way to return custom ~tags~ directives.

dae:
Thanks Rumo, works well. Maybe we can put off the renaming until we have a non-audio directive :-)

PR #1561, opened 2021-12-15
Make Hard use current step's interval if it's not the first one

### LINKED ISSUE #1555: The inconsistency of the hard button delay with the scheduler 2.1 specifications (learning cards)
According to the documentation, when you press the hard button you expect to get the average of the first and second steps on the first step and the repetition of the current step on the next steps:

> Hard repeats the current step after the first step, and is the average of Again and Good on the first step. [Anki Manual](https://docs.ankiweb.net/studying.html#learningrelearning-cards)

> Hard will repeat the current learning step (except on the first step, where it is the average of first and second steps). [Anki FAQ](https://faqs.ankiweb.net/hard-good-interval-longer-than-good-easy.html)

However, when I set the following learning steps for my deck "2h 2h 1d 3d 6d", I always get the average of the current and the next steps, for example, 13 hours on step 2 instead of 2 hours, and 2 days instead of 1 day on step 3. This behavior occurs both for Anki 2.1.49 and AnkiDroid 2.15.6.

I checked the code of the scheduler and found out that it always averages the current and the next steps (line 654) which contradicts the specification of the scheduler:

https://github.com/ankitects/anki/blob/eb839f0ba29192c3fa41f4dd0c71712a3b358082/pylib/anki/scheduler/v2.py#L1042-L1050

https://github.com/ankitects/anki/blob/eb839f0ba29192c3fa41f4dd0c71712a3b358082/pylib/anki/scheduler/v2.py#L647-L655

RumovZ:
Closes #1555.
The Rust tests show nicely how the intervals for different buttons remain distinct.

### DIFF
--- a/pylib/anki/scheduler/v2.py
+++ b/pylib/anki/scheduler/v2.py
@@ -645,14 +645,18 @@ limit ?"""
         return int(delay * 60)
 
     def _delayForRepeatingGrade(self, conf: QueueConfig, left: int) -> Any:
-        # halfway between last and next
         delay1 = self._delayForGrade(conf, left)
-        if len(conf["delays"]) > 1:
-            delay2 = self._delayForGrade(conf, left - 1)
-        else:
-            delay2 = delay1 * 2
-        avg = (delay1 + max(delay1, delay2)) // 2
-        return avg
+        # first step?
+        if len(conf["delays"]) == left % 1000:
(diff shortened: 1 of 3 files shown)


dae:
Nice clean fix, and tests included! Thanks Rumo.

PR #1562, opened 2021-12-15
 Support drag & drop in main window to import files

abdnh:
I find myself sometimes dragging files into the main window, expecting Anki to process them.

By the way, does `AnkiWebView.allowDrops` do anything at all? I first thought it's a QT thing, but I couldn't find any documentation about it, so it's probably a leftover or something.

Question: are JS drag & drop events expected to work in the reviewer? I think only `dragover` works currently, but `drop` doesn't. I tested these to make sure my change doesn't affect them.


dae:
I think you're right about allowDrops being legacy code that's no longer being used.

I can't seem to get the drop event from an external window to work in JS either, so it may be (yet another) Qt issue.

I was concerned this might have unintended side-effects like breaking card templates, but it looks like you've limited it to the deck browser 👍 

I'll give the JS experts a few days to chime in, and if I don't hear from anyone, will merge this in. The JS I tried for reference:

```html
<div id=area style="height: 5em; background: blue;"></div>
<script>
function dragenter(e) {
    console.log("enter");
    e.stopPropagation();
    e.preventDefault();
}

function dragover(e) {
    console.log("over");
    e.dataTransfer.dropEffect = 'copy';
    e.stopPropagation();
    e.preventDefault();
}

function drop(e) {
    console.log("drop");
    e.stopPropagation();
    e.preventDefault();

    const dt = e.dataTransfer;
    const files = dt.files;
    console.log(files);
}


const el = document.getElementById("area");
document.documentElement.addEventListener("dragenter", dragenter, false);
document.documentElement.addEventListener("dragover", dragover, false);
document.documentElement.addEventListener("drop", drop, false);
</script>
```

dae:
Thanks Abdo!

dae:
Hi Abdo,

I just tried out importing multiple files at once based on your comment on the forums, and realised it's not going to work correctly. The importing code runs in a background thread, so the import call will return before the import has actually completed, and thus we can end up in a state where we're trying to import multiple files in parallel. This tends to trigger a message about not being able to start a transaction, but could potentially cause bigger problems if the decks have anything in common. So for now I've disabled everything except the first import: https://github.com/ankitects/anki/commit/e49496abaea968882f0f6ccab6bdd1317fd63b93

When the importing code gets moved into Rust, the import will happen in one atomic step, so this should no longer be an issue. So I'd be tempted to leave things this way until then, instead of trying to add some separate 'already importing' flag.

abdnh:
Ah, I only managed to trigger an error with collection packages when working on this, so I excluded them.

PR #1563, opened 2021-12-17
Fix some issues with closing previewer

hikaru-y:
This fixes the following issues with closing the previewer.

- Media playback does not terminate when the previewer is closed.
  https://forums.ankiweb.net/t/anki-2-1-50-beta/15608/78
- `_on_preview_closed` is called twice when the preview button is clicked while the previewer is open.
- https://github.com/ankitects/anki/blob/3e0c9dc86688c531dba95f406239a643d93dc7b6/qt/aqt/browser/browser.py#L633-L635
This seems to be leftover code from before the editor toolbar was migrated to svelte, and causes the following error to be shown in the console when closing the previewer:
`Uncaught TypeError: Cannot read property 'classList' of null`
In this PR, the state of the preview button is toggled by updating the `active` prop of `LabelButton`, instead of toggling `class` directly.


dae:
Thanks for staying on top of these issues Hikaru, it's a big help.

PR #1564, opened 2021-12-18
 Allow customization of add-on config help path

abdnh:
This is useful for loading translated versions of the help file according to the user's language, for example.

I also changed some occurrences of `dir` and `addon` to `module` to be consistent.

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -600,7 +600,7 @@ class AddonManager:
 
     _configButtonActions: dict[str, Callable[[], bool | None]] = {}
     _configUpdatedActions: dict[str, Callable[[Any], None]] = {}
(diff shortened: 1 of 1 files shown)


dae:
Thanks for doing that rename in a separate commit, that makes it easier to review :-)

Instead of setting a string in that dict, what about having a dict[module, Callable[[], str]]? That way the add-on can calculate the desired filename right as it's called, instead of in advance. You could potentially even go one step further and have it return the file contents instead of the filename, which would allow add-ons to construct the documentation programmatically.

abdnh:
> You could potentially even go one step further and have it return the file contents instead of the filename, which would allow add-ons to construct the documentation programmatically.

Done.

dae:
Thanks as always Abdo!

Issue #1536, opened 2021-12-06
Deleting certain elements in the editor leaves <div><br></div> behind


kleinerpirat:
Here are two actions on an empty field that have the same result visually, but not in HTML:
(I removed the `ul` margins for better alignment of the images :)
#### Toggle a list button on, then off
<img src="https://user-images.githubusercontent.com/62722460/144900207-1634fdf4-161c-4e48-b3f9-dd98c000be95.png" width="48%"> <img src="https://user-images.githubusercontent.com/62722460/144903315-0be5626d-61ab-4b51-9cb4-6b44df189f6d.png" width="48%">

#### Insert a list, hit <kbd>Backspace</kbd>
<img src="https://user-images.githubusercontent.com/62722460/144900207-1634fdf4-161c-4e48-b3f9-dd98c000be95.png" width="48%"> <img src="https://user-images.githubusercontent.com/62722460/144903203-bb2c8bfd-b579-44e2-9581-a5d456c9d04c.png" width="48%">

This `<div>` can be removed by pressing <kbd>Backspace</kbd> a second time. But who would do that?
There is no visual indication that there's still a `<div>` inside the field, because it doesn't have any margins (see caret position).

### Why is this problematic?
Repeating the second action multiple times results in nested `<div>`s:

![image](https://user-images.githubusercontent.com/62722460/144905393-9fcb4617-d2cb-4f64-892a-cf3e67e98b70.png)

It's probably not uncommon to rethink the choice to use a list, or for a list to be deleted after it was preserved in a sticky field. Users will mark the list (some will hit <kbd>Ctrl</kbd>+<kbd>A</kbd>) and hit <kbd>Backspace</kbd> until the field is empty and the caret is at the starting position. Most people won't notice that there's still HTML inside the field and continue their work. Decks will (continue to) be cluttered with preventable nested `div` structures (https://forums.ankiweb.net/t/automatic-removal-of-div-div/9071).

---
A margin for `<div>` inside fields would help, but the much better fix would be to prevent `<div><br></div>` from happening in the first place. I don't know the editor code well enough to fix this myself, sadly.

kleinerpirat:
[This forum post](https://forums.ankiweb.net/t/anki-2-1-50-beta/15608/82?u=kleinerpirat) by @nwwt describes a similar issue and makes a good point:
Invisible, left-behind HTML can trigger unwanted conditional card creation.

dae:
[...]

PR #1565, opened 2021-12-21
Prevent unwanted <div> from being left behind when clearing field

hikaru-y:
This is an attempt to address #1536.

### DIFF
--- a/ts/sveltelib/input-manager.ts
+++ b/ts/sveltelib/input-manager.ts
@@ -71,14 +71,21 @@ function getInputManager(): InputManager {
         cancelInsertText();
     }
 
-    function onInput(this: HTMLElement): void {
-        // prevent unwanted <div> from being left behind when clearing the contents
-        // of a field with 'Backspace' or 'Delete'
-        if (/^\n?$/.test(this.innerText)) {
-            const children = this.children;
-            if (children.length === 1 && children.item(0) instanceof HTMLDivElement) {
-                this.innerHTML = "";
-            }
+    function onInput(event: Event): void {
+        if (
+            !(event instanceof InputEvent) ||
+            !(event.currentTarget instanceof HTMLElement)
+        ) {
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


dae:
Thanks Hikaru! @hgiesel what do you think?

ts/sveltelib/input-manager.ts
```diff
@@ -71,6 +71,17 @@ function getInputManager(): InputManager {
         cancelInsertText();
     }
 
+    function onInput(this: HTMLElement): void {
```
hgiesel, line 74: I'm somewhat opposed to using `this` like this, as I find it to be a bit magical.
First it doesn't behave the same in arrow functions, and second it's not really clear which target it is actually binding to (`currentTarget` vs `target`).

hgiesel:
Looks good to me. I've left on comment, but that's more like a preference, so I cannot really make a call :)

I also just noticed that I haven't added any comments to the input manager yet. I think I'll prepare one PR, just with some explanations to the interfaces.

ts/sveltelib/input-manager.ts
hikaru-y: Thanks @hgiesel,
I've changed to use `event.currentTarget` instead of `this`.

ts/sveltelib/input-manager.ts
```diff
@@ -71,6 +71,25 @@ function getInputManager(): InputManager {
         cancelInsertText();
     }
 
+    function onInput(event: Event): void {
+        if (
+            !(event instanceof InputEvent) ||
```
hikaru-y, line 76: Related: https://github.com/microsoft/TypeScript/issues/39925#issuecomment-669550088

hikaru-y:
In addition to the change from `this` to `event.currentTarget`, I realized that it is not efficient to get `innerText` for every input event and do a regex comparison, so I added a check to see if `event.data` is empty.

dae:
Thanks Hikaru!

PR #1566, opened 2021-12-23
Minor changes to graphs

### LINKED ISSUE #1560: a few minor stats changes
These should be fairly easy: https://forums.ankiweb.net/t/minor-ui-improvements-for-stats/15899

Voovs:
Fixes #1560 by adding thousand-comma separation and percent signs to some graphs, for consistency


RumovZ:
Would it be possible to use locale-specific separators? In many countries, `,` is the decimal separator, Japanese uses groups of fours, not threes, and I'm sure there are even more variations I don't know of.
This is not only relevant to this PR. The existing stats already seem to use hard-coded number representations. This is especially confusing as other parts of Anki, like the overview stats also displayed in the stats screen, are correctly localised. So when I use Anki with a German interface, I always have to look twice to know what `.` or `,` in a number means.

grafik

abdnh:
I think it's possible to use [`Number.toLocaleString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString), passing it the configured locales in [`i18n/utils.ts`](https://github.com/ankitects/anki/blob/21fde1b59ee53de5384d1ff0eb8d9b190d85549a/ts/lib/i18n/utils.ts#L42)

Voovs:
I've also made a few trivial UI changes on the stats page

shrink_screenshot_out

ts/graphs/TableData.svelte
```diff
@@ -29,6 +30,10 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         text-align: end;
     }
 
+    .space {
+        padding-left: 5px;
+    }
+
```
dae, line 36: Defining an extra column for spacing feels like a bit of a hack - I imagine this could probably be accomplished by just styling the table or td tags instead?

ts/graphs/CardCounts.svelte
```diff
@@ -111,6 +111,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     .right {
         text-align: right;
+        padding-left: 40px;
```
dae, line 114: Maybe this could be more cleanly done by styling the table instead of a particular class - the CSS gurus here may have a suggestion for you.

One issue we need to consider is how this will appear or a narrow Phone screen. Here's what it looks like after this change:
Screen Shot 2021-12-24 at 9 38 37 am

And before (which was not ideal either):

Screen Shot 2021-12-24 at 9 39 39 am

Bearing in mind that some languages can be up to 2x longer than English, I think we probably need to reduce the padding from 40 to something smaller, or use styling that is dependent on the width of the screen.

ts/graphs/buttons.ts
```diff
@@ -134,6 +134,20 @@ export function renderButtons(
         return { total, correct, percent };
     };
 
+    const totalPressedStr = (data: Datum): string => {
+        const groupTotal = totalCorrect(data.group).total;
+        const buttonTotal = data.count;
+
+        const buttonTotalCommaSeparated = buttonTotal
+            .toString()
+            .replace(/\B(?=(\d{3})+(?!\d))/g, ",");
```
dae, line 143: Should this be using toLocaleNumber?

ts/lib/i18n/utils.ts
```diff
@@ -48,6 +48,10 @@ export function toLocaleString(
     return date.toLocaleDateString(langs, options);
 }
 
+export function toLocaleNumber(n: number): string {
```
dae, line 51: I recommend we call this "localizedNumber", and rename "toLocaleString" to "localizedDate"

dae (commented):
Thanks for working on this Vova! Mostly looks good; I've left some comments below.

ts/graphs/CardCounts.svelte
Voovs: `./script/ts-run` was very helpful, thanks! Is there a way to also have it hot-reload, so we don't have to restart anki every time?

Using `padding-left: 4vw` seems to work pretty nicely from desktop to iPhone SE (attached). Ends up like an average of the two above, which is probably the best we can get with this layout

shrink_screenshot_out

Alternatively, we'd probably achieve something much better by just switching to `flex-direction: column` on narrow devices. Something like this, though this mockup is a bit ideal

shrink_screenshot_out

dae: The spacing there looks good! Switching to column might be a better long term solution, but no need to tackle that in this PR.

Running scripts/ts-watch in a separate terminal will automatically rebuild on file change, then you can just refresh the page in Chrome

ts/graphs/buttons.ts
```diff
@@ -134,6 +135,18 @@ export function renderButtons(
...
     };
 
+    const totalPressedStr = (data: Datum): string => {
+        const groupTotal = totalCorrect(data.group).total;
+        const buttonTotal = data.count;
+
+        const buttonTotalCommaSeparated = localizedNumber(buttonTotal);
+        const percent = buttonTotal
+            ? ((buttonTotal / groupTotal) * 100).toFixed(2)
+            : "0";
```
dae, line 145: We probably need to localize the percent number as well, as otherwise the thousands separator doesn't match. You can test with `scripts/ts-run -l de`. There are other areas that are inconsistent and need updating as well, but that belongs in a separate PR (if you're interested please let us know; if not it may be something @RumovZ might like to tackle)

ts/graphs/CardCounts.svelte
dae: The automatic repositioning seems to work well; thanks for that.

ts/graphs/buttons.ts
Voovs: I wouldn't mind going for a hunt for these. Seems like a good way to explore the codebase

Voovs:
Is the "correct" statistic for the Answer Buttons graph helpful? It's already in the subtitle of the graph, though it's honestly kind of confusing without context as to what 93.01% means. It makes more sense in the tooltip, however we now have the percent counts for each individual button, so the correct count is easy to calculate: (100% - "button 1 %"). That's (100% - 6.99% = 93.01%) in the example below

Window

It might be nicer to put the total presses for that card type in the subtitle. Then the last line of the tooltip could be removed, or restate the same information with "Group total:" for context:

shrink_screenshot_out
shrink_screenshot_out

dae:
The total pass rate ("retention rate") is a useful metric for judging how efficient the algorithm is, and is used as a reference when tweaking the interval multiplier - it's one of the main motivations behind that graph. While the user could extrapolate from just the Again percentage in a tooltip, that makes life a bit more complicated, and I suspect some users would prefer the percentage to remain immediately visible.

ts/graphs/buttons.ts
```diff
@@ -140,9 +140,7 @@ export function renderButtons(
         const buttonTotal = data.count;
 
         const buttonTotalCommaSeparated = localizedNumber(buttonTotal);
-        const percent = buttonTotal
-            ? ((buttonTotal / groupTotal) * 100).toFixed(2)
-            : "0";
+        const percent = localizedNumber((buttonTotal / groupTotal) * 100);
```
dae, line 143: Removing the check here causes this to happen:
Screen Shot 2021-12-28 at 10 47 25 am

Otherwise looks good!

Voovs:
Should be good to go. Can't spot any other bugs with my collection, though it'd be nice if someone who reviews >1000 cards per day could double check

dae:
Thanks for all your work on this! 🚀

PR #1569, opened 2021-12-30
Add Create Copy to reviewer and use current card's deck

RumovZ:
https://forums.ankiweb.net/t/anki-2-1-50-beta/15608/103


dae:
Thanks as always Rumo :-)

PR #1573, opened 2022-01-03
Increment sched.reps for preview cards in V2

abdnh:
This is to make the timebox dialog report the correct count. The V3 scheduler already does that.

Minor question: I'm curious where `sched.reps` is actually reset to 0 between review sessions?

### DIFF
--- a/pylib/anki/scheduler/v2.py
+++ b/pylib/anki/scheduler/v2.py
@@ -462,11 +462,12 @@ limit ?"""
         card.flush()
 
     def _answerCard(self, card: Card, ease: int) -> None:
+        self.reps += 1
+
         if self._previewingCard(card):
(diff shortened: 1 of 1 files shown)


dae:
Thanks Abdo! IIRC this code is just used for the timeboxing, and as it records the current number at the start of a session, reps doesn't need to be reset .

abdnh:
> Thanks Abdo! IIRC this code is just used for the timeboxing, and as it records the current number at the start of a session, reps doesn't need to be reset .

Yes, I also thought of that and I remember testing it in the debug console and found the count is actually reset, but I probably just made a mistake.

dae:
It's possible - it's been a long time since that code was originally written :-)

PR #1574, opened 2022-01-03
Guarantee execution order of <script> elements

hikaru-y:
https://forums.ankiweb.net/t/anki-2-1-50-beta/15608/119
I tried to find a simple way to guarantee the execution order of dynamically inserted external scripts, but couldn't find one, so I implemented the method you mentioned in the post above:
> making the function asynchronous, and awaiting the loading of any remote scripts before continuing the loop.

### Disadvantage of this PR
If a template contains multiple external js files, it will not move on to loading the next file until the first file has been loaded, parsed, and executed, which results in a slight performance loss compared to the current code.  I'm currently working on another PR that includes the feature to preload external js files to improve performance.

### DIFF
--- a/ts/reviewer/index.ts
+++ b/ts/reviewer/index.ts
@@ -59,7 +59,16 @@ export function _queueAction(action: Callback): void {
     _updatingQueue = _updatingQueue.then(action);
 }
 
-function setInnerHTML(element: Element, html: string): void {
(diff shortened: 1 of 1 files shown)


dae:
Thanks Hikaru! Both this PR, and future caching sound great.

PR #1576, opened 2022-01-04
Fix padding and alignment inside revlog table

RumovZ:
This addresses https://github.com/ankitects/anki/issues/1516#issuecomment-1004420601 by using nested tables to have both, evenly padded columns and aligned units. Some remarks:

- The first column is still left-aligned to ensure alignment with the other stats above. It can be seen as a kind of row header, so I think the additional right padding is tolerable.
- The sub-tables would wrap independently, so wrapping has to be disabled. However, I'd argue that this is for the best anyway. The content is pretty limited for all cells, regardless of localisation, and if a row doesn't fit the window, it breaks at some odd positions, like between a value and its unit. That makes it extremely hard to read, so it's better to have a horizontal scrollbar in my opinion.

### DIFF
--- a/ts/card-info/Revlog.svelte
+++ b/ts/card-info/Revlog.svelte
@@ -77,111 +77,97 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 </script>
 
 {#if revlog.length > 0}
-    <table class="revlog-table">
-        <tr>
-            <th class="left">{tr2.cardStatsReviewLogDate()}</th>
-            <th class="center hidden-xs">{tr2.cardStatsReviewLogType()}</th>
-            <th class="center">{tr2.cardStatsReviewLogRating()}</th>
-            <th class="center">{tr2.cardStatsInterval()}</th>
-            <th class="center hidden-xs">{tr2.cardStatsEase()}</th>
-            <th class="center">{tr2.cardStatsReviewLogTimeTaken()}</th>
-        </tr>
-        <tr>
-            <td
-                ><div class="column">
-                    {#each revlogRows as row, _index}
-                        <div><b>{row.date}</b> @ {row.time}</div>
-                    {/each}
-                </div></td
-            >
-            <td class="hidden-xs"
-                ><div class="centered-cell">
-                    <div class="column">
-                        {#each revlogRows as row, _index}
-                            <div class={row.reviewKindClass}>
-                                {row.reviewKind}
-                            </div>
-                        {/each}
+    <div class="revlog-table">
+        <div class="column">
+            <div class="column-head">{tr2.cardStatsReviewLogDate()}</div>
+            <div class="column-content">
+                {#each revlogRows as row, _index}
+                    <div><b>{row.date}</b> @ {row.time}</div>
+                {/each}
+            </div>
+        </div>
+        <div class="column hidden-xs">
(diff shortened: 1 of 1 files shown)


RumovZ:
We could also split units and values into separate columns for better alignment. Here's a quick demo (for the interval column):

grafik

Compared to the status quo:

grafik

We could even split off the decimal part to have all numbers aligned.

<hr>

I don't know how to emulate a narrow screen device, but I hope it works as before with b29f743 now.

dae:
Looks nice on a desktop, and seems to work correctly on small displays too - great!

I wonder if the nested tables are strictly necessary, or whether this could be accomplished with some internal divs instead? Nested tables brings back memories of the earlier days of HTML :-) If the solution is not immediately obvious, and none of the HTML gurus chime in over the next few days, I'm fine with merging it in as is though.

The demo feels more readable, but also slightly odd - could just be that I'm not used to it. No strong feelings here; would be interesting to see what others think.

RumovZ:
Had to brush up on my flexbox skills, but it's done.
Would be interested to hear what others think about separate columns for units as well. @kleinerpirat @hgiesel 🙂

RumovZ:
I ended up replacing the outer table as well. Padding is now applied uniformly everywhere and the special casing for the first column could be dropped, because the header can now be centered ignoring the whitespace on the right.

dae:
Thanks Rumo, looks good.

ts/card-info/Revlog.svelte
```diff
@@ -77,47 +77,97 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-        text-align: start;
+<style lang="scss">
+    .revlog-table {
+        width: 100%;
+        max-width: 50em;
+        display: flex;
+        flex-direction: row;
+        justify-content: space-between;
+        align-items: center;
+        gap: 0.5em;
```
hgiesel, line 142: @dae Wasn't `flex-gap` still unsupported on some Anki versions?

dae: Good catch, I missed that.

@RumovZ gap is unavailable on our Qt 5.14/5.15 builds, and on older iOS versions:

https://wiki.qt.io/QtWebEngine/ChromiumVersions
https://caniuse.com/flexbox-gap

Here's how a resized window appears on Qt 6:

Screen Shot 2022-01-08 at 9 40 31 am

And on Qt 5:

Screen Shot 2022-01-08 at 9 39 29 am

In this case though, I'm not sure it's a big deal - on the desktop, if the user increases the window width a bit, the text will space out, and on mobile we're hiding some of the columns, so the text only gets bunched-up on the narrowest of screens, on devices using old iOS versions. So we can probably get away with it here. :-)

RumovZ: Thanks for the info. It's not a big deal to use something else instead, as the new approach is extremely flexible, so I've pushed a fix. 🙂

RumovZ:
Ready to merge from my side. The others don't seem to have a strong opinion on aligning numbers and units separately, so I think maybe we should skip that for now. It would involve using a regex to split the localised strings, and without a mono-spaced font, numbers will never line up perfectly anyway.
I'll keep it in mind, maybe a less messy method will present itself in the future.

dae:
Thanks Rumo! 🚀

PR #1577, opened 2022-01-05
Fix new preview card's position being interpreted as a date

abdnh:
Can be reproduced by opening the Card Info screen of a new preview card not answered yet.

Screenshot 2022-01-05 141305

Note: I'm not sure if the `card.original_due != 0` part here is necessary.

### DIFF
--- a/rslib/src/stats/card.rs
+++ b/rslib/src/stats/card.rs
@@ -71,7 +71,7 @@ impl Collection {
             ),
             CardQueue::Review | CardQueue::DayLearn => (
                 {
-                    if card.ctype == CardType::New && card.original_due != 0 {
+                    if card.ctype == CardType::New {
                         // new preview card not answered yet
                         None
                     } else {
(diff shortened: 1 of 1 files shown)


rslib/src/stats/card.rs
```diff
@@ -66,10 +71,15 @@ impl Collection {
             ),
             CardQueue::Review | CardQueue::DayLearn => (
                 {
-                    let days_remaining = due - (self.timing_today()?.days_elapsed as i32);
-                    let mut due = TimestampSecs::now();
-                    due.0 += (days_remaining as i64) * 86_400;
-                    Some(pb::generic::Int64 { val: due.0 })
+                    if card.ctype == CardType::New && card.original_due != 0 {
```
dae, line 74:
```suggestion
                    if card.ctype == CardType::New {
```

I think you're right that it's not necessary - it shouldn't do any harm, but might as well remove it.

dae:
Thanks as always Abdo :-)

PR #1581, opened 2022-01-08
Split/Merge editor.py for its three use cases

hgiesel:
My starting motivation here is: I want to put more UI elements surrounding the Svelte Editor into the web view (e.g. the notetype/deck chooser).

However going further than we have currently means porting elements that are not shared among the three instances (AddCards / EditCurrent / BrowserEditor). I'd like to avoid introducing something like `addMode` from editor.py into the Svelte editor, as I think it will make things more complicated in the long run. Instead I'd rather have three new modules in `/ts` for each of them, which each use `/ts/editor` inside of them.

In order for this to work, however, I'll need to revamp editor.py quite a bit. I'm not quite sure yet what the end result will be, but my guess at this point is that I'll end up with three subclasses of `AnkiWebView`.


dae:
In isolation, I don't have strong feelings about the architecture here - if you feel separate driver modules make more sense than some internal variables, that works for me. We do need to be mindful of the amount of churn we're creating for add-on authors though - this change sounds like it might break a bunch of code that they may have only just updated to work with the editor changes in 2.1.50. Are you thinking this would go into the next major update after 2.1.50, or did you have some other timing in mind? 

(As an aside, is removing objects/iframes working around an issue? What effect will this change have on users who already have iframes or object references in their fields?)

hgiesel:
Okay, I kept the changes to editor.py minimal. I probably overthought it a bit.
Breaking changes would only occur, if somebody copy/pasted the code from `setupWeb` I assume, as now there's no editor.js anymore.

What do you think?

> As an aside, is removing objects/iframes working around an issue? What effect will this change have on users who already have iframes or object references in their fields?

I added this after stumbling upon the following line:

https://github.com/ankitects/anki/blob/4360fd16fb20b647ca2597d38129024d952f996e/qt/aqt/editor.py#L818

It seems like Anki used to remove before.

hgiesel:
Not to unnecessarily delaying the PR, I removed the lines concerning `object` and `iframe`. Also, because this is specifically applied during pre-paste. I wonder whether you know/remember why that was added.

dae:
> I added this after stumbling upon the following line:

Maybe I misunderstood what the code above was doing - is it only affecting paste handling, or does it also apply when people edit fields which already contains such tags? The former is probably still fine, but if we did the latter, it would break cases like https://forums.ankiweb.net/t/anki-2-1-50-beta/15608/129

I seem to recall that stripping iframes/objects was due to concerns that they wouldn't work correctly - either because QtWebKit wouldn't support them correctly, or because they'd cause issues when using the editor. QtWebKit was rather fragile - I seem to recall it being possible to crash it at various points when feeding it kana tags, audio tags, etc.

(will follow with a code review)

qt/aqt/editor.py
```diff
@@ -91,13 +91,20 @@ class Editor:
...
+        addMode: bool = False,
+        *,
+        browserMode: bool = False,
     ) -> None:
         self.mw = mw
         self.widget = widget
         self.parentWindow = parentWindow
         self.note: Note | None = None
         self.addMode = addMode
+        self.browserMode = browserMode
```
dae, line 107: Assuming you're not planning to support add=True when browse=True, this might be better done with an enum - how about something like this?

```diff
diff --git a/qt/aqt/browser/browser.py b/qt/aqt/browser/browser.py
index 4469f2663..cf7f28275 100644
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -18,7 +18,7 @@ from anki.notes import NoteId
 from anki.tags import MARKED_TAG
 from anki.utils import is_mac
 from aqt import AnkiQt, gui_hooks
-from aqt.editor import Editor
+from aqt.editor import Editor, EditorMode
 from aqt.exporting import ExportDialog
 from aqt.operations.card import set_card_deck, set_card_flag
 from aqt.operations.collection import redo, undo
@@ -419,7 +419,7 @@ class Browser(QMainWindow):
 
         gui_hooks.editor_did_init.append(add_preview_button)
         self.editor = aqt.editor.Editor(
-            self.mw, self.form.fieldsArea, self, browserMode=True
+            self.mw, self.form.fieldsArea, self, mode=EditorMode.BROWSE
         )
         gui_hooks.editor_did_init.remove(add_preview_button)
 
diff --git a/qt/aqt/editor.py b/qt/aqt/editor.py
index 4d1d7dba5..7ffe3c20b 100644
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -13,6 +13,7 @@ import urllib.error
 import urllib.parse
 import urllib.request
 import warnings
+from enum import Enum
 from random import randrange
 from typing import Any, Callable, Match, cast
 
@@ -80,6 +81,12 @@ audio = (
 )
 
 
+class EditorMode(Enum):
+    ADD = 0
+    EDIT_CURRENT = 1
+    BROWSE = 2
+
+
 class Editor:
     """The screen that embeds an editing widget should listen for changes via
     the `operation_did_execute` hook, and call set_note() when the editor needs
@@ -95,16 +102,19 @@ class Editor:
         mw: AnkiQt,
         widget: QWidget,
         parentWindow: QWidget,
-        addMode: bool = False,
+        addMode: bool | None = None,  # legacy
         *,
-        browserMode: bool = False,
+        mode: EditorMode = EditorMode.EDIT_CURRENT,
     ) -> None:
         self.mw = mw
         self.widget = widget
         self.parentWindow = parentWindow
         self.note: Note | None = None
-        self.addMode = addMode
-        self.browserMode = browserMode
+        # legacy argument provided?
+        if addMode is not None:
+            mode = EditorMode.ADD if addMode else EditorMode.EDIT_CURRENT
+        self.addMode = mode is EditorMode.ADD
+        self.mode = mode
         self.currentField: int | None = None
         # Similar to currentField, but not set to None on a blur. May be
         # outside the bounds of the current notetype.
@@ -131,9 +141,9 @@ class Editor:
         self.web.set_bridge_command(self.onBridgeCmd, self)
         self.outerLayout.addWidget(self.web, 1)
 
-        if self.addMode:
+        if self.mode == EditorMode.ADD:
             file = "note_creator"
-        elif self.browserMode:
+        elif self.mode == EditorMode.BROWSE:
             file = "browser_editor"
         else:
             file = "reviewer_editor"
```

dae:
> Okay, I kept the changes to editor.py minimal. 

This looks considerably less likely to cause problems than I'd originally feared, so presumably we can slip this part into 2.1.50 if that makes things easier. The notetype/deck selection work is a larger task which is going to be more disruptive though, so I presume we'll want to put off merging that until after 2.1.50?

qt/aqt/editor.py
hgiesel: This looks better indeed.

ts/browser-editor/tsconfig.json
```diff
@@ -0,0 +1,11 @@
...
+    "extends": "../tsconfig.json",
+    "include": ["*"],
+    "references": [
+        { "path": "../lib" },
+        { "path": "../domlib" },
+        { "path": "../sveltelib" },
+        { "path": "../components" },
+        { "path": "../editor" }
+    ]
+}
```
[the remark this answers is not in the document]
hgiesel: Some thoughts:
1. Having more separate modules means we can move out logic from editor.ts. One small example: the logic for `StickyBadge` is only needed for the editor in the AddCards window. However all other editor use cases will also bundle and load it whenever it is invoked. `StickyBadge` is small, so it doesn't matter that much, but they will only get more from here on out: One is the notetype/deck buttons, another is the [card preview](https://github.com/ankitects/anki/pull/969) I've been meaning to implement for a very long time now, which could open in a split pane, but won't be needed in EditCurrent. 
2. I don't intend to increase the amount of modules forever, in fact at some point we could merge modules together: Once I manage to open a previewer in a split pane from the browser and the AddCards editor, we could also open the EditCurrent editor in a split pane from the reviewer, which means that `reviewer-editor` could be merged with the `reviewer` module (hence the name). The same is true for `browser-editor` and a hypothetical `browser` ts module (even though I don't intend to tackle this for a very long time).
3. Is there any way to parallelize module compilation?
4. Would it make sense to try and make `ts/lib` smaller? So that any change in `ts/lib` is guaranteed to actually affect every module. E.g. I'm thinking of `lib/keys` and `lib/shortcuts` here, which are only used in the editor and could be factored out into a `ts/editinglib`.

> do we need esbuild() in ts/editable?

I'm abusing esbuild here to generate the `.css` file, which is included from within editor [here](https://github.com/ankitects/anki/blob/4e4d12a62a261ea90afba88d761a2aad50baf12b/ts/editor/RichTextStyles.svelte#L49). It is kind of a hack, but it's the only way I know to combine the CSS from the Svelte components and the css files together.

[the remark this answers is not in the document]
dae: Will reply below

dae:
> Having more separate modules means we can move out logic from editor.ts. One small example: the logic for StickyBadge is only needed for the editor in the AddCards window. However all other editor use cases will also bundle and load it whenever it is invoked. 

I wonder if we actually need separate modules to accomplish that, or whether we could just use a single typescript module with multiple entrypoints? For example, if we had editor/adding/index.ts or similar import StickyBadge, and the other browser/editcurrent entrypoints did not import it, then presumably it would be omitted from the latter two by esbuild's dead code elimination.

Each extra module we add requires another invocation of the typescript compiler (twice if it's a separate Svelte module as well). Each time the TypeScript compiler is started up, it needs to load all the typescript compiler code before it even starts processing our code. That initial startup is fairly expensive. The older TypeScript worker approach rules_nodejs had was faster since it avoided the repetitive startup delay, but it tended to be a lot flakier.

Our build process can do multiple things at once, but the problem here is that each of our modules in that chain depends on the previous one. It can't start working on `components` until `sveltelib` has been compiled for example.

You can see the critical path visualised in this graph:

Screen Shot 2022-01-12 at 8 57 30 am

The critical path is showing the steps that must complete before the build can proceed. We build a TS lib, then the Svelte lib that depends on it, then the next TS libs that depends on the first, and so on. 

Since bundling only needs to happen for the leaf nodes, it can be better parallelized - most of the bundling is not in the critical path:

Screen Shot 2022-01-12 at 9 00 11 am

(interestingly, that graph also reveals that on macOS at least, the bundling step suffers from a lot of overhead from Bazel's sandboxing)

> Would it make sense to try and make ts/lib smaller?

I previously split the protobuf and i18n code into separate modules, but then realised we were worse off that way - most of the code uses them both anyway, and each extra module has that extra overhead. For more obscure code that is only used by a few other modules, it's not quite so clear cut. If that code didn't depend on anything in ts/lib, then it could be compiled in parallel and won't delay the build. But there are only a finite number of CPU cores available, and we need to consider whether the overhead of launching tsc again is really worth what we'd gain there.

> I'm abusing esbuild here to generate the .css file, which is included from within editor here. It's a hack, but it's the only way I know to combine the CSS from the Svelte components and the css files together.

 Gotcha. No problem keeping it in if it's serving a purpose, I was just worried it was an unneeded step.

dae:
And with bazel analyze-profile:

```
Critical path (19.655 s):
       Time Percentage   Description
    18.0 ms    0.09%   action 'Copying files'
    1.582 s    8.05%   action 'Compiling TypeScript project //ts/lib lib [tsc -p ts/lib/tsconfig.json]'
    1.226 s    6.24%   action 'Compiling TypeScript project //ts/sveltelib sveltelib [tsc -p ts/sveltelib/tsconfig.json]'
    2.443 s   12.43%   action 'Compiling Svelte ts/components svelte'
    1.124 s    5.72%   action 'Compiling TypeScript project //ts/components components [tsc -p ts/components/tsconfig.json]'
    2.337 s   11.89%   action 'Compiling Svelte ts/editable svelte'
    1.630 s    8.29%   action 'Compiling TypeScript project //ts/editable editable_ts [tsc -p ts/editable/tsconfig.json]'
    3.064 s   15.59%   action 'Compiling Svelte ts/editor svelte'
    1.274 s    6.48%   action 'Compiling TypeScript project //ts/editor editor [tsc -p ts/editor/tsconfig.json]'
    1.845 s    9.39%   action 'Compiling Svelte ts/note-creator svelte'
    1.390 s    7.07%   action 'Compiling TypeScript project //ts/note-creator note_creator_ts [tsc -p ts/note-creator/tsconfig.json]'
    1.718 s    8.74%   action 'Bundling Javascript ts/note-creator/index_wrapper.ts [esbuild]'
    0.07 ms    0.00%   action 'Copying files'
    3.20 ms    0.02%   runfiles for //qt runanki
```

More info if you're interested: https://docs.bazel.build/versions/main/skylark/performance.html

hgiesel:
Thanks for all the explanations.
However doesn't that imply though that if we just had one big `/ts/src` directory with one `/ts/src/BUILD.bazel`, we could boil down the compilation down by probably more than half? I might try to investigate that tomorrow...

dae:
It is a tradeoff. For a change to the start of the graph (eg modifying ts/lib), the fewer separate modules there are, the faster it's likely to be. But one giant module means that everything needs to be recompiled when a change is made, so changes to files at the other end of the graph (eg note-creator) would mean everything else needed to be recompiled as well, where currently it doesn't need to be. I suspect a balanced approach might work best - currently the above is a chain of 6; maybe we should be aiming for a chain closer to 3 (eg core libs, libs that depend on core, leaf components). But that's a gut feeling; I have not benchmarked it.

dae:
(change looks good and shaves a few seconds off the build - thanks!)

qt/aqt/editor.py
```diff
@@ -91,13 +98,23 @@ class Editor:
...
         self.mw = mw
         self.widget = widget
         self.parentWindow = parentWindow
         self.note: Note | None = None
-        self.addMode = addMode
+        # legacy argument provided?
+        if addMode is not None:
+            editorMode = EditorMode.ADD_CARDS if addMode else EditorMode.EDIT_CURRENT
+        self.addMode = editorMode is EditorMode.ADD_CARDS
+        self.editorMode = editorMode
```
dae, line 117: Didn't notice this before merging, but ideally this would be snake_case, not camelCase. We can't fix our old code at the moment, but should probably try to use Python convention in the new code we introduce, when we're not referencing the Qt API.

PR #1582, opened 2022-01-08
Change X-GNOME-SingleWindow key to SingleMainWindow

FHEK789:
X-GNOME-SingleWindow was upstreamed to be an XDG spec with the name
"SingleMainWindow" in
https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/53

### DIFF
--- a/qt/package/lin/anki.desktop
+++ b/qt/package/lin/anki.desktop
@@ -10,4 +10,6 @@ Terminal=false
 Type=Application
 Version=1.0
 MimeType=application/x-apkg;application/x-anki;application/x-ankiaddon;
+#should be removed eventually as it was upstreamed as to be an XDG specification called SingleMainWindow
 X-GNOME-SingleWindow=true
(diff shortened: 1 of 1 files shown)


dae:
This appears to have only just been accepted in the standard, and the Gnome/KDE versions in people's distros presumably don't support it yet. Wouldn't it make more sense to wait until this is supported in distros, or provide both old and new keys at the same time?

FHEK789:
Added both the new and old keys and a comment for clarity :)

dae:
Thanks!

The spec seems to indicate comments should be on their own line:

https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#comments

I probably wouldn't bother with the comment - it will likely be quite a few years before we can think of removing it.

FHEK789:
The comment is not on it's own line

dae:
Looks good - thanks for contributing!

PR #1583, opened 2022-01-09
Webview margins

RumovZ:
Addresses two small issues in  #1516. I got the feeling that margins behave differently on Macs, so I hope this works out on your machine as well.


dae:
When the screen is wide enough, this PR looks better than what we had before, as the sides match the top margin:
Screen Shot 2022-01-10 at 12 57 45 pm

Unfortunately it results in the text on mobile scrolling off the page:
Screen Shot 2022-01-10 at 1 02 07 pm

Where it didn't previously:

Screen Shot 2022-01-10 at 1 02 52 pm

On such narrow screens, the smaller margin is probably a better choice. Maybe we could apply separate styling on mobile?

@hgiesel any thoughts?

RumovZ:
> Maybe we could apply separate styling on mobile?

I think it does look better on mobile, too, but of course, fitting the available space is more important, and it's not as much of a problem here, because there is no crowded backdrop, instead it's always shown in fullscreen.
Alternatively, what if we hide the time of day? With or without margins, it wouldn't be such a balancing act that way.

hgiesel:
I remember trying it on my iPhone when I set those margins, and making sure that the lines fit.

Actually, the only that sticks out to me are the non-monospace dates...

RumovZ:
> I remember trying it on my iPhone when I set those margins, and making sure that the lines fit.

Did you try for other languages, too? The width of the table seems to be directly affected by the length of the translation for "Rating".

> Actually, the only that sticks out to me are the non-monospace dates...

I agree, but a lot of users would probably rather notice an unfamiliar monospace font.

dae:
> Alternatively, what if we hide the time of day? With or without margins, it wouldn't be such a balancing act that way.

@RumovZ Sounds like a pretty good solution to me - the time is usually not so important, and the user could rotate their device to see the wider output if they wanted to see it. WDYT @hgiesel?

> I remember trying it on my iPhone when I set those margins, and making sure that the lines fit.

Don't want to undo your previous efforts - if we're heading in the wrong direction, please let us know :-)

RumovZ:
I've taken the silence as acquiescence and made the change.

hgiesel:
Sorry for no answer. Generally I'm no fan of hiding information, just because the viewport is too small. Reminds me how YouTube hides the publish date of a video until you increase the width. But if it leads to a better display, I'm all for it.

In the tag editor, we shorten tags, and make the whole tag visible on hover via a tooltip. The bootstrap tooltips work on mobile as well, but activate on tap. Maybe this could work in this scenario?

RumovZ:
Personally, I never had any use for the revlog time of day, but it would be very intuitive to show it on hovering or tapping the date. Not sure how that would work for the other currently hidden columns though.
As I'm not an AnkiMobile user, I don't want to push any opinion here.

hgiesel:
> Not sure how that would work for the other currently hidden columns though.

Possibly a horizontally scrollable container.
That is something that would work both on desktop, if the screen width is small, and on mobile, too.

For example I've seen them on the Netflix:
Screenshot 2022-01-14 at 21 24 17
Clicking on the right side of (where the white arrow is) scrolls a bit to the right, and it also makes it obvious, that it's possible to scroll horizontally.

But we could also wait until we get a user request for it ;)

dae:
Not 100% sure about the use of monospace there - does it improve readability enough to justify the different appearance compared to the top part? Also a bit concerned about how it will appear on systems with an uglier monospace font. Anyone have any thoughts either way?

Screen Shot 2022-01-16 at 1 49 17 pm
 
Aside from an answer to that, this looks good to go

RumovZ:
Doesn't look too bad on your Mac, but with the default Win 10 monospace `@` is almost unrecognisable.

grafik

ts/card-info/Revlog.svelte
```diff
@@ -133,6 +136,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
 <style lang="scss">
     .revlog-table {
+        font-family: monospace;
```
dae, line 139:
```suggestion
```

Going to remove this as it's holding up the PR. It's not a long-term no, but if you think it's worth pursuing, it might be worth doing a brief poll on the forums showing before/after to see if it's a change people are keen on.

Thanks Rumo!

dae:
> But we could also wait until we get a user request for it ;)

👍

ts/card-info/Revlog.svelte
RumovZ: Sorry, that was completely unintentional! I just wanted to try it out and wasn't aware that I had actually committed it.
Now your last comment makes a lot more sense. 😅

dae: 😄

PR #1584, opened 2022-01-09
Fix IME input after tab

hgiesel:
This PR adresses the bug reported [here](https://forums.ankiweb.net/t/anki-2-1-50-beta/15608/124). As it is mentioned in the post, it only happens in Qt6.
I've figured out the issue of the bug: It happens when trying to type after the caret has been placed programmatically at the end of the node (`placeCaretAfterContent` in ts/domlib/place-caret).
I'm not entirely sure how to circumvent it yet.

### DIFF
--- a/ts/editor/OldEditorAdapter.svelte
+++ b/ts/editor/OldEditorAdapter.svelte
@@ -29,7 +29,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     const currentField = writable<EditorFieldAPI | null>(null);
 
     function updateFocus() {
-        get(activeInput)?.moveCaretToEnd();
+        /* get(activeInput)?.moveCaretToEnd(); */
     }
 
     registerShortcut(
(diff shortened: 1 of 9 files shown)


dae:
I'd be pleasantly surprised if we can solve this problem entirely in our code - switching fields without committing the pending IME string has been problematic for years, and I seem to recall it's even an issue in Safari on iOS as well.

hgiesel:
I found a workaround: Basically, right after placing the caret after the content, we save & restore the location with my low-level `ts/domlib/location` library.

I have also come upon another bug while fixing this one: After moving out of a Mathjax element with the keyboard, this whole "set caret to end of field upon Tab" is broken. I'll fix it in the next PR.

ts/editor/OldEditorAdapter.svelte
```diff
@@ -29,7 +29,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     const currentField = writable<EditorFieldAPI | null>(null);
 
     function updateFocus() {
-        get(activeInput)?.moveCaretToEnd();
+        /* get(activeInput)?.moveCaretToEnd(); */
```
dae, line 32: Seems to work well - thanks for digging into this! Just a minor point about this and the .disconnect() comment  down below. If the code or some variation of it is intended to be added back in the future, maybe a comment explaining why it's been commented out for now would be good. If the code won't be coming back, maybe it would be better to just delete it?

PR #1585, opened 2022-01-10
Use consistent colors across years in calendar graph.

rye761:
With the turn of the year, I noticed that the calendar graph changes palette when flipping between years, limiting it's usefulness for comparing review activity over time where multiple years are involved. The issue is demonstrated in the two screenshots below, where you can see a similar level of reviews being much lighter in 2021 due to more activity earlier in the year.

 
image
image

This pull request fixes that. I never filed a bug report as I wanted to confirm the issue in the latest code and once I had built, it was a really simple fix. Here's a screenshot of 2022 under the fix.

image

As you can see, my review activity is much more in line with what was seen at the end of 2021.


dae:
Seems reasonable; thanks for the contribution!

PR #1586, opened 2022-01-10
Browser sidebar icon improvements

hgiesel:


dae:
Thanks Henrik. Would be nicer if the flag were a bit less pixelart-ish, but at least it's consistent now :-)

hgiesel:
I've requested for the `flag-variant` icon, which looks less like pixelart, to be extended with modifiers:

https://github.com/Templarian/MaterialDesign/issues/6415

Let's see how it goes :)

PR #1587, opened 2022-01-12
fix: try catch excepton in get_windows_dark_mode and global cache the check

qxo:
```
Qt warning: DirectWrite: CreateFontFaceFromHDC() failed (指示输入文件 (例如字体文件) 中的错误。) for QFontDef(Family="Fixedsys", pointsize=9, pixelsize=13, styleHint=5, weight=400, stretch=100, hintingPreference=0) LOGFONT("Fixedsys", lfWidth=0, lfHeight=-13) dpi=120
Qt fatal: Traceback (most recent call last):
  File "D:\Python\Python39\lib\site-packages\aqt\main.py", line 154, in __init__
    self.setupUI()
  File "D:\Python\Python39\lib\site-packages\aqt\main.py", line 189, in setupUI
    self.setupStyle()
  File "D:\Python\Python39\lib\site-packages\aqt\main.py", line 1047, in setupStyle
    theme_manager.apply_style()
  File "D:\Python\Python39\lib\site-packages\aqt\theme.py", line 179, in apply_style
    self.night_mode = self._determine_night_mode()
  File "D:\Python\Python39\lib\site-packages\aqt\theme.py", line 163, in _determine_night_mode
    return get_windows_dark_mode()
```


dae:
Thank you for contributing! Was there an error message underneath the traceback? I'm guessing that the registry key is missing on your machine? Are you using Windows 10?

Caching is not appropriate here, as we want the UI to update when the theme changes. If we expect the call to fail on some machines, I think we can just silently return False without logging.

dae:
Thanks qxo!

PR #1588, opened 2022-01-13
Remove individual .html files + other refactorings

hgiesel:
I'm sorry, this PR ended up being all over the place:

1. AddCards-specific code is moved from `OldEditorAdapter` to `NoteCreator.svelte`
2. I gave the toggle buttons for the editing inputs (on the editor fields) new strings (which are a bit more specific) and adjusted their vertical alignment to make them dead center
3. I removed the `appendInParentheses` helper. I remember that when I initially wrote it, I thought, we'd always have to write: ```<span title={`${tooltip} (${shortcut})`} />``` instead of `<span title="{tooltip} ({shortcut})" />`.
4. That is the big one: I removed all `.html` files in ts/*, and they are now generated from ts/page.html and ts/generate_page.bzl instead. My motivation here is to remove some complexity, and also make sure that the pages are run under more similar conditions. For example, some pages still include bootstrap in their .html file, which happened to made code dealing with Bootstrap imported from Svelte components behave differently.


dae:
> That is the big one: I removed all .html files in ts/*, and they are now generated from ts/page.html and ts/generate_page.bzl instead.

Seems reasonable. Looks like deck-options is still manually created though - it can't easily be converted like the others?

hgiesel:
Weird, I could've sworn I had them all removed. Oh well..

dae:
I had to fix the change notetype screen, but otherwise looks good - thanks Henrik.

qt/aqt/browser/card_info.py
```diff
@@ -63,15 +63,11 @@ def _setup_ui(self, card_id: CardId | None) -> None:
...
-
-        self.web.eval(
-            "const cardInfo = anki.cardInfo(document.getElementById('main'));"
-        )
         self.update_card(card_id)
 
     def update_card(self, card_id: CardId | None) -> None:
         self.web.eval(
-            f"cardInfo.then((c) => c.$set({{ cardId: {json.dumps(card_id)} }}));"
+            f"anki.cardInfoPromise.then((c) => c.$set({{ cardId: {json.dumps(card_id)} }}));"
```
dae, line 70: Since we've gone to the trouble of changing these references already, would it make sense to switch from a global to using our register-package functionality as well? Eg so calls like the above would become something like

```
require("anki/card-info").loaded.then(...)
```

I presumably that would give us a bit more flexibility in terms of renames and deprecations, and would be more discoverable than global variables. WDYT? Would there be disadvantages to doing it that way?

hgiesel: Actually I had the same thought.
But rather than doing it per package, I wanted to expose one single package (in lib), that exposes, let's say, `uiLoaded` (or we could keep the iOS naming scheme: `uiDidLoad`).

My thought is that we're always going to have _one_ Svelte entry point per webview, so it would be easy to maintain going forward.

However this promise would not return any API, but rather `void`. To access the Editor API, we could expose it in an interface similar to `aqt.dialogs`:

```ts
const { uiDidLoad, uiComponents } = require("anki/ui")

await uiDidLoad;

// My intention is to rename `OldEditorAdapter.svelte` to
// `BaseEditor`, and have it contain all functionality shared between all
// editor instances.
const editor = uiComponents.get("baseEditor" /* or maybe a symbol */);
```

This would make things easier for add-ons and also safe going forward, however it'll create quite a bit boilerplate for `web.eval` calls from Python.

dae: No preferences on the name - on the Qt end where we often have hooks before and after some action, the awkward naming made some sense, but we don't necessarily need to keep it here.

A single promise for all screens sounds like it could work. If 'await ui, then load component' is a common use case, then perhaps we could add some helper routine to reduce the boilerplate in such cases. Symbols might be a better choice there, or some other approach that allows us to strongly type the input and output of uiComponents.get()

PR #1598, opened 2022-01-16
fix: AttributeError: 'QMouseEvent' object has no attribute 'pos'

qxo:
```
Caught exception:
  File "D:\Python\Python39\lib\site-packages\aqt\browser\sidebar\tree.py", line 328, in mouseReleaseEvent
    if (index := self.currentIndex()) == self.indexAt(event.pos()):
d
```


dae:
Thank you, looks good! For future reference, you can test your changes locally with '.\bazel.bat test ...'. You may find that is faster than having to push your changes after each small change, and it will save me being sent lots of check failure emails :-)

qxo:
> bazel.bat

ok, I'll try.
I'm new to bazel and python is not my main development language.

PR #1600, opened 2022-01-17
Backend Custom Study

### LINKED ISSUE #1176: 'custom study' needs to be moved to the backend and made undoable

RumovZ:
Closes #1176.
I somewhat struggled with a clean implementation, so please let me know if you have any ideas for improvement. In detail:

- Building of search strings feels a bit awkward. I wrote a small macro to reduce some boilerplate and am manually pushing `Node::And`s to concatenate nodes. Maybe we need some kind of API in the search module?
- Not sure if `scheduler::filtered::custom_study.rs` is the right place for the code, but it was convenient as I'm using a private method from the parent module (`add_or_update_filtered_deck_inner()`).
- Previously, the dialog closed if it failed with the "Existing custom study deck" error. However, there now is no indicator (apart from the localized description) which custom study error occured, so the dialog remains open after a failure, because that was the behaviour of the "No matching cards" error and I gathered that one will account for 99.9% of the errors.

Other remarks:
- The check for no matching cards was fixed as a side effect.
- Seems like a few unrelated changes from `rslib:format` made it into this PR. (Or maybe `cargo fmt`.)
- I think `TagLimit` should return two lists of tags, so we can build the search string on the backend. No add-ons seem to be using it, should I make that change?


dae:
Thanks Rumo, will get back to you in more detail tomorrow morning.

> I think TagLimit should return two lists of tags, so we can build the search string on the backend. No add-ons seem to be using it, should I make that change?

Yep, sounds good.

rslib/src/scheduler/filtered/custom_study.rs
```diff
@@ -0,0 +1,226 @@
...
+            }
+            id = did;
+        }
+
+        self.add_or_update_filtered_deck_inner(FilteredDeckForUpdate {
+            id,
+            human_name,
+            config,
+        })
+        .map(|_| ())
```
dae, line 81: Very minor, and may come down to personal preference, but I find it can make things a little clearer if longer arguments are first aliased to a temporary variable, eg

```rust
        let deck = FilteredDeckForUpdate {
            id,
            human_name,
            config,
        };
        self.add_or_update_filtered_deck_inner(deck)
            .map(|_| ())
```

rslib/src/scheduler/filtered/custom_study.rs
```diff
@@ -0,0 +1,226 @@
...
+    })];
+    custom_study_config(deck_name, true, nodes, FilteredSearchOrder::Due, None)
+}
+
+fn preview_config(deck_name: String, days: u32) -> FilteredDeck {
+    let nodes = vec![
+        Node::Search(SearchNode::State(StateKind::New)),
+        Node::And,
+        Node::Search(SearchNode::AddedInDays(days)),
+    ];
```
dae, line 137: The existing macros and TryIntoSearch trait could make this a bit easier:

```diff
@@ -130,11 +132,11 @@ fn ahead_config(deck_name: String, days: u32) -> FilteredDeck {
 }

 fn preview_config(deck_name: String, days: u32) -> FilteredDeck {
-    let nodes = vec![
-        Node::Search(SearchNode::State(StateKind::New)),
-        Node::And,
-        Node::Search(SearchNode::AddedInDays(days)),
-    ];
+    let nodes = match_all![
+        SearchNode::State(StateKind::New),
+        SearchNode::AddedInDays(days)
+    ]
+    .into_node_list();
     custom_study_config(
         deck_name,
         false,
```

If custom_study_config() were reworked to avoid the concatenate and additional deck argument, we could just add the deck to the list, and make things a bit simpler again:

```diff
diff --git a/rslib/src/scheduler/filtered/custom_study.rs b/rslib/src/scheduler/filtered/custom_study.rs
index 269b3ce16..e428d3678 100644
--- a/rslib/src/scheduler/filtered/custom_study.rs
+++ b/rslib/src/scheduler/filtered/custom_study.rs
@@ -13,7 +13,8 @@ use crate::{
     error::{CustomStudyError, FilteredDeckError},
     prelude::*,
     search::{
-        concatenate_searches, BoolSeparator, Node, PropertyKind, RatingKind, SearchNode, StateKind,
+        concatenate_searches, writer::write_node, BoolSeparator, Node, PropertyKind, RatingKind,
+        SearchNode, StateKind,
     },
     text::escape_anki_wildcards_for_search_node,
 };
@@ -113,6 +114,24 @@ fn custom_study_config(
     }
 }
 
+fn custom_study_config_example(
+    reschedule: bool,
+    node: Node,
+    order: FilteredSearchOrder,
+    limit: Option<u32>,
+) -> FilteredDeck {
+    FilteredDeck {
+        reschedule,
+        search_terms: vec![FilteredSearchTerm {
+            search: write_node(&node),
+            limit: limit.unwrap_or(99_999),
+            order: order as i32,
+        }],
+        delays: vec![],
+        preview_delay: 10,
+    }
+}
+
 fn forgot_config(deck_name: String, days: u32) -> FilteredDeck {
     let nodes = vec![Node::Search(SearchNode::Rated {
         days,
@@ -130,18 +149,12 @@ fn ahead_config(deck_name: String, days: u32) -> FilteredDeck {
 }
 
 fn preview_config(deck_name: String, days: u32) -> FilteredDeck {
-    let nodes = vec![
-        Node::Search(SearchNode::State(StateKind::New)),
-        Node::And,
-        Node::Search(SearchNode::AddedInDays(days)),
+    let node = match_all![
+        SearchNode::State(StateKind::New),
+        SearchNode::AddedInDays(days),
+        SearchNode::Deck(deck_name),
     ];
-    custom_study_config(
-        deck_name,
-        false,
-        nodes,
-        FilteredSearchOrder::OldestReviewedFirst,
-        None,
-    )
+    custom_study_config_example(false, node, FilteredSearchOrder::OldestReviewedFirst, None)
 }
 
 macro_rules! state_node {
diff --git a/rslib/src/search/writer.rs b/rslib/src/search/writer.rs
index 0b700d915..a87f333ac 100644
--- a/rslib/src/search/writer.rs
+++ b/rslib/src/search/writer.rs
@@ -69,7 +69,7 @@ pub fn write_nodes(nodes: &[Node]) -> String {
     nodes.iter().map(write_node).collect()
 }
 
-fn write_node(node: &Node) -> String {
+pub fn write_node(node: &Node) -> String {
     use Node::*;
     match node {
         And => " ".to_string(),
```

Maybe instead of making write_node() public, we should implement a write() method on Node.

rslib/src/scheduler/filtered/custom_study.rs
```diff
@@ -0,0 +1,226 @@
...
+        false,
+        nodes,
+        FilteredSearchOrder::OldestReviewedFirst,
+        None,
+    )
+}
+
+macro_rules! state_node {
+    (not $state_kind:expr) => {
+        Node::Not(Box::new(Node::Search(SearchNode::State($state_kind))))
```
dae, line 149: Neat. This makes me wonder why I boxed the inner argument in the Not case though - presumably it would have worked without the extra layer of indirection.

dae (commented):
Thanks Rumo, the code is lovely and neat, and worked well in my brief testing. I've left some comments below.

rslib/src/scheduler/filtered/custom_study.rs
RumovZ: Isn't it necessary because we would have an potentially infinite chain with unknown size otherwise?
I like the macro too, but I guess it's not very scalable. Might be a better idea to impl some traits and use `Node::negated()`.

rslib/src/scheduler/filtered/custom_study.rs
RumovZ: Thanks, I was missing these tools. I think I'll expand on them in a separate file.

RumovZ:
Sorry, I ended up making quite a few changes to backend search building. Hopefully this will be easier to maintain and save us from some errors, as it exposes less implementation details and is quite ergonomic to use in my opinion.

rslib/src/scheduler/filtered/custom_study.rs
dae: Ah yep, the recursion issue - not sure why that slipped my mind :-)

dae:
There are a fair few changes there, but the new API looks good. Thanks for all your work on this Rumo! 🚀

PR #1601, opened 2022-01-17
feat: missing media file list it's note nids

qxo:
so we can check the note on searchbox by nid:xxx
ie:
```
以下文件在卡片中被提及但在媒体文件夹里没找到：
缺少：20495.gif --> nid:1449637010685
缺少：20521.gif --> nid:1449637010472
缺少：20711.gif --> nid:1449637009902
缺少：20864.gif --> nid:1449637010439
缺少：20901.gif --> nid:1449637010628

```

### DIFF
--- a/rslib/src/media/check.rs
+++ b/rslib/src/media/check.rs
@@ -353,13 +353,14 @@ where
     fn check_media_references(
         &mut self,
         renamed: &HashMap<String, String>,
-    ) -> Result<HashSet<String>> {
-        let mut referenced_files = HashSet::new();
+    ) -> Result<HashMap<String, HashSet<String>>> {
+        let mut ret: HashMap<String, HashSet<String>> = HashMap::new();
         let notetypes = self.ctx.get_all_notetypes()?;
         let mut collection_modified = false;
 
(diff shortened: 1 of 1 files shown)


dae:
I'm afraid I'm not fond of this change. `missing` is intended to be a list of filenames, not arbitrary strings, which calling code may want to do something with. I'm also not really sure this provides a lot of value - if you're copying and pasting the nid link into the browser, you can copy&paste the filename into the browser instead, and most of the time it will accomplish the same thing.

qxo:
> I'm afraid I'm not fond of this change. `missing` is intended to be a list of filenames, not arbitrary strings, which calling code may want to do something with. I'm also not really sure this provides a lot of value - if you're copying and pasting the nid link into the browser, you can copy&paste the filename into the browser instead, and most of the time it will accomplish the same thing.

I mean "the anki note browser" not the web browser...
The idea of the feature is "I want to known which note missing which media"
if we cloud just click the missing media link to card not that's will be great!

dae:
Yep, I understand which browser you mean. My point is, if you are copying&pasting nid:1449637010685 into the browser, why don't you copy&paste 20495.gif instead, and then search for that?

Issue #985, opened 2020-11-24
Selectively enable pinch-to-zoom


dae:
Currently webview.py's eventFilter() captures native gesture events and ignores them to prevent pinch-to-zoom,  because:

- the top and bottom webviews, and places like the deck list and editing screen, do not make sense to zoom
- issues have been reported in the past where the reviewing screen behaved oddly after zooming in and then answering a card

If the latter issue could be fixed/confirmed to no longer exist, selectively enabling pinch during review might be worth looking into.

Of further note, the event filter doesn't appear to work reliably anymore in recent Qt versions. On a Mac here for example, pinching in the main screen starts off, but once the congratulations screen is shown, pinching is now possible, even when returning to the deck list.

Related: https://forums.ankiweb.net/t/feature-request-enable-pinch-to-zoom-on-trackpad/5333

georgez-zgeorge:
Just a note on this. This is my current set up:
```
    Anki 2.1.38 (355e4cd5) Python 3.8.6 Qt 5.14.2 PyQt 5.14.2
    Platform: Mac 10.16
    Flags: frz=True ao=True sv=2
    Add-ons, last update check: 2021-02-17 15:38:56
```
I'm unable to pinch to zoom at all until I see an congratulations screen. I can't tell you how long I've been working to figure out what combination of settings, addons, closing my computer, sitting facing north, and other crazy configurations I tried to figure out why pinch to zoom only worked _sometimes_ until I found this post. Thank you for figuring out this bug and I look forward to being able to zoom at any time on the review screen.

RumovZ:
Can be closed I guess.

PR #1602, opened 2022-01-17
Selectively disable zoom

RumovZ:
This disables zoom for the top and bottom bars in the main view. Unlike the previous eventfilter, it also disables zooming via scrolling.

- Would be great if this could be tested by a mac user. It is my understanding that zooming can also be done on a mac by holding <kbd>Control</kbd> and a two finger swipe. This should also be disabled.
- I've kept the filter for `QNativeGestureEvent`, but isn't that a bit unspecific? Are there no other gestures we also end up disabling?
- I couldn't reproduce any of the issues described in #985.

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -869,12 +869,14 @@ title="{}" {}>{}</button>""".format(
         # toolbar
         tweb = self.toolbarWeb = AnkiWebView(title="top toolbar")
         tweb.setFocusPolicy(Qt.FocusPolicy.WheelFocus)
+        tweb.disable_zoom()
         self.toolbar = aqt.toolbar.Toolbar(self, tweb)
         # main area
         self.web = MainWebView(self)
         # bottom area
         sweb = self.bottomWeb = AnkiWebView(title="bottom toolbar")
         sweb.setFocusPolicy(Qt.FocusPolicy.WheelFocus)
+        sweb.disable_zoom()
         # add in a layout
         self.mainLayout = QVBoxLayout()
         self.mainLayout.setContentsMargins(0, 0, 0, 0)
--- a/qt/aqt/utils.py
+++ b/qt/aqt/utils.py
@@ -1002,6 +1002,22 @@ def no_arg_trigger(func: Callable) -> Callable:
     return pyqtSlot()(func)  # type: ignore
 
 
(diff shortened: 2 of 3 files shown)


dae:
Thanks Rumo, seems to work well on a Mac+touchpad here (just pinching - I'm guessing the option key is only required when using a mousewheel).

The only other gesture that might possibly make sense is the pan one, which didn't exist in Qt5: https://doc.qt.io/qt-6/qnativegestureevent.html. Might be worth waiting for someone to say they miss it first. :-)

One remaining question - should we be always allowing zooming in the main webview / editor / etc, or should we limit it to particular views (eg review) where it might be useful?

RumovZ:
>seems to work well on a Mac+touchpad here (just pinching - I'm guessing the option key is only required when using a mousewheel).

My mistake, I meant <kbd>Control</kbd> (mapped to <kbd>Meta</kbd> in Qt). Swiping with two fingers should trigger a scroll event, ergo you should be able to zoom with <kbd>Control</kbd> (on all OSes) and this gesture.
But even if I'm right, it's not that important because you would usually just pinch if you have a trackpad. I'd like to fix this if I should be wrong though.

> One remaining question - should we be always allowing zooming in the main webview / editor / etc, or should we limit it to particular views (eg review) where it might be useful?

Not sure. I can't imagine why you would zoom on the deck tree instead of increasing the interface size, but it doesn't do any harm either, does it?
Disabling zoom in the deck tree, but not in review would be a bit more tricky, as we would need to dis-/enable and reset zoom when the main window state changes.

dae:
Maybe you're referring to an accessibility feature? When I turn it off, I don't see ctrl+two fingers up/down doing anything, and with it on, the entire screen zooms in/out (a handy feature that I use a fair bit!)

Screen Shot 2022-01-18 at 7 10 14 pm

You're right that it doesn't do much harm. It might be slightly confusing if users do it accidentally, but the same thing could be said about most of Anki's features :-) Thanks for your work on this Rumo.

RumovZ:
Then it's different for Mac. On my machine, <kbd>Ctrl</kbd>+Swipe does the same as <kbd>Ctrl</kbd>+Mouse wheel. Thanks for clarifying!
I better fix `is_zoom_event()` then, or we might get surprised in the future.

PR #1606, opened 2022-01-19
Make `note:` and `card:` search require full match

RumovZ:
Note sure if this changed at some point, but I think requiring a full match is the only reasonable option here.


dae:
Yep, a partial match was not intended - good catch!

PR #1607, opened 2022-01-19
Fix `is_zoom_event()`

RumovZ:
Cleanup for #1602.


dae:
Thanks Rumo. I double-checked the behaviour here, and even when plugging in a mouse with a scrollwheel (which is fairly rare on Macs), it seems ctrl+scroll does nothing, regardless of the accessibility setting.

PR #1608, opened 2022-01-20
Add last version check for add-on updates

RumovZ:
https://forums.ankiweb.net/t/tell-users-to-restart-after-un-installing-and-toggling-enable-disable-add-ons/16771/

This revealed that there were two update checks. I removed the one in `_deckBrowserState()`.
I couldn't figure out how that method gets called in the first place.

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -31,7 +31,16 @@ from anki.decks import DeckDict, DeckId
 from anki.hooks import runHook
 from anki.notes import NoteId
 from anki.sound import AVTag, SoundOrVideoTag
-from anki.utils import dev_mode, ids2str, int_time, is_lin, is_mac, is_win, split_fields
+from anki.utils import (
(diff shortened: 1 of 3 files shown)


qt/aqt/main.py
```diff
@@ -1112,6 +1120,7 @@ def onStudyKey(self) -> None:
     ##########################################################################
 
     def closeEvent(self, event: QCloseEvent) -> None:
+        self.pm.set_last_run_version()
```
RumovZ, line 1123: I figured `last_run_version` is the most useful, when it is set as late as possible. But is this the right place?

dae: The only other place I can think of is at the end of _run() in aqt/__init__.py, but I can't think of a big reason to prefer one over the other, so until one arises, this should do :-)

dae:
Great, thanks for this Rumo.

PR #1613, opened 2022-01-21
Separate input components into their own directories / Remove `WithShortcut`

hgiesel:
This PR is a mostly just moving around code, but it also does two important things:

1. It creates separate directories for PlainTextInput, RichTextInput, and the EditorToolbar. One major thing I wanted to work on before a possible 2.1.50 release was the Undo/Redo feature in the editor. I'll assume it'll be common point of complaining if 2.1.50 delivers without it. The good thing, if we manage to build one ourselves is: we could even sustain it between sessions, maybe even save it to disk? That's why right now, putting them into separate directories might seem a bit overkill, but soon they might all contain a bit more logic of feeding an Undo stack.

2. I removed the `WithShortcut` component, and replaced their uses in favor `Shortcut`. It was one of the components, that relied on `on:mount`, which I'd like to refactor away in the future. `Shortcut` causes a bit of duplication, but it avoids one level of nesting, and it's more general purpose. The duplication can be more easily dealt with, once we can update Svelte to its newest version and use the new `{@const ...}` feature.


dae:
- Shortcut changes seem reasonable; reducing the nesting a bit will be nice, and @const sounds handy.
- Separate subfolders inside the same module are almost free, so if they improve clarity, no objections from me. I seem to recall re-exports breaking tree shaking in the past - this isn't going to be an issue for editing functionality that is always included anyway, but we may want to check at one point to see if index_* files are excluding code specific to one case like adding. Not a huge priority though.
- I'm a bit torn on trying to implement our own undo stack before a 2.1.50 release. How long do you estimate it will take? 2.1.50 does not have a hard deadline, but we're already about 850 commits away from the 2.1.49 release, and we've accumulated a fair few bugfixes that will be held up by a delayed release. Will it require further changes to the add-on-facing API? If it will mean further API breakages, that is an argument in favour of doing it now rather than post 2.1.50.

hgiesel:
> I seem to recall re-exports breaking tree shaking in the past.

Yep. That's why I really only kept it to one component and its exclusive dependencies.

Concerning the Undo feature: in that case, it can probably wait. Even though I don't think it'll break any outward-facing API, but it'll probably be a bigger PR. Concerning the outward-facing add-on API, I'll have a PR ready tomorrow.

PR #1614, opened 2022-01-21
Remove outdated info about VS Code

roxgib:
Pylance is now the default for Python on VS Code, so it should no longer be necessary to suggest switching to it.

I haven't used PyCharm so maybe there is something I'm missing, but for the record I haven't had any issues with Pylance understanding the type annotations. Maybe it's worth checking if this is still true? It's pretty new so it may have had improvements since this was written.


dae:
Thanks for the heads up. I gave it a quick test, and things do appear to have improved quite a bit. Will give it a try over a longer period; if I don't run into issues I'll update the docs further.

PR #1618, opened 2022-01-21
Fix field content overflowing its container

hikaru-y:
This fixes https://forums.ankiweb.net/t/text-can-exit-the-file/16903.

### DIFF
--- a/ts/editable/ContentEditable.svelte
+++ b/ts/editable/ContentEditable.svelte
@@ -55,7 +55,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         display: block;
         padding: 6px;
         overflow: auto;
-        overflow-wrap: break-word;
+        overflow-wrap: anywhere;
+        /* fallback for iOS */
+        word-break: break-word;
 
         &:focus {
             outline: none;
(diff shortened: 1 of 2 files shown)


hgiesel:
Seems like

```css
overflow-wrap: anywhere;
word-break: break-word;
```

also does the trick.

hikaru-y:
Thanks, @hgiesel. I also thought of `overflow-wrap: anywhere`, but it seems that iOS doesn't support the value `anywhere`, so I used `min-width: 0`.
https://caniuse.com/mdn-css_properties_overflow-wrap_anywhere

hgiesel:
That's why I also added `word-break: break-word`, which seems to fulfil the same purpose but is officially deprecated: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break.

dae:
Thanks Hikaru, I appreciate you thinking about the iOS side of things as well! break-word is successfully used in AnkiMobile's current editor, so maybe that might be a better choice? It does make the intention a bit more explicit.

hikaru-y:
> It does make the intention a bit more explicit.
I agree with you.

dae:
Thank you both! :heart:

PR #1620, opened 2022-01-21
Simplify name

Arthur-Milchior:
While considering https://forums.ankiweb.net/t/issue-with-new-change-note-in-add-note/16909 I realized that "field" was used sometime to mean the value and sometime the key. This small change simply ensure that it becomes explicit and do some small improvement in the same part of the code


dae:
Nice and clear, thanks Arthur.

PR #1622, opened 2022-01-22
Optimise searching in (all) fields

RumovZ:
https://forums.ankiweb.net/t/wrong-error-message-in-search/16914

With my test collection, the first commit brings down search time for `*:` from 0.23s to 0.18s, and the second one to 0.11s. The tree depth error shouldn't occur either.

Unfortunately, 79140cf4b519c634798be5b36d9dabb9fca0787c doesn't work for regex searches, because if there is an `^` or `$`, it will fail to match.
@dae, what is the reason regex field searches behave differently from plain field searches and do not require a full match? I'm asking because if they did, `^` and `$` wouldn't match anyway and the shortcut wouldn't make a difference.


TheBlackBeans (approved):

RumovZ:
Hope you don't mind the additional code and SQL function. Some reasoning:

- Searches in all or many fields are now much, much faster.
- Searches in all fields and/or regexp searches only require a couple of SQL nodes now. A non-regex search where the first part matches a lot of field names could still bloat the SQL statement, but that's somewhat reduced too, and I didn't want to sacrifice the speed of the native SQL comparison.
- The matching behaviour remains unchanged: plain searches require a full match, regex ones don't. (I'd still be interested to know why the difference though.)
- In the all-fields-no-regexp case, 69b921f is probably a tiny a bit faster than the new code, but I removed it for the sake of simplicity.

rslib/src/search/sqlwriter.rs
```diff
@@ -419,57 +421,119 @@ impl SqlWriter<'_> {
...
+                                    "field_at_index(n.flds, {}) like ?{} escape '\\'",
+                                    ord, arg_idx
+                                )
+                            })
+                            .join(" or "),
+                    )
+                })
+                .join(" or ")
+        )
+        .unwrap();
```
dae, line 509: This is getting a bit unwieldy with the nesting; it may make sense to declare some of the closures above the write!() statement to break it up a bit. Now that we're on Rust 1.58, you can also take advantage of inline replacements: 
https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings

rslib/src/search/sqlwriter.rs
```diff
@@ -419,57 +421,119 @@ impl SqlWriter<'_> {
...
         Ok(())
     }
 
+    fn field_map(&mut self, field_name: &str) -> Result<Vec<(NotetypeId, Vec<u32>)>> {
+        let notetypes = self.col.get_all_notetypes()?;
+        let matches_glob = glob_matcher(field_name);
+
+        let mut field_map = vec![];
+        for nt in notetypes.values() {
+            let mut nt_fields = vec![];
```
dae, line 520:
```suggestion
            let mut matched_fields = vec![];
```

rslib/src/search/sqlwriter.rs
```diff
@@ -654,10 +718,10 @@ mod test {
...
             (
                 concat!(
-                    "(((n.mid = 1581236385344 and field_at_index(n.flds, 0) like ?1 escape '\\') or ",
-                    "(n.mid = 1581236385345 and field_at_index(n.flds, 0) like ?1 escape '\\') or ",
-                    "(n.mid = 1581236385346 and field_at_index(n.flds, 0) like ?1 escape '\\') or ",
-                    "(n.mid = 1581236385347 and field_at_index(n.flds, 0) like ?1 escape '\\')))"
+                    "(((n.mid = 1581236385344 and (field_at_index(n.flds, 0) like ?1 escape '\\')) or ",
+                    "(n.mid = 1581236385345 and (field_at_index(n.flds, 0) like ?1 escape '\\')) or ",
+                    "(n.mid = 1581236385346 and (field_at_index(n.flds, 0) like ?1 escape '\\')) or ",
+                    "(n.mid = 1581236385347 and (field_at_index(n.flds, 0) like ?1 escape '\\'))))"
```
dae, line 724: Coverage of the regex case would be nice too.

dae (commented):
While we probably don't want to add SQL functions too frequently, it makes this case considerably simpler, so it's probably worth it. Thanks Rumo, just some minor bikeshedding below.

dae:
> The matching behaviour remains unchanged: plain searches require a full match, regex ones don't. (I'd still be interested to know why the difference though.)

I don't recall my exact thought process at the time, but I imagine I was going for consistency with the standard regex search, instead of trying to be consistent with the glob-based field search. It's also marginally easier to add ^/$ than it is to add .* at the start and end.

dae:
I've made some minor tweaks to the wording - what do you think? Feel free to revert/change them if you feel they should be done differently. Let me know when you're ready and I'll merge this in :-)

RumovZ:
Looks great, happy to see this merged. 🙂

iamllama:
It's been a while, but could i ask why `regexp_fields` was used for wildcard-all-field searches like `*:blah` instead of treating them as an unqualified search? The latter would easily allow for per-field nc (e.g. `Front:nc:blah`), unless there was a specific reason not to that im not seeing

PR #1623, opened 2022-01-22
qt/aqt/webview: handle opening href with target=_blank

### LINKED ISSUE #1339: Links in cards with `target="_blank"` don't open when clicked
I use links with the target attribute set to `_blank` to open them in new tabs when clicked on AnkiWeb. It looks like such links don't work in Anki desktop, when I click them there, nothing happens.

An example of such a link is:

```html
<a href="https://ankiweb.net/" target="_blank">link</a>
```

danishprakash:
Turns out, if using `target="_blank"`, the `createWindow` method is called and not `acceptNavigationRequest`. But similar to `acceptNavigationRequest`, we're supposed to override `createWindow` and handle the behavior accordingly. I was trying to extract the URL and invoke the `openLink` method, but this turned out to be the preferred approach as per the community.

Note: I was only able to test this on GNU/Linux.

From https://doc.qt.io/qt-5/qwebengineview.html#createWindow:
```
This function is called from the createWindow() method of the associated QWebEnginePage each 
time the page wants to create a new window of the given type. For example, when a JavaScript 
request to open a document in a new window is issued.
```

Fixes #1339

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -91,6 +91,7 @@ git9527 <github.com/git9527>
 Vova Selin <vselin12@gmail.com>
 qxo <49526356@qq.com>
 Spooghetti420 <github.com/spooghetti420>
+Danish Prakash <github.com/danishprakash>
 
 ********************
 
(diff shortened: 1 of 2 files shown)


dae:
Thank for looking into this Danish, it seems to work well!

PR #1625, opened 2022-01-24
Tweak search highlight color in the templates screen

abdnh:
The default highlight color was hard to see, especially in light mode.

Before:
Screenshot 2022-01-24 193816
Screenshot 2022-01-24 193834

After:
Screenshot 2022-01-24 184610
Screenshot 2022-01-24 184544

Using `QPalette.ColorGroup.Inactive` will also cause normally selected text to have the highlight background color if the template editor is not focused, but this is the simplest solution I could think of.

image

I had to move the setup_edit_area() call because it was preventing the palette change to take effect for some reason.

### DIFF
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -240,6 +240,7 @@ class CardLayout(QDialog):
 
     def setup_edit_area(self) -> None:
         tform = self.tform
+        editor = tform.edit_area
 
         tform.front_button.setText(tr.card_templates_front_template())
         tform.back_button.setText(tr.card_templates_back_template())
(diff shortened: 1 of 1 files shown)


qt/aqt/clayout.py
```diff
@@ -233,34 +234,42 @@ def setupMainArea(self) -> None:
...
         tab_width = self.fontMetrics().horizontalAdvance(" " * 4)
-        self.tform.edit_area.setTabStopDistance(tab_width)
+        editor.setTabStopDistance(tab_width)
+
+        palette = editor.palette()
+        palette.setColor(
+            QPalette.ColorGroup.Inactive,
+            QPalette.ColorRole.Highlight,
+            QColor("#4169e1" if theme_manager.night_mode else "#FFFF80"),
+        )
```
dae, line 271: We may want to move these into _vars.scss at one point, but that can wait until we have more semantic color names (eg @hgiesel's comment on https://github.com/ankitects/anki/issues/1471#issuecomment-973130720)

dae (commented):
Thanks Abdo!

dae:
Ah, I just found a problem when giving it a test - on a Linux machine, it seems the text color is white: 

image

Adding this seems to fix it:

        palette.setColor(
            QPalette.ColorGroup.Inactive,
            QPalette.ColorRole.HighlightedText,
            QColor("#ffffff" if theme_manager.night_mode else "#000000"),
        )

dae:
Thanks Abdo!

Issue #1567, opened 2021-12-28
highlight color of buttons is orange on qt6


dae:
Would be nice to return to the blue of before.

https://forums.ankiweb.net/t/anki-2-1-50-beta/15608/96

RumovZ:
Is this a Mac thing? It's blue on Windows, but there is another small issue: The outline is square, but the border is rounded.
![grafik](https://user-images.githubusercontent.com/48286681/149745483-e84145bf-c03e-48e2-b491-de20b9f1f764.png) ![grafik](https://user-images.githubusercontent.com/48286681/149746073-88b6ab89-be71-417c-a2fa-6c1320001da1.png)

As for the colour, maybe the Mac code just needs something like this too:

https://github.com/ankitects/anki/blob/0e06a1c6797d8ac99a7b1b0c2041d9265d03ce35/qt/aqt/webview.py#L408

dae:
Unfortunately that doesn't seem to help. An orange outline is also visible if you tab in the deck list to move the highlight to each deck in turn. It's starting to feel like Qt have (perhaps accidentally?) change the hard-coded color in Qt 6.2.

roxgib:
mw.palette().highlight().color().getRgb() returns 119, 204, 255, which is blue, but if you flip the order of the RGB elements to 255, 204, 119, you get orange. A hack to fix this might involve manually reversing them back to the correct order, but it'd require checking exactly which platforms are affected.

dae:
More discussion on #1627

dae:
Appears to be resolved.

PR #1627, opened 2022-01-24
Fix button focus highlight on Windows

RumovZ:
Fixes the issue described in https://github.com/ankitects/anki/issues/1567 and a similar one for the reviewer buttons at the bottom.
Before:
grafik grafik grafik grafik
After:
grafik grafik grafik grafik

This took me forever and I still don't fully understand the interplay of the three files, so most likely someone will have to clean things up. I've probably broken things for other systems, and didn't do things how they're supposed to be done. (E.g. `--focus-color` is declared twice now, but I assumed there was a reason it's not in `_vars.scss`.)

Heads-up, @hgiesel.

### DIFF
--- a/qt/aqt/data/web/css/reviewer-bottom.scss
+++ b/qt/aqt/data/web/css/reviewer-bottom.scss
@@ -52,11 +52,25 @@ button {
 .focus {
     outline: 5px auto var(--focus-color);
 
+    .nightMode & {
+        outline: unset;
+        box-shadow: 0 0 0 2px var(--focus-color);
+    }
+
     #innertable:focus-within & {
         outline: unset;
 
+        .nightMode & {
+            box-shadow: unset;
+        }
+
         &:focus {
             outline: 5px auto var(--focus-color);
+
+            .nightMode & {
+                outline: unset;
+                box-shadow: 0 0 0 2px var(--focus-color);
+            }
         }
     }
 }
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -413,8 +413,9 @@ class AnkiWebView(QWebEngineView):
         if is_win:
             # T: include a font for your language on Windows, eg: "Segoe UI", "MS Mincho"
             family = tr.qt_misc_segoe_ui()
-            button_style = "button { font-family:%s; }" % family
-            button_style += "\n:focus { outline: 1px solid %s; }" % color_hl
+            button_style = f"""
+button {{ font-family: {family}; }}
+button:focus {{ outline: 5px auto {color_hl}; }}"""
             font = f"font-size:12px;font-family:{family};"
         elif is_mac:
             family = "Helvetica"
--- a/sass/buttons.scss
+++ b/sass/buttons.scss
@@ -1,5 +1,13 @@
 @use "fusion-vars";
 
+:root {
+    --focus-color: #0078d7;
+
+    .isMac {
+        --focus-color: rgba(0 103 244 / 0.247);
+    }
+}
+
 .isWin {
     button {
         font-size: 12px;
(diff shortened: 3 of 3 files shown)


dae:
Would you mind giving this a quick look-over when you have a chance @hgiesel? You know more CSS than I do :-)

hgiesel:
https://github.com/RumovZ/anki/compare/button-focus...hgiesel:button-focus

I also made sure that macOS has blue highlights now again. The reason for the color change is that Webkit changed the value for `-webkit-focus-ring-color`, which is the outline-color that is applied on :focus.

@RumovZ you should have another look, that it still looks fine on Windows, and then we're ready to go from my side. (One of these days, I'll try setting up my Anki dev env on Windows again...)

For a better understanding of what `unset` does, I recommend [this article](https://elad.medium.com/understanding-the-initial-inherit-and-unset-css-keywords-2d70b7121695), which also explain all the other _mysterious values_, like `inherit` or `initial`. 

> --focus-color is declared twice now, but I assumed there was a reason it's not in _vars.scss.

I think the main reason why I didn't put it into _vars.scss is that we also have different focus colors per OS. I wonder if that is necessary?

Another thing: it might be simpler to aim to use a combination of `border-color` and `box-shadow` to recreate the focus highlight, instead of using `outline`. The browser defaults seem to do some funny things to the `outline` value, e.g. you cannot set the width of the outline, but you have a radius applied to the corners. If you reset the outline explicitly with `outline: none`, you can set the width, but there's no way to apply any radius.

RumovZ:
> @RumovZ you should have another look, that it still looks fine on Windows, and then we're ready to go from my side. (One of these days, I'll try setting up my Anki dev env on Windows again...)

Looks the same. 👍 By the way, building Anki (can only speak for Win) has become really painless. Thanks, @dae!

> For a better understanding of what unset does, I recommend this article, which also explain all the other mysterious values, like inherit or initial.

Interesting, thanks.

> I think the main reason why I didn't put it into _vars.scss is that we also have different focus colors per OS. I wonder if that is necessary?

I can only say that the current transparent MacOS colour is hardly visible on my machine.

> Another thing: it might be simpler to aim to use a combination of border-color and box-shadow to recreate the focus highlight, instead of using outline. The browser defaults seem to do some funny things to the outline value, e.g. you cannot set the width of the outline, but you have a radius applied to the corners. If you reset the outline explicitly with outline: none, you can set the width, but there's no way to apply any radius.

I struggled with that as well. In light mode, I couldn't get the box-shadow to have the same corners as the button, so I used (or kept using) outline. In night mode, it was the other way around.

Just noticed a minor issue: In light mode, changing focus via mouse doesn't set any visible focus. This seems to have been the case on main for the Good button already, but since 38cebe8 it affects all buttons in light mode. Maybe you have an idea what could be the cause?
Otherwise, this can be merged from my side.

hgiesel:
> I can only say that the current transparent MacOS colour is hardly visible on my machine.

Actually, it isn't that well visible on macOS either (anymore?). I resisted from adjusting it, because I'd rather move it to _vars.scss before and do it in another PR.

> changing focus via mouse doesn't set any visible focus
Hm, seems to work fine here. Clicking on any buttons in the bottom bar will leave the button with a visible focus.

dae:
> I also made sure that macOS has blue highlights now again. The reason for the color change is that Webkit changed the value for -webkit-focus-ring-color

Nice! This seems to work well for the reviewer bottom bar, but the orange highlights are still present everywhere else - when using tab on the deck list to select a deck, the highlight of the study now button on the overview screen, etc. Maybe this needs to be in our core styling?

> I think the main reason why I didn't put it into _vars.scss is that we also have different focus colors per OS. I wonder if that is necessary?

@glutanimate might have some thoughts here, as I think this was code he originally contributed. On Linux, the highlight color can vary depending on desktop environment and theme, and I think this code was trying to get Anki to blend in better with other apps running on the system. If using blue across all platforms would make things simpler, then maybe that's also an option worth considering.

dae:
It would be nice to solve the other focus cases in the future, but this is a step in the right direction, so might as well merge it in now. Thank you both!

PR #1630, opened 2022-01-27
Fix no sound after collection export

kelciour:
https://forums.ankiweb.net/t/plese-answer-me-mp3-files-in-card-not-play-after-export/16991


dae:
Thanks for digging into this Kelciour! Anki is supposed to be setting the working directory to the media folder when a collection is opened, and resetting it to the previous location when the collection is closed. The root cause of this issue appears to be that col.reopen() is not calling col.media.connect() in this case. Rather than try to figure out what changing reopen() might break, it might be worth biting the bullet and looking into getting rid of the directory changing completely, as it's mostly legacy at this point. I mentioned this to @RumovZ yesterday, so he may look into it in the not too distant future.

RumovZ:
Yes, will do. 👍

PR #1636, opened 2022-02-02
Fix active field not being updated after some operation

### LINKED ISSUE #1628: editor does not refresh active field after note undone
When changes are made to a field, and the user uses Edit>Undo, the note is reverted to its previous state in the database, but the current field text remains. This may cause the user to think the undo hasn't been applied, and may cause the new text to be saved again if any keys are pressed in that field.

hikaru-y:
Fixes #1628.

The issue occurs not only when undoing, but also when performing a "Find and Replace".


hikaru-y:
Closing this as #1639 fixes the issue.

PR #1637, opened 2022-02-02
Use eslint for sorting our imports

hgiesel:
One overhead I often have while working on the TS code is sorting imports. This would leave it to eslint to sort our imports (with `bazel run eslint -- --fix`)

In the process I also applied eslint generally to .svelte files, which revealed some issues. I would use this PR, to fix those and fit all requirements. Preferably after #1626 :)


dae:
Seems reasonable, though it looks like it also triggers an 'exports must be at bottom' warning that needs to be manually addressed (eg in card-info)

dae:
.. and it looks like a bunch of other issues we weren't aware of need addressing too ^_^;

package.json
```diff
@@ -89,6 +89,6 @@
         "not op_mini all",
         "not < 1%",
         "Chrome 77",
-        "iOS 12"
+        "iOS 13.4"
```
hgiesel, line 92: That's the minimum version to support ResizeObserver. If the editor makes it AnkiMobile before 13.4 is more spread, we could just generally turn off the image resizing and Mathjax editor for AnkiMobile (that's the two cases were we use `ResizeObserver`).

dae: While it will probably be a while until I get the editor ported over, a bump to 13.4 is not too far off, and may come in the next update, so this should be fine. :+1:

dae:
Thanks for sorting those out Henrik, it's nice to have the extra linting coverage, and the more the code formatting can be automated, the better.

hgiesel:
I broke the editor overlays (Mathjax editor, Image resizer) in this PR 🤷‍♂️ I'm working on it.

PR #1638, opened 2022-02-02
V3 parent limits

RumovZ:
This may not work yet and a lot will probably change soon. I already opened this PR because I ran into a problem:
```
error[E0432]: unresolved import `id_tree`
 --> rslib/src\scheduler\queue\builder\context.rs:6:5
  |
6 | use id_tree::{InsertBehavior, Node, NodeId, Tree, TreeBuilder};
  |     ^^^^^^^ use of undeclared crate or module `id_tree`

error: aborting due to previous error
```
Did something change, and more work is now required to add a new dependancy? Everything works fine inside VSC. I even ran `update.py` this time, but (maybe unsurprisingly) this didn't help.


rslib/src/decks/tree.rs
```diff
@@ -216,20 +216,17 @@ fn sum_counts_and_apply_limits_v3(
     for child in &mut node.children {
         individually_capped_total += sum_counts_and_apply_limits_v3(child, limits);
     }
+    // so we can calculate this total
     node.total_including_children = individually_capped_total.total;
```
RumovZ, line 220: Does this field serve any purpose? I couldn't find any other reference to it, and don't know value it is supposed to have with the new approach.
If we could drop it, this code could be simplified.

dae:
The dependency needs to be listed in rslib/BUILD.bazel as well. Here's a branch you can use as a base, that avoids updating other dependency versions: https://github.com/ankitects/anki/commits/id-tree

Will review the code then follow up

rslib/src/decks/tree.rs
dae: When the deck list screen is refactored, I was thinking it would be nice to have a hover tooltip like we do with the graphs. "Why am I not seeing my cards?" is one of the most common questions, and if the tooltip showed the total cards in the deck with and without children/limits, I thought that might make it easier for users to understand what is going on. The numbers are supposed to be exported via decks.proto:DeckTreeNode, but not yet used by the UI. Would a tally of the child deck's totals without limits be easier?

dae:
Thanks Rumo, the basic approach seems reasonable, and looking forward to seeing how it performs :-)

RumovZ:
> The dependency needs to be listed in rslib/BUILD.bazel as well. Here's a branch you can use as a base, that avoids updating other dependency versions: https://github.com/ankitects/anki/commits/id-tree

Sorry, I simply forgot. Thanks for the branch.

rslib/src/scheduler/queue/builder/mod.rs
```diff
@@ -204,134 +207,126 @@ fn sort_learning(mut learning: Vec<DueCard>) -> VecDeque<LearningQueueEntry> {
...
-        // do more work than necessary
-        let mut remaining = remaining_limits_map(decks.iter(), &config, timing.days_elapsed, true);
-        let selected_deck_limits_at_start = *remaining.get(&deck_id).unwrap();
-        let mut selected_deck_limits = selected_deck_limits_at_start;
-        for limit in remaining.values_mut() {
-            limit.cap_to(selected_deck_limits);
-        }
+        let mut ctx = Context::new(self, deck_id)?;
+        self.storage.update_active_decks(ctx.root_deck())?;
+        let mut queues = QueueBuilder::new(ctx.sort_options.clone());
```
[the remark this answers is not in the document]
RumovZ: Yes, maybe it would. I'll look into it.

rslib/src/storage/deck/mod.rs
```diff
@@ -210,6 +211,19 @@ impl SqliteStorage {
...
     }
 
+    pub(crate) fn immediate_child_decks(&self, parent: &Deck) -> Result<Vec<Deck>> {
+        let prefix_start = format!("{}\x1f", parent.name);
+        let prefix_end = format!("{}\x20", parent.name);
+        let child_descendant = format!("{}%\x1f%", escape_sql_wildcards(&prefix_start));
+        self.db
+            .prepare_cached(concat!(
+                include_str!("get_deck.sql"),
+                " where name >= ? and name < ? and not name like ? escape '\\'"
```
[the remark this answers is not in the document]
RumovZ: Isn't the first point a show-stopper then?
What would be the alternative? Using `storage.child_decks()` and then figuring out the hierarchy in Rust? I guess this wouldn't be easy, but it would solve both points of concern.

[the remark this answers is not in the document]
dae: If you wanted to keep close to the current approach, I imagine you could emulate it by doing the not comparison in Rust after the rows are returned, or by using an extra sqlite function. But before going down that route, it may be worth considering whether we might be able to build our with a single query, similar to the way the deck list code does (IIRC it also deals with possible missing parents)

rslib/src/decks/tree.rs
RumovZ: I don't think either is particularly complicated to implement, but I don't quite understand the usefulness of the first alternative.
Let's say the user has clicked on deck A with a limit of 100. It contains 200 eligible cards, but due to descendant limits it only shows 50.

- Alternative 1: Count all cards, only considering the respective source deck's limit. This may be any number between 50 and 200.
- Alternative 2: Count all cards, only considering deck A's limit. This is a number between 50 and 100.

Have I got this right? If so, I'd say 2 is more useful, because it explains why there are less cards than the clicked-on deck's limit would allow for.

RumovZ:
This is a first: a check (`test_newLimits`) is failing locally, but not here. 🤯 Will look into that tomorrow.

Here are some quick numbers for @Arthur-Milchior's deck _Chanson_ with 39,075 unsuspended new cards. (_Theory_ has more, but throws a db error with v3.)

| order | old | new |
| --- | --: | --: |
| deck | 5 ms | 5 ms |
| pos asc | 5 ms | 25 ms |
| pos desc | 7 ms | 35 ms |

Here is the branch I ran this against: https://github.com/RumovZ/anki/tree/v3-child-limits-baseline
If you have more intricate timer logic, feel free to add it.

rslib/src/storage/deck/mod.rs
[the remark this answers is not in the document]
RumovZ: So indeed using `storage.child_decks()` and then figuring out the hierarchy in Rust, only that you've already done the work which I can copy. 🙂
Wasn't easy to unify with the id_tree building logic, but it forced to to make some other simplifications.

dae:
This should fix @Arthur-Milchior's deck:

  update cards set mod = cast(mod as integer);

I moved all other decks under the theory deck, bumped the new card limit to 9999, and applied it to all subdecks to make the query as long as possible. I then tried out the tests with pos asc. I saw quite different results:

baseline around 77-88ms
this code around 63-74ms

Unfortunately this is because the new code is not yet correct :-) 

The following in the debug console shows that the due numbers are not in ascending order:

```python
cards = mw.col.sched.get_queued_cards(fetch_limit=10000)
for card in cards.cards:
    if card.card.queue == 0:
        print(card.card.due)
```

rslib/src/scheduler/queue/builder/mod.rs
```diff
@@ -204,134 +210,132 @@ fn sort_learning(mut learning: Vec<DueCard>) -> VecDeque<LearningQueueEntry> {
...
+        queues: &mut QueueBuilder,
+        ctx: &mut Context,
+        reverse: bool,
+    ) -> Result<()> {
+        self.storage
+            .for_each_new_card_in_active_decks(reverse, |card| {
+                let bury = ctx.bury_mode(card.original_deck_id.or(card.current_deck_id));
+                if let Some(node_id) = ctx.limits.remaining_node_id(card.current_deck_id) {
+                    if queues.add_new_card(card, bury) {
+                        ctx.limits.decrement_node_and_parent_limits(&node_id, true);
```
dae, line 330: This code looks to be exiting early when limits have been reached.  That won't work with the approach that the code currently uses, which is to gather max(deck, ...parent decks) from each deck, sort the received cards, and then truncate them to the global limit.

RumovZ:
> The following in the debug console shows that the due numbers are not in ascending order:

This fits the result I got from the failing test. Curiously, it didn't fail today nor did the GitHub action. Same code, of course.

rslib/src/scheduler/queue/builder/mod.rs
RumovZ: Just stumbled across the same thing. :)

RumovZ:
Ah, now it passes locally but fails here. 🙃 
No worries, I'll take care of the bugs when the refactoring is more or less done. I'll also add some tests.

rslib/src/decks/tree.rs
dae: From decks.proto:

  // card counts without children or limits applied
  uint32 intraday_learning = 9;
  uint32 interday_learning_uncapped = 10;
  uint32 new_uncapped = 11;
  uint32 review_uncapped = 12;

These would allow the user to see how many cards may be in that particular deck beyond the set limits

  uint32 total_in_deck = 13;
  uint32 total_including_children = 14;

These were so the user could see the entire number of cards in the deck, regardless of if they are due or not. Not sure it would make sense to apply limits there? The UI could show "total cards in this deck" and total_including_children-total as "total cards in child decks" for example. These two are not so much about answering the daily limits questions, but for things like when cards are due on a different day and users wonder where they have gone.

RumovZ: > The UI could show "total cards in this deck" and total_including_children-total as "total cards in child decks" for example. These two are not so much about answering the daily limits questions, but for things like when cards are due on a different day and users wonder where they have gone.

Ah, this also makes sense.

Arthur-Milchior:
Unless you need my input, would you mond not using the @ behind my name, so that I don’t get mentioned 

Thanks again for your work on anki and taking care of big strange collections

dae:
@Arthur-Milchior I tagged you because your collection has a float in an integer column, and I thought you might want to know about it so you could fix it.

rslib/src/storage/deck/mod.rs
```diff
@@ -198,13 +198,15 @@ impl SqliteStorage {
...
         let prefix_start = format!("{}\x1f", parent.name);
         let prefix_end = format!("{}\x20", parent.name);
         self.db
-            .prepare_cached(concat!(
+            .prepare_cached(&format!(
+                "{}{}{}",
                 include_str!("get_deck.sql"),
-                " where name >= ? and name < ?"
+                " where name >= ? and name < ?",
+                if sorted { " order by name" } else { "" }
```
dae, line 209: That ordering should be free due to the use of the index, and I'd expect you'd get it even if you didn't explicitly request it:

```
>>> pp(mw.col.db.all("explain query plan select id, name from decks where name >= 'foo' and name < 'foo1' order by name"))
[[3,
  0,
  0,
  'SEARCH decks USING COVERING INDEX idx_decks_name (name>? AND name<?)']]

>>> pp(mw.col.db.all("explain query plan select id, name from decks where name >= 'foo' and name < 'foo1'"))
[[2,
  0,
  0,
  'SEARCH decks USING COVERING INDEX idx_decks_name (name>? AND name<?)']]
```

It's good to be explicit about our intentions here to avoid a repeat of 9daa0c9acb3019c719a282c81888707b465bc9ce, but since this order should be free, I think we can probably omit the boolean and just always sort that way.

rslib/src/scheduler/queue/builder/mod.rs
```diff
@@ -270,3 +271,111 @@ impl Collection {
...
+        let cards = vec![
+            (grandchild.id, 0),
+            (grandchild.id, 1),
+            (child_2.id, 0),
+            (child_2.id, 1),
+            (child.id, 0),
+            (parent.id, 0),
+            (parent.id, 1),
+        ];
+        assert_eq!(col.queue_as_deck_and_template(parent.id), cards);
```
dae, line 377: Nice!

Arthur-Milchior:
Sorry, I didn't realize the implication of hours message. Indeed, that is strange. I can't even start to imagine how it occured

dae:
@RumovZ this code appears to be giving correct results now, and still seems to benchmark a little faster than the old code :tada:

rslib/src/storage/deck/mod.rs
RumovZ: > I'd expect you'd get it even if you didn't explicitly request it:

This doesn't seem to be always the case. I only made that change because the tree was malformed for certain decks.
But no objections to always return a sorted list, which matches the bahaviour of `get_all_deck_names()`.

RumovZ:
@dae there are at least two issues I still need to sort out:

- Slight discrepancies between the new counts in the deck list and the deck overview screen for certain decks.
- When gathering stops between two siblings. the wrong one may be gathered. I think the current implementation also suffers from this, but as usually siblings appear in the correct order, this is a rare edge case. However, due to the SQL sorting in the new implementation, this occurs regularly. I think this explains the inconsistently failing tests.

As soon as I've dealt with these, I'll post new benchmarks. By the way, I used `runopt` for the previous ones.

RumovZ:
I tried to reproduce your modifications to the collection as described above. Then I also disabled burying and changed the new card sort order to none (order gathered).

| order | old | new |
| --- | --: | --: |
| deck | 25 ms | 18 ms |
| pos asc | 30 ms | 75 ms |
| pos desc | 35 ms | 70 ms |

I think the code is largely bug free by now, but the ordering of siblings, and its implications for burying, need some more thought.

rslib/src/storage/card/mod.rs
```diff
@@ -249,7 +249,7 @@ impl super::SqliteStorage {
         F: FnMut(NewCard) -> bool,
     {
         let mut stmt = self.db.prepare_cached(&format!(
-            "{} ORDER BY due ASC",
+            "{} ORDER BY due, ord ASC",
```
RumovZ, line 252: I don't understand why this is required for the new test to pass. Shouldn't the cards (siblings) of the grandchild deck be in the right order anyway?

Either way, this also eliminates the nasty special case handled in `add_new_card()`. Is there any reason why the index isn't on (did, queue, due, ord)? Then this new query should be possible without any slowdown.

rslib/src/storage/card/mod.rs
```diff
@@ -266,16 +266,23 @@ impl super::SqliteStorage {
...
     where
         F: FnMut(NewCard) -> bool,
     {
         let mut stmt = self.db.prepare_cached(&format!(
-            "{} ORDER BY due {}, ord asc",
+            "{} ORDER BY {}",
             include_str!("active_new_cards.sql"),
-            if reverse { "desc" } else { "asc" }
+            if random {
+                "random()"
```
dae, line 280: Review card sorting uses a different approach:

            ReviewOrderSubclause::Random => "fnvhash(id, mod)",

This keeps things mostly random, but will give reproducible results when starting from the same state, which can be useful in debugging and the close+reopen cases. It would be nice to have the same behaviour here.

The old random insertion order+due gather order ensured siblings were introduced together in template order, whereas currently this code would show templates in a random order. We'll probably want to offer both options, but that presents us with a bit of a challenge:

- `order by fnvhash(nid), ord` would always show notes in the same order
- `order by fnvhash(nid, mod), ord` would separate the notes, as each of the cards will have a different mtime.
- using the mtime of the note would keep the notes together, but we don't have easy access to it, and notes change infrequently
- `fnvhash(nid, $daysElapsed), ord` would mean the position of a given note changes per day. This would have the disadvantage of showing new cards in the same order again if the user were to study new cards, reset them to new, and study them again. But in the absence of a better solution, it may be our best bet. WDYT?

dae: (you might also want to check out the enum use in that review sorting code, as it scales a bit better than separate boolean flags)

rslib/src/storage/card/mod.rs
dae: > I don't understand why this is required for the new test to pass. Shouldn't the cards (siblings) of the grandchild deck be in the right order anyway?

`select * from cards` should (but as far as I'm aware, is not guaranteed to) deliver cards in card id order, as it's walking the primary key id btree. In the common case, that will also deliver cards in template order, but that is not always the case (such as when some siblings are generated later). 

In this case though we're relying on a separate index, so presumably the cards are coming back in order that index happens to be stored in. In the case of identical due values, sqlite would be free to store them in either order, so that's probably where the unpredictability is coming from.
 
> Either way, this also eliminates the nasty special case handled in add_new_card(). Is there any reason why the index isn't on (did, queue, due, ord)? Then this new query should be possible without any slowdown.

Every extra indexed column adds a storage cost, and extra work when that row is updated. Anki 1 was a bit too liberal in its use of indexes, and I tried to reduce them in Anki 2 to the minimum possible.

Adding ord ASC here would allow us to fetch new cards in the default order faster, but we'd pay the extra column index cost for every card in the collection. We could perhaps work around it with a pair of conditional indexes (including ord if queue=0, and a separate index without ord where queue!=0), but I don't know what sort of performance implications that would entail. And unfortunately the index would only help in the default sort case - if the user wanted decreasing card positions for example, we either couldn't use the index, or we'd be giving the user decreasing ordinals as well.

dae:
I found the variability disappeared when running multiple times in succession; I'm guessing it's CPU powersaving related.

```
for i in range(20):
  mw.col.decks.select(1)
  mw.col.decks.select(1592002960463)
  mw.col.sched.get_queued_cards()
```

With latest code on an Linux/AMD machine:

| order | old | new |
| --- | --: | --: |
| deck | 10 ms | 18 ms |
| pos asc | 48 ms | 45 ms |
| pos desc | 56 ms | 46 ms |
| random | NA | 56 ms |

And on an ARM64 Mac:

| order | old | new |
| --- | --: | --: |
| deck | 10 ms | 20 ms |
| pos asc | 47 ms | 50 ms |
| random | NA | 60 ms |

So I still can't seem to be able to reproduce the difference with your baseline results. Are you on an Intel machine? Maybe something in our stack has some Intel-only optimisations.

In any case, I agree, these numbers look very promising. If we allow for a 10x speed decrease on slower devices, performance is still quite reasonable at these large counts, and smaller collections will of course perform better. I think the performance hit (even in your case) is small enough that it may make sense to be taking this approach for all sort orders except the deck order special case. This would allow us to merge the separate gather and sort orders into a single order that has fully correct burying, and simplify the code. WDYT?

dae:
(that would mean we'd lose the ability to sort after gathering in the deck order case, but I'm not sure whether that's something we really want to retain, as the difference between sorting before and after the gather can be confusing to users)

dae:
Curious if there was any room to shave time off the new limit handling in the deck order case, I added the following:

```diff
diff --git a/rslib/examples/perf.rs b/rslib/examples/perf.rs
new file mode 100644
index 000000000..a62490ce5
--- /dev/null
+++ b/rslib/examples/perf.rs
@@ -0,0 +1,12 @@
+use anki::{collection::CollectionBuilder, prelude::*};
+
+fn main() -> Result<()> {
+    let mut col =
+        CollectionBuilder::new("/home/dae/.local/share/Anki2/t1/collection.anki2").build()?;
+    for _i in 0..500 {
+        col.set_current_deck(DeckId(1))?;
+        col.set_current_deck(DeckId(1592002960463))?;
+        col.get_queued_cards(1, false)?;
+    }
+    Ok(())
+}
```

I generated a flamegraph with:

```
CARGO_PROFILE_RELEASE_DEBUG=true cargo flamegraph --example perf
```

(not sure how well it will work on Windows; I did this on a Linux machine)

That results in the following file, which you can interact with:

flamegraph

I don't see any easy wins there - there's nothing that jumps out as taking longer than it should.  :tada:

dae:
GitHub have re-encoded the file which broke the interactivity. This should work: https://apps.ankiweb.net/downloads/flamegraph.svg

rslib/src/storage/card/mod.rs
RumovZ: > In this case though we're relying on a separate index,

Without the `ord asc`? But I guess this is the same scenario as with the deck name sorting above. Maybe it's because it requires a tiny bit additional work to find the first entry fitting the `where` clause in the b-tree, whereas with arbitrary order SQLite just has to find any fitting entry and can expand from there.

Okay, pity about the extra sorting table, but that's probably not even noticeable in terms of performance.

rslib/src/storage/card/mod.rs
RumovZ: > fnvhash(nid, $daysElapsed), ord would mean the position of a given note changes per day. This would have the disadvantage of showing new cards in the same order again if the user were to study new cards, reset them to new, and study them again. But in the absence of a better solution, it may be our best bet. WDYT?

Makes sense. This user should use a filtered deck. 🙂

RumovZ:
> So I still can't seem to be able to reproduce the difference with your baseline results. Are you on an Intel machine? Maybe something in our stack has some Intel-only optimisations.

No, AMD as well. I don't have hardware acceleration, and it would be cool if the DB could make use of the GPU, but I can't imagine it does.

> This would allow us to merge the separate gather and sort orders into a single order that has fully correct burying, and simplify the code. WDYT?

Sounds good to me. 

The graph is awesome! It looks like the db operations dwarf any Rust optimisations we could make, as would be expected, so I won't make the logic any more complicated than necessary.

RumovZ:
How do you imagine the new settings exactly? I can't come up with a completely satisfying solution, but here is an idea: We could offer three different settings:

1. Order by deck (yes/no)
2. Sort order (ascending position, descending position, random)
3. Template order (group siblings, group template ordinals, random)

This would mean increasing the number of settings, but at least they should be easier to understand than the gathering/sorting distinction. Another less than ideal thing is that not all combinations are equally useful.

How does the insertion order come into play? Now with on-the-fly sorting, I guess it's lost a lot of its purpose. Would dropping the setting and only assigning ascending positions be an option?

We also lost the ability to gather all siblings, but have them spaced within a session. Should there be an option for this, or should this even be done implicitly?

dae:
Good point, losing sibling spacing would not be great, so we may need to retain the post-gather sorting in some form. Will gave these questions some thought and get back to you.

Re insertion order, we'll need to keep that around for now for users who are on the v2 scheduler, and some users may have v3 enabled but be using v2 on some of their devices. Maybe the best bet for now would be to add something to the tooltip explaining how adjusting the display order is recommended instead in v3?

dae:
I was a bit concerned about the performance implications of adding 'ord ASC' to the query in the deck-order case, but the performance impact does not seem to be as bad as I had initially feared, and this change solves the partially gathered sibling case you mentioned above, and allows us to drop that complicated ordering code, so I think it's a win :-)

Out of curiosity, I checked what sort of performance hit we'd be looking at if we did the deck order case in SQL, like how it is done for reviews:

| order | no bury | bury |
| --- | --: | --: |
| deck (current) | 18 ms | 39 ms |
| pos asc | 46 ms | 50 ms |
| deck then due (SQL) | 65ms | 73 ms |

Being able to treat the deck order case the same as the other cases might be nice, though we would be paying a performance hit for it, and I'm not sure it's worth it.

(also out of curiosity, I benchmarked the fnvhash change mentioned earlier - it's basically identical to random()).

On the topic of sorting and sibling shuffling, I retract my earlier comments about removing the post-gather sorting - I didn't think that one through enough. :-) 

- For users with burying disabled, I think spacing is the sensible default - showing a front->back and then back->front immediately after it is not very useful most of the time, and a post-gather sort seems the simplest way to accomplish that.
- Some users do want to be able to see siblings one after the other, as they have cards constructed in such a way that each sibling builds off the previous one (sometimes made with an add-on). I don't think this is an approach we really want to encourage, as the cards will not appear in that order during later review, but some users do want it.
- There are also some requests to see cards in random sibling order, or to be more specific, when they select random they expect the card templates to appear randomly as well. Your current random() approach (or an equiv. like fnvhash(id, mod)) would provide them that, but I think we'll also want a separate fnvhash(nid, $today) to retain the existing functionality.

Aside from burying/siblings, a post-gather sort is also useful for cases like "take the 5 oldest cards from each deck, then show them to me in a random order", or "take the 200 oldest cards, and show them to me in a random order".

> We could offer three different settings:

    Order by deck (yes/no)
    Sort order (ascending position, descending position, random)
    Template order (group siblings, group template ordinals, random)

> This would mean increasing the number of settings, but at least they should be easier to understand than the gathering/sorting distinction. Another less than ideal thing is that not all combinations are equally useful.

I am still kind of fond of "gather", as I feel it better conveys the fact that it's the order prior to a limit being applied. But whether that is clear to other people or just me, I'm not sure :-)

For "order by deck", maybe "Sort by deck first"?

The separate option does reduce the permutations we need to display, though it does make the performance messaging a little harder. Previously we could say the "deck" option is the fastest; now we need a qualifier like "deck first enabled, with ascending position selected in the next option, is the fastest"

Alternatively the first two options could remain a single option, where we list out the various combinations, eg

"Deck, then ascending position"
"Ascending position"
"Descending position"
"Random note"
"Random card"

We could potentially add "Deck, then descending/random note/random card" options as well, though I don't recall anyone asking for these in the past, so it could be a case of YAGNI.

Because the post-gather sort is not just about siblings/templates, maybe something like "Post-gather order" or "Post-gather sort" might be clearer than "Template order"?

Currently we list out quite a few options in "New card sort order", and it makes for some combinations people are unlikely to use, such as gathering in ascending order then sorting in descending order. And the sort orders partly repeat the ordering we're doing in SQL. I like your idea of reducing the available options there. Presumably we could implement these as a stable sort like we do in the current TemplateOnly case?

For "Group siblings", were you thinking of that as a no-op, and relying on the implicit ord ASC in the gather stage? There'd be the corner case of collections where siblings don't have the same due# - users could potentially report a "bug" that the siblings are not grouped in that case. Whether that's worth a more complicated sort, or just a change of wording to something like "Order gathered", I don't know.

We might need a few more permutations though, eg a full list like

- Group siblings
- Group template ordinals (maybe "Separate siblings" or "Card template order" instead?)
- Random note, then group siblings
- Random note, then card template
- Random note, then random card (probably superfluous with the option below)
- Random card

WDYT?

RumovZ:
> For "Group siblings", were you thinking of that as a no-op, and relying on the implicit ord ASC in the gather stage?

Yes. Maybe the label says "Order gathered", but there is a tooltip somewhere that this will usually mean consecutive siblings?

I think that covers the expectable use cases, however there are again a few superfluous combinations. Here are your suggestions numbered and without the second to last one:

1.  Order gathered (Group siblings)
2.  Group template ordinals (maybe "Separate siblings" or "Card template order" instead?)
3.  Random note, then group siblings
4.  Random note, then card template
5.  Random card

They all make sense when gathering by (deck and) position. But for "random note" only 1, 2 and 5 do, whereas for "random card", only 1 does. Users would probably waste time trying to figure out what combination gives them the desired result, when there is no perceptible difference, and we would have to compare gather and sort orders to avoid redundant sorting.
I feel like we don't get around conditional settings. So for a given choice of gather order, some options for sort order are greyed out.

dae:
For the sake of discussion, here's an approximation of what that would look like if we combined the permutations into a single dropdown:

- Deck, then ascending position (siblings apart)
- Deck, then ascending position (siblings together)
- Deck, then ascending position (shuffled notes, siblings together)
- Deck, then ascending position (shuffled notes, siblings apart)
- Deck, then ascending position (shuffled cards)
- Ascending position (siblings apart)
- Ascending position (siblings together)
- Ascending position (shuffled notes, siblings together)
- Ascending position (shuffled notes, siblings apart)
- Ascending position (shuffled cards)
- Descending position (siblings apart)
- Descending position (siblings together)
- Descending position (shuffled notes, siblings together)
- Descending position (shuffled notes, siblings apart)
- Descending position (shuffled cards)
- Random notes, siblings together
- Random notes, siblings apart
- Random notes, shuffled cards
- Random cards

It's a bit unwieldy :-) I think your suggestion is better. Can we grey out individual items in a dropdown? If not, perhaps we could change the wording. We could also change the number of elements dynamically, though that makes the UI code a bit more complicated.

RumovZ:
Wasn't that hard to achieve fortunately.
I also allowed sorting by template in the random card case, because on second thought spacing siblings may be a good idea there as well. It's also nice to have the default sort order available regardless of the gather order.
Currently, the mapping from the pb enums to the TS arrays relies on the indices of the variants, which seems a bit fragile to me. But I don't what would be the most idiomatic solution here.

This PR is starting to feel mature! Are there other things left to do or benchmark than what's listed below?

- [ ] Remove queue building timer.
- [ ] Update deck option tooltips.

For the tooltips, I guess the only option is to create new ftl strings and have the translators start from scratch?

By the way, I had also benchmarked different new limits, and as you predicted 999 or 9999 didn't really make a difference.

proto/anki/deckconfig.proto
```diff
@@ -32,22 +32,33 @@ message DeckConfig {
...
+      // Ascending card template ordinal.
+      // For a given ordinal, cards appear in random order.
       NEW_CARD_SORT_ORDER_TEMPLATE_THEN_RANDOM = 2;
-      NEW_CARD_SORT_ORDER_LOWEST_POSITION = 3;
-      NEW_CARD_SORT_ORDER_HIGHEST_POSITION = 4;
-      NEW_CARD_SORT_ORDER_RANDOM = 5;
-      // Sorts by template, preserving original gather order.
-      NEW_CARD_SORT_ORDER_TEMPLATE_ONLY = 6;
-      // Preserves original gather order (eg deck order)
-      NEW_CARD_SORT_ORDER_NO_SORT = 7;
```
dae, line 49: If the user has selected something like option 7 in a previous Anki version, the selection box will display as blank in our deck options.  One way to work around it would be something like the following inside the $: { ... } of DisplayOrder.svelte, after making the changes mentioned in the previous comment:

        if (!($config.newCardSortOrder in SortOrder)) {
            $config.newCardSortOrder = 0;
        }

dae: (we'll probably at least also want to add a note to the change notes directing users to check their options after updating)

rslib/src/storage/card/mod.rs
```diff
@@ -619,6 +642,33 @@ fn review_order_sql(order: ReviewCardOrder) -> String {
...
+    RandomCards(u32),
+}
+
+impl NewCardSorting {
+    fn write(self) -> String {
+        match self {
+            NewCardSorting::LowestPosition => "due ASC, ord ASC".to_string(),
+            NewCardSorting::HighestPosition => "due DESC, ord ASC".to_string(),
+            NewCardSorting::RandomNotes(salt) => format!("fnvhash(nid, {salt}), ord ASC"),
+            NewCardSorting::RandomCards(salt) => format!("fnvhash(cid, {salt})"),
```
dae, line 667:
```suggestion
            NewCardSorting::RandomCards(salt) => format!("fnvhash(id, {salt})"),
```

dae:
Looking good Rumo! Regarding the tooltips:

- for insertion order, we could preserve the existing one, and add a new string like "With the V3 scheduler, leaving this set to sequential, and adjusting the order in the Display Order section is recommended." Then we could show it as a yellow warning when random is selected, which will be more discoverable.
- for gather and sort order, we'll need new strings. How about something like this?

```
deck-config-new-gather-priority-tooltip =
    `Deck`: gathers cards from each deck in order, starting from the top. Cards from each deck are
    gathered in ascending position. If the daily limit of the selected deck is reached, gathering
    may stop before all decks have been checked. This order is fastest in large collections, and
    allows you to prioritize subdecks that are closer to the top.
    
    `Ascending Position`: gathers cards by ascending position (due #), which is typically
    the oldest-added first.

    `Descending Position`: gathers cards by descending position (due #), which is typically
    the latest-added first.

    `Random Notes`: gathers cards of randomly selected notes. When sibling burying is
    disabled, this allows all cards of a note to be seen in a session (eg. both a front->back
    and back->front card)

    `Random Cards`: gathers cards completely randomly.

deck-config-new-card-sort-order-tooltip =
    `Card Template`: Displays cards in card template order. If you have sibling burying
    disabled, this will ensure all front->back cards are seen before any back->front cards.

    `Order Gathered`: Shows cards as they were gathered. If sibling burying is disabled,
    this will typically result in all cards of a note being seen one after the other.

    `Card Template, then Random`: Like `Card Template`, but shuffles the cards of each
    template. This can be useful for showing the oldest 20 cards in a random order for example.

    `Random Note, then Card Template`: Picks notes at random, then shows all of their siblings
    in order.

    `Random`: Fully shuffles the gathered cards.
```

- I dropped the ", then order gathered" part, since it felt a bit superfluous
 
WDYT?

ts/deck-options/DisplayOrder.svelte
```diff
@@ -43,6 +43,32 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+                disabledNewSortOrders = [2, 3, 4];
+                break;
+            default:
+                disabledNewSortOrders = [];
+                break;
+        }
+        if (disabledNewSortOrders.includes($config.newCardSortOrder)) {
+            $config.newCardSortOrder = 0;
+        }
+    }
```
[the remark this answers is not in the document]
RumovZ: Sorry, I should have been clearer. What gives me trouble are the choices arrays. I'd like to associate every variant with exactly one string, and get a type error if one variant is missing (or a non-exisiting one referenced of course). I did a bit of research and there seems to be the [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) type, but it doesn't seem to work with enums in our TS version.

Makes sense to still use enums directly where possible, of course, and I'll commit your diff shortly.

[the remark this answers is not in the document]
dae: Gotcha. I imagine it might be possible if we switched the choices array into a record, and used something like keyof. @hgiesel might know more. No need to hold up this PR over it though. Run out of time tonight, but will give this a final review tomorrow morning. :-)

RumovZ:
I thought a bit about the tooltips, but I think they look good. Maybe we can put some illustrative examples in the manual.

dae:
Looks good Rumo. Thank you for all your work on this - it has solved a number of pain points, and I imagine this is going to make quite a few users happy!

PR #1639, opened 2022-02-02
Fix field not restored if active

### LINKED ISSUE #1628: editor does not refresh active field after note undone

hgiesel:
Fixes #1628, alternative for #1636.

@hikaru-y Sorry for picking up all those issues after you already started working on them. I've never had notifications turned for open issues on the anki repo, which I just turned on. As I just wrote the editor code, I know when something is more of a bug of an existing function (in this case `focusField`), rather than a missing feature.

### DIFF
--- a/ts/editor/EditingArea.svelte
+++ b/ts/editor/EditingArea.svelte
@@ -9,10 +9,17 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     export interface EditingInputAPI {
         readonly name: string;
-        focus(): void;
-        refocus(): void;
         focusable: boolean;
-        moveCaretToEnd(): void;
+        /**
+         * The reaction to a user-initiated focus, e.g. by clicking on the
+         * handle, or clicking Tab.
+         */
+        focus(): void;
+        /**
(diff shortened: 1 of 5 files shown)


ts/editor/EditingArea.svelte
```diff
@@ -8,10 +8,17 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
         readonly name: string;
+        focusable: boolean;
+        /**
+         * The reaction to a user-initiated focus, e.g. by clicking on the
+         * handle, or clicking Tab.
+         */
         focus(): void;
+        /**
+         * Behaves similar to a refresh, e.g. sync with content, put the caret
+         * into a neutral position, a/o undo selections.
```
dae, line 19: a/o?

dae:
@hikaru-y sorry for the duplicated effort here - any objections to me merging this approach?

hikaru-y:
@hgiesel @dae No worries!

@dae No objections from me.

PR #1642, opened 2022-02-06
Add forget card to review context "more" menu

aracelix:
I hope I did this correctly. Development notes say that there is a transition from python to rust for the business logic - forgetting cards seems like business logic...what I did in this PR is reference `forget_cards` from `scheduling.py` in the context menu:

https://github.com/ankitects/anki/blob/0c81bbce04c1998b92ed0c1fa08d1cd5d13f3f56/qt/aqt/operations/scheduling.py#L66-L75

Which looks like it eventually calls the rust backend?
https://github.com/ankitects/anki/blob/0c81bbce04c1998b92ed0c1fa08d1cd5d13f3f56/pylib/anki/scheduler/base.py#L162-L164

Re: tests, it seemed like there were tests for `forget_cards` already, but didn't see anything specifically for the review view. I did check that clicking the menu and using the shortcut both work to forget a card. Let me know if that's not enough/expected.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -92,6 +92,7 @@ Vova Selin <vselin12@gmail.com>
 qxo <49526356@qq.com>
 Spooghetti420 <github.com/spooghetti420>
 Danish Prakash <github.com/danishprakash>
+Araceli Yanez <github.com/aracelix>
 
 ********************
 
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -30,6 +30,7 @@ from aqt.operations.scheduling import (
     answer_card,
     bury_cards,
     bury_notes,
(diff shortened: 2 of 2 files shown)


dae:
Thanks Araceli, that's exactly what I would have done, and as this is just hooking the GUI up to existing functionality, I don't think we need a test here. Most of the biz logic is implemented in Rust at this point, with import/export the last major remaining hurdle.

zcorley24:
I'm sorry... but how in the world would do i get this to work? I've been looking for something like this forever!

I have python and VSCode but am still learning


Also thank you so much, even if i'm not able to use it i think it's so cool you have the ability to do this and also post it for everyone!

dae:
If you don't see it in your Anki, maybe you need to update it to the latest version.

Issue #987, opened 2020-12-24
Avoid exec_() in studydeck.py


dae:
https://forums.ankiweb.net/t/intermittent-hanging-on-add-screen-when-attempting-to-change-note-type/6003 reports that the screen is hanging, and I think there's a decent chance it's caused by the call to self.exec_(), which has been the source of crashes in the past. Taking a different approach by calling .setWindowModality() to make the window modal, and removing the .exec_() line may fix the issue. But as all users of StudyDeck currently expect .name to be set by the time the constructor finishes, the code will need to be refactored to accept a callback that is called once the dialog is closed.

charltonaustin:
Hi @dae  and team. Is anyone working on this? I was going to take a look. I'd like to get more involved in the project and though this might be a good place to start. Please let me know if there is somewhere else I should be looking.

charltonaustin:
It looks likes the window has been set to modal already here, `qt/aqt/studydeck.py:83`, and that the success call back is run in onAddDeck here, `qt/aqt/studydeck.py:179`. Is that correct? Should this be closed?

dae:
The .exec() call is still there - it was renamed from .exec_() since this issue was originally posted. This might not be the best ticket to get started with I'm afraid - all existing callers will need to be updated to handle a callback, and it will likely break some add-ons that rely on the current behaviour, so it's something we probably won't want to merge in immediately.

charltonaustin:
Sounds good. Do you have a suggest for a ticket that would be better suited to start with?

dae:
Nothing springs immediately to mind I'm afraid, but if you follow along on the forums, I imagine something will come up.

roxgib:
The use of exec in QDialogs and QMenus is causing issues with an [add-on I'm developing](https://github.com/roxgib/anki-contanki), so I've tried to develop a fix. If it's accepted I'll implement it for other QDialogs and QMenus.

PR #1643, opened 2022-02-06
Add callback and non-blocking option to StudyDeck

roxgib:
### Background

QDialog has a few different ways of displaying itself, including:

```
QDialog.exec() # runs in own loop, blocks program until finished
QDialog.open() # non-blocking
```

 The blocking version is useful if your function needs to get a value out of QDialog, but this causes issues because:
- If QDialog causes an error, it will also cause an error in the function that called it (or simply hang)
- Other code might need to run in the background, e.g. an add-on that makes a selection from the dialog

A similar problem exists with QMenu, and ideally all subclasses of QDialog and QMenu will be changed to their respective non-blocking options, but StudyDeck is known to be causing issues (see #987) already. However, since existing code (and likely add-ons as well) relies on QDialogs being blocking, this will cause issues if the behaviour is changed.

### Fix

A `callback` kwarg is added to StudyDeck, accepting a callable and defaulting to `None`, and the class is modified so that:
- if `callback == None` the class retains its old, blocking, behaviour. As this is the default, existing code that calls StudyDeck will be unaffected. 
- if a callable is passed to callback, the callable is called with the StudyDeck instance as its only argument by `StudyDeck.accept`. New code that requires the result of StudyDeck should use this behaviour.

This option allows existing code (and addons) to be updated while preserving legacy behaviour in the interim. Once the new behaviour is confirmed to be suitable, a depreciation warning can be added to emit when values of the instance (usually `StudyDeck.name`) are referenced after the dialog has been closed to warn add-ons of the need to update, and eventually the blocking option can be removed.

Changes have been tested in MacOS (Apple, PyQt6) and StudyDeck is confirmed to behave normally when called using either approach.

If this approach is successful for StudyDeck, the same changes should be made to other objects that block the main loop.

### DIFF
--- a/qt/aqt/studydeck.py
+++ b/qt/aqt/studydeck.py
@@ -36,7 +36,7 @@ class StudyDeck(QDialog):
         dyn: bool = False,
         buttons: Optional[list[Union[str, QPushButton]]] = None,
         geomKey: str = "default",
-        callback: Union[Callable, bool] = False,
+        callback: Callable = None,
     ) -> None:
         QDialog.__init__(self, parent or mw)
         self.mw = mw
--- a/qt/aqt/studydeck.py
+++ b/qt/aqt/studydeck.py
@@ -36,7 +36,7 @@ class StudyDeck(QDialog):
         dyn: bool = False,
         buttons: Optional[list[Union[str, QPushButton]]] = None,
         geomKey: str = "default",
-        callback: Callable = None,
+        callback: Callable | None = None,
     ) -> None:
         QDialog.__init__(self, parent or mw)
         self.mw = mw
--- a/qt/aqt/studydeck.py
+++ b/qt/aqt/studydeck.py
@@ -36,7 +36,7 @@ class StudyDeck(QDialog):
         dyn: bool = False,
         buttons: Optional[list[Union[str, QPushButton]]] = None,
         geomKey: str = "default",
-        callback: Callable | None = None,
+        callback: Union[Callable, None] = None,
     ) -> None:
         QDialog.__init__(self, parent or mw)
         self.mw = mw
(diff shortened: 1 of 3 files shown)


roxgib:
For an example of how the use of `QDialog.exec()` affects add-ons, see [here](https://github.com/roxgib/anki-contanki/issues/1).

qt/aqt/studydeck.py
```diff
@@ -36,6 +36,7 @@ def __init__(
         dyn: bool = False,
         buttons: Optional[list[Union[str, QPushButton]]] = None,
         geomKey: str = "default",
+        callback: Callable = None,
```
dae, line 39:
```suggestion
        callback: Callable | None = None,
```

qt/aqt/studydeck.py
```diff
@@ -152,6 +157,8 @@ def accept(self) -> None:
             showInfo(tr.decks_please_select_something())
             return
         self.name = self.names[self.form.list.currentRow()]
+        if callable(self.callback):
```
dae, line 160: A minor point, but you've tested for truthiness above, so you could probably do so here to be consistent.

dae (commented):
Thanks Sam, this looks reasonable, and I appreciate that you've thought about compatibility with existing add-ons. Some minor points below:

dae:
Popping this at the top of the file below the top comments should fix that CI error:

```
```

roxgib:
I already changed it to Union before I saw you comment, but I'll keep that in mind for the future. 

It's not quite clear to me why the None type hint is required for this line but not e.g. this one...

https://github.com/roxgib/anki/blob/3439483e26511bd036ed681a6c56afaaad1e203c/qt/aqt/studydeck.py#L29

...but I'll read up on type hints more tomorrow

qt/aqt/studydeck.py
roxgib: Updated as suggested

dae:
That line would ideally use a union too, but it wasn't part of your change, and I didn't notice it :-) mypy isn't warning us when we forget it, as we have a bunch of legacy code that will need fixing first. Thanks for the fixes and the contribution!

PR #1650, opened 2022-02-08
Make dialogs non-blocking

roxgib:
This makes two main changes:
- implements the callback option for StudyDeck for the remaining callers from #987 
- switches CustomStudy, DeckConf, Models dialogs to be non-blocking via open()

Notes:
- open() causes the dialog to be modal, preventing interaction with the rest of the program, but is non-blocking, whereas exec() blocks the main loop until the dialog exits
- self.setWindowModality(Qt.WindowModality.WindowModal) has been removed since open makes the dialog modal
- There are some other dialogs to be changed to non-blocking, and the QMenus, but these three are fair straightforward as they are unlikely to be called in a way that expects them to be blocking in the same way as StudyDeck
- As discussed in #987 the use of blocking dialogs can cause issues with add-ons that rely on being able to run while the dialog is open, perhaps because they implement a method to interact with the UI and therefore the dialog. The use of exec has also been identified as causing other problems
- The possibility that an add-on relies on being blocked while a dialog is open can't be ruled out other than when trying to get a return value, but it seems unlikely for these dialogs

### DIFF
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -673,6 +673,13 @@ class Browser(QMainWindow):
         cids = self.table.get_selected_card_ids()
         did = self.mw.col.db.scalar("select did from cards where id = ?", cids[0])
         current = self.mw.col.decks.get(did)["name"]
+
+        def callback(ret: StudyDeck) -> None:
+            if not ret.name:
+                return
+            did = self.col.decks.id(ret.name)
+            set_card_deck(parent=self, card_ids=cids, deck_id=did).run_in_background()
+
         ret = StudyDeck(
             self.mw,
             current=current,
@@ -680,12 +687,8 @@ class Browser(QMainWindow):
             title=tr.browsing_change_deck(),
             help=HelpPage.BROWSING,
             parent=self,
+            callback=callback,
         )
-        if not ret.name:
-            return
-        did = self.col.decks.id(ret.name)
-
-        set_card_deck(parent=self, card_ids=cids, deck_id=did).run_in_background()
 
     # legacy
 
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -680,7 +680,7 @@ class Browser(QMainWindow):
             did = self.col.decks.id(ret.name)
             set_card_deck(parent=self, card_ids=cids, deck_id=did).run_in_background()
 
-        ret = StudyDeck(
+        StudyDeck(
             self.mw,
             current=current,
             accept=tr.browsing_move_cards(),
--- a/qt/aqt/customstudy.py
+++ b/qt/aqt/customstudy.py
@@ -32,10 +32,9 @@ class CustomStudy(QDialog):
         self.created_custom_study = False
         f.setupUi(self)
         disable_help_button(self)
-        self.setWindowModality(Qt.WindowModality.WindowModal)
         self.setupSignals()
         f.radioNew.click()
-        self.exec()
+        self.open()
 
     def setupSignals(self) -> None:
         f = self.form
--- a/qt/aqt/deckchooser.py
+++ b/qt/aqt/deckchooser.py
@@ -91,7 +91,17 @@ class DeckChooser(QHBoxLayout):
         from aqt.studydeck import StudyDeck
 
         current = self.selected_deck_name()
-        ret = StudyDeck(
+
+        def callback(ret: StudyDeck) -> None:
+            if not ret.name:
+                return
+            new_selected_deck_id = self.mw.col.decks.by_name(ret.name)["id"]
+            if self.selected_deck_id != new_selected_deck_id:
+                self.selected_deck_id = new_selected_deck_id
+                if func := self.on_deck_changed:
+                    func(new_selected_deck_id)
+
+        StudyDeck(
             self.mw,
             current=current,
             accept=tr.actions_choose(),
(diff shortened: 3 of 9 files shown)


roxgib:
While I'm looking at it, should StudyDeck not be called with `cancel=True` in `DeckChooser.choose_deck`? In the add cards dialog the dialog for choosing the note type has a cancel button, but the dialog for choosing a deck doesn't. 
https://github.com/ankitects/anki/blob/30427fb540f9d48a371c00ee7ddbc008954936c1/qt/aqt/deckchooser.py#L100

dae:
Thanks Sam, looks good, and I was not aware of .open().

It's probably fine to add a cancel to the deck selector too. IIRC, the reason it originally didn't have one is because the current deck was changed immediately, so we would have had extra code to undo that change.

I did notice one change in behaviour in this PR - the notetype selector was previously modal, but it now isn't. It my not cause any issues, but keeping it modal would be the conservative choice.

roxgib:
The note type selector is still modal for me (at least on the add cards dialog, where are you looking at it?). If it's different for you it could be a platform difference (I'm on MacOs). 

Regardless, perhaps this line is the culprit:
https://github.com/ankitects/anki/blob/30427fb540f9d48a371c00ee7ddbc008954936c1/qt/aqt/models.py#L51

I've removed the last argument in my latest commit - can you check on your platform if it's now modal?

Since parent is optional I also updated the second arg to match the line in deckchooser

dae:
.exec() is effectively ApplicationModal, so the user can't interact with the main window while the notetype selector is visible. With this change, the dialog is WindowModal, so it blocks the add screen, but the main window is still accessible.

roxgib:
Gotcha - I was just seeing it as modal with the add screen.

It doesn't seem to be possible to make it application model without using .exec() - `self.setWindowModality(Qt.WindowModality.ApplicationModal)` has no effect when using .open(). I can have a longer look tomorrow, and of course it's probably possible to directly disable interaction with the main window manually while the dialog is open.

However, I'd still argue to go ahead with the change, because:
- Currently the notetype selector is ApplicationModal, but the DeckSelector is WindowModal, because the DeckSelector has `self.setWindowModality(Qt.WindowModality.WindowModal)`. These should behave identically, since they sit next to each other in the add cards dialog and otherwise function quite similarly
- It's probably not unusual for users to have the add cards dialog open and still want to interact with other windows, particularly the browser
- Having other windows unresponsive while a modal dialog is open inside a separate window that is otherwise unblocking is unintuitive, and it probably won't be clear to users why the main window isn't responding
- Poking around a bit, it seems interactions with the main window aren't actually prevented, just delayed until the dialog is closed, at which point they all fire at once (this is the case in both 49 and 50), which seems a bit dangerous as many users' instincts are to start clicking around to get it going again
- It doesn't block all interactions, and this causes some strange behaviour - it's still possible to open the browser with b, but not to close it, and possible to select rows or items from the side bar, but the effects of doing so are delayed 
- The disadvantages of .exec() that have already been discussed
- The Qt docs [recommend against the use of .exec()](https://doc.qt.io/qt-5/qdialog.html#exec) where possible, pointing to bugs 

That said, reversing the changes to NotetypeChooser so that it uses the old behaviour of StudyDeck (without a callback) is all that's required to make it ApplicationModal again, so if you're sure it should be ApplicationModal I can reverse that particular change with ease.

dae:
Does a .show() instead of an open() work with application modal?

roxgib:
It seems that:
- `self.setModal(True)` with `self.show()` results in:
    👍 Input is blocked to other windows
    👍 No weird behaviour with e.g. the browser or queuing inputs
    👍 Add ons still operate
    👎 The dialog displays as non-modal
- `self.setModal(True)` with `self.open()` results in:
    👎 Input is not blocked to other windows 
    👍 The dialog displays as modal
    👍 Add ons still operate
- `self.exec()` results in:
    👍 Most (but not all) input is blocked to other windows
    👍 The dialog displays as modal
    👎 Weird behaviour with e.g. the browser and queuing inputs
    👎 Add ons are blocked or queue actions

I'll play around some more and get back to you - I think that the first option should be working, but something might be tripping it up.

dae:
If by "displays as non-modal", you mean that the window is not dropping down as a sheet of the parent window, that behaviour will only happen with windowmodal AFAIK, and it's a distinction only seen on macOS. I'm not too concerned about that, and happy to merge this in as-is if it works for you.

roxgib:
Yep, sheet is what I mean. Perhaps I've been conflating the two until now since they often occur together. 

It turns out to be possible to make it a sheet and still ApplicationModal by calling the following when the dialog opens:
```
self.parent().window().setWindowModality(Qt.WindowModality.ApplicationModal)
self.parent().window().hide()
self.parent().window().show()
```
And then when the dialog closes:
```
self.parent().window().hide()
self.parent().window().setWindowModality(Qt.WindowModality.NonModal)
self.parent().window().show()
```
Which sets the parent window to ApplicationModel, blocking access to other windows, and unsets it when the dialog is closed. This doesn't strike me as a good solution and would require a lot more testing, so **I'm happy for you to go ahead and merge what's there now**.

I also think the main window should fade out when it's being blocked (and I think there's code to achieve that already) but I'll test that out and consider submitting a separate PR.

Thanks for your patience, I'm still getting familiar with Qt and GUI programming in general isn't my strong suit.

dae:
Not a problem, thank you for your work on this. :rocket: 

- Agreed on that code not being a good solution, as it could potentially break in a Qt update.
- The current fading behaviour is done when changes have been made that will require a UI refresh - it's there to indicate that the displayed content is potentially stale. I'm not sure whether it would be worth trying to reuse it for modal windows, as the approach wouldn't work outside of the webviews, and it would mean the fading code would have to keep track of more state.

PR #1651, opened 2022-02-08
Intersperse more evenly, mainly at start and end

### LINKED ISSUE #1645: could new/review card interspersing be improved?
https://forums.ankiweb.net/t/mix-new-with-reviews-new-card-after-one-review/17398

The current behaviour is more an artifact of the implementation than a deliberate choice.

```rust
        let a = &[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
        let b = &[111, 222, 333];
```

currently gives us

```
scheduler::queue::builder::intersperser::test::interspersing' panicked at
  left: `[1, 111, 2, 3, 4, 5, 6, 222, 7, 8, 9, 10, 11, 333, 12, 13, 14, 15]
```

RumovZ:
Closes #1645.

I actually liked the old behaviour, because the earlier the last new card is shown, the more review cards remain to space out its learning steps. But this new formula may feel a bit more balanced.

### DIFF
--- a/rslib/src/scheduler/queue/builder/intersperser.rs
+++ b/rslib/src/scheduler/queue/builder/intersperser.rs
@@ -23,7 +23,7 @@ where
     pub fn new(one: I, two: I2) -> Self {
         let one_len = one.len();
         let two_len = two.len();
-        let ratio = one_len as f32 / two_len as f32;
+        let ratio = (one_len + 1) as f32 / (two_len + 1) as f32;
         Intersperser {
             one,
             two,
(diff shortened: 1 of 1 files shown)


dae:
Thanks for the quick fix! I think we should probably aim to what is intuitive by default, but we could possibly consider a separate option in the future that biases the introduction of new cards more towards the front.

RumovZ:
To be fair, there already is the "show new before reviews" option. Maybe I'll just use that.

PR #1653, opened 2022-02-09
Allow dropping external content on the HTML editor

hgiesel:
Inspired by https://forums.ankiweb.net/t/new-html-editor-works-very-inconvenient-for-adding-pic-addresses/16730/22.


dae:
I may have misread what they said, but I was under the impression they had only been trying to use the HTML editor to work around the fact that drag&drop in a normal field wasn't working. I suspect it's a Linux/Qt issue - dropping works for me in Xwayland, but not when running Anki with native Wayland.

Currently none of our other functionality like bold/italics/mathjax etc work when the HTML editor is focused, so this change would make drag and drop different to everything else, including regular copy+paste., and that may lead to future questions when users are surprised some things work there and some don't. WDYT?

hgiesel:
Upon reading his reply, I think we can conclude, that he used the DnD on the old HTML editor, because it inserted the online link, rather than downloading and putting it into the media folder.

> none of our other functionality like bold/italics/mathjax etc work

I don't think that quite applies, because you're still able to make something bold, you just have to type `<strong>`, same with mathjax, you can create Mathjax by typing `<anki-mathjax>`, whereas the insertion of an image does not work any way whatsoever.

However I don't feel that strong about it.

dae:
(same behaviour after rolling back webengine)

hgiesel:
I'll close this as "probably not worth it"

PR #1655, opened 2022-02-10
Remove unused actions from ui files

RumovZ:
Looks like the add dialog was mostly unused actions. 😅
I also found a few elsewhere. Removing them should help to detect unused ftl strings.


dae:
Thanks for that :-) Did you write a script to detect the other ones? If so, maybe we should think about throwing it in the tools folder to help clean things up in the future too. If not, no need to create one :-)

RumovZ:
No, I just used the designer, which shows a list of actions and whether they're used. I think the three main windows are the only files that need checking anyway, and there shouldn't be a lot of obsolete actions, unless someone carelessly copy pastes a new main window. 🙃

PR #1660, opened 2022-02-11
Add a default logging configuration for addons

telotortium:
It would be good for addons that log debug messages using the Python
`logging` module to have a good default configuration.

TODOs:
- [ ] Build and test Anki using this (currently I'm having issues with Bazel
   on my machine).
- [ ] Confirm that these defaults make sense.


dae:
Would you mind elaborating on your use case a bit? As Anki is a GUI program, the console output tends to only be viewed by developers, and tends to show warnings that are relevant to what you're currently doing. I'm not sure timestamps make much sense in this context, as one does not typically go and revisit the history of the console, and the history is lost when the terminal is closed.

If you're writing an add-on that needs to store logs for later reference, logging to a specific file might be a better choice, and it perhaps makes more sense to do that for each add-on that needs it, instead of having a global log stored somewhere. WDYT?

@glutanimate I seem to recall you've used logging in your add-ons before; do you have any thoughts?

dae:
It seems nobody has any thoughts :-)

PR #1661, opened 2022-02-12
Round calculated Hard days

RumovZ:
https://forums.ankiweb.net/t/change-hard-interval-on-new-cards-with-the-custom-scheduler/17290/7

### DIFF
--- a/rslib/src/scheduler/states/steps.rs
+++ b/rslib/src/scheduler/states/steps.rs
@@ -80,7 +80,8 @@ impl<'a> LearningSteps<'a> {
     }
 }
 
-/// If the given interval in seconds surpasses 1 day, rounds it in days.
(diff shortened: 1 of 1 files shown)


rslib/src/scheduler/states/steps.rs
```diff
@@ -79,6 +80,16 @@ impl<'a> LearningSteps<'a> {
     }
 }
 
+/// If the given interval in seconds surpasses 1 day, rounds it in days.
```
dae, line 83:
```suggestion
/// If the given interval in seconds surpasses 1 day, rounds it to a whole number of days.
/// Ensures that the user gets the same results earlier and later in the day.
```

dae:
Seems like a good solution - thanks Rumo!

PR #1662, opened 2022-02-12
Submodules

### LINKED ISSUE #1631: VS Code requires us to be more explicit with submodule imports
Take the following example code:

```python
import aqt

def myfunc(mw: aqt.main.AnkiQt) → None:
  mw.col.reopen()
```

PyCharm is able to provide completions and jump-to-definition support on the mw object, but Pyright (and by extension, VS Code's Python support) can't handle it. If the import is instead changed to `import aqt.main` so that it imports the submodule directly, instead of relying on a reference in the parent module, the code works correctly in VS Code. Pyright's approach is the more principled one, but unfortunately a fair bit of Anki's existing code relies on this pattern.

Perhaps it is worth writing a script to fix up our imports, to improve the editing story in VS Code.

RumovZ:
Should close #1631.
(I assume you don't have any use for my half-baked Rust script.)

### DIFF
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -7,6 +7,8 @@ import pprint
 import time
 
 import anki  # pylint: disable=unused-import
+import anki.collection
+import anki.notes
 from anki import cards_pb2, hooks
 from anki._legacy import DeprecatedNamesMixin, deprecated
 from anki.consts import *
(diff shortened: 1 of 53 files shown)


dae:
That was fast - thank you! If you were planning to delete the script, it might be worth stashing somewhere just in case we ever need it in the future, but probably not worth committing to the repo (could just inline it in this comment section if you didn't want to store it locally)

RumovZ:
No, I didn't plan on deleting it. The file walking and I/O structure alone is a useful snippet to have at hand.

PR #1664, opened 2022-02-13
Show buried until daily limits in overview screen

### LINKED ISSUE #1633: Decks configured to 'Bury review siblings until next day' show buried cards in deck card count
A deck which has a pair of sibling cards shows both cards in the deck list:
image
Even when it has its options set to not show both of them (if they are new or reviewing):
image
Though they will not show when opening the deck or going through the cards:
image

I have testing and this bug happens on both Anki for Desktop version 2.1.49 and Anki for iOS 2.0.83. I have not tested any other platforms or versions. This bug happens with both 'Bury new siblings until the next day' and 'Bury review siblings until the next day' (when the cards satifsy one of the triggers).

If the deck card overview screen (figure 1) used the same buried sibling management code as the opened deck card overview screen (figure 2), then the bug would not occur.

I would also like to ask whether cards in the 'learning' phase follow the Bury new siblings until the next day' or 'Bury review siblings until the next day' option, or whether a 'Bury learning siblings until the next day' option is needed, though a new issue may be better suited for this despite the fact that it will affect the solution of this bug.

RumovZ:
Closes #1633.
What you wrote elsewhere made a lot of sense. What do you think about the graphical implementation? I think it's discreet and the familiar bury colour, but it can't be called pretty.
Ideally, there would be a tooltip or something explaining what the numbers actually mean, but I guess it's only a stopgap anyway.

### DIFF
--- a/qt/aqt/overview.py
+++ b/qt/aqt/overview.py
@@ -228,27 +228,31 @@ class Overview:
         buried_new = tree_counts.new_count - counts[0]
         buried_learning = tree_counts.learn_count - counts[1]
         buried_review = tree_counts.review_count - counts[2]
-        return """
+        buried_label = tr.browsing_buried()
+
+        def number_row(title: str, klass: str, count: int, buried_count: int) -> str:
+            return f"""
+<tr>
+    <td>{title}:</td>
+    <td>
+        <b>
+            <span class={klass}>{count}</span>
+            <span class=bury-count title="{buried_label}">{buried_count or ""}</span>
+        </b>
+    </td>
+</tr>
+"""
+
+        return f"""
 <table width=400 cellpadding=5>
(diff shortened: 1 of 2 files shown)


dae:
I added a small tooltip reusing one of our existing strings, which will hopefully help with identifying what the extra numbers are - WDYT? I guess consistency with our existing bury color is a good thing, though I'm not sure how many people will immediately read it as a buried count, and you're right, it's not super pretty :-) Another option would be a grey which de-emphasises the numbers, sort of like when something is disabled.

RumovZ:
Looking good! Grey also works for me.

dae:
That's nicer - thanks Rumo!

jiqz:
Awesome! Thanks!

PR #1666, opened 2022-02-14
Fix ValueError when exported files have wrong mtime

gnnoh:
Set the `strict_timestamps` argument to `False`, so the media files which have a wrong mtime(like `1970-01-01 00:00:00`) can be normally added to the zipfile instead of raising an error.


dae:
This code is due for a rewrite soon, but I can't see a reason not to include this change until that happens. Thanks for the contribution!

PR #1667, opened 2022-02-14
Add option to ignore accents in search by default

abdnh:
This PR adds a preferences option to ignore accents in unqualified and word boundary searches by default. The "Ignore accents in browser search" add-on supports ignoring accents in unqualified search, but it doesn't work with `w:` as far as I know.

Regarding demand of this feature, the add-on seems quite popular and there are requests to [make this use case easier](https://anki.tenderapp.com/discussions/ankidesktop/41875-proposal-card-browser-search-change-default-search-behaviour-in-respect-to-accents-nc-syntax-as-default) than having to type `nc:` before every search, or to [extend it to word boundary searches](https://forums.ankiweb.net/t/ignore-accent-and-case-while-searching/17305).

I didn't look much into implementing this for field searches. Maybe it can be implemented in the future if there is demand.

Perhaps the Fluent string can be improved to communicate the extent of the feature more clearly.

### DIFF
--- a/ftl/core/preferences.ftl
+++ b/ftl/core/preferences.ftl
@@ -42,4 +42,4 @@ preferences-theme-follow-system = Follow System
 preferences-theme-light = Light
 preferences-theme-dark = Dark
 preferences-v3-scheduler = V3 scheduler
-preferences-ignore-accents-in-search = Ignore accents in search
+preferences-ignore-accents-in-search = Ignore accents in search (slower)
--- a/ftl/core/preferences.ftl
+++ b/ftl/core/preferences.ftl
@@ -42,3 +42,4 @@ preferences-theme-follow-system = Follow System
 preferences-theme-light = Light
 preferences-theme-dark = Dark
 preferences-v3-scheduler = V3 scheduler
+preferences-ignore-accents-in-search = Ignore accents in search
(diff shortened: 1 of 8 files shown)


ftl/core/preferences.ftl
```diff
@@ -42,3 +42,4 @@ preferences-theme-follow-system = Follow System
 preferences-theme-light = Light
 preferences-theme-dark = Dark
 preferences-v3-scheduler = V3 scheduler
+preferences-ignore-accents-in-search = Ignore accents in search
```
dae, line 45:
```suggestion
preferences-ignore-accents-in-search = Ignore accents in search (slower)
```

dae (commented):
It would be nice if we could support this in the future without a performance penalty. But you're right that the demand is there, and we're probably better off providing this as an option. Thanks Abdo.

dae:
> Perhaps the Fluent string can be improved to communicate the extent of the feature more clearly.

I had a few ideas:

- "Ignore accents in unqualified searches" (ignoring the fact that w: is supported)
- "Ignore accents in simple searches" (vague)

Not sure if they're worth it or not, and will require re-translation if we do add support for more in the future.

abdnh:
> Not sure if they're worth it or not, and will require re-translation if we do add support for more in the future.

I'll keep the string as it is then. We can explain the option in detail in the manual.

dae:
Thanks! :rocket:

PR #1668, opened 2022-02-14
View menu

### LINKED ISSUE #1530: Add View menu to Qt toolbars
It's a game of luck to find the fullscreen mode currently (as mentioned by [cqg on the forum](https://forums.ankiweb.net/t/anki-2-1-50-beta/15608/22)). A `View` toolbar menu would be the go-to place for that toggle. Some add-ons like [Browser: Table/Editor side-by-side](https://ankiweb.net/shared/info/831846358) already implement that.

For the browser, I can think of three items that would fit into this category:
- Toggle Full Screen
- Toggle Showing Cards/Notes
- _Toggle Layout_ (potential future Table/Editor side-by side view)

Depending on future changes to the main window, it might benefit from a separate `View` menu as well.

RumovZ:
Qt won't let us define a shortcut sequence twice, so we can't have an app-wide full screen shortcut and a another one in the browser. I went with the simplest solution and just left out the browser shortcut. That means, the menu action doesn't say "F11" or whatever, but that key still works. Open to other suggestions.
I also discovered the StandardKey enum. Let me know if it doesn't work to your liking on another platform.

Closes #1530.


ftl/qt/qt-accel.ftl
```diff
@@ -33,3 +33,10 @@ qt-accel-undo = &Undo
 qt-accel-redo = &Redo
 qt-accel-set-due-date = Set &Due Date...
 qt-accel-forget = &Forget
+qt-accel-view = &View
+qt-accel-full-screen = Toggle &full screen
+qt-accel-zoom-in = Zoom &in
+qt-accel-zoom-out = Zoom &out
```
dae, line 39: Menu items need to use Title Case

dae (commented):
On a Linux machine here, I don't see any shortcut listed next to the main window's fullscreen menu item. Maybe Qt don't have a standard shortcut for this on (some) Linux distros? We might need to hard-code a shortcut in this case (maybe F11)

qt/aqt/forms/browser.ui
```diff
@@ -665,6 +676,34 @@
...
+   <property name="text">
+    <string>qt_accel_reset_zoom</string>
+   </property>
+   <property name="shortcut">
+    <string notr="true">Ctrl+0</string>
+   </property>
+  </action>
+  <action name="actionbrowsing_toggle_showing_cards_notes">
+   <property name="text">
+    <string>browsing_toggle_showing_cards_notes</string>
```
dae, line 704: This felt a bit long as a menu item - would it be worth tweaking the string to be "Toggle Cards/Notes"?

dae:
Thanks as always Rumo :-)

abdnh:
Could <kbd>Ctrl+=</kbd> also be assigned as a shortcut for Zoom In so that the shortcut works on more keyboard layouts [without having to press Shift](https://doc.qt.io/qt-5/qkeysequence.html#keyboard-layout-issues) (e.g. US and UK QWERTY keyboards, I think)?

The Anki Zoom add-on, which is quite popular, assigns both <kbd>Ctrl++</kbd> and <kbd>Ctrl+=</kbd>. See https://github.com/ccz-2/Anki-Zoom/blob/6553d7b4aa6c2b1704466a334912a3c9048a7b03/__init__.py#L155

dae:
It looks like the editor is already using ctrl+= and ctrl+shift+= for subscript/superscript, so currently the subscript shortcut is conflicting with the zoom shortcut, and adding ctrl+= would make superscript conflict as well. :sweat_smile: I guess we're going to need to rethink one pair of these shortcuts; ideas welcome.

RumovZ:
<kbd>Ctrl++/-</kbd> for zoom are really prevalent, so I wouldn't change these shortcuts lightly. However, as Anki already has established conflicting ones, maybe we shouldn't assign them at all? I assume most users use either a mouse or a trackpad, with which zooming is much more ergonomic anyway.

RumovZ:
Then again, <kbd>Ctrl+=</kbd> and <kbd>Ctrl+Shift+=</kbd> are already broken for keyboards like mine, where `=` is <kbd>Shift+0</kbd>, as I just found out. 😅

dae:
Yeah, that issue has cropped up multiple times in the past when the subscript/superscript shortcuts were implemented in Python. Henrik's taken a different approach to handling shortcuts in the JS code that allows this case to work correctly IIRC, but not sure that's something we can do in Qt. Would removing the shortcuts and using ctrl+scrollwheel work for you instead @abdnh?

abdnh:
> Would removing the shortcuts and using ctrl+scrollwheel work for you instead @abdnh?

Yes, I'm fine with using ctrl+scrollwheel.

Issue #1589, opened 2022-01-14
memory leak in browse screen


dae:
browse screen open/close leaks memory: https://forums.ankiweb.net/t/opening-and-closing-browse-permanently-increases-memory-usage/16655

dae:
Thanks, I accidentally cut off the last digit.

dae:
Looks like the sidebar might be the main culprit:

```
>>> import tracemalloc
... aqt.snapshot1 = tracemalloc.take_snapshot()
<no output>
>>> aqt.snapshot2 = tracemalloc.take_snapshot()
... 
... top_stats = aqt.snapshot2.compare_to(aqt.snapshot1, 'lineno')
... 
... print("[ Top 10 differences ]")
... for stat in top_stats[:10]:
...     print(stat)
[ Top 10 differences ]
qt/aqt/browser/sidebar/item.py:71: size=121 KiB (+72.7 KiB), count=1069 (+642), average=116 B
qt/aqt/browser/sidebar/tree.py:785: size=36.6 KiB (+21.9 KiB), count=390 (+234), average=96 B
qt/aqt/browser/sidebar/item.py:79: size=29.3 KiB (+17.6 KiB), count=535 (+321), average=56 B
qt/aqt/browser/sidebar/tree.py:798: size=18.9 KiB (+11.5 KiB), count=199 (+120), average=97 B
qt/aqt/browser/sidebar/tree.py:794: size=18.3 KiB (+11.0 KiB), count=195 (+117), average=96 B
qt/aqt/main.py:1594: size=22.5 KiB (+10.7 KiB), count=71 (+27), average=325 B
<string>:3: size=16.1 KiB (+9048 B), count=318 (+174), average=52 B
pylib/anki/collection.py:625: size=14.5 KiB (+8928 B), count=155 (+93), average=96 B
qt/aqt/browser/sidebar/item.py:73: size=14.1 KiB (+8634 B), count=180 (+108), average=80 B
qt/aqt/__init__.py:632: size=30.8 KiB (+8336 B), count=211 (+1), average=150 B
```

RumovZ:
I spent a long time looking for leaks in the browser and sidebar, and actually found a couple. However, they didn't fix the noticeable memory leak. I then made a curious discovery that isn't limited to the browser.
To reproduce, execute the following:

```
import gc
import weakref

d = aqt.dialogs.open("FilteredDeckConfigDialog", mw)
aqt.r = weakref.ref(d)
[...]

PR #1672, opened 2022-02-16
Fix various minor leaks

RumovZ:
Here are a couple of fixes for other issues I found while investigating #1589.
Additionally, `ProgressManager.timer()` is creating `QTimer`s with mw as the parent. That means when the browser is using it to set a timer for instance, that timer doesn't get destroyed along with the browser. Couldn't this lead to issue? Should `timer()` maybe take an obligatory `parent` arg?


dae:
Will dig into this and your other comments on #1589 and get back to you

RumovZ:
Before we spend time twice, here's a quick update: It turns out the QTimers really do cause issues. In the case of the filtered deck settings dialog, `aqt.utils.ensureWidgetInScreenBoundaries()` (used in `restoreGeom()`) created a timer that holds a reference to the widget.
For the browser, there seems to be more than one issue. Of the three cells referencing it after deletion, I could get rid of two by passing the correct parents to timers. Haven't found the cause for the third one, yet.

RumovZ:
Found it! 🥳🎉🎈

While the leaks are plugged, more could be done to prevent similar issues in the future.
A required parent arg for `progress.timer()` as proposed above doesn't seem advisable, as this routine is heavily used by add-ons, but we could at least print something in the likes of "calling `timer()` without `parent` is deprecated". To make it a bit more ergonomic, we could expose a second helper for the common single shot case only. This wouldn't require a parent as we could use the static `QTimer.singleShot()`.
As for the editor, it's awkward to have the browser mutate a 'private' field of it. Maybe the editor should expose a method for that? I also don't really understand why it would have to be done asynchronously.

RumovZ:
> If it was parented to the dialog then it should get deleted, but if it's parented to mw, it may end up firing after the dialog has been removed, leading to an error. WDYT?

Ah, that makes sense.

>Was it only the preview button that was causing a leak? If so, could we perhaps preserve functionality for existing add-ons that are doing the equiv. of 'editor._links["foo"] = standalone_function', by supporting both the previous approach and the one you have here?

I'm not sure I understood you correctly, but the previewer case doesn't require putting an instance method in `_links`. As long as `_links` is an instance variable, we can store class methods there just as well. I just made more changes than necessary in the heat of the moment.
I agree that we shouldn't worry about add-ons mutating `*E*ditor._links`, because this may already cause issues, if a new Editor instance is created, and the functions stored in its class point to deleted objects.

> As for a public function instead of modifying _links directly, there is already addButton(). We could potentially add another public function, but it might be worth waiting a bit for this, as @hgiesel is still figuring out the add-on story for the editor.

Sounds reasonable, too.

> I imagine there are more issues lurking that we'll discover in the future.

I'll leave our snippets here (top comment), where they may be found by anyone having need for them in the future.

dae:
Great, thanks for those changes, and thanks again for getting to the bottom of this! :tada:

PR #1673, opened 2022-02-17
add GitHub actions workflow for build the code

qxo:
* A workflow for build on linux/masos/window
 with this out-of-the-box feature, it will improve the develop process:
    developers only should care about the source code, 
   the github actions cloud will do the other jobs.
   Before make a PR(pull request), please make sure build-on-all-os passed 
    ***Manually run build-on-all-os workflow on PR branch by github actions UI "Run workflow"***

* A unique branch 'tmp4build' which used only for trigger the workflow by git push
ie: if we want build on a commit($commit_id) at github actions cloud,
   then we just `git push origin $commit_id:tmp4build --force`, which will trigger the workflow:)
   such as a PR code build failed on upstream git repo,  we can change the code locally, 
   before push to PR branch ,  we could force push the commit to the 'tmp4build' branch to check code on all os (linux/macos/windows)
   direct push to PR branch without check will a noise for someone watch the PR.

 ***with this hack we easy build code on web (cloud), not locally***

ref: https://github.com/qxo/anki/actions

### DIFF
--- a/.buildkite/linux/entrypoint
+++ b/.buildkite/linux/entrypoint
@@ -6,13 +6,14 @@ set -e
 echo "--- Checking CONTRIBUTORS"
 .buildkite/linux/check_contributors
 
-BAZEL="bazel --output_user_root=/state/bazel --output_base=/state/bazel/anki"
-BUILDARGS="--config=ci --disk_cache=/state/bazel/disk --repository_cache=/state/bazel/repo"
+test -z "$STATE" && STATE=/state
+BAZEL="bazel --output_user_root=$STATE/bazel --output_base=$STATE/bazel/anki"
+BUILDARGS="--config=ci --disk_cache=$STATE/bazel/disk --repository_cache=$STATE/bazel/repo"
 
 echo "+++ Building and testing"
 
 # move existing node_modules into tree
-test -e /state/node_modules && mv /state/node_modules .
+test -e $STATE/node_modules && mv $STATE/node_modules .
 
 $BAZEL test $BUILDARGS ... //rslib/linkchecker
 
@@ -21,4 +22,4 @@ python tools/copyright_headers.py
 
 echo "--- Cleanup"
 # if tests succeed, back up node_modules folder
-mv node_modules /state/
+mv node_modules $STATE/
--- a/.buildkite/linux/entrypoint
+++ b/.buildkite/linux/entrypoint
@@ -6,14 +6,13 @@ set -e
 echo "--- Checking CONTRIBUTORS"
 .buildkite/linux/check_contributors
 
-test -z "$STATE" && STATE=/state
-BAZEL="bazel --output_user_root=$STATE/bazel --output_base=$STATE/bazel/anki"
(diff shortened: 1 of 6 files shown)


dae:
Thank you for contributing!

We already have a different CI system set up which is considerably faster - typically 1-2 minutes, instead of ~30 minutes, which allows PR authors to be alerted to errors much more quickly. It only runs on Linux for PRs, with checks for Windows and Mac running after the PR is merged. Most changes are not OS-specific, so this usually not a problem.

While I don't think we have a need for this to be enabled by default, I'd be happy to include these yml files in /docs/github in this repo if you'd like - then users could copy them into .github/workflows if they wanted to use them. If you'd like to do that, it would be great if you could change them to only run them manually (eg workflow_dispatch), and remove any unused code (eg conda, wsl?)

qxo:
> Thank you for contributing!
> 
> We already have a different CI system set up which is considerably faster - typically 1-2 minutes, instead of ~30 minutes, which allows PR authors to be alerted to errors much more quickly. It only runs on Linux for PRs, with checks for Windows and Mac running after the PR is merged. Most changes are not OS-specific, so this usually not a problem.
> 
> While I don't think we have a need for this to be enabled by default, I'd be happy to include these yml files in /docs/github in this repo if you'd like - then users could copy them into .github/workflows if they wanted to use them. If you'd like to do that, it would be great if you could change them to only run them manually (eg workflow_dispatch), and remove any unused code (eg conda, wsl?)





> Thank you for contributing!
> 
> We already have a different CI system set up which is considerably faster - typically 1-2 minutes, instead of ~30 minutes, which allows PR authors to be alerted to errors much more quickly. It only runs on Linux for PRs, with checks for Windows and Mac running after the PR is merged. Most changes are not OS-specific, so this usually not a problem.
> 
> While I don't think we have a need for this to be enabled by default, I'd be happy to include these yml files in /docs/github in this repo if you'd like - then users could copy them into .github/workflows if they wanted to use them. If you'd like to do that, it would be great if you could change them to only run them manually (eg workflow_dispatch), and remove any unused code (eg conda, wsl?)

ok, I'll do something about it:)

But I don't think we should move these workflows yml files to /docs/github 

1. Event the workflow enabled by default on yml files, we also can disable on the repo " Settings" ( Settings -> Actions -> General -> Disable Actions)
2. When someone fork the repo, the github actions workflow will not be enabled by default.
3. It's not about time, it's about build on everyone's forked repo which the github actions bring everyone  the out-of-box  CI feature.  -- so we can focus on the code, no need worry about the env.

dae:
You have done a good job getting the build time down, and I can see how having these available by default would be more convenient. Could you revert the changes to the entrypoint files though? They are not intended to be called with arguments with buildkite, so the change makes things less clear. A small README file in .github/actions that describes how to use these files would be good too.

dae:
Those changes look good. Once you've added a README and made any other changes you want to make, please let me know and I'll merge it in.

dae:
Please feel free to reply if you want to pursue this further.

PR #1674, opened 2022-02-18
Remove .exec() from QMenus in Deck Browser, Reviewer, and Card Template screens

roxgib:
- Replaces the .exec() calls with .popup() for a number of QMenus, including:
  - 'Gear' menu on the deck browser
  - 'More' menu in the reviewer
  - Two 'more' menus in the card template editor
- Replaces the .exec() calls with .open() for the deck options deck prompt (triggered when you try to open the deck options when studying a sundeck, allowing you to choose which deck's options will open). 

For background on the motivations for these changes, see #1643, #1650, and [this issue](https://github.com/roxgib/anki-contanki/issues/1). Since these menus contain buttons already connected to the functions they triggers, callbacks do not need to be implemented.

I'm fixing these a few at a time to avoid making a PR with too many changes at once, but I'm happy to do them in bigger batches if that's preferred.


dae:
Thanks Sam!

PR #1677, opened 2022-02-21
Original position

RumovZ:
I'll leave some notes below.

### DIFF
--- a/proto/anki/cards.proto
+++ b/proto/anki/cards.proto
@@ -40,8 +40,8 @@ message Card {
   uint32 remaining_steps = 14;
   sint32 original_due = 15;
   int64 original_deck_id = 16;
-  uint32 flags = 18;
-  generic.UInt32 original_position = 17;
+  uint32 flags = 17;
+  generic.UInt32 original_position = 18;
 }
 
 message UpdateCardsRequest {
--- a/proto/anki/cards.proto
+++ b/proto/anki/cards.proto
@@ -40,8 +40,8 @@ message Card {
   uint32 remaining_steps = 14;
   sint32 original_due = 15;
   int64 original_deck_id = 16;
-  uint32 flags = 17;
-  string data = 18;
+  uint32 flags = 18;
+  generic.UInt32 original_position = 17;
 }
 
 message UpdateCardsRequest {
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -89,11 +89,11 @@ class Card(DeprecatedNamesMixin):
         self.odue = card.original_due
         self.odid = anki.decks.DeckId(card.original_deck_id)
         self.flags = card.flags
-        self.data = card.data
+        self.original_position = card.original_position.val
 
     def _to_backend_card(self) -> cards_pb2.Card:
         # mtime & usn are set by backend
-        return cards_pb2.Card(
+        pb_card = cards_pb2.Card(
             id=self.id,
             note_id=self.nid,
             deck_id=self.did,
@@ -109,8 +109,10 @@ class Card(DeprecatedNamesMixin):
             original_due=self.odue,
             original_deck_id=self.odid,
             flags=self.flags,
-            data=self.data,
         )
+        if self.original_position:
+            pb_card.original_position.val = self.original_position
+        return pb_card
 
     def flush(self) -> None:
         hooks.card_will_flush(self)
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -11,7 +11,7 @@ import anki.collection
 import anki.decks
 import anki.notes
 import anki.template
-from anki import cards_pb2, hooks
+from anki import cards_pb2, generic_pb2, hooks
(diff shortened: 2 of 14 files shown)


proto/anki/cards.proto
```diff
@@ -40,8 +40,8 @@ message Card {
   uint32 remaining_steps = 14;
   sint32 original_due = 15;
   int64 original_deck_id = 16;
-  uint32 flags = 17;
-  string data = 18;
+  uint32 flags = 18;
+  generic.UInt32 original_position = 17;
```
RumovZ, line 44: The build fails if `original_position` is assigned the highest field number, and unlike with a syntax error there is no specific error message. Do you know why? I couldn't find anything about that in the official docs.

pylib/anki/cards.py
```diff
@@ -89,11 +89,11 @@ def _load_from_backend_card(self, card: cards_pb2.Card) -> None:
         self.odue = card.original_due
         self.odid = anki.decks.DeckId(card.original_deck_id)
         self.flags = card.flags
-        self.data = card.data
+        self.original_position = card.original_position.val
```
RumovZ, line 92: Am I correct that the Python implementation will default to `original_position=UInt32(val=0)`? That would defeat the purpose of using the generic wrapper.

rslib/src/scheduler/new.rs
```diff
@@ -119,12 +120,16 @@ impl Collection {
             let cards = col.storage.all_searched_cards_in_search_order()?;
             for mut card in cards {
                 let original = card.clone();
-                card.schedule_as_new(position);
+                if let Some(original_position) = card.original_position {
+                    card.schedule_as_new(original_position);
+                } else {
+                    card.schedule_as_new(position);
+                    position += 1;
+                }
```
RumovZ, line 128: While you said the main motivation is shared decks, I guess this could also be convenient and won't hurt?

proto/anki/cards.proto
dae: Hmm, I don't seem to have a problem using 18 here. Once the other build error is resolved, could you try pushing the original order to CI? If it passes, I can try on a Windows machine.

rslib/src/scheduler/new.rs
dae: It may depend on the user's preference - I suspect some users like to be able to throws cards to the back of the queue, but I imagine others would want the original position restored instead. May be worth giving users the choice in the future?

proto/anki/cards.proto
RumovZ: Maybe it's a cache problem. 🤔 I also get this:
Unbenannt

Building works fine otherwise, and `bazel clean` didn't help, though. I'll look into it further.

dae: Rust Analyzer may need you to ctrl+shift+p: rust analyzer: run, then run a cargo check, in order for the build script to get re-run.

pylib/anki/cards.py
RumovZ: > Presumably it would make sense to have self.card_position being Optional[int]; for that,

Yes, that's what I'm looking for. However, `card.original_position` is always truthy, even if we pass `original_position: None` in Rust. In that case, `card.original_position.val` gives 0, indistinguishable from `original_position: Some(pb::generic::UInt32 { val: 0 })`.
The only solution seems to be the `HasField()` method.

I'll adopt what you've written for the reverse case. Not sure if what I had written is valid as well, or if it only works for oneofs.

rslib/src/scheduler/new.rs
RumovZ: Letting the users decide sounds good, I've reverted it for now.

dae:
LGTM :-) Thanks for working on this Rumo.

proto/anki/cards.proto
RumovZ: Main builds just fine... All right then. Keep your secrets, protobuf.

dae:
So, for the other part of this - allowing the user to restore cards to the original position, what do you think about adding an arg to schedule_as_new() that controls whether the original positions are used or not? And presumably we'll want to make Forget pop up a dialog asking the user what they want? Maybe a checkbox like "Restore original position where possible"? We'll also want to force that mode when exporting an .apkg, in sched.reset_cards()

RumovZ:
Re #1700 you wrote on the forum:

> I’m not super keen on adding an option, as that changes an instantaneous action into one that requires confirmation.

That would have been my concern here as well, but now with two settings a popup seems more jusitifiable.

dae:
Yep, that's a good point. Part of me wants a solution that doesn't require confirmation (eg binding lapse reset+original position into one action, and another action for no reset+end of queue), but that's probably not going to meet everyone's needs, and I guess a checkbox is best here.

dae:
This may already be on your radar, but did you want to tackle the remaining parts of this next? Once we get the rest of this feature landed, and give it a brief time in beta testing, it's probably time for a feature freeze and a switch over to release candidate status for a few weeks.

RumovZ:
Will do!

PR #1678, opened 2022-02-21
Fix bury count

RumovZ:
Apparently, `deck_due_tree(did)` doesn't do what I thought it does.

### DIFF
--- a/qt/aqt/overview.py
+++ b/qt/aqt/overview.py
@@ -221,13 +221,14 @@ class Overview:
 
     def _table(self) -> str | None:
         counts = list(self.mw.col.sched.counts())
-        tree_counts = self.mw.col.sched.deck_due_tree(
-            self.mw.col.decks.get_current_id()
-        )
+        current_did = self.mw.col.decks.get_current_id()
+        deck_tree = self.mw.col.sched.deck_due_tree(current_did)
+        deck_node = self.mw.col.decks.find_deck_in_tree(deck_tree, current_did)
+
         but = self.mw.button
-        buried_new = tree_counts.new_count - counts[0]
-        buried_learning = tree_counts.learn_count - counts[1]
-        buried_review = tree_counts.review_count - counts[2]
+        buried_new = deck_node.new_count - counts[0]
+        buried_learning = deck_node.learn_count - counts[1]
+        buried_review = deck_node.review_count - counts[2]
         buried_label = tr.browsing_buried()
 
         def number_row(title: str, klass: str, count: int, buried_count: int) -> str:


dae:
Do you know what was going wrong? I would have thought the approach here would yield the same results, but presumably it wasn't doing so. I can't remember why I added the ability to limit to a particular branch, and suspect I just thought it might be useful at the time, but we don't seem to be using it anywhere. Maybe it's worth just dropping the feature instead if it's not giving the correct results.

Will merge this in as-is for now, so it can make it into the next beta

RumovZ:
I thought passing a top level id would lead to the respective subtree being returned, where the deck of the passed id is the root node. Instead, the whole tree is returned either way, but without the counts for other decks.
It's documented in Python and Rust:

```python
    def deck_due_tree(self, top_deck_id: int = 0) -> DeckTreeNode:
        """Returns a tree of decks with counts.
        If top_deck_id provided, counts are limited to that node."""
```

```rs
    /// - If top_deck_id is provided, only the node starting at the provided
    ///   deck ID will have the counts populated. Currently the entire tree is
    ///   returned in this case, but this may change in the future.
    pub fn deck_tree(
```

I assume this was meant as an optimisation?
It looks like every caller passing an id immediately afterwards extracts the according node, so maybe this should be done in `deck_tree()` to give less surprising results.

dae:
Yep, but the count of the top deck we limited to should propagate up to the top node, which is why the code was (partly) working even before this change. So I was wondering whether I am overlooking something that would cause the top counts in that case not to match what this PR is getting

RumovZ:
Ah, now I get the idea. Then maybe this usage broke with #1638, because now intermediate parent limits get imposed?

dae:
Possibly, I'm not sure. Let's see how beta 4 goes; if we see any reports of incorrect bury counts we can tally up the buried cards as part of the queue build process instead, and it's probably more efficient that way anyway.

RumovZ:
Why wouldn't the new approach work? I can see how the new limit handling foils the propogation of the counts to the root, because previously it was enough to make sure the nodes above don't add any counts, and only now they impose their limits.
But I don't see what could go wrong if we extract the node of interest.

dae:
The tree code wasn't fresh in my head - I've revisited it, and get what you're saying now. :-) I think that's another argument for stripping out `top_deck_id`, as not only is it not used, but its top values are misleading.

dae:
This looks odd: https://forums.ankiweb.net/t/anki-2-1-50-beta5-buried-bug/17969

RumovZ:
Yes, indeed, but the bury count may only be an indicator. I found a deck that has the same anomaly, and even on 2.1.49 its learning count in the deck tree is lower than the one in the overview screen.
At a first glance, it looks like burying is causing some learning cards to be counted as review cards (or the other way around). Maybe due to the changes to how interday learning cards are queued? I'll look into it later.

RumovZ:
It seems like the issue I'm experiencing is different from the one reported on the forum. This:

grafik grafik

... is actually correct. (Ignoring the redundant '+'.) The deck tree counted 22 intraday and 4 interday learning cards. The review limit is 10, so only 6 review cards can be shown. Now 3 interday learning cards get buried, reducing the learning count to 23. But in turn, **3 more review cards may be shown**.
So it's just that mine/our assumption, that burying can only lead to lower counts, was wrong.
Should we just fix the sign and maybe expand the tooltip? You have already documented the dilemmas of the interday queue and count (8830d338264017712a9c964121735574d3aa7f9b, b9402b5c4782cd401eb0141092a0256cb234dfff), and it doesn't sound like there is a perfect solution.

The bug reported by hanbin must have a different cause, because the above doesn't explain how the *learning* count could be higher after burying.

dae:
Ah, that makes sense, though it is a bit tricky to intuit. The downside of offering these different options is the emergent behaviour it can sometimes produce :sweat: 

One alternative to changing to "-3" would be to hide the negative numbers, and have the grey numbers only represent buried cards, with increases left unexplained. I wonder if that might be slightly clearer, or make things worse?

RumovZ:
Since the reason for this feature was transparency, I wouldn't hide it. Users with default settings won't encounter it anyway, and more adventurous users may notice discrepancies by themselves sooner or later. I'd rather show a tooltip like "Counts in the deck tree may differ, because they cannot take into account burying."

dae:
:+1: 

I assume we're going to keep the tooltip on the grey number? How about something like this for the wording?

"Burying is enabled. Some cards have been excluded, and others may have taken their place."

or

"Counts differ from the deck list, because burying is enabled. Some cards have been excluded, and others may have taken their place."

PR #1680, opened 2022-02-22
Config for burying interday learning cards

RumovZ:
Do the changes to `rslib/src/deckconfig/schema11.rs` cover it or is more required?


dae:
I think that should do it, and the rest all looks good - thanks Rumo :-)

PR #1681, opened 2022-02-22
Fix new deck not being selected

hikaru-y:
This will do the following:
1. Fix the issue: https://forums.ankiweb.net/t/new-deck-is-not-selected-anymore/17712
1. Explicitly delete `StudyDeck` widget after closing the dialog to release memory when `parent` is `None`.
    - Memory will not be automatically released until Anki is closed with the current approach of setting `mw` as parent when `parent` is `None`.
1. Minor refactoring and tidyups in studydeck.py


dae:
Thanks Hikaru! The cleanups are appreciated too. :-)

Arthur-Milchior:
Thanks @hikaru-y, that was frustrating, happy you fixed it!

PR #1682, opened 2022-02-22
Move all buttons to our custom inline surrounding

### LINKED ISSUE #1624: paperclip should insert media at previous cursor position
https://forums.ankiweb.net/t/images-are-included-in-the-start-of-the-field/16967

hgiesel:
Phew, this PR originally started in trying to fix #1624. However, it turned into a different beast.

With this PR, all _inline_ formatting buttons will use our custom algorithm, including the eraser button.

I had to rewrite the algorithm for this purpose, as it couldn't deal with something like text color (the color is like a parameter to the text color, and the old algorithm couldn't keep track of those parameters).

I'd say the algorithm is in a much better shape now: it has way less logic than the old algorithm, much more documentation, and the algorithmic differences between surrrounding and unsurrounding are much smaller. Now for all operations it's basically: 1. Build a formatting tree, and 2. Apply format to the tree nodes in post-order. 

I've reorganized the buttons slightly, to make this distinction between "inline formatting", and "block formatting" easier:

Screenshot 2022-02-22 at 11 36 46

Another advantage of splitting off the bold/italic/underline buttons is that the wrapping behavior is nicer (smaller button groups).

A new feature is that you can now select which formats to clear, when using the eraser:

Screenshot 2022-02-22 at 11 39 04

It also opens up modifying the eraser for add-ons. If you look into e.g. `BoldButton.svelte`, you can see how it pushes its own surround format onto removeFormats.

The original problem mentioned in the first paragraph still exists, but I'll address it in a different PR (that won't take a hundred commits ^^;)

One open issue are the unit tests: I've commented a bunch of them, when they threw up wrong positives. Jest v28 was released last week, and it should fix some of the wrong positives, at which point I would try uncommenting some of them again.


dae:
Looks great Henrik! I only did a brief review, but it seems to work well, and it was great to see some more comments in the code. I'll merge this in now as I'd like to get a beta out tonight or tomorrow. One minor point for later: do you think people will ever need to remove subscripts but not superscripts or vice versa? Even the bold/italic/underline separate choices feels a little niche, at least compared to MS Word where you just get a 'clear all' :-)

hgiesel:
I agree it's probably a little niche. We could certainly sum up superscript/subscript, as they are mutually exclusive. Otherwise it's hard to say, but I don't think it hurts to have more options. It's probably going to be most useful, when users add custom surround formats (like spans with a class). I certainly remember situations, where I would have welcomed something like this.

dae:
Yep, totally see the appeal of being able to add further options in the future, but that means the list will probably grow larger than what it already is, and I just thought it might be better to keep similar options together, to avoid overwhelming the user with too many choices - at least until we see a bunch of demand for being able to configure them separately. :-) But I don't feel that strongly about it.

dae:
Jest 28's now available to test with

PR #1683, opened 2022-02-23
Add progress.single_shot()

RumovZ:
Previous discussion in #1672. Fixes https://forums.ankiweb.net/t/anki-2-1-50-beta-3-4/17501/22.

Sometimes, as in the reported previewer case, there is no reasonable parent to pass to the timer, because 1) the timer must outlive the calling widget, and 2) the calling widget's parent (usually `mw`) is so long-lived that it would cause a memory leak. The solution is the self-destroying static `QTimer.singleShot`.

<hr>

I see two more potential problems with the current timers:

1. Single shots may reference a destroyed widget, because they don't have a parent they get destroyed along with.
2. If the progress manager is busy, it spawns new timers. Those don't react if a user calls `stop()` on the original QTimer they got from `progress.timer()`, so again deleted objects could get called.

These shortcomings existed before #1672, so apparently they don't matter in practice. I just wanted to document them here, in case we run into them in the future.

### DIFF
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -560,7 +560,7 @@ class Browser(QMainWindow):
 
         # schedule sidebar to refresh after browser window has loaded, so the
         # UI is more responsive
-        self.mw.progress.single_shot(10, self.sidebar.refresh)
+        self.mw.progress.timer(10, self.sidebar.refresh, False, parent=self.sidebar)
 
     def showSidebar(self) -> None:
         self.sidebarDockWidget.setVisible(True)
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -560,7 +560,7 @@ class Browser(QMainWindow):
 
         # schedule sidebar to refresh after browser window has loaded, so the
         # UI is more responsive
-        self.mw.progress.timer(10, self.sidebar.refresh, False, parent=self.sidebar)
+        self.mw.progress.single_shot(10, self.sidebar.refresh)
 
     def showSidebar(self) -> None:
         self.sidebarDockWidget.setVisible(True)
--- a/qt/aqt/progress.py
+++ b/qt/aqt/progress.py
@@ -49,13 +49,15 @@ class ProgressManager:
         timer to fire even when there is no collection, but will still
         only fire when there is no current progress dialog.
 
+
         Issues and alternative
         ---
         The created timer will only be destroyed when `parent` is destroyed.
         This can cause memory leaks, because anything captured by `func` isn't freed either.
         If there is no QObject that will get destroyed reasonably soon, and you have to
         pass `mw`, you should call `deleteLater()` on the returned QTimer as soon as
-        it's served its purpose, or use `single_shot()`.\n
+        it's served its purpose, or use `single_shot()`.
+
         Also note that you may not be able to pass an adequate parent, if you want to
         make a callback after a widget closes. If you passed that widget, the timer
         would get destroyed before it could fire.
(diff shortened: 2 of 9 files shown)


RumovZ:
(c05cd71 is an unrelated fix.)

qt/aqt/browser/browser.py
```diff
@@ -560,7 +560,7 @@ def setupSidebar(self) -> None:
 
         # schedule sidebar to refresh after browser window has loaded, so the
         # UI is more responsive
-        self.mw.progress.timer(10, self.sidebar.refresh, False, parent=self.sidebar)
+        self.mw.progress.single_shot(10, self.sidebar.refresh)
```
dae, line 563: This (and other similar cases below) will now throw an error if the widget is closed before the timer has a chance to fire, so we're introducing a bit of a regression here. Perhaps we should be reserving singe_shot() for cases where no appropriate parent exists? Any code that uses them should probably be defensively written, eg testing sip.is_deleted() before doing what it wants to do, and we could mention that in the docstring.

dae (commented):
Thanks for digging into this Rumo. A single_shot() for such cases sounds like a good idea, but one comment below.

qt/aqt/browser/browser.py
RumovZ: That's what I meant with problem 1. Before #1672, all timers were parented to mw, so also weren't dropped if the widget got destroyed in the meantime.

RumovZ:
What do you think of the docs? Should `single_shot()` use `sip.isdeleted(mw)`, or is that not necessary?

qt/aqt/progress.py
```diff
@@ -47,16 +47,61 @@ def timer(
...
+        qtimer.start(ms)
+        return qtimer
+
+    def single_shot(
+        self,
+        ms: int,
+        func: Callable[[], None],
+        requires_collection: bool = True,
+    ) -> None:
+        """Create and start a one-of Anki timer. For an alternative and more
```
dae, line 84:
```suggestion
        """Create and start a one-off Anki timer. For an alternative and more
```

qt/aqt/progress.py
```diff
@@ -47,16 +47,61 @@ def timer(
...
+        documentation see `timer()`.
```
dae, line 85:
```suggestion
        documentation, see `timer()`.
```

qt/aqt/progress.py
```diff
@@ -47,16 +47,61 @@ def timer(
...
+
+        Issues and alternative
+        ---
+        `single_shot()` cleans itself up, so a passed closure won't leak any memory.
+        However, if `func` references a QObject other than `mw`, which gets deleted before the
+        timer fires, an Exception is raised. To avoid this, either use `timer()` passing
+        that object as the parent, or check in `func` with `sip.isdeleted(object)` if
+        it still exists.\n
```
dae, line 93: \n? There's another one above too.

qt/aqt/progress.py
```diff
@@ -47,16 +47,61 @@ def timer(
...
         collection has been unloaded. Setting it to False will allow the
         timer to fire even when there is no collection, but will still
-        only fire when there is no current progress dialog."""
+        only fire when there is no current progress dialog.
+
+        Issues and alternative
+        ---
+        The created timer will only be destroyed when `parent` is destroyed.
+        This can cause memory leaks, because anything captured by `func` isn't freed either.
+        If there is no QObject that will get destroyed reasonably soon and you have to
```
dae, line 56:
```suggestion
        If there is no QObject that will get destroyed reasonably soon, and you have to
```

dae (commented):
Thanks Rumo, those doc changes do a good job of describing the issues at play here, and will hopefully encourage better code moving forward :-)

I'm guessing we can get away without needing to check mw in single_shot(), as that case is already covered by .timer(repeat=False, parent=mw), and that situation is probably going to be pretty rare anyway. But no strong feelings here.

qt/aqt/progress.py
RumovZ: I wanted a line break in the rendered markdown, but no empty line in the plain text, because that looks disconnected from the heading. But if it causes confusion I better remove it.

RumovZ:
> I'm guessing we can get away without needing to check mw in single_shot(), as that case is already covered by .timer(repeat=False, parent=mw), and that situation is probably going to be pretty rare anyway.

My thinking was that `.timer(repeat=False, parent=mw)` shouldn't be used just to ensure the mw still exsists, because then the timer either won't get cleaned up or you'll have to do it manually. But it's probably not a concern in practice, because there is little need to access `mw.super()` (i.e. the underlying QMainWindow).

dae:
> shouldn't be used just to ensure the mw still exsists, because then the timer either won't get cleaned up

Yep, good point. 

>  But it's probably not a concern in practice

I suspect so. :-)

qt/aqt/progress.py
dae: If one newline makes it clearer, how about two instead :-) Then we can just use standard newlines.

PR #1684, opened 2022-02-23
Fix insert media always insert at the start (on Windows)

### LINKED ISSUE #1624: paperclip should insert media at previous cursor position

hgiesel:
Also fixes coloring text upon picking color.
Closes #1624.


qt/aqt/editor.py
```diff
@@ -687,23 +687,22 @@ def saveAddModeVars(self) -> None:
...
         filter = f"{tr.editing_media()} ({extension_filter})"
 
-        def accept(file: str) -> None:
-            self.addMedia(file)
-
         file = getFile(
             parent=self.widget,
             title=tr.editing_add_media(),
-            cb=cast(Callable[[Any], None], accept),
+            cb=None,
```
dae, line 699: Was this just tidying things up, or was the change necessary to fix the issue? @roxgib has been trying to move us away from .exec() and synchronous dialogs (eg https://github.com/ankitects/anki/commit/4b9b923775b7033853ff8f58441a6f0783a06ef1), as it blocks the UI, and has been the source of crashes in the past. Currently getFile() unconditionally uses .exec(), so this code is totally valid at the moment, but as we're trying to migrate away from synchronous dialog calls, changing this function to return a value synchronously is sort of moving us a bit further away from where we want to be.

dae (commented):
Thanks for the quick fixes Henrik!

ts/sveltelib/handler-list.ts
```diff
@@ -0,0 +1,165 @@
...
+
+type Handler<T> = (args: T) => Promise<void>;
+
+interface HandlerAccess<T> {
+    callback: Handler<T>;
+    clear(): void;
+}
+
+class TriggerItem<T> {
+    #active: Writable<boolean>;
```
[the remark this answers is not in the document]
hgiesel: Hm, I think Typescript compiled it into something compatible. Shouldn't have browserslist also complained otherwise?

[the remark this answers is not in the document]
dae: Ah, you're right - looks like TypeScript is transpiling it for us, which is great. Browserslist is a useful tool, but it doesn't have 100% coverage, so it can sometimes pass even when we have code that is incompatible.

qt/aqt/editor.py
hgiesel: The change was necessary to get the timing for inserting media right. I will have another look though.

dae: I seem to recall having similar issues on iOS in the past, and the way I solved it there was to have the code save the cursor position before presenting the window, and then restore the focus/cursor before attempting to insert the media link.

roxgib: Is it possible that the timing issue is because QFileDialog.accept() is being called prior to the callback, and therefore releasing the UI too early? Compare this to StudyDeck, which does the callback first:
https://github.com/ankitects/anki/blob/5eefb9bea7cebe2455c2768419ca0b0b76871c89/qt/aqt/studydeck.py#L166-L169

To be honest getFile isn't high on my list for removing the exec call because it's likely to involve a lot more testing that the other dialogs I've changed so far. I can take a look at improving the callback functionality over the weekend, but it's also likely to need a lot of testing, so I don't want to hold up this fix in the meantime.

dae: I'm not in a hurry to change getFile() either, and when we do, we may be better off creating a new function instead of trying to modify the existing one. I just thought we probably shouldn't be introducing more code that depends on a synchronous dialog return, as that will make things harder in the future.

hgiesel: The problem is indeed that the UI is released before it could set the caret position.

ts/sveltelib/export-runtime.ts
```diff
@@ -6,7 +6,9 @@
 // would not work.
 
 import * as svelteRuntime from "svelte/internal";
+import * as svelteStore from "svelte/store";
 
 import { registerPackageRaw } from "../lib/runtime-require";
 
 registerPackageRaw("svelte/internal", svelteRuntime);
+registerPackageRaw("svelte/store", svelteStore);
```
hgiesel, line 14: This change is unrelated to the PR, but it's quite helpful to debugging the UI in the webview with a quick `require('anki/store').get(ourStore)`

dae:
Thanks as always for all your editor work Henrik. :-)

PR #1685, opened 2022-02-23
Backups

RumovZ:
Here's a start on the backup porting. A few question/notes:

- Building works fine, although I did get a lot of these on the very first build (non-English parts translated):
  ```
  INFO: From Running Cargo build script zstd_sys:
  Build Script Warning: cl : Command line warning D9002 : Unknown option "-fvisibility=hidden" is ignored.
  ```
- Backups are preserved if either
  1. they are from within the last 48h, or
  2. there are no backups from the following 24h period.
 
  So as you suggested. I think that's easy to communicate to the user. The setting label only needs to be changed to: "Keep backups from the last x days."
- Where should this code go? Maybe `collection` isn't such a good idea after all? And should the frontend calls into the backend be made directly from `main.py`?


rslib/src/collection/backup.rs
```diff
@@ -0,0 +1,100 @@
...
+
+pub fn backup(col_path: &str, out_dir: &str) -> Result<()> {
+    let out_file = File::create(out_path(out_dir))?;
+    let mut zip = ZipWriter::new(out_file);
+    let options = FileOptions::default().compression_method(CompressionMethod::Stored);
+
+    let col_file = File::open(col_path)?;
+    let encoded_col = zstd::encode_all(col_file, 0)?;
+
+    zip.start_file("collection.anki2", options)?;
```
dae, line 31: Older clients will try to read this file, and they copy it over the existing collection without checking it. I tried with 2.1.15, and after importing a .colpkg with random data in it, it throws a 'collection corrupt' message and the interface hangs. Once restarted, the collection can't be opened and the profile needs to be deleted. :sweat:  Newer Anki versions at least don't hang the interface, but still copy the file into place without checking it. That should get addressed by the import/export work, but we can't fix older clients, so I think it would be better if we avoid the issue by using a new filename inside the zip.

Anki 2.1 already uses collection.anki21, and using that will cause the same problems. How about collection.anki22? We should probably also throw in a 'meta' file with '{"ver": 3}'. If our restore backup/import code checks to make sure the version isn't greater than what it supports, it will allow us to avoid having to use any more filenames in the future.

dae:
Looking good so far!

zstd is probably using a compile option that should ideally be excluded by them when building with MSVC, but it should be safe to ignore them. I think cargo may hide these by default.

Looks like this increases our library size by about 700k, which seems reasonable for this functionality. :+1:

dae:
For the actual backing up, we may want to do so in close_collection(), after we've closed the collection but before we've dropped the mutex, so we ensure a backup is not corrupted by immediately reopening the profile. We could take an approach similar to the Python code, where we read the closed collection into memory, send the buffer to a separate thread for writing asynchronously, and then return from close_collection() after dropping the mutex.

- We'd want to make it optional, as there may be times we want to skip making a backup on close - maybe it could take a backup_folder: Option<String> arg or similar?
- Notifying the UI if the background thread fails is a complicating factor, and I'd be tempted to just log the error to the console for now, as users with single profiles would probably have the UI shut down by the time the error was triggered anyway.

We could also perhaps automatically trim at the end of a successful backup operation - if we did that, we wouldn't need either of those two protobuf methods.

We'll also need a method to restore the collection to a provided colpkg. While the method wouldn't require an open collection, it's still a collection-related activity, so collection.proto doesn't seem like a bad place for it. Perhaps it should:

- hold the collection mutex, and make sure there's no open collection. I think we should make the UI responsible for closing first, since the UI will know if a backup should be made on close.
- read the provided colpkg file, which may be in an earlier format, and extract the collection file
- attempt to open the collection, and run an integrity check on it. This covers corruption, and checks for an unrecognized schema before we attempt to replace the existing collection file. I said above that we could add these checks in the import/export work, but on reflection, I think it probably makes more sense to do that as part of this work.
- if the above succeeds, close the temporary collection object, and overwrite the original location with its contents

WDYT?

rslib/src/collection/backup.rs
```diff
@@ -0,0 +1,100 @@
...
+pub fn thin_backups(backup_dir: &str) -> Result<()> {
+    let backups = read_dir(backup_dir)?
+        .filter_map(|entry| entry.ok().and_then(Backup::from_entry))
+        .sorted_unstable_by_key(|entry| entry.timestamp)
+        .rev();
+    let mut last_kept_backup_time = TimestampSecs(i64::MAX);
+
+    for backup in backups {
+        let keep_this_backup = backup.timestamp.elapsed_secs() < NO_THINNING_SECS
+            || last_kept_backup_time.elapsed_secs_since(backup.timestamp) > THINNING_INTERVAL_SECS;
```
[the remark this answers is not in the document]
RumovZ: Yes, I thought about that, but deliberately decided against using calendar days. It's advised to study before going to sleep, which is around midnight for a significant amount of users. That means that you could study every 'Anki day', but not every calendar day, because you might study on Monday 11 pm, Wednesday 1 am, and have the day boundary set to 4 am. If you study again at 11 pm on Wednesday, you end up with a 48h gap, because there is no backup for the calendrical Tuesday.

Of course, in this scenario, one of the Wednesday backups is discarded with my implementation as well. I'm not sure what the best solution is here. A few ideas:

1. Keep the current implementation but with 12h intervals. If we assume that people tend to study at around the same times of the day, there should rarely be gaps in the daily backups. On the other hand, at most two per day will be stored, which seems moderate.
2. Use Anki days. This may be a bit fragile if users change the rollover time. Also, some seem to have set it to an unreasonable time.
3. For every day, keep the backup closest to 12 am, even if it's across a day boundary.

Do you have other ideas? I think I'd prefer no. 3.

On a different note, should we encode the timezone in the file name? We can always fall back on the local one for older backups, of course.

P.s. Hope I didn't confuse am/pm anywhere. I'm not used to the 12-hour clock. 😅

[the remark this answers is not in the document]
dae: > For every day, keep the backup closest to 12 am, even if it's across a day boundary.

This feels a bit complicated to code, and to convey to the user. Couldn't we just group the backups into days, then keep the first or last backup of each ordered group, to achieve roughly similar results?

I probably wouldn't bother with a timezone - moves to different timezones are rare, these files are not intended to be shared with others, and since we only start thinning after 2 days, a timezone move shouldn't interrupt any immediate backups. If we add a TZ to the filename and start trying to read it back in when thinning/displaying the list, it adds a bit of complexity to the code, and adds some new failure cases like an invalid TZ name. WDYT?

rslib/src/collection/backup.rs
RumovZ: > I think it would be better if we avoid the issue by using a new filename inside the zip.

Yes, I was planning to do that, but wanted to set up the frontend first for easier testing, therefore the PR. Sorry, I should have said so!

I'd think `.anki22` was used by Anki 2.2.x. It doesn't really matter to me, but maybe `.anki215`? Or `.anki21a`?
Adding meta sounds good!

dae: 21b? Any of them work for me though.

rslib/src/collection/backup.rs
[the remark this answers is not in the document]
RumovZ: What exactly do you want to convey to the user? I thought we'd just offer a setting "backups for the last x days", and optionally a tooltip like "Anki keeps all backups from the last two days, and about 1 for older days." I can't imagine a user will mind if they look into the backup folder and there are two backups for the same date.
Wouldn't grouping result in the same gaps as when using ce days? How would it work in the above example? (Mon 11pm, Wed 1am, Wed 11pm.)

I just thought we better add the timezone now, if we possibly have use for it in the future, but as you point out, that probably won't be the case.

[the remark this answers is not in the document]
dae: >  How would it work in the above example? (Mon 11pm, Wed 1am, Wed 11pm.)

My thinking was that if we use the first backup of each day, they get Mon 11, and Wed 1am. If we use the last backup, they get Wed 11pm instead of Wed 1am.

I do see the problem you are trying to solve, and it's totally valid, I'm just worried that trying to me "smart" about how we handle this case might end up causing confusion for users. On mobile devices where space tends to be tight, backups for a 200MB collection will add up quickly, and I suspect some users may prefer that multiple ones aren't taken in a given day, as it reduces the total number of days they're willing to keep to stay under their space budget. Having said that though, there are probably other users who are more concerned with getting a backup in cases like the one you presented above. Hmm.

Could you give me some pseudo-code  for how you see approach 3 working? Maybe I have not understood what you are proposing fully.

And on the topic of mobile, mobile presents an added complication, as the user "closes" the app each time they switch away to a different one. We need to be able to take backups when they do it, but we don't want hundreds of backups a day, so we probably need some way to skip the backup if a minimum amount of time has not passed since the previous one. This could potentially be useful for the computer too, for people Anki devs and add-on authors, who may be opening and closing the app a lot :-) But there are cases like before a full sync where we want to ensure a backup is taken, regardless of how much time has elapsed.

> I just thought we better add the timezone now, if we possibly have use for it in the future, but as you point out, that probably won't be the case.

It's a reasonable point. My gut tells me YAGNI here, as it hasn't come up so far with the existing timestamp setup that has been in place for years. But I there have been times in the past where I've felt that way, and later come to regret it, so I can't say with certainty :wink:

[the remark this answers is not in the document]
RumovZ: Wrote the pseudocode and realised that it doesn't even solve the problem in all cases. Maybe the solution is much simpler: We just use the last backup of each day, but on days we have no backup for, we take the first one of the following day instead. If we want, we can add an additional condition, like it must be from within the first x hours of the new day, or it mustn't be closer than x hours to the backup picked for the new day.

<details><summary>Pseudocode (feel free to ignore it)</summary>

```python
# first two days are handled separately
day_to_backup = DateTime.today().set_time(12, 0, 0).advance_by_days(-2)
last_backup = None
for backup in backups_in_newest_first_order:
    if backup.datetime.is_earlier_than(day_to_backup):
        # day_to_backup exceeded; ideal backup is this or the last one
        if backup.datetime.until(day_to_backup) < day_to_backup.until(last_backup.datetime):
            backup.keep_it = True
        else:
            last_backup.keep_it = True
        day_to_backup.advance_by_days(-1)
    last_backup = backup
```

</details>

> so we probably need some way to skip the backup if a minimum amount of time has not passed since the previous one.

Or do some conservative thinning to ensure the most recent version is backed up? Either way, it's tricky because the user may deliberately close the app to force a backup before doing something risky. If it goes wrong, they will not be amused if Anki was smart and skipped or deleted that one backup. 😃
Maybe that's more likely to happen on the desktop client though, where people do more complex operations, and know what counts as a "close".

proto/anki/collection.proto
```diff
@@ -30,7 +28,12 @@ message OpenCollectionRequest {
 }
 
 message CloseCollectionRequest {
+  message BackupPaths {
+    string col_path = 1;
```
dae, line 32: Since the Collection object should already know its own path, I think we probably don't need to pass it in when closing - I'd probably drop the inner message and pass backup_dir directly.

dae: (Collection opening already uses empty strings to represent optional args, so we could do that here too)

rslib/src/collection/backup.rs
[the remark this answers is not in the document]
dae: AnkiMobile provides a 'create backup now' option in the preferences/backup screen, which users can use if they want to ensure a backup is taken before they do something experimental.

The problem with unconditionally creating backups and then thinning them afterwards is that we're using the user's flash memory and CPU/battery each time - with a 200MB collection and switching to and from the app every 30 seconds to look words up in a dictionary as they're being created, that's a lot of wasted resources. And even with a throttle, the user is in a better place than they currently are with the computer version, where hours or days can go by without a backup if they don't close their profile. At one point we should probably think about adding timer-based backup creation as well. With the backing up happening in the backend, we could potentially do it while holding the mutex the entire time, preventing any issues with timers or GUI code that tries to access the DB while we close it and copy the data.

I think we need to strike a balance here between potential work lost, and the computing resources required to prevent it. Users have automatic saving and an undo system as a first line of defense, and having to restore from a backup because of a bug is rare - it's usually more about recovering from an accidental mistakes, where the user hasn't thought to make a manual backup first, and has already lost their undo history. Losing 30 minutes of work in that situation is frustrating, but isn't the end of the world. If you feel 30 minutes is too long, maybe we should be making it configurable.

If we're going to add in some smarts to try to cover the early morning study case, perhaps it's worth again considering providing a 'keep x weeks of backups' option as well, as after the first set of conditions, it should be pretty easy to add more. I'm conscious that users sometimes fail to detect problems like deleted cards for some time, and having longer term backups can occasionally save people. One possible approach we could take:

- we gather all available backups into a Vec<Backup>, where Backup has .day and .week ordinals, and a .keep bool
- we apply each of our limits in turn, eg:
- first setting keep=true on all backups within 48 hours (especially if we're throttling them, extra thinning is probably not necessary)
- then setting keep=true on the backups within a 2-7 days that we want to keep
- then setting keep=true on one backup per week
- finally, removing any items without the keep flag set

WDYT?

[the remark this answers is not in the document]
RumovZ: What you say about mobile backups makes sense. I won't try to argue what the best value for the minimum interval between backups is.

> I'm conscious that users sometimes fail to detect problems like deleted cards for some time, and having longer term backups can occasionally save people.

Definitely. I didn't realise it at first, but of course you can restore a deleted deck from a backup and import it into your actual collection, without losing any progress made since. With that in mind, there might even be value in keeping extremely old backups, extremely thinned accordingly. However, as we want the user to decide how much backups to keep, that's not easily possible.
Maybe week-based thinning is the right compromise, I don't know. What setting(s) would you present to the user in this case? One of the reasons I went with days, was that it makes for an easy user decision: "How many days to keep?"
Directly asking for the number doesn't seem practical anymore. If you trigger a lot of backups on one day, that will dispose of the old ones. But once these new backups are old enough to get thinned, you end up with less backups than you asked for.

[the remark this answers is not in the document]
dae: Couldn't we just have "keep up to x daily backups" and "keep up to y weekly backups"? As long as x > 0 and the user has studied in the given week, there should always be at least one daily backup in the weekly period that can be used as the weekly backup. 5 might be a sensible default for x if we have a 48 hour exclusion zone, which should give ~1 week of daily backups. If we wanted to keep the options down, the daily setting could potentially be hidden, and just set to 5 when y>0, and backups disabled completely when y=0. Doesn't hurt much to keep it visible though, as it's sort of documenting what Anki is doing (and most people never tweak the backup options anyway)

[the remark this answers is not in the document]
dae: (adding a 'up to z monthly backups' wouldn't be hard either)

RumovZ:
The code to restore backups and to import are deeply linked currently, so I didn't get around handling media etc.
I'll leave questions below.

Error Management
---
I've taken some shortcuts here. What is the goal? Some aspects:

- Restoring collection fails:
  ❌ Fatal, but probably more detail is needed (I/O, corruption, too new, ...). Maybe we need `AnkiError::ImportError`? There's some overlap with `DbError` though.
- Restoring media fails:
  ✔️❓ I assume this is rare, and we'd want to keep the new collection in this case. Still, we should report back the error, shouldn't we?
- Restoring media *file* fails:
  ✔️ Skip this file and continue? Should we report back any skipped files? Again, probably rare.

rslib/src/config/mod.rs
```diff
@@ -286,6 +292,27 @@ pub(crate) enum Weekday {
...
+    pub daily: u32,
+    #[serde(rename = "w")]
+    pub weekly: u32,
+    #[serde(rename = "m")]
+    pub monthly: u32,
+}
+
+impl Default for BackupLimits {
+    fn default() -> Self {
+        // 2d + 12d + 10w + 9m ≈ 1y
```
RumovZ, line 307: Feel free to tweak these to your liking.

qt/aqt/main.py
```diff
@@ -631,63 +637,9 @@ def apply_collection_options(self) -> None:
...
-            backups.append(file)
-        backups.sort()
-
-        # remove old ones
-        while len(backups) > nbacks:
-            fname = backups.pop(0)
-            path = os.path.join(dir, fname)
-            os.unlink(path)
-
-        self.taskman.run_on_main(gui_hooks.backup_did_complete)
```
RumovZ, line 689: Can we just drop this hook? No add-on seems to use it.

rslib/src/collection/backup.rs
```diff
@@ -0,0 +1,348 @@
...
+            }
+            BackupStage::Monthly => {
+                self.limits.monthly -= 1;
+                self.last_kept_month = backup.month();
+            }
+        }
+    }
+
+    /// Tries to delete the [Backup], discarding any errors.
+    fn delete(&mut self, backup: &Backup) {
```
RumovZ, line 247: It'd be nice to able to test this without writing any files. Should I make `BackupThinner::new` take a `dyn Fn(&Backup)`? Then in testing we could pass in some closure that allows to check which `Backup`s would get deleted.

pylib/anki/collection.py
```diff
@@ -235,15 +235,22 @@ def autosave(self) -> None:
         elif time.time() - self._last_checkpoint_at > 300:
             self.save()
 
-    def close(self, save: bool = True, downgrade: bool = False) -> None:
+    def close(
+        self,
+        save: bool = True,
+        downgrade: bool = False,
+        backup_folder: str = "",
```
dae, line 242: Sorry, I know I only just suggested copying the "empty string = undefined" approach from OpenCollection(), but it looks like we can do this in a better way now:

https://github.com/ankitects/anki/commit/6941bccde4212a15816d07852ab4f6e0ef6395c0

We should be able to pass a str | None into _backend.close_collection(), as the message constructor will accept a None argument for an optional field.

rslib/src/collection/backup.rs
dae: How about separating out the logic and the I/O? Eg we have a function that takes a list of backups and returns a list of ones that need deletion (or mutates a delete/keep flag on them). That's easy to test, and then the actual deletion is just a matter of looping over the selected files and removing them.

qt/aqt/importing.py
```diff
@@ -460,39 +451,11 @@ def _replaceWithApkg(mw: aqt.AnkiQt, filename: str, backup: bool) -> None:
...
-            shutil.copyfileobj(source, target)
-
-        d = os.path.join(mw.pm.profileFolder(), "collection.media")
-        for n, (cStr, file) in enumerate(
-            json.loads(z.read("media").decode("utf8")).items()
-        ):
-            mw.taskman.run_on_main(
-                lambda n=n: mw.progress.update(  # type: ignore
-                    tr.importing_processed_media_file(count=n)
-                )
```
dae, line 484: This is potentially quite slow when users have a lot of media, so we probably need to provide progress info. check_media_folder() and checkmedia.py is an example of how we can do this.

qt/aqt/main.py
dae: @Arthur-Milchior I seem to recall you originally added this for long term backups - will the changes discussed above mean you no longer need this?

rslib/src/collection/backup.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    Ok(())
+}
+
+pub fn restore_backup(col_path: &str, backup_path: &str, media_folder: &str) -> Result<()> {
+    let backup_file = File::open(backup_path)?;
+    let mut archive = ZipArchive::new(backup_file)?;
+    let new_col_data = extract_collection_data(&mut archive)?;
+    let old_col_data = fs::read(col_path)?;
+    fs::write(col_path, new_col_data)?;
```
dae, line 59: This approach has a couple of race conditions:

- if the new data fails the check and Anki or the system crashes, the user may be stuck with the new data. Likewise if a crash happens after it's written but before a check is done.
- fs::write() is not atomic, so if a crash happens before all data is written, you can end up with partially written file.

Reading the old file into memory first also increases our memory usage here.

What I'd recommend doing instead is to use NamedTemporaryFile::new_in() to write a temporary file in the same folder as the existing collection (can't use standard temp location, as cross-drive/cross-partition will break things). We can run the tests on that temporary file, and if anything goes wrong, the named file can juts be dropped. Once we're ready to use it, the original file can be replaced atomically via https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html#method.persist, ensuring it doesn't get half-written.

We can probably get away with reading the entire new collection into memory as we are currently doing with decode_all(), but it's not ideal - at one point it might be nice to tweak the code so the collection is decoded and written into the temporary file in chunks, instead of loading everything into memory.

dae (commented):
I haven't had a chance to look though everything yet, but seems to be coming along nicely. Some initial comments:

Arthur-Milchior:
@dae thanks for mentioning me. If I get it right, you'll essentially keep a back-up for every single day anki was opened. In this case, my add-on indeed will be deprecated.
Personally, I decided to keep only a back-up by months for every back-up which is more than a month old and one by year for the preceding year. Even if you don't preserve media, I'm still worried that it will lead to too much space taken. So ideally, I guess I would keep using a similar add-on, but to restrict the number of back-up. However, this can be independant of any specific feature you implements, so don't mind me here

dae:
Re error handling:

>x Fatal, but probably more detail is needed (I/O, corruption, too new, ...). Maybe we need AnkiError::ImportError? There's some overlap with DbError though.

Which cases aren't already covered by DbErrorKind?

> Restoring media fails:

Do you mean the `media` file is missing/unparsable? I'd just treat that as fatal/corrupt, as without it we can't know any of the filenames, and if the user is desperate to recover their collection they could manually modify the file.

> Restoring media file fails:
> Skip this file and continue? Should we report back any skipped files? Again, probably rare.

The errors I can think of:
- corrupt .zip file
- permission or I/O issue reading or writing to the media folder

I'd be tempted to abort in both those case, leaving the user with the new collection and a message telling them the error that stopped media files from being imported. WDYT?

On the topic of media files, it may make sense to compress media files as well. While mp3s/jpg files don't tend to compress well, other files like png and wav files do benefit. But that raises some questions we'll need to dig into:

- using our existing API calls, what does zstd do when fed non-compressible data, like random data? Is it producing an output larger than the input, or does it have a way of representing that the data is original? If the former, we may need to keep track of this ourselves, using some flag on files to indicate if they are compressed or not.
- does the above change if we switch to chunk-based encoding/decoding in the future, instead of trying to do it all in one operation?

qt/aqt/main.py
dae: @Arthur-Milchior re your comment below:

> Personally, I decided to keep only a back-up by months for every back-up which is more than a month old and one by year for the preceding year. Even if you don't preserve media, I'm still worried that it will lead to too much space taken. 

We are looking into options for # of weekly and daily backups to keep. No setting for yearly backups at this point.

rslib/src/config/mod.rs
dae: The way I'd been thinking it might work is that the ranges would overlap, rather than be one after another. That would allow a particular backup to satisfy multiple criteria at once (eg a given day's backup, but also a given month's backup), and wouldn't require adding the different periods up to know the maximum interval that will be kept. I'm not pushing for this solution over the existing one at the moment (haven't thought it through carefully yet :-), but wondered what you think.

dae:
I did a bit of digging into the zstd calls, and it seems to be ok:

- compress_all() appears to use the standard streaming/frame API that the command line tools use
- from what I can tell from poking through the source, it will store non-compressible data as raw (with slight overhead for frame/block headers)

rslib/src/collection/backup.rs
RumovZ: Oh, that's even better. 🙂

pylib/anki/collection.py
RumovZ: No problem. It's great to be able to ditch these nasty wrappers. Too bad that the Python implementation is still so cumbersome.

rslib/src/config/mod.rs
RumovZ: This approach simplifies calculating the number of stored backups, though.
Anyway, the main reason I chose it was the overlap of weeks and months: The backup we'd like to keep for January is the one from the 31st, which was a Monday. However, the backup we'd like to keep for that week is from Sunday, February 6. So we'd have to keep two backups from the same week.
I also thought that was your plan, too, because you wrote this:

> 5 might be a sensible default for x if we have a 48 hour exclusion zone, which should give ~1 week of daily backups.

Treating the grace period differently might increase confusion.

rslib/src/preferences.rs
```diff
@@ -149,3 +155,23 @@ impl Collection {
         Ok(())
     }
 }
+
```
RumovZ, line 158: A bit boilerplaty. Should I ditch the native struct in favour of the auto-generated pb one?

ftl/core/preferences.ftl
```diff
@@ -43,3 +43,7 @@ preferences-theme-light = Light
 preferences-theme-dark = Dark
 preferences-v3-scheduler = V3 scheduler
 preferences-ignore-accents-in-search = Ignore accents in search (slower)
+preferences-backup-explanation = Anki will create a backup of your collection each time it is closed. Backups older than 2 days will be thinned, which means only one per the last backed up days, weeks, and months is preserved, according to the following values.
```
RumovZ, line 46: Whatever behaviour you decide on, this probably needs tweaking. I have a hard time describing it concisely.

pylib/anki/collection.py
RumovZ: This seems to break passing keywords directly into an RPC.
04b209c

proto/anki/collection.proto
```diff
@@ -107,5 +116,6 @@ message Progress {
     FullSync full_sync = 4;
     NormalSync normal_sync = 5;
     DatabaseCheck database_check = 6;
+    string media_processing = 8;
```
RumovZ, line 119: This again. 😞

qt/aqt/main.py
dae: @RumovZ to answer your question, I think dropping it's fine, as we can't easily provide it anymore anyway

rslib/src/config/mod.rs
dae: That all sounds reasonable; let's run with your solution.

proto/anki/collection.proto
dae: ?

If you're referring to the blank=null discussion from above, oneofs are a different story (and the new optional functionality is actually handled internally like a hidden oneof)

rslib/src/preferences.rs
dae: We could probably just reuse the pb here; it's a simple struct and a native copy of it doesn't buy us anything.

dae:
https://github.com/ankitects/anki/pull/1685/commits/22e1fbeebc6c56ab60be5f412b5ca6c70f31d860 has revealed a problem - backups are likely to be skipped when quitting the program. Python by default will wait for background threads to finish running before terminating the program, but that doesn't extend to threads we create in Rust, so we are responsible for .join()ing them ourselves.

One way we could solve this is to store the spawned thread handle in Backend, and then provide another backend method to wait for a backup to complete if there is one. The frontend could then call that before closing the program. When creating a backup, if there's an existing backup thread handle, I guess we should wait for it before continuing. WDYT?

rslib/src/collection/backup.rs
```diff
@@ -341,3 +348,65 @@ fn get_zip_file_contents(file: &mut ZipFile) -> Result<Vec<u8>> {
...
+            // weekly
+            (backup!(2022, 2, 7), true), // Monday, week already backed up
+            (backup!(2022, 2, 6, 1, 0, 0), true),
+            (backup!(2022, 2, 6, 2, 0, 0), false),
+            (backup!(2022, 1, 6), false),
+            // monthly
+            (backup!(2022, 1, 5), true),
+            (backup!(2021, 12, 24), false),
+            (backup!(2021, 12, 1), true),
+            (backup!(2021, 11, 1), true),
```
dae, line 428: These macros are great for cutting down on the boilerplate :-) It might be worth adding one more test here, synthesizing backups for every day for a year in a loop, and then confirming thinning works as expected in that case too.

ftl/core/preferences.ftl
dae: Does the updated wording work for you? I used 'periodically ' to allow for the previously-discussed throttling of backup creation.

proto/anki/collection.proto
RumovZ: Sorry, that was rather vague. 😅
The build fails if I write `= 7`. This time I can only get around this by skipping this number, but without a doubt it's the same issue as in https://github.com/ankitects/anki/pull/1677#discussion_r811070799.

ftl/core/preferences.ftl
RumovZ: Sure!

RumovZ:
Thanks for the help!

> I'd be tempted to abort in both those case, leaving the user with the new collection and a message telling them the error that stopped media files from being imported. WDYT?

Then we'll need to return some status message in the success case, and for the failure case, the existing db errors should be sufficient.
Currently a failing import is breaking the UI, because it has trashed the old collection and loaded a new profile. More decoupling of the backup code from the import code is needed.
Will tackle that next and gradually coming back to you on the other points. (The same goes for my outstanding answers on previous PRs.)

proto/anki/collection.proto
dae: Ah, I get you now :-) I tried to reproduce it on a Windows machine, and got an error about a missing backend_pb2.py file or similar one time, but then the build completed fine the second time. Do you use --disk_cache in your user.bazelrc? I don't use it on Windows, and it's possible that there's a race condition in the build process on Windows leading to a bad output, which then gets cached on your system. If that's the case, you probably don't need to change the number, and a minor change like a comment or extra newline would probably fix it too.

RumovZ: > got an error about a missing backend_pb2.py file or similar one time

I got used to these, as restarting once or twice always resolves them.

> Do you use --disk_cache in your user.bazelrc? I don't use it on Windows, and it's possible that there's a race condition in the build process on Windows leading to a bad output, which then gets cached on your system.

You're right! Deleting that cache fixed it. I would have tried that sooner, but I thought `bazel clean --expunge` would have the same effect.

> If that's the case, you probably don't need to change the number, and a minor change like a comment or extra newline would probably fix it too.

I'll try that the next time. I understand this means it was just chance that tweaking the numbers worked, and I actually only needed to keep making changes until a build succeeded at first pass (without the usual error mentioned at the top).

dae: If the backend_pb2.py thing keeps coming up, you might want to try commenting `stderr=subprocess.DEVNULL` in protoc_wrapper.py to see if that gives us any hint as to why the file is missing - it just makes the build a bit noisier. It might also give us a hint as to what exactly is going wrong in the bad cache case.

RumovZ:
I'll try to bundle the remaining issues in the following list.

- [ ] Was the following meant to be addressed in this PR? The new code doesn't write any media into archives, because it is only used for automatic backups yet. We could add code to decompress media from v3 archives, but it won't get executed until there is actual export code to write v3 archives. When that's happening, this code here could be adjusted without bumping the version.
  > On the topic of media files, it may make sense to compress media files as well. 
- [ ] Errors and aborting (see above).
- [ ] Currently, opening Anki by clicking a backup file will prompt the user to open a profile, then make a full import. As the profile might be corrupt, ideally no profile would be attempted to open. But since the user is also required to specify a profile first, I'm not sure how this would work. Maybe we should just leave that for now, but I wanted to mention it.

rslib/src/collection/backup.rs
```diff
@@ -0,0 +1,480 @@
...
+
+fn version_3_reader(archive: &mut ZipArchive<File>) -> Option<impl Read + '_> {
+    archive
+        .by_name("collection.anki21b")
+        .ok()
+        .and_then(|file| Decoder::new(file).ok())
+}
+
+fn legacy_reader<'a>(archive: &'a mut ZipArchive<File>) -> Option<impl Read + '_> {
+    if archive.file_names().contains(&"collection.anki21") {
```
RumovZ, line 350: This seems needlessly wasteful, but I didn't get it to compile without searching first. The problem is that the mutable borrow of `archive` is not dropped at the end of an `if let` clause (instead only at the end of the block). Likewise, it's not dropped after a conditional return (https://users.rust-lang.org/t/solved-borrow-doesnt-drop-returning-this-value-requires-that/24182)

rslib/src/collection/backup.rs
```diff
@@ -0,0 +1,477 @@
...
+            yesterday: today.num_days_from_ce() - 1,
+            last_kept_day: i32::MAX,
+            last_kept_week: i32::MAX,
+            last_kept_month: u32::MAX,
+            limits,
+            obsolete: Vec::new(),
+        }
+    }
+
+    /// Returns a [Vec] with the obsolete [Backup]s.
```
dae, line 206: We're kind of restating the type signature here. How about:

```suggestion
    /// Returns obsolete backups that can be deleted.
```

We should probably tweak the method name too, since it no longer does what it original did. Maybe "obsolete_backups()"? Then we no longer need the docstring at all.

rslib/src/collection/backup.rs
```diff
@@ -0,0 +1,477 @@
...
+            if self.is_recent(&backup) {
+                self.keep(None, backup);
+            } else if self.remaining(Daily) {
+                self.keep_or_delete(Daily, backup);
+            } else if self.remaining(Weekly) {
+                self.keep_or_delete(Weekly, backup);
+            } else if self.remaining(Monthly) {
+                self.keep_or_delete(Monthly, backup);
+            } else {
+                self.delete(backup);
```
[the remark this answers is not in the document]
dae: Plus maybe keep_or_flag_for_deletion()? A bit wordy though :sweat_smile:

rslib/src/collection/backup.rs
dae: Yeah, that's unfortunate. zip-rs is actually creating a hashmap of the files up front, but it's not exposed publicly. IIRC Anki clients will place the collection file in the zip before the media, so assuming the index is being read in order, this check should at least be able to terminate within a few filenames.

dae: ... actually no, the order would be lost when the filenames are added to the hashmap. How about this?

```rust
fn legacy_reader(archive: &mut ZipArchive<File>) -> Option<impl Read + '_> {
    let have21 = archive.by_name("collection.anki21").is_ok();
    if have21 {
        archive.by_name("collection.anki21").ok()
    } else {
        archive.by_name("collection.anki2").ok()
    }
}
```

rslib/src/collection/backup.rs
```diff
@@ -0,0 +1,477 @@
...
+            return Err(AnkiError::Interrupted);
+        }
+
+        if let Ok(mut zip_file) = archive.by_name(&archive_file_name) {
+            let file_path = Path::new(&media_folder).join(normalize_to_nfc(&file_name).as_ref());
+            let files_are_equal = fs::metadata(&file_path)
+                .map(|metadata| metadata.len() == zip_file.size())
+                .unwrap_or_default();
+            if !files_are_equal {
+                io::copy(&mut zip_file, &mut File::open(file_path)?)?;
```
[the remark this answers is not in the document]
RumovZ: Since you're sure that media will be compresed in the future, it makes sense to handle that now, and make this code as forward compatible as possible.
I'll take care of it.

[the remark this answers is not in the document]
RumovZ: But the current approach would still work, wouldn't it? I don't think speed should be a concern. Let's assume a user tries to open a .colpkg from a future Anki release with this one (presumably 2.1.50).
Option 1: Process it, sluggishly at worst.
Option 2: Refuse to process it.
The latter inconveniences the user much more.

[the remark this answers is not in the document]
dae: Just worried about committing to something and then finding out later than the raw compression outputs in a different format, eg without wrapping in frames or without some other header. If we can confirm they will be compatible, then agreed that option one is a much better choice.

rslib/src/collection/backup.rs
```diff
@@ -0,0 +1,497 @@
...
+    }
+
+    fn mark_obsolete(&mut self, backup: Backup) {
+        self.obsolete.push(backup);
+    }
+}
+
+impl Meta {
+    /// Extracts meta data from an archive and checks if its version is supported.
+    fn from_archive(archive: &mut ZipArchive<File>) -> Result<Self> {
```
dae, line 281: Nice

dae:
Will come back to this either later tonight or tomorrow, but looking pretty great at this point :-) It looks like we might not be throttling backup creation yet - could we get that into this PR? We could just check for the latest mtime in the backups folder, or could parse the filenames. We'll also need a way to force creation regardless of this previous creation time.

dae:
I assume File::open(dir) is POSIX, but could you give it a quick test to make sure restoring still works on Windows?

rslib/src/collection/backup.rs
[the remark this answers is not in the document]
RumovZ: I had a quick glance, and it doesn't look like reusing decoding/encoding contexts is currently supported (https://github.com/gyscos/zstd-rs/issues/41). What about putting media files into a nested archive and compressing that?
Either way, it shouldn't be a big change, so I'd just implement whatever you think is the most likely scenario for the future, and minimise chances for breaking forward compatibility.

RumovZ:
> It looks like we might not be throttling backup creation yet - could we get that into this PR? We could just check for the latest mtime in the backups folder, or could parse the filenames. We'll also need a way to force creation regardless of this previous creation time.

Done. Are there any places where we want to force it, yet?

> I assume File::open(dir) is POSIX, but could you give it a quick test to make sure restoring still works on Windows?

Unfortunately, it doesn't. I suppose we'll use walkdir then to recursively sync the files? Does this cover the `persist` issue?

Quick reminder:

> Instead we could poll for an abort signal once before restoring media. Then if the user wanted to abort, we could completely rollback the import. Would that be the most intuitive behaviour?

dae:
> Done. Are there any places where we want to force it, yet?

full_download_inner() should probably be forcing a backup

We also need to consider the colpkg import case, as we don't want the user to lose 30 minutes of work if they decide an import was a mistake. I first thought about just forcing the 'import while collection open' case, but opening a backup from the profile screen is a problem too - the user could potentially do 25 minutes of work, return to the profile screen, then import a backup.

Not sure what the best way to handle this is. One idea is to switch from a boolean flag to allowing the .close_collection() caller to pass in a desired_backup_secs argument? The cases like full syncs and colpkg imports could pass in, say, 300 to ensure a backup is made if more than 5 minutes have elapsed since the last backup. That would hopefully be short enough that people don't feel they've lost much work, without potentially creating hundreds of backups if the user is rapidly retrying imports.

Taking it a bit further, maybe we could make ~5 minutes the default, and just pick longer backup durations for the mobile clients when the user is backgrounding the app. WDYT?

> Unfortunately, it doesn't. I suppose we'll use walkdir then to recursively sync the files? Does this cover the persist issue?

The operations are a bit different. Think of the directory as a hashmap of filenames, with pointers to different file objects. When we sync a file, we ensure that a given file object is up to date on disk. When we persist(), we're renaming the temporary filename to the desired one, effectively deleting the temporary file entry in the hashmap, and creating a new entry with the desired filename, pointing to the same file object. The directory sync is to ensure that after that point, the directory will point to the new file, even if the computer crashes/loses power at that point. Without the sync, a crash/power loss at that exact point could cause the previous collection to come back again. 

The directory sync is not super important in our case - we're not dealing with bank transactions that we need to ensure don't get undone, and our other syncing/persist code ensures that we either get the full old file, or the full new file, and not a half-written one. And the directory changes will likely be persisted in a few seconds if we take no action. So I think it would be fine to just skip the directory sync on Windows - we could write a helper function that does the sync, and use a #[cfg(windows)] to control the behaviour. 

> Instead we could poll for an abort signal once before restoring media. Then if the user wanted to abort, we could completely rollback the import. Would that be the most intuitive behaviour?

Sorry, forgot to reply there. Sounds reasonable to me :-)

rslib/src/collection/backup.rs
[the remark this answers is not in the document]
RumovZ: After thinking about it a bit, I take that back. The archive approach would likely be somewhat more involved, and without having the generating counterpart, it would be hard to test.

RumovZ:
Thanks for taking the time to explain about syncing!

Allowing abortion was harder than I thought, because it's not enough to call it once, and then we need to send a presentable label.

> full_download_inner() should probably be forcing a backup

That's a collection function, so it couldn't coordinate with other running backups. Maybe it can be done in `full_sync_inner()` instead?

rslib/src/collection/backup.rs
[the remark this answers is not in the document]
dae: We can do it, but we need to drop down to a lower level:

```rust
use std::time::Instant;

use anyhow::Result;
use zstd::stream::raw::Encoder as RawEncoder;
use zstd::stream::zio::Writer;
use zstd::zstd_safe;
use zstd::Encoder;

const INPUT_BYTES: &[u8] = include_bytes!("main.rs");
const CYCLES: usize = 10000;

fn simple_compress() -> Result<Vec<u8>> {
    let mut reader = std::io::Cursor::new(INPUT_BYTES);
    let writer = Vec::new();
    let mut encoder = Encoder::new(writer, 0)?;
    encoder.multithread(16)?;
    std::io::copy(&mut reader, &mut encoder)?;
    encoder.finish().map_err(Into::into)
}

fn setup_reusable_encoder() -> Result<RawEncoder<'static>> {
    let mut encoder = RawEncoder::with_dictionary(0, &[])?;
    encoder.set_parameter(zstd_safe::CParameter::NbWorkers(16))?;
    Ok(encoder)
}

fn reusing_compress(encoder: RawEncoder<'static>) -> Result<(Vec<u8>, RawEncoder<'static>)> {
    let mut writer = Writer::new(Vec::new(), encoder);
    let mut reader = std::io::Cursor::new(INPUT_BYTES);
    std::io::copy(&mut reader, &mut writer)?;
    writer.finish()?;
    Ok(writer.into_inner())
}

fn main() -> Result<()> {
    // ensure we get identical results
    let data1 = simple_compress()?;
    let reusable_encoder = setup_reusable_encoder()?;
    let (data2, reusable_encoder) = reusing_compress(reusable_encoder)?;
    assert_eq!(data1, data2);
    let (data3, mut reusable_encoder) = reusing_compress(reusable_encoder)?;
    assert_eq!(data1, data3);

    // compare performance
    let start = Instant::now();
    for _i in 0..CYCLES {
        let (data, encoder) = reusing_compress(reusable_encoder)?;
        assert_eq!(data, data1);
        reusable_encoder = encoder;
    }
    println!(
        "reuse took {}",
        Instant::now().duration_since(start).as_millis()
    );

    let start = Instant::now();
    for _i in 0..CYCLES {
        let data = simple_compress()?;
        assert_eq!(data, data1);
    }
    println!(
        "simple took {}",
        Instant::now().duration_since(start).as_millis()
    );

    Ok(())
}
```

On a 50MB collection file with 100 cycles, the simple approach is actually a little faster:

reuse took 3933
simple took 3842

But when we process lots of little files, reusing the context is a big win:

reuse took 144
simple took 2697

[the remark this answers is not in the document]
dae: (because the output is identical, we can take a simple approach for decompression for now, and deal with encoding/optimizing in the future)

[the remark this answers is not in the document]
dae: One other point of interest - when turning off multithreading, the two cases are much closer together, and the reuse case is about twice as fast:

reuse took 68
simple took 147

Switching to a 1MB test file, single threading is still better performing than multithreading. At 10MB, multithreading is performing better than single threading. So we probably don't want to enable it for the media case, as media files tend to be smaller than 10MB (and we'd likely achieve better performance by decompressing multiple files in parallel if we wanted to push that any further).

dae:
> That's a collection function, so it couldn't coordinate with other running backups. Maybe it can be done in full_sync_inner() instead?

Good point. full_sync_inner() sounds like it might work, if moving the location the collection closed isn't too disruptive.

rslib/src/collection/backup.rs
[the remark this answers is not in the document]
RumovZ: That looks great! Both for this PR and future import work.
Seems like I should have looked a level deeper as well, so to speak.

RumovZ:
Hm, if it's done on the backend, would that be to replace the `close(), reopen()` on the frontend where backing up before a full download is currently handled (`_close_for_full_download()`)?
I've adjusted the frontend for now, which was trivial, but can move it to the backend anytime.

dae:
Keeping it on the frontend probably makes sense at the moment, as otherwise we need to pass the backup details into the full sync request. Looking great Rumo - thanks for working on this!

ftl/core/importing.ftl
```diff
@@ -72,10 +72,10 @@ importing-note-updated =
         [one] { $count } note updated
        *[other] { $count } notes updated
     }
-importing-processed-media-file =
```
RumovZ, line 75: This one wasn't introduced by this PR, so shouldn't it be kept around and garbage collected later?

PR #1686, opened 2022-02-23
Match deck name truncation behavior to prevent long deck names from obscuring stats

sachingooo:
When deck names are very long, they have no limit as to how much is displayed, resulting in the long deck name obscuring the stats on the right-hand side of the screen by pushing them off to the side. This issue is not present at https://ankiweb.net/decks/. 

For example, here is AnkiWeb truncating the name of a long deck so that stats are still visible:
image

Here is the desktop client's current handling of a long deck name:
image
The stats are pushed off to the side. This is undesirable functionality and does not match AnkiWeb's handling. 

This pull request addresses the issue in the desktop client by matching AnkiWeb's handling of the deck name - truncating it if it's too long. This is accomplished by computing an appropriate maximum width for the deck name table cell and truncating anything larger than the cell size. 

The new behavior of the desktop client:
image

The computation is done based on the current window size and adapts as the window is resized.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -95,6 +95,7 @@ Danish Prakash <github.com/danishprakash>
 Araceli Yanez <github.com/aracelix>
 Sam Bradshaw <samjr.bradshaw@gmail.com>
 gnnoh <gerongfenh@gmail.com>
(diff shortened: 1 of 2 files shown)


dae:
With an adaptive limit this seems quite reasonable - thank you for the contribution!

PR #1690, opened 2022-02-25
Add silent option for uninstaller

PatricCunhaforWork:
Change to allow skipping of confirmation prompt for unattended uninstall.

Let me know if I should update the documentation somewhere to reflect the change.


PatricCunhaforWork:
@dae Hey Damien, the use case is really only uninstalling, not upgrading.
Storage is limited, and sometimes it's necessary to do an unattended uninstall of apps that are no longer used.

dae:
Appreciate you taking the time to elaborate!

PR #1691, opened 2022-02-25
Using the Rust backend in the editor

hgiesel:
So far a draft. It works well, but there is an issue in the editor in the browser. You will notice that going up and down will make the editor noticeably lag. The problem is that were sequentially getting 1. the note, and then 2. the notetype.

1. I think we should cater the backend more to the individual views. Right now we have e.g. `NotesService`, `NotetypesService`, etc., i.e. the services are oriented on the data objects in Anki. But to have a responsive UI, we should fit some endpoints directly to what the front-end needs.

Two other points causing me a headache:

2. It seems like protobufjs has very unhelpful/damaging typing. You must have noticed it already, when you wrote the deck options screen, e.g. statements like this:

```typescript
this.defaults = data.defaults!.config! as DeckConfig.DeckConfig.Config;
```

It seems like, when the type of `data.defaults` should theoretically be `Deck.DeckConfig` (which is its type at runtime), it actually is `Deck.IDeckConfig | null | undefined`, and so on. At that point I would actually prefer everything to be `any`, rather than wrong types that require type assertions and casting all the time.

Do you know the reason for this? Is this really an oversight on the side of protobufjs? Or is this something within our grasp?

3. I cannot use `async` functions in svelte files. It gives me an error like so:

```
compile failed: Cannot declare same variable name which is imported inside <script context="module"> (22:9)
20:
21: <script lang="ts">var _a;
22: import { __awaiter } from "tslib";
             ^
23: import { onMount, tick } from "svelte";
24: import { get, writable } from "svelte/store";
```

Do you have any clue what causes this? If no, my first guess is that there's something wrong with the logic in `ts/svelte/svelte.ts`. Does that seem right?

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -397,11 +397,7 @@ require("anki/ui").loaded.then(() => require("anki/NoteEditor").instances[0].too
                 self.currentField = None
                 # run any filters
                 if gui_hooks.editor_did_unfocus_field(False, self.note, ord):
-                    # something updated the note; update it after a subsequent focus
-                    # event has had time to fire
-                    self.mw.progress.timer(
-                        100, self.loadNoteKeepingFocus, False, parent=self.widget
-                    )
+                    self.update_note_fields_and_tags()
                 else:
                     self._check_and_update_duplicate_display_async()
             else:
@@ -486,20 +482,34 @@ require("anki/ui").loaded.then(() => require("anki/NoteEditor").instances[0].too
         elif hide:
             self.widget.hide()
 
+    @deprecated()
     def loadNoteKeepingFocus(self) -> None:
-        self.loadNote(self.currentField)
+        self.update_note_fields_and_tags()
+
+    def update_note_fields_and_tags(self) -> None:
+        if not self.note:
+            return
+
+        update_fields_js = "".join(
+            [
+                f'require("anki/NoteEditor").instances[0].fields[{index}].editingArea.content.set({json.dumps(value)});\n'
+                for index, value in enumerate(self.note.values())
+            ]
+        )
+
+        update_tags_js = f'require("anki/NoteEditor").instances[0].tagEditor.tags.set({json.dumps(self.note.tags)});\n'
+
+        # this is important in case the focused field was updated
+        # because having focus prevents the update from showing to the user
+        refocus_js = 'require("svelte/store").get(require("anki/NoteEditor").instances[0].focusedInput)?.refocus();'
+
+        js = update_fields_js + update_tags_js + refocus_js
+
+        self.web.eval(js)
 
     def loadNote(self, focusTo: int | None = None) -> None:
         if not self.note:
             return
-
-        data = [
-            (fld, self.mw.col.media.escape_media_filenames(val))
-            for fld, val in self.note.items()
-        ]
-
-        flds = self.note.note_type()["flds"]
-        descriptions = [fld.get("description", "") for fld in flds]
 
         self.widget.show()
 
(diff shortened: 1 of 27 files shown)


dae:
Let me say up front that I do think this will be a nice improvement over what we currently have - it will make reasoning about load/save easier when we don't have to worry about data flowing through Python, and it will be nice to reduce the amount of json strings we're build up to communicate with the editor.

I am a bit worried about what this may break though. How will the existing Python editor hooks fit into this? Could we perhaps hack in a way to continue supporting them (with some extra load/save calls of the note in Python) to ease the impact here? Or is it going to require add-on authors switch to a different approach, and mean that a bunch of unmaintained add-ons suddenly become unusable?

Your idea of creating new_format_pack to demo the features of the new editor/private a template was a great one, and I think we'd do well to put effort into making something like that for these changes as well, both as a way to test our changes, and to show others how things should be ideally done.

It would also be nice to think about whether we can perhaps introduce these changes behind a feature flag, especially if they're going to break all the existing hooks. We could perhaps start by separating out the changes that can be made now without causing a change in behaviour (eg mediasrv, proto.ts, etc), and perhaps that will provide a clearer view as to what may be possible to do with the remaining code. If the remaining code proves to be impractical to put behind a feature flag without a great deal of work then that's ok - at the very least, it's less code that needs to be reviewed at once. :-)

hgiesel:
All python hooks should be called properly after this. The old hook-related thing that is changed is the string that is passed into `editor_will_load_note` in editor.py:loadNote, and that few JS functions are not available anymore: `setFields`, `setFonts`, `setTags`, and `setDescriptions`.

The `getNote` logic is now achieved by directly accessing the backend, however the bridge commands are still sent to python, which then save the note there.

I basically tried to get the possibility of a `onLoadNote` hook in the fewest steps. This could be used for use cases like mentioned by @glutanimate in #1733.

(I'm not advocating for merging it now. I'm just saying I don't think this will break as much as you fear.)

dae:
That sounds promising, but I ran into an issue when I gave it a go with my [Japanese Support](https://ankiweb.net/shared/info/3918629684) add-on, which reads text from the first field on blur, writes altered text into the third field, and then expects the editor to show the inserted text in the third field. If I add a 'Japanese' notetype and try that with this PR, the text typed into the first field of the add screen disappears when I hit tab.

The 100ms delay before calling loadNoteKeepingFocus may be connected here - I believe it was done so that currentField could be updated. I don't mind updating this add-on, but it's a pattern that a number of other add-ons use as well, so it would be good if we could preserve it.

hgiesel:
I had a look into this:
- Normally (without any add-ons) the editor inside the AddCards dialog only calls loadNote when a new note is set. This PR builds upon this behavior. It also seems to me that the `editor_did_unfocus_field` is the only case where this happens.
- With this PR, the communication always has to go over the backend. So it's either: Python pushes to backend; tells JS that it should update; JS fetches from backend. In the case of this hook, the communication has to be Python -> JS, without going over the backend.

Instead of calling `loadNote` in the hook, I will add a new method in editor.py, which will update note fields and tags via web.eval. In fact, you could even say this is a bug fix: Currently using `editor_did_unfocus_field` means that the plain text history is reset.

qt/aqt/editor.py
```diff
@@ -483,20 +479,34 @@ def set_note(
         elif hide:
             self.widget.hide()
 
+    @deprecated()
     def loadNoteKeepingFocus(self) -> None:
-        self.loadNote(self.currentField)
+        self.update_note_fields_and_tags()
 
-    def loadNote(self, focusTo: int | None = None) -> None:
+    def update_note_fields_and_tags(self) -> None:
```
hgiesel, line 486: This method can be used to push the python `self.note` to Svelte.

hgiesel:
I've managed to find a way.
Instead of loadNote, the `editor_did_unfocus_field` hook now does one big `web.eval`, accessing the component tree to update the UI.

dae:
Sorry for the delay Henrik. Really happy to hear this use case is supported now! I'm still a bit nervous about potential other breakages, so this may still need to wait for a while. Once ~2.1.52 is out, I'll start being more liberal with merges into main again, and start backporting fixes into a stable branch, so hopefully not too much longer to wait until developers can test this out.

dae:
Just wondering what the best timing/approach is for merging this. I had been thinking about merging this into main after 2.1.52, and then starting to backport relevant changes to a separate stable branch (skipping the inclusion of this one in stable until later). If we merge this in after 2.1.52, how likely do you think it is to conflict with other editor-related changes destined for the stable branch?

Other alternatives would be:

b) wait for another 1-2 stable releases before merging it in. When it is merged in, avoid backporting it to stable until 2.1.55+
c) merge it after 1-2 releases, backporting as soon as merged

c) will minimize the chances of conflicts when backporting changes, but may introduce regressions in a stable release that are not caught by beta testing. If that happens, we'd either need to fix the issues promptly and put out another stable release, or revert the changes for now (to be able to do so, we'd want to avoid merging other changes that build off these ones for now).

Do you have any thoughts/preferences?

hgiesel:
The PR paves the way for some feature PRs I'm interested in, so it doesn't make sense to merge it in before we're heading for the next major update. So I'd be fine with option b.

dae:
I'm pretty happy with the state of 2.1.5x now, so I think 54 will be the last stable release before we can start focusing on a larger update again. The new Qt is due out in about 10 days, and I'll want to give it a week or so of testing in an RC. After .54 is out as stable, I can merge this in, if you think it's ready and will have the free time to deal with any issues that crop up (an alpha/beta for .55 would probably be a few weeks after merge at the least)

dae:
Sorry to drag this out Henrik, but I think we should get another smaller update out before we get stuck into a longer development round. I have a feeling once we start on this stuff it's going to take some months to get it all polished, and there is some lower-hanging fruit/bugs that is probably worth tackling first.

dae:

PR #1693, opened 2022-02-26
Use async functions in .svelte files

hgiesel:
This is an example of point 3 in this message: https://github.com/ankitects/anki/pull/1691#issue-1150683175.

Related: https://github.com/sveltejs/svelte-preprocess/issues/373


dae:
Would the change I pushed work instead? The next AM release will bump the minimum target to 13.4, so unless I've overlooked something, targeting 2019 should be safe at this point?

hgiesel:
Yes, that also works.

PR #1694, opened 2022-02-26
Consider using --force-message for ts/protobuf.bzl

hgiesel:
This would force us to be more explicit when providing input for the protobuf services, and avoids as have to cast the type when consuming output.

The first is arguable, but the second is much more ergonomic.
This means we don't have cast from `IOurType` to `OurType` anymore, but non-null assertion is still needed (which makes sense after reading your explanation [here](https://github.com/ankitects/anki/pull/1691#issuecomment-1051474992))


dae:
Thanks Henrik, even with the extra code at message creation time, this looks like a win. :+1:

PR #1695, opened 2022-02-26
Replace (some) Bootstrap dropdowns with Floating UI

hgiesel:
[Floating UI](https://github.com/floating-ui/floating-ui/) is the rebranding of Popper JS, and in terms of functionality somewhere between PopperJS and TippyJS.

I've mentioned before that using Bootstrap dropdowns caused me headaches before. I think with floating ui, we could write some dropdown/popover components, which would work cooperate much better with Svelte.


hgiesel:
I've now ported a single dropdown menu to Floating UI: the Latex menu in the editor. The new dropdown actually has some features, which the old Bootstrap dropdowns don't have:
- They have a small arrow pointing to their reference element (the toggle button).
- The are shifted away from the edges of the screen.

Probably the biggest convenience is that we have more control over when to show/hide the popover.

I've also added an [elevation](https://material.io/go/design-elevation) effect to the popover, to hopefully give it a more modern look.

I've also fixed the Cmd-Shift-T shortcut / the button, as it was just one line.

dae:
Seems to work well, and the corner avoidance is nice. Thanks for fixing the shortcut issue too!

While testing this I noticed a minor issue with the MathJax editor: the shift+enter mentioned in the placeholder text doesn't seem to work, at least on the Linux system I'm using here. It seems to be unrelated to this PR though; issue also exists in main.

PR #1698, opened 2022-02-28
Extend maximum answer time...

bruceharris:
Previously the time allowed to answer a question was capped at 10 minutes.
While this makes sense for fact recall, it limits the utility of Anki when
used for solving problems that can take more time to work through. This
extends the maximum answer time to 2 hours (7,200s), which seems to be
a reasonable upper limit for solving a math or algorithm question.

Screen Shot 2022-02-27 at 10 05 37 PM


bruceharris:
While this certainly isn't a classic use case, I use Anki to practice solving coding algorithm problems. I find that spaced repetition helps me get better at solving a wide range of problems. Often I can solve these in 2-3 minutes, but some can take more than 10 minutes, especially if I am less familiar with that class of problem. Breaking these into smaller problem cards would defeat the purpose, as the challenge of solving a more complex problem is to figure out precisely how to break it down. This max answer time is configurable by the user, and this PR doesn't change the default value, it just raises the max allowable value, which gives flexibility to the user to choose the tradeoff of making Anki less effective at detecting when you've stepped away.

dae:
I agree it's preferable to give users the option if they know what they're doing, but I'd also like to provide some guide rails. How about tweaking this PR to display a yellow warning if the configured time exceeds 10 minutes? It could say something like "Anki can schedule your reviews more efficiently when you keep each question short." It would need a new entry in deck-config.ftl, and a check like the ones done in DailyLimits.svelte.

bruceharris:
Sounds great, will do, thanks!

bruceharris:
Added warning when max answer seconds > 600
Screen Shot 2022-03-12 at 11 51 30 PM
Screen Shot 2022-03-12 at 11 51 23 PM

dae:
Thanks Bruce! One minor change and then it should be good to go - would you mind moving the warning below the field, to match the other warnings?

bruceharris:
Done, thanks!
Screen Shot 2022-03-14 at 8 19 23 AM

dae:
Great, thanks Bruce!

PR #1704, opened 2022-03-02
Categorize colors / skew existing color schemes

hgiesel:
In this PR I would like to tackle categorizing our colors into color shades with names such as `red-1`, `red-2`, `red-3`, and then use those for the semantic color names (such as `window-bg`), as we've discussed before (Sorry, couldn't find the original discussion anymore).

If possible, I would also like to realize what was mentioned [here](https://forums.ankiweb.net/t/brainstorming-for-modern-ui-anki-3-0/17510#h-7-use-a-white-background-for-light-theme-8) by @kleinerpirat and skew the base color (`window-bg`) into a darker tone for night mode/dark theme, and a lighter tone for light theme.

Here is a comparison of Anki's window background in dark theme, compared to some other apps/websites:
https://coolors.co/2f2f31-1e1e1e-1d1e20-191919-202123-1f2328
The first color is Anki, then follows VS Code, the Google website, a macOS window, Slack, and last GitHub. You can see that Anki is particular bright in dark mode. For design purposes, it would be nicer, if the base color would be darker, because that would allow more in range in brighter colors.

Here is a comparison of Anki's window background in light theme, compared to some other apps/websites:
https://coolors.co/ececec-f5f7f9-f7f7f7-ffffff
The first color is Anki, the second is on some GitHub sites (like the main page, where you see notification), the third is the Slack app, however most websites/apps actually are entirely white.

WDYT? Do you I don't think this needs to make it into 2.1.50, and I don't think changes in CSS files will conflict with any changes upstream any time soon, so I'll take this slowly.

### DIFF
--- a/sass/_vars.scss
+++ b/sass/_vars.scss
@@ -1,6 +1,228 @@
 /* Copyright: Ankitects Pty Ltd and contributors
  * License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html */
 
+/*
+- cyan tones
+5ccfca-00d1b5-399185
+
+5ccfca (medium turquiose)
+00d1b5 (turquiose)
+399185 (celadon green)
+
+
+
+
+
+- green tones
+93e066-86ce5d-5ccc00-54c414-559238
+
+00aa00 (carmin)
+93e066 (inchworm)
+86ce5d (mantis)
+5ccc00 (lime green)
+54c414 (kelly green)
+559238 (may green)
+
+
+
+- blue / cyan tones
+ccccee (lavender blue)
+77ccff (light sky blue)
+506aa3 (blue yonder)
+7edbd7 (middle blue green)
+
+- truely blue
+9dbcff (baby blue)
+6f9dff (cornflower blue)
+578cff (corn flower blue)
+194380 (dark cornflower blue)
+316dca (celtic blue)
+0969da (celtic blue)
+0000aa
+7777cc (violet blue crayola)
+
+ccccee-77ccff-506aa3-9dbcff-6f9dff-578cff-316dca-0969da-0000aa
+
+
+
+
+
+
+
+
+- red tones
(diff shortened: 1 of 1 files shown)


hgiesel:
I think the color schemes provided by AnkingMed in his add-on: https://github.com/AnKingMed/AnkiRecolor/wiki, can serve as a model to some degree. 

I'd imagine a new default light theme to look closer to his "Nord" light theme, and a dark theme to look closer to his "One Darkest theme".

dae:
:+1: for semantic names

I have no strong attachment to the current theme colors, so no objections to changes there, though it would be nice to see some screenshots / see what the community thinks before we commit to anything.

The themes on that wiki don't clearly state their license, so to avoid any potential licensing issues it would be best to create our own instead of using other ones as a reference.

kleinerpirat:
I don't feel comfortable choosing colors myself, because I suffer from deuteranopia. But I found this open source color scheme (MIT License) that looks very promising: https://yeun.github.io/open-color/

image

@hgiesel If you don't mind, I'd take up on your idea and propose a PR soon.

kleinerpirat:
I just found out `_vars.scss` isn't really compiled as SCSS. Sass variable syntax remains plain text:

image

Is this intentional? Ideally, the color shades should not be exposed as CSS variables, only the semantic names. This input:
```scss
@use "open-color";

:root {
    --text-fg: $oc-gray-9;
}
```
would give the following output if the `.scss` were actually compiled:
```css
:root {
  --text-fg: #212529;
}
```
---
With vanilla CSS variables, we'd have something like this, which is less readable and exposes way too many CSS variables:
```css
:root {
  --oc-gray-9: #212529;
}
:root {
  --text-fg: var(--oc-gray-9);
}
```

dae:
https://github.com/sass/libsass/issues/2621

dae:

PR #1705, opened 2022-03-02
Fix newline shortcut in MathjaxMenu

hgiesel:


dae:
Always nice when removing code fixes things :-)

PR #1706, opened 2022-03-03
Add a flag to allow loading the fields dialog in a web view

hgiesel:
I'd like to defer https://github.com/ankitects/anki/pull/1691 and other impactful changes in order for 2.1.50 to finally arrive.

Instead I'd like to work on the fields web view. I'd like to implement it as sidebar, which knows about all fields, and a field editor, which only knows about a single field object. As such it would have similar setup as the parent container and NoteEditor, and I could try out how things would work best without possibly delaying 2.1.50.

### DIFF
--- a/qt/aqt/fields.py
+++ b/qt/aqt/fields.py
@@ -1,6 +1,7 @@
 # Copyright: Ankitects Pty Ltd and contributors
 # License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
+import os
 from typing import Optional
 
 import aqt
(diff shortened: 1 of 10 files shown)


qt/aqt/fields.py
```diff
@@ -41,28 +42,53 @@ def __init__(
...
+            self.webview = form.webview
+            self.webview.set_title("fields")
+
+            self.show()
+            self.refresh()
+            self.webview.set_bridge_command(self._on_bridge_cmd, self)
+            self.activateWindow()
+
+        else:
+            self.form = aqt.forms.fields.Ui_Dialog()
```
dae, line 63: Minor nit: if you return inside the if clause, then the original code doesn't need to be indented, and it will result in a smaller diff/less chance of a conflict.

dae:
Happy with this approach, we should be doing more development behind feature flags. :-) You may be able to fix the mypy error by adding this to the top of the file:

dae:
Strange, this is passing when I check it out locally

dae:
:man_shrugging:

hgiesel:
Hm, after this PR, the fields web view still doesn't seem to cleaned up properly. Opening and closing the fields web view a few times, and then opening the web inspector, it will still list all of them.
Do you have any idea what might cause this?

dae:
On reflection, I suspect why CI was failing is because it's missing some library that's required at runtime when importing PyQt6. That would have caused it to fall back to PyQt5, and since we weren't allowing for that case, it was breaking the test with a vague error.

I've pushed a fix to main that should ensure the webviews get cleaned up.

PR #1707, opened 2022-03-03
Add regex tag search

RumovZ:
Like in field searches, this is case-insensitive and doesn't require a full match.

In https://forums.ankiweb.net/t/how-to-search-for-head-tags-without-sub-tags/17788 I was talking about the non-regex case, but I see why it wouldn't work. Were you also talking about that and propose to use `regexp_tags` for it as well?


dae:
Thanks Rumo, this all looks good to me.

Re converting the glob behaviour to use regex_tags, assuming we keep the same behaviour (exact matching, implicit children inclusion), I don't have much of a preference. Using the same splitting approach might slightly simplify the code, but the existing approach isn't particularly complicated. Either works for me. :-)

PR #1709, opened 2022-03-04
Disable full screen for OpenGL on Windows

RumovZ:
https://forums.ankiweb.net/t/anki-2-1-50-beta-3-4/17501/45

### DIFF
--- a/ftl/core/actions.ftl
+++ b/ftl/core/actions.ftl
@@ -53,6 +53,7 @@ actions-previous-card-info = Previous Card Info
 # By convention, the name of a menu action is suffixed with "..." if additional
 # input is required before it can be performed. E.g. "Export..." vs. "Delete".
 actions-with-ellipsis = { $action }...
+actions-unsupported-video-driver = Not supported for your video driver
 
 ## Flags
 
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -49,6 +49,7 @@ from aqt.utils import (
     current_window,
     ensure_editor_saved,
     getTag,
+    maybe_disable_full_screen_action,
     no_arg_trigger,
     openHelp,
     qtMenuShortcutWorkaround,
@@ -191,6 +192,7 @@ class Browser(QMainWindow):
 
         # view
         qconnect(f.actionFullScreen.triggered, self.mw.on_toggle_full_screen)
+        maybe_disable_full_screen_action(f.actionFullScreen, f.menu_view)
         qconnect(
             f.actionZoomIn.triggered,
             lambda: self.editor.web.setZoomFactor(self.editor.web.zoomFactor() + 0.1),
--- a/qt/aqt/forms/browser.ui
+++ b/qt/aqt/forms/browser.ui
@@ -298,7 +298,7 @@
     <addaction name="separator"/>
     <addaction name="actionDelete"/>
    </widget>
-   <widget class="QMenu" name="menuqt_accel_view">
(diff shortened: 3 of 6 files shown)


dae:
I suspect users won't think to hover over the item, and it doesn't work when holding the mouse down. Maybe we should add the check inside the action, and show a popup instead?

Also, if we assume Qt6 is the future, and assume we can't find a way to detect the active driver, that means users won't be able to work around the issue by adjusting their settings. Maybe it might be worth tweaking the error message to something like the following?

"This action is not currently supported on your system."

or even

"This action is not currently supported on Windows."

or maybe we should just be hiding the menu item?

RumovZ:
I didn't know prospects were so bleak. If the only solution is Qt5 and you don't plan to provide this alternative going forward, I agree that hiding the option might be best.

dae:
There may be a way to detect the driver, I'm not sure - the approach we were using for Qt5 (opengl_vendor()) doesn't work in Qt6, and I haven't dug into it further.

Life was so much easier before 3D acceleration!

RumovZ:
Sorry, I don't understand why detecting the driver is so essential. Isn't it just OpenGL-proper by default, and _software_ if `QT_OPENGL=software` is set? I can only see `opengl_vendor()` being used on Linux.
And aren't there regularly other issues beside full screen mode that require changing the driver to fix them?

dae:
Currently in Qt6, driver selection is not available in the Preferences screen, and QT_OPENGL is ignored. If it's working on your software-only system, presumably it's because Qt6 is falling back to software automatically, and the question then becomes - can we tell? If we know the issues don't exist with software rendering, and find out a way to detect the current rendering mode, then enabling this feature selectively sounds fine.

There is a separate ANKI_SOFTWAREOPENGL=1 flag that is still honored.

As for why QT_OPENGL is currently ignored, when I was initially experimenting with Qt6, I seem to recall that setting QT_OPENGL=auto lead to different behaviour to just leaving it unset, in the case that graphics acceleration was available. Moving forward, we may have more success setting flags that control the behaviour of Chromium directly, such as the ones mentioned here:
https://forums.ankiweb.net/t/anki-opens-but-all-windows-are-blank/8920/17

I'm not yet sure what the optimum flags would be; once we get 2.1.50 into the hands of more users, it should give us a better idea of what options we can offer users to work around issues.

RumovZ:
> If it's working on your software-only system, presumably it's because Qt6 is falling back to software automatically, and the question then becomes - can we tell?

Both `QT_OPENGL=software` and `ANKI_SOFTWAREOPENGL=1` work for me. With either of them set, `Using Software Dynamic GL: yes` is printed to the console (`no` otherwise), and the various full screen glitches are gone.
I don't have the full picture since it depends on graphics accleration, but I'd selfishly propose to enable full screen mode if either of those variables is set. If you want to collect more information with 2.1.50, maybe a pop-up with a link to a manual page could be shown when trying to enter full screen on Qt6-Windows. The manual page would contain instructions to set `QT_OPENGL=software`.

dae:
Ok, sorry, I was mixing up the native Qt env vars with our gl driver settings - we don't explicitly set QT_OPENGL ourselves in Qt6, but that doesn't stop Qt from reading env vars the user has set. I'd been assuming you were running Qt6 without any env vars set and it was defaulting to software; I didn't realise you'd forced it into software more. In that case, no objections to conditionally showing/enabling the menu item when the driver is forced to software. If it later turns out that some users get issues even when software mode is forced, we can revisit the decision, and maybe put it behind a separate flag.

PR #1710, opened 2022-03-04
Fix html_media_regexps

BrayanDSO:
Some regexes on `html_media_regexps` were flawed, matching other tags, as you can see on:

src element quoted case: https://regex101.com/r/JXSruk/1

unquored case: https://regex101.com/r/F263nV/1

---

If anyone wants to test the fixed ones:

quoted: https://regex101.com/r/8KiURM/1

unquoted: https://regex101.com/r/ps45IC/1

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -97,6 +97,7 @@ Sam Bradshaw <samjr.bradshaw@gmail.com>
 gnnoh <gerongfenh@gmail.com>
 Sachin Govind <sachin.govind.too@gmail.com>
 Patric Cunha <patricc@agap2.pt>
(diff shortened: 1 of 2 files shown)


dae:
Thanks Brayan! The majority of Anki's code uses HTML_MEDIA_TAGS defined in text.rs, and these regexps will likely be deprecated once the import/export code gets an update. No issue with correcting them for now, though please note you also need to add object to the list there, and please update the CONTRIBUTORS file.

@nwwt heads up

BrayanDSO:
Done

dae:
Thanks Brayan! And sorry, please ignore the comment about adding object; I was not reading carefully enough.

PR #1713, opened 2022-03-06
Do not prevent drop event in AnkiWebView

hgiesel:
Is there a reason we straight-out prevent it? It does not seem to interfere with DnD in the deck overview or the editor.

### DIFF
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -321,9 +321,6 @@ class AnkiWebView(QWebEngineView):
         gui_hooks.webview_will_show_context_menu(self, m)
         m.popup(QCursor.pos())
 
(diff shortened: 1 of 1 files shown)


dae:
That line was added 11 years ago, when we were using Qt4/WebKit. I think it was added because dragging an image onto the webview would cause it to change the location to file:///path/to/image, replacing whatever we were showing in the webview. That no longer seems to happen, but that's because of our URL handler - with this change, dropping an image on the overview screen causes the image to load in a separate browser window. We probably don't want that, so maybe it's better to keep this, and explicitly enable a drop event when we need it (like the editor does)?

hgiesel:
Closed, this will be dealt with in #1716

PR #1714, opened 2022-03-08
Optionally restore original position and reset counts when forgetting

### LINKED ISSUE #1700: consider changes to 'forget'
- Some users would like to reset the lapse count to delay leech flagging
- Some users aren't sure what 'Forget' will do

Associated thread:

https://forums.ankiweb.net/t/improving-forget/8644/7

RumovZ:
Closes #1700.


qt/aqt/browser/browser.py
```diff
@@ -862,10 +862,13 @@ def set_due_date(self) -> None:
     @skip_if_selection_is_empty
     @ensure_editor_saved
     def forget_cards(self) -> None:
-        forget_cards(
+        if op := forget_cards(
             parent=self,
             card_ids=self.selected_cards(),
-        ).run_in_background()
+            restore_position_key=Config.Bool.RESTORE_POSITION_BROWSER,
+            reset_counts_key=Config.Bool.RESET_COUNTS_BROWSER,
```
RumovZ, line 869: This doesn't scale very well. With every new option, callers need to pass in one more argument.
Maybe there should be a `ConfigContext` enum instead? Then callers would only need to pass `ConfigContext.BROWSER`.
There'd still need to be two new keys for every new option, unless there is some `ConfigBoolWithContext` routine, though.

dae: Agreed that a single context enum would be better that passing multiple keys in. In the long term we may want something like ConfigBoolWithContext, but until we see 1-2 more similar uses, I'd be tempted to just have the backend fn look up specific keys based on the provided context.

dae: One downside of this new approach that I neglected to consider is that the frontend still needs to know the key names so it can look up the previously-set values. I'll add a separate method for fetching the defaults, so the calling code only has to think about the context, and not individual keys.

I'd also been thinking that the context would be specific to this method, instead of one that is shared with other routines. My reply didn't make that clear - apologies for the confusion. The problem with a shared context is that different routines may have different valid values. Say we have a different action in the future that wants a Context.REVIEWER and Context.PREVIEWER. If we have various routines sharing the same context, we will lose some of the advantages of enums, as then we run the risk of the calling code accidentally passing in an enum variant that a routine doesn't support, and we don't get notified about it at compile time.

If you feel things should be done differently, please let me know - I'm making these changes now to get a beta build out, and we can change them again if necessary.

pylib/anki/scheduler/base.py
```diff
@@ -15,6 +15,8 @@
 UnburyDeck = scheduler_pb2.UnburyDeckRequest
 BuryOrSuspend = scheduler_pb2.BuryOrSuspendCardsRequest
 CustomStudyRequest = scheduler_pb2.CustomStudyRequest
+ScheduleCardsAsNew = scheduler_pb2.ScheduleCardsAsNewRequest
+ScheduleCardsAsNewDefaults = scheduler_pb2.ScheduleCardsAsNewDefaultsResponse
```
dae, line 19: Renaming these makes it a bit more friendly for the frontend code, and less likely that they'll accidentally import from our semi-private _pb2 files.

qt/aqt/operations/scheduling.py
```diff
@@ -65,10 +66,37 @@ def set_due_date_dialog(
...
+    disable_help_button(dialog)
+    form = aqt.forms.forget.Ui_Dialog()
+    form.setupUi(dialog)
+
+    if context is not None:
+        defaults = aqt.mw.col.sched.schedule_cards_as_new_defaults(context)
+        form.restore_position.setChecked(defaults.restore_position)
+        form.reset_counts.setChecked(defaults.reset_counts)
+
+    if not dialog.exec():
```
dae, line 86: Ideally we should avoid .exec() calls, instead calling .open() after attaching to the `accepted` slot. It's a bit more cumbersome, but avoids some issues:

https://github.com/ankitects/anki/pull/1684#discussion_r813434377

Ideally we should also be querying the defaults with QueryOp(), but unfortunately that doesn't play well with our current approach of returning a CollectionOp so that the caller can adjust the success callback.

PR #1716, opened 2022-03-08
Vision for new Fields dialog

hgiesel:
This is not yet close to being finished, but I think it can be discussed:

To get to the new fields screen, you need to start Anki with the environment variable `ANKI_EXPERIMENTAL_FIELDS_WEB` set.

This is what it currently looks like. The colours are just placeholders:

Screenshot 2022-03-08 at 20 52 20

There are two major "structural" components: The drawer (aka sidebar) and the panes to the right.

Things that work:
- [x] scroll the drawer and each pane individually
- [x] click on panes to make them active (their icon will be blue)
- [x] click on drawer items to make open in the current pane
- [x] ctrl-click on drawer items to open in a v split
- [x] ctrl-shift-click on drawer items to open in a h split
- [x] ctrl-alt-click to open it in the current pane and close all others
- [x] collapse according to hierarchical structure in the drawer
- [x] rearrange items in the drawer via drag'n'drop (However only the first reorder will behave correctly, because the underlying fields are not mutated yet by this action).
- [x] close panes by clicking the top right "x" 

Things that do not work yet:
- [ ] reorder the panes by drag'n'drop the pane header
- [ ] open drawer items by dragging them onto the pane area
- [x] resize the drawer
- [ ] hide the drawer
- [ ] there are currently ways to have no current pane. This should be avoided. There should be always exactly one current pane.
- [ ] indicate active drawer item (which associates with current pane) in drawer
- [ ] I've simply allowed drop events like [in this PR](#1713) 
- [ ] Our own on:contextMenu
- [ ] Key shortcuts for navigating the drawer items
- [ ] actually edit the underlying field object

While this is currently only hosts the fields, my intention is to provide the functionality of every dialog that can be currently reached via the Manage Notetypes window in this view. My vision is to provide several things which are currently only reachable via sub dialogs directly in the drawer as an item. (related: https://forums.ankiweb.net/t/brainstorming-for-modern-ui-anki-3-0/17510/3#my-own-points-3)

@kleinerpirat, I think you might also be interested in this.


ts/fields/FieldsPage.svelte
```diff
@@ -1,8 +1,212 @@
...
+    import { randomUUID } from "../lib/uuid";
+    import type { DrawerItemType } from "./drawer";
+    import { Drawer } from "./drawer";
+    import DrawerTitle from "./DrawerTitle.svelte";
+    import FieldConfigEditor from "./FieldConfigEditor.svelte";
+    import { fieldIcon } from "./icons";
+    import InsertionPoint from "./InsertionPoint.svelte";
+    import { Panes } from "./panes";
+
+    const ntid = 1641830698585;
```
hgiesel, line 21: Heads-up for those who want to try out the PR: this notetype id is hardcoded! You need to exchange it with one that exists on the current profile.

dae:
Will check this out in the next few days and get back to you

hgiesel:
> Horizontal/vertical splits don't seem so useful for the fields list, but I presume you have other things like the card templates in mind for them.

Yes. It does have some consequences for how the communication with the server should work though. E.g. our current deck options window maintains its own "options" state, which can be discarded by closing the window.

But now imagine two panes open with the same deck options. If both panes have unsaved changes to the same menu point, and now one saves, what should happen? Should the other one discard the other change, or keep it? 

From my observation, most web sites or web apps do it similarly how we do it in the note editor: Entered changes are immediately saved to the database.

My idea is that we solve this conflict here on the small scale :)

> Ctrl+click on an item

Of course that should have Cmd-Click on a Mac. However our own onContextMenu should probably also be part of this PR.

> some users may want to be able go to next/previous entries

I'll also take this into the list.

dae:
> But now imagine two panes open with the same deck options. If both panes have unsaved changes to the same menu point, and now one saves, what should happen? Should the other one discard the other change, or keep it?

> From my observation, most web sites or web apps do it similarly how we do it in the note editor: Entered changes are immediately saved to the database.

A note update is quite different to a schema change though. Writing out note changes requires only a small amount of database I/O, and the backend takes care of coalescing multiple changes made in short succession into a single undo entry.

With fields, if the user changes the position of a field, all notes of that notetype need to be rewritten with an updated flds entry, which is potentially slow. Older Anki versions applied the changes as the user made them, as that was easier to implement, but it meant that when you wanted to reposition multiple fields, you had to wait for it complete after each field was repositioned. 2.1.28 moved to an accumulate-and-apply-in-bulk approach, which meant the I/O only needs to be performed once at the end, and the individual UI operations can be fast. If we change the UI to apply changes immediately, that feels like it would be a step backwards in terms of performance.

One other thing to bear in mind is that the current backend code exposes a method to update fields, and one to update templates. The former will automatically update field references inside templates when saving, for cases where fields have been renamed. If we want to show both fields and templates on the same screen, that is going to require a rework of that code, so that the "in memory" copy of the templates is updated correctly as fields are renamed.

Concurrent updates are a complicated problem, and Anki makes liberal use of modal dialogs at the moment to reduce the number of areas we need to deal with them. In your example, can't we keep that screen modal? Then we don't have to worry about separate windows modifying the same data, and provided we're using stores for the data, the same data being displayed twice in a split should update at the same time.

hgiesel:
From a UI/user perspective however I think "automatic saving" is the kind of behaviour users have become to expect. I don't know any mobile app or website, which does it any other way. However it's probably not in the scope of this PR to do anything about it.
The closest we can get to it currently is probably this:

Screenshot 2022-03-16 at 02 35 51

You can discard changes, or save any pane individually. Panes which originate from the same drawer item do so in sync. If we ever can allow it (by making the sync more append-only, or maybe lazily update notes), we could just remove those. 

> If we want to show both fields and templates on the same screen, that is going to require a rework of that code, so that the "in memory" copy of the templates is updated correctly as fields are renamed.

This is also not in the scope of this PR, but something I'll keep in mind. We could probably even solve that UI-wise, as the drawer will build the notetype in a hierarchical manner.

> the same data being displayed twice in a split should update at the same time.

Yep, that was my brain-fart. Opening the same webview twice is also something which is _interesting_, but I don't see anytime soon. It's also a feature that becomes more helpful, the more functionality is merged into the same webview.

dae:
> From a UI/user perspective however I think "automatic saving" is the kind of behaviour users have come to expect. I don't know any mobile app or website, which does it any other way.

Immediate application is certainly common, but I think you're overstating things a little :-) Think of the various wizard screens that apps present, that allow you to proceed/go back, and don't typically apply changes until you reach the end. And there are screens like the iOS IP settings page that have options to save or to go back without making changes:
https://photos5.appleinsider.com/gallery/25722-35518-ios-dns-change-iPhone-iPad00005-xl.jpg

> You can discard changes, or save any pane individually.

Are you proposing a general approach to the UI here Henrik, or are you suggesting something to be used specifically with fields? The way the backend code works is that all changes made to all fields (edits, deletions, repositions, sort field changes, etc) are applied in one go - it doesn't support changing one field at a time, because it wants to make all changes at once to avoid multiple heavy operations. And even if we changed it to support updating individual fields, how would individual save buttons work with things like field reordering, or deleting fields?

The current field/template backend code performs well, and is robust. While changes to it are possible, they will come at the cost of engineering time, and possible regressions. I guess the question is - is this such a pain point for users at the moment that the downsides would be worth it? Or might it perhaps be better to make UI changes within the constraints of the current API for now, and use the time we save to work on other areas/features?

hgiesel:
> Are you proposing a general approach to the UI here Henrik

It was meant as a general approach, but once again you totally make sense.

However I can still work on the PR, even without it. 

> is this such a pain point for users at the moment that the downsides would be worth it?

I'd certainly say it's a pain point for everybody who edits/maintains their templates. Having ultimately 4 dialog windows merged into one web view, and being able to work on templates+fields or templates of different notetypes at the same time would be a great feature.

> it [might] perhaps be better to make UI changes within the constraints of the current API for now, and use the time we save to work on other areas/features?

I think I can work with the constraints, yes. It would mean that fields editing and templates editing live separately in their own webview, and there would have to be a save button somewhere (possibly in the drawer, or an additional top bar), just like in the deck options.

dae:
> It would mean that fields editing and templates editing live separately in their own webview

Could we perhaps start with that, and then merge them in at a later date? Would the merging process be reasonably simple from the UI end, or would that require a great deal of extra work?

From the backend side of things, while moving to saving individual field/template changes immediately doesn't seem too practical at the moment, merging the separate "save fields" and "save templates" actions into a single action doesn't seem _too_ complicated - the main issue that comes to mind being that templates need rewriting when fields are renamed/deleted. Our template handling code is in Rust, so to be able to apply transforms to the templates, we'd need to either:

- send the fields and templates back to the backend, asking it to apply field changes to the templates in memory, and sending them back. 
- or implement template parsing on the frontend. We could possibly reuse our Rust template code to do this, by factoring it out into a separate module and compiling it with wasm. That would avoid roundtrips to the backend, and avoid us having to reimplement the parsing code in typescript, but it's not something Anki has done before, and it would need experimentation to determine whether it's practical or not in terms of required changes/wasm size/etc. Definitely worth exploring at one point, but could be a bit of a rabbit hole that I'm not sure we have time to fall into at the moment :-)

If merging the two screens later wouldn't require throwing out a lot of work, I'd be tempted to run with that for now, and return to the merging at a later date when more of the UI has been updated. WDYT?

Will shoot you an email with some more thoughts

hgiesel:
> Could we perhaps start with that, and then merge them in at a later date? Would the merging process be reasonably simple from the UI end, or would that require a great deal of extra work?

That would certainly work, and not cause much overhead to merge. 

> merging the separate "save fields" and "save templates" actions into a single action doesn't seem too complicated

If that allowed us to merge a hypothetical fields and templates webview, that would be great, but I think there'd still be a big danger, that a user forgets the final "Save" click, and discards a lot of unsaved changes.

I agree that having the template adjusting on the front end could make sense. After all, it could also be configured. E.g. delete "Field1", and replace all appearances with "Field2", etc.

However I'd say that I have a good enough perspective to get this PR somewhere, and a few after that :)

dae:
>  that a user forgets the final "Save" click, and discards a lot of unsaved changes.

Could we perhaps solve that by asking the user to confirm they want to discard the changes when they attempt to close the window, like we do for the add screen?

kleinerpirat:
What do you guys think about a batch-editing feature? Full disclosure, I have personal interest in this because my main notetype has 200+ fields.

I'm thinking of selecting multiple fields in the sidebar with a shift-click, then a special pane called "multiple fields" pops up and you can perform an action on all of them at once, e.g. remove them, change the font size or do a RegEx rename. I'm aware this is a niche requirement, but with this beautiful sidebar, it seems like the natural thing to add.

dae:

Issue #1708, opened 2022-03-03
Tag suggestions do not show up when adding a new card in fullscreen


computeristic:
When I am adding a new card and I start typing the first few letters of existing tags in the 'Tag' input bar, suggestions show up (then I can press 'Enter/Return' to autocomplete). However, when I make the window fullscreen, these suggestions do not show up anymore. This never hapenned before today but I did not change or update anything (anki or OS). I tried restarting anki and restarting my macbook. I also tried disabling all add-ons and then restarting anki. 

Anki: Version ⁨2.1.49 (dc80804a)⁩ Python 3.8.6 Qt 5.14.2 PyQt 5.14.2
OS: macOS Catalina 10.15.7 (19H1715)

dae:
Can you reproduce the issue with the latest beta?

computeristic:
The issue is not reproducible in the latest beta, it works just fine. I'll be using the beta until the next update, thank you!

dae:
Thanks for testing!

computeristic:
@dae no problem! However, I noticed a change in behavior (which may be intentional). In 2.1.49, pressing enter/return after getting a tag suggestion autocompletes my partially entered tag. In the latest beta, I have to arrow up or mouse click the suggestion for autocomplete. This can be cumbersome when adding many cards and multiple tags, but I understand if it is intentional in order to prevent accidental autocompletions.

hgiesel:
@computeristic
You can also use the Tab key to navigate through the items, so to select the first item, you can press Tab, and then Enter.

This was a deliberate change from the old behaviour, because under the old behaviour, there was no visual indication that the first item of the suggestions was "selected". However we're still opening to adjusting the behavior, and making it easier to use.

computeristic:
@hgiesel ohh this is even better. I used to keep on typing until I eliminate all suggestions but the one I wanted. Navigating with Tab and Enter is better. Thank you for your work on the tag editor! I really appreciate FOSS developers and contributors.

computeristic:
An issue that I stumbled upon while adding new cards in the latest **beta** is that I cannot access the tags via the Tab key (and I cannot Tab out of the tags either).

PR #1717, opened 2022-03-10
Fix outstanding tag editor issues

hgiesel:
- Fixes the suggestions scrolling the window. There are still some weird things that can happen if you use very, very long tags. I'm not sure if we can eliminate them all, without dramatically changing how the tag editor works.
- Clicking on a tag suggestion now accepts it, rather than just updating the input
- You can not tab **out** of the tag editor (https://github.com/ankitects/anki/issues/1708#issuecomment-1061156768). However you can still not tab into it. Instead the user should use Ctrl+Shift+T.
- Both tag menus (autocompletion menu and selected tag button) now use `WithFloating` instead of `WithDropdown`.
- There were actually some hardcoded strings in the tag editor


dae:
Thanks Henrik! Seems to work well, and the new popup seems to look a bit nicer.

The popup width changes depending on the available tags, which I seem to remember happened in the past too - could we just use the full width always to prevent the jumping as the user types?

The tab handling is not ideal - once the user has tabbed out into the bottom buttons, tab just rotates through those buttons, and the user can't return to the webview fields or the tags area without clicking on them. If this is not an easy fix, we can postpone it for a future release, but it would be nice to address at one point.

Some other minor issues I spotted:

- there's a console.log() in TagEditor.svelte
- In night mode on Linux, the popup triangle appears as a diamond

hgiesel:
I've fixed the autocompletion menu at 80vw. However it will still jump, if the new tag exceeds the window width in the tag input.

> The tab handling is not ideal

This would be indeed easier once we move the remaining editor buttons into the web view.

PR #1721, opened 2022-03-15
Fix add-on usages of `addMedia`

### LINKED ISSUE #1720: addMedia() has broken
While fixing attaching of media on Windows in 0d83581ab08e1a92c5c55c99965e9f054b3f43e6, we accidentally broke programmatic access by add-ons, as addMedia() assumes that the focus handler has already been set up when it is called.

(reported on https://forums.ankiweb.net/t/anki-2-1-50-beta-6-stable-release/18181/18)

One way we could make the separate setup/apply steps clearer would be something like this:

```typescript
    let mediaPromise: Promise<string>;
    let resolve: (media: string) => void;

    /// Setup media to be inserted into the current field
    /// after `insertMedia` is called.
    function prepareToInsertMedia(): void {
        if (!editingInputIsRichText($focusedInput)) {
            return;
        }

        [mediaPromise, resolve] = promiseWithResolver<string>();
        $focusedInput.focusHandler.focus.on(
            async () => setFormat("inserthtml", await mediaPromise),
            { once: true },
        );
    }

    /// Insert media into the field that was active when
    /// `prepareToInsertMedia` was called.
    function insertMedia(media: string): void {
        resolve(media);
    }

    registerPackage("anki/TemplateButtons", { prepareToInsertMedia, insertMedia });

    function onAttachment(): void {
        bridgeCommand("attach");
    }
```

Then the attach command could call prepareToInsertMedia() prior to presenting the dialog, and insertMedia() after the user had made a selection:

```python
    def onAddMedia(self) -> None:
        self.web.evalWithCallback(
            'require("anki/TemplateButtons").prepareToInsertMedia()',
            self._after_media_prepare,
        )

    def _after_media_prepare(self, val: Any) -> None:
        extension_filter = " ".join(
            f"*.{extension}" for extension in sorted(itertools.chain(pics, audio))
        )
        filter = f"{tr.editing_media()} ({extension_filter})"

        def accept(file: str) -> None:
            self.addMedia(file, prepared=True)

        file = getFile(
            parent=self.widget,
            title=tr.editing_add_media(),
            cb=cast(Callable[[Any], None], accept),
            filter=filter,
            key="media",
        )

        self.parentWindow.activateWindow()

    def addMedia(
        self, path: str, canDelete: bool = False
    ) -> None:
        """canDelete is a legacy arg and is ignored."""

        try:
            html = self._addMedia(path)
        except Exception as e:
            showWarning(str(e))
            return

        self.web.eval(
            f'require("anki/TemplateButtons").insertMedia({json.dumps(html)})'
        )
```

There are two remaining problems though:

- addMedia() still expects an extra preparation step now. If it's fairly simple to fix, making it do the preparation step as well when called by legacy callers might be nice. But if it's not a simple fix, maybe we should be renaming the function and adding a docstring explaining how it needs to be used.
- A focus change is required for the image to be inserted, so if an add-on inserts an image without popping up a dialog, the user needs to click on a different window and then return to the editing window before the image is actually inserted. Maybe we could have insertMedia() automatically trigger the insertion in that case.

Other ideas welcome!

hgiesel:
Alternative solution to close #1720.

This basically restores the old version of `addMedia`, but also deprecates it and only exposes it for add-ons.


dae:
Thanks Henrik, this seems like the better approach for now.

PR #1722, opened 2022-03-16
Colpkg fixes

dae:
@RumovZ I'd appreciate it if you could give this a review, as it was a bit of a rush job


RumovZ (approved):
LGTM. I should have written some tests sooner.
As we'll soon build on this code, I can take care of the fixmes and refactor a bit after the merge if you like.

dae:
Ultimately the issues are my fault - I was in a hurry to get a new beta out, and rushed this through a bit quickly. Hopefully ok now anyway, and appreciate the review. :-)

qt/aqt/importing.py
```diff
@@ -491,14 +491,12 @@ def on_done(future: Future) -> None:
         timer.deleteLater()
 
         try:
-            soft_error = future.result()
+            future.result()
         except Exception as error:
             if not isinstance(error, Interrupted):
                 showWarning(str(error))
             callback(False)
```
RumovZ, line 498: My motivation was to differentiate between these two sorts of errors:
1. Fatal, everything should be reverted to the state before the import attempt.
2. Exit early, the collection has been committed and only some media files are missing.

In the second case, the callback should be called with `True`, so the GUI can act accordingly. E.g. if a media is missing, the new collection won't open now.

dae: I think we may have been looking at separate things? I was looking at the File>Import case, which opens the collection regardless. Backup opening does depend on the callback result, but given that backups do not contain media, I wonder if this is a situation we would actually hit?

In any case, exposing 'partial success' to the GUI is totally reasonable, I just think it might be better done with a separate error variant instead of a string return value, as it's easy to accidentally miss a failure in the latter case, as I did earlier today :-)

RumovZ: Right, it's only the tooltip in the File>Import case, but that's still essential feedback in my opinion. I totally agree that my string solution wasn't great and a flag on the error is much better.
Just so we're on the same page, is there any reason why only automatic backups should be opened from the profile window? Until now I thought it wouldn't matter from where an import is triggered. Surely manual full backups would be imported from the profile window, and even if it's a transfer between devices, the user might save a click to open the profile and import the collection from there.

dae: I've updated the code to pass the error variant in to the GUI. The messaging is not perfect - the user is notified about a single error but then told the import completed, which could easily be mistaken as it going fine apart from one media file. We might need to use a custom error string for that case to make it clearer. Alternatively, I still don't think returning an error and skipping the final collection rename would be so bad. Depending on the design of a deck, missing media files could be considered "fatal" by the user, and not replacing the collection sends a strong signal that importing was aborted. They may end up with some imported media files, but those can be cleaned up with a media check. But no strong feelings here, arguments could be made for either approach :-)

> is there any reason why only automatic backups should be opened from the profile window?

Originally there was only File>Import. But it had two downsides:

- users with a corrupt collection weren't able to open their profile, so they couldn't import
- opening a collection normally triggered a backup, and as there was no throttling, trying multiple times could end up destroying all the old backups

Open Backup was added to address those, and my assumption was that people would continue to use the old workflow for importing non-backup colpkgs, as it's the same workflow as for importing other types of files.

Also bear in mind that many users don't use multiple profiles, so they may never see the profiles screen unless they explicitly navigate to it. It also pops up when the collection is correct and can't be loaded.

dae: (but to answer your question, it's mainly that I expected users to use "open backup" only for ..opening backups :-)

RumovZ: > Alternatively, I still don't think returning an error and skipping the final collection rename would be so bad.

You mean treating media errors like all other errors and completely rolling back? I wouldn't mind that. The current behaviour may only exist because I didn't use a tempfile in the beginning and rolling back wasn't trivial.

Shall I take care of that and continue with the export code?

dae: If that works for you, that might be the best course of action at this point :+1:

PR #1723, opened 2022-03-17
Refactor export-import code and resolve fixmes

RumovZ:
Not at all sure about 4124571af654d606020b8145828033f3fdbf7499.
This PR is supposed to address issues that should make it into 2.1.50. Let me know if there is more.


dae (commented):
Lovely and neat! I just had two comments; everything else looks great.

RumovZ:
Both done! 👍

rslib/src/storage/upgrades/mod.rs
```diff
@@ -81,9 +79,10 @@ mod test {
     use super::*;
 
     #[test]
+    #[allow(clippy::assertions_on_constants)]
     fn assert_18_is_latest_schema_version() {
```
dae, line 83: Nice

dae:
Thanks Rumo :-)

PR #1725, opened 2022-03-18
Fix `NotFoundError` when opening browser after switching profiles

hikaru-y:
Steps to reproduce:
1. Go to the review screen
2. <kbd>File</kbd> -> <kbd>Switch Profiles</kbd>, then open another profile
4. Open the browser before going to the review screen
5. `anki.errors.NotFoundError` will be thrown if the browser is in `Notes` mode


dae:
Thanks Hikaru!

PR #1728, opened 2022-03-20
Backup improvements

dae:
- Backups without closing the collection/losing v3 undo
- Option to backup immediately in the file menu
- Periodic automatic backups
- Configurable backup duration in preferences

@RumovZ would you mind giving the code a review when you get a chance? If you feel anything should be done differently, please don't hold back. :-) Feedback from anyone else is of course welcome too.

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -615,6 +615,7 @@ class AnkiQt(QMainWindow):
                 # default 5 minute throttle
                 self.col.create_backup(
                     backup_folder=self.pm.backupFolder(),
+                    force=False,
                     wait_for_completion=False,
                 )
             self.col.close(downgrade=False)
@@ -644,8 +645,8 @@ class AnkiQt(QMainWindow):
         Waits until the backup completes. Caller should run this on a background thread."""
         assert self.col.create_backup(
             backup_folder=self.pm.backupFolder(),
+            force=True,
             wait_for_completion=True,
-            minimum_backup_interval=0,
         )
 
     def on_create_backup_now(self) -> None:
@@ -1380,9 +1381,8 @@ title="{}" {}>{}</button>""".format(
         def backup(col: Collection) -> bool:
             return col.create_backup(
                 backup_folder=self.pm.backupFolder(),
+                force=False,
                 wait_for_completion=False,
-                # 30 minutes
-                minimum_backup_interval=30 * 60,
             )
 
         def after_backup_started(created: bool) -> None:
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -640,28 +640,6 @@ class AnkiQt(QMainWindow):
             Config.Bool.INTERRUPT_AUDIO_WHEN_ANSWERING
         )
 
-    def create_backup_now(self) -> None:
-        """Create a backup immediately, regardless of when the last one was created.
-        Waits until the backup completes. Caller should run this on a background thread."""
-        assert self.col.create_backup(
-            backup_folder=self.pm.backupFolder(),
-            force=True,
-            wait_for_completion=True,
-        )
-
-    def on_create_backup_now(self) -> None:
-        def on_backup_created(val: None) -> None:
-            # legacy checkpoint may have been cleared
-            self.update_undo_actions()
-
-            tooltip(tr.profiles_backup_created(), parent=self)
-
-        QueryOp(
-            parent=self,
-            op=lambda _: self.create_backup_now(),
-            success=on_backup_created,
-        ).with_progress().run_in_background()
-
     # Auto-optimize
     ##########################################################################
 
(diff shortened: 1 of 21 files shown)


dae:
Oh, we'll probably want to figure out a shortcut key for the create backup action.

RumovZ:
Looks great!

> \- If force is true but the collection has not been modified, the backup
will be skipped. If the user manually deleted their backups without
closing Anki, they wouldn't get a new one if the mtime hadn't changed.

I think that's fine because they will always have the option to use _export_, but it should be communicated (like a tooltip "Current state already backed up"). On the other hand, not much harm in creating an identical backup, and this behaviour would probably be the easiest to understand.

> \- Force takes preference over the configured backup interval - should
we be ignored the user here, or take no backups at all?

That's the behaviour I'd expect. Maybe just change the setting's description to "Minutes between _automatic_ backups"?

<hr>

With the increase of created backups, it might get more difficult for the user to find the right one, i.e. the latest before the breaking operation. Maybe more meta info could be stored to help with that? E.g. an enum specifying the reason for a given backup:

```rs
enum BackupReason {
    Manual,
    Automatic,
    FullDownload,
    Import,
```

I know there's currently no opportunity to display that information, and am not sure this idea makes sense, just throwing it out there. 🙂

rslib/src/collection/backup.rs
```diff
@@ -14,34 +14,51 @@ use itertools::Itertools;
...
 };
 
 const BACKUP_FORMAT_STRING: &str = "backup-%Y-%m-%d-%H.%M.%S.colpkg";
-/// Default seconds after a backup, in which further backups will be skipped.
-const MINIMUM_BACKUP_INTERVAL: u64 = 5 * 60;
 
-pub fn backup(
-    col_path: impl AsRef<Path>,
+pub fn maybe_backup(
+    col: &mut Collection,
```
RumovZ, line 24: Could be placed in an impl block. (Or is this deliberate?)

qt/aqt/main.py
```diff
@@ -643,6 +640,28 @@ def apply_collection_options(self) -> None:
             Config.Bool.INTERRUPT_AUDIO_WHEN_ANSWERING
         )
 
+    def create_backup_now(self) -> None:
+        """Create a backup immediately, regardless of when the last one was created.
+        Waits until the backup completes. Caller should run this on a background thread."""
+        assert self.col.create_backup(
```
RumovZ, line 646: You're probably aware as it ties to the remaining decicions you mentioned, but this currently causes Anki to hang.

rslib/src/collection/backup.rs
```diff
@@ -14,34 +14,51 @@ use itertools::Itertools;
...
-fn has_recent_backup(backup_folder: &Path, recent_secs: u64) -> Result<bool> {
+fn should_skip_backup(
+    force: bool,
+    minimum_interval_mins: u32,
+    backup_folder: &Path,
+) -> Result<bool> {
+    if force {
+        Ok(false)
+    } else if minimum_interval_mins == 0 {
+        Ok(true)
```
RumovZ, line 54: I believe it should be `Ok(false)` to match the previous behaviour.

RumovZ:
> Did a sneaky edit of the existing ftl string, as it hasn't been live
long.

Doesn't Pontoon have a feature to mark strings as fuzzy, so translators get notified, but the old translations remain live until they're updated? Seems like that could come in handy often.

rslib/src/collection/backup.rs
dae: This is coming from the user config now, and my thinking here was the user could set it to 0 if they didn't want any automatic periodic/on close backups.

qt/aqt/main.py
dae: Do you mean that you're seeing the UI become unresponsive when doing a DB check or a full download? They are currently using taskman directly, so for the first part of the operation where the collection lock is held, if the browse screen tries to refresh, it's going to get stuck querying the DB as these two screens aren't using QueryOp().

Or do you mean something else? For the user-initiated create backup case, I've changed the code to do the second half in the background, then monitor for errors. If the user creates a backup and then immediately makes a breaking change before the backup completes, they could potentially be unable to restore the backup if the second half of the process fails, but the window for that to happen would be pretty short.

dae:
Unfortunately as far as I'm aware, Pontoon only supports fuzzy strings with its legacy gettext support. Maybe they've made improvements in the last 18 months though - updating the installation is one of the many things on the todo list. 😅

dae:
> With the increase of created backups, it might get more difficult for the user to find the right one

I think we've actually made it a bit easier than it was in the past - the user will get a backup about 5 minutes after opening their profile, and then every 30 minutes after that. If the modification times show more than one backup created within a 30 minute window, the second one will have been because of a colpkg import, full sync or DB check.

RumovZ:
That's true, although it presumes some insight by the user.
I thought in the future it might be nice to present a list of backups with parsed file names and meta, instead of just a file picker. But the backup reason would be most interesting for short-term backups (latest 48h) anyway, so even if this will happen, there's no rush to add that meta information before it is exposed.

qt/aqt/main.py
RumovZ: I meant when triggering a manual backup when there have been no changes since the last one, but that's solved with a42c511ceb59c912ea6b476b5487c07637732e96.

rslib/src/collection/backup.rs
RumovZ: This should be explained then, shouldn't it? Otherwise I'd assume, lower *minutes between backups* -> more backups.

dae: I'd been thinking we could do that by tweaking the description or adding a separate checkbox that sets it to 0 behind the scenes. I just tried adding the checkbox to the preferences screen, but having a single switch that could accidentally get disabled feels risky, as backups are so important. Could the following work instead?

- remove the special case for 0, but set the minimum in the spinner to 5 so the user doesn't waste disk/get confused by the 5 minute duration of the time
- increase the maximum for the minimum interval to 99999, so users can't fully disable automatic backups, but can have them created only once every 2 months

RumovZ: Sounds good!

dae: On second thought, even 9999 is probably reasonable - once a week is not going to consume large amounts of space. Might start with that and see how many people want more (and they could always use the debug console if they know what they're doing)

RumovZ: Also, there are the other options which effectively allow to have only one backup stored at any point in time.

dae:
I can see how that context could be useful when combined with a separate screen to list the backups, but that's a fair bit of extra work :-) I'd be tempted to wait for a while to see how the new backup handling is received, and then consider that again later.

dae:
Appreciate all the feedback! Will merge this in now and hopefully get another beta out in an hour or two.

PR #1738, opened 2022-03-23
Add tooltip that you can double-click image for restoring its size

hgiesel:
Users might miss it, as indicated [in this post](https://forums.ankiweb.net/t/anki-2-1-50-beta-6-9-stable-release/18181/53)


ftl/core/editing.ftl
```diff
@@ -13,6 +13,7 @@ editing-couldnt-record-audio-have-you-installed = Couldn't record audio. Have yo
 editing-customize-card-templates = Customize Card Templates
 editing-customize-fields = Customize Fields
 editing-cut = Cut
+editing-double-click-image = Double-click image
```
dae, line 16: How about something like the following?

```suggestion
editing-double-click-image = double-click to toggle
```

or

```suggestion
editing-double-click-image = double-click image
```

PR #1742, opened 2022-03-26
Add Deleted error and disable all bad browser rows

### LINKED ISSUE #1729: Cards with inexisting ord appears as deleted
How to reproduce:
* manually change the ord of a card of a normal note type in the db. Make it big enough that it does not correspond to a card type ord
* Open the browser and search for the note with this text

Expected result:
* really not clear honestly. Probably an invitation to check the database

Actual result:
* the card appear as deleted. While it still exists actually. 

This is a trouble mainly because it mean that I can't access the note content (fields, tags) so I can't even check which note it is. Also it means I can't just  use the advanced browser feature that allow to edit card properties from the browser (not that it's a feature that I regularly use. But it would allow for manual correction)

RumovZ:
Closes #1729.
As far as I'm aware, `is_deleted` wasn't used by any add-on, and this can be merged into 2.1.50 safely.

### DIFF
--- a/proto/anki/backend.proto
+++ b/proto/anki/backend.proto
@@ -58,6 +58,7 @@ message BackendError {
     SEARCH_ERROR = 14;
     CUSTOM_STUDY_ERROR = 15;
     IMPORT_ERROR = 16;
+    DELETED = 17;
   }
 
   // localized error description suitable for displaying to the user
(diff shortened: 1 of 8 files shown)


dae:
Do those changes work for you?

RumovZ:
I guess it's not strictly correct—semantically and because rows don't get checked while the backend is blocked. But that would be an edge case of an edge case, so never mind. 🙂

dae:
Nice, thanks for the fix :-)

Issue #1017, opened 2021-02-12
flags need to be reset on export, possibly ignored on import


dae:
https://forums.ankiweb.net/t/flags-are-kept-when-exporting-decks/7450/2

dae:
From #1743, cleared on export now when scheduling excluded.

PR #1743, opened 2022-03-27
Add apkg import/export on backend

RumovZ:
This is pretty much untested. My plan is to do the apkg import part next, then add extensive roundtrip tests and fix it up.
What do you think so far? Should I do anything differently?


dae:
The approach I had in mind was a little different, but that's my fault for not communicating it properly - sorry :-(

Rather that copying rows directly and then mutating the resulting database like the current Python code does, it might make things easier to reason about and maintain if we instead deserialize items into our standard structs like Note, Noteype and so on, and then serialize them again to the output DB - eg the equivalent of for_each_note_in_search(|note| output.storage.add_or_update_note(note)). It won't be as efficient as copying rows directly, though we would win back some of the performance by not having to apply updates after rows have been copied (to modifying tags/fields, reset scheduling/flags, etc). 

The gathering and transforms are also something we potentially might want to reuse. For example, at one point in the future it might be nice to have a separate import/export format that is more easily read by third-party apps. Logic like resetting cards, excluding revlog entries, renaming out-of-tree decks or blanking certain fields is something that could be applied to such other formats as well.

A simple approach would be to gather and transform all the data up-front, then have the exporter just responsible for writing out the gathered data. A more memory-friendly approach would be to have the exporter able to deal with the items in a streaming fashion, but that does add more complexity, and decks over 500MB are exceedingly rare, so a decent case could be made for the former approach too.

For either approach, if the apkg exporter was then implemented as an Exporter trait that accepts the gathered+modified data and writes it out, that would make adding future formats easier. We may even want to use it for csv exports as well - as while the scheduling info is currently irrelevant, removing the marked tag and redacting some fields are things we may still want to do in the future.

WDYT?

dae:
(Getting a minimum implementation of apkg importing done will ease testing, and I think it's a good idea to consider both import and export sides of the equation at the same time, so please feel free to focus on that first before considering the suggestions above)

rslib/src/import_export/package/colpkg/export.rs
```diff
@@ -67,6 +68,24 @@ impl Collection {
...
+impl MediaIter {
+    pub(crate) fn from_folder(path: &Path) -> Result<Self> {
+        Ok(Self(Box::new(
+            read_dir_files(path)?.map(|res| res.map(|entry| entry.path())),
+        )))
+    }
+
+    pub(crate) fn from_file_list(list: impl IntoIterator<Item = PathBuf> + 'static) -> Self {
+        Self(Box::new(list.into_iter().map(Ok)))
+    }
```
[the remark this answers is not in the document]
RumovZ: My intention was to reduce the spread of broken decks with missing media, but I see how it doesn't work. I'll just add a `filter(|path| path.exists())`.

rslib/src/import_export/package/apkg/export.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+impl CardTemplate {
+    fn extract_media_paths(&self, names: &mut HashSet<PathBuf>, media_folder: &Path) {
+        for template_side in [&self.config.q_format, &self.config.a_format] {
+            for name in extract_underscored_references(template_side) {
+                if is_local_base_name(name) {
+                    let path = media_folder.join(name);
+                    // shotgun approach, so check if paths actually exist
+                    if path.exists() {
```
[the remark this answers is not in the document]
RumovZ: Not sure if it's worth it. If we only allow reasonable extensions, we'd need to differentiate between JS strings and HTML refs more strictly, and repeat the different quotation cases for both of them. And for every case, repeat the list of extensions.
If we just check for paths to exist, and ignore them if they don't, do we lose that much through false positives?

Anecdotally, I have a notetype with several kB of JS, generated by an add-on, and there's not a single match.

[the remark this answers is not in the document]
dae: I was more thinking of what would happen when we no longer require a leading underscore - we'd likely get a lot more matches then. Looking the files up when it's not a local folder is somewhat more expensive, so it would be good to avoid thousands of lookups that are likely to fail. But we can deal with that in the future when removing the underscores; for now it's probably fine.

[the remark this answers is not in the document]
dae: Actually, ignore me - we could relax the restriction but continue to require it for the bare reference case. We're in agreement then. :-)

dae:
With the Python code, there is something a bit like that. apkg is treated specially, but the other importers like CSV, Mnemosyne and SuperMemo all are based on NoteImporter. They gather the data into a common format, which NoteImporter then applies to the DB. Ideally our new code would unify the apkg import case with the others, if it is practical to do so.

For the export trait, I'd just been thinking of a serialization/visitor-like interface, eg:

fn export_notes(&self, notes: Vec<Note>)
fn export_cards(&self, cards: Vec<Card>)

Then each format exporter could decide how it would turn the gathered and mutated cards/notes/etc into the desired format (whether it places all the provides objects in an .anki2 file and zips it up, writes them out to a csv file, etc). Each exporter could ignore parts not relevant (eg no notetype info in a csv), and would probably needs to retain state throughout the export (eg a json exporter might need to build up a single giant dict with the info).

Skipping the trait while prototyping sounds wise; we can easily add it later.

rslib/src/storage/revlog/mod.rs
```diff
@@ -120,6 +120,13 @@ impl SqliteStorage {
             .collect()
     }
 
+    pub(crate) fn get_revlog_entries_for_card_ids(&self, cids: String) -> Result<Vec<RevlogEntry>> {
+        self.db
+            .prepare_cached(&format!("{} where cid in {cids}", include_str!("get.sql"),))?
+            .query_and_then([], row_to_revlog_entry)?
+            .collect()
+    }
+
```
dae, line 129: See comment on gather_cards()

rslib/src/import_export/gather.rs
```diff
@@ -0,0 +1,284 @@
...
+                .map(|did| self.storage.get_deck(*did)?.ok_or(AnkiError::NotFound))
+                .collect()
+        } else {
+            Ok(self
+                .storage
+                .get_all_decks()?
+                .into_iter()
+                .filter(|deck| deck.id != DeckId(1))
+                .collect())
+        }
```
dae, line 210: To support the selected notes use-case, we'll need to gather decks based on the identified cards, optionally discarding ones that are are outside the selected deck.

dae:
> Is this more like what you had in mind?

Yep, looking good :-)

> For importing, I imagine the apkg would just use the current gather code. Other formats would need to do more work, but share some routines among them, of course. Then there'd be a preparation step for handling ids, usns etc, which should be pretty much the same for all formats. After that, the insert module can handle the rest.

Sounds good as a high-level plan. The preparation step will be the tricky bit - the existing apkg import logic is somewhat complicated and badly designed. For example:

- Notes are matched with existing ones based on a string guid instead of id. This was done to reduce the chances of a conflict that a millisecond timestamp would have, but the 64 bit guid should have been larger. It was stored as a string because interoperability with Javascript was a concern at the time.
- When an existing note is found, importing will refuse to update it if the notetype's fields or template names differ, as we don't want to accidentally write a note out with 2 fields if the notetype says it should have 3.
- When a new note is to be added, we try to copy the notetype over from the source collection, preserving its id.  But if the notetype of source collection and current collection have different schemas, we need to copy the source notetype into the collection with a new id, and rewrite the ntid of the note. Instead of assigning a random new number, the code instead tries to increment the ntid by one - the logic being that if an updated deck is imported again later, we could walk the ntids starting at the original one until we locate the copied notetype, allowing it to be reused/updated, instead of a new notetype being created each time. But the current way this is done feels a bit complex, and I wonder whether there might be a better way to do it.
- When importing shared decks, two different decks may have used "banana.jpg" with different file content in each. If that file already exists in the collection, we need to adjust any notes that reference it to point to the new filename instead.

> To be honest, I still don't understand the benefit of an export trait, but maybe it'll dawn on me as I proceed.

We may not need one. I just thought if we ended up with each exporter looking like

fn some_exporter() {
  ... lots of shared setup code
  specific_exporter_logic()
  ... lots of shared teardown code
}

then in that case, being able to pass in the custom logic with a fn or trait might be convenient. Let's YAGNI until we see a pattern. :-)

rslib/src/import_export/gather.rs
```diff
@@ -0,0 +1,284 @@
...
+       }
+    */
+    pub(super) fn gather_data(
+        &mut self,
+        col: &mut Collection,
+        deck_id: Option<DeckId>,
+        with_scheduling: bool,
+    ) -> Result<()> {
+        self.decks = col.gather_decks(deck_id)?;
+        let search = optional_deck_search(deck_id);
```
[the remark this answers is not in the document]
RumovZ: Makes sense. Maybe `export_apkg()` can just accept a search? The backend code could expose a oneof of reasonable options (a deck id, all, note ids...), and pass on the according search node.

rslib/src/tags/matcher.rs
```diff
@@ -17,7 +16,7 @@ pub(crate) struct TagMatcher {
 ///
 /// Tracks seen tags during replacement, so the tag list can be updated as well.
 impl TagMatcher {
-    pub fn new(space_separated_tags: &str) -> Result<Self> {
+    pub fn new(space_separated_tags: &str) -> Self {
```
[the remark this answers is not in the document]
RumovZ: Sorry, my bad. Since I ended up not using TagMatcher after all (at least at the moment), I'll just revert all changes to that module.

rslib/src/import_export/package/colpkg/export.rs
```diff
@@ -67,6 +68,27 @@ impl Collection {
...
+    pub(crate) fn from_folder(path: &Path) -> Result<Self> {
+        Ok(Self(Box::new(
+            read_dir_files(path)?.map(|res| res.map(|entry| entry.path())),
+        )))
+    }
+
+    /// Skips missing paths.
+    pub(crate) fn from_file_list(list: impl IntoIterator<Item = PathBuf> + 'static) -> Self {
+        Self(Box::new(
+            list.into_iter().filter(|path| path.exists()).map(Ok),
```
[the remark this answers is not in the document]
RumovZ: If I'm not mistaken, that's what's happening, because the iterator isn't consumed until `write_media_files()`.

rslib/src/import_export/gather.rs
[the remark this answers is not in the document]
RumovZ: Just remembered why I chose a deck-based approach: The user might want to export a top-level deck where all cards are in subdecks. Then we would also want to export the "empty" parent.
If we accepted a SearchNode, we could check if it's a single deck search, and export empty parent decks accordingly. Not the cleanest solution, but accepting both note ids and a deck id isn't pretty, either.

rslib/src/import_export/package/colpkg/export.rs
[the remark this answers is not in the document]
dae: Whoops, missed that 👍

rslib/src/import_export/gather.rs
```diff
@@ -0,0 +1,284 @@
...
+            card.interval = 0;
+            card.ease_factor = 0;
+            card.reps = 0;
+            card.lapses = 0;
+            card.original_deck_id = DeckId(0);
+            card.original_due = 0;
+            card.original_position = None;
+            card.queue = CardQueue::New;
+            card.ctype = CardType::New;
+            card.flags = 0;
```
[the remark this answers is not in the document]
RumovZ: Just noticed an unhandled edge case: If a card from a filtered deck is rescheduled as new, shouldn't it's original position (odue) also be reused if it's of type new?

rslib/src/import_export/gather.rs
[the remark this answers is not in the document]
dae: Accepting a SearchNode sounds good.

Good point about needing the parents, but that should apply in the specific notes case as well - we must implicitly include all parent decks of the decks that have been used, as the deck tree code in the importing Anki will expect the parents to exist.

rslib/src/import_export/gather.rs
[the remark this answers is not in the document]
RumovZ: Hm, even if a card is new and in a regular deck, it's assigned a new position. Should I make a PR that can be merged into 2.1.50?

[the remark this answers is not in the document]
dae: For our export case, it sounds like we should be doing that. In your follow-up comment, are you saying that reschedule_cards_as_new() assigns new positions if there is no original position or restore_position is false? I thought that was intended behaviour, so that users could send new cards to the back of the queue. Have I missed something?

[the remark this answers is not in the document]
RumovZ: I'm only talking about the `restore_position == true` case here. In that case, I wouldn't expect the current position of new cards to change. Rather something like the following should be used:

```rs
    fn last_position(&self: Card) -> Option<u32> {
        if self.ctype == CardType::New {
            Some(if self.original_deck_id.0 != 0 {
                self.original_due
            } else {
                self.due
            } as u32)
        } else {
            self.original_position
        }
    }
```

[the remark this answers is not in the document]
dae: Ok, I think I get you - you're saying you think the behaviour should be 'restore original position if available (otherwise, leave current position alone if it's already new)', instead of '(otherwise, place card at end of queue)'? That does seem to make more sense; a PR would be great.

rslib/src/import_export/gather.rs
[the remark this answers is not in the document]
RumovZ: 97c9dd4 removes the feature to specify a deck, and move siblings outside that deck into it. Can we get away with that?
First of all, it won't affect the majority of users who don't have siblings in different decks. And for those who do, including cards outside the selected deck may not be the expected behaviour anyway.
We could show a warning if an export includes decks and cards outside the selected decks.

rslib/src/import_export/gather.rs
```diff
@@ -0,0 +1,211 @@
...
+                .collect();
+        }
+    }
+
+    fn reset_deck_config_ids(&mut self) {
+        for deck in self.decks.iter_mut() {
+            if let Ok(normal_mut) = deck.normal_mut() {
+                normal_mut.config_id = 1;
+            } else {
+                // TODO: scheduling case
```
RumovZ, line 90: I couldn't figure out how filtered decks are currently converted to regular ones if scheduling is included. 🤔 
Do I have to run the below code for all filtered decks in the scheduling case as well?

rslib/src/import_export/gather.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    fn remove_scheduling_information(&mut self, col: &Collection) {
+        self.remove_system_tags();
+        self.reset_deck_config_ids();
+        self.reset_cards(col);
+    }
+
+    fn remove_system_tags(&mut self) {
+        // TODO: case folding? child tags?
```
RumovZ, line 76: Anki assigns these tags with consistent lower case. Should I still match unicase? And what if a user used them as parent tags?

dae: They're ascii and not localized, so to_ascii_lowercase() should suffice. I don't think I've ever seen subtags used here, so wouldn't worry about it.

rslib/src/import_export/gather.rs
dae: Unintuitively, the current code handles this at import time (anki2.py:287, :356). It means renaming an apkg file to colpkg and importing it results in errors. Ideally we'd do it on the export side instead, but since we'll need to keep doing it on the import side to support legacy decks, we could get away with just doing that for now.

In the scheduling=reset case, we reset the config id to 1 so that the user's default deck preset/config is used. In the scheduling=keep case for normal cards, we preserve+copy the deck presets over, so the user can use this mode for moving/copying their cards/decks/presets from one profile to another. In the filtered deck case I guess the current code is discarding the deck config id? In the scheduling=keep case, it might be preferable if we preserved the filtered deck and made sure the home deck was included at the same time, as that would allow users to shuffle filtered decks between profiles as they can do with normal decks. However, the first paragraph may come back to bite us there - if we went down that path, we'd need to make sure we're not breaking legacy exports. Do you have any thoughts/ideas/preferences on what we do here? We could potentially defer this part until the code to deal with the importing side is implemented.

rslib/src/import_export/gather.rs
```diff
@@ -0,0 +1,211 @@
...
+}
+
+impl Collection {
+    fn gather_notes(&mut self, search: impl TryIntoSearch) -> Result<Vec<Note>> {
+        self.search_notes_into_table(search)?;
+        self.storage.all_searched_notes()
+    }
+
+    fn gather_cards(&mut self) -> Result<Vec<Card>> {
+        self.storage.search_notes_cards_into_table()?;
```
dae, line 158:
```suggestion
        self.storage.search_cards_of_notes_into_table()?;
```

```suggestion
        self.storage.search_cards_into_table_from_notes_table()?;
```

```suggestion
        self.storage.search_notes_table_into_cards_table()?;
```

rslib/src/import_export/gather.rs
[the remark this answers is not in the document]
dae: I was mainly thinking of the shared decks case, where it's the users who import the deck who tend to encounter the problems/confusion, rather than the person exporting them (similar to our template checks). Showing an error on export if the user has elected to export a particular deck and cards are outside it may even be a better option than moving cards, as it avoids surprising behaviour, and is simpler than attempting to move the cards to a different deck. And for local exports, users could do them via the browse screen if they wanted to avoid the error.

rslib/src/import_export/gather.rs
dae: Consider the case where a user has created two separate profiles, and then wants to follow the advice on https://faqs.ankiweb.net/synchronizing-multiple-profiles.html. They need some way of being able to transfer content from one profile to the other in order to be able to merge the profiles, and currently as filtered decks can't be transported, it means the user will need to recreate them and their settings in the importing profile. Not a big deal for the majority of users who either don't use filtered decks, or use them ephemerally. But some users have a bunch of filtered decks with complex queries, and ideally in that case we'd provide a way for them to transport them. But to be fair, it is a niche use case, and probably doesn't come up much.

rslib/src/import_export/gather.rs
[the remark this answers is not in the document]
RumovZ: I assume then there'd be a separate routine to check for these things in advance? I think something like this would be required anyway if we want the export process to be more customisable. E.g. we could detect notetype conflicts in advance, and let the user decide on a resolution strategy (mapping or skipping).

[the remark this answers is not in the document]
dae: Wouldn't notetype conflict checking need to be done at import time? Maybe you have something else in mind than I'm imagining?

I'm not too fussed about the timing of the check, though it would be nice to at least do prior to all media being exported.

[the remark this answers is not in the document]
RumovZ: Sorry, I got confused with import/export. For both cases, some routine for preparing/checking that is called before the import/export call by the frontend seems to make sense. Then we can be more naive in the core routines. In this case, we can keep just exporting all decks of matching cards.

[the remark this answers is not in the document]
dae: Yep, sounds good 👍

RumovZ:
As discussed above, I was trying to implement a three-step approach (gather, prepare, insert), compared to the current code, which does everything at once, to reduce complexity, and increase code reusability. However, it feels like an uphill battle. All ids and names need to be handled manually, while we'd (at least in part) get that for free with the existing add methods.
Also, the approach seems impossible to implement strictly. Two examples: Canonifying tags needs to both register new tags, and look up the case of existing ones, so the note can be adjusted accordingly. So we need to either make changes to the db during the  preparation step, or modify the imported data during the insertion step.
Similarly with media: We need to read incoming media while preparing notes, so we can check for equality with existing media, and potentially use an alternative file name (which then needs to be reflected on the notes). So it makes sense to immediately copy the file while it's in memory. On the other hand, I think media should be copied as late as possible, because we can't roll it back if an error occurs. (Or should we copy to temp files, and persist these in bulk at the end of a successful import?)

I wonder if this is still the way to go. Should I stick more closely to the current Python code, or do you have another idea for dealing with this? If you think this might still be our best bet, I'll continue. I just want to avoid writing all this new code if it's doomed from the beginning.

Another consideration is undoing. I assume that's a goal at some point? Not sure if this will be easier to achieve with one approach or another.

I've pushed my work so far, hopefully it helps a little bit to understand what I'm rambling about. It's not even remotely finished, though, so there's no point in reviewing anything in detail.

dae:
I probably won't get to reviewing the code until tomorrow, but just some brief comments based on what you've said above.

My main goals with separate steps for gathering+writing were:

- allowing the prepare/write steps to be partly/mostly shared by other importers, so that we are decoupling the incoming format (anki2, json, csv, etc) from the common handling of things like renaming media references, dealing with conflicts, etc), and actual insertion.
- allowing us to handle the filtering/transforming based on our Rust data structures like Note, Notetype and so on, instead of having to use database-mutating functions to modify either the source anki2 or the target anki2 file after we've inserted content in it.

The separate gather step achieves those goals, and whether we do preparation+writing in two discrete steps or partially mix them together doesn't seem to matter (at least at from what I can tell at this point :-). So if it makes things easier to write and reason about, I have no objections to combining the two. It think it may still make sense to do some preparation up front, but we don't need to have a separate distinct preparation step; those can just be some calls at the start of the prepare+write routine.

For media, this is how I imagined it might work:

- at the start of the prepare/write stage, we call ChangeTracker::register_changes (like in media/sync.rs:register_changes()), to ensure we have up to date checksums of all the existing files in the media folder. The progress can be mapped to part of our  import progress. We should then grab all existing entries into existing_files: HashMap<Filename, Sha1>
- we then inspect the Vec<MediaEntry> from the imported file, and see if sha1s were included. In the happy path (new-style format), the checksums are already there for us to use. For legacy imports/other future formats that don't include a checksum, we can generate the checksums at that point, by reading data from the zip. We don't need to do it for every file - only files where the filename is already in existing_files.
- We should normalize the incoming filenames as we do this, as we can't trust them to be normalized.
- If we create a modified version of add_data_to_folder_uniquely() that works with our MediaEntry list and existing_files, we should be able to predict the resulting filename in the case of conflicts, without actually adding the media. We could store our media entries with an Option<RenamedFilename> in a HashMap<OriginalFilename, MediaEntryWithPossiblyRenamedFilename>
- At the note processing stage, for each media entry we encounter in a field, we can look up the filename and replace it if it has been renamed.

How does that sound?

Re undo, I'd definitely like to support undo if at all possible. Reusing our existing routines to add things like notes and notetypes is probably going to make that easier.

RumovZ:
Thanks, that helped a lot! I'll make the changes. Reviewing 6836da0 is probably not a good use of your time then.

> I think it may still make sense to do some preparation up front, but we don't need to have a separate distinct preparation step; those can just be some calls at the start of the prepare+write routine.

I wonder if it wouldn't make sense to do initial preparation in collection format, though. E.g. apkgs are exported without parent decks. This could easily be resolved with a call to `check_database()`.
I know, this somewhat contradicts the current idea of a common import format. It would probably mean collecting from all formats into a temporary collection (apkgs already have one, of course), then preparing the data, and finally copying it into the target collection, while finishing off preparation.
As a side effect, this should make switching to a streaming, more memory-friendly approach much easier.
I don't have much of an idea what is required for csv-based formats, but maybe they too would be easier to prepare inside a collection? The Vec-based format isn't ergonomic to work with if you need random access to, let's say, notetypes in order to prepare notes.

It seems to me the two goals you stated above would still be achieved if the common format is a plain collection, but maybe I've missed something. Anyway, I felt like I should bring this up as early as possible.

dae:
I've had a brief look through the current code, and it seems to be coming along nicely, and feels quite a bit easier to follow than the old Python implementation :-) The Cow helper looks like a nice improvement too.

>  It would probably mean collecting from all formats into a temporary collection (apkgs already have one, of course), then preparing the data, and finally copying it into the target collection, while finishing off preparation.
> As a side effect, this should make switching to a streaming, more memory-friendly approach much easier.

It's a decent idea, and would certainly make it easier to keep the working set size manageable. The main downside is the data would need to be written out to disk twice, which will impact performance to some extent. Worse so if we start doing a bunch of database-mutating operations on the temporary collection instead of transforming what we can "on the fly". A database check on the temporary collection for example would be a bit too heavy-weight :-) But if we can minimize such things, I can see it working as an alternative to a memory-based approach. 

Perhaps it might be worth thinking through/prototyping a non-apkg implement before we make a decision there?

A bit of background on the current (messy) implementation:

- The various importers subclass NoteImporter
- NoteImporter has a `mapping` property that specifies how fields in the file should be mapped to the fields in the current notetype. The GUI instantiates the importer and expects it to pre-fill the mapping based on the file and active notetype, lets the user adjust it, and then uses the adjusted mapping when importing proceeds. The csv importer relies on this functionality, and thus relies on a notetype that's already in the target collection. Same with decks - the active deck is used.
- One way we could support such a workflow in the new prepare/import code be to inject the current notetype/deck into the temporary in-memory or on-disk gathered data, and mutate the notes to the the user's desired mapping.
- Importers like the Mnemosyne and SuperMemo one create their own notetypes, insert them into the target collection, and then use the same mapping approach as above.
- Properties like Note.id and Note.guid are currently synthesized; no attempt is made to import them from the original file
- Uniqueness matching is done based on the first field, like when adding via the Add screen. This prevents the user from modifying the first field and re-importing to update, and has lead to awkward workarounds like https://ankiweb.net/shared/info/1672832404. There have also been instances where users imported an apkg file and expected the first-field behaviour instead; perhaps we should think about making it configurable if practical to do so.

RumovZ:
> The main downside is the data would need to be written out to disk twice, which will impact performance to some extent.

Alternatively, could we possibly use an in-memory database? That completely negates the memory advantage, but would still allow us to reuse existing collection routines.

> Perhaps it might be worth thinking through/prototyping a non-apkg implement before we make a decision there?

Sounds like the wise thing to do. I'll continue with the existing gather format, and skip the initial preparation step for now, which seems to depend the most on that format.

dae:
> Alternatively, could we possibly use an in-memory database?

Yep, that's certainly possible, and we could potentially even select memory/file based on the amount of incoming data, provided we know the approximate amount up front.

RumovZ:
Slowly getting to a point where I can test some stuff. What do you think of adding a test collection file with some more or less representative data points? I find it bothersome to populate a new collection from scratch every time something needs testing, and  sometimes difficult to figure out how to reach a certain collection state programatically.
If we had a source-controled `collection.anki2`, we would only need to add cards, decks, etc. in various states once (or incrementally as required) and could reuse that for different tests.
Such a file could be inspected with the Anki GUI, whereas a programatically constructed collection, although being inspectable directly from the source code, is quite hard to reason about.

dae:
There is precedent for it - you'll find a few test apkg and anki2 files in pylib/tests/support. 

There are a few downsides with embedding collections in git though:

- Any minor change to the test file will require a new copy to be stored in the repo, which increases the size of the repo over time.
- Any changes are opaque, and can't be easily reviewed
- The unit tests are one of the only places we have a chance to "dogfood" our Rust API, and when things are not ergonomic, it can sometimes give us a hint that the API needs improving.

For those reasons, it would be good if we could try to minimize the use of binaries, and use alternative approaches where it is practical to do so. But I realise sometimes it may not be practical.

A middle ground may also be worth considering in some cases - we could use '.dump' in the sqlite3 program to get the SQL commands necessary to create a test collection, and then feed them into a helper which creates the collection via execute_batch(). The dumped sql is more efficient and easy to review than a binary blob, though still harder to read/adjust compared to programmatic creation.

RumovZ:
I was afraid it would be something like this, but I understand.
For now, I'll try again with a programatic approach, but put it in `src/tests.rs` or something, so hopefully it can be reused and improved in the future. I guess it should also be possible to write a helper that creates a profile folder from that collection, which can then be opened with Anki.

rslib/src/import_export/package/apkg/import.rs
```diff
@@ -248,12 +255,23 @@ impl<'a> Context<'a> {
...
+
+        self.target_col.add_note_only_with_id_undoable(note)?;
+        self.existing_notes.insert(note.id);
         Ok(())
     }
 
+    fn uniquify_note_id(&mut self, note: &mut Note) {
+        let original = note.id;
+        while self.existing_notes.contains(&note.id) {
+            note.id.0 += 999;
```
dae, line 268: At a later stage when this code is functional, it might be interesting to benchmark this approach compared to a simple call to storage.add_note(), which assigns the next-greatest available id in the case of a conflict. The latter is a bit easier to reason about, but if it introduces a significant performance regression, then we're better off sticking with what we've got.

rslib/src/import_export/package/apkg/import.rs
```diff
@@ -393,21 +398,32 @@ impl<'a> Context<'a> {
...
+
+    /// The number of days the source collection is ahead of the target collection.
+    fn collection_delta(&mut self) -> Result<i32> {
+        Ok(self.data.days_elapsed as i32 - self.target_col.timing_today()?.days_elapsed as i32)
+    }
+
+    fn add_card(&mut self, card: &mut Card) -> Result<()> {
+        card.usn = self.usn;
+        if self.target_col.add_card_if_unique_undoable(card)? {
+            self.added_cards.insert(card.id);
```
dae, line 426: The original code also does 'insert or ignore', but if I've read it correctly, it is first giving the card a unique id, so theoretically the conflict case should never happen for cards, and the if_unique here makes things a bit more confusing.

rslib/src/import_export/package/apkg/import.rs
```diff
@@ -426,8 +432,7 @@ impl<'a> Context<'a> {
 
     fn import_revlog(&mut self) -> Result<()> {
         for mut entry in mem::take(&mut self.data.revlog) {
-            if let Some(cid) = self.remapped_cards.get(&entry.cid) {
-                entry.cid = *cid;
+            if self.added_cards.contains(&entry.cid) {
```
dae, line 445: Just to double-check we're on the same page here, the way the Python code works is:

- a given card is imported if its (note.guid, ordinal) isn't in the target collection
- it's assigned a new ID if its existing one conflicts with an ID in the collection
- because it may have a new ID, the revlog entries that reference it need to be updated to point to the new card id

It looks like the cid update has been removed here; are you planning to handle it a different way?

dae:
> For now, I'll try again with a programatic approach, but put it in src/tests.rs or something, so hopefully it can be reused and improved in the future.

That'd be great - the easier tests are to write, the more likely we are to add an appropriate number of them :-)

rslib/src/import_export/package/apkg/import.rs
RumovZ: Isn't the problem with assigning the next available id that it messes up note creation time?

rslib/src/import_export/package/apkg/import.rs
RumovZ: Yes, I've messed this up. So many objects, all with different equality checks and merge strategies. 🙂

rslib/src/import_export/package/apkg/import.rs
RumovZ: See above. Thanks for reiterating. I came to the conclusion that we don't need the guid here. Either the card's note could not be imported, then its original id is in `conflicting_notes`. Or it was imported successfully, then its possibly remapped id is unique in the target collection.

rslib/src/import_export/package/apkg/import.rs
dae: Yes, you're quite right, scratch that idea.

rslib/src/import_export/package/apkg/import.rs
dae: Yep, it's a mess - can you tell why this ended up being the last pylib area to get ported to Rust? :-)

RumovZ:
At first I thought I'd just write a function for obtaining a test-ready collection, but that would probably not be very reusable, because most tests don't require a collection on disk. Also, even if it was so rich that it could be used by all kinds of tests, the time saved by not having to insert data points for a test yourself would probably be somehwat spoiled by having to look at the construction code to find usable data points.
So it seems it would make more sense to write tools to simplify constructing a test collection. Maybe a `TestCollectionBuilder` and some macros for passing in arguments? The idea is to do something like `builder.add_note(note!("Basic", "Some Deck", ["field1", "field2", ...]))`.
Does that make sense, or do you see a better way? 
<hr>

Some other points:
- The old import code is copying all latex and underscored media. Should we be more selective, and use the same criteria as in exporting apkgs?
- We may import unused decks and notetypes, if the notes and cards that use them end up not being imported. Should we keep track of users, and if necessary remove unused objects again? Ideally, we wouldn't add them in the first place, but I think that would add even more complexity.

dae:
Does a builder and a macro buy you much there? Could it perhaps be done more simply using something like the following?

```rust
let mut col = open_test_collection();
col.add_test_note("Basic", "Some Deck", &["field1", "field2", ...])
```

If add_test_note were `#[cfg(test)]`, it stays out of our official API.

> The old import code is copying all latex and underscored media. Should we be more selective, and use the same criteria as in exporting apkgs?

I'd be tempted to just copy any unreferenced files verbatim like the old code for now, as the exporting side should be limiting it for us, and we have enough on our plate to deal with already :-)

> We may import unused decks and notetypes, if the notes and cards that use them end up not being imported.

It's a bit of a corner case and relatively harmless, so we can probably just ignore this issue for now.

RumovZ:
> Does a builder and a macro buy you much there?

The builder would help with stuff like media that needs to set up some context.
Tests usually only care about one specific aspect of an object, and macros would allow to pass in only the relevant arguments (without having to use lots of `None`s).
However, both can also be dealt with differently. You're probably right that we're better served with something simpler.

+1 for skipping those edge cases. 😅

RumovZ:
One of the last things I have to take care of is the log. Would it make sense to unify that with the progress callback? So in steps that log something (currently only during note handling I think), instead of updating the progress message we send log entries (or could do both). When the frontend periodically queries the progress status, it can take these entries and display them in realtime.
Hopefully, this would increase perceived responsiveness without us having to add new progress strings, and help with troubleshooting if an import fails, because a partial log would be provided.

Re testing, after a lot of moving stuff around, I realised many things can better be tested module-wise, instead of in an all-encompassing roundtrip test. Let me know if you'd like to see more tests for anything in particular.

dae:
I haven't stepped through all the logic of the latest changes, but from a brief read it's looking great Rumo :-)

Would it be worth plumbing into the GUI side of things now? The existing Python code could continue to be used by default, with this new code being enabled if an env var is defined.

When you talk about log entries, I presume you mean things like "[Skipped] some example note", rather than the summary counts like "Skipped 4 notes whose note type has changed"? The former could be tricky, as currently there is a single active progress state. If multiple long-running operations are happening at the same time, they'll overwrite each other's progress  periodically. Each screen that monitors progress looks out only for relevant progress, and ignores anything else it receives. If the media sync screen were open while an import was happening for example, it would inadvertently end up fetching those log entries, which is not ideal as there are potentially thousands of them.

The progress callback was intended mainly for small amounts of data, and we typically call it ~10 times a second, so it doesn't feel like the best place to be storing a large log of changes. Might it perhaps be better to store just summary counts there, and pass back the full log on completion? You mentioned that a partial log would be available on failure, but in most cases wouldn't we be rolling back all changes if something goes wrong?

Re tests, playing around with it in the GUI may reveal some things that have been overlooked. I was also going to suggest an undo test, but it looks like you've done that already. :-)

RumovZ:
> Would it be worth plumbing into the GUI side of things now?

Sounds reasonable. Should I start with reusing the existing UI, or create a new svelte page? I expect we'll need a new screen with a webview anyway, once we want to add the discussed features.

> When you talk about log entries, I presume you mean things like "[Skipped] some example note", rather than the summary counts like "Skipped 4 notes whose note type has changed"?

Yes, exactly.

> You mentioned that a partial log would be available on failure, but in most cases wouldn't we be rolling back all changes if something goes wrong?

The idea was that the user would be able to see what did work, which might help narrow down the issue.
Anyway, that would only have been a small advantage in theory, and is by far outweighed by the arguments you make against streaming log entries.

I'll just add logging in the style of the Python code then.

dae:
A Svelte page would be good, but I'd be tempted to get csv importing going before we do that, so we have a better idea of how the field mapping will be represented. In the short term, maybe plugging into the existing UI would be simplest? No strong feelings though; it all needs doing at one point.

dae:
While this isn't hooked up to the GUI yet, I couldn't resist having a little play with it :-) You may have already fixed this locally, but we can't assume a particular filename on import:

```diff
diff --git a/rslib/src/import_export/package/apkg/import/mod.rs b/rslib/src/import_export/package/apkg/import/mod.rs
index d5399804b..bb5dc7d9f 100644
--- a/rslib/src/import_export/package/apkg/import/mod.rs
+++ b/rslib/src/import_export/package/apkg/import/mod.rs
@@ -96,7 +96,8 @@ impl ExchangeData {
 }

 fn collection_to_tempfile(archive: &mut ZipArchive<File>) -> Result<NamedTempFile> {
-    let mut zip_file = archive.by_name(Meta::new_legacy().collection_filename())?;
+    let meta = Meta::from_archive(archive)?;
+    let mut zip_file = archive.by_name(meta.collection_filename())?;
     let mut tempfile = NamedTempFile::new()?;
     io::copy(&mut zip_file, &mut tempfile)?;
     Ok(tempfile)
```

Aside from that, from my very brief testing so far, it seems to be working well :tada: 

For the logging code, while it's what the existing code was doing, we might not want to be dumping the entire note contents on import, especially if it contains heaps of HTML. Instead of nested lists, what do you think about doing something like fields.map(strip_html).join(",")[:80] instead? It may also be worth including the resulting note id as part of the message in the future - that way we could make each line clickable to reveal the imported note in the browse screen. Not urgent, just a possible usability improvement for the future.

RumovZ:
> You may have already fixed this locally, but we can't assume a particular filename on import:

I hadn't, thanks!

> Instead of nested lists, what do you think about doing something like fields.map(strip_html).join(",")[:80] instead?

Forgot about html stripping, which the Python code is already doing.
The idea was for the GUI to take care of displaying separate fields. I think some sort of table would be much more legible than a continuous string with embedded separators (`,` or any other).
As for the slicing, maybe we could truncate each field instead? Truncating at a variable position like `80 / fields.len()` would still allow for limiting the total size. Of course, in reality this is more complictated, because we have to take into account char boundaries. I assume 80 is supposed to be the number of chars, not bytes.
Logging the id sounds useful.

dae:
Another option would be to only display the first field - if the user can click on a row to reveal it in the browse screen, the extra fields may not add a lot of value. 

One potential problem with displaying a table is if there are multiple notetypes with different number of fields - were you thinking the headers would simply be numbers and not try to convey what each field was called? We also need to think about how to show the table - I'd prefer not to go down the QTableView route, as then the mobile clients will need to implement their own solutions. We could do it with HTML, but a `<table>` with 100,000 rows will be slow to render, and we'd probably need some solution that renders to the rows only when they're shown (eg https://svelte.dev/repl/f78ddd84a1a540a9a40512df39ef751b?version=3.47.0)

RumovZ:
> Another option would be to only display the first field - if the user can click on a row to reveal it in the browse screen, the extra fields may not add a lot of value.

Good point. It's only available for successfully imported notes, though.

> were you thinking the headers would simply be numbers and not try to convey what each field was called?

Yes, wouldn't we need to log notetypes as well to get field names?

I didn't think of rendering performance, but it sounds doable, doesn't it? And of course, we can just join the fields on the frontend until we have something more intricate set up.
Should I just leave it for now? I guess a Vec is the most flexible, and if it turns out to be superfluous when we get to implementing the new webview, it should still be easy to return a String instead.

dae:
> It's only available for successfully imported notes, though.

Good point, I overlooked that.

> Should I just leave it for now? I guess a Vec is the most flexible,

Yeah, it's not urgent; I was mainly thinking about performance. Back when originally moving DB access to Rust (04ca8ec038ebfddc474ea9cbfdf9bfec46a5c140), I found that there was quite a bit of overhead with nested arrays in the Python protobuf implementation, so I was thinking that joining the string on the backend (or returning a single field) would end up faster.

RumovZ:
Setting `ANKI_BACKEND_IMPORT_EXPORT` will now enable the new import routines. Exporting will be next.

Implementing `CollectionOpWithBackendProgress` was more complicated than I thought, because for it to replace the old progress code in `replace_with_apkg()`, there also needed to be a CollectionOp that works with a closed collection. Hope you don't mind the additional classes.

RumovZ:
Some more questions and notes:

- The current release offers both the new and old colpkg export. Should we continue to do so? Then maybe with the new as the default, and the old one labeled legacy?
- I think you said that there will need to remain an "upgrade path" for third party imports. What exactly do you have in mind? Will this be provided by vanilla Anki? Is the old Python code going to stay for the foreseeable future? Or is that code intended to be ported as well?
- Exporting is awfully slow. Not sure why, yet.
- How shall I proceed? I guess CSV is next. You mentioned JSON—will this be the new plaintext standard? Should CSV be transformed to JSON first then, or should there be dedicated deserialisers?

Before I continue with new features, I'll wait until you've had a chance to review this in detail, because the current state is relatively self-contained, and should offer an opportunity for optimisations and refactoring (not to mention bug fixes). The only things that come to mind that will still certainly change, are the import log and the export progress.
I know you're still busy dealing with the aftermath of 2.1.50, but there's no hurry.

dae:
Will give this a good look over tomorrow and get back to you. Was going to suggest transactions as the issue, but you beat me to it :-) Without an outer one, one or more fsyncs are required with each inner operation, leading to slow performance like in d100f7a2c8681b2d4dd1bf03797ef8fb3157358b

dae:
Hi Rumo,

I fear I'm going to run out of time today and not have a chance to review this fully (sorry :frowning: ),  but I'll try to at least make a start, and will follow up with one or two more messages. Apologies if things are a bit scattered :-)

Initial impressions: 
- On a moderately large collection without media, exporting went from 8s to 4s - nice! :tada: 
- When I adjust the code to use zstd instead, it drops to about 2.6s - even nicer! At one point perhaps we should add a "Anki 2.1.55+ (faster)" or similar checkbox, like we have for colpkg files. No rush though, as long as the importing side will be ready for it.
- Importing is even more impressive: 12.5s->4s on a legacy package.
- "Notes skipped: 0" and similar messages could perhaps be omitted in the 0 case on import? Lets the user more easily focus on what actually needs attention.

qt/aqt/import_export/exporting.py
```diff
@@ -0,0 +1,219 @@
...
+    def export(mw: aqt.main.AnkiQt, options: Options) -> None:
+        QueryOpWithBackendProgress(
+            parent=mw,
+            op=lambda col: col.export_anki_package(
+                out_path=options.out_path,
+                selector=options.note_ids or options.deck_id or None,
+                with_scheduling=options.include_scheduling,
+                with_media=options.include_media,
+            ),
+            success=lambda fut: tooltip(tr.exporting_note_exported(count=fut)),
```
dae, line 217: The way QueryOpWithBackendProgress() is currently implemented unfortunately means we're losing typechecking here. 

- mypy has no way of knowing that the result of export_anki_package() will be passed to the success fn. You've labelled the argument 'fut' there, but it's actually an int. I think some code below is also incorrectly typed as a future, but we're not being notified about it currently.
- code completion/hovering in export_anki_package() is broken because the editor doesn't know that col is a Collection

I presume the current design was motivated by the fact that collection and query ops take different op callables, and you wanted to avoid code duplication. I think typechecking/code completion is more important here, and would be fine with some of that code being duplicated if it meant typechecking worked like standard collection/query ops.

Rather than subclassing the original classes, would it make sense to enable this additional behaviour via an extra argument to the original CollectionOp/QueryOp classes instead? Eg if a progress_key arg were provided, the timer could be started up, and ._run() could take a different codepath. Perhaps some/all of OpWithBackendProgress._run() could be turned into a free-function that both QueryOp and CollectionOp could call?

One other point to note is that currently we need to teach OpWithBackendProgress() about each new progress type we want to handle. The approach I took in AnkiMobile instead delegates extraction/label updating to the calling code. Eg, instead of passing in a label directly, it passes in a function that extracts the data it wants, and if it finds it, it calls the equivalent of mw.progress.update(...) itself. That way the calling code has control over how the data is rendered, and the code related to a specific operation lies in the file dealing with that operation, instead of in our shared operations code.

qt/aqt/import_export/importing.py
```diff
@@ -0,0 +1,124 @@
...
+
+def show_import_log(future: Future) -> None:
+    log = future.log  # type: ignore
+    total = len(log.conflicting) + len(log.updated) + len(log.new) + len(log.duplicate)
+
+    text = f"""{tr.importing_notes_found_in_file(val=total)}
+{tr.importing_notes_that_could_not_be_imported(val=len(log.conflicting))}
+{tr.importing_notes_updated_as_file_had_newer(val=len(log.updated))}
+{tr.importing_notes_added_from_file(val=len(log.new))}
+{tr.importing_notes_skipped_as_theyre_already_in(val=len(log.duplicate))}
```
dae, line 112: Repeating my earlier comment here for reference: might be better to hide lines that have zero counts.

qt/aqt/main.py
```diff
@@ -402,15 +404,14 @@ def doOpen(path: str) -> None:
...
-
-        import aqt.importing
-
         self.restoring_backup = True
         showInfo(tr.qt_misc_automatic_syncing_and_backups_have_been())
-        aqt.importing.replace_with_apkg(self, path, on_done)
+
+        import_collection_package_op(self, path).success(
+            lambda _: self.onOpenProfile(
+                callback=lambda: self.col.mod_schema(check=False)
```
dae, line 412: I think we can safely remove the mod_schema() line now, due to 6735b23e5bd3257c1c36d48f4068b991a3c24556

rslib/src/import_export/package/apkg/import/media.rs
```diff
@@ -0,0 +1,158 @@
...
+    fn new(progress_fn: F) -> Self {
+        Self {
+            progress_fn,
+            counter: 0,
+        }
+    }
+
+    fn increment(&mut self) -> Result<()> {
+        self.counter += 1;
+        if self.counter % 17 == 0 {
```
dae, line 50: I've copied this number in https://github.com/ankitects/anki/pull/1743/commits/23eda19677eb880f6468370e11c492f425c9d144 - I presume you did this so the rightmost digit is not always zero?

dae (commented):
Still working my way through the Rust code

rslib/src/import_export/package/apkg/import/cards.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn map_to_imported_note(&self, card: &mut Card) -> bool {
+        if let Some(nid) = self.imported_notes.get(&card.note_id) {
+            card.note_id = *nid;
+            true
+        } else {
+            false
+        }
+    }
+
+    fn target_already_exists(&self, card: &Card) -> bool {
```
dae, line 117:
```suggestion
    fn card_ordinal_already_exists(&self, card: &Card) -> bool {
```

RumovZ:
I hadn't done any benchmarking myself, yet. Great to read these numbers! 🎉

> "Notes skipped: 0" and similar messages could perhaps be omitted in the 0 case on import? Lets the user more easily focus on what actually needs attention.

This was only supposed to be a debug aid, until we implement the new webview screen. Same with the export screen.

> You previously mentioned potentially using sqlite as an intermediate format. That could work too, and means we don't need to worry about working memory. Like protobuf, it makes things a fair bit more complicated for third-party authors though.

I had rather thought about sqlite replacing the current `ExchangeData` Rust struct, not the common frontend format. For the latter, a self-describing, ubiquitous format like JSON makes sense.

> My thinking was that perhaps we could build some sort of adapter that mimicked the old NoteImporter API, but converted the data into our new generic format, so we can feed it into our new importing code.

I have only looked at `NoteImporter` superficially, but I guess writing such an adapter would be a challenge, due to the nature of Python. There is no clear-cut API, and inheritors may just call the db directly if they feel like it.

rslib/src/import_export/package/apkg/import/decks.rs
```diff
@@ -0,0 +1,202 @@
...
+        for config in &mut configs {
+            config.usn = self.usn;
+            self.target_col.add_deck_config_if_unique_undoable(config)?;
+        }
+        Ok(())
+    }
+
+    fn import_decks(&mut self, mut decks: Vec<Deck>) -> Result<()> {
+        // ensure parents are seen before children
+        decks.sort_unstable_by(|d1, d2| d1.name.as_native_str().cmp(d2.name.as_native_str()));
```
dae, line 49: What happens if the order is wrong? A child deck is not guaranteed to match the case of its parent, and we can't use simple ASCII case folding here.

dae (commented):
Almost done...

rslib/src/notes/mod.rs
```diff
@@ -55,6 +55,10 @@ impl Note {
         &self.fields
     }
 
+    pub fn take_fields(self) -> Vec<String> {
+        self.fields
```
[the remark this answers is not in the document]
RumovZ: Exactly, I used `take()` before, and didn't update the name! 😃 (Reasoning was that a note without fields is useless, so we'd better consume it.)

rslib/src/import_export/package/apkg/import/media.rs
RumovZ: Yes, looks more dynamic to me. Also, the number might be a bit low. I'll probably genralise `ProgressHandler`, so it can be used at some other places as well, and make the frequence customisable.

rslib/src/import_export/package/apkg/import/cards.rs
```diff
@@ -0,0 +1,178 @@
...
+    imported_notes: &'a HashMap<NoteId, NoteId>,
+    remapped_decks: &'a HashMap<DeckId, DeckId>,
+
+    /// The number of days the source collection is ahead of the target collection
+    collection_delta: i32,
+    scheduler_version: SchedulerVersion,
+    /// Cards in the target collection as (nid, ord)
+    targets: HashSet<(NoteId, u16)>,
+    /// All card ids existing in the target collection
+    target_ids: HashSet<CardId>,
```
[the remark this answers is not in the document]
RumovZ: For the former, maybe `existing_cards` and a type alias?

rslib/src/import_export/package/apkg/import/notes.rs
```diff
@@ -0,0 +1,415 @@
...
+        LogNote {
+            id: Some(self.id.into()),
+            fields: self
+                .take_fields()
+                .into_iter()
+                .map(|field| {
+                    strip_html_preserving_media_filenames(&field)
+                        .get_owned()
+                        .unwrap_or(field)
+                })
```
dae, line 76: Perhaps for now instead of trying to enforce a total size of a record, we could just limit any given stripped field to 80 chars? That should at least avoid issues where users have inserted megabytes of text (copying webpages verbatim for example).

rslib/src/import_export/package/apkg/import/notes.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    fn import_notes(&mut self, notes: Vec<Note>, progress_fn: &mut ProgressFn) -> Result<()> {
+        for (idx, mut note) in notes.into_iter().enumerate() {
+            // FIXME: use ProgressHandler.increment()?
+            if idx % 17 == 0 {
+                progress_fn(ImportProgress::Notes(idx))?;
+            }
+            if let Some(notetype_id) = self.remapped_notetypes.get(&note.notetype_id) {
+                if self.target_guids.contains_key(&note.guid) {
+                    // TODO: Log ignore
```
dae, line 190: todo

dae (commented):
Ok, that's all the files reviewed :-)

One thing that might be nice that I forgot about - can we reset flags when exporting with scheduling off?: https://github.com/ankitects/anki/issues/1017

Overall this is looking great, and I'm looking forward to integrating this. After the above points are worked through, what do you think about switching from an env var to checking the shift key? To start with, we could invert the usual behaviour - the old Python code being used by default, except if the user holds down shift. This will make it easier for people to try out the new code, and once it's been used for a bit in the wild, we can look at switching the default so that shift shows the legacy importer instead. There aren't a large number of changes to existing codepaths, so I don't feel very anxious about including this in a release before .55 (with the default remaining the Python codepath to start with). The webview handling/CSV importer/etc could follow in future commits. 

Does that work for you? If you would prefer to keep this out of main for a while that's fine too. :-) We might want to at least backport that Interrupted fix though.

dae:
>I had rather thought about sqlite replacing the current ExchangeData Rust struct, not the common frontend format. For the latter, a self-describing, ubiquitous format like JSON makes sense.

Yep, that makes more sense. What is your thinking about moving ExchangeData to SQL at this point? Is it still something you'd like to do, or are you happy with the current implementation for now? From my end, at first glance it looks like it wouldn't be too hard to swap out in the future if we decide we need to do so, so I'm happy to keep things as they are for now if you are.

> I have only looked at NoteImporter superficially, but I guess writing such an adapter would be a challenge, due to the nature of Python. There is no clear-cut API, and inheritors may just call the db directly if they feel like it.

The core of the API is an importNotes() method call, which takes the provided ForeignNotes, and the provided current model+mapping of fields to do an import. Importers may make direct DB calls prior to or after that call, but the import process is basically all in that function - so adapting it to munge things into our JSON format may not necessarily be impossible (at least, if importers haven't overriden NoteImporter's internal implementation). I'm not suggesting this is the best course of action, just that we may not need to rule it out just yet. 

That said, we may be best off leaving the old code lying around for now (eventually behind a shift+click) - it offers the best compatibility with existing add-ons, and it's less work for us as well.

rslib/src/import_export/package/apkg/import/decks.rs
RumovZ: Not so sure myself. I think a default deck would be added, and then updated when it's finally the parent's turn to be imported. Maybe we can sort by level instead.

RumovZ:
Flags should already get removed: https://github.com/ankitects/anki/pull/1743/files#diff-839ec35e7520e8ddc5460e4a3e91b3cbb7fcb8fd0a1b999b4155f07a18f8b84dR109

This PR is starting to get unwieldy, so no objections to merging it in, as long as it's a bit hidden. That does mean putting some more effort in mere stopgaps, though (as noted above).
I thought about making a PR for the Interrupted error, but it doesn't look like it's currently exposed. Some ops throw exceptions when canceled for other reasons, though.

I currently don't see any need to switch to SQLite, but (spontaneous intuition) maybe it could be used for NoteImporter. Could we exectue `importNotes()` on a fresh Collection, and then leverage the apkg importer?

qt/aqt/import_export/exporting.py
```diff
@@ -0,0 +1,219 @@
...
+    tr,
+)
+
+
+class ExportDialog(QDialog):
+    def __init__(
+        self,
+        mw: aqt.main.AnkiQt,
+        did: DeckId | None = None,
+        nids: Sequence[NoteId] | None = None,
```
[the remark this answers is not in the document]
RumovZ: I found a slightly different solution. Hope it's also okay for you.

dae:
Re flags, nice, sorry I missed that.

> I currently don't see any need to switch to SQLite, but (spontaneous intuition) maybe it could be used for NoteImporter. Could we exectue importNotes() on a fresh Collection, and then leverage the apkg importer?

I'm not sure I understand what you're proposing here. Are you suggesting we keep the existing NoteImporter code, run it on a temporary collection, and then import that collection into the main collection? If that's what you meant, what would that gain us over just keeping the existing NoteImporter code that imports directly?

rslib/src/import_export/package/apkg/import/notes.rs
```diff
@@ -414,4 +358,95 @@ mod test {
...
+        note.fields_mut()[0] = "<img src='foo.jpg'>".to_string();
+
+        let mut media_map = MediaUseMap::default();
+        let entry = SafeMediaEntry::from_legacy(("0", "bar.jpg".to_string())).unwrap();
+        media_map.add_checked("foo.jpg", entry);
+
+        let mut log = import_note!(col, note, media_map);
+        assert_eq!(col.get_all_notes()[0].fields()[0], "<img src='bar.jpg'>");
+        assert_note_logged!(log, new, &[" bar.jpg ", ""]);
+    }
```
dae, line 451: These unit tests are lovely and readable!

rslib/src/import_export/package/apkg/import/decks.rs
```diff
@@ -46,7 +46,7 @@ impl DeckContext<'_> {
 
     fn import_decks(&mut self, mut decks: Vec<Deck>) -> Result<()> {
         // ensure parents are seen before children
-        decks.sort_unstable_by(|d1, d2| d1.name.as_native_str().cmp(d2.name.as_native_str()));
+        decks.sort_unstable_by_key(|deck| deck.level());
```
dae, line 49: This was a clever way of avoiding case issues, and performs well to boot - well done!

pylib/anki/collection.py
```diff
@@ -92,6 +92,33 @@ class LegacyCheckpoint:
 LegacyUndoResult = Union[None, LegacyCheckpoint, LegacyReviewUndo]
 
 
+class ExportLimit:
+    """Limit to what will be exported. Either specific notes, or a deck, or the
+    whole collection (if neither is set). Only the last set value is preserved.
+    """
```
dae, line 98: This is similar to how protobuf oneofs perform, so there is precedent, but I suspect the multiple-wrapper-classes is a bit more Pythonic. Some disadvantages of this approach:

- It requires some more boilerplate, especially as each added case requires updates to the code for the other cases. 
- The implicit clearing of other fields is not normal/intuitive, so it makes it a bit easier to make mistakes.

RumovZ:
> Are you suggesting we keep the existing NoteImporter code, run it on a temporary collection, and then import that collection into the main collection? If that's what you meant, what would that gain us over just keeping the existing NoteImporter code that imports directly?

Yes. It might allow us to wrap it in some of the new logic. E.g., it could enable undoing, because we wouldn't write to the db directly, it would be easier to merge with the other new routines (progress and so on), and future customisation options would be supported for free. Of course, it would be slower, but maybe not by much if we used a in-memory collection.
That said, I don't know how well it would work in practice. It's only been an idea.

pylib/anki/collection.py
RumovZ: It was an attempt to make this more ergonomic and avoid instance checks, but the fact that I used my own API wrong convinces me, that it failed. 😄 I'll revert it to what you suggested.

qt/aqt/import_export/exporting.py
RumovZ: You're right, and it's a bad use of inheritance. One more drawback: The signature of the child classes is rather unhelpful, because they have to accept arbitrary args and kwargs.
Implementing this in a free function is definitely possible (if state is maintained in a returned closure or so), but there already is a class `ProgressHandler` that we call. Would make it sense to place the backend progress logic there as well?

dae: Do you mean ProgressManager? I only see ProgressHandler in our Rust code. If you meant putting the generic backend progress handling code rather than area-specific code into ProgressManager, that's basically what I did in AnkiMobile, so it works for me :-) The AnkiMobile equiv looks roughly like this - a routine that the calling code can call after it calls progress.start():

```
    func monitorProgress(
        secs: Double,
        callback: (BackendProgress.OneOf_Value) -> Void
    ) {
        self.progressTimer = Timer.scheduledTimer(
            withTimeInterval: secs,
            repeats: true,
            block: { _ in
                // if there's an active progress value, send it to the callback, which will check it's the
                // type it expects, and then act on it if it is
                if let progressEnum = try? am.col.backend.latestProgress(),
                    let progress = progressEnum.value {
                    callback(progress)
                }
            }
        )
    }
```

The timer is automatically cleaned up on progress.finish().

RumovZ: Yes, sorry for the typo. And thanks for the snippet!

dae:
Ok, I see what you mean. My initial reaction is that this might not be the best path forward, as I'd be a bit worried about it not working in all cases NoteImporter currently does, and it would mean we're spending energy maintaining the old code. Keeping the old importing screen as-is means the old importers don't benefit from undo support, but in exchange, it a) is the least likely to break compatibility/introduce regressions, and b) requires very little work from us :-)

The downside of keeping the old code around is that in the long term it will become a maintenance burden, so we don't want to do so indefinitely, and will want to figure out some upgrade path (such as a modern equivalent of NoteImporter). I thought an automatic adaptor might be a partial solution there (the nasty logic of NoteImporter could be removed, and only a shim would remain), but keeping the old code around for now is easier.

(As you are probably already aware), NoteImporter is not the only reason for keeping around the old importing screen - we also need it to support the Specials Fields apkg import case, as our Rust code currently doesn't provide such functionality.

dae:
Minor issue: "Importing collection... " is shown when importing apkg files as well; could give users a fright. Maybe we could rename it to ImportProgress::File and make it "Importing file..." (or "Importing {filename}") so that we could use it with various different importers?

qt/aqt/import_export/exporting.py
RumovZ: Good call, that was a big improvement. Do you still see any issues?

rslib/src/import_export/package/apkg/import/notes.rs
```diff
@@ -185,11 +185,10 @@ impl<'n> NoteContext<'n> {
...
 
     fn import_notes(&mut self, notes: Vec<Note>, progress_fn: &mut ProgressFn) -> Result<()> {
-        for (idx, mut note) in notes.into_iter().enumerate() {
-            // FIXME: use ProgressHandler.increment()?
-            if idx % 17 == 0 {
-                progress_fn(ImportProgress::Notes(idx))?;
-            }
+        let mut progress = IncrementalProgress::new(|u| progress_fn(ImportProgress::Notes(u)));
+
+        for mut note in notes.into_iter() {
```
dae, line 190: Very minor: can be note in notes

qt/aqt/import_export/exporting.py
dae: Much nicer! 

Having the callback return a string is the most ergonomic in cases like this, but there are cases where we want to update the progress bar as well, such as dbcheck.py, and full syncs in sync.py. I'm not suggesting we rewrite those existing cases, just that it's a use case we might want to more explicitly support for future code. We could do this by having the callback be passed in the mw.progress instance, and have it make the update call itself, eg:

```python
# before
def export_progress_label(progress: Progress) -> str | None:
    if not progress.HasField("exporting"):
        return None
    return tr.exporting_exported_media_file(count=progress.exporting)

# after
def on_backend_progress(manager: ProgressManager, progress: Progress):
    if not progress.HasField("exporting"):
        return
    manager.update(label=tr.exporting_exported_media_file(count=progress.exporting))
```

Code could then pass in value= and max= if it wanted to update the progress bar.

We could theoretically do both - return a label, and allow the routine to also call .update() itself. That just feels a little less efficient and a bit less clear. WDYT?

RumovZ: As your function doesn't return anything, can we `set_wants_abort()` then regardless of whether the closure processed the progress message?
Instead of passing the whole ProgressManager and expecting the user to call the right method, what do you think of taking the arguments for `update()`? As its signature is a bit cumbersome, we could accept a dataclass instead.

```py
def progress_update(progress: Progress) -> ProgressUpdate | None:
    if not progress.HasField("exporting"):
        return None
    return ProgressUpdate(label=tr.exporting_exported_media_file(count=progress.exporting))
```

rslib/src/import_export/package/apkg/export.rs
```diff
@@ -0,0 +1,98 @@
...
+        },
+    },
+    io::{atomic_rename, tempfile_in_parent_of},
+    prelude::*,
+    storage::SchemaVersion,
+};
+
+impl Collection {
+    /// Returns number of exported notes.
+    #[allow(clippy::too_many_arguments)]
```
RumovZ, line 27: Should we add a new options struct to avoid this lint, or pass the proto message directly?

qt/aqt/import_export/exporting.py
dae: Don't have a strong preference here; returning a dataclass sounds fine too.

> can we set_wants_abort() then regardless of whether the closure processed the progress message?

At one point in the future we may want to handle multiple long-running ops at once (at the moment if a media sync is running at the same time, what get cancelled will depend on which progress happens to run next), and at that time we'll probably want to be only calling that in a routine-specific code (so that it can eg. pass an enum back to set_wants_abort() specifying what is to be aborted). So an argument could be made for moving the check out of progress.py and into the individual progress handlers instead, at the cost of some more work in the short term.

rslib/src/import_export/package/apkg/export.rs
dae: I can live with that warning for now, but if you'd like to avoid it, the native struct might be better, as then we wouldn't need to divide the args up between the proto struct and separate ones, and could make a builder. At one point it might be worth looking into something like https://crates.io/crates/derive_builder as well - but probably better not in this PR :-)

qt/aqt/import_export/exporting.py
RumovZ: Not sure I got this. Doesn't the GUI prevent the user from aborting any spcific op anyway, because it's only a single dialog?
If that's not relevant for this PR, feel free to just resolve this conversation.

RumovZ:
> [String::truncate() is a bit of a footgun

Indeed, I mentioned the pitfalls of truncating Unicode somewhere above only recently, but apparently immediately forget about it when implementing. 🤦‍♂️ 

Keeping NotImporter as-is for now sounds reasonable. There are some options on the table now, and a modern plaintext importer has to be implemented first, anyway.

Apart from what I've last commented above, is there anything else that should be addressed in this PR?

qt/aqt/progress.py
```diff
@@ -185,11 +186,11 @@ def on_progress() -> None:
             assert self.mw
 
             progress = self.mw.backend.latest_progress()
-            if not (label := label_from_progress(progress)):
+            if not (update := progress_update(progress)):
                 return
             if dialog and dialog.wantCancel:
                 self.mw.backend.set_wants_abort()
-            self.update(label=label)
+            self.update(**asdict(update))
```
dae, line 193: I'd rather we explicitly pass in each arg, so we don't lose typechecking. We only need to worry about label, value and max. process is unused by update() and can be ignored, and I don't think maybeShow is used either.

rslib/src/import_export/mod.rs
```diff
@@ -1,10 +1,98 @@
...
+    MediaCheck(usize),
+    Notes(usize),
+}
+
+/// Wrapper around a progress function, usually passed by the [crate::backend::Backend],
+/// to make repeated calls more ergonomic.
+pub(crate) struct IncrementableProgress<P> {
+    progress_fn: Box<dyn FnMut(P, bool) -> bool>,
+    count_map: Option<Box<dyn FnMut(usize) -> P>>,
+    count: usize,
```
dae, line 23: I was a bit concerned about the extra overhead of Box<dyn>, as we may be invoking increment() in a hot loop. Whether it actually matters or not in our current code I don't know - the other stuff we're doing in the loop like SQL lookups or sha1 calculations probably dwarfs it.

One possible alternative would be to change set_count_map() to zero(), get rid of count_map, and pass an impl into increment instead. But either is ok with me.

rslib/src/import_export/mod.rs
```diff
@@ -1,10 +1,98 @@
...
+}
+
+const UPDATE_INTERVAL: usize = 17;
```
dae, line 26: We could include this as an associated value, though not sure how much of an improvement that is.

```diff
diff --git a/rslib/src/import_export/mod.rs b/rslib/src/import_export/mod.rs
index 8f9825e59..b6afc04a4 100644
--- a/rslib/src/import_export/mod.rs
+++ b/rslib/src/import_export/mod.rs
@@ -23,9 +23,9 @@ pub(crate) struct IncrementableProgress<P> {
     count: usize,
 }

-const UPDATE_INTERVAL: usize = 17;
-
 impl<P> IncrementableProgress<P> {
+    const UPDATE_INTERVAL: usize = 17;
+
     /// `progress_fn: (progress, throttle) -> should_continue`
     pub(crate) fn new(progress_fn: impl 'static + FnMut(P, bool) -> bool) -> Self {
         Self {
@@ -47,7 +47,7 @@ impl<P> IncrementableProgress<P> {
     /// Must have called `set_count_map()` before calling this.
     pub(crate) fn increment(&mut self) -> Result<()> {
         self.count += 1;
-        if self.count % UPDATE_INTERVAL != 0 {
+        if self.count % Self::UPDATE_INTERVAL != 0 {
             return Ok(());
         }
         let progress = self.mapped_progress()?;
```

Alternatively we could optionally pass it into new. There may be a day that we have a routine where each iteration of the loop is long, and we don't want to wait 17 iterations to update.

qt/aqt/import_export/exporting.py
dae: If some cancellable operation + media sync is running at the same time, closing the progress window will cancel one of the two operations - whichever happens to post progress next. As that's not ideal, we'll probably want to fix it at one point in the future. One possible way to do so would be to change set_wants_abort() to take an enum or string specifying the action that we want to cancel. I'm not suggesting we do that now, but if we leave the the set_wants_abort() call in start_with_backend_updates(), that's going to make it harder to pass an enum in the future. If instead progress_update() were passed a wants_abort, it could then call set_wants_abort() itself, after it's determined that the progress message is the one it's interested in. That wouldn't change anything at the moment, but would make it easier to pass in a relevant enum in the future.

Does that make sense? Alternative ideas welcome :-)

dae:
> Apart from what I've last commented above, is there anything else that should be addressed in this PR?

Looking good to me! Once you're ready, please remove the draft tag and I'll do one final check.

rslib/src/import_export/mod.rs
RumovZ: Yeah, with all the I/O and allocations going on, it's hard to imagine that some derefs will carry weight.
But I added a new struct that is generic over `count_map`. That's an improvement with regards to the API, because users can't forget to call `set_count_map()` first.
However, I don't see how we can ergonomically use a generic for `progress_fn`, because we'd like to pass that struct around, and each time annotating it with `&mut IncrementableProgress<ImportProgress, impl FnMut(P, bool) -> bool>` seems pretty cumbersome. Here's hoping that `impl`s in type aliases will be stabilised at some point.

rslib/src/import_export/mod.rs
RumovZ: I made it a field of `Incrementor`. Should be easy enough to add a setter when the need arises. I think requesting it in `incrementor()` might be a bit irritating, because it would require consideration by the programmer each time, and in reality, we hardly have an idea what a reasonable value is, or do we?

qt/aqt/import_export/exporting.py
RumovZ: Thanks, got it. I was referring to the fact that the GUI currently only has a single cancel signal, so the user can't decide which op to abort anyway, but I see that's not the point here.

As we're getting data back from the callback anyway, couldn't we just add a `ProgressUpdate.abort_variant` that gets passed on to `set_wants_abort()` if appropriate? Alternatively, `start_with_backend_updates()` could also take an `abort()` callback.
My issue with passing a `wants_abort` (or primitives in general) into the callback is, that it's hard to communicate its meaning. This may not be a problem in single-purpose callbacks, but here the type would be `Callable[[Progress, bool], ProgressUpdate | None]`. Then you'll have to explain in comments what the caller is supposed to do with that bool. (I guess you could use a newtype, but well...)

rslib/src/import_export/mod.rs
dae: That seems like a good compromise - thanks Rumo. 

>  Here's hoping that impls in type aliases will be stabilised at some point.

Agreed!

rslib/src/import_export/mod.rs
dae: We could pass in Option<usize> with None meaning 17, but no problems keeping it as it is for now.

qt/aqt/import_export/exporting.py
dae: The way I do it in AnkiMobile is pass in a mutable object that also contains the cancellation flag from the GUI, eg something like:

```python
    def progress_update(progress: Progress, update: ProgressUpdate) -> None:
        if not progress.HasField("exporting"):
            return None
        update.label = tr.exporting_exported_media_file(count=progress.exporting)
        if update.user_wants_abort:
            update.abort = True       
```

The invoking progress code then invokes set_wants_abort() if abort is true. That saves us from having to create an enum at the moment, and when we want to add one in the future, only the update.abort line + the set_wants_abort() invocation needs to be changed . Would that work for you?

> couldn't we just add a ProgressUpdate.abort_variant that gets passed on to set_wants_abort() if appropriate?

Unconditionally assigning it each time progress_update() is called feels a bit unclean, though the performance difference would be negligible.

Passing something into start_with_backend_updates() could work, but it might be slightly nicer to have all the logic inside the callback. WDYT?

RumovZ: Yes, makes sense. I've pushed a commit.

dae:
Still some testing to do, but I've pushed a few tweaks I've made so far. Please let me know if there's anything you'd prefer done differently or if you spot any problems.

dae:
Ready from my end, pending your feedback/approval

RumovZ:
Great!
I've replaced [40ba291](https://github.com/ankitects/anki/pull/1743/commits/40ba29118a42b3a14773d4fd48465807e69bbb8c), so it hopefully does the same thing, but is easier to follow.

dae:
:rocket: 

Thanks for all your hard work on this Rumo!

PR #1754, opened 2022-03-28
Fix SearchBuilder

RumovZ:
Found more issues as in e2eff7deaee5d7c380d0cb930186c62fa0309de5, and took the opportunity to simplify things a bit.

### DIFF
--- a/rslib/src/notetype/notetypechange.rs
+++ b/rslib/src/notetype/notetypechange.rs
@@ -295,7 +295,7 @@ impl Collection {
             let ords =
                 SearchBuilder::any(map.removed.iter().map(|o| TemplateKind::Ordinal(*o as u16)));
             self.search_cards_into_table(
-                SearchBuilder::from(nids).and_join(&mut ords.group()),
+                SearchBuilder::from(nids).and(ords.group()),
                 SortMode::NoOrder,
             )?;
             for card in self.storage.all_searched_cards()? {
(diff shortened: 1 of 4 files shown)


dae:
Ouch :-( I'm glad these were caught before a wider release. I have double-checked all the match_all[] conversions now, which I should have done last time this came up.

While the docstrings do point out that no explicit grouping is done, this behaviour has turned out to be a bit of a footgun, and I think we should be considering ways to make the API harder to use incorrectly. How frequently do we actually need to inline a multi-term argument to .and() or .or() without it being in a group? Perhaps implicit grouping should be the default, with a separate way of opting out of it in the few cases where it might be required?

dae:
It's a single class of mistake, but there were 3 separate locations it was made. I reviewed the code and did not notice the issues either, so it's caught out multiple people multiple times at this point - it's not such a stretch to start calling it a pattern :-)

When we call .and(), we're providing the argument inside the brackets of the function call, so I think intuitively, there's an expectation that the argument will be treated as a group, and that is what tripped us up. 

I just started reviewing #1743, and came across this line:

```
SearchBuilder::from(SearchNode::NoteIds(nids)).and(SearchNode::CardIds(cids).negated())
```

With the group issue fresh in my mind, that line jumps out at me, as we're passing multiple ids into .and(). I found myself having to double-check that CardIds does in fact resolve to a single query clause. With the passage of time though, the group issue will likely fade from memory, and I worry that we may accidentally hit it again in the future, and curse our past selves for not fixing it earlier. :-)

RumovZ:
You're right, I'll prepare a PR.
It won't help with the above line though, because we skip grouping for single nodes. Instead, I'd say it's the SQL writer's responsibility to ensure that every search node corresponds to a single SQL node (i.e. parenthesising as needed).

dae:
Agreed. Might be worth adding a comment at the top of write_search_node_to_sql() as a reminder to ourselves to make sure brackets are used when adding any future terms, in case we don't notice all the ones in our sql() unit test.

Thanks Rumo :-)

dae:
Ah, just saw you've put through a PR already! Will take a look and add that comment too.

Issue #1733, opened 2022-03-23
Update demo/docs for editor changes


dae:
Aside from making life easier for add-on authors, an updated demo will help us check the API is working as intended.

Related: https://forums.ankiweb.net/t/2-addons-conflicting-with-svelte-components-editortoolbar-toolbar-insertgroup/18019

luc-languagetools:
many thanks for creating this, I'd love to test any changes, i'm happy to do some of the work if you set me in the right direction

dae:
@luc-languagetools does rc2 resolve the duplicate group issue you were experiencing?

luc-languagetools:
@dae no, I'm not able to get anki_new_format_pack working, so I don't know how to hook into the new editor API:
https://github.com/hgiesel/anki_new_format_pack/issues/8
let me know if more details are needed to reproduce. It could be the problem is with anki_new_format_pack.

luc-languagetools:
If I could have my wish, I would love to see an example addon which creates a whole new component on the editor (not just adding a button like anki_new_format_pack), similarly to https://uploads-ssl.webflow.com/602f82f1abdf727c415dfca9/621f7b75edda7ddf95fa1705_2022-03-02_update_preview_audio.png
I'm happy to sponsor this development with money (please reach out languagetools@mailc.net), as I realize it may be quite specific to my addons.

dae:
This should get you going with the add-on: https://github.com/hgiesel/anki_new_format_pack/pull/9

I'd recommend using its build system as a template for your build, since you need to exclude svelte & anki from bundling, and because your current bundle seems to be including import statements. The following change to your code may help, but I wasn't able to test them.

```
import * as NoteEditor from "anki/NoteEditor";

NoteEditor.lifecycle.onMount(({ toolbar }) => {
    toolbar.toolbar.append({component: LanguageTools, id: "languagetools"});
});
```

luc-languagetools:
I can confirm your latest merge to anki_new_format_pack fixes those new toolbar buttons (like strikethrough) on 2.1.50 rc2, now let me try to update my addons.

luc-languagetools:
I got everything working in HyperTTS now with 2.1.50 rc2 thanks to those updates. thanks a lot for updating the documentation as well, it wasn't clear at all to me that the anki/ dir is an git submodule.

glutanimate:
Thanks for working on updating the docs!

[...]

PR #1756, opened 2022-03-28
Add extra button group for cloze commands

hgiesel:
I was thinking about adding separate buttons for incremented cloze number vs. same cloze number, as an "alt" modifier wouldn't work on mobile.

However I'm struggling with adding a media directory for custom icons under `/ts/icons`. Could you give me a pointer what I'm doing wrong? Currently I just get

> text: 'Could not resolve "../../icons/contain-plus.svg"'


dae:
A filegroup() references files in their source location, but eslint expects everything to be under a single root folder in bazel-bin. By using js_library(), the files are copied into the bazel-bin folder at build time, and they are then accessible by eslint.

FWIW, AnkiMobile currently uses a long-press on the cloze button to make it repeat the number, but having a separate button for it seems reasonable too.

hgiesel:
This is what it looks like now:

Screenshot 2022-03-31 at 00 22 43

hgiesel:
Tell me if this is to be be merged after 2.1.50, because it actually contains a bug fix regarding the editor buttons (and thus relevant for #1733)

dae:
Ideally we'd be just fixing the bug now, and waiting until we have a larger testing window before merging in new features, as there's always a risk something will get accidentally overlooked. I don't want to take up your time asking you to change it now, but for future PRs in the next few weeks, it would be great if we could have any bugfixes separated from feature/refactoring changes. I haven't been very clear about what I was hoping for, and a bit inconsistent with what does/doesn't get merged, so I apologize for any confusion there :-)

Issue #1746, opened 2022-03-28
Consider adding 'relative overdue' to review order


dae:
Re: https://forums.ankiweb.net/t/add-all-filtered-deck-sorting-options-to-the-new-display-order-setting-introduced-with-v3-scheduler/18630

With our ordering being done at query time these days, it wouldn't be difficult to add this, though the existing query was designed for filtered decks and could probably be improved a bit.

PR #1757, opened 2022-03-29
Add relative overdueness to review order

abdnh:
Closes https://github.com/ankitects/anki/issues/1746

Only briefly tested with some dreadful math cards that I abandoned last year 😅

Is there any place where the other review options are tested?


dae:
Thanks Abdo! The Rust part looks ok, but the actual calculation doesn't seem to do what one would expect:

```
[dae] ~  % python test.py
-2.0000	ivl=100	days late=-50
-0.6667	ivl=100	days late=-150
-0.4000	ivl=20	days late=-50
-0.1333	ivl=20	days late=-150
-0.1000	ivl=5	days late=-50
-0.0333	ivl=5	days late=-150
-0.0200	ivl=1	days late=-50
-0.0067	ivl=1	days late=-150
0.0067	ivl=1	days late=150
0.0100	ivl=1	days late=100
0.0200	ivl=1	days late=50
0.0333	ivl=5	days late=150
0.0500	ivl=5	days late=100
0.1000	ivl=5	days late=50
0.1333	ivl=20	days late=150
0.2000	ivl=20	days late=100
0.4000	ivl=20	days late=50
0.6667	ivl=100	days late=150
1.0000	ivl=100	days late=100
2.0000	ivl=100	days late=50
1000.0000	ivl=1	days late=0
5000.0000	ivl=5	days late=0
20000.0000	ivl=20	days late=0
100000.0000	ivl=100	days late=0
[dae] ~  % cat test.py
today = 50
def calc(ivl, due):
    return [ivl / (today - due + 0.001), ivl, due]

res = []
for ivl in 1, 5, 20, 100:
    for due in -100, -50, 0, 50, 100, 200:
        res.append(calc(ivl, due))

res.sort()
for (pos, ivl, due) in res:
    print("%0.4f\tivl=%s\tdays late=%s" %(pos, ivl, today-due))
```

queue/builder.rs has some tests for the new card ordering, which you could use as a template for a separate review-based test, but it would require a bit of boilerplate to set the cards up with relevant values. One alternative would be to create a function like I did in the test above and then test that, then either replicate the logic in the sql statement, or create a new sql function in sqlite.rs that calls it

abdnh:
Isn't `due <= today` guaranteed by due_cards.sql, so the top negative positions can't occur? The other results look fine to me, and in fact, the query here expectedly returns almost identical results to the filtered deck query according to my tests with this deck:
[relative-overdue-test.zip](https://github.com/ankitects/anki/files/8368415/relative-overdue-test.zip)

I'll look into writing some tests.

dae:
You're right, thinko on my part :man_facepalming:

dae:
Thanks Abdo! Great to have a test with it. Will merge this in once 2.1.50 is out the door.

PR #1759, opened 2022-03-30
Implicitly group when joining searches

RumovZ:
Now what about redundant grouping? I can think of a few ways to deal with it:

1. Add separate `.and_flat()`, `.or_flat()`.
2. Add mehod to `.flatten()`.
3. Skip redundant grouping (e.g. only group if there are both `And`s and `Or`s.)
4. (((((Ignore it,) because) it's) syntactically) the) same.


dae:
(associated discussion for future reference: https://github.com/ankitects/anki/pull/1754)

Thanks Rumo, IMHO this feels quite intuitive now, and the reduction in boilerplate is a nice bonus!

I've gone through all the .and() / .or() calls in the codebase, and it looks like 4) would fine for those - the few extra brackets on the left side of expressions shouldn't do any harm (and I didn't spot any changes in meaning in the existing code), and they add clarity to the output. The only thing I was worried about is adding hundreds of them, but the custom study tag search uses any/all() instead for that, so that should be fine. For our own code, the explicit/safe option feels best.

For the end user creating searches in the browse screen using the shortcut keys though, it does represent a change in behaviour, and I imagine some users will not be fond of the extra brackets being inserted there. Maybe we should retain the old behaviour there? 1) feels like the simplest way to do that. WDYT?

RumovZ:
Oh, I thought users would only notice in custom study filters etc., but of course, it's also used in the browser. Agreed, 1) is the only acceptable option there.

dae:
It would be a bit neater - do those changes look ok to you?

RumovZ:
Perfect!

PR #1760, opened 2022-03-30
Also restore/keep position of new cards

RumovZ:
https://github.com/ankitects/anki/pull/1743#discussion_r838397834

### DIFF
--- a/rslib/src/scheduler/new.rs
+++ b/rslib/src/scheduler/new.rs
@@ -18,9 +18,29 @@ use crate::{
 };
 
 impl Card {
-    fn schedule_as_new(&mut self, position: u32, reset_counts: bool) {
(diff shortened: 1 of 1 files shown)


rslib/src/scheduler/new.rs
```diff
@@ -18,9 +18,31 @@ use crate::{
 };
 
 impl Card {
-    fn schedule_as_new(&mut self, position: u32, reset_counts: bool) {
+    pub(crate) fn original_or_current_due(&self) -> i32 {
+        self.is_filtered()
+            .then(|| self.original_due)
```
dae, line 23: Neat, I wasn't aware of .then()

dae:
Thanks Rumo!

PR #1766, opened 2022-04-02
Speed up editor by lazily loading CodeMirror

hgiesel:
When populating the fields with content, CodeMirror is preloaded even if the PlainTextInput isn't open. This populating takes up a good chunk of the entire loading time, and it is noticeable in the Browser, when there are a lot of fields. On a particular notetype with 20 fields, this shortened the time to interactive from ~750ms to ~250ms.


dae:
That's a great improvement - thanks Henrik!

PR #1769, opened 2022-04-03
Template order

RumovZ:
https://github.com/ankitects/anki/commit/d55f0807330805bfb036c6014b7135e4d7fdd072#r70149815

I didn't change the string _names_ containing 'template'. Was that correct?
Also, I feel like we should emphasise that this about the _number_ of card types. Not sure if I expressed this well.

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -122,7 +122,7 @@ deck-config-new-gather-priority-random-notes = Random notes
 deck-config-new-gather-priority-random-cards = Random cards
 deck-config-new-card-sort-order = New card sort order
 deck-config-new-card-sort-order-tooltip-2 =
-    `Card type`: Displays cards in order of card type ordinals. If you have sibling burying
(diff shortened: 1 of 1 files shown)


dae:
Ordinal is not a term that comes up much in everyday speech, and it's not a term we expose to the user in the interface, so I think it might be a bit confusing. I gather you don't like 'in card type order', but the comment about showing all front->back cards first is intended to clarify that. Does moving it closer help at all, eg the following?

'in card type order (eg front->back/cloze 1 cards before back->front/cloze 2 cards)'

Otherwise looks good, and I think it's probably better not to break translations for these wording tweaks :+1:

RumovZ:
What about number instead of ordinal? If you think that wouldn't work either, I'll just revert it to "card type order". As you said, there already is a clarifying example, and I don't feel like expanding that would improve things.

dae:
Number works for me :-)

dae:
Thanks Rumo!

PR #1779, opened 2022-04-07
Add some component comment 

hgiesel:
I remember at some points, where we've added comments in the script section, but this is how you actually add a comment to a component, so that it shows on hover in VSCode (or other editors).

I think this kind of uneasiness about how to exactly write comments for Svelte components prevented me from writing some in the past, so I opened this PR.


hgiesel:
Hm, what's not ideal is that comment needs to be above the component template, which means in a lot of cases it will be in the middle of the file, rather than on the top. WDYT?

dae:
It looks like putting it at the top works too:

```html
<!--
Copyright: Ankitects Pty Ltd and contributors

@component
Testing
-->
<script context="module" lang="ts">
```

A separate HTML comment for the copyright header would be nicer, so that it doesn't get included in the docstring, but unfortunately prettier-plugin-svelte seems to get confused when two comments come one after the other.

hgiesel:
However in that VSCode includes the copyright notification in the hover info for me.

dae:
Yep :-)

> so that it doesn't get included in the docstring

The question is, what is the lesser of two evils? An unwanted copyright statement on hover, or module documentation in the middle of the file? 

I don't have strong feelings about it. If you prefer the latter, the comment at the top could tell people to look for the @component comment below.

dae:
I'll just merge this in as-is for now.

PR #1780, opened 2022-04-08
Fix a number of bugs with add-on conflict resolution

glutanimate:
Looking into migrating Review Heatmap users back to AnkiWeb, I stumbled on a number of bugs in the add-on conflict resolution system that could lead to add-ons being disabled when they shouldn't be. In one scenario in particular, I found that updating via AnkiWeb could lead to two conflicting add-ons both being disabled without notice. This could be pretty bad, given that the path to recovery likely won't be immediately obvious to users.

The changes in this PR address three distinct scenarios focusing on different release channels of the same add-on, but they should also extend to other situations where multiple add-ons are involved. Simply put, they try to ensure that:

- Whenever an add-on is installed via user interaction (be it from a local file or from AnkiWeb), it always ends up enabled.
- Whenever add-ons are updated via AnkiWeb, existing active/disabled settings are followed fully, with add-ons not being disabled unexpectedly.

----

To make it easier to illustrate the problem and replicate the aforementioned scenarios, I've attached a couple of dummy add-ons and also uploaded a sample listing to AnkiWeb (for the local builds, make sure to rename .zip to .ankiaddon):

- [review-heatmap-ankiweb.zip](https://github.com/ankitects/anki/files/8448081/review-heatmap-ankiweb.zip)
- [review-heatmap-github.zip](https://github.com/ankitects/anki/files/8448082/review-heatmap-github.zip)
- [AnkiWeb ID 1108829](https://ankiweb.net/shared/info/1108829)

###  Scenario 1: Installing conflicting add-ons from local filesystem

Steps to reproduce:

1. Install review-heatmap-ankiweb.ankiaddon
2. Install review-heatmap-github.ankiaddon
3. Install review-heatmap-ankiweb.ankiaddon

**Expected**: review-heatmap-ankiweb enabled, review-heatmap-github disabled

**Actual**: Both disabled

### Scenario 2: Migrating from local build of add-on to previously disabled AnkiWeb build

Steps to reproduce:

1. Install 1108829
2. Install review-heatmap-github.ankiaddon
3. Install 1108829

**Expected**: 1108829 enabled, review-heatmap-github disabled

**Actual**: Both disabled

### Scenario 3: Receiving updates for previously disabled add-ons when a conflicting add-on is installed and enabled

1. Install 1108829
2. Install review-heatmap-github.ankiaddon
3. Simulate add-on update by setting the mod date for 1109929 to `0` in its meta.json, then checking for updates.

**Expected**: 1108829 remains disabled, review-heatmap-github remains active

**Actual**: Both disabled

-----

I realize that these changes are coming in at the very tail-end of the 2.1.50 test period, but if we could someone still squeeze them in, I'd very much appreciate it. They just eliminate a lot of potential headaches that migrating users over to AnkiWeb could otherwise bring (but are of course also important for any other scenarios where conflict handling is key, e.g. different stable/beta add-on release tracks, or plain conflicts between different add-ons).


glutanimate:
(hit shift+return by accident :sweat_smile: , description above is still WIP)

dae:
With the release planned for tomorrow, I'm afraid you've left this a bit too late. 2.1.51 will likely follow within a few weeks, so these changes won't need to wait long.

glutanimate:
Totally understandable. Thanks for the quick merge (and follow-up – not sure how I missed that)!

PR #1781, opened 2022-04-08
Add minimumContentsLength property to templatesBox

### LINKED ISSUE #1777: Card layout window size depends on card template names
https://forums.ankiweb.net/t/anki-2-1-50-beta-6-release-candidate/18181/122

roxgib:
Fixes #1777 by adding the `minimumContentsLength` property to the comboBox, which causes the size policy to not require expanding beyond that amount while still expanding if the user stretches the window manually.

Note that users who were previously experiencing large window widths due to the previous behaviour might see the same size again due the geometry being restored.

### DIFF
--- a/qt/aqt/forms/clayout_top.ui
+++ b/qt/aqt/forms/clayout_top.ui
@@ -49,6 +49,9 @@
          <verstretch>0</verstretch>
         </sizepolicy>
        </property>
+       <property name="minimumContentsLength">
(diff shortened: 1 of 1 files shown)


roxgib:
Sorry for the force pushes, apparently I'm extra prone to typos today

dae:
Great, thanks for the quick fix!

PR #1798, opened 2022-04-13
Fix non-character keys resetting text style

### LINKED ISSUE #1795: Pressing shift cancels pending bold status
Appears to have been introduced in either 

52dc889e7fe35a950b8f3f94e9f5c2824830039a or 4117fbea6c9d0457fe24ae75d601eb99785d2a7a (the former fails to build).

https://forums.ankiweb.net/t/anki-2-1-51-release-candidate/18942/5

If this can be fixed easily/without risk of breaking other things, it would be nice to have for 2.1.51

Vycton:
Issue #1795 seems to have been caused in 0d83581ab08e1a92c5c55c99965e9f054b3f43e6 
The `clearInsertText()` call from line 68 of `ts/sveltelib/input-handler.ts` seems to remove any bold/italics/underlined etc. modifiers.
I am not sure what the intended behavior was, since the original commit seems to have to do with Mathjax which I know little about. However, the comment on line 66 seems to indicate that `!keyboardEventIsPrintableKey(event)` was intended to return true if any arrow key was pressed, when it in fact returns true when any key is pressed that does not have a 1 character name.

I have simply replaced the function call with a check if the key is an arrow key, which should hopefully retain the originally intended behavior as per the comment, while fixing the unintended behavior while editing regularly. The `keyboardEventIsPrintableKey` function is no longer used, so I also removed that.

Closes #1795.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -99,6 +99,7 @@ Sachin Govind <sachin.govind.too@gmail.com>
 Bruce Harris <github.com/bruceharris>
 Patric Cunha <patricc@agap2.pt>
 Brayan Oliveira <github.com/BrayanDSO>
+Vincent van Aalten <vincent.aalten@hotmail.com>
 
 ********************
 
--- a/ts/lib/keys.ts
+++ b/ts/lib/keys.ts
@@ -23,10 +23,6 @@ export function checkIfInputKey(event: KeyboardEvent): boolean {
     return event.location === GENERAL_KEY || event.location === NUMPAD_KEY;
 }
 
-export function keyboardEventIsPrintableKey(event: KeyboardEvent): boolean {
-    return event.key.length === 1;
-}
-
 export const checkModifiers =
     (required: Modifier[], optional: Modifier[] = []) =>
     (event: KeyboardEvent): boolean => {
--- a/ts/sveltelib/input-handler.ts
+++ b/ts/sveltelib/input-handler.ts
@@ -63,8 +63,7 @@ function useInputHandler(): [InputHandlerAPI, SetupInputHandlerAction] {
 
     function onKeydown(event: KeyboardEvent): void {
         /* using arrow keys should cancel */
-	
-    if (["ArrowUp", "ArrowLeft", "ArrowDown", "ArrowRight"].includes(event.key)){
+        if (["ArrowUp", "ArrowLeft", "ArrowDown", "ArrowRight"].includes(event.key)) {
             clearInsertText();
         }
     }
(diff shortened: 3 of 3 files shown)


Vycton:
Sorry for the messy commits... tests are a pre-commit hook now :)

hgiesel:
Thanks for looking into this!

This fixes the Shift issue, however the behavior is still a bit weird:
E.g. you can click Ctrl-b, to have bold turned on, and then press Ctrl-a, and still have the bold command turned on, which means that typing now creates bold text, even though you changed the selection quite a bit.

It might be better to restrict the call to `clearInsertText()` a bit further.

```typescript
if (checkIfInputKey(event) && !keyboardEventIsPrintableKey(event)) {
    clearInsertText();
}
```

`checkIfInputKey` is defined in `ts/lib/keys`.
We might also need special treatment for dead characters, but that should go in the right direction.

Vycton:
I'm not quite sure how text editors are generally expected to behave in this regard. Do you mean that if I type 'abc', then hit Ctrl-b and Ctrl-a, then type 'efg', the 'efg' should not be bold? For reference I tried this in LibreOffice and the 'efg' came out bold, so the behavior aligns with that at least. 

As for how much the call to `clearInsertText()` is restricted: are there keys other than the arrow keys which should trigger the call? The comment above it seems to suggest explicitly restricting it to just the arrow keys should be correct.

hgiesel:
Typically we either use either text editors as a model (convention), or the improvement in efficiency warrants it, we'll deviate from it (convenience).

Typically I'll use the CKEditor as a model on how it should behave, as it's also based on contentEditable, just like the Anki editor.

What you said might be true, however there's also other cases:
1. Press Ctrl+B and then create a cloze via Ctrl+Shift+C: You'll type bold inside of the cloze.
2. On macOS, you can use Ctrl+f and Ctrl+b to simulate the arrow keys, which also won't trigger the resetting in this case.

> The comment above it seems to suggest explicitly restricting

Comment can be a bit misleading sometimes. Source: I wrote that comment :)
Probably when I wrote that code (+ that comment) I didn't think about it too much.

dae:
Thanks for looking into this Vincent! @hgiesel knows this part of the code best, so happy to merge this in once he's happy with it.

Vycton:
I see. Your comment made me believe it was an easy fix :)
I will familiarize myself with the code some more and see if I can get it to better match the behavior of CKEditor.

dae:
I may have misinterpreted what Henrik was suggesting, but my reading of the above was that the code he provided above would solve the immediate problem, with room for improvement in a future PR. But perhaps Henrik can clarify :-)

Vycton:
Sorry, I should have mentioned that I tried Henrik's suggestion and it didn't work for me. On my system every key I tried would cause `checkIfInputKey` to return true, so the unintended behavior remained.

dae:
Sorry this ended up being a bit involved Vincent, but I appreciate the contribution in any case.

Vycton:
No worries! Sometimes a simple fix turns out to be more complicated. I will continue to try and get a grasp of the code and see if I can help out somewhere down the line.

PR #1800, opened 2022-04-13
Store last position when scheduling as review

RumovZ:
I think you mentioned this case at the very beginning, but somehow it slipped through!


dae:
Thanks Rumo!

PR #1802, opened 2022-04-14
Load page script after body

dae:
document.body must exist by the time the top level Svelte component
initializes.

https://forums.ankiweb.net/t/2-1-50-editor-wont-show-when-addons-load-many-js-files/19036

@hgiesel look ok to you?

### DIFF
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -499,7 +499,7 @@ html {{ {font} }}
 
         from aqt import mw
 
-        head = mw.baseHTML() + csstxt + jstxt + web_content.head
+        head = mw.baseHTML() + csstxt + web_content.head
         body_class = theme_manager.body_class()
 
         if theme_manager.night_mode:
@@ -520,7 +520,9 @@ html {{ {font} }}
 {head}
 </head>
 
-<body class="{body_class}">{web_content.body}</body>
+<body class="{body_class}">
+{jstxt}
+{web_content.body}</body>
 </html>"""
         # print(html)
         self.setHtml(html)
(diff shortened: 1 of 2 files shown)


BlueGreenMagick:
Can the <script> be after <body>? [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/html) says the <html> tag may only have a <head> and a <body>

By the way [svelte template](https://github.com/sveltejs/template/blob/master/public/index.html) uses a defer attribute to execute script after the document is parsed.

dae:
Ok, there are multiple problems here :-)

- cdc60ef didn't actually fix the problem, since the editor does not use page.html; it uses stdHtml()
- as you rightly pointed out, script after the body is not technically valid

There are a few issues with switching to defer however:

- the `js` arg to stdHtml() currently does not provide a way for scripts to elect to be deferred. We can work around this by putting the script tag in the `head` argument instead, but it then runs after add-on provided JS. This will break add-ons like the new format pack that assume svelte/internal will be available when they run
- even if we were to rework the code so that the editor js is inserted before add-ons, if it is made defer and the addons are not defer, the add-on js may still run first and break.

I'm not sure what the best way to solve this is. If we made all scripts passed in via `js` as defer, then it should solve the new format pack case, but could potentially break other cases where something has injected inline script tags into the document, which will run before the deferred scripts.

Anyone have any ideas on our least bad option here?

BlueGreenMagick:
So to summarize, this is the order we want things to run, right?
1. body html
2. svelte script
3. addon js script
4. inline body script

Then what if we put the script at the start of body? As in:

```
<body>
  <script svelte>
  <script addon> 
  <inner body elements>
</body>
```

dae:
I didn't expect that to work, as I thought Svelte would overwrite the children of `body`, removing the script tags. It doesn't seem to though - b7042ad seems to work. Can anyone see any problems with this approach?

dae:
Let's give this a try in rc2 then.

dae:
Unfortunately I forgot to pull this in before building rc2, so it will probably need to wait for .52 now.

PR #1808, opened 2022-04-18
chore: update Bazel version

gregorias:
This update addresses https://github.com/bazelbuild/rules_rust/issues/226, which was causing Anki builds to fail on macOS.


dae:
I'll give this a test after 2.1.51 is out

Issue #1114, opened 2021-04-09
Exporting stats to PDF produces white on white text if Anki is in night mode.


Filius-Patris:
## Steps to reproduce
1. Make sure Anki is in night mode (Preferences > Basic > Night mode; Restart if needed)
2. Go to Stats, "Save PDF" (very bottom)
3. Inspect the PDF. (Using LibreOffice Draw, for example)

## Expected behaviour
The text renders black on white (like in light mode).

## Observed behaviour
The text renders white on white.

dae:
Looks fine to me. Did you test with the latest version and without add-ons?

<img width="1100" alt="Screen Shot 2021-04-17 at 10 01 42 am" src="https://user-images.githubusercontent.com/180542/115095260-fc652d80-9f63-11eb-820f-19a18d75e550.png">

Filius-Patris:
Without addons: yes, latest version: latest stable yes. You mean master?

Filius-Patris:
Hmm weid. I tried again this morning and now it worked.

Sorry for the inconvenience.

lukawarren:
Sorry to bump a year-old issue but I'm having this issue too, but only when I begin to scroll down.

In other words if I open stats with the dark theme and immediately click "Save PDF" it works just fine, but if I first scroll down so that the top is out of view and _then_ click save, then the top of the PDF turns white. The more I scroll down, the more of the final PDF turns white.

Maybe that is why it was hard to reproduce?

(I'm on version ⁨2.1.49 - dc80804a)

dae:
Thanks, that does seem to trigger the issue here too. It's not totally white, but the background is white, making it hard to see.

dae:
Fixed in #1810

RotyBR:
Is it possible to export stats to pdf on mobile? If so, where do I find it? Please, help me!

dae:
If you're asking about AnkiMobile, it does not presently have such a feature, but I'll keep an eye on demand.

PR #1810, opened 2022-04-18
Fix issue #1114 by scrolling the user to the top of the page

lukawarren:
I toyed with changing the CSS of the page instead, but I wasn't sure if that would have any adverse consequences on anything else. For now I believe it's better than nothing.

From the first commit's description:
`It's obviously a bit of a "hacky" solution, since it's slightly jarring for users to scroll down, click export, then find themselves all the way at the top again, but it's somewhat less confusing than wondering why your PDF is broken :-)`


dae:
Thanks Luka, it seems like a reasonable solution for now.

PR #1811, opened 2022-04-18
Fix resetting of formatting modes when pressing modifer or dead keys

hgiesel:
Closes #1798 and #1795.

I gave this a go, seeing as it's tagged for 2.1.51.


dae:
Great, thanks for looking into this Henrik. I've moved the remaining editor issues to 2.1.52, and will try to get 2.1.51rc2 out today or tomorrow.

PR #1813, opened 2022-04-20
Fix pasting of local links to unsupported types that contain spaces

abdnh:
This change causes local links to unsupported formats like PDF to be
handled by _processUrls() instead of _processText(), converting them to
correct hyperlinks in case they contain spaces.

Before:

https://user-images.githubusercontent.com/41397710/164309575-1a75fcf9-563d-4699-86ef-34e98b435068.mp4

After:

https://user-images.githubusercontent.com/41397710/164309599-5424a7c4-dd55-47eb-abdd-b59b6a559e7e.mp4


dae:
Seems reasonable. Thanks Abdo!

PR #1815, opened 2022-04-21
Remove night mode rule for typeans

kleinerpirat:
https://forums.ankiweb.net/t/shaddow-below-cloze-type-area-in-night-mode/19287/2


dae:
I think the original goal was to adjust the input box (https://forums.ankiweb.net/t/night-mode-does-not-affect-input-tags/10116), but we no longer need that due to https://github.com/ankitects/anki/commit/c3b5902da04cf9e75777995431132505f5884794. Thanks Matthias!

PR #1817, opened 2022-04-21
Fix progress of chained operations

RumovZ:
The new behaviour is more in line with `CollectionOp` which uses `mw.taskman.with_progress()`, which also calls `.finish()` _before_ the callback.

Motivation
---
4515c41d2cd2196e8624d5dd2f331f15228 broke aborting colpkg imports.
QueryOp starts a new progress, which is only finished after its success callback is run. But in `replace_with_apkg` we want to start a new progress inside that callback. However, `mw.progress.start()` returns `None` because of the already running progress.
It took me a while to figure this out due to the `except AttributeError: pass`. I had copied this from somewhere. Is it supposed to handle this very `dialog == None` case?

Somewhat related
---
Should I subclass `CollectionOp` with an `InteractiveCollectionOp` that wraps the progress handling from `replace_with_apkg()`? I need that same logic for apkg importing in #1743.

### DIFF
--- a/qt/aqt/importing.py
+++ b/qt/aqt/importing.py
@@ -14,7 +14,7 @@ import aqt.modelchooser
 from anki.errors import Interrupted
 from anki.importing.anki2 import V2ImportIntoV1
 from anki.importing.apkg import AnkiPackageImporter
-from aqt import AnkiQt, gui_hooks
+from aqt.main import AnkiQt, gui_hooks
 from aqt.operations import QueryOp
 from aqt.qt import *
 from aqt.utils import (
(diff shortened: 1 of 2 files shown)


dae:
Thanks Rumo, change looks good.

It looks like you copied the except line from mediacheck.py. I don't remember the rationale for the current approach, but dialog==None looks like the likely explanation. I thought it might be connected to the immediate=False case where the window hasn't been shown yet (eg what mw.progress.want_cancel() checks for), but that shouldn't matter if .start() returned the dialog.

The progress code is a bit of a mess unfortunately. If we had access to async/await and asynchronous context managers it would probably make things simpler, but unfortunately the asyncio/Qt integrations don't look very mature/well supported.

> Should I subclass CollectionOp with an InteractiveCollectionOp that wraps the progress handling from replace_with_apkg()

Sounds good. For the name though, how about CollectionOpWithProgress?

RumovZ:
> Sounds good. For the name though, how about CollectionOpWithProgress?

Since the current classes already offer a progress window, I wanted to emphasise its interactive nature, i.e. the automatic queries for label updates, and the propogation of the cancel signal. But I don't mind either way.

dae:
Good point. CancellableCollectionOp? CollectionOpWithBackendProgress? ProgressMonitoringCollectionOp? :-) Not in love with any of them and open to other ideas. "Interactive" just feels a bit strange to me, as you can't really interact with it other than cancelling it.

RumovZ:
Yes, I see what you mean. I think I'll go with CollectionOpWithBackendProgress, because it makes clear that this only works for certain ops, for which progress updates are explicitly supported by the backend.

PR #1818, opened 2022-04-24
Instead of transforming to {\lt}, transform to &lt;

hgiesel:
In Mathjax editor. This way you can also use Mathjax convenience shortcuts like <=> in chemistry mode like so: `\ce{<=>}`

Mentioned here: https://forums.ankiweb.net/t/on-anki-2-1-50-arrows-are-not-properly-rendered-when-writing-chemical-reactions/19341/6

### DIFF
--- a/ts/editable/mathjax-element.ts
+++ b/ts/editable/mathjax-element.ts
@@ -20,7 +20,7 @@ const mathjaxInlineDelimiterPattern = /\\\((.*?)\\\)/gsu;
  * be first translated to entities.
  */
 function translateEntitiesToMathjax(value: string) {
-    return value.replace(/&lt;/g, "{\\lt}").replace(/&gt;/g, "{\\gt}");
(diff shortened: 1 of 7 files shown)


ts/editable/mathjax-element.ts
```diff
@@ -20,7 +20,7 @@ const mathjaxInlineDelimiterPattern = /\\\((.*?)\\\)/gsu;
  * be first translated to entities.
  */
 function translateEntitiesToMathjax(value: string) {
-    return value.replace(/&lt;/g, "{\\lt}").replace(/&gt;/g, "{\\gt}");
+    return value; // value.replace(/&lt;/g, "{\\lt}").replace(/&gt;/g, "{\\gt}");
```
dae, line 23: Since this function now appears to be a no-op, would it be worth removing? I'll do that now on the assumption the answer is yes (so I can get a new beta build out today), but if my assumption is wrong please feel free to add it back :-)

dae (commented):
Thanks Henrik, these fixes are much appreciated.

PR #1819, opened 2022-04-24
Remove onInput call in input handler

### LINKED ISSUE #1805: backspace on a div deletes multiple images
https://forums.ankiweb.net/t/2-1-50-deleting-images-in-editor/19095/

If this can be easily fixed for .51 that would be great, but we can push it back if necessary.

hgiesel:
Fixes #1805 

This was implemented by me to prevent seemingly empty fields,
which however still contain some HTML (e.g. `<div></div>`) and
thus will trigger in the templates.
However it only worked in very limited cases, maybe we could
solve this better by indicating whether a field has content somehow.

### DIFF
--- a/ts/sveltelib/input-handler.ts
+++ b/ts/sveltelib/input-handler.ts
@@ -61,21 +61,8 @@ function useInputHandler(): [InputHandlerAPI, SetupInputHandlerAction] {
         insertText.clear();
     }
 
-    function onInput(this: HTMLElement, event: InputEvent): void {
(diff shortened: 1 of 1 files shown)


dae:
If we wanted to retain the current behaviour, perhaps we could restrict it to only applying when the div has a single child node, which is an empty text node? I guess it depends on how common other cases are - can you think of another common way the field ends up "empty but not empty" aside from an empty div?

Fixing the image loss is the priority in any case - thank you digging into that :-)

PR #1821, opened 2022-04-26
Extends the SearchNode interface to support field and unqualified searches

abdnh:
Context: https://forums.ankiweb.net/t/is-it-possible-to-construct-field-foo-searches-using-searchnode/19365

This adds two new options to SearchNode to support field searches and unqualified searches with automatic escaping.

Maybe the `text` option can use a more descriptive name (maybe `literal_text`?). I'm open to suggestions.

Can we somehow attach a docstring to SearchNode (in case there is any need for that)?

### DIFF
--- a/rslib/src/backend/search/search_node.rs
+++ b/rslib/src/backend/search/search_node.rs
@@ -9,7 +9,7 @@ use crate::{
     search::{
         parse_search, Negated, Node, PropertyKind, RatingKind, SearchNode, StateKind, TemplateKind,
     },
-    text::escape_anki_wildcards_for_search_node,
+    text::{escape_anki_wildcards, escape_anki_wildcards_for_search_node},
 };
 
 impl TryFrom<pb::SearchNode> for Node {
@@ -98,6 +98,11 @@ impl TryFrom<pb::SearchNode> for Node {
                         Node::Group(nodes)
                     }
                 }
+                Filter::Field(field) => Node::Search(SearchNode::SingleField {
+                    field: escape_anki_wildcards_for_search_node(&field.field_name),
+                    text: escape_anki_wildcards(&field.text),
+                    is_re: field.is_re,
+                }),
             }
         } else {
             Node::Search(SearchNode::WholeCollection)
--- a/rslib/src/backend/search/search_node.rs
+++ b/rslib/src/backend/search/search_node.rs
@@ -103,7 +103,7 @@ impl TryFrom<pb::SearchNode> for Node {
                     text: escape_anki_wildcards(&field.text),
                     is_re: field.is_re,
                 }),
-                Filter::Text(text) => {
+                Filter::LiteralText(text) => {
                     let text = escape_anki_wildcards(&text);
                     Node::Search(SearchNode::UnqualifiedText(text))
                 }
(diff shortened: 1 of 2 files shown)


abdnh:
Completely unrelated rant: For anyone on building on Windows with PowerShell and getting those pesky [LNK1104](https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-error-lnk1104?view=msvc-170) linker errors, you need to have a `windows.bazelrc` file in the root dir with the following line:
```
startup --output_user_root=\\bazel\\anki
```

`setup-env.bat` already writes that file, but I adapted a similar `setup-env.ps1` file because the former doesn't completely work for PowerShell, and the thing is, `echo` in PowerShell apparently separates its arguments by newlines, so somthing like `echo startup --output_user_root=\\bazel\\anki > windows.bazelrc` will result in a file with the following contents:

```
startup
--output_user_root=\\bazel\\anki 
```

Which doesn't work, and you're left to wonder what is the cause of those annoying errors.

dae:
@RumovZ uses PowerShell as well IIRC. I'd prefer not to remove the existing .bat files, but I'm happy to accept PRs that add PS scripts when they're needed.

If you add `// something` above an enum variant in the .proto, it'll be visible when hoving over it in .py files, eg n.text:

```
n = SearchNode(text="foo")
n.text
```

You can also add docstrings to particular messages.

`literal_text` works for me - it better communicates that wildcards are ignored.

dae:
@RumovZ uses PowerShell as well IIRC. I'd prefer not to remove the existing .bat files, but I'm happy to accept PRs that add PS scripts when they're needed.

If you add `// something` above an enum variant in the .proto, it'll be visible when hoving over it in .py files, eg n.text:

```
n = SearchNode(text="foo")
n.text
```

You can also add docstrings to particular messages.

`literal_text` works for me - it better communicates that wildcards are ignored.

RumovZ:
Might callers be surprised that passing `literal_text="foo bar"` is not the same as typing `foo bar` into the brower? If so, maybe `literal_phrase` would better communicate the implicit quoting.

I've never encountered such a linker error. And since I did [this](https://github.com/ankitects/anki/blob/main/docs/windows.md#more), I never had any reason to run setup-env.bat directly.

rslib/src/backend/search/search_node.rs
```diff
@@ -98,6 +98,15 @@ impl TryFrom<pb::SearchNode> for Node {
                         Node::Group(nodes)
                     }
                 }
+                Filter::Field(field) => Node::Search(SearchNode::SingleField {
+                    field: escape_anki_wildcards_for_search_node(&field.field_name),
+                    text: escape_anki_wildcards(&field.text),
```
RumovZ, line 103: You could argue for any of these two functions, but maybe it would be more intuitive if the same was used in both cases?

RumovZ (approved):

rslib/src/backend/search/search_node.rs
dae: Is there a particular use case you had in mind for _for_search_node() @abdnh? If not, maybe the wildcards() option would be a better choice?

dae:
I thought that is what "literal" is conveying :-) But either one works for me.

abdnh:
`literal_text` works for me too. Maybe it's less likely to be misunderstood with the presence of `parsable_text` (which I once used incorrectly, thinking that it treats its input literally).

abdnh:
> I've never encountered such a linker error. And since I did [this](https://github.com/ankitects/anki/blob/main/docs/windows.md#more), I never had any reason to run setup-env.bat directly.

Will start using that script. Thanks!

dae:
Thank you both!

PR #1833, opened 2022-05-02
Fix scrollbars showing on bottom bar

kleinerpirat:
This is a quick fix for https://forums.ankiweb.net/t/anki-2-1-52-release-candidate/19412/2

It does change the spacing of the buttons by a few pixels, but I don't think it looks worse. Open to more conservative approaches of course.


dae:
Doesn't look too out of place in either qt6 or qt5 here. Thanks Matthias :-)

PR #1836, opened 2022-05-05
Fix some apkg import issues

### LINKED ISSUE #1835: legacy apkgs with subdeck root fail to import
To repro:

- create a card and put it in a foo::bar deck, then export that deck using the legacy exporter
- attempt to import it using the new importer, and be told you should run "check database" :-)

Because this is not such an uncommon case, we should ideally handle this instead of throwing an error. Maybe a low-effort solution would be, when this error is encountered, to run the part of the DB check that adds missing parents. I think you mentioned something like this in the past Rumo.

RumovZ:
Closes #1835. That was an unintended side effect of using the gather code for both im- and export.
One more issue I realised: Unexported siblings don't get generated on import. I'll deal with that after plaintext imports, because it won't be easy to add this, but it needs to be taken care of when importing text at the very least.

### DIFF
--- a/rslib/src/import_export/gather.rs
+++ b/rslib/src/import_export/gather.rs
@@ -187,12 +187,10 @@ impl Collection {
     ) -> Result<()> {
         if let Some(parent_name) = immediate_parent_name(name) {
             if parent_names.insert(parent_name.to_owned()) {
-                parents.push(
-                    self.storage
-                        .get_deck_by_name(parent_name)?
-                        .ok_or(AnkiError::DatabaseCheckRequired)?,
-                );
-                self.add_parent_decks(parent_name, parent_names, parents)?;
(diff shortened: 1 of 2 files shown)


dae:
Thanks for the quick fixes :-)

Just out of curiosity on the generation issue:

1) I forget, are we ensuring this doesn't happen in the non-legacy export case already?
2) Presumably we could solve this (inefficiently) by just calling after_note_updates() at the end of the import, or generate_cards_for_existing_note()? Not super urgent though.

RumovZ:
1. We export all siblings of a note, so it shouldn't happen in well-formed collections, but existing issues could propogate.
2. Yes, would be nice to have a more efficient solution, at least for the happy path. This is complicated by the circumstance that the current import code is consuming. Since we need to generate after having imported cards, we would need to refetch notes and notetypes.

PR #1837, opened 2022-05-06
Fix for crash with invalid dates on Windows

dae:
https://forums.ankiweb.net/t/bug-report-crashing-when-opening-deck-browser/19768

Caused by a note mtime that was 1000x larger than it should have been.
Ideally we should show an appropriate message only on the failed column,
and fix this automatically in Check DB, but this change will at least
prevent the crashes.

https://docs.rs/chrono/0.4.19/x86_64-pc-windows-msvc/src/chrono/sys/windows.rs.html#128

### DIFF
--- a/rslib/src/dbcheck.rs
+++ b/rslib/src/dbcheck.rs
@@ -206,7 +206,8 @@ impl Collection {
         let nids_by_notetype = self.storage.all_note_ids_by_notetype()?;
         let norm = self.get_config_bool(BoolKey::NormalizeNoteText);
         let usn = self.usn()?;
-        let stamp = TimestampMillis::now();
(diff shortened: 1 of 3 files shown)


RumovZ (approved):

RumovZ:
> Ideally we should show an appropriate message only on the failed column,

It seems one way to achieve this would be to keep returning Strings from `time_string()` and `data_string()`, but make that an error message, if `Local.timestamp()` has failed. This would have the benefit of avoiding all these new Results.

dae:
Yep, that's a better idea - it also means the user can still access that browser row instead of it being completely disabled. Thanks :-)

PR #1838, opened 2022-05-06
Make TagAddButton focusable and simulate click on focus

### LINKED ISSUE #1832: Find a way to support tab navigation into/out of tags area?
Maybe we could accomplish it by marking part of the tag editor tabindex=0? Or providing a focusable proxy element that when receiving focus, applies focus to the correct location. Not sure how feasible it is though.

https://forums.ankiweb.net/t/tab-key-skips-tag-section-in-editor/19581

hgiesel:
Closes #1832


dae:
Thanks Henrik, this will make some people quite happy. It looks like we can 'tab' into the tags area and then 'tab' again to return to the first field, but 'shift+tab' to return to the previous field doesn't seem to work. Not a big deal (this is already better than it was before), but if it's easily fixed, that would be the icing on the cake. :-)

PR #1841, opened 2022-05-07
Fix formatting issues

RumovZ:
666c2e0f2469bdf55441101c24b3c7c15a29e2c3 is a bit controversial, and could be removed while keeping 84a68a6dd671c29d7449aa9f605572104943b81c without failing tests. Though personally, I don't find forcing multiline lists particularly helpful.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,6 +11,7 @@
     },
     "python.analysis.extraPaths": ["./pylib"],
     "python.formatting.provider": "black",
+    "python.formatting.blackArgs": ["--skip-magic-trailing-comma"],
     "python.linting.mypyEnabled": false,
     "python.analysis.diagnosticSeverityOverrides": {
         "reportMissingModuleSource": "none"
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,8 @@
 {
     "editor.formatOnSave": true,
+    "editor.codeActionsOnSave": {
+        "source.organizeImports": true
+    },
     "files.watcherExclude": {
         "**/.git/objects/**": true,
         "**/.git/subtree-cache/**": true,
--- a/pylib/anki/_backend/__init__.py
+++ b/pylib/anki/_backend/__init__.py
@@ -70,10 +70,7 @@ class RustBackend(RustBackendGenerated):
         if langs is None:
             langs = [anki.lang.current_lang]
 
-        init_msg = backend_pb2.BackendInit(
-            preferred_langs=langs,
-            server=server,
-        )
+        init_msg = backend_pb2.BackendInit(preferred_langs=langs, server=server)
         self._backend = rsbridge.open_backend(init_msg.SerializeToString(), log_file)
 
     def db_query(
@@ -116,19 +113,13 @@ class RustBackend(RustBackendGenerated):
         }
 
         input = i18n_pb2.TranslateStringRequest(
-            module_index=module_index,
-            message_index=message_index,
-            args=args,
+            module_index=module_index, message_index=message_index, args=args
         )
 
         output_bytes = self.translate_string_raw(input.SerializeToString())
         return anki.generic_pb2.String.FromString(output_bytes).val
 
-    def format_time_span(
-        self,
-        seconds: Any,
-        context: Any = 2,
-    ) -> str:
+    def format_time_span(self, seconds: Any, context: Any = 2) -> str:
         traceback.print_stack(file=sys.stdout)
         print(
             "please use col.format_timespan() instead of col.backend.format_time_span()"
--- a/pylib/anki/_legacy.py
+++ b/pylib/anki/_legacy.py
@@ -94,8 +94,7 @@ class DeprecatedNamesMixin:
 
     @classmethod
     def register_deprecated_attributes(
-        cls,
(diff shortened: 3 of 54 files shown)


dae:
Could you explain to me the issue fe1fcde is solving?

After the 2.1.52 release, the next few releases will probably involve cherry-picking changes from master onto a stable branch, so if we make large formatting changes at the moment, it's going to increase the chances of conflicts. Such a change would be better done closer to a large update (eg before 2.1.55beta1), when we're reasonably confident changes won't need to be backported anymore.

That said, I wonder if this approach is the best way forward anyway? My understanding of that black feature is that it was intended for the opposite purpose: by default an expression is not split into multiple lines, unless you add a trailing comma on
it. Isn't the issue here that vscode is adding a trailing comma without asking, which in turn results in black splitting the expression over multiple lines? Can we perhaps prevent that from happening instead?

RumovZ:
> Could you explain to me the issue https://github.com/ankitects/anki/commit/fe1fcde25da9e5637ded03b676b02ccafc371368 is solving?

If any imports in pylib are unordered, tests will fail, yet `bazel run //pylib:format` won't fix it, because `run_format.py` fails with a system exception, before isort can run. So you have to fix imports by hand.

> Such a change would be better done closer to a large update (eg before 2.1.55beta1), when we're reasonably confident changes won't need to be backported anymore.

Gotcha, I've opened #1848.

> by default an expression is not split into multiple lines, unless you add a trailing comma on it.

Or, of course, if the expression is too long. Then a trailing comma is added as well. But trailing commas are only ever *added* implicitly, not removed. So the formatter will never restore one-liners, even if the expression ends up being very short.

> Isn't the issue here that vscode is adding a trailing comma without asking, which in turn results in black splitting the expression over multiple lines?

Not sure what you mean by that. On save, black is invoked. What else is there to add a comma?

In my opinion, the diff shows a lot of poorly formatted code, and I assume there are basically two causes:

1. As mentioned above, expressions often change in size during development, but the formatter will only split them if they grow, not inline them if they shrink.
2. The programmer added a trailing comma.

The only disadvantage of the new settings I can see would be if programmers wanted to deliberately split short expressions across multiple lines, because they expect them to grow, and find it easier that way. Not sure how many people do this, I don't.
And even in that case, I assume this would mostly be used during active development, and the expression could be inlined, at the latest before it is merged into main.

dae:
> If any imports in pylib are unordered

Ah, so it was a fix for an issue introduced by the other commit. Got it :-)

> Not sure what you mean by that. On save, black is invoked. What else is there to add a comma?

Sorry, I confused myself by testing two separate examples in one file, one that was longer than the other, and I mistakenly attributed the difference to the two editors I tested in.

> because they expect them to grow, and find it easier that way.

That's the use case Black describe in their documentation.

I don't have strong feelings about this. There is an argument for keeping as close to the default code settings as possible, and I have gotten somewhat used to the vertical-heavy wrapping. In cases like the following the vertical style a bit easier to quickly locate a specific kw argument:

https://github.com/ankitects/anki/commit/84a68a6dd671c29d7449aa9f605572104943b81c#diff-49def8d6aa1c23a82430ecf37736688b8ebe70c31e5524efad00f3f0a3776cc3L1122

On the other hand, the presence/absence of the trailing comma is subtle, and as you fairly point out it may end up getting left around from refactoring. So not sure what the best course of action is there. Anyone else want to give their 2c? :-) @hgiesel @abdnh @hikaru-y @kleinerpirat @glutanimate @Arthur-Milchior and others

RumovZ:
Interested to hear what others have to say.

> Ah, so it was a fix for an issue introduced by the other commit. Got it :-)

Now I'm confused. What other commit? 🤔 You don't mean another one of this PR, do you? [fe1fcde](https://github.com/ankitects/anki/pull/1841/commits/fe1fcde25da9e5637ded03b676b02ccafc371368) was my main motivation for this PR, as it fixes a long existing issue on main.

dae:
:man_facepalming: 

I was confused because I've never run into this issue before. I suspect it's another Windows-only/lack-of-bazel-sandbox issue. Thank you for the fix, and sorry for the confusion there. :-)

RumovZ:
Oh, I didn't think it would be platform-specific. That explains the confusion, glad it's sorted out!

Arthur-Milchior:
No preference.
Intuitively, I believe both makes sense depending on the use case. If there are so many arguments, e.g. in a call to dict(), or any other part using **kwargs a lot, a trailing comma is nice, as it'll allow to add more elements without changing the line.
On more general function call, on a single line, I don't think they add value.
And it's a weak opinion in either case

It's not as if I were contributing to the desktop version recently.
However, I'd suggest adding in the first post a clarification of the question. Because it seems to be clear for Dae; but it took me minutes to understand what was this about.

kleinerpirat:
I find arguments split over multiple lines easier to read. Since I'm mainly using <kbd>Ctrl + F</kbd> to navigate files, I don't see a disadvantage in the extra space that's used.

But I'm also used to leaving such decisions completely up to the formatter and wouldn't think of deliberatly interfering with the preferred format. Also, I rarely do stuff on the Python side.. so no strong opinion here.

hgiesel:
I personally prefer more vertical wrapping + trailing commas. It's also gives you nicer diffs.

dae:

PR #1842, opened 2022-05-08
Add the av_player_will_play_tags hook

abdnh:
This is just a general version of the `reviewer_will_play_question_sounds` and `reviewer_will_play_answer_sounds` hooks that works on more screens.
I thought of removing the former hooks, but they are used by 4-5 add-ons now.


dae:
Seems reasonable - thanks Abdo.

Issue #1823, opened 2022-04-26
small tag editor suggestions


dae:
https://forums.ankiweb.net/t/on-anki-2-1-50-in-note-editor-you-cant-add-a-tag-after-typing-its-full-name-and-selecting-it-using-tab/19436

PR #1843, opened 2022-05-08
align tag icon and text

BlueGreenMagick:
Before:
image
After:
image

Related: https://github.com/ankitects/anki/issues/1823


dae:
Thanks for these fixes BGM!

PR #1847, opened 2022-05-09
select all tags with Ctrl+A

### LINKED ISSUE #1824: provide a way to select tags in the tag editor
https://forums.ankiweb.net/t/2-1-50-tts-bug-not-addressed-in-2-1-51/19307/7

BlueGreenMagick:
Pressing Ctrl + A while in tag input selects all tags if input is empty.

Fixes #1824


BlueGreenMagick:
Also, are there some other commands I should run other than `bazel run //ts:format`? It wouldn't fix eslint complaining about import order.

dae:
Thanks for all these fixes BGM. No objections from my end; do these two PRs work for you @hgiesel?

BlueGreenMagick:
Got it! Thanks!

Should it be added to https://github.com/ankitects/anki/blob/main/docs/development.md#fixing-formatting ?

PR #1850, opened 2022-05-09
Plaintext import/export

RumovZ:
As discussed, this implements rudimentary CSV and JSON (Mnemosyne) importing. I've also explored CSV metadata encoding, because it's tightly coupled with the import GUI.
Below is a list of questions/todos that will need an answer at some point, but for some it probably makes sense to collect more information first. I will keep updating this list.
I wonder what makes the most sense to implement next. The GUI? Exporting? Or polish up the existing routines first?

<hr>

General
---

-  [ ] It now seems to me that a unified import screen isn't really necessary: colpkg and JSON can probably do without one, and the requirements for apkg, and CSV are pretty different. So I guess we will need two separate ones.
-  [ ] Currently, the import routine is chosen by looking at the file ending. As we have at least two different plaintext formats now, this might be a bit too naive. I see two alternatives:
    1. Ask first what the user wants to import. Then the file picker window could filter out files without the expected ending, but also allow to show all files.
    2. Analyse the chosen file on the backend. The response could already contain the meta information required by the apkg, or CSV import screens.

JSON
---

-   [ ] Nesting over references? Currently, cards are nested inside notes, and notes reference their notetypes. The former is unambiguous (an id may refer to a source, or a target object), and 'parents' are guaranteed to exist by design. However, it would require a different approach for 'children' that are to be imported with an already existing parent, and isn't always possible (cards can't be nested in notes _and_ decks).
-   [ ] Insert and/or update. How can the user decide if a potentially existing object should be updated? Alternatively, they may want to insert a unique object, or skip the conflicting one. Options I see:
    1. Only allow a decision for the whole file.
    3. Accept an enum variant/integer for every object.
    4. Update objects with an id field, import everything else uniquely.
       For notes, I assume we also need to preserve the old behaviour that looks at the first field.
-   [ ] Validation checks. I guess we want to be as thorough as possible, which will require a lot of logic.
-   [ ] Error handling. How forgiving do we want to be? Should we use a reasonable default for an invalid value, skip the object, or abort the whole process?
-   [ ] Error messages. Depending on the two issues above, there may a be a lot of different errors. Then maybe we'll have to compromise, and use either unlocalised, or very generic messages ("invalid value 'x' for 'y'").
-   [ ] Skip serialising default values? Would allow for much smaller files, but maybe explicitness is to be preferred if JSON exports are mainly used for external modification.

CSV
---

-  [ ] Is there any use case for arbitrary delimiters? Aren't space, comma, semicolon, and tab enough?


RumovZ:
Hm, probably an issue with [ae940f1](https://github.com/ankitects/anki/pull/1850/commits/ae940f1faae261769a0f9b8869835ebaa1414aee) I need to look into.

dae:
Thanks Rumo, will give this all a good look over tomorrow. Test failure probably requires tweaks to /Cargo.toml such as compile_data_attr. If you don't have any luck after a few minutes, feel free to leave it to me, as it'll be easier to debug on a Linux/Mac machine.

RumovZ:
Would be grateful if you could take care of it. I wouldn't even know where to begin.

rslib/src/import_export/text/import.rs
```diff
@@ -0,0 +1,216 @@
...
+
+impl<'a> Context<'a> {
+    fn new(data: &ForeignData, col: &'a mut Collection) -> Result<Self> {
+        let usn = col.usn()?;
+        let normalize_notes = col.get_config_bool(BoolKey::NormalizeNoteText);
+        let mut notetypes = HashMap::new();
+        notetypes.insert(
+            String::new(),
+            col.notetype_for_string(&data.default_notetype)?,
+        );
```
dae, line 42: This zero-value-as-default approach is a bit subtle, and might be worth documenting briefly in the struct's field docstring.

rslib/src/import_export/text/import.rs
```diff
@@ -0,0 +1,216 @@
...
+            let did = self.col.deck_id_for_string(&note.deck)?;
+            self.deck_ids.insert(note.deck.clone(), did);
+            did
+        })
+    }
+
+    fn import_foreign_notes(&mut self, notes: Vec<ForeignNote>) -> Result<NoteLog> {
+        let mut log = NoteLog::default();
+        for foreign in notes {
+            if let Some(notetype) = self.notetype_for_note(&foreign)? {
```
dae, line 86: Warning or failing might be better than silently ignoring

rslib/src/import_export/text/import.rs
```diff
@@ -0,0 +1,216 @@
...
+        Ok(())
+    }
+
+    fn generate_missing_cards(
+        &mut self,
+        notetype: &Notetype,
+        deck_id: DeckId,
+        note: &Note,
+    ) -> Result<()> {
+        let card_gen_context = CardGenContext::new(notetype, Some(deck_id), self.usn);
```
dae, line 130: This parses all the templates in the notetype, so ideally we don't want to be doing this for every note.

dae (commented):
Looking great so far! A review of the code is below; will get back to your questions hopefully later today.

rslib/src/import_export/text/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+    tags: Vec<String>,
+    notetype: String,
+    deck: String,
+    cards: Vec<ForeignCard>,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Default, Serialize, Deserialize)]
+#[serde(default)]
+pub struct ForeignCard {
+    pub due: i32,
```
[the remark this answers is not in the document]
RumovZ: Yes, this needed some more thought. I assume `ForeignCard` may be a new card, too, in which case this is the new position? This case would be indicated by factor and interval being 0.

rslib/src/import_export/text/csv/metadata.rs
```diff
@@ -0,0 +1,302 @@
...
+    if let Some(delim) = delimiter {
+        metadata.delimiter = delim as u32;
+    } else if metadata.delimiter == 0 {
+        // XXX: should '#delimiter:[NUL]' be supported?
+        metadata.delimiter = delimiter_from_line(line) as u32;
+    }
+}
+
+fn delimter_from_value(value: &str) -> Option<u8> {
+    // FIXME: bytes like '\n', '#' and '"' will likely cause issues
```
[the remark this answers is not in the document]
RumovZ: Do you mean that the CSV is invalid, but we falsely detect a valid delimiter? There's probably no easy way around that, but we could offer a preview of the first few lines, like spreadsheet programs do.

dae:
I like the extra functionality you've added to the csv headers! Some users have expressed in the past the desire to be able to change things like deck or notetype partway through a file, so that cards/notes for more than one deck/notetype can be imported at once. I'm guessing that's going to be tricky though - presumably we can't easily get access to the comment lines, and there are questions about how mapping and deck/notetype selection would work. Maybe best left to the JSON format?

> It now seems to me that a unified import screen isn't really necessary

Agreed.

> Currently, the import routine is chosen by looking at the file ending. As we have at least two different plaintext formats now, this might be a bit too naive.

Since JSON importing is a new feature with no legacy cases to import, it's not unreasonable for us to require the text file to have an .anki-json or similar extension.

One minor usability issue I noticed while playing around with this code is that by default only colpkg/apkg files are selectable in the import screen, and you have to manually choose one of the other formats at the bottom to be able to import it. The old importer deliberately used * for text import and made it the default, specifically so that all available file types could be selected at once. An alternative approach would be for us to stop listing the various types separately, and just have a (generated) "All supported formats (*.colpkg *.apkg *.zip *.txt *.csv *.tsv *.db *.json)" or similar instead. This would prevent users from importing csv files with arbitrary names, but it is easy enough to rename a file, and means the importing screen won't show irrelevant stuff. There have been a fair few reports over the years from confused users who tried to import random files like .xls as a text file.

> Nesting over references?

Nesting of cards in notes is appealing for the reasons you mentioned. I think it's also conceptually a bit simpler and easy for people to follow than separate lists where objects in one list reference objects in the other. We do sort of have to do that for decks anyway, but we can treat their name as their id, which makes things a bit simpler.

If we go down this route, presumably we'd want to nest revlog entries under each card as well. 

If a card were reviewed in some external system and the card changes+revlog entry wanted to be imported, it would mean either all the note data needs to be reimported as well, or we need to have things like fields as optional, eg Note(note_id=xxx, fields=None, cards=[...]) is valid on the update case, but throws an error if a note is provided without fields that does not already exist.

> How can the user decide if a potentially existing object should be updated? Alternatively, they may want to insert a unique object, or skip the conflicting one. Options I see:
a) Only allow a decision for the whole file.
b) Accept an enum variant/integer for every object.
c) Update objects with an id field, import everything else uniquely.
d) For notes, I assume we also need to preserve the old behaviour that looks at the first field.

a) will probably cover the majority of cases. We could potentially have options like "only add", "always update", "update if newer" (which would require a timestamp to be provided).

In addition to that, optional fields like described above could potentially give us more flexibility. eg if factor is omitted or null, we use the defaults in the add case, or leave the existing value alone in the update case. Our code could avoid writing out the object if none of its properties changed. (ps: let's call it ease_factor).

For identifying specific objects, we could potentially support multiple approaches. Eg. a note_id used if available, and guid used if note_id not available but guid available. If neither are available, it means a new note needs to be added. If one is provided but does not match an existing object, we'd need to add, but use the provided id.

Cards could be identified either via card_id or by their index in the provided list.

WDYT?

> Validation checks. I guess we want to be as thorough as possible, which will require a lot of logic.

Yeah, this will unfortunately take some time to do well. Any provided property needs to be bounds and sanity checked, and we need to ensure reference integrity is maintained.

> Error handling. How forgiving do we want to be? Should we use a reasonable default for an invalid value, skip the object, or abort the whole process?
> Then maybe we'll have to compromise, and use either unlocalised, or very generic messages ("invalid value 'x' for 'y'").

The .anki-json authoring is intended to be done by power users and third-party software authors, so it's not unreasonable to abort when provided invalid input, as long as we're giving them enough of a hint of what they've done wrong. Errors that are in the file itself and don't depend on the current state of the importing collection (eg programming errors or garbage data) are probably ok to treat this way. 

If we have errors that the json author couldn't anticipate however (referencing a notetype that has a different structure than expected in the importing collection, etc), then maybe that needs an error that an end-user could interpret, and perhaps we should warn+skip rather than aborting?

> Skip serialising default values? Would allow for much smaller files, but maybe explicitness is to be preferred if JSON exports are mainly used for external modification.

Explicit is probably better here, as it makes it much easier for users to understand the format+modify.

rslib/src/import_export/text/csv/metadata.rs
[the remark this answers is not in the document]
dae: Yep, the user has actually provided one column, but we show 2 or more if the first line happens to have a space. A preview of the first few lines would help users.

RumovZ:
> Some users have expressed in the past the desire to be able to change things like deck or notetype partway through a file, so that cards/notes for more than one deck/notetype can be imported at once. 

Yeah, I thought about that, too. As you said, meta rows in the middle of a file would be a challenge, but I don't see why we couldn't support deck and notetype columns. Anki allows to export notes of different notetypes and decks, and it would be great to be able to also reimport such a CSV file. Some notes on that:

1. If a deck or notetype can't be resolved, the question is again whether to skip the note, use the default deck/notetype, or abort.
2. To allow mapping the fields of all referenced notetypes from the GUI, we'd have to read the whole file when gathering the metadata. If the turns out to be too slow, notetypes could be required to be declared in the header.
3. I think the current mapping logic is flawed. For every field or meta value, the user should select a column, not the other way around. (Just like the old vs. new notetype change dialog I guess.) This change should also help with multiple notetypes. E.g., you could map Column 1 to both 'Front' of 'Basic', and 'Front' of 'Basic (and reverse)'.

<hr>

The other points all sound good to me as well! Updating etc. sounds like it will become complex, but we have a plan of attack there.

dae:
> but I don't see why we couldn't support deck and notetype columns

How would that work when notetypes have differing numbers of fields? How would the importer know which column corresponded to the notetype in a given row?

RumovZ:
The backend would receive a map for for every notetype, as well as for meta values, where each field/value is assigned a column or None. Let's have an example:

CSV:
```
#columns:notetype,tags,Front,Back,Text,Extra
Basic,Capitals,Peru,Lima,,
Cloze,Capitals,,,{{c1::Lomé}} is the capital of Togo,
Basic (and reversed card),Capitals,Luxemburg,Luxemburg,,
...
```

Map:
```
Meta: { notetype: 0, tags: 1 }
Basic: { Front: 2, Back: 3 }
...
```

What I was planning to do but haven't done, yet, is automatically mapping columns according to their names in the import dialog. This should be helpful here, because we can't just map column 1 to field 1 etc., but the user shouldn't have to setup the map manually. There could also be a new tag like `#notetype-map` to allow more flexible mapping in the file head.

rslib/src/import_export/text/csv/metadata.rs
```diff
@@ -122,55 +133,49 @@ impl Collection {
...
-    Some(if value.as_bytes().len() == 1 {
-        value.as_bytes()[0]
-    } else {
-        match value.trim().to_ascii_lowercase().as_str() {
-            "tab" | "\\t" => b'\t',
-            "semicolon" => b';',
-            "comma" => b',',
-            "space" => b' ',
-            _ => return None,
+fn delimter_from_value(value: &str) -> Option<Delimiter> {
```
dae, line 152: typo

dae:
Changes look good! When we get to the updating part we may want to change those properties on ForeignCard to None, and rely on the backend to inject the defaults in for us instead.

> The backend would receive a map for for every notetype, as well as for meta values, where each field/value is assigned a column or None

Ok, I can see how that could potentially work for notetypes and tags, provided the meta columns come before the variable number of notetype columns.

Card-based properties like deck, interval and so on seem trickier though, since they'd applied to all cards of the note. This would prevent us from being able to round-trip them correctly in a csv export+import, unless we made things even more complicated with packed values or similar approaches.

There is still some value in supporting multiple notetypes in csv files, but as you pointed out, it means we need to read the entire file or predeclare them, and it makes our mapping code more complicated. That makes me wonder a bit about whether it's worth the extra effort? But no strong feelings about it at this point.

RumovZ:
> There is still some value in supporting multiple notetypes in csv files, but as you pointed out, it means we need to read the entire file or predeclare them

Only if we expose mapping in the GUI, though. But maybe we should only show the new settings for files with `#columns:...` etc., anyway. And if a file does describe its own columns, I wonder if there is any use case for overrriding that.
So maybe we should keep allowing to map the default notetype only, but support varying notetypes and decks inside files?

> Card-based properties like deck, interval and so on seem trickier though, since they'd applied to all cards of the note. This would prevent us from being able to round-trip them correctly in a csv export+import, unless we made things even more complicated with packed values or similar approaches.

I've mostly thought about CSV as a means to edit or add content. Indeed, supporting scheduling information, or any properties of objects other than notes, would increase complexity by a lot. I think `deck` is a bit different, though, because it's set on the note-level when adding.

dae:
> So maybe we should keep allowing to map the default notetype only, but support varying notetypes and decks inside files?

That keeps extra complexity out of the GUI, but still allows power users to support multiple notetypes at once, so it seems like a decent compromise.

> I think deck is a bit different, though, because it's set on the note-level when adding.

Good point. Maybe the best option is to just ignore the provided deck for existing cards?

RumovZ:
Foreign imports
---

I've been thinking about how to continue for a while now. As every object property that should be adjustable via JSON will need to be implemented explicitly, an incremental approach probably makes sense. I think third-party importers will only ever be interested in adding new data, and CSV will add or update notes only. So shall I focus on these aspects first? I assume exporting has a higher priority as well, because it's an already existing feature.
Then there are the additional features and fixes desired for apkgs. Maybe they should come before more extensive JSON support, too?

CSV
---

I've started implementing a new screen and it is almost fully functional, but there's one problem in `FieldMapper.svelte`: I can't get the field mapping to work (except for the first one). When the GUI changes the column index, it is immediately overwritten with the default setup. Strangely, it seems Svelte is reacting as if `columnNames` had been changed.
Do you know what I'm doing wrong? It's probably to do with the array binding. If you don't, maybe I'll have more luck using events.

Will take me a long time to make the UI presentable, though...

As mentioned above, it probably doesn't make sense to allow manually remapping meta columns. Maybe I should remove the selectors for notetype and deck? If there are notetype and deck columns in the CSV, they could still be listed in the dropdown menu, but unselectable.

dae:
Not sure exactly what was going on there, but breaking it into a separate function (so we can be sure only the provided two args trigger an update) seems to work:

```diff
diff --git a/ts/import-csv/FieldMapper.svelte b/ts/import-csv/FieldMapper.svelte
index 41590ac4f..687c88279 100644
--- a/ts/import-csv/FieldMapper.svelte
+++ b/ts/import-csv/FieldMapper.svelte
@@ -16,17 +16,20 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     $: options = [tr.changeNotetypeNothing()].concat(columnNames);

     let fieldNames: string[] = [];
-    $: {
+    let otherFieldIndices: number[] = [];
+
+    function onNotetypeChange(notetypeId: number, columnNames: string[]): void {
         getNotetypeFields(notetypeId).then((newFieldNames) => {
             fieldNames = newFieldNames;
+            otherFieldIndices = Array(Math.max(0, fieldNames.length - 1))
+                .fill(0)
+                .map((_, i) => (i + 1 < columnNames.length ? i + 2 : 0));
         });
     }

-    let firstFieldIndex: number = 0;
-    $: otherFieldIndices = Array(Math.max(0, fieldNames.length - 1))
-        .fill(0)
-        .map((_, i) => (i + 1 < columnNames.length ? i + 2 : 0));
+    $: onNotetypeChange(notetypeId, columnNames);

+    let firstFieldIndex: number = 0;
     $: fieldColumnIndices = [firstFieldIndex, ...otherFieldIndices.map((i) => i - 1)];
 </script>
```

In a a bit of a hurry today; will look through the code in more detail tomorrow. Feel free to review/offer any feedback if you're interested @hgiesel :-)

In terms of priorities, it would be nice to get the existing apkg import/export code you've written into the hands of more users - with it being behind a feature flag at the moment which prevents import/export of other formats, it's a bit difficult for users to test. What do you think about adding another feature flag that causes the old import/export path to use the new import/export apkg code, like we've done with colpkg handling? That way users could more easily try it out without breaking other workflows, and we could consider making that the default in the future like we did for colpkg, even if the rest of the import/export work is not done yet. 

For the timing of adding additional apkg functionality, I'm not sure what the best course of action is. I think I slightly lean towards getting the existing functionality out there first, as it will be easier to compare to the old Python code in the current state, and with real world use, we'll be more confident that we have a stable base on which to build. I'd also like to switch AnkiMobile over to the new importer sooner rather than later, as the old objective C implementation is a real maintenance headache.

That said, some of those proposed changes may benefit from or require some refactoring, so there is a potential risk that bugs will be introduced when the new functionality is added. What do you think?

Taking a gradual approach for updating/JSON sounds good, and non-CSV text import/export feels like it should be the last stage. As long as we continue to provide an option for the old code paths/add-ons until we're done, there's no rush here. 

Will get back to you on the GUI points tomorrow after I've had a chance to explore/review the code.

RumovZ:
Thanks for the fixes! I had tried a lot of different stuff in FieldMapper, but somehow missed this solution.

> What do you think about adding another feature flag that causes the old import/export path to use the new import/export apkg code, like we've done with colpkg handling?

As for importing, I think there is not much left preventing us

RumovZ:
Thanks for the fixes! I had tried a lot of different stuff in FieldMapper, but somehow missed this solution.

> What do you think about adding another feature flag that causes the old import/export path to use the new import/export apkg code, like we've done with colpkg handling?

As for the new import logic, I think there is not much left to do before it is on a par with the old one: tidying up the GUI, handling note updates, logging updates/conflicts. Did I miss something?
Exporting is a walk in the park compared to importing anyway. Maybe if I just focus on the existing functionality now, the new code is fully usable soon enough?
The requested apkg features will probably take some time, on the other hand, and I'd vote to postpone them, too.

rslib/src/import_export/text/import.rs
```diff
@@ -0,0 +1,239 @@
...
+        let card_gen_context = self
+            .card_gen_ctxs
+            .entry((notetype.id, deck_id))
+            .or_insert_with(|| CardGenContext::new(notetype, Some(deck_id), self.usn));
+        self.col
+            .generate_cards_for_existing_note(card_gen_context, note)
+    }
+}
+
+pub(super) trait NotetypeForString {
```
RumovZ, line 146: I've tried something new here, what do you think? Implementing the required Collection API in traits allows us to setup a mock for testing, which is cheaper to build, and whose state is easier to control. It it also helps a bit with namespaces.
The main motivation was to avoid setting up a whole collection for every unit test. That always felt overkill, but I didn't do any serious benchmarking. Do you think it's worth the addtional boilerplate?

dae:
Remains of the existing functionality now/soon, and new apkg features later sounds good. :-) Will take a look through the code now.

rslib/src/import_export/text/import.rs
dae: Is setting up a collection for each test that requires one such a big deal? Many of our unit tests don't require a collection, and the total runtime of all our Rust tests is only 0.6s currently:

```
[dae] code/desktop/anki [main]  % bazel run rslib:anki_tests
...
test result: ok. 155 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.60s
```

I don't mind such an approach if we need it, but I just wonder whether we really need it :-)

ts/import-csv/FieldMapper.svelte
```diff
@@ -29,18 +27,17 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     $: onNotetypeChange(notetypeId, columnNames);
 
+    // first index is treated specially, because it must be assigned some column
     let firstFieldIndex: number = 0;
+    // other indices are shifted by -1, because the first option is n/a, which
+    // is represented by a negative value
```
dae, line 33: We could potentially simplify our code a bit here by not special-casing the first field here, and instead showing an error at submission time if the first field is not mapped.

ts/import-csv/MetaMapper.svelte
```diff
@@ -6,14 +6,22 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
 
     $: options = [tr.changeNotetypeNothing()].concat(columnNames);
 </script>
 
-<MapperRow label={tr.editingTags()} {options} bind:index={tagsColumn} />
-<MapperRow label={tr.decksDeck()} {options} bind:index={deckColumn} />
-<MapperRow label={tr.notetypesNotetype()} {options} bind:index={notetypeColumn} />
+<MapperRow label={tr.editingTags()} {options} bind:index={tagsIndex} />
+<MapperRow label={tr.decksDeck()} {options} bind:index={deckIndex} />
+<MapperRow label={tr.notetypesNotetype()} {options} bind:index={notetypeIndex} />
```
dae, line 27: From your earlier comment:

> Maybe I should remove the selectors for notetype and deck? If there are notetype and deck columns in the CSV, they could still be listed in the dropdown menu, but unselectable.

Yeah, it feels a bit confusing at the moment - having both manual selection in the import options, but also supporting mapping makes it a bit unclear (at least to me) how things will work. Maybe instead of showing these rows, we could just show an info line if deck or notetype have been set in the file? And for the field/tag mapping, are we able to disable certain columns from selection? If not, that might make things tricky if we're relying on array indices.

dae (commented):
One idea I had while looking through this: would it make sense/be easy to extract the fields of the first row, and add them to the column names to help the user? Eg instead of showing "Column 1", it could show "Column 1 ([text from first field])". Perhaps the first row could be included as part of the metadata?

ts/import-csv/HtmlSwitch.svelte
```diff
@@ -0,0 +1,21 @@
...
+</script>
+
+<Row --cols={2}>
+    <Col --col-size={1}>
+        {tr.importingAllowHtmlInFields()}
+    </Col>
+    <Col --col-size={1} --col-justify="flex-end">
+        <Switch id={undefined} bind:value={isHtml} />
+    </Col>
+</Row>
```
dae, line 21: As we only have a single switch at the moment this is not a big deal, but in the long run we probably don't want to be creating separate files for each switch we add - it would be a bit nicer to have a reusable SwitchRow that allows passing in the boolean and text. We have a SwitchRow in deck-options that is similar, though we don't really need the revert button in this case. We could potentially make the revert button optional, or create a separate SwitchRow here. @hgiesel is the most qualified in this area, and he may have a preference on the way we should do it.

rslib/src/import_export/text/import.rs
RumovZ: You're right, it doesn't seem to be necessary. 🙂 Tests are dwarfed by compilation time. Maybe it would be helpful in cases where a more exotic collection state is required for testing.

ts/import-csv/FieldMapper.svelte
RumovZ: Do you mind this complexity? I'd have said that it's worthwhile, if it prevents invalid user input.

ts/import-csv/MetaMapper.svelte
RumovZ: An info sounds good. Should we also disable the deck/notetype selectors if such columns are present? Currently, we use them as fallbacks, if a row leaves these columns empty, but I guess that might be confusing, too. Either all or no notetypes/decks should be allowed to be remapped.
Consequentially, every option that is set in the file should be disabled then. We'd need a `force` flag for every value in the CsvMetadata message, though.

ts/import-csv/HtmlSwitch.svelte
RumovZ: Agreed, I'd generalise HtmlSwitch, if there was more than one use case, and ideally SwitchRow would be reused, but I was afraid to mess up existing pages. I'm not very skilled with the styling of different-level components.

RumovZ:
> Eg instead of showing "Column 1", it could show "Column 1 ([text from first field])". Perhaps the first row could be included as part of the metadata?

Wouldn't a preview achieve something to the same effect?
I'm also considering changing the label for named columns to "Column 'Front'" instead of just "Front".
One more idea: Instead of using the last setting for the HTML checkbox (which the new code doesn't do, yet) we could try to detect if the file contains HTML. This wouldn't always be perectly accurate, but maybe it's still more useful than the old behaviour?

ts/import-csv/MetaMapper.svelte
RumovZ: Just thought, we probably don't even need info boxes if we go this route. If a widget is greyed out and says "This setting is defined in the chosen file", it's probably clear enough.

As for disabling select options, EnumSelector from deck-options supports this already. But of course, we could also replace the array with an object.

rslib/src/import_export/text/import.rs
```diff
@@ -40,6 +40,24 @@ struct Context<'a> {
...
+    Ignored,
+    Updated,
+}
+
+impl NoteLog {
+    fn log_result(&mut self, note: Note, result: ImportResult) {
+        match result {
+            ImportResult::Added => &mut self.new,
+            ImportResult::Ignored => &mut self.duplicate,
+            ImportResult::Updated => &mut self.updated,
```
RumovZ, line 55: The messages generated from the log are currently apkg-specific. Not yet sure how to handle this.

ts/import-csv/FieldMapper.svelte
dae: I don't feel strongly about it, so no need to change if you prefer it this way.

ts/import-csv/MetaMapper.svelte
dae: +1 on all/none - I don't think the complexity of allowing a fallback is worth it

> We'd need a force flag for every value in the CsvMetadata message, though.

I'm not sure I'm following here - isn't it just deck/notetype we're concerned about?

ts/import-csv/HtmlSwitch.svelte
dae: Sounds good. And me either :-)

dae:
> Wouldn't a preview achieve something to the same effect?

Sure, that could work too. Showing a few examples may make it clearer to users that just the first row, though it does have the disadvantage of making the controls for selecting each field further away from the example content.

> One more idea: Instead of using the last setting for the HTML checkbox (which the new code doesn't do, yet) we could try to detect if the file contains HTML.

Could be useful; sounds like it's worth a try.

rslib/src/import_export/text/import.rs
dae: I'm not sure I'm following - do you mean importing.py:210+? What is specific to apkgs there?

rslib/src/import_export/text/import.rs
```diff
@@ -104,22 +113,73 @@ impl<'a> Context<'a> {
...
+            .col
+            .storage
+            .note_fields_by_checksum(note.notetype_id, note.checksum.expect("set by caller"))?
+            .into_iter()
+            .filter(|(_, field)| strip_html_preserving_media_filenames(field) == first_stripped)
+            .map(|(nid, _)| nid)
+            // FIXME: There may be multiple duplicates and the old Python code updated them all in this
+            // case. However, this is probably not generally useful and worth the additional complexity.
+            // The note should probably be skipped, or an error could be raised.
+            .next())
```
dae, line 187: What about .last() instead of .next()? That way we update everything in this unlikely case, but don't concern ourselves with multiple note ids.

rslib/src/import_export/text/import.rs
```diff
@@ -241,3 +301,70 @@ impl ForeignTemplate {
...
+        data.dupe_resolution = DupeResolution::Update;
+
+        data.clone().import(&mut col).unwrap();
+        assert_eq!(col.storage.notes_table_len(), 1);
+
+        data.notes[0].fields[1] = "new".to_string();
+        data.import(&mut col).unwrap();
+        assert_eq!(col.storage.get_all_notes()[0].fields()[1], "new");
+    }
+}
```
dae, line 420: Love seeing more tests being added

rslib/src/storage/note/mod.rs
```diff
@@ -313,6 +313,15 @@ impl super::SqliteStorage {
...
     }
+
+    #[cfg(test)]
+    pub(crate) fn notes_table_len(&mut self) -> usize {
+        self.db
+            .prepare("SELECT COUNT(*) FROM notes")
+            .unwrap()
+            .query_row([], |row| row.get(0))
+            .unwrap()
+    }
```
dae, line 324:
```diff
diff --git a/rslib/src/storage/note/mod.rs b/rslib/src/storage/note/mod.rs
index e8f68e57c..df9d1e170 100644
--- a/rslib/src/storage/note/mod.rs
+++ b/rslib/src/storage/note/mod.rs
@@ -316,11 +316,7 @@ impl super::SqliteStorage {

     #[cfg(test)]
     pub(crate) fn notes_table_len(&mut self) -> usize {
-        self.db
-            .prepare("SELECT COUNT(*) FROM notes")
-            .unwrap()
-            .query_row([], |row| row.get(0))
-            .unwrap()
+        self.db_scalar("select count(*) from notes").unwrap()
     }
 }
```

ts/import-csv/MetaMapper.svelte
RumovZ: I meant if the file specifies `#delimiter:Colon` or `#is_html:true`, then it doesn't make much sense to let the user change these settings. But currently, `get_csv_metadata()` just returns an enum/boolean, so the frontend code can't know whether it's auto-detected, thus allowed to be changed, or explicitly set in the file, therefore disallowed to be changed by the user.
So the message would need a separate flag `force_is_html` bool to inform about that.

rslib/src/import_export/text/import.rs
RumovZ: Sorry, it's confusing because I couldn't comment the actual lines of interest.
The problem is in `stringify_log()`, which uses messages like "Notes skipped, as they're already in your collection: { $val }". Compared to "First field matched: { $val }", which is used for CSV imports in the old code.
Maybe we need more different log queues like `full_duplicate` and `duplicate_first_field`.

RumovZ: But I see, you found the relevant lines importing.py:210. Alternatively, we could use more generic messages, of course. E.g. "Added duplicate with first field: { $val }" would be just subsumed into "Notes added from file: { $val }".

ts/import-csv/DupeResolutionSelector.svelte
```diff
@@ -0,0 +1,56 @@
...
+                value: DupeResolution.UPDATE,
+                label: tr.importingUpdateExistingNotesWhenFirstField(),
+            },
+        ];
+    }
+</script>
+
+<Row --cols={2}>
+    <Col --col-size={1}>
+        {tr.importingDuplicateResolution()}
```
RumovZ, line 46: Is this understandable? I also think the combobox option labels below are cumbersome. I know they were made for a larger box without label in the old screen. But I couln't come up with something better.

ts/import-csv/MetaMapper.svelte
dae: Ah, I see what you mean. I don't have strong feelings here - disabling the items that are provided in the file is probably a bit more user friendly, but directives in text files are probably going to be used mainly by power users, who should theoretically understand that it doesn't make sense to change them in the GUI later. The deck/notetype selectors were a bit more of a priority, since we were displaying two different ways to input them, which was somewhat confusing. But no objections to disabling these fields if you think that's best.

rslib/src/import_export/text/import.rs
dae: Good point, it would probably be better if we can preserve the mention of the first field. Separate log queues could work. One other alternative would be to add an enum field to ImportResponse that tells the GUI which of the two translation strings it should pick, eg 

```
  enum DuplicateMatchingStrategy {
    GUID = 0;
    FIRST_FIELD = 1;
  }
```

Might be a bit easier to implement the separate queues approach in the GUI though.

ts/import-csv/DupeResolutionSelector.svelte
dae: How about "Duplicate resolution" -> "Existing notes"

If removing the label and using the entire row for the selection box doesn't look good, maybe we could add some new translations instead, that were shorter, eg "Update", "Preserve", "Duplicate". We could add a tooltip with more detailed explanations if the short forms were not clear enough.

rslib/src/import_export/text/import.rs
RumovZ: I think this is note-specific, though. I a JSON file you could have both, notes with and without a guid/id. Then different queues are probably simpler.

ts/import-csv/MetaMapper.svelte
RumovZ: > but directives in text files are probably going to be used mainly by power users,

I was also planning to write some metadata when exporting. At least `html` and `delimiter` depend solely on the file and not the state of the target collection. If we know these settings for certain we can spare the user the chance to mess things up.

ts/import-csv/DupeResolutionSelector.svelte
RumovZ: Sounds good! I'll leave tooltips out for now, but keep it in mind.

rslib/src/import_export/text/import.rs
RumovZ: I ended up with something in between: a single queue for all first-field matches and an enum. It's not perfect, because the fluent template strings for CSV and apkg import are constructed differently, but it seemed too early to add new ones.

ts/import-csv/index.ts
```diff
@@ -0,0 +1,83 @@
...
+export async function setupImportCsvPage(path: string): Promise<ImportCsvPage> {
+    const gettingMetadata = getCsvMetadata(path);
+    const [notetypes, decks, metadata] = await Promise.all([
+        gettingNotetypes,
+        gettingDecks,
+        gettingMetadata,
+        i18n,
+    ]);
+
+    const [deckId, deckColumn] =
```
RumovZ, line 46: Ugh. Is there a better way to deal with oneofs in TS?
One idea I had was to use one-based column indices. Then `deckColumn == 0` would always mean unset, instead of now where it can mean either unset or set to 0.

Apart from that, should we use one-based indices in the CSV file, anyway? I.e. `#deck column:1` for the first column in the file.

RumovZ:
Test failures are caused by visibility issues again I assume, but [81d321a](https://github.com/ankitects/anki/pull/1850/commits/81d321a8bdbffee768902345a6213bc53383c8ed) and [530a296](https://github.com/ankitects/anki/pull/1850/commits/530a296ecbba07c0ddc42788e90cb9ac7c48f0ff) are chiefly to get your feedback anway, @dae and @hgiesel.
What do you think of using the tag editor there? Apart from the top border and possibly the breaking behaviour, it pretty much works great out of the box. Is it a bad idea to have two on the same screen?

RumovZ:
Trying to come to a conclusion so this PR can be merged. I've updated the todo list at the top, but most of it would be new features. Do you see anything else that prevents a merge, apart from the tag editor stuff?

By the way, I won't be able to do anymore dev work before Monday.

hgiesel:
I cannot get the current version to compile, I get a lot of resolution errors for tag-editor, here's an excerpt:

```
Use --sandbox_debug to see verbose messages from the sandbox
✘ [ERROR] Could not resolve "../helpers"

    bazel-out/darwin-fastbuild/bin/ts/editor/tag-editor/TagEditor.svelte.mjs:36:28:
      36 │ import { execCommand } from "../helpers";
         ╵                             ~~~~~~~~~~~~

✘ [ERROR] Could not resolve "./tag-options-button"

    bazel-out/darwin-fastbuild/bin/ts/editor/tag-editor/TagEditor.svelte.mjs:37:33:
      37 │ import { TagOptionsButton } from "./tag-options-button";
         ╵                                  ~~~~~~~~~~~~~~~~~~~~~~

✘ [ERROR] Could not resolve "./tags"

    bazel-out/darwin-fastbuild/bin/ts/editor/tag-editor/TagEditor.svelte.mjs:46:7:
      46 │ } from "./tags";
         ╵        ~~~~~~~~

✘ [ERROR] Could not resolve "./tags"

    bazel-out/darwin-fastbuild/bin/ts/editor/tag-editor/TagInput.svelte.mjs:28:7:
      28 │ } from "./tags";
         ╵        ~~~~~~~~
```

I would also suggest that we move the tag-editor into its own directory, or into components at this case, as its no longer an exclusive editor component.

I actually remember checking out this PR before, but I couldn't find a way to invoke the dialog that you programmed in the UI. Do I have to set a flag?

dae:
@hgiesel the following in the debug console will enable it: mw.pm.set_new_import_export(True)

@RumovZ sorry, things have been a bit hectic - I'll try to give this a thorough review tomorrow. Just one brief comment for now: I gather you moved the import button down the bottom based on the related forum discussion. We may want to change that back before merging in unless we're sure we want to make the same change to the other screens, as otherwise it's inconsistent. The main reason why I'd lean towards keeping it on top is that it's also where we tend to put options that affect everything below (eg the preset selection in the deck options, or note type selection in the change notetype screen) - having those selectors on the bottom might feel a bit odd.

RumovZ:
@hgiesel, unfortunately I can't easily sort out these errors, because I don't encounter them on my Windows machine. Damien has fixed it for now, but as you said, this must finally be fixed by moving the tag editor to a different folder.

@dae, no worries, I'm off until Monday anyway.
It makes sense for the deck options screen, because the save button refers to the preset selector, which must be located at the top. But for most dialogs, it feels weird to me to have that button anywhere but at the bottom right. The forum discussion indeed reinforced my intuition.
Apart from convention, I think it's also not ideal to have two levels of sticky headers pushing the interactive content further down.

dae:
> no worries, I'm off until Monday anyway.

Ok, I'll try and chip away at this and aim to have everything replied to/reviewed by the time you get back :-) 

> While I have your attention, may I ask if the Svelte/TS analysis works as poorly for you as it does for me? It keeps crashing (all the time if Rust Analyzer is enabled), often tells me paths like "../lib/ftl" could not be found, complains about incompatible assignments, where it compares the type in the source code with the type in the build folder, but doesn't report obvious bugs like when I set a non-existing prop on a component. I guess the last one is a limitation of Svelte, because that doesn't show up when testing, either.

I find I get a bunch of errors if I switch to this branch while VS Code is already open. Closing it and re-opening after switching branches seems to make those go away. I also try to make sure I've run Anki from the source folder first, so that things like the generated protobuf code are available when VSCode does the initial index. I seem to recall there have been other times where I had to restart VSCode due to it no longer typechecking, or showing errors.

Adding an invalid property to a component does result in an error inside vscode for me (eg changing `<Component>` to `<Component foo=5>` in ImportCsvPage.svelte). One possibility is that the issues you're experiencing are Windows-specific - I do all development on a Linux or Mac machine and then rsync code over to a Windows machine to actually build/run it, so haven't tried running VSCode on Windows. If that turns out to be the problem, one way you could work around it is to do development with WSL.

ts/import-csv/index.ts
dae: Like in Python, I think we can access the fields directly and rely on them being null/undefined in the case that it's a different oneof, eg something like:

```typescript
    const deckId = metadata.deckId ?? null;
    const deckColumn = metadata.deckColumn ?? null;
```

Or if you wanted to enforce deckColumn being null if deckId is defined, something like:

```typescript
    const deckColumn = deckId === null ? metadata.deckColumn ?? null : null;
```

1-based indexing is the more user-friendly option, and deferring the translation to 0-based indexing until the last moment could potentially save us having to deal with oneofs. It does increase the risk of off-by-one bugs - if not in the initial implementation, then when we return to the code at a later date and have forgotten how we chose to implement it. :-) But either approach seems reasonable, and I don't have a strong preference.

dae:
> What do you think of using the tag editor there? Apart from the top border and possibly the breaking behaviour, it pretty much works great out of the box. Is it a bad idea to have two on the same screen?

My initial impression wasn't the best, mainly due to the top bar that changes size as the entered text changes. After playing with it a bit more, I don't mind it, and can see how having the tags autocomplete could be potentially useful. There are a few issues that would need addressing though:

- the aforementioned top border
- the displayed tags and tag icon appear on top of the sticky headers when scrolling

There may be other things that @hgiesel can touch upon. One option we might want to consider is using a simple text input box for the initial implementation, and then tackling these issues in a future PR.

> But for most dialogs, it feels weird to me to have that button anywhere but at the bottom right. The forum discussion indeed reinforced my intuition. Apart from convention, I think it's also not ideal to have two levels of sticky headers pushing the interactive content further down.

I'm not so sure about the sticky headings either - is it really useful to keep showing "Import Options" to the user as they scroll down on a small screen? But maybe I'm in the minority here.

dae (commented):
I haven't reviewed most of the code yet, and will keep working through it. While there are some issues to work through, I think you've done a great job on this, and I'm excited to see it come together.

A few things I noticed:

- The Mnemosyne importer is currently broken ("skipped as deck was missing")
- Perhaps 'include html' should be the default on tsv export, so that content round-trips by default?
- The HTML checkbox label is a bit confusing, as it talks about media references, but sound tags are not stripped. Maybe we could either change the wording, or strip sound tags too.
- When entering a filename without extension in the export screen, 'txt' instead of '.txt' is appended
- The default option for "Existing notes" should probably be 'update', not 'duplicate'.
- Ideally, when 'update' is selected and the other fields/tags have not changed, we'd show '[ignored]' rather than '[updated]'
- When the text file has more fields than the currently-selected notetype, it's not obvious that the extra fields will be discarded. Maybe we could shown a warning message, or deal with this in the preview?

ts/import-csv/StickyFooter.svelte
```diff
@@ -0,0 +1,59 @@
...
+    import Shortcut from "../components/Shortcut.svelte";
+    import * as tr from "../lib/ftl";
+    import { getPlatformString } from "../lib/shortcuts";
+
+    export let path: string;
+    export let onImport: () => void;
+
+    function doImport(): void {
+        if (document.activeElement instanceof HTMLElement) {
+            document.activeElement.blur();
```
dae, line 19: Just curious, why is this needed?

rslib/src/import_export/text/csv/metadata.rs
```diff
@@ -0,0 +1,590 @@
...
+            }
+            true
+        } else {
+            false
+        }
+    }
+
+    fn parse_meta_value(&mut self, key: &str, value: &str, metadata: &mut CsvMetadata) {
+        match key.trim().to_ascii_lowercase().as_str() {
+            "delimiter" => {
```
dae, line 89: If using 'separator' in the GUI, we might want to use it here as well.

dae (commented):
Ok, I've finished going through the code. The code was lovely and neat, and I struggled to find any issues with it. :-)

ts/import-csv/StickyFooter.svelte
RumovZ: I kept it around because I thought it might become necessary. But I guess it's not. 🙂

ts/import-csv/index.ts
RumovZ: > Like in Python, I think we can access the fields directly and rely on them being null/undefined in the case that it's a different oneof

Don't we use `HasField()` in Python because that's exactly not so and unset fields yield a default value?
I ran into that exact issue here: `metadata.deckColumn` is 0 if unset.

> It does increase the risk of off-by-one bugs

I agree, hopefully it could be minimised with tests and comments. Since we label columns starting with "Column 1" in the GUI, I think it would be really confusing to expect zero-based indices in the file.

RumovZ:
Happy that you like the code!

Thanks for your comments on environment setup! Regarding the missing property error, I should have been more specific: An unknown property name generally does cause an error, but didn't in one particular case. Also, I often get an unhelpful syntax error instead ("missing `>` on line 1").
I didn't note down the exact ways to reproduce either of these issues, but will do in the future.

>  is it really useful to keep showing "Import Options" to the user as they scroll down on a small screen?

Probably not, but it might on a page with more sections, where it's not immediately obvious which section a setting does belong to. Then it would make sense to have headers behave the same on all pages. Arguably, it's also less aesthetic to have a low-level text element directly beneath the window top bar.
But I really wouldn't mind either way.

> When the text file has more fields than the currently-selected notetype, it's not obvious that the extra fields will be discarded. Maybe we could shown a warning message, or deal with this in the preview?

You mean like "not all columns have been mapped"? There might also be unmapped fields if a row has more fields than the first one, which we can't (or choose not to) detect. I'm not sure there should be a warning, if it's not completely reliable.

I'll look into fixing the tag editor. If that doesn't work out quickly, I'll split it into a separate PR and fallback to a plaintext field.

ts/import-csv/index.ts
dae: Ah, you're right. Not the first time that behaviour has tripped me up. Life would be easier if we only had to deal with tagged unions in Rust!

> I think it would be really confusing to expect zero-based indices in the file.

:+1:

dae:
> You mean like "not all columns have been mapped"? There might also be unmapped fields if a row has more fields than the first one, which we can't (or choose not to) detect. I'm not sure there should be a warning, if it's not completely reliable.

This actually highlights a difference between the old and new importers - in the old importer, if a later row had a different number of columns from the first row, it was ignored and a warning was shown. We can't do that when we're allowing the notetype to be specified on a per-row basis, but it could potentially have some value in the common case where the user is targeting a single notetype. The reason the old importer rejected differing columns in this case is that it tended to be accidental (eg by specifying the wrong field separator, or by including an unquoted newline in a field).

Not super important in any case, and we could presumably add it later if we notice it proves to be a common point of confusion.

Thanks for addressing all those other issues. If you're ready for me to merge this in, please remove the draft tag to let me know.

RumovZ:
> Not super important in any case, and we could presumably add it later if we notice it proves to be a common point of confusion.

I've added it to the todo list.

Would you take care of moving the tag editor to components? I think it requires a new build script (would be the first multi-file component) and it would take me some time to figure this out and debug on my Windows machine.

dae:
I've placed it into a separate module instead of components, as it felt complicated enough to justify it. @hgiesel feel free to comment or push changes if you think it should be done differently

dae:
> We could also add the old third-party importers to the new screen. 

That sounds like a better idea - if that's not too hard, it would be good to get that in place before we switch the default to the new import/export code in the next feature update.

> should there be a new webview for the export screen as well? A number of new options need to be added, but nothing as complex as the import mapping

Svelte is a better choice for new screens, but if these changes just involve a few checkboxes/dropdowns without special logic, updating the existing screen would probably be fine too, as it should be fairly easy to convert in the future if need be.

Issue #1792, opened 2022-04-13
Reduce font size of editor toolbar/field labels?


dae:
https://forums.ankiweb.net/t/anki-2-1-50-inconvenient-interface-size/19024

dae:
@hgiesel do you think we should adjust the field label size at all, or just call it 'done' since #1853 was merged in?

dae:
@hgiesel ping

hgiesel:
The value I changed did not affect the field labels, so they're still at the same size. I think we could hook the up to the `--base-font-size` var we also have floating around.

dae:
With the following, the field name feels a bit tiny on Windows at non-2x. On a Linux machine it seems ok. On a Mac it feels a bit small as well. Maybe we only should be applying this on certain DPI configurations? I'm tempted to just call this complete for now, and wait until we see further feedback before moving further.

```diff
diff --git a/ts/editor/LabelName.svelte b/ts/editor/LabelName.svelte
index 527acdbda..20eb41f5c 100644
--- a/ts/editor/LabelName.svelte
+++ b/ts/editor/LabelName.svelte
@@ -10,5 +10,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 <style lang="scss">
     .label-name {
         user-select: none;
+        font-size: var(--base-font-size);
     }
 </style>
```
![2022-06-20T12:35:26,510441453+10:00](https://user-images.githubusercontent.com/180542/174518590-e38e51f7-bdc7-4238-bb3d-1c6128c7b9c4.png)

dae:
Let's call it complete

PR #1853, opened 2022-05-09
Set editor size to 1.6 on Windows as well

hgiesel:
Partially adresses #1792


dae:
At least on the Linux system I'm using here, this seems like an improvement for Linux as well.

PR #1854, opened 2022-05-09
Change sequence of Mathjax escaping

hgiesel:
Follow-up to #1818.

I noticed that there was still an issue when using `<`, `>` directly.
Could swear this did not happen when writing the original PR, but the issue is clear: escaping transformed `<` into `&amp;lt;` instead of just `&lt;`, which could not be undone by `unescapeSomeEntities`.


dae:
Thanks Henrik, this was my fault: bc6ede7c11aaf3fcaf62f00605fba97b151024b1

PR #1855, opened 2022-05-09
Fix review queue if limit is reached

RumovZ:
Quite an obvious bug... I'm surprised it hasn't been complained about sooner. 
Tests are messy, unfortunately, but at least I've added one to demonstrate the issue.

https://forums.ankiweb.net/t/sums-calculated-wrong/19770/5


dae:
Thanks Rumo, both the fix and the added test are appreciated.

PR #1856, opened 2022-05-10
remove image.style.height on resize

BlueGreenMagick:
Image style editor add-on may also set `image.style.height`.

Related: #1579 https://forums.ankiweb.net/t/macos-anki-2-1-51-qt5-incorrect-image-resizing/19469

### DIFF
--- a/ts/editor/image-overlay/ImageHandle.svelte
+++ b/ts/editor/image-overlay/ImageHandle.svelte
@@ -152,12 +152,13 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         }
 
         /**
-         * Image resizing add-ons previously used image.style.width to set the
-         * preferred width of an image. In these cases, if we'd only set
-         * image.width, there would be no visible effect on the image.
-         * To avoid confusion with users we'll clear image.style.width (for now).
+         * Image resizing add-ons previously used image.style.width/height to set the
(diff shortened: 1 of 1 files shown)


ts/editor/image-overlay/ImageHandle.svelte
```diff
@@ -152,12 +152,13 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         }
 
         /**
-         * Image resizing add-ons previously used image.style.width to set the
+         * Image resizing add-ons previously used image.style.width/height to set the
          * preferred width of an image. In these cases, if we'd only set
          * image.width, there would be no visible effect on the image.
          * To avoid confusion with users we'll clear image.style.width (for now).
```
hgiesel, line 158: The comment is a bit confusing now. Maybe reword it somewhat like this?:

* Image resizing add-ons previously used image.style.width/height to set the
* preferred dimensions of an image. In these cases, if we'd only set
* image.[dimension], there would be no visible effect on the image.
* To avoid confusion with users we'll clear image.style.[dimension] (for now).

Open for suggestions.

BlueGreenMagick: I didn't read the next lines. Fixed!

hgiesel (approved):

PR #1857, opened 2022-05-10
Change behaviour or image handles

BlueGreenMagick:
This PR is quite opinionated. 

Personally, I feel that the image resizing handles are very unintuitive when shrinking images. As can be seen in the below video, for nw (northwest), ne, and sw handles you have to drag _outside_ the image to shrink it, when dragging outside should make the image larger. It's difficult to change width with the ne handle, or height with sw handle.

This PR changes the ne handle into a width handle, sw into height, and disables nw handle. 

Original:
image_handle_behaviour

This PR:
Screen Recording 2022-05-10 at 1 42 06 PM

### DIFF
--- a/ts/editor/HandleControl.svelte
+++ b/ts/editor/HandleControl.svelte
@@ -107,10 +107,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         &.nw {
             top: calc(0px - var(--offsetY));
             left: calc(0px - var(--offsetX));
-
-            &.active {
-                cursor: nw-resize;
-            }
         }
 
         &.ne {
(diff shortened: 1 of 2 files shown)


hgiesel:
Hm, I'm not a big fan of this change.

The resize handles suggest the direction in which the user should resize, moving the cursor _into_ the center of the picture to make it smaller, _away_ from the center to make it bigger.

Disabling the nw handle makes things awkward once you're dealing with right-floating pictures or rtl text.

I also mostly copied the behavior from CKEditor, which you can test [here](https://ckeditor.com/ckeditor-5/demo/).

Maybe as a compromise we could also give the edges of the picture a resize handle, which would resize more like you expected it to.

BlueGreenMagick:
I think you may be right. I looked into how other softwares implemented image resizing, and surprisingly MS Word behaves the same as the current implementation. If even word does that, I am probably the minority that don't like how it is.

I think having side resize handle may make it too cluttered. I can just resize using the se handle so it's no big problem

PR #1859, opened 2022-05-10
Change how resizable images work

### LINKED ISSUE #1791: image resizing ergonomics
- Images shrink on focus change, so they can't be referenced while typing in a different field
- Should the behaviour be configurable?

https://forums.ankiweb.net/t/2-1-50-images-in-editor-too-small/18977

### LINKED ISSUE #1804: image changes size when pressing enter
https://forums.ankiweb.net/t/2-1-50-images-in-editor-changing-size-seemingly-randomly/19096

hgiesel:
Closes #1804, closes #1791 (maybe arguable).

Size constrained images, now have some HTML attributes, marking them so, e.g.:

```html
<img
    src="Screenshot%202022-05-10%20at%2010.07.30.png"
    width="469"
    data-editor-size-constrained="true"
    style="--editor-width:250px;"
>
```

In this case, the natural width is 158px (not visible in HTML), the resized width to be used in the reviewer is 469px, and in the editor it will be displayed with 250px width (the algorithm for detecting the editor width is that it will find an aspect ratio, so that its width does not exceed 250px and its height does not exceed 125px).

Whether an image is displayed full size or constrained is kept when you open/close the editor.

When pasting in a new picture, it will now be full size by default, to constrain its width, the user can use the button, or double-click. So images are no longer constrained by default. It's arguable whether this is the better behavior, but its easier to implement, as constraining the image takes an extra calculation step (to set `data-editor-size-constrained` and`--editor-width`).


dae:
I like this - expanded by default feels the most natural to me, and means for users who are pasting images of text, the text is usually readable without any work on their part. The state preservation seems to work well too, and it's great that images no longer resize on focus change.

Some minor suggestions from playing around with it:
- When the user clicks on a shrunken image, the image dimensions that are used during review are shown instead of the displayed dimensions, which is a bit confusing. What do you think about removing the first dimension (or both) in the shrunken case, and instead showing something like "Double-click to expand" or "(shrunken in editor)" or similar?
- Maybe a button to restore an image to its original size when it has been adjusted?

And some questions:
- The code changes don't look too big - presumably you'd be happy to have this go into 2.1.53?
- For the sizing of shrunken images, could we achieve similar results with min-width/min-height styling instead? Packing a separate width in as a css variable feels a little hacky.
- data-editor-size-constrained -> data-editor-shrink?

dae:
Thanks Henrik, this looks great. All good to go once 2.1.52 is out.

PR #1860, opened 2022-05-10
Implement cloze buttons in Mathjax editor

### LINKED ISSUE #1822: support cloze button while mathjax editor active?
How difficult would this be? I'm assuming https://forums.ankiweb.net/t/can-live-preview-of-mathjax-be-tunred-off-in-the-card-editor/19414/5 was talking about the MathJax macros and not the cloze button. :-)

hgiesel:
Closes #1822.

Currently the cloze buttons will show even in non-Cloze notetypes. Fixing this would require some additional logic in Python with web.eval, which I would like to avoid, because after #1691 this could be solved much more easily.

This mean we could put this on hold until after #1691 or if you deem it important enough, we could merge it now.


dae:
I was a bit worried about having the cloze button always available, but when the user adds one to a basic type, a warning message pops up at the bottom of the window, so hopefully there won't be too much confusion until we can implement a better solution with #1691.

While playing around with this, the bad rendering of clozes in the preview bothered me enough to try to fix it. I'll push a commit with those draft changes into a new PR shortly.

PR #1861, opened 2022-05-10
Move focus into HTML editor when shown 

### LINKED ISSUE #1778: Consider switching focus when enabling HTML editor
https://forums.ankiweb.net/t/anki-2-1-50-beta-6-release-candidate/18181/124

@hgiesel WDYT? Would this be easy to do, and do you think it's a good idea?

hgiesel:
+ Move focus into visual editor when closing HTML editor with focus in it.

Closes #1778.


kvn1351:
Hey @hgiesel,

I was wondering why `_addFocusCheck()` is still a thing after your implementation of the focusTrap. 

Please have a look:
https://github.com/ankitects/anki/blob/21812556a6a29c7da34561e58824219783a867e7/qt/aqt/editor.py#L332-L338

If I'm not misunderstanding [8270ce1](https://github.com/ankitects/anki/pull/1861/commits/8270ce1eb87f74d6d1c55d0cd1a7019173ca7ff7), there's no need for this check anymore right?

I'm asking because it's causing problems for [Liresol/anki-custom-shortcuts](https://github.com/Liresol/anki-custom-shortcuts) because it's effectively breaking the shortcut logic.

I'll try to explain:

If you click anywhere on the Editor's background, then the focus stays intact on the field. This is the intended new behavior courtesy of your focusTrap.

However, `self.currentField` is now set to `None` (tbh, I haven't checked this with a debugger but it has to be that way). Now, getting currentField set again requires the activation of another field. It's the only way.

So either that has to be addressed in the focusTrap, or `_addFocusCheck` and `self.currentField` have to be removed. Because as it currently stands, it's not calling `fn()` when you've clicked on the editor's background. **And thus, although the focus is still on the field, no shortcuts can be invoked.**

From my testing, removing the check works fine and it solves the bug with [Liresol/anki-custom-shortcuts](https://github.com/Liresol/anki-custom-shortcuts).
```python
    for row in cuts:
        if len(row) == 2:
            keys, fn = row
            #fn = self._addFocusCheck(fn)
        else:
            keys, fn, _ = row
        scut = QShortcut(QKeySequence(keys), self.widget, activated=fn)
```

PR #1870, opened 2022-05-16
Allow style tags in HTML editor

hgiesel:
Motivated by [this post](https://forums.ankiweb.net/t/revert-style-stripping-in-html-editor/19966/4)

The reason I stripped the style tags was because I don't think they would be compatible with card-preloading. But that was probably a bad decision and a case of premature optimization...

### DIFF
--- a/ts/editor/plain-text-input/remove-prohibited.ts
+++ b/ts/editor/plain-text-input/remove-prohibited.ts
@@ -11,7 +11,7 @@ function removeTag(element: HTMLElement, tagName: string): void {
     }
 }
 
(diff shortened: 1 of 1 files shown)


dae:
Thanks Henrik. Impossible to make the right call all the time, but as long as we're listening to feedback and can adjust when necessary, I think we'll be ok. :-)

dae:
It would break existing cards until the user figured out there was an option available to stop that from happening. As there's no strong technical reason for doing this at the moment, it's probably better we don't do it at all.

PR #1872, opened 2022-05-16
Fix graphs tooltip incorrect position

### LINKED ISSUE #1871: graphs tooltip shiftLeftAmount is incorrect on first render
## Description

When you first open the stats window and hover on the right-most bar, the tooltip is shown on the right of the mouse and extends to outside the page, leading to horizontal scroll bar appearing. 
This is especially visible on AnkiMobile where pressing on the bar shifts the whole page left-ward.

image

## Debugging

The bug happens here:
https://github.com/ankitects/anki/blob/a451fec5a1c0c594a5fa33b4b42e2a0fcb8b85cd/ts/graphs/Tooltip.svelte#L16-L18

The reactive statement is triggered when the tooltip data (`html` and `x`) is set. But at this point, the container hasn't rendered yet with updated `html`. So `shiftLeftAmount` is set using previous tooltip width, which is 0 when first shown. 

The `shiftLeftAmount` value is also incorrect when tooltip on different graph is first triggered, but not to a noticeable degree.

BlueGreenMagick:
Fixes #1871

I'm not happy with this PR, but I can't think of a better way to fix it

### DIFF
--- a/ts/graphs/Tooltip.svelte
+++ b/ts/graphs/Tooltip.svelte
@@ -3,6 +3,8 @@ Copyright: Ankitects Pty Ltd and contributors
 License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 -->
 <script lang="ts">
+    import { tick } from "svelte";
+
     export let html = "";
     export let x: number = 0;
     export let y: number = 0;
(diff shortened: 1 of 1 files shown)


dae:
The use of tick seems reasonable, as I gather it's cases like this that it was intended for. Ok on your end @hgiesel?

hgiesel:
Looks good to me. My instinct would have probably gone for [afterUpdate](https://svelte.dev/docs#run-time-svelte-afterupdate) instead of `tick`, but it doesn't matter so much.

dae:
Thanks BGM!

PR #1877, opened 2022-05-17
Add the reviewer_will_replay_recording hook

abdnh:
This adds the reviewer_will_replay_recording hook to inspect and modify recordings recorded by "Record Own Voice".

I want to use this in [Record Own Voice History](https://ankiweb.net/shared/info/1508039970) in the following way to reduce monkey-patching a bit:

```python
def on_replay_recording(path: str) -> str:
    if not path:
        path = get_most_recent_recording(mw.reviewer.card.id)
    return path


gui_hooks.reviewer_will_replay_recording.append(on_replay_recording)
```


dae:
Thanks Abdo, the use-case examples are always appreciated.

PR #1881, opened 2022-05-18
Notetype and Deck selector in Svelte

hgiesel:
This builds on #1691.

This is how they currently look:

Screenshot 2022-05-18 at 08 05 16

Screenshot 2022-05-18 at 08 05 37

I have made them visually distinct from the editing toolbar buttons, in a similar way to what I described [here](https://forums.ankiweb.net/t/position-of-the-tag-editor-in-2-1-50/16608/5?u=hengiesel).

They are already connected to the backend, but upon a change, they do not update the note yet.

As you can see, these pictures are taken in the Edit Current window. I wanted to test if we could put them there as well to update the note type of the card / change the deck, or otherwise just show the info to the user.

Basically expose some functionality to the user, that otherwise would only be available through a right-click on the rows in the browser, but only for the current note.

What do you think about it?


kleinerpirat:
Here are my 2 cents (mostly about visuals):

I like how the icons are consistent with the corresponding browser sidebar items. The dropdown for notetypes is also a great improvement, altough I'm missing a quick way to go to "Manage Notetypes" there.

### Card template button
I'd prefer "Templates" over "Cardtype", because A: you can edit **all** the cardtypes via this button and B: "type" is more of a tech-savy term which people without programming knowledge might not associate as well with the concept as they would with "template" (maybe that's just me though). The tooltip says "Customize Card Templates", which is a mix of the two.

### Design
The new elements don't invoke the same *clickable* feeling as buttons (except when they're hovered) and they feel a bit *detached* from the UI. The notetype and deck dropdowns have a border, but the buttons in-between don't. Personally, I would prefer a uniform style that looks more like a button.

Also, having them grayed out by default might cause confusion because that's usually associated with a *deactivated* state, whereas a button can have other hover effects which don't require a text color change.

hgiesel:
@kleinerpirat

Thanks for the feedback
 
> I'd prefer "Templates" over "Cardtype".

I don't have that strong of a preference here, it's mostly important for me that we settle on a term. Currently the terms "Cardtype", "Card type", "Template", "Card template" are used to describe the same object.

<s>I could imagine that cardtype is a bit clearer, as it's more than just _a template_, in fact it's up to 4 templates, and some meta data (like deck override). This could lead to confusing statements such as "A card template consists of a front and a back template".

Also cardtype could be easier for teaching, saying: "One note has multiple cards, one notetype has multiple cardtypes". </s>

However...

Looking at `proto/anki/notetypes.proto`, I noticed that indeed "Template" is used, and what in the UI is called front template is here called front format, whereas `proto/anki/cards.proto` card type is actually used to describe in which queue the card currently is (new/learning/review). So I guess you're right here.

> I'm missing a quick way to go to "Manage Notetypes"

I could add it as a menu item in the dropdown. Although I also have a few reservations here:

- In the future, I would like to display most windows that originate in the editor as JS modals on top of the editor, this includes the ChangeNotetypePage, and also a future Fields and Templates window. I don't think this could ever work well with the Manage Notetypes window, as it would probably require nested modals (which are just awkward in JS).
- I think it's a bit of an awkward operational path: "Reviewer (that currently displays a card) -> Editor (that shows the note associated with the card: makes sense) -> Manage Notetypes (which show all notetypes, and allows you to edit unrelated notetypes: huh?)". I'm aware that could use it to add a new notetype, and then add notes to that, but then again, this screen does not allow to edit the templates or fields of unrelated notetypes. There probably is a reason this is not allowed (@dae ?)
- _Logically_, it would make more sense to me, if pressing "Control+Shift+N" (the Manage Notetypes shortcut) while in the editor would trickle down to the main window, and then open the Manage Notetypes window from there. This version of the Manage Notetypes window also has all the functionality.

> don't invoke the same clickable feeling [...] feel a bit detached from the UI

They indeed are supposed to feel _a tad_ detached, as they basically frame the editor. I guess in the degree of attachment, they are supposed to be similar to the toolbar on top of the reviewer (containing "Decks", "Add", "Browse", etc.).

I agree with you on the clickable feeling. They are too greyed out at the moment. I also don't like how they can turn grey again, even though they have a dropdown attached.

The notetype and deck buttons having borders is supposed to communicate that they are editable. I thought it would be weird if you had all these buttons next to each, and some opened a modal, and others were editable inline. I'm open for suggestions on how to communicate that better.

dae:
Nice to see this coming along! @kleinerpirat offered some good feedback which mostly echos my own thoughts.

I like the fields/templates/history options. Add Note may need some emphasis to make it more prominent/clear that it's the primary button on this screen; perhaps it could be styled like 'save' in the deck options? If we used that split button/arrow approach, we could potentially bury the history in the dropdown to save a bit of horizontal space, but no strong feelings there - just throwing out an idea.

These buttons look good on the desktop, but will likely needs some tweaks to work on phones - currently they just disappear off the right side of the screen if the window is made narrower.

>  this screen does not allow to edit the templates or fields of unrelated notetypes.

It was designed to be invoked from the editing screen and manage notetype screens where there was a "current" notetype, and adding a selector for different notetypes would have required refactoring, and taken up screen space.

Nested modals does sound problematic. But we could display a 'manage notetypes' button that sent a bridge command, which in turn activated the dialog from the main window. That also allows us to delegate control to each individual client, as the mobile clients will likely need to take a different approach (pushing a new window on top).

> The notetype and deck buttons having borders is supposed to communicate that they are editable.

Two suggestions for the completion:

- a case-insensitive comparison is needed
- when no matches are found, currently the dropdown disappears, making it look like arbitrary text input is allowed. Maybe a single "(no matches found)"  line or similar could be shown instead?

Even with the latter, it still feels a little strange to be editing text in that area. It made sense with the tag input area, because users are allowed to input arbitrary text that may not match anything that has come before, but in the top area, we require the user to choose from an available option.

What if instead of entering text in that area directly, it was a more normal-looking button that when pressed, shifted focus into a text input field at the top of the pop-up, similar to how "filter:" is shown in the existing modal dialog? That would allow for more normal-looking buttons, and makes it clearer that the text you type filters the available results, and is not used directly.

qt/aqt/addcards.py
```diff
@@ -33,21 +33,16 @@
 )
 
 
-class AddCards(QMainWindow):
+class AddCards(QDialog):
```
hgiesel, line 36: This is probably a problem (?)
I think I remember discussions about whether EditCurrent should also become a QMainWindow instead of a QDialog? Is the only difference the menu bar? However I don't see any other menu items in the AddCards menu bar.

dae: #980 describes why we changed it to a main window - it resulted in conflicting shortcuts. I presume we could make EditCurrent a main window in the future too, there just wasn't a pressing need or the change, as it didn't have field/notetype selectors.

dae:
I'm not so sure about that expanding/contracting toolbar - it causes all the fields to move up and down as you navigate (eg enter tags, change deck, switch to another window), which is a bit distracting.

hgiesel:
I agree. I was thinking about how to save that space if you're just browsing and not editing, but I also think that the the layout shift is too distracting. Currently thinking about alternatives or just dropping it.

kleinerpirat:
I too think the expanding/collapsing toolbar is too distracting. Also, there are add-ons with buttons that do not require a field to be focused - they would lose functionality.

> Add Note may need some emphasis to make it more prominent/clear that it's the primary button on this screen; perhaps it could be styled like 'save' in the deck options?

Agreed. Perhaps it also should be fixed to a corner like it used to be. It currently wraps with the other buttons, which also makes it feel less significant than it actually is.

glutanimate:
> because they are very rarely used

Have to disagree with this notion. I definitely think there's value in revisiting long-standing UI elements and re-evaluating whether they're still necessary. However, I think there is a sizable portion of the user base that actually switches between note types and decks fairly frequently. [*Quick note and deck buttons*](https://ankiweb.net/shared/info/2181333594) used to be one of the most popular add-ons on 2.0 and its [2.1 fork](https://ankiweb.net/shared/info/1331140127) is still fairly popular. Impossible to say what the download counts are, but with the ratings on both listings merged, the add-on would likely be in the ≈ top 20 on AnkiWeb.

The add-on aside, I think it's reasonable to say that even users new to Anki will want to e.g. switch between basic and cloze fairly frequently and/or regularly create a new deck to add cards to. With the buttons present in the editor, the UX paths to these actions are fairly obvious. If we bury them in a nested menu structure, relegated to the top of the screen and thus completely removed from the AddCards window, that does quite a bit of harm to the discoverability of these features. The few dozen pixels of screen space and decluttering we gain from this (especially in the revamped design above), simply doesn't seem worth it to me.

Also looking at how the UX flow would change for existing users: There is a significant difference between moving your mouse a few pixels from text entry to the *Fields* or *Decks* button and having to move it all the way up the screen, make sure the window is actually in focus as the top bar menu will otherwise show the wrong options, then dig through a menu to switch note types / decks.

Last, moving the options to the menu bar, users would also no longer be able to tell at a glance which deck cards are being added to (and to a lesser degree, which note type is in use in case of note types with similar field set-ups).

kleinerpirat:
> Also, to be honest, if the user chose a DE where the watchacallit menu is located far from the window, it probably means it's ok for them. If not, they would just have changed their DE.

@glutanimate probably spoke for the Mac userbase there. Considering the average Mac user doesn't even know what a folder hiearchy is (because the default Finder view doesn't show it), it's pretty safe to say most of them don't choose their UI deliberately, they just use what Apple provides them with :)

I'm afraid the Apple menu bar is a major show stopper for your idea @TheBlackBeans.

ijgnd:
@TheBlackBeans: 

> I used to think "forcing" users to learn basic shortcuts wasn't such a good design, but then it happened on Google Drive (for copy / cut / paste shortcuts, when you actually right click > copy / cut / paste, it will open a pop-up saying "use these shortcuts instead: ..."), and it has been like this for some years now,

I hope that websites can't just access my clipboard.

Google Docs uses a custom context menu so clicking their paste entry should be an untrusted operation from the view of the browser. So showing you an info box that tells you to use "Ctrl+v" is a stopgap/workaround to me. According to [this post](https://stackoverflow.com/a/41832982) this pop-up is only shown if you don't have a google extension installed in your browser. 

In Microsoft Word online (at onedrive.live.com) you get the same info box after selecting paste from the custom context menu. Though the Microsoft pop up also tells you that it's shown because the browser can't access the clipboard.

This pop-up solution is not used in other google products like mail. In GMail I have the built-in browser context menu (with a trusted paste option) that works as expected without falling back on such a pop-up. If the Google Docs solution were a deliberate strategy I would expect to see the popup window in other Google products.

So I would suggest to not adopt this strategy to Anki.

dae:

PR #1884, opened 2022-05-22
Add getNoteId to exports

hgiesel:
I'm currently updating my Closet add-on to work with current versions, which is where I noticed this oversight. It would be nice if this could make it into 2.1.53.


abdnh:
I just hit a situation where I needed this too. It would be nice also to expose a way to get the field's ordinal. I posted a question about this on the forums: https://forums.ankiweb.net/t/how-to-get-the-ordinal-of-the-current-editor-field-from-javascript/20195

dae:
Will defer to Henrik on that one

PR #1885, opened 2022-05-23
workaround on the ibus first character issue

wisherhxl:
I'm using ibus with candidate characters, when adding cards, the first keystroke goes double. For example, if I type "a" it becomes "aa".

`placeCaretAfterContent(richText);` When `richText` is empty will cause this problem when adding cards.

Since when it is empty, there is no need to place the caret. I think add a check can solve this problem.


wisherhxl:
wrong code.

PR #1886, opened 2022-05-23
workaround on ibus first character twice issue

wisherhxl:
I'm using ibus-rime which has candidate characters, when adding cards, the first keystroke goes double. For example, if I type "a" it becomes "aa".

This problem exists in many Linux operating systems, I've checked in Ubuntu 22.0, Arch Linux, with multiple Anki versions.

call `placeCaretAfterContent(richText);` When `richText` is empty will cause this problem, I found it when adding cards.

Since when `richText` it is empty, there is no need to place the caret, I add the check as a workaround.

### DIFF
--- a/ts/editor/rich-text-input/RichTextInput.svelte
+++ b/ts/editor/rich-text-input/RichTextInput.svelte
@@ -72,6 +72,14 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     async function moveCaretToEnd(): Promise<void> {
         const richText = await richTextPromise;
+        if (richText.textContent?.length === 0) {
+            // Calling this method when richText is empty will cause the first keystroke of
+            // ibus-based input methods with candidates to go double. For example, if you
+            // type "a" it becomes "aa". This problem exists in many linux distributions.
+            // When richText is empty, there is no need to place the caret, just return.
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


ts/editor/rich-text-input/RichTextInput.svelte
```diff
@@ -72,7 +72,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     async function moveCaretToEnd(): Promise<void> {
         const richText = await richTextPromise;
-        placeCaretAfterContent(richText);
+        if (richText.textContent) {
+            placeCaretAfterContent(richText);
+        }
```
hgiesel, line 77: This looks good, but it will be hard to understand it later on.
How about something like this:

```typescript
if (richText.textContent.length === 0) {
    // your explanation why this is necessary
    return;
}

placeCaretAfterContent(richText);
```

hgiesel (changes requested):

ts/editor/rich-text-input/RichTextInput.svelte
wisherhxl: Revised, is this ok?

dae: Looks good - thank you for the fix!

PR #1887, opened 2022-05-23
Add TTS flag `skip_parenthesis`

RumovZ:
https://forums.ankiweb.net/t/modify-string-before-tts-reads-the-field-advanced/20075/

Arbitrary replacements (with regex I presume) would require some work, because the parser would need to be taught escaping. But even if it was implemented, this use case seems common enough to support a simpler solution.

### DIFF
--- a/rslib/src/card_rendering/mod.rs
+++ b/rslib/src/card_rendering/mod.rs
@@ -72,6 +72,7 @@ struct TtsDirective<'a> {
     voices: Vec<&'a str>,
     speed: f32,
     blank: Option<&'a str>,
+    skip_parenthesis: bool,
     options: HashMap<&'a str, &'a str>,
 }
 
@@ -113,7 +114,11 @@ mod test {
     fn av_extracting() {
         let tr = I18n::template_only();
         let (txt, tags) = extract_av_tags(
-            "foo [sound:bar.mp3] baz [anki:tts lang=en_US][...][/anki:tts]",
+            concat!(
+                "foo [sound:bar.mp3] baz ",
+                "[anki:tts lang=en_US skip_parenthesis=true cloze_blank=bar]",
+                "(foo)[...][/anki:tts]"
+            ),
             true,
             &tr,
         );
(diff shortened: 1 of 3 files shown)


dae:
Have there been cases in the past where this was requested that I've forgotten about, or do we only have that one instance in the link above so far? It will be difficult to drop support once added, so I'd like to be sure there's enough demand for it before pulling the trigger here, and at least in this case it does appear that field splitting might be an alternative solution.

RumovZ:
Here is a similar request: https://forums.ankiweb.net/t/awesome-tts-configuration/4597 (Awesome TTS).
I think "noun (gender initial)" is a common pattern, even though people would ideally use two separate fields.
May also be useful in other situations: https://forums.ankiweb.net/t/tts-plays-despite-being-commented-out/15954

RumovZ:
I would only support round brackets. It's only meant to enable one of the most common use cases (presumably) without the use of a more complex feature, which would enable arbitrary replacements in the future.

dae:
I think we'd need nom for that - without back references in the Rust regex lib, I don't think we could exclude unbalanced inner parens. And now that I think about it, the contents of HTML comments inside TTS text are stripped, so that's probably already an option for users using the new-style [anki:tts].

I agree that 'word (gender)' is something that that can be seen in a number of shared decks, but so far we have only one person actually requesting that we omit '(gender)' from TTS (if we don't count the person who wanted us to omit '[gender]'). Based on how infrequently this comes up, it makes me wonder how many users would actually take advantage of this if we added it, and whether it would be worth the (admittedly not particularly large) maintenance burden.

It's not so much about this particular feature as it is the general principal, but I realise I too am guilty of thinking "that's a good idea" and adding features even when there hasn't been much demand shown :sweat_smile: If you'd still like to go ahead with this, perhaps it might be worth changing the flag name slightly, to eg "skip_parentheses" or "skip_parens" or "skip_round_brackets".

RumovZ:
I'll just leave it then. Maybe when the new TTS syntax is officially documented, there'll be a clearer picture of demand for this features or the lack thereof.

dae:
Thanks for understanding; let's leave this around as a reminder.

dae:
I don't think we've seen enough requests for this at this point, so let's close this for now.

PR #1890, opened 2022-05-27
rust impl stdError trait

dobefore:
get AbkiError as from error using crate `thiserror`,error appears
```
#[derive(Error, Debug)]
pub enum ApplicationError {
  #[error("Anki lib error: {0}")]
    AnkiError(#[from] anki::error::AnkiError ),
}
```
```
doesn't satisfy `AnkiError: StdError`
```

this is a rather simple fix.

orginal forum post https://forums.ankiweb.net/t/rust-error-handle-get-ankierror-as-a-from-error/16703


dae:
Thanks!

PR #1891, opened 2022-05-27
add addons_dialog_will_toggle_enable_addons hook

BlueGreenMagick:
This adds `addons_dialog_will_toggle_enable_addons` hook.

I'm developing an add-on that adds functionality to note types, and I'd like to use this hook to remove the script from note types when the add-on is disabled.

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -874,8 +874,12 @@ class AddonsDialog(QDialog):
         return self.addons[idxs[0]]
 
     def onToggleEnabled(self) -> None:
-        for module in self.selectedAddons():
+        selected = self.selectedAddons()
+        gui_hooks.addons_dialog_will_toggle_enable_addons(self, selected)
+        for module in selected:
             self.mgr.toggleEnabled(module)
+        if not selected:
+            return
         self._require_restart = True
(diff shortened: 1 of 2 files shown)


BlueGreenMagick:
For context, the addon I'm making implements Bionic Reading, **wh**ich **bo**lds **t**he **begg**ining **pa**rt **o**f **a**ll **wo**rds.

What I was trying do do is to sell an illusion that the add-on is actively bolding the words, even though what it's actually doing is adding a script to the note type (so it works on all devices) 

As such, the addon would automatically add the script on install, and remove it on disable/delete. And to the user, it would seem like it's the addon doing it and not the note type.

Do you think such illusion would be more confusing to the user? It's unavoidable that the illusion breaks on some situations (deleting addon manually, seeing script tag when editing note types), so maybe it's trading big loss for small gain?

@Ankingmed Maybe you have some inputs as well?

AnKingMed:
I believe the AMBOSS addon does this

https://ankiweb.net/shared/info/1044112126

AnKingMed:
That's probably the ideal situation

BlueGreenMagick:
Thanks for the feedbacks! I'll switch to that strategy

PR #1898, opened 2022-06-03
CSV import/export fixes and features

RumovZ:
The layout is currently broken, because the table can overflow horizontally. I guess we want a scrollbar instead? Do you have a pointer on how to achieve this with those flexbox containers for me?
Fields are truncated at 80 chars, which can lead to very different lengths when rendered. But truncated fields should take up all available space or it looks odd:
grafik
Should we set a small max-width for columns so there are always less than 80 characters visible and CSS can take care of showing ellipses?


dae:
Table looks great when importing with a small number of fields. We'll need to allow scrolling, as otherwise notetypes with lots of fields look ridiculous :-) I'll push some tweaks that get the scrollbar going; some HTML/CSS guru may know a better way to accomplish it. Agreed on capping columns to a certain length, which should be done now.

One downside of this approach is the user sees 'Column 1', 'Column 2' and so on in the header, instead of the actual names of the fields. It might be a bit nicer if the actual field names were used in the headers, so that the user could more easily confirm they've got things mapped correctly (currently they need to cross-reference the mapping area and preview area). We wouldn't be able to do it in the case when there's a notetype column, but in the single notetype case, perhaps the preview area could be passed the current map details, and replace the column names with the field names?

It might be nice to pass include_html into the preview generation code. We probably want the opposite behaviour here: if include_html is active, we strip the HTML tags so the user just sees the text, and if it's inactive, the HTML tags are shown as-is, indicating they'll be imported as text. Does that make sense?
2022-06-06T14:48:22,958977414+10:00
2022-06-06T13:27:09,820341648+10:00

dae:
Thoughts on moving the preview to its own section, and moving the other file options into Import Options?

dae:
(I used the new 'include notetype' functionality in exporting for this test, which was handy :-)

RumovZ:
Looks great, thanks!

> if include_html is active, we strip the HTML tags so the user just sees the text, and if it's inactive, the HTML tags are shown as-is, indicating they'll be imported as text. Does that make sense?

I also thought about that, but wasn't sure we should hide that much information. I assume we preserve file names and decode entities in the HTML case?

> Thoughts on moving the preview to its own section, and moving the other file options into Import Options?

My vision was that you'd first tweak the file options until the preview is rendered correctly, then use it to set the other options. The file options are a bit special because they're objective, universal and the only ones to influence the preview. Also, I didn't like the look of the table directly beneath a header.
Not all of these are strong reasons, but I rather like the current setup.

> perhaps the preview area could be passed the current map details, and replace the column names with the field names?

Wouldn't they need to cross-reference anyway, because they'd look at the preview, scroll down and remap a field, then scroll up again and ascertain the mapping is correct?
What would be shown if a column is mapped to none or several fields? Some files may also define their own column labels, making it hard for users to understand where the displayed label is coming from.
Earlier, you've suggested to use the first field of a column as its title to make mapping easier IIRC. What if we show that (or the entire respective preview column) as a tooltip for every select option? Then you'd have the relevant information right at hand when browsing the mapping options. For collapsed selectors, the tooltip for the current option would be shown to allow to quickly confirm its correct setting.
Hope that's understandable. I'd make a quick prototype, but I've failed to make tooltips work before.

ftl/core/exporting.ftl
```diff
@@ -38,3 +38,5 @@ exporting-processed-media-files =
         [one] Processed { $count } media file...
        *[other] Processed { $count } media files...
     }
+exporting-include-deck = Include deck
+exporting-include-notetype = Include notetype
```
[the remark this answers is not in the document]
dae: I think I slightly prefer 'include', but either works for me

dae:
> I assume we preserve file names and decode entities in the HTML case?

Perhaps we could reuse html_to_text_line(preserve=True)?

> Earlier, you've suggested to use the first field of a column as its title to make mapping easier IIRC. What if we show that (or the entire respective preview column) as a tooltip for every select option? Then you'd have the relevant information right at hand when browsing the mapping options. For collapsed selectors, the tooltip for the current option would be shown to allow to quickly confirm its correct setting.

Yep, some good points - maybe it's better done in the mapping area.

The only issue there is that users can't 'hover' on devices without a mouse, so we'd need an icon like the (i) in the deck options screen that they can tap on. That makes it somewhat cumbersome to confirm, which is why I thought it might be better to embed the first row of the preview into the actual dropdowns (eg instead of "Column 1", it could be "Column 1 (foo)" or even "1: foo"). That makes the matching process considerable easier, as they can think in terms of field names+column content, instead of field names+column numbers. WDYT?

RumovZ:
Makes sense, will do. 👍 

I guess people won't appreciate [5ece10a](https://github.com/ankitects/anki/pull/1898/commits/5ece10ad05f331be350aa27c29899127168fcc5b), as it makes working with the CSV plaintext more cumbersome, but I couldn't figure out a better way to fix that issue. Any ideas?

rslib/src/storage/deck/all_decks_of_search_notes.sql
```diff
@@ -0,0 +1,9 @@
+SELECT nid,
+  did
```
RumovZ, line 2: Would it make more sense to use the odid here? Currently, if you export notes in a filtered deck with deck names, the reimport will fail. (With a "deck missing" message; reporting that the specified deck is filtered instead would require some more work.)
I guess this usually doesn't matter because a deck column is only really useful if you want to add cards from outside Anki, but who knows.

dae: No strong preference on my end - it's a bit inconsistent with our apkg handling, but if it avoids a confusing error then perhaps it's the better choice.

RumovZ: > it's a bit inconsistent with our apkg handling

Good point. Maybe we should leave it for now and come back to it after there's been some real-world usage?

dae:
I've sent a fix for that upstream: https://github.com/BurntSushi/rust-csv/pull/283

I suspect it'll be quite some time until it gets merged in (if at all), but we could potentially use the fork in the mean time.

ftl/core/exporting.ftl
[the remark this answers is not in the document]
dae: deck name/notetype name works for me too. For GUID, maybe "unique identifier"?

ftl/core/importing.ftl
```diff
@@ -9,7 +9,6 @@ importing-appeared-twice-in-file = Appeared twice in file: { $val }
 importing-by-default-anki-will-detect-the = By default, Anki will detect the character between fields, such as a tab, comma, and so on. If Anki is detecting the character incorrectly, you can enter it here. Use \t to represent tab.
 importing-change = Change
 importing-colon = Colon
-importing-column = Column { $val }
```
RumovZ, line 12: This hasn't been in a previous release, so I think it's safe to remove?

ftl/core/exporting.ftl
```diff
@@ -38,5 +38,6 @@ exporting-processed-media-files =
         [one] Processed { $count } media file...
        *[other] Processed { $count } media files...
     }
-exporting-include-deck = Include deck
-exporting-include-notetype = Include notetype
+exporting-include-deck = Include deck name
+exporting-include-notetype = Include notetype name
+exporting-include-guid = Include unique note identifier
```
dae, line 43: Is note a little superfluous here, given that the context is a row/note? Not a big deal though.

dae:
This has turned out really nice @RumovZ! The preview works great, and should make things considerably easier for users. The GUID support is also great. Thanks for all your hard work :-)

dae:
> I've neglected the card export a bit, because I don't know the use cases for this.

It exists to provide users with complex notetypes a way to import the questions/answers into simpler flashcard apps that only support 2 fields. I'm not sure how much it's actually used in practice, so probably not worth putting much effort into it until we see demand.

dae:
(minor trivia: with this PR, Rust is finally shown first in the list of languages GitHub provides for the repo :tada:)

ftl/core/exporting.ftl
RumovZ: Yeah, guess you're right.

PR #1905, opened 2022-06-06
Update get_linux_dark_mode to use freedesktop.org standard

### LINKED ISSUE #1903: Detect dark mode setting on Linux with `org.freedesktop.appearance.color-scheme`
Thanks to f0fedf6ab8cf18aceedde0150422e43a43a0d108 (see #1116), there is some dark mode detection on Linux.

However, there is now a Free Desktop standard for detecting dark mode. Since GNOME 42 the current implementation is not even supported anymore, because said Free Desktop standard is being used instead. See: https://blogs.gnome.org/alexm/2021/10/04/dark-style-preference/

Other apps that have already implemented this standard:

- Telegram Desktop: https://github.com/telegramdesktop/tdesktop/issues/17061
- Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1734934

louwers:
Fixes #1903

Instead of relying on unreliable GNOME-specific dark mode detection (that does not work anymore in the most recent version of GNOME), use freedesktop.org standard (see link in comments in code).

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -102,7 +102,7 @@ Brayan Oliveira <github.com/BrayanDSO>
 Luka Warren <github.com/lukawarren>
 wisherhxl <wisherhxl@gmail.com>
 dobefore <1432338032@qq.com>
-Bart Louwers <bart.git@emeel.net>
+Bart Louwers <bart.git@emeel.net> 
 
 ********************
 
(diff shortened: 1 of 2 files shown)


dae:
Thanks for looking into this Bart!

Gnome 42 appears to be very recent, so I am concerned that this change will break dark mode detection on Linux distros from the past few years. Would it perhaps be better to add the new code above the previous approach, and fall back on gsettings if the color-scheme key can't be read?

louwers:
@dae Thank you for replying so quickly on this! 

Good point, I updated the PR. :+1:

dae:
Thanks for making those changes; looks good.

PR #1906, opened 2022-06-07
Fix JS drop event not firing in the reviewer

abdnh:
I noted in #1562 that the `drop` event is not firing in JS. I recently revisited the issue because I needed drag & drop to work on a template I'm designing. It turned out the issue is caused by `AnkiWebView.dropEvent`. I vaguely remember testing if removing it fixes the issue previously, but maybe I didn't...

You can confirm this works by trying the code you posted in the previous PR. I'm not sure if this will have unintended effects on other screens, but I didn't notice any issues after some brief testing. These two lines are too ancient that I had to work on my git-blame skills to find when they were originally introduced: they were [added in 2008](https://github.com/ankitects/anki/commit/7391cef4f2).

### DIFF
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -222,6 +222,8 @@ class WebContent:
 
 
 class AnkiWebView(QWebEngineView):
+    allow_drops = False
+
     def __init__(
         self,
         parent: Optional[QWidget] = None,
@@ -245,7 +247,6 @@ class AnkiWebView(QWebEngineView):
         self._disable_zoom = False
 
         self.resetHandlers()
-        self.allow_drops = False
         self._filterSet = False
         QShortcut(  # type: ignore
             QKeySequence("Esc"),
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -245,6 +245,7 @@ class AnkiWebView(QWebEngineView):
         self._disable_zoom = False
 
         self.resetHandlers()
+        self.allow_drops = False
         self._filterSet = False
         QShortcut(  # type: ignore
             QKeySequence("Esc"),
@@ -322,7 +323,8 @@ class AnkiWebView(QWebEngineView):
         m.popup(QCursor.pos())
 
     def dropEvent(self, evt: QDropEvent) -> None:
-        pass
+        if self.allow_drops:
+            super().dropEvent(evt)
 
     def setHtml(self, html: str) -> None:  #  type: ignore
         # discard any previous pending actions
@@ -331,6 +333,7 @@ class AnkiWebView(QWebEngineView):
         self._queueAction("setHtml", html)
         self.set_open_links_externally(True)
         self.setZoomFactor(1)
+        self.allow_drops = False
         self.show()
 
     def _setHtml(self, html: str) -> None:
@@ -357,6 +360,7 @@ class AnkiWebView(QWebEngineView):
     def load_url(self, url: QUrl) -> None:
         # allow queuing actions when loading url directly
         self._domDone = False
+        self.allow_drops = False
         super().load(url)
 
     def app_zoom_factor(self) -> float:
--- a/qt/aqt/browser/previewer.py
+++ b/qt/aqt/browser/previewer.py
@@ -132,6 +132,7 @@ class Previewer(QDialog):
             ],
             context=self,
         )
+        self._web.allow_drops = True
         self._web.set_bridge_command(self._on_bridge_cmd, self)
 
     def _on_bridge_cmd(self, cmd: str) -> Any:
--- a/qt/aqt/browser/previewer.py
+++ b/qt/aqt/browser/previewer.py
@@ -133,6 +133,7 @@ class Previewer(QDialog):
             context=self,
         )
         self._web.allow_drops = True
+        self._web.eval("_blockDefaultDragDropBehavior();")
         self._web.set_bridge_command(self._on_bridge_cmd, self)
 
     def _on_bridge_cmd(self, cmd: str) -> Any:
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -356,6 +356,7 @@ class CardLayout(QDialog):
             context=self,
         )
         self.preview_web.allow_drops = True
+        self.preview_web.eval("_blockDefaultDragDropBehavior();")
         self.preview_web.set_bridge_command(self._on_bridge_cmd, self)
 
         if self._isCloze():
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -355,6 +355,7 @@ class CardLayout(QDialog):
             ],
             context=self,
         )
+        self.preview_web.allow_drops = True
         self.preview_web.set_bridge_command(self._on_bridge_cmd, self)
 
         if self._isCloze():
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -313,6 +313,7 @@ class Reviewer:
             ],
             context=self,
         )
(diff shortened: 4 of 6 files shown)


dae:
I seem to recall that when using WebKit, dragging an image onto the review screen replaced the active page with the image, causing the reviewing interface to get lost. I think that method was added to prevent that. That no longer seems to happen in WebEngine, though it's still a bit odd - it opens the image in an external window (presumably due to acceptNavigationRequest()). Before removing the method, perhaps we could block drag/drop requests on the JavaScript/CSS end by default?

abdnh:
> it opens the image in an external window

I've pushed a change that should fix this in all webviews that I could test.

qt/aqt/webview.py
```diff
@@ -654,11 +651,12 @@ def hide_while_preserving_layout(self) -> None:
         self.setSizePolicy(sp)
         self.hide()
 
-    def inject_dynamic_style_and_show(self) -> None:
-        "Add dynamic styling, and reveal."
+    def inject_dynamic_css_js_and_show(self) -> None:
```
dae, line 654: This is a public method; renaming it could potentially break an add-on.

qt/aqt/data/web/js/webview.ts
```diff
@@ -18,3 +18,14 @@ document.addEventListener("keydown", function (evt: KeyboardEvent) {
...
+// prevent Qt's default handling of drag & drop
+document.addEventListener("dragover", function (evt: DragEvent) {
+    evt.preventDefault();
+});
+document.addEventListener("dragenter", function (evt: DragEvent) {
+    evt.preventDefault();
+});
+document.addEventListener("drop", function (evt: DragEvent) {
+    evt.preventDefault();
+});
```
dae, line 31: If a page decides it wants to opt-in to the default drag/drop behaviour (not prevent the default), how would it do it? Isn't it impossible to remove the registered event listeners without a handle to the anonymous function?

dae (commented):
Thanks for looking into this Abdo. This approach feels a little untidy - I think I was hoping it would be possible with a single line of CSS, but it looks like CSS properties don't control drop behaviour.

One alternative that might be worth considering - keeping the existing dropEvent() function, but having it check a flag on the webview, which when set, delegates to super(). That way a page could do web.allow_drops = True to enable the default functionality. We'd probably want to set it to False on a call to setHtml()/load_url(), so that moving from one screen in the main window to another resets the functionality. Screens/add-ons that want drops would set it to true after loading a page (eg in webview_will_set_content in the add-on case).  What do you think?

abdnh:
> keeping the existing dropEvent() function, but having it check a flag on the webview, which when set, delegates to super().

This solution is certainly a lot cleaner.

> That way a page could do web.allow_drops = True to enable the default functionality

So we would presumably set allow_drops = True in the reviewer and previewer to allow JS in templates to receive drop events like this?

```diff
--- a/qt/aqt/browser/previewer.py
+++ b/qt/aqt/browser/previewer.py
@@ -132,6 +132,7 @@ class Previewer(QDialog):
             ],
             context=self,
         )
+        self._web.allow_drops = True
         self._web.set_bridge_command(self._on_bridge_cmd, self)
 
     def _on_bridge_cmd(self, cmd: str) -> Any:
diff --git a/qt/aqt/clayout.py b/qt/aqt/clayout.py
index 5d56c1646..60582fdc8 100644
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -355,6 +355,7 @@ class CardLayout(QDialog):
             ],
             context=self,
         )
+        self.preview_web.allow_drops = True
         self.preview_web.set_bridge_command(self._on_bridge_cmd, self)
 
         if self._isCloze():
diff --git a/qt/aqt/reviewer.py b/qt/aqt/reviewer.py
index 3e223b83f..7256de38f 100644
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -313,6 +313,7 @@ class Reviewer:
             ],
             context=self,
         )
+        self.web.allow_drops = True
         # show answer / ease buttons
         self.bottom.web.show()
         self.bottom.web.stdHtml(
diff --git a/qt/aqt/webview.py b/qt/aqt/webview.py
index 4f24274aa..70a3b37cb 100644
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -245,6 +245,7 @@ class AnkiWebView(QWebEngineView):
         self._disable_zoom = False
 
         self.resetHandlers()
+        self.allow_drops = False
         self._filterSet = False
         QShortcut(  # type: ignore
             QKeySequence("Esc"),
@@ -322,7 +323,8 @@ class AnkiWebView(QWebEngineView):
         m.popup(QCursor.pos())
 
     def dropEvent(self, evt: QDropEvent) -> None:
-        pass
+        if self.allow_drops:
+            super().dropEvent(evt)
 
     def setHtml(self, html: str) -> None:  #  type: ignore
         # discard any previous pending actions
@@ -331,6 +333,7 @@ class AnkiWebView(QWebEngineView):
         self._queueAction("setHtml", html)
         self.set_open_links_externally(True)
         self.setZoomFactor(1)
+        self.allow_drops = False
         self.show()
 
     def _setHtml(self, html: str) -> None:
@@ -357,6 +360,7 @@ class AnkiWebView(QWebEngineView):
     def load_url(self, url: QUrl) -> None:
         # allow queuing actions when loading url directly
         self._domDone = False
+        self.allow_drops = False
         super().load(url)
 
     def app_zoom_factor(self) -> float:
```

One minor issue remains: the default behavior of image drag & drop will still be triggered in these screens if no event handler is registered. Not a big deal.

qt/aqt/data/web/js/webview.ts
abdnh: Yes, I think it's impossible. I wish there was a function to undo any .preventDefault() calls in other event handlers.

abdnh: Yes, I think it's impossible. I wish there was a function to undo any .preventDefault() calls in other event handlers.

abdnh:
I have no idea why mypy suddenly started throwing an unrelated error now.

dae:
It looks like a mypy bug - moving allow_drops to the class definition level appears to resolve it:

```diff
diff --git a/qt/aqt/webview.py b/qt/aqt/webview.py
index 70a3b37cb..f68b58926 100644
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -222,6 +222,8 @@ class WebContent:


 class AnkiWebView(QWebEngineView):
+    allow_drops = False
+
     def __init__(
         self,
         parent: Optional[QWidget] = None,
@@ -245,7 +247,6 @@ class AnkiWebView(QWebEngineView):
         self._disable_zoom = False

         self.resetHandlers()
-        self.allow_drops = False
         self._filterSet = False
         QShortcut(  # type: ignore
             QKeySequence("Esc"),
```

It appears I misunderstood your goal here - I assumed you were trying to do this in an add-on, not in a standard card template. The review screen is the screen users probably spend the most time, so ideally we'd fix the image drop issue there too.

I think this current approach is probably still worth keeping, as it's the least likely to introduce regressions on other screens. But to fix the review screen, maybe we could combine the allow_drops change with some reviewer-specific JS? If we set document.ondrop to a function that calls preventDefault(), reviewer code should be able to override it. Once any template does this, the dropping issue would come back, but at least most users would not see it.

abdnh:
I've pushed a change to fix the drop issue in all review screens. Template code can now restore the default behavior using something like this:

```js
document.ondragenter = null;
document.ondragover = null;
document.ondrop = null;
```

dae:
Great, thanks Abdo.

PR #1908, opened 2022-06-08
Add legacy importers to new import screen

RumovZ:
Doing this for the exporters as well would be more involved, with an increased chance that add-ons would break anyway. And as there aren't any native legacy exporters, I skipped it for now.
As a matter of fact, I couldn't find any reference to the `exporters_list_created` and `importing_importers` hooks.

### DIFF
--- a/qt/aqt/import_export/importing.py
+++ b/qt/aqt/import_export/importing.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import re
 from abc import ABC, abstractmethod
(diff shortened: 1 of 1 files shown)


dae:
It looks like there are a few add-ons that directly modify the older importing.Importers, so this at least provides them a (temporary) upgrade path if they switch over to using the hook. I tested this with the SuperMemo and Pauker importers and they worked well, and it's great that the new code path will be usable for most users now :tada:  (Special Fields users excluded)

PR #1912, opened 2022-06-15
Make field description a placeholder inside EditingArea

kleinerpirat:
In my personal Anki use, I feel like the feature introduced in #1476 isn't discoverable enough to be useful.
This PR based on forum user [Ammar's idea](https://forums.ankiweb.net/t/field-descriptions/10686/6?u=kleinerpirat) should improve the situation:

## Before
image
## After
image

It should also be more mobile-friendly.

Method
---
The description is added as a CSS variable which is then used by a pseudo-element:

```css
&:empty::after {
   content: var(--description);
   opacity: 0.4;
   cursor: text;
}
```


kleinerpirat:
@hgiesel Can you explain the purpose of this function?

https://github.com/ankitects/anki/blob/2be1f4c56da3cd4964f590dcfeded85340aa4d74/ts/editor/rich-text-input/transform.ts#L20-L24

Removing it doesn't seem to make a difference, but surely there was a good reason for it. If it is indispensable, I'll have to find a different method than `:empty` to conditionally show the placeholder.

hgiesel:
When users online only has inline content in a field (i.e. no block elements like `<div>` or `<ul>`), we want to make sure it's not accidentally displayed as block content in the templates.

A trailing `<br>` is not visible in the editor, but it will be during review, this is why it is trimmed, when written to the DB.

However if it would also be trimmed in the visual editor, the user would not be able to enter a newline in the editor. After a lot of playing around, I found this to be the safest way to avoid the problem mentioned in the second paragraph.

hgiesel:
Btw, I like the concept of this PR. It also mirrors how we did in the Mathjax editor.

kleinerpirat:
How do you feel about exposing the description in EditorFieldAPI? Maybe someone (me) wants to provide descriptions in multiple languages for his notetype via an add-on, or provide instructions on fields with a certain name, regardless of user settings.

dae:
Nice - thanks Matthias!

dae:
When I open the browse screen, I'm seeing the default text displayed underneath the contents of the first field, making it difficult to read. Do you see that too? Once I start editing it's fine; the issue is only on browser open.

kleinerpirat:
I can't reproduce this :/ Could you share a dev tools screenshot of the editable and its contents when it happens?

hgiesel:
I also noticed this, even in the EditCurrent window. It shows up as a ::before pseudo element on the anki-editable.

ts/editable/ContentEditable.svelte
```diff
@@ -33,10 +35,18 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+    $: descriptionCSSValue = `"${$description}"`;
+
+    let innerHTML = "";
+    $: empty = ["", "<br>"].includes(innerHTML);
 </script>
 
 <anki-editable
+    class:empty
     contenteditable="true"
+    bind:innerHTML
```
hgiesel, line 49: I'm pretty sure, this is a one-way bind: Changes to `innerHTML` within the ContentEditable component will be pushed to the element, however not the other way around.

kleinerpirat: Thanks, I thought the `:bind` directive would always create two-way bindings. With the CSS `:empty` pseudo-class this wasn't an issue, so I forgot to test this again with the new approach.

kleinerpirat: Is there an idiomatic way of doing this? I don't want to pollute the component with a vanilla JS two-way binding. Also, if there's an alternative to appending `<br>` to contenteditable in order to allow newlines without crashing, I would prefer the CSS (`:empty`) solution.

kleinerpirat:
@dae I think you might have to revert the PR for the 2.1.54 release, because I do not have the time to fix these issues right now.

hikaru-y:
Instead of binding to `innerHTML` or the CSS `:empty` method, I think one easy fix would be to subscribe to the `content` store of the `EditingAreaAPI` and toggle the `empty` class according to the value of the store. This way we don't have to worry about `<br>`.

@kleinerpirat  Would you mind if I submit a PR for this issue?

kleinerpirat:
@hikaru-y please go ahead and create a PR. I'm very thankful for your help! Once again you save the day 🥳👌🏻

PR #1918, opened 2022-06-20
Join RichTextAPI and RichTextContextAPI + Expose `anki/RichTextInput`

hgiesel:
After this PR, setting custom CSS will be very easy:

```ts
// This uses onMount, so it will only affect rich text inputs which are newly mounted.
// To affect already mounted rich text inputs, you need to iterate over `require(...).instances`.
require("anki/RichTextInput").lifecycle.onMount(async ({ customStyles }) => {
    const { addStyleTag } = await customStyles; 
    const { element } = await addStyleTag('myStyles');
    // element is a <style> element, which gives you access to a CSSStyleSheet
    element.sheet.addRule('anki-editable', 'color: red');
});
```


dae:
I presume you're happy for this to be included in the next rc?

hgiesel:
Yes, that would be nice. This does modify an external API, but I'm pretty quite sure this has not been used by add-ons before.

I personally want to use the functionality to restore [anki__editor__apply__font_color__background_color__custom_class__custom_style.](https://github.com/ijgnd/anki__editor__apply__font_color__background_color__custom_class__custom_style/pull/58) for 2.1.50+.

dae:
Thanks Henrik!

PR #1919, opened 2022-06-20
Fix field description not toggled properly

hikaru-y:
Follow-up to #1912 

### Issue
The field description is not properly toggled in the following cases:
1. when opening a *Browser* or *Edlit Current*, as mentioned in #1912
2. when editing HTML in the HTML editor.

### Proposed solution
This PR fixes the issue by toggling the `empty` class according to changes in the value of the  `content` store.

### DIFF
--- a/ts/editable/ContentEditable.svelte
+++ b/ts/editable/ContentEditable.svelte
@@ -39,14 +39,12 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     const description = getContext<Readable<string>>(descriptionKey);
     $: descriptionCSSValue = `"${$description}"`;
 
-    let innerHTML = "";
-    $: empty = ["", "<br>"].includes(innerHTML);
+    export let content: Writable<string>;
 </script>
 
 <anki-editable
-    class:empty
+    class:empty={!$content}
     contenteditable="true"
-    bind:innerHTML
     use:resolve
     use:setupFocusHandling
     use:preventBuiltinShortcuts
--- a/ts/editor/rich-text-input/RichTextInput.svelte
+++ b/ts/editor/rich-text-input/RichTextInput.svelte
@@ -148,6 +148,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                     mirrors: [mirror],
                     inputHandlers: [setupInputHandler, setupGlobalInputHandler],
                     api: api.editable,
+                    content,
                 },
                 context: allContexts,
             });


dae:
IIRC, editable is a dependency of editor, so editable can't access items from the editor. Maybe we could pass the context in as a property instead?

dae:
(thank you very much for looking into this, by the way!)

hikaru-y:
Thanks for the feedback. Please give me a few hours to change the code and test it before updating the PR.

hikaru-y:
Updated.

Instead of the `editingAreaContext` you suggested, I have passed `content` as a prop. If `editingAreaContext` is preferred, I will change it soon.

> editable is a dependency of editor, so editable can't access items from the editor.

BTW, is this an issue with the Typescript/Svelte test settings? Here on Win10, I successfully built and started Anki without any errors with the first code in this PR:
`import { context as editingAreaContext } from "../editor/EditingArea.svelte";`

dae:
Bazel's TypeScript support on Windows is not able to isolate the building of each module, so it is possible for one module to read files from another module, even if those dependencies haven't been declared. Clearing the build products will prevent some failures, but IIRC, references to source files will still work on Windows but not on other platforms, so the only way to be sure things are working is to run the Linux CI. These sort of issues are fairly rare, and clearing the cache is cumbersome, so I'd recommend you only do that if CI fails.

Thanks again for this Hikaru :-)

hgiesel:
I will have a look later today

kleinerpirat:
Thought [this comment](https://github.com/ankitects/anki/commit/3eb8a6ce9de7290c13748e0a6ecee2e0e83d2565#r76583693) would appear here automatically:

> `attachContentEditable` is called only once on mount with:
> 
> https://github.com/ankitects/anki/blob/3eb8a6ce9de7290c13748e0a6ecee2e0e83d2565/ts/editor/rich-text-input/RichTextInput.svelte#L197
> 
> ... so only the initial string value of `$content` will be passed to the component. If instead the store is passed as a prop, it will be reactive.

hikaru-y:
Sorry for introducing a serious bug. I rarely use shortcut keys when doing Anki, so I was not aware of the bug. I have been trying to find the cause of the bug and a fix for it since reading [the forum report](https://forums.ankiweb.net/t/selected-field-is-not-clearel/20923), and have finally found them now. I know the approach in the PR will not be adopted anymore, as #1923 has already been merged, but I'm leaving a note here in case someone tries a similar approach in the future and encounters the same problem.

### Cause of the bug
https://github.com/ankitects/anki/blob/a8e34ce419699b8e61a055173e1ffb411f3fa501/ts/sveltelib/dom-mirror.ts#L13
Due to this setting, when an `<anki-editable>` has focus, DOM mirroring is performed even when the `class` attribute is changed. When loading a new note in the editor, the respective values of the `fieldStores` are updated in the `setFields()`, which in turn updates the `class` of the `<anki-editable>`s, so that the field with focus is synchronized back from the corresponding field in the previous note and the bug occurs. In the forum report, he says that the problem does not occur when pressing the <kbd>Add</kbd> button or selecting a note with a mouse click in the browser. This is because those actions take the focus off all fields before `setFields()` is executed.

### Fix
Exclude `class` by listing only the attributes that need to be observed in the `attributeFilter` option.

```diff
--- a/ts/sveltelib/dom-mirror.ts
+++ b/ts/sveltelib/dom-mirror.ts
@@ -7,11 +7,17 @@ import { writable } from "svelte/store";
 import { on } from "../lib/events";
 import storeSubscribe from "./store-subscribe";

-const config = {
+/**
+ * Do not add 'class' to attributeFilter, otherwise updating the class
+ * attribute of <anki-editable> in response to changes in fieldStore will
+ * cause problems when changing notes.
+ */
+const config: MutationObserverInit = {
     childList: true,
     subtree: true,
     attributes: true,
     characterData: true,
+    attributeFilter: ["style"],
 };

 export type MirrorAction = (
```

dae:
No need to apologise; I did not notice the issue either, and I appreciate all the contributions you have made. :-)

PR #1923, opened 2022-06-21
Fix field description

hgiesel:
This is my solution :)

However there's one minor thing to keep in mind: Once we allow for e.g. all inputs to be collapsed by default, or have the HTML editor uncollapsed instead of the visual editor, the description will not be visible.

Fixes regressions introduced by #1912 and #1919

### DIFF
--- a/ts/editable/ContentEditable.svelte
+++ b/ts/editable/ContentEditable.svelte
@@ -7,11 +7,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 </script>
 
 <script lang="ts">
-    import { getContext } from "svelte";
-    import type { Readable, Writable } from "svelte/store";
+    import type { Writable } from "svelte/store";
 
     import { updateAllState } from "../components/WithState.svelte";
-    import { descriptionKey } from "../lib/context-keys";
     import actionList from "../sveltelib/action-list";
     import type { MirrorAction } from "../sveltelib/dom-mirror";
     import type { SetupInputHandlerAction } from "../sveltelib/input-handler";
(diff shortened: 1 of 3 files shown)


ts/editor/rich-text-input/RichTextInput.svelte
```diff
@@ -160,6 +171,15 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         $editingInputs = $editingInputs;
     }
 
+    function setFocus(): void {
+        $focusedInput = api;
+
+        // We do not unset focusedInput here.
+        // If we did, UI components for the input would react the store
+        // being unset, even though most likely it will be set to some other
+        // field right away.
```
hgiesel, line 180: The comment is unrelated to this PR, but should be helpful

dae:
Thanks Henrik, appreciate you looking into it. 

> the description will not be visible.

I presume that's only if the user had deliberately collapsed the fields? If so, that doesn't seem like a big deal.

ts/editable/ContentEditable.svelte
```diff
@@ -70,17 +61,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-        &.empty::before {
-            content: var(--description);
-            opacity: 0.4;
-            cursor: text;
-            /* stay on single line */
-            position: absolute;
-            max-width: 95%;
-            overflow-x: hidden;
-            white-space: nowrap;
-            text-overflow: ellipsis;
```
kleinerpirat, line 82: The reason I restricted the description width was to prevent the pseudo-element from expanding the editable. Your implementation doesn't expand it either, but you should still cut off the text-overflow:

image

ts/editor/rich-text-input/RichTextInput.svelte
```diff
@@ -208,7 +241,17 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+        margin: 6px;
+    }
+
+    .rich-text-placeholder {
+        position: absolute;
+        color: var(--disabled);
+
+        /* Adopts same size as the content editable element */
+        width: 100%;
+        height: 100%;
```
kleinerpirat, line 255: I recommend adding these:
```
overflow-x: hidden;
white-space: nowrap;
text-overflow: ellipsis;
```

kleinerpirat (commented):
This new implementation introduces a regression for long descriptions. I recommend fixing them with plain CSS.

PR #1925, opened 2022-06-22
Fix descriptions overflowing field

kleinerpirat:
Fixes the regression mentioned in https://github.com/ankitects/anki/pull/1923#pullrequestreview-1014690204.

### DIFF
--- a/ts/editor/rich-text-input/RichTextInput.svelte
+++ b/ts/editor/rich-text-input/RichTextInput.svelte
@@ -253,5 +253,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         /* Adopts same size as the content editable element */
         width: 100%;
         height: 100%;
+        /* Keep text on single line and hide overflow */
+        white-space: nowrap;
+        overflow-x: hidden;
+        text-overflow: ellipsis;
     }
 </style>


ts/editor/rich-text-input/RichTextInput.svelte
```diff
@@ -253,5 +253,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         /* Adopts same size as the content editable element */
         width: 100%;
         height: 100%;
+        /* Keep text on single line and hide overflow */
+        white-space: nowrap;
```
hgiesel, line 257: Why would we want to restrict the placeholder to a single line though?

kleinerpirat: If the description is super long and expands the field to multiple lines, it's an awkward experience when it collapses to one line when the user starts typing.

The input field in the field description is very short as well, already nudging users towards shorter descriptions.

The feature - at least how I understood it - is intended for short, concise descriptions.

hgiesel: I see. I didn't consider that the placeholder would actually make the editor field grow bigger than a single line...

kleinerpirat:
These placeholders should not be intrusive in my opinion. If they can affect field height, they are somewhat intrusive.
I see how cutting off long descriptions could be problematic for mobile use with limited horizontal space. Two options:

1. Set a maximum character length for newly set descriptions in the field dialog
(if users choose a big font-size, they might still overflow)
3. Allow to show full description (e.g. expand placeholder with a click on the ellipsis)

Might be over-engineering a non-issue though. If we personally do not have a use for long descriptions, we should probably listen for community feedback first.

hgiesel:
I agree we should wait for user feedback. Enough other things to fix :)

kleinerpirat:
We still got this situation, in case you overlooked it [here](https://github.com/ankitects/anki/pull/1923#discussion_r903399543):

image

... so either this PR or another fix has to get merged in for the next rc, otherwise we'll an easy-to-predict bug report in the forum :)

hgiesel:
Maybe there was a misunderstanding, but I'm all for merging this one :)

dae:
I'll get this in before the release, but there may not be another rc if nothing else comes up

dae:
Seems to work well, thank you both :-)

PR #1930, opened 2022-06-23
Restore and save last dupe resolution setting

RumovZ:
Is this the way to do it? If not, how do I access the profile meta from within Svelte?


dae:
I've made some tweaks to do the saving/loading in the Rust layer. This has a few advantages:
- the default setting syncs between devices, as it's stored in the collection instead of the profile
- it minimizes the extra work that is required to integrate this code into the mobile clients - any code we place in the Python layer needs to be reimplemented on each system

PR #1931, opened 2022-06-23
Refactor how toolbar buttons get to surround within editor fields

hgiesel:
I was fixing up the custom styles add-on (https://github.com/ijgnd/anki__editor__apply__font_color__background_color__custom_class__custom_style/pull/58), and this PR was created alongside with it.

If you look at a single button component (e.g. `BoldButton.svelte`), you will see there is way less boilerplate now.

Having way less subscribers to `focusedInput` also seemed to have sped up the NoteEditor, when having a lot of formatting buttons hooked up, e.g. look at this screenshot:

Screenshot 2022-06-23 at 22 51 14

Other changes:
- Adapt exports of directories wrapping components
- Decouple editor/surround from editor/rich-text-input.

This should be safe to merge for 2.1.54, but I could also wait.


dae:
This seems to have introduced a bug - when I press ctrl+b, all the buttons get pressed down at once.

ijgnd:
I might miss something but this commit seems to work for me: When I press ctrl+b I don't notice anything unusual, see this screencast: 

https://raw.githubusercontent.com/ijgnd/temp/main/works_for_me.mp4

hgiesel:
You need to press Ctrl+B without selecting any text. This triggers a different pathway (a hook is registered that formats on next text insert)

hgiesel:
Phew, I finally got to it. I did not need any hacks, in fact, I could remove a lot of `any`s and non-null assertions.

dae:
Thanks for pushing that through Henrik. I won't pretend to understand it all, but the code looks sane at first glance, and I can't reproduce the bug I encountered earlier. :rocket:

PR #1932, opened 2022-06-24
Disregard manual reschedulings in `introduced:x`

RumovZ:
https://forums.ankiweb.net/t/no-new-cards-being-shown-with-introduced-feature/20979

I think setting the due date of a new card would be considered as introduction by most people, but it also makes it harder to understand what's going on, because you can't easily distinguish it from forgetting. Not sure what's the best solution here.

### DIFF
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -578,8 +578,11 @@ impl SqlWriter<'_> {
         write!(
             self.sql,
             concat!(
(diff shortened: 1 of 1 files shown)


dae:
Not sure the performance hit of a perfect solution is worth it. This is an improvement already - thanks Rumo :-)

dae:
I'm not sure what the best answer is here :-) Some users might still be confused that it would go off the date they were rescheduled, instead of the date they were scheduled to appear?

PR #1933, opened 2022-06-24
Update typescript to 4.7

hgiesel:
You locked Typescript in 661f7697.
I could not reproduce the issue with bootstrap, maybe it has been solved?


hgiesel:
Unfortunately I cannot reproduce the build error locally, you have any idea @dae?

dae:
The build fails for me on both a Mac and Linux machine - maybe you need to clear out your Bazel cache with 'bazel clean --expunge'? And remove the disk cache if you manually turned that on in the config file. Presumably the issue is that the Svelte files are being compiled without value imports, and that is breaking the protobuf references (and may break other stuff too).

dae:
In current main, if we look at one of the generated typing files:

```
import { SvelteComponentTyped } from "svelte";
declare const __propDef: {
    props: {
        info: any;
    };
...
```

with this branch:

```
cat .bazel/out/k8-fastbuild/bin/ts/congrats/CongratsPage.svelte.d.ts
import { SvelteComponentTyped } from "svelte";
declare const __propDef: {
    props: {
        info: Scheduler.CongratsInfoResponse;
    };
...
```

It appears to have correctly identified the type, but not included the import line for it. Presumably the issue is related to the way protobufjs creates nested namespaces instead of exporting things at the root level - maybe they are not accessible unless imported as a value.

dae:
(are there specific changes that are required/would be nice to have that would make this a priority?)

hgiesel:
No required changes here, just many nice-to-haves (like `import { type };`

PR #1935, opened 2022-06-26
Fix Study Deck in Sway spawning tiled window instead of floating window

Yutsuten:
In Sway (a tiling wayland compositor), if creating a `QDialog` without a parent widget, a new tiled window is created instead of a floating window. Probably because Sway is picky about the window properties or whatever.

I had the [same problem in my add-on](https://github.com/Yutsuten/anki-lifedrain/pull/127), and fixed it by passing the `mw` variable to `QDialog`. And to fix this Study Deck window issue I did the same thing.

By the way the Study Deck window is that window that pops up when you press the `/` shortcut anywhere.

Hope this can be merged upstream, this bug was annoying me for a while. Thanks!


dae:
Thanks Mateus!

PR #1944, opened 2022-07-02
Add hook: overview_will_render_bottom

Yutsuten:
This is my first time adding a hook, pardon me if I did anything wrong here.

# Use case

In my add-on [Life Drain](https://ankiweb.net/shared/info/715575551) I add two buttons to the Overview screen. (screenshots [here](https://github.com/Yutsuten/anki-lifedrain/wiki/Deck-Settings))

- `Life Drain`: This is a configuration at a deck level. Not every deck have the same difficulty so I allowed customization per deck.
- `Recover`: Fully restore life.

For a long time I monkey patched `aqt.toolbar.BottomBar.draw` like [this](https://github.com/Yutsuten/anki-lifedrain/blob/53ff173c7de9118004f89ba6fd7112b03a7e32e2/src/main.py#L82-L107), but I think it would be better to have a hook for this.

Another thing that motivated me to open this Pull Request is that another add-on creator inadequately monkey patched `Overview._renderBottom` and caused add-on incompatibility.
- https://github.com/Yutsuten/anki-lifedrain/issues/129
- https://github.com/telotortium/anki-reset-limits/issues/5

# About the change

I call `gui_hooks.overview_will_render_bottom` inside `_renderBottom` with 2 arguments: `links` and `self._linkHandler`.

Manipulating `links` from an add-on is not a problem.

But `self._linkHandler` is a function. Inside it we only have a chain of `if-elif` and I see no easy way to handle new `url`s it if not monkey paching it.

So a simple usage would be like:

```python
def add_hello_button(links, link_handler):
    links.append(['H', 'hello', 'Click me!'])
    def custom_link_handler(url):
        if url == 'hello':
            print('Hello World!')
        return link_handler(url=url)
    return custom_link_handler

gui_hooks.overview_will_render_bottom.append(add_hello_button)
```

Not so beautiful but works. The best would be to refactor the link handling in my opinion, but that's not the objective of this Pull Request.

Any feedback is greatly appreciated.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -75,7 +75,6 @@ Meredith Derecho <meredithderecho@gmail.com>
 Daniel Wallgren <github.com/wallgrenen>
 Kerrick Staley <kerrick@kerrickstaley.com>
 Maksim Abramchuk <maximabramchuck@gmail.com>
-Mateus Etto <mateus.etto@gmail.com>
 Benjamin Kulnik <benjamin.kulnik@student.tuwien.ac.at>
 Shaun Ren <shaun.ren@linux.com>
 Ryan Greenblatt <greenblattryan@gmail.com>
@@ -105,6 +104,7 @@ wisherhxl <wisherhxl@gmail.com>
 dobefore <1432338032@qq.com>
 Bart Louwers <bart.git@emeel.net>
 Sam Penny <github.com/sam1penny>
+Mateus Etto <mateus.etto@gmail.com>
 
 ********************
 
--- a/qt/aqt/overview.py
+++ b/qt/aqt/overview.py
@@ -287,8 +287,8 @@ class Overview:
             links.append(["U", "unbury", tr.studying_unbury()])
         links.append(["", "description", tr.scheduling_description()])
         link_handler = gui_hooks.overview_will_render_bottom(
-            links,
             self._linkHandler,
+            links,
         )
         if not callable(link_handler):
             link_handler = self._linkHandler
--- a/qt/aqt/overview.py
+++ b/qt/aqt/overview.py
@@ -286,6 +286,12 @@ class Overview:
         if self.mw.col.sched.have_buried():
             links.append(["U", "unbury", tr.studying_unbury()])
         links.append(["", "description", tr.scheduling_description()])
+        link_handler = gui_hooks.overview_will_render_bottom(
+            links,
+            self._linkHandler,
+        )
+        if not callable(link_handler):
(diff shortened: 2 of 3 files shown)


Yutsuten:
My name is in the CONTRIBUTORS file, why the check is failing... :thinking:

dae:
When you originally added your name, you appear to have done so via the GitHub UI, so the commit was using this email: `Yutsuten@users.noreply.github.com`.

I'm happy to accept this change, but please be aware that it may be short-lived - the deck browser/overview and review screens will likely be rewritten in the next major update, and the separate upper/lower toolbars may even be removed in favour of a single page.

Yutsuten:
> When you originally added your name, you appear to have done so via the GitHub UI

I don't use the GitHub UI. Every change I made were done locally and then pushed. What I did using the GitHub UI was to fetch changes is `main`, and only that. Shouldn't affect anything.
My name is in the list in my branch and `main`, and it is with the correct e-mail. What I can do is to move my name in the CONTRIBUTORS file to see if the CI fixes itself.

> please be aware that it may be short-lived

I'm okay with that.

Yutsuten:
Sorry for the several commits. I think next time I should setup these check locally. By the way I'm unable to understand what the tests are complaining.

In the `genhooks_gui.py` when I put `list[list[str]]` it complains `Incompatible types in assignment (expression has type "Union[Callable[[str], bool], List[Any]]", variable has type "List[List[str]]")`

Then I changed to `Union[Callable[[str], bool], List[Any]]` as requested. Now I get `error: Name "Union" is not defined`.

Whats happening here...

dae:
Filters should take the value they return in the first argument. I suspect if you switch the order of the two args, it will allow you to simplify the type signature as well.

Yutsuten:
Running the tests locally with `bazel test qt:mypy` now.

> I suspect if you switch the order of the two args

Tried changing `overview.py` to

```python
link_handler = gui_hooks.overview_will_render_bottom(
    self._linkHandler,
    links,
)
```

And `genhooks_gui.py` to

```python
args=[
    "link_handler: Callable[[str], bool]",
    "links: list[list[str]]",
],
```

I'm getting

```
aqt/hooks_gen.py:3188: error: Incompatible types in assignment (expression has type "Union[Callable[[str], bool], List[Any]]", variable has type "Callable[[str], bool]")
```

Yutsuten:
I think its good to go now.

Yutsuten:
But there is too much useless commit in this PR, I'll open a new one with only one commit.

PR #1949, opened 2022-07-05
Skip card generation if fronts remain unchanged

### LINKED ISSUE #1945: Frozing field should be constant time
How to reproduce:
* tested on fe302a5d1b8ba2b8c398f22c28bdd67357c6f339 with no add-on
* take a collection, with a note type N with a lot of notes. In the example of my collection on www.milchior.fr/anki , take "Definition".
* open "add note"
* click on the button to freeze/unfreeze a field

Expected result:
* button change status immediately and allow to continue editing

Actual result:
* the action takes multiple seconds

Other test:

I tested on a clone of the Definition field, with 0 notes, and the change was instantaneous. So I'm convinced it's related to the number of notes and note the number of fields/card type. I suspect some sanity check iterates over each note, but in this case it's absolutely useless

RumovZ:
Closes #1945

### DIFF
--- a/rslib/src/notetype/mod.rs
+++ b/rslib/src/notetype/mod.rs
@@ -695,7 +695,7 @@ impl Collection {
                 original.config.sort_field_idx,
                 normalize,
             )?;
-            self.update_cards_for_changed_templates(notetype, original.templates.len())?;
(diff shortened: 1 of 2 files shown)


dae:
Thanks Rumo!

PR #1953, opened 2022-07-08
Template err improvements

RumovZ:
https://forums.ankiweb.net/t/template-overrides-are-not-mentioned-in-the-error-message-when-they-reference-a-non-existent-field/21245

I would also like to check the browser templates on save and export, but I'm not sure how this could work. The error prompt currently says 'See preview ...' and those templates don't have one. We could display the error messages in the regular preview window or embed it in the error prompt, but even then I'm afraid there would be users who get stuck, because they can't track down the *Browser Appearance* dialog.

I think in a future redesign of the cards dialog, the browser templates should be located more closely to the regular ones.


dae:
Thanks for taking care of these issues Rumo!

Issue #1669, opened 2022-02-15
Typo in "Card Types for Basic" screen


ErykPiasecki07:
## Introduction
Unless it is some inside joke I don't get, there's a typo in "Card Types for Basic" screen.

![Screen Shot 2022-02-15 at 11 01 13 AM](https://user-images.githubusercontent.com/59264707/154038711-c21ba484-0689-4e5f-ad8a-da6fe3996161.png)

If you can't spot it at the first glance, it's "exomple" instead of example.

## To reproduce

1. Open Anki desktop app
2. Click "Add" button
3. Select "Basic (type in the answer)" Type
4. Click "Cards..."
5. Here it is in the "Preview" window!

**If it is indeed unintended, let me know and I will prepare a fix.**

ErykPiasecki07:
Oh, didn't catch it anywhere in the docs at my initial research, my bad! Thanks for "explonation" ;)

PR #1954, opened 2022-07-08
Update clayout.py

andrew-field:
Spelling


abdnh:
This is intentional. Please see https://github.com/ankitects/anki/issues/1669

Issue #1834, opened 2022-05-03
Consider providing a way to bypass review limits on new cards


dae:
Associated forum thread:

https://forums.ankiweb.net/t/v3-scheduler-new-cards-and-custom-study-behavior-question-requests/13870

Two possible approaches spring to mind:

- A "new cards ignore review limit" checkbox
- A "minimum new cards" option, that lets you specify the number of new cards that should be shown even if the review limit has been reached. Such a use case was anticipated when writing the v3 scheduler, and a `new_per_day_minimum` field already exists in the deck config.

The latter is more flexible - while a bit more complicated, it allows users to control how many new cards they want to be introducing when they're near or at the daily review limit.

Would need to figure out whether this should simply offset the review limit (eg if set to 5, the new limit is capped to min(reviews_remaining+5, new_remaining)), or whether it should override the the configured new limit. If the user specifies 0 new cards a day, but a minimum of 5, what should we do?

Other things that need consideration:

- How this will interact with extending the daily limits with custom study. Currently the user can extend their new card limit and find nothing changes, because their new cards are capped by the review limit already. Could the decisions we make here perhaps address that issue at the same time?
- How this will fit with the per-deck limits proposed on https://forums.ankiweb.net/t/set-daily-limits-separately-from-the-other-deck-options/17511

Dolphyn5:
Some people (including myself) regard "new" and "review" as separate categories, so another way to resolve this issue would be to implement three separate daily limit settings:

- Maximum new cards per day
- Maximum review cards per day (not including new cards)
- Maximum total number of cards per day (new + review) OR (new + review + learning)

dae:
I don't think burying is relevant to this issue, as the daily limits are adjusted based only on cards that weren't buried. I can't reproduce the issue in your latest forum reply; if I test with your AnkiWeb collection and disable burying of new cards in that deck's options, all 35 new cards are included.

Specifying a separate total instead of having an offset for new cards is a potential alternative solution, though I'm not sure that's the best approach - it's more disruptive as it alters the way reviews work as well, and it wouldn't be possible to tell Anki you want eg. a minimum of 5 new cards per day regardless of how many reviews are due.

LordxRequiem:
[...]

PR #1955, opened 2022-07-08
Deck-specific Limits

RumovZ:
https://forums.ankiweb.net/t/set-daily-limits-separately-from-the-other-deck-options/17511

That was easier than expected.
The dialog started as a debug aid, but now I actually don't see any reason not to release it. Do you?


dae:
That was nice and easy :-) Thanks for looking into this Rumo.

The dialog is a useful starting point, so no objections to keeping it until we have something better in place, but I think we'll probably want to improve on it in a separate PR before a new release. First some minor quibbles with it as it stands:

- it's not possible to override one without overriding the other
- one needs to click the checkbox and then enter a number - maybe we could avoid the separate checkbox by having an "unset" value in the input area instead? eg instead of a spinner, we could use a text box that allowed blank text, or allow -1 to represent "off"?

More importantly, I think it's going to be hard for users to see how this relates to the settings in the deck options at the moment, and it doesn't support the existing workflow where users adjust the daily limits via the deck options while in the review screen/congrats screen.

Could we perhaps extend the deck options screen to support customizing the current deck's limits as well? That way the user would be able to see at a glance what the defaults and overridden values were, and it would avoid users wondering why their chosen deck preset values were not being honored.

For this PR, there are two things I'd like to see done:

- Checking to make sure these settings don't get exported in an apkg with scheduling disabled (both old and new exporters), as that will cause a lot of confusion. We won't be able to fully avoid this though, as a user could sync to an older client, and export from there, which is another argument for why the defaults+overriden values need to be visible in the same place.
- The clear_other_duplicates() approach used in deckconfig.rs will also need to be used here

dae:
(Thinking about how it would fit into the deck options screen, maybe we would want to keep a checkbox, but have one for each item?)

dae:
I was looking at https://github.com/ankitects/anki/issues/1834 and it reminded me that we previously discussed single-day overrides as an alternative way of implementing custom study. What if instead of a checkbox, it was a group box with "use deck preset", "override today", "override permanently"? I still agree with you about preserving the permanent override when the user makes a temporary one, so we'd need some extra protobuf fields, eg a msg with (today, temporary_limit). That would allow us to remove the limit extension from the custom study screen, making custom study solely about filtered decks, and keeping all limit-modifying options in one place. WDYT?

RumovZ:
I thought these settings would finally replace the config limits, but it sounds like you intend to keep around config limits as well as deck limits? Then I'd agree that the flexibility and ergonomics of disabling these should be improved.

I'm not convinced they should be added to the other deck options. Some users already seem to have a hard time wrapping their heads around presets. Conflating the concepts of decks and their configs further will likely increase confusion.
Also, when the reviewer gets reworked and deck limits integrated more prettily and accessibly, the separate dialog can easily be removed again.
I understand your concerns, though. What if we add a warning instead? Something like "This deck has a limit configured overriding this value: { $val }". That would have the benefits you mention above, while still keeping deck and config settings separated.
Of course, I can add the dialog to the congrats, deck overview and study screens, too. Until now, I just didn't bother because I thought it's only a stopgap anyway.

Removing limit extensions from the custom study screen still sounds like a good idea. I wonder, maybe if modifying deck limits is really effortless, like hovering and scrolling / tapping and swiping without changing screen or opening a dialog, maybe we can even do entirely without it? That would still lack the automatic reset, but also simplify things for users. Three sets of limits might be overdoing it.

RumovZ:
I'm not sure if I've fully grasped the issue of serialization when moving between versions, so please let me know if I'm wrong, but don't we also need to check the `other` hashmap for the new settings when deserialising a deck? Otherwise, it seems to me that settings will be lost when switching to an older client, that has the same schema but not the new pb fields, and back.

dae:
> but it sounds like you intend to keep around config limits as well as deck limits

I think we probably at least need to keep them around until the majority of users are on an Anki version that supports the per-deck limits. I also think there's an argument for keeping them regardless, as they allow users to control what the default values of new decks will be, and will allow users to change the limits for a bunch of decks at once if they have not customized them individually. I do appreciate it makes things more complicated though. If we do decide to drop them in the future, it would require a schema upgrade that applies the current config values to each deck.

> I'm not convinced they should be added to the other deck options. Some users already seem to have a hard time wrapping their heads around presets. Conflating the concepts of decks and their configs further will likely increase confusion.

My main concern is that if the config limits and deck limits are not shown in one place, it will be confusing for the user. If some future iteration of the decks list allows users to see/customize the config limits as well as the deck limits, that would partially solve the problem, but if we're only showing config limits in the deck options, then there's still the potential for users to wonder "why aren't my settings being honored?"

Showing the deck limits as a read-only value in deck options, and config limits as a read-only value in the deck list would reduce confusion, though it's not very convenient for the user to have to jump between screens to change the two - wouldn't it be better to allow them to control everything in one place? The approach I'd originally been thinking of is that the deck list would provide some read-only visibility into the limits that are being applied, and users could open the opens if they wanted to change them.

> I wonder, maybe if modifying deck limits is really effortless, like hovering and scrolling / tapping and swiping without changing screen or opening a dialog, maybe we can even do entirely without it? Three sets of limits might be overdoing it.

For extending the limits of a single deck, it wouldn't be too hard for the user to adjust the settings again the next day. But if they do that for multiple decks in a session, or in bulk (eg they want to override to the limits of every subdeck via something like the existing "save to subdecks", then study the top deck), it starts to become a bit unreasonable to expect them to fix the limits the next day, especially if each of their decks had a different value previously. Three limits does seem a bit much, but maybe it's unavoidable?

> I'm not sure if I've fully grasped the issue of serialization when moving between versions, so please let me know if I'm wrong, but don't we also need to check the other hashmap for the new settings when deserialising a deck? Otherwise, it seems to me that settings will be lost when switching to an older client, that has the same schema but not the new pb fields, and back.

You are correct, any options preserved via 'downgrade and quit' would not be applied when updating to a new version with the same schema until the user used 'downgrade and quit' or did something else that downgrades the collection to v11, such as a full sync upload. Not sure how much of a problem it in is practice (don't recall anyone complaining so far :-). clear_other_duplicates() is more about preventing an error in older clients than it is about preserving the values.

RumovZ:
Isn't one of the goals of a future UI overhaul to reduce the number of dialogs? This would make getting to a specific part of the software simpler and more convenient. Complex and less often used dialogs like the deck options may still make sense, but I would hope regularly accessed options like deck limits would somehow get embedded in the main screen.
I don't think splitting the limits is really less convenient. Either you want to adjust deck limits or you want to adjust config limits. As they don't interact, I can't think of a use case where you'd want to adjust them simultaneously.

How would you avoid the confusion of having deck-level settings inside the preset widget?

I guess you're right about needing all three sets of limits. (Although dropping config limits would not necessarily require a schema upgrade. They could just be hidden from the UI and the code would continue to use them where no other limits apply.)
A combobox allowing to switch between them sounds good then.

dae:
> How would you avoid the confusion of having deck-level settings inside the preset widget?

I was thinking of something like:

New limit: xxx
[Preset limit/Deck-specific limit/Limit for today] 

Review limit: yyy
[Preset limit/Deck-specific limit/Limit for today] 

Initially I was thinking a dropdown, but maybe a group of radio options would be faster for the user to select?

I'm not strongly set on this idea, it just seems like it might be the best for now:

1) it continues to allow users to adjust the limits of multiple decks at once - maybe they want to vary each deck's new cards for example, but have all decks have a review limit of 9999
2) it would be added to an existing screen, minimizing the amount of immediate work required to expose this functionality to users on desktop and mobile clients
3) we could use the 'save to all subdecks' option to allow users to set the per-deck limits to subdecks

Putting these settings in the deck list in the future seems reasonable, and we could potentially remove the preset limits if the UI made it easy enough for users to apply changes in bulk (eg, applying review limits to all decks or all child decks, while preserving new cards. And maybe also providing a way to set defaults for new decks). But since that part of the UI will likely take some time, maybe it's better left for now?

RumovZ:
Ok, I'll implement that. Mobile clients and 'save to all subdecks' are good points.
Maybe when this is released someone will provide the ultimate solution for a new UI. 🙂

RumovZ:
> New limit: xxx
    [Preset limit/Deck-specific limit/Limit for today]

I'm struggling with this. An enum selector doesn't fit on the same row, not to mention a radio button group, and it looks too detached on the next one. And if I put it on its own row regardless, it kind of needs a label, because evey other row has one and we might need to attach a tooltip.
Maybe we could use a tab widget instead? That would also allow to preserve state for all three limits. That makes it not only more convenient to operate, it would even allow to change all limits in one go, whereas other solution would require closing and reopening the dialog each time.

dae:
Bit hard to visualize; happy to give it a try and see how that works out.

> That would also allow to preserve state for all three limits. That makes it not only more convenient to operate, it would even allow to change all limits in one go, whereas other solution would require closing and reopening the dialog each time.

One nice thing about using the currently selected tab to denote "active" is that we can use regular spinboxes - we wouldn't need to allow blank entries or a separate checkbox to toggle temporary/deck limits on/off. And since only one can apply at a given time, would there be any need for the user to adjust the limits in one go?

But I'm still thinking in radio button terms, and was thinking there'd be two sets of tabs, one for new and one for review. If you were thinking of combining them, then we'd need some other approach to allow eg. preset new limits and per-deck review limits.

Might be harder to see at a glance which settings are active when done with a single set of tabs?

dae:
Another option would be to show all 6 settings on the one screen without tabs, and revert to allowing deck-specific and day-specific settings to be blanked out or activated with a checkbox - it would take up more room, but the settings would be visible at a glance.

RumovZ:
> New limit: xxx
    [Preset limit/Deck-specific limit/Limit for today]

I'm struggling with this. An enum selector doesn't fit on the same row, not to mention a radio button group, and it looks too detached on the next one. And if I put it on its own row regardless, it kind of needs a label, because evey other row has one and we might need to attach a tooltip.
Maybe we could use a tab widget instead? That would also allow to preserve state for all three limits. That makes it not only more convenient to operate, it would even allow to change all limits in one go, whereas other solution would require closing and reopening the dialog each time.

dae:
You posted the same comment again, did I miss something?

RumovZ:
Sorry! When I opened GitHub today, my comment was still a draft, so I thought I forgot to sent it yesterday and sent it again. Usually first of all, I refresh the page to avoid these caching issues with GitHub, but today they got me.  😅 

<hr>

> And since only one can apply at a given time, would there be any need for the user to adjust the limits in one go?

It's not a must-have, but I can imagine some use cases: You could increase a parent deck's limit and the preset limit applying to its children by a smaller amount at the same time. Or if you realise that you can introduce more new cards, you might want to double the day limit, but increase the preset or deck limit more moderately.

I would set the higher-precedence limits to the lower-precedence values to imply they're inactive. So if you only had a preset limit configured, the other two tabs would show the same value. Then the *deck-specific today* tab would always show the final value. However, then changing one limit would require to also change the other ones for them to stay inactive.
So your idea of only applying the active tab might be the better one. I don't see a problem with using two sets of tabs.

But I can start with the 'flat' solution. I'll make a quick mock and then we'll see if the 6 extra rows it takes are too much.

RumovZ:
I don't think it takes up too much vertical space, but the experiment revealed other issue: It's not easily recognisable what the applying values are, especially if all values are set (`> -1`). If then a warning is shown about a reviews/new mismatch, it's not at all clear what that refers to.
Also, using `-1` to mean inactive may be unintuitive. For one thing, it might lead users to believe that the value will be added to the preset limits, because that's how the current custom study feature works.

dae:
Dividing it up and tweaking the wording makes it a bit more comprehensible, but I agree, this is probably not the best path forward. I think users will struggle with the -1, and we probably should only be presenting the relevant information instead of all three values
2022-07-14T21:43:10,197551756+10:00
.

I tried mocking up radio buttons and a group box - the latter seems to work ok, though the styling will need tweaking from someone with more experience in this area:
2022-07-14T22:05:05,370053793+10:00

The code change is a complete hack, but here if you wanted to play with it:

```diff
diff --git a/ts/deck-options/DailyLimitRow.svelte b/ts/deck-options/DailyLimitRow.svelte
new file mode 100644
index 000000000..1222b198f
--- /dev/null
+++ b/ts/deck-options/DailyLimitRow.svelte
@@ -0,0 +1,36 @@
+<!--
+    Copyright: Ankitects Pty Ltd and contributors
+    License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+-->
+<script lang="ts">
+    import Col from "../components/Col.svelte";
+    import Row from "../components/Row.svelte";
+    import RevertButton from "./RevertButton.svelte";
+    import SpinBox from "./SpinBox.svelte";
+    import TooltipLabel from "./TooltipLabel.svelte";
+    import Container from "../components/Container.svelte";
+    import EnumSelector from "./EnumSelector.svelte";
+
+    export let value: any;
+    export let defaultValue: any;
+    export let min = 0;
+    export let max: number | undefined = undefined;
+    export let markdownTooltip: string;
+</script>
+
+<Row --cols={12}>
+    <Col --col-size={6} breakpoint="sm">
+        <TooltipLabel {markdownTooltip}><slot /></TooltipLabel>
+    </Col>
+    <Col --col-size={6} breakpoint="sm">
+        <Container --gutter-block="0" --gutter-inline="0">
+            <SpinBox bind:value {min} {max} />
+            <!-- todo <RevertButton bind:value {defaultValue} /> -->
+            <br />
+            <EnumSelector
+                bind:value
+                choices={["Shared Preset", "Deck-Specific", "Today Only"]}
+            />
+        </Container>
+    </Col>
+</Row>
diff --git a/ts/deck-options/DailyLimits.svelte b/ts/deck-options/DailyLimits.svelte
index 1e5b1de35..cd0accc4e 100644
--- a/ts/deck-options/DailyLimits.svelte
+++ b/ts/deck-options/DailyLimits.svelte
@@ -10,6 +10,7 @@
     import SpinBoxRow from "./SpinBoxRow.svelte";
     import TitledContainer from "./TitledContainer.svelte";
     import Warning from "./Warning.svelte";
+    import DailyLimitRow from "./DailyLimitRow.svelte";
 
     export let state: DeckOptionsState;
     export let api: Record<string, never>;
@@ -44,13 +45,13 @@
 <TitledContainer title={tr.deckConfigDailyLimits()}>
     <DynamicallySlottable slotHost={Item} {api}>
         <Item>
-            <SpinBoxRow
+            <DailyLimitRow
                 bind:value={$config.newPerDay}
                 defaultValue={defaults.newPerDay}
                 markdownTooltip={tr.deckConfigNewLimitTooltip() + v3Extra}
             >
-                {tr.schedulingNewCardsday()}
-            </SpinBoxRow>
+                New cards/day
+            </DailyLimitRow>
         </Item>
 
         <Item>
@@ -58,61 +59,21 @@
         </Item>
 
         <Item>
-            <SpinBoxRow
-                bind:value={$config.reviewsPerDay}
-                defaultValue={defaults.reviewsPerDay}
-                markdownTooltip={tr.deckConfigReviewLimitTooltip() + v3Extra}
-            >
-                {tr.schedulingMaximumReviewsday()}
-            </SpinBoxRow>
-        </Item>
-
-        <Item>
-            <Warning warning={reviewsTooLow} />
+            <br /><br />
         </Item>
 
         <Item>
-            <SpinBoxRow
-                value={-1}
-                defaultValue={-1}
-                markdownTooltip={tr.deckConfigNewLimitTooltip() + v3Extra}
-                min={-1}
-            >
-                {tr.deckConfigDeckNewLimit()}
-            </SpinBoxRow>
-        </Item>
-
-        <Item>
-            <SpinBoxRow
-                value={-1}
-                defaultValue={-1}
+            <DailyLimitRow
+                bind:value={$config.reviewsPerDay}
+                defaultValue={defaults.reviewsPerDay}
                 markdownTooltip={tr.deckConfigReviewLimitTooltip() + v3Extra}
-                min={-1}
             >
-                {tr.deckConfigDeckReviewLimit()}
-            </SpinBoxRow>
+                Reviews/day
+            </DailyLimitRow>
         </Item>
 
         <Item>
-            <SpinBoxRow
-                value={-1}
-                defaultValue={-1}
-                markdownTooltip={tr.deckConfigNewLimitTooltip() + v3Extra}
-                min={-1}
-            >
-                {tr.deckConfigTodayNewLimit()}
-            </SpinBoxRow>
-        </Item>
-
-        <Item>
-            <SpinBoxRow
-                value={-1}
-                defaultValue={-1}
-                markdownTooltip={tr.deckConfigReviewLimitTooltip() + v3Extra}
-                min={-1}
-            >
-                {tr.deckConfigTodayReviewLimit()}
-            </SpinBoxRow>
+            <Warning warning={reviewsTooLow} />
         </Item>
     </DynamicallySlottable>
 </TitledContainer>
```

dae:
Yeah, that's definitely better. In the screenshot, I think I prefer the former appearance, as it might blend in with the rest of the screen a bit better. WDYT?

RumovZ:
Yes, agree. It's clear enough what the tab components belong to, so we can do without the border.

pylib/anki/exporting.py
```diff
@@ -270,6 +270,7 @@ def exportInto(self, path: str) -> None:
                 # scheduling not included, so reset deck settings to default
                 d = dict(d)
                 d["conf"] = 1
+                d["reviewLimit"] = d["newLimit"] = None
```
dae, line 273: Perhaps for the daily limits too?

rslib/src/decks/limits.rs
```diff
@@ -5,12 +5,43 @@ use std::{collections::HashMap, iter::Peekable};
...
+use super::{Deck, NormalDeck};
 use crate::{
     deckconfig::{DeckConfig, DeckConfigId},
+    pb::decks::deck::normal::DayLimit,
     prelude::*,
 };
 
+impl NormalDeck {
+    /// The deck's review limit for today or its general one, if any is configured.
+    pub fn day_review_limit(&self, today: u32) -> Option<u32> {
```
dae, line 17:
```suggestion
    pub fn current_review_limit(&self, today: u32) -> Option<u32> {
```

rslib/src/decks/limits.rs
```diff
@@ -5,12 +5,43 @@ use std::{collections::HashMap, iter::Peekable};
...
-use super::Deck;
+use super::{Deck, NormalDeck};
 use crate::{
     deckconfig::{DeckConfig, DeckConfigId},
+    pb::decks::deck::normal::DayLimit,
     prelude::*,
 };
 
+impl NormalDeck {
+    /// The deck's review limit for today or its general one, if any is configured.
```
dae, line 16:
```suggestion
    /// The deck's review limit for today, or its regular one, if any is configured.
```

rslib/src/decks/limits.rs
dae: (same for new case)

rslib/src/decks/schema11.rs
```diff
@@ -116,6 +116,10 @@ pub struct NormalDeckSchema11 {
     extend_new: i32,
     #[serde(default, deserialize_with = "default_on_invalid")]
     extend_rev: i32,
+    #[serde(default, deserialize_with = "default_on_invalid")]
+    review_limit: Option<u32>,
+    #[serde(default, deserialize_with = "default_on_invalid")]
+    new_limit: Option<u32>,
```
dae, line 122: I presume this was a conscious decision to exclude the daily limits from the schema11 representation? That seems reasonable*, I just wanted to double check it was intentional.

*My comment elsewhere about preserving default for today may influence this though.

pylib/anki/exporting.py
dae: Actually, I see you've excluded them from schema11, making that unnecessary.

rslib/src/decks/schema11.rs
```diff
@@ -383,6 +396,13 @@ impl From<Deck> for DeckCommonSchema11 {
     }
 }
 
+/// See [crate::deckconfig::schema11::clear_other_duplicates()].
+fn clear_other_duplicates(other: &mut HashMap<String, Value>) {
+    for key in ["review_limit", "new_limit"] {
```
dae, line 401: Needs to be camelCase, due to `#[serde(rename_all = "camelCase")]`

rslib/src/deckconfig/update.rs
```diff
@@ -83,15 +88,18 @@ impl Collection {
...
         Ok(CurrentDeck {
             name: deck.human_name(),
-            config_id: deck.normal()?.config_id,
+            config_id: normal.config_id,
             parent_config_ids: self
                 .parent_config_ids(&deck)?
                 .into_iter()
                 .map(Into::into)
                 .collect(),
+            limits: Some(normal.to_limits(today)),
```
dae, line 102: One minor downside of doing it this way is that we don't remember what the previously set 'today' limit was after the day rolls over, so if the user frequently extends the limits, they'll have to enter in the desired value each time. We could achieve it by passing the configured limit regardless of day, along with a bool or enum indicating whether it happened to be active at the time. Whether that's worth bothering with at the moment though, I don't know - wouldn't be too hard to add later if we see enough demand for it. If we do add it in the future, we'd probably want to preserve the limits in schema11 as well, or downgrade-and-quit would forget the defaults.

dae (commented):
Looking good! I tested this on a phone screen, and it rendered well.

rslib/src/decks/schema11.rs
RumovZ: Strange, I thought I had checked, but makes sense. Good catch!

rslib/src/deckconfig/update.rs
RumovZ: Yes, I thought about the same thing, but took the easy path. If you think it's worth it, I'll take care of it.

RumovZ: Who'd have thought the trickiest part would be mapping day limits between the frontend and backend representations? But I think everything works as expected now.

dae: Thanks for looking into that :-)

dae:
Thanks Rumo, it's looking good. I noticed one remaining issue: when changing the preset at the top, the newly-selected preset's counts aren't shown in the limits area. Presumably we need a reactive statement that will refresh the tabs when one of preset values change.

RumovZ:
Thanks for pointing out and fixing my other overlook.
It seems impossible to know if the preset changed just by comparing the config values, so I had to add some logic especially for this case. Please give it another go. I can also add some tests if you like.

ftl/core/deck-config.ftl
```diff
@@ -39,8 +39,8 @@ deck-config-limit-interday-bound-by-reviews =
 ## Daily limit tabs: please try to keep these as short as the English version,
 ## as longer text will not fit on small screens.
 
-deck-config-shared-preset = Shared preset
-deck-config-deck-only = Deck only
+deck-config-shared-preset = Preset
+deck-config-deck-only = This deck
 deck-config-today-only = Today only
```
RumovZ, line 44: I think this could just as well apply to the preset. Maybe "This deck today" would be clearer? Hypothetically, Anki could add "Preset today" in the future.
When that's decided we should pronably adjust the keys as well.

dae: Just trying to keep the names as succinct as possible, so they fit well. How about we explain in more detail in the tooltip?

dae:
I noticed one more issue: we're showing these options when the user is using the v2 scheduler, which doesn't support them. Could we exclude the 2nd and 3rd tab if state.v3Scheduler is false?

(looking forward to the day when we can drop v2!)

PR #1956, opened 2022-07-10
Adjust remaining steps after config update

### LINKED ISSUE #1927: Adjust learning card steps when altering deck options
Users get repeated or skipped steps when they change the number of learning steps while cards are in learning. We could reduce user confusion if we adjusted the remaining steps on cards in learning when saving the deck options.

RumovZ:
Closes #1927.
Should we also be doing this when moving a learning card to a different deck?

### DIFF
--- a/rslib/src/deckconfig/update.rs
+++ b/rslib/src/deckconfig/update.rs
@@ -13,6 +13,7 @@ use crate::{
     pb,
     pb::deck_configs_for_update::{ConfigWithExtra, CurrentDeck},
     prelude::*,
+    search::{JoinSearches, SearchNode, SortMode, StateKind},
 };
 
 #[derive(Debug, Clone)]
@@ -147,14 +148,15 @@ impl Collection {
         // loop through all normal decks
         let usn = self.usn()?;
         let selected_config = input.configs.last().unwrap();
+        let mut decks_by_steps_len_delta: HashMap<i32, Vec<DeckId>> = HashMap::new();
         for deck in self.storage.get_all_decks()? {
             if let Ok(normal) = deck.normal() {
                 let deck_id = deck.id;
 
                 // previous order
                 let previous_config_id = DeckConfigId(normal.config_id);
-                let previous_order = configs_before_update
-                    .get(&previous_config_id)
+                let previous_config = configs_before_update.get(&previous_config_id);
+                let previous_order = previous_config
                     .map(|c| c.inner.new_card_insert_order())
                     .unwrap_or_default();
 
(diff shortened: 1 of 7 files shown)


rslib/src/deckconfig/update.rs
```diff
@@ -171,20 +173,74 @@ impl Collection {
...
+    fn adjust_remaining_learning_steps(
+        &mut self,
+        decks_by_steps_len_delta: &HashMap<i32, Vec<DeckId>>,
+    ) -> Result<()> {
+        let usn = self.usn()?;
+        for (&delta, decks) in decks_by_steps_len_delta {
+            for mut card in self.all_learning_cards_in_decks(decks)? {
+                let original = card.clone();
+                // strip "remaining today"
+                let remaining = (card.remaining_steps % 1000) as i32;
```
RumovZ, line 208: I understand `remaining_steps / 1000` is the number of learning steps that can be completed before the day rolls over in v2. Is this still used? I couldn't find this value being updated after a config change, so I reasoned it wouldn't aggrevate things to simply strip it.

dae:
Ideally, we'd handle it for deck changing as well. Is it an easy change? If not, fine to skip it - I suspect deck option changes are the main cause of confusion.

RumovZ:
Probably not that hard.
I also just realise that it's the same for relearning cards.

dae:
The only thing I was going to point out was that we should be caching deck configs, but then I noticed you're already doing that. Code looks great :-)

Having thought about this a bit more during the review, I've realised that this approach may not be optimal in all cases. Consider the case where the user has steps 1 10 100, and a card currently on 10. If a step is added after 1, then ideally the remaining count should stay the same, since the user is already past that step, and the old behaviour of leaving the remaining count alone would have worked better. Perhaps one way we could solve this is to note down the current actual delay of the card, and then try to find the best place that delay fits inside the new steps? eg walking the steps backwards, and stopping at the first step <= the previous delay. WDYT?

dae:
I can only offer anecdotal evidence here, but in the decks I've encountered when investigating problems, I don't recall ever seeing repeated steps. Not saying they don't exist, but they're probably fairly rare.

RumovZ:
It's probably only a small step to support that as well after implementing delay matching, though.

RumovZ:
I think I could do it in that time, but I won't obfiscate the code with edge-case handling if you think it's that rare. ;)

As you can see, I've also added some test helpers like we've started to discuss some time ago, most notably `DeckAdder`. If it proves its worth, maybe that kind of helpers can be incrementally expanded and adopted in future refactorings.

dae:
Love the brevity of those tests! No objections to adding more cfg(test)-gated helpers where they make things easier. Thanks Rumo!

PR #1960, opened 2022-07-16
Load previously loaded profile

sam1penny:
See this [forum post](https://forums.ankiweb.net/t/dont-ask-for-profile-upon-startup-every-time/839) for more details.

Open to suggestions / improvements

### DIFF
--- a/qt/aqt/forms/preferences.ui
+++ b/qt/aqt/forms/preferences.ui
@@ -104,13 +104,6 @@
         </widget>
        </item>
        <item>
-        <widget class="QCheckBox" name="load_last_loaded_profile">
-         <property name="text">
-          <string notr="true">Load last loaded profile by default</string>
-         </property>
-        </widget>
-       </item>
-       <item>
         <widget class="QComboBox" name="useCurrent">
          <item>
           <property name="text">
@@ -692,7 +685,6 @@
   <tabstop>paste_strips_formatting</tabstop>
   <tabstop>ignore_accents_in_search</tabstop>
   <tabstop>new_import_export</tabstop>
-  <tabstop>load_last_loaded_profile</tabstop>
   <tabstop>useCurrent</tabstop>
   <tabstop>default_search_text</tabstop>
   <tabstop>uiScale</tabstop>
--- a/qt/aqt/forms/preferences.ui
+++ b/qt/aqt/forms/preferences.ui
@@ -104,6 +104,13 @@
         </widget>
        </item>
        <item>
+        <widget class="QCheckBox" name="load_last_loaded_profile">
+         <property name="text">
+          <string notr="true">Load last loaded profile by default</string>
+         </property>
+        </widget>
+       </item>
+       <item>
         <widget class="QComboBox" name="useCurrent">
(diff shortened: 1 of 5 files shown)


qt/aqt/forms/preferences.ui
```diff
@@ -103,6 +103,13 @@
          </property>
         </widget>
        </item>
+       <item>
+        <widget class="QCheckBox" name="load_last_loaded_profile">
+         <property name="text">
+          <string notr="true">Load last loaded profile by default</string>
```
dae, line 109: Because of https://forums.ankiweb.net/t/set-html-editor-as-a-default-editor-instead-of-visual-editor/20988/7, perhaps it would be better to start off by switching from the old functionality to the new functionality, without an option to change back to the old way. Hopefully, people will like the new behaviour, and we won't need another option. If that turns out to be wrong, we can always add an option later.

(if/when we do so, we would need to make that string translatable by putting it in an .ftl file)

qt/aqt/main.py
```diff
@@ -284,6 +284,11 @@ def setupProfile(self) -> None:
             profs = self.pm.profiles()
             if len(profs) == 1:
                 self.pm.load(profs[0])
+            elif self.pm.load_last_loaded_profile() and (
+                name := self.pm.last_loaded_profile_name()
+            ):
+                self.pm.load(name)
```
dae, line 290: What happens when the user deletes the last-loaded profile then starts Anki again?

qt/aqt/profiles.py
```diff
@@ -544,6 +545,18 @@ def new_import_export(self) -> bool:
     def set_new_import_export(self, enabled: bool) -> None:
         self.meta["new_import_export"] = enabled
 
+    def load_last_loaded_profile(self) -> bool:
+        return self.meta.get("load_last_loaded_profile", False)
+
+    def set_load_last_loaded_profile(self, val: bool) -> None:
+        self.meta["load_last_loaded_profile"] = val
+
+    def last_loaded_profile_name(self) -> Optional[str]:
```
dae, line 554: Modern syntax is `str | None`, which saves you having to import Optional.

dae (commented):
Thanks for looking into this Sam!

qt/aqt/profiles.py
sam1penny: Thanks, fixed!

qt/aqt/main.py
sam1penny: Thanks for spotting that, should be fixed in the latest commit.

Btw, is there any reason that the type hint for `profiles()` in `aqt/profiles.py` is:
`def profiles(self) -> list:` rather than `def profiles(self) -> list[str]:` ?

dae: Some of the typings were added using an automatic tool that did not create perfect types

dae:
Thanks Sam!

nyxaria:
I am using Anki on Mac, how exactly should I go about enabling this behaviour to automatically load previously used profile on startup? I can't seem to find this in the settings

PR #1961, opened 2022-07-16
Improve eslint formatting

hgiesel:
After this PR, `prettier` would be executed as part of `eslint` via `eslint-plugin-prettier`.

This would mean that, we need to run one "testing tool" less (eslint and svelte-check).

The only thing I stumbled upon is that prettier seems to also be used for formatting files in `/docs`? However they don't seem to actually be executed? Simply running `prettier --write` within the directory does change them.

Background: I worked on my tooling a bit, and configured running `eslint --fix` on save.


dae:
(I presume this is still a draft at this point)

I see the appeal of being able to run one fix command instead of having to run two separate ones. One downside of this change is that fixing formatting is slower now, though if the editor is formatting on the fly, I guess it won't be a problem.

After you get the build/test failures fixed, it would be interesting to run `time bazel test //... --cache_test_results=no` before and after this change to see what sort of speed benefit we gain from it. 

The -svelte plugin receives 2k weekly downloads vs the offical one's 100k, which makes me a bit nervous about its long term support, but by the looks of the changes you've made so far, it seems to be doing a better job than the official one.

hgiesel:
> The -svelte plugin receives 2k weekly downloads vs the offical one's 100k

The numbers are a bit skewed, because the npm package just changed. This is the deprecated old entry, which still received 10k downloads this week:
https://www.npmjs.com/package/@ota-meshi/eslint-plugin-svelte

Maybe we'll have some kind of official statement [soon](https://github.com/sveltejs/eslint-plugin-svelte3/issues/184).

JounQin:
Hi, I'm the active maintainer of https://github.com/prettier/eslint-plugin-prettier for now, I have to say the performance of it is a bit not idea although I'll improved it a lot at https://github.com/prettier/eslint-plugin-prettier/pull/415 and https://github.com/prettier/eslint-plugin-prettier/pull/485.

So for my own cases, I'm supporting env based choice at https://github.com/1stG/configs/pull/128, so that it can be chosen to prefer `prettier` cli or `eslint-plugin-prettier` instead locally or on CI.

---

Besides, for `svelte`, you should just use `eslint-plugin-svelte` instead of the *official* one.

See also https://github.com/prettier/eslint-plugin-prettier#svelte-support

hgiesel:
Seems like using `eslint-plugin-svelte` does not cooperate with `eslint-plugin-simple-sort-imports` :sigh:

JounQin:
> Seems like using `eslint-plugin-svelte` does not cooperate with `eslint-plugin-simple-sort-imports` :sigh:

You could raise an issue with reproduction.

dae:
Keeping an eye on https://github.com/ota-meshi/eslint-plugin-svelte/issues/199

dae:

PR #1968, opened 2022-07-17
Expose cloze text as HTML attribute on question side

kleinerpirat:
Exposes clozed text as a data-attribute on the front side.
Use case: https://forums.ankiweb.net/t/an-improved-implementation-of-cloze-to-enable-sequential-uncovering/21489/6

# Example
Input:
```
{{c1::hidden text}}
```

Output with current implementation:
```html
<span class=cloze>[...]</span>
```

After this PR:
```html
<span class="cloze" data-cloze="hidden text">[...]</span>
```

### DIFF
--- a/rslib/src/cloze.rs
+++ b/rslib/src/cloze.rs
@@ -68,7 +68,7 @@ pub fn reveal_cloze_text(text: &str, cloze_ord: u16, question: bool) -> Cow<str>
         let text_attr;
         let replacement;
         if question {
-            text_attr = format!(r#" data-cloze="{}""#, htmlescape::encode_minimal(&text));
+            text_attr = format!(r#" data-cloze="{}""#, htmlescape::encode_attribute(&text));
             // hint provided?
             if let Some(hint) = caps.get(cloze_caps::HINT) {
                 replacement = format!("[{}]", hint.as_str());
--- a/rslib/src/cloze.rs
+++ b/rslib/src/cloze.rs
@@ -57,26 +57,30 @@ pub fn reveal_cloze_text(text: &str, cloze_ord: u16, question: bool) -> Cow<str>
             .parse()
             .unwrap_or(0);
 
+        let text = caps.get(cloze_caps::TEXT).unwrap().as_str().to_owned();
         if captured_ord != cloze_ord {
             // other cloze deletions are unchanged
-            return caps.get(cloze_caps::TEXT).unwrap().as_str().to_owned();
+            return text;
         } else {
             cloze_ord_was_in_text = true;
         }
 
+        let text_attr;
         let replacement;
         if question {
+            text_attr = format!(r#" data-text="{}""#, text);
             // hint provided?
             if let Some(hint) = caps.get(cloze_caps::HINT) {
                 replacement = format!("[{}]", hint.as_str());
             } else {
-                replacement = "[...]".to_string()
+                replacement = "[...]".to_string();
             }
         } else {
-            replacement = caps.get(cloze_caps::TEXT).unwrap().as_str().to_owned();
+            text_attr = "".to_string();
+            replacement = text;
         }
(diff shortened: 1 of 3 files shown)


kleinerpirat:
Perhaps it could be useful to expose the hint as `data-hint` as well and show it as a CSS pseudo-element. People could then make the hints show on hover/click or create tooltips:
```
{{c1::hidden text::hint}}
```
```html
<span class="cloze" data-cloze="hidden text" data-hint="hint"></span>
```

A disadvantage of this approach is that pseudo-elements are not selectable since they're not part of the DOM.

rslib/src/cloze.rs
```diff
@@ -57,26 +57,30 @@ pub fn reveal_cloze_text(text: &str, cloze_ord: u16, question: bool) -> Cow<str>
...
-            return caps.get(cloze_caps::TEXT).unwrap().as_str().to_owned();
+            return text;
         } else {
             cloze_ord_was_in_text = true;
         }
 
+        let text_attr;
         let replacement;
         if question {
+            text_attr = format!(r#" data-cloze="{}""#, htmlescape::encode_minimal(&text));
```
dae, line 71: minimal is not appropriate here: https://docs.rs/htmlescape/latest/htmlescape/fn.encode_minimal.html#safety-notes

dae:
No objections to adding hint as well, though let's avoid pseudo elements for now.

rslib/src/cloze.rs
kleinerpirat: Alright, then I'll create a helper function for the pylib tests. `html.escape` doesn't encode as thoroughly as Rust's `htmlescape::encode_attribute`:)

kleinerpirat:
Considering the hint is available as the innerHTML of a cloze anyway, I think I'll hold back on adding it as an attribute until a proper cloze overhaul is in the works. If the attribute isn't used by Anki itself (i.e. as pseudo-element content), it is redundant.

gomarcus:
Thanks @kleinerpirat for the implementation! Can hardly wait for the update to reach me. Added another idea to my suggestion in our last chat to expand the 'Cloze', which certainly has many practical applications. Would be very happy if these could be incorporated as well. Although this rather requires a revision of the backend.

https://forums.ankiweb.net/t/an-improved-implementation-of-cloze-to-enable-sequential-uncovering/21489/9

**Proposal: A text cloze should be able to be assigned to several groups.**
```
current: {{group::value::hint}}
proposal: {{group1,group2,group3::value::hint}}
```

Template:
```anki
The {{c1,c2::quick}} brown {{c2,c3::fox}} jumps over the lazy dog
```

Cards:
```anki
The [...] brown fox jumps over the lazy dog
```
```anki
The [...] brown [...] jumps over the lazy dog
```
```anki
The quick brown [...] jumps over the lazy dog
```

TRIAEIOU:
To follow up with an additional suggestion that should be easy to implement with few maintenance issues: increase the meta data to allow easier addon creation:
- Hint (as @kleinerpirat suggested)
- Cloze ordinal
- Wrap inactive clozes (i.e. the "other" clozes) in a `<span>` as well with the same meta data and a specific class,  eg `inactive`.

So something like this
`The {{c1::quick::speed}} brown {{c2::fox}} jumped over the lazy {{c2::dog}}` 
⇓
`The <span class="cloze" data-cloze="quick" data-hint="speed" data-ordinal="1">quick</span> brown <span class="cloze inactive" data-cloze="fox" data-hint="" data-ordinal="2">fox</span> jumped over the lazy <span class="cloze inactive" data-cloze="dog" data-hint="" data-ordinal="2">dog</span>`

It would be easy enough to make such a PR if there is an indication of it being welcome?

dae:
I'd be fine with accepting such a PR.

PR #1969, opened 2022-07-17
Expose side of card in body className

kleinerpirat:
Makes wrapping the front template in `<div class="front">` or similar hacks to allow side-specific CSS redundant.

I find "front/back" more fitting than "question/answer", because it is more generic.


dae:
This won't be useful until all of the clients support it, so it might be better to hold off on this until we have all the clients using the same reviewer JS, then we can make the change in one go.

PR #1970, opened 2022-07-18
Fix legacy import switch

glutanimate:
Just a super minor typo/remnant from the old logic that prevents the switch from working atm


dae:
Whoops - thanks for catching that.

PR #1971, opened 2022-07-18
Add deck/collection export hooks

glutanimate:
Adds a number of hooks that allow add-ons to perform actions before/after decks are exported and – in the case of the new export handling – modify the options of the export.

Previously, add-ons could achieve something similar by wrapping `Exporter.exportInto`. However, with exports now being handled by the Rust backend in the background and `on_success` out of easy reach, my hope is that these hooks will allow add-on authors to retain the ability to act upon deck exports without major monkey-patches.

**Use case**

Although I think this could be useful in many scenarios, the concrete use case I'm targeting is mobile / AnkiWeb support for the AMBOSS add-on: To make the medical definition pop-ups work on these platforms, in its current beta release the add-on appends a small one-liner to card templates – in an idempotent operation that's fully reversible. However, keeping the one-liner around when exporting decks is not ideal as the user importing the deck might not have the add-on installed, nor want the explanations to show up in their cards.

So to address this scenario, with legacy export handling, the add-on currently monkey-patches legacy exporters to remove the snippet ahead of exports, and then re-introduce it afterwards. With the new export code this is much more difficult to do because of the asynchronous background operation, so I thought it would be best to handle this by extending the hook API.

**Some notes**:

- I tried to keep the footprint as minimal as possible, but did end up introducing some minor refactoring changes to improve typing and clarity of the API when used in add-ons (e.g. use of `ExportFormat`, `Options` → `ExportOptions`). Hope that's ok still in this early phase of the new exporter code
- `exporter_will_export` being a filter for `ExportOptions` isn't necessary for the AMBOSS use case, but I figured it could be helpful to other add-ons, so I opted to add it now rather than the hook API having to change in the future. Would be happy to grade this down to a regular hook, however, if you think it's pre-mature and add-ons wanting to modify the export settings at this stage is not as likely.
- To make it easier to give the hooks a try, I've added a debug console snippet in the comment below

----

### DIFF
--- a/qt/aqt/import_export/exporting.py
+++ b/qt/aqt/import_export/exporting.py
@@ -49,14 +49,14 @@ class ExportDialog(QDialog):
         self.open()
 
     def setup(self, did: DeckId | None) -> None:
-        self.exporter_classes: list[Type[Exporter]] = [
+        self.exporters: list[Type[Exporter]] = [
             ApkgExporter,
             ColpkgExporter,
             NoteCsvExporter,
             CardCsvExporter,
         ]
         self.frm.format.insertItems(
-            0, [f"{e.name()} (.{e.extension})" for e in self.exporter_classes]
+            0, [f"{e.name()} (.{e.extension})" for e in self.exporters]
         )
         qconnect(self.frm.format.activated, self.exporter_changed)
         if self.nids is None and not did:
@@ -86,7 +86,7 @@ class ExportDialog(QDialog):
             self.frm.includeSched.setChecked(False)
 
     def exporter_changed(self, idx: int) -> None:
-        self.exporter = self.exporter_classes[idx]()
+        self.exporter = self.exporters[idx]()
         self.frm.includeSched.setVisible(self.exporter.show_include_scheduling)
         self.frm.includeMedia.setVisible(self.exporter.show_include_media)
         self.frm.includeTags.setVisible(self.exporter.show_include_tags)
--- a/qt/aqt/import_export/exporting.py
+++ b/qt/aqt/import_export/exporting.py
@@ -126,14 +126,14 @@ class ExportDialog(QDialog):
             break
         return path
 
-    def options(self, out_path: str) -> Options:
+    def options(self, out_path: str) -> ExportOptions:
         limit: ExportLimit = None
         if self.nids:
             limit = NoteIdsLimit(self.nids)
         elif current_deck_id := self.current_deck_id():
             limit = DeckIdLimit(current_deck_id)
 
-        return Options(
+        return ExportOptions(
             out_path=out_path,
             include_scheduling=self.frm.includeSched.isChecked(),
             include_media=self.frm.includeMedia.isChecked(),
@@ -166,7 +166,7 @@ class ExportDialog(QDialog):
 
 
 @dataclass
-class Options:
+class ExportOptions:
     out_path: str
     include_scheduling: bool
     include_media: bool
@@ -200,7 +200,7 @@ class Exporter(ABC):
 
     @classmethod
     @abstractmethod
-    def export(cls, mw: aqt.main.AnkiQt, options: Options) -> None:
+    def export(cls, mw: aqt.main.AnkiQt, options: ExportOptions) -> None:
         pass
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


glutanimate:
Debug console snippet / sample add-on: 

```python
from aqt.gui_hooks import (
    exporter_did_export,
    exporter_will_export,
    legacy_exporter_did_export,
    legacy_exporter_will_export,
)
from aqt.import_export.exporting import ExportFormat, ExportOptions


def on_exporter_did_export(export_options: ExportOptions, export_format: ExportFormat):
    print(
        f"Exporter did export data with {export_format} format. Options: {export_options}"
    )


def on_exporter_will_export(
    export_options: ExportOptions, export_format: ExportFormat
) -> ExportOptions:
    print(
        f"Exporter will export data with {export_format} format. Options: {export_options}"
    )
    return export_options


def on_legacy_exporter_did_export(file_extension: str):
    print(f"Legacy exporter did export data with {file_extension} extension.")


def on_legacy_exporter_will_export(file_extension: str):
    print(f"Legacy exporter will export data with {file_extension} extension.")


exporter_will_export.append(on_exporter_will_export)
exporter_did_export.append(on_exporter_did_export)
legacy_exporter_will_export.append(on_legacy_exporter_will_export)
legacy_exporter_did_export.append(on_legacy_exporter_did_export)

```

glutanimate:
Hmm, both very good points, yeah. Might have been a bit overambitious with DRY. My main goal with the enum initially was to find a way to tell the two CSV exporters apart without e.g. passing the exporter class to the add-on (as that always feels like it introduces some extra coupling).

What do you think of this approach? We could:

- Go back to the `extension` property as it was for interop between `Exporter` and `ExportDialog`, and easy implementation of new `Exporter`s by add-ons
- Keep `ExportFormat`, but relegate it to only be used in the hook, and change it to an integer enum to avoid the `csv` duplication. If add-ons need to access the extension, they can simply look at `ExportOptions.out_path`
- Modify the hooks to accept `export_format: ExportFormat | str`, where `str` can be used by add-ons to identify their custom exporters with a string key (if they want to call these hooks)

glutanimate:
Just a quick note on the updated implementation - decided to go with passing in `Exporter` instances (even though it meant initializing them in the first place) for a couple of reasons:

- most hooks that are constructed similarly usually accept an instance, so add-on authors are likely familiar with `isinstance` checks in their handlers
- should `Exporter` need to become stateful later down the line, we won't have to break the hook, or resort to awkwardly passing in `self.exporter.__class__` or something similar

RumovZ:
> @RumovZ any objections/comments on your end?

Happy too see someone tweak this code to the needs of add-ons!

qt/aqt/import_export/exporting.py
```diff
@@ -86,7 +86,7 @@ def setup(self, did: DeckId | None) -> None:
             self.frm.includeSched.setChecked(False)
 
     def exporter_changed(self, idx: int) -> None:
-        self.exporter = self.exporters[idx]
+        self.exporter = self.exporter_classes[idx]()
```
dae, line 89: There are a couple of add-ons that reference exporters - is it worth breaking them to rename this?

dae: (otherwise looks good)

glutanimate: No, no need to cause any extra breakages I'd say. Pushed a change to revert the naming.

However, out of curiosity: Were the add-ons you found add-ons that already modify the new `ExportDialog` implementation? I did try grepping for potentially affected add-ons before the PR, but the ones I found were targeting the old dialog, and correct me if I'm wrong, but they would have to be updated anyway to support the new code, right? (i.e. there's no reference that Anki used to keep to the old `ExportDialog` (e.g. via the dialog manager) that's now switched out to the new one and could trip up add-ons if they access `.exporters`?)

If we aren't breaking compatibility of existing add-ons, but just introducing an extra step on top of other necessary adjustments, then I would slightly favor going with `exporter_classes` as it highlights the difference to `.exporter` better. But this is super minor, and at the end of the day both would be absolutely fine by me.

dae: You are right - I just did a quick grep without really thinking about it, but this is new code that's unlikely to have been used yet. Sorry for the wasted effort; feel free to revert the revert.

glutanimate: No worries, thanks for checking! ok, done

PR #1973, opened 2022-07-19
Fix missing limits field in tests

RumovZ:
Strange this wasn't caught by the PR action.


dae:
Yeah, I'm puzzled by why the PR check didn't fail too. Thanks for the quick fix :-)

PR #1976, opened 2022-07-19
Improve temporary table handling

RumovZ:
I've investigated RAII guards for temporary tables and here's a start on the refactoring. My conclusions so far:

- Guards make a lot of sense in `gather.rs`, where there are several operations relying on the tables. Responsibility for the table is explicitly passed around and for any call in `gather_data()` it's clear if the table is still alive.
- Functions like `gather_revlog()`, that rely on the cards table already exisisting, can and probably should be implemented on `CardTableGuard`, so it's clear inside the callee as well.
- However, every other table user is either immediately collecting or iterating over the searched cards/notes. Here, it seems sensible to move the setup→collect→clear logic into a helper, but inside such a small function a RAII guard gains us little.
- Even without a guard, adhering more to RAII might help reduce missing cleanups. That is to say, `setup_searched_cards_table()` should be called in the same scope as `clear_searched_cards_table()` as demonstrated in 9d85400a567804ea4d2676cccfbafb8ca414cd83.

WDYT? Are guards overkill if they're only really useful at one place? Should I continue to implement the other refactorings?

### DIFF
--- a/rslib/src/card/mod.rs
+++ b/rslib/src/card/mod.rs
@@ -274,12 +274,11 @@ impl Collection {
         ))?;
         let config = self.get_deck_config(config_id, true)?.unwrap();
         let mut steps_adjuster = RemainingStepsAdjuster::new(&config);
-        self.storage.set_search_table_to_card_ids(cards, false)?;
         let sched = self.scheduler_version();
         let usn = self.usn()?;
         self.transact(Op::SetCardDeck, |col| {
             let mut count = 0;
-            for mut card in col.storage.all_searched_cards()? {
+            for mut card in col.all_cards_for_ids(cards, false)? {
                 if card.deck_id == deck_id {
                     continue;
                 }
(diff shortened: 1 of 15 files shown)


dae:
One potential issue with the current guard approach is if we wanted to add another one for searched notes, a future routine that made use of both tables at once presumably would not be able to hold both guards, due to the &mut col. 

One alternative approach would be to give the guard a private cleaned_up bool, and panic on drop if cleaned_up is not true. The clear table function could be passed in the guard, and would take care of setting the flag to true. It would mean a shift from a compile time to a runtime check, but presumably we'd still get a warning if guard was assigned and then not used again.

I don't have a strong preference on whether we keep the guard - the helpers you've added look to cover the majority of the cases already, but it presumably wouldn't take long to fix either. Up to you.

RumovZ:
As you can see in 310c6c3, guards can be chained without a problem. I'll leave them in then.
By the way, a lot of db functions take `&mut self` where `&self` would suffice. Of course, there is a number of reasons why this would happen inadvertently, but just to be sure: Is this ever intentional or is it always okay to remove the mut?

dae:
My thinking was that while the second guard is held, the first one can't be touched, so you couldn't eg. pass it into a helper routine that wanted to drop it early. Might not be an issue in practice though, so no objections from me.

I can't recall ever deliberately using a &mut where & would have sufficed - I'd say these are all oversights.

RumovZ:
Ah, that's true. As you can't access the first guard, it would be inconvenient to have functions implemented on it, like I had originally imagined. But as long as we don't do that, I think it's fine. And for very complex use cases, there's always the option to manage tables manually.

PR #1977, opened 2022-07-20
Populate media DB on import

### LINKED ISSUE #1831: Populate media DB on import
Currently importing does not update the media db, meaning all the files need to be checksummed the next time the user syncs or imports again. It might be a nice optimization to update the entries as we import, so that the next sync or import is faster.

RumovZ:
Closes #1831.

### DIFF
--- a/rslib/src/import_export/package/apkg/import/media.rs
+++ b/rslib/src/import_export/package/apkg/import/media.rs
@@ -11,10 +11,7 @@ use crate::{
         package::media::{extract_media_entries, SafeMediaEntry},
         ImportProgress, IncrementableProgress,
     },
-    media::{
-        files::{add_hash_suffix_to_file_stem, sha1_of_reader},
-        MediaManager,
-    },
+    media::files::{add_hash_suffix_to_file_stem, sha1_of_reader},
     prelude::*,
 };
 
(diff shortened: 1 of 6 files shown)


dae:
Thanks for looking into this Rumo! The refactoring is nice too.

Would be nice to have a small unit test for this that confirms entries are being added correctly. Currently it seems to be broken in the legacy apkg case - the db2 file ends up with all-zero checksums, presumably due to from_legacy(). We could either calculate the checksums on the fly in that case, or skip adding the entries and rely on change scanning to pick them up later - but the former would be more efficient as the files would already be in memory.

If not electing to calculate the sums at SafeMediaEntry creation time, it might also be worth considering changing SafeMediaEntry::sha1 to an optional, so code that deals with it is prepared for this case.

RumovZ:
Whoops, thanks for checking!
I've added the test for this to the roundtrip test, as that only took a handful of lines and a unit test would require much of the same setup. 

It's a bit unfortunate that the hash has this implicit null form. I'm not sure if an `Option` would be better or worse. On the one hand, `None` is superior to `Some([0; 20])`. But on the other hand, we'd then have to ensure `[0; 20]` is always mapped to `None`.
We could introduce a new type that takes care of this, though.

dae:
There wasn't a great deal that needed changing to make it optional. It avoids having to compare bytes, and will ensure we don't accidentally allow bad data in. Does the commit I've just pushed work for you?

dae:
>  But on the other hand, we'd then have to ensure [0; 20] is always mapped to None.

I might be missing something here - why do we need to check for this? With a legacy import we know we need to populate the checksums.

RumovZ:
I was thinking that it could be confusing in in a broader context where there are both options and plain arrays, which may be zeroed. But I see now that the import code was the only place using `[0; 20]`. An option is much better then!

PR #1978, opened 2022-07-20
Introduce exporters_list_created hook to new export code

glutanimate:
Basically just bringing the old `hooks.exporters_list_created` hook to the new export code. Thought it would be a good follow-up to #1971 in order to further flesh out the exporter add-on API.

*(filing as myself again, so no disclosure)*

Sample add-on (presupposing #1971):

```python
from typing import List, Type

from aqt.gui_hooks import exporters_list_created
from aqt.import_export.exporting import Exporter, ExportOptions
from aqt.main import AnkiQt


class MyExporter(Exporter):
    extension = "foo"
    show_deck_list = True
    show_include_html = True

    @staticmethod
    def name():
        return "My Export File Format"

    def export(self, mw: AnkiQt, options: ExportOptions) -> None:
        print(f"Exporting data with options {options}")


def extend_exporters_list(exporters_list: List[Type[Exporter]]):
    exporters_list.append(MyExporter)


exporters_list_created.append(extend_exporters_list)

```

### DIFF
--- a/qt/aqt/import_export/exporting.py
+++ b/qt/aqt/import_export/exporting.py
@@ -55,7 +55,7 @@ class ExportDialog(QDialog):
             NoteCsvExporter,
             CardCsvExporter,
         ]
-        gui_hooks.exporters_list_created(self.exporters)
+        gui_hooks.exporters_list_did_initialize(self.exporter_classes)
         self.frm.format.insertItems(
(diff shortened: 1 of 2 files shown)


dae:
While sorting out the conflict, could we perhaps tweak the name to something like `exporters_list_did_initialize` for consistency with other hooks?

PR #1982, opened 2022-07-22
Fix sidebar not being restored properly

### LINKED ISSUE #1902: Browser sidebar appears as small box in top left
Appears to be a regression in Qt 6.3. Reports on Windows and Linux; seems fairly rare. Maybe DPI-related?

https://forums.ankiweb.net/t/browser-sidebar-is-not-enlarging/20207
https://forums.ankiweb.net/t/the-browse-window-is-not-rendered-correctly-on-version-2-1-52-and-ubuntu-22-04

hikaru-y:
Fixes #1902

I was able to reproduce the issue with Qt6 builds on Windows 10 and Linux (Lubuntu 22.04).

#### Steps to reproduce the issue
1. Open the browser and make the window size as small as possible
1. Click on the maximize button
1. Close the browser and re-open it

It appears that calling `restoreState()` before setting up the sidebar(`QDockWidget`) is causing the issue. I am not sure why the issue did not happen before even though the code has not been changed, but I suspect that [this patch](https://codereview.qt-project.org/c/qt/qtbase/+/375521) has probably brought the issue to the surface.

### DIFF
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -124,13 +124,14 @@ class Browser(QMainWindow):
         self.form = aqt.forms.browser.Ui_Dialog()
         self.form.setupUi(self)
         restoreGeom(self, "editor", 0)
-        restoreState(self, "editor")
         restoreSplitter(self.form.splitter, "editor3")
         self.form.splitter.setChildrenCollapsible(False)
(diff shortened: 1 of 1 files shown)


dae:
Great job on tracking this down Hikaru!

PR #1987, opened 2022-07-26
Default input setting in fields dialog

kleinerpirat:
Adds a checkbox to the "Options" section of the fields dialog that allows users to show the HTML editor by default. This is a preparation for the Editor design changes I've been planning for some time. I decided to split the project into mulitple PRs.

I went for `"rich" | "plain"` instead of a boolean because it is easier to understand for devs/add-on authors.

## Why introduce this setting?
I'd like to get rid of the trio of buttons on the top left of `LabelContainer` and scatter them to more intuitive places.

image
###### For brainstorming on the new editor look and feel, see [here](https://forums.ankiweb.net/t/brainstorming-for-modern-ui-anki-3-0/17510).

Expected behavior:

- chevron collapses the whole field (`richTextInput` AND `plainTextInput`)
- the text on the bottom corner shows up on highlight and toggles the non-default input

The latter is why I want to introduce a default input toggle: for toggled fields it could say "OUTPUT" instead of "HTML", for example. Side note: With the new fields design, a "[collapsed by default](https://forums.ankiweb.net/t/toggle-visual-editor-collapse-permanently-by-default-like-collapsible-fields/20845)" setting would fit nicely underneath this one in the fields dialog.

@dae I have read your "1000 settings later" comment and understand your hesitancy towards this. No hard feelings if you reject it - I only made this to cater to the people who would enjoy this feature.

## Forum requests:
https://forums.ankiweb.net/t/set-html-editor-as-a-default-editor-instead-of-visual-editor/20988
https://forums.ankiweb.net/t/html-editor-by-default-for-ankimobile-ios/20223

---
Related: https://forums.ankiweb.net/t/default-editor-settings-instead-of-persistent/12761

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -499,7 +499,7 @@ require("anki/ui").loaded.then(() => require("anki/NoteEditor").instances[0].too
         ]
 
         flds = self.note.note_type()["flds"]
-        plain_texts = [fld.get("plain_text", False) for fld in flds]
+        plain_texts = [fld.get("plainText", False) for fld in flds]
         descriptions = [fld.get("description", "") for fld in flds]
 
         self.widget.show()
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -499,7 +499,7 @@ require("anki/ui").loaded.then(() => require("anki/NoteEditor").instances[0].too
         ]
 
         flds = self.note.note_type()["flds"]
-        inputs = [fld.get("input", "") for fld in flds]
+        plain_texts = [fld.get("plain_text", False) for fld in flds]
         descriptions = [fld.get("description", "") for fld in flds]
 
         self.widget.show()
@@ -522,7 +522,7 @@ require("anki/ui").loaded.then(() => require("anki/NoteEditor").instances[0].too
 
         js = """
             setFields({});
-            setInputs({});
+            setPlainTexts({});
             setDescriptions({});
             setFonts({});
             focusField({});
@@ -531,7 +531,7 @@ require("anki/ui").loaded.then(() => require("anki/NoteEditor").instances[0].too
             setTags({});
             """.format(
             json.dumps(data),
-            json.dumps(inputs),
+            json.dumps(plain_texts),
             json.dumps(descriptions),
             json.dumps(self.fonts()),
             json.dumps(focusTo),
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -499,6 +499,7 @@ require("anki/ui").loaded.then(() => require("anki/NoteEditor").instances[0].too
         ]
 
         flds = self.note.note_type()["flds"]
+        inputs = [fld.get("input", "") for fld in flds]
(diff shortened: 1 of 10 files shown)


kleinerpirat:
Closing in favor of #1990.

kleinerpirat:
@hgiesel I was just about to delete the branch, but saw you reacted with :confused: :D
Do you think this setting is worth it? One thing to keep in mind is that loading CodeMirror has a huge performance impact even on a moderate amount of fields.

rslib/src/notetype/schema11.rs
```diff
@@ -211,6 +214,9 @@ pub struct NoteFieldSchema11 {
     #[serde(default, deserialize_with = "default_on_invalid")]
     pub(crate) description: String,
 
+    #[serde(default, deserialize_with = "default_on_invalid")]
+    pub(crate) input: String,
```
dae, line 218: This will make it default to "", which will confuse the front-end code. Instead of using a string here which is not very space efficient, how about either a boolean, or if you anticipate there to be other kinds of editors in the future, an enum?

rslib/src/notetype/schema11.rs
```diff
@@ -164,6 +164,9 @@ fn clear_other_field_duplicates(other: &mut HashMap<String, Value>) {
     for key in &["description"] {
         other.remove(*key);
     }
+    for key in &["input"] {
+        other.remove(*key);
+    }
```
dae, line 169: input could be placed after description in the previous for statement instead

rslib/src/notetype/schema11.rs
kleinerpirat: I initially thought the string would be easier for others who see the field data for the first time, but actually a boolean is just as obvious, as long as the name is telling.

rslib/src/notetype/schema11.rs
```diff
@@ -161,10 +161,7 @@ impl From<Notetype> for NotetypeSchema11 {
 
 /// See [crate::deckconfig::schema11::clear_other_duplicates()].
 fn clear_other_field_duplicates(other: &mut HashMap<String, Value>) {
-    for key in &["description"] {
-        other.remove(*key);
-    }
-    for key in &["plain_text"] {
+    for key in &["description", "plain_text"] {
```
kleinerpirat, line 164: Like this?

dae: Pretty much, yep. For consistency with the other items like decks, recommend you change it to "plainText", and then update NoteFieldSchema11 like so:

#[derive(Serialize, Deserialize, Debug, Clone)]
#[serde(rename_all = "camelCase")]
pub struct NoteFieldSchema11 {

qt/aqt/editor.py
```diff
@@ -499,6 +499,7 @@ def loadNote(self, focusTo: int | None = None) -> None:
         ]
 
         flds = self.note.note_type()["flds"]
+        plain_texts = [fld.get("plain_text", False) for fld in flds]
```
dae, line 502: this will also need updating

dae:
Seems to work well - thanks Matthias!

PR #1989, opened 2022-07-27
Rename vars.scss to colors.scss / Introduce props.scss / Unify scrollbar design

kleinerpirat:
This is a groundwork-PR for design work down the line.

## Changes
- rename `_vars.scss` to `_colors.scss`
- introduce `_props.scss` for CSS property values other than colors
  - This will be the place where default sizes, box-shadows, transitions etc. are defined.
  - I included a redundant night mode definition for demo purposes. If a value isn't set for night mode, `extract_sass_props` assumes the day value.

### Included sample use cases
- make scrollbar appearance consistent with a single custom style for all platforms and themes
- use CSS variables for UI elements that change opacity

---
Let me know if this is a valid way forward or if you had other plans in mind. I'd also love to take up the work on #1704 to have a solid system before making actual design changes.

### DIFF
--- a/qt/aqt/theme.py
+++ b/qt/aqt/theme.py
@@ -145,13 +145,13 @@ class ThemeManager:
         "Returns body classes used when showing a card."
         return f"card card{card_ord+1} {self.body_class(night_mode)}"
 
-    def color(self, colors: tuple[str, str]) -> str:
-        """Given day/night colors, return the correct one for the current theme."""
+    def value(self, values: tuple[str, str]) -> str:
+        """Given day/night values, return the correct one for the current theme."""
         idx = 1 if self.night_mode else 0
-        return colors[idx]
+        return values[idx]
 
     def qcolor(self, colors: tuple[str, str]) -> QColor:
-        return QColor(self.color(colors))
+        return QColor(self.value(colors))
 
     def _determine_night_mode(self) -> bool:
         theme = aqt.mw.pm.theme()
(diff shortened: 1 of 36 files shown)


qt/aqt/theme.py
```diff
@@ -145,13 +145,13 @@ def body_classes_for_card_ord(
         "Returns body classes used when showing a card."
         return f"card card{card_ord+1} {self.body_class(night_mode)}"
 
-    def color(self, colors: tuple[str, str]) -> str:
-        """Given day/night colors, return the correct one for the current theme."""
+    def value(self, values: tuple[str, str]) -> str:
```
dae, line 148: An add-on may have used this, so we should probably add a deprecated alias with the old name.

dae:
The addition of new vars is nice. A separate file for props seems reasonable, though I wouldn't have minded if you'd stuffed them in _vars either.

I'm not sure that using the same scrollbar in day mode is actually a good thing. In day mode we use the platform theme, which is what the user is used to. In night mode we use a Qt theme so that we can properly control the colors, and the default scrollbar looks bad without custom styling. On the mobile clients, we'll want to avoid custom scrollbars altogether (so I'll need to fix the editor including it at one point). Which is more important in day mode - being consistent with the platform theme, or being consistent with night mode?

kleinerpirat:
Still sketching out ideas for proper theming support and might end up with a different approach. I'll keep this as a draft for now.

kleinerpirat:
I'm closing this in favor of #2016.

PR #1990, opened 2022-07-28
Introduce setting to collapse field by default

kleinerpirat:
image

#1987 should probably be a global setting somewhere in the preferences (or with an add-on), because it is a very niche requirement. A field-specific setting would not only litter the fields dialog with something most people will never use, but also be more cumbersome for the three people that want to have it as default input on every notetype.

So I repurposed the code for collapsing by default. If the setting is checked, the field will be fully collapsed (`richTextInput` and `plainTextInput`). Again, here's my future field vision:

image

Once the chevron is introduced, the term "collapsed" will make more sense.


dae:
@hgiesel I know we talked about this sort of thing in the past - do you have thoughts/preferences on how/when this is implemented?

rslib/src/notetype/schema11.rs
```diff
@@ -211,6 +214,9 @@ pub struct NoteFieldSchema11 {
     #[serde(default, deserialize_with = "default_on_invalid")]
     pub(crate) description: String,
 
+    #[serde(deserialize_with = "deserialize_bool_from_anything")]
```
dae, line 217:
```suggestion
    #[serde(default)]
```

Since this is a new key, we don't need to worry about legacy code accidentally using 0/1, and we don't want to error if the key is missing.

rslib/src/notetype/schema11.rs
```diff
@@ -164,6 +164,9 @@ fn clear_other_field_duplicates(other: &mut HashMap<String, Value>) {
     for key in &["description"] {
         other.remove(*key);
     }
+    for key in &["collapsed"] {
+        other.remove(*key);
+    }
```
dae, line 169: Same as other PR

dae (commented):
Since @hgiesel doesn't appear to have objections, and we already have a similar approach for plain_text, this LGTM.

hgiesel:
Sorry for not reacting 🤦‍♂️
Some months ago, I would have probably said, I'd prefer to put it after #1691, but by now, it probably doesn't matter, because I'll have to write the PR again anyway :)
It's also nice to get out some editor features out again.

kleinerpirat:
I've resolved the merge conflicts, but still have to ensure proper functionality within `NoteEditor`. I'm on holiday rn, so this will probably take a few more days.

dae:
No problem, please let me know when you're ready for a review. Enjoy the holiday :-)

kleinerpirat:
I noticed the API property `focusable` of PlainTextInput and RichTextInput isn't exactly true anymore due to `Collapsible` having an async transition, so I made `focusable` depend on that transition being finished.

Using `tick()` in a while-loop waiting for the inputs to be focusable again still results in an infinite loop, so there must be some inner Svelte stuff at play that I don't fully grasp yet. I used two successive `tick()` calls as a temporary workaround.

ts/editor/helpers.ts
```diff
@@ -19,3 +22,13 @@ export function withFontColor(
 
     return false;
 }
+
+/***
+ * Required for field inputs wrapped in Collapsible
+ */
+export async function refocusInput(input: RichTextInput | PlainTextInput) {
```
hgiesel: In the end you're just using `.api`, which is an object of type `EditingInputAPI`.

kleinerpirat: Thanks! Why does the check fail though? I've added the return type :eyes:

hgiesel: `api` is of type `EditingInputAPI`, not the entire RichTextInput / PlainTextInput

kleinerpirat: Yes, but I already change that with https://github.com/ankitects/anki/pull/1990/commits/87fb5f6c7c32f0deb5fc69a78e7f022b82cb3e8e...

kleinerpirat: Oh sorry i got it. I'll fix that once I'm home.

kleinerpirat: Turns out I didn't get it after all. TS is still uncharted terrain for me. I don't know why the function would expect an argument of type `EditingArea` if I typed it as `EditingInputAPI`.

kleinerpirat:
@dae I'm ready for a review now :)

dae:
I've only briefly skimmed over the TS/HTML (not my strongest area), but the Rust/Python changes look good and it seems to work well. Thanks Matthias!

PR #1992, opened 2022-07-29
added editor hook: add_cards_might_add_note

### LINKED ISSUE #1983: Adding optional reasons to reject in add_cards_will_add_note
Currently, the `add_cards_will_add_note hook` does not have an option for add-ons to give a warning to a user. An example is an add-on to check if a note does not contain any tags before it is added. If so, then it should ask the user to confirm that they want to add a card without tags. (An example of this add-on using monkey patching is [here](https://www.reddit.com/r/Anki/comments/coc6wp/comment/ewi7m2m/?utm_source=share&utm_medium=web2x&context=3))

I can think of two possible ways to achieve this. The first option would be to add another hook, e,g add_cards_might_add_note. The second option would be to provide another argument to add_cards_will_add_note (a list of strings, initialised to be empty) and opening an askUser prompt for each string in the list. However this would break current add-ons that use the hook. 

Let me know which way is preferred and I'll open a PR :)

sam1penny:
Resolves #1983 

Let me know if the naming is confusing - I run the hook after `add_cards_will_add_note` since there is no point asking the user for confirmation about an optional problem if there will be a problem found in the `add_cards_will_add_note` hook.

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -296,6 +296,11 @@ class AddCards(QMainWindow):
             showWarning(problem, help=HelpPage.ADDING_CARD_AND_NOTE)
             return False
 
+        optional_problems: list[str] = []
+        gui_hooks.add_cards_might_add_note(optional_problems, note)
+        if not all(askUser(op) for op in optional_problems):
+            return False
+
         return True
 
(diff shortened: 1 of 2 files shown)


dae:
Seems reasonable; thanks Sam.

PR #1995, opened 2022-07-31
Use wavy flag icons (flag-alternative)

hgiesel:
We used the blocky flag icon, instead of the wavy one, because it had more modifiers.
Well, now the wavy flag has just as much :)


dae:
Thanks for pushing that through :-)

PR #1998, opened 2022-08-01
fix progress.update(value=0) to not increment progress

BlueGreenMagick:
This fixes a bug where calling `progress.update(value=0)` multiple times increments `self._counter`.


dae:
Thanks BGM!

PR #2000, opened 2022-08-02
Serialize modal popups at startup

### LINKED ISSUE #1988: Check we're not presenting multiple modal dialogs at once on startup
The sync on startup, add-on update notification, and Anki update notification dialogs are all modal. We should make sure we don't present more than one at once, perhaps by only triggering the update checks after an initial sync, and only one after the other.

https://forums.ankiweb.net/t/dialogue-box-collision-on-startup/21724

sam1penny:
Fixes #1988 by calling `maybe_check_for_addon_updates()` after an initial sync

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -980,7 +980,8 @@ title="{}" {}>{}</button>""".format(
 
         def after_sync_and_call_addon_update() -> None:
             after_sync()
-            self.maybe_check_for_addon_updates()
+            if not self.safeMode:
+                self.maybe_check_for_addon_updates()
 
         if self.can_auto_sync():
             self._sync_collection_and_media(after_sync_and_call_addon_update)
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -894,15 +894,15 @@ title="{}" {}>{}</button>""".format(
             self.addonManager.loadAddons()
 
     def maybe_check_for_addon_updates(
-        self, on_done: Callable[[list[DownloadLogEntry]], None] | None = None
+        self, on_done: Callable[[], None] | None = None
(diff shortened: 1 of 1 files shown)


dae:
Thanks Sam! I'm not sure we can close #1988 yet though, as the "anki updated" popup still needs to be delayed until the others have run (or vice versa). Would you like to try handle that too in this PR?

sam1penny:
> Thanks Sam! I'm not sure we can close #1988 yet though, as the "anki updated" popup still needs to be delayed until the others have run (or vice versa). Would you like to try handle that too in this PR?

Sure - how about like this?

I might've overcomplicated it a bit by making it as general as possible - e.g providing an optional `on_done` argument to `maybe_check_for_addon_updates` even though it's only called once. Let me know if you want me to make any changes :)

qt/aqt/main.py
```diff
@@ -893,17 +892,23 @@ def setupAddons(self, args: list | None) -> None:
...
+            if on_done:
+                on_done(log)
+            self.on_updates_installed(log)
+
         if elap > 86_400 or self.pm.last_run_version() != point_version():
             check_and_prompt_for_updates(
                 self,
                 self.addonManager,
-                self.on_updates_installed,
+                wrap_on_updates_installed,
```
dae, line 911: This skips calling on_done when time hasn't elapsed. Not a big deal in the grand scheme of things, but it's inconsistent with the auto sync code that always calls the callable.

dae (commented):
on_done seems like a reasonable approach to me :-)

sam1penny:
Thanks for the review :)

dae:
Thanks Sam!

PR #2001, opened 2022-08-03
Updated permalink to architecture visualization

phwoo:
The existing link in the docs 404's, so I've just updated to what seems to be the new url.
[The docs](https://githubnext.com/projects/repo-visualization#integrate-into-your-own-projects) for that project suggest using GitHub actions to generate an svg file, but it might be possible to do this with BuildKite?

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -8,6 +8,7 @@ At the highest level, Anki is logically separated into two parts.
 
 A neat visualization of the file layout is available here:
 <https://mango-dune-07a8b7110.1.azurestaticapps.net/?repo=ankitects%2Fanki>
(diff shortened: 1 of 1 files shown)


dae:
Maybe it would be less likely to break if we were to link to https://githubnext.com/projects/repo-visualization#explore-for-yourself and tell the user to input ankitects/anki? It could be done with a GitHub action, but I'd rather avoid that complexity if possible.

phwoo:
I've included both, with the link to the project as a backup. Hopefully at least one of those stays up.
And I agree about the complexity, probably not worth it unless there were a number of pieces of generated documentation.

dae:
Thanks phwoo!

PR #2002, opened 2022-08-03
Field redesign

kleinerpirat:
This is pretty much finished, except the CSS rules of `PlainTextBadge`, where positions are not yet fully calculated, but hard-coded.

image

I'm off for the next three days, so I wanted to put this out there quickly for you to test.


kleinerpirat:
Getting an error on ` bazel run eslint -- --fix` from [development.md](https://github.com/ankitects/anki/blob/main/docs/development.md#fixing-formatting):

```
ERROR: Skipping 'eslint': no such target '//:eslint': target 'eslint' not declared in package '' defined by /home/kleinerpirat/Documents/GitHub/anki/BUILD.bazel
WARNING: Target pattern parsing failed.
ERROR: no such target '//:eslint': target 'eslint' not declared in package '' defined by /home/kleinerpirat/Documents/GitHub/anki/BUILD.bazel
```

dae:
How does the user access the HTML editor now?

kleinerpirat:
By clicking anywhere on the bottom border of `RichTextInput`. There's a triangular indicator that shows on hover. For mobile I'd probably go for a swipe gesture.

dae:
I'm afraid I'm not sure about this change Matthias. On the one hand, it is smoothly implemented, and slightly reduces the button count, which makes things feel less cluttered. But that comes at the cost of discoverability, and even on desktop I had to ask you how to do it :-) Without your help, I probably would have stumbled across it at one point, but generally speaking, I think it's best if the available actions are immediately apparent, and not hidden.

Another downside of having removed the button is that there's no text that explains what the feature does, or its associated shortcut anymore.

And one final thought is that this new approach assumes the user will always want to have the visual editor shown. We have received some requests in the past from users who wanted to see the HTML editor without the visual editor, and that doesn't seem to be possible with this approach. To be fair though, it's not something that is frequently requested.

kleinerpirat:
> I think it's best if the available actions are immediately apparent, and not hidden.

Fair point. Then again, most interactions on mobile apps are hidden and only discoverable through intuition. One could of course argue that my proposal is unintuitive and therefore isn't a suitable option to declutter the fields.

> Another downside of having removed the button is that there's no text that explains what the feature does, or its associated shortcut anymore.

That would be easy to implement. I published this PR for you to test the functionality - polish would have come if it were approved.

> We have received some requests in the past from users who wanted to see the HTML editor without the visual editor, and that doesn't seem to be possible with this approach.

This was meant to be adressed with https://github.com/ankitects/anki/pull/1987.

---

WDYT about the other changes @dae? So far I have only received feedback on the HTML editor toggle, which is the latest thing I had been working on, but there are other changes too:

- chevron icon to collapse whole field instead of "Toggle visual editor"
- buttons only show on field focus and hover instead of all the time
- different border (only around actual input)
- slight box-shadow for field
- different focus effect (thinner, inset)
- different tag icon (outline instead of filled)

dae:
> chevron icon to collapse whole field instead of "Toggle visual editor"

It looks good, though there is the issue mentioned above that this no longer allows the HTML editor to be shown without the visual editor.

> buttons only show on field focus and hover instead of all the time

While it looks neater, this is another discoverability issue, especially on mobile. And even on desktop, I'm not sure it's a great user experience that users have to hover their mouse around the screen trying to reveal the various locations that a hidden toggle has been added. :-)

> different border (only around actual input)

It feels a bit less cluttered that what we had before, so that's a :+1: from me.

> slight box-shadow for field

I like that it's subtle; I didn't really notice it.

> different focus effect (thinner, inset)

Looks clean, though I get a little artifacting on the corners on this non-HiDPI system:
2022-08-08T11:51:09,386280944+10:00

> different tag icon (outline instead of filled)

I think it looks cleaner.

Anyway, I'm no UI expert. Any other people want to weigh in? :-)

dae:
> It's quite common in today's software to reveal extra options only on hover.

The difference between the old implementation and your examples is that a) the hover zone is much larger (eg, not just active in a little corner), and b) when the area is focused, the hover elements are also visible. I see you've applied a similar approach in your latest changes, and I like it a lot more :-)

Some minor comments:

- the distance between the HTML button and the pin below it is quite small, and users may struggle to tap on one over the other
- do you have the hover set up to fade in or display after a delay? It feels a bit "laggy"
- if the user clicks on one field, then tabs into another, both hovered and focused fields show the icons, which feels a bit busy. VS Code does this too, though the different frames tend to be spaced further apart

Not sure if it's a good idea, but what do you think about dropping the hover and relying solely on focus? It would mean toggling the pin or HTML mode would require a click into the field, then another on the revealed button. Not sure whether that would be a pain or not in practice.

kleinerpirat:
I have now partially reverted to the previous HTML icon position. The main reason was me testing the approach with the bottom border position on actual filled out notes.

Some fields are filled with so much content that it overflows - in such cases it feels really weird when the controls are spread out between top and bottom borders. I then tried a responsive solution with an `IntersectionObserver` (HTML toggle switches to the top when the field is clipped), but that didn't feel right either.

So I decided to revert to the old badge position, but still fixed the issue of long overflowing fields by making `LabelContainer` sticky. This way the `FieldState` controls are always accessible, meaning you can toggle the HTML editor on and off anytime.

dae:
The smaller icon in a single place looks cleaner. We still have an issue with the icons being too close to one another for tapping, but that was an issue that existed previously too. :-)

Minor bug report: toggling visibility on/off for a field shows this in the console:

JS error /_anki/js/reviewer_editor.js:78218 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'refresh')

> Currently I have implemented animations only in the "out" direction - they appear instantly and fade out slightly. This feels responsive and slick to me, WDYT?

Feels much more responsive, and the smaller icon is less distracting that "HTML" - that works for me. Thanks for all your efforts on improving this screen Matthias :-)

@hgiesel would be great if you could weigh in as well, as this screen is largely your baby :-)

kleinerpirat:
> `JS error /_anki/js/reviewer_editor.js:78218 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'refresh')`

I think I'll have to replace the Svelte transitions, which are triggered by `{#if}` blocks and require destroying/re-mounting the components, with CSS animations or use `create_out_transition` from `svelte/internal`.

kleinerpirat:
From my side, this is ready for review @hgiesel.

---
There's one little issue left:

- [ ] The first expansion of `Collapsible.svelte` isn't animated. The animation relies on the `clientHeight` property, which is initially 0 for elements that are collapsed by default.

ts/editor/rich-text-input/RichTextInput.svelte
```diff
@@ -200,45 +201,48 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-            use:attachStyles
-            use:attachContentEditable={{ stylesDidLoad }}
-            on:focusin
-            on:focusout
-        />
-
-        {#await Promise.all([richTextPromise, stylesDidLoad]) then _}
-            <div class="rich-text-widgets">
-                <slot />
+<Collapsible collapsed={hidden}>
```
hgiesel, line 204: I wonder if instead of using Collapsible inside RichTextInput (and PlainTextInput), you could also apply it in the user component, and pass in `hidden` with `bind:`. It's something I'd like to try out.

ts/components/WithTooltip.svelte
```diff
@@ -38,17 +38,6 @@
     }
 
     onDestroy(() => tooltipObject?.dispose());
-
-    // hack to update field description tooltips
-    let previousTooltip: string = tooltip;
-    $: if (tooltip !== previousTooltip) {
```
hgiesel, line 44: Just a reminder: We should test every location, where we use WithTooltip. These components have proven to be very fuzzy, and they broke on me in cases where I thought they shouldn't.

ts/editor/EditingArea.svelte
```diff
@@ -189,10 +194,25 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+
+        box-shadow: 0px 0px 2px 0px var(--border);
+        transition: box-shadow 80ms cubic-bezier(0.33, 1, 0.68, 1);
 
-        &:focus {
+        &:focus-within {
             outline: none;
+            &::after {
+                content: "";
+                position: absolute;
```
hgiesel, line 207: This also asks for a comment

ts/editor/SecondaryInputBadge.svelte
```diff
@@ -9,15 +9,16 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
 
     const editorField = editorFieldContext.get();
     const keyCombination = "Control+Shift+X";
     const dispatch = createEventDispatcher();
 
     export let off = false;
+    export let defaultInput: "plain" | "rich" = "rich";
 
-    $: icon = off ? htmlOff : htmlOn;
+    $: icon = defaultInput == "rich" ? plainTextIcon : richTextIcon;
```
hgiesel, line 21: I think you could leave this as the `PlainTextBadge`.
It seems a bit silly to generalize this component to serve as both the badge for the plain text and the rich text, while still having other badges like the sticky badge, which also mostly look the same.

hgiesel (changes requested):
First of all: It looks absolutely grand! I love it.

Feature wise:
- I wonder if this is a bug:
  1. Open the code input
  2. Collapse the field
  3. The code badge will still indicate an open code input, which would be a bug, but I guess you could also interpret it as "the code input once you uncollapse the field again", which would not be a bug :)
- I don't like the collapsing if you click anywhere on the label container. In fact, you can accidentally collapse the field if you click between the badges. I'd prefer if it was only on the chevron, or possibly the field name.

I've also added quite a few comments. It's nothing serious however, just some questions and refactorings.

kleinerpirat:
Thank you for this thorough review, Henrik!

> The code badge will still indicate an open code input, which would be a bug, but I guess you could also interpret it as "the code input once you uncollapse the field again", which would not be a bug :)

Originally the chevron would collapse both inputs, with https://github.com/ankitects/anki/pull/2002/commits/3dca559f88e1eda6cb217e26430f8b3db7d26298, these states got decoupled. Like you said, it feels a bit unintentional and the more natural behavior would be to collapse the secondary input together with the field - https://github.com/ankitects/anki/pull/2002/commits/75b2bfe5974a15a8b11c57b22cd1af4e62096352.

Also, the secondary input toggle should be hidden when the field is collapsed, because it has no apparent effect in this state. https://github.com/ankitects/anki/pull/2002/commits/2b3d86804f8ca3db93d092c80ce2ea6cba09148c and https://github.com/ankitects/anki/pull/2002/commits/37472dcea757cc1e3b55413d4933a1b82670b62e take care of that.

ts/editor/SecondaryInputBadge.svelte
```diff
@@ -37,20 +38,30 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-        opacity: 0.4;
-
+        cursor: pointer;
+        opacity: 0;
         &.highlighted {
             opacity: 1;
         }
+    }
+    :global(.editor-field) {
+        &:focus-within,
```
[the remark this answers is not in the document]
kleinerpirat: I'd prefer to define such variables on the *higher level* (e.g. _props.scss from https://github.com/ankitects/anki/pull/1989). The 0.4 - 0.8 - 1.0 steps are used for all badges, so I'd like to define them in one place.

This will be done in one sweep with a coming up PR.

I went for a `visible` boolean that is passed down from `NoteEditor.svelte`: https://github.com/ankitects/anki/pull/2002/commits/37472dcea757cc1e3b55413d4933a1b82670b62e.

ts/components/Collapsible.svelte
```diff
@@ -0,0 +1,73 @@
...
+
+    let style: string;
+    const setStyle = (el: HTMLElement) => {
+        height = el.clientHeight;
+        style = `--collapse-height: -${height}px; --duration: ${Math.sqrt(
+            height * 80,
+        )}ms`;
+    };
+
+    const transition = async (collapse: boolean) => {
```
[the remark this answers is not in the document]
kleinerpirat: Done: https://github.com/ankitects/anki/pull/2002/commits/825170ea882da247d4273f6c81807a4c2efbe072

ts/components/Collapsible.svelte
```diff
@@ -0,0 +1,73 @@
...
+        isCollapsed = true;
+
+        if (collapse) {
+            setStyle(inner);
+
+            inner.addEventListener(
+                "transitionend",
+                () => {
+                    // DOM-manipulation is required here
+                    inner.setAttribute("hidden", "");
```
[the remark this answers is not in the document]
kleinerpirat: The issue I faced with Svelte transitions is that they're only triggered when an element enters or leaves the DOM (i.e. with `{#if}` blocks). `Collapsible` wouldn't be a safe-to-use component if it had to destroy and re-mount its contents.

I tried it before (https://github.com/ankitects/anki/pull/2002/commits/9a2b3410d0ead37ae1da408d68e14507a058a613) and it lead to this error @dae mentioned in a [comment above](https://github.com/ankitects/anki/pull/2002#issuecomment-1212577220):

> JS error /_anki/js/reviewer_editor.js:78218 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'refresh')

because an array used in `NoteEditor.svelte` got shorter with each toggle until there were no elements left.

ts/editor/LabelContainer.svelte
```diff
@@ -3,38 +3,78 @@ Copyright: Ankitects Pty Ltd and contributors
...
-    class:rtl={$direction === "rtl"}
     on:mousedown|preventDefault
+    on:click|stopPropagation={toggle}
 >
+    <span class="chevron" class:collapsed>
+        <Badge {tooltip} iconSize={80} --icon-align="text-bottom"
+            >{@html chevronDown}</Badge
+        >
+    </span>
+    <slot name="field-name" />
```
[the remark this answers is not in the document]
kleinerpirat: If we're making the collapse click-area smaller like you suggested, I think these slots need to be separated to apply `cursor: pointer` only to clickable elements.

ts/editor/LabelContainer.svelte
```diff
@@ -3,38 +3,78 @@ Copyright: Ankitects Pty Ltd and contributors
...
+        position: sticky;
+        top: 0;
+        z-index: 3;
+        background: var(--window-bg);
+        cursor: pointer;
+
+        &::before {
+            content: "";
+            z-index: -1;
+            position: absolute;
```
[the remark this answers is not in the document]
kleinerpirat: https://github.com/ankitects/anki/pull/2002/commits/8697e460f73ca890a132b67c964365b8e19daddb

ts/editor/LabelContainer.svelte
```diff
@@ -3,38 +3,78 @@ Copyright: Ankitects Pty Ltd and contributors
...
     class="label-container"
-    class:rtl={$direction === "rtl"}
     on:mousedown|preventDefault
+    on:click|stopPropagation={toggle}
 >
+    <span class="chevron" class:collapsed>
+        <Badge {tooltip} iconSize={80} --icon-align="text-bottom"
+            >{@html chevronDown}</Badge
+        >
+    </span>
```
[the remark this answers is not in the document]
kleinerpirat: True! There you go: https://github.com/ankitects/anki/pull/2002/commits/bd5873b5327d0e8c9168842e1aad07368b2db4b5

ts/editor/EditingArea.svelte
kleinerpirat: I had to delete that one to find out why it was required.

The highlight box-shadow is inset now (to not mess up alignment with field borders) and has to be displayed above the field contents. `PlainTextInput` covers the shadow when it's applied to `EditingArea` directly, so I went for a pseudo element.

Comment added in: https://github.com/ankitects/anki/pull/2002/commits/171cc063a22a7efbbeb5b34e2416816553fc1afc

ts/editor/plain-text-input/PlainTextInput.svelte
```diff
@@ -164,15 +166,12 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-            display: none;
+            padding: 8px 0;
         }
     }
 
-    .light-theme :global(.CodeMirror) {
-        border-top: 1px solid #ddd;
+    :global(.CodeMirror) {
+        border-top: 1px solid var(--border);
+        background: var(--code-bg);
```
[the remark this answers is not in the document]
kleinerpirat: https://github.com/ankitects/anki/pull/2002/commits/dc8a0370ef4976adbf89308dcaf84bf4925e99ac

ts/editor/StickyBadge.svelte
```diff
@@ -37,20 +35,31 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-        opacity: 0.4;
-
+        cursor: pointer;
+        opacity: 0;
         &.highlighted {
             opacity: 1;
         }
+    }
+    :global(.editor-field) {
+        &:focus-within,
```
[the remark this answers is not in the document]
kleinerpirat: Applying the hover/focus boolean to this one seems a bit more tricky because `StickyBadge` is located in `NoteCreator`. Can I move it down to `NoteEditor`?

ts/editor/rich-text-input/RichTextInput.svelte
kleinerpirat: That was easier than expected :smile: https://github.com/ankitects/anki/pull/2002/commits/9988c000c16f5ec3d7fbbc84abc1265f35aead9d

ts/editor/EditingArea.svelte
kleinerpirat: If there is a more elegant way to do it, I'm all ears. This hard-coded approach will need adjusting any time the gutters are changed in `Fields.svelte`, so it's far from ideal.

ts/components/WithTooltip.svelte
kleinerpirat: They're currently only used in the deck options and tag editor. Both seem to have weathered the change :)

ts/editor/LabelContainer.svelte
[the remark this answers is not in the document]
kleinerpirat: I found a way to do it without pseudo-elements: https://github.com/ankitects/anki/pull/2002/commits/2b27724e6723b1d152311c59712491bd790a2ba9

hgiesel:
Unfortunately, currently the in-line Mathjax seems to be broken by this. The empty Mathjax takes the whole width of the screen, and the overlay is constrained by editor field.

The overlay should probably be fixed by using the `WithFloating` component for the menu instead of bootstrap.

ts/editor/StickyBadge.svelte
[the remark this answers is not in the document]
kleinerpirat: Nevermind, I just found out about slot props :)

But I think I found a Svelte bug with https://github.com/ankitects/anki/pull/2002/commits/2806c617b31fa3f7475150fc3397278c5f9e47db.

The `active` prop of `StickyBadge` resets to false as soon as the new `visible` prop is updated. This makes the sticky badge disappear on mouseleave/blur, regardless of its back-end value. I can't for the life of me figure out what I've done wrong to cause this issue.

kleinerpirat:
> The empty Mathjax takes the whole width of the screen

It does this on main too, I guess since #1859.

ts/components/Collapsible.svelte
```diff
@@ -0,0 +1,76 @@
...
+        {style}
+    >
+        <slot />
+    </div>
+</div>
+
+<style lang="scss">
+    .collapsible-container {
+        position: relative;
+        overflow: hidden;
```
kleinerpirat, line 64: The MathJax overlay getting constrained is an unintentional side-effect of this line, which means I'll just have to use a different method for the slide effect.

hgiesel: I wouldn't be too fast here. The MathJax overlay being nested that deeply into the the HTML tree is the actual problem here. `WithFloating` works quite differently to that, by removing the component from the tree and adding it at the bottom of `document.body`:

Bildschirmfoto 2022-08-14 um 12 22 40

kleinerpirat: Beccause the margin-top + clipping approach is more natural-looking than other solutions involving height or scaleY(), I decided to keep it and hide the overflow only during transition (https://github.com/ankitects/anki/pull/2002/commits/721ab0a23a7e208ea96f01939e9d6496c347255f).

kleinerpirat: @hgiesel Sorry, the GitHub page I had open didn't update so I missed your comment.

kleinerpirat: If you don't mind, I would leave the MathJax handle untouched in this PR because my recent changes don't make this an immediate issue anymore.

hgiesel: No worries, I already started working on it: #2011

hgiesel:
I will try to give this another thorough review soon, but one thing I already noticed is, is that you cannot unselect fields anymore by clicking on empty space without fields.

kleinerpirat:
> one thing I already noticed is, is that you cannot unselect fields anymore by clicking on empty space without fields.

I can't make out a difference to current main :thinking:

ts/editor/FieldState.svelte
```diff
@@ -11,9 +11,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         display: flex;
         justify-content: flex;
         flex-grow: 1;
-
-        & > :global(*) {
-            margin-left: 2px;
-        }
+        gap: 6px;
```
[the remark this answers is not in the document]
dae: Unfortunately not - the minimum target is currently Chromium 77 (for the Qt5 Mac build), and iOS 13.4. Bumping iOS to 14.5 may be possible in a few months when iOS 16 comes out, but even if we switched the Mac build to Qt 5.15, it would still not be enough for this, as 5.15 is Chromium 83: https://caniuse.com/?search=flex-gap

hgiesel:
> Is this something we should support again?
Hm, I think the current behavior feels quite buggy. However, in that case, it is not in the scope of this PR.

hgiesel:
I've merged in main for your PR (as I've created most of the conflicts):
https://github.com/kleinerpirat/anki/compare/fresh-fields...hgiesel:anki:fresh-fields

kleinerpirat:
Thanks a lot Henrik, that was a great help!

I don't think there are any major showstoppers left. One thing to keep in mind: super-users and people with low-end machines or high # of fields will most likely demand an opt-out for the CSS transitions. But that can be done in a separate PR too.

ts/editor/FieldState.svelte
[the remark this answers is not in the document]
hgiesel: @kleinerpirat This is still needs to be adressed unfortunately. In the past I've solved it by using `... > :global(*) { margin-left: npx }`

hgiesel:
Actually I thought this could be merged before the Mathjax PR. I also noticed that the Mathjax overlay is more broken again, have you reverted some fixes here?

ts/editor/FieldState.svelte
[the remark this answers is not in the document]
kleinerpirat: Forgot about that one, thanks :v:

kleinerpirat:
> Mathjax overlay is more broken again, I wonder why that is

Me too. Back when I adressed it with https://github.com/ankitects/anki/pull/2002/commits/a3a26bc70ef879f20d582013131b7c03411f08c8 the controls were still displaying fine. I'll investigate in the evening.

ts/editor/FieldState.svelte
```diff
@@ -11,6 +11,27 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         display: flex;
         justify-content: flex;
         flex-grow: 1;
-        gap: 6px;
+
+        /* replace these global selectors with "grid-gap: 6px" ASAP */
```
hgiesel, line 15: I think you mean `flex-gap`, or rather just `gap` (as gap sets both `grid-gap` and `flex-gap`)

ts/editor/Fields.svelte
```diff
@@ -10,7 +10,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     .fields {
         display: grid;
         grid-auto-rows: min-content;
-        grid-gap: 6px;
```
hgiesel: You don't need to replace grid-gap.

`grid-gap` is well supported, the only thing not supported is `flex-gap`, or in other words, using `gap` with `display: flex`

ts/editor/FieldState.svelte
kleinerpirat: Oh, I was copy-pasting this comment and didn't notice. Fixing it right now.

kleinerpirat:
@hgiesel the MathJax controls got buggy with #1931.

ts/components/Collapsible.svelte
```diff
@@ -0,0 +1,99 @@
...
+            { once: true },
+        );
+    }
+
+    /* prevent transition on mount for performance reasons */
+    let blockTransition = true;
+
+    $: {
+        if (blockTransition) blockTransition = false;
+        else transition(collapsed);
```
dae, line 73: Please always use {} for if/else

ts/components/Collapsible.svelte
```diff
@@ -0,0 +1,99 @@
...
+
+        if (!collapse) {
+            inner.removeAttribute("hidden");
+            isCollapsed = false;
+        }
+
+        inner.addEventListener(
+            "transitionend",
+            () => {
+                if (collapse) inner.setAttribute("hidden", "");
```
dae, line 61: Please always use {} for if/else

ts/editor/FieldState.svelte
```diff
@@ -12,8 +12,26 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         justify-content: flex;
         flex-grow: 1;
 
-        & > :global(*) {
-            margin-left: 2px;
+        /* replace with "gap: 6px" as soon as possible */
```
dae, line 15: Might be worth mentioning in the comment that it's blocked on an update to Chromium 84 (Qt6 only) and iOS 14.1 Same in FieldsEditor

dae (commented):
Some minor style comments, otherwise seems good to me. Once the merge conflict is fixed, happy to merge assuming @hgiesel has no objections

kvn1351:
Wow, I love the look of this. Can't wait.

dae:
:rocket:

kleinerpirat:
Thanks Damien! I just noticed the weird commit mess I caused with that last merge ^^ Thankfully you always squash those 😅

dae:
Yep, that was impressive :-)

PR #2008, opened 2022-08-12
Make note view switch knob smaller than its path

kleinerpirat:
@RumovZ We have discussed the look of this knob extensively on #1182. I remember you said:

> I feel that the knob should either be smaller or larger than the path.

Once we're using a brighter background one or two PRs down the road, I think we should switch to the smaller variant, otherwise there's no contrast:

### current state | PR
 

*Actually I enlarged the path rather than decrease the knob's size, but everything is relative ;)*

The switches in the deck options also look like this, so it will make things more consistent, too.


dae:
Thanks Matthias!

Issue #1774, opened 2022-04-05
Rationalising info and confirmation dialogs in aqt.utils


roxgib:
I've been looking at a number of the dialogs in `qt/aqt/utils.py` as the next target for removing `exec()`, specifically those showing information or getting a response from the user, but looking over them I think a bigger refactoring might be worthwhile. The main issues I see are:
- A lot of overlap between functions
- Many functions doing too much 
- Ineffective split between class and function code

To fix this I'd like to:
-  Create a new `InfoDialog` class for displaying a message dialog with a confirmation button, and a few convenience functions along the lines of `showWarning`, `showCritical`, `showInfo`, `showText`. Existing callers of those functions that simply need to show some information can be easily transitioned. 
- Rewrite `ButtonedDialog` to utilise callbacks (to avoid the use of exec), with new convenience functions that replace `askUser`, `askUserDialog`, as well as `showInfo` where that is used currently to get a confirmation from the user (e.g. in `aqt.addons.installAddonPackages`). 

I think a cleaner split between showing information and getting confirmation/answer from the user would greatly simplify this part of the code, although I realise these functions are used in a lot of places and there may need to be more variations than listed above. In those cases, at least, things can still be improved with simpler and smaller scoped functions, letting the classes do the heavy lifting behind the scenes where necessary.

Some of the calls to showInfo and showText could instead be replaced with tooltips e.g. when a user tries to rebuild or empty a regular deck, or tries to undo when there's nothing to undo - these situations probably don't warrant a dialog popup in my opinion. This can be discussed in the PRs of course.

I assume a fair few add-ons use these functions so they'll have to stick around for a while, but I still think they are worth replacing rather than refactoring. New functions makes it easier for existing callers to be moved incrementally, and they'd otherwise end up more complicated rather than less.  

I'll submit the changes incrementally, implementing the new functions and transitioning a few callers at a time to keep things manageable. I'm sure you're busy with 2.1.50, so no rush with this, but I wanted to get the go ahead and any guidance you might have before I dive in it since the changes are larger in scope and will take a bit of work.

dae:
> I assume a fair few add-ons use these functions so they'll have to stick around for a while, but I still think they are worth replacing rather than refactoring. New functions makes it easier for existing callers to be moved incrementally, and they'd otherwise end up more complicated rather than less.

[...]

PR #2010, opened 2022-08-14
Add MessageBox class and associated funcs to aqt.utils and update the first few callers

roxgib:
First step in #1774. 

Implements a new `MessageBox` class, and the functions `ask_user`, `ask_user_dialog`, `show_info`, `show_warning`, `show_critical`, which all return an instance of MessageBox. This class is used to display a message to the user with some buttons that optionally trigger a callback which receives the name of the button pressed. This avoids the use of `exec()` in favour of `open()` for the reasons discussed in #987 and #1774 along with cleaning up this section of the code a bit.

Some design considerations to note:
- It turned out to be easiest to have a single class to handle both messages to the user and asking questions of the user. These were previously handled by separate but overlapping functions/classes. 
- Most of convenience functions wrap the callback function in a lambda to achieve more natural types for the callback's argument. By default MessageBox expects the callback to take a single str argument (the button pressed), but this becomes a bool for `ask_user` and is removed for all the info functions (since the only purpose of the callback in those functions is to delay an action until the user clicks 'Ok'). This is a little clunky, but should result in a smoother experience when using these functions.
- I've kept the arguments for the convenience functions to a minimum. If the other arguments need to be accessed it's possible to pass additional keyword arguments through to MessageBox or simply call MessageBox directly.
- Arguments for the icon and text format now expect Qt enums rather than strings. This gives more flexibility to the caller and removes the need for an unnecessary conversion. An alternative for the text format would to use `Qt.TextFormat.AutoText` instead of `PlainText` as the default.
- I haven't tried to have this class handle the role of `showText`, but this could probably be done without too much bloat

Happy to reconsider any of the above points as well the names of the class or functions.

I believe these functions should be able to handle all the existing callers, so if this is accepted I'll start converting the callers a few at a time - there's always the potential for bugs since these functions previously blocked the main loop while the dialog was open, which callers sometimes rely upon in unexpected ways. In this PR three calling functions (two in `addons.py`, one in `sync.py`) have been converted as a demonstration.

One known issue - the buttons are now left aligned, whereas before they were right aligned. I need to dig a bit more into what's causing that. I also haven't done much testing with the Qt5 version since I'm not sure how long that's going to stick around - can any guidance be provided on that? I'm marking this as draft while these two points are addressed.

### DIFF
--- a/qt/aqt/sync.py
+++ b/qt/aqt/sync.py
@@ -26,20 +26,9 @@ from aqt.qt import (
     QVBoxLayout,
     qconnect,
 )
-from aqt.utils import (
-    askUser,
-    askUserDialog,
-    disable_help_button,
-    showText,
-    showWarning,
-    tr,
-)
+from aqt.utils import askUser, disable_help_button, showText, showWarning, tr
 
-
-class FullSyncChoice(enum.Enum):
-    CANCEL = 0
-    UPLOAD = 1
-    DOWNLOAD = 2
+from .utils import ask_user_dialog
 
 
 def get_sync_status(
@@ -135,13 +124,26 @@ def full_sync(
     elif out.required == out.FULL_UPLOAD:
         full_upload(mw, on_done)
     else:
-        choice = ask_user_to_decide_direction()
-        if choice == FullSyncChoice.UPLOAD:
-            full_upload(mw, on_done)
-        elif choice == FullSyncChoice.DOWNLOAD:
-            full_download(mw, on_done)
-        else:
-            on_done()
+        button_labels = [
+            tr.sync_upload_to_ankiweb(),
+            tr.sync_download_from_ankiweb(),
+            tr.sync_cancel_button(),
+        ]
+
+        def callback(choice):
+            if choice == button_labels[0]:
+                full_upload(mw, on_done)
+            elif choice == button_labels[1]:
+                full_download(mw, on_done)
+            else:
+                on_done()
+
+        ask_user_dialog(
+            tr.sync_conflict_explanation(),
+            callback=callback,
+            buttons=button_labels,
+            default_button=button_labels[2],
+        )
 
 
 def confirm_full_download(mw: aqt.main.AnkiQt, on_done: Callable[[], None]) -> None:
@@ -275,23 +277,6 @@ def sync_login(
         lambda: mw.col.sync_login(username=username, password=password),
         on_future_done,
     )
-
-
-def ask_user_to_decide_direction() -> FullSyncChoice:
-    button_labels = [
-        tr.sync_upload_to_ankiweb(),
-        tr.sync_download_from_ankiweb(),
-        tr.sync_cancel_button(),
-    ]
-    diag = askUserDialog(tr.sync_conflict_explanation(), button_labels)
-    diag.setDefault(2)
-    ret = diag.run()
-    if ret == button_labels[0]:
-        return FullSyncChoice.UPLOAD
-    elif ret == button_labels[1]:
-        return FullSyncChoice.DOWNLOAD
-    else:
-        return FullSyncChoice.CANCEL
 
 
 def get_id_and_pass_from_user(
--- a/qt/aqt/sync.py
+++ b/qt/aqt/sync.py
@@ -145,7 +145,7 @@ def full_sync(
         ask_user_dialog(
             tr.sync_conflict_explanation(),
             callback=callback,
-            buttons=button_labels,  # type: ignore
+            buttons=button_labels,
             default_button=2,
         )
 
--- a/qt/aqt/sync.py
+++ b/qt/aqt/sync.py
@@ -134,10 +134,10 @@ def full_sync(
             tr.sync_cancel_button(),
         ]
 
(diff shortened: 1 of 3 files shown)


roxgib:
Apologies, I had disabled the mypy tests while I worked on this, and forgot to re-enable them before submitting it.

While fixing the annotations, I realised that since the existing classes used QMessageBox.StandardButton, the translations might not be available for all of them, and also that I had used untranslated literals in some cases, so I've fixed that as well. MessageBox now accepts list[str | QMessageBox.StandardButton] for the buttons.

I should test the last few changes a bit more before proceeding.

qt/aqt/utils.py
```diff
@@ -70,6 +120,107 @@ def openLink(link: str | QUrl) -> None:
         QDesktopServices.openUrl(QUrl(link))
 
 
+class MessageBox(QMessageBox):
+    def __init__(
+        self,
+        text: str,
+        callback: Callable[[str], None] = None,
```
sam1penny, line 127: fyi, these type hints should include the `None` default value - in this case it should be:
`callback: Callable[[str], None] | None = None, `

roxgib: No problem, thanks for the review. I've added them.

Personally I think explicit Optionals are redundant, but I should have stayed consistent with the rest of the codebase.

qt/aqt/sync.py
```diff
@@ -135,13 +128,26 @@ def full_sync(
...
-            on_done()
+        button_labels: list[str] = [
+            tr.sync_upload_to_ankiweb(),
+            tr.sync_download_from_ankiweb(),
+            tr.sync_cancel_button(),
+        ]
+
+        def callback(choice: str) -> None:
+            if choice == button_labels[0]:
+                full_upload(mw, on_done)
```
dae, line 139: I regret returning the button text in the old API - the index of the chosen button would have been sufficient, and it would allow us to avoid this hack:

```
        # work around KDE 'helpfully' adding accelerators to button text of Qt apps
        return txt.replace("&", "")
```

What do you think about returning the index instead in the new routine? The default_button arg could become an index then too.

dae (commented):
Looks much nicer!

qt/aqt/utils.py
dae: Currently mypy doesn't complain if they're omitted because of the `no_strict_optional = true` in mypy.ini, but once that is changed in the future, omitting them would be an error.

qt/aqt/sync.py
roxgib: Agreed - this hack shouldn't be necessary with the new class since the return values are extracted from the originating argument values rather than the button text, but I can't think of a reason not to return the index except perhaps if the buttons are being dynamically generated, which seems unlikely.

roxgib: One thing that's annoying about using the index is that Yes returns 0, No returns 1, and Cancel returns the highest value, if the typical button order is used, but I suppose we'll have to live with that

dae: The calling code should be checking the index rather than truthiness, so that should be ok.

roxgib: I've left `ask_user` as using bools, which required compensating for this by not-ing the value, but I think in that function it would be confusing to return the index since it only asks a Yes/No question.

qt/aqt/utils.py
```diff
@@ -70,6 +120,107 @@ def openLink(link: str | QUrl) -> None:
...
+        if help is not None:
+            b = self.addButton(QMessageBox.StandardButton.Help)
+            qconnect(b.clicked, lambda: openHelp(help))
+        self.open()
+
+
+def ask_user(
+    text: str,
+    callback: Callable[[bool], None],
+    default_button: bool = True,
```
dae, line 165: How about `default_yes` or `default_no` to make it clearer?

qt/aqt/sync.py
dae: That seems reasonable - thanks!

qt/aqt/utils.py
roxgib: Done!

dae:
Re dark mode change, nice! This PR looks to be in good shape, so happy to merge when you're ready.

roxgib:
> Re dark mode change, nice!

I did look at integrating this with Anki's existing solution, but this was quicker... I figure it will do for now

> This PR looks to be in good shape, so happy to merge when you're ready.

Yep I think I've done enough testing now so I've taken it off draft

dae:
Thanks for your work on this Sam!

dae:
I've just noticed I overlooked one thing in review: it would be good if routines like show_info() take a parent argument, as displaying on top of whichever window happens to be open can be a bit unreliable.

roxgib:
It's possible to pass a parent argument through to MessageBox via `**kwargs` but happy to add it explicitly if you prefer. It's just that the argument list had just gotten a bit long. I've started working on a PR converting the next batch of callers, so I can include it with that.

dae:
IMHO it's best practice to always pass in the parent, so it probably is worth explicitly stating instead of relying on kwargs here. In a future PR would be great :-)

PR #2011, opened 2022-08-14
Use WithFloating for MathjaxOverlay

hgiesel:
This will make it easier for #2002


dae:
Is this still a draft or are you waiting for me to review it?

hgiesel:
Unfortunately there's still one major issue.

I use an implementation of a [portal](https://reactjs.org/docs/portals.html) to move the DOM that belongs to the dropdown / overlay from its deeply nested position to being a direct child of the `document.body` element. However I noticed that this prevents scrolling the content underneath the overlay, which makes sense. It wasn't that noticeable with the dropdowns in the editor toolbar, but it becomes a problem with the mathjax or image overlay, which take a lot of space.
The React documentation actually mentions, that portals are useful for:

> dialogs, hovercards, and tooltips.

I guess they didn't mention dropdowns here for a reason. Now, removing the portal causes some scrolling artifacts in return, which I need to have a look at.

dae:
Gotcha, thanks for the update.

hgiesel:
This closes #2048 

It also fixes some less severe issues:

- Changing the mathjax in the mathjax editor from 1 to zero characters would not update the preview
- You can now directly click from one mathjax to another.
- The mathjax overlay is much nicer now:

Bildschirmfoto 2022-09-05 um 00 26 25

- Removes all uses of WithDropdown and replaces them with WithFloating, which use floating-ui under the hood instead of bootstrap.

dae:
Seems to work well, and I trust in your decisions here. Thanks Henrik!

kleinerpirat:
@hgiesel I just got to test the new `Collapsible`. First of all, glad you got the tweened svelte animation working, looks nice!

There are two striking visual regressions though:

### No animation for expansion
This seems like a deliberate change. I'd prefer the animation going both ways - otherwise it looks like a bug prevents the expected animation from starting.

### Large margin between `LabelContainer` and editing inputs
image

At the same time, the margin between the toolbar and first field got smaller. IMO it looks worse now and makes the fields take up more space than necessary, as seen in this side-by-side comparison:

 

If you don't mind, I'll prepare a PR to solve abovementioned issues.

PR #2013, opened 2022-08-15
Allow passing debug=True to jest_test for debugger support

hgiesel:
Now, if you set `debug = True` in `jest.bzl`, you can use `debugger` in test files, and stop execution.

It is deactivated by default: I don't know what would be the best practice of passing flags to bazel commands, and I doubt that's a good idea anyway.

Having it in the repo at least can serve as documentation.


dae:
Nice!

PR #2014, opened 2022-08-16
Add option in math dropdown to toggle MathJax rendering

### LINKED ISSUE #1942: Provide opt-out of MathJax live preview
Until we can fix issues such as https://github.com/ankitects/anki/issues/1941 and things like [loading of macros](https://forums.ankiweb.net/t/can-live-preview-of-mathjax-be-tunred-off-in-the-card-editor/19414) and focus change issues, maybe we should provide an option to treat the MathJax code like we did previously.

dae:
Closes #1942

@hgiesel Does this seem like a reasonable approach to you as a quick fix? Is there an easy way to redraw the fields after toggling? Currently users need to switch to a different note and back again, or re-open the editor.

### DIFF
--- a/ftl/core/editing.ftl
+++ b/ftl/core/editing.ftl
@@ -58,6 +58,7 @@ editing-toggle-visual-editor = Toggle Visual Editor
 editing-underline-text = Underline text
 editing-unordered-list = Unordered list
 editing-warning-cloze-deletions-will-not-work = Warning, cloze deletions will not work until you switch the type at the top to Cloze.
+editing-toggle-mathjax-rendering = Toggle MathJax Rendering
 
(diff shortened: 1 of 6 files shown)


hgiesel (commented):
As a quickfix, this works fine for me.

In the long run, it would also be nice if disabling Mathjax would update the visual and html editor, so that `toUndecorated` is not even called, and also that the MathjaxOverlay is unmounted (and cleaned up).

Forcing a rerender of the content seems like a good use case. Typically, the svelte implementation of store prevents updates with identical content, but maybe we could extend the contract and add `content.forceSet(...)`.

dae:
I couldn't figure out how to do it elegantly, so I've just resorted to reloading the entire page for now. :-)

PR #2018, opened 2022-08-16
Fix field descriptions blocking cursor from being set

hgiesel:
Happens when focus is in HTML editor and you try to put caret into visual editor (I've also used inset by accident)


kleinerpirat:
This also fixes another bug I had noticed today: Not being able to scroll when the cursor hovers the EditingArea.

It helped me find the CSS cause of a long-standing issue: #2019.

PR #2019, opened 2022-08-17
Fix not being able to scroll when mouse hovers PlainTextInput

kleinerpirat:
#2018 inspired me to look for the cause of this long-standing issue.

Initial tests show no regressions.


dae:
This reintroduces a bug when scrolling on macOS - the "rubber band" scrolling is too slow. Does changing * to body help at all?

https://github.com/ankitects/anki/commit/396f583e2a971ce58d965566cdf008960d6bdc50

dae:
Thanks Matthias!

PR #2023, opened 2022-08-20
Keep content of unmapped fields when importing

RumovZ:
https://forums.ankiweb.net/t/importing-cannot-partially-update-existing-cards/22051

Also, tags are now merged when updating notes. Before, any old tags were discarded. I guess we'll want to make that customisable at some point.

### DIFF
--- a/rslib/src/tags/register.rs
+++ b/rslib/src/tags/register.rs
@@ -20,6 +20,7 @@ impl Collection {
     ) -> Result<(Vec<String>, bool)> {
         self.canonify_tags_inner(tags, usn, true)
     }
+
     pub(crate) fn canonify_tags_without_registering(
         &mut self,
         tags: Vec<String>,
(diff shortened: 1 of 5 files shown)


rslib/src/tags/register.rs
```diff
@@ -17,14 +17,37 @@ impl Collection {
         &mut self,
         tags: Vec<String>,
         usn: Usn,
+    ) -> Result<(Vec<String>, bool)> {
+        self.canonify_tags_inner(tags, usn, true)
+    }
+    pub(crate) fn canonify_tags_without_resgistering(
```
dae, line 23: minor typo

dae (commented):
Sorry this turned into a bigger project than originally anticipated Rumo. I noticed that the "Updated" log is showing the original field content instead of the updated text. Was that intentional?

dae:
> Also, tags are now merged when updating notes. Before, any old tags were discarded. I guess we'll want to make that customisable at some point.

It would be good if we could address this before the .55 release, as I suspect we'll break some people's workflows if we force the new behaviour.

dae:
Thanks for the quick fixes :-)

PR #2028, opened 2022-08-22
Make all Anki-native exceptions inherit from the same base class

glutanimate:
Allows add-ons to easily catch all custom Anki exceptions without having to import each exception individually, decoupling add-ons from the exact exception definitions in the catch-all use case.

----

### DIFF
--- a/pylib/anki/errors.py
+++ b/pylib/anki/errors.py
@@ -10,7 +10,24 @@ if TYPE_CHECKING:
     import anki.collection
 
 
-class LocalizedError(Exception):
+class AnkiException(Exception):
+    """
+    General Anki exception that all custom exceptions raised by Anki should
+    inherit from. Allows add-ons to easily identify Anki-native exceptions.
+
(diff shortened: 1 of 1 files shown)


glutanimate:
A bit puzzled why `pylint` didn't trigger locally with `PYTEST=test_bury bazel run //pylib:pytest`, but here we go :man_shrugging:

dae:
- 'bazel test //...' will run all tests, including pylint
- 'PYTEST=test_bury bazel run //pylib:pytest' will only run pytest, and only run the test_bury test

glutanimate:
> * 'bazel test //...' will run all tests, including pylint
> * 'PYTEST=test_bury bazel run //pylib:pytest' will only run pytest, and only run the test_bury test

 :man_facepalming: double-whammy

I swear I `Ctrl+R`'d for pylib and not pylint :sweat_smile: - turns out that blindly running the first command that comes up in your bash history is not a good idea. I'll put it down to being hungry and in a hurry that noon :P

PR #2029, opened 2022-08-22
Move VS Code settings to .vscode.dist and update docs

### LINKED ISSUE #1981: [Suggestion] Can we move the shared VS Code settings out of the project root?
I think it's great that the project now ships with a series of handy presets for VS Code! However, since Code [doesn't support local workspace settings](https://github.com/microsoft/vscode/issues/40233), the fact that `.vscode` is part of version tracking does complicate things a bit in some scenarios. For instance, with my venv set-up, I have to add `"./qt"` to `python.analysis.extraPaths` in order to get suggestions from the local source tree rather than the venv-installed `aqt` build. But even more generally speaking, users might want to / need to use the workspace settings to set all kinds of project-specific settings that are particular to their preferences (ranging from custom themes to extension settings).

With that in mind, I would suggest the following change: Let's move `.vscode/settings.json` out of the project root and into a subfolder (e.g. `IDE`), adding the root-level `. vscode/settings.json` to `.gitignore`. Users who want to follow along with the shared workspace settings could then symlink `IDE/.vscode/settings.json` into the project root `.vscode` folder, while users who need to customize their workspace settings could simply copy the defaults over, and then adjust them as needed (and update them manually when upstream pushes changes).

If you think this is reasonable, I'd be happy to file a PR with the changes above and also update the [IDE docs](https://github.com/ankitects/anki/blob/main/docs/editing.md).

glutanimate:
Closes #1981, following @hgiesel's suggestion

(sorry, meant to file this much earlier, but lost track)

The only caveat I see with this change is that existing contributors will have to run `cp -r .vscode.dist .vscode` one time once they pull the changes.


dae:
Thanks Aristotelis!

RumovZ:
> The only caveat I see with this change is that existing contributors will have to run cp -r .vscode.dist .vscode one time once they pull the changes.

Isn't that rather `cp .vscode.dist/* .vscode`?

glutanimate:
> Isn't that rather `cp .vscode.dist/* .vscode`?

Maybe I'm missing something, but assuming that the pull deletes `.vscode`, both commands would functionally do the same, no? The local copy should be deleted by the pull AFAIK, despite the added `.gitignore` entry. But didn't give this a spin, so would be curious to hear if it works differently in practice.

dae:
That command worked for me after a git pull, but maybe windows is different?

RumovZ:
Ah, okay. My `.vscode` apparently didn't get deleted due to some `.ropeproject` folder inside created by a Python extension.

glutanimate:
Ah, that makes sense, yeah. In that case `cp .vscode.dist/* .vscode` might be the better command to recommend as it covers both cases, whether `.vscode` was deleted or not.

glutanimate:
Ah, that would explain it, yeah. I guess the same goes if people have set up custom `launch.json`s and similar and excluded them locally.

PR #2030, opened 2022-08-24
Use extern error crate

RumovZ:
Here is a showcase of what an error crate like snafu could bring to the table. Curious to hear what you think. 

## Why snafu
1. anyhow / eyre, thiserror and error_stack seem mostly concerned with an error's string representation, which is not as useful for Anki. snafu seems the most convenient for assembling typed errors as well.
2. Backtraces on the stable toolchain.
3. Support for Options.

## Caveats
1. Tuple structs or enums with tuple variants, like `AnkiError`, cannot derive `Snafu`. To set up a quick demonstration, I had to use nested structs like `AnkiError::InvalidInputError(InvalidInputError)`. This should later be inlined to `AnkiError::InvalidInputError { message, backtrace }`.
2. `Backtrace` doesn't implement `PartialEq`, so I've implemented it manually for the error structs. But equality checks with `AnkiError`s are probably completely expendable.

## Conclusions
I think we can roughly differentiate between two kinds of errors.

#### 1. Unexpected Errors
`NotFound` is the prime example. As we don't expect this error to happen, we don't want to spend a lot of effort attaching context. With snafu we can simply replace `.ok_or(AnkiError::NotFound)` with `.context(NotFoundSnafu)` to at least have a backtrace.
`InvalidInput` is another one. We can make this the `whatever` variant to create string errors very concisely like in `ensure_whatever!(!self.name.is_empty(), "Empty note type name");`.
(Side note: `InvalidInput` should probably not inherit `LocalizedException` and be shown with `show_exception()`.)

#### 2. Localized Errors
This kind of errors needs some manual context handling. Since currently most fallible functions in the codebase return an `AnkiError` as the error variant, all context needs to be available in the lowest failing function. This often counteracts proper separation of concerns. I think more modules should define their own errors which then get contextualised as they pass through the call stack. Snafu makes this quite painless as in `ensure_valid_parsed_templates()`.
They still may happen under unexpected circumstances and we can add a backtrace to help with debugging here as well, but snafu admonishes not to generate backtraces in a tight loop.

### DIFF
--- a/rslib/src/error/mod.rs
+++ b/rslib/src/error/mod.rs
@@ -6,7 +6,7 @@ mod filtered;
 mod network;
 mod search;
 
-use std::{fmt::Display, io, path::Path};
+use std::{fmt::Display, io, path::PathBuf};
 
 pub use db::{DbError, DbErrorKind};
 pub use filtered::{CustomStudyError, FilteredDeckError};
@@ -123,9 +123,7 @@ impl AnkiError {
             | AnkiError::NotFoundError(_)
             | AnkiError::Existing
             | AnkiError::UndoEmpty => format!("{:?}", self),
-            AnkiError::FileIoError(err) => {
-                format!("{}: {}", err.path, err.error)
-            }
+            AnkiError::FileIoError(err) => err.message(),
         }
     }
 
@@ -280,24 +278,44 @@ impl ImportError {
     }
 }
 
-#[derive(Debug, PartialEq, Clone)]
-
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub(crate)))]
 pub struct FileIoError {
-    pub path: String,
-    pub error: String,
+    pub path: PathBuf,
+    pub op: FileOp,
+    pub source: std::io::Error,
 }
 
-impl AnkiError {
-    pub(crate) fn file_io_error<P: AsRef<Path>>(err: std::io::Error, path: P) -> Self {
-        AnkiError::FileIoError(FileIoError::new(err, path.as_ref()))
+#[derive(Debug, PartialEq, Clone)]
+pub enum FileOp {
+    Open,
+    Create,
+    Write,
+    CopyFrom(String),
+}
+
+impl FileOp {
+    pub fn copy(from: &str) -> Self {
+        Self::CopyFrom(from.to_owned())
+    }
+
+    fn verb(&self) -> String {
+        match self {
+            Self::Open => "read".into(),
+            Self::Create => "create file in".into(),
+            Self::Write => "write".into(),
+            Self::CopyFrom(s) => format!("copy from {s} to"),
+        }
     }
 }
 
 impl FileIoError {
-    pub fn new(err: std::io::Error, path: &Path) -> FileIoError {
-        FileIoError {
-            path: path.to_string_lossy().to_string(),
-            error: err.to_string(),
-        }
+    fn message(&self) -> String {
+        format!(
+            "failed to {} {}: {:?}",
+            self.op.verb(),
+            self.path.to_string_lossy(),
+            self.source
+        )
     }
 }
--- a/rslib/src/error/mod.rs
+++ b/rslib/src/error/mod.rs
@@ -319,3 +319,9 @@ impl FileIoError {
         )
     }
 }
+
+impl From<FileIoError> for AnkiError {
+    fn from(err: FileIoError) -> Self {
+        Self::FileIoError(err)
+    }
+}
--- a/rslib/src/error/mod.rs
+++ b/rslib/src/error/mod.rs
@@ -19,7 +19,7 @@ use crate::{i18n::I18n, links::HelpPage};
 
 pub type Result<T, E = AnkiError> = std::result::Result<T, E>;
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug)]
 pub enum AnkiError {
     InvalidInput(String),
     InvalidInputError(InvalidInputError),
--- a/rslib/src/error/mod.rs
+++ b/rslib/src/error/mod.rs
@@ -12,7 +12,7 @@ pub use db::{DbError, DbErrorKind};
 pub use filtered::{CustomStudyError, FilteredDeckError};
 pub use network::{NetworkError, NetworkErrorKind, SyncError, SyncErrorKind};
 pub use search::{ParseError, SearchErrorKind};
-use snafu::Snafu;
+use snafu::{Backtrace, ErrorCompat, Snafu};
 use tempfile::PathPersistError;
 
 use crate::{i18n::I18n, links::HelpPage};
@@ -22,6 +22,7 @@ pub type Result<T, E = AnkiError> = std::result::Result<T, E>;
 #[derive(Debug, PartialEq)]
 pub enum AnkiError {
     InvalidInput(String),
+    InvalidInputError(InvalidInputError),
     TemplateError(String),
     CardTypeError(CardTypeError),
     IoError(String),
@@ -36,6 +37,7 @@ pub enum AnkiError {
     CollectionNotOpen,
     CollectionAlreadyOpen,
     NotFound,
+    NotFoundError(NotFoundError),
     /// Indicates an absent card or note, but (unlike [AnkiError::NotFound]) in
     /// a non-critical context like the browser table, where deleted ids are
     /// deliberately not removed.
@@ -93,7 +95,8 @@ impl AnkiError {
             }
             AnkiError::DbError(err) => err.localized_description(tr),
             AnkiError::SearchError(kind) => kind.localized_description(tr),
-            AnkiError::InvalidInput(info) => {
+            AnkiError::InvalidInput(info)
+            | AnkiError::InvalidInputError(InvalidInputError { message: info, .. }) => {
                 if info.is_empty() {
                     tr.errors_invalid_input_empty().into()
                 } else {
@@ -117,6 +120,7 @@ impl AnkiError {
             | AnkiError::CollectionNotOpen
             | AnkiError::CollectionAlreadyOpen
             | AnkiError::NotFound
+            | AnkiError::NotFoundError(_)
             | AnkiError::Existing
             | AnkiError::UndoEmpty => format!("{:?}", self),
             AnkiError::FileIoError(err) => {
@@ -138,6 +142,45 @@ impl AnkiError {
             }),
             _ => None,
         }
+    }
+
+    pub fn backtrace(&self) -> String {
+        if let AnkiError::InvalidInputError(ref err) = self {
+            if let Some(bt) = ErrorCompat::backtrace(err) {
+                return bt.to_string();
+            }
+        }
+        if let AnkiError::NotFoundError(ref err) = self {
+            if let Some(bt) = ErrorCompat::backtrace(err) {
+                return bt.to_string();
+            }
+        }
+        String::new()
+    }
+}
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub), whatever)]
+pub struct InvalidInputError {
+    pub message: String,
+    pub backtrace: Backtrace,
+}
+
+impl PartialEq for InvalidInputError {
+    fn eq(&self, other: &Self) -> bool {
+        self.message == other.message
+    }
+}
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub))]
+pub struct NotFoundError {
+    pub backtrace: Backtrace,
+}
+
+impl PartialEq for NotFoundError {
+    fn eq(&self, _other: &Self) -> bool {
+        true
     }
 }
 
--- a/rslib/src/error/mod.rs
+++ b/rslib/src/error/mod.rs
@@ -12,6 +12,7 @@ pub use db::{DbError, DbErrorKind};
 pub use filtered::{CustomStudyError, FilteredDeckError};
 pub use network::{NetworkError, NetworkErrorKind, SyncError, SyncErrorKind};
 pub use search::{ParseError, SearchErrorKind};
+use snafu::Snafu;
 use tempfile::PathPersistError;
 
 use crate::{i18n::I18n, links::HelpPage};
@@ -77,12 +78,14 @@ impl AnkiError {
             AnkiError::CardTypeError(err) => {
                 let header =
                     tr.card_templates_invalid_template_number(err.ordinal + 1, &err.notetype);
-                let details = match err.details {
+                let details = match err.source {
                     CardTypeErrorDetails::TemplateError | CardTypeErrorDetails::NoSuchField => {
                         tr.card_templates_see_preview()
                     }
                     CardTypeErrorDetails::NoFrontField => tr.card_templates_no_front_field(),
-                    CardTypeErrorDetails::Duplicate(i) => tr.card_templates_identical_front(i + 1),
+                    CardTypeErrorDetails::Duplicate { index } => {
+                        tr.card_templates_identical_front(index + 1)
+                    }
                     CardTypeErrorDetails::MissingCloze => tr.card_templates_missing_cloze(),
                     CardTypeErrorDetails::ExtraneousCloze => tr.card_templates_extraneous_cloze(),
                 };
@@ -124,11 +127,11 @@ impl AnkiError {
 
     pub fn help_page(&self) -> Option<HelpPage> {
         match self {
-            Self::CardTypeError(CardTypeError { details, .. }) => Some(match details {
+            Self::CardTypeError(CardTypeError { source, .. }) => Some(match source {
                 CardTypeErrorDetails::TemplateError | CardTypeErrorDetails::NoSuchField => {
                     HelpPage::CardTypeTemplateError
                 }
-                CardTypeErrorDetails::Duplicate(_) => HelpPage::CardTypeDuplicate,
+                CardTypeErrorDetails::Duplicate { .. } => HelpPage::CardTypeDuplicate,
                 CardTypeErrorDetails::NoFrontField => HelpPage::CardTypeNoFrontField,
                 CardTypeErrorDetails::MissingCloze => HelpPage::CardTypeMissingCloze,
                 CardTypeErrorDetails::ExtraneousCloze => HelpPage::CardTypeExtraneousCloze,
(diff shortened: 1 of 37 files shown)


dae:
I've read through the snafu docs to better understand this, and it does look promising.

> Tuple structs or enums with tuple variants, like AnkiError, cannot derive Snafu. To set up a quick demonstration, I had to use nested structs like AnkiError::InvalidInputError(InvalidInputError). This should later be inlined to AnkiError::InvalidInputError { message, backtrace }

Inlining the inner error into a struct variant would mean that routines couldn't deal with the inner error separately - they'd have to return an AnkiError instead of being able to return an InvalidInputError. This seems somewhat counter to the desire to improve separation of concerns. InvalidInput is used throughout the codebase so it's not a big deal in this case, but what about more specific errors such as card template ones? Maybe we could handle that with an outer struct variant like CardTemplateError { source: CardTemplateError }?

> Backtrace doesn't implement PartialEq, so I've implemented it manually for the error structs. 

I'm not too bothered about losing automatic equality, but at same time, I wonder whether we'd really need to? 

- I suspect the variants where a backtrace actually adds value are fairly limited, so we'd presumably only need to manually implement them for those variants
- If we follow best practice and attach enough context to ambiguous errors, that will reduce the need for backtraces further

> NotFound is the prime example. As we don't expect this error to happen

Having seen a lot of broken collections over the years, I do sort of expect it to happen :-) 

Some ideas for how we could improve it:

- Instead of showing NotFound to the user, tell them that their collection is in an inconsistent state, and recommend they run Check DB
- Add a message field to NotFoundError as well, and use it with with_context, eg:

```diff
diff --git a/rslib/src/error/mod.rs b/rslib/src/error/mod.rs
index 3bad2c8d3..444a96c21 100644
--- a/rslib/src/error/mod.rs
+++ b/rslib/src/error/mod.rs
@@ -12,7 +12,7 @@ pub use db::{DbError, DbErrorKind};
 pub use filtered::{CustomStudyError, FilteredDeckError};
 pub use network::{NetworkError, NetworkErrorKind, SyncError, SyncErrorKind};
 pub use search::{ParseError, SearchErrorKind};
-use snafu::{Backtrace, ErrorCompat, Snafu};
+use snafu::{Backtrace, Error, ErrorCompat, IntoError, NoneError, OptionExt, ResultExt, Snafu};
 use tempfile::PathPersistError;
 
 use crate::{i18n::I18n, links::HelpPage};
@@ -55,6 +55,75 @@ pub enum AnkiError {
     InvalidId,
 }
 
+#[derive(Debug, Snafu)]
+pub enum FutureAnkiError {
+    // avoids conflicting name, and allows use of '?' to convert from NotFoundSnafu to FutureAnkiError
+    #[snafu(context(false))]
+    NotFound { source: NotFoundError },
+    #[snafu(context(false))]
+    InvalidInput { source: InvalidInputError },
+    #[snafu(context(false))]
+    LegacyError { source: AnkiError },
+}
+
+
+pub trait OptionExt2<T>: Sized {
+    fn ensure_some<F, C>(self, context: F) -> Result<T, NotFoundError>
+    where
+        F: FnOnce() -> C,
+        C: Into<String>;
+}
+
+impl<T> OptionExt2<T> for Option<T> {
+    #[cfg_attr(feature = "rust_1_46", track_caller)]
+    fn ensure_some<F, C>(self, context: F) -> Result<T, NotFoundError>
+    where
+        F: FnOnce() -> C,
+        C: Into<String>,
+    {
+        match self {
+            Some(v) => Ok(v),
+            None => Err(NotFoundSnafu { message: context() }.into_error(NoneError)),
+        }
+    }
+}
+
+macro_rules! ensure_some {
+    ($option:expr, $info:expr) => {
+        match $option {
+            Some(v) => v,
+            None => return Err(NotFoundSnafu { message: format!($info) }.into_error(NoneError).into())
+        }
+    }
+}
+
+
+fn example(col: &mut crate::collection::Collection) -> Result<(), FutureAnkiError> {
+    let name = "no such name";
+    col.get_notetype_by_name(name)?
+        .with_context(|| NotFoundSnafu {
+            message: format!("no such notetype: {name}"),
+        })?;
+    Ok(())
+}
+
+fn example2_with_helper_impl(
+    col: &mut crate::collection::Collection,
+) -> Result<(), FutureAnkiError> {
+    let name = "no such name";
+    col.get_notetype_by_name(name)?
+        .ensure_some(|| format!("no such notetype: {name}"))?;
+    Ok(())
+}
+
+fn example3_with_macro_impl(
+    col: &mut crate::collection::Collection,
+) -> Result<(), FutureAnkiError> {
+    let name = "no such name";
+    ensure_some!(col.get_notetype_by_name(name)?, "no such notetype: {name}");
+    Ok(())
+}
+
 impl std::error::Error for AnkiError {}
 
 impl Display for AnkiError {
@@ -175,6 +244,7 @@ impl PartialEq for InvalidInputError {
 #[derive(Debug, Snafu)]
 #[snafu(visibility(pub))]
 pub struct NotFoundError {
+    pub message: String,
     pub backtrace: Backtrace,
 }
 ```

> (Side note: InvalidInput should probably not inherit LocalizedException and be shown with show_exception().)

What do you propose instead? While it's supposed to be for programming errors, we may accidentally make mistakes, and add-ons may be triggering this error too.

>  I think more modules should define their own errors which then get contextualised as they pass through the call stack. Snafu makes this quite painless as in ensure_valid_parsed_templates()

Agreed, it does appear to make things easier.

> They still may happen under unexpected circumstances and we can add a backtrace to help with debugging here as well, but snafu admonishes not to generate backtraces in a tight loop.

I'd rather add them only when required (generic errors, and when a need becomes apparent in the future), as they're sort of a kludge around best practices, and they come with a performance cost.

I'd recommend we use `Option<Backtrace>` - snafu will then only populate it when RUST_BACKTRACE is set. The difference in performance in a non-release build when an error occurs is quite noticeable.

Note that our standard ./run and release builds will not get line numbers, which somewhat reduces the useful of the backtraces. We can get them in testing with the following:

RUST_BACKTRACE=1 BUILDARGS='-c dbg' ./run

File: rslib/src/error/mod.rs [hunk 1]
```diff
@@ -12,15 +12,17 @@ pub use db::{DbError, DbErrorKind};
...
 pub use network::{NetworkError, NetworkErrorKind, SyncError, SyncErrorKind};
 pub use search::{ParseError, SearchErrorKind};
+use snafu::{Backtrace, ErrorCompat, Snafu};
 use tempfile::PathPersistError;
 
 use crate::{i18n::I18n, links::HelpPage};
 
 pub type Result<T, E = AnkiError> = std::result::Result<T, E>;
 
-#[derive(Debug, PartialEq)]
```
dae, line 21: After you added the manual PartialEq implementations, this can be derived again, so presumably those equality changes are no longer required.

rslib/src/notetype/mod.rs
```diff
@@ -504,14 +496,33 @@ impl Notetype {
...
     }
 
+    fn ensure_unempty_values(&mut self) -> Result<(), InvalidInputError> {
+        let bad_chars = |c| c == '"';
+        if self.name.contains(bad_chars) {
+            self.name = self.name.replace(bad_chars, "");
+        }
+        ensure_whatever!(!self.name.is_empty(), "Empty note type name");
+        ensure_whatever!(!self.fields.is_empty(), "1 field required");
+        ensure_whatever!(!self.templates.is_empty(), "1 template required");
```
dae, line 512: `whatever` seems to not work well with the coercion into a containing error, so if this routine had to handle some other type of error as well, I don't think we could use this. The following code fails for example:

```rust
fn map_example() -> Result<(), FutureAnkiError> {
    // legacy style
    let ctype = CardType::try_from(7)
        .map_err(|_| AnkiError::invalid_input("invalid card type"))?;
    // whatever
    let ctype = whatever!(CardType::try_from(7), "invalid card type");
Ok(())
}
```

Maybe we'd be better off with a helper routine and avoiding `whatever`?

RumovZ:
> Maybe we could handle that with an outer struct variant like CardTemplateError { source: CardTemplateError }?

Sure, I just meant to explain why I've implemented it like this for now.

> I'm not too bothered about losing automatic equality, but at same time, I wonder whether we'd really need to?

I guess either works.

> Instead of showing NotFound to the user, tell them that their collection is in an inconsistent state, and recommend they run Check DB
   Add a message field to NotFoundError as well, and use it with with_context, eg:

Sounds good! I think we don't even have to provide the format strings if we implement a trait like:
```rust
trait AnkiObject { fn name() -> &'static str; }
impl AnkiObject for Note { fn name() -> &'static str { "note" } }
```

> What do you propose instead? While it's supposed to be for programming errors, we may accidentally make mistakes, and add-ons may be triggering this error too.

`LocalizedError` is a misleading name. Like `.localized_description()` which does not in all cases return a localized description. They could just be called `DescribedError` and `.description()`. If the frontend wants to distinguish between localized and unlocalized descriptions, there could be a `class LocalizedError(DescribedError)`.
Then I would only accept `DescribedError`s in `show_exception()`. The rest, like InvalidInput, would be picked up by the ErrorHandler and displayed with the usual debugging information suitable for a forum report.

> Note that our standard ./run and release builds will not get line numbers, which somewhat reduces the useful of the backtraces. We can get them in testing with the following:

Ah, I didn't know that. I found the name of the function the error originated in to be the most valuable information, though.
Makes sense to use an `Option<Backtrace>`.

rslib/src/error/mod.rs
RumovZ: Yeah, I removed the equality checks afterwards to see if it was practical.

rslib/src/notetype/mod.rs
RumovZ: Not sure I understand your example. Wouldn't InvalidInput always be a top-level error anyway?

dae: By that, I guess you mean defining it as a variant of FutureAnkiError directly? What if we want more than one variant with string errors there?

RumovZ: Yes. I can't think of another use case for string errors apart from unexpectedly invalid input, but should one arise in the future, we could still add a helper routine for that and keep using `whatever` for invalid input, couldn't we?
This seems to require the least amount of work for now and in the long run.

dae:
> The rest, like InvalidInput, would be picked up by the ErrorHandler and displayed with the usual debugging information suitable for a forum report.

Ok, I see what you mean. I'm not sure whether it would be best to separate the errors on the frontend though:

- We implement the backend once, but the frontend needs 3-4 different implementations for the different clients, so wherever possible it's best to push the complexity into the shared code, and keep the platform-specific UI code simple
- For performance reasons we'll want to run without RUST_BACKTRACE=1 by default, which means no backtrace will be available in the common case - it will just be an option of last resort if we get a report where it's not clear what's failing
- An error that we haven't attempted to localize is no different to the user than an error we have made localizable but which hasn't been translated into the user's language yet. Maybe we could 'solve' this by renaming `localized` to something more generic like `message` instead? :-)

rslib/src/notetype/mod.rs
dae: I can't think of another of our existing variants that would benefit from string errors at the moment. Presumably it would also rule out the usage of .(with_)whatever_context()? Not necessarily a bad thing, as I guess that would complicate later determining the class of error if I've understood things correctly (we'd need to hunt for it in the source error?)

The examples above focused on throwing an error based on a condition - perhaps it would be worth also experimenting a bit with how we'd deal with existing errors that need to be mapped to a different error with context added, since that's also common?

RumovZ:
I'm not proposing to introduce any new separation. That's just the (inevitable) status quo: Some errors get caught because we anticipated them, others are only picked up by the error handler because we didn't – and some errors get caught because we anticipated a differented one and were too generous with our `except` clause. Now my assumptions are:

1. We don't want the last kind of errors, because we'd rather show debugging information for surprising errors. (The additional Rust backtrace would only be the cherry on the cake.)
2. InvalidInput and NotFound are never anticipated, because they only happen if a dev, an add-on author or a superuser made a mistake.

That's why I've suggested that InvalidInput should not be a LocalizedError: Then we could write `except LocalizedError` without catching it inadvertently (NotFound doesn't inherit LocalizedError anyway).
Of course, ideally we would be even more specific with our except clause.

<hr>

I had assumed I had more or less grasped the current error handling strategy, but apparently that's not so. Could you explain what errors you want to catch and what the purpose of LocalizedError is?

rslib/src/notetype/mod.rs
RumovZ: > The examples above focused on throwing an error based on a condition - perhaps it would be worth also experimenting a bit with how we'd deal with existing errors that need to be mapped to a different error with context added, since that's also common?

I don't know what you mean exactly. Is this still about InvalidInput? Do you have an example?

dae: Sorry, not about InvalidInput, just about using snafu in general. For example, the way we include a filename in data_for_file() and atomic_rename(), and things like 

```rust
        let decks = self
            .get_schema11_decks()
            .map_err(|e| AnkiError::JsonError(format!("decoding decks: {}", e)))?;
```

dae:
I see it a little differently - I consider any error raised from the Rust code as "anticipated", as we've explicitly marked it with a ? or dealt with it in a match expression. Some errors will be more likely than others, but it's clear from the code that they can occur. 

Contrast the Python code, where it's not obvious whether a particular line will throw or not - in many cases you either need to rely on doc strings, or reading the source of the function you're calling to know whether it will fail or not. It's error prone, and the legacy code didn't even try to handle all the possible errors, instead relying on a global error handler to catch them. Because the errors were not dealt with in the place they occurred, there's often no useful message that can be printed, so we had to dump the stack trace to figure out what was going on. It's a crux, and not great to be dumping a wall of text at the user.

> InvalidInput and NotFound are never anticipated, because they only happen if a dev, an add-on author or a superuser made a mistake.

As discussed above, we could fairly easily turn NotFound into a localized error that points out the id/kind that was missing, and prompts the user run Check Database, so I guess that just leaves InvalidInput?

Even for InvalidInput, I'm not sure a Python stack trace/things like platform type would be particularly useful in the majority of cases (I'm assuming that's what you meant by debugging info) - in most cases of a backend error, the Python part of the stack does not seem to be interesting.

I guess LocalizedError should be considered synonymous with BackendError - in my ideal world, any error returned from the backend could be displayed to the user as-is (translated or not), which keeps the complexity down on the UI side of things. For the backend string errors that don't currently provide enough context (eg "NotFound"), ideally we'll fix them so they do, and the ability to add context seems to be one of the appealing points of snafu. For exceptions raised by Python code, presumably we'll want to continue using the existing traceback-based message.

Does that make sense?

dae:
Works for me. I was going to suggest we leave LocalizedError around as an alias, but from yesterday's snapshot it doesn't look like any add-ons are using it yet.

rslib/src/notetype/mod.rs
RumovZ: > The examples above focused on throwing an error based on a condition - perhaps it would be worth also experimenting a bit with how we'd deal with existing errors that need to be mapped to a different error with context added, since that's also common?

I've added `FileIoError` now.

rslib/src/import_export/package/apkg/import/mod.rs
```diff
@@ -113,11 +115,15 @@ fn collection_to_tempfile(meta: &Meta, archive: &mut ZipArchive<File>) -> Result
...
+        copy_decode(&mut zip_file, &mut tempfile)
     } else {
         io::copy(&mut zip_file, &mut tempfile).map(|_| ())
     }
-    .map_err(|err| AnkiError::file_io_error(err, tempfile.path()))?;
+    .with_context(|_| FileIoSnafu {
+        path: tempfile.path(),
+        op: FileOp::copy(zip_file.name()),
+    })
+    .map_err(AnkiError::FileIoError)?;
```
dae, line 126: This feels a little boilerplatey - it would be nice if we could just use with_context(...)? instead. I think we could do that if we make the top-level error derive Snafu and switch to struct enums? https://github.com/ankitects/anki/pull/2030#issuecomment-1226779795.

RumovZ: Yes, that's what I meant in caveat 1. However, maybe that original plan isn't such a good idea after all.
You made an argument against inlining error structs in the case of more specific errors, like CardTemplateError. I've realised that it might also be a problem in universal low-level errors like FileIoError: Let's say we have a module that returns its own error, but also works with input, so it must handle io errors. If AnkiError::FileIoError was an inline struct, we couldn't reuse it in the module error enum.
e0fb628d4d2f616a53d947e271fd5359aded890c gets rid of most of the boilerplate.

dae: Just to confirm we're on the same page - in the `FutureAnkiError` I proposed above, I was not proposing that the error context be directly inlined into the enum variant. By only using a single `source` field for each struct variant, it's effectively the same as the existing tuple struct approach, and each of the embedded errors can be used separately.  And the #[snafu(...)] line gives us the from Impl with 3 less lines. WDYT?

RumovZ: Sorry, I lost track a bit during the hiatus. Of course, your patch above will get rid of the boilerplate anyway and e0fb628d4d2f616a53d947e271fd5359aded890c is superfluous.
I mean this PR only to be a demo and playground. Once we are decided on the basic points (and I guess we are about now?) I will replay the relevant parts, including your patch, on a clean branch. My bad for not clarifying this sooner!

dae: No problem, sounds like we agree on the basic points :-)

RumovZ:
> You said before that you don't want to dump the backtrace on the console. However, if the backtrace requires env variable, it isn't very accessible to users anyway. Do you still think it should be displayed in the GUI?

Still wondering about that.
Let's take NotFoundError for example. There are three pieces of information:

1. Localized user message: "Inconsistent state ... check database ..."
2. Unlocalized details: "No such deck: 'Some Name'"
3. Backtrace

I understand you don't want them passed to the frontend separately, as that would require each frontend to handle it individually. However, I also don't think we should show the user unlocalized, technical information without need. Here are some other ideas:

1. Only attach details if some env var is set. (True for backtrace anyway.)
2. Use HTML to hide details. Would require a webview, as Qt's rich text doesn't support `<details>` or similar.
3. Never send details to the frontend, but log it or dump it on the console instead.

WDYT?

dae:
My concern was mainly that it can be difficult/impossible to see the console output in the mobile clients, so the traceback would need to be passed back to the frontend for appropriate handling. Since the tracebacks will tend to be pretty unwieldy due to the number of Python/pyo3 stack frames, perhaps the best option would be to include the backtrace in a separate protobuf field, so the desktop client can dump it to the console instead of adding it to the dialog box, which would end up as a wall of text. When RUST_BACKTRACE is not set, the backtrace field would be empty, so nothing would be dumped to the console. How does that sound?

On a related note, apparently backtrace support has been stabilized, and will be in 1.64 without needing an extra crate.

RumovZ:
Sounds good, but what about the second level of information?
With "No such deck: 'Some Name'" I have given a concrete example for NotFoundError, but this is also where further context information from other stack frames will go if we collect some in the future. That is to say, it might get a lot more verbose.
Also, this is the info the user will see if no localized message is available, like for InvalidInput.

dae:
Sorry, I was too focused on the backtrace portion :-). Maybe we could start with dumping the context to the console as well (via protobuf+the frontend), and then revisit it in the future when we have a better idea of the sort of messages and length that will tend to be output? In the future, instead of `<details>`, another option would be to have the error dialog have a `debug info` button.

PR #2031, opened 2022-08-25
add shortcut to close addcards window

### LINKED ISSUE #2026: Cmd+w to close add window no longer works
Caused by making it a main window; we'd need to add an explicit shortcut or a menu item to fix it.

https://forums.ankiweb.net/t/problem-closing-add-window-in-newest-mac-os-version/22423

sam1penny:
Fixes #2026 

(This will show a confirmation window to the user if the window has text in the field, since closeEvent() is overridden in AddCards)


dae:
Great, thanks for looking into this Sam!

PR #2032, opened 2022-08-25
More template checks

### LINKED ISSUE #2024: More template checks
Based on recurring forum questions, here are two more ideas for template checks:

1. Multiple `type` filters
I believe there are add-ons supporting this. Is there a way to add a check for this without disrupting add-on users? Maybe we could mark the error as uncritical, so only a warning is shown, but the user may still save the template.
2. [Media references to fields](https://docs.ankiweb.net/templates/fields.html?highlight=img#field-references)
Is there any valid use case for this?

RumovZ:
Closes #2024.
Is this like what you had in mind? Feel free to tweak the wording to your liking.
I wasn't sure what media to check for. Currently, only img, object and audio tags are found. I guess we'll also want to check for Anki sounds and latex? What about script tags with src attributes?

### DIFF
--- a/ftl/core/card-templates.ftl
+++ b/ftl/core/card-templates.ftl
@@ -56,3 +56,4 @@ card-templates-this-will-create-card-proceed =
         [one] This will create { $count } card. Proceed?
        *[other] This will create { $count } cards. Proceed?
     }
+card-templates-type-boxes-warning = Only one type box per card template is supported!
--- a/ftl/core/card-templates.ftl
+++ b/ftl/core/card-templates.ftl
@@ -56,4 +56,4 @@ card-templates-this-will-create-card-proceed =
         [one] This will create { $count } card. Proceed?
        *[other] This will create { $count } cards. Proceed?
     }
-card-templates-type-boxes-warning = Only one type box per card template is supported!
+card-templates-type-boxes-warning = Only one typing box per card template is supported.
--- a/ftl/core/media-check.ftl
+++ b/ftl/core/media-check.ftl
@@ -22,7 +22,9 @@ media-check-oversize-header = Files over 100MB can not be synced with AnkiWeb.
 media-check-subfolder-header = Folders inside the media folder are not supported.
 media-check-missing-header = The following files are referenced by cards, but were not found in the media folder:
 media-check-unused-header = The following files were found in the media folder, but do not appear to be used on any cards:
-media-check-template-refs-header = Files referenced by templates directly will not be considered in use, unless their names start with an underscore. Referencing templates:
+media-check-template-references-field-header = Anki can not detect used files when you use { "{{Field}}" } references in media/LaTeX tags. The media/LaTeX tags should be placed on individual notes instead.
+
+Referencing templates:
 
 ## Shown once for each file
 
--- a/ftl/core/media-check.ftl
+++ b/ftl/core/media-check.ftl
@@ -22,6 +22,7 @@ media-check-oversize-header = Files over 100MB can not be synced with AnkiWeb.
 media-check-subfolder-header = Folders inside the media folder are not supported.
 media-check-missing-header = The following files are referenced by cards, but were not found in the media folder:
 media-check-unused-header = The following files were found in the media folder, but do not appear to be used on any cards:
+media-check-template-refs-header = Files referenced by templates directly will not be considered in use, unless their names start with an underscore. Referencing templates:
 
 ## Shown once for each file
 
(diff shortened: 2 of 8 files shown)


ftl/core/card-templates.ftl
```diff
@@ -56,3 +56,4 @@ card-templates-this-will-create-card-proceed =
         [one] This will create { $count } card. Proceed?
        *[other] This will create { $count } cards. Proceed?
     }
+card-templates-type-boxes-warning = Only one type box per card template is supported!
```
dae, line 59:
```suggestion
card-templates-type-boxes-warning = Only one typing box per card template is supported.
```

ftl/core/media-check.ftl
```diff
@@ -22,6 +22,7 @@ media-check-oversize-header = Files over 100MB can not be synced with AnkiWeb.
 media-check-subfolder-header = Folders inside the media folder are not supported.
 media-check-missing-header = The following files are referenced by cards, but were not found in the media folder:
 media-check-unused-header = The following files were found in the media folder, but do not appear to be used on any cards:
+media-check-template-refs-header = Files referenced by templates directly will not be considered in use, unless their names start with an underscore. Referencing templates:
```
dae, line 25:
```suggestion
media-check-template-references-field-header = Anki can not detect used files when you use { "{{Field}}" } references in media/LaTeX tags. The media/LaTeX tags should be placed on individual notes instead.

Referencing templates:
```

dae (commented):
The media check is a big too eager at the moment - it reports a problem for static references like ``, even if the filename does have a leading underscore. It should probably only report if it sees a field reference inside, and not mention underscores. We will want to check for sounds and LaTeX as well, as this issue is commonly seen with those.

As an aside, I seem to recall you detect the used filenames in the template in the apkg exporter? At one point it might be nice to drop the underscore restriction and extract the filenames directly in a media check, though we'll probably want to wait for AnkiDroid to switch over to the new backend before doing that.

RumovZ:
> it reports a problem for static references like &lt;img src=logo.jpg>, even if the filename does have a leading underscore.

Actually, that was intentional, as I've assumed a list of templates would be handy in both cases: Media files prefixed with underscores need to be managed manually and the template list can help to find out if a certain file is used by anything.
But if you don't think that's helpful, I can remove the report in this case.

> As an aside, I seem to recall you detect the used filenames in the template in the apkg exporter? At one point it might be nice to drop the underscore restriction and extract the filenames directly in a media check

Yes. How would that work? Wouldn't that require rendering all the exported cards?

dae:
I'd lean towards skipping the underscores for now, and hopefully dealing with them properly by making them no longer required in the future.

> How would that work? 

We're only concerned with literal references here, eg [sound:foo.mp3], not [sound:{{Field}}], so we only need to scan the templates, like UNDERSCORED_CSS_IMPORTS but also for sounds, script tags etc. We could start with only looking for underscore-prefixed files, then remove the underscore requirement in the future.

ftl/core/media-check.ftl
RumovZ: I just noticed the changed text doesn't mention static tags like `` anymore. I assume this is in anticipation of the mentioned change to Check Media below? Should I then first update Check Media and later return to this PR to adjust it to the new reality?

dae: I don't have a preference on the order we tackle these - just think it's probably not worth mentioning static references if we'll handle them better in the future anyway.

RumovZ:
Ready.
Only field references are reported now to match the new message. I'll open an issue for the other task.

rslib/src/notetype/checks.rs
```diff
@@ -0,0 +1,169 @@
...
+        |
+            # inline math
+            \[\$\]\{\{.+?\}\}\[/\$\]
+        |
+            # math environment
+            \[\$\$\]\{\{.+?\}\}\[/\$\$\]
+        "#
+    )
+    .unwrap();
+}
```
dae, line 63: Two related issues:

- It would be nice if we were able to reuse HTML_MEDIA_TAGS, LATEX and SOUND_TAG instead of duplicating them here. 
- The field reference may be only part of the reference, eg:

[sound:{{Front}}.mp3]

Maybe we could loop through the existing regexs, extracting the referenced filenames, and then checking if they contain a field reference?

RumovZ: I thought too much in terms of speed. I see how this makes more sense.

dae:
Thanks Rumo! :rocket:

PR #2033, opened 2022-08-26
fix wrong rich-text-widget offset

BlueGreenMagick:
Fixes incorrect rich-text-widget (image/mathjax overlay) positioning

image

### Details

```html
<div class="rich-text-input">
  <div class="rich-text-editable"><img src="..."></div>
  <div class="rich-text-widgets">
    <div class="image-overlay"/>
  </div>
</div>

<style>
.rich-text-input{
  position: relative;
  padding: 6px;
}

.image-overlay{
  position: absolute;
  left: 0px;
  top: 0px;
}
</style>
```

Above is the basic html structure. `.image-overlay`'s `left` and `top` is rendered based on `.rich-text-input` position, which is the position without padding. However, its value is calculated relative to `.rich-text-editable` position which is the position after padding.

This PR fixes it by making `.rich-text-input` and `.rich-text-editable` position identical, using `margin` instead of `padding`.

### DIFF
--- a/ts/editor/FieldDescription.svelte
+++ b/ts/editor/FieldDescription.svelte
@@ -40,9 +40,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         opacity: 0.4;
         pointer-events: none;
 
-        /* same as in ContentEditable */
-        padding: 6px;
-
         /* stay a on single line */
         overflow-x: hidden;
         white-space: nowrap;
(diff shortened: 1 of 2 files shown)


hgiesel:
This ahs bothered me for quite a while :)
Note that I also addressed the issue of misaligned overlays (and strange behavior) in https://github.com/ankitects/anki/pull/2011.

dae:
Thanks BGM!

PR #2036, opened 2022-08-29
Make dvipng use transparent background by default

gnnoh:
As a result, it will look better, and strange black background in dark mode will be gone. Note that if using all-lowercase 'transparent', it will give "a simple fully transparent background with non-transparent antialiased pixels", which is unnecessary.


dae:
This has the potential to break things for some users who rely on the background to read the text against their card background color, but hopefully that is rare in practice, and it does seem like a more sensible default. Let's give it a try - thanks!

PR #2040, opened 2022-08-31
Add card meta for persisting custom scheduling state

### LINKED ISSUE #2039: Support persistent state in custom scheduler code
Allowing custom code to save/restore metadata about a card would allow people to experiment with schedulers further removed from the standard SM2 algorithm. We could do this by packing an object with arbitrary keys into card.data["meta"], and exposing a symbol like `meta` that the JS code could read from/write to.

https://forums.ankiweb.net/t/some-problems-in-implementing-a-state-of-the-art-srs-scheduler-on-anki/22705

@RumovZ feel free to tackle this if it interests you

RumovZ:
Closes #2039.
A welcome opportunity to look into custom scheduling!

### DIFF
--- a/proto/anki/cards.proto
+++ b/proto/anki/cards.proto
@@ -44,7 +44,7 @@ message Card {
   int64 original_deck_id = 16;
   uint32 flags = 17;
   optional uint32 original_position = 18;
-  string meta = 19;
+  string custom_data = 19;
 }
 
 message UpdateCardsRequest {
--- a/proto/anki/cards.proto
+++ b/proto/anki/cards.proto
@@ -44,6 +44,7 @@ message Card {
   int64 original_deck_id = 16;
   uint32 flags = 17;
   optional uint32 original_position = 18;
+  string meta = 19;
 }
 
 message UpdateCardsRequest {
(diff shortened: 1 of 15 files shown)


dae:
Thanks Rumo. I've made some minor tweaks - do they work for you?

RumovZ:
Sure!

dae:
:rocket:

RumovZ:
I just wanted to document `customData` and realised it might not be very useful in its current implementation. I understand the goal is to facilitate persistent custom variables which will be used for calculating scheduling states. Let's say we have something like this:

```javascript
const fooFactor = customData.fooFac || 1;
if (states.good.normal?.review) {
  states.good.normal.review.scheduledDays *= fooFactor;
}
customData.fooFac = calculateNewFooFactor(fooFactor); // what's the point?
```

As the custom variable must be calculated in advance, it cannot take into account the answer rating. But without it, what scheduling variable worth persisting can we possibly calculate?
Shouldn't `customData` be stored in every member of `states`, so that we can pick the new one depending on the rating?

Am I missing something?

dae:
You're right 🤦 I guess we could pack it into the top level of SchedulingState so we don't have to update every enum variant?

RumovZ:
Sounds good, I'll take care of it.

PR #2041, opened 2022-08-31
Add animation toggle to preferences

kleinerpirat:
image

The CSS transitions I used for Collapsible can cause lags on older systems or huge notetypes. Also, power users tend to lean towards a purist experience, so they might want to disable animations too.

I would probably use a `no-transition` class for body and add a `transition: none !important;` rule to the base stylesheet. @dae @hgiesel do you have other ideas on how to implement the effect of the toggle?

I've moved the settings affecting appearance together, but they do not quite deserve their own tab yet. The backend and Svelte parts aren't implemented yet, but I wanted to let you know I'm working on this feature.


kleinerpirat:
Seems to work well for the editor. It feels super snappy with animations off... I'll decrease the time of existing transitions a bit now.

ftl/core/preferences.ftl
```diff
@@ -46,3 +46,6 @@ preferences-daily-backups = Daily backups to keep:
 preferences-weekly-backups = Weekly backups to keep:
 preferences-monthly-backups = Monthly backups to keep:
 preferences-minutes-between-backups = Minutes between automatic backups:
+preferences-appearance = <b>Appearance</b>
```
dae, line 49: The only other bold text in that screen is the Synchronization label. Could we do without the label? You could argue other things like "show play buttons" also alter appearance too.

Semi-related: 
https://translating.ankiweb.net/anki/developers.html#avoid-html-where-possible

Like the fields screen, I'd prefer we add new options below instead of to the right

ftl/core/preferences.ftl
```diff
@@ -46,3 +46,6 @@ preferences-daily-backups = Daily backups to keep:
...
+preferences-animations = Animations
```
dae, line 50: Since we're not currently stopping all of them (eg graphs screen), maybe "Reduce animations" or Apple's "Reduce motion" would be more accurate?

dae:
Thanks Mattthias!

PR #2044, opened 2022-09-01
Break long words in CodeMirror

kleinerpirat:
Long words or codes inside `PlainTextInput` clip the fields currently:

image

I got the selector from this answer by Marko Letic: https://stackoverflow.com/a/57377527


hgiesel:
Maybe this can go into the more general `CodeMirror.svelte`, as we also want the word breaking for other uses of code mirror like the Mathjax editor?

kleinerpirat:
Done. Thanks for the quick reviews Henrik!

PR #2045, opened 2022-09-01
Break long words in pre tags

### LINKED ISSUE #1985: Ensure text in editor wraps
Looks like we're not using break-word at the moment:

https://forums.ankiweb.net/t/report-bug-for-version-2-1-54/21674/3

kleinerpirat:
image

Closes #1985.

### DIFF
--- a/ts/editable/ContentEditable.svelte
+++ b/ts/editable/ContentEditable.svelte
@@ -61,6 +61,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         &:focus {
             outline: none;
         }
+        & :global(pre) {
+            white-space: pre-wrap;
+        }
     }
 
     /* editable-base.scss contains styling targeting user HTML */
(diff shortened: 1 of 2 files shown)


ts/editable/ContentEditable.svelte
```diff
@@ -61,6 +61,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         &:focus {
             outline: none;
         }
+        & :global(pre) {
+            white-space: pre-wrap;
+        }
     }
 
     /* editable-base.scss contains styling targeting user HTML */
```
hgiesel, line 69: As I've mentioned here, you should put css targeting the user's input into editable-base.scss

kleinerpirat: Ah yes, thanks. I've searched for something like this but somehow overlooked it ^^

hgiesel (approved):

kleinerpirat:
Btw, see that black vertical line to the right of the field in the screenshot? It appears sometimes when I resize the window horizontally and persists between sessions - so it seems to be caused by certain width values.

I'd be curious if you can replicate this.

PR #2046, opened 2022-09-01
Make tags editor resizable using Henrik's components

### LINKED ISSUE #1940: tags area may need a maximum size or visibility toggle?
https://forums.ankiweb.net/t/toggle-off-tags-in-viewer-anki-2-1-5/21044

kleinerpirat:
Intended to close #1940.

This PR uses some of the components @hgiesel created for #1716. The vertical and horizontal resizers will be useful for the main window and browser too, so I moved them to `../components`.

- [x] default height for tags editor (probably 1 line-height)
- [x] max-height for tagsPane = all tags visible
- [x] fully expand/collapse tags editor with double click (or a separate button) *Edit: already possible, but no time for a good design right now.*
- [x] maybe swap `<hr>` for `<div>` and import handle icons via TS


kleinerpirat:
Alternatively, we could just use `Collapsible.svelte`, but I wanted to try what @firejake308 suggested in the issue. Especially @hgiesel if you are currently working on the fields dialog, I would use `Collapsible` to not interrupt your work.

hgiesel:
I actually it's a good way to get some code of the fields dialog PR field tested, so I'm all for it.

Without having tested it, there are still some things I'm concerned about, like the resizing behavior when resizing the window itself.
I'll give this a review, hopefully tonight.

kleinerpirat:
I've just merged something I've been working on for a while with your `PaneContent` component:

image

It was originally called `ScrollArea` - a component that shows an inset box-shadow in every direction content is overflowing to. When you reach the end, the shadow disappears. This integrates nicely with your panes, because those serve the same purpose: containing scrollable content.

dae:
A splitter seems like a good way of solving this. I don't have strong preferences on hr vs handle icon, but currently the handle takes up 2-3 lines worth of tags, and the bottom margin is quite large too, so there's a lot of wasted space: 

2022-09-02T11:35:26,741105016+10:00

kleinerpirat:
This is just a quick draft. I'll make it smaller once I've figured out which element's margin/padding causes this :) 

The resizer itself is quite slim already and we want to keep the clickable area at least around 8-10px. I could also make the element itself 1px and increase the clickable area with a pseudo-element.

kvn1351:
I really like this idea and it looks great imo!

May I suggest to add a small transparency gradient where the tags are cut off? I also think that the handle takes a bit too much space on the vertical axis. Other than that, I think it's very nice!

kvn1351:
188040533-e243e757-b244-4554-91fb-95cb4303672f

Here's a quick mockup of what I intended.

hgiesel:
> May I suggest to add a small transparency gradient where the tags are cut off?

The awkward look (tags are cut off before the end of the window) are caused by the webview only extending so far. A transparency gradient at the bottom of the page is something very uncommon on websites, so I'm afraid I'll have to disagree here.

hgiesel:
One base assumption when writing the Panes and Resizers for the Fields PR is that when resizing the window, all panes will lose/gain the same amount of space.

However in this scenario, the tag editor should keep the same amount of space, and only the fields should gain/lose any extra space. This means that they all have to be somewhat aware of each other. In the case of the tag editor, this is not necessary.

Maybe the Resizers could helpful as a starting point on how to deal with pointer lock, but the panes may be more confusing than helpful in this case.

kleinerpirat:
A few days ago I told you I would resort to `Collapsible` to avoid the hard route of reworking Henrik's resizers.

But looking ahead at the main window rework, where I want to implement the deck overview as a side-pane, I figured it will be worth the effort. The components saw some critical changes:

- Panes communicate their height to user components.
- Panes have a max- and min-height value, which can be used to prevent scaling above/below certain sizes.
- The height of `HorizontalResizer` can be set manually, allowing collapse/expand actions. It was also necessary for the initiation because otherwise the editor panes would take up arbitrary heights.

So far I only reworked `HorizontalResizer`, but the vertical one can be written after the horizontal example at a later time, as it isn't used yet.

---

## Todo

- [ ] fix type issues

Also, maybe I'll add buttons to expand a pane when it's height is 0, à la:
```svelte
{#if toolbarPane.height == 0}
    <div on:click={() => upperResizer.expand(toolbarPane)}>Toolbar</div>
{/if}
```
The button would need to be subtle though, because people who collapse these views do so to prevent distractions. Probably absolutely positioned too to not affect the resizer.

dae:
Thanks for your work on this Matthias. Some thoughts/issues after playing around with it for a bit:

- When using the AnKing deck for testing (due to it using so many tags), if I click on a number of different rows in the browse screen to quickly switch between notes, I find after doing so 5-10 times that the editor hangs, and WebEngine starts chewing up 100% cpu. This seems to happen more quickly when the editor is on the bottom instead of the right.
- I'm not sure about having a splitter for the toolbar area. It makes the interface more cluttered, and as you resize the window so the toolbar is shown on 1 or 2 lines, it cuts off the bottom of the buttons.
- When you resize things to a single line and then select new notes, the size increases. I suspect what the typical user wants is to be able to shrink the size down and keep it that way - both in a single session, and probably across browser opens as well (though we could tackle that later)

kleinerpirat:
Let me know if we're faring better with the single split (especially the CPU issue, which I couldn't reproduce yet).

> When you resize things to a single line and then select new notes, the size increases.

I made the tags pane reset to full height whenever the clientHeight changes, because A: I thought it would be awkard if you get into a new line while typing but don't see the cursor anymore and B: because an initial resize of the pane is needed and it was an easy way to do that. But I can experiment with other approaches too.

Now I got to go to bed, sorry for the long delay.

dae:
Good news - I went to record a video of the hang and discovered it's a particular card that's causing the issue, and the issue appears to have been introduced in #2066 instead of this PR.

Sleep well :-)

dae:
I am looking into the typing issue

dae:
Thanks Matthias! :rocket:

kleinerpirat:
Awesome, thanks! 🙏🏻

Issue #2043, opened 2022-09-01
Consider if a switch to protobuf-es is practical


dae:
https://github.com/bufbuild/protobuf-es

The code is quite new, but it passes the protobuf conformance tests, and the company backing it appears to have a good reputation. The smaller bundles, nicer API and more active development would be a nice change over protobufjs, which does not have very active development, uses namespaces for the generated TypeScript which has bitten us a few times, and needs patching to avoid trying to install npm packages at run time.

One potential issue with protobuf-es is the use of bigint for (u)int64 fields - we'd need to make sure the string fallback works correctly, or wait until iOS 14 becomes the minimum target version.

dae:
One thing we'll need to think about is how the v3 custom scheduling code is handled. We expose the protobufjs API at the moment, so we'll need to decide whether to make a breaking change, or create a compat proxy. Given it's a niche feature, we probably won't break things for too many users if we make it a breaking change.

dae:
I've done a quick performance test on the graphs screen, and unfortunately it's significantly slower. On a collection with a revlog of about 150k, data fetch is 100ms, protobufjs decoding is 100ms, and protobuf-es decoding is about 350ms. Larger collections or slower computers will of course perform worse. Most of our screens are not as performance-sensitive, but the fact that we can't use this across the ts codebase is somewhat disappointing.

PR #2049, opened 2022-09-02
Enable state-dependent custom scheduling data

RumovZ:
https://github.com/ankitects/anki/pull/2040#issuecomment-1235288957

I started by adding a `custom_data` field to `SchedulingState`, but since this message is mapped to a Rust enum and included in a whole bunch of data types (which are mostly `Copy`), this would have required a huge refactoring.
The frontend can handle it with a lot less code and redundancy. It's not perfect, but the best I could come up with.

### DIFF
--- a/proto/anki/scheduler.proto
+++ b/proto/anki/scheduler.proto
@@ -305,8 +305,19 @@ message RepositionDefaultsResponse {
   bool shift = 2;
 }
 
-// Data required to support the v3 scheduler's custom scheduling feature
-message CustomScheduling {
+// Data required to support the v3 scheduler's custom scheduling feature.
+// Consumed by the custom scheduling mutator.
+message CurrentCustomScheduling {
   NextCardStates states = 1;
   string custom_data = 2;
+}
+
+// Data required to support the v3 scheduler's custom scheduling feature.enum
+// Returned by the custom scheduling mutator.
+message NextCustomScheduling {
+  NextCardStates states = 1;
+  string again_custom_data = 2;
+  string hard_custom_data = 3;
+  string good_custom_data = 4;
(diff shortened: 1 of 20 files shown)


dae:
Keeping custom_data only in the proto message sidesteps the copy issues and reduces the amount of code required in our platform-specific UI layer. Downsides are slightly more complicated mapping, and the fact that customData is both passed in as an extra argument, and also accessible on the original states as a string, so we'll need to explain that the strings shouldn't be modified when documenting this. We could potentially avoid that by mutating the string property into an object and back again, though there's a risk that would not work correctly if/when we switch TS protobuf implementations. What do you think?

RumovZ:
> We could potentially avoid that by mutating the string property into an object and back again

Not sure if this would gain us much. We'd just replace telling users "there is this string, but don't use it" with "there is this string, but actually it's an object".
A custom interface sounds like the most sustainable solution. Apart from avoiding future breaking changes and offering a more comprehensible source to JS scripters, maybe it would even allow adding some utility functions to make common scheduling interactions easier? E.g. to help with abiding by the size restrictions. Not something we'd want to pursue right now, just meant to highlight the potential.
Of course, having to reimplement the current interface that's _almost_ perfectly fine is a bit off-putting.

dae:
> Not sure if this would gain us much. We'd just replace telling users "there is this string, but don't use it" with "there is this string, but actually it's an object".

My thinking here is that users will tend to focus on the output of the console.log() statement rather than the .proto file:
https://faqs.ankiweb.net/the-2021-scheduler.html#add-ons-and-custom-scheduling

#2043 is going to break the existing API, so when looking into that, we'll need to decide on the API we want to present going forward

PR #2051, opened 2022-09-06
Insert symbols overlay

hgiesel:
Alternative title: Emoji support

This implements the functionality of [Symbols as you type](https://ankiweb.net/shared/info/2040501954) as a built-in functionality.

Inserting Unicode characters (and sometimes even images) via typing special codes is quite established in productivity apps, I'd say, so, I think it only makes sense to add it to Anki core.

https://user-images.githubusercontent.com/7188844/188695536-f98be212-4d23-4df0-96d8-7a9a69aa6208.mov

First off: this feature is behind a feature flag. To enable, you need to have `ANKI_EDITOR_INSERT_SYMBOLS` set in your environment. That also means it can be merged as-is.

This will only work for the visual editor (rich text input). A insert symbol overlay for the html editor is thinkable, but the implementation will have to be quite differnet (contentEditable vs codemirror)

The UI itself stands. There are several open questions, which I think should be discussed:

- [x] There need to be some sources, from where to fetch symbols and their names/codes. I found two good sources:
  - For emojis: https://api.github.com/emojis
  - For maths/linguistics related characters: https://html.spec.whatwg.org/entities.json

We could simply parse those, and get a good amount of symbols and emojis to pick from. It might actually contain everything we need for now. This saves us the effort of trying to invent something ourselves.

- [x] Should these sources be served from the frontend, or in the backend via API calls? I guess there's no reason to serve from the backend, except for maybe improved speed when filtering the symbols? On the other hand, we could serve from the frontend, and use fuzzy searching javascript lib, like fuse.js?

- [x] Currently, when the symbol dropdown is open, pressing "Tab" will do nothing. That is intentional, as I haven't decided on a behaviour yet.
  - In all implementations of a "symbol picker" I've seen (Notion, Github, Slack), "Tab" picks the currently highlighted item (same as "Enter").
  - In the tag editor dropdown, tab selects the next item. However the tag editor also is a bit different behavioral-wise, as it also can have "no item selected" state, and it also modifies the input by selection an item in the menu.
 
Do you have an opinion here? 

### The following points address specific features of Symbols as you type

- [ ] One big feature of Symbols as you type, is that you can define your own symbols. Basically all symbols are "user-defined", and can be deleted/edited. For this PR however, I see them as two distinct categories: Our preset symbols on the one side, which cannot be edited or deleted, and user-defined symbols on the other side. In fact, I would suggest to keep user-defined symbols out of the MVP, and I'd implement them in a follow-up PR.

- [ ] Symbols as you type allows for symbols to be replaced without delimiters (`:`), by simply typing it (all the way). As an example, let's say you define `:->:` to be replaced with →. Instead of typing `:->:`, you could then simply type `->`, and it would be replaced. There would also no suggestion menu open during this process. I would reserve this for user-defined symbols only.

- [ ] Symbols as you type allows for more than just one symbol as the replacement. Basically going from "insert symbol" to "insert template". I think this might point to an underlying issue, if the user wants an easy way to insert a lot of boilerplate, but who are we to judge 😅.

- [ ]  It also optionally allows for HTML inside the replacements. 

The PR also addresses two minor things, beside the main feature:

- On macOs can use Ctrl+F/B/N/P in some situations now, where previously only arrow keys would work (picking a suggestion item in the tag editor / moving into mathjax editor)
- The inline padding in some dropdown menu lists (tag editor, template buttons), which was introduced with the `WithFloating` PR is fixed.


hgiesel:
I have added https://html.spec.whatwg.org/entities.json and https://raw.githubusercontent.com/github/gemoji/master/db/emoji.json as sources now, which means there are a lot of symbols to play with. 3323 in fact.

Search still works nicely, however are symbols with short names very hard to find, and usually require scrolling. Maybe a dedicated fuzzy search lib could help out here.

qt/aqt/editor.py
```diff
@@ -551,6 +551,10 @@ def oncallback(arg: Any) -> None:
             sticky = [field["sticky"] for field in self.note.note_type()["flds"]]
             js += " setSticky(%s);" % json.dumps(sticky)
 
+        print(os.getenv("ANKI_EDITOR_INSERT_SYMBOLS"))
```
dae, line 554:
```suggestion
```

dae:
Seems to work well!

> Should these sources be served from the frontend, or in the backend via API calls?

Doing it on the frontend seems ok to me. When adding support for custom replacements though, it would mean we'd either need to do those on the backend, or the frontend would need to fetch them when the editor is initialized. That would probably be fine - users probably wouldn't add hundreds of them.

> Currently, when the symbol dropdown is open, pressing "Tab" will do nothing.
> In all implementations of a "symbol picker" I've seen (Notion, Github, Slack), "Tab" picks the currently highlighted item (same as "Enter").

When in doubt, copying what others do is usually a safe bet :-)

>  In fact, I would suggest to keep user-defined symbols out of the MVP, and I'd implement them in a follow-up PR.

That works for me

> Symbols as you type allows for symbols to be replaced without delimiters

My only concern here is performance - ideally we'd only be scanning through the user-defined choices here, as we don't really want to be searching 3k+ records for every keystroke.

> Symbols as you type allows for more than just one symbol as the replacement.

Limiting to a single character would reduce the usefulness of this feature I think. Might be nice to allow the users to use HTML too.

I've only skimmed the source changes so far, but am happy to merge this in if you're ready.

hgiesel:
I gave tab the same behavior as enter. I'll revisit the tab completion behavior in the tag editor as well soon.

I'm happy withing merging it :)
I'll try out a fuzzy search algorithm in another PR.

dae:
> I'll try out a fuzzy search algorithm in another PR.

As long as it doesn't cause a noticeable performance impact, sounds good.

PR #2052, opened 2022-09-06
Collapsible patch

kleinerpirat:
This patches the issues that cropped up with #2011.

### Todo:

- [x] implement reduce motion setting for the animations
- [x] improve method of awaiting the required height of the collapsible content for the expand animation


dae:
Sorry Henrik, I should have waited until @kleinerpirat was ready before requesting a review

kleinerpirat:
@hgiesel I'm ready :)

Regarding the reduced motion mode: Is there anything other than stores that comes close to body classes in terms of global accessibility? The store still has to be initialized and retrieved by components, meaning lots of extra complexity for animated UI. 

Perhaps it's better to just check for the body class ("reduced-motion") from inside the animated component.. wdyt?

hgiesel:
I would push the review for this until after #2058, WDYT?

ts/editor/Fields.svelte
```diff
@@ -12,6 +12,7 @@ Contains the fields. This contains the scrollable area.
 
 <style lang="scss">
     .fields {
+        padding-top: 5px;
```
hgiesel, line 15: Bildschirmfoto 2022-09-13 um 00 57 08

This will expose some area above the label container. Should probably be margin-top.
Alternatively, we could drop the grid-gap a few rows below, and add the padding-top in the LabelContainer component.

hgiesel (changes requested):
Just out of curiosity, there's another svelte function, called [`spring`](https://svelte.dev/tutorial/spring), which is also in the svelte/motion package. It actually says, that it as opposed to tween, it doesn't have a fixed duration, but rather adjusts to the "velocity". I thought that maybe releases us from having to calculate the duration ourselves?

I tried it, and even though it seems the duration isn't fixed, it seems to open/close the fields much quicker if they're bigger, than what the dynamicDuration function did. I'll let you make the judgement here.

kleinerpirat:
I've now tried `spring`, but I couldn't quite achieve the same effect we get with `tweened`. It seems to produce more of an `easeOut` function, which doesn't look that nice when collapsing (slows down right before it disappears).

> it [spring] seems to open/close the fields much quicker if they're bigger

That doesn't sound ideal to me, let me explain why. When I tested the original slide animation on fields with large content such as images for the first time, they seemed to rush by a bit too violently, because speed has to increase when the duration is constant. I figured increasing the duration ever so slightly with content height would give a more natural effect.

Perhaps this isn't as big of a deal for high-refresh-rate screens, but on my 60hz laptop monitor, super fast closing animations look a bit janky.

hgiesel (approved):

hgiesel:
Btw, the collapsing is really smooth now, it looks great

dae:
Yeah, it feels nice and responsive. Thanks for this Matthias, and thanks for the review Henrik.

Issue #2056, opened 2022-09-10
Regressions caused by removal of overflow: hidden (PR #2011)


kleinerpirat:
I noticed some more regressions that were introduced with #2011 and figured it's best to discuss them here.

- [x] long field descriptions overflow horizontally again (i.e. are not ellipsed)
caused by removal of `overflow-x: hidden` CSS rule
- [x] field label containers do not stick anymore
The fields area isn't constrained anymore, meaning the whole document is scrollable rather than just the fields. Therefore, labels start sticking underneath the toolbar. Probably also caused by the removal of CSS rules written for #2002.

@hgiesel sometimes overflow has to be clipped to achieve certain effects. Can we keep such rules without disrupting your floating overlays?

hgiesel:
> long field descriptions overflow horizontally again

That's an easy fix, I pushed a fix

> field label containers do not stick anymore

Ah, that's a bigger oversight. Sorry for destroying your feature. I guess I wasn't used enough to the sticking labels yet, so I didn't even notice when they were gone again :(

I'll give this a try and try to refactor it. I have a long-standing suspicion, what might be causing our overflow woes, and see if I can confirm.

PR #2057, opened 2022-09-10
Add back overflow:hidden to field descriptions

hgiesel:
When I was trying to cleanup unnecessary overflow:hidden in #2011, this one got under the bus.

Mentioned in #2056


hgiesel:
Is also addressed by #2058

PR #2058, opened 2022-09-10
Fix sticky field labels - Make editor toolbar+tag editor non sticky in editor

### LINKED ISSUE #2056: Regressions caused by removal of overflow: hidden (PR #2011)

hgiesel:
Closes #2056

@kleinerpirat The issue was not the removal of display: hidden, but that the editor toolbar was sticky even though it didn't need to be.

I think I didn't exactly understand what position:sticky meant when I first implemented the note editor. You only need position:sticky, when you can scroll past the element, but you want the element to "stick by", but in our design you don't need to scroll past the editor toolbar (like you scroll past the label containers). And in fact - you never scroll past the tag editor either, so I could remove the sticky container there as well.


kleinerpirat:
> you never scroll past the tag editor either, so I could remove the sticky container there as well.

Yep, I've actually worked on a PR this summer that removes almost all uses of `sticky` in Anki's webviews, but decided to wait until the theme changes are done.

Limiting scroll to the components that really need it also simplifies offset calculations and makes layouts simpler. The body never needs to exceed 100vh in height.

ts/editor/LabelContainer.svelte
```diff
@@ -37,16 +37,13 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-        position: sticky;
         justify-content: space-between;
-        top: 0;
-        padding-bottom: 1px;
+        padding: 0 3px 1px;
 
-        /* slightly wider than EditingArea
-           to cover field borders on scroll */
-        left: -1px;
-        right: -1px;
```
hgiesel, line 48: By moving the inline padding into the label container and the editor field, I came to the conclusion we don't need this anymore

PR #2059, opened 2022-09-10
Fuzzy search in symbol insertion overlay

hgiesel:
This uses [fuse.js](https://fusejs.io/) to fuzzily search through symbols.

It also refactors the query input in many ways. For example you press delete during searching, like entering `:alpha`, and deleting back to `:al` and still have the symbols overlay behave correctly.

The logic for queries to contain html and being automatically insert already stands as well, however there's no API to access it yet, but the _lookup_ already takes place. "Auto-insertion" codes can have a maximum of 5 characters to make sure there's no performance impact while typing.

Edit: There still seems to be an issue sometimes when trying to update the query (the overlay will now show up).


hgiesel:
Currently, entering `<b>:for</b>al` and then inputting `l`, will not trigger a search for `forall`, because of the <b> formatting, but I think that should be fine as the next stage.
The current state is ready for merging. The feature here is the fuzzy search and more practical showing of the symbols overlay (will keep showing, if you delete a character, while the overlay is open)

dae:
Thanks Henrik!

PR #2060, opened 2022-09-10
Fix ButtonGroupItem margin and remove hard-coded radius

kleinerpirat:
## Main
image
- `margin-left: -1px` is only applied to one button, the others aren't selected by the CSS rule `not(:first-child)` because they're nested

## PR
image
- border radius and margin are dynamically applied
- hard-coded definitions have been removed from user components


ts/components/ButtonGroup.svelte
```diff
@@ -22,7 +31,15 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-    <slot />
+    <DynamicallySlottable
+        slotHost={ButtonGroupItem}
+        {createProps}
+        {updatePropsList}
+        {setSlotHostContext}
+        {api}
+    >
+        <slot />
+    </DynamicallySlottable>
```
hgiesel, line 42: `DynamicallySlottable`, or rather its predecessor, used to be within ButtonGroup, but I moved it outside.

DynamicallySlottable only makes sense if it is used with ButtonGroupItem (or Item), and is supposed to allow add-ons devs to, well, dynamically slot components into it. I believe this could be a very powerful method of writing anki add-ons. Even features like the symbols insertion overlay I just implemented, uses APIs which might as well be used by add-ons.

**Because** it only makes sense if used with ButtonGroupItem, I would like them only ever to be used together, in the same component, otherwise they might get out of sync. 

This add-on API is still experimental however, mostly because it probably goes over the head of most (all ^^;) Anki add-on devs.

hgiesel (changes requested):

ts/editor/editor-toolbar/TextColorButton.svelte
```diff
@@ -126,31 +128,37 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-            keyCombination={pickCombination}
-            on:input={(event) => {
-                color = setColor(event);
-                bridgeCommand(`lastTextColor:${color}`);
-            }}
-            on:change={() => setTextColor()}
-        />
-    </IconButton>
+    <ButtonGroup>
+        <ButtonGroupItem>
```
hgiesel, line 132: For a similar reason, I'd like to keep the `ButtonGroup` out of the button components. The name does not imply that it's a button group, and more importantly, the text color button itself doesn't need the ButtonGroup to function, so the user should wrap it in a button group instead (or not).

ts/components/ButtonGroup.svelte
kleinerpirat: Okay, thanks for the explanation. I thought I'd reduce some overhead by moving it into `ButtonGroup`  because I need `DynamicallySlottable` everytime, otherwise I get this error:
```
ButtonGroupItem: should always have a slotHostContext
```

ts/editor/editor-toolbar/TextColorButton.svelte
kleinerpirat: Should I resort to hard-coded radii instead? `ButtonGroup` handles the radius of its items nicely, it would be a pity to lose this functionality.

hgiesel: See #2069

kleinerpirat:
Closing in favor of #2069.

PR #2068, opened 2022-09-14
Fix Windows dark mode detection: OpenKey can fail

ZoomRmc:
#1497 introduced reading hardcoded Windows Registry key, which assumes key exists. This is not true on Windows 7. Later addition of `try-except` block missed that winreg.OpenKey might fail.

This fix allows launching current version of Anki on Windows 7 when installed with Pip with the [modified Python 3.9](https://github.com/NulAsh/cpython) installation.


dae:
Thanks!

PR #2069, opened 2022-09-14
Remove per-button border-radius property

hgiesel:
This is in response to @kleinerpirat #2060

I think this is the way we want to go.
The idea is simple: Rather than trying to apply the border to each and every button, you give the *ButtonGroup* a border, and make the buttons themselves entirely without border.

In the ButtonGroup, you also apply a gap (with flex or grid), and set a background property. This background property will now be visible in the gaps and serves as the visual border *between* buttons.

image

If you don't mind, I'd continue working on this PR, once #2016 is merged. The button components are some of the very first I've written, so it'll be fun to refactor them :)


kleinerpirat:
Sure, go ahead! Thanks for looking into this. I was too boxed in to see this obvious solution :)

kleinerpirat:
Henrik, glad to see activity from you again!

I've been working on a PR that makes your work here compatible with main. I really liked your changes on `_button-mixins.scss`. Since I broke your PR with my design changes in the meantime, I felt it's fair I fix it myself. Let me know if that's alright.

dae:

PR #2070, opened 2022-09-14
Single overlay instead of per field

hgiesel:
This refactors our three overlays (Mathjax + Image + Symbols insert) to be initialized once per the note editor instance, rather than once per field, which makes much more sense anyway.

By moving them outside of the individual field, I could set EditorField to be overflow:hidden. This means I can use a similar strategy as [here](#2069) to avoid having to set border radii on the editing inputs (e.g. see how much css I could remove in [PlainTextInput.svelte](https://github.com/ankitects/anki/compare/main...hgiesel:single-overlay#diff-aeefc715d07ebf71ae6457c22eab5aff6c705a4e72b6b05ca0b287a9d17a7404)).

It also fixes some focus behavior in the note editor: You can unfocus fields again by clicking on empty space :)

Caveat:
You will notice a thicker bottom-border on the field:

Bildschirmfoto 2022-09-15 um 00 56 34

That's because the grid-gap trick doesn't work out quite yet for the editing inputs. The HTML editor is only hidden, but not unmounted, so grid-gap will draw a single pixel border between the visual editor and the hidden html editor, creating this thicker bottom border.
I'd like to try and actually unmount the editor when hiding it, but wanted to avoid this PR getting too large.

@kleinerpirat I moved the box-shadow from the EditingArea to EditorField, and also changed how it looks, for several reasons:
1. I want to use a similar trick as [here](#2059), to avoid having to define individual border radii on editing inputs.
2. The editor field defines a border, but the editing inputs actually define their own background color, so any inset box-shadow on the editor field would be overdrawn.
3. I would like to have a single source of box-shadows, so we can compare them, rather defining box-shadow for each component that uses one. I implemented the sass/elevation some time ago, and I'm not sure you were aware of it. I basically intended to be the _vars.scss of box shadows :) . I've created it by referencing a material-ui css lib, and it's basically draws several box-shadows, to create a nice shadow dropback effect.

P.S. I'm happy with waiting until after #2016 to avoid conflicts


dae:
Looks like some nice improvements! The bottom border is a bit weirder on a Linux machine here, but I'm ok with merging this if you plan to address that in a follow-up PR.

2022-09-16T14:15:19,085350107+10:00

kleinerpirat:
I suggest making the box-shadow a bit softer:



A border-based focus indication would look cleaner imo because you're restricted to a vertically offset box-shadow because of the sticky labels, whereas a border is perfectly symmetrical:

 -> 

But don't feel obliged to deal with finicky stuff like this yourself. I can always patch rough edges (in this case literally) later.

hgiesel:
> I suggest making the box-shadow a bit softer:

Sounds good to me. If we make more components use the elevation sass package, we can adjust it there for all of them.

> A border-based focus indication would look cleaner imo

Like I said, that's very subjective, I actually like the left version better (except for the double bottom-border of course) :) I think the thick border is a bit overkill. Maybe we can meet somewhere in the middle, as I also don't feel that strong about it.

dae:
Happy to merge this when the conflicts are resolved. Would be nice to have this in the first beta build if you have time.

dae:
Any chance you'll have the time to push this over the finish line soon?

dae:
Superseded by #2144

PR #2071, opened 2022-09-17
Add option to shrink editor images by default

### LINKED ISSUE #1894: Global switch for small/large images by default in the editor?
https://forums.ankiweb.net/t/improving-image-resizing-in-editor/20200

I wonder if we could support a toggle that changed whether images without a data-shrink=true/false tag are shown as small or large? Eg true=small, false=large, missing=whatever the current global switch specifies. This would allow users to quickly switch between all-small and all-large, though if they manually adjusted an image to be different to the current default, it would "stick" and need manual adjustment from that point forward, which is potentially confusing. 

Other ideas welcome too :-)

dae:
+ Don't persist shrinking toggle

Closes #1894

@hgiesel this is what I had in mind on #1894 - WDYT?

### DIFF
--- a/ts/editor/editor-toolbar/OptionsButton.svelte
+++ b/ts/editor/editor-toolbar/OptionsButton.svelte
@@ -1,0 +1,49 @@
+<!--
+Copyright: Ankitects Pty Ltd and contributors
+License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+-->
+<script lang="ts">
+    import CheckBox from "../../components/CheckBox.svelte";
+    import DropdownItem from "../../components/DropdownItem.svelte";
+    import IconButton from "../../components/IconButton.svelte";
+    import Popover from "../../components/Popover.svelte";
+    import WithFloating from "../../components/WithFloating.svelte";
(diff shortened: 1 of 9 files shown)


dae:
@hgiesel I've exposed the toggle in the editor via an options button, and fixed the earlier issues. Do you have any objections to me merging this?

ts/editor/editor-toolbar/OptionsButton.svelte
```diff
@@ -0,0 +1,49 @@
...
+
+<WithFloating
+    show={showFloating}
+    placement="bottom"
+    inline
+    on:close={() => (showFloating = false)}
+    let:asReference
+>
+    <span use:asReference>
+        <IconButton
```
hgiesel, line 33: I think `<IconButton slot="reference">` should work just fine here

hgiesel (approved):
Looks good to me, just a minor comment :)

ts/editor/editor-toolbar/OptionsButton.svelte
dae: Dang, made the change but forgot to commit it :sweat_smile:. Pushed a follow-up commit.

PR #2073, opened 2022-09-17
Color Palette Patch 1

kleinerpirat:
In the rush to get #2016 merged, I introduced some bad styling that I'd like to fix with this PR.

## Fixes:
- Several supposedly faint colors not being faint enough (forgot to adjust after change to 20 gray tones)
- QTableView headers not having separators (due to `border-subtle` change)
- Preferences QTabWidget buttons not having visible borders (also due to `border-subtle` change)
- button gradients getting dark too early (moved y-coordinate for the end color further down)

## New:
- There's now a clearly visible handle for both separators in the browser

### DIFF
--- a/qt/aqt/browser/table/table.py
+++ b/qt/aqt/browser/table/table.py
@@ -12,7 +12,7 @@ from anki.collection import Collection, Config, OpChanges
 from anki.consts import *
 from anki.notes import Note, NoteId
 from anki.utils import is_win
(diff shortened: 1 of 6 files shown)


dae:
One minor issue I noticed is that the splitter icon is black in dark mode, making it hard to see. Otherwise looks good; the handles should blend in nicely with #2046

PR #2074, opened 2022-09-17
Add orientation toggle to browser view menu

kleinerpirat:
Makes https://ankiweb.net/shared/info/831846358 redundant.

Based on personal preference, I made the splitter default to `Qt.Orientation.Horizontal` (i.e. the one that was only achievable with the add-on). I'll create a forum poll to see if I'm with the majority here. The shortcut is taken from the popular add-on. Heads-up @ijgnd to set the max-version number to 2.1.54 to avoid an ambiguous shortcut error.


ijgnd:
I deliberately used an uncommon shortcut so that my add-on would not conflict with another add-ons's shortcut so that my add-on would not break others. 

I suggest to consider using a shorter shortcut.

Maybe @AnKingMed already has popularized an alternative shortcut to use in some video?

AnKingMed:
I have been using Ctrl+shift+8 I believe (8 because it's two circles on top of each other.. helped me remember it). I used to use Ctrl+8 until we added all the tags

kleinerpirat:
> I think a button would be more practical

I didn't expect toggling orientation to be as frequently used to justify a simpler shortcut or button. Isn't this something you'd do once when installing the software and then never use again?

Perhaps the better approach is to make the orientation respond to window width (vertical for slim window, horizontal for landscape ratio) :thinking:

AnKingMed:
I would still give an option for the user to change it. I toggled frequently enough I would personally use a button. Might also consider a way to click a note and pull up the editor on the sidebar and then either an X or re-clicking the note to close the editor so that you can better view the columns for browsing

ijgnd:
I hope this gets merged because it's one add-on less to maintain.

1. Maybe just add a button to the [fastbar addon](https://ankiweb.net/shared/info/46611790). People who really like buttons probably know this add-on because of all the anking videos? 
2. About the shortcut: I don't think that this is important but why not think about it for some time? The nicest list of built-in shortcuts I know is the [default config of the customize shortcuts addon](https://github.com/Liresol/anki-custom-shortcuts/blob/dev/custom_shortcuts/config.json). I think "Ctrl+Alt+8" is neither easy to press nor to remember because "Ctrl+Alt" is usually about modifying "Ctrl+" as in Search->Replace. Maybe something shorter like "Alt+8" (though "Alt" is rarely used as the sole modifier key in Anki) or something more memorable like "Alt+H" ("h" for horizontal) or "Ctrl+Shift+H"? "Ctrl+Shift" because it seems to be commonly used for shortcuts that are related to the UI in the browser (though it's usually about bringing up dialogs like "add tag", preview, reschedule etc.)

&nbsp;

Anking wrote: 

> Might also consider a way to click a note and pull up the editor on the sidebar and then either an X or re-clicking the note to close the editor so that you can better view the columns for browsing

I agree that having a better version of of my add-on [Browser Maximize/Hide Table/Editor/Sidebar](https://ankiweb.net/shared/info/1819291495) built-in would be nice. But this add-on seems to be much less popular (iirc it has 20x less downloads than the side-by-side add-on).

kleinerpirat:
> I would still give an option for the user to change it.

Ideally, we'll avoid adding another option to the preferences screen. This three way switch I built from a `QTabWidget` inside a `QWidgetAction` should do the trick:

image

The widget a bit of a hack, but I find it more intuitive to use than a tri-state checkbox. A shortcut could then cycle between the three states and show the currently selected mode with a tooltip message.

dae:
Thanks for working on this Matthias, and for putting up a poll. I'm happy to change the default to horizontal, as that seems to be the more popular, and anyone who prefers the status quo can revert back. Agreed that this option is best put in the view menu, but I'm not a big fan of the current widget - it makes the menu unnaturally wide, and the buttons seem to be finicky to click on (depending on when I aim, they seem not to trigger half the time when testing on Linux). And it looks quite bad on macOS - it's very unnatural to embed widgets at the top. How about a more traditional approach of making Orientation a submenu of the View menu, with the three options listed as three menu items, and a tick on the active one?

Regarding a shortcut key, alt+... has a rather high chance of conflicting with default macOS shortcuts. With an auto option, I wonder how frequently people will need to switch between them?

dae:
Screen Shot 2022-09-19 at 11 09 44 am

AnKingMed:
I like the submenu idea.  I also don't know that a shortcut is totally necessary with this availability. I was never frequently switching, but I didn't want it to be hidden in the preferences menu and require restarting the browser or something like that

dae:
Would "Layout" be a better word than "Orientation"?

kleinerpirat:
Time to get an Apple test machine I guess. I replaced the widget with a submenu and renamed "Orientation" to "Layout". 

@dae Regarding the new `enums.py`: I created this file to avoid circular imports. Does anything speak against moving all existing Python enums into this file in a follow-up PR?

dae:
Catch-all files like "enums" and "consts" are not great - I'd rather keep the structures close to where they're actually used. The cyclic imports are a pain, but can be worked around. You could fairly point out that I just added some strings to a catch-all qt-misc.ftl, but in that case the file already existed ^_^;

The new menu works well - thank you as always!

PR #2078, opened 2022-09-19
Make mdi icons for Qt themeable

### LINKED ISSUE #2063: Change fill color of SVG used in Qt stylesheet
In `qt/aqt/data/qt/icons/BUILD.bazel` you can make mdi icons available to Qt by simply appending their name.
For example, I used this method to set the triangle icons for table headers:

image

https://github.com/ankitects/anki/blob/8f8f3bd465ed7e5c47108374106b8ec6ec2bd072/qt/aqt/data/qt/icons/BUILD.bazel#L37-L39
https://github.com/ankitects/anki/blob/8f8f3bd465ed7e5c47108374106b8ec6ec2bd072/qt/aqt/stylesheets.py#L254-L256

The icons show in black by default, but ideally it would have the default text color. After a lot of research I've concluded it isn't possible to set or invert an SVG's fill color in a Qt stylesheet.

I can think of two options:
- create two separate icons from Bazel, one for dark and one for light theme
- edit the svg fill color at runtime in `theme.py`
(preferred because it would allow theme-devs to use arbitrary colors)

Sadly, I don't know how to achieve either of the two.

kleinerpirat:
Closes #2063.


qt/aqt/theme.py
```diff
@@ -81,8 +75,43 @@ def set_night_mode(self, val: bool) -> None:
...
+            data = re.sub(r"fill=\"#.+?\"", f'fill="{self.var(color)}"', data)
+        else:
+            data = re.sub(r"<svg", f'<svg fill="{self.var(color)}" ', data, 1)
+
+        copy_path = os.path.join(
+            os.path.join(aqt_data_folder(), "qt", "icons"), f"{key}.svg"
+        )
+        # create copy with specified fill color
+        with open(copy_path, "w", encoding="UTF-8") as file:
+            file.write(data)
```
dae, line 109: Anki may be installed in a location where its data files are not writable, which will cause this to fail. The files will either need to be placed somewhere like a temp folder (which will introduce problems with temp file cleaners), or generated up-front. I'd lean towards the latter, as the extra files aren't going to contribute a large amount to the total bundle size, and having them generated in advance means there's no risk of impacting startup time. I'll push a commit that shows how you can do this with Bazel.

qt/aqt/data/qt/icons/BUILD.bazel
```diff
@@ -56,9 +57,32 @@ copy_mdi_icons(
...
+    color = "FG_SUBTLE",
+)
+
 filegroup(
     name = "icons",
-    srcs = ["mdi-icons"] + glob([
+    srcs = [
+        "mdi-icons",
+        "chevron-down",
+        "drag-horizontal",
```
dae, line 84: This is currently including base.svg as well as base-light.svg and base-dark.svg. We could avoid that by creating a separate mdi-icons that holds the files we plan to modify, and then not including it in this filegroup.

qt/aqt/theme.py
kleinerpirat: Thanks for showing me this! I modified the script to allow an arbitrary amount of color variants for each icon.

dae:
:rocket:

PR #2082, opened 2022-09-23
Redesign deck options inputs

### LINKED ISSUE #2076: Dropdown in options screen is too small
@hgiesel #2011 reduced the size of the dropdown to the right of the Save button - was that deliberate? The new size is a bit too small to be comfortably tapped with a finger.

kleinerpirat:
The main goal of this PR was making the deck options inputs look more like the new Qt stylesheet overrides for Windows and Linux.

- `SelectButton` has been replaced with a new `Select` component.
  - This component uses the new floating elements and has a fancy animation - as a side-effect all other floating elements are animated now.
  - [x] Known issue: The popover doesn't overlay the sticky containers in the Change Notetype screen yet.

- `SpinBox` got a rework too, it now looks more like `QSpinBox` for Windows and Linux. It also got a new feature:
  - When the input is focused, you can change the value with the scrollwheel.
- The save button dropdown is now wider (Closes #2076).


kleinerpirat:
Some parts involving dropdown selections are still broken. I'll address the issues tomorrow.

dae:
The animations look nice, and it's great that the flash on dropdowns in night mode is gone. Some minor issues I noticed:

- spinbox controls are too small on iOS to tap - we should either hide them on mobile, or make them horizontal
- the down chevron is not vertically centered in the dropdowns

kleinerpirat:
A design like
```
<    2.50    >
```
is probably the better choice then. I already implemented it but not at home rn, so I'll push that tomorrow morning.

dae:
The new spinners look good, but when I tried them out on iOS, I realised we probably won't want to show them anyway. The issue is that a double-tap on a page will zoom in or out, so if you try to increment/decrement quickly, it ends up zooming in or out. So at the least, we'd probably want to hide them on mobile. As that was the only reason for splitting them up, maybe it would be better to revert to the previous design, so things can be left-aligned again? If you agree, sorry for the wasted work :-(

Two other minor issues I noticed:

2022-09-27T11:34:00,384370017+10:00

- The text inside the dropdowns appears to be slightly above vertical center
- The chevrons on the smaller dropdowns are also above center

kleinerpirat:
> on iOS [...] we probably won't want to show them anyway.

My plan for all new input designs is to hide a conventional HTML `<select>`, `<input>` etc. within the component that will be activated only on mobile, because mobile OSs usually have their own in-your-face UI for these elements. I'm not sure holding back the beta is worth it for something I'll address in the future anyway.

Personally I see the centered numbers as an improvement, because now you can distinguish spinners from plain text inputs.

dae:
Ok, I'm happy to run with this for now; it'll give us a chance to see what people think in the beta.

PR #2091, opened 2022-09-26
Fix pm.night_mode()

abdnh:
`pm.night_mode()` is used by some add-ons, but it now always returns False. Looks like it should have been updated in #1497. I've also removed `set_night_mode()` as it doesn't appear to be used by any add-ons.


dae:
Thanks Abdo!

PR #2095, opened 2022-09-28
Add API shim for webview.get_window_bg_color

glutanimate:
@dae, it looks like there was a small breakage with the AMBOSS add-on after all, but in `WebView` APIs rather than `ThemeManager`. Hope this is OK to add for now while the add-on transitions away from this API.

----

### DIFF
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -717,8 +717,6 @@ html {{ {font} }}
 """
         )
 
-    @deprecated(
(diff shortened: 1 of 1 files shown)


dae:
The 'deprecated' part is automatically included - currently this prints:

file.py:line:get_window_bg_color() is deprecated: webview.get_window_bg_color() is deprecated. Use theme_manager.qcolor() instead

glutanimate:
Sorry about that one, should be fixed now

dae:
Thanks!

PR #2096, opened 2022-09-28
Set line-height = 1 on inline floating-reference

kleinerpirat:
Fixes
https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/9
https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/12


dae:
Thanks for the quick fix!

Issue #1952, opened 2022-07-08
Investigate shortcut issues in review screen in Qt 6.3/Windows/Linux


dae:
https://forums.ankiweb.net/t/bug-updated-to-2-1-54-and-e-to-edit-shortcut-no-longer-works/21109

meehl:
I wanted to chime in and point out that this doesn't seem to be a Windows-specific problem. I started encountering the same issue with Qt6 on my Linux machine since 2.1.50 (Qt5 & Qt6 split) and it still persists in 2.1.54.

My issue is similar to the one described in this [comment](https://forums.ankiweb.net/t/bug-updated-to-2-1-54-and-e-to-edit-shortcut-no-longer-works/21109/5). I use type-in-the-answer cards which require me to input an answer in hangul(korean). After pressing Enter to confirm, the back card gets displayed but none of the hotkeys like 1/2/3/4 work. Pressing TAB once or clicking into the area at the bottom, where the ease level buttons are located, lets me use the hotkeys again.

versions and system info:
Anki Version ⁨2.1.54 (88e539aa)⁩
Python 3.10.5 Qt 6.3.1 PyQt 6.3.1
IME: fcitx5 5.0.18
WM: sway 1.7

vintotal:
Do you think the following problem is related? On Windows, my keyboard shortcuts for language switching (Alt+Shit+1 to 6) make Anki crash. They work in other apps. To avoid the problem, I have to use Windows key + space.
    
Anki 2.1.54 (b6a7760c) Python 3.9.7 Qt 5.15.2 PyQt 5.15.5

dae:
That's a long-standing Qt issue.

PR #2099, opened 2022-09-30
Fix scrolling with keys / keyboard event listeners not working on answer side

hikaru-y:
### Issue
Page scrolling with arrow keys or PageUp/Down does not work on the answer side. Also, keyboard event listeners set in the card template do not work on the answer side.

The issue appears to have been introduced in the commit 5bf031f. I have come up with a better workaround for #1952, inspired by 5bf031f. Not only does this not cause the above issue, it also works around #1952 a little better. With 5bf031f, once you turn the IME on, clicking somewhere other than the type box to take focus away from the type box does not turn the IME off, so you cannot use the shortcut keys. However, with this workaround, the IME will be  turned off.

### Tested with
| OS | Input Method | Note |
| ---- | ---- | ---- |
| Win10 | MS Japanese IME | |
| Win10 | Google Japanese Input | |
| Lubuntu 22.04 | iBus + Mozc | Even after this PR, when the focus is lost from the type box, the task tray icon of Mozc remains <kbd>あ</kbd> (i.e., it does not change to <kbd>A</kbd>) so it does not look like the IME is turned off, but the shortcut keys will actually work. |

### DIFF
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -653,12 +653,6 @@ class Reviewer:
     def _onTypedAnswer(self, val: None) -> None:
         self.typedAnswer = val or ""
         self._showAnswer()
-        self.unfocus_typing_box()
-
-    def unfocus_typing_box(self) -> None:
-        # shifting focus to the bottom area works around a WebEngine/IME bug
-        # https://github.com/ankitects/anki/issues/1952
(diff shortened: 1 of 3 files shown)


dae:
Great, thanks for this Hikaru, and thanks for testing on both platforms.

PR #2102, opened 2022-10-01
Fix duplicate highlight and improve its look

### LINKED ISSUE #2098: regression: duplicates are no longer red
After a recent update:
```
[2022-09-27T06:31:38+0200] [ALPM] upgraded anki-git (r9620.3ec7e23d7-1 -> r9640.fe9b0aecf-1)
```
I noticed that duplicates are no longer highlighed with red:
image
This header with "Show duplicates" used to be red in previous versions. I understand it could be a matter of personal preferences and tastes, but big read header was just sooo easier to notice, than a bleak "Show duplicates" label. I wonder if that could be reverted or configured. Thank you.

kleinerpirat:
Closes #2098.

The big red highlight bar looks off on the new fields. To keep that look, I would have had to manage border-radii with variables (or cut overflow, but that's no good either). I went for a box-shadow color change on `EditingArea` instead, which looks more elegant anyway.

### DIFF
--- a/ts/editor/EditingArea.svelte
+++ b/ts/editor/EditingArea.svelte
@@ -199,7 +199,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
             inset: 0;
             pointer-events: none;
             border-radius: 5px;
-            box-shadow: 0 0 2px 1px var(--shadow);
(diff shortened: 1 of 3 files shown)


dae:
Agreed - looks good. Thanks!

PR #2103, opened 2022-10-02
Use showText to show add-on import errors

abdnh:
Long error messages can be completely unreadable as QMessageBox doesn't show a scrollbar and the text can't be selected and copied by default (on Windows at least).

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -45,6 +45,7 @@ from aqt.utils import (
     send_to_trash,
     show_info,
     showInfo,
+    showText,
     showWarning,
     tooltip,
     tr,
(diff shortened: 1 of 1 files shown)


qt/aqt/addons.py
```diff
@@ -246,11 +247,15 @@ def loadAddons(self) -> None:
...
-                showWarning(
-                    tr.addons_failed_to_load(
-                        name=addon.human_name(),
-                        traceback=traceback.format_exc(),
-                    )
+                error = tr.addons_failed_to_load(
+                    name=addon.human_name(),
+                    traceback=traceback.format_exc(),
+                )
+                txt = f"<h1>{tr.qt_misc_error()}</h1><div style='white-space: pre-wrap'>{error}</div>"
```
dae, line 254: I think this may fail to render properly if the traceback happens to contains `<` or `>`? What about using the plain type instead, and passing qt_misc_error() in as the title= arg?

abdnh: I opted to just use html.escape here, like it's done in errors.py (which I tried to follow but didn't notice the html.escape call).

dae: That works too - thanks!

PR #2107, opened 2022-10-06
fix css var not renamed

BlueGreenMagick:
This PR fixes a variable that should have been renamed in #2016

I think you also missed `--highlight-fg` and `--highlight-bg` in `DropDownItem.svelte`, but I wasn't sure which color this should reference instead.


dae:
Thanks BGM!

@kleinerpirat what do you want to do about highlight-fg/bg? The highlighting still seems to work for some reason.

kleinerpirat:
> The highlighting still seems to work for some reason.

The vars `--highlight-fg/bg` have never been removed.

dae:
Well that would explain it :-)

PR #2110, opened 2022-10-07
Decrease minimum value of interval modifier

dinoboy197:
I'm proposing a change to the minimum value of the interval modifier in the advanced settings for a deck.

[The interval modifier documentation suggests that retaining 90% of knowledge of mature cards is a reasonable target](https://docs.ankiweb.net/deck-options.html#interval-modifier). Further, it suggests that by computing `log(desired retention%) / log(current retention%)`, the resulting value can be used as an interval modifier. For example, if the current retention is 60%, and the target is 90%, the interval modifier should be `log(0.9)/log(0.6) = 0.21`.

The minimum value of the interval modifier is currently 0.5. This minimum prevents using an appropriately calculated interval modifier for any current retention value less than 81% (`log(0.9)/log(0.81) = 0.5`).

I believe it's reasonable that some users might have a current retention value less than 81% (I do on some decks). Rather than forcing a user to restructure or ignore certain notes rather than set an artificially high interval modifier, I propose that the minimum value be lowered. I've arbitrarily chosen a new minimum value of 0.1. This would allow a user whose target is 90% and current retention of 35% or higher to set an appropriately calculated interval modifier (`log(0.9)/log(0.35)=0.1`).

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -106,6 +106,7 @@ Bart Louwers <bart.git@emeel.net>
 Sam Penny <github.com/sam1penny>
 Yutsuten <mateus.etto@gmail.com>
 Zoom <zoomrmc+git@gmail.com>
+Taylor Raack <taylor@raack.info>
 
 ********************
 
--- a/ts/deck-options/AdvancedOptions.svelte
+++ b/ts/deck-options/AdvancedOptions.svelte
@@ -62,7 +62,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
             <SpinBoxFloatRow
                 bind:value={$config.intervalMultiplier}
                 defaultValue={defaults.intervalMultiplier}
-                min={0.5}
+                min={0.1}
                 max={2}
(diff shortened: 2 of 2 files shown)


dae:
I'm afraid such lower multipliers do not make sense - answering Good would decrease the interval instead of increasing it. If you find your retention is very low, instead of trying to brute force it, you'll probably benefit more from simplifying your cards, creating mind-maps/mnemonics, etc.

dinoboy197:
Thanks for the note. Considering that the default value for the interval modifier is 1.0, and the minimum value is 0.5, is there something special about 0.5 that makes that the minimum? I'm trying to understand if there is a mathematical significance that I'm missing in the log calculation, or if that number was just chosen because it was half way in between 0 and 1, or maybe something else?

dae:
With 0.5x, the default starting ease of 2.5x becomes 1.3x, which is only marginally above the hard button of 1.2x, and after failing the card or marking it as hard a few times, the multiplier would creep down to 1.0x and only a day's progress would be made at a time.

PR #2115, opened 2022-10-08
Add min-width to QPushButton stylesheet

kleinerpirat:
[Quote](https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/5?u=kleinerpirat) from forum user jcznk:
> Regarding the design, I like most changes, but I I think some buttons are a little bit too small. To make a comparison:
>
>image

## Before
image
## After
image

I couldn't set up a dev build on Windows so far (hard to debug under bombardment of forced updates and loud fan noises :), but I'm confident this works regardless.

### DIFF
--- a/qt/aqt/stylesheets.py
+++ b/qt/aqt/stylesheets.py
@@ -63,6 +63,9 @@ QToolTip {{
 
 def button_styles(tm: ThemeManager) -> str:
     return f"""
+QPushButton {{
+    min-width: 75px;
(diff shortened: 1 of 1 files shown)


dae:
This is a nice improvement, thanks. Could we do the same for the webview buttons? The horizontal padding on the bottom buttons of the deck list is quite a bit smaller than screens like the Add screen now.

PR #2117, opened 2022-10-09
Experiment with labelled note view switch

kleinerpirat:
This [comment](https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/30?u=kleinerpirat) by @RumovZ made me revisit the switch:
> I would make the knob smaller or the path larger, like the toggles in deck options. Curved hair lines can’t be rendered evenly.

I'm struggling to see the issue here. The line looks even to me and I think it's more elegant when the knob fits the path with minimal margin (like a well-fitted clockwork). But I'm happy to discuss this here.

## The idea of this PR
Initial letters don't convey the message to first-time users that well, so I thought a full text label would be preferable.

## Light
image image
image image

## Dark
image image
image image

The longer of the two words sets the path length, so the switch won't change width on toggle.

I'm a bit torn because on the one hand, the status quo is smaller and drags less attention (better for advanced users), but on the other hand a full text label might make it easier to provide help to inexperienced users. WDYT?

### DIFF
--- a/ftl/core/browsing.ftl
+++ b/ftl/core/browsing.ftl
@@ -12,6 +12,7 @@ browsing-browser-appearance = Browser Appearance
 browsing-browser-options = Browser Options
 browsing-buried = Buried
 browsing-card = Card
+browsing-cards = Cards
 # Exactly one character representing 'Cards'; should differ from browsing-note-initial.
 browsing-card-initial = C
 browsing-card-list = Card List
@@ -59,6 +60,7 @@ browsing-new-note-type = New note type:
 browsing-no-flag = No Flag
 browsing-no-selection = No cards or notes selected.
 browsing-note = Note
+browsing-notes = Notes
 # Exactly one character representing 'Notes'; should differ from browsing-card-initial.
 browsing-note-initial = N
 browsing-optional-filter = Optional filter:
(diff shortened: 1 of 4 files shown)


RumovZ:
Of course, it's subjective to a certain extent, but doesn't the margin in your first dark mode screenshot for example look smudgy? If you zoom in you can see the grey-green pattern caused by anti-aliasing.
A quick image search seems to confirm that such fine margins are rather unusual in toggle widgets.

I like the written out labels. No text at all would look best, but under the assumptions that some verbal indicator is necessary, full words are superior, as they don't look any worse than initials in my opinion.

glutanimate:
:+1: on going with written out labels. We have enough space in the toolbar and IMHO expressive is better than implicit, especially for concepts like notes vs cards that new users tend to struggle with.

dae:
I think using a full word to describe things is a decent idea, but putting the text on green/blue leads to a not-so-great contrast ratio in both light and dark mode. Would it make sense to ditch the colors in favour of a neutral grey, or move the text outside the slider (eg making the slider very short, and putting the text underneath it)?

Looking at the longest 'cards/note' text in the translations, it's Tikaṛḍiwin/Jegyzettípus. Will that leave enough space for input?

kleinerpirat:
> putting the text on green/blue leads to a not-so-great contrast ratio

I thought low contrast would be preferable here, because people who want to read it can still relatively easily do so, while others who already know the functionality experience less visual noise.

> Tikaṛḍiwin/Jegyzettípus. Will that leave enough space for input?

## Horizontal aspect ratio
image
image

## Vertical aspect ratio
image

dae:
Looks like that's not too much space, and happy to defer to the community on feedback about contrast. Thanks Matthias :-)

PR #2119, opened 2022-10-09
Darken border colors in dark theme

kleinerpirat:
Point 3 of https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/30
## Before
image
## After
image

I made the default border one shade darker here. @dae, you once suggested to make borders brighter than the background. Personally, I'd like to keep borders darker than the background. WDYT about this PR?

### DIFF
--- a/sass/_vars.scss
+++ b/sass/_vars.scss
@@ -67,11 +67,11 @@ $vars: (
         border: (
             default: (
                 light: palette(lightgray, 6),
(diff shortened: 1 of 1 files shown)


dae:
I usually run my monitor at about 30% brightness, and the black-on-grey is pretty difficult to see for me. I'm no UI expert though, so I'm happy to run with this and see what the community thinks.

PR #2120, opened 2022-10-09
Fix QToolTip styling issue on Windows

kleinerpirat:
Point 4 of https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/30

image

## PR
image

@RumovZ does this fix the issue you mentioned on the forum?


dae:
On Windows with this PR:

kleinerpirat:
Ouch. Thanks for testing, Damien. Solving the border-radius issue with stylesheets only isn't possible, so I'd say the best move is to just get rid of the custom style entirely.

PR #2122, opened 2022-10-09
Add stylesheet for QMenu

kleinerpirat:
Following forum user MIZMU's suggestion: https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/23

image

## Changes
- `cursor: pointer` for QMenu
- border-radius applied to items and checkboxes
- improved visibility of checkboxes and highlights
- increased QMenuBar height for touch accessibility

---
MIZMU, if you're reading this: Sadly it [isn't possible to apply a border-radius to the QMenu via stylesheet](https://www.pythonfixing.com/2021/11/fixed-rounded-corners-for-qmenu-in-pyqt.html), as it is a top-level element. It would look like this:

image

### DIFF
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -345,27 +345,19 @@ class AnkiApp(QApplication):
     ##################################################
 
     def eventFilter(self, src: Any, evt: QEvent) -> bool:
+        pointer_classes = (
+            QPushButton,
+            QCheckBox,
+            QRadioButton,
+            QMenu,
+            # classes with PyQt5 compatibility proxy
+            without_qt5_compat_wrapper(QToolButton),
+            without_qt5_compat_wrapper(QTabBar),
+        )
         if evt.type() in [QEvent.Type.Enter, QEvent.Type.HoverEnter]:
-            if (
-                (
-                    isinstance(
-                        src,
-                        (
-                            QPushButton,
-                            QCheckBox,
-                            QRadioButton,
-                            QMenu,
-                            # classes with PyQt5 compatibility proxy
-                            without_qt5_compat_wrapper(QToolButton),
-                            without_qt5_compat_wrapper(QTabBar),
-                        ),
-                    )
-                )
-                and src.isEnabled()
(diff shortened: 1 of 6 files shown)


dae:
I'm afraid I'm not sure about this one:

- It increases the height of the menu bar, the size of the font, and the spacing between menu items, making the menu bar more prominent, thus reducing emphasis on the top toolbar
2022-10-10T14:29:56,628501914+10:00
- the empty checkbox, while consistent with what's shown in dialogs, looks a bit odd to me
- I use SwayWM on Linux, and this change causes the menus to bug out. If I click on Help, and move the mouse down to Guide for example, it selects the Tools menu and highlights the first item instead

Do other apps on your system apply styling to the top menubar that provide a precedent?

dae:
Looks better, but the SwayWM issue still seems to exist unfortunately.

dae:
Sorry, false alarm - this appears to have been a Qt 6.3 issue which was fixed by updating to 6.4.

kleinerpirat:
> making the menu bar more prominent, thus reducing emphasis on the top toolbar

Yep. People would likely complain, also because our fullscreen mode continues to show the menu bar. Removing the bar entirely one day would be fine by me as well, so no hard feelings.

image

Do the checkboxes still look off to you? This is how it looked on my DE before the PR:

image

Judging from your reaction I assume Sway doesn't show them at all by default. That would be fine too, but a user complained about the low visibility of the checkboxes on the beta thread...

> I use SwayWM on Linux, and this change causes the menus to bug out.

That's interesting. Did you run Wayland? On Wayland, the menus have been buggy from the very beginning on my DE - similar to your description. This seems to have been solved though, now that II'm testing it again. QToolTip remains buggy on Wayland though.

dae:
Checkboxes are better. Not 100% sold on them and the highlight of the File/Edit/etc menu, but let's see what the community thinks. One minor issue: when hovering over a menu like File, the text shifts down by about a pixel.

dae:
Thanks Matthias!

PR #2125, opened 2022-10-11
Fix daily limits spinboxes showing 1 instead of passed value

kleinerpirat:
- https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/10
- https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/19

For some reason, only the instances for Daily Limits were affected - probably because they use `TabbedValue`. Sorry for not fixing this showstopper earlier.


dae:
No problem - thank you for this!

dae:
I've just noticed a (fairly minor) issue with this change - zero values (eg 'new interval' at the bottom) appear as blank instead of 0.00 like before.

PR #2129, opened 2022-10-12
Make resizer snap to closest end and improve toggle

kleinerpirat:
image
## Changes
- resizer snaps to closer end on pointer release
- resizer adjusts on window resize event
- there is no more stage in the resizing process where the tag icon is hidden
- tag icon is now absolutely positioned because it messed with resizing calculations
- tag icon is animated and tag editor fades when it reaches snapping size

## Fixes
- https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/20
- https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/37

### DIFF
--- a/ts/components/HorizontalResizer.svelte
+++ b/ts/components/HorizontalResizer.svelte
@@ -3,6 +3,9 @@ Copyright: Ankitects Pty Ltd and contributors
 License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 -->
 <script lang="ts">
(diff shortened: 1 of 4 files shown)


dae:
Nice, this feels lovely and smooth.

Cursor issues may be API misuse? In the console:

> Pointer is already locked.

and before this PR:

> WrongDocumentError: The root document of this element is not valid for pointer lock.

Perhaps related: https://stackoverflow.com/questions/36903442/js-pointer-lock

PR #2130, opened 2022-10-12
Add background to tag editor and de-uglify autocomplete items

kleinerpirat:
## Motivation
The button gradient looked awful on the Autocomplete items and the font size was pretty small.

## Changes
- tag editor has `canvas-inset` background color and a border radius
- autocomplete popup can take up to `30vh` in height now
- autocomplete items do not have a gradient except on hover and select
- font size increased from 11 to 13
- increased padding

image


dae:
Not 100% sure about the different background, but let's see what the community thinks. Thanks Matthias.

PR #2132, opened 2022-10-13
Fix maximized browser window not being restored

hikaru-y:
Fixes the first issue of https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/51


dae:
Thanks as always Hikaru!

Issue #2134, opened 2022-10-15
Rust Analyzer broken due to missing protoc


RumovZ:
Since the last prost update Rust Analyzer isn't working for me anymore. I assume it's because of this ([prost](https://github.com/tokio-rs/prost)):

> With prost-build v0.11 release, protoc will be required to invoke compile_protos (unless skip_protoc is enabled). Prost will no longer provide bundled a protoc or attempt to compile protoc for users. For install instructions for protoc please check out the [protobuf install](https://github.com/protocolbuffers/protobuf#protocol-compiler-installation) instructions.

I fixed it for me by downloading the protoc binary, but maybe this can be fixed more elegantly by Anki's build scripts? I wouldn't know.
But if not, I would drop a note in development.md.

dae:
I didn't test the fix on Windows; please send a follow-up if it doesn't work.

PR #2135, opened 2022-10-17
Fix protoc for Rust Analyzer on Windows

RumovZ:
Thanks for the fix. Didn't work on Windows, but I could take it from there.
#2134


dae:
Thanks Rumo :-)

Issue #2133, opened 2022-10-14
Various color issues in dark mode after Qt 6.4.0 update (Manjaro Linux)


kleinerpirat:
Related to #2124.
The screenshots are from a build **without custom Qt stylesheets.**

---

- [x] progress bar is black:

![image](https://user-images.githubusercontent.com/62722460/195803873-917e5243-5c31-4718-9a20-c5a809b5c4f0.png)

- [x] highlight background for items of QListView, QTableView, QTreeView, QComboBox, QMenu

![image](https://user-images.githubusercontent.com/62722460/195804207-c45d0d2c-08cc-4fbb-8333-dc2e3a5ffd0f.png)
![image](https://user-images.githubusercontent.com/62722460/195806039-69055336-de8b-42f9-8efe-47043c9e35d8.png)
![image](https://user-images.githubusercontent.com/62722460/195804823-b40c8fbe-14e3-4fce-bb23-96b868e20f05.png)

The [ColorRole](https://doc.qt.io/qt-6/qpalette.html#ColorRole-enum) enum of QPalette doesn't contain entries for these elements.

Again, setting these explicitly via stylesheet would fix this, but it does come with caveats. For example, it's not possible to animate things in Qt stylesheets, so native animations for QProgressBar would be lost.

dae:
Have only had a chance to test on a Mac so far, but if I comment out this line:

```
        palette.setColor(QPalette.ColorRole.Highlight, hlbg)
```

Then the highlight changes from black to blue.

lennart02:
I don't know if it belongs to this issue, but if you have the Dark Mode enabled, a card with code section (<pre></pre>) also appears stil black.

kleinerpirat:
Thanks @lennart02, yeah it's the right place :)

`QColor` doesn't accept rgba strings in CSS format, e.g. `rgba(12, 34, 56, 0.5)`. Since I introduced rgba values in the Sass maps a few PRs ago, these colors don't work for Qt elements anymore. I'm working on the fix for this.

kleinerpirat:
Closed with #2137.

Issue #2124, opened 2022-10-10
Qt 6.4.0 update broke light theme on Manjaro Linux


kleinerpirat:
After 276c42e25057e6064c05ecc2556279aa7850f79e Anki defaults to night-mode colors ~~for widgets affected by stylesheets~~, resulting in illegible text and other atrocities:

![image](https://user-images.githubusercontent.com/62722460/194854090-1284b9ca-014f-4264-8b69-36b1900f3672.png)
(Screenshot from b4293573ddee1f9d4e9dd29200b3afe50b365efb)

Defining the faulty properties explicitly in the stylesheets should solve the issue. But I wonder why these defaults don't seem to work anymore:

https://github.com/ankitects/anki/blob/3d47c9547a1d18ca54d1858cff7712de8a3a67af/qt/aqt/theme.py#L276-L280

kleinerpirat:
Actually, it isn't even related to stylesheets. The issue persists without them:

![image](https://user-images.githubusercontent.com/62722460/194856089-27bb4ad9-69de-4c67-9e4e-46f2a02bbb03.png)

dae:
Light mode looks fine on my system - is your desktop environment set up to use a dark Qt theme?

kleinerpirat:
Yes, if you mean that with "Follow System" it switches to dark mode. I'm busy tonight, but I'll investigate tomorrow.

dae:
The light theme doesn't explicitly set the background IIRC, so that Anki matches the default background color on different platforms/themes. The downside is that it only works when the default background is light - if dark, you get illegible text.

kleinerpirat:
Closed with #2137.

PR #2137, opened 2022-10-20
Add comments to Sass variables and tweak main window

kleinerpirat:
The main purpose of this PR is to make descriptive comments available to theme authors.

E.g.:

```sass
  colors: (
      fg: (
          ...
          subtle: (
              "Placeholder text, icons in idle state",
              (
                  light: palette(darkgray, 6),
                  dark: palette(lightgray, 3),
              ),
          ),
          ...
    ),
),
```
Results in:
image

Add-on authors will also be able to access these descriptions with `color.comment`, as the colors are now dictionaries instead of tuples in Python. I imagine it would be practical for theme-editing GUIs.

## Other improvements
- Previously the whole set of variables was included for each Svelte component that imported them.
I removed those imports and added them to the base stylesheet of the webview.
- The main window got some visual improvements to bridge the gap to the real main window rework.


dae:
Are you aware that you can run the tests locally? I have it set up so the tests get run when I try to push, so I never accidentally push broken changes:

```
[dae] code/desktop/anki [main]  % cat .git/hooks/pre-push
#!/bin/zsh -l

set -e

# ensure path matches login shell
source ~/.zshrc

if [ "$SKIP_CI" != "1" ]; then
    unset GIT_DIR
    bazel test ...
fi
```

kleinerpirat:
I thought so, but was too lazy to set it up so far :sweat_smile:

The formatter complains about the auto-generated files, but I'm not sure I can achieve what it wants with `pprint.pformat`. That module doesn't seem to care about the opening/closing bracket style.

dae:
You can exclude certain files in [pylib|qt]/tools/run_format.py

kleinerpirat:
I'll try testing locally for the next PR.

qt/aqt/webview.py
```diff
@@ -448,15 +448,31 @@ def standard_css(self) -> str:
...
+:root[class*=night-mode] {{
+  {br.join([f"/* {color['comment']} */{br}--{name.lower().replace('_', '-')}: {color['dark']};" for name, color in colors.__dict__.items() if name != "__builtins__" and isinstance(color, dict)])}
+}}
+:root {{
+  {br.join([f"/* {prop['comment']} */{br}--{name.lower().replace('_', '-')}: {prop['light']};" for name, prop in props.__dict__.items() if name != "__builtins__" and isinstance(prop, dict)])}
+}}
+:root[class*=night-mode] {{
+  {br.join([f"/* {prop['comment']} */{br}--{name.lower().replace('_', '-')}: {prop['dark']};" for name, prop in props.__dict__.items() if name != "__builtins__" and isinstance(prop, dict)])}
+}}
+        """
```
kleinerpirat, line 474: Yesterday I had this shower-thought that we could go full-circle with the Sass variables and actually add them to our webviews from the Python side.

At first it sounds kind of dumb, but having a single source of truth will make it much easier for us and add-on authors to apply new themes.

kleinerpirat:
Closes #2133 and #2124. For some reason the check for light theme returns a different result since the recent Qt update. There was an if block with a return that prevented the application of our custom palette in `theme.py`. I removed that block entirely.

qt/aqt/webview.py
dae: Agreed that a single source of truth is good, but it's best if that is the Rust or TypeScript, which is shared by all clients. Any complexity we push into the Python code means that the two mobile clients need up having to reimplement it, which wastes developer resources.  Could these comments not be included via our JS/sass tooling instead?

Likewise for the rgba solution - you've added extra parsing code into Python that the mobile clients will now need to mirror; couldn't we instead limit ourselves to hex colors of the form #aabbcc or #aabbccdd in our color definitions, to keep the parsing on the clients simpler?

dae:
Regarding the styling changes in the main window:

- The rounding feels a little over the top, and reminds me of old-school macOS: http://macgateway.com/featured-articles/a-decade-of-mac-os-x-a-retrospective/
- Not sure about putting the card counts to the left of the answer button; it feels a bit unbalanced.
- The answer buttons are clipping the bottom of the window/showing a scroll bar.
- How do the changes to the top toolbar affect add-ons that add new entries to the top, like AMBOSS?

The new look seems similar to the "anki redesign" add-on, which I gather some people much prefer, so maybe my opinions on style don't reflect the average user :-) But I am a bit concerned about making too many more UI changes at this point in the .55 cycle - it would be nice if we could start focusing on getting .55 ready for release, and pushing larger changes back until we can get that out.

kleinerpirat:
> The rounding feels a little over the top
> ...
> The new look seems similar to the "anki redesign" add-on

That was intentional as I also thought the majority would prefer this due to the add-on's popularity, but of course that's just a gut feeling and we should probably ask the community, how much of that look should be integrated in Anki. The slight gradient + drop shadow might be responsible for the old school mac feeling more than the radius itself.

In the future I want to allow setting a custom wallpaper. For that reason I thought giving the toolbar some substance would be important. When a wallpaper is set, I'd give the UI an (optional) [glassmorphism](https://css.glass/) look. This can look really cool, I've used in in my templates in the past:

image

> Not sure about putting the card counts to the left of the answer button; it feels a bit unbalanced.

Which screen you're referring to here? Do you mean the "Study now" button in the overview?

> How do the changes to the top toolbar affect add-ons that add new entries to the top, like AMBOSS?

They should not be affected at all, since these are just CSS changes on the center element.

---
> it would be nice if we could start focusing on getting .55 ready for release

Note taken. I'll create bugfix PRs from here on out.

dae:
Background blur does look nice; hopefully it performs reasonably on people with bad graphics drivers (or we provide a way to turn it off). 

Re the card counts:
Screen Shot 2022-10-24 at 7 35 36 pm
Screen Shot 2022-10-24 at 7 35 31 pm

dae:
Here's a before/after on macOS:

- This is changing the buttons from the native macOS ones to non-native ones, which look somewhat out of place compared to other apps.
- The switch also breaks changing from light->dark mode - dropdowns end up with white text on white background.
- The Edit/More and Show Answer buttons are no longer vertically aligned
- The extra padding on the bottoms leads to truncation at certain widths
Screen Shot 2022-10-25 at 4 25 31 pm
Screen Shot 2022-10-25 at 4 26 01 pm
Screen Shot 2022-10-25 at 4 31 30 pm

kleinerpirat:
> This is changing the buttons from the native macOS ones to non-native ones, which look somewhat out of place compared to other apps.

The reason for the box-shadow was a change of the button gradient to a darker tone in dark mode. I thought the whole UI would look better with a slightly less striking gradient. Drawback of the darker shade was contrast to the window, so I added the elevation effect.


> The switch also breaks changing from light->dark mode - dropdowns end up with white text on white background.

I didn't think of mac when I fixed the color issues. Looking at the screenshot of the preferences screen, I wonder why we handle macOS separately in the first place. It looks exactly like the Qt GUI of Win/Lin that I patched with the stylesheets.

I'd like to ask the community whether they want to have Anki buttons looking natively or if they're ok with a custom, more modern look. I believe most people are used to web apps having their own style nowadays.

---
I'll take a look at this PR as soon as possible (sadly got uni the whole day, so you'll see the fix tomorrow).

dae:
This is what the preferences screen looks like on main:
Screen Shot 2022-10-25 at 5 02 16 pm

dae:
(and no problem - your studies should always take precedence!)

kleinerpirat:
Ah, I see. So I broke the native look - got it :)

kleinerpirat:
Does this fix the issues on mac? Please also check dark mode, as this seems to have been handled separately in the past.

dae:
Thanks Matthias:

- Light/dark switching fixed 👍 
- macOS Qt appearance fixed 👍 
- Button vertical alignment fixed 👍 

Only remaining issue I can see is that the webview buttons stand out a bit. They didn't match the native appearance before either, but the larger size and drop shadow has made it stand out a bit more:
 
Screen Shot 2022-10-26 at 11 42 32 am
Screen Shot 2022-10-26 at 11 39 43 am

WDYT?

kleinerpirat:
I decreased the elevation effect for the main view buttons (but kept it for the toolbar at the top).

Personally I'm a fan of box-shadows. It never ceases to amaze me how some computer screen pixels can create the illusion of real, plastic elements with such a simple method.

image
image

The slight shadow at the bottom makes the buttons look a lot more real to me. It does however come at the cost of increased visual noise - and some will consider the lower screenshot cleaner.

Maybe we should merge it and get some feedback in the next beta. Then I can still change it back.

kleinerpirat:
I can emulate the effect with Qt stylesheets quite convincingly with `qlineargradient`:

image

dae:
It may be my imagination, but I feel like the dark theme buttons look a bit nicer than the light theme ones. Could it be related to the gradient?

On the light theme, the Qt buttons feel like they're slightly truncated at the bottom - maybe the text is just a pixel or two too low?

>  It never ceases to amaze me how some computer screen pixels can create the illusion of real, plastic elements with such a simple method.

For better or for worse, iOS 7 marked the start of a move across the industry towards flat UI, so I fear that prominent 3D UI may been seen by some as "old school" at this point. But let's see what users think :-)

2022-10-27T08:44:59,853459384+10:00
2022-10-27T08:46:17,557814921+10:00

qt/aqt/webview.py
dae: This is the only thing I'd like clarified before merging. It looks like the comments are included in .bazel/bin/sass/_root_vars.css. Is this Python code serving some other purpose? And if so, can we not accomplish this in our scss->css transform step instead?

kleinerpirat:
You were right, the gradient was too shallow for light theme. Imitating the shadow with the background gradient alone also made the edges look different than the web counterparts, so I used the border-color property for the last pixel. I think now it looks better.

I see where you're coming from with the old-school look, but to me there's a difference between a semi-flat look with subtle box-shadow versus old skeuomorphic macOS buttons with much more striking gradients - which I also find old-fashioned.

dae:
Did you see my reply on https://github.com/ankitects/anki/pull/2137#discussion_r1006259607 ?

qt/aqt/webview.py
kleinerpirat: > Could these comments not be included via our JS/sass tooling instead?

Yes, I've already done that before I went for the Python route. The only reason for doing this is future theme support. If our whole UI consisted of web components, there'd be no need for this step - but we also need to modify Qt widgets via Python with the same variables.

kleinerpirat: If there's a way to do this via Rust/Typescript, I'm all ears.

dae: I fear I'm not yet understanding the problem/goal here.  Is your thinking here that the theme author will mutate the values in colors.py, and then we'll automatically regenerate vars.css based on those? Why can't we go in the opposite direction instead, where the theme author provides an alternative vars.css which we then extract the colors from, by including/copying extract_sass_vars.py?

dae: (if we want to make it easy to tweak only a few values in the scss file, another option would be to make the source of truth a proto message with defaults provided by the backend, and provide a backend method to generate vars.css from a provided collection of colors)

kleinerpirat: > Is your thinking here that the theme author will mutate the values in colors.py, and then we'll automatically regenerate vars.css based on those?

I thought we'd apply theme changes via `ThemeManager` and `ProfileManager`, so I'd have added a colors dictionary in `ThemeManager`, which would then be modified on theme change.

> Why can't we go in the opposite direction instead, where the theme author provides an alternative vars.css which we then extract the colors from, by including/copying extract_sass_vars.py?

That was my plan for our own theme support. Users would provide a CSS or JSON file from which we extract the colors. Add-on authors (e.g. ReColor, Anki Redesign) on the other hand might want to control things via Python instead - at least that's what I was thinking.

>  make the source of truth a proto message with defaults provided by the backend

That sounds a bit cumbersome for us developers. We'd lose all Sass features such as `color.scale`or `.adjust`. I also wanted to auto-generate a `gradient-middle` variable from the `gradient-start` and `gradient-end` values via Sass functions in the future.

I'm not familiar enough with the backend to see this problem from your perspective, yet.d

kleinerpirat: I've reverted the changes, because I don't want to hold back this update any longer than necessary. We should continue this discussion in a proper theming PR in the future :)

dae: Agreed on continuing this later :-)

dae:
I've pushed a change that improves the appearance on macOS for the main window, but the changes I made aren't applied into the editor, presumably because of the more specific styling inside the svelte component. Can you think of a clean/simple way to work around it?

dae:
Before the latest commit:

Screen Shot 2022-10-28 at 9 33 38 pm

After the latest commit:

Screen Shot 2022-10-28 at 9 34 01 pm

But editor still looks like this:

Screen Shot 2022-10-28 at 9 34 17 pm

Also, is the congrats screen supposed to be lower now?

Screen Shot 2022-10-28 at 9 39 17 pm

kleinerpirat:
> I've pushed a change that improves the appearance on macOS for the main window, but the changes I made aren't applied into the editor, presumably because of the more specific styling inside the svelte component.

Yes, if you mean the editor toolbar buttons, these got specific styling. As they serve a quite different role than the buttons in the main view, I think it's fair they got their own stylesheet. They need much less padding and margins etc.

> Also, is the congrats screen supposed to be lower now?

Yes, that was deliberate. It was very close to the top and made the screen very unbalanced. But it's probably too low right now.

kleinerpirat:
> Before the latest commit:
>
> image

Did these buttons look like that throughout my PR? On my machine it looks like this:

image

I guess you got a bigger resolution than 1920x1080 on your mac? Perhaps radii shouldn't be defined in `px` then.

dae:
> I've said this a couple of times before, but I find OS-specific handling rather annoying. Not only does it complicate the design process, it also creates a divide between the user bases.

I think of all the platforms, Anki on macOS looks the best, and I think you've said similar in the past too. While I certainly understand the appeal of a unified theme, until our custom styling feels as polished and consistent as the stock macOS styling, I fear switching over to it on macOS will be a step in the wrong direction, and users may not appreciate it.

The editor buttons are not a blocker; we had inconsistent styling with them on macOS in the past as well, so I'll merge this in now so we can get a new beta into people's hands and some feedback.

> Did these buttons look like that throughout my PR?

Since the "improve appearance of main window" commit.

I use 4k and 5k monitors with the Mac, and the same 4k monitor with the Linux machine.

> Currently on holiday

Hope you have a good time :-)

dae:
The appearance does appear to change depending on resolution - I use a Windows VM with 100% DPI on the 4k monitor so everything looks small, and to me the styling looks better there than under Wayland, where the drop shadows feel too harsh:
2022-10-29T10:58:25,811250140+10:00

dae:
Sorry if the new beta came as a surprise; I'd been waiting for this PR to be merged before releasing, and figured the feedback could accumulate over the weekend. I hope it did not disrupt your holiday.

kleinerpirat:
Holiday was great, no worries! :)

The feedback was mostly new information and very helpful. Fixing everything will take a few more days, but should be possible till thursday evening.

dae:
No rush; I really appreciate you looking into these issues.

PR #2138, opened 2022-10-22
Center add-on icons and allow text buttons to expand in width

kleinerpirat:
On some forum screenshots, I saw that the icons of add-on buttons were too big and not centered properly. This PR aligns them better with the native buttons and also fixes text buttons not expanding in width.

### Before
image

### After
image


dae:
Thanks as always Matthias!

PR #2139, opened 2022-10-23
Redesign deck options screen, swap tooltips for help modals

kleinerpirat:
In the forum I [proposed a redesign](https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/57?u=kleinerpirat), which I now implemented.

image

# Changes
- redesign `TabbedValue` component
- add elevation effect to containers
- make `RevertButton` more subtle
- improve layout for vertical aspect ratio
- replace all tooltips with wiki-style modals

## Modals

image

 There's one modal for each config section. It can either be opened via the info badges next to the section header, or by clicking a setting title. Doing so will open the specific page of the help modal. All relevant manual pages have been linked. You can open the main chapter by clicking on the modal header, or use the link at the footer, if a URL is provided for that section.

I created a new `.ftl` file called `help`, because I think such modals will be useful in other screens too.

## Known issue

- [x] the tabbed spinboxes do not reflect changes yet

---
Overall, I think the deck options are much cleaner now and interaction with the former tooltip text is encouraged a bit more.


dae:
Various thoughts:

- I like the styling of the tab widget.
- Moving the individual info buttons to a single one seems like a reasonable compromise. The user has to hunt a bit more for the description of the item they're after, but it makes the UI cleaner. The links to sections of the manual are a nice bonus.
- Maybe personal opinion, but I find the rounding and drop shadow of each section to be a bit strong. Could it be made a bit more subtle?
- The margins are too large on small mobile devices, leading to wasted screen space. This applies to both the main screen, and the help boxes.

dae:
IMG_1604
IMG_1603

ftl/core/deck-config.ftl
```diff
@@ -202,6 +206,9 @@ deck-config-show-answer-timer-tooltip =
 
 deck-config-audio-title = Audio
 deck-config-disable-autoplay = Don't play audio automatically
+deck-config-disable-autoplay-tooltip =
+    If you check this setting, Anki will not play audio
```
dae, line 210:
```suggestion
   When enabled, Anki will not play audio
```

ftl/core/deck-config.ftl
```diff
@@ -202,6 +206,9 @@ deck-config-show-answer-timer-tooltip =
...
+    until you press the replay audio key, r or F5.
```
dae, line 211:
```suggestion
    automatically. It can be played manually by clicking/tapping on an audio icon, or by using the replay audio action.
```

(we need to avoid desktop-specific terminology)

ftl/core/help.ftl
```diff
@@ -0,0 +1,20 @@
...
+# Newly introduced settings may not have an explanation yet
+help-no-explanation =
+    Whoops! There doesn't seem to be an explanation for this setting yet.
+    
+    You can help us complete this help page on { $link }.
+
+## Footer
+
+# Link to more detailed information in the manual
+help-for-more-info = For more information, see { $link } in the manual.
```
dae, line 20: Can we cut out the other two if they're not used?

dae:
Minor: the highlight on hover over the (i) icon appears laggy, presumably because it fades in.

ftl/core/help.ftl
kleinerpirat: I'm sorry, but I'm not sure what you mean with "other two". The only string that isn't used currently is the one that appears, when no `help` string is provided for a modal item. Technically all of them are used in the code.

kleinerpirat:
> The user has to hunt a bit more for the description of the item they're after

Have you seen that you can jump directly to the respective section with a click on the setting label?

ftl/core/help.ftl
dae: The first two don't seem to be used - what am I missing?

```
[dae] code/desktop/anki [main]  % gh pr checkout 2139
remote: Enumerating objects: 165, done.
remote: Counting objects: 100% (140/140), done.
remote: Compressing objects: 100% (49/49), done.
remote: Total 111 (delta 89), reused 77 (delta 62), pack-reused 0
Receiving objects: 100% (111/111), 21.91 KiB | 3.65 MiB/s, done.
Resolving deltas: 100% (89/89), completed with 20 local objects.
From github.com:ankitects/anki
   c7b51ef2a..9ae039b2b  refs/pull/2139/head -> deck-options-rework
Switched to branch 'deck-options-rework'
[dae] code/desktop/anki [deck-options-rework]  % rg openManual
zsh: exit 1     rg openManual
[dae] code/desktop/anki [deck-options-rework]  % rg open_manual
zsh: exit 1     rg open_manual
[dae] code/desktop/anki [deck-options-rework]  % rg noExplan
zsh: exit 1     rg noExplan
[dae] code/desktop/anki [deck-options-rework]  % rg no_explan
zsh: exit 1     rg no_explan
```

dae:
Yeah, I just meant that you have to locate the correct label on the left again, instead of being able to immediately click on the (i). Not a big deal.

kleinerpirat:
> Yeah, I just meant that you have to locate the correct label on the left again

No. Take a look:

image
> It can either be opened via the info badges next to the section header, or by clicking a setting title.

ftl/core/help.ftl
kleinerpirat: https://github.com/kleinerpirat/anki/blob/9ae039b2b7a4178c00b16b355a04d81a6d509a4b/ts/deck-options/HelpModal.svelte#L72

image


---
The other one: https://github.com/kleinerpirat/anki/blob/9ae039b2b7a4178c00b16b355a04d81a6d509a4b/ts/deck-options/HelpSection.svelte#L32

dae: 🤦 Silly me on searching incorrectly.

I didn't notice the clickable headings before - they're not the most discoverable. What about having a "View in manual" link below the heading in smaller text instead of making the heading itself clickable? Alternatively we could just rely on the links in individual sections, and not have one to the chapter itself.

dae:
Ah, that's neat - sorry I missed it.

Apart from the heading comment above needing clarification, this is all looking good.

dae:
Having tried out the latest code, I think my "view this chapter in manual" suggestion was a bad idea - the page feels a bit cluttered now. If you agree, I guess that leaves us with either reverting to the previous difficult-to-discover behaviour, or removing the top link and relying on a link in individual sections. What do you think?

Won't block merging on this; the rest looks good. Thanks for your work on it. :-)

kleinerpirat:
I think it won't hurt to have the main chapter link, even if it isn't that discoverable. I'll gladly revert to the previous behavior in a follow-up PR. Maybe we could also use a little manual icon next to the title?

dae:
We could give it a go - easier to decide after seeing it :-)

PR #2140, opened 2022-10-23
Wrap inactive clozes in CSS class

RumovZ:
https://forums.ankiweb.net/t/different-styles-for-clozes-that-are-not-asked/23553/1

I think it makes sense to preserve that information.


dae:
This seems reasonable, but will create conflicts with #2141. Since more work has gone into that, perhaps we're better off focusing on that one?

PR #2143, opened 2022-10-25
Use book icon to link manual chapters

kleinerpirat:
Follow-up to #2139.

image


dae:
Icon was a good idea - thanks! Any objection to me changing "Open chapter" to "Open" in the .ftl?

kleinerpirat:
Thanks. No objection!

PR #2144, opened 2022-10-26
Finish #2070: Single overlay instead of per field 

### LINKED ISSUE #2089: Field name appears in the middle of the field
This occur when there is many fields and so I need to scroll

2022-09-25-200247_1235x1040_scrot

kleinerpirat:
Makes Henrik's work of #2070 mergeable.
Closes #2089.

## My changes
- fix double-border issue
- tweak appearance of focused fields: box-shadow doesn't change color, only border
- fix field labels overlapping with field content

---
@hgiesel miss you man. I hope everything is alright.


dae:
Thanks for looking into this Matthias. And agreed, hope you're well and in good spirits @hgiesel, and have just become busy.

abdnh:
This has broken the fields.

image

Maybe this a clue (though adjusting the height to a valid value doesn't seem to make a difference).
image

dae:
They work for me on both Qt 5+6. Ruled out add-ons?

kleinerpirat:
The `height = NaNpx` probably stems from my use of `display: none` to fix the double-border issue Henrik mentioned in his original PR. Does Qt use a different web engine on Windows?

abdnh:
> Ruled out add-ons?

Yes.

abdnh:
I found an actual clue. The issue only happens with the "Reduce motion" option enabled.

kleinerpirat:
Thanks for investigating, Abdo!
I'll look into this.

PR #2146, opened 2022-10-27
Fix some issues with tag reparenting

abdnh:
This fixes two issues with tag reparenting:

## First issue

If a source tag is a prefix of the target tag (e.g. `a` and `ab`), reparented_name() incorrectly returns None, resulting in a panic in reparent_tags_inner(). This can be reproduced easily from the sidebar.

<details>
<summary>Stack trace</summary>

```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', rslib/src\tags\reparent.rs:63:78
Caught exception:
  File "aqt.taskman", line 122, in _on_closures_pending
  File "aqt.taskman", line 71, in <lambda>
  File "aqt.taskman", line 90, in wrapped_done
  File "aqt.operations", line 120, in wrapped_done
  File "concurrent.futures._base", line 438, in result
  File "concurrent.futures._base", line 390, in __get_result
  File "concurrent.futures.thread", line 52, in run
  File "aqt.operations", line 105, in wrapped_op
  File "aqt.operations.tag", line 81, in <lambda>
  File "anki.tags", line 105, in reparent
  File "anki._backend.generated", line 1389, in reparent_tags
  File "anki._backend", line 140, in _run_command
pyo3_runtime.PanicException: called `Option::unwrap()` on a `None` value

thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }', rslib/src\backend\mod.rs:181:18
Error in atexit._run_exitfuncs:
Error in atexit._run_exitfuncs:
```

</details>

## Second issue

If a source tag is the parent of the target tag (e.g. `foo` and `foo::bar`), reparent_tags_inner() also panics at the same line it panics at in the first issue.

### DIFF
--- a/rslib/src/tags/reparent.rs
+++ b/rslib/src/tags/reparent.rs
@@ -38,7 +38,9 @@ impl Collection {
         let usn = self.usn()?;
         let mut matcher = TagMatcher::new(&join_tags(tags_to_reparent))?;
         let old_to_new_names = old_to_new_names(tags_to_reparent, new_parent);
-
+        if old_to_new_names.is_empty() {
+            return Ok(0);
+        }
         let matched_notes = self
             .storage
             .get_note_tags_by_predicate(|tags| matcher.is_match(tags))?;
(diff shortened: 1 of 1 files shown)


dae:
Thanks for these fixes Abdo!

PR #2148, opened 2022-10-28
Fix grandchild tag not being reparentable under the same root

abdnh:
#2146 broke the case where a grandchild is dropped into its grandparent (e.g. `a::b::c` into `a`). I only noticed this after I looked into the similar deck parenting code to see if it suffers from similar issues (it suffers from the first issue), and it was not covered by any test.


dae:
Thanks for your thorough work Abdo :-)

Issue #2150, opened 2022-10-29
[Question] Development process for fast iteration on changes to Anki's web stack


glutanimate:
@dae @kleinerpirat @hgiesel 

Not sure if here or the forums is the right place to post this, but I'm trying to gain a bit more insight into your development process when iterating on Anki's web stack as I would like to contribute more in that area and bazel is making things a bit opaque for me.

Say you perform a small change in `toolbar.scss` and want to see the results in practice, do you actually restart Anki via `bazel run`, or is there an easier way to compile the web stack (across both `ts` and `qt/aqt/data/web`) without having to launch a new Anki session? In other words is there a bazel command you typically run?

(Reloading a web view is easily done via the dev tools, so that's not the issue)

Similarly, for pages that have already been migrated to Svelte / aren't rendered at runtime in Python, do you actually develop inside of Anki, or do you use an outside browser session? Are there any launch scripts you have set up for this that would e.g. run Anki, setting `ANKI_API_PORT`, etc.?

I'm sure these are things that were already covered in other issues / PRs at some point, but they seem to be hard to dig up via GitHub's search. Generally speaking, I think it would be great if we worked on building up the development docs in this area a bit as I think this would help reduce the barriers of entry towards contributing to Anki's web stack (which is both becoming ever more important to Anki's future and likely also the best entry point for new contributors).

Would of course be happy to pick up part of that documentation work as I'm becoming more familiar with the stack myself (or more specifically, with working on it in a more efficient way than I've done so far).

glutanimate:
OK, to partially answer my own question, `bazel build  "//qt/aqt:aqt_with_data"` seems to be workable way to compile the web stack (alongside Qt). Is that what you use, or do you use a different target?

Would still very much be interested in your Anki web dev workflow in general, however :)

hikaru-y:
> `"//qt/aqt:aqt_with_data"`

I think `//qt/aqt/data/web` is probably adequate for that purpose, that's what I usually use.

glutanimate:
@hikaru-y that works perfectly, thanks! I could have sworn I gave this a try before, but I probably just messed up the bazel package path / specified a (wrong) target where there shouldn't have been one.

To be completely honest, I think I'll just need to spend more time playing around with bazel and reading through its docs (never dove too deeply into its mechanics as it wasn't necessary for past PRs). Already starting to understand it much better now as a result of this discussion.

dae:
[...]

PR #2151, opened 2022-10-30
Add dev tools for live-reloading Anki's web views

glutanimate:
Spawned off from #2150, this is a proof-of-concept PR that extends the existing `ts` monitoring tools with a few more tools that:

- monitor changes in the `ts`, `sass`, and `qt/aqt/data/web` folders
- trigger a rebuild of the web stack if changes are detected
- and then finally trigger a reload of the corresponding web views in Anki via the Chrome DevTools Protocol

In action:

https://user-images.githubusercontent.com/5459332/198858969-266e5b63-417b-4c4c-b054-ed10ccc81c44.mp4

---

Filing this as a draft for now as I wasn't sure about a few things:

- Is this helpful, generally?
- If so, should there be both a `web-watch` and `ts-watch` script, or should we combine them?
- What's the best way to add `PyChromeDevTools` as a dev dependency? (still struggling with Bazel's management of Python dev dependencies)
- Reloading some web views doesn't work as neatly as others (e.g. the editor and reviewer where triggering a reload results in a blank screen). So it might make sense add a blacklist of web views that shouldn't be reloaded programmatically.


dae:
In the long term I'd like most of our code to be testable outside of the Qt GUI, but that's not going to happen in the near future, so this looks like a nice improvement. Blacklisting certain screens could work, or adding extra logic for them to reload (even with lost state) could be an option too.

Adding a dev dependency would require changing requirements.in/txt, defining the script with py_binary(), then using bazel run to invoke it. The pip package could then be listed as a requirement of the script, and would get automatically pulled in on run. I would not bother with this at the moment though, as there may be changes to the build system after .55 is out.

glutanimate:
@dae I've updated the PR to work with the new build system and adjusted the dev dependencies. Also took the liberty to remove `ts-watch`. Hope that's OK. I saw that it wasn't updated to `ninja` yet and it made sense to me to have `web-watch` supersede it.

I haven't had time yet to look into a solution for reloading webviews that don't play as well as others, but I think the tools here already improve the DX quite a bit, so if you're OK with this, I would vote to merge them as is and follow up later (also in regards to documentation).

tools/rebuild-web
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# Copyright: Ankitects Pty Ltd and contributors
+# License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+
+# Manually trigger a rebuild and reload of Anki's web stack
+
+set -e
+
+PARENT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+ROOT_DIR=$( cd -- "$( dirname -- "${PARENT_DIR}" )" &> /dev/null && pwd )
```
dae, line 10: Could this (and the similar line in the other file) be done more elegantly with the following?

```
PARENT_DIR=$(dirname $0)/..
```

dae (commented):
Thanks for looking into this and fixing the watch script - looks like a nice improvement! Just a minor point on style.

tools/rebuild-web
glutanimate: It does look ugly, doesn't it? First time using this form actually, `dirlink` is my usual choice, typically combined with `readlink` to avoid symlink headaches. But I ran into some `coreutils` differences between Mac and Linux recently, so I wanted to see what SO thinks is the best / most universal practice, and that [lead me](https://stackoverflow.com/a/246128/1708932) to the snippet here.

But arguably this might be overly defensive for our purposes. Either way is fine by me, happy to change it if you want

dae: We're not using symlinks here, so I'd favor the more readable version if we don't need the more elaborate one.

Also, aren't parent_dir and root_dir pointing to the same place?

glutanimate: `parent` would be pointing to `./tools`, and root to `.` 

But revisiting this again, I think we can just drop the relative paths here completely. I think my original intent here was to make `rebuild-web` callable both from the project root and when `cd`ed into tools (or from any where else, for that matter). I'm not sure if this ever worked, TBH, but it might have because specifying the target with `bazel build "//qt/aqt/data/web"` might have been absolute.

Either way, all of this no longer makes sense as `ninja` expects to be run from the project root, so I switched back to just specifying the paths absolutely, which is also consistent with other scripts under `tools`.

tools/web-watch
```diff
@@ -0,0 +1,25 @@
...
+
+# Monitor all web-related folders and rebuild and reload Anki's web stack
+# when a change is detected.
+
+set -e
+
+MONITORED_FOLDERS=("ts/" "sass/" "qt/aqt/data/web/")
+MONITORED_EVENTS=("Created" "Updated" "Removed")
+
+parent_dir=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
dae, line 13: Sorry, still this straggler

glutanimate: There's always one! Fixed.

dae:
Thanks Aristotelis!

PR #2154, opened 2022-10-31
Use custom scrollbar in stats screen, use grid layout and tweak CSS

kleinerpirat:
https://forums.ankiweb.net/t/anki-2-1-55-beta-3/24295/3

---
I always thought the single column with `max-width = 60em` was a huge waste of space in landscape mode, so I added a responsive grid layout.

To better separate the graphs, I used the same `TitledContainer` component as in the deck options. This also paves the way for help modals in the stats screen in the future.

### DIFF
--- a/sass/base.scss
+++ b/sass/base.scss
@@ -55,6 +55,7 @@ button {
     /* override transition for instant hover response */
     transition: color 0.15s ease-in-out, box-shadow 0.15s ease-in-out !important;
     border-radius: prop(border-radius);
+    @include button.base;
 }
 
 pre,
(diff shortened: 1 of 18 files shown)


kleinerpirat:
Also sorry for the hypocrisy re:
> ... no additional UI changes

But this is just reusing what we already got in other screens, so I don't think it will cause any trouble.

dae:
Two issues I spotted:
- when there's only a single column, there's too much padding on top
- if I flick the mouse wheel to scroll down to the bottom, the last graph is truncated
2022-11-01T11:34:15,261609569+10:00
2022-11-01T11:34:23,317686838+10:00

dae:
:rocket:

dae:
Minor regression: there's a strange outline when typing tags into the tags area.

PR #2156, opened 2022-10-31
Flip popover placement automatically to prevent overflow

kleinerpirat:
https://forums.ankiweb.net/t/anki-2-1-55-beta-3/24295/26


dae:
There's a bit of flicker where it's first animated downwards, but still an improvement, thanks :-)

PR #2159, opened 2022-11-01
Flatten buttons to match macOS style

kleinerpirat:
In an effort to unify the platform styles a bit better, I made all buttons flat by default. It should address complaints about the difference between Qt and webview buttons on macOS.

The inspiration came from this screenshot by forum user garrettm30:

Looking at the Type and Deck `QPushButtons`, I noticed the bottom border is a tad bit darker than the rest, which creates a subtle elevation effect without box-shadow.

## Sample from this PR
image

The button gradient is only used as a hover-effect now.

### DIFF
--- a/qt/aqt/data/web/css/toolbar.scss
+++ b/qt/aqt/data/web/css/toolbar.scss
@@ -3,7 +3,7 @@
 
 @use "root-vars";
 @use "sass/vars" as *;
-@use "sass/elevation";
+@use "sass/elevation" as *;
 @use "sass/button-mixins" as button;
 
 #header {
@@ -18,9 +18,9 @@
     @include button.base($with-hover: false, $with-active: false);
     overflow: hidden;
     padding: 0;
-    @include elevation.elevation(1, $opacity-boost: -0.08);
+    @include elevation(1, $opacity-boost: -0.08);
     &:hover {
-        @include elevation.elevation(2);
+        @include elevation(2);
         transition: box-shadow 0.2s linear;
     }
 }
--- a/qt/aqt/data/web/css/toolbar.scss
+++ b/qt/aqt/data/web/css/toolbar.scss
@@ -15,9 +15,14 @@
     border-radius: prop(border-radius);
     border-bottom-left-radius: prop(border-radius-large);
     border-bottom-right-radius: prop(border-radius-large);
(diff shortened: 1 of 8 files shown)


kleinerpirat:
That test needs to be rerun. The PR passes locally.

dae:
Please double-check, as it's failing when I test the build on my machine as well.

kleinerpirat:
Sass is a bit lazy and only compiles when style definitions in a user file are changed. That's why it continues to fail here. Is there any command to force Sass compilation?

kleinerpirat:
See? :D

hikaru-y:
#2167 should probably prevent local `bazel test` and `buildkite/anki-ci` from failing.

kleinerpirat:
@hikaru-y thanks for looking into this, but the reason for the failed checks is that there needs to be a change (e.g. a newline) in an output producing file to start the Sass compilation process anew.

As you can see, the newline added in the latest commit caused a fresh compilation, which in turn results in a new `colors.py`, so the variables are available for the Python tests.

hikaru-y:
I don't think so. The dependencies for the `//qt/aqt:extract_sass_vars` target are `//sass:_root-vars.css` and `//qt:extract_sass_vars`, and then `_root-vars.css` is generated here:
https://github.com/ankitects/anki/blob/bf34ac77672b615ea1def2a1088ac0ced0c4fc20/sass/BUILD.bazel#L106-L110
But there are no other labels or files listed in `srcs` or `deps`  except for `_root-vars.scss`. That' s why `colors.py` will not recompile unless you make changes to `_root-vars.scss`.

https://github.com/ankitects/anki/blob/bf34ac77672b615ea1def2a1088ac0ced0c4fc20/sass/_root-vars.scss#L5-L6
You can see here that we are importing `_vars.scss` and `_functions.scss`, so I think we need to pass all of these to `compile_sass` as deps.

Try checking out #2167 and you should see that even if `_root-vars.scss` is unchanged, if `_vars.scss` or `_functions.scss` is changed, `_root-vars.css` will be recompiled.

kleinerpirat:
Ah, now I see what you mean. Thanks for the clarification, that makes sense!

dae:
I might be biased, but I think it looks considerably better :-) Thanks for all your work on polishing this up Matthias.

PR #2165, opened 2022-11-01
Maybe preload images

kelciour:
As it seems to have some visible effect, I'd like to suggest to maybe delay setInnerHTML for up to 100ms to preload images (if necessary) to prevent flickering & layout shift on image load. It's for the front side of the card.

I tested a different delay on SDD and HDD. The delay of 50ms looked good on SSD, but didn't help much with flickering on HDD where the same image could take about 50-90ms to load.

As a side affect, the code will be also executed on the answer side too, but by that time any answer images will be already preloaded and the code shouldn't wait much time (maybe 1-5ms according to console.time). For cards with no images, the delay probably will be < 1ms or maybe 1-5 ms.

I didn't notice any visible difference in responsiveness except for some cards with images if they take some time to load. In this case, the card will be displayed with a bit of delay and the image will be already visible or will be loaded later.

### Test apkg file from the screencast

[Custom Study Session - Issue with Images.zip](https://github.com/ankitects/anki/files/9913135/Custom.Study.Session.-.Issue.with.Images.zip)

### Before

[Screencast from 01.11.2022 19_29_40.webm](https://user-images.githubusercontent.com/16260735/199311691-b895d376-b305-4049-9f18-df124687d01d.webm)

### After

[Screencast from 01.11.2022 19_56_00.webm](https://user-images.githubusercontent.com/16260735/199311721-6226f589-5485-4038-a550-96594a6522e7.webm)

### DIFF
--- a/ts/reviewer/images.ts
+++ b/ts/reviewer/images.ts
@@ -49,3 +49,17 @@ export function preloadAnswerImages(qHtml: string, aHtml: string): void {
         diff.forEach((src) => injectPreloadLink(src, "image"));
     }
 }
+
+export async function maybePreloadImages(html: string): Promise<void> {
+    const srcs = extractImageSrcs(html);
+    await Promise.race([
+        Promise.all(
+            srcs.map((src) => {
+                const img = new Image();
+                img.src = src;
+                return imageLoaded(img);
+            }),
+        ),
+        new Promise((r) => setTimeout(r, 100)),
+    ]);
+}
--- a/ts/reviewer/index.ts
+++ b/ts/reviewer/index.ts
@@ -16,7 +16,7 @@ globalThis.anki.mutateNextCardStates = mutateNextCardStates;
 
 import { bridgeCommand } from "../lib/bridgecommand";
 import { maybePreloadExternalCss } from "./css";
-import { allImagesLoaded, preloadAnswerImages } from "./images";
+import { allImagesLoaded, maybePreloadImages, preloadAnswerImages } from "./images";
 
 declare const MathJax: any;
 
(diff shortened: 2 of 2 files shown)


dae:
I'm a bit worried about possible regressions here, as we had a similar approach in the past which caused problems and got reverted: https://github.com/ankitects/anki/commit/8e5e277ee45778521e47dc646b4e2e2ce93f6870. But that said, I tried this PR and added a time.sleep(0.1) into mediasrv.py, and didn't see a problem, so perhaps something else has changed since then?

@hikaru-y any thoughts?

hikaru-y:
LGTM.

I have also tested this and do not see the problem mentioned in 8e5e277. I don't think this is because something has changed, but because the approach itself is different (albeit similar).

With a598066a, `setInnerHTML` is called, the page is updated, and then waits for the images in the page to load. During that loading, the page opacity is set to 0, so the above problem occurs.

On the other hand, this PR dynamically generates image elements with the same `src` as the images about to be displayed on the card and preloads them before calling `setInnerHTML`. Therefore, there can be no period of time when the page opacity is 0, which is caused by waiting for the images to load. (If the card has a `<script>` element, there will be a period of time when the page opacity is 0 while waiting for the script to execute.)

ts/reviewer/images.ts
```diff
@@ -49,3 +49,17 @@ export function preloadAnswerImages(qHtml: string, aHtml: string): void {
         diff.forEach((src) => injectPreloadLink(src, "image"));
     }
 }
+
+export async function maybePreloadImages(html: string): Promise<void> {
+    const srcs = extractImageSrcs(html);
+    await Promise.race([
+        Promise.all(
+            Array.from(srcs).map((src) => {
```
hikaru-y, line 57: Just a minor thing: Isn't `Array.from` redundant as the return value of `extractImageSrcs` is already an array of string?

dae:
Thanks for looking into it in more detail Hikaru; I retract my concerns :-)

dae:
Thank you both!

PR #2166, opened 2022-11-01
Prevent deck options switches from toggling on label click

kleinerpirat:
I just saw that clicking the label of a `SwitchRow` toggles the switch. This is of course unwanted because the new expected action is to open the help modal.


dae:
I understand the desire for consistency and to make the docs for these elements quick to access, but I'm a bit worried this change will be sacrificing accessibility: 
- https://www.a11ymatters.com/pattern/checkbox/
- https://stackoverflow.com/questions/52408594/is-it-a-wcag-2-0-aa-violation-if-the-label-of-a-checkbox-is-a-hyperlink

kleinerpirat:
How about this compromise? There'll still be a `<label>` element for screen readers, but the mouseclick event is muted.

dae:
That's better, though I should note it's not just about screen readers - having the text clickable gives a larger target area which you can click on, which can also be helpful for people who struggle to make precise mouse movements. To be honest, even though I don't struggle with the mouse, I find myself preferring to click on text when it is supported :-). To be fair though, our bigger toggle switches are easier to hit than the standard checkbox, and I note that iOS does not allow tapping on labels to toggle a switch unless you have the accessibility tools turned on. So maybe it's fine? Any thoughts @hikaru-y?

dae:
(another point: a lot of our other UI currently allows clicking on labels (eg the group boxes for front/back/styling in the card layout screen, the dropdown options, the checkboxes in the preferences), so maybe it would be surprising to the user to behave differently here?)

hikaru-y:
If a label text is right next to a standard small checkbox, many people, including me, would expect the checkbox to be toggled when the label is clicked. However, on Anki's deck options screen, there is a fair distance between a switch and its label, and also a switch is large enough to be easily clickable. So I am in favor of making it so that clicking on a label does not toggle its switch.

#### Regarding the code
I wonder if it might be better to make `id` in `Switch.svelte` and `forId` in `Label.svelte` optional and then remove `for` and `id` from `SwitchRow.svelte`, rather than using an event listener. Not only because it would make the code a bit cleaner, but also because the following line is useless if we don't need to tie labels to inputs.
https://github.com/ankitects/anki/blob/27eff122357f5378be75d193c85c0fc7374feef8/ts/deck-options/SwitchRow.svelte#L16

(Just my thoughts - feel free to ignore.)

dae:
Thanks for commenting Hikaru. The fact that the text is split up is a good point. Won't we still need the id though? We still want people with screen-reading software to be able to toggle the checkbox while they're focused/reading the label.

hikaru-y:
I don't know much about screen readers and accessibility, so I did some googling and found that apparently `id` and `for` need to be set on `label` and `input` for accessibility: https://www.w3.org/WAI/tutorials/forms/labels/
So I withdraw the part of my previous comment about code suggestions.

dae:
And I should have noticed https://github.com/ankitects/anki/pull/2165#issuecomment-1301515631 by reading the PR more carefully. These things happen :-)

So it sounds like your compromise is a go @kleinerpirat - thank you!

PR #2167, opened 2022-11-01
Add missing sass deps

hikaru-y:


kleinerpirat:
Thanks for this, Hikaru. You just made the build process seem a bit less like a magic box to me.

dae:
Ideally with Bazel the build will fail to compile at all when we forget to include dependencies, but the way we specify the load path to sass unfortunately allows it to see files that we haven't explicitly listed.

PR #2171, opened 2022-11-02
Set same font size for all platforms and make editor UI a bit smaller

kleinerpirat:
https://forums.ankiweb.net/t/anki-2-1-55-beta-3/24295/4

# Changes
- use base font size of 13px (macOS value) for all platforms
- decrement `1.6em` in various editor components to `1.5em`

## Before
image

## After
image

---
@dae Could you send me similar screenshots of Mac and Windows for comparison? :pray:

### DIFF
--- a/sass/base.scss
+++ b/sass/base.scss
@@ -44,6 +44,7 @@ html {
 }
 
 body {
+    --base-font-size: 13px;
     overscroll-behavior: none;
     &:not(.isMac),
     &:not(.isMac) * {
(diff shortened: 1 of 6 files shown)


dae:
Windows before->after, with native window for comparison:

2022-11-02T20:59:53,445536301+10:00
2022-11-02T21:00:55,644947092+10:00

kleinerpirat:
Thanks! This looks to me like we should keep the separate size for Windows. On Linux, I prefer the macOS size.

dae:
macOS before→after
shot-main
shot-pr

kleinerpirat:
From my side this would be a go. How do you feel about the slightly smaller editor UI (`1.5em`)? Other than that, nothing will change on mac.

dae:
The difference is quite subtle - I think the PR might be closer to the macOS defaults, but I'm not 100% sure :-)

PR #2172, opened 2022-11-02
Fix various RTL issues

kleinerpirat:
Fixes everything (except the odd sidebar highlight) mentioned in https://forums.ankiweb.net/t/anki-2-1-55-beta-3/24295/20

## Todo

- [ ] Rearrange browser views for RTL mode. The sidebar should be to the right, browser table in the middle and editor on the left:
image
- [ ] Fix buggy sidebar highlight
image

### DIFF
--- a/qt/aqt/data/web/css/deckbrowser.scss
+++ b/qt/aqt/data/web/css/deckbrowser.scss
@@ -58,6 +58,26 @@ tr:hover:not(.top-level-drag-row) {
         }
     }
 }
+[dir="rtl"] {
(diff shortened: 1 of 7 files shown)


dae:
You've been smashing out these fixes Matthias - most impressive! I'll hold off on another beta until you're ready - please let me know when you are.

Regarding the RTL dock layout, commenting out restoreState() seems to fix it, so I guess we should include rtl/ltr in the browser state key name so that the states are saved separately.

PR #2175, opened 2022-11-03
Equalize browser panes on splitter double-click

glutanimate:
Adds the ability to quickly resize the editor and table view to equal width/height by double-clicking on the splitter handle (similar to window tiling on Windows and various Linux WMs):

https://user-images.githubusercontent.com/5459332/199632911-8cef9e49-6367-48aa-9a5a-51495cfe5f6c.mp4


dae:
Looks handy - thanks!

PR #2176, opened 2022-11-03
Fix RTL arrangement of browser views

abdnh:
This just implement your suggestion in https://github.com/ankitects/anki/pull/2172#issuecomment-1300190664 so that @kleinerpirat has less on his plate.

> Regarding the RTL dock layout, commenting out restoreState() seems to fix it, so I guess we should include rtl/ltr in the browser state key name so that the states are saved separately.

I fixed this a year ago in #1453 but it was reintroduced recently in #2132

The sidebar highlight issue seems to be a Qt issue; it disappears after scrolling.

### DIFF
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -138,9 +138,14 @@ class Browser(QMainWindow):
 
         # restoreXXX() should be called after all child widgets have been created
         # and attached to QMainWindow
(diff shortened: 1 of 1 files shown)


dae:
Thanks for doing this Abdo. I'd actually forgotten about that previous fix, though maybe it was still somewhere in my subconscious. :-)

PR #2177, opened 2022-11-03
Hide note/card switch label during animation

glutanimate:
This introduces a tiny change to the card/note switch animation that hides the label while the knob is traversing over it. With the switch to the full text label in #2117, to me this behavior feels slightly smoother.

Here's a quick comparison to show the difference (the previews don't work for these tiny videos, so you'll have to open them in a separate tab / start playback with space):

Before: https://user-images.githubusercontent.com/5459332/199638786-c88188b2-f289-49a7-9413-7411a701be6a.mp4

After: https://user-images.githubusercontent.com/5459332/199638676-5ba72840-3e0c-4741-ad42-828fa850a240.mp4


dae:
Nice!

PR #2181, opened 2022-11-04
Fix HTML editor not showing content initially

### LINKED ISSUE #2179: HTML does not appear immediately in field
Running on dba4925aba6f3ecbb6980651a82e64b1ffe62eb7 without addon

How to reproduce:
* Open a note editor
* Ensure you have a field with some content
* click on <> above the field

Expected result:
* the html of the field content appears below

Actual result:
* the new text space is empty 
2022-11-03-204249_688x91_scrot

The content appear either when you edit the field or when you put the focus in the html version of the field.

Maybe related:
* Click on <>
* Click in the field
* use  tab

Expected result:
* focus is in the html part of the field

Actual result:
* focus seems to disappear. If you type, then the text appear. 

Finally, I'd like to note that the HTMl part of the field has "1" in a grey box on the left of the html field. And it took me a long time to understood that it's counting lines. I had to actually type a new line in the html field editor.
I suggest that, when there is a single line (which, I assume, will be most of the time), the 1 should not appear. It does not provide any relevant information and is mostly confusing to be honest.

I can split this in multiple issues if you prefer. Last one, I don't even know if it's actually an issue, so if I were not already testing code in this part of the app, I probably would have put it in the forum instead

kleinerpirat:
Closes #2179.

---
There's one minor issue left: When `PlainTextInput` is collapsed, `RichTextInput` isn't refocused.

### DIFF
--- a/ts/components/Collapsible.svelte
+++ b/ts/components/Collapsible.svelte
@@ -11,7 +11,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     export let toggleDisplay = false;
     export let animated = !document.body.classList.contains("reduced-motion");
 
-    let collapsed = false;
     let contentHeight = 0;
 
     function dynamicDuration(height: number): number {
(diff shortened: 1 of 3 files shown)


dae:
Thanks Matthias. I must admit I currently prefer the HTML editor appearance when 'reduce motion' is on - it appears instantly, instead of appearing after the area has animated into view, which feels slightly clunky.

Arthur-Milchior:
Thanks a lot @kleinerpirat. And by the way, nice to meet you

kleinerpirat:
@Arthur-Milchior nice to meet you too!

PR #2182, opened 2022-11-04
Make dupes highlight persist on field blur

kleinerpirat:
Before, the highlight would disappear on blur, leaving only the link.


dae:
Nice

PR #2183, opened 2022-11-04
Adjust QTableWidget stylesheet

### LINKED ISSUE #2109: Browser table styling issues
I've investigated a bit regarding the [issues I found](https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/30?u=rumo):

1. First fix is trivial: Give column labels some horizontal padding, so they remain separated from the borders if the column is very narrow.
2. In order to center the column labels, we could add negative padding to the right to compensate for the sort indicator. That looks fine for the non-sort columns, but will lead to the label and sort indicator of the sort column overlapping, if it is very narrow. There doesn't seem to be any way to style a column depending on whether it is sorted by, so I don't think this is solvable with stylesheets. Maybe we have to return to the old indicators?
3. The vertical header can easily be hidden. In fact, Anki is already explicitly disabling it on all platforms except Windows for some reason. However, then the left table border is missing (screenshot 1). As a hack, we could keep showing it, but use negative padding, so only the right border is visible (screenshot 2).
    grafik
    grafik

WDYT, @kleinerpirat?

kleinerpirat:
Closes #2109.

---
## Changes
- remove vertical header on Windows
- center table headers
- add border to empty area below table cells

@RumovZ this is as good as I could get it. The reason I didn't use a border on `QTableWidget` itself is that it would wrap around the scrollbar. That would not look good with its custom minimal design.

### DIFF
--- a/qt/aqt/browser/table/table.py
+++ b/qt/aqt/browser/table/table.py
@@ -11,7 +11,6 @@ from anki.cards import Card, CardId
 from anki.collection import Collection, Config, OpChanges
 from anki.consts import *
 from anki.notes import Note, NoteId
-from anki.utils import is_win
 from aqt import gui_hooks
 from aqt.browser.table import Columns, ItemId, SearchContext
 from aqt.browser.table.model import DataModel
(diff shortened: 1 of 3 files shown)


RumovZ:
It's not perfect (screenshot), but barely noticeable anymore. Nice one! 🙂 
grafik

dae:
Thanks Matthias, I'll get a new one out today.

PR #2184, opened 2022-11-07
UI size tweaks

kleinerpirat:
Relevant forum posts:
- Indicator font size: https://forums.ankiweb.net/t/anki-2-1-55-beta-3/24295/75
- Table row height: https://forums.ankiweb.net/t/anki-2-1-55-beta-3/24295/80


dae:
Thanks for looking into all these issues Matthias!

I'm afraid we can't use such an approach for the table height - it causes a massive performance regression when showing lots of rows. See _update_font() for how we set the height. A user could presumably tweak the browser height of an unused notetype in order to adjust the padding there.

Changing the text in the dropdown seems like a good idea, but is it my imagination, or is the text slightly too large now compared to eg. the deck options text?
2022-11-08T23:09:09,110642175+10:00

kleinerpirat:
> See _update_font() for how we set the height. A user could presumably tweak the browser height of an unused notetype in order to adjust the padding there.

Ah yes, you mean that hidden "Browser Appearance" setting for browser entries in the template editor. Are there any good reasons why we allow users to set different fonts and font-sizes per notetype? In my eyes a global option in the preferences would make more sense, as it ensures a consistent look.

kleinerpirat:
> is it my imagination, or is the text slightly too large now compared to eg. the deck options text?

The dropdown text is the same size as the other deck options text. It's probably a slight optical illusion.

kleinerpirat:
What do you think about the new approach? I experimented with a `QSlider` for the new setting, but it felt awkward not knowing the exact integer it is set to (sadly the default widget doesn't show it).


If the change is okay, I'm wondering how to proceed with the schema entry for the font size. I guess we'd keep it in some form to ensure legacy notetype support?

dae:
As you guessed, the browser font override exists because some text may not be readable at all if the wrong font is specified, or the wrong variant may be chosen (eg Chinese shown instead of Japanese). The font size override exists because some users want the ability to increase the size of the font depending on the language being shown - eg showing kanji in a larger font than Latin text, since there are thousands of variants to recognize instead of ~50. I fear there would be backlash if we removed that option. The behaviour is also not new to 2.1.55, so it may not be the best time to be looking into that - I'd prefer we focus on issues that 2.1.55 has introduced such as https://github.com/ankitects/anki/issues/2190 and https://github.com/ankitects/anki/issues/2195, so we can get a release out the door :-)

kleinerpirat:
I reverted all the changes on browser row height. I hope the other changes of this PR are good to go.

dae:
One minor thing that won't block a merge: the font size on dropdowns feels a bit big. For example, in the editor, clicking on the cog shows text that is noticeably bigger than the text on buttons.

Another minor issue I noticed that is not related to this PR: in the deck options screen, the labels appear to be top-aligned rather than center-aligned, eg "new cards per day" or "learning steps".
2022-11-23T16:49:24,211347776+10:00

PR #2185, opened 2022-11-07
Fix reduce motion bottom border issues

kleinerpirat:
Fixes https://forums.ankiweb.net/t/anki-2-1-55-beta-3/24295/81 first issue:

image


kleinerpirat:
Now also addresses https://forums.ankiweb.net/t/anki-2-1-55-beta-3/24295/82.

dae:
Would it be worth doing a survey for the pin hiding? It does make the UI cleaner, so it would be interesting to see how many people want the old behaviour.

dae:
Oh, nevermind - I see now you've only hidden it when reduce motion is on. Sounds good.

PR #2187, opened 2022-11-09
Fix error while compiling rslib

dobefore:
1. pass `--experimental_allow_proto3_optional` to protoc
2. add missing derive trait `Eq` to protobuf-generated struct `DayLimit`
3. delete derive trait Eq` from `ForeignNote`


dae:
Thanks, #2 and #3 are useful fixes for Rust 1.65. #1 is due to an outdated protoc.exe, so I'd suggest updating it instead of changing the code.

dobefore:
I've remove the argument from protoc.

dae:
Thanks @dobefore!

PR #2191, opened 2022-11-12
Fix inverted Ctrl+right/left handling in RTL fields (again)

abdnh:
This reintroduces the fix added in #730


dae:
Thanks Abdo. Out of curiosity, do you happen to know which PR broke it? If you don't know off the top of your head, no need to go bisecting - it's not important, I'm just curious :-)

abdnh:
It's #1403 

I noticed this broke in April (according to my todo list) but never bothered fixing it until now.

dae:
Ah, that makes sense, thanks.

PR #2197, opened 2022-11-19
Fix TagsSelectedButton's popup appears in wrong position

hikaru-y:
This fixes point 2 of https://forums.ankiweb.net/t/anki-2-1-55-beta-3/24295/91.


dae:
Fantastic, thanks Hikaru!

PR #2198, opened 2022-11-19
Call browser_will_show before restoring state

abdnh:
This is to fix a recent issue that started to appear in the Fastbar add-on where the layout of the toolbar is broken.

image

Looks like this area of the browser is a minefield. Hopefully this one-line change won't cause more problems.


dae:
Yep, it does feel like a bit of whack-a-mole sometimes :-) Thanks for looking into this.

PR #2199, opened 2022-11-20
fix body class names when changing theme

BlueGreenMagick:
body uses 'night_mode' class instead of 'night-mode'. Also adds and removes 'macos-dark-mode'.


dae:
Thanks BGM!

PR #2205, opened 2022-11-22
Pass addon name to config edit hook

zjosua:
Only the config schema is currently available to check which add-on was edited. [Example](https://github.com/zjosua/anki-mc/commit/a9aca96acfe263314d2b8323c5291a6076af0b49)
This is cumbersome and not quite safe (similar add-ons might have the same config schema).

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -1601,7 +1601,7 @@ class ConfigEditor(QDialog):
 
     def accept(self) -> None:
         txt = self.form.editor.toPlainText()
-        txt = gui_hooks.addon_config_editor_will_save_json(txt)
+        txt = gui_hooks.addon_config_editor_will_save_json(txt, self.addon)
         try:
             new_conf = json.loads(txt)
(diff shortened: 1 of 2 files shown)


dae:
This seems reasonable, but to avoid breaking existing add-ons using the old hook, I think you'll need to create a new one, then call both old+new (maybe adding a deprecation warning to the former as well).

zjosua:
Does something like the `@deprecated` decorator from anki._legacy exist for hooks? Or should I just make the deprecated hook print a warning?
And is there a naming convention for updated hooks?

dae:
I don't think we have a tool for deprecating hooks yet, or a scheme for creating new hooks. You could use addon_config_editor_will_save_json2, or call it something like addon_config_editor_will_update_json instead, which is a bit cleaner. When invoking the new hook, you can check gui_hooks.addon_config_editor_will_save_json.count() to see if any legacy users exist, and if they do, print a warning before calling the old hook.

dae:
Looks good - thanks!

PR #2206, opened 2022-11-23
Fix Floating Ui error causing freezes on image resize

### LINKED ISSUE #2190: Hang when resizing images in the editor
From https://github.com/ankitects/anki/commit/ef933f997584cac743afcb77579d1dc1506dcbee#commitcomment-89311319:

> @kleinerpirat When resizing images in the editor with a double-click, the following appears in the console:
> 
> Error: Floating UI: duplicate flip and/or autoPlacement middleware detected. This will lead to an infinite loop. Ensure only one of either has been passed to the middleware array.
> 
> That seems like the same issue that this patch was trying to fix. I can't see any other references to flip() or autoPlacement() in the code, but assume there's something more subtle going on in WithFloating. Any ideas?
> 

Originally introduced by the fix in #2156

kleinerpirat:
Fixes #2190.
Also fixes elements briefly moving in the wrong direction before flipping.

According to Floating UI's docs, `flip()` is an alternative to `autoPlacement()` and both mustn't be used together.

> You’ll want to use autoPlacement() instead of flip() if you **don't want** to give the floating element a “preferred” placement and let it choose a placement for you.
https://floating-ui.com/docs/autoplacement

Flip is the better option for us, so I made the following changes:

- replace the usage of `autoPlacement()` with `flip()`
- adjust user components of `<WithFloating>` accordingly

### DIFF
--- a/ts/components/Select.svelte
+++ b/ts/components/Select.svelte
@@ -26,7 +26,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
 <WithFloating
     show={showFloating}
-    placement="bottom"
+    preferredPlacement="bottom"
     offset={0}
     shift={0}
     hideArrow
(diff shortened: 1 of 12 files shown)


dae:
Great, thanks for looking into Matthias!

PR #2207, opened 2022-11-23
Make OptionsButton more prominent

kleinerpirat:
https://forums.ankiweb.net/t/anki-2-1-55-beta-3/24295/112

image

I figured giving it its own group with an extra api export could be useful for some add-ons.


dae:
I think I'm more used to seeing option buttons on the right, but don't have a strong preference, and exposing it for add-ons makes sense.

PR #2209, opened 2022-11-23
Revert to pseudo-element approach for field description

kleinerpirat:
Fixes https://forums.ankiweb.net/t/anki-2-1-55-beta-3/24295/115.

## Issue:
Field descriptions don't update
- when they're changed via the Fields dialog
- when changing notetypes in Add-mode

## Backstory
When I first introduced [in-field descriptions](https://github.com/ankitects/anki/pull/1912), I used a pseudo-element inside `ContentEditable` and used a very primitive approach to get an `.empty` class. Both were bad ideas, so @hgiesel pushed an [improved PR](https://github.com/ankitects/anki/pull/1923), where he used a `FieldDescription` component that would appear *inside* `<anki-editable>`.

It definitely was the more elegant approach, but among other things it introduced the issue mentioned in the forum.

There seems to be a disconnect between `NoteEditor` and the `ContentEditable` elements. Although `field.description` would update inside `NoteEditor`, that change wouldn't go through to the actual `<anki-editable>`, unless an update was forced by typing text into the field. My best guess is it's because the editables are appended manually.

## A healthy mix
This PR takes the best of both approaches: Henrik's way of getting an `.empty` class and moving editor-specific logic out of `ContentEditable`, and my tried-and-tested CSS variable approach, which might be a cheap hack from a coder's perspective, but the end result is the same and it does update properly.


kleinerpirat:
I reckon you've been working on buildkite for the new build system? Checks are passing locally.

dae:
Looks like a flake - I just restarted the test and it went away. :sweat_smile:

dae:
A working hack beats an elegant solution that doesn't work :-) Thanks Matthias!

Issue #2127, opened 2022-10-12
Empty tag added when completing tags


dae:
https://forums.ankiweb.net/t/anki-2-1-55-beta/23470/28

PR #2211, opened 2022-11-24
Add platform-specific classes to body of ts pages

kleinerpirat:
Sets the dynamic body class on all screens that use `load_ts_page` instead of `std_html`, namely:
- Change Notetype
- Deck Options
- Overview
- Stats
- Card Info
- Import CSV

This should remove the custom scrollbar styling on macOS. It will also affect the default button font size.

### DIFF
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -647,11 +647,12 @@ html {{ {font} }}
         self.setSizePolicy(sp)
         self.hide()
 
-    def inject_dynamic_style_and_show(self) -> None:
-        "Add dynamic styling, and reveal."
+    def add_dynamic_css_and_classes_then_show(self) -> None:
+        "Add dynamic styling, set platform-specific body classes and reveal."
         css = self.standard_css()
+        body_classes = theme_manager.body_class().split(" ")
 
-        def after_style(arg: Any) -> None:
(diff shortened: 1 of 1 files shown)


dae:
Thanks for the quick fixes Matthias! Can confirm it's working correctly now :-)

kleinerpirat:
How long will you stay up tonight? I'm working on a few small changes that I'd like to have in the next beta - might take a few more hours.

dae:
I was planning to wait at least another day or two for a new beta - they're somewhat time-consuming to build (and for users to download), and Linux users have a workaround for the current packaging issue. I should be around for a few more hours if you need a quick reply to anything.

In terms of getting a release out, I'm hoping the next beta is the last one (or second last if more issues are found that are important). I think the only existing issue still on my radar is the https://github.com/ankitects/anki/issues/2127, and perhaps merging the ninja changes in). Thanks again for all your hard work throughout 2.1.55 - it's much appreciated, and the users seem to really like the visual changes. :-)

kleinerpirat:
Another day sounds good to me. I'm currently polishing the tag editor. After that, the CSV import window and some minor tweaks to font-sizes and colors are the last remaining projects for me. Should be doable by tonight, Austrian time :)

PR #2214, opened 2022-11-25
Fix shortcuts for tag actions not working

### LINKED ISSUE #2080: copying all tags has broken
ctrl+a still selects all tags, but ctrl+c no longer has an effect. Appears to have regressed since 2.1.54

### LINKED ISSUE #2081: allow deleting all tags at once
Similar to #2080, it would be nice if once all tags were selected, it was possible to delete them.

hikaru-y:
Fixes #2080
Fixes #2081

I have confirmed that this works in the editor (Add, Edit Current, Browse) and in the import file dialog (both on the **Tag updated notes** and **Tag all notes** sections).


dae:
Much nicer - thanks Hikaru!

PR #2215, opened 2022-11-26
Fix some issues with the Tag Editor

kleinerpirat:
Fixes:
- focus border showing on left side of tag input
- tag editor slightly changing height when it's focused or when activating tag input mode
- tag editor pane cutting off bottom pixel of tag editor
- second empty tag getting appended when an autocomplete suggestion is selected via <kbd>Enter</kbd>
- newly added tag with no content briefly showing before input mode is activated

I also made some strings translatable.


dae:
Thanks for all of these Matthias. There are still some minor glitches with the tag editor, but we can tackle them later. (when adding a new tag, there's a brief flash, and when selecting items from the autocomplete menu, the differing length of selections can cause the currently-displayed input to wrap/unwrap. I brought this up previously, but don't recall if we ended up fixing it for it to later regress, or whether it was always an issue: https://github.com/ankitects/anki/pull/1264#issuecomment-917530590)

PR #2216, opened 2022-11-26
Make dropdown text smaller than page text by default

kleinerpirat:
This should address some complaints about font size being too big.


dae:
Thanks, that looks a bit more natural to me. :-)

PR #2217, opened 2022-11-28
use BUTTON_BG for button background

BlueGreenMagick:
Using `BUTTON_BG` seems more appropriate here.

It doesn't change anything in vanilla Anki since `BUTTON_BG` and `BUTTON_GRADIENT_START` is identical in default palette, but it's slightly confusing when using [ReColor add-on](https://ankiweb.net/shared/info/688199788) to modify theme.


dae:
Look good to you @kleinerpirat?

kleinerpirat:
Yes :) This is a leftover from when everything had gradients. Thanks @BlueGreenMagick!

dae:
Thank you both!

PR #2218, opened 2022-11-28
Use less specific html selector to not override user CSS

kleinerpirat:
There will be no harm on our side and it prevents issues with existing notetypes that modify the base font size, like https://forums.ankiweb.net/t/anki-2-1-55-beta-5/24870/31.

Generally speaking, I'd say we shouldn't micro-manage all breakages that stem from the CSS changes, but this one is easy and prevents significantly more effort on the user side.


dae:
Thanks Matthias, I was going to suggest we do as much, and agree with your assessment.

PR #2220, opened 2022-11-28
Give containers brighter color than canvas

kleinerpirat:
Making containers stand out from the background has been on my to-do list for some time. In order to make this work with inputs I had to tweak the CSS variables a bit.

image

I also gave the deck table on the main page the *elevated* look:

image

---

## Extra
Re: https://forums.ankiweb.net/t/anki-2-1-55-beta-5/24870/45

I gave the `.field-container` element a background and rounded the corners of `LabelContainer`, so the fields will not look as awful with custom backgrounds as they otherwise would. Making the webviews transparent proved difficult, because some pages would only render on resize, so I don't want to pursue this any further right now.

### DIFF
--- a/qt/aqt/data/web/css/deckbrowser.scss
+++ b/qt/aqt/data/web/css/deckbrowser.scss
@@ -3,6 +3,20 @@
 @use "sass/root-vars";
 @use "sass/vars" as *;
 @use "sass/card-counts";
+@use "sass/elevation" as *;
+
+table {
+    padding: 1rem;
+    background: var(--canvas-elevated);
+    border: 1px solid var(--border-subtle);
+    border-radius: var(--border-radius-large);
+
+    @include elevation(1, $opacity-boost: -0.08);
+    &:hover {
+        @include elevation(2);
+    }
+    transition: box-shadow 0.2s ease-in-out;
+}
 
 a.deck {
     color: color(fg);
@@ -44,7 +58,7 @@ body {
 .current,
 tr:hover:not(.top-level-drag-row) {
     td {
-        background: color(canvas-inset);
+        background: color(border-subtle);
         &:first-child {
             border-top-left-radius: prop(border-radius-large);
             border-bottom-left-radius: prop(border-radius-large);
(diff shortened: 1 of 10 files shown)


dae:
I'm assuming you'd like this in 2.1.55? I understand you could argue this is addressing a regression in the handling of custom backgrounds, though it makes me a bit nervous to be making styling changes this close to a release. What do you want to do?

dae:
(I do think it looks good :-)

kleinerpirat:
It was in response to point 4 here: https://forums.ankiweb.net/t/anki-2-1-55-beta-3/24295/93

These are just minor CSS changes. I didn't change any color names, nor did I add any new ones. IMO it should be safe to merge before the release.

As you know, I've been working on a bigger PR (CSV import + other screens) the last couple of days. Sadly real life got in the way this weekend. I assume I'm now too late to push larger changes (if the ones in this PR are too major already).

I can try to pump out alternative isolated fixes instead of the overhaul...

dae:
There is no hard deadline, just please bear in mind that the more non-trivial changes we make, the more the release is likely to slip. :-) I don't think we'll be able to fix everything for the 2.1.55 release, I just want to address the major issues and get it out, and then we can follow up with one or more subsequent releases that address the remaining issues/new ones that are reported. What sort of changes are you making in that other PR?

dae:
Minor issue I noticed after merging: ideally the borders would collapse at thinner window sizes, and not show a scrollbar.
2022-11-29T17:56:28,068232034+10:00

kleinerpirat:
Here's a list of the changes in the other PR:

- CSV Import screen refactoring:
  - It now uses `TitledContainer`s with the elevated tiles look and has a better layout overall.
  - I swapped out the `<select>` elements for our custom `Select.svelte` component and made it keyboard accessible (also relevant for the ones in the deck options).
- `ScrollArea` and `Page` component:
  - `ScrollArea` indicates overflow with a scroll-shadow and makes `StickyContainer` redundant.
  -  the component is used in the CSV Import window, Stats screen, Editor and Deck Options
  - `Page` is intended to be the default outermost element for every screen, so we don't need to set `body { height: 100vh; width: 100vw; }` every time
  - These changes do not need to be made all at once. We can gradually implement the components in our screens.

dae:
The first one sounds fairly safe. The second one sounds a bit more involved.  If we pushed that back until after 2.1.55, would we lose out on any fixes for issues people reported? The CSV option flowing under the bottom label perhaps?

PR #2222, opened 2022-11-30
CSV Overhaul | Add Page/ScrollArea components

kleinerpirat:
I need to go to bed now, but I'll provide a detailed description on what this PR fixes tomorrow morning.

The main point was to rework the CSV window:

image

... but it uses some components that I wanted to update in other screens too.

You should test it on the following screens:
- CSV import
  - new layout with `Page.svelte` and `ScrollArea.svelte`
  - swapped `<select>` for `Select.svelte`
- Editor
  - Fields pane affected due to the `ScrollArea` changes (scroll shadow now works as intended)
  - Tag Editor doesn't have a distinct background color anymore
- Deck Options
  - new layout with `Page.svelte` and `ScrollArea.svelte`
  - refactored/fixed `SpinBox.svelte`
  - affected by the `Select.svelte` changes
  - tweaked some margins and other CSS
- Change Notetype
  - new layout with `Page.svelte` and `ScrollArea.svelte`
- Stats
  - new layout with `ScrollArea.svelte`
  - large screens now show three columns

The main new feature of the `Page` layout is the scroll-shadow on overflow in any direction, as it wraps the content in a `ScrollArea`. You can see that component in action in the CSV import window, where it is used both for the `Page` layout and the preview table (horizontal scroll indication).

The new paradigm is to set `body { height: 100vh; }` and use Flexbox to designate screen areas. This makes things like `StickyContainer` redundant, as it can now simply be a `Container` outside of the `ScrollArea`.


dae:
Thank you for all of your hard work Matthias - I hope the effort you have been putting in has not been affecting your studies.

After some testing, I think we may need to put this off until after 2.1.55. While there's some good stuff here, there are some issues to work through, especially on mobile. Would it be much work to cherry-pick some of the fixes, while leaving the rest for now? From testing a few of the screens:

Stats screen:

- At wider widths, looks good.
- Minimum width feels a bit wide, and forces users to see two graphs side by side even at the smallest width. This is particularly bad on mobile, where the screen width is less than the specified minimum width.

CSV screen:

- I like the new select style.
- Not sure if this is a new issue, but the scroll wheel seems inconsistent - sometimes it scrolls up and down, and sometimes it doesn't, depending on where I move the mouse cursor in the window.
- The import button is not visible on mobile until you scroll down, despite there being empty space at the bottom. This may be due to the use of 100vh and the way margins work - in order to have content start below the top bar but to allow it to scroll under the bar as you scroll down, the webview on iOS is effectively the entire screen size, but with the top margin such that content is shifted down below the bar.
- (minor) Putting the filename on a separate line is more symmetrical, but wastes a bit of vertical space. Given the user generally knows which file they opened, I wonder if it's worth the cost?

Deck options:

- Seems to work well on desktop
- On mobile, initially the bottom of the page is cut off when scrolling down, and top of page when scrolling back - maybe more issues related to the scroll area change? If you don't have access to an iPhone but have an iPad, you should be able to reproduce this by running this PR with

ANKI_API_HOST=0.0.0.0 ./tools/ts-run

then surfing to http://your.computer.ip.address:40000/_anki/pages/deck-options.html#test on the iPad. To emulate the narrow width, you can use split screen mode. It might be worth checking on your Android phone as well. You can also use this to see how graphs.html behaves with a narrow width.
- The dropdown buttons have become a bit small, making it hard to tap on them.
- The spinners on mobile have improved, but left and right buttons have differing heights.

kleinerpirat:
> (minor) Putting the filename on a separate line is more symmetrical, but wastes a bit of vertical space. Given the user generally knows which file they opened, I wonder if it's worth the cost?

In the future we might want to add a file input into that line, so users can select a different file from the already open window. It's what I'd naturally expect from such an import screen.

dae:
Hmm, is that common in other apps? To be honest I'm not even sure we need to show the filename (and have brought this up with @RumovZ in the past), as isn't it reasonable for the user to know what file they selected? When in doubt I usually go looking for examples in other apps, and the only one that came to mind easily was libreoffice, where when you open a csv file it opens a dialog with options for how the file should be read like we're doing here. That dialog does not show the filename or give an option to select a different file. Can you think of any examples of other apps that show filename/allow you to choose a different file after presenting a dialog with options after opening a file?

dae:
Sorry, I was partly wrong - I had my WM set up to not show window titles :-) It does show the filename, but doesn't provide a way to select another.

kleinerpirat:
This is almost mergable. I'm still trying to find the cause for the inconsistent scroll.

kleinerpirat:
If you got the time to take a look at this, maybe it can fit inside the next beta. Overflow detection is not perfect yet in the CSV import window (minor cosmetic issue), but all the other screens work fine.

kleinerpirat:
Nevermind, Damien. It's too rushed right now and I messed up the merge (automatic was a bad idea). I've whirled up enough dust up with all these PRs. This can wait :)

dae:
Yep, let's not rush. Thanks for the effort in any case :-)

kleinerpirat:
Superceded by #2248.

PR #2223, opened 2022-11-30
Refactor SpinBox component

kleinerpirat:
Cherry-picked from #2222.

This fixes the increment button not showing.

---

I'll do my best to pick out the other high-priority changes from that PR as well. Hopefully I separated my commits orderly enough this morning :crossed_fingers:


dae:
Thanks for taking the time to extract all these fixes Matthias. :-)

PR #2224, opened 2022-11-30
Fix glitchy animation of floating elements

kleinerpirat:
Cherry-picked from #2222.

I'll keep this as a draft until I fixed the janky animation when the container initially goes downwards until it overflows and flips.


dae:
This seems to work well, and the issue you mentioned seems minor - I suspect most of the time people won't even notice if they're not looking out for it.

PR #2226, opened 2022-12-01
Use third grid column in graphs layout on wide screens

kleinerpirat:
Cherry-picked from #2222.

You said you had issues on mobile with this change. But compared to main, it doesn't change anything below 1400px width. Perhaps you didn't notice it in the previous betas?


dae:
Nope, it was definitely regressing. Looks like the 'px' is required - once I added that, it's working as intended.

kleinerpirat:
Ah! I didn't even notice that I had left the px out. Long nights 😄

PR #2227, opened 2022-12-01
Refactor Select component | Fix CSV import issue

kleinerpirat:
Cherry-picked from #2222.

This adds the refactored `Select` component to the CSV import window. Additionally, I made a temporary fix for the layout issue that isn't nearly as involved as the overhaul.

### DIFF
--- a/sass/_button-mixins.scss
+++ b/sass/_button-mixins.scss
@@ -105,8 +105,6 @@ $focus-color: var(--shadow-focus);
 
 @mixin select($with-disabled: true) {
     width: 100%;
-    white-space: nowrap;
-    text-overflow: ellipsis;
 
     pointer-events: all;
     cursor: pointer;
--- a/sass/base.scss
+++ b/sass/base.scss
@@ -57,42 +57,9 @@ samp {
     unicode-bidi: normal !important;
 }
 
-[dir="rtl"] {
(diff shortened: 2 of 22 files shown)


dae:
The selects look good, and happy to see the filename in the same area as the button :-) I presume this PR also addresses the reported issue where one of the options was hidden by default?

Regarding the changes to the import bar, there are two issues:

1) As it's no longer sticky, if the user scrolls down it disappears, and they need to scroll back up in order to be able to proceed, which is not the best UI experience. What's the rationale for making this change? 
2) The filename and button appear to be at different heights, and the area takes up a bit too much space:

2022-12-01T17:08:18,807329557+10:00

kleinerpirat:
> As it's no longer sticky, if the user scrolls down it disappears, and they need to scroll back up in order to be able to proceed, which is not the best UI experience. What's the rationale for making this change?

No rationale there, just an oversight :) I forgot to change `bottom` for `top`.

The misalignment should also be fixed now.

kleinerpirat:
I see. This is because I reverted to `position: fixed`. That was also the culprit for the truncation of the tag option previously. Fixed elements do not have any effect on the document flow.

dae:
Seems to work well now - thanks!

PR #2228, opened 2022-12-01
Allow forcing fusion style via environment variable

kleinerpirat:
Setting `FORCE_FUSION_STYLE` will make Anki look really old-school now. This has been requested by @AnKingMed.


dae:
I don't mind exposing this, but currently it doesn't handle switches between light/dark mode properly, which will likely lead to support questions :frowning_man:

kleinerpirat:
It seems that at least on my system only `QComboBox` gets buggy after a theme switch. I wonder how we've handled that in the past.

kleinerpirat:
> My assumption was that @AnKingMed was [...] not so much asking for stock fusion over your custom stylings.

I really do think he wants the native fusion look, but I also think it's a good idea to allow forcing the custom styles. Perhaps two separate env vars would be best.

Applying only the `QComboBox` and `QSpinBox` changes does fix the theme change issues (no idea why those happen in the first place, though), but they do not fit with the old look. I can try to emulate the fusion style for those elements.

kleinerpirat:
Alright, now we got two options:
- `FORCE_FUSION_STYLE` for an almost-perfect imitation of the 2.1.49 look
- `FORCE_CUSTOM_STYLE` to enable custom Qt stylesheets on macOS

In addition I tweaked the default button gradient a bit. I'm really happy with how this turned out.

kleinerpirat:
> This PR went from being a very simple change to +100 lines and two separate options

Most of those lines are just CSS. The logic in `theme.py` is still as simple as it was at the start of the PR.

> How would you feel about turning this back into a single option, and just solving the day/night issue to start with?

Solving the day/night issue required those extra 100+ lines. `QComboBox` and `QSpinBox` require stylesheet overrides to work with day/night mode. They should not have any maintenance cost.

I don't really see the problem with providing two options. One of them is in use for Linux/Windows by default anyway. To prevent user complaints, one could just keep this out of the docs. I think @AnKingMed is one of around 2 people in the world who would prefer the Fusion style over the new one (which cannot be called "Fusion" anymore as pretty much every widget Anki uses is custom-styled).

> This might be better done with a var in the profile, which can be enabled/disabled via the debug console.

Agree. I didn't know where to search on how to do that. Could you link an example?

AnKingMed:
> I think @AnKingMed is one of around 2 people in the world who would prefer the Fusion style over the new one

The survey I did on instagram was 60% new style, 40% fusion style with 1000+ responses. So at least 400+ ;)

dae:
@AnKingMed Presumably your survey was showing the older Anki vs the newer one with native macOS styling? If so, that tells us that a portion of users prefer the old fusion styling over the macOS native styling, but it doesn't tell us what people think about Matthias' customizations, which aren't visible on Macs at the moment. 

@kleinerpirat 

> one could just keep this out of the docs

If only that would solve the problem!

https://www.hyrumslaw.com/

> Solving the day/night issue required those extra 100+ lines.

But only when avoiding your custom styling, right? A simple option to make Macs display like Windows/Linux machines is low cost, and uses the same code paths as the other platforms, minimizing the chance of breakages. A hidden option to make the interface display in a way that no platforms do by default is likely to cause support headaches down the road.

> Could you link an example?

You can define a getter/setter in profiles.py:
 
```python
    def legacy_import_export(self) -> bool:
        return self.meta.get("legacy_import", False)

    def set_legacy_import_export(self, enabled: bool) -> None:
        self.meta["legacy_import"] = enabled
```

That can then be accessed via mw.pm.legacy_import_export()

kleinerpirat:
Closing in favor of #2233.

PR #2229, opened 2022-12-02
Cap calculated hard delay secs at again delay + 1d

RumovZ:
https://forums.ankiweb.net/t/interval-for-hard-is-longer-than-intervals-for-good-and-easy/24860


dae:
Thanks for this Rumo!

PR #2230, opened 2022-12-03
Rework spinner buttons

dae:
Move to the right, and hide on mobile.

### DIFF
--- a/ts/components/SpinBox.svelte
+++ b/ts/components/SpinBox.svelte
@@ -3,6 +3,8 @@ Copyright: Ankitects Pty Ltd and contributors
 License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 -->
 <script lang="ts">
+    import { isDesktop } from "@tslib/platform";
+
     import IconConstrain from "./IconConstrain.svelte";
     import { chevronDown, chevronUp } from "./icons";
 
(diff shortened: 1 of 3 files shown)


dae:
@kleinerpirat does this seem ok to you?

dae:
Before/after (with dev tools highlight)

IMG_1671
IMG_1670

dae:
Actually, I'm thinking it may be better just to disable these on mobile. As I think I mentioned in the past, double-tapping on one of the buttons (to decrease by 2) triggers the page zoom action, and I'm not sure that can be disabled without removing the user's ability to zoom at all.

dae:
Looks like they also cause problems for focus - when tapping on a spinbox field, the first tap reveals the increment/decrement, but doesn't focus the field/bring up the keyboard.

kleinerpirat:
Yeah, disabling them on mobile sounds best to me. As I mentioned some time ago, we should use a standard `<select>` element for mobile, because it's hard to recreate the expected behavior on a custom component.

My thought was to trigger such a hidden element programmatically and use the custom style on the surface.

dae:
@kleinerpirat I've reworked this to move the buttons to the right and hide them on mobile - does this work for you? I wasn't sure what you meant about select, since this is an input field rather than a select field.

kleinerpirat:
I'm fine with the arrows on one side. I noticed the spinboxes have a larger height than the other elements:

image

So I suggest removing this rule:

```diff
diff --git a/ts/components/SpinBox.svelte b/ts/components/SpinBox.svelte
index 3f0cb055c..dc3e3f585 100644
--- a/ts/components/SpinBox.svelte
+++ b/ts/components/SpinBox.svelte
@@ -171,10 +171,5 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         @include button.base($border: false);
         border-radius: 0;
         height: 100%;
-
-        &.increment,
-        &.decrement {
-            min-height: 2em;
-        }
     }
 </style>
```

> I wasn't sure what you meant about select

Nevermind, I confused this with the select component :)

dae:
Good catch, that was left over from my previous change :-)

dae:
I presume the hover change is non-contentious, but if you disagree with it feel free to let me know why :-)

dae:
(was there anything else you wanted to address, or am I ok to build another beta?)

kleinerpirat:
The only thing left on my todo list is the scroll area stuff... if I don't push an update to #2222 within the next one or two hours, you can assume I've given up and went to sleep.

PR #2232, opened 2022-12-03
Make button color and hover gradient fit together

kleinerpirat:
These are cherry-picked commits from #2228 + some additinal tiny CSS changes that I'd like to see in the release.

I thought since the other PR is still being discussed, I don't want to risk not having those in the next beta.

## Changes

- QComboBox and Select components are now slightly brighter than their container, which usually has `--canvas-elevated` as background color. This makes them seem more clickable. The new color also ensures a better transition to the hover effect, as it's the same value as `--button-gradient-end`.
- The QComboBox arrow has slightly more padding
- Qt buttons don't have the aggressive border change in the pressed state anymore

## Before vs After
image

### After
image

### DIFF
--- a/qt/aqt/stylesheets.py
+++ b/qt/aqt/stylesheets.py
@@ -107,13 +107,17 @@ QPushButton {{
 QPushButton,
 QTabBar::tab:!selected,
 QComboBox:!editable,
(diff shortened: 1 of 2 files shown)


dae:
Nice!

PR #2233, opened 2022-12-04
Add profile manager setting to force custom qt stylesheets

### LINKED ISSUE #2204: Provide a (perhaps hidden) option to force the fusion theme on macOS?
For users who prefer it over the standard macOS theme

kleinerpirat:
Closes #2204.

I used `custom_styles` over `fusion`, because altough the base style is Fusion, the widgets styled in `stylesheets.py` are fully custom. Changing one property requires styling every other property too:

> **Note:** With complex widgets such as [QComboBox](https://doc.qt.io/qt-6/qcombobox.html) and [QScrollBar](https://doc.qt.io/qt-6/qscrollbar.html), if one property or sub-control is customized, all the other properties or sub-controls must be customized as well.
https://doc.qt.io/qt-6/stylesheet-syntax.html

WDYT?

### DIFF
--- a/qt/aqt/profiles.py
+++ b/qt/aqt/profiles.py
@@ -546,6 +546,12 @@ create table if not exists profiles
     def set_theme(self, theme: Theme) -> None:
         self.meta["theme"] = theme.value
 
+    def force_custom_styles(self) -> bool:
+        return self.meta.get("force_custom_styles", False)
+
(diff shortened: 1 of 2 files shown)


dae:
Thanks Matthias, this is much easier to approve :-) It works fine on Mac, and will hopefully be enough to satisfy users who don't like the Mac native styling.

PR #2236, opened 2022-12-04
Allow Nested filtered decks

ShootingKing-AM:
mostly from here: https://forums.ankiweb.net/t/allow-nested-filter-decks/25089

This will give extreme flexibility to custom study schedules. 
One can also use many `search_terms` for finetuning their study, many `limits` too !

Note: Emptying the filter deck(at any level) will always revert cards to the original deck

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -107,6 +107,7 @@ Sam Penny <github.com/sam1penny>
 Yutsuten <mateus.etto@gmail.com>
 Zoom <zoomrmc+git@gmail.com>
 Stefan Kangas <stefankangas@gmail.com>
+Harsha Raghu N. <narnindi.raghu@gmail.com>
 
 ********************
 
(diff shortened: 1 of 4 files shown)


dae:
I'm afraid I don't think this is a good idea. The handling when shifting between a rescheduling and non-rescheduling deck has not been considered, and the current behaviour has been around a long time, meaning that some users have grown to depend on it (using a filtered deck to exclude items from another one).

ShootingKing-AM:
Yeah i forgot about rescheduling option!
So what's your recommendation? Fix rescheduling option somehow or implement -limit:n to limit cards from a block "()" in search string for filtered decks.

I don't think we are making any breaking changes i guess

ShootingKing-AM:
@dae ?

dae:
I'm afraid as I mentioned above, I don't think changing this is a good idea.

PR #2237, opened 2022-12-05
Close MathJax editor when Esc is pressed

### LINKED ISSUE #2235: Javascript can't capture the escape key
We currently register a shortcut handler in webview.py to take care of closing the active dialog when Esc is pressed, but this prevents the webview from being able to capture it in a keydown handler. This prevents us from doing custom actions when Esc is pressed, such as closing the MathJax editor.

https://forums.ankiweb.net/t/anki-2-1-55-beta-mathjax-preview/23529/18

abdnh:
Closes #2235

Maybe the field should be refocused after the editor is closed. I didn't look into that in this PR.

### DIFF
--- a/qt/aqt/data/web/js/webview.ts
+++ b/qt/aqt/data/web/js/webview.ts
@@ -1,9 +1,11 @@
 /* Copyright: Ankitects Pty Ltd and contributors
  * License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html */
 
-// prevent backspace key from going back a page
 document.addEventListener("keydown", function(evt: KeyboardEvent) {
-    if (evt.keyCode !== 8) {
+    if (evt.keyCode === 27) {
+        pycmd("close");
+    } // prevent backspace key from going back a page
+    else if (evt.keyCode !== 8) {
         return;
     }
     let isText = 0;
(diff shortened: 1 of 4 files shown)


dae:
Thanks Abdo, this seems to work well!

dae:
Deck options can no longer be closed with Esc - was that intentional?

dae:
Ah, it's a bit more subtle than that, and reminds me why onEsc was originally introduced. Focus starts on the window, so pressing Esc immediately works. But if you click in the webview to focus something, then esc can no longer be used to close the dialog. It also affects the card info screen, graphs, and so on.

abdnh:
Oh, I have a bigger problem here: the deck options screen doesn't even open for me. This is not caused by my change and I just found about it while trying to properly test this. The console shows Anki is making the `/_anki` requests but nothing is showing up.

dae:
Working on a Windows machine here. Is the window failing to appear, or is the window appearing blank? If the former, does commenting out restoreGeom() help? If the latter, if you view the console with Chrome does it reveal any errors?

dae:
One of these days we'll get around to adding a 'reset window positions' option to the preferences. If you use different monitors I guess the window was off-screen; if you don't, maybe it was an invalid size or minimized?

abdnh:
I don't use different monitors. My guess is that it has something to do with some programs I use that tweak the system tray and the taskbar (RBTray and 7+ Taskbar Tweaker).

PR #2239, opened 2022-12-05
Improve Select component and add it back to Change Notetype screen

kleinerpirat:
Re: https://github.com/ankitects/anki/commit/65fd1a677248d66d28c307ee6336cd72ab0ff666#commitcomment-92030676

The issues that made you revert to the `<select>` element stemmed from a misunderstanding about the variable names `oldIdx` and `newIdx` in `MapperRow.svelte`. I thought they were referring to index states before and after a change event - but they are just referring to the old/new notetype.

The Change Notetype screen works according to my tests, but please test it yourself too, to be sure.

---
In addition I added a highlight for the `.active` class of `DropdownItem`:

image


dae:
The check notetypes screen seems to work correctly now, which is good. The currently-active marker looks good. It's subtle and a bit easy to miss, but maybe that's not a problem.

There is an error printed to the console every time you select an option in a dropdown:

JS error /_anki/pages/change-notetype.js:19188 Uncaught TypeError: Cannot read properties of null (reading 'scrollIntoView')

kleinerpirat:
Fixed the error caused by referencing a removed element in a `setTimeout` callback.

dae:
Thanks Matthias!

I'm guessing we no longer need the setTimeout - do you happen to remember why you added it in the first place?

kleinerpirat:
I don't remember, sadly. Maybe it was a try to get the smooth scroll working, but it didn't work regardless and I probably just forgot to remove it.

dae:
This has caused a minor regression: the text when pressing the Fx button now wraps.


https://user-images.githubusercontent.com/180542/206599977-fae25b39-8690-49fd-bb79-95d18d0256df.mp4

PR #2241, opened 2022-12-06
Persist field states with SessionOptions object

### LINKED ISSUE #2200: All fields reopen after adding a card
Running on dc4ac1da6630a90ebf506235eef1c2df0c1192fa

How to reproduce:
* open a "new note"
* close a field
* fill other field so that the note can be created
* click on "add" to add the note

Expected result:
* seing the window allowing to add a new card. Same fields open/closed

Actual result:
* all fields are now open

I'm quite certain it's a new behavior. I see the interest when you change note type. But, for use case such as mine, it makes it hard to use, because, depending on the deck, I'll have some fields I never use and opening them again makes hard to find the one I actually want to use

kleinerpirat:
Closes #2200.

Following @hgiesel's suggestion in [Default Editor settings instead of persistent](https://forums.ankiweb.net/t/default-editor-settings-instead-of-persistent/12761), I implemented a `SessionOptions` map that stores an `EditorOptions` object for each new notetype that gets loaded.

If there's no entry for the current mid (=`notetypeId`) yet, the defaults are loaded.

---
### Known issue:
- open HTML editor gets carried over to different notetype


ts/editor/NoteEditor.svelte
```diff
@@ -120,17 +135,24 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     let fieldsCollapsed: boolean[] = [];
     export function setCollapsed(fs: boolean[]): void {
-        fieldsCollapsed = fs;
```
dae, line 123: It seems these routines are only ever called with {} at the moment. Presumably the thinking at the time was that the current states will be passed in to these routines. Adding separate state persisted by the JS editor makes things more confusing - I think we might be better off either using these routines as intended (passing the state in instead of {}), or changing the routines to more clearly reflect what they do (eg something like updateCollapsedStateForNotetype(ntid))

kleinerpirat: I'm sorry, but I don't understand what you mean by "called with {}". You mean only from the Python side?

> Adding separate state persisted by the JS editor makes things more confusing

How else would we persist the states in a webview session?

dae: When I grep for `setCollapsed`, all I can find is `setCollapsed({});`, so we're passing an object to a routine that's typed as accepting an array of booleans. Is it used somewhere else that I've missed?

The Python code could potentially poll for the current state before replacing the note, but renaming the routine sounds like it would be cleaner.

kleinerpirat: That `{}` is syntax for a formatted string and get's replaced with:

https://github.com/ankitects/anki/blob/a702641b589ef22b66e8e4a801bd7db435e1e65f/qt/aqt/editor.py#L545

kleinerpirat: > The Python code could potentially poll for the current state before replacing the note

Yeah, it could. But that would require a lot more logic (back-and-forth messaging) than just keeping track on the JS side. Currently, Python is oblivious to the current state of the webview.

kleinerpirat: > updateCollapsedStateForNotetype(ntid)

... would not reflect what the routine does, though. The functions `setCollapsed`, `setPlainTexts` etc. set the defaults defined in the Fields dialog. If they did in fact update the collapsed state like above-quoted name suggests, we wouldn't have session persistence.

dae: Can we start again? :-) My confusion over the format strings seems to have made things unclear. Have I read the code correctly that it uses the provided `fs` argument when initially setting the state, and then ignores the provided argument when called again if the state has been persisted for that notetype? If so, then I still think the method signature could be clearer - for example, how about renaming fs to initialStateForNotetype?

kleinerpirat: Yes, exactly :) I can update the variable names in those routines of course (tbh I don't know what `fs` stands for).

dae: Yeah, I'm guessing "fieldStates".

dae:
Thanks Matthias!

PR #2243, opened 2022-12-07
Make SpinBox chevrons more subtle

kleinerpirat:
... and keep them visible when the input has focus.

## Before
image

## After
image

They're closer together now, but since they're hidden on mobile anyway, this should not be an accessibility issue.


dae:
Styling changes look good. I presume you saw my rationale for removing the focus styling on https://github.com/ankitects/anki/commit/3d8c7f5ea15aa0a89655264bd73420770debc807 and disagree?

kleinerpirat:
Ah yes.. Originally I used `focus-within` to make the component viable for mobile - but that's no longer an argument to keep it.

But it still has its purpose for keyboard-only use. When someone tabs into a spinbox and sees the arrows appear, they'll probably intuitively know that the value can be changed with the arrow keys.

kleinerpirat:
Thanks, Damien! Then let's stick to the default Chromium behaviour.

Due to the time difference I'm always tempted to push changes quickly so I don't keep you up too long. Also, I trust your opinion a lot.

dae:
No problem, I'll get to this soon - the webview flicker issue turned into a bit of a yak shave. Apart from your pending PRs is there anything you wanted to get done, or am I ok to make a new beta after they're merged?

dae:
They feel so polished now :-)

PR #2244, opened 2022-12-08
Fix some more RTL issues

kleinerpirat:
Fixes points 2 and 3 of https://forums.ankiweb.net/t/anki-2-1-55-beta-5-6/24870/102

The point about the editor gear menu in https://forums.ankiweb.net/t/anki-2-1-55-beta-5-6/24870/90 has to wait a bit, though. Removing the hard-coded `dir="ltr"` from `ButtonGroup` means all buttons in the editor toolbar and elsewhere are rounded on the wrong side. This will be fixed with a continuation of #2069 that I'm working on.

### DIFF
--- a/pylib/anki/stats.py
+++ b/pylib/anki/stats.py
@@ -63,7 +63,7 @@ class CardStats:
         self.txt += self.makeLine(k, v)
 
     def makeLine(self, k: str, v: str | int) -> str:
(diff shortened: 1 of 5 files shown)


dae:
Thanks Matthias! The gear menu is not a serious issue, so it should be fine to punt until after 2.1.55.

PR #2245, opened 2022-12-08
Replace pct-str with percent_encoding crate

RumovZ:
Whereas pct-str seems to encode everything possible and use opaque Unicode code point ranges to do so, percent_encoding only has a single predefined encoding set and encourages you to define your own, pinpoint set.
I experimented with the characters listed on https://url.spec.whatwg.org/#fragment-percent-encode-set (at least the ones Windows lets me put in a filename) and arrived at the set below. Maybe unlike me you're actually knowledgable in this area and know if we indeed only need the query and fragment sets.

### DIFF
--- a/.deny.toml
+++ b/.deny.toml
@@ -54,10 +54,6 @@ ignore = true
 unknown-registry = "warn"
 unknown-git = "warn"
 allow-registry = ["https://github.com/rust-lang/crates.io-index"]
-allow-git = [
-  # https://github.com/timothee-haudebourg/pct-str/issues/5
-  "https://github.com/timothee-haudebourg/pct-str.git",
-]
 
 [sources.allow-org]
 github = ["ankitects"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -108,7 +108,7 @@ dependencies = [
  "num_cpus",
  "num_enum",
  "once_cell",
- "pct-str",
+ "percent-encoding",
  "pin-project",
  "prost",
  "prost-build",
@@ -141,7 +141,7 @@ dependencies = [
  "which",
  "workspace-hack",
  "zip",
- "zstd 0.12.0+zstd.1.5.2",
+ "zstd 0.12.1+zstd.1.5.2",
 ]
 
 [[package]]
@@ -210,7 +210,7 @@ dependencies = [
  "workspace-hack",
  "xz2",
  "zip",
- "zstd 0.12.0+zstd.1.5.2",
+ "zstd 0.12.1+zstd.1.5.2",
 ]
 
 [[package]]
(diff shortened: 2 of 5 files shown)


dae:
Hmm, I don't think it was encoding everything? We were using 'Segment': https://github.com/timothee-haudebourg/pct-str/blob/4adccd8d4a222ab2672350a102f06ae832a0572d/src/lib.rs#L714. The stable release had less features though, which may be where the confusion lies.

I think we may have actually hit the opposite problem here - it looks like percent_encoding encodes any non-Latin text. That's the correct approach for a URI, but not for an IRI ("international"). The original motivation to moving to pct-str from the Python implementation was to be able to support IRIs. When Henrik introduced the new HTML editor, it showed the percent-encoded filenames, so users with non-latin text in their filenames were getting a bunch of percent-escapes instead of something readable.  Sorry for not making this clearer up front :-)

I guess we have two options:

- Stick with pct-str for now.
- Implement our own encoder. percent_encoding's one operates on bytes. If we wrote our own that operated on str.chars() instead, we could rewrite ascii chars that matched our asciiset, and leave codepoints above the ascii set alone. It would hopefully be fairly simple: https://docs.rs/percent-encoding/2.2.0/src/percent_encoding/lib.rs.html#253

Either is a valid choice, as there's no pressing need to migrate away from pct-str. How would you like to proceed?

RumovZ:
Oh, I see. 
Sorry for the confusion, I meant pct-str *can* encode everything you could wish for.
I agree that writing an own function should be simple and would vote to proceed with this, now that we've already started looking into it. I think I'll only need to figure out how to match the right bits. Could we even do without supporting control chars? Then it should really be trivial.

dae:
I was thinking we might want to keep using the percent_encoding for its AsciiSet functionality and for decoding, as then all we really need to do is map each unicode char to a u8 if it's in the ASCII range and then test if it's in the set. We probably want to be more liberal with what we decode than what we encode, so an off-the-box solution there would make sense. Presumably if we used AsciiSet, we'd get control character handling for free?

RumovZ:
Hm, makes sense for decoding, but I'm not sure how we would use AsciiSet for encoding, as its field is private.
Maybe we could write a wrapper for the PercentEncode iterator that yields non-ascii chars, but uses percent_encoding for ascii. However, this would probably be a lot of overhead if ascii and non-ascii chars alternate often.

dae:
Dang, didn't realise it was private. It's a tiny library; I'd be happy to copy the required parts and vendor the resulting source into our repo if that works for you (maybe as a separate local crate).

RumovZ:
Sounds good. I'll look into making it a local crate.

PR #2246, opened 2022-12-08
Fix popover animation causing placement switching loop

kleinerpirat:
Should fix https://forums.ankiweb.net/t/mathjax-flash-error/25231.

This uses the initially true `hidden` variable as a guard to prevent a loop, which was probably caused by asynchronous loading of the CodeMirror instance.


dae:
Seems to work well - thanks for the quick fix!

Issue #2121, opened 2022-10-09
Can't (easily) change colors of old cards (PC)


Redysz:
Once some colors are used in a card, changing them later is difficult. It is not enough to select the word, but you have to select the word and the surrounding spaces/brackets, which color you do not necessarily want to change.
![Untitled](https://user-images.githubusercontent.com/25427869/194773807-7ca3dff3-f1c9-4fcf-9203-58d2af3ac15f.gif)

    
    Anki 2.1.54 (b6a7760c) Python 3.9.7 Qt 6.3.1 PyQt 6.3.1
    Platform: Windows 10
    Flags: frz=True ao=True sv=2
    Add-ons, last update check: 2022-10-09 19:14:54
    
    
    ===Add-ons (active)===
    (add-on provided name [Add-on folder, installed at, version, is config changed])
    '' ['BezAkcentow', 0, 'None', '']
    BetterSearch ['1052724801', 2022-01-25T23:25, 'None', '']
    The KING of Button Add-ons ['374005964', 2020-04-18T00:33, 'None', mod]
    
    ===IDs of active AnkiWeb add-ons===
    1052724801 374005964
    
    ===Add-ons (inactive)===
    (add-on provided name [Add-on folder, installed at, version, is config changed])

dae:
I can't seem to reproduce this. Please try the latest beta version, and post back if the issue persists for you.

Redysz:
Hi @dae Thank you for your reply!

It persists on the latest beta as well. It's not easy to reproduce since it doesn't happen on all cards and the card had to be saved firstly and you need to try change a color after a while. Also the problem occurs only when you want to change the color near to punctuation without it or a space (just a word) and the color that is (and doesn't want to be changed) was set with space.

1. " some_text."  <--- color is set for " some_text" (with a space)
2. save it, restart anki. Probably update anki version meanwhile.
3. you want to change a color of the word "some_text" (without space) because it is default behavior of double click
4. now changing the color doesn't work (sadly: sometimes)

Video:

https://user-images.githubusercontent.com/25427869/205491925-ac8ebe63-664b-4a5d-8898-e0d0588f7709.mp4



What I discovered is: if I can't change a color in one card, the situation for that card won't change until I finally change the color  with surrounding space and/or punctuation. It means that the problem is in the card, not changing-color-feature (I guess). I have that card and because I didn't change the color I could "somehow" share this card with you if you tell me how.

dae:
[...]

PR #2247, opened 2022-12-08
Fix overwrite surround

hgiesel:
Related: https://github.com/ankitects/anki/issues/2121.

The faulty behavior was actually documented there:

> Assumes that there are no matching ancestor elements above
`range.commonAncestorContainer`. Make sure that the range is not placed
inside the format before using this.

However I don't blame you for not understanding it :) Unfortunately I didn't add why I chose to do it like that. I _think_ I simply had no idea which behavior was the correct one, and opted to just documented it instead.

I also have a suggestion: I could take the whole surrounding logic, and simply turn it into a npm library with a BSD-3 license. Maybe I could even get some other people interested, because of its very general-purpose functionality similiar to Rangy.js. I think most Anki contributors would be intimidated by this feature anyway or it'll take them a long time to get into the code. WDYT?


dae:
Thanks for the fix Henrik! There was a similar issue in 1a444f30e9b91aaebafbe341c604f0d48777841b that you may feel should be done differently.

No objections to you publishing a copy of the library if that's something you're interested in and feel would benefit others, but I'd prefer to keep a copy inside Anki's tree as well for now, as if any changes do need to be made in the future, having it in the same repo is less friction than needing to submit fixes upstream or make use of patch-package.

PR #2248, opened 2022-12-08
Refine layout of CSV Import, Graphs, Deck Options and Change Notetype

kleinerpirat:
... using the new components `ScrollArea` and `Page` in conjunction with CSS grid.

Sorry this took so long. I didn't have the time to separate this into multiple commits, but most changes are small design tweaks.

# Notable changes
## Rework Import CSV layout
Like shown in #2222:

image

## Rework Change Notetype screen
Instead of sticky headers, it now uses a responsive grid layout that is much more stable.
  
image
image

## Scroll shadow effect for all users of `ScrollArea`
namely the editor fields area and all screens mentioned above

image

## Fullscreen Toggle for Graphs

Can be triggered either via the badge or by double-clicking anywhere on the graph container.

image

## Fixes
- Fix header of graphs page disappearing after a certain scroll value
- Fix header button misalignments in Change Notetype and Deck Options screens

The scroll inconsistencies of #2222 have been fixed with adjustments to `overflow-behavior` within `ScrollArea`.


kleinerpirat:
I've run into a weird situation with `icons.ts`.

`./ninja format` and `./ninja fix` seem to have different opinions regarding the import sort order. The check fails no matter which order I choose.

dae:
There are some nice improvements here, but it's quite a large change, and we're late in the beta testing cycle, so I'm afraid I think we should probably wait until after 2.1.55 is released to merge this.

Some initial thoughts:

- The icon that expands a graph to the full screen is an interesting idea, but needs a bit of polish. Some of the graphs look a bit messy with a large height, or they overflow the box behind them:

2022-12-09T10:35:36,878260270+10:00
2022-12-09T10:36:03,406196364+10:00

- Because popovers don't share the same area as the "main" content anymore, it can lead to somewhat confusing behaviour with the scrollbars if you scroll down to the bottom graph, then hover over the graph so the popover extends off the screen, causing the outer area to resize instead of the inner area.

https://user-images.githubusercontent.com/180542/206596735-fbd64516-6d6d-48e9-a900-ed8527a8160a.mp4

- The bottom of the deck options and graphs screens are truncated on iOS, as mentioned on https://github.com/ankitects/anki/pull/2222#issuecomment-1331760301. Possibly related: when maximizing a graph and scrolling down, the top bar scrolls off the screen.

kleinerpirat:
> The bottom of the deck options and graphs screens are truncated on iOS

I couldn't connect to the local server with the iPad, sadly...

> Possibly related: when maximizing a graph and scrolling down, the top bar scrolls off the screen.

Is this also meant for iOS? I don't see that issue on Desktop.

---
Since the issues you mentioned are related to the fullscreen graphs, I'd disable that for now, but maybe we can think about incorporating the other things? There are lots of little improvements in this PR that would be really cool to have in the release.

I'm up all night for other work, so I'd be available for a quick feedback loop.

dae:
> I couldn't connect to the local server with the iPad, sadly...

Did you run ts-run with ANKI_API_HOST=0.0.0.0, and use http://x.x.x.x with your computer's IP address?

> Is this also meant for iOS? 

Desktop. You may need a short window size.

>  but maybe we can think about incorporating the other things?

The mobile issues are a blocker - both AnkiMobile and AnkiDroid are waiting on a desktop update (eg https://github.com/ankidroid/Anki-Android-Backend/pull/246#issuecomment-1331668085), and changes that cause issues for the ts pages will prevent us from updating to the new version.

Please also bear in mind that every improvement we add could potentially introduce new problems accidentally, and the larger the change, the more likely that becomes. This release has already been pushed back a considerable amount, and I'm afraid I need to start being stricter about accepting changes if we're ever to get it out.

kleinerpirat:
> Did you run ts-run with ANKI_API_HOST=0.0.0.0, and use http://x.x.x.x/ with your computer's IP address?

Yeah, but Safari loaded for a while and then displayed this message (translated from german): "Safari couldn't open the page, because the server doesn't respond anymore."

kleinerpirat:
Ah, nevermind Damien. It was a firewall issue. I disabled it and now it works on both devices.

kleinerpirat:
> This release has already been pushed back a considerable amount, and I'm afraid I need to start being stricter about accepting changes if we're ever to get it out.

I understand and I'm sorry about that long delay. Beta testers have been very strict too (which is a good thing, but also a bit nerve-wracking). It's up to you. I can only promise that this is the last project I had in the pipeline for this release :-)

> The mobile issues are a blocker

I could reproduce the bottom of the page getting cut off reliably on Android. On iOS it was only at the start and then it displayed fine after scrolling up and down.

Does https://github.com/ankitects/anki/pull/2248/commits/ba6efc734458aeaa9dd1c3781d2d14a127995524 fix it for you? It did for me on Android. On the iPad I'm not sure if `isDesktop()` is actually returning false, because I don't see the extra margin I added.

Is there anything else blocking?

kleinerpirat:
Alright :)

It's nothing critical anyway, just some polish that would have made me feel better about the release. But not everyone is as perfectionist as me, so it'll probably be fine.

dae:
Perhaps relevant: https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good :-)

I really appreciate all the time and effort you've put into both making these improvements and then polishing them/addressing regressions Matthias, but I'm sorry you've felt a bit of pressure to do so. Take as long as you need - Anki will be here when you feel like contributing again, and you've certainly earned the break :-)

dae:

PR #2252, opened 2022-12-09
Fix dropdown items wrapping

kleinerpirat:
Re: https://github.com/ankitects/anki/pull/2239#issuecomment-1343702450

When I removed `bootstrap/scss/dropdown` in #2239, two important styles got lost: `padding` and `white-space`. This PR adds those back.


dae:
Thanks for the super quick fix :-)

PR #2253, opened 2022-12-09
Fix field labels not sticking (again)

kleinerpirat:
I think this time it was me who broke it: https://github.com/ankitects/anki/pull/2220/commits/6664bf893dd4ff9d1b7c96c88792906ff48302b2.

### DIFF
--- a/ts/editor/EditorField.svelte
+++ b/ts/editor/EditorField.svelte
@@ -124,7 +124,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     .field-container {
         background: var(--canvas);
         border-radius: var(--border-radius);
-        overflow: hidden;
     }
 
     .editor-field {
--- a/ts/editor/Fields.svelte
+++ b/ts/editor/Fields.svelte
@@ -23,6 +23,7 @@ Contains the fields. This contains the scrollable area.
         /* Move the scrollbar for the NoteEditor into this element */
         position: relative;
         overflow-y: auto;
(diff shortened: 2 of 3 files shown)


dae:
I noticed a few problems with this:

- when scrolling with the mouse wheel, it makes it look like the top field is collapsed
- the sticky field overlaps the field content, so when shift-tabbing you end up with the cursor being hidden

No rush on a fix - I think we can probably punt this one to post 2.1.55 as well.

https://user-images.githubusercontent.com/180542/206621934-9938fb36-e90f-488c-ba44-f907990a120b.mp4

kleinerpirat:
> when scrolling with the mouse wheel, it makes it look like the top field is collapsed

This is a drawback of the approach in general. I don't know how I would fix this, because `position: sticky` is a bit limited in terms of customizability. AFAIK there's no such thing as a `sticky-offset` to change the point where it ceases to stick.

In my opinion the advantage of always seeing the field name and being able to collapse the field no matter how large the content outweighs that issue. Unwieldy HTML content was the motivation for it:

## With Sticky Labels
image

## Without
image

dae:
While I take your point, I think the ability to see the text you're editing is more important than seeing the currently active field :-) We could potentially achieve both in the future by checking the position of the caret and scrolling the webview up or down as necessary if it's too close to the screen margins. AnkiMobile does something like that, but I've found it to be a source of bugs in the past, as webview updates changed previously-working behavior.

dae:

PR #2255, opened 2022-12-09
Replace pct-str with local percent_encoding crate

RumovZ:
Replacement for #2245.

### DIFF
--- a/rslib/ascii_percent_encoding/Cargo.toml
+++ b/rslib/ascii_percent_encoding/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "ascii_percent_encoding"
+publish = false
+description = "Like percent_encoding, but does not encode non-ASCII characters."
+
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+
+authors = ["The rust-url developers"]
+license = "MIT OR Apache-2.0"
+
+[lib]
+name = "ascii_percent_encoding"
+path = "src/lib.rs"
+
+[features]
+default = ["alloc"]
+alloc = []
--- a/rslib/ascii_percent_encoding/src/lib.rs
+++ b/rslib/ascii_percent_encoding/src/lib.rs
@@ -1,0 +1,468 @@
+// Copyright 2013-2016 The rust-url developers.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! URLs use special characters to indicate the parts of the request.
+//! For example, a `?` question mark marks the end of a path and the start of a query string.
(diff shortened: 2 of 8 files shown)


rslib/ascii_percent_encoding/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+    /// Not used for non-ASCII bytes.
+    const fn contains(&self, byte: u8) -> bool {
+        let chunk = self.mask[byte as usize / BITS_PER_CHUNK];
+        let mask = 1 << (byte as usize % BITS_PER_CHUNK);
+        (chunk & mask) != 0
+    }
+
+    fn should_percent_encode(&self, byte: u8) -> bool {
+        // MODIFIED
+        byte.is_ascii() && self.contains(byte)
```
RumovZ, line 78: I copied https://github.com/servo/rust-url/blob/master/percent_encoding/src/lib.rs and only modified this line. This gives the smallest diff, but I can also remove redundant stuff if you prefer.

RumovZ: * And added the file header...

dae:
Thanks Rumo. Would it be possible to split it into 3 commits? If the first one adds the crate as-is, second one adds our local changes, and then third switches over from pct-str, that should make it easier to track changes to the crate in the future. Also because this is mainly someone else's work, I think we should keep their copyright header, and add an exclusion to copyright_headers.py instead of adding our own.

dae:
(I'll then do a non-squash merge for this, so ninja check will need to pass for each of the commits)

rslib/ascii_percent_encoding/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "ascii_percent_encoding"
+publish = false
+description = "Like percent_encoding, but does not encode non-ASCII characters."
+
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
```
dae, line 9: We should probably define authors and license using the original crate's values rather than using our own names/license here.

RumovZ:
This should be it then!

rslib/ascii_percent_encoding/src/lib.rs
```diff
@@ -82,7 +82,8 @@ impl AsciiSet {
     }
 
     fn should_percent_encode(&self, byte: u8) -> bool {
-        !byte.is_ascii() || self.contains(byte)
+        // MODIFIED
+        byte.is_ascii() && self.contains(byte)
```
dae, line 86: I'm afraid this doesn't look correct. If a unicode char is converted to utf8, some of the individual bytes will be below 128. Eg the last one in this kanji:

u"今".encode("utf8")
b'\xe4\xbb\x8a'

Instead of checking the bytes of the utf8, we need to operate on the unicode chars, and focus on only chars which are <128. Effectively string.chars().map(|c| if c.is_ascii { percent_escape(...) }).collect::<String>()

A unit test that checks a few test cases might be a good idea here :-)

RumovZ: Yeah, UTF-8 is great. 🙂

dae (approved):

PR #2257, opened 2022-12-09
Remove dark theme exclusive inset shadow from FloatingArrow

kleinerpirat:
I must have been in light mode when I changed the popover look, so I didn't notice this at first:

## Before
image

## After
image

General note: We should be able to safely remove most uses of `$pageTheme.isDark` because all colors are defined in `_vars.scss` now.


dae:
Thanks Matthias. 

> We should be able to safely remove most uses of $pageTheme.isDark

Feel free to create an issue noting the work that needs to be done if you think it's a good idea. :-)

kleinerpirat:
> Feel free to create an issue

It's not a high priority to do it all at once. I just mentioned it in case you or others stumble upon this in other screens - as a best practice.

PR #2259, opened 2022-12-09
Do not show description button in overview if deck is filtered

kleinerpirat:
... because in that case it's not supposed to do anything.

https://forums.ankiweb.net/t/anki-2-1-55-beta-7-rc1/25130/14

### DIFF
--- a/qt/aqt/overview.py
+++ b/qt/aqt/overview.py
@@ -8,10 +8,12 @@ from typing import Any, Callable
 import aqt
 import aqt.operations
 from anki.collection import OpChanges
+from anki.decks import DeckDict
 from anki.scheduler import UnburyDeck
 from aqt import gui_hooks
 from aqt.deckdescription import DeckDescriptionDialog
 from aqt.deckoptions import display_options_for_deck
+from aqt.operations import QueryOp
 from aqt.operations.scheduling import (
     empty_filtered_deck,
     rebuild_filtered_deck,
@@ -285,7 +287,17 @@ class Overview:
             # links.append(["F", "cram", _("Filter/Cram")])
         if self.mw.col.sched.have_buried():
             links.append(["U", "unbury", tr.studying_unbury()])
-        links.append(["", "description", tr.scheduling_description()])
+
+        QueryOp(
+            parent=self.mw,
+            op=lambda col: col.decks.current(),
+            success=lambda deck: self._check_deck_and_draw(deck, links),
+        ).run_in_background()
+
+    def _check_deck_and_draw(self, deck: DeckDict, links) -> None:
+        if not deck["dyn"]:
+            links.append(["", "description", tr.scheduling_description()])
+
         link_handler = gui_hooks.overview_will_render_bottom(
             self._linkHandler,
             links,
--- a/qt/aqt/overview.py
+++ b/qt/aqt/overview.py
@@ -65,6 +65,22 @@ class Overview:
     def refresh(self) -> None:
         self._refresh_needed = False
         self.mw.col.reset()
+
+        def fetch_data(
+            col: Collection,
+        ) -> Tuple[bool, bool]:
+            is_dyn = col.decks.current()["dyn"]
+            have_buried = col.sched.have_buried()
+            return (is_dyn, have_buried)
+
+        QueryOp(
+            parent=self.mw,
+            op=fetch_data,
+            success=lambda data: self._render(data),
+        ).run_in_background()
+
+    def _render(self, data: Tuple[bool, bool]) -> None:
+        self._is_dyn, self._have_buried = data
         self._renderPage()
         self._renderBottom()
         self.mw.web.setFocus()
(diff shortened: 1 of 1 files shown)


qt/aqt/overview.py
```diff
@@ -285,7 +287,17 @@ def _renderBottom(self) -> None:
             # links.append(["F", "cram", _("Filter/Cram")])
         if self.mw.col.sched.have_buried():
             links.append(["U", "unbury", tr.studying_unbury()])
-        links.append(["", "description", tr.scheduling_description()])
+
+        QueryOp(
+            parent=self.mw,
+            op=lambda col: col.decks.current(),
+            success=lambda deck: self._check_deck_and_draw(deck, links),
+        ).run_in_background()
```
dae, line 295: QueryOp is the best way to solve this, but there are two other existing DB queries in the same function (have_buried(), and another dyn check). Rather than add an extra query, I think we either need to a) assign the dyn check above to a variable and reuse it (pros: minimum code change before a release, cons: keeps accessing data on the main thread), or b) rewrite the whole routine so the QueryOp fetches all the needed data up-front.

dae: (thank you for the fix by the way :-)

kleinerpirat:
Like this?

qt/aqt/overview.py
```diff
@@ -63,6 +64,22 @@ def show(self) -> None:
...
+            have_buried = col.sched.have_buried()
+            return (is_dyn, have_buried)
+
+        QueryOp(
+            parent=self.mw,
+            op=fetch_data,
+            success=self._render,
+        ).run_in_background()
+
+    def _render(self, data: Tuple[bool, bool]) -> None:
```
dae: This is not wrong and doesn't need to be changed, but just a note for next time: in recent Pythons we can use tuple[bool, bool] instead, which saves having to import Tuple.

kleinerpirat:
I noticed a lot of other calls of `self.mw.col.xxxx` in the file.

Should I reduce the amount of queries, but keep a single query inside each method, or should I really query all of them in advance and store them in `self`?

qt/aqt/overview.py
```diff
@@ -63,6 +65,23 @@ def show(self) -> None:
...
+            return (deck, deck_id, have_buried)
+
+        QueryOp(
+            parent=self.mw,
+            op=fetch_data,
+            success=self._render,
+        ).run_in_background()
+
+    def _render(self, data: Tuple[DeckDict, DeckId, bool]) -> None:
+        self._current_deck, self._deck_id, self._have_buried = data
```
dae, line 84: Ideally we'd pass the fetched state into the functions that need it, instead of storing it as attributes on self. But if we changed the signature of functions like _renderBottom(), we'd break some add-ons that are overwriting it :-( So storing it as hidden properties seems reasonable.

I was about to point out the separate deck query in _renderPage(), but it looks like you already caught it. Had I noticed that at the start, I would have more strongly pushed for the do-the-bare-minimum approach, as a minor bugfix has turned into a bit of a refactor which we probably don't need right now :-) How about we put this PR on hold until 2.1.55 is out the door, and apply the following fix in the mean time?

```patch
diff --git a/qt/aqt/overview.py b/qt/aqt/overview.py
index 8d79c0034..9d64b0db6 100644
--- a/qt/aqt/overview.py
+++ b/qt/aqt/overview.py
@@ -277,7 +277,8 @@ class Overview:
         links = [
             ["O", "opts", tr.actions_options()],
         ]
-        if self.mw.col.decks.current()["dyn"]:
+        is_dyn = self.mw.col.decks.current()["dyn"]
+        if is_dyn:
             links.append(["R", "refresh", tr.actions_rebuild()])
             links.append(["E", "empty", tr.studying_empty()])
         else:
@@ -285,7 +286,8 @@ class Overview:
             # links.append(["F", "cram", _("Filter/Cram")])
         if self.mw.col.sched.have_buried():
             links.append(["U", "unbury", tr.studying_unbury()])
-        links.append(["", "description", tr.scheduling_description()])
+        if not is_dyn:
+            links.append(["", "description", tr.scheduling_description()])
         link_handler = gui_hooks.overview_will_render_bottom(
             self._linkHandler,
             links,
```

Regarding fetch-everything-at-once vs lots-of-little-queries, each background op has a slight performance overhead, and a code overhead as a function has to be split into fetch and do-after-fetch steps. That makes it more efficient to do everything in one go.

dae: (if there are still more mw.col calls that haven't been handled, another option is to just abandon this PR for now. This is part of the codebase we'll eventually want to move into a ts page, so it's not worth spending a lot of effort into improving it as it will be obsolete in the future)

kleinerpirat: Yeah, I'll create a new PR with the small fix and abandon this one for now.

dae: Thanks Matthias, sorry for not being clearer from the start.

kleinerpirat:
Superceded by #2266.

qt/aqt/overview.py
kleinerpirat: No worries. I learned some stuff about QueryOp, so it wasn't wasted time :)

PR #2262, opened 2022-12-10
Auto-hide toolbar in Reviewer

kleinerpirat:
I know, perfect is the enemy of good and all...

---

# Make reviewer toolbar behave like a dock

[Screencast from 2022-12-11 19-33-30.webm](https://user-images.githubusercontent.com/62722460/206922094-cc974fe2-65f8-472f-93e4-28d82a9032d6.webm)


kleinerpirat:
Even though it works as intended, I'll keep this as a draft until I've moved all the logic where it belongs. Right now it's all over the place, mainly because the `EventFilter` has to be done on the webview. Perhaps I can use inheritance to add specific filtering to the main and toolbar webviews.

It cannot be completely encapsulated in the toolbar, because the `Leave` and `Enter` events for the main webview are more suitable than the counterparts on the toolbar for this job.

dae:
I'm not 100% enthusiastic about this change, as the the multiple-webviews-in-the-main-window is something I think we should move away from in the future, and this PR is adding more code to that old path. If you'd asked me before starting work on it, I would have suggested not working on this at the moment.

That said, you have come up with something that seems to work fairly well. I think it probably needs an option to turn off the behaviour, as some users will prefer the static menus, and that should probably even be the default, so new users don't need to go hunting for the menus. Having an option will also give an escape hatch to users with add-ons that break because of this change.

kleinerpirat:
As you know, I'm also eager to unify the main webviews, but I figured it will take a while until we get there. So I tried this, saw it works nicely and thought it can't hurt if we bridge the gap with this.

Should I add an extra option or do you think reduce motion is suitable for this?

dae:
In order to have this be opt-in instead of opt-out, we'll need a separate option, or we would need to disable all animations by default which is probably not the best idea :-)

dae:
I initially thought the 'minimize distractions' setting was accidentally inverted, because I was thinking of the animation as a distraction :-) While I think it's a good idea to try to combine settings where we can, I'm not sure we can do so here. AnkiMobile has separate settings for hiding the top bar and answer buttons/learning count, because some users want to hide one but not the others, and I suspect desktop users will be the same.

Even when the animation/hiding is disabled, it still seems to animate when Anki is first opened - was that intentional? It might possibly break some add-ons?

kleinerpirat:
As an alternative to #2287 I added two modes for the toolbar:

- `outset` : protruding design we're used to
image

- `inset`: no box-shadow, no background, just a subtle border
image

In review mode, the toolbar is inset to minimize distractions.

---
@glutanimate I experimented with your code to dynamically adjust the background color. It definitely looked better for default templates, but I found it to be too risky for cards with image- or colorful backgrounds. I think this will have to wait until we unify the webviews - then we could use a background-blur filter to achieve a milk-glass effect that looks good on any background.

dae:
I like that the transition is quite subtle - it gets out of the way without a big animation/delay, allowing the user to focus on their content. It doesn't address the "toolbar is too tall" complaint - how do you see this change fitting in with the proposal to provide an older-style appearance with an option?

Because I'd like to get a release out soon to address some issues, I think the automatic hiding should probably be pushed back until we have the time to give it a longer beta testing period than a week. It introduces a fair bit of new code, and there are probably corner cases to work through, particularly with add-ons. WDYT?

kleinerpirat:
Like Aristotelis said in the other PR, vertical space isn't such a rare resource in the reviewer. I believe the main issue with the new style was the box shadow and background color drawing too much attention.

An extra option to change the appearance doesn't seem worthwhile to me, but of course we can tweak the toolbar style according to community feedback.

kleinerpirat:
Update @glutanimate:

Following the idea of your POC code to adjust the background color, I made a function to create a background image for the toolbar. We can create a seamless transition between the webviews by offsetting the background of the reviewer by the toolbar's height. As you can see here, this will work with complex backgrounds:

image

tatsumoto-ren:
> inset: no box-shadow, no background, just a subtle border

The size of the toolbar on the screenshot is 41px which is a regression compared to 2.1.54.

kleinerpirat:
That depends on how you define "regression".

tatsumoto-ren:
> That depends on how you define "regression".

There's less space left for a card.

Although, on a second thought, changing the current theme/look isn't the best path. Better keep it as is. The new theme is made in a very specific style which is not neutral and thus frustrates some users. But from what I've seen there are people who like it. For everyone else an option in Preferences to go back to the previous (2.1.54 and older) design would be ideal. 

Being able to turn it off will also help when you run into a visual glitch that prevents widgets or parts of the UI from functioning correctly, obstructs them or makes them unusable, which is unfortunately something I've run into a lot.

dae:
It looks great in the screenshots, and I imagine people with backgrounds like that will really like it. Propagating the main window color to the top takes a second or so though, which makes the move from overview->study a bit janky, and it's particularly bad if the user has different backgrounds depending on the card:

https://user-images.githubusercontent.com/180542/210716409-85885f55-1b54-4640-b657-b0630ba89ce2.mp4

dae:
(also, this approach doesn't appear to work in the Qt 5.14 Mac builds:)

2023-01-05T16:34:14,398515353+10:00

kleinerpirat:
> Propagating the main window color to the top takes a second or so though, which makes the move from overview-study a bit janky

Yeah, that's cause I used a simple `single_shot` timer to await the card background change. Originally I thought the delay wouldn't be a big deal considering the other advantages of the PR, but it was easy to fix (https://github.com/ankitects/anki/pull/2262/commits/04ebdfb62c7d6aa08cc56b4da1a55b8e989ed52c). Initiating the background update via `bridgeCommand` from `ts` also has the advantage of respecting template input via `onUpdateHook`.

> also, this approach doesn't appear to work in the Qt 5.14 Mac builds

This should be fixed in Qt 5.15+: https://bugreports.qt.io/browse/QTBUG-82842
I can confirm the PR works fine on my Qt 5.15.2, PyQt 5.15.5 build.

glutanimate:
@kleinerpirat I really like this implementation, very nicely done! +1 on using `bridgeCommand` to schedule the update. I see no noticeable delay when walking through cards, only an ever-so-slight desync between the toolbar background update and `mw` state switches, but I would say it's subtle enough not to be a problem. If we wanted to address this, I presume that triggering the update via the js→python bridge would also work for other mw states.

On that note, I very much like the elevation change between the reviewer and other mw states. Coupled with the transparent background, I really do think that this will address pretty much all concerns about the toolbar being too obtrusive when reviewing cards.

The height difference still stands, which a minimal/classical theme could address, but @tatsumoto-ren, I would give this updated design a chance and see if you don't end up liking it after all. Keeping an alternate design updated as Anki undergoes further changes seems like a not-so-insignificant maintenance cost, and if the height difference is the only remaining pain point, then perhaps we could ship an option that simply tweaks the height instead (maybe alongside the size of other UI elements like e.g. the deck list, which I've seen some users also comment on).

One change I noticed (I think it's a change?), is the deck table shifting up slightly when launching Anki:

https://user-images.githubusercontent.com/5459332/210879405-2cde6bf6-5111-4e8c-99c6-f7097cf822e2.mp4

But again, barely noticeable, and – if unintentional – smooth enough to still look good. 

I can confirm the Qt 5.14.x issues here as well on Linux, and also that Qt 5.15.x works fine.

dae:
The background updating seems to work really well now!

> This should be fixed in Qt 5.15

The Qt5 Mac build is stuck on .14 due to some regressions in .15 on macOS. We don't need to support the new functionality on that build, but we need some sort of fallback, so that users on that build don't get stuck with ugly black borders.

tatsumoto-ren:
@glutanimate

> @tatsumoto-ren, I would give this updated design a chance and see if you don't end up liking it after all.

It's okay, I'm sure there's something about it. But I've seen posts where people say that they want to be able to disable it.

I have an entire thread on the forum dedicated to flaws of this design. The toolbar is just one little part of it. Fixing so many flaws is a lot of effort so I figure it's better to choose something else instead. Wasn't it already perfect in Anki `2.1.24`? Maybe there should be at least one more theme shipped by default and a `gui_hook` to add your own themes. There's already a `gui_hook` that allows you to modify widget styles, but apparently it has no effect on the webview screens.

I also don't want to hide the toolbar because I need to use my custom buttons during review. 

> Keeping an alternate design updated as Anki undergoes further changes seems like a not-so-insignificant maintenance cost

That's why it's fair to ask why you'd want to make changes in the first place. Plus, I might be able to maintain it alone, without any extra cost required from others.

tatsumoto-ren:
I checked out  `kleinerpirat:auto-hide-toolbar` and tested it a bit.

This is the **AMBOSS** button. It obstructs some of the default buttons.

screenshot-2023-01-06-00-56-54

And here's a screencast of the animation:

https://user-images.githubusercontent.com/69171671/210908459-47db953b-140c-4371-9272-c2f6ce4ffcbb.mp4

I see a gray bar that splashes both at the bottom and on top of the card. There's also a noticeable delay before the toolbar disappears. 

And lastly, I think the margin between the times (1d, 3d, etc,) and the edge of the card was not so small. There's hardly any margin now. But the change doesn't bother me that much.

glutanimate:
> This is the AMBOSS button. It obstructs some of the default buttons.

Thanks, appreciate the report. Saw this as well a bit ago and been looking into a solution. Should be doable, but a bit more tricky as it's about unhacking the hacks that add-ons use to position themselves arbitrarily. Will post a follow-up PR that will hopefully also make these hacks unnecessary.

kleinerpirat:
> I see a gray bar that splashes both at the bottom and on top of the card.

Qt seems to have trouble with fast animations. If you set the animation duration to 2000ms, the gray background won't show. One workaround I've played around with is transitioning the toolbar via CSS and collapsing the webview separately.

> There's also a noticeable delay before the toolbar disappears.

That is an intentional delay set on `hide_timer` and can be reduced. My thought was that it might be frustrating if the toolbar disappeared too quickly when you accidentally move the mouse down a bit.

> The Qt5 Mac build is stuck on .14 due to some regressions in .15 on macOS. 

Alright, I'm working on a fallback!

---
Thanks @glutanimate and @tatsumoto-ren for looking over this, I appreciate both of your efforts on improving the toolbar.

kleinerpirat:
@dae To prevent glitchy animations, I decided to not alter the toolbar height and use a CSS transition instead. While this eliminates the advantage of gaining extra vertical space with the "Collapse toolbar" setting, it does look a whole lot cleaner. WDYT?

dae:
I feel more comfortable with this new approach, as there's less that can go wrong. The automatic copying of image backgrounds from the main window is clever, and makes it look almost as if it's a single webview. The illusion is not perfect, and the slight delay as the two different webviews refresh makes it feel a bit clunky, but perhaps the average user cares more about how it looks after that than the transition? And if we have a less fancy option that tatsumoto is pushing for, then users can pick their tradeoff.

Unfortunately I'm still seeing black borders like in my screenshot above when I use tools/run-qt5.14 - do you not see them on your machine?

kleinerpirat:
> The illusion is not perfect

I don't even think there needs to be an illusion in the first place. I would have been contempt with an average-color background too, but copying the background turned out to be easier than that. Even if the transition is clunky, there is more breathing room visually and that's what matters.

dae:
Thanks Matthias, this is looking pretty polished now! Reducing the hiding delay seems to have make it feel more snappy too. :-)

PR #2268, opened 2022-12-12
Fix MathJax popup fails to appear

### LINKED ISSUE #2250: MathJax popup fails to appear when adding MathJax via Fx button
Works correctly when using a shortcut. Appears to have regressed in #2011.

hikaru-y:
Fixes #2250

### DIFF
--- a/ts/editor/editor-toolbar/LatexButton.svelte
+++ b/ts/editor/editor-toolbar/LatexButton.svelte
@@ -94,7 +94,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     <Popover slot="floating" --popover-padding-inline="0">
         {#each dropdownItems as [callback, keyCombination, label]}
-            <DropdownItem on:click={callback}>
+            <DropdownItem on:click={() => setTimeout(callback, 100)}>
                 <span>{label}</span>
                 <span class="ms-auto ps-2 shortcut"
                     >{getPlatformString(keyCombination)}</span
--- a/ts/editor/editor-toolbar/LatexButton.svelte
+++ b/ts/editor/editor-toolbar/LatexButton.svelte
@@ -6,6 +6,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import * as tr from "@tslib/ftl";
     import { getPlatformString } from "@tslib/shortcuts";
     import { wrapInternal } from "@tslib/wrap";
+    import { tick } from "svelte";
 
     import DropdownItem from "../../components/DropdownItem.svelte";
     import IconButton from "../../components/IconButton.svelte";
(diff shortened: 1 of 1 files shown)


dae:
Thanks for looking into this Hikaru! I did some brief experimentation and it looks like we can also solve the issue with the following:

```diff
diff --git a/ts/editor/editor-toolbar/LatexButton.svelte b/ts/editor/editor-toolbar/LatexButton.svelte
index b281e9acd..f511c3b31 100644
--- a/ts/editor/editor-toolbar/LatexButton.svelte
+++ b/ts/editor/editor-toolbar/LatexButton.svelte
@@ -94,7 +94,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html

     <Popover slot="floating" --popover-padding-inline="0">
         {#each dropdownItems as [callback, keyCombination, label]}
-            <DropdownItem on:click={callback}>
+            <DropdownItem on:click={() => setTimeout(callback, 100)}>
                 <span>{label}</span>
                 <span class="ms-auto ps-2 shortcut"
                     >{getPlatformString(keyCombination)}</span
```

Do you think we could use that instead, or do you think it's a bad idea?

hikaru-y:
I don't think it's a bad idea, and I also considered using `setTimeout()` as an option.

1. Do not worry about the delay of a few dozens of milliseconds, but give priority to keeping the code simple.
2. Detect exactly when the pop-up menu disappears and eliminate as much delay as possible in command execution, even if it makes the code a bit more complicated.

Weighing these two options, I chose the latter when I first submitted this PR. On second thought, however, there is no point in worrying about the slightest delay here, so we may as well choose the former.

----
I am getting an error in Anki CI, but I can't figure out how to fix it by looking at the error message.

dae:
Thanks Hikaru! Out of curiosity, what was the problem?

hikaru-y:
After reading your comment, I added and pushed two commits with exactly the same content as a266d40c1243478219a11d3d325fbe3d59316be1 and 3532448dafcd142538129166f0638868e765f707, but I got an error in Anki CI. I don't remember exactly, but I think the message was something like `commit 123456789 could not be found in hikaru-y:fix-mathjax-popup`. I must have done something wrong when I did those two commits, but I didn't realize it at the time. After that, I pushed several times by trial and error, redoing commits, merging updates from the `main` branch, etc., and each time I got a similar error message. So I carefully read the entire error log and noticed that `FETCH_HEAD` was commit 11115f59bcbde22bbc0448bfd6b86e887a8a56a7, but the hash of the first commit of this PR at that point was different from that. So I ran the following:
`git reset --hard 11115f59bcbde22bbc0448bfd6b86e887a8a56a7`
, then added the remaining commits again, and the Anki CI error stopped appearing. Sorry for the trouble. I need to become more familiar with git operations.

dae:
No need to apologise, I'm not entirely sure what's happened there either! Glad the issue sorted itself out in any case.

PR #2274, opened 2022-12-16
Fix possible unexpected flash of pop-up menu

hikaru-y:
In some pop-up menus on the editor toolbar, the display state of the pop-up is not properly managed, which can unexpectedly cause the pop-up to flash momentarily.

### Steps to reproduce:
1. Click on one of the following icons:
    - Fx
    - Right-most icon in BlockButtons
    - Chevron down icon next to RemoveFormatButton
2. Click on an empty space outside the fields to close the pop-up
3. Click on a field

https://user-images.githubusercontent.com/47855854/208202726-f3a6c823-e0b3-4572-89b0-f8d85bad3243.mp4


dae:
Thanks Hikaru!

PR #2281, opened 2022-12-21
Use a webview to show add-on's config help

abdnh:
This allows add-ons to embed images for example. Some of @AnKingMed's add-ons has local links to social media icons in [config.md](https://raw.githubusercontent.com/AnKingMed/The-KING-of-Button-Add-ons/master/config.md) that somehow used to work in older Anki versions.

This also brings other improvements like zooming support and gives add-ons more flexibility to modify the display using the webview_will_set_content hook for example.

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -1550,6 +1550,7 @@ class ConfigEditor(QDialog):
         self.updateHelp()
         self.updateText(self.conf)
         restoreGeom(self, "addonconf")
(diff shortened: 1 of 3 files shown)


dae:
This seems like a reasonable change, but the out-of-the-box appearance could be improved. With default geom/splitter, it used to look like:

2022-12-21T12:00:00,460891978+10:00

It now looks like:

2022-12-21T12:00:39,757150983+10:00

dae:
Thanks Abdo!

PR #2287, opened 2022-12-28
Tweak toolbar styles to fix previous regressions

tatsumoto-ren:
This PR fixes a regression with the toolbar that occurred in 2.1.55.

**Main problems with the current look:**

* It has grown noticeably taller.
* There’s a gap between the end of the shadow and the upper edge of the card

notch

**New version (screencast):**

https://user-images.githubusercontent.com/69171671/209826215-75684d24-4811-4353-b02a-7d1a1a7b2e9f.mp4

The toolbar buttons behave more similarly to what was in 2.1.54 and earlier. Because add-ons often print on the toolbar, it is important to keep it compact, so the paddings were reduced. 

I have seen a PR that makes it possible to hide the toolbar. I think it's fine for someone who don't use any addons that add extra buttons to the toolbar. However, without this PR those who want the toolbar to stay visible would have to accept its intrusive look.

There are probably alternative ways to fix the issue, but I would assume they require more work. For example, the toolbar could be moved to a completely different location.

### DIFF
--- a/qt/aqt/data/web/css/toolbar.scss
+++ b/qt/aqt/data/web/css/toolbar.scss
@@ -6,25 +6,18 @@
 @use "sass/elevation" as *;
 @use "sass/button-mixins" as button;
 
-#header {
-    padding-bottom: 4px;
-    margin-top: -3px;
+#header,
+#header tbody {
+    padding: 0px;
+    margin: 0px;
+    border-spacing: 0px;
 }
 
 .tdcenter {
     white-space: nowrap;
-    border-radius: prop(border-radius);
-    border-bottom-left-radius: prop(border-radius-large);
-    border-bottom-right-radius: prop(border-radius-large);
-    @include button.base($with-hover: false, $with-active: false);
     overflow: hidden;
     padding: 0;
-
-    @include elevation(1, $opacity-boost: -0.08);
-    &:hover {
-        @include elevation(2);
-    }
-    transition: box-shadow 0.2s ease-in-out;
+    background-color: transparent;
 }
 
 body {
@@ -40,17 +33,22 @@ body {
 
 .hitem {
     font-weight: bold;
-    padding: 8px 14px;
+    padding: 4px 6px;
     text-decoration: none;
     color: color(fg);
     display: inline-block;
-    @include button.base;
     border: none;
     &:first-child {
         padding-left: 18px;
     }
     &:last-child {
         padding-right: 18px;
+    }
+    border-radius: 5px;
+    background-color: transparent;
+    &:hover {
+        background-color: hsl(0, 0, 0, 0.1);
+        cursor: pointer;
     }
 }
 
--- a/qt/aqt/data/web/css/toolbar.scss
+++ b/qt/aqt/data/web/css/toolbar.scss
@@ -31,17 +31,11 @@ body {
 
 .hitem {
     font-weight: bold;
-    padding: 4px 12px;
+    padding: 4px clamp(4px, 1.2vw, 1.5rem);
     text-decoration: none;
     color: color(fg);
     display: inline-block;
     border: none;
-    &:first-child {
-        padding-left: 18px;
-    }
-    &:last-child {
-        padding-right: 18px;
-    }
     background-color: transparent;
     transition: box-shadow 0.2s ease-in-out, background-color 0.2s ease-in-out;
     &:hover {
--- a/qt/aqt/data/web/css/toolbar.scss
+++ b/qt/aqt/data/web/css/toolbar.scss
@@ -39,7 +39,7 @@ body {
     background-color: transparent;
     transition: box-shadow 0.2s ease-in-out, background-color 0.2s ease-in-out;
     &:hover {
-        background-color: white;
+        background: var(--button-bg);
         @include button.impressed-shadow(0.35);
         cursor: pointer;
     }
--- a/qt/aqt/data/web/css/toolbar.scss
+++ b/qt/aqt/data/web/css/toolbar.scss
@@ -31,7 +31,7 @@ body {
 
 .hitem {
     font-weight: bold;
-    padding: 4px clamp(4px, 1.2vw, 1.5rem);
+    padding: 6px clamp(4px, 1.2vw, 1.5rem) 5px;
     text-decoration: none;
     color: color(fg);
     display: inline-block;
--- a/qt/aqt/data/web/css/toolbar.scss
+++ b/qt/aqt/data/web/css/toolbar.scss
@@ -13,6 +13,10 @@
     border-spacing: 0px;
 }
 
+center#outer {
+    border-block-end: 1px solid color(border);
+}
+
 .tdcenter {
     white-space: nowrap;
     overflow: hidden;
@@ -33,7 +37,7 @@ body {
 
 .hitem {
     font-weight: bold;
-    padding: 4px 6px;
+    padding: 4px 10px;
     text-decoration: none;
     color: color(fg);
     display: inline-block;
@@ -44,10 +48,11 @@ body {
     &:last-child {
         padding-right: 18px;
     }
-    border-radius: 5px;
     background-color: transparent;
+    transition: box-shadow 0.2s ease-in-out, background-color 0.2s ease-in-out;
     &:hover {
-        background-color: hsl(0, 0, 0, 0.1);
+        background-color: white;
+        @include button.impressed-shadow(0.35);
         cursor: pointer;
     }
 }
--- a/qt/aqt/data/web/css/toolbar.scss
+++ b/qt/aqt/data/web/css/toolbar.scss
@@ -6,24 +6,6 @@
 @use "sass/elevation" as *;
 @use "sass/button-mixins" as button;
 
-#header,
-#header tbody {
-    padding: 0px;
-    margin: 0px;
-    border-spacing: 0px;
-}
-
-center#outer {
-    border-block-end: 1px solid color(border);
-}
-
-.tdcenter {
-    white-space: nowrap;
-    overflow: hidden;
-    padding: 0;
-    background-color: transparent;
-}
-
 body {
     margin: 0;
     padding: 0;
@@ -35,9 +17,21 @@ body {
     -webkit-user-drag: none;
 }
 
+.top_toolbar {
+    padding: 0;
+    margin: 0;
+    border-block-end: 1px solid color(border);
+    white-space: nowrap;
+    overflow: hidden;
+    background-color: transparent;
+    display: flex;
+    flex-flow: row nowrap;
+    justify-content: center;
+}
+
 .hitem {
     font-weight: bold;
-    padding: 4px 10px;
+    padding: 4px 12px;
     text-decoration: none;
     color: color(fg);
     display: inline-block;
--- a/qt/aqt/toolbar.py
+++ b/qt/aqt/toolbar.py
@@ -195,12 +195,9 @@ class Toolbar:
     ######################################################################
 
     _body = """
-<center id=outer>
-<table id=header>
-<tr>
-<td class=tdcenter align=center>%s</td>
-</tr></table>
-</center>
+<div class="top_toolbar">
+%s
+</div>
 """
 
 


dae:
With the default background color, the lack of contrast between the top area and the main area on pages like the deck list feels a bit off. When the main area has a different background color, it's not an issue though. Might it be worth sacrificing a pixel to add a thin bottom border?

2022-12-30T13:12:06,557553768+10:00

dae:
@kleinerpirat would be interested in your thoughts

kleinerpirat:
I'm fine with removing the elevation effect if that's what the community prefers - we haven't polled them yet.

The buttons are squeezed together a bit too much in the horizontal axis now. I'd prefer a bit more padding overall, but here it is especially needed.

Regarding the implementation:
You used a hard-coded gray value for the hover effect. You could test the CSS variables --canvas-elevated or --button-bg, which should be slightly more striking, but usable background colors.

tatsumoto-ren:
> The buttons are squeezed together a bit too much

Maybe I over-corrected a bit. Added slightly more padding.

> Might it be worth sacrificing a pixel to add a thin bottom border?

Good idea.

> You used a hard-coded gray value for the hover effect.

It was black color with transparency. But I decided to change it to something similar to the 2.1.55 look.

Second version:

https://user-images.githubusercontent.com/69171671/210093456-f6bc4625-274c-47ca-a0b1-8311961f25fe.mp4

tatsumoto-ren:
Horizontal padding between buttons now shrinks with the viewport. The rationale is that the more buttons you add to the toolbar (when subscribing to `top_toolbar_did_init_links`), the more space they need, especially when the main window has been resized.

glutanimate:
@tatsumoto-ren I understand your points and have to say that it initially took me a bit of time to get used to the new toolbar as well. However, having used 2.1.55 for a bit now, I do think it's a net improvement over the old design and would be hesitant to have us go back to something similar to before.

So to throw my thoughts into the discussion:

### What I like about the new design

- It communicates interactivity much better than the plain-styled links we had before. With no underlines or other design signals, the old controls were indistinguishable at a glance from non-interactive text, which seems problematic from a UX standpoint.
- It's (mostly) consistent with the rest of the web components we have now, thanks to all the work done by @kleinerpirat. Or at least, we're moving into a direction where the design language for interactivity is fairly similar across all Anki views. Going back to the plain controls we had before without changing the rest of the UI would take away from that consistency.

### What I think we could tweak

#### Toolbar controls bottom edge

I agree that the bottom edge gap in the reviewer is a bit jarring visually.

Auto-hiding the toolbar as in #2262 will ameliorate the issue, of course, but as this will be optional (which I think is the right choice), I think there is a need still for improving the default experience for users.

@kleinerpirat, please correct me if I'm wrong, but I think your original vision with the toolbar was to have it seamlessly flow into the reviewer once the consolidation of toolbar webviews and mw is done. As far as I'm concerned this would fully resolve the bottom-edge gap for card templates with custom background colors.

However, it bears mentioning that there is a chance that we still end up in the same situation if the card content is going to be displayed either via an iframe or shadow DOM component going forward (or the other way around, Anki's controls being wrapped in a shadow DOM). Any solution with the goal to prevent card template styles from flowing over onto Anki's UI could of course also get us back to a situation where we wouldn't be able to get a seamlessly integrated toolbar "for free", i.e. where we would still need to add some custom styling logic. Fully depends on the actual implementation, though, of course.

Putting that aside for now, to bridge the gap until that consolidation is done, what about this solution: We could have the toolbar's background color dynamically follow whatever background color is set on the card. We would of course still run into edge cases with this that wouldn't work as neatly, e.g. when cards have a background image, but it seems like a solution that would get us 95% of the way towards a seamless toolbar for most users. As far as the implementation is concerned, I would see two paths:

1. Naively parse the card template for the background color set on `body` or `card`.
2. Determine the actual color of the mw region adjacent to the toolbar. The easiest way for this would likely be to render the first pixel row of `mw` to a QPixmap, and then use the most predominant pixel color in that row as the toolbar's background color.

#### Toolbar controls top edge

Staying on the topic of toolbar edges, I'm not a huge fan of the toolbar's top edge touching the Qt menu bar, either. Given the toolbar's elevation, to me it feels a bit crammed and like the toolbar is getting pushed against the edge.

Here's how I think we could tweak this:

- Let's actually slightly increase the toolbar height, in order to
- introduce a margin between the toolbar controls and Qt menu bar.
- Finally, adjust the top border-radius to switch to a pill-shaped design (I suppose that gets us back to iPhone designs, but I do think this works reasonably well 🙃)


#### Other areas

(e.g. touching on hover states, current table markup, add-on support, etc. Will continue in a separate comment, both due to post length, but also because there are less incremental changes and will make much more sense to address as part of the mw refactor)

----

Here's what all of the proposed changes so far would look like, taken together:

|     View     |                                                             Screenshots                                                             |
|--------------|-------------------------------------------------------------------------------------------------------------------------------------|
| Deck Browser |  |
| Reviewer w/ dynamic background     |                                                                                                                                   |

Branch: https://github.com/glutanimate/anki/tree/poc-toolbar-styling-adjustments

glutanimate:
Regarding dynamically adjusting the toolbar background color, here's a POC in add-on form:

```python
from collections import Counter

from aqt import mw
from aqt.gui_hooks import (
    reviewer_did_show_answer,
    reviewer_did_show_question,
    state_did_change,
)
from aqt.main import MainWindowState
from aqt.qt import QPixmap, QRect, QRegion, QSize


def update_toolbar_background():
    if mw is None or mw.toolbar is None or mw.web is None:
        return
    toolbar_web = mw.toolbar.web
    main_web = mw.web
    main_web_rect = main_web.rect()

    first_row_rect = QRect(main_web_rect.topLeft(), QSize(main_web_rect.width(), 1))
    first_row_pixmap = QPixmap(main_web_rect.width(), 1)

    main_web.render(first_row_pixmap, sourceRegion=QRegion(first_row_rect))
    first_row_image = first_row_pixmap.toImage()

    pixel_colors = []
    for index in range(first_row_image.width()):
        pixel_colors.append(first_row_image.pixelColor(index, 0).name())

    if not pixel_colors:
        return

    most_common_color = Counter(pixel_colors).most_common(1)[0][0]

    if not most_common_color:
        return

    print(most_common_color)

    toolbar_web.eval(
        f"""
        document.body.style.backgroundColor = "{most_common_color}";
    """
    )


def on_reviewer_render_card(card):
    mw.progress.single_shot(1, update_toolbar_background)  # type: ignore


def on_main_window_state_change(new_state: MainWindowState, old_state: MainWindowState):
    mw.progress.single_shot(1, update_toolbar_background)  # type: ignore


reviewer_did_show_answer.append(on_reviewer_render_card)
reviewer_did_show_question.append(on_reviewer_render_card)
state_did_change.append(on_main_window_state_change)
```

This works ok-ish, but is currently hampered by following Python events which aren't in-sync with the actual web view render state. A proper solution would likely have to be triggered from the web view, which we do have existing architecture for in some mw states (e.g. `onShownHook` in the reviewer), but not in others.

glutanimate:
Regarding reducing the toolbar height: This currently breaks the button styling for the AMBOSS add-on and will likely also affect other add-ons that introduce custom buttons to the toolbar, e.g. Study Timer.

Screenshot_20221231_060712

It would be nice to avoid small breakages like this until the larger toolbar overhaul that will inevitably have to come as part of the mw refactor.

tatsumoto-ren:
@glutanimate if you make the toolbar detached and pill-shaped, it will sure look pretty, but your screenshots still leave about 42px of useless space on the top. It was only 31px in 2.1.54. 

If you really wanna go this route though, directly injecting the toolbar code into the review screen (MainWebView) will be more straightforward. The toolbar could be placed just before `<div id="qa">`. Technically it should allow the users to customize its look in their card templates.

> It's (mostly) consistent with the rest of the web components we have now

Not really if you look closely. It's the only component that has inset shadows, for example.

> It communicates interactivity much better than the plain-styled links we had before.

Did you have any trouble using the older toolbar? It's hard for me to imagine. But if so, adjusting the styles of individual buttons a bit to make them stand out more is what I would do.

> This currently breaks the button styling for the AMBOSS add-on 

Adding max-height rules to the children could solve this. However, even right now my version of the toolbar doesn't appear to break it. 

screenshot-2022-12-31-05-21-34

> Auto-hiding the toolbar will ameliorate the issue, of course, but as this will be optional.

I would view it as an unrelated feature. By itself it's nice to have. But if you don't want to hide the toolbar, then it won't do anything for you.

> We could have the toolbar's background color dynamically follow whatever background color is set on the card

There's a set of corner cases where that would end up looking ugly. For example, when you have image as a background. But it won't be an issue if the toolbar simply floats on top of a card inside the same webview.

glutanimate:
> if you make the toolbar detached and pill-shaped, it will sure look pretty, but your screenshots still leave about 42px of useless space on the top. It was only 31px in 2.1.54.

I'm not a huge fan of too much whitespace, either, but generally, I think we should be open to sacrificing some screen real estate in the spirit of having more freedom with visual designs. I would actually say that we are in a fairly privileged position to do so, given that Anki cards are fairly compact by design, so we're not really hurting that much for vertical screen space, and even in more cramped areas of the UI like the deck browser, a few pixels don't seem significant.

> If you really wanna go this route though, directly injecting the toolbar code into the review screen (MainWebView) will be more straightforward. The toolbar could be placed just before `<div id="qa">`. Technically it should allow the users to customize its look in their card templates.

Essentially this is the direction the `mw` refactor would take us in, but merging the three web views will entail a host of other changes in terms of how the `mw` content is rendered and compositioned (moving away from rendering in Python towards client-side rendering via Svelte components). Injecting the toolbar in a classical sense without a larger overarching refactor of `mw` would also be possible in theory, but IMHO would take too much effort and come with too many pitfalls for – what would essentially  be – a transitional solution until the inevitable `mw` refactor.

As far as styling Anki's UI controls via templates is concerned, I think that's something we'd have to ponder on once we decide to move towards the unified `mw`. To me this feels like a support nightmare that I don't think we should touch. I would rather have us place the controls out of reach of templates as best as we can to avoid unintentional spillover of template styling and logic. IMHO, we should keep UI customization squarely in the domain of add-ons and/or a native app-wide theming system.

> Not really if you look closely. It's the only component that has inset shadows, for example.

Agreed, yes, it's one of the areas I think we should tweak. I'm also not 100% a fan of the double-hover interaction that the toolbar currently has, i.e. both the toolbar as a whole changing in elevation on hover, but also the individual labels. But in general I would say that we're much more consistent now with e.g. the recent-ish changes to the editor that introduced more button-y buttons.

> Did you have any trouble using the older toolbar? It's hard for me to imagine. But if so, adjusting the styles of individual buttons a bit to make them stand out more is what I would do.

No, but I think as Anki power users we have to be particularly mindful of the UX for new users and the expectations / trained patterns of UI interactions they're coming in with. To be honest, I'm not hard-set on any particular toolbar design, I just think that the old design has quite a few problems that we should seek to address at some point to make Anki more intuitive (cf. below for more general thoughts on that, though as someone who's not a UI/UX designer, it's always much tougher to come up with solutions than pointing out problems).

> However, even right now my version of the toolbar doesn't appear to break it.

Hmm, that's strange. I can still reproduce this here, testing against 95397b5480b1b0727073c1cdfb1838f737517396. Are there any local changes you might have not pushed yet?

> I would view it as an unrelated feature. By itself it's nice to have. But if you don't want to hide the toolbar, then it won't do anything for you.

:+1:

---

Continuing on with some general thoughts on the toolbar (*sorry for somewhat hijacking this PR, just seems like the best place to discuss this with all of us in here*):

- I think in general we're all battling a bit with finding a design that, at the same time:
    - is easy to visually parse
    - is intuitive and tactile to interact with
    - fits in nicely with all the states `mw` can be in
    - is not overlay intrusive while reviewing cards
- To be completely honest, I don't think that any of our designs so far manage to hit all of these goals. E.g. for the adjusted design I proposed above, while I do like its seamlessness, I think it suffers from drawing too much attention to itself in the reviewer. With the classic toolbar design, OTOH, I do like its non-intrusiveness when reviewing cards, but as mentioned above, am really not a fan of its lack of "tactility" from a UX standpoint.
- IMHO, I don't think we will be able to address these issues in satisfactory way without a much more encompassing overhaul of the toolbar and navigation in Anki in general. For now, I think that going with one of these iterative designs is reasonable, but once we seriously start looking into the mw refactor, I would really like for us to consider breaking free from this design and visit questions like:

    + Does it actually make sense to group all of these controls in one component?
       + Consider that all of these trigger completely different UI interactions
           + *Decks* allows navigating between mw states
           +  *Add*, *Browse*, and *Stats* open completely new windows
           +  *Sync* is not really navigational and also communicates state
           +  Add-ons can do all kinds of things, adding a confusing mish-mash of state indicators, navigational elements, and contextual actions
    - Does it actually make sense to display all of these controls at all times?
    - Would it make sense to switch to icons for some of these controls/indicators?
          - Consider that the UX here is very different for different internationalizations. What might look OK-ish in English can actually look very noisy in other languages where labels for different controls can vastly differ in their length. It's one of the reasons I don't particularly like to use Anki in German:
          Screenshot_20221231_074443
    - etc.

tatsumoto-ren:
> Would it make sense to switch to icons for some of these controls/indicators?

screenshot-2022-12-31-05-12-40

Yes, why not. This is Geany, just a random program with a toolbar. The toolbar can be hidden. Really if Anki just copied this design, that would be an improvement.

> Add-ons can do all kinds of things

It's likely impossible to restrict addon devs from doing any particular thing. They can override default code and shove something completely foreign in any part of the UI. I guess the users should complain when it happens.

> Does it actually make sense to group all of these controls in one component? 

If they're changed into icons, probably won't be a problem. I don't see where else they could be moved.

> Are there any local changes you might have not pushed yet?

No.

> I'm not hard-set on any particular toolbar design, I just think that the old design has quite a few problems that we should seek to address at some point.

It wasn't broken before, and it wasn't counter-intuitive either. That's why my idea was to make the change less drastic. But in the future it can be changed to something else.

> place the controls out of reach of templates

Card templates can already interfere with a number of UI elements. The most obvious is the play button. But since card templates could in theory break the entire UI, there should be a checkbox to restrict that.

I currently don't know how to do that though. Any element inside the main webview can be affected by a card template unless it is placed in a different webview (but we already have it this way).

>  I think we should be open to sacrificing some screen real estate in the spirit of having more freedom with visual designs.

This idea could eventually lead to very dangerous outcomes in my opinion.

tatsumoto-ren:
https://github.com/tatsumoto-ren/anki/commits/top_toolbar_integration

We can inject the toolbar into the main webview by doing this. However, this approach suffers from the complications mentioned above.

glutanimate:
> It's likely impossible to restrict addon devs from doing any particular thing. They can override default code and shove something completely foreign in any part of the UI. I guess the users should complain when it happens.

Yes, absolutely, and I also don't think we should take away from that freedom, just build add-on APIs that are helpful enough to naturally create that structure. E.g. if the next iteration of the toolbar has dedicated areas for navigational controls and status indicators, create APIs that allow add-ons to place elements/components in either area.

> If they're changed into icons, probably won't be a problem. I don't see where else they could be moved.

Probably not a drastically different location, but one first thought could be to separate the controls by the types of actions they perform, i.e. navigation within the same view, spawning new views, or contextual actions & indicators.

Simple mock-up:

Screenshot_20221231_093647

> I currently don't know how to do that though. Any element inside the main webview can be affected by a card template unless it is placed in a different webview (but we already have it this way).

If we wrap the card content in an iframe, it might be possible to sandbox it from the rest of the web view, but that could end up breaking templates that do require access to window globals, and we'd have to make sure that add-ons can still easily inject scripts and content into cards.

Another solution would be to wrap all UI controls in shadow DOMs, which would at least prevent (or make difficult) accidental spillovers from templates to the UI, and hopefully discourage template authors to attempt modifying the UI.

But you're right, it's not a trivial problem to solve and there's a lot considerations that would have to go into it. https://github.com/ankitects/anki/issues/1120 lists quite a few more.

tatsumoto-ren:
When it comes to replacing labels with icons, we could add this CSS and adjust titles a bit:

https://github.com/tatsumoto-ren/anki/commit/64495e846b455d6b89029b293cb70444fe5e49ad

The `svg`s should be more carefully selected of course. These are somewhat random, just do demonstrate.

https://user-images.githubusercontent.com/69171671/210131141-3c922670-7a61-4f11-baf1-dfb233bd398d.mp4

tatsumoto-ren:
> Simple mock-up

I think it sacrifices usability for the looks. Unless you want to automatically shrink it vertically when entering the Reviewer, then it's fine. It's also easier to replace all labels with icons at once than leave some of them in text form. Geany's toolbar is compactly designed and allows adding many buttons and other elements which I think I nice.

Separating buttons by type would require more work, but it's probably not very difficult to implement.

> If we wrap the card content in an iframe, it might be possible to sandbox it from the rest of the web view
> Another solution would be to wrap all UI controls in shadow DOMs

A little hacky. Honestly I'd leave it as is then. If it's not broken, no need to fix it.

dae:
Just to set expectations: I'd like to get a new rc out within a couple of days, and .56 out in not much later than a week, as the macOS issue in .55 is going to be bothersome. I expect we'll need at least one more release after that focused on bugfixes, and testing Qt 6.4.2 which is due out around mid month. Based on the feedback above, I think the changes in this PR will need some time in beta testing, and a poll on the forums to see how popular they are. In the shorter term, perhaps there is a less contentious middle ground that could be reached? Is there room to shave a few pixels off the top and bottom of the current design without breaking add-ons/requiring a poll?

tatsumoto-ren:
The focus of this PR is mostly to fix a visual bug, not to introduce new features. We could discuss everything else on the forum in more detail.

If this is merged, I can make a separate PR that would implement toolbar icons.

dae:
Fixes for regressions are good, but you're also undoing some of the styling that was tested over quite a few weeks, and we can see an example of where this fix introduces a new problem with an add-on above. As the person who has to pull the trigger, my concerns are "what is this going to break?" and "how many people are going to wish this change wasn't made?". The more conservative you are in your changes, the less I have to worry about those things, and less time/community feedback is required.

Anki 2.0 moved from icons to text because some users couldn't intuit what some of the icons meant. That could potentially be ameliorated with better icons, and I'm not fundamentally opposed to a move back to using icons, but it's something that again needs a testing period and community feedback.

tatsumoto-ren:
>  this fix introduces a new problem with an add-on above.

I'm afraid the only way to ensure that the **AMBOSS** button has enough vertical space is to go back exactly to the previous height of 31px which is achieved by increasing `.hitem` padding from `4px` to `6px` on the top and `5px` on the bottom (or the other way around). The `amboss-indicator` has absolute positioning so it basically lives its own life.

On the current version it looks even stranger.
screenshot-2023-01-01-21-23-13
screenshot-2023-01-01-21-24-21

Are there any other potential problems?

> Anki 2.0 moved from icons to text because some users couldn't intuit what some of the icons meant. That could potentially be ameliorated with better icons.

Without tooltips there's probably no way to be sure that a certain icon does a good job at conveying what the button means. Tooltips are essential.

glutanimate:
To regain some vertical space in the original .55 design, we could reduce the `.hitem` padding to `6px 12px;` which would lower the toolbar height to 36px. Anything smaller than that and the design in its current form starts looking off, IMHO.

Generally speaking, I think it's important that we avoid ping-ponging between designs too much. The changes in this PR would essentially revert all of the .55 design choices in relation to the toolbar, and I don't think the community feedback is there yet to support that move. Giving the current .55 design a bit more time in users' hands will allow us to narrow in more closely on the actual sentiment in the wider community and allow us to make more informed choices.

With that said, I definitely do think that some of the feedback so far – and also our own takes in this discussion – show that there is a need to further tweak the toolbar's design, especially in the reviewer context. However, I think this is something that will require more deliberation and back-and-forth, ideally in a way that doesn't affect the UX of Anki's wider user base until we narrow in on the right choice.

glutanimate:
(the comment above only focuses on the user-facing changes. I do like some of the technical changes in this PR in terms of e.g. moving away `table` and `center`, but I don't think it makes sense for us to cherry-pick these refactoring changes in isolation without the design changes to go alongside with them. I would imagine that the main reason this old code persisted throughout the .55 redesign was to keep the diff surface small until a larger overhaul of the toolbar)

tatsumoto-ren:
> Generally speaking, I think it's important that we avoid ping-ponging between designs too much. The changes in this PR would essentially revert all of the .55 design choices in relation to the toolbar

In order to prevent the bottom shadow from creating an awkward gap between the toolbar and a card, the shadow had to be removed. But then the rest of the notch would look somewhat incomplete (white on gray), so the white background had to go too. The height of the toolbar was reduced back to 31px. Other than that, I kept the 2.1.55 design.

glutanimate:
@tatsumoto-ren I think that's a perfectly sensible series of steps, but to me it also perfectly shows that this is not a problem that can be tackled with small iterative changes (if these small iterative steps eventually lead back to a toolbar design that – to a user looking at it without knowing about the under-the-hood changes – basically looks like 2.1.54). I don't think this is an inevitability. I think we can arrive at a design that both continues to look nice in the deck browser & overview screen (and harmonizes with the rest of the components there), while also not being too obtrusive in the reviewer.

To clarify my take on the .55 toolbar design in general (putting aside smaller issues with hover styling, inset shadows, etc.): I do like how it looks in the deck browser and overview screen and how it fits in there with the rest of the components. It's just the reviewer where I think it currently stands out too much. With the design in this PR it's the exact inverse to me. So the naive solution might be to use both designs in their respective contexts. But that switch between mw states would then be jarring in its own right (if not done in a more elegant way), circling back to my point that that IMHO there are no low-hanging fruit solutions here.

tatsumoto-ren:
@glutanimate

>  to a user looking at it without knowing about the under-the-hood changes – basically looks like 2.1.54

I would be happy to see the old toolbar back in one way or the other. The users are likely to welcome this change.

<details>

screenshot-2023-01-02-18-05-05

</details>

>  I think we can arrive at a design that both continues to look nice 

You are implying that what we have now looks nice, but at the very least what we have now is *controversial*. In addition, there are quite a few visual glitches that have been reported.

Another way this issue can be solved is by providing more presets in Preferences, or at least an option to return the 2.1.54 interface. Currently, you can choose between "system", "dark" and "light", and there could be one or two more options under the same settings.

<details>

screenshot-2023-01-02-18-13-16

</details>

chezfo:
Perhaps a slightly bold idea can be suggested...
When testing, the bar will be automatically hidden, and opened when the mouse is placed on it. Similar to the Windows taskbar.
אנימציה

glutanimate:
> You are implying that what we have now looks nice, but at the very least what we have now is controversial

Fair point, that wasn't the best phrasing. I like the looks of the design in the overview and deck browser and it seems like during beta testing most users did so as well, but that doesn't mean that this reflects the general consensus in the wider community. I saw that comment you mentioned, alongside other feedback on the forums and elsewhere, hence why I said above:

> With that said, I definitely do think that some of the feedback so far – and also our own takes in this discussion – show that there is a need to further tweak the toolbar's design, especially in the reviewer context

My main stance on this PR is that it feels like prematurely reversing design choices that *did* already go through a community feedback round, and I would rather have us play around more with the .55 design based on user feedback than go full reverse throttle and revert to a .54-like design.

Perhaps going back will end up being the consensus we arrive at, but IMHO it's too early to properly gauge the community's sentiment, and in particular understand which parts of the designs are well received and which ones aren't (even though I think a picture is starting to form around the pain points primarily lying with the reviewer and obtrusiveness of the toolbar there).

> Another way this issue can be solved is by providing more presets in Preferences

That would be nice, but I think we have to arrive at a good default first, as that's what the large majority of users will be using.

tatsumoto-ren:
>  the bar will be automatically hidden, and opened when the mouse is placed on it

I think there's a pull request that adds such feature.

tatsumoto-ren:
> My main stance on this PR is that it feels like prematurely reversing design choices that did already go through a community feedback round

Yes, this is the biggest thing I don't get. You're saying that it was *tested* and there was *community feedback* but I knew nothing, other people (who aren't frequent visitors of the forum) also knew nothing, and then it became mandatory. 

To give you an analogy, we can still toggle the v3 scheduler on and off, and it's not mandatory yet. If issues were to arise with it, people could just disable it.

From here we have two routes:

1) Remove the current UI design completely and make a new one.
2) Fix the current styles.

We can do both: provide an alternative theme in Preferences and apply small tweaks to the current theme.

dae:
You both have valid points. I agree with Aristotelis that it feels premature to be simply undoing the changes made in .55 - a fair number of users seem to prefer the new look, and these changes went through a beta testing phase where add-on authors could test/make tweaks to their add-ons to support them. But as Ren points out, there do seem to be a fair number of users who prefer the old style as well. While I don't like to add extra options when we can avoid them, perhaps that is the best move here, as it will placate both sides, and allow further tweaks to the new styles to be made without being in such a rush. If Ren were to alter this PR to add such an option and try to keep the toolbar styling as close as possible to the .54 version (https://github.com/ankitects/anki/pull/2287#issuecomment-1368565373), is that something you guys would be ok with?

Issue #2251, opened 2022-12-09
Duplicated character after entering IME-text into MathJax editor


dae:
Issue appears to exist with both Qt5 and Qt6 builds, and both when closing via the enter key, and clicking outside the editing area.

https://forums.ankiweb.net/t/chinese-input-after-mathjax-error/25187

PR #2288, opened 2022-12-28
Work around issue with entering text around MathJax via IME

hikaru-y:
This works around #2251.

I don't think this is an ideal solution, and it would be better to find a way to eliminate the need to move the text node in the first place, but I have not been able to come up with one at this time, so I have created this as a tentative workaround.

( I used the phrase 'around MathJax' in the title of the PR because the issue occurs not only when entering text after a MathJax element, but also before one. )

I tested this on Win10 with the following three IMEs and it worked with all of them.
- Microsoft Pinyin (Simplified Chinese)
- Microsoft Japanese IME
- Google Japanese Input

I also tested this on Linux with **iBus+Mozc**, and the issue did not occur even on the current `main` branch. This is probably because unlike the three IMEs above, no changes are made to `<anki-editable>` while the IME composition session is active.

### DIFF
--- a/ts/editable/frame-handle.ts
+++ b/ts/editable/frame-handle.ts
@@ -53,7 +53,6 @@ function restoreHandleContent(mutations: MutationRecord[]): void {
             }
 
             const handleElement = target;
-            const placement = handleElement instanceof FrameStart ? "beforebegin" : "afterend";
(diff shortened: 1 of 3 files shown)


dae:
A tentative workaround is better than nothing! Really appreciate you looking into this Hikaru.

PR #2289, opened 2022-12-29
Revamp Preferences, implement Minimalist Mode and Qt widget gallery to test GUI changes

kleinerpirat:
This adds a gallery similar to [this one](https://doc.qt.io/qt-6/gallery.html) to the debug dialog.

image

image

It allows users to change the styling approach across the application via a dropdown and see the effects immediately on various Qt widgets.

There are three choices:
- Anki (new code path with custom stylesheets)
- Fusion (code path of 2.1.54 and below)
- Native (same code path as current macOS Qt6)

If the "Force Style" box is checked, the style will be kept across profiles.

---
@tatsumoto-ren I don't know how much of the GTK theme Anki should pick up. Please tell me if the "Native/Fusion" options are working for you. The UI looks the same as on 2.1.49 for me. Missing styles for the "Anki" option will be added in future PRs.


tatsumoto-ren:
Your branch doesn't build for me. The dialog looks handy though.

<details>

```
./run

Finished dev [unoptimized + debuginfo] target(s) in 0.19s
[1/21; 1 active; 1.604s] ftl:repo:core
FAILED: /home/ren/Downloads/anki_git/out/git/ftl/core-repo
/home/ren/Downloads/anki_git/out/rust/debug/runner run --stamp=/home/ren/Downloads/anki_git/out/git/ftl/core-repo git submodule update --init ftl/core-repo
Command failed:

fatal: transport 'file' not allowed
fatal: Fetched in submodule path 'ftl/core-repo', but it did not contain 2d13bfbe9e60dd365c4300f1e0589100b8b044af. Direct fetching of that commit failed.

ninja: build stopped: subcommand failed.
Finished dev [unoptimized + debuginfo] target(s) in 0.17s
[1/21; 1 active; 0.583s] ftl:repo:core
FAILED: /home/ren/Downloads/anki_git/out/git/ftl/core-repo
/home/ren/Downloads/anki_git/out/rust/debug/runner run --stamp=/home/ren/Downloads/anki_git/out/git/ftl/core-repo git submodule update --init ftl/core-repo
Command failed:

fatal: transport 'file' not allowed
fatal: Fetched in submodule path 'ftl/core-repo', but it did not contain 2d13bfbe9e60dd365c4300f1e0589100b8b044af. Direct fetching of that commit failed.

ninja: build stopped: subcommand failed.

Build failed.
```

</details>

When it comes to changing the theme, my idea is that the selector in "Tools" > "Preferences" could allow the user to choose between not only "system", "dark" and "light", but also other presets. Also, **system** currently doesn't do anything so I'd like to fix that.

dae:
@tatsumoto-ren Matthias has not changed anything that would cause that error. It seems the error is caused by a recent change to git's behaviour. From the repo root, if you run `git submodule update --init ftl/core-repo`, do you get the same error? What about if you run `git -c protocol.file.allow=always submodule update --init ftl/core-repo`?

dae:
@kleinerpirat I get an error on startup, and had to run the following first:

```
>>> mw.pm.meta["force_anki_styles"] = False 
<no output>
 >>> mw.pm.meta["force_fusion_styles"] = False 
<no output> 
>>> mw.pm.meta["force_native_styles"] = False
 <no output>
```
```
    if self.meta[f"force_{AnkiStyles(member).name.lower()}_styles"]:
KeyError: 'force_native_styles'
```

If those three options are mutually exclusive, I think we should probably be storing them in a single key instead of three separate booleans, eg meta.get("widget_style") would return "anki" | "fusion" | "native" or 0 | 1 | 2.

Rest of the PR looks good; I can see how this would be helpful for debugging styling issues moving forward.

kleinerpirat:
Yeah, I think a single key will be the more elegant implementation, although it's less intuitive to the user.

tatsumoto-ren:
@dae thanks for the tip. After running the second command, i was able to build successfully.

The `KeyError: 'force_anki_styles'` also affected me, but I fixed that too. @kleinerpirat, maybe if you use `dict.setdefault`, you could prevent the errors.

For testing it would be handy if the table widget had a few rows with some text in them. Earlier I ran into a problem with text being unreadable when using a custom table widget in 2.1.55.

Since Anki doesn't pick up my GTK theme currently, I don't see much difference between "fusion" and "native".

When testing custom Qtwidgets, I'd really like this menu to appear globally under "tools" > "preferences" too. I'm sure addon devs would agree.
menu

dae:
The choices change based on platform. On Mac, "Anki" is mostly native widgets, with "Fusion" being the less-native styling that was used in dark mode in old releases, and I presume we don't need a separate native option. System Qt themes are not a concern on Windows either, so presumably we'd only need 2 options - or even just 1? The push for access to the old fusion theme seems to have come from macOS users IIRC, so perhaps we don't even need a setting on Windows?

@kleinerpirat would it make sense to expose this as a dropdown in the prefs screen, with 2 options on Mac, 3 on Linux, and either hide it on Windows, or have 2 options? That removes any confusion about manually changing enum settings. I'd originally hoped to keep the macOS fusion option hidden in the debug console, but AnKing went and promoted it to his followers, so any hope of keeping that obscure rapidly went out the door :-)

kleinerpirat:
A little update on what I'm working on:

image

The Preferences screen will be revamped a bit. I intend to make more use of `QGroupWidget` to bring more order into the tabs.

`Minimalist mode` will add a body class to `AnkiWebView`. Properties like large border radii, glass effects or box-shadows of containers are disabled in that mode.

dae:
Thanks for working on this. I think we'll probably want to include the UI language there too, and move that page to the first tab.

kleinerpirat:
# Update

## Styles

- reduced styling choices:
  - `Anki`: Fusion + Palette + Stylesheets, default on Linux and Windows
  - `Native`: Fusion + Palette, default on macOS
- added custom styling for `QSlider` and `QGroupBox`, prefixed a few missing selectors to existing rules, which I also tweaked a bit
- simplified the stylesheet imports in `theme.py`
- added  `border-radius-medium` prop (12px) and used it for containers where `border-radius-large` (15px) had been previously used

## Preferences

The preferences are reordered quite a bit now.

- added `Appearance` tab, which contains settings related to the UI
- used `QGroupBox` to categorize related settings
- added a dropdown to select the widget style (Anki, Native)
- added a minimalist mode checkbox
- made settings involving class changes to webview bodies respond immediately



## Minimalist Mode

This mode removes the body class `fancy` from Anki's webviews. Good-looking, but unnecessary CSS can now be defined like this:

```scss
.sample-element {
  // common definitions
  padding: 1rem;
  font-weight: bold;

  .fancy & {
    // extra definitions omitted in minimalist mode
    border-radius: var(--border-radius-medium);
    filter: drop-shadow(0 0 0.75rem var(--border-focus));
  }
}
```



etc.

The mode also unsets `box-shadow` and `background-blur` definitions globally with `!important`.

qt/aqt/main.py
```diff
@@ -1068,9 +1068,9 @@ def setupStyle(self) -> None:
         if is_lin:
             # On Linux, the check requires invoking an external binary,
             # which we don't want to be doing frequently
-            interval_secs = 300
+            interval_secs = 10
```
kleinerpirat, line 1071: 300s seemed overkill. I don't know how much of a performance impact invoking the external binary has, but I didn't notice anything when I set it to `2` on my Gnome system. `5` or `10` would also suffice to achieve a relatively smooth transition between system themes.

Automatic theme adjustments are a very nice thing to have, especially if you switch themes regularly via shortcut or with a sunset timer.

kleinerpirat:
Unfortunately the Preferences screen crashes when using <kbd>Shift+Tab</kbd> to focus the previous widget. This does not happen in the widget gallery, so I'm positive it can be resolved with some testing.

qt/aqt/main.py
dae: Yes, I was probably a little too conservative there :-) They appear to take a few ms at most, so once every 5-10s isn't going to contribute much to global warming/exhaust user's batteries.

dae:
It looks like the shift+tab issue exists in the main branch as well, so that doesn't seem to be your fault. Wierd!

The preferences screen looks cleaner - those groups do help with things. I do find it a bit strange to have a Reviewer section in Appearance and then a separate Review section in Basic. One idea would be to move the review and scheduler things into a single Review tab, and rename Basic to Editing, with browsing and import/export covered under that tab. WDYT?

Regarding the minimal mode, I don't have personal preferences here, so we'll need to see how people react to it. Some possible objections they might raise are that buttons don't have outlines anymore, and that there's no border at the top. @tatsumoto-ren any feedback at this stage?

kleinerpirat:
> move the review and scheduler things into a single Review tab, and rename Basic to Editing, with browsing and import/export covered under that tab. WDYT?

Sounds good. I'll rearrange it.

> Some possible objections they might raise are that buttons don't have outlines anymore, and that there's no border at the top.

That's sort of the point of this mode :-) I figured the same users that want no outlines on the upper buttons (toolbar) would also prefer borderless buttons at the bottom. Which border at the top are you referring to? The toolbar does have a border:

image

dae:
Ah, I usually use the interface at 150% - it seems the bar disappears when it's at 120%+

tatsumoto-ren:
> Some possible objections they might raise are that buttons don't have outlines anymore, and that there's no border at the top. 

People who commented on the change stated that they wanted the old design back. Following the old design should be a safe bet.

I think the buttons would look way better if they had outlines. They used to look like this:

screenshot

When I review, the toolbar copies the card's background (and has a 1px border at the bottom). This doesn't look pretty.

screenshot-2023-01-13-13-24-29

Is "hide top bar" on by default? Wouldn't it be better if it was always shown by default instead?

tatsumoto-ren:
I tried with default settings (where the top bar collapses) and I still see an empty gap. 

https://user-images.githubusercontent.com/69171671/212335624-ff0edde5-1d7f-4b1c-bbfe-f58d0dc8533c.mp4

kleinerpirat:
> When I review, the toolbar copies the card's background (and has a 1px border at the bottom). This doesn't look pretty.

My first impulse was to unset the dynamic background in this mode because the likes of you surely do not approve of it, but then I thought there might be a significant portion of users who want the old look, but still have the dynamic background.

I don't have hard feelings about any of this, but I'd bet there will come a complaint, one way or the other. If we do not want to end up with an endless tree of fine-tuning options, we'll best ask the community in a poll once I'm done with this PR.

> Is "hide top bar" on by default?

If it is right now, that's not wanted. I'll take a look.

> I also find that the deck list doesn't look the same anymore.

That's right.

tatsumoto-ren:
> Yes. I squashed the entries a bit to satisfy another frequent complaint about my design changes in 2.1.55+.

Having a `border-bottom` like before would make it closer to the previous look.

> dynamic background 

I doubt in this mode there is a need for it. The toolbar is supposed to be one monolith entity, in other words it has to look as in 2.1.54. 

> Options groups

Removing the white background only makes things harder to separate visually. But removing box-shadow and reducing/removing border-radius is a good idea.

kleinerpirat:
> Having a border-bottom like before would make it closer to the previous look.

> The toolbar is supposed to be one monolith entity, in other words it has to look as in 2.1.54.

I think you might have misunderstood my intentions with minimalist mode. It isn't meant to exactly copy the 2.1.24 look - if you want that, you should have stopped updating Anki some time ago. Its sole purpose is to reduce distractions and if there's a better solution than copying the look of previous Anki versions, I'll take that road. That's another reason why I decided to keep the dynamic background in that mode: a thick white bar can be very distracting on cards with image backgrounds.

I'll take your suggestions into consideration, but please, for the love of god, reduce your demanding tone a bit. I'm risking my semester just to make you stop complaining all over the forum. Perhaps you don't know how much effort all of this is. A simple thank you would be nice at some point.

tatsumoto-ren:
> It isn't meant to exactly copy the 2.1.24 look - if you want that, you should have stopped updating Anki some time ago.

There are other changes that are worth upgrading for. I'm surprised you don't know or don't care about them.

>  a thick white bar can be very distracting on cards with image backgrounds.

I assumed it's not needed since you already added an option to hide the bar. I'm afraid the copying is not perfect. I attached a few screenshots above.

> a thick white bar can be very distracting on cards with image backgrounds.

I don't think people complained about it when it was that way. Another thing to consider is that the buttons will be nearly invisible if the background is too dark.

>  reduce your demanding tone a bit.

Sorry to come off as demanding haha. All I ask is to be able to opt out of "improvements" that ruined Anki. But at the same time I care about actual improvements, e.g. the scheduler, so I can't just use an old version. I'll thank you when you're done 👍
You don't have to do it alone. I offered my help in the issue.

kleinerpirat:
> I assumed it's not needed since you already added an option to hide the bar. 

The toolbar webview isn't hiding as a whole, see https://github.com/ankitects/anki/pull/2262#issuecomment-1373477292. I assume the dynamic background doesn't work in your sample screenshot, because that template's background is set on some element other than the body.

tatsumoto-ren:
I see. I have a background set for `.card` and a different background set for a div that rests inside `.card`. This is to make cards look more like pages in a book. I have uploaded sample cards here https://ankiweb.net/shared/info/1557722832

dae:
>  @dae, what do you think the purpose of the mode should be?

Your minimalist mode here is effectively asking a question to the users who said they prefer the old look: "you say you like the old look more because it's less distracting; how about this instead, which gets even further out of the way?". I think your general approach here of trying to read between the lines/get to the root of the issue is a good one. I suspect in this case the objections are more about people wanting what they're used to than wanting further flattening of the UI, but we'll have a better idea of what the majority feel once you put it to a poll on the forums.

kleinerpirat:
Full screen switching still looks a bit janky, but other than that I think this PR is ready for review.

dae:
Quite a lot going on here! The top/bottom toolbar hiding seems to work well, and looks good as a whole.

Minor issues that won't block merging:
- would be nice if the top border doesn't disappear in minamal mode with UI > 100%
- the top group looks a little odd when the video driver option is hidden - not sure if that's easy to fix or not, since it seems to be Qt related?
2023-01-16T10:08:25,518685609+10:00

More serious issues:
- the native macOS styling has been lost - the native looks like Linux now instead of the standard macOS
- if the user enters full screen via the menu without remembering the shortcut, they have no way to get back to the menu without going through all their cards. We should perhaps be listening for Esc, or provide some UI element to return.
- when in full screen,

dae:
When Anki is first opened, the bottom bar starts near the top and then animates down towards the bottom of the screen, which is somewhat odd.

kleinerpirat:
Please let me know if https://github.com/ankitects/anki/pull/2289/commits/792b834109637834d73f0a1df21dd8810395d87c has fixed the macOS issue. I had removed those lines because theme switching in native mode didn't update the colors correctly, so I figured forcing our palette would be required.

My guess is that macOS is fixed now, but on Windows/Linux, theme switching might not work as smoothly as before when native style is active (app restart and OS theme switch required).

kleinerpirat:
Alright, I think I've figured it out. Because the theme switching issue is caused by the OS palette, limiting the behaviour to `Theme.FOLLOW_SYSTEM` ensures Anki will always be in the correct theme for the native colors.

Also, apparently macOS and Windows Vista do not use the palette for everything:

> Note: Some styles do not use the palette for all drawing, for instance, if they make use of native theme engines. This is the case for the Windows Vista and macOS styles.
https://doc.qt.io/qt-6/qapplication.html#setPalette

This might explain some differences we've experienced in the past.

kleinerpirat:
I removed the check for macOS specifically, because the style is set to `NATIVE` on macOS by default. We might need to check if Qt5 still needs separate handling, though.

dae:
Still broken - it looks better if I comment out this line:

app.setStyle(QStyleFactory.create("fusion"))  # type: ignore

dae:
Before commenting:

Screenshot 2023-01-16 at 7 19 51 pm

After commenting:

Screenshot 2023-01-16 at 7 26 39 pm
Screenshot 2023-01-16 at 7 26 34 pm

dae:
> Because the theme switching issue is caused by the OS palette, limiting the behaviour to Theme.FOLLOW_SYSTEM ensures Anki will always be in the correct theme for the native colors.

I'm not following the logic here. Previously we did not need to check FOLLOW_SYSTEM for this code path, and the change makes me nervous.

dae:
To reiterate on the old logic, because it's confusing even to me: on macOS:

- in day mode, for both qt5 and qt6, fusion should not be set (unless the user has that custom widgets setting on; older Anki releases never set it)
- in night mode, fusion should only be set on qt5 (because the old qt doesn't display properly when switching between light/dark otherwise, IIRC), or if the user has forced the custom widgets

kleinerpirat:
Ah sorry, I forgot about the manual theme cases. Now it should work as intended on mac.

> I'm not following the logic here

On Qt6, if we do not manually override the palette when switching to a theme that differs from the current system theme, the UI will look like this:

image

> the change makes me nervous.

For what reason? :D Nothing will change other than `FOLLOW_SYSTEM` actually obeying system colors.

qt/aqt/theme.py
```diff
@@ -281,7 +282,8 @@ def _apply_palette(self, app: QApplication) -> None:
             app.setPalette(self.default_palette)
             return
 
-        app.setStyle(QStyleFactory.create("fusion"))  # type: ignore
+        if not is_mac:
+            app.setStyle(QStyleFactory.create("fusion"))  # type: ignore
```
kleinerpirat, line 286: Does this check not work for you?

dae:
On dark/mac/qt5:
- fusion needs to be enabled; the native theme doesn't switch between light/dark correctly, and has some rendering issues

On both light/dark on qt5 mac:
- the scrollbars are being forced, when the native system ones should be used
- something strange is going on with the tab groups

On both qt5/qt6:
- the webview buttons look different

dae:
Screenshot 2023-01-16 at 8 12 03 pm
Screenshot 2023-01-16 at 8 11 37 pm
Screenshot 2023-01-16 at 8 09 33 pm

dae:
Hmm, it looks like .56/qt5 uses fusion in light mode. It didn't used to do that, so this appears to have been an accidental change as part of the other theme work. .55 appears more like in this PR, and it has the same blue lines in the tab groups, so it appears that was an existing issue

kleinerpirat:
To prevent further back and forth, I'll revert some of the logic according to your requirements above. Afterwards I'll get my hands on a friend's MacBook to test the issues myself. It will take a day or two to come up with a follow-up PR.

dae:
The joins of platform-specific styles :-( Sorry to make things harder.

kleinerpirat:
Or maybe you can figure it out yourself? There are two paths we need to choose between:

- use slightly modified system default palette (was default for macOS light theme in past versions)
- apply our own theme
  - this path also has the choice between forcing fusion and using the system default Qt theme (which is Fusion on Linux and Windows)

dae:
> For what reason? :D Nothing will change other than FOLLOW_SYSTEM actually obeying system colors.

I don't think we should be treating the "detected from system" path any differently to "manually set by user". As it stands, if my system is in light mode, when I switch between "follow system" and "light", which you'd think would be a no-op, the background color slightly changes. Could you help me understand the motivation for treating follow system differently?

dae:
The webview-based scrollbar changes still seem to be active on qt5 I'm afraid, so screens like the editor and deck list still show a permanent bar on qt5. If you don't have any immediate ideas, I can try dig into this further tomorrow.

kleinerpirat:
> Could you help me understand the motivation for treating follow system differently?

It's for the reason mentioned above. If we do not use the first path (system palette) in native mode, we end up with UI issues when the OS theme changes.

Also, it makes sense to apply the system colors in a theme mode called `Follow System`. The theme is responsible for the colors, not the style, after all.

kleinerpirat:
Edited my comment above.

dae:
Ok, that's where my confusion lay - it seems we have different ideas about what that setting means. My thinking is that it simply means "pick light or dark mode depending on if the system is currently light or dark", not "try to match the system color even if it's different to manually choosing light or dark". The former is how it behaved in the past, so we didn't end up with issues like in your screenshot. Do you strongly believe it should be the latter, even if it means differences when manually specifying light/dark?

kleinerpirat:
> Do you strongly believe it should be the latter, even if it means differences when manually specifying light/dark?

Not strongly, but I tend towards the latter. Our own palette looks off in the Fusion theme (the tab background is too bright, for example), and I imagine there could be lots of other issues with custom GTK themes like Tatsumoto uses.

I think I messed up the logic with my recent commits and confused myself, so I will need another day to get this right, I'm afraid.

dae:
No problem :-) Just to throw gasoline on the fire, I just tested switching from light→dark on Mac via the system instead of within Anki, and it results in this: 
Screenshot 2023-01-16 at 9 10 19 pm

Works fine when in the 'anki' mode, so it's just an issue with native on macOS.

kleinerpirat:
Always applying our own palette fixes those system theme issues. But in Fusion, dark mode looks unpleasant compared to the system palette.

### left: our palette, right: system default palette:



That's because Fusion uses `ColorRole.Button` for the tab background and our `button-bg` variable is relatively bright. If we set that role to use `canvas` instead, we get a result that looks much more in line with our custom `ANKI` style:

kleinerpirat:
I do not encounter any issues on Linux anymore. I'd be interested to know if this approach fixes the issues on macOS too.

dae:
Just gave it a quick test with qt6:

- when starting from native, switching light/dark via system works correctly now
- there's still a difference between follow system & explicit setting - is that something that would be easy to avoid on macOS? The "follow system" in dark mode actually looks less like the system background than "dark" does :-)
- after switching to the Anki style, switching back to Native does not restore the theme - it gets stuck on fusion

Qt5 is mostly the same, except with the forced scrollbars in the webviews

dae:
Screenshot 2023-01-17 at 12 07 52 pm
Screenshot 2023-01-17 at 12 07 40 pm

kleinerpirat:
> there's still a difference between follow system & explicit setting 

The only thing I can think of is this line:

https://github.com/ankitects/anki/blob/1be30573e1e2de4b009c72ae38316f3c17be78d3/qt/aqt/theme.py#L271

I don't know what it does, but perhaps commenting it out would fix this?

> Qt5 is mostly the same, except with the forced scrollbars in the webviews

Oh, so it wasn't about Qt scrollbars? I don't know how that would happen. There's no part where I check for the Qt version when setting the scrollbar. Perhaps there's something going wrong with `is_mac`?

dae:
I've pushed a fix for the conflicts; will try to return to this later today.

dae:
Sorry, I need to apologize about the scrollbar comments. I was in a rush when testing, and having looked at it more closely now, the same issue exists in 2.1.54/qt5. I thought that the scrollbar auto-hid in Qt 5.14, but it seems my memory was wrong. :sweat: 

I'll do some more testing now and post back in a bit. One issue I noticed so far: when switching between light/dark mode with style set to native on Linux, the distractions area becomes difficult to read.
2023-01-18T15:46:25,066641193+10:00

dae:
Well, that was fun. Qt is being really weird here - it seems like `self._apply_palette(app)` alternates between two different appearances shades for the canvas. I've checked the code and we're setting the same RGB values each time, so this seems like a Qt bug. If I modify this function to run every time the theme check is made, I get this:

```
    def apply_style_if_system_style_changed(self) -> None:
        theme = aqt.mw.pm.theme()
        if theme != Theme.FOLLOW_SYSTEM:
            return
        if True: # self._determine_night_mode() != self.night_mode:
            self.apply_style()
```

dae:
https://user-images.githubusercontent.com/180542/213108109-1fbdbba1-83b3-4e78-9c6d-1da6fb3216d6.mov

dae:
So, the question is how to resolve it. The following seems to work for me:

```diff
diff --git a/qt/aqt/theme.py b/qt/aqt/theme.py
index 9979d70b9..ca8c36c4d 100644
--- a/qt/aqt/theme.py
+++ b/qt/aqt/theme.py
@@ -61,6 +61,7 @@ class ThemeManager:
     _icon_size = 128
     _dark_mode_available: bool | None = None
     _default_style: str | None = None
+    _current_widget_style: AnkiStyles | None = None
 
     def rtl(self) -> bool:
         return is_rtl(anki.lang.current_lang)
@@ -218,16 +219,19 @@ class ThemeManager:
                 return get_linux_dark_mode()
 
     def apply_style_if_system_style_changed(self) -> None:
-        theme = aqt.mw.pm.theme()
-        if theme != Theme.FOLLOW_SYSTEM:
-            return
-        if self._determine_night_mode() != self.night_mode:
-            self.apply_style()
+        self.apply_style()
 
     def apply_style(self) -> None:
         "Apply currently configured style."
         app = aqt.mw.app
-        self.night_mode = self._determine_night_mode()
+        new_theme = self._determine_night_mode()
+        theme_changed = self.night_mode != new_theme
+        new_style = aqt.mw.pm.get_widget_style()
+        style_changed = self._current_widget_style != new_style
+        if not theme_changed and not style_changed:
+            return
+        self.night_mode = new_theme
+        self._current_widget_style = new_style
         if not self._default_style:
             self._default_style = app.style().objectName()
         self._apply_palette(app)
```

Do you think that would work? Apologies if it seems like an obvious question; I fear I haven't yet fully understood all the issues you're trying to deal with here.

Incidentally, I think we could improve on the naming - "style" and "theme" are easily confused. How about something like AnkiStyles->WidgetStyle?

kleinerpirat:
> I fear I haven't yet fully understood all the issues you're trying to deal with here.

Do you mean the whole PR? I apologize for my poor documentation. If you want, I can create a comprehensive list of changes.


> How about something like AnkiStyles->WidgetStyle?

Sounds good.

dae:
I think I'm ok with the outline of everything, it's just the specific handling of follow_system vs light/dark mode that I don't seem to have fully understood yet. Would the patch I posted above cause problems? If not, would you like me to make that change or would you like to?

kleinerpirat:
I'm getting an error on ./run after 0d945cff3:

```
    Updating crates.io index
error: failed to select a version for the requirement `async-compression = "^0.3.15"` (locked to 0.3.15)
candidate versions found which didn't match: 0.3.2, 0.3.1, 0.3.0, ...
location searched: crates.io index
required by package `anki v0.0.0 (/home/kleinerpirat/Documents/GitHub/anki/rslib)`
```

dae:
The write-up will come in handy when writing a summary for the beta anyway :-) 

Hmm, that error is strange - https://crates.io/crates/async-compression is available on crates.io. What does 'cargo --version' print?

kleinerpirat:
It prints `cargo 1.66.0`.

dae:
Presumably it prints 1.65 if you run it inside Anki's source folder? Presumably Cargo.lock hasn't been modified since checking out the latest commit? If you remove out/ and build again, does that make a difference? If not, how about if you remove ~/.cargo/registry/index?

kleinerpirat:
Deleting out/ and build didn't fix the issue, but removing `~/.cargo/registry/index` did. Anki is building right now.

Prior to the error, something had gone wrong on a build due to low system storage space. Perhaps that caused this glitch in the matrix. Thanks for the help!

kleinerpirat:
I applied your changes and it seems to work well. Regarding the Qt5 text issue: It seems a bit random, because I got different issues than you. Dark theme displays fine, but Light looks like this:

image

It seems to be isolated to the Distractions group in the preferences, so it's likely just a dirty `.ui` file :thinking:

kleinerpirat:
Now I can't reproduce it anymore :eyes:

dae:
This is looking in pretty good shape now on Mac/Linux. Are you happy for me to merge it, or was there anything you wanted to do first? 

A couple of points that will need addressing, but can be done in a follow-up PR:

- the buttons are somewhat inconsistent at the moment - see the bottom of the deck list vs preferences vs top of editor in the following screenshot
- native+dark is broken on Windows. Perhaps we should just hide the native option on Windows?

dae:
Screenshot 2023-01-18 at 8 23 46 pm

dae:
2023-01-18T20:32:53,280080207+10:00

kleinerpirat:
> Perhaps we should just hide the native option on Windows?

I hid it now. We might get requests from Windows users who'd like to set Anki to Fusion, though. I guess we'll have to see. For now I'd be very happy if you could merge this in.

tatsumoto-ren:
Thank you! It's better now.

I still see some padding at the bottom of the toolbar but it could probably be fixed by adjusting height to fit.

screenshot-2023-01-19-02-08-57

PR #2290, opened 2022-12-30
Integrate AnkiDroid's backend patches into the repo

dae:
Most of the work here was done by David in the Backend repo; integrating it here will mean AnkiDroid doesn't need to maintain a separate patch and update it as the code changes.

Based on https://github.com/ankidroid/anki/commit/5d9f262f4c1153156b3f5fc46eaf273dce980357
with some tweaks:

- Protobuf imports have been fixed to match the recent refactor
- FatalError has been renamed to AnkidroidPanicError
- Tweaks to the desktop code to deal with the extra arg to open_collection,
and exclude AnkiDroid service methods from our Python code.

@david-allison did you see my forum message?


mikehardy:
I think this would be fantastic of course. Just curious if there is a licensing fidelity issue between the Anki-Android-Backend repo where the code was, and this repo where you're hoping to re-home.

dae:
This was from the fork of the anki repo rather than Anki-Android-Backend, but it does require the standard contributor license. I reached out to David about this about 5 days ago, but haven't heard back from him yet - guess it's that time of year :-)

dae:
David got back to me, so we're good to go. :rocket:

Issue #2203, opened 2022-11-21
Multiple users can not run Anki in different Windows accounts at the same time


dae:
https://forums.ankiweb.net/t/anki-with-two-windows-accounts/24680/

I'm guessing the named pipe we use on Windows is in a global namespace instead of per-user.

```
Error 
An error occurred. Please start Anki while holding down the shift key, which will temporarily disable the add-ons you have installed. 
If the issue only occurs when add-ons are enabled, please use the Tools>Add-ons menu item to disable some add-ons and restart Anki, repeat until you discover the add-on that is causing the problem. 
When you've discovered the add-on that is causing the problem, please report the issue on the add-ons section of our support site. 
Debug info:
Anki 2.1.54 (b6a7760c) Python 3.9.7 Qt 5.15.2 PyQt 5.15.5
Platform: Windows 10
Flags: frz=True ao=True sv=3
Add-ons, last update check: 2022-11-17 06:37:32

Exception in thread Thread-2:
Traceback (most recent call last):
  File "threading", line 973, in _bootstrap_inner
  File "threading", line 910, in run
  File "aqt.mpv", line 233, in _reader
AttributeError: 'MpvManager' object has no attribute '_sock'
Exception in threading.excepthook:
Exception ignored in thread started by: <bound method Thread._bootstrap of <Thread(Thread-2, started daemon 34656)>>
Traceback (most recent call last):
  File "threading", line 930, in _bootstrap
  File "threading", line 975, in _bootstrap_inner
  File "threading", line 1244, in invoke_excepthook
AttributeError: 'ErrorHandler' object has no attribute 'flush'
Exception ignored in sys.unraisablehook: <built-in function unraisablehook>
AttributeError: 'ErrorHandler' object has no attribute 'flush'
```

Issue #1730, opened 2022-03-21
Audio cutting off early on Windows Desktop due to mpv, forking mpv while they solve this issue should fix it


Garethp:
Since Anki swapped over to mpv for Windows audio I've been having an issue with audio being cut off at the end of cards and decided to dig into it. As far as I can tell it's an issue with the audio draining for mpv not working correctly for the windows driver (and for other drivers but for this bug report I'm specifically talking about the windows driver). 

I've got an [issue logged against mpv](https://github.com/mpv-player/mpv/issues/9535) and I've got my own patchwork fix for this running locally. The patchwork solution is just to add a 150ms sleep [when the audio driver tried to shutdown](https://github.com/mpv-player/mpv/blob/master/audio/out/ao_wasapi.c#L229) and this lets windows properly play the audio before shutting the thread down. 

```c
        case WASAPI_THREAD_SHUTDOWN:
            Sleep(150);
            thread_reset(ao);
            goto exit_label;
```

The reason I'm not posting this in the forums is that I don't know how to compile mpv with all the dll's statically linked, so instead I wanted to know if you compiled mpv yourself or got it from someone else. If you compile it yourself, I can fork mpv at a specific version and you can compile from my forked repo to solve this problem while mpv works on it?

As far as I can see there have been numerous other people running into this issue posting on ankweb or reddit but because of how difficult this problem is to diagnose for someone else and how there's no clear workaround there doesn't appear to be much response to people asking for help, but I do think that this would be quite a wide-spread issue

dae:
I'd prefer to avoid having to recompile mpv if at all possible. Did you try the various -ao options, and rule out third-party software such as Sonic Studio? https://forums.ankiweb.net/t/desktop-anki-2-1-35-no-audio-playback/3990

Garethp:
I did try that, yes. This is roughly the steps I went through to try and pinpoint the issue as actually being with how mpv handles the audio driver

* I put logging in the `ao-wasapi` drivers to double check that it was making the correct calls
* Tested that just making the wasapi driver shut down 150ms slower fixes it, 
* I think I confirmed the root cause of it (MPV tells the thread with the wasapi driver to shutdown when it reaches the EOF of the sound file, but due to how wasapi works that shuts it down before all the audio is finished playing and therefore the audio gets cut off).
* Confirmed that documentation for the wasapi driver by Microsoft shows that you need to wait extra time for the audio to play before shutting down the thread, which MPV does not do
[...]

PR #2294, opened 2023-01-02
Fix two mpv issues on Windows

kelciour:
**Issue 1: Audio cutting off early on Windows** (https://github.com/ankitects/anki/issues/1730)

I attached a sample Anki deck to reproduce the issue ([mpv sound test.zip](https://github.com/ankitects/anki/files/10332470/mpv.sound.test.zip)) (after renaming .apkg to .zip). The front side of the card should play 3 identical short beep sounds but the last one will be cut short due to the bug.

As a temporary workaround, I added a bit of silence at the end with [apad](https://ffmpeg.org/ffmpeg-filters.html#apad) FFmpeg filter. After recording the output with Audacity using [WASAPI loopback](https://manual.audacityteam.org/man/tutorial_recording_computer_playback_on_windows.html), it looks like the last 50 ms will be lost, and after adding 150ms silence (just in case), I couldn't reproduce it anymore.

**Issue 2: Two Windows accounts** (https://github.com/ankitects/anki/issues/2203)

I used os.getpid() to get unique pipe names and tested it by running Anki from pip.

### DIFF
--- a/qt/aqt/mpv.py
+++ b/qt/aqt/mpv.py
@@ -143,7 +143,7 @@ class MPVBase:
         --input-unix-socket option.
         """
         if is_win:
-            self._sock_filename = "ankimpv"
+            self._sock_filename = "ankimpv{}".format(os.getpid())
             return
         fd, self._sock_filename = tempfile.mkstemp(prefix="mpv.")
         os.close(fd)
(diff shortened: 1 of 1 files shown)


dae:
Looks good - thanks as always for your work in this area Kelciour!

PR #2301, opened 2023-01-06
Fix toolbar add-on breakages and introduce toolbar tray layout & API

glutanimate:
Pending merge of #2262 ([actual diff](https://github.com/kleinerpirat/anki/compare/auto-hide-toolbar...glutanimate:anki:extend-toolbar-layout-and-addon-api)).

This PR has two main goals:

- work around breakages that add-ons using absolute positioning in the toolbar would otherwise experience after #2262
- introduce proper APIs for extending Anki's header area beyond just the center toolbar

To do so it performs the following changes:

- refactor the header layout into a grid with three areas: The existing center toolbar and two tray areas to the left and right of it
- introduce Python hooks that allow add-ons to inject their own HTML into both tray areas – these could be replaced by a JS API in the future
- add a workaround to migrate over add-on components which were previously using CSS hacks that no longer work past the toolbar redesign

AMBOSS add-on and Study Timer before these changes:

Screenshot_20230106_061001

afterwards:

Screenshot_20230106_061045

To explain the custom logic in `toolbar.ts`: I wasn't able to find a reasonable CSS-only solution that would maintain the structure of the current design, while also positioning these addon-injected elements properly. So I opted for adding this workaround as short-term solution to avoid add-on breakages, while also providing a new canonical API for the affected add-ons to slowly transition to.

These changes do fundamentally modify the composition of the toolbar body, but from what I can see, the only working add-on that would be affected by this is [Move Stats and sync buttons to the right (2.1.20)](https://ankiweb.net/shared/info/1594467075).

----

### DIFF
--- a/qt/aqt/data/web/css/toolbar.scss
+++ b/qt/aqt/data/web/css/toolbar.scss
@@ -27,7 +27,7 @@
     align-self: start;
     display: flex;
     flex-direction: row;
-    align-items: center;
+    align-items: start;
 }
 
 .toolbar {
--- a/qt/aqt/data/web/css/toolbar.scss
+++ b/qt/aqt/data/web/css/toolbar.scss
@@ -6,8 +6,33 @@
 @use "sass/elevation" as *;
 @use "sass/button-mixins" as button;
 
+.header {
+    height: 41px;
+    display: grid;
+    grid-template-columns: repeat(3, 1fr);
+    align-items: start;
+    align-content: space-between;
+}
+
+.left-tray {
+    justify-self: start;
+}
+
+.right-tray {
+    justify-self: end;
+}
+
+.left-tray,
+.right-tray {
+    align-self: center;
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+}
+
 .toolbar {
-    display: inline-block;
+    height: 31px;
+    justify-self: center;
     white-space: nowrap;
     overflow: hidden;
     border-bottom-left-radius: prop(border-radius-large);
--- a/qt/aqt/data/web/css/toolbar.scss
+++ b/qt/aqt/data/web/css/toolbar.scss
@@ -24,7 +24,7 @@
 
 .left-tray,
 .right-tray {
-    align-self: center;
+    align-self: start;
     display: flex;
     flex-direction: row;
     align-items: center;
--- a/qt/aqt/data/web/js/toolbar.ts
+++ b/qt/aqt/data/web/js/toolbar.ts
@@ -27,3 +27,58 @@ function updateSyncColor(state: SyncState) {
             break;
     }
 }
+
+// Dealing with legacy add-ons that used CSS to absolutely position
+// themselves at toolbar edges
+
+function isAbsolutelyPositioned(node: Node): boolean {
+    if (!(node instanceof HTMLElement)) {
+        return false;
+    }
+    return getComputedStyle(node).position === "absolute";
+}
+
+function isLegacyAddonElement(node: Node): boolean {
+    if (isAbsolutelyPositioned(node)) {
+        return true;
+    }
+    for (const child of node.childNodes) {
+        if (isAbsolutelyPositioned(child)) {
+            return true;
+        }
+    }
+    return false;
+}
+
+function getElementDimensions(element: HTMLElement): [number, number] {
+    const widths = [element.offsetWidth];
+    const heights = [element.offsetHeight];
+    for (const child of element.childNodes) {
+        if (!(child instanceof HTMLElement)) {
+            continue;
+        }
+        widths.push(child.offsetWidth);
+        heights.push(child.offsetHeight);
+    }
+    return [Math.max(...widths), Math.max(...heights)];
+}
+
+function moveLegacyAddonsToTray() {
+    const rightTray = document.getElementsByClassName("right-tray")[0];
+    const links = document.querySelectorAll<HTMLAnchorElement>(".toolbar > *");
+    const legacyAddonElements: HTMLAnchorElement[] = Array.from(links)
+        .reverse()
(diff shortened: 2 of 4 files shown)


glutanimate:
Demo add-on showcasing use of existing toolbar APIs (including hacky use that's corrected by this PR) and new toolbar APIs:

Screenshot_20230106_063648



```python
from typing import Callable

from aqt.gui_hooks import (
    top_toolbar_did_init_links,
    top_toolbar_will_set_left_tray_content,
    top_toolbar_will_set_right_tray_content,
)
from aqt.toolbar import Toolbar

# Add-ons using top_toolbar_did_init_links as intended:

Handler = Callable[[list[str], Toolbar], None]


def on_top_toolbar_did_init_links_factory(
    label: str,
) -> Handler:
    def on_top_toolbar_did_init_links(links: list[str], toolbar: Toolbar):
        my_link = toolbar.create_link(
            cmd=label, label=label, func=lambda: print(label), tip=label, id=label
        )
        links.append(my_link)

    return on_top_toolbar_did_init_links


on_top_toolbar_did_init_links_hacky = on_top_toolbar_did_init_links_factory(
    label=f"normal_addon"
)
top_toolbar_did_init_links.append(on_top_toolbar_did_init_links_hacky)


# Add-ons using top_toolbar_did_init_links to inject absolutely positioned
# elements to the toolbar:


def on_top_toolbar_did_init_links_factory_hacky(
    label: str,
) -> Handler:
    def on_top_toolbar_did_init_links_hacky(links: list[str], toolbar: Toolbar):
        my_element = (
            f"""<div style='position: absolute; right: 4px; top: 4px;'>{label}</div>"""
        )
        links.append(my_element)

    return on_top_toolbar_did_init_links_hacky


on_top_toolbar_did_init_links_hacky = on_top_toolbar_did_init_links_factory_hacky(
    label=f"hacky_addon"
)
top_toolbar_did_init_links.append(on_top_toolbar_did_init_links_hacky)

# Add-ons using new APIs to add components to the right and left tray,
# respectively:


def on_top_toolbar_will_set_tray_content_factory(label: str) -> Handler:
    def on_top_toolbar_will_set_tray_content(content: list[str], toolbar: Toolbar):
        my_content = f"""<button>{label}</button>"""
        content.append(my_content)

    return on_top_toolbar_will_set_tray_content


for i in range(2):
    on_top_toolbar_will_set_tray_content = on_top_toolbar_will_set_tray_content_factory(
        label=f"new_addon{i}"
    )

    if i % 2 == 0:
        top_toolbar_will_set_left_tray_content.append(
            on_top_toolbar_will_set_tray_content
        )
    else:
        top_toolbar_will_set_right_tray_content.append(
            on_top_toolbar_will_set_tray_content
        )

```

dae:
If you're confident the workaround for legacy add-ons will fix more than it breaks, that works for me. The new hooks make sense in isolation, though there's a good chance a future one-webview refactor will stop them from working. Would it perhaps be worth adding a disclaimer in the docstrings that this solution is provided for now, and a future main screen refactor will probably require changes? Though I guess you could argue such a disclaimer applies to many of the hooks.

glutanimate:
> If you're confident the workaround for legacy add-ons will fix more than it breaks

I think so, yes. I've just run another few tests with pretty much all still working add-ons using `top_toolbar_did_init_links`   and ones monkey-patching the toolbar directly (though these are more difficult to find) and wasn't able to find any breakages.

> Would it perhaps be worth adding a disclaimer in the docstrings that this solution is provided for now, and a future main screen refactor will probably require changes?

Done!

> Though I guess you could argue such a disclaimer applies to many of the hooks

Have you had a chance yet to think through how a primarily JS-based add-on API could look like for `mw` in the future? I really think it's crucial that we have this fleshed out before shipping the eventual one-webview refactor to users. Other screens which have been ported to Svelte have been fairly painful to work with for add-on authors, ranging from no clear APIs (graphs screen, congrats screen) to ones which are fairly good, but still very much in flux, and suffering from lack of documentation (editor, deck options). For the latter, at this point only a handful of add-ons or so actually make use of these APIs which I don't think is a good sign.

Given that `mw` likely is the UI that's targeted the most by add-ons, I think it would be very valuable both for Anki contributors and add-on authors to set some time aside and brainstorm on how extensibility should look like in the future, so that e.g. the `mw` refactor could be designed from the ground up with those ideas in mind.

(Plainly, I fear that, without going into this with a clear plan, we will end up with a chaotic situation of wild DOM-modifying code, hacky Svelte lifecycle boilerplate, and a lot of unnecessary Python glue just to inject content scripts via `web.eval()`. And to users, this would manifest as a pandora's box of add-on incompatibilties, frequent breakages, and general instability.)

glutanimate:
@noobj2 quick heads-up :exclamation: This PR will break [Move Stats and sync buttons to the right](https://ankiweb.net/shared/info/1594467075), but you should be able to use the new tray hooks (together with `top_toolbar_did_init_links`) to implement the same changes to the UI without the need for monkey patching the toolbar body composition.

dae:
"Soon" is relative - it'll probably be at least a few months before I have the time and energy to properly focus on the TS/Svelte side of things again, and I'd like us to address some of the performance/usability regressions and take stock of things before we move ahead with much more.

dae:
(Though I should add, just getting something implemented is hard enough, without the added burden of making it extensible from the get-go. So it is nice when add-on developers actually contribute code to implement the necessary hooks, as you've done here and on previous PRs)

glutanimate:
@dae I hear you! Hoping to put more time into that this year and just be more active on Anki development in general. And let me also say that the – oftentimes invisible – work you and many contributors put into maintaining add-on compatibility is also very much appreciated.

Just rebased the PR and tested everything again. Should be good to go now.

dae:
:rocket:

PR #2306, opened 2023-01-09
Fix negated `introduced:x` search

RumovZ:
https://forums.ankiweb.net/t/anki-2-1-56-release-candidate/26001/14


dae:
Thanks for spotting the other issue as well :-)

PR #2307, opened 2023-01-09
Fix spaces not handled properly when entering text around MathJax

hikaru-y:
https://forums.ankiweb.net/t/2-1-56-qt6-pressing-space-after-mathjax-inserts-nbsp-with-escaped/26147

I overlooked that the `data` property of a text node and the `innerHTML` value of its parent node are different when the text node contains whitespaces.


dae:
Thank you for the quick fix Hikaru!

PR #2308, opened 2023-01-10
Vertically center text inside toolbar

kleinerpirat:
After #2301 I found the text to be a little too low inside the toolbar. This tweak centers the text:

## Before
image

## After
image


glutanimate:
Good eye, thanks!

PR #2314, opened 2023-01-11
Make QCheckBox more visible

kleinerpirat:
Uses `border` instead of `border-subtle` for the border
and `canvas-elevated` instead of `canvas-inset` for the background.

|  Before | After   |
|---|---|
| image  |  image  |
| image |  image |
|  image  | image  |

In the third screenshot (Preferences), visibility is actually slightly decreased on the different background, but the stronger border partly compensates for that.


user1823:
What about using a white colour for the border? It would make it stand out even more.

PR #2319, opened 2023-01-14
Update Options Groups dialog

tatsumoto-ren:
I noticed that there are additional Options Groups settings that weren't added to the Options Groups dialog, so I have added them in.

screenshot-2023-01-14-00-40-23
screenshot-2023-01-14-00-40-26

I'm not sure why [here](https://github.com/ankitects/anki/blob/main/pylib/anki/consts.py#L115) a dict is returned instead of a list, but I have followed the pattern anyway.

Added missing tooltips, except for one which has no appropriate translation. Not sure what to do with it.

There's a new field called "custom scheduling", but I couldn't find any references to it in the Options Groups dict. Maybe it would be better to have it under Preferences instead?

It's a little annoying to have to press <kbd>shift</kbd> every time, so I have added a checkbox that swaps the Webview-based dialog with the Qtwidgets-based dialog.

Finally, the dialog used to always fail to save its geometry. I have moved `saveGeom()` under `done()`

### DIFF
--- a/ftl/core/preferences.ftl
+++ b/ftl/core/preferences.ftl
@@ -49,3 +49,4 @@ preferences-minutes-between-backups = Minutes between automatic backups:
 preferences-reduce-motion = Reduce motion
 preferences-reduce-motion-tooltip = Disable various animations and transitions of the user interface
 preferences-collapse-toolbar = Hide top bar during review
+preferences-enable-webview-based-deck-options-dialog = Use WebView-based Options Groups dialog
--- a/pylib/anki/consts.py
+++ b/pylib/anki/consts.py
@@ -142,6 +142,19 @@ def new_gather_priority_choices(
     }
 
 
+def new_sort_order_choices(
+    col: anki.collection.Collection | None,
+) -> dict[int, Any]:
+    tr = _tr(col)
+    return {
+        0: tr.deck_config_sort_order_template_then_gather(),
+        1: tr.deck_config_sort_order_gather(),
+        2: tr.deck_config_sort_order_card_template_then_random(),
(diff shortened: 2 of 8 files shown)


dae:
I wish you'd reached out prior to working on this. I'm afraid this PR is not the direction I'd like to take things. The legacy deck options screen has only been kept around for the benefit of add-ons that use it, and it will eventually be removed once we have better documentation on how to update add-ons to use the new screen and have given them time to migrate over. The same applies to the old graphs - they are effectively frozen in time and will be removed in the future, so it does not make sense to sink any further engineering time into them.

tatsumoto-ren:
I hope that the old dialog stays longer. If it is removed, it's going to be sad. The new deck options dialog doesn't really satisfy everybody, right? It's very large and essentially takes up the whole screen. If people use the old one, I think we could at least provide the newly added options to them as well for the time being.

dae:
Updating the old screen means the risk of possible regressions, and means we need to make changes in two screens moving forward. I'm afraid I don't think it's worth it. If you're happy to spend time maintaining a parallel implementation, you're of course welcome to release these changes in an add-on - either by dynamically adding the items to the legacy code, or by providing a completely replaced screen that copy/pastes the existing code.

tatsumoto-ren:
I couldn't find a way to reach out prior. The forum is meant for solving issues, and github seems to be the only other way.

I guess if the work has already been done, "any further engineering" is not required. And if you want to delete it later, at least it will be up to date and could be converted to an addon instead. Though I think the old dialog has always done the job better than what's default now.

dae:
The forum has a 'development' section you could use, or creating an issue here about a potential future PR would be fine too.

Issue #2305, opened 2023-01-09
Investigate switch to difflib


dae:
1e0be26b7e11640c6381f7bdf2f686275a21ed08 migrated from the Python difflib to the dissimilar crate so we can use the code across all platforms, but it unfortunately does not seem to perform as well for some languages/inputs. There appears to be a Rust port of difflib available, so it might be worth looking into whether we can use that instead, if it gives better output.

https://crates.io/crates/difflib

https://forums.ankiweb.net/t/error-highlighting-after-typing-the-answer/25755/6

@RumovZ feel free to take this if it interests you

dae:
We'll want to look into this as a matter of priority, as it appears that dissimilar will panic with certain non-Latin text input, eg

```
        let ctx = DiffContext::new("Сущность должна быть ответственна только за одно дело", "Single responsibility Сущность выполняет только одну задачу. Повод для изменения сущности только один.");
```

RumovZ:
To be clear, that means the investigation part is over and we want to switch without question now?

dae:
If the library isn’t suitable, there’s still the option of finding another
one, or trying to fix the panic in the one we currently use as a last resort

dae:
Closed via #2322

popyoung:
Still incorrect in 2.1.57

dae:
Please create a new issue and describe the problems you're having/provide reproduction steps.

dae:
Nevermind, I see you've posted in the forums. Please provide a card ID and the exact text you're entering, so I can try to reproduce the problem.

https://forums.ankiweb.net/t/bug-cloze-correction-doesnt-work-as-expected/26302/3

PR #2321, opened 2023-01-14
Replace dissimilar crate with difflib

RumovZ:
Here is an experiment with difflib. It makes a lot of allocations:
1. A new string is created for every `Opcode`. 😮 
2. It only works with slices, so we have to create a `Vec<char>`.
3. For the output, I currently collect the chars into a string again, but this can be optimised to slicing the original strings.

However, I guess performance is not of the essence here and it works well enough. I've added unit tests for the two issues with dissimilar mentioned in #2305 and difflib handles them fine.

WDYT? Should I clean this up for a merge?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -92,7 +92,7 @@ dependencies = [
  "convert_case 0.6.0",
  "criterion",
  "csv",
- "dissimilar",
+ "difflib",
  "env_logger",
  "flate2",
  "fluent",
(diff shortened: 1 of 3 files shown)


dae:
I'm not too worried about allocations, as we only ever compare one string at a time, and `Vec<char>` at least avoids the possibility of slicing into the middle of a character - sounds good to me.

I'm not strongly set on that missed_chars_only_shown_in_provided_when_after_good() behaviour - if you want to revisit that/think it makes sense to do differently in light of comments on the forum, feel free :-) I seem to recall I originally changed it because in certain inputs, it made things confusing, but I don't remember why - maybe I wasn't keeping track of input vs output accumulated length at the time.

RumovZ:
I'm not sure I'm following. Please let me know if I'm missing a point, but the current behaviour seems odd in this case:

`56` is filled in, but `34` isn't. I checked and this is also the behaviour on main. I have assumed that this is a side effect of avoiding something like this (main without that if condition):
1200056 without if
However, it's not necessary in this PR because for difflib this is not an 'insert', but a 'replace'. This is how it looks on this branch *without* the if condition:
grafik grafik

It seems reasonable to use the expected text instead some arbitrary character like `-`, but I've never used type-in cards, so what do I know? 🙂

I'll go ahead with this PR then. The `-` thing will be a trivial change either way.

PR #2322, opened 2023-01-14
Replace dissimilar crate with difflib

RumovZ:
More polished version of #2321.


dae:
Looks great - thanks Rumo!

PR #2323, opened 2023-01-14
Create field options to store media file path

evandrocoan:
Just opened and closed this pull request to save the code somewhere. A better solution was found after writing this pull-request. The new solution would be directly embedding media on the page, the same way images are. For example, I can embed images with ``, therefore I can also embed other HTML5 media with `<audio src='some.mp3'></audio>`.

Initially, this pull-request would be a replacement for https://github.com/ankitects/anki/pull/540 - Added arguments as `[sound:filename|argument1 argument2...]`. Instead of creating options for existent `[sound:file]` tags, it would allow marking a field to store media only.


evandrocoan:
Closing as this pull-request is not required anymore, as https://github.com/ankitects/anki/pull/604 audio elements can already be used with Anki standard tools for handling `` tags.

PR #2326, opened 2023-01-17
Explicitly set web view to visible

### LINKED ISSUE #2285: Add a ctrl+enter shortcut to proceed with csv import
https://forums.ankiweb.net/t/home-ui-is-fixed-in-position/25682

blueputty01:
This PR attempts to resolve #2285 by proposing:

- Explicitly setting web view of CSV import window to visible

I added a `self.web.setVisible(True)` call after `self.show()` in the constructor instead of changing `self.web.setVisible(False)` to `self.web.setVisible(True)` in the `_setup_ui` function since the former seems to more closely follow the existing design patterns.


dae:
Thanks for looking into it! The reason we don't call show() immediately is because we want to wait for the page to load first, which can reduce flicker.  Your change does appear to fix the problem, but removes the delayed showing we had previously, so a different approach might be better - I'll push one now.

PR #2331, opened 2023-01-18
Bump Linux theme refresh timer back up to 5 minutes

glutanimate:
On Linux distributions that do not yet support `org.freedesktop.appearance.color-scheme` (most distros released prior to 2022), querying DBus for the missing property produces verbose journal entries over Anki's entire runtime.

I first noticed this in the stdout and was considering to submit a PR that would simply gate printing the error behind an env var, but then discovered that `dbus-send` will also log errors to the journal. Because this happens every five seconds now, this is a bit suboptimal.

Stdout:

Screenshot_20230118_130422

System journal entries:

Screenshot_20230118_123303-1

Admittedly, this is a bandaid solution, but it's what we were using previously, so I think it's ok to opt for a conservative timer again. We can potentially follow up with a proper solution that first checks whether the corresponding DBus property exists before querying it (started looking into this a bit, but it's not incredibly straightforward).

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -1074,8 +1074,10 @@ title="{}" {}>{}</button>""".format(
         theme_manager.apply_style()
         if is_lin:
             # On Linux, the check requires invoking an external binary,
+            # and can potentially produce verbose logs on systems where
+            # the preferred theme cannot be determined,
             # which we don't want to be doing frequently
-            interval_secs = 5
+            interval_secs = 300
         else:
             interval_secs = 2
         self.progress.timer(


dae:
Sounds like the best course of action for now - shame we can't just redirect it to /dev/null. Thanks Aristotelis.

@kleinerpirat heads up

glutanimate:
@dae or @kleinerpirat, does the following command succeed for you? (don't have a 2022+ Linux VM handy atm to check)

```bash
dbus-send --print-reply --dest=org.freedesktop.DBus  /org/freedesktop/DBus org.freedesktop.DBus.ListNames | grep appearance
```

If so, we could use this to check whether the property is actually available.

dae:
I'm on Debian 11, which doesn't appear to have that text.

dae:
I may not have understood the question? ListNames returns a bunch of text, but 'appearance'  is not in the list. The dbus query Anki runs doesn't work on my system, so that seems to be what you expected?

glutanimate:
Sorry, the confusion was on my part. Somehow I had it set in my mind that Debian 11 was newer than it actually is. `appearance` not being grepped is what I expected on older distros, yep.

However, giving this a spin myself now on a KDE Neon image, which does support the DBus appearance query, it looks like the command doesn't work in the first place. `org.freedesktop.appearance` is not an actual DBus interface, just the name of a property returned by `org.freedesktop.portal.Settings`'s interface methods. As `org.freedesktop.portal.Settings` does not have a method that would allow getting all property keys without trying (and failing) to read them, I think we're out of luck here in terms of easy solutions.

I did try calling `ReadAll` instead, but while it doesn't return an error, it still prints a warning to the journal. Here's the corresponding command:

```bash
dbus-send --session --dest=org.freedesktop.portal.Desktop --type=method_call --print-reply /org/freedesktop/portal/desktop org.freedesktop.portal.Settings.ReadAll array:string:"org.freedesktop.appearance"
```

`Settings.version` is also the same for implementations that do support `org.freedesktop.appearance` and do not.

So I'm a bit at a loss here of how we can determine whether `org.freedesktop.appearance` is supported without raising errors/warnings. I'm also not sure if the errors/warnings printed to the journal is a KDE-specific problem, which could very well be the case as the log entries seem to explicitly mention KDE's implementation of the freedesktop portal.

If that were the case, we could perhaps add a heuristic that would check whether the distro is running KDE, and if so check for Plasma >= 5.24 before querying the appearance property.

dae:
@glutanimate that's a shame, but thanks for digging into it.

@kleinerpirat Is it really useless? If the system automatically switches at a certain time of day, the user may not even notice the delay if Anki happens to be closed/obscured by other windows.

Your icon is pretty, but is this important enough that we want to be taking up valuable eyeball real-estate any time the user is looking at the main window? Can you think of some popular apps that have such a toggle placed so prominently? What about placing it in the view menu instead?

kleinerpirat:
I find it off-putting when I switch the OS theme manually and the window decorations respond, but Anki itself doesn't. It feels like a bug. Because I know Anki can do better, I always have the urge to manually adjust the theme to fit with the rest :sweat_smile: 
image

As for scheduled theme switches, I think your argument only holds when users are in full-screen study mode (and only have one monitor). Outside of full screen one would notice because the window decorations change immediately.

> is this important enough that we want to be taking up valuable eyeball real-estate any time the user is looking at the main window?

I'd make it as small and subtle as possible. The only mode where it could grab attention too much is review and we just added an option to hide the toolbar. I do not know of other apps that have such a button, but a lot of websites do. I might be in a minority to whom this seems useful, but I'd like to create a proposal and ask the community.

dae:
My point about automatic mode is that if the user starts up Anki, it should pick up the correct color immediately, and if they don't happen to be looking at the window when a change occurs (eg because they're in another app that's on top, or they're in a different desktop/workspace, or they're away from the computer), then they may well not notice the delayed change either. So I wouldn't call it useless (which implies it has no value and should be removed). Not ideal, sure, but not useless :-)

If the suggestion to add a toggle in the main window is to work around the slow update period, I think that might be the wrong way to solve that problem. Until we can handle everything automatically, maybe an env var or temporary preference to alter the update frequency would be a solution for users who are not concerned with/experiencing the log spam?

Regarding the toggle in the main window, by all means discuss it with the community - just bear in mind the answers there will tend to be biased towards power users who are forum regulars, and newcomers who are already overwhelmed with the options present in the main window may not be adequately represented.

Personally, I'm not sure placement in the main screen is a great idea. The main webpages I can think of with dark toggles in a prominent place are ones that tend to be focused on reading (such as doc websites, RSS readers and so on), and they tend not to have separate preferences or a menubar where such an option could be placed (at least in my limited experience). Wouldn't adding a section to the View menu be sufficient? Power users could access it with a shortcut key, and it means we don't clutter the UI with something many users may never need/use.

glutanimate:
Regarding the switch in general:

I do like the design of the toggle (it looks nice, and the non-auto state is clever!), but I would tend to agree with Damien in that I'm not sure if users would find a lot of utility in having this visible all the time. It seems to me like the trend is moving more towards apps and websites implicitly following the system theme and switches like this being slightly more hidden, e.g. tucked away behind a hamburger/dropdown menu. Our closest equivalent to that in Anki would be the View menu IMHO.

Regarding auto mode on Linux:

I think we should test whether the log spam is also an issue on GNOME, and then decide on whether we want to ship some DE-specific heuristics for KDE (and possibly GNOME). I've only researched this briefly, but it looks like there are two main implementations of `xdg-desktop-portal`, `xdg-desktop-portal-kde` (used by KDE) and `xdg-desktop-portal-gnome` (used by GNOME, XFCE, and others). So after we've tested this on GNOME, we can probably be fairly confident about how checking for the theme behaves on the majority of Linux set-ups, which – after potentially introducing some DE-specific logic – would allow us to bump up the frequency of checks again. Should this turn out to be an issue for more exotic set-ups, then we'd still have the option to go for the env var approach.

kleinerpirat:
> I think we should test whether the log spam is also an issue on GNOME

I am on GNOME 43.2 and the build of 943dddf28 (before this PR was merged) doesn't show log spam for me in stdout or systemd.

PR #2334, opened 2023-01-18
Fix non-Latin text wrapping in deck options dropdown

### LINKED ISSUE #2316: Investigate non-Latin text wrapping in deck options dropdown
https://forums.ankiweb.net/t/anki-2-1-56-release-candidate/26001/30

winterhazel:
Fixes #2316.


dae:
Thanks for looking into this Fabricio!

PR #2335, opened 2023-01-18
fix: remove empty menubar for addcards (#2189)

### LINKED ISSUE #2189: The Add dialog doesn't have the Edit menu on macOS
I'm using Anki v2.1.54 Qt6 on macOS 13.0. There's no **Edit** menu on the **Add** dialog. This is unfortunate because the **Emoji & Symbols** shortcut — **Control-Command-Space bar** — doesn't work without it. The dialog that it's supposed to bring up is very useful when adding notes with diacritics.

There's the **Edit** menu and the **Emoji & Symbols** shortcut on the **Browse** dialog, so it works there.

altafard:
Fix #2189

Removing the empty menu bar allows you to use the menu from the parent widget. As a result  It seems that this menubar is a remaining from #1655.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -108,6 +108,7 @@ Yutsuten <mateus.etto@gmail.com>
 Zoom <zoomrmc+git@gmail.com>
 TRIAEIOU <github.com/TRIAEIOU>
 Stefan Kangas <stefankangas@gmail.com>
+altafard <altafard.dev>
 
 ********************
 
--- a/qt/aqt/forms/addcards.ui
+++ b/qt/aqt/forms/addcards.ui
@@ -82,16 +82,6 @@
     </item>
    </layout>
   </widget>
-  <widget class="QMenuBar" name="menubar">
-   <property name="geometry">
-    <rect>
-     <x>0</x>
-     <y>0</y>
-     <width>750</width>
-     <height>22</height>
-    </rect>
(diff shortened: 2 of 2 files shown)


dae:
Unfortunately I think this may re-introduce the problems mentioned here: https://forums.ankiweb.net/t/anki-2-1-50-apple-silicon-add-new-cards-window-does-not-behave-normally/18929/4

altafard:
I'm not sure that I understood your point. Seems that ticket you shared describes exactly the same problem with menu bar. By the way, I couldn't reproduce the closing issue with CMD+W either on the main branch or with my changes. Removing useless markup for menubar in addcards screen may help to show Edit tab at least. Please, correct me if I'm wrong.

dae:
For example, if you open the browse screen, and then switch to the add cards screen, cmd+d will change the deck of the selected cards in the browse screen, instead of switching the target deck which new cards will be added to.

altafard:
Oh yeah, now I got you. That's sad, it looked like an easy fix. Any insides why menu actions were removed from add cards screen (according to #1655)?

dae:
Are you sure #1655 caused a change in behaviour here? The way I remember it is we switched the add cards screen to be a main window so it defines its own (empty) menubar, instead of displaying whatever window was selected previously. That was done to fix the issue mentioned in the forum post I linked above.

PR #2336, opened 2023-01-18
Preferences: Hide style label too on Windows

kleinerpirat:
Forgot about that.

https://forums.ankiweb.net/t/anki-2-1-57-beta/26494/9


dae:
I missed it too :-)

PR #2337, opened 2023-01-18
Add separate styling for template editor QGroupBoxes

kleinerpirat:
Point 3 of this comment: https://forums.ankiweb.net/t/anki-2-1-57-beta/26494/4

I thought since those two instances of `QGroupBox` are a bit of an edge case, we can simply remove the background and border definitions and things will look normal again.


dae:
Thanks Matthias - seems like a reasonable solution if this is the only place it's causing problems. If further problems are found in the future, we might want to revisit this.

PR #2338, opened 2023-01-18
Change names of edited translations to notify translators

kleinerpirat:
https://forums.ankiweb.net/t/anki-2-1-57-beta/26494/8

This should work, right?


dae:
Yep. It's a bit borderline whether we need to get the translators to translate the other ones that didn't have the HTMl issue, as they were minor tweaks, but all good.

PR #2343, opened 2023-01-22
Improve invalid argument info message

evandrocoan:


dae:
That reads "...to import this file.: /some/path", which looks a bit awkward. How about " (path)" at the end instead?

(A proper fix would be to update the translation, but this is a niche issue, so I'm ok with a hack here)

evandrocoan:
I updated it (path). Was this what you meant?

dae:
Yep, thanks!

PR #2345, opened 2023-01-23
Use a ninja variable for Protoc binary

krmanik:
Use installed local protoc binary instead of downloaded version.


dae:
LGTM pending a successful CI

dae:
Thanks Mani. That svelte file should have been caught earlier; I will investigate.

PR #2350, opened 2023-01-26
Fix 'View Files' not working with a relative base folder

abdnh:
Probably introduced in 0570cfdf48dbab21e561bcbc0516721d08bdc025
I use the `-b` option often enough to notice this.


dae:
Looks good, thanks Abdo!

PR #2351, opened 2023-01-26
Allow burying cards in browser

### LINKED ISSUE #2295: Allow burying of cards in the browse screen
https://forums.ankiweb.net/t/bury-cards-from-browser/25829

kaben:
Resolves #2295. This code is based on existing "toggle suspend" command in browser.

- Adds "toggle bury" command to browser cards menu.
- Adds "browsing-toggle-bury" to core translation. Only English-language.
- Adds "buried" coloring to rows for buried cards in browser table.

Not yet done:

- Needs discussion: keyboard shortcut for "toggle bury" action.
- Needs non-English translations.

### DIFF
--- a/rslib/src/browser_table.rs
+++ b/rslib/src/browser_table.rs
@@ -575,7 +575,9 @@ impl RowContext {
                         Color::Marked
                     } else if self.cards[0].queue == CardQueue::Suspended {
                         Color::Suspended
-                    } else if self.cards[0].queue == CardQueue::UserBuried || self.cards[0].queue == CardQueue::SchedBuried {
+                    } else if self.cards[0].queue == CardQueue::UserBuried
+                        || self.cards[0].queue == CardQueue::SchedBuried
+                    {
                         Color::Buried
                     } else {
                         Color::Default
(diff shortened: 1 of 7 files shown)


rslib/src/browser_table.rs
```diff
@@ -575,6 +575,10 @@ impl RowContext {
                         Color::Marked
                     } else if self.cards[0].queue == CardQueue::Suspended {
                         Color::Suspended
+                    } else if self.cards[0].queue == CardQueue::UserBuried
+                        || self.cards[0].queue == CardQueue::SchedBuried
+                    {
```
dae, line 580: This section could be simplified a little with:

```rust
                _ => {
                    if self.note.is_marked() {
                        Color::Marked
                    } else {
                        match self.cards[0].queue {
                            CardQueue::Suspended => Color::Suspended,
                            CardQueue::UserBuried | CardQueue::SchedBuried => Color::Buried,
                            _ => Color::Default,
                        }
                    }
                }
```

dae (commented):
Thanks for contributing Kaben!

Regarding the shortcut, maybe ctrl+shift+j? Other ideas welcome.

You only need to add the English translation for now - once it's added and the translations synced, it will become available for translators to translate.

rslib/src/browser_table.rs
kaben: I like ctrl-shift-j for "toggle bury". It's similar to ctrl-j for "toggle suspend". Commit suggested changes.

dae:
Thanks Kaben!

PR #2354, opened 2023-01-29
Fix deck options config selector's starting value

winterhazel:
This PR fixes a visual issue.

The Select component's value prop always started as 0 here. This meant that, when opening the deck options window, the first option would always show as active (has a blue highlight on the dropdown), even when the current entry was another option.


dae:
Thanks for contributing another fix Fabricio! LGTM.

PR #2356, opened 2023-01-30
Refactor CSS/image preloading; implement custom font preloading

### LINKED ISSUE #2355: Preload custom fonts?
Perhaps we could load font files in advance, in a similar way to @hikaru-y's code that loads images in advance in images.ts

https://forums.ankiweb.net/t/bug-font-flicker-on-reviews/26912

hikaru-y:
Closes #2355

This is currently an early draft. I have done a little testing with a card that has multiple custom fonts with very large file sizes, and it seems to work as expected.

( I will not have access to my main development PC for a week or two starting tomorrow. After that period, I'll try to refine this PR, if @dae or someone else hasn't completed another implementation by then. I may not be able to log into GitHub during that period, so I may not be able to reply here. )


ts/tsconfig.json
```diff
@@ -23,7 +23,7 @@
         "target": "es2019",
         "module": "es2020",
         "lib": [
-            "es2017",
+            "es2020",
```
dae, line 26: Could we make this es2020.string instead? String.matchAll appears to be supported in Qt 5.14 + iOS 14, but I'm not sure all of es2020 is.

dae (commented):
Thanks for looking into this Hikaru!

ts/tsconfig.json
hikaru-y: Yes, sure we can. I will update it that way.

dae:
Was much required to push this over the finishing line?

hikaru-y:
Actually, regarding font preloading, the functionality is almost complete with the current code. However, I would like to clean up the entire preloading code, including image and CSS preloading, but there were a few areas where I was a little confused as to how best to do this. I was also testing how long the timeout should be for image preloading on the front side of a card, as the current timeout setting of 100ms sometimes causes preloading to fail on low spec machines. I have also been putting off working on this PR by prioritizing work on a few PRs I recently submitted.

I will try to complete this PR in the next few days.

dae:
No rush Hikaru, just wanted to give you a nudge if you'd happened to have forgotten about it. :-) A 2.1.63 release is going to be at least a few weeks away.

dae (commented):
Thanks for your hard work Hikaru, the code looks nice and clean. 

Re Jest, I've run into troubles with it in the past too. Unfortunately a solution closer to a browser would probably be a lot more heavyweight, as it would probably have to invoke chromium or similar.

PR #2357, opened 2023-02-01
droid-ify when nothing {{type}}d

twwn:


dae:
Thanks nwwt. While reviewing this, I noticed that we probably want to be doing the check on the Rust end instead, as that will ensure each platform behaves identically. Can you confirm that the latest main branch behaves as you wanted?

@RumovZ as an aside, while testing this I noticed that providing "1123" to a correct answer of "123" would not show both provided and expected, which I've attempted to fix with https://github.com/ankitects/anki/commit/32ec6e06f6dce8ab70e5b63b4656186c2b94a4db. Does that work for you, or was there some use case you were trying to address with `no_mistakes()` that I've overlooked?

twwn:
Oh, I didn't look at the Rust part. That works too, ofc.

RumovZ:
I didn't write that part and the behaviour doesn't seem to have changed with the migration to difflib.

dae:
Sorry, I seem to have completely forgotten writing that function ^_^;

PR #2360, opened 2023-02-03
Support UTF-8-BOM when importing CSV

RumovZ:
https://forums.ankiweb.net/t/control-csv-file-import-via-header-key-value/27031/

By the way, I noticed that it is possible for the user to override the delimiter configured in the file. I'm sure this wasn't possible before and overriding the HTML flag is still disallowed. Did we change this deliberately and I forgot about it?


dae:
Great, thanks for this Rumo. I've logged the issue on https://github.com/ankitects/anki/issues/2362

PR #2361, opened 2023-02-04
Disable burying of previously gathered cards

### LINKED ISSUE #2352: Bury order in v3 queue build is incorrect, leading to "bug: card modified" messages
I suspect the issue mentioned on https://forums.ankiweb.net/t/card-modified-without-updating-queue/26052 was caused by the following configuration:

- bury reviews: yes
- bury new: no
- mix reviews and new cards

In gathering.rs, we gather the review cards first, and check if they should be buried at that point. We then gather the new cards. Only after gathering do we sort the cards in the order they'll appear in to the user. If a new sibling happens to be shown first, that will result in the review card being buried, but it's not removed from the queue, as the code expects it to have never been included.

To solve this, we'll presumably need to shift the bury check to apply as we collect the merge_new() iterator into a vec, so that we bury in exactly the same order as cards will appear. I think we may be ok with keeping the intraday learning code as is, since it comes first, and intraday cards are not subject to burying themselves.

@RumovZ interested in tackling this one?

RumovZ:
Closes #2352.

By the way, I noticed that you've added a deprecation warning to some translation strings, so they don't get translated needlessly. Should I add a binary so you can automate this whenever updating translations? Shouldn't be hard, just a variation of garbage_collection.rs.

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -108,17 +108,21 @@ deck-config-bury-review-tooltip = Whether other `review` cards of the same note 
 deck-config-bury-interday-learning-tooltip =
     Whether other `learning` cards of the same note with intervals > 1 day
     will be delayed until the next day.
-deck-config-bury-siblings = Bury siblings
-deck-config-none = None
-deck-config-only-new = Only new
-deck-config-only-new-review = Only new and review
-deck-config-new-review-interday = New, review, and interday learning
-deck-config-bury-tooltip =
-    What kind of siblings will be delayed until the next day. New, review and learning
-    cards with intervals > 1 day can be gradually included.
-    
-    Siblings are other cards of the same note (eg reverse cards, adjacent cloze deletions).
 
+deck-config-bury-siblings = Bury siblings
+deck-config-do-not-bury = Do not bury siblings
+deck-config-bury-if-new = Bury if new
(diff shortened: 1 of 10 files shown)


dae:
Sure, that sounds useful, thanks.

One minor issue with the change to a dropdown is that it forces v2 scheduler users to enable new card burying when enabling reviews, which the v2 scheduler doesn't technically require. But since v2 won't be around for too much longer anyway, it's probably not a big deal.

RumovZ:
Thanks for the improvements! I missed the scheduler version check for the interday learning burying.

dae:
Any preference on merge timing? It would be nice to clean up the now-unused translations, but that can be done in a follow-up PR if you'd like to handle that as part of your new binary.

RumovZ:
Just pushed #2364. I guess if you incorporate that script into your translation update workflow (not sure how it works), this PR is also good to go, right?
I would have run it myself, but I don't have access to the Swift codebase.

dae:
Since we're not deleting the strings in this case but just marking them as deprecated, it's not so important if AnkiMobile is using them or not - I suspect such cases will be rare if they exist at all, and they can be resolved at garbage collection time.

dae:
(if you're curious, the update workflow is run by ftl:sync in rust.rs)

dae:
Thanks for this Rumo!

PR #2364, opened 2023-02-05
Add Rust bin to deprecate unused ftl entries

RumovZ:


dae:
I ran this with the following command:

```
cargo run --bin deprecate_ftl_entries ftl ftl/usage
```

It looks like we have some false-positives at the moment. Eg qt-misc-layout-* and media-check-template-references-field-header. I think the original intention with the json files was to declare strings used by AnkiMobile, and by older Anki versions, so that we don't cut strings while they're still in use. We'll need to supplement that with whatever's used currently, since those json files will be from a previous point in time.

dae:
(and in fact, since we're not deleting old entries, it may be that only the current state is relevant)

RumovZ:
I was assuming you would be running `write_ftl_json` before that. Don't you also do that with `garbage_collect_ftl_entries`, because it dosn't take into account the current repositoy state, either?

Are you saying `deprecate_ftl_entries` shouldn't read any jsons, but parse the source roots directly? Should it be run automatically whenever `ftl/sync.rs` runs then?

dae:
I could do that if that's easier for you - it's been a fair while since I've visited this code, and its usage is not fresh in my memory :-) I think my original thinking is that garbage collection would only happen every ~6 months, so that we don't delete translations as soon as they're removed in git head, in case a stable release based on earlier commit will be needed. But deprecating unused strings as part of ftl:sync does sound good.

RumovZ:
It's not a problem to collect the references along the way.
On a different note, I see `ftl` now also contains the submodules which we don't want to mess with. So I guess the scripts should now also have the ability to take multiple ftl roots? (Currently core and qt.)

dae:
Yep, that'd be good

RumovZ:
Haven't managed to add it to the build process, yet. Am assuming correctly that I must add something like the following to `build/configure/src/rust.rs:update_translations`?
```rs
    build.add(
        "rslib/i18n_helpers",
        CargoRun {
            binary_name: "deprecate_ftl_entries",
            cargo_args: "-p anki_i18n_helpers",
            bin_args: "ftl/core ftl/qt -- pylib qt rslib ts",
            deps: inputs!["ftl/core", "ftl/qt", "pylib", "qt", "rslib", "ts"],
        },
    )?;
```
I probably got (at least) the deps wrong, I'm getting `ninja: error: out\build.ninja:131: multiple rules generate out/phony`.

Also noticed a bug I think: In garbage collection we infer the used entries by parsing the provided jsons and grepping nested messages and terms from ftl files in the repository. I guess we should either included nested entries in the JSONs or take into account all entries of the repository.
Let me know if you want me to tackle that. We can also wait until you need to use it the next time. Then the ftl shell scripts will also need to be migrated to the new build system.

dae:
The phony issue was a bug in CargoRun - I've pushed a fix. In your snippet above, instead of "rslib/i18n_hepers", the name "ftl:deprecate" or similar would be easier to remember. Then I could ./ninja ftl:deprecate ftl:sync.

By nested messages, do you mean strings that reference other strings? Hmm. I wonder if that would be an issue in practice?

Eg if a source file references string A, and string A references strings B, then only A would be listed in the json file. But when we do a GC run, presumably the code would mark A as required as it's in the json file, and B as required since it referenced by A? If it's currently resulting in B being marked unused, that is something we'd want to fix.

dae:
Having played around with this for a while, it will be nice to avoid deprecating strings used by the mobile clients for now. Ideally we'd migrate them from the desktop repo into an AnkiMobile or AnkiDroid-specific repo, but AnkiDroid is not currently set up for translating Fluent files, and moving strings from one repo to another is a bit of a pain. The workflow involves using duplicate-string.py to create a new key with a different name, then updating the source files(s) to reference the new key. Easier to punt that to garbage collection time for now :-)

dae:
This seems to work well in any case :-) :rocket:

dae:
(I'm happy to merge this; is that ok from your end?)

RumovZ:
Ok from my end now. 🙂

dae:
Thanks Rumo, this is a nice improvement :-)

PR #2366, opened 2023-02-06
Rework & unify webview identification and title setting

glutanimate:
### Rationale

Reading through #2340  again reminded me that the way individual webviews identify themselves across Anki is fairly inconsistent, which is a source of confusion for add-on developers and potentially annoying future coupling for Anki contributors. Namely the problems are:

- When interacting with webviews with legacy page content, a few (but frequently used) add-ons currently use the webview title to identify webviews to interact with. I assume that page titles were never meant to become a canonical part of the add-on API, but it was a good way for add-on authors to reliably identify webviews [in some code paths](https://github.com/glutanimate/popup-dictionary/blob/main/src/popup_dictionary/reviewer.py#L72-L75), and so add-on dev hijinks ensued. This is not immediately obvious to contributors and could be the cause of breakages in the future.
- When interacting with webviews with Svelte/TS page content, the current recommended practice is to check the URL path to identify the webview in the context of hooks like `webview_did_inject_style_into_page`. This binds Anki's Svelte/TS output to a set file naming and directory structure, which could be a problem in the future. In particular, as we don't explicitly mention that this is the case, contributors could easily miss this fact in the future and introduce add-on breakages when e.g. refactoring a single page into multiple pages under a common subpath.

In addition there are a number of smaller inconsistencies which, while not as crucial, do potentially slightly detract from the developer experience:

- Legacy web views do have a title in the Chromium debugging view, but Svelte/TS pages do not
- Some webviews with Svelte/TS content do set a webview title, but this does not actually work (e.g. deckOptions and changeNoteType)
- With some pages setting a title and others not, the logic for identifying web views via the Chromium DevTools protocol, which could be useful for custom treatment as discussed in #2151, also becomes slightly more complicated

### Solution

To address these problems, this PR implements the following changes:

- It introduces a new enum registry for all web views called `AnkiWebViewOrigin`. This serves both as a property that add-ons can use to identify a web view or that we can equip hooks as in #2340 with, and as a unified place to define the page titles for web views.
- It extends `load_ts_page` with the ability to set the page title for Svelte/TS pages

**Effects on the debugger screen**

| Before | After |
| ------ | ----|
|    Screenshot_20230206_003833         |    Screenshot_20230206_004133       |

### Additional thoughts / questions

- How do you feel about the `AnkiWebViewOrigin` naming? FWIW, I'm not hard-set on it and actually started with `AnkiWebViewType`, but `origin` made it easier to avoid using `type` as a parameter and also felt like a slightly better fit
- If this is merged we can use `AnkiWebViewOrigin` to replace the more addon-specific `WebviewDidInitContext` in #2340
- Add-on compat relevant changes: This PR extends `AnkiWebView.__init__` with an optional argument, renames `add_dynamic_css_and_classes_then_show`, and also renames a couple of web view titles to achieve consistent naming. I checked and wasn't able to find any potential add-on breakages for any of these changes.


dae:
This seems like a nice improvement, and I like that WebviewDidInitContext is being addressed in a more general way.

I'm not so fond of 'origin', as it makes me think of https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin. How about 'context'?

evandrocoan (approved):

glutanimate:
> I'm not so fond of 'origin', as it makes me think of https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin. How about 'context'?

Fair point. I do like `context` quite a bit, but I worry that it could lead to confusion with our use of `context` in hooks like `webview_did_receive_js_message` (where it's not just an enum describing the context, but – for the most part – the actual parent object of the webview).

We could also go with `AnkiWebViewOriginType`, but that's quite long.

I switched to `AnkiWebViewKind` for now, WDYT?

evandrocoan:
As I can see, `context` is already of `Any` type on the `webview_did_receive_js_message` signature and of type `object` on `webview_will_set_content`. So, I would prefer to use `context` instead of `origin` or `kind`.

dae:
I have a slight preference for context, but kind works for me too. Would you prefer to keep it or change it @glutanimate?

glutanimate:
I would vote to keep `AnkiWebViewKind` or go with a different name like `AnkiWebViewType`. `context` would be more fitting, I agree, but I truly do think it would end up confusing add-on authors. While it's true that we type `context` as `Any` in other hooks, the implicit expectation is that it holds the parent object of the webview. So I could see new add-on authors conflate the different types of webview context, expecting to be able to use the same logic across `webview_will_set_content`, `webview_did_receive_js_message`, `webview_did_init`, and checks against `webview.context()`.

evandrocoan:
Well, kind it goes then 😄

PR #2371, opened 2023-02-11
Implement TTS using windows crate

RumovZ:
I started with directly porting the Python code, which lead to a huge delay. However, this was trivial to improve as that code only wrote 1 byte at a time. Reading in, then writing out the whole stream had the best performance overall:

| Implementation | Time |
| --- | --- |
| Python API (main) | 4.3 s |
| Direct Rust port | 29.4 s |
| This PR | 1.5 s |

These benchmarks are for synthesizing 1000 words of Lorem Ipsum. Afterwards I noticed that the chosen voice also has a significant impact: Hedda was twice as fast as Stefan (German voices). The benchmarks should still hold up since I used the latter voice for all of them.

Open questions / todos:
- [x] Reading in the whole stream requires allocating a lot of memory, 15 MiB in the above case. Should we cap this for the case users put excessive amounts of text in TTS fields? We could also cap the text that gets synthesized. E.g. of 2000 words, only the first 1000 would get read out.
- [x] Error handling needs to be improved. Example: Japanese voices are listed as available on my system, but TTS crashes when I try to use them. Maybe because I uninstalled the language pack some time ago. The error is handled by the frontend, but the backend message is not very informative. Would be good to manually add some context instead of just `into`ing to an AnkiError.
- [x] The Python code checks for the Windows 10 October 2018 update. I couldn't confirm or disprove that this is the minimum required patch. Do you remember what the source is?
- [x] I put the tts in the card_rendering module because I didn't want to add a new backend service just for two methods. Not sure if it's a good fit.
- [x] Passed text is currently pasted into the [SSML](https://learn.microsoft.com/en-us/azure/cognitive-services/speech-service/speech-synthesis-markup-voice) document verbatim. We should probably escape it, but wouldn't it be cool if the user could write their own SSML? Maybe we can add a flag to the new TTS syntax to enable raw input.

### DIFF
--- a/qt/aqt/tts.py
+++ b/qt/aqt/tts.py
@@ -583,7 +583,7 @@ if is_win:
         @staticmethod
         def _get_available_voices(validate: bool) -> list[TTSVoice]:
             assert aqt.mw
-            voices = aqt.mw.backend.all_tts_voices(val=validate)
+            voices = aqt.mw.backend.all_tts_voices(validate=validate)
             return list(map(WindowsRTVoice.from_backend_voice, voices))
 
         def _play(self, tag: AVTag) -> None:
--- a/qt/aqt/tts.py
+++ b/qt/aqt/tts.py
@@ -28,11 +28,9 @@ expose the name of the engine, which would mean the user could write
 
 from __future__ import annotations
 
-import asyncio
 import os
 import re
 import subprocess
-import threading
 from concurrent.futures import Future
 from dataclasses import dataclass
 from operator import attrgetter
@@ -40,7 +38,9 @@ from typing import Any, cast
 
 import anki
 import anki.template
+import aqt
 from anki import hooks
+from anki.collection import TtsVoice as BackendVoice
 from anki.sound import AVTag, TTSTag
 from anki.utils import checksum, is_win, tmpdir
 from aqt import gui_hooks
@@ -545,35 +545,26 @@ if is_win:
 
     @dataclass
     class WindowsRTVoice(TTSVoice):
-        id: Any
+        id: str
 
-    class WindowsRTTTSFilePlayer(TTSProcessPlayer):
-        voice_list: list[Any] = []
-        tmppath = os.path.join(tmpdir(), "tts.wav")
-
-        def import_voices(self) -> None:
-            import winrt.windows.media.speechsynthesis as speechsynthesis  # type: ignore
-
-            try:
-                self.voice_list = speechsynthesis.SpeechSynthesizer.get_all_voices()  # type: ignore
-            except Exception as e:
-                print("winrt tts voices unavailable:", e)
-                self.voice_list = []
-
-        def get_available_voices(self) -> list[TTSVoice]:
-            t = threading.Thread(target=self.import_voices)
-            t.start()
-            t.join()
-            return list(map(self._voice_to_object, self.voice_list))
-
-        def _voice_to_object(self, voice: Any) -> TTSVoice:
-            return WindowsRTVoice(
+        @classmethod
+        def from_backend_voice(cls, voice: BackendVoice) -> WindowsRTVoice:
+            return cls(
                 id=voice.id,
-                name=voice.display_name.replace(" ", "_"),
+                name=voice.name.replace(" ", "_"),
                 lang=voice.language.replace("-", "_"),
             )
 
+    class WindowsRTTTSFilePlayer(TTSProcessPlayer):
+        tmppath = os.path.join(tmpdir(), "tts.wav")
+
+        def get_available_voices(self) -> list[TTSVoice]:
+            assert aqt.mw
+            voices = aqt.mw.backend.all_tts_voices()
+            return list(map(WindowsRTVoice.from_backend_voice, voices))
+
         def _play(self, tag: AVTag) -> None:
+            assert aqt.mw
             assert isinstance(tag, TTSTag)
             match = self.voice_for_tag(tag)
             assert match
@@ -582,12 +573,16 @@ if is_win:
             self._taskman.run_on_main(
                 lambda: gui_hooks.av_player_did_begin_playing(self, tag)
             )
-            asyncio.run(self.speakText(tag, voice.id))
+            aqt.mw.backend.write_tts_stream(
+                path=self.tmppath,
+                voice_id=voice.id,
+                speed=tag.speed,
+                text=tag.field_text,
+            )
 
         def _on_done(self, ret: Future, cb: OnDoneCallback) -> None:
-            try:
-                ret.result()
-            except RuntimeError:
+            if exception := ret.exception():
+                print(str(exception))
                 tooltip(tr.errors_windows_tts_runtime_error())
                 return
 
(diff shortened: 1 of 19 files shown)


dae:
Thanks Rumo, this looks great!

Associated bug report for reference: https://forums.ankiweb.net/t/slow-tts-and-duplicated-voices/25157. The user was experiencing big slowdowns even with a "hello world" when we were using winsdk, so even if the Rust performance is on par with the current winrt package, it will hopefully be an improvement over winsdk.

- Some sort of limit probably does make sense, though we should probably make it character based to be easier to implement. Various other tasks Anki does can take much more than 15MB with larger collections, so we probably don't need to be too stingy here - the equivalent of 100MB of memory for example is probably fine.
- Context on errors sounds good. Is it failing when creating the audio file, or at an earlier step? I'm wondering whether pre-validating the voices is something we could do quickly, or whether it could lead to startup delays.
- The restriction on Windows version was originally introduced in https://github.com/ankitects/anki/pull/855/commits/9a0b112a0fca8b2924885069297b862167e210bb. Most of the API looks like it works on any Windows version, but speaking rate in the Python implementation seems to require a later version: https://learn.microsoft.com/en-us/uwp/api/windows.media.speechsynthesis.speechsynthesizeroptions?view=winrt-22621. I did not check every API call, so there might be one that I missed. @rye761 Do you happen to remember what originally prompted that change?
- card_rendering sounds good
- Re SSML: was there a technical reason for using that instead of raw strings (I see it supports specifying a speaking rate, which may circumvent the added-in-a-later-API rate option), or was the change just to be able to support SSML tags in the future? My main concern here is ensuring things work well across platforms. If we expose the ability to add SSML and then the user switches to a Mac/Linux/mobile device and suddenly the SSML tags they or a shared deck author added are being read out instead of parsed, then that's going to make for unhappy users and more support requests. So I think we'd probably want to be escaping the text, at least until we ensure that other platforms degrade gracefully (and even then, it could cause confusion if we just strip it and don't parse it on other platforms, since the user will probably pick up on differences in the way the text is read across the platforms)

RumovZ:
> Some sort of limit probably does make sense, though we should probably make it character based to be easier to implement. 

Sorry, that wasn't clear from my side. Proposal 1 was to limit the buffer size.

dae:
(and sorry, I wasn't clear either - I meant a byte rather than a character limit)

RumovZ:
(Hit 'Comment' here instead of in the other PR. 😅)

> Some sort of limit probably does make sense, though we should probably make it character based to be easier to implement.

Sorry, that wasn't clear from my side. Proposal 1 was to limit the buffer size, but don't enforce a hard cap on text length. So instead of blindly allocating a buffer for the entire stream, we would allocate a buffer of at most 100MB, then read and write in a loop until the stream is written to disk.

> Is it failing when creating the audio file, or at an earlier step? I'm wondering whether pre-validating the voices is something we could do quickly, or whether it could lead to startup delays.

I'll investigate where the error occurs exactly. Validating the voices in advance sound like it's worth a try.
I'll also check what's with those duplicates cqg reported and that I'm also seeing. Maybe we should be using one voice with the same name over the other.

> was there a technical reason for using that instead of raw strings

No, I just missed that it was possible to set a rate without SSML. 😃 I'll use the API call then. I can see that eventual SSML support would need more thought.

rye761:
>     @rye761 Do you happen to remember what originally prompted that change?

Unfortunately I don't remember where I read that at the time. I recall finding some information saying the RT voices were introduced in that version of Windows, but I can't find anything to confirm that now.

RumovZ:
I only found resources saying that the current TTS is supported since Win 7, but as there don't seem to have been any complaints from users, I have no objections to just sticking with the current logic.

I added some context to the more interesting API calls.
Great idea about the validation! The extra delay should be imperceptible. (The error occurs at the async synthesize step.)
Regarding the duplicates, that's just from the other implementation, WindowsTTSPlayer.

rslib/src/error/windows.rs
```diff
@@ -0,0 +1,32 @@
...
+pub struct WindowsError {
+    details: WindowsErrorDetails,
+    source: windows::core::Error,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum WindowsErrorDetails {
+    SettingVoice(windows::Media::SpeechSynthesis::VoiceInformation),
+    SettingRate(f32),
+    Synthesizing(String),
```
dae, line 19: To avoid giant errors, we probably want to truncate the included text to a maximum of something like 40 characters, or just omit it (without using SSML, I would be surprised if the content of the text had any influence over whether the request failed or not).

dae:
I got confused above, and assumed we were talking about 15MB of input text, instead of 15MB of output audio. The latter could easily grow into hundreds of megabytes, so reading it in chunks instead of a single operation does sound wise.

Regarding validation, I wonder if we can always rely on it being quick. If attempting to synthesize an empty string requires loading the TTS engine of each voice into memory, and the user is on a HDD and is opening Anki for the first time since reboot, and has a bunch of third-party engines installed, I wonder whether performance would still be acceptable, and whether there might be a memory usage hit? Not sure what the best course of action here is - it may not make much of a difference to the typical user, so I may be overthinking things. The alternative would be to revert back to the old approach, and perhaps add a suggestion about installing language packs when an error is shown. WDYT?

RumovZ:
Validation does increase the time to retrieve available voices from 0.00 s to 0.14 s on my machine (3 working, 4 broken voices). Making it async brings it down to 0.11 s and will scale better with more voices of course.

If you want to drop validation, where do we display that suggestion? Currently when an error occurs, a tooltip is shown that already has a lot of text in it. Maybe we could add it to the `{{tts-voices:}}` filter instead? The tooltip already mentions choosing a different voice, and if a voice unexpectedly doesn't work, the user user would presumably check out the voice list.
We could even use validation there to display next to each voice if it can actually be used.

dae:
Doing it in the filter is a good idea - that's something users are unlikely to have permanently on their cards, so the delay would only be when they're setting things up. Maybe we could append something like (unavailable) to the broken voices, and add text to the effect of "One or more voices were unavailable. Installing a Windows language pack may help"?

RumovZ:
I think this is ready now.

qt/aqt/tts.py
```diff
@@ -205,7 +213,9 @@ def _parse_voice_line(self, line: str) -> TTSVoice | None:
 
         original_name = m.group(1).strip()
         tidy_name = f"Apple_{original_name.replace(' ', '_')}"
-        return MacVoice(name=tidy_name, original_name=original_name, lang=m.group(2))
+        return MacVoice(
+            name=tidy_name, original_name=original_name, lang=m.group(2), available=None
```
RumovZ, line 217: In case you're wondering why I didn't make `available` a default attribute: The non-default attributes of the child classes would appear after it in the `__init__` signature, making it invalid. It would work with `@dataclass(kw_only=True)`, but that would require Python 3.10.

dae:
Unfortunately this isn't currently working for me - I get an The operation attempted to access data outside the valid range (Exception from HRESULT: 0x8000000B) message when I try to play any of the sounds. They don't show up as unavailable in the tts list. When I revert the maximum buffer size change things work, so it seems to be an issue there. I tested with a field that just contained "hello world".

dae:
I've pushed an alternate version that seems to work for me; does that look ok to you?

qt/aqt/tts.py
dae: Interesting, I was not aware subclassed dataclasses behaved that way.

RumovZ:
Oh, I see my mistake now. Unfortunately I only tested with large strings.

I wrote it this way to make the allocation as small as possible, because presumably synthesizing a single word is a common case and for that you don't need a whole MB. But then let's keep the buffer static and use an array.

dae:
One thing to be mindful of here is that the default stack size is only a few megabytes (varies by platform), so it can be a bit dangerous to allocate large arrays. I've reduced the array to 128kb as a precaution - while I didn't benchmark it, I suspect the performance difference is fairly small. Feel free to follow up with tweaks if you feel a different approach is warranted - pushing this through now as I need to get an important bugfix out.

dae:
(Thanks as always for your great work :-)

PR #2372, opened 2023-02-12
Deck scoped dupe check

RumovZ:
This allows to restrict the duplicates check when importing a CSV file to the deck a note is imported into.

By the way, as we've recently talked about configuring the dupe resolution setting from inside a CSV file, it's already part of CsvMetadata. Consquently I've also added this new setting to CsvMetadata. This makes it easy us to support CSV headers for these settings, too. Should I do that?

### DIFF
--- a/rslib/src/import_export/text/csv/metadata.rs
+++ b/rslib/src/import_export/text/csv/metadata.rs
@@ -161,13 +161,17 @@ impl Collection {
                     metadata.guid_column = n;
                 }
             }
-            "limit dupe check to deck" => {
-                if let Ok(limit) = value.to_lowercase().parse() {
-                    metadata.limit_dupe_check_to_deck = limit;
+            "match scope" => match value {
+                "notetype" => {
+                    metadata.limit_dupe_check_to_deck = false;
                 }
-            }
-            "dupe resolution" => {
-                if let Some(resolution) = DupeResolution::from_str_name(&value.to_uppercase()) {
+                "notetype + deck" => {
+                    metadata.limit_dupe_check_to_deck = true;
+                }
+                _ => (),
+            },
+            "if matches" => {
+                if let Some(resolution) = DupeResolution::from_text(value) {
                     metadata.dupe_resolution = resolution as i32;
                 }
             }
@@ -261,6 +265,15 @@ impl CsvMetadata {
 impl DupeResolution {
     fn from_config(col: &Collection) -> Self {
         Self::from_i32(col.get_config_i32(I32ConfigKey::CsvDuplicateResolution)).unwrap_or_default()
+    }
+
+    fn from_text(text: &str) -> Option<Self> {
+        match text {
+            "update current" => Some(Self::Update),
+            "keep current" => Some(Self::Preserve),
+            "keep both" => Some(Self::Duplicate),
+            _ => None,
+        }
     }
 }
 
--- a/rslib/src/import_export/text/csv/metadata.rs
+++ b/rslib/src/import_export/text/csv/metadata.rs
@@ -161,6 +161,16 @@ impl Collection {
                     metadata.guid_column = n;
                 }
             }
+            "limit dupe check to deck" => {
+                if let Ok(limit) = value.to_lowercase().parse() {
+                    metadata.limit_dupe_check_to_deck = limit;
+                }
+            }
+            "dupe resolution" => {
+                if let Some(resolution) = DupeResolution::from_str_name(&value.to_uppercase()) {
+                    metadata.dupe_resolution = resolution as i32;
+                }
+            }
             _ => (),
         }
     }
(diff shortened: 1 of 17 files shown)


rslib/src/tests.rs
```diff
@@ -173,3 +173,37 @@ impl DeckAdder {
...
     }
 }
+
+#[derive(Debug, Clone)]
+pub(crate) struct NoteAdder {
+    note: Note,
+    deck: DeckId,
+}
+
+impl NoteAdder {
```
dae, line 183: Looks handy - I need to get into the habit of using this/DeckAdder.

rslib/src/import_export/text/import.rs
```diff
@@ -83,6 +89,36 @@ struct DeckIdsByNameOrId {
     default: Option<DeckId>,
 }
 
+/// Notes in the collection indexed by notetype, checksum and optionally deck.
+/// With deck, a note will be included in as many entries as its cards
+/// have different original decks.
+#[derive(Debug)]
+enum ExistingChecksums {
+    WithoutDeck(HashMap<(NotetypeId, u32), Vec<NoteId>>),
+    WithDeck(HashMap<(NotetypeId, u32, DeckId), Vec<NoteId>>),
```
dae, line 98: I'd originally assumed we'd just ignore any cards outside of the currently-selected deck, but of course that doesn't work for users who have per-row decks specified. This implementation seems like a better approach.

dae (commented):
I don't really have an opinion on whether we should expose that in the file - please do what you think is best :-)

RumovZ:
Ok, done. I've also aligned the pb names with the GUI labels to avoid confusion.

rslib/src/import_export/text/csv/metadata.rs
```diff
@@ -168,6 +161,16 @@ impl Collection {
                     metadata.guid_column = n;
                 }
             }
+            "limit dupe check to deck" => {
```
dae, line 164: This feels a bit verbose to me, but I'm struggling to come up with better alternatives. Maybe "duplicate scope: deck/collection"?

rslib/src/import_export/text/csv/metadata.rs
```diff
@@ -168,6 +161,16 @@ impl Collection {
...
+                if let Ok(limit) = value.to_lowercase().parse() {
+                    metadata.limit_dupe_check_to_deck = limit;
+                }
+            }
+            "dupe resolution" => {
```
dae, line 169: Brainstorming - how about "if duplicate: update current/keep current/keep both"?

rslib/src/import_export/text/csv/metadata.rs
RumovZ: No objections, but since this is hard-wired to a GUI option, I would keep the two very similar. Should I replace the GUI switch with a selector as well then?

rslib/src/import_export/text/csv/metadata.rs
RumovZ: Same here: Do we also change the ftl strings then?
By the way, the label says "Existing notes". An argument could be made to use that here as well.

rslib/src/import_export/text/csv/metadata.rs
dae: Thinking about it more, deck/collection is a bit misleading too. notetype/notetype+deck might better reflect what's going on, but it's wordy. I'm not sure what's best. Any other ideas? :-)

rslib/src/import_export/text/csv/metadata.rs
dae: I don't mind 'existing notes' either, and it would save time. I am conscious that we're deciding on a file format here though and won't be able to change it later, so maybe it's worth the extra effort of bikeshedding the wording/updating the strings.

One other possible alternative:

`If matches:` instead of `If duplicate`. Thoughts? :-)

rslib/src/import_export/text/csv/metadata.rs
RumovZ: I like `duplicate scope:notetype+deck` because it's close to the implementation and could easily support more options in the future, like `duplicate scope:deck` or `duplicate scope:all` (or `collection`).
So should we use this in the GUI, too?

rslib/src/import_export/text/csv/metadata.rs
RumovZ: Either is fine for, though we may want to align this with the naming above: If we go with `if matches`, I would also prefer `match scope`.

rslib/src/import_export/text/csv/metadata.rs
dae: Ok, sounds good. Maybe 'notetype + deck' to avoid triggering people with OCD? :-)

rslib/src/import_export/text/csv/metadata.rs
dae: Sounds good.

RumovZ:
I didn't change the exisitng fluent strings for the dupe resolution selector, yet. If you want I can do that, too. Otherwise this is good to go from my side.

dae:
No strong feelings either way re translations, and we can always add them later. Let's get this in. Thanks Rumo :-)

PR #2374, opened 2023-02-16
Add deck quick actions to deck browser

sachingooo:
Adds quick actions - icons near the gear for each deck - to the deck screen. Quick actions are meant to remove some of the extra clicking for a deck-specific action. 

Example of the quick actions, showing options, rebuild, and empty:

image

Quick actions can be defined as unconditional (present for every deck) or conditional (present for only some decks), and a hook `deck_browser_did_generate_quick_actions_for_deck` has been added for addons to customize new quick actions. 

The quick actions can also be collapsed using an icon in the options column header in the deck browser and are _collapsed by default_, so users will not experience a change by default with this addition. 

The available quick actions are customizable in the user preferences, and preferences are global (saved across profiles):

image

Functions to modify and check these preferences have also been added. 

A minor UI change was made: changing the gear icon for the options menu into a hamburger menu, and reserving the gear (which traditionally implies changes _settings_ for configuring options. This was done so that the gear icon can be more readily associated with configuration, and a quick action has been added for jumping directly to options. 

Quick actions available as part of this enhancement:
- Options
- Rename
- Stats
- Rebuild (conditional)
- Empty (conditional)

Future quick actions might include things like jumping directly to study (skipping the overview screen), browsing the deck, or deleting the deck. 

Modifications and comments are welcome!


sachingooo:
Closing - will update with fixed tests

PR #2377, opened 2023-02-18
Fix HTML editor hiding when selected as default

winterhazel:
Fixes a bug introduced in d2fa50dd9.

Since `plainTextDefaults` and `richTextsHidden` referenced the same array, toggling the visual editor on a field that uses the HTML editor by default would also change whether the HTML editor is the default editor for that field. 

This caused the badge that toggles the visual editor to turn into a badge that toggles the HTML editor when clicking it for the first time.


dae:
Thank you Fabricio!

PR #2378, opened 2023-02-18
Revert backwards incompatible changes to TTSVoice

RumovZ:
https://forums.ankiweb.net/t/2-1-58-release-candidate/27532/11


dae:
Thanks for the quick fix Rumo :-)

PR #2379, opened 2023-02-18
Add option to tag notes with missing media

RumovZ:
https://forums.ankiweb.net/t/tag-notes-that-have-missing-media-when-running-check-media/27317

### DIFF
--- a/ftl/core/media-check.ftl
+++ b/ftl/core/media-check.ftl
@@ -73,3 +73,7 @@ media-check-empty-trash = Empty Trash
 # button to move deleted files from the trash back into the media folder
 media-check-restore-trash = Restore Deleted
 media-check-check-media-action = Check Media
+# a tag for notes with missing media files (must not contain whitespace)
+media-check-missing-media-tag = missing-media
+# add a tag to notes with missing media
+media-check-add-tag = Add tag '{ media-check-missing-media-tag }'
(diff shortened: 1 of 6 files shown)


ftl/core/media-check.ftl
```diff
@@ -73,3 +73,7 @@ media-check-empty-trash = Empty Trash
 # button to move deleted files from the trash back into the media folder
 media-check-restore-trash = Restore Deleted
 media-check-check-media-action = Check Media
+# a tag for notes with missing media files (must not contain whitespace)
+media-check-missing-media-tag = missing-media
+# add a tag to notes with missing media
+media-check-add-tag = Add tag '{ media-check-missing-media-tag }'
```
dae, line 79:
```suggestion
media-check-add-tag = Tag Missing
```

dae: Your approach does it make it more obvious to the user which tag they should look for, but we may be displaying a number of different buttons at once, and the buttons can end up getting truncated already, so brevity is probably the better option here. About to build a new rc so I'll push this through without waiting for your feedback, but feel free to make adjustments/discuss if you disagree.

dae (commented):
Seems to work well, and nothing jumped out at me in the code. Thanks Rumo!

PR #2381, opened 2023-02-19
Add option to create subdeck from selected deck

t2b3:
In the deck browser, creating a subdeck for a specific deck with multiple deck levels involved is rather inconvient (1. create a new deck and 2. than drag and drop it to the specific deck).

This commmit adds the option to add a new subdeck directly to the specific deck.


sachingooo:
Just thinking through the proposed solution - you've doubled the number of places that a user can create a new deck. Why add this to the options menu, instead of as a checkbox (something like "Create subdeck of current deck") to the existing create deck dialog that prepends the current deck? 

It also looks like you need to add yourself to CONTRIBUTORS as part of this PR

t2b3:
Thanks for your feedback. I closed this PR for now as it needs further polishment.

PR #2383, opened 2023-02-19
Set deck to current deck when clicked on table row

t2b3:
When a user clicks on the table row in the deck list (neither click on the deck link nor on the options menu), then set the deck in that table row as the current deck.

This gives the user some quick feedback and enables the user to easily create subdecks (as it automatically fills the add deck dialog with the deck name of the selected deck) and to create notes belonging to the selected deck.

### DIFF
--- a/qt/aqt/deckbrowser.py
+++ b/qt/aqt/deckbrowser.py
@@ -118,6 +118,10 @@ class DeckBrowser:
             self._confirm_upgrade()
         elif cmd == "v2upgradeinfo":
             openLink("https://faqs.ankiweb.net/the-anki-2.1-scheduler.html")
+        elif cmd == "select":
+            set_current_deck(
+                parent=self.mw, deck_id=DeckId(int(arg))
+            ).run_in_background()
         return False
 
     def set_current_deck(self, deck_id: DeckId) -> None:
@@ -210,7 +214,14 @@ class DeckBrowser:
         else:
             klass = "deck"
 
-        buf = "<tr class='%s' id='%d'>" % (klass, node.deck_id)
+        buf = (
+            "<tr class='%s' id='%d' onclick='if(event.shiftKey) return pycmd(\"select:%d\")'>"
+            % (
+                klass,
+                node.deck_id,
+                node.deck_id,
+            )
+        )
         # deck link
         if node.children:
             collapse = (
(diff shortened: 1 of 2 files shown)


qt/aqt/deckbrowser.py
```diff
@@ -346,7 +354,9 @@ def _onShared(self) -> None:
         openLink(f"{aqt.appShared}decks/")
 
     def _on_create(self) -> None:
-        if op := add_deck_dialog(parent=self.mw):
+        if op := add_deck_dialog(
```
sachingooo, line 357: Is this a remnant from your other PR?

dae:
I'm on the fence with this one. On the positive side, it's a small change, and I do see how this is making the create-subdeck workflow easier. On the negative, I'm not sure this is the behaviour the typical user would expect - I worry that it would lead to bug reports where users complain that sometimes when they click on the deck the highlight changes but it doesn't open up the study screen. I wonder whether the typical user would prefer they could just click anywhere to start studying that deck.

Anyone else want to weigh in?

dae:
(One other possible option would be to use the hooks mentioned on https://github.com/ankitects/anki/pull/2375#issuecomment-1437191459 to add a button on the right to pop up a dialog to add a subdeck)

RumovZ:
I think an option to change the current deck without leaving the deck browser is a good idea, but assigning a different action to 'clicking a deck, but not quite' indeed feels a bit accidental.
What if instead the trigger is <kbd>Alt+Click</kbd> (or another modifier)?

qt/aqt/deckbrowser.py
t2b3: Yes, this is a facilitator for creating sub-decks.

t2b3:
> 

I think that this is a good idea. Having taken a quick look into Svelte, there does not seem to be an easy way to combine the `onclick` event/binding with modifier keys? It seems to be a little more complicated (i.e. fetching the modifier key pressed as extra event).

RumovZ:
No, sorry, that's not my area of expertise. I was only speaking from a user perspective there. 🙂

dae:
Doing it with a keyboard modifier would alleviate most of my concerns. The deck list doesn't use Svelte, so you'd need to check the modifier keys using standard Javascript. https://javascript.info/mouse-events-basics#modifiers-shift-alt-ctrl-and-meta

t2b3:
Thanks for your feedback! I adjusted the code likewise. Now, the deck is only selected in case of the Shift modifier pressed.

dae:
Thanks Tobias!

kleinerpirat:
@dae I noticed this change while merging my WIP main window rework into a new branch. In case I forget it when I publish the PR (hopefully in a few days), I'll document this here:

As the former deck browser and overview will be shown side to side on a single webview, selecting decks without starting the reviewer will be the new standard and won't require need any special treatment (such as a keyboard modifier). For those in a hurry, double-clicking a deck will start the review directly.

PR #2387, opened 2023-02-20
Fix deck options selector not updating the current entry

winterhazel:
Fixes a regression in 766d1fdb5.

The current entry wasn't being visually updated in the selector when adding/cloning/renaming a preset.


dae:
Thanks Fabricio.

PR #2388, opened 2023-02-20
Allow async callbacks in lifecycle.onMount/onDestroy

### LINKED ISSUE #2384: JS error if lifecycle.onMount doesn't return a callback
As described [here](https://github.com/kleinerpirat/anki-tooltips/issues/4), the webview may crash if add-on authors do not return a callback in their `lifecycle.onMount` function.

Also, our current code assumes the callback is synchronous. Most Svelte operations need to be done asynchronously, so hacks like async IIFE's inside a synchronous function are required. We should allow `Promise<Callback>` too.

kleinerpirat:
# Issue

Currently, API users cannot use the Svelte lifecycle hooks the following way:

```ts
require("anki/RichTextInput").onMount(async (api: RichTextInputAPI) => {
    const editable = await api.element;
    editable.addEventListener("focus", doThatThing);

    // cleanup callback
    return () => {
        editable.removeEventListener("focus", doThatThing);
    };
});
```

Because an async function always has a return value (at least `Promise<void>`) and our code in [lifecycle-hooks.ts](https://github.com/ankitects/anki/blob/ec15f1176c12c7290ccbc76553eed3c850becedc/ts/sveltelib/lifecycle-hooks.ts#L35-L37) only checks for the existence of a value, this would result in the following error when switching between notetypes in the browser:

```
Uncaught (in promise) TypeError: cleanup is not a function
    at browser_editor.js:64505:13
    at run (browser_editor.js:61762:12)
    at Array.forEach (<anonymous>)
    at run_all (browser_editor.js:61768:9)
    at destroy_component (browser_editor.js:63416:7)
    at Object.d (browser_editor.js:109016:9)
    at Object.d (browser_editor.js:65379:24)
    at destroy_component (browser_editor.js:63417:34)
    at Object.d (browser_editor.js:109065:9)
    at Object.d (browser_editor.js:89279:32)
```

A workaround might look like this:

```ts
// synchronous callback
require("anki/RichTextInput").onMount((api: RichTextInputAPI) => {
     // async IIFE to access the element
    const callbackPromise = (async () => {
        const editable = await api.element;
        return () => {
            editable.removeEventListener("focus", doThatThing);
        };
    })();
    return () => {
        // Another IIFE, let's go!
        (async () => {
            (await callbackPromise)();
        })();
    };
});
```

That's really bad.

We can't expect all add-on authors to recognize the problem and be able to fix it. Also it makes the add-on code very obfuscated.

# Fix
This PR makes the type of the callback `Callback | Promise<Callback>` and handles all of this asynchronously, no matter whether the original callback was sync or async. It also makes the return of a cleanup function optional, like it was originally intended.

---
Closes #2384.

### DIFF
--- a/ts/sveltelib/lifecycle-hooks.ts
+++ b/ts/sveltelib/lifecycle-hooks.ts
@@ -11,8 +11,8 @@ type ComponentAPIDestroy<T> = (api: T) => void;
 type SetLifecycleHooksAction<T> = (api: T) => void;
 
 export interface LifecycleHooks<T> {
-    onMount(callback: ComponentAPIMount<T>): Callback;
-    onDestroy(callback: ComponentAPIDestroy<T>): Callback;
+    onMount(callback: ComponentAPIMount<T>): Callback | Promise<Callback>;
+    onDestroy(callback: ComponentAPIDestroy<T>): Callback | Promise<Callback>;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


dae:
Thanks Matthias!

PR #2390, opened 2023-02-21
Fix type box losing focus when switching IMEs

hikaru-y:
https://forums.ankiweb.net/t/bug-field-type-loses-focus-when-leaving-the-window/27680


dae:
Thanks as always Hikaru :-)

Issue #2376, opened 2023-02-17
require("anki/shortcuts").registerShortcut doesn't disable QShortcuts (but probably should)


kleinerpirat:
Anki exports the `anki/shortcuts` package to set a user-defined keyboard shortcut in webviews. Currently such shortcuts do not prevent QShortcuts registered on the editor from firing, so keys like <kbd>Esc</kbd> or <kbd>Ctrl+Enter</kbd> ([or my default shortcut for Anki Tooltips](https://github.com/kleinerpirat/anki-tooltips/issues/1#issuecomment-1434212561)) cannot be used without unwanted side effects like closing the window, adding a new note etc.

## Proposal
We could send a
```ts
bridgeCommand(`registerShortcut:${keyCombinationString}`);
```
on `registerShortcut` from `shortcuts.ts` to `AnkiWebView` and `setEnabled(False)` on any conflicting shortcut of the parent widget.

(This is an over-simplification. In reality it would be a bit more involved, as we're currently assigning shortcuts to individual widgets, not the QMainWindow/QDialog.)

Add-ons already have the ability to override shortcuts in Python, so it seems reasonable to allow it via JS too.

dae:
If an add-on author adds a Qt shortcut that conflicts with an existing one, both shortcuts fail to work, so it's up to the add-on author to disable the existing one first. So isn't the situation the same for JS? Add-on authors can disable the existing Qt shortcut if required.

kleinerpirat:
> So isn't the situation the same for JS?

Not really. If an add-on sets a shortcut via the JS API, the Qt shortcut continues to work.

If an add-on author allows users to set a custom shortcut, he'd have to know all shortcuts and which widget of each window they're set to and then provide a complex method to handle all cases. It's not like all shortcuts are set to the QMainWindow or QDialog: a lot of them are set on subwidgets.

I suggest cleaning house and creating a standard way to register shortcuts. E.g. subclass `QMainWindow` to `AnkiWindow` and `QDialog` to `AnkiDialog`, then give those subclasses a `ShortcutHandler` with `register_shortcut` and `unregister_shortcut` as well as a `set_shortcut_enabled` method for temporary suppression. Then move all subwidget shortcuts to the parent window.

It would also make the code cleaner:

```diff
-        self.compat_add_shorcut = QShortcut(QKeySequence("Ctrl+Enter"), self)
-        qconnect(self.compat_add_shorcut.activated, self.addButton.click)
[...]

PR #2391, opened 2023-02-22
Make registerShortcut from anki/shortcuts package override conflicting shortcuts

kleinerpirat:
Sparked by:
- https://github.com/kleinerpirat/anki-tooltips/issues/1

We discussed the Qt side in #2376 and came to the conclusion it's not worth the effort. But doing it on the JS side might be. This PR allows the following:

```ts
registerShortcut(
    (event: KeyboardEvent) => {
        // handle event
    },
    "Control+T",
    { event: "keydown", override: true},
);
```
As a result, all JS shortcuts (including emacs-style sequences) that conflict with the given keyCombination are disabled.

The example above would disable the LaTeX shortcuts:
 - Control+T, T
 - Control+T, E
 - Control+T, M

---
While it is (almost) reasonable to let add-on authors disable QShortcuts themselves, it is currently impossible to do the same in JS - how would they access the `remove` callback of native Svelte components that do not expose it?


kleinerpirat:
`./ninja format` and `./ninja check` seem to have different opinions again.

kleinerpirat:
An alternative to the `override` parameter would be to export a different function to the API than what our native components are using. Then add-on shortcuts would take precedence over native ones and existing add-ons would not have to be updated.

hikaru-y:
At this time, there appear to be a few issues with this.

1. It is not possible to register multiple listeners to a single shortcut, which is currently possible.
    - An add-on may want to register multiple listeners with different event targets or different `AddEventListenerOptions` for the same shortcut.

1. If a `<Shortcut>` component is mounted after an add-on executes `registerShortcut()`, the shortcut key of that component cannot be overridden.
    1. Suppose an addon wants to override <kbd>Ctrl + Shift + C</kbd>.
    1. Open an editor with a non-Cloze note type, then change the note type to a Cloze one.
    1. Pressing that shortcut key will execute the original callback as well as the add-on's callback.

kleinerpirat:
Thanks for testing this, Hikaru!

We could fix the first issue by simply not storing add-on shortcuts in the `registeredShortcuts` array. We should probably rename it to `nativeShortcuts` then.

The second issue is more involved.. But I feel like that could be solved on the add-on side. I'm at work the next 10 hours, so I won't be able to fix this today @dae.

kleinerpirat:
@hikaru-y do you think it's reasonable to let add-on authors manually handle issue 2? I guess the other option is to store calls to `require("anki/shortcuts").registerShortcut` and reiterate the disabling process when the native `registerShortcut` is called later.

dae:
(I'm tied up with AnkiWeb work at the moment, so I likely won't be able to get to the pending PRs until early next week I'm afraid)

dae:
We've been focusing on *how can we*, but I'd like to take a step back and think about the *should we* first. From what I can tell, the example use case here is turning off the LaTeX shortcuts so that the add-on can use Ctrl+t instead. Do we really want to be encouraging add-ons to replace the default shortcuts with unrelated actions? This is potentially surprising/annoying for users, as it means they can no longer access built-in functionality easily when they install the add-on. Wouldn't it be better to pick an unused shortcut combination instead?

dae:
(as an aside, I tried out your add-on for the first time today, and it's quite sleek)

kleinerpirat:
If it were a fixed shortcut, I would simply fix the issue from the add-on side. But a recent update allows users to set their own shortcut for the tooltips.

So any time an add-on allows to set custom shortcuts, there will be trouble with conflicts. I wouldn't exactly call the PR "encouragement" to override native shortcuts, but I get why you're nervous about it.

Alternatively, we could provide a list of editor shortcuts for add-on authors to compare user input against and create an alert in case a user sets a conflicting one.

It's probably best to get some feedback on the issue from other devs.

dae:
Ah, if the replacement is because the user deliberately chose a shortcut and not because it was the default, that makes more sense. One other option would be to tweak this code so that instead of disabling the existing shortcuts if found, it throws an error that the shortcut is already in use. The add-on could catch it, and report a message like "Ctrl+T is already used; please pick another shortcut". No strong feelings here. WDYT?

.eslintrc.js
```diff
@@ -33,6 +33,8 @@ module.exports = {
             ],
             rules: {
                 "@typescript-eslint/no-non-null-assertion": "off",
+                "simple-import-sort/imports": "off",
+                "simple-import-sort/exports": "off",
```
dae, line 37: This should not be necessary with the steps mentioned here: https://github.com/ankitects/anki/pull/2395#issuecomment-1445777657

kleinerpirat: Thanks!

kleinerpirat:
Throwing an error is a good idea! However, if a user doesn't care about a particular native shortcut, they may still want to override it. I'll revisit this later.

dae:

PR #2392, opened 2023-02-22
Button hover effect: swap impressed shadow for border color

kleinerpirat:
Like @tatsumoto-ren once mentioned, the current button hover effect (introduced with #2267) is too drastic and makes buttons seem clicked when they're only hovered. This changes the visual indication to a border-color change. Also makes the toolbar feel more snappy again.

## Before
image

## After
image


dae:
Looks good, and I liked the improved snappiness. Thanks!

PR #2393, opened 2023-02-22
Fix open editors getting carried over to different notetypes

winterhazel:
This fixes the issue mentioned in #2241. Field states persistence should be working as intended now.

Before, `Collapsible`s would only react to the changes made in `setPlainTexts` when the user clicked to show/collapse the content of any `Collapsible`.
___
Other changes in this PR:
- Fixed an issue that made the editor take a noticeably longer time to load notetypes with many fields defaulted to the HTML editor when reduced motion was enabled. The loading time of a notetype with 90 fields was reduced from about 18 seconds to 1.7 second in my machine.


kleinerpirat:
Thanks for all these fixes, Fabricio! I knew I had to go over my work one day and polish it, but you beat me to it and it's quite relieving.

Heads up: I'm currently reducing the complexity of `NoteEditor.svelte` to make it more stable, so if you were planning to work on that area I'd hold off until I got my PR out.

winterhazel:
Alright, thanks for the heads up! I was actually planning on making further changes to this this file, but I'll hold off for now.

kleinerpirat:
I just finished the editor work I mentionied above. You can build your next PR on #2395.

winterhazel:
I just realized that I broke the automatic focus on the first field when the editor gets opened, so I'm converting this to a draft.

dae:
Thanks as always for the fixes Fabricio - that's quite a speed improvement!

Issue #1085, opened 2021-03-22
consider extending sort field to all fields and/or changing HTML stripping


dae:
Currently the sort field stores the contents of a single field with HTML stripped. This enables faster sorting, and allows the user to find text that has embedded HTML in it, if that text occurs in the sort field.

There are some downsides to the current approach:

- if the user tries to search for text with embedded HTML that is not in the sort field, it is not found
- substring matches require a full table scan
- the current HTML stripping replaces tags with an empty string, so `foo<br>bar` becomes `foobar`. Because the lack of spaces is hard to read, the browse screen currently loads the full field content and uses a different HTML stripping routine that replaces the tags with a space, which is more expensive.

If instead of inserting just a single field into the sort field, we inserted sort_field + all_other_fields, using the standard field separator, it would solve the searching issue, at the cost of extra disk space.

A more comprehensive alternative would be to use sqlite's FTS, which is something I haven't really looked into yet. I gather it does quite well with space-separated text, but it may not be as useful for languages like Japanese? I'm also a bit concerned by the previous security issues it had, which could be triggered with malformed databases.

dae:
Tagged as a reminder to at least give it some consideration

PR #2394, opened 2023-02-22
Add option to exclude fields from search

abdnh:
This adds an option in the Fields dialog to exclude fields from unqualified searches.

image

This is certainly not the fastest implementation, but I guess we can only go so far before having to modify the schema. Do you think the Fluent string should say something like "Exclude from search (slow)"?

*Disclosure: Commissioned by @AnKingMed*


RumovZ:
Sorry if that's hijacking this thread, but is a schema change out of the question? Not only this feature, but also existing searches could perform better if there was a separate fields table where fields weren't joined together.
Of course, I'm talking long-term here.

Slightly related:  #1085

abdnh:
> but is a schema change out of the question?

Certainly not out of question. I just went for the solution I thought is the simplest.

I'd like to look into tackling this either in this PR or as a follow-up after hearing Damien's feedback. I wonder though what will be the storage cost of a separate table with both HTML/stripped contents, and the `sfld` and `flds` columns (which I guess we can't just get rid of immediately for the sake of backward compatibility), but maybe [external virtual tables](https://www.sqlite.org/fts5.html#external_content_tables) can help here.

RumovZ:
> > but is a schema change out of the question?
> 
> Certainly not out of question. I just went for the solution I thought is the simplest.

Sorry, I meant to mention @dae here. 🙂

I think the additional storage cost would be negligible because you'd just move the data from one column to another (like from `notes.flds` to `fields.content`). Granted, you'd also need `note` and `ord` columns for the mapping. As far as I know, we could drop `notes.flds` then, because backward compatibility is what downgrading is for. Restoring `notes.flds` from a fields table and vice versa would be easy.
The downside, apart from the schema change itself, is that it would require a major refactoring. However, all parts I can think of would be better for it.

Allowing more than one sort field is a different matter. But if we made that configurable per field and defaulted to off, there would be no additonal storage cost unless you opt in.
I linked that issue because it would make sense to do that with the same schema change and it would probably be easier to implement with a new fields table.

---

Now, to actually say something about this PR, have you considered reusing the existing field search logic? I also would have assumed that excluding a field applies to field searches as well.

abdnh:
> As far as I know, we could drop notes.flds then, because backward compatibility is what downgrading is for. 

I was mainly concerned about add-ons that directly access sfld and flds. I found around 50 grepping [addons_src_only_31aug22](https://apps.ankiweb.net/downloads/addons_src_only_31aug22.zip). But maybe we can raise the issue to the authors and drop the columns if this proves to offer significant gains.

> But if we made that configurable per field and defaulted to off, there would be no additonal storage cost unless you opt in.

Good point.

> have you considered reusing the existing field search logic? I also would have assumed that excluding a field applies to field searches as well.

I originally planned to only implement this for unqualified searches, but yes, we should probably extend it to field searches too. Will look into that.

rslib/src/notetype/schema11.rs
```diff
@@ -167,7 +167,12 @@ impl From<Notetype> for NotetypeSchema11 {
 
 /// See [crate::deckconfig::schema11::clear_other_duplicates()].
 fn clear_other_field_duplicates(other: &mut HashMap<String, Value>) {
-    for key in &["description", "plainText", "collapsed"] {
+    for key in &[
+        "description",
+        "plainText",
+        "collapsed",
+        "exclude_from_search",
```
dae, line 174: This should be camelCase

rslib/src/storage/notetype/mod.rs
```diff
@@ -392,4 +392,41 @@ impl SqliteStorage {
             .query_and_then([notetype_id], |row| Ok(row.get(0)?))?
             .collect()
     }
+
+    pub fn populate_excluded_fields(&self) -> Result<bool> {
+        self.db.execute_batch(include_str!("excluded_fields.sql"))?;
+        let mut any_excluded = false;
+        self.db
+            .prepare_cached(
+                "SELECT nt.id, f.ord, f.config FROM notetypes nt, fields f where f.ntid = nt.id",
```
dae, line 401: The notetypes table is superfluous here

RumovZ:
> > have you considered reusing the existing field search logic? I also would have assumed that excluding a field applies to field searches as well.

> I originally planned to only implement this for unqualified searches, but yes, we should probably extend it to field searches too. Will look into that.

My first question might have been unclear, so let me rephrase it:  Logic to only search a subset of fields already exists. I don't see any difference between an unqualified search with excluded fields and a field search (possibly with wildcards in the field name). So shouldn't we use the same logic to solve both problems?
The current field search logic already supports regexes, too.

---

@dae, thanks for explaining, some interesting points there.

abdnh:
> I don't see any difference between an unqualified search with excluded fields and a field search (possibly with wildcards in the field name). So shouldn't we use the same logic to solve both problems?

I see now. That looks like a better solution. Thanks for the suggestion!

dae:
The new approach is simpler, but I'm afraid I have concerns about its performance.

- You're calling .get_all_notetypes() instead of reading just the fields table. Some users embed large css/js in their templates and have hundreds of templates, so this is a potentially slow operation that's being run on every search without caching
- If a notetype has 100 fields and 1 is excluded, that's 99 separate subclauses that need to be evaluated, with the fields split each time, which is not very efficient. @RumovZ was that what you had in mind, or were you thinking of a different approach?

RumovZ:
I deliberately didn't advocate for a particular algorithm because I hadn't looked deeply enough into @abdnh's code to tell if it was superior to the current one.
All I'm saying is that there doesn't seem to be a reason to have two different routines to accomplish the same thing. We would need to do all optimisations and enhancements (like regex) twice.
I get your point that the ratio of included to excluded fields will likely tend to be different for unqualified versus field searches. However, that's not something you can rely on, so it seems more reasonable to do the optimisation based on the number of actually matched fields, rather than the kind of search.
Yes, you do need to get all notetypes for that, but `Collection::get_all_notetypes` does use a cache, or am I missing something? If not, then that shouldn't be slower than the other solution that set up a new table for each query.

How does the following sound to unify the best from both approaches: Keep identifying the relevant field indices by notetype in advance, but replace the currently used `field_at_index` function with the new `exclude_fields`.

Note, though, that this would entail one change in behaviour. Currently wildcards in unqualified searches can cross field boundaries, while that's not possible in field searches. So `foo*bar` matches a note where field 1 contains "foo" and field 2 contains "bar".
One could say, if wildcards in field searches crossed field boundaries, too, that would be less correct, but one could also argue for consistency.
Either way, in practice speed probably matters much more often here.

dae:
> Collection::get_all_notetypes does use a cache, or am I missing something?

Sorry, that was me missing something - I saw the self.storage call at the top and didn't read the rest of the function.

> All I'm saying is that there doesn't seem to be a reason to have two different routines to accomplish the same thing. We would need to do all optimisations and enhancements (like regex) twice.

If there's negligible performance impact, then I agree treating the no-exclusions case in the same way as the exclusions case is obviously better for maintenance. But if there is a measurable performance hit, we may not want to be forcing that on users that don't use this new feature.

abdnh:
Here are some poor man's benchmarks for unqualified search that I've done with a big collection:
~260K notes
78 notetypes
437 fields

## no excluded

- main (404a6c5):
n=10, total=2.2868592739105225, average=0.22868592739105226
n=100, total=27.324347496032715, average=0.27324347496032714
n=1000, total=294.95963406562805, average=0.29495963406562803
- excluded_fields:
n=10, total=2.3167638778686523, average=0.23167638778686522
n=100, total=26.96027398109436, average=0.2696027398109436
- field_at_index:
n=10, total=2.335484743118286, average=0.23354847431182862
n=100, total=26.34629249572754, average=0.2634629249572754

## 1 excluded field

- excluded_fields:
n=10, total=18.369256496429443, average=1.8369256496429442
n=100, total=207.8554105758667, average=2.078554105758667
- excluded_fields (one excluded_fields row per notetype):
n=10, total=18.058056354522705, average=1.8058056354522705
n=100, total=184.7814426422119, average=1.8478144264221192
- field_at_index:
n=10, total=26.172335386276245, average=2.6172335386276244
n=100, total=296.0932524204254, average=2.9609325242042543

## 226 excluded fields

- excluded_fields:
n=10, total=45.72403025627136, average=4.572403025627136
n=100, total=426.807009935379, average=4.26807009935379
- excluded_fields (one excluded_fields row per notetype):
n=10, total=31.09799599647522, average=3.109799599647522
n=100, total=321.28474140167236, average=3.2128474140167236
- field_at_index:
n=10, total=14.94304084777832, average=1.494304084777832
n=100, total=161.61398792266846, average=1.6161398792266846


The current field_at_index approach seems to fare better with less included fields (maybe expectedly).

## Code

Benchmark code (BTW @dae @RumovZ , if you recommend a specific method/tool/library you use for benchmarking Anki's code let me know about it):
```diff
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -604,6 +604,22 @@ class Collection(DeprecatedNamesMixin):    
         query: str,
         order: bool | str | BrowserColumns.Column = False,     
         reverse: bool = False,
+    ) -> Sequence[CardId]:
+        if n := int(os.environ.get("BENCHMARK_SEARCH", 0)):    
+            start = time.time()
+            for _ in range(n):
+                ret = self._find_cards(query, order, reverse)  
+            total = time.time() - start
+            print(f"n={n}, total={total}, average={total/n}")  
+        else:
+            return self._find_cards(query, order, reverse)     
+        return ret
+
+    def _find_cards(
+        self,
+        query: str,
+        order: bool | str | BrowserColumns.Column = False,
+        reverse: bool = False,
     ) -> Sequence[CardId]:
         """Return card ids matching the provided search.
```

I ran the following in the debug console to exclude a random set of fields:
```python
import random

notetypes = mw.col.models.all()
total_excluded = 0
total = 0
for notetype in notetypes:
    for field in notetype["flds"]:
        excluded = random.choice([False, True])
        field["excludeFromSearch"] = excluded
        total_excluded += int(excluded)
        total += 1
    mw.col.models.update_dict(notetype)

print(f"Excluded {total_excluded} out of {total} fields")
```

RumovZ:
Thanks, great to have some numbers!
I don't quite understand what you have benchmarked exactly, though. Is 'field_at_index' the current state of this PR, f4f3837? Are 'excluded_fields' and 'excluded_fields (one excluded_fields row per notetype)' also commits on your remote?

---

There is also way to do this in native SQL, which should be fastest. At the time, I had dismissed it, because 

1. it's not reusable for regex and
2. it would introduce cross-field matching.

But as 

1. we need a different approach for regex, anyway, and
2. this change in behaviour seems to be less controversial now that unqualified searches, which already behave that way, basically become field searches, too,

maybe it's now worth considering.

It would work like this: Assuming you have a notetype with 5 fields and the last 2 are excluded (i.e. the first 3 are matched), then we could write `n.flds like ?{n} || '\x1f%\x1f%'`.
Note that this would need to be repeated once for each set of consecutive matched fields. So if instead the first and last fields would be matched, we'd have `n.flds like ?{n} || '\x1f%\x1f%\x1f%\x1f%' or n.flds like '%\x1f%\x1f%\x1f%\x1f' || ?{n}`.

---

You can use Criterion for benchmarking in Anki like this:

```diff
diff --git a/rslib/benches/benchmark.rs b/rslib/benches/benchmark.rs
index 451d9e7d8..e94138d32 100644
--- a/rslib/benches/benchmark.rs
+++ b/rslib/benches/benchmark.rs
@@ -1,13 +1,20 @@
 // Copyright: Ankitects Pty Ltd and contributors
 // License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html

-use anki::card_rendering::anki_tag_benchmark;
+use anki::collection::CollectionBuilder;
+use anki::search::SortMode;
 use criterion::criterion_group;
 use criterion::criterion_main;
 use criterion::Criterion;

 pub fn criterion_benchmark(c: &mut Criterion) {
-    c.bench_function("anki_tag_parse", |b| b.iter(|| anki_tag_benchmark()));
+    c.bench_function("find_notes", |b| {
+        let mut col =
+            CollectionBuilder::new("/path/to/your/collection.anki2")
+                .build()
+                .unwrap();
+        b.iter(|| col.search_notes("foo", SortMode::NoOrder))
+    });
 }

 criterion_group!(benches, criterion_benchmark);
```

(The existing benchmark function was renamed, so you need to remove it in order to compile.) Then run with `cargo bench -F bench -p anki -- find_notes`.

abdnh:
> I don't quite understand what you have benchmarked exactly, though. Is 'field_at_index' the current state of this PR, https://github.com/ankitects/anki/commit/f4f3837c1b204e5a6b32c2ac91f32fb7bb767df7? Are 'excluded_fields' and 'excluded_fields (one excluded_fields row per notetype)' also commits on your remote?

Yes, 'field_at_index' is for the current state of the PR (I should have put the commit hashes right there).
'excluded_fields' is the old approach using a new SQL function and temporary tables (407682bdd11090191cd4a9ae23ba872f56b46219).
'excluded_fields (one excluded_fields row per notetype)' is an uncommitted change applied on top 407682bdd11090191cd4a9ae23ba872f56b46219:

```diff
diff --git a/rslib/src/search/sqlwriter.rs b/rslib/src/search/sqlwriter.rs
index bcc031fc7..a9e73076c 100644
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -191,7 +191,7 @@ impl SqlWriter<'_> {
         let any_excluded = self.col.storage.populate_excluded_fields()?;
         if any_excluded {
             let exclude_clause =
-                "coalesce((select group_concat(ord) from excluded_fields where ntid = n.mid), '')";
+                "coalesce((select ords from excluded_fields where ntid = n.mid), '')";
             write!(self.sql, "(exclude_fields(cast(n.sfld as text), {exclude_clause}, (select ord from sort_fields where ntid = n.mid)) like ?{n} escape '\\' or exclude_fields(n.flds, {exclude_clause}) like ?{n} escape '\\')",
             n = self.args.len()).unwrap();
         } else {
diff --git a/rslib/src/storage/notetype/excluded_fields.sql b/rslib/src/storage/notetype/excluded_fields.sql
index cbb683a31..ebdf7c338 100644
--- a/rslib/src/storage/notetype/excluded_fields.sql
+++ b/rslib/src/storage/notetype/excluded_fields.sql
@@ -1,10 +1,12 @@
 DROP TABLE IF EXISTS excluded_fields;
+
 CREATE TEMPORARY TABLE excluded_fields (
-  ntid integer NOT NULL,
-  ord integer NOT NULL,
-  PRIMARY KEY (ntid, ord)
+  ntid integer PRIMARY KEY NOT NULL,
+  ords text NOT NULL
 );
+
 DROP TABLE IF EXISTS sort_fields;
+
 CREATE TEMPORARY TABLE sort_fields (
   ntid integer PRIMARY KEY NOT NULL,
   ord integer NOT NULL
diff --git a/rslib/src/storage/notetype/mod.rs b/rslib/src/storage/notetype/mod.rs
index f05e3c733..3b5a9e514 100644
--- a/rslib/src/storage/notetype/mod.rs
+++ b/rslib/src/storage/notetype/mod.rs
@@ -396,6 +396,7 @@ impl SqliteStorage {
     pub fn populate_excluded_fields(&self) -> Result<bool> {
         self.db.execute_batch(include_str!("excluded_fields.sql"))?;
         let mut any_excluded = false;
+        let mut excluded_fields: HashMap<i64, Vec<String>> = HashMap::new();
         self.db
             .prepare_cached("SELECT ntid, ord, config FROM fields")?
             .query_and_then([], |row| -> Result<()> {
@@ -404,14 +405,15 @@ impl SqliteStorage {
                 let config = NoteFieldConfig::decode(row.get_ref_unwrap(2).as_blob()?)?;
                 if config.exclude_from_search {
                     any_excluded = true;
-                    self.db
-                        .prepare_cached("INSERT OR IGNORE INTO excluded_fields values(?, ?)")?
-                        .execute(params![ntid, ord])?;
+                    excluded_fields.entry(ntid).or_insert(vec![]).push(ord.to_string());
                 }
                 Ok(())
             })?
             .last();
         if any_excluded {
+            for (ntid, ords) in excluded_fields.iter() {
+                self.db.prepare_cached("insert into excluded_fields values(?, ?)")?.execute(params![ntid, ords.join(",")])?;
+            }
             self.db
                 .prepare_cached(include_str!("get_notetype.sql"))?
                 .query_and_then([], |row| -> Result<()> {
```


> There is also way to do this in native SQL, which should be fastest.

Looks clever! Will explore this approach.

RumovZ:
I still think the suggestion above (keep identifying the relevant field indices by notetype in advance, but replace the currently used field_at_index function with the new exclude_fields) could combine the best performances from your benchmarks, and it should be fairly simple to implement.
But here is my draft for the pure-SQL and probably fastest solution:

```diff
diff --git a/rslib/src/search/sqlwriter.rs b/rslib/src/search/sqlwriter.rs
index 2b6014277..15d7f5411 100644
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -3,6 +3,7 @@

 use std::borrow::Cow;
 use std::fmt::Write;
+use std::ops::Range;

 use itertools::Itertools;

@@ -514,6 +515,48 @@ impl SqlWriter<'_> {
         Ok(())
     }

+    fn write_single_field_new(&mut self, field_name: &str, val: &str) -> Result<()> {
+        let field_indicies_by_notetype =
+            self.num_fields_and_fields_indices_by_notetype(field_name)?;
+        if field_indicies_by_notetype.is_empty() {
+            write!(self.sql, "false").unwrap();
+            return Ok(());
+        }
+
+        self.args.push(to_sql(val).into());
+        let arg_idx = self.args.len();
+        let field_idx_str = format!("' || ?{arg_idx} || '");
+        let other_idx_str = "%".to_string();
+
+        let notetype_clause =
+            |(mid, num_fields, fields): &(NotetypeId, usize, Vec<Range<u32>>)| -> String {
+                let field_index_clause = |range: &Range<u32>| {
+                    let f = (0..*num_fields)
+                        .filter_map(|i| {
+                            if i as u32 == range.start {
+                                Some(&field_idx_str)
+                            } else if range.contains(&(i as u32)) {
+                                None
+                            } else {
+                                Some(&other_idx_str)
+                            }
+                        })
+                        .join("\x1f");
+                    format!("n.flds like '{f}' escape '\\'")
+                };
+
+                let all_field_clauses = fields.iter().map(field_index_clause).join(" or ");
+                format!("(n.mid = {mid} and ({all_field_clauses}))",)
+            };
+        let all_notetype_clauses = field_indicies_by_notetype
+            .iter()
+            .map(notetype_clause)
+            .join(" or ");
+        write!(self.sql, "({all_notetype_clauses})").unwrap();
+
+        Ok(())
+    }
+
     fn fields_indices_by_notetype(
         &mut self,
         field_name: &str,
@@ -535,7 +578,31 @@ impl SqlWriter<'_> {
         }

         // for now, sort the map for the benefit of unit tests
-        field_map.sort();
+        // field_map.sort();
+
+        Ok(field_map)
+    }
:...skipping...
diff --git a/rslib/src/search/sqlwriter.rs b/rslib/src/search/sqlwriter.rs
index 2b6014277..15d7f5411 100644
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -3,6 +3,7 @@

 use std::borrow::Cow;
 use std::fmt::Write;
+use std::ops::Range;

 use itertools::Itertools;

@@ -514,6 +515,48 @@ impl SqlWriter<'_> {
         Ok(())
     }

+    fn write_single_field_new(&mut self, field_name: &str, val: &str) -> Result<()> {
+        let field_indicies_by_notetype =
+            self.num_fields_and_fields_indices_by_notetype(field_name)?;
+        if field_indicies_by_notetype.is_empty() {
+            write!(self.sql, "false").unwrap();
+            return Ok(());
+        }
+
+        self.args.push(to_sql(val).into());
+        let arg_idx = self.args.len();
+        let field_idx_str = format!("' || ?{arg_idx} || '");
+        let other_idx_str = "%".to_string();
+
+        let notetype_clause =
+            |(mid, num_fields, fields): &(NotetypeId, usize, Vec<Range<u32>>)| -> String {
+                let field_index_clause = |range: &Range<u32>| {
+                    let f = (0..*num_fields)
+                        .filter_map(|i| {
+                            if i as u32 == range.start {
+                                Some(&field_idx_str)
+                            } else if range.contains(&(i as u32)) {
+                                None
+                            } else {
+                                Some(&other_idx_str)
+                            }
+                        })
+                        .join("\x1f");
+                    format!("n.flds like '{f}' escape '\\'")
+                };
+
+                let all_field_clauses = fields.iter().map(field_index_clause).join(" or ");
+                format!("(n.mid = {mid} and ({all_field_clauses}))",)
+            };
+        let all_notetype_clauses = field_indicies_by_notetype
+            .iter()
+            .map(notetype_clause)
+            .join(" or ");
+        write!(self.sql, "({all_notetype_clauses})").unwrap();
+
+        Ok(())
+    }
+
     fn fields_indices_by_notetype(
         &mut self,
         field_name: &str,
@@ -535,7 +578,31 @@ impl SqlWriter<'_> {
         }

         // for now, sort the map for the benefit of unit tests
-        field_map.sort();
+        // field_map.sort();
+
+        Ok(field_map)
+    }
+
+    fn num_fields_and_fields_indices_by_notetype(
+        &mut self,
+        field_name: &str,
+    ) -> Result<Vec<(NotetypeId, usize, Vec<Range<u32>>)>> {
+        let notetypes = self.col.get_all_notetypes()?;
+        let matches_glob = glob_matcher(field_name);
+
+        let mut field_map = vec![];
+        for nt in notetypes.values() {
+            let matched_fields = nt
+                .fields
+                .iter()
+                .filter_map(|field| {
+                    matches_glob(&field.name).then(|| field.ord.unwrap_or_default())
+                })
+                .collect_ranges();
+            if !matched_fields.is_empty() {
+                field_map.push((nt.id, nt.fields.len(), matched_fields));
+            }
+        }

         Ok(field_map)
     }
@@ -646,6 +713,43 @@ impl RequiredTable {
     }
 }

+trait CollectRanges {
+    type Item;
+    fn collect_ranges(self) -> Vec<Range<Self::Item>>;
+}
+
+impl<
+        Idx: Copy + PartialOrd + std::ops::Add<Idx, Output = Idx> + From<u8>,
+        I: IntoIterator<Item = Idx>,
+    > CollectRanges for I
+{
+    type Item = Idx;
+
+    fn collect_ranges(self) -> Vec<Range<Self::Item>> {
+        let mut result = Vec::new();
+        let mut iter = self.into_iter();
+        let next = iter.next();
+        if next.is_none() {
+            return result;
+        }
+        let mut start = next.unwrap();
+        let mut end = next.unwrap();
+
+        for i in iter {
+            if i == end + 1.into() {
+                end = end + 1.into();
+            } else {
+                result.push(start..end + 1.into());
+                start = i;
+                end = i;
+            }
+        }
+        result.push(start..end + 1.into());
+
+        result
+    }
+}
+
 impl Node {
     fn required_table(&self) -> RequiredTable {
         match self {
```

dae:
Interested to hear how the native SQL approach compares. If I'm reading the initial benchmarks correctly, any excluded fields means a 10x performance decrease on searches? That's a fairly hefty price to pay for all searches.

dae:
(I seem to recall regex searches are 2-4x, and nc 4-8x, but my memory could be wrong here. And apart from the 'ignore accents by default' option, these are opt-in features, rather than ones that penalize every search)

abdnh:
The update I just pushed (which I think what Rumo was suggesting) performs better than the ones I benchmarked so far, but still has a 8x performance hit. I'll look into the pure-SQL approach and post the updated benchmarks.

rslib/src/search/sqlwriter.rs
```diff
@@ -184,16 +185,69 @@ impl SqlWriter<'_> {
...
+    /// Return excluded fields for notetypes that have some but not all fields
+    /// excluded
+    #[allow(clippy::type_complexity)]
+    fn excluded_fields_indices_by_notetype(
+        &mut self,
+    ) -> Result<Option<Vec<(NotetypeId, u32, HashSet<u32>)>>> {
+        let notetypes = self.col.get_all_notetypes()?;
+        let mut any_excluded = false;
+        let mut field_map = vec![];
+        for nt in notetypes.values() {
```
danielzgtg, line 197: In Rust it's preferred to map (or flatmap) the iterator then collect it, instead of creating a separate vec[] and looping over an iterator.

abdnh:
Here are the updated benchmarks:

<details>
<summary>Benchmarks</summary>

~260K notes
78 notetypes
437 fields

## no excluded

- main (404a6c5):
n=10, total=2.2868592739105225, average=0.22868592739105226
n=100, total=27.324347496032715, average=0.27324347496032714
n=1000, total=294.95963406562805, average=0.29495963406562803
- exclude_fields (407682b):
n=10, total=2.3167638778686523, average=0.23167638778686522
n=100, total=26.96027398109436, average=0.2696027398109436
- exclude_fields (8664bcee56199e09a38ffda66f555de9b537df75):
n=10, total=2.2749314308166504, average=0.22749314308166504
n=100, total=23.49696683883667, average=0.2349696683883667
- field_at_index (f4f3837c1b204e5a6b32c2ac91f32fb7bb767df7):
n=10, total=2.335484743118286, average=0.23354847431182862
n=100, total=26.34629249572754, average=0.2634629249572754
- pure SQL:
n=10, total=2.324925184249878, average=0.2324925184249878
n=100, total=25.410995960235596, average=0.25410995960235594

## 1 excluded field

- exclude_fields (407682b):
n=10, total=18.369256496429443, average=1.8369256496429442
n=100, total=207.8554105758667, average=2.078554105758667
- exclude_fields (one excluded_fields row per notetype):
n=10, total=18.058056354522705, average=1.8058056354522705
n=100, total=184.7814426422119, average=1.8478144264221192
- exclude_fields (8664bcee56199e09a38ffda66f555de9b537df75):
n=10, total=14.659024238586426, average=1.4659024238586427
n=100, total=155.02491235733032, average=1.5502491235733031
- field_at_index (f4f3837c1b204e5a6b32c2ac91f32fb7bb767df7):
n=10, total=26.172335386276245, average=2.6172335386276244
n=100, total=296.0932524204254, average=2.9609325242042543
- pure SQL:
n=10, total=5.571667671203613, average=0.5571667671203613
n=100, total=62.19403600692749, average=0.6219403600692749

## 226 excluded fields

- exclude_fields (407682b):
n=10, total=45.72403025627136, average=4.572403025627136
n=100, total=426.807009935379, average=4.26807009935379
- exclude_fields (one excluded_fields row per notetype):
n=10, total=31.09799599647522, average=3.109799599647522
n=100, total=321.28474140167236, average=3.2128474140167236
- exclude_fields (8664bcee56199e09a38ffda66f555de9b537df75):
n=10, total=19.350993871688843, average=1.9350993871688842
n=100, total=209.0182328224182, average=2.090182328224182
- field_at_index (f4f3837c1b204e5a6b32c2ac91f32fb7bb767df7):
n=10, total=14.94304084777832, average=1.494304084777832
n=100, total=161.61398792266846, average=1.6161398792266846
- pure SQL:
n=10, total=6.385999441146851, average=0.638599944114685
n=100, total=69.96412920951843, average=0.6996412920951843

</details>

Looks like the pure-SQL solution is the way to go, for non-regex search at least. It's Rumo's code modified for unqualified search:

<details>
<summary>Code</summary>

```diff
diff --git a/rslib/src/search/sqlwriter.rs b/rslib/src/search/sqlwriter.rs
index 7672ea9cb..9918bdef1 100644
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -2,8 +2,8 @@
 // License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
 use std::borrow::Cow;
-use std::collections::HashSet;
 use std::fmt::Write;
+use std::ops::Range;
 
 use itertools::Itertools;
 
@@ -185,55 +185,47 @@ impl SqlWriter<'_> {
         Ok(())
     }
 
-    /// Return excluded fields for notetypes that have some but not all fields
-    /// excluded
-    #[allow(clippy::type_complexity)]
-    fn excluded_fields_indices_by_notetype(
-        &mut self,
-    ) -> Result<Option<Vec<(NotetypeId, u32, HashSet<u32>)>>> {
-        let notetypes = self.col.get_all_notetypes()?;
-        let mut any_excluded = false;
-        let mut field_map = vec![];
-        for nt in notetypes.values() {
-            let mut matched_fields: HashSet<u32> = HashSet::new();
-            for field in &nt.fields {
-                any_excluded |= field.config.exclude_from_search;
-                if field.config.exclude_from_search {
-                    matched_fields.insert(field.ord.unwrap_or_default());
-                }
-            }
-            if nt.fields.len() != matched_fields.len() {
-                field_map.push((nt.id, nt.config.sort_field_idx, matched_fields));
-            }
-        }
-        if any_excluded {
-            Ok(Some(field_map))
-        } else {
-            Ok(None)
-        }
-    }
-
     fn write_unqualified(&mut self, text: &str) -> Result<()> {
         // implicitly wrap in %
         let text = format!("%{}%", &to_sql(text));
         self.args.push(text);
         let arg_idx = self.args.len();
 
-        if let Some(field_indicies_by_notetype) = self.excluded_fields_indices_by_notetype()? {
-            let notetype_clause =
-                |(mid, sortf, fields): &(NotetypeId, u32, HashSet<u32>)| -> String {
-                    let mut clauses = vec![];
-                    let mut joined_fields = String::from(if fields.is_empty() { "" } else { ", " });
-                    joined_fields.push_str(&fields.iter().join(", "));
-                    clauses.push(format!("exclude_fields(n.flds, {sortf}{joined_fields}) like ?{arg_idx} escape '\\'"));
-                    if !fields.contains(sortf) {
-                        clauses.push(format!("n.sfld like ?{arg_idx} escape '\\'"));
-                    }
-                    format!(
-                        "(n.mid = {mid} and ({clauses}))",
-                        clauses = clauses.join(" or ")
-                    )
+        if let Some(field_indicies_by_notetype) = self.included_fields_by_notetype()? {
+            let field_idx_str = format!("' || ?{arg_idx} || '");
+            let other_idx_str = "%".to_string();
+
+            let notetype_clause = |(mid, num_fields, sortf_excluded, fields): &(
+                NotetypeId,
+                usize,
+                bool,
+                Vec<Range<u32>>,
+            )|
+             -> String {
+                let field_index_clause = |range: &Range<u32>| {
+                    let f = (0..*num_fields)
+                        .filter_map(|i| {
+                            if i as u32 == range.start {
+                                Some(&field_idx_str)
+                            } else if range.contains(&(i as u32)) {
+                                None
+                            } else {
+                                Some(&other_idx_str)
+                            }
+                        })
+                        .join("\x1f");
+                    format!("n.flds like '{f}' escape '\\'")
                 };
+                let mut all_field_clauses: Vec<String> =
+                    fields.iter().map(field_index_clause).collect();
+                if !sortf_excluded {
+                    all_field_clauses.push(format!("n.sfld like ?{arg_idx} escape '\\'"));
+                }
+                format!(
+                    "(n.mid = {mid} and ({all_field_clauses}))",
+                    all_field_clauses = all_field_clauses.join(" or ")
+                )
+            };
             let all_notetype_clauses = field_indicies_by_notetype
                 .iter()
                 .map(notetype_clause)
@@ -594,6 +586,38 @@ impl SqlWriter<'_> {
         Ok(field_map)
     }
 
+    #[allow(clippy::type_complexity)]
+    fn included_fields_by_notetype(
+        &mut self,
+    ) -> Result<Option<Vec<(NotetypeId, usize, bool, Vec<Range<u32>>)>>> {
+        let notetypes = self.col.get_all_notetypes()?;
+        let mut any_excluded = true;
+        let mut field_map = vec![];
+        for nt in notetypes.values() {
+            let mut sortf_excluded = false;
+            let matched_fields = nt
+                .fields
+                .iter()
+                .filter_map(|field| {
+                    (!field.config.exclude_from_search).then(|| {
+                        any_excluded = false;
+                        let ord = field.ord.unwrap_or_default();
+                        sortf_excluded |= ord == nt.config.sort_field_idx;
+                        ord
+                    })
+                })
+                .collect_ranges();
+            if !matched_fields.is_empty() {
+                field_map.push((nt.id, nt.fields.len(), sortf_excluded, matched_fields));
+            }
+        }
+        if any_excluded {
+            Ok(Some(field_map))
+        } else {
+            Ok(None)
+        }
+    }
+
     fn write_dupe(&mut self, ntid: NotetypeId, text: &str) -> Result<()> {
         let text_nohtml = strip_html_preserving_media_filenames(text);
         let csum = field_checksum(text_nohtml.as_ref());
@@ -700,6 +724,43 @@ impl RequiredTable {
     }
 }
 
+trait CollectRanges {
+    type Item;
+    fn collect_ranges(self) -> Vec<Range<Self::Item>>;
+}
+
+impl<
+        Idx: Copy + PartialOrd + std::ops::Add<Idx, Output = Idx> + From<u8>,
+        I: IntoIterator<Item = Idx>,
+    > CollectRanges for I
+{
+    type Item = Idx;
+
+    fn collect_ranges(self) -> Vec<Range<Self::Item>> {
+        let mut result = Vec::new();
+        let mut iter = self.into_iter();
+        let next = iter.next();
+        if next.is_none() {
+            return result;
+        }
+        let mut start = next.unwrap();
+        let mut end = next.unwrap();
+
+        for i in iter {
+            if i == end + 1.into() {
+                end = end + 1.into();
+            } else {
+                result.push(start..end + 1.into());
+                start = i;
+                end = i;
+            }
+        }
+        result.push(start..end + 1.into());
+
+        result
+    }
+}
+
 impl Node {
     fn required_table(&self) -> RequiredTable {
         match self {

```

</details>

dae:
Much nicer! It might be worth adding a (slower) to the ignore option in the fields dialog, like for the accent ignoring in the preferences, but the penalty is considerably lower than before which is great.

RumovZ:
Thanks for running all these benchmarks, @abdnh! Very instructive.
Since regex searches have to call into Rust anyway the performance hit might even be less decisive there.

abdnh:
The `nc`, `re`, and `w` operators should support field exclusion now. I'm going to test things more and maybe benchmark those too.

abdnh:
Here are some numbers for the search operators. Field search performs better (whether regex or not), but the `nc` suffers badly with a lot of excluded fields.

## Unqualified search ignoring accents ("nc:foo")

### no excluded

- main (404a6c5):
n=10, total=75.3080677986145, average=7.53080677986145
- pure SQL:
n=10, total=79.09307670593262, average=7.909307670593262

### 1 excluded field

- pure SQL:
n=10, total=86.28219151496887, average=8.628219151496888


### 226 excluded fields

- pure SQL:
n=10, total=180.5921745300293, average=18.05921745300293

## Regex search ("re:foo")

### no excluded

- main (404a6c5):
n=10, total=2.81099796295166, average=0.281099796295166
- pure SQL:
n=10, total=2.7760345935821533, average=0.27760345935821534

### 1 excluded field

- pure SQL:
n=10, total=8.504792928695679, average=0.8504792928695679

### 226 excluded fields

- pure SQL:
n=10, total=24.39304542541504, average=2.439304542541504


## Word boundary search ("w:foo")

### no excluded

- main (404a6c5):
n=10, total=9.33099889755249, average=0.933099889755249
- pure SQL:
n=10, total=9.537001371383667, average=0.9537001371383667

### 1 excluded field

- pure SQL:
n=10, total=15.338004112243652, average=1.5338004112243653

### 226 excluded fields

- pure SQL:
n=10, total=28.030372381210327, average=2.803037238121033


## Field search ("Front:foo")

### no excluded

- main (404a6c5):
n=10, total=0.3690526485443115, average=0.03690526485443115
- pure SQL:
n=10, total=0.10098981857299805, average=0.010098981857299804

### 1 excluded field

- pure SQL:
n=10, total=0.02393198013305664, average=0.002393198013305664

### 226 excluded fields

- pure SQL:
n=10, total=0.12103962898254395, average=0.012103962898254394


## Regex field search ("Front:re:foo")

### no excluded

- main (404a6c5):
n=10, total=1.0740084648132324, average=0.10740084648132324
- pure SQL:
n=10, total=1.0590095520019531, average=0.10590095520019531

### 1 excluded field

- pure SQL:
n=10, total=0.14502978324890137, average=0.014502978324890137

### 226 excluded fields

- pure SQL:
n=10, total=1.0159916877746582, average=0.10159916877746582

dae:
Those numbers look reasonable - adding extra excluded fields appears to be a sublinear slowdown, which is about as much as you could hope for. If you're ready for me to review/merge this, please let me know.

A question: does it make sense to apply exclusions when users are searching in a particular field?

abdnh:
> If you're ready for me to review/merge this, please let me know.

Yes, I think it's ready for review.

> A question: does it make sense to apply exclusions when users are searching in a particular field?

Exclusions are now applied here, but I don't have a strong opinion about this. I imagine users will have different expectations here, some expecting excluded fields to be absolutely excluded from every search, while others might occasionally want a simple way to search for a field across all notetypes without having to go check all notetypes that have the field and toggle off the option. Thinking about it more, maybe it's more common to want to exclude a certain field name from all notetypes (like references/resources fields named similarly across notetypes - I tend to add a field named "References" to a lot of my notetypes for example), so in this case I guess it makes sense to make the latter scenario easier.

rslib/src/storage/sqlite.rs
```diff
@@ -190,6 +192,31 @@ fn add_regexp_tags_function(db: &Connection) -> rusqlite::Result<()> {
...
+fn add_exclude_fields_function(db: &Connection) -> rusqlite::Result<()> {
+    db.create_scalar_function(
+        "exclude_fields",
+        -1,
+        FunctionFlags::SQLITE_DETERMINISTIC,
+        move |ctx| {
+            assert!(ctx.len() > 1, "not enough arguments");
+
+            let joined_fields = ctx.get_raw(0).as_str()?;
+            let fields: Vec<&str> = joined_fields.split('\x1f').collect();
```
dae, line 206: The .collect() creates a vector that you don't need - you could remove this line and replace 'fields.iter()' below with joined_fields.split('\x1f'). This may improve performance somewhat.

rslib/src/search/sqlwriter.rs
```diff
@@ -514,6 +579,35 @@ impl SqlWriter<'_> {
         Ok(())
     }
 
+    #[allow(clippy::type_complexity)]
```
dae, line 582: Rather than silencing this warning, it would be nicer to create a struct for the return value.

dae:
I don't have strong feelings, but I slightly lean towards ignoring the setting when searching in a given field. That allows the user to search in that field without having to change their field settings back and forth, and I suspect it's what the majority of users would expect when they've explicitly specified a field. We could potentially change the option name to 'exclude this field from unqualified searches' if you thought that made things clearer. WDYT?

rslib/src/search/sqlwriter.rs
```diff
@@ -598,25 +766,47 @@ impl SqlWriter<'_> {
...
         self.args.push(format!(r"(?i){}", word));
-    }
+        let arg_idx = self.args.len();
+        if let Some(field_indices_by_notetype) = self.excluded_fields_by_notetype()? {
+            let notetype_clause = |UnqualifiedRegexSearchContext{mid, fields}: &UnqualifiedRegexSearchContext| -> String {
+                let clause = if fields.is_empty() {
+                    format!("{flds_expr} regexp ?{arg_idx}")
+                } else {
+                    let indices = fields.iter().join(",");
+                    format!("exclude_fields({flds_expr}, {indices}) regexp ?{arg_idx}")
```
RumovZ, line 788: This calls two Rust functions. Wouldn't it make sense to use the existing `regexp_fields` function instead?

rslib/src/search/sqlwriter.rs
```diff
@@ -184,30 +184,79 @@ impl SqlWriter<'_> {
...
+            let other_idx_str = "%".to_string();
+
+            let notetype_clause =
+                |UnqualifiedSearchContext {
+                     mid,
+                     num_fields,
+                     sortf_excluded,
+                     fields,
+                 }: &UnqualifiedSearchContext|
+                 -> String {
```
dae, line 221: This may come down to personal preference, but I find destructuring in a closure argument makes the code a bit hard to read. How about the following?

```diff
diff --git a/rslib/src/search/sqlwriter.rs b/rslib/src/search/sqlwriter.rs
index 0d3a3fc9d..f5dc84718 100644
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -211,38 +211,32 @@ impl SqlWriter<'_> {
             let field_idx_str = format!("' || ?{arg_idx} || '");
             let other_idx_str = "%".to_string();

-            let notetype_clause =
-                |UnqualifiedSearchContext {
-                     mid,
-                     num_fields,
-                     sortf_excluded,
-                     fields,
-                 }: &UnqualifiedSearchContext|
-                 -> String {
-                    let field_index_clause = |range: &Range<u32>| {
-                        let f = (0..*num_fields)
-                            .filter_map(|i| {
-                                if i as u32 == range.start {
-                                    Some(&field_idx_str)
-                                } else if range.contains(&(i as u32)) {
-                                    None
-                                } else {
-                                    Some(&other_idx_str)
-                                }
-                            })
-                            .join("\x1f");
-                        format!("{flds_expr} like '{f}' escape '\\'")
-                    };
-                    let mut all_field_clauses: Vec<String> =
-                        fields.iter().map(field_index_clause).collect();
-                    if !sortf_excluded {
-                        all_field_clauses.push(format!("{sfld_expr} like ?{arg_idx} escape '\\'"));
-                    }
-                    format!(
-                        "(n.mid = {mid} and ({all_field_clauses}))",
-                        all_field_clauses = all_field_clauses.join(" or ")
-                    )
+            let notetype_clause = |ctx: &UnqualifiedSearchContext| -> String {
+                let field_index_clause = |range: &Range<u32>| {
+                    let f = (0..ctx.num_fields)
+                        .filter_map(|i| {
+                            if i as u32 == range.start {
+                                Some(&field_idx_str)
+                            } else if range.contains(&(i as u32)) {
+                                None
+                            } else {
+                                Some(&other_idx_str)
+                            }
+                        })
+                        .join("\x1f");
+                    format!("{flds_expr} like '{f}' escape '\\'")
                 };
+                let mut all_field_clauses: Vec<String> =
+                    ctx.fields.iter().map(field_index_clause).collect();
+                if !ctx.sortf_excluded {
+                    all_field_clauses.push(format!("{sfld_expr} like ?{arg_idx} escape '\\'"));
+                }
+                format!(
+                    "(n.mid = {mid} and ({all_field_clauses}))",
+                    mid = ctx.mid,
+                    all_field_clauses = all_field_clauses.join(" or ")
+                )
+            };
             let all_notetype_clauses = field_indicies_by_notetype
                 .iter()
                 .map(notetype_clause)
```

rslib/src/search/sqlwriter.rs
```diff
@@ -490,18 +539,39 @@ impl SqlWriter<'_> {
...
+                            Some(&field_idx_str)
+                        } else if range.contains(&(i as u32)) {
+                            None
+                        } else {
+                            Some(&other_idx_str)
+                        }
+                    })
+                    .join("\x1f");
+                format!("n.flds like '{f}' escape '\\'")
+            };
```
dae, line 573: I think I'm missing something here - if we ignore the exclusions for single field searches, why does this routine need to change?

abdnh: I thought maybe we want to keep using this approach since it's faster?

rslib/src/search/sqlwriter.rs
abdnh: Makes sense. It's even 3 functions in the no_combining case.

rslib/src/search/sqlwriter.rs
dae: Ah, that's what I was missing :-)

dae:
Thanks Abdo, I've made some minor naming tweaks to make things a bit clearer, but it seems to be in a pretty good place. Will merge this in and get a new beta out. Thanks also for the feedback/suggestions, Rumo.

PR #2395, opened 2023-02-23
Simplify NoteEditor by replacing Pane components with Collapsible

### LINKED ISSUE #2296: tags are slightly truncated at bottom
https://forums.ankiweb.net/t/minor-issues-tag-layouts/25942/3

kleinerpirat:
Reverts most changes introduced with #2046 and uses `Collapsible.svelte` for the tag editor.

---
Since we've already gone from arbitrary resizing to snapping the tag area, we might as well make it a `Collapsible` instead. This greatly reduces the complexity of the editor and should make it feel more polished, because there were some glitches with the previous approach. A new component `ScrollArea` takes over the scroll indication of removed `PaneContent.svelte`.

Also, the tag editor has a focus outline and border now.

## New Look:

image

---
Closes #2296.


kleinerpirat:
Eslint is unhappy with the import order set by `./ninja format` again. I reckon if you rerun this check after #2391 is merged, it will pass.

dae:
Thanks Matthias, this looks slick, and any complexity reduction is much appreciated.

For future reference, the tests can be fixed by running ./ninja fix (which reorders the imports) and then ./ninja format (which fixes the whitespace that eslint outputs)

PR #2398, opened 2023-02-24
Add option to show new cards regardless of the review limit

### LINKED ISSUE #1834: Consider providing a way to bypass review limits on new cards

RumovZ:
Closes #1834.

I must admit, this was harder than I would have thought. The code was tightly coupled to the previous behaviour.
I took the opportunity to simplify the limits code a bit (or so I hope). In a nutshell, the most important change is that the previous code enforced the review limit on new cards by capping the new limit after the review gathering, whereas the new code decrements the two limits in lockstep _during_ the review gathering.

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -39,6 +39,11 @@ deck-config-tab-description =
     - `Preset`: The limit is shared with all decks using this preset.
     - `This deck`: The limit is specific to this deck.
     - `Today only`: Make a temporary change to this deck's limit.
+deck-config-new-cards-ignore-review-limit = New cards ignore review limit
+deck-config-new-cards-ignore-review-limit-tooltip =
+    By default, the review limit also applies to new cards and no new cards will be
+    shown when the review limit has been reached. If this option is enabled, new cards
+    will be shown regardless of the review limit.
 
 ## Daily limit tabs: please try to keep these as short as the English version,
 ## as longer text will not fit on small screens.
--- a/rslib/src/deckconfig/schema11.rs
+++ b/rslib/src/deckconfig/schema11.rs
@@ -51,6 +51,7 @@ pub struct DeckConfSchema11 {
     // NOTE: if adding new ones, make sure to update clear_other_duplicates()
     #[serde(default)]
     new_mix: i32,
+    /// This has been repurposed as a boolean: `new_ignore_review_limit`
     #[serde(default)]
     new_per_day_minimum: u32,
     #[serde(default)]
@@ -293,7 +294,7 @@ impl From<DeckConfSchema11> for DeckConfig {
                 relearn_steps: c.lapse.delays,
                 new_per_day: c.new.per_day,
                 reviews_per_day: c.rev.per_day,
-                new_per_day_minimum: c.new_per_day_minimum,
+                new_ignore_review_limit: c.new_per_day_minimum > 0,
                 initial_ease: (c.new.initial_factor as f32) / 1000.0,
                 easy_multiplier: c.rev.ease4,
                 hard_multiplier: c.rev.hard_factor,
@@ -403,7 +404,7 @@ impl From<DeckConfig> for DeckConfSchema11 {
             },
             other: top_other,
             new_mix: i.new_mix,
-            new_per_day_minimum: i.new_per_day_minimum,
+            new_per_day_minimum: i.new_ignore_review_limit as u32,
             interday_learning_mix: i.interday_learning_mix,
             review_order: i.review_order,
             new_sort_order: i.new_card_sort_order,
--- a/rslib/src/decks/limits.rs
+++ b/rslib/src/decks/limits.rs
@@ -16,6 +16,12 @@ use crate::deckconfig::DeckConfigId;
 use crate::pb::decks::deck::normal::DayLimit;
 use crate::prelude::*;
 
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum LimitKind {
+    Review,
+    New,
+}
+
 impl NormalDeck {
     /// The deck's review limit for today, or its regular one, if any is
     /// configured.
@@ -50,8 +56,9 @@ impl DayLimit {
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub(crate) struct RemainingLimits {
-    pub review: u32,
-    pub new: u32,
+    pub(crate) review: u32,
+    pub(crate) new: u32,
+    pub(crate) cap_new_to_review: bool,
 }
 
 impl RemainingLimits {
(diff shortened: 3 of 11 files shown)


ftl/core/deck-config.ftl
```diff
@@ -39,6 +39,11 @@ deck-config-tab-description =
     - `Preset`: The limit is shared with all decks using this preset.
     - `This deck`: The limit is specific to this deck.
     - `Today only`: Make a temporary change to this deck's limit.
+deck-config-new-cards-ignore-review-limit = New cards ignore review limit
+deck-config-new-cards-ignore-review-limit-tooltip =
+    By default, the review limit also applies to new cards and no new cards will be
+    shown when the review limit has been reached. If this option is enabled, new cards
+    will be shown regardless of the review limit.
```
RumovZ, line 46: The documentation is why I've marked it as a draft.
How do we proceed with the `deck-config-limit-new-bound-by-reviews` message? Should we remove it because it is now covered in its own section and does not apply universally anymore? Not sure if the `deck-config-limit-interday-bound-by-reviews` message would look a bit out of place then.

dae: new-bound-by no longer seems necessary. interday-bound-by still has some value, though we should probably only show it when showing the review limit, and perhaps remove the ", and finally new cards". WDYT?

rslib/src/deckconfig/schema11.rs
```diff
@@ -51,6 +51,7 @@ pub struct DeckConfSchema11 {
     // NOTE: if adding new ones, make sure to update clear_other_duplicates()
     #[serde(default)]
     new_mix: i32,
+    /// This has been repurposed as a boolean: `new_ignore_review_limit`
```
dae, line 54: Since this field was never used, another option would be to just drop the existing key and create a new one that's a bool.

rslib/src/decks/limits.rs
```diff
@@ -71,40 +78,89 @@ impl RemainingLimits {
...
         }
     }
 
     pub(crate) fn cap_to(&mut self, limits: RemainingLimits) {
         self.review = self.review.min(limits.review);
         self.new = self.new.min(limits.new);
     }
+
+    /// True if some limit was decremented to 0.
+    fn decremented_to_zero(&mut self, kind: LimitKind) -> bool {
```
dae, line 141: I find this name a bit confusing. How about 'decrement' with a docstring that explains what the bool represents?

dae (commented):
I gave this a quick try and think I may have hit a bug. I started with a large deck that had 9999+ pending reviews, and all the (sub)decks were using the same preset. It was showing no new cards, because the review limit was reached. When I turned the new-ignores-review option on, the new card number increased from zero to the expected amount, so all good so far. Then I switched the option off again, and gave the top level deck a new preset, turning on the option for only that deck. Then things didn't work - it showed the new count in the overview as 0 (+ 200), despite the fact that burying was disabled in all presets. The deck list correctly showed 200.

rslib/src/scheduler/queue/builder/mod.rs
```diff
@@ -482,4 +484,37 @@ mod test {
...
+    #[test]
+    fn review_limit_only_applies_to_new_cards_from_the_same_deck() {
+        let mut col = Collection::new_v3();
+        // parent deck with more new than reviews, but new cards ignore the review limit
+        let parent = DeckAdder::new("parent")
+            .with_config(|config| {
+                config.inner.reviews_per_day = 0;
+                config.inner.new_per_day = 1;
+                config.inner.new_ignore_review_limit = true;
+            })
```
dae, line 510: These sort of helpers are great, and should help encourage us to write more tests.

ftl/core/deck-config.ftl
RumovZ: Sounds good. Should I add a new message to drop ", and finally new cards" or does changing the English one suffice?

rslib/src/decks/limits.rs
RumovZ: I actually started out with `decrement`, but ain't happy with either name. I guess that means there's something wrong with the function!
Maybe I can return a struct instead and then call something like `limits.decrement(kind).decremented_to_zero`.

RumovZ:
> I gave this a quick try and think I may have hit a bug.

Dang. I thought I had it, but apparantly I just lost my patience. It's not a trivial thing to fix, so back to the drawing board.  🙁

dae:
If no better solutions present themselves, would having the option be taken only from the clicked-on-deck like the display order options make things easier? Presumably only for the queue building code, and not the deck list.

RumovZ:
Yes, it seems that for the queue building code it's currently impossible to handle certain cases where parent and child have this setting set differently. Let's take the following setup:
| Deck | Review Limit | New Limit | Review Cards | New Cards | New Cards Ignore Review Limit |
| :-: | :-: | :-: | :-: | :-: | :-: |
| parent | 0 | 1 | 0 | 0 | ✔️ |
| parent::child | 1 | 1 | ? | 1 | ❌  |

In this example, the parent should queue the child's new card if the child has no review cards, but not if the child does have review cards.
To make this work we would need to keep iterating the due review cards until all children's limits are exhausted, too, before moving on to queue new cards. Not sure if that's practical. We would basically need to keep track of an addtional queue with bury state and all.

Unfortunately, applying the parent's setting to all children doesn't seem to work, either.

| Deck | Review Limit | New Limit | Review Cards | New Cards | New Cards Ignore Review Limit |
| :-: | :-: | :-: | :-: | :-: | :-: |
| parent | 1 | 1 | 0 | 0 | ✔️ |
| parent::child | 1 | 1 | 1 | 1 | ❌  |

Here the child count wouldn't show any new cards, but the parent count would include the child's new card. I feel like this phantom new card would confuse things a lot. And this too  could turn out messy because in the summation logic we would need to keep track not only of the capped counts for each deck, but also the uncapped counts, so we can pass them up to the respective parents.
Did you mean this when you said this solution would not probably not be applied to the deck list?

Could we make this setting collection-global?

dae:
I was thinking that the deck list can already be inconsistent - users get surprised when the sum of child decks exceeds the parent for example. But I think you're right here, making this a collection-wide setting is going to remove a bunch of added complexity and confusion, and the ability to configure it on a per-deck basis is not something anyone's asked for. The only downside is it becomes less discoverable if placed in the preferences, or needs an explanation in the tooltip in the deck options that it applies collection wide.

PR #2405, opened 2023-02-27
Add a button in the preferences to reset window sizes

### LINKED ISSUE #2403: Add a button in the preferences screen that resets all window sizes
Will be helpful for issues like https://forums.ankiweb.net/t/filtered-deck-in-2-1-60-attributes-window-is-opening-in-full-screen-and-goes-off-screen/27837, and other cases where windows fail to appear on crash on opening.

abdnh:
Closes #2403

This also resets windows and splitters state to be safe.

image


dae:
Thanks for tackling this so quickly Abdo!

PR #2406, opened 2023-02-28
Fix double scrollbars in deck options

### LINKED ISSUE #2401: Double scrollbars in deck options
If the tooltip is bigger than the screen, we end up with two sets of scrollbars.
2023-02-27T16:56:29,735247160+10:00

winterhazel:
Closes #2401.

This change also prevents the sticky element from hiding and the page from shifting to the left while a modal is open.

### DIFF
--- a/sass/base.scss
+++ b/sass/base.scss
@@ -40,11 +40,11 @@ body {
 }
 
 html {
-    overflow-x: hidden;
     overscroll-behavior: none;
 }
 
 body {
+    overflow-x: hidden !important;
     &:not(.isMac),
     &:not(.isMac) * {
         @include scrollbar.custom;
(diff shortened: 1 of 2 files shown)


dae:
Thanks for the quick fix! Does this work or you @kleinerpirat?

winterhazel:
I did some more testing and the only unintended change that I found was the `StickyHeader` in the import CSV page hiding when there's too many fields and it reaches the bottom of the body, but that actually happens because the body is smaller than its content.

Screenshot_20230228_011832

Adding something like `body { min-height: 100vh; height: auto; }` to `import-csv-base.scss` so that the body expands vertically solves this problem.

kleinerpirat:
Should be good 👍🏻 I don't know if the `!important` is needed though.

PR #2410, opened 2023-03-02
Ensure fields were created before saving them

winterhazel:
This PR fixes an issue introduced in bf5bcd3f5.

Since now `fieldsCollapsed`, `richTextsHidden`, `plainTextsHidden` and `plainTextDefaults` are only set after `tick()` is resolved, but`notetypeId` is still set immediately, the editor would sometimes save empty values and restore them, messing the fields. This happened frequently when opening the browser by clicking a date on the stats page's calendar.

Checking `fieldsData` should ensure there's actually fields to save.

### DIFF
--- a/ts/editor/NoteEditor.svelte
+++ b/ts/editor/NoteEditor.svelte
@@ -81,7 +81,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     const sessionOptions: SessionOptions = {};
     export function saveSession(): void {
-        if (notetypeId) {
+        if (notetypeId && fieldsData.length > 0) {
             sessionOptions[notetypeId] = {
                 fieldsCollapsed,
                 fieldStates: {


dae:
I'm not very familiar with this part of the codebase, so I don't have much insight to offer I'm afraid. I presume the check fixes the initially-showing-editor case. Is there any chance a similar race condition exists when moving between different notes (possibly with a different number of fields)?

winterhazel:
I'm not entirely sure but, as far as I'm aware, there shouldn't be other race conditions like this when moving between notes.

___

(explanation, sorry this got quite long)

### Context

When the current note changes, the following functions in NoteEditor are executed (I'm only mentioning the relevant ones):
- Immediately: `saveSession(),` then `setNotetypeId()`
- After the previous commands are executed and pending changes are applied to the DOM, `tick()` resolves and the code inside `setCollapsed()` runs, then the code inside `setPlainTexts()` runs. This sets what we're saving.

This code is evaluated from `set_note()` in `editor.py`. When updating the current row, it runs `set_note()`. However, when updating the current row more than once too quickly (by holding the down arrow, for example), it runs `set_note()` more than once only for the currently selected note, but this wasn't causing any problems since the code in NoteEditor is executed in the proper order:
- `saveSession()` (saves the previous notetype), then `setNotetypeId(A)` (from the 1st evaluation)
- `tick()` resolves: `setCollapsed()`, then `setPlainTexts()` (from the 1st evaluation)
- `saveSession()` (saves notetype A), then `setNotetypeId(A)` (from the 2nd evaluation)
- `tick()` resolves: `setCollapsed()`, then `setPlainTexts()` (from the 2nd evaluation)

### The problem this PR addresses

When opening the browser through the calendar graph, it would also run `set_note()` twice for the currently note. However, since NoteEditor wasn't mounted yet, what sometimes happened was the following:
- `saveSession()` (doesn't save), then `setNotetypeId(X)` (from the 1st evaluation), then `saveSession()` (saves empty values) and `setNotetypeId(X)` (from the 2nd evaluation)
- `tick()` resolves: `setCollapsed()`, then `setPlainTexts()` (from the 1st), then `setCollapsed()` and `setPlainTexts()` (from the 2nd). These would restore the empty values.

This change prevents it from saving in this 2nd evaluation.

### Possible problems when moving between notes?

The only case I could think where we may arrive at some problems when moving between different notes would be something like:
- Load a note with notetype A
- Change to a note with notetype B: `saveSession()` (`notetypeId=A` and we have A's fields, so we save them), then `setNotetypeId(B)`
- Before `tick()` resolves and B's fields are loaded, change back to notetype A: `saveSession()` (`notetypeId=B` and we still have A's fields, but it saves them on notetype B), then `setNotetypeId(A)`

However, since `tick()` seems to always resolve before the next evaluation if it gets evaluated after NoteEditor is mounted, and this scenario requires NoteEditor to be mounted, I don't think it is possible. I've attempted to recreate this situation by manually changing the notes through code but, as expected, it didn't work.

I'm not completely sure about this though, and there could definitely be things I'm missing and some cases I haven't thought.

winterhazel:
I may have found another way to fix the open editors getting carried to different notetypes that doesn't rely on `tick` and `window.requestAnimationFrame`, which I think should make this code more stable. If it works as intended, I'll revert most of the changes in #2393 and here.

winterhazel:
I'm closing this PR in favor of #2422.

PR #2414, opened 2023-03-03
Fix black on grey minimalist buttons in KDE

danielzgtg:
This PR fixes an color contrast accessibility problem where the overview screen button text is illegible on KDE. Minimalist mode seems to be an effort to restore the theme before the redesign. The button text color was working fine both in the redesigned theme and the version before it, but not in this "minimalist mode"'s attempt to bring it back. This PR restores the button color properly like it was before.

This does not affect and is not reproducible by `./run` as that is using a bundled version of Qt instead of KDE's Qt. I reproduced and confirmed the fix using `tools/build` and installing to a venv created by `python3 -m venv venv --system-site-packages`. I am on Linux KDE 5.27.0 with the Breeze Dark theme. Anki is set to the Dark Theme and the Anki Style.

This is achieved by always doing `@include button.base` whether in minimalist mode or not, as the name ".base" suggests by convention. The unintended side effect, which I think turned out nicely, was that the mouse pressed styles are also added back.

# Before

image

# After

image

### DIFF
--- a/sass/buttons.scss
+++ b/sass/buttons.scss
@@ -35,8 +35,8 @@ button {
     padding: 8px 10px;
     margin: 0 4px;
 
+    @include button.base;
     .fancy & {
-        @include button.base;
         border-radius: var(--border-radius-large);
         @include elevation(1, $opacity-boost: -0.08);
         &:hover {


dae:
Any objections @kleinerpirat?

dae:
(This can be reproduced with ./tools/run-qt5.15)

kleinerpirat:
Looks good to me. I will keep your interpretation of `button.base` in mind for future PRs.

dae:
Thanks Daniel!

danielzgtg:
> (This can be reproduced with ./tools/run-qt5.15)

I just tried that and no qtwebengine components of Anki work. The processes shows up as `/home/home/PycharmProjects/anki/out/pyenv-qt5.15/lib/python3.9/site-packages/PyQt5/Qt5/libexec/QtWebEngineProcess --type=zygote --no-zygote-sandbox --application-name=Anki --webengine-schemes=qrc:sLV --lang=en-US` and others in `ps`, but it's all grey.

image

Probably another Qt version conflict with KDE. Also happens with other applications such as closed-source applications that insist on bundling their own Qt. It's also not loading my input method only installed to my system Qt. I'll just keep on using the version of Qt used with KDE like before.

dae:
https://docs.ankiweb.net/platform/linux/blank-window.html

danielzgtg:
Thank you. `QTWEBENGINE_CHROMIUM_FLAGS="–no-sandbox" ./tools/run-qt5.15` works. This qtwebengine components show up and I was able to reproduce the issue before this PR.

> I believe it’s fixed in Qt 6.2. They fixed it in 5.15 as well, but only provided an updated binary build to their commercial customers.

I think KDE might have patched this in their own version of Qt, and that might have been why the system version worked. KDE 6 is coming out soon so I'll be updating to Qt6 when that arrives in Ubuntu or a PPA. I wouldn't do so before that in case installing Qt6 breaks KDE.

Also, I think `./tools/run-qt5.15` tests a clean install of Qt5 instead of Qt5 + KDE. For one, platform plugins such as `/usr/lib/x86_64-linux-gnu/qt5/plugins/platformthemes/KDEPlasmaPlatformTheme.so` and `/usr/lib/x86_64-linux-gnu/qt5/plugins/platforminputcontexts/libfcitx5platforminputcontextplugin.so` (my input method that I need for regular use) aren't loaded. The style when set to "Native" also looks completely different without the platform theme plugin loaded.

image

image

But I do see the value of the additional testing environment that is a Qt5 clean install.

dae:
There are two advantages of testing with run/tools-run*:

- The Qt version exactly matches what's used for the binary bundles.
- Developers can reproduce issues regardless of their distro's Qt version, and typically across platforms as well. This allowed me to reproduce the issue you were suffering, without having to change my distro/desktop environment.

It's not intended as a daily driver for the reasons you mentioned, but useful when reporting bugs :-)

PR #2415, opened 2023-03-03
Fix black on grey deck options on KDE

danielzgtg:
This PR fixes another color contrast accessibility problem. The deck options input fields were illegible on KDE.

This time it's not specific to minimalist mode. See #2414 for the runtime OS environment used.

# Before

image

# After

image


kleinerpirat:
No objections. Thanks @winterhazel!

PR #2416, opened 2023-03-04
Workaround for CodeMirror starting too big when the browser is opened

winterhazel:
Since f3b2dbd31, when a field has the HTML editor shown by default, CodeMirror sometimes starts with the `min-height` attribute set to a value way higher it should be. It returns back to normal when the field is edited or focused.

Peek 2023-03-03 20-31

CodeMirror's docs recommend calling `cm.refresh()` after hiding/unhiding it to ensure it still looks as intended, but PlainTextInput's code already calls it when the Collapsible shows/collapses.
> If your code does something to change the size of the editor element (window resizes are already listened for), or unhides it, you should probably follow up by calling this method to ensure CodeMirror is still looking as intended. See also the autorefresh addon.

I've tried calling it in other places too, but when it works, a new problem appears. These problems are usually: (1) the height is adjusted but the width becomes smaller than the PlainTextInput and (2) the height is adjusted but you can see some frames with CodeMirror bigger than it should be before that.

The only way I found to solve this without causing more problems was changing `PlainTextInput.svelte`'s style. However, this is just a workaround and it's not fixing the root cause. CodeMirror still starts taller than it should be, which didn't happen before.

Does anyone know if there's a better way to solve this?

### DIFF
--- a/ts/editor/plain-text-input/PlainTextInput.svelte
+++ b/ts/editor/plain-text-input/PlainTextInput.svelte
@@ -168,6 +168,13 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
             padding-inline: 4px;
         }
 
+        :global(.CodeMirror-sizer) {
+            // Workaround for CodeMirror starting with a height bigger than it should  
+            // when the browser is opened on a notetype that has fields with the HTML
+            // editor selected by default
+            min-height: 0 !important;
+        }
+
         :global(.CodeMirror-lines) {
             padding: 8px 0;
         }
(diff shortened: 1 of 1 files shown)


hikaru-y:
> Since f3b2dbd,

I cannot reproduce the issue with  version 2.1.60, which includes the changes in f3b2dbd. Doing a `git bisect`, it appears to have been introduced in bf5bcd3f5.

winterhazel:
I'm able to reproduce this in 2.1.60, but it seems `bf5bcd3` also contributes to this issue.

winterhazel:
I did some more investigating and it seems that after f3b2dbd31 this only happens when reduced motion is disabled, but bf5bcd3f5 makes it happen whether it is enabled or disabled.

I'm sending a new PR that reverts bf5bcd3f5 and solves the open editors getting carried over through a more stable way, which partially fixes this.

hikaru-y:
Sorry, you are correct, @winterhazel.

I just tested with "Reduce motion" disabled, and was able to reproduce the problem with 2.1.60.

winterhazel:
Thanks for looking into this @hikaru-y. I'm closing this PR now.

PR #2419, opened 2023-03-05
Fix packaged audio tools path on Windows

abdnh:
There is no audio subfolder as of 2.1.55+. The wrong path was not causing any issues in Anki as far as I can tell. I just found about it while investigating an issue in an add-on that uses _packagedCmd.


dae:
Thanks!

PR #2420, opened 2023-03-05
Export static references from sound tags

RumovZ:
https://forums.ankiweb.net/t/sound-mp3-added-to-template-and-media-collection-not-exporting/28051

### DIFF
--- a/rslib/src/text.rs
+++ b/rslib/src/text.rs
@@ -168,13 +168,15 @@ lazy_static! {
     /// Strings, src and data attributes with a leading underscore.
     static ref UNDERSCORED_REFERENCES: Regex = Regex::new(
         r#"(?x)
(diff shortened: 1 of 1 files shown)


rslib/src/text.rs
```diff
@@ -312,31 +314,24 @@ pub fn replace_media_refs(
...
-        })
+        .map(extract_match)
         .collect()
 }
 
+/// Returns the first matching group as a str. This is intended for regexes
+/// where exactly one group matches, and will panic for matches without matching
+/// groups.
+fn extract_match(caps: Captures) -> &str {
+    caps.iter().skip(1).find_map(|g| g).unwrap().as_str()
```
dae, line 332: Nice

dae (commented):
Thanks Rumo!

PR #2421, opened 2023-03-05
Ensure state mutator runs after card is rendered

### LINKED ISSUE #2409: Don't run custom scheduling code until cards have finished rendering
https://forums.ankiweb.net/t/enhancement-block-review-js-until-the-cards-front-is-rendered/27984

RumovZ:
Closes #2409.

I wonder if there was a reason `_run_state_mutation_hook()` was called so early before. But then again, it usually ended up running after the rest anyway, so it should be fine.

I tested with a card template containing `<div id="deck">{{Deck}}</div>`, custom scheduling `console.log("The deck is " + document.getElementById("deck").innerHTML);` and the following patch:
```diff
diff --git a/ts/reviewer/index.ts b/ts/reviewer/index.ts
index c382f5182..3116a5239 100644
--- a/ts/reviewer/index.ts
+++ b/ts/reviewer/index.ts
@@ -121,6 +121,7 @@ export async function _updateQA(
     onupdate: Callback,
     onshown: Callback,
 ): Promise<void> {
+    await new Promise((resolve) => setTimeout(resolve, 1000));
     onUpdateHook.length = 0;
     onUpdateHook.push(onupdate);
```

This works now.

Heads-up, @L-M-Sherlock.

### DIFF
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import hashlib
 import pprint
 import time
 
@@ -83,7 +84,7 @@ class Card(DeprecatedNamesMixin):
         self.due = card.due
         self.ivl = card.interval
         self.factor = card.ease_factor
-        self.reps = card.reps
+        self.reps: int = card.reps
         self.lapses = card.lapses
         self.left = card.remaining_steps
         self.odue = card.original_due
@@ -215,6 +216,15 @@ class Card(DeprecatedNamesMixin):
         if not 0 <= flag <= 7:
             raise Exception("invalid flag")
         self.flags = (self.flags & ~0b111) | flag
+
+    def review_hash(self) -> str:
+        """Sha1 of this card's id and repetition count."""
+        hash = hashlib.sha1()
+        # CardId is an i64 in Rust
+        hash.update(self.id.to_bytes(8, "little", signed=True))
+        # reps are a u32 in Rust
+        hash.update(self.reps.to_bytes(4, "little"))
+        return hash.hexdigest()
 
     @deprecated(info="use card.render_output() directly")
     def css(self) -> str:
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -3,7 +3,6 @@
 
 from __future__ import annotations
 
-import hashlib
 import pprint
 import time
 
@@ -84,7 +83,7 @@ class Card(DeprecatedNamesMixin):
         self.due = card.due
         self.ivl = card.interval
         self.factor = card.ease_factor
-        self.reps: int = card.reps
+        self.reps = card.reps
         self.lapses = card.lapses
         self.left = card.remaining_steps
         self.odue = card.original_due
@@ -216,15 +215,6 @@ class Card(DeprecatedNamesMixin):
         if not 0 <= flag <= 7:
             raise Exception("invalid flag")
         self.flags = (self.flags & ~0b111) | flag
-
-    def review_hash(self) -> str:
-        """Sha1 of this card's id and repetition count."""
-        hash = hashlib.sha1()
-        # CardId is an i64 in Rust
-        hash.update(self.id.to_bytes(8, "little", signed=True))
-        # reps are a u32 in Rust
-        hash.update(self.reps.to_bytes(4, "little"))
-        return hash.hexdigest()
 
     @deprecated(info="use card.render_output() directly")
     def css(self) -> str:
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -83,6 +83,7 @@ class V3CardInfo:
 
     queued_cards: QueuedCards
     states: SchedulingStates
+    context: SchedulingContext
 
     @staticmethod
     def from_queue(queued_cards: QueuedCards) -> V3CardInfo:
@@ -90,8 +91,7 @@ class V3CardInfo:
         states = top_card.states
         states.current.custom_data = top_card.card.custom_data
         return V3CardInfo(
-            queued_cards=queued_cards,
-            states=states,
+            queued_cards=queued_cards, states=states, context=top_card.context
         )
 
     def top_card(self) -> QueuedCards.QueuedCard:
@@ -268,16 +268,12 @@ class Reviewer:
     def get_scheduling_states(self) -> SchedulingStates | None:
         if v3 := self._v3:
             return v3.states
-        else:
-            return None
+        return None
 
     def get_scheduling_context(self) -> SchedulingContext | None:
-        if not self.card:
-            return None
-        return SchedulingContext(
-            deck_name=self.mw.col.decks.name(self.card.current_deck_id()),
-            seed=self.card.review_hash(),
-        )
+        if v3 := self._v3:
+            return v3.context
+        return None
 
     def set_scheduling_states(self, key: str, states: SchedulingStates) -> None:
         if key != self._state_mutation_key:
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -125,6 +125,51 @@ class V3CardInfo:
             return CardAnswer.EASY
 
 
+class State(Enum):
+    WAITING = 0
+    READY = 1
+    DELAYED = 2
+
+
+class EventualCallback:
+    """Enables suspension of execution until a callback is ready.
+
+    After a run singal, the callback is executed as soon as a ready signal is received,
+    which may be immediately. A ready signal is also emitted automatically a certain
+    time after a wait signal.
+    """
+
+    _state = State.READY
+
+    def __init__(self, parent: QObject, callback: Callable[[], None]) -> None:
+        self._timer = QTimer(parent)
+        self._timer.setSingleShot(True)
+        qconnect(self._timer.timeout, self.ready)
+        self._callback = callback
+
+    def wait(self) -> None:
+        """Suspend any runs until the next ready signal or timeout.
+        Must only be called on the main thread.
+        """
+        self._state = State.WAITING
+        self._timer.start(500)
+
+    def ready(self) -> None:
+        """Mark the callback as ready, and run it immediately if a run is pending."""
+        if not self._timer.isActive() and self._state is not State.READY:
+            print("timed out waiting for ready signal")
+        if self._state is State.DELAYED:
+            self._callback()
+        self._state = State.READY
+
+    def run_when_ready(self) -> None:
+        """Run the callback as soon as it is ready, possibly immediately."""
+        if self._state is State.READY:
+            self._callback()
+        else:
+            self._state = State.DELAYED
+
+
 class Reviewer:
     def __init__(self, mw: AnkiQt) -> None:
         self.mw = mw
@@ -143,6 +188,7 @@ class Reviewer:
         self.bottom = BottomBar(mw, mw.bottomWeb)
         self._card_info = ReviewerCardInfo(self.mw)
         self._previous_card_info = PreviousReviewerCardInfo(self.mw)
+        self._ease_button_callback = EventualCallback(mw, self._showEaseButtons)
         hooks.card_did_leech.append(self.onLeech)
 
     def show(self) -> None:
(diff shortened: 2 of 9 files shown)


L-M-Sherlock:
Something wrong happened. The interval displayed above the `easy` button is 3 months.

image

But after pressing `easy` and checking the precious card info, the interval is 2.2 months.

image

It frequently happens if I press the rating buttons quickly. And I'm sure that the 3 months interval is given by Anki's built-in algorithm, and the 2.2 months interval is given by FSRS. I think it is due to the async process of custom scheduling, which is finished after `describe_next_states`. So the interval displayed is the old one.

https://github.com/ankitects/anki/blob/e5d5d1d4bdecfac326353d154c933e477c4e3eb8/qt/aqt/reviewer.py#L757-L759

L-M-Sherlock:
The intervals are displayed correctly if I click the `show answer` slowly.

image

dae:
> I wonder if there was a reason _run_state_mutation_hook() was called so early before.

Just oversight :-)

Thanks for testing @L-M-Sherlock. It sounds like we may need to delay revealing the answer buttons until we've confirmed the mutation hook has run. One thing to be careful of here is that we mustn't get stuck if the hook fails to run, due to eg. invalid JS (aside from infinite loops).

RumovZ:
Thanks, @L-M-Sherlock.

So the python code needs to make sure the main webview is done before calling the bottom webview. This will be tricky, at least without slowing things down a lot.

dae:
If images/mathjax take half a second to render, then updating the bottom bar would need to wait that long - not sure how laggy it would feel in practice. Maybe one way we could avoid this is to change the mutation hook to pass in the deck id (or maybe a context object with just a deck id at the moment)? We'd still need to defer rendering of the bottom section until the mutation hook ran, but it would no longer depend on the question being rendered.

At one point it will probably make sense to shift the FSRS code into Rust and avoid all the added complexity doing the calculations in JS, but it may make sense to hold off on that until you're fairly happy with the implementation and don't see it changing too frequently, as once it's in our Rust code users would need to wait for new Anki releases to get updates. Did you have any thoughts on that, @L-M-Sherlock?

L-M-Sherlock:
> Maybe one way we could avoid this is to change the mutation hook to pass in the deck id (or maybe a context object with just a deck id at the moment)?

In my option, the deck name is better than the deck id because the deck name is readable and user-friendly. And it is easy to apply the same parameters in the sub-decks via deck name.

In addition, the current implementation of FSRS uses the question content in two ways. 

First, FSRS uses it as the initial seed for Fuzz (which is `cid`+ `reps` in Anki's built-in method). 

```javascript
let seed = !customData.again.seed | !customData.hard.seed | !customData.good.seed | !customData.easy.seed ? document.getElementById("qa").innerText : customData.good.seed;
const generator = new Math.seedrandom(seed);
const fuzz_factor = generator();
seed = Math.round(fuzz_factor * 10000);
```

Second, FSRS extracts the deck name from it to choose parameters for specific decks and their sub-decks.

```javascript
for (let i = 0; i < deckParams.length; i++) {
  if (deck_name.startsWith(deckParams[i]["deckName"])) {
    params = deckParams[i];
    break;
  }
}
```

L-M-Sherlock:
> At one point it will probably make sense to shift the FSRS code into Rust

Will it allow users to save the parameters in deck options? It is a big project and has many benefits. If we can integrate the FSRS into Anki, it will be easier to use. 

> until you're fairly happy with the implementation and don't see it changing too frequently

The main function of FSRS has been stable for three months (since https://github.com/open-spaced-repetition/fsrs4anki/commit/e6b086b10d0aef5c7064ee3d71848c6c24866ae5). Recently, I am fixing some issues related to current PR. @MNastri helped me to decouple the configuration and main code. So it is a good time to start the integration. But I don't know how to contribute to Anki's scheduler module which is a little complicated.

L-M-Sherlock:
By the way, my friend implement FSRS in Rust here: https://github.com/open-spaced-repetition/rs-fsrs

dae:
Passing the deck name instead of deck id makes sense. Maybe the context should also pass in a seed derived from cid+reps?

Bit busy at the moment so will need to get back to you on the Rust integration in the future. Just a quick question for now - how much data is required per deck? If it's not a large amount, we could potentially store it in the deck and avoid having to check deck names.

L-M-Sherlock:
> Just a quick question for now - how much data is required per deck?

Here is an example data for a deck and its sub-decks:

```javascript
  {
    // Example 1: User's custom parameters for this deck and its sub-decks.
    // Need to add <div id=deck deck_name="{{Deck}}"></div> to your card's front template's first line.
    "deckName": "ALL::Learning::English::Reading",
    "w": [1.1475, 1.401, 5.1483, -1.4221, -1.2282, 0.035, 1.4668, -0.1286, 0.7539, 1.9671, -0.2307, 0.32, 0.9451],
    "requestRetention": 0.9,
    "maximumInterval": 36500,
    "easyBonus": 1.3,
    "hardInterval": 1.2,
  }
```

The `maximumInterval`, `easyBonus`, and `hardInterval` have been stored in deck configs. Only `w` and `requestRetention` are required. In addition, the deck config maybe needs to contain a checkbox to enable FSRS, because some users also want to use the built-in algorithm in some decks.

L-M-Sherlock:
> Maybe the context should also pass in a seed derived from cid+reps?

Using cid+reps as the seed is better than the content of the card's front because some cards contain a lot of HTML. The render would slow down the scheduler.

RumovZ:
It's hard to gauge the implications with other parts of the reviewer code (the joys of async the programming!), but I tested this with blocking, broken and working custom scheduling and it worked as expected in all scenarios.
It doesn't solve the potential issue with slow question rendering, but maybe we can wait until it proves to be, in fact, a problem?

dae:
I feel bad for saying this after you've put in that effort Rumo, but might it not perhaps be better to pass the context into the hook instead? The added complexity of delaying the status bar update is going to have to be mirrored on the two mobile clients, whereas if we pass the details in to the mutation hook, that's a simpler change. And having the context readily available makes things easier for custom scheduling authors, as they can read the values directly instead of trying to extract them from the DOM.

qt/aqt/reviewer.py
```diff
@@ -125,6 +125,51 @@ def rating_from_ease(ease: int) -> CardAnswer.Rating.V:
...
+        self._timer.setSingleShot(True)
+        qconnect(self._timer.timeout, self.ready)
+        self._callback = callback
+
+    def wait(self) -> None:
+        """Suspend any runs until the next ready signal or timeout.
+        Must only be called on the main thread.
+        """
+        self._state = State.WAITING
+        self._timer.start(500)
```
dae, line 155: Timeouts might not be the best way to approach this - I can see it leading to difficult bug reports in the future where users report that their custom study settings are failing to be applied on very rare occasions, and our future selves may not realise it's caused by some temporary lag in DOM rendering.

Because we're passing data between the webview and Python even if we pass the context in, we do still need to wait - but a way to do that which should be more reliable is by evaluating the custom js with evalWithCallback() - provided any errors are caught and swallowed, that should guarantee we get a callback that tells us when the bottom section should be rendered, and we don't need to worry about timeouts (infinite loops in JS notwithstanding - a timeout doesn't help there anyway, as the webview will effectively be hung)

dae:
Agreed on the two distinct issues, and sorry for the confusion about evalWithCallback() - I'd forgotten that mutateNextCardStates() is async. That does complicate things, but I suspect there are still ways we could handle this. For example:

```
anki.mutateNextCardStates(...).catch(() => false).then(() => bridgeCommand("statesMutated") 
```

Unless I've overlooked something, I think that should send a notice to the UI once the mutator has run, even if it happens to fail. If in addition to that, we set a can_show_answer=False flag in the UI when the question is revealed, and have the bridgeCommand set it to true. If showAnswer() then waited until it was set (by repeatedly calling itself 50ms later until the boolean changes), that should ensure the values are correct.

Another option would be to do the work in setSchedulingStates() call in mediasrv.py. It's not currently called in the error case though, so we'd either need to ensure it's always called, or add some other endpoint for that case.

WDYT? Other ideas welcome too.

RumovZ:
> Unless I've overlooked something, I think that should send a notice to the UI once the mutator has run, even if it happens to fail.

Only if it throws, not if there's a syntax error.

> If showAnswer() then waited until it was set (by repeatedly calling itself 50ms later until the boolean changes), that should ensure the values are correct.

I don't understand what problem recursion would solve here. Wouldn't a while loop with a `sleep(0.05)` accomplish the same thing? That would be simpler than the current tristate solution, but I don't think we should block the main thread.

dae:
> Only if it throws, not if there's a syntax error.

Thanks, I overlooked that. Yuck :-) We should be able to catch that case by checking the return value with evalWithCallback - if it's None, that implies a syntax error :

```
                f"anki.mutateNextCardStates('{self._state_mutation_key}', (states, customData) => {{ {js} }}).catch(() => false).then(() => console.log('mutated')); true",
```

> I don't understand what problem recursion would solve here. Wouldn't a while loop with a sleep(0.05) accomplish the same thing? 

Sorry, my wording was unclear there - when I said "calling itself", I meant via a mw.progress.timer() or similar. Sleeping in a background thread should work too, but might be a bit more complicated.

RumovZ:
> if it's None, that implies a syntax error

Wonderful! I had basically given up on a callback solution. 😃

> I meant via a mw.progress.timer() or similar.

I see, will do that instead if you prefer.

dae:
> I see, will do that instead if you prefer.

No preference, whatever you think is easier.

RumovZ:
Phew, a lot of back-and-forth, but this is a nice solution, so it was worth it in my opinion. 🙂
Is the context object fine like this? 
Regarding 4e07f9c, is there any reason against it? I added it for testing, but then reckoned the additional flexibility may also come in handy for users. For one thing, they can now choose to await question rendering should the context object not be enough.

pylib/anki/cards.py
```diff
@@ -216,6 +217,15 @@ def set_user_flag(self, flag: int) -> None:
...
         self.flags = (self.flags & ~0b111) | flag
 
+    def review_hash(self) -> str:
+        """Sha1 of this card's id and repetition count."""
+        hash = hashlib.sha1()
+        # CardId is an i64 in Rust
+        hash.update(self.id.to_bytes(8, "little", signed=True))
+        # reps are a u32 in Rust
+        hash.update(self.reps.to_bytes(4, "little"))
+        return hash.hexdigest()
```
dae, line 227: Could we move this into Rust, so that we don't need to reimplement it/keep it up to date across the various clients? Maybe something like col.backend.get_scheduling_context(c.id, c.reps)?

A full sha1 hex digest is also probably excessive, as FSRS tucks the seed into the custom card data, and this is going to add a fair bit to the storage of each row. How about a u64, like we use for fuzz_seed? JS won't be able to represent values exactly if it's larger than 52 bits, but since it's a random number, perhaps that doesn't matter?

ts/reviewer/answering.ts
```diff
@@ -53,11 +59,18 @@ function packCustomData(
...
+        states: Scheduler.SchedulingStates,
+        customData: CustomDataStates,
+        ctx: Scheduler.SchedulingContext,
+    ) => void,
 ): Promise<void> {
-    const states = await getSchedulingStates();
+    const [states, ctx] = await Promise.all([
+        getSchedulingStates(),
+        getSchedulingContext(),
+    ]);
```
dae, line 71: I'd like to try keep the number of separate requests to a minimum, as there's potentially a large round trip time (eg, if AnkiWeb adds support for this, there can be a RTT of 300-400ms in some parts of the world). Doing the requests in parallel does significantly improve the situation, but it's still an extra request that needs to be processed.

How about packing these two values into a container struct and returning that in a single method? We could also potentially return the context in get_queued_cards(), like we already do for the states.

dae (commented):
Thanks for your work on this Rumo. I've left a comment regarding the seed. Keeping the method async sounds good to me.

pylib/anki/cards.py
RumovZ: Makes sense.
However, it seemed just adding id and reps as for the fuzz seed is too prone to collisions so I've added a little tweak.

ts/reviewer/answering.ts
RumovZ: That's interesting. I've never thought about how it'd work on AnkiWeb.

dae:
Thanks Rumo, looks great. :-)

PR #2422, opened 2023-03-05
Improved fix for open editors getting carried to different notetypes

winterhazel:
This is an improved version of #2393 that doesn't cause issues like the one #2410 addresses and partially solves the issue in #2416.

It also seems b12476de9 solved the performance issue mentioned in #2393, so it's okay to revert the changes in it.


dae:
Thanks for your work on this Fabricio. Would appreciate your thoughts on this @hikaru-y

hikaru-y:
I just did a quick test, and it seems to work fine.

Incidentally, once this PR, which includes reverting bf5bcd3, is merged, it will be very easy to get a proper fix for the issue that #2416 is trying to work around, that the CodeMirror is not showing up in the correct size when opening the editor. I will hopefully push a PR for that later tonight.

dae:
Thank you both!

PR #2423, opened 2023-03-05
Fix backup creation for collections > 1 GiB

### LINKED ISSUE #2123: Backup creation fails when collection over 1GB on Windows
https://forums.ankiweb.net/t/the-process-cannot-access-the-file-because-another-process-has-locked-a-portion-of-the-file/23679

Caused by https://www.mail-archive.com/sqlite-users@mailinglists.sqlite.org/msg08081.html

Working around it will require switching our std::fs::read() call to a custom copy that ignores that region, or closing and reopening the collection. If ignoring, we presumably should just skip that section rather than padding with zeros, but will need to confirm.

RumovZ:
Closes #2123.
I confirmed the offsets and that the locked section must be padded with NULs by trial and error.

### DIFF
--- a/rslib/src/io.rs
+++ b/rslib/src/io.rs
@@ -2,6 +2,8 @@
 // License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
 use std::fs::File;
+use std::io::Read;
+use std::io::Seek;
 use std::path::Component;
 use std::path::Path;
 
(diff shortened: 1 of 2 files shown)


rslib/src/io.rs
```diff
@@ -63,6 +65,34 @@ pub(crate) fn read_file(path: impl AsRef<Path>) -> Result<Vec<u8>> {
...
+    let size = std::fs::metadata(&path)?.len() as usize;
+    if size < LOCKED_SECTION_END_BYTE {
+        return std::fs::read(path);
+    }
+
+    let mut file = File::open(&path)?;
+    let mut buf = vec![0; size];
+    file.read_exact(&mut buf[..LOCKED_SECTION_START_BYTE])?;
+    file.seek(std::io::SeekFrom::Current(LOCKED_SECTION_LEN_BYTES as i64))?;
+    file.read_exact(&mut buf[LOCKED_SECTION_END_BYTE..])?;
```
dae, line 91: The use of slices is a nice succinct way of implementing this. I've tested this on a Linux system and the sha256 of a file and its exported version matches. Thanks Rumo!

dae:
Sorry, I usually remove the co-authored lines when making minor edits as it makes it look like I contributed more than I did, but I forgot to this time. :-(

PR #2426, opened 2023-03-07
Fix CodeMirror not properly sized when opening editor

hikaru-y:
This is an alternative to #2416.


dae:
Thanks Hikaru! Unfortunately merging #2422 has caused a conflict.

dae:
Thank you! Appreciate all the work you and @winterhazel have put into this area.

PR #2427, opened 2023-03-08
Fix silent file save failures and bad default save path in Flatpak

jrpear:
Fix the issue I reported on the forum [here](https://forums.ankiweb.net/t/bug-desktop-export-fails-in-flatpak-when-file-extension-is-not-provided-by-user/27115).

I've tested that this fixes the failure to save and the weird default save path in Flatpak.

My patch checks for Snap too because that's what I saw Qt does, but I haven't tested that this issue exists or is fixed in snap. I don't even know if there's an Anki snap.

I appreciated how easy it was to build and lint Anki. Makes me want to keep working with it.

Let me know if there're any changes I should make!

### DIFF
--- a/qt/aqt/utils.py
+++ b/qt/aqt/utils.py
@@ -638,7 +638,7 @@ def getFile(
     return ret[0] if ret else None
 
 
-def checkNeedPortalSupport():
+def running_in_sandbox():
     """Check whether running in Flatpak or Snap. When in such a sandbox, Qt
     will not report the true location of user-chosen files, but instead a
     temporary location from which the sandboxing software will copy the file to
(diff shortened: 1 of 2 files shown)


qt/aqt/utils.py
```diff
@@ -638,6 +638,23 @@ def accept() -> None:
     return ret[0] if ret else None
 
 
+def checkNeedPortalSupport():
```
dae, line 641: Minor nit: we try to use PEP8 snake_case in new code. It's an easy mistake to make when the surrounding functions are still camelCase; I've fallen into it a number of times myself.

Suggestion: how about running_in_sandbox() instead, or something similar?

dae (commented):
Thanks for contributing Jack! I'm glad to hear the build system (which is fairly new) worked well for you.

qt/aqt/utils.py
jrpear: Yep, I just copied the format from around where I was working 😅

Yeah, that rename seem great.

dae:
Thanks Jack!

PR #2428, opened 2023-03-08
Make search in stats page trigger onChange

yellowjello:
### Context/Issue:
AnkiDroid uses Anki's stats/graph page when the new backend is enabled, but the "on enter key up" trigger for search does not work on software keyboards on Android, so search essentially does not work unless a hardware keyboard is connected.

The reason is because many software keyboards on Android do not send out key codes by design.

### Approach:
This PR addresses this issue by triggering the search update with the onchange event. The onchange event fires if enter key is pressed or if the field loses focus. Pressing the enter key on Android appears to trigger this event as well.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -114,6 +114,7 @@ Kaben Nanlohy <kaben.nanlohy@gmail.com>
 Tobias Predel <tobias.predel@gmail.com>
 Daniel Tang <danielzgtg.opensource@gmail.com>
 Jack Pearson <github.com/jrpear>
+yellowjello <github.com/yellowjello>
 
 ********************
 
--- a/ts/graphs/RangeBox.svelte
+++ b/ts/graphs/RangeBox.svelte
@@ -55,11 +55,8 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         }
     }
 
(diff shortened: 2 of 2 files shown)


dae:
This seems like a reasonable solution, and I am happy to merge it, but a question first: is there no way to detect that enter is pressed on the software keyboard?

yellowjello:
> is there no way to detect that enter is pressed on the software keyboard?

It looks like on Android, key codes sent by the keyboard used and the default keyboard by design does not send out keycodes:
https://bugs.chromium.org/p/chromium/issues/detail?id=118639#c242

This means that we cannot rely on keyboard events at all for doing detection.

I guess a potential alternative to get the enter key working would be to wrap the input in a form. Pressing the enter key in the textbox will submit the form and then we can use the form submit instead of an enter press to trigger the search.

Let me know if this alternative sounds better.

dae:
Please try this on your phone: https://svelte.dev/repl/bf21b14cb29e41f28efc802b56e7f152?version=3.23.1

If you type some text into the input on the software keyboard and hit enter, does the change count only update when hitting enter? If so, perhaps we can use on:change instead of looking at keyboard codes. It should also increment the number if you tap away from the input.

dae:
Thank you!

PR #2432, opened 2023-03-10
Revert bury restriction

### LINKED ISSUE #2397: Revisit independent review card burying
There have been a number of users posting how they miss independent review card burying (eg https://forums.ankiweb.net/t/option-to-bury-sibling-review-cards-without-burying-sibling-new-cards-on-ankimobile/27495), so we may need to revisit the decision in https://github.com/ankitects/anki/issues/2352, and see if we can't find some way of supporting this use case without making things much more complicated.

RumovZ:
Closes #2397.

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -104,22 +104,22 @@ deck-config-leech-action-tooltip =
 ## Burying section
 
 deck-config-bury-title = Burying
-deck-config-bury-siblings = Bury siblings
-deck-config-do-not-bury = Do not bury siblings
-deck-config-bury-if-new = Bury if new
-deck-config-bury-if-new-or-review = Bury if new or review
-deck-config-bury-if-new-review-or-interday = Bury if new, review, or interday learning
-deck-config-bury-tooltip =
-    Siblings are other cards from the same note (eg forward/reverse cards, or
-    other cloze deletions from the same text).
-    
-    When this option is off, multiple cards from the same note may be seen on the same
-    day. When enabled, Anki will automatically *bury* siblings, hiding them until the next
-    day. This option allows you to choose which kinds of cards may be buried when you answer
-    one of their siblings.
-    
-    When using the V3 scheduler, interday learning cards can also be buried. Interday
-    learning cards are cards with a current learning step of one or more days.
+deck-config-bury-new-siblings = Bury new siblings
+deck-config-bury-review-siblings = Bury review siblings
+deck-config-bury-interday-learning-siblings = Bury interday learning siblings
+deck-config-bury-new-tooltip =
+    Whether other `new` cards of the same note (eg reverse cards, adjacent cloze deletions)
+    will be delayed until the next day.
+deck-config-bury-review-tooltip = Whether other `review` cards of the same note will be delayed until the next day.
+deck-config-bury-interday-learning-tooltip =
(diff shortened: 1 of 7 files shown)


ftl/core/deck-config.ftl
```diff
@@ -104,22 +104,22 @@ deck-config-leech-action-tooltip =
...
+deck-config-bury-review-tooltip = Whether other `review` cards of the same note will be delayed until the next day.
+deck-config-bury-interday-learning-tooltip =
+    Whether other `learning` cards of the same note with intervals > 1 day
+    will be delayed until the next day.
+deck-config-bury-priority-tooltip =
+    Which one of multiple siblings will be shown if burying is enabled, depends on the *gather*
+    order. The card gathered first will be shown, its siblings will be buried.
+    Since learning cards are gathered first, then review cards, and finally new cards,
+    a review card can never be buried by a new sibling. So if you enable burying new cards,
+    but not review cards, you may see a new and a review sibling in the same session.
```
RumovZ, line 122: A bit lengthy perhaps. Please tweak it to your liking.

rslib/src/scheduler/bury_and_suspend.rs
```diff
@@ -138,15 +138,26 @@ impl Collection {
...
+            .all_siblings_for_bury(card.id, nid, bury_mode)?;
         self.bury_or_suspend_cards_inner(cards, BuryOrSuspendMode::BurySched)
     }
 }
 
+impl BuryMode {
+    /// Disables burying for queues gathered before `queue`.
+    fn exclude_earlier_gathered_queues(&mut self, queue: CardQueue) {
+        self.bury_interday_learning &= matches!(queue, CardQueue::DayLearn);
+        self.bury_reviews &= matches!(queue, CardQueue::DayLearn | CardQueue::Review);
```
dae, line 157: Is this perhaps a bit more restrictive that it needs to be? I think we may only need to enforce this when the user has some burying options enabled but not others. For example, if burying of both new cards and reviews is enabled, it's fine for new cards to bury reviews, since if there's a sibling review card that will take precedence. The issue comes when only some of the burying options are enabled, and eg. a new card ends up burying a review card that we've already gathered.

dae: Sorry, my thinking was wrong. Of course, relaxing this restriction if new card burying is enabled would buy us nothing, since the fetch ordering would always ensure that the review sibling takes precedence. Please ignore me. :-)

dae: Hmm, thinking a bit more about this, should CardQueue::Learn also be in each of these lines? Intraday learning cards are fetched before anything else, and they should bury siblings (except if they're also intraday learning)

ftl/core/deck-config.ftl
dae: Does the new wording work for you?

rslib/src/scheduler/bury_and_suspend.rs
RumovZ: Of course, you're right. I missed that one. It also seems like `PreviewRepeat` cards should be taken into account.

ftl/core/deck-config.ftl
dae: I'm going to assume that's a yes. Thanks for your work on this Rumo :-)

PR #2433, opened 2023-03-11
Fallback to Wayland if X11 is not available

ingemarberg:
When Anki is running in a Wayland-only session and `ANKI_WAYLAND` is not set, Anki will crash on startup. It may be more helpful to print a warning and then fallback to Wayland instead. This might not be a very common case, but it can occur when Anki is running in a Flatpak sandbox. In this case `WAYLAND_DISPLAY` would be set, while `DISPLAY` is not. This pull request should fix that issue.

I had to temporarily change `getattr(sys, "frozen", False)` to `getattr(sys, "frozen", True)` in order to test this, but I assume that is to be expected when running a non-packaged version of Anki.

Please let me know if there is a better way to do this!

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -115,6 +115,7 @@ Tobias Predel <tobias.predel@gmail.com>
 Daniel Tang <danielzgtg.opensource@gmail.com>
 Jack Pearson <github.com/jrpear>
 yellowjello <github.com/yellowjello>
+Ingemar Berg <github.com/ingemarberg>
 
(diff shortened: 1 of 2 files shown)


dae:
Thanks Ingemar!

ingemarberg:
It feels great to be able to contribute something, however minor, after using Anki for so long. Even though I'm not a programmer, I found it easy to get started thanks to the build system and well-written documentation. Thanks for all your hard work, Damien!

PR #2435, opened 2023-03-12
Improve debug console

RumovZ:
This is still a bit rough around the edges, but I think the ability to easily persist scripts can already be a huge time saver.

### DIFF
--- a/qt/aqt/debug_console.py
+++ b/qt/aqt/debug_console.py
@@ -1,0 +1,269 @@
+# Copyright: Ankitects Pty Ltd and contributors
+# License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+
+from __future__ import annotations
+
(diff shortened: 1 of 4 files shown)


dae:
Looks like a nice improvement - thanks Rumo!

dae:
Currently the debug console opens with focus on the top dropdown - what do you think about changing that to the text box so you can type immediately?

dae:
Nevermind - I see it already does it on Mac, so presumably this is a Linux-specific issue. Will investigate when I have a chance.

RumovZ:
No, it seems to be on Windows, too. Makes sense to ensure the text edit is focused by default again.

dae:
Should be fixed in https://github.com/ankitects/anki/commit/77a45365cb64284dd4a8cd2535d6cbdc2676472c

PR #2436, opened 2023-03-12
Check URLs in TS code

### LINKED ISSUE #2380: Literal links in ts are unchecked
There is a test to check for broken URLs, but it can only cover members of the HelpPage enum. For that reason, the new literal links in the ts code should be replaced.
To efficiently fetch URLs from the backend en masse, as is required for the deck options screen, I guess we also need a new RPC that returns (a map of) all available pages from the backend?

RumovZ:
Closes #2380.
No broken links detected. 🙂


rslib/linkchecker/tests/links.rs
```diff
@@ -1,94 +1,137 @@
...
-                        page.to_link_suffix(),
-                    ))
+impl From<&'static str> for CheckableUrl {
+    fn from(value: &'static str) -> Self {
+        Self::String(value)
+    }
+}
+
+fn ts_help_pages() -> impl Iterator<Item = &'static str> {
+    lazy_static! {
```
dae, line 67: No need to change this, but just FYI: I've recently started using once_cell's Lazy instead of lazy_static, as it seems to be the preferred solution these days, and may one day end up in the Rust stdlib.

dae (commented):
Looks good - thanks Rumo!

PR #2437, opened 2023-03-13
Check for division by zero when calculating browser aspect ratio

Ben-Kerman:
Not doing so caused a (non-crash) error under X11 with i3 when opening the browser while the main window is in fullscreen mode


dae:
Happy to merge this once tests pass.

Ben-Kerman:
I forgot to add myself to the contributor list, sorry about that

dae:
Thanks Ben!

PR #2441, opened 2023-03-15
Fix current tag text not committed when changing note via shortcut key

### LINKED ISSUE #1924: Changing note via shortcut key does not commit current tag text
https://forums.ankiweb.net/t/cursor-dont-remain-in-tag-and-tag-not-registered/20937

It works when clicking on a different row instead

@Arthur-Milchior a reminder you can post bug reports here if you're confident it's a bug

hikaru-y:
Fixes #1924

The issue also happens when closing editor with <kbd>Esc</kbd>.


dae:
Thanks Hikaru, much appreciated.

hikaru-y:
I've confirmed that this has also fixed the issue reported in the forum today:
https://forums.ankiweb.net/t/bug-navigate-to-the-next-card-with-keyboard-shortcut-also-removes-its-tag/28507

PR #2442, opened 2023-03-15
Fix tags popover becoming separated from the tags field

winterhazel:
This fixes the issue in https://forums.ankiweb.net/t/anki-2-1-61-beta/28361/2.

`"top-start"` doesn't seem to be a valid placement anymore.


dae:
What a nice, small fix. Thanks Fabricio :-)

PR #2445, opened 2023-03-17
Fix invalid ids on db check

RumovZ:
Since fixing ids requires a full sync and is not undoable (`todo!()`) should we do it outside the actual db check transaction? We could report the number of invalid ids, create a backup and ask for permission, after the db check is finished.
Or should it be undoable after all? I realise I should have asked sooner now...


dae:
Various other checks already proceed without confirmation, are not undoable, and will force a full sync if required, so that should be fine here too. We make a backup before the check runs, so users can revert to the backup if required.

RumovZ:
Should've seen that myself. Thanks for pointing it out. 🙂

rslib/src/storage/dbcheck/invalid_ids_create.sql
```diff
@@ -0,0 +1,27 @@
...
+        SELECT max_id
+        FROM max_existing_valid_id
+      ) + 1
+    END AS id
+)
+SELECT id,
+  (
+    SELECT id
+    FROM first_new_id
+  ) + row_number() OVER (
```
dae, line 23: Neat, I wasn't aware of row_number()

dae (commented):
Thanks Rumo, all looks good. I was going to suggest that maybe those storage routines should take TimestampMillis, but then realised we're dealing with CardId/NoteId/etc as well.

dae:
As an optional follow-up to this, perhaps we should change the 'contact the deck author' error users get when importing to instead say something like "This deck contains timestamps in the future. Please let the deck author know. You can still import it, but we'll need to do a database check and force a one-way sync afterwards. Cancel/Import"? And then if the user proceeds, we run the DB check automatically in the Rust code. I imagine it could pass a bool in to the import that if unset will throw the error (which we catch + prompt), and if set allows importing to proceed+runs the DB check. WDYT?

RumovZ:
We could also run the `invalid_ids` routines directly on the opened apkg collection. Then we'd only need one pass and no db check.

dae:
If that's easy enough, sounds great.

PR #2446, opened 2023-03-17
Don't download nodejs if NODE_BINARY is set, add YARN_BINARY

euank:
Some build environments, such as nixpkgs, restrict network access and
thus would prefer to not download anything at all. Setting PROTOC_BINARY
and friends makes the build system not download stuff, but the same was not true for nodejs.
This PR updates it so nodejs behaves similarly in that regard.

While I was at it, I also added YARN_BINARY which does what you'd expect.

For a little more context, these patches are what let this build system work over here: https://github.com/NixOS/nixpkgs/pull/221229

I, unfortunately, can't actually verify everything works without the \*_BINARY env vars set as easily because the binaries it downloads don't run on nixos, so the build of course fails.

<sup>(thanks for the excellent project btw <3)</sup>

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -117,6 +117,7 @@ Jack Pearson <github.com/jrpear>
 yellowjello <github.com/yellowjello>
 Ingemar Berg <github.com/ingemarberg>
 Ben Kerman <ben@kermanic.org>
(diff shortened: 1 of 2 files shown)


dae:
Looks good - thanks!

PR #2447, opened 2023-03-17
Fix new limit being decremented unduly

RumovZ:
😬 
https://forums.ankiweb.net/t/anki-2-1-61-beta/28361/6


dae:
Thanks for the quick fix :-)

PR #2449, opened 2023-03-19
Fix invalid ids on import

RumovZ:
See #2445.


dae:
Thanks Rumo, that's a nice simple solution. I was worried that this would break duplicate detection when ids get changed, but since we rely on guid+ord, that still appears to work fine, which is great.

PR #2450, opened 2023-03-19
Add Backend routines for cross-platform user theme support [Help needed]

kleinerpirat:
**Please note:** This is a super rough sketch that doesn't build as of right now. I do not have the time to fix all issues with this currently, but thought I'd share the WIP anyway.

---

I've been working on a theme editor for a while now:

It supports editing of user backgrounds and the color palette with a live preview. Lots of changes across the app made the diff very large, so I decided to reach out for help with the backend problems in a separate PR so I don't lose overview.

## I need help

I have no clue of SQL and very limited understanding of Rust. Since the frontend interaction will be quite similar to the current deck options implementation, I figured a good way to start is to use that part of the backend as a reference. Apart from variables, the only difference in the backend interaction is that we won't query the themes by the current deck id, but the current mode (light/dark). So most changes are 1:1 copies of the `deckconfig` implementation - exceptions being `themes.proto` and `themes/default.rs`. Since it's a new feature, I also assumed `Schema11` files and functions aren't needed.

Unfortunately, copying, renaming and stripping out unrequired functions left a mess of errors I'm afraid I cannot solve on my own in a reasonable time frame :sweat_smile:

### Data Structure

- A `Theme` stores two sets of vars: `factors` and `colors`
  - `factors`: positive floating-point numbers
    - Props like `blur-radius` or `transition-duration` will be generated in Rust from these factors
  - `colors`: a set of base colors
    - Helper colors like `button-gradient_start` or `scrollbar_hover` will be derived from the base colors with the Rust crate `palette`.

One could define a color as

```protobuf
message Color {
  RgbaColor light = 1;
  RgbaColor dark = 2;
}
```
... but users will probably want to mix and match themes of different authors for light and dark, so I decided against that. A `Theme` is *not a light/dark combination*, just one of them.

#### Build / Development
- Derive full set of variables from default factors and base colors
- Create `.py` and `.scss` files that contain the names of all:
  - props (generated from factors)
  - base colors + derived helper colors
- These files should enable code completion in Python and SCSS. They do not need to have actual values (those are fetched from the backend at runtime). In python they act as keys and in SCSS they resolve to CSS variables, e.g. `$some-color: var(--some-color);`

#### Runtime
For simplicity's sake, I'd fetch the variables in Python and include them via `webview.py`.

## Sharing Palettes

Themes could be shared as `.ankitheme` JSON files. We could offer an option to create a single file for a light+dark combination.

---

@dae What do you think of this concept?


dae:
I won't have much time for this until early April I'm afraid, so just some brief thoughts for now:

- I don't think we should use a separate table. Schema changes are expensive - they require upgrade/downgrade code, and changes to the sync protocol. I think we should pack the themes into the config table instead, which is intended to be used as a general key/value store, and will sync only changed keys in the future. We could use keys like themeLight[name]. We're not set up to store binary values in config at the moment, so the most compact way to store the values would probably be to base64 the binary output from the protobuf. JSON would also work, but is probably going to be quite a bit larger to store.
- I don't think you can do things like change the button color when using the native theme on MacOS - how do you intend to handle that?
- For theme sharing we can get JSON for free by adding a line to the protobuf building code.

kleinerpirat:
> Schema changes are expensive - they require upgrade/downgrade code, and changes to the sync protocol.

Thanks for the info, I wasn't aware of that. Dropped out of CS before it got real with database systems ^^

> most compact way to store the values would probably be to base64 the binary output from the protobuf.

Whatever you think is best. I might play around with this a bit, but will probably still require your help in April :-)

> I don't think you can do things like change the button color when using the native theme on MacOS - how do you intend to handle that?

I thought about this and figured we'll have to leave it to the user to decide whether user theme support is worth switching to our custom Qt style.

proto/anki/themes.proto
```diff
@@ -0,0 +1,148 @@
...
+
+/* User-modifiable base colors
+from which helper colors (gradient, hover, etc.) are derived */
+message Colors {
+  // background
+  Color canvas = 1;
+  Color canvas_secondary = 2;
+  Color canvas_code = 3;
+  reserved 4;
+  reserved 5;
```
dae, line 64: I recommend cutting out these reserved lines - it doesn't matter if the numbers are out of order when we add new fields in the future. The main reason to reserve numbers is when we anticipate repeated fields to be required, as fields in 1-15 take 1 byte per entry, and 16+ take 2 bytes per entry, so for large lists it's more space efficient to place fields at the start.

rslib/src/pb.rs
```diff
@@ -29,3 +29,4 @@ protobuf!(search, "search");
 protobuf!(stats, "stats");
 protobuf!(sync, "sync");
 protobuf!(tags, "tags");
+protobuf!(themes, "themes");
```
dae, line 32: I'd probably just put this in config.proto, as that's where the rest of the preferences stuff lies.

rslib/Cargo.toml
```diff
@@ -109,6 +109,7 @@ utime = "0.3.1"
 workspace-hack = { version = "0.1", path = "../tools/workspace-hack" }
 zip = { version = "0.6.3", default-features = false, features = ["deflate", "time"] }
 zstd = { version = "0.12.2", features = ["zstdmt"] }
+palette = "0.6.1"
```
dae, line 112: Not currently used?

dae (commented):
I am available now if you get stuck.

rslib/Cargo.toml
kleinerpirat: The plan is to use it to generate colors that can be derived, e.g. `BUTTON_GRADIENT_START` and `...-END` from the base color `BUTTON` by making it slightly lighter/darker. The crate seems like a good fit for that purpose.

Generating helper colors from a set of base colors allows us to keep the amount of colors users need to tweak for new themes low, while also ensuring basics like hover feedback work at all times.

dae:

PR #2456, opened 2023-03-25
fix stats calendar incorrect due to daylight savings time

### LINKED ISSUE #1025: Due counts on calendar graph can change after midnight
https://forums.ankiweb.net/t/stats-bug-v2-0-67/4248

And perhaps the same issue:

https://forums.ankiweb.net/t/bug-extra-day-inserted-when-learning-over-midnight/7521

kieranlblack:
Previously, when computing counts for the calendar in the stats menu, it was assumed that days had 86,400 seconds. However, this assumption does not hold true on the day when daylight savings occurs. As such, there is an hour period where all reviews occurring after daylight savings are computed as belonging to the wrong day.

This screenshot was taken between 00:00 - 1:00 on Saturday. At that time no reviews had been completed for Saturday yet the shift by one day is visible in the graph with March 12th (DST went into effect) being recorded as having no reviews.
image

The issue on the frontend is simpler to fix since the included library, `D3`, has a built-in time library which can account for DST. It  is trickier on the backend half of things.

I am currently working on fixing this issue.

Fixes #1025

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -118,6 +118,7 @@ yellowjello <github.com/yellowjello>
 Ingemar Berg <github.com/ingemarberg>
 Ben Kerman <ben@kermanic.org>
 Euan Kemp <euank@euank.com>
+Kieran Black <kieranlblack@gmail.com>
 
 ********************
 
--- a/proto/anki/stats.proto
+++ b/proto/anki/stats.proto
@@ -141,6 +141,7 @@ message GraphsResponse {
   FutureDue future_due = 7;
   Added added = 8;
   ReviewCountsAndTimes reviews = 9;
+  uint32 rollover_hour = 10;
 }
 
 message GraphPreferences {
--- a/qt/aqt/about.py
+++ b/qt/aqt/about.py
@@ -230,6 +230,7 @@ def show(mw: aqt.AnkiQt) -> QDialog:
             "hafatsat anki",
             "Carlos Duarte",
             "Edgar Benavent Català",
+            "Kieran Black",
         )
     )
 
--- a/rslib/src/stats/graphs/mod.rs
+++ b/rslib/src/stats/graphs/mod.rs
@@ -71,6 +71,7 @@ impl Collection {
             hours: Some(ctx.hours()),
             buttons: Some(ctx.buttons()),
             card_counts: Some(ctx.card_counts()),
+            rollover_hour: self.rollover_for_current_scheduler()? as u32,
         };
         Ok(resp)
     }
--- a/ts/graphs/calendar.ts
+++ b/ts/graphs/calendar.ts
@@ -85,7 +85,7 @@ export function renderCalendar(
     const dayMap: Map<number, DayDatum> = new Map();
     let maxCount = 0;
     for (const [day, count] of sourceData.reviewCount.entries()) {
-        const date = new Date(now.getTime() + day * 86400 * 1000);
+        const date = timeDay.offset(now, day);
         if (count > maxCount) {
             maxCount = count;
         }
@@ -110,7 +110,7 @@ export function renderCalendar(
     const oneYearAgoFromNow = new Date(now);
     oneYearAgoFromNow.setFullYear(now.getFullYear() - 1);
     for (let i = 0; i < 365; i++) {
-        const date = new Date(startDate.getTime() + i * 86400 * 1000);
+        const date = timeDay.offset(startDate, i);
         if (date > now) {
             // don't fill out future dates
             continue;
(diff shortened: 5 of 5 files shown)


kieranlblack:
I wasn't sure what the protocol was for adding one's name to `about.py` so please let me know if I should remove my name from it!

kieranlblack:
There are actually a lot of places throughout the codebase where a day is assumed to have 86,400 seconds instead of using time libraries to compute day offsets. A simple search for `86.?400` reveals 93 results across 32 files.

This PR could be expanded in scope to perform a more comprehensive fix.

kieranlblack:
With a quick poc it looks like by working with the rollover hour in the calendar code on the frontend we can account for everything and correctly display the data. Will look at the best way to get access to that tomorrow.

dae:
Thanks for looking into this Kieran! Anyone who's contributed something is welcome in the about screen.

Does the change you made to the UI fix the graphs issue, or are you saying more is required? I'd recommend we don't try and fix every instance of seconds arithmetic in the backend unless we have clearly identified an issue it's causing, as the more that gets changed, the higher the chance of an accidental regression.

If you do need to perform these calculations on the backend as well, we can presumably accomplish that with the chrono library that we use.

kieranlblack:
The change made to the UI fixes the issue with DST. In more detail: the issue before was when iterating backwards from the current date and crossing the boundary of DST, adding 86,400 seconds would actually push a little bit past an extra day boundary since the daylight savings day was shorter by a calendar hour (and so subtracting a full 86,400 seconds placed [0, 1) hours into the prior calendar day). This trap of going a little bit over was always present, it's just that after 1 am, moving back by an extra hour didn't push us over into the previous day so it wasn't visible.

There is more to the issue than just the DST component though that requires some more work. It is already tracked in the github issue, but it is the one reported here:

https://forums.ankiweb.net/t/bug-extra-day-inserted-when-learning-over-midnight/7521

Since Anki days don't necessarily roll over at midnight, mapping an Anki day into a calendar day needs to have a linear shift applied. This shift is not currently accounted for by the frontend, so when served data for a day offset 0 from the current Anki day, the frontend code assumes it must be 0 offset from this calendar day. As such, between midnight and the user configured start of new day, all data in the stats calendar is shown as being shifted one calendar day forward from how it should actually be displayed.

It makes sense to me to keep the backend as still producing the graph data wrt Anki days especially since not just the calendar consumes that data. Consequently, the frontend needs access to the user configured rollover hour. I'm not familiar with the codebase but two solutions which jump out to me are we could either insert it to be returned from this chain:

https://github.com/ankitects/anki/blob/bbaa6c24ecddb5bf64fece08fe7f2d08ac3d3a69/rslib/src/stats/graphs/mod.rs#L40

or alternatively expose preference information to the typescript part of things maybe in a fashion similar to how the `GraphPreferences` information is exposed or more low down like directly in `CalendarGraph.svelte`.

**Does one of these approaches seem better to you than the other? / maybe something else entirely?** I'm partial to exposing the preference information to the typescript end of things but playing around with it for a little bit shows it to not be super simple to do using the existing `PreferenceStore` (but definitely doable) whereas inserting it into the `GraphsResponse` would be quicker and require a lot fewer moving parts. Then if someone in the future needed more rich preferences support, they could come along and implement something like the first option.

Also, I agree this PR isn't the place for a change like trying to modify every place in the code the 86,400 assumption is made. From what I've seen, the other usages throughout the code will only impact people who use the app on the day which DST goes into effect / goes out of effect.

dae:
Let's keep it simple until there's a compelling case not to - GraphsResponse is what I'd suggest. It used to be included; if you locate the two lines that reference next_day_at_secs in https://github.com/ankitects/anki/commit/37151213cd9d431f449ba4b3bc4c0329a1d9af78#diff-fa512376ac51a0c1654e92f17ec912ffbd9ad76999746d814e76da7df2981dbd it should be trivial to add back.

kieranlblack:
`self.get_v2_rollover` or `self.rollover_for_current_scheduler`? The latter seems more appropriate.

https://github.com/ankitects/anki/blob/0e0436f8508825616e997353ad8afc65a29476e3/rslib/src/scheduler/mod.rs#L117-L122

dae:
It's not super important, since users are required to update to V2 before they're able to study, so a user would only hit this if they viewed graphs without studying. But you're right that the latter is the better choice.

Is this about ready for review/merge? I'd like to put out a release candidate in the next few days, and it would be good to slip this in if we can.

kieranlblack:
Okay, thank you for the context.

Yes, this is ready for review.

dae:
Looks good - thank you for all your work on this Kieran!

bm-w:
Possibly related: I caught a missing day on my calendar view. At first glance it seems like it’s a Saturday that’s missing (towards the bottom right), but in reality that whole week is shifted “up” by 1 day. That is, the day in the top row (labeled “S” for Sunday) is actually a Monday, and below that are Tuesday through Saturday, the latter in the second row from the bottom (labeled “F” for Friday).

As it happens, the Sunday that’s missing there was 3 September 2023, which happens to be the start day of DST in my current timezone (CLST, Chilean Summer Time, GMT -3:00).

Version ⁨2.1.66 (70506aeb)⁩ on macOS 14.0 (also seen on 13.6).

image

PR #2457, opened 2023-03-25
Prevent the sticky from hiding in the stats page

winterhazel:
Same issue as in https://github.com/ankitects/anki/pull/2406#issuecomment-1447564639, but this one was already present before #2406.


dae:
Thanks for figuring this out Fabricio!

PR #2461, opened 2023-03-28
Add a deck selector to Stats

### LINKED ISSUE #2434: Adding a Deck Selector to Statistics Page
## Issue:
Right now, it is pretty awkward to switch the deck being used for the Statistics page.

From what I can tell, there are currently two ways to switch the deck, both of which are awkward:
1. Close the statistics page, click on the deck in the main screen, then open stats again.
2. Typing in the deck name as part of a search query. This is especially awkward if the deck name is very long.

## Proposed Solution:
We can add in a deck selector to the statistics page to make it quick and easy to switch decks.

To kickstart the discussion, I have made rough mock-ups for some UI layout options we can potentially chose.

### Option 1: Add new option in existing UI
Screenshot 2023-03-10 at 5 28 39 PM

With this option, we show the deck selector when the "deck" option is selected and hide the search bar. The search bar comes back when the new "custom" option is selected.

### Option 2: Simplified UI
Screenshot 2023-03-10 at 5 31 50 PM

With this option, we just have the deck selector and the search bar. The deck selector will have a "Whole Collection" option for seeing stats and searching across all decks.

The search bar will be used to add additional filter options for the selected deck/collection.

yellowjello:
### What this PR does
Add a deck selector to the UI of the stats screen to make it easier to switch between decks.

Closes #2434

### Approach
- Reuse existing DeckChooser and put in in the bottom bar of the (new) stats screen
- Change current deck on deck selection and reload stats page
- Left align the deck chooser and right align the save pdf/close buttons

### Questions
1. Currently, the size of the deck chooser corresponds to the length of the deck name. Should we add a constraint to the size like min/max/fixed size?
Screenshot 2023-03-28 at 1 55 37 PM
Screenshot 2023-03-28 at 1 56 00 PM

2. Removing the spacers and right-aligning button box made it so that the elements are also left/right aligned instead of centered in the legacy stats screen. Is this fine?
Screenshot 2023-03-28 at 1 57 31 PM

### DIFF
--- a/qt/aqt/stats.py
+++ b/qt/aqt/stats.py
@@ -8,7 +8,9 @@ from typing import Any
 import aqt
 import aqt.forms
 import aqt.main
+from anki.decks import DeckId
 from aqt import gui_hooks
+from aqt.operations.deck import set_current_deck
 from aqt.qt import *
 from aqt.theme import theme_manager
 from aqt.utils import (
@@ -43,6 +45,15 @@ class NewDeckStats(QDialog):
         f.groupBox.setVisible(False)
         f.groupBox_2.setVisible(False)
         restoreGeom(self, self.name, default_size=(800, 800))
+
+        from aqt.deckchooser import DeckChooser
+
+        DeckChooser(
+            self.mw,
+            f.deckArea,
+            on_deck_changed=self.on_deck_changed,
+        )
+
         b = f.buttonBox.addButton(
             tr.statistics_save_pdf(), QDialogButtonBox.ButtonRole.ActionRole
         )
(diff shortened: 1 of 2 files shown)


qt/aqt/stats.py
```diff
@@ -69,6 +80,11 @@ def closeWithCallback(self, callback: Callable[[], None]) -> None:
         self.reject()
         callback()
 
+    def on_deck_changed(self, deck_id: int) -> None:
+        set_current_deck(parent=self.mw, deck_id=DeckId(deck_id)).success(
+            lambda _: self.refresh()
+        ).run_in_background(initiator=self)
```
yellowjello, line 86: Is this the right way to change the current deck?

qt/aqt/stats.py
```diff
@@ -43,6 +45,15 @@ def __init__(self, mw: aqt.main.AnkiQt) -> None:
         f.groupBox.setVisible(False)
         f.groupBox_2.setVisible(False)
         restoreGeom(self, self.name, default_size=(800, 800))
+
+        from aqt.deckchooser import DeckChooser
```
yellowjello, line 49: Have to do the import here to avoid a circular import

qt/aqt/stats.py
dae: Yep, with some minor tweaks I've applied below.

dae:
Thanks yellowjello! Seems to work well. Regarding your questions:

1. I'm not too concerned about this - Qt will avoid expanding the button out if it doesn't fit in the window, and one character deck names are rare.
2. I'm not bothered by the alignment either - those old graphs are on life support and will eventually go away.

Your code looks fine - are you happy for me to merge this?

yellowjello:
That would be great!

PR #2464, opened 2023-04-06
Add emojis for graphs in hourly breakdown tooltip

### LINKED ISSUE #2444: "Hourly Breakdown" percentage axis is always 0-100%, even if maximum percentage is less
In the "Hourly Breakdown" graph on the stats page, the percentage axis seems to be always scaled from 0-100%, even if the maximum par height is less than 100%. See the following example, where the maximum percentage is around 76%, but the axis on the right hand side of the graph shows it as 100%.

image

Possible solutions:

1. Adjust the upper limit of the percentage axis to the maximum percentage among the bars
2. Scale all bars down accordingly to a 100% maximum

Versions in use:
Anki Version ⁨2.1.60 (76d88073)⁩
Python 3.10.10 Qt 6.4.2 PyQt 6.4.2

RumovZ:
Closes #2444.

### DIFF
--- a/ftl/core/statistics.ftl
+++ b/ftl/core/statistics.ftl
@@ -149,6 +149,10 @@ statistics-intervals-day-single =
 # hour range, eg "From 14:00-15:00"
 statistics-hours-range = From { $hourStart }:00~{ $hourEnd }:00
 statistics-hours-correct = { $correct }/{ $total } correct ({ $percent }%)
+# the emoji depicts the graph displaying this number
+statistics-hours-reviews = 📊 { $reviews } reviews
+# the emoji depicts the graph displaying this number
+statistics-hours-correct-reviews = 📈 { $percent }% correct ({ $reviews })
 statistics-hours-title = Hourly Breakdown
 statistics-hours-subtitle = Review success rate for each hour of the day.
 # shown when graph is empty
--- a/ts/graphs/hours.ts
+++ b/ts/graphs/hours.ts
@@ -170,12 +170,12 @@ export function renderHours(
             hourStart: d.hour,
             hourEnd: d.hour + 1,
         });
-        const correct = tr.statisticsHoursCorrect({
-            correct: d.correctCount,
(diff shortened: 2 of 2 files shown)


dae:
I felt the emoji inside the brackets was a little off - any objections to the following change, so the symbols line up?

```diff
diff --git a/ftl/core/statistics.ftl b/ftl/core/statistics.ftl
index 8024c234f..cc512f667 100644
--- a/ftl/core/statistics.ftl
+++ b/ftl/core/statistics.ftl
@@ -150,9 +150,9 @@ statistics-intervals-day-single =
 statistics-hours-range = From { $hourStart }:00~{ $hourEnd }:00
 statistics-hours-correct = { $correct }/{ $total } correct ({ $percent }%)
 # the emoji depicts the graph displaying this number
-statistics-hours-reviews = { $reviews } reviews 📊
+statistics-hours-reviews = 📊 { $reviews } reviews
 # the emoji depicts the graph displaying this number
-statistics-hours-correct-reviews = { $reviews } correct ({ $percent }% 📈)
+statistics-hours-correct-reviews = 📈 { $reviews } correct ({ $percent }%)
 statistics-hours-title = Hourly Breakdown
 statistics-hours-subtitle = Review success rate for each hour of the day.
 # shown when graph is empty
```

RumovZ:
I feel like this doesn't make it clear enough that the line graph displays the correct *ratio*. I agree that it looks better with the emojis at the line start, though.
What do you think of the following alternatives?
1. > 📊 50 reviews, 40 correct
    📈 80% correct
1. > 📊 50 reviews
    40 correct
    📈 80% correct
1. > 📊 50 reviews
    📈 80% (40) correct

dae:
I think I like 3 the best, since showing 40 in the top section is a bit confusing as it's not represented on the bar graph. It's fine as-is, or one possible minor tweak if it works for you:

:bar_chart: 50 reviews
:chart_with_upwards_trend:  80% correct (40)

PR #2472, opened 2023-04-14
Add ability to restore a notetype to its original configuration

dae:
This look ok to you @RumovZ?

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -45,4 +45,4 @@ good-names =
   ip,
 
 [IMPORTS]
-ignored-modules = anki.*_pb2, anki.sync_pb2, win32file,pywintypes,socket,win32pipe,pyaudio,anki.scheduler_pb2
+ignored-modules = anki.*_pb2, anki.sync_pb2, win32file,pywintypes,socket,win32pipe,pyaudio,anki.scheduler_pb2,anki.notetypes_pb2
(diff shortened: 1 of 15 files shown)


dae:
I'd like to move forward on this, but would still appreciate it if you could give it a quick look over when you have a chance. :-)

dae:
Thanks Abdo, I added the cast just before committing, and should have tested it again instead of just relying on CI.

RumovZ:
Sorry for the late reply!
Looks good to me, only maybe the dialog should also warn about the potential loss of note data? To me personally, that seems like a greater risk than losing styling information.
And to be honest, I didn't understand the difference between Kind and OriginalStockKind. E.g. why doesn't Kind has an ImageOcclusion member?

dae:
How about this?

```
card-templates-restore-to-default-confirmation = This will reset all fields and templates in this notetype to their default
    values, removing any extra fields/templates and their content, and any custom styling. Do you wish to proceed?
```

> E.g. why doesn't Kind has an ImageOcclusion member?

Backwards compatibility: older clients (and perhaps some of the current code too) assume kind is a binary choice, and do things like if kind == CLOZE: branch_A else branch_B, so adding a third option could lead to unexpected behaviour. They also expect all the basic notetypes to use the same kind.

RumovZ:
Perfect. :)

I get it now, thanks for explaining!

PR #2478, opened 2023-04-19
Various tweaks to I/O code

dae:
@krmanik heads up


glutanimate:
> Allow user to select I/O notetype instead of enforcing a specific name

:confetti_ball: much appreciated, thank you!

> Show the header on the front side as well (I presume this is what users expect; if not am happy to revert)
> Don't show comments on card (again, I presume users expect to use this field to add notes that aren't displayed during review, as they can use back extra for that)

:+1: I think both of these decisions make sense and are in line with users' expectations coming from IOE (where the header acts as a way for users to add a prompt to the image / other context that is helpful in answering the card, and where there are both supplementary fields that do appear on the back and fields that don't)

PR #2480, opened 2023-04-21
fix image height issues in iOS safari

### LINKED ISSUE #2477: Toggle Masks and Back Extra can't be seen on small screens
@krmanik no rush; don't let this interrupt your studies

It looks like currently 'toggle masks' and 'back extra' can't be seen when reviewing on mobile, as the container is being resized to the original image size, so the user has to scroll past a bunch of blank area to see it:

Screenshot 2023-04-19 at 2 47 58 pm
Screenshot 2023-04-19 at 2 48 08 pm

I've made some minor tweaks to the code as I work on integration, and will push them in the next day or two.

krmanik:
It may be different values in apple webkit when using `image.height`. I have changed it to `image.clientHeight`. It needs to test if fixed or not.

Fixes #2477

### DIFF
--- a/ftl/core/notetypes.ftl
+++ b/ftl/core/notetypes.ftl
@@ -50,3 +50,4 @@ notetypes-hide-one-guess-one = Hide One, Guess One
 notetypes-error-generating-cloze = An error occurred when generating an image occlusion note
 notetypes-error-getting-imagecloze = An error occurred while fetching an image occlusion note
 notetypes-error-loading-image-occlusion = Error loading image occlusion. Is your Anki version up to date?
+notetype-error-no-image-to-show = No image to show.
--- a/ts/image-occlusion/ImageOcclusionPage.svelte
+++ b/ts/image-occlusion/ImageOcclusionPage.svelte
@@ -32,13 +32,17 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 </script>
 
 <Container class="image-occlusion">
-    <ul>
+    <div class="tab-buttons">
         {#each items as item}
-            <li class={activeTabValue === item.value ? "active" : ""}>
-                <span on:click={tabChange(item.value)}>{item.label}</span>
-            </li>
+            <button
+                class="tab-item {activeTabValue === item.value ? 'active' : ''} 
+                    {item.value === 1 ? 'left-border-radius' : 'right-border-radius'}"
+                on:click={tabChange(item.value)}
+            >
+                {item.label}
+            </button>
         {/each}
-    </ul>
+    </div>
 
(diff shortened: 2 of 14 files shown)


krmanik:
The fix for safari is
1. Change `load` to `DOMContentLoaded`
2. Need to call again `setupImageCloze` method.

I hope it may work.

dae:
Thanks for looking into this Mani.  I gave this a test, and discovered the following:

- ios16 was working correctly already; the issue was with ios 14 and 15
- load→domcontentloaded and adding readystatechange did not seem to help
- if I logged height/clientHeight/naturalHeight in setupImageCloze(), they were 2k for all 3
- with the following patch, `height` becomes about 200, and the problem is fixed:

```diff
diff --git a/ts/reviewer/image_occlusion.ts b/ts/reviewer/image_occlusion.ts
index 5c7fedbfe..1d310811d 100644
--- a/ts/reviewer/image_occlusion.ts
+++ b/ts/reviewer/image_occlusion.ts
@@ -6,6 +6,10 @@ window.addEventListener("load", () => {
 });
 
 export function setupImageCloze(): void {
+    window.requestAnimationFrame(setupImageClozeInner);
+}
+
+function setupImageClozeInner(): void {
     const canvas: HTMLCanvasElement = document.querySelector("canvas")! as HTMLCanvasElement;
     canvas.style.maxWidth = "100%";
     canvas.style.maxHeight = "95vh";
```

My guess is that iOS is running the JavaScript before the CSS styling has been applied.

Perhaps we should revert the last two commits?

krmanik:
I updated with the suggested change.
Thanks

dae:
Thanks Mani, the redesign looks good. Just a couple of minor issues I noticed:

Screenshot 2023-04-25 at 3 21 15 pm

- Mask/Note are slightly smaller and slightly truncated. This is because AnkiMobile injects the following CSS:
```
body { font-size: 16px; }
:root { --font-size: 16px; }
```

The font size is set to 16px because that's the minimum size required for iOS not to try to zoom in on the page when the user taps on things like an input box. Would it be possible to adjust the toolbar icons based on --font-size?

- The toolbar doesn't scroll left and right when you try to scroll it if some of the buttons are cut off. This only happens on iOS 14/15, and seems to work correctly on iOS 16. If you know an easy fix, great, but don't spend much time on it - it's fairly minor, as the remaining buttons are obscure, and users can rotate their screen.

dae:
Update Note is cut off - maybe we should hide it on small screens?
Screenshot 2023-04-25 at 3 35 14 pm

krmanik:
- Added font size to css styling
- scrolling fixed, (tested on iOS 14/15)
- Removed `Update Note` label

PR #2481, opened 2023-04-23
Add a hook that allows to edit a note before it is added.

tatsumoto-ren:
This hook lets us change a note before it is added to the collection. 

It can be used for intercepting notes that AnkiConnect adds and adding new information to them.

### DIFF
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -562,6 +562,7 @@ class Collection(DeprecatedNamesMixin):
         return Note(self, notetype)
 
     def add_note(self, note: Note, deck_id: DeckId) -> OpChanges:
+        hooks.note_will_be_added(self, note, deck_id)
         out = self._backend.add_note(note=note._to_backend_note(), deck_id=deck_id)
         note.id = NoteId(out.note_id)
         return out.changes
--- a/pylib/tools/genhooks.py
+++ b/pylib/tools/genhooks.py
@@ -38,7 +38,12 @@ hooks = [
             "note: anki.notes.Note",
             "deck_id: anki.decks.DeckId",
         ],
-        doc="Can modify the note before it is added to the collection.",
+        doc="""Allows modifying a note before it's added to the collection.
+
+        This hook may be called both when users use the Add screen, and when
+        add-ons like AnkiConnect add notes. It is not called when importing. If
+        you wish to alter the Add screen, use gui_hooks.add_cards_will_add_note
+        instead.""",
     ),
     Hook(
         name="media_files_did_export",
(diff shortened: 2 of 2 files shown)


dae:
From the top of genhooks.py:

> In most cases, hooks are better placed in genhooks_gui.py.

I will add this to the add-on docs to make it more discoverable.

If we look at genhooks_gui, there is already an `add_cards_will_add_note` hook. Is there a reason you cannot use that?

dae:
Nevermind, I just saw you mentioned AnkiConnect above. I will give it some more thought.

tatsumoto-ren:
> In most cases, hooks are better placed in genhooks_gui.py.

Yes, I know about this. But I thought this is not a gui hook because it doesn't deal with gui. It is supposed to be used for modifying a note before it is added. For example, for adding furigana to sentences or editing images.

> there is already an add_cards_will_add_note hook. Is there a reason you cannot use that?

When AnkiConnect adds a new note, it calls `collection.addNote`, which calls `collection.add_note`. It goes past the existing hook.

dae:
I'm not totally thrilled about adding another hook that performs identically to the existing one except when the user happens to be using AnkiConnect, but I do appreciate that it's a simple solution, and can't think of a better approach.

You haven't given me permission to push into your branch, so please apply the following patch before I merge this:

```diff
commit f2c98c677fd7cc29aae0d621ece266b33919be5a (HEAD -> add_note_hook)
Author: Damien Elmes <gpg@ankiweb.net>
Date:   Thu Apr 27 13:29:24 2023 +1000

    Improve hook documentation

diff --git a/pylib/tools/genhooks.py b/pylib/tools/genhooks.py
index fd248dafd..40d25e1e2 100644
--- a/pylib/tools/genhooks.py
+++ b/pylib/tools/genhooks.py
@@ -38,7 +38,12 @@ hooks = [
             "note: anki.notes.Note",
             "deck_id: anki.decks.DeckId",
         ],
-        doc="Can modify the note before it is added to the collection.",
+        doc="""Allows modifying a note before it's added to the collection.
+
+        This hook may be called both when users use the Add screen, and when
+        add-ons like AnkiConnect add notes. It is not called when importing. If
+        you wish to alter the Add screen, use gui_hooks.add_cards_will_add_note
+        instead.""",
     ),
     Hook(
         name="media_files_did_export",
```

tatsumoto-ren:
> You haven't given me permission to push into your branch

I must have missed it somewhere. Now I don't know how enable it. But I will run git apply with your diff.

dae:
For future reference: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

PR #2485, opened 2023-04-27
image occlusion button in note editor

krmanik:
Added a button in note editor for adding/editing image occlusion notes.

The current implementation will be updated with option have mask editor and note editor in same window not in separate window.


dae:
A few tweaks are required to get this to a minimum viable level:

- Older notetypes won't have that key:

```
    setOriginalStockKind({json.dumps(self.note.note_type()["originalStockKind"])});
KeyError: 'originalStockKind'
```

- Notetype will be missing:

```
diff --git a/pylib/anki/collection.py b/pylib/anki/collection.py
index 4b924fe17..ae1dd7a54 100644
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -466,6 +466,10 @@ class Collection(DeprecatedNamesMixin):
     def get_image_for_occlusion(self, path: str | None) -> GetImageForOcclusionResponse:
         return self._backend.get_image_for_occlusion(path=path)

+    def add_image_occlusion_notetype(self) -> None:
+        "Add if missing."
+        self._backend.add_image_occlusion_notetype()
+
     def add_image_occlusion_note(
         self,
         notetype_id: int,
diff --git a/qt/aqt/addcards.py b/qt/aqt/addcards.py
index ef8a11500..1d0124320 100644
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -55,6 +55,7 @@ class AddCards(QMainWindow):
         self._last_added_note: Optional[Note] = None
         gui_hooks.operation_did_execute.append(self.on_operation_did_execute)
         restoreGeom(self, "add")
+        self.col.add_image_occlusion_notetype()
         gui_hooks.add_cards_did_init(self)
         self.show()
```

- Image occlusion window should be closed after user adds/updates an occlusion

> The current implementation will be updated with option have mask editor and note editor in same window not in separate window.

Does that mean you plan to change the code to avoid showing an extra window? If so, did you intend for this PR to be a draft at the moment, or did you want the current approach merged in and plan to make those changes in a future update?

ftl/core/notetypes.ftl
```diff
@@ -51,3 +51,5 @@ notetypes-error-generating-cloze = An error occurred when generating an image oc
 notetypes-error-getting-imagecloze = An error occurred while fetching an image occlusion note
 notetypes-error-loading-image-occlusion = Error loading image occlusion. Is your Anki version up to date?
 notetype-error-no-image-to-show = No image to show.
+notetype-create-edit-io-message = Tap the two squares to create or edit an occlusion.
+notetype-io-switch-type = Tap 'Type: Image Occlusion' to switch to normal text/media cards.
```
dae, line 55: I didn't think too carefully about the wording when adding this to the AnkiMobile beta. Maybe this would be better. 

```suggestion
notetype-io-switch-type = Change 'Image Occlusion' to 'Basic' or 'Cloze' to add normal text/media cards.
```

qt/aqt/imageocclusion.py
```diff
@@ -0,0 +1,62 @@
...
+        layout.setContentsMargins(0, 0, 0, 0)
+        layout.addWidget(self.web)
+        self.setLayout(layout)
+
+        if mode == "add":
+            options = {"kind": mode, "imagePath": path, "notetypeId": notetype_id}
+        elif mode == "edit":
+            options = {"kind": mode, "noteId": note_id}
+
+        self.web.eval(f"anki.setupImageOcclusion({options});")
```
dae, line 55: Adding an explicit json.dumps() call here would be better than relying on Python's string repr happening to be valid JS.

qt/aqt/editor.py
```diff
@@ -1460,4 +1486,14 @@ def set_cloze_button(editor: Editor) -> None:
     )
 
 
+def set_image_occlusion_button(editor: Editor) -> None:
+    action = "show" if editor.note.note_type()["originalStockKind"] == 6 else "hide"
```
dae, line 1490: This should use the enum name, not a number. pylib should export the symbol, and then aqt can reference it via StockNotetypes.ORIGINAL_*

```diff
diff --git a/pylib/anki/models.py b/pylib/anki/models.py
index 7fe27dba3..fe9c9b5f2 100644
--- a/pylib/anki/models.py
+++ b/pylib/anki/models.py
@@ -27,6 +27,7 @@ NotetypeNameIdUseCount = notetypes_pb2.NotetypeNameIdUseCount
 NotetypeNames = notetypes_pb2.NotetypeNames
 ChangeNotetypeInfo = notetypes_pb2.ChangeNotetypeInfo
 ChangeNotetypeRequest = notetypes_pb2.ChangeNotetypeRequest
+StockNotetype = notetypes_pb2.StockNotetype

 # legacy types
 NotetypeDict = dict[str, Any]
```

ts/editor/NoteEditor.svelte
```diff
@@ -456,138 +462,149 @@ the AddCards dialog) should be implemented in the user of this component.
...
-                                    on:toggle={() => toggleRichTextInput(index)}
-                                />
-                            {:else}
-                                <PlainTextBadge
-                                    show={!fieldsCollapsed[index] &&
-                                        (fields[index] === $hoveredField ||
-                                            fields[index] === $focusedField)}
-                                    bind:off={plainTextsHidden[index]}
-                                    on:toggle={() => togglePlainTextInput(index)}
+    {#if originalStockKind == 6}
```
dae, line 465:
```diff
diff --git a/ts/editor/NoteEditor.svelte b/ts/editor/NoteEditor.svelte
index 2b4e249a4..adb0aa32d 100644
--- a/ts/editor/NoteEditor.svelte
+++ b/ts/editor/NoteEditor.svelte
@@ -42,6 +42,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 <script lang="ts">
     import { bridgeCommand } from "@tslib/bridgecommand";
     import * as tr from "@tslib/ftl";
+    import { Notetypes } from "@tslib/proto";
     import { onMount, tick } from "svelte";
     import { get, writable } from "svelte/store";

@@ -462,7 +463,7 @@ the AddCards dialog) should be implemented in the user of this component.
         </Absolute>
     {/if}

-    {#if originalStockKind == 6}
+    {#if originalStockKind == Notetypes.StockNotetype.OriginalStockKind.ORIGINAL_STOCK_KIND_IMAGE_OCCLUSION}
         <div class="io-message">
             <div>
                 {tr.notetypeCreateEditIoMessage()}
```

qt/aqt/imageocclusion.py
krmanik: I have changed the implementation to include setupImageOcclusion method in note editor.

ftl/core/notetypes.ftl
krmanik: TODO

krmanik:
We should go with current implementation.
The inclusion of image occlusion in note editor may need more time to implement and I will implement.

dae:
Some of your previously-mentioned concerns have already been ameliorated by the approach AnkiMobile uses and Mani has copied here - you can try it out in a fresh profile by opening the add screen and selecting the I/O notetype. It's not perfect, and the necessity to switch to the Note tab to be able cards is still a bit difficult to discover, but there's less room for accidents.

>  I don't see us critically pressured for time here. I think the most important and time-sensitive thing was to introduce IO to AnkiMobile and satisfy the long-standing feature request behind that.

Yes, that was my primary concern. There would be advantages to having support in the desktop version as well - a larger testing base for one, and allowing users to make edits/new cards on whatever platform they happen to be on at the time, but those things are not super urgent from my perspective.

How do you feel about things Mani? Are you happy to hold off until the editor integration is done, or are you keen to get this into the hands of desktop users sooner rather than later? If you don't want to wait, maybe we could reach some compromise where this requires a debug console command to opt into, with a warning that it's only a temporary solution?

krmanik:
If it can wait, then I will complete the editor integration. I have viewed the forum and thought that this feature is needed to be merged but it can be wait because on desktop IOE can be used for now.

So, I will complete the integration and push the changes in few weeks. The integration will be better, instead of implementing two features.

dae:
We're currently focused on an MVP; feature requests are best posted in the suggestions section of the forums (and will probably get more traction after we get the MVP into the hands of users).

krmanik:
I have integrated the mask editor in note editor, next I will implement update button in edit mode and hide occlusion and image field.

dae:
The integration seems to work well! Two issues I noticed:

- The first is likely the fault of my recent changes: the image is constrained to the original dimensions of the window, so if you start occluding in a small window and then resize it to become larger, the image is stuck at the smaller size. The size to fit button seems to force that size, instead of adjusting to the new window dimensions.
- Instead of note/mask buttons/labels, what if we had a single icon that acts a a toggle? In the add screen, clicking it would prompt to select an image, switch to the mask mode, and the button would be depressed. If users press it again, it switches back to notes mode and becomes undepressed. Press it again and it switches back to masks / is depressed again. WDYT?

dae:
Rebasing over the main branch should fix the compile error

krmanik:
What could be possible fix for this? I have rebased and tried to fix.

```
./tools/web-watch 
033cn    Finished dev [unoptimized + debuginfo] target(s) in 0.15s
ninja: error: unknown target 'qt/aqt', did you mean 'qt_aqt'?
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
ninja: error: unknown target 'qt/aqt', did you mean 'qt_aqt'?
```

```
~/anki$ ./ninja pylib/anki qt/aqt check
    Finished dev [unoptimized + debuginfo] target(s) in 0.26s
ninja: error: unknown target 'pylib/anki', did you mean 'pylib_anki'?
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
ninja: error: unknown target 'pylib/anki', did you mean 'pylib_anki'?
```

dae:
I've pushed a fix for web-watch; I forgot to update it when I made related changes.

krmanik:
The integration is almost completed.

dae:
Thanks Mani, it seems to work well. I needed to make a change to be able to open the add screen from an older collection; I've pushed a fix for that.

The integration in the editor could be a bit cleaner from a code perspective (e.g. global styling and a somewhat fragile adjustMaskEditorPosition()), but I presume you've done it that way to continue supporting the use outside of the editor case for now.

Our Python/TS editor bridge is a bit of a mess at the moment, and is going to need some serious attention in the near future to tidy it up and make it easier for the mobile clients to integrate. Hopefully I'll find time for that in the not-too-distant future.

Also in case you hadn't seen yet, a new AnkiMobile release came out yesterday, and it includes the new I/O support. It's mentioned in the change notes, but I'll mention it here too - thank you again for all the hard work you've put into this :-) I'm sure users are going to love it.

krmanik:
I have used CSS for adjusting position for maskeditor, during resize the position does not update so I have moved it to JS. The code can also be moved to image-occlusion ts dir.

I've seen AnkiMobile's IO feature update. Users will love this awaited mobile client feature.

dae:
Once https://github.com/ankidroid/Anki-Android-Backend/pull/294 gets merged, I/O should be ready for exposing in AnkiDroid as well, though they'll need to get through their scoped storage update first.

> The code can also be moved to image-occlusion ts dir.

To confirm, are you saying this is a further change you'd like to make, or that you've already made? Are you ready for this to be considered for merging, or would you like to make further changes?

@glutanimate heads up

krmanik:
> To confirm, are you saying this is a further change you'd like to make, or that you've already made? Are you ready for this to be considered for merging, or would you like to make further changes?

It is completed.
I want to change some UI related stuff and check if more code can be moved from note editor svelte (related to io) to io svelte then push the changes.

krmanik:
It can be merged now.

dae:
Having played with it a bit more, I'm not sure about hiding the image by default in editing mode. I suspect many image occlusions do not have any associated text, and the information is only contained in the image. Currently when the user clicks on a new note in the browser, they just see the blank fields, and have to manually click on the image occlusion button to see the image. Perhaps it should either show the image occlusion editor by default, or not hide the image in the fields list?

krmanik:
Both can be done. Showing mask editor by default need to show a message to click on button to add image and create occlusion, in case of showing image field, it needs to show message to switch to mask editor then input image.
So, I have implemented first case because image should be changed from mask editor not by editing field in note editor. Also remove the function to adjust mask editor position and used target element to init.

dae:
"Select image" works well for the add screen when an image occlusion notetype is selected, but it was the browsing screen that I was more worried about - I think users browsing their existing cards will want to be able to see the image as soon as they select a note. There are a few issues with the latest commit:

- "Select image" is shown in the add screen even if a normal notetype is selected
- "Select image" is shown in the browse screen when you select an existing note. It's shown even if it's not an image occlusion note. If you click on it, the image appears, but if you then select a different note, the image doesn't change.

krmanik:
I have only viewed/implemented the add and edit mode but now we have to check browse mode implementation also.
It will be fixed in next commit.

glutanimate:
Thanks again from me as well for all your work on this @krmanik 

Starting to test on again now on desktop. First quick observation, not sure if this is because things are work-in-progress on the current branch HEAD, but "Select Image" currently appears on all note types for me:

https://github.com/ankitects/anki/assets/5459332/74e32105-8386-4874-b495-5505cef7b8c8

krmanik:
I have implemented for add, edit and browse mode, the latest changes fix above issues.

qt/aqt/editor.py
```diff
@@ -1177,6 +1199,40 @@ def toggleCloseHTMLTags(self) -> None:
...
+        )
+        filter = f"{tr.editing_media()} ({extension_filter})"
+
+        def accept(file: str) -> None:
+            try:
+                html = self._addMedia(file)
+                options = {"kind": "add", "imagePath": file, "notetypeId": 0}
+                # pass both html and options
+                options = {"html": html, "mode": options}
+                self.web.eval(f"setupMaskEditor({options})")
```
glutanimate, line 1215: It would probably be safer to use `json.dumps()` here instead of stock Python format string serialization

glutanimate:
Showing the mask editor by default when editing and in the browser is a nice improvement, thanks! I would agree that this is what users will likely find most helpful in most cases.

A couple of small bugs I've found with the current implementation:

- Editing a shape will sometimes cause that shape to disappear. It likely has to do with shapes intersecting a screen edge.

    https://github.com/ankitects/anki/assets/5459332/d8a3ee24-1b3e-42f0-ab3c-0ca18eb4b302

- The fields <-> mask editor toggle sometimes requires multiple clicks to work. Not sure what triggers this exactly:

    https://github.com/ankitects/anki/assets/5459332/21b657f7-94df-4a0d-ada1-35b200e3720c

glutanimate:
A couple more general thoughts on the current design:

**Unifying mask editor and fields**

While I like the idea of giving the mask editor as much screen space as possible, I'm wondering if it wouldn't be better to display both the mask editor and fields in the same view. My gut feeling is that the toggle will turn out to be a fairly hard to discover feature and make the fields even less accessible than in IOE. The add-on has two advantages here:

1. In note adding & editing mode, the UI path to the fields editor is super straightforward as users just have to find the _Fields_ tab (similar to the previous multi-window implementation of native IO). I think the icon-only button is a lot more cryptic here, but of course a given limitation of the restricted space in the toolbar.
2. In the browser, users can see the image, masks, and text fields right next to each other. This means that users can essentially grok the entirety of the note at a glace, quickly locate text matches when performing searches in the browser, easily edit one note after another by using the note-jump hotkeys.
   
Putting both the mask editor and fields in the same view would also make the UX of adding IO notes consistent with any other note in that you have just one vertically scrollable entry form that contains all editable note data. With fields being collapsible, the field list scrollable, and having the luxury of desktop screen sizes, I don't think we need to be concerned about usable area for the masks editor.

Here is a quick mock-up (similar to Matthias' mockup from the previous PR, but with mask editing not integrated into the image overlay and instead being a separate component as it currently is)

io-mockup-one-screen

To save further space, we could collapse some fields, or all fields, by default.

WDYT? I quite strongly believe that this would make IO significantly more straightforward to use as users would essentially use the same user input flow as for any other note type, and this is pretty much what I always hoped we could achieve with native IO at some point.
    
How difficult do you think it would be to morph the current implementation into something like this @krmanik? An easier step might be not to split the fields up and have all of them positioned below the masks editor.

**Saving edits in the browser on focus switch**

Along the same lines of bringing uniformity in UI interactions to IO, I think we should strongly consider changing the way the browser implementation works. Right now users have to confirm their edits by using the same "Hide ..." buttons as they do for creating new cards. That introduces a special behavior that's not present for other note types in that form. Users will be used to just being able to edit a field and switch to the next note / close the browser for their edits to apply. So I could see this new behavior introducing quite a bit of confusion.

My suggestion here would be to save on focus loss, as we do for other note types. To switch between occlusion modes, I would propose adding a dropdown selector in place of the buttons. This is what IOE does at the moment, so that would also have the benefit of being familiar to existing users. But more importantly, it would fit better with the idea of just being able to have your changes be saved without any explicit confirmation step.

dae:
Re: Unifying mask editor and fields

While I can see how it would be advantageous to show both fields and image at the same time when space is available, there are some things to bear in mind:

- the plan is to integrate the desktop editor in the mobile clients as well, where space will be constrained
- users may not want the editing screen to take up the full screen on desktop, because they're using a tiling WM, macOS split screen mode, etc, and have other things like a browser open at the same time

In such cases, even showing the fields collapsed can be consuming screen real-estate that is scarce, so I'm not sure just making the fields collapsible is a good-enough solution. We could potentially solve this with a more reactive design that switches between all-in-one and on-two-pages approaches, but this was originally meant to be an MVP :-) Perhaps this is best left as-is for now?

Re: Saving edits in the browser on focus switch

I found it a bit confusing at first too. Auto-saving + a dropdown sound like a more user-friendly solution. The auto save could perhaps be hooked into saveNow() and saveOnPageHide(). Some global function would also need to be exported to save, so that the mobile clients can trigger a save until we can get them shifted over to the new editor.

dae:
One minor issue I noticed with the current code: there's a brief flash as a new note is selected in the browse screen - it looks like the image is initially shown at full size, and then shrunk down to fit .

krmanik:
> Editing a shape will sometimes cause that shape to disappear. It likely has to do with shapes intersecting a screen edge.

This is interesting, I will check it.

> The fields <-> mask editor toggle sometimes requires multiple clicks to work. Not sure what triggers this exactly:

The boolean get reset in next note, earlier used svelte store, and removed, but again check with store implementation.

> Unifying mask editor and fields

The current implementation will be changed when note editor reimplemented/ported for mobile client, so the current implement is temporary solution. So, mask editor can be same page or can be in another page accessed by toggle button. But I think the toggle button will be better in case more area for adding/editing shapes and separate area for only text-based editing.

> I think the icon-only button is a lot more cryptic here

The documentation will be helpful or message similar to AnkiMobile can be used.

> How difficult do you think it would be to morph the current implementation into something like this?

It is not difficult to implement, it can be implemented, I will try this implementation and check which can be used or may be a prefs can be added so that it will be upto user which way they want to create io notes.

> Right now, users have to confirm their edits by using the same "Hide ..." buttons as they do for creating new cards. That introduces a special behavior that's does not present for other note types in that form.

It will not be easy to guess one option from two option (occlude inactive and not occlude inactive). So, two buttons used here in case of IO notetypes.

> To switch between occlusion modes, I would propose adding a dropdown selector in place of the buttons.

It will be added.


> the plan is to integrate the desktop editor in the mobile clients as well, where space will be constrained

The mask editor will be accessed with toggle in case of smaller screen and for large screen the mask editor can be in same page or another and that will be upto user to make default from prefs.

> Auto-saving + a dropdown sound like a more user-friendly solution. 

The dropdown will be implemented. Autosaving implementation need to check further, I will check it.

> there's a brief flash as a new note is selected in the browse screen - it looks like the image is initially shown at full size, and then shrunk down to fit.

Will be fixed in next commit.

dae:
> It will not be easy to guess one option from two option 

Can't we just check for 'oi=1' in the note text?

krmanik:
It can also be done.

dae:
Please let me know when you're ready for another review

krmanik:
I tried to add io page in note editor, it is not properly integrating, it needs to re-implement note editor from scratch.
The flicker issues also tried to fix using this, but the props not updating so the issue still not fixed.
```js
if (page) {
        page.$set({ mode: options.mode });
    } else {
        page = await setupImageOcclusion(options.mode, ioMaskEditor);
}
```

Other issues have been fixed, and it is ready for review.

ftl/core/notetypes.ftl
```diff
@@ -51,3 +51,5 @@ notetypes-error-generating-cloze = An error occurred when generating an image oc
 notetypes-error-getting-imagecloze = An error occurred while fetching an image occlusion note
 notetypes-error-loading-image-occlusion = Error loading image occlusion. Is your Anki version up to date?
 notetype-error-no-image-to-show = No image to show.
+notetypes-no-occlusion-created = No occlusion note created.
```
dae, line 54:
```suggestion
notetypes-no-occlusion-created = You must make at least one occlusion.
```

qt/aqt/addcards.py
```diff
@@ -48,13 +48,16 @@ def __init__(self, mw: AnkiQt) -> None:
...
         add_close_shortcut(self)
         self._load_new_note()
+        self.setupButtons()
         self.history: list[NoteId] = []
         self._last_added_note: Optional[Note] = None
         gui_hooks.operation_did_execute.append(self.on_operation_did_execute)
         restoreGeom(self, "add")
+        self.col.add_image_occlusion_notetype()
+        # hide io buttons for note type other than image occlusion
+        self.show_hide_add_buttons()
```
dae, line 60: Maybe this could be called in setupButtons()?

qt/aqt/addcards.py
```diff
@@ -112,6 +115,12 @@ def setupButtons(self) -> None:
         self.compat_add_shorcut = QShortcut(QKeySequence("Ctrl+Enter"), self)
         qconnect(self.compat_add_shorcut.activated, self.addButton.click)
         self.addButton.setToolTip(shortcut(tr.adding_add_shortcut_ctrlandenter()))
+
+        # add io button
+        self.ioAddButton = bb.addButton(f"{tr.actions_add()} {downArrow()}", ar)
```
dae, line 120: You've used a mix of camelCase and snake_case in this file - it would be nice to stick to snake_case for newly-added items. I do realise it's an easy mistake to make, as Qt uses camelCase, and our legacy code does too.

qt/aqt/addcards.py
```diff
@@ -348,6 +371,24 @@ def doClose() -> None:
 
         self.ifCanClose(doClose)
 
+    def onAddIo(self) -> None:
+        m = QMenu(self)
+        a = m.addAction(tr.notetypes_hide_all_guess_one())
+        qconnect(a.triggered, self.add_io_hide_all_note)
+        a = m.addAction(tr.notetypes_hide_one_guess_one())
+        qconnect(a.triggered, self.add_io_hide_one_note)
+        m.exec(self.ioAddButton.mapToGlobal(QPoint(0, 0)))
```
dae, line 380:
```suggestion
        m.popup(QCursor.pos())
```

qt/aqt/editcurrent.py
```diff
@@ -33,9 +30,39 @@ def __init__(self, mw: aqt.AnkiQt) -> None:
...
         self.show()
 
+    def setupButtons(self) -> None:
+        if self.editor.current_notetype_is_image_occlusion():
+            bb = self.form.buttonBox
+            ar = QDialogButtonBox.ButtonRole.ActionRole
+            # add io hide all button
+            self.ioAddButton = bb.addButton(tr.actions_update_note(), ar)
+            qconnect(self.ioAddButton.clicked, self.onAddIo)
+            self.ioAddButton.setShortcut(QKeySequence("Ctrl+Shift+I"))
```
dae, line 44: I'm confused by the different way this is handled in the editcurrent and browse dialogs. Why do we have an explicit update button in one, but update automatically in the other? Maybe updates could just be automatic in both?

ts/editor/NoteEditor.svelte
```diff
@@ -603,4 +729,37 @@ the AddCards dialog) should be implemented in the user of this component.
...
+    }
+    :global(.image-occlusion .tool-bar-container) {
+        top: unset !important;
+        margin-top: 2px !important;
+    }
+
+    :global(.active-io-btn) {
+        background: var(--button-primary-bg) !important;
+        color: white !important;
+    }
```
dae, line 755: This style belongs in the svelte file that uses it, and then doesn't need to be global. Maybe the same applies to some of the others too?

ts/editor/editor-toolbar/ImageOcclusionButton.svelte
```diff
@@ -0,0 +1,59 @@
...
+        {api}
+    >
+        <ButtonGroupItem>
+            <IconButton
+                id="io-mask-btn"
+                on:click={() => {
+                    if (get(maskEditorButtonPressed)) {
+                        bridgeCommand(`toggleMaskEditor:${false}`);
+                    } else {
+                        bridgeCommand(`toggleMaskEditor:${true}`);
```
dae, line 38: Instead of an action, this should be a store value.

ts/editor/tsconfig.json
```diff
@@ -6,7 +6,8 @@
         "mathjax-overlay/*",
         "plain-text-input/*",
         "rich-text-input/*",
-        "editor-toolbar/*"
+        "editor-toolbar/*",
+        "../image-occlusion/*"
```
dae, line 10: This should be in the references section below instead.

ts/image-occlusion/Toolbar.svelte
```diff
@@ -27,8 +31,53 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+        instance.pause();
+        stopDraw(canvas);
+        canvas.selectionColor = "rgba(100, 100, 255, 0.3)";
+    };
 </script>
 
+<div class="tool-bar-container">
+    {#each tools as tool}
+        <IconButton
+            class="tool-icon-button {activeTool == tool.id ? 'active-tool' : ''}"
```
dae, line 69: Another candidate for class shortcut

ts/image-occlusion/Toolbar.svelte
```diff
@@ -189,4 +239,31 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+        width: 32px;
+        height: 32px;
+        margin: unset;
+        padding: 6px !important;
+    }
+
+    :global(.active-tool) {
+        color: white !important;
+        background: var(--button-primary-bg) !important;
+    }
```
dae, line 268: Do any of these need to be global?

dae (commented):
A few more issues I'm afraid - mainly just cleaning things up to make it more maintainable. When working on these, it would be great if you could use new commits rather than updating the old ones, as that makes it easier to provide further feedback.

krmanik:
I have read all the reviewed message and will implement considering the suggested changes. Then push the changes in new commit. Thanks for the review.

dae:
Thanks as always Mani :-)

ts/image-occlusion/Toolbar.svelte
krmanik: Getting this error when using class shortcut.
`Classes can only be applied to DOM elements, not componentssvelte(invalid-class)`

ts/image-occlusion/Toolbar.svelte
krmanik: If class it on HTML element then global cannot be used but when class need to add to component then following issues is showing 
`Unused CSS selector ".active-tool"svelte(css-unused-selector)`

ts/editor/editor-toolbar/ImageOcclusionButton.svelte
krmanik: Implemented using store

ts/editor/NoteEditor.svelte
krmanik: `Unused CSS selector ".active-tool"svelte(css-unused-selector)`

But class needed for hiding the element, so HTML element does not need global but component needs global.

ts/editor/NoteEditor.svelte
```diff
@@ -372,12 +390,97 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+
+        const ioMaskEditor = document.getElementById("io-mask-editor") as HTMLElement;
+        const ioSelectImageDiv = document.getElementById(
+            "io-select-image-div",
+        ) as HTMLElement;
+        if (ioSelectImageDiv) {
+            ioSelectImageDiv.remove();
+        }
+
+        await setupImageOcclusion(options.mode, ioMaskEditor);
```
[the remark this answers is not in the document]
krmanik: Implemented using suggested approach.

ts/editor/NoteEditor.svelte
```diff
@@ -372,12 +390,97 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
     import { mathjaxConfig } from "../editable/mathjax-element";
     import CollapseLabel from "./CollapseLabel.svelte";
     import * as oldEditorAdapter from "./old-editor-adapter";
 
+    $: isShowMaskEditor = false;
+    $: isIOImageLoaded = false;
+    function toggleMaskEditor(show: boolean) {
+        isShowMaskEditor = show;
+        const element = document.getElementById("io-mask-editor") as HTMLElement;
+        const ioMaskBtn = document.getElementById("io-mask-btn") as HTMLElement;
```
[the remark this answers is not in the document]
krmanik: Use the class shortcut and it is more cleaner than JS implementation.

qt/aqt/editor.py
```diff
@@ -476,6 +477,11 @@ def onBridgeCmd(self, cmd: str) -> Any:
         elif cmd in self._links:
             return self._links[cmd](self)
 
+        elif cmd.startswith("toggleMaskEditor"):
+            (_, show_str) = cmd.split(":", 1)
+            show = show_str == "true"
+            self.onToggleMaskEditor(show)
+
```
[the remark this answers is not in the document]
krmanik: Remove toggle mask editor and use store to handle the visibility of mask editor.

qt/aqt/editcurrent.py
krmanik: Now the add button only added to Add Cards, the other mode edit and browse mode does not have add button. The only downside for this is that if user want to change the `hide all` to `hide one` note then they have to delete and add again.
Or a more button in ImageOcclusionPage can be added to switch between `hide all` and `hide one` notetype.

krmanik:
The last commit has changes for recent reviews.

qt/aqt/addcards.py
```diff
@@ -142,13 +139,17 @@ def setupButtons(self) -> None:
         b.setEnabled(False)
         self.historyButton = b
 
+        self.col.add_image_occlusion_notetype()
```
dae, line 142: A minor point, but as the notetype doesn't relate to the buttons, this call might be better left in its old location.

ts/editor/NoteEditor.svelte
```diff
@@ -392,57 +387,29 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-            ioMaskBtn?.classList.remove("active-io-btn");
-        }
-        if (element) {
-            element.style.display = show ? "block" : "none";
-        }
-    }
-
+    let isIOImageLoaded = false;
+    const showMaskEditor = isShowMaskEditor;
+    let imageOcclusionMode: IOMode;
```
dae, line 401: You later test for `if (imageOcclusionMode ...)`, so the type should be `IOMode | undefined`.

ts/editor/NoteEditor.svelte
```diff
@@ -572,11 +538,14 @@ the AddCards dialog) should be implemented in the user of this component.
         </Absolute>
     {/if}
 
-    <!-- for init maskeditor -->
-    <div id="io-mask-editor" style="display: none;" />
+    {#if imageOcclusionMode && isImageOcclusion}
```
dae, line 541: Could we perhaps avoid the need for a separate isImageOcclusion boolean, and just check for imageOcclusionMode not being undefined?

ts/editor/NoteEditor.svelte
```diff
@@ -572,11 +538,14 @@ the AddCards dialog) should be implemented in the user of this component.
...
+        <div style="display: {$showMaskEditor ? 'block' : 'none'}">
+            <ImageOcclusionPage mode={imageOcclusionMode} />
+        </div>
+    {/if}
 
-    {#if isShowMaskEditor && isImageOcclusion && !isIOImageLoaded}
-        <div id="io-select-image-div" style="padding-top: 60px; text-align: center">
+    {#if $showMaskEditor && isImageOcclusion && !isIOImageLoaded}
```
dae, line 547: Minor: $ioMaskEditorVisible would be slightly more clear, as 'show' is a verb and sounds like an action.

ts/editor/editor-toolbar/ImageOcclusionButton.svelte
```diff
@@ -31,29 +30,21 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+                class={$showMaskEditor ? "active-io-btn" : ""}
                 on:click={() => {
-                    if (get(maskEditorButtonPressed)) {
-                        bridgeCommand(`toggleMaskEditor:${false}`);
-                    } else {
-                        bridgeCommand(`toggleMaskEditor:${true}`);
-                    }
-                    maskEditorButtonPressed.set(!get(maskEditorButtonPressed));
+                    $showMaskEditor = !$showMaskEditor;
+                    isShowMaskEditor.set($showMaskEditor);
```
dae, line 36: Do we need showMaskEditor and the set call here? Why not just directly refer to $isShowMaskEditor?

ts/image-occlusion/Toolbar.svelte
dae: Whoops, good point.

ts/editor/NoteEditor.svelte
dae: Yep, because of https://github.com/ankitects/anki/pull/2485#discussion_r1271678478, I see this will need to stay global.

dae:
Thanks Mani, looking good.

ts/editor/editor-toolbar/ImageOcclusionButton.svelte
krmanik: This is called on button click and handle the switch between mask editor and note editor.

dae: Why not remove the alias and access/set the store directly?

```diff
diff --git a/ts/editor/NoteEditor.svelte b/ts/editor/NoteEditor.svelte
index 9ff14bcae..dd5b96237 100644
--- a/ts/editor/NoteEditor.svelte
+++ b/ts/editor/NoteEditor.svelte
@@ -390,14 +390,13 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import ImageOcclusionPage from "image-occlusion/ImageOcclusionPage.svelte";
     import type { IOMode } from "image-occlusion/lib";
     import { exportShapesToClozeDeletions } from "image-occlusion/shapes/to-cloze";
-    import { ioMaskEditorVisibleStore } from "image-occlusion/store";
+    import { ioMaskEditorVisible } from "image-occlusion/store";
 
     import { mathjaxConfig } from "../editable/mathjax-element";
     import CollapseLabel from "./CollapseLabel.svelte";
     import * as oldEditorAdapter from "./old-editor-adapter";
 
     let isIOImageLoaded = false;
-    const ioMaskEditorVisible = ioMaskEditorVisibleStore;
     let imageOcclusionMode: IOMode | undefined;
     async function setupMaskEditor(options: { html: string; mode: IOMode }) {
         imageOcclusionMode = options.mode;
diff --git a/ts/editor/editor-toolbar/ImageOcclusionButton.svelte b/ts/editor/editor-toolbar/ImageOcclusionButton.svelte
index 550f86d1b..3b17ad121 100644
--- a/ts/editor/editor-toolbar/ImageOcclusionButton.svelte
+++ b/ts/editor/editor-toolbar/ImageOcclusionButton.svelte
@@ -6,7 +6,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import ButtonGroup from "components/ButtonGroup.svelte";
     import DynamicallySlottable from "components/DynamicallySlottable.svelte";
     import IconButton from "components/IconButton.svelte";
-    import { ioMaskEditorVisibleStore } from "image-occlusion/store";
+    import { ioMaskEditorVisible } from "image-occlusion/store";
 
     import ButtonGroupItem, {
         createProps,
@@ -16,7 +16,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import { mdiViewDashboard } from "./icons";
 
     export let api = {};
-    const ioMaskEditorVisible = ioMaskEditorVisibleStore;
 </script>
 
 <ButtonGroup>
@@ -32,7 +31,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                 id="io-mask-btn"
                 class={$ioMaskEditorVisible ? "active-io-btn" : ""}
                 on:click={() => {
-                    ioMaskEditorVisibleStore.set(!$ioMaskEditorVisible);
+                    $ioMaskEditorVisible = !$ioMaskEditorVisible;
                 }}
             >
                 {@html mdiViewDashboard}
diff --git a/ts/image-occlusion/store.ts b/ts/image-occlusion/store.ts
index c37736d43..bf56b688f 100644
--- a/ts/image-occlusion/store.ts
+++ b/ts/image-occlusion/store.ts
@@ -10,4 +10,4 @@ export const zoomResetValue = writable(1);
 // it stores the tags for the note in note editor
 export const tagsWritable = writable([""]);
 // it stores the visibility of mask editor
-export const ioMaskEditorVisibleStore = writable(true);
+export const ioMaskEditorVisible = writable(true);
```

dae:
Just one comment left above; rest looks good.

dae:
Thanks for all your work on this Mani. I think it's about time we merge this in! :tada:  I expect to build beta 2 in around 2 weeks if everything goes smoothly. If any issues/tweaks are required, follow-up discussions or PRs are welcome.

PR #2488, opened 2023-05-02
Prevent skewing of resized images in editor

kleinerpirat:
Issue explained here: https://forums.ankiweb.net/t/is-there-a-way-to-prevent-the-squish-skew-of-the-image-in-the-browser-editor/29925

### DIFF
--- a/ts/editable/editable-base.scss
+++ b/ts/editable/editable-base.scss
@@ -33,14 +33,14 @@ img:not(.mathjax) {
         :host-context(.shrink-image) & {
             max-width: var(--editor-default-max-width);
             max-height: var(--editor-default-max-height);
+            // prevent inline width/height from skewing aspect ratio
+            width: unset;
+            height: unset;
         }
     }
 
     &[data-editor-shrink="true"] {
         max-width: var(--editor-shrink-max-width);
         max-height: var(--editor-shrink-max-height);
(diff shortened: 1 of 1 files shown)


kleinerpirat:
We need to make a choice:

1. unset inline sizes for all images when shrinking is enabled.
This might produce unwanted side effects for people who use images that are smaller than the default max size.
2. unset inline size for images with the `data-editor-shrink` attribute (i.e. revert https://github.com/ankitects/anki/pull/2488/commits/f91155be0535e8f01f192441c4a73469f7a65a3f)
This isolates the fix for images resized with the Anki image handle, but keeps skewing manually edited images.

WDYT?

dae:
> This might annoy people who use images that are smaller than the default max size (e.g. symbols/logos)

I'm not sure what you mean here - are you talking about users who set a width greater than the actual image dimensions, to make it larger?

Pending an answer to that, A seems preferable, as there's a fair amount of decks floating around that have been adjusted via add-ons in the past.

kleinerpirat:
I meant that there's a possibility that someone uses inline size constraints smaller than the editor shrink constraint (e.g. `width="20px"`). Such images would be displayed bigger than intended if the width/height attribute is unset.

But perhaps I'm over-thinking this. If you're ok with this, we could merge it and see if any reports crop up.

PR #2489, opened 2023-05-03
Enchance reviewer with new settings

tatsumoto-ren:
A small patch that ports to vanilla Anki some features that are popular among users of addons. 

* Place grading keys on the home row. This should be more convenient than pressing 1,2,3,4. Less movement.
* Add "u" for undo, for the same reasons.
* An option to disable grading with space (and enter). It is possible to accidentally press Space twice when grading. It's also confusing when one key serves two different purposes at the same time.


dae:
Please bring up any features you are planning to add in the future before sending through a PR, to avoid any wasted work if they do not get approved.

> Place grading keys on the home row. This should be more convenient than pressing 1,2,3,4. Less movement.
> Add "u" for undo, for the same reasons.

I am ok with these. I know how strongly you feel about breaking add-ons though, so please bear in mind this will break any other add-ons that reference these keys. I found https://ankiweb.net/shared/info/1371005125 with a brief grep; I do not know if there are others.

> An option to disable grading with space (and enter). It is possible to accidentally press Space twice when grading. It's also confusing when one key serves two different purposes at the same time.

In the past the enter key used to perform this way, but it got lost at one point. I am ok with such an option, but a couple of minor issues:

- Instead of "Grade with space", maybe something like "Spacebar rates card" or "Spacebar also rates card"? "Rating" is used in other places; I don't think we use "grade". Open to other suggestions.
- The tab order in the .ui file needs updating, so that (shift)+tab moves in the correct order.

tatsumoto-ren:
>  so please bear in mind this will break any other add-ons that reference these keys. 

I think adding new shortcuts is relatively safe to do. Add-ons modify the default list of shortcuts, so any default shortcut can be masked/replaced. Additionally, there is `gui_hooks.state_shortcuts_will_change` that allows doing it without calling `wrap()`.

>  I found https://ankiweb.net/shared/info/1371005125 

Did it fail for you? On my machine it overrides the default shortcuts as expected.

> Please bring up any features you are planning to add in the future before sending through a PR, to avoid any wasted work if they do not get approved.

I made a topic on the forum a while ago, but I believe you already closed it. 

>  maybe something like "Spacebar rates card" or "Spacebar also rates card"?

Does "answer" make more sense than "rate"?

> The tab order in the .ui file needs updating, so that (shift)+tab moves in the correct order.

I will investigate this.

ftl/core/preferences.ftl
```diff
@@ -19,6 +19,7 @@ preferences-preferences = Preferences
 preferences-scheduling = Scheduling
 preferences-show-learning-cards-with-larger-steps = Show learning cards with larger steps before reviews
 preferences-show-next-review-time-above-answer = Show next review time above answer buttons
+preferences-spacebar-rates-card = Press Spacebar to rate card "Good"
```
dae, line 22: This might be a bit misleading - if the user has changed the focused answer button to eg. hard, space will not answer good. How about 'Spacebar also answers card'? It gives a hint that spacebar serves another purpose.

dae (commented):
> Did it fail for you? On my machine it overrides the default shortcuts as expected.

No, but if users grow to depend on the new shortcuts, they'll probably ask that add-on author to pick a new shortcut.

> I made a topic on the forum a while ago, but I believe you already closed it.

You can of course create a new thread. If you have a bunch of questions at once, it would be good to put them all in the same thread though.

> Does "answer" make more sense than "rate"?

That works for me too.

ftl/core/preferences.ftl
tatsumoto-ren: sounds good

dae:
Thanks tatsumoto, looks good.

PR #2490, opened 2023-05-08
Fix sync client ignoring directories

XeR:
Current implementation of the sync client does not properly send requests to a self-hosted sync server if the URL has directories.

If the self-hosted sync URL is `https://example.org/foo/bar/`, Anki is expected to send requests to `https://example.org/foo/bar/sync/hostKey`. Instead, it will discard the directories and wrongly send requests to `https://example.org/sync/hostKey`.

Fixes: e5d5d1d ("Fix panic with invalid sync server URL with port")


dae:
Thanks @XeR

PR #2493, opened 2023-05-11
Fix custom font preloading sometimes not working

hikaru-y:
This fixes an issue with my previous PR #2356 where custom font preloading would not work if both a reference to an external style sheet and a description of a custom font are present in a style element.


dae:
Thanks Hikaru - always nice to have bugs fixed before users even hit them :-)

PR #2495, opened 2023-05-12
Add pycmd to play all card audio in sequence

BtbN:
This adds a pycmd for playing all card audio in sequence.
The primary use for this is to add a "Play"-Button to Card templates, which just replays all the audio on the card.
As opposed to just one single clip.

It also allows for more complex Card-Side JavaScript, which can base the decision on whether to play or not to play the audio on more complex factors, like the entered answer or a bit of data from the corresponding Note.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -120,6 +120,7 @@ Ben Kerman <ben@kermanic.org>
 Euan Kemp <euank@euank.com>
 Kieran Black <kieranlblack@gmail.com>
 XeR <github.com/XeR>
+BtbN <btbn@btbn.de>
 
 ********************
 
--- a/qt/aqt/sound.py
+++ b/qt/aqt/sound.py
@@ -824,14 +824,21 @@ def av_refs_to_play_icons(text: str) -> str:
 
 
 def play_clicked_audio(pycmd: str, card: Card) -> None:
-    """eg. if pycmd is 'play:q:0', play the first audio on the question side."""
+    """eg. if pycmd is 'play:q:0', play the first audio on the question side.
+    Likewise, if it is 'play:a:all', play all card audio in sequence."""
     play, context, str_idx = pycmd.split(":")
-    idx = int(str_idx)
+    try:
+        idx = int(str_idx)
+    except ValueError:
+        idx = 0
     if context == "q":
         tags = card.question_av_tags()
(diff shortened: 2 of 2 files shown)


dae:
I'm afraid I'm not keen on this change - pycmd is an internal API not intended for use outside Anki. Users who want to replay audio already have access to the replay audio action (shortcut key r), and that action works across the different Anki platforms.

BtbN:
The background here is that the deck in question has meaning and reading cards.
And only on the reading cards should the audio play. On the Meaning-Sibling-Cards, it'd spoil the reading which is not quizzed on those cards.
Right now the only way to do that in the card JS is to first turn off auto-play and then iterate over the play-button HTML elements to get the number of audio elements, and then issue the relevant pycmds with a guessed delay.
Those cards are in the same deck, so the setting for auto-play is too broad.

Additionally it allows for the card JS to only play the reading audio automatically if the correct answer was typed in.

dae:
I'm not sure what you mean - the replay audio action replays what is currently shown on the front/back of the card, not all audio in the note, and autoplay will only play audio that it's included on the front or back of that card template. If what you are trying to accomplish is to show play icons but not have the audio autoplay on one card template, I'm afraid that's not something Anki currently tries to support, and the approach in this PR is not the best way to resolve this. We may add more sophisticated audio control via Javascript at some point in the future, but it's not at the top of the priority list at the moment I'm afraid.

BtbN:
Yes, the templates for all cards are identical, and are altered on the fly with a bunch of JavaScript on the template.
On the cards that teach the reading of a word, the audio should play.
On the sibling-card that teaches the meaning, audio should not play (unless the user explicitly requests it).

I guess a possible other solution for this would be to include a way to add audio-buttons to a card that do not add their audio to the autoplay list. Those would then be used on the meaning-template.
That would also enable the user to still have a working autoplay-audio setting for the reading cards.

I'll see if this can be implemented without too much fuss.

dae:
I'm afraid I'd prefer not to make changes to the way autoplay is handled at this time. Per-template or per-audio autoplay control is not something that is requested too frequently, and I'd prefer developer resources were focused on other areas at this time. A workarounds you could use for now is to put the audio file only on the card you want to autoplay, and omit it from the other card, or place it on the answer side.

PR #2496, opened 2023-05-13
Allow `k` key to flip cards during review

SpicyRicecaker:
I’m loving that Anki is getting native support for (h,j,k,l,u) bindings to assign a score to each card during reviews (thanks @tatsumoto-ren!). 

## The Missing Feature of the Current (h,j,k,l,u) Implementation
However, @tatsumoto-ren 's change only allows for the `k` key (used to give cards a "good" ease level) to be used during the `answer` state of reviewing a card. This means that if the user wants to 1) use the (h,j,k,l,u) keybindings and 2) give a card a "good" ease level (which I believe to be a majority of cards), they must first flip the card with `space` and then rate it with `k`. 

This makes the `k` key function differently than the `space` key currently does in the latest anki release, where `space + space` can be used to both flip the card and rate the card as "good". 

Thus having to use `space + k` to review a card would be both unintuitive (compared to the current `space + space` keybinding) and slow.

## What This PR Does
This PR allows the user to press the `k` key to flip the card while the card is in `question` state. This means that users can easily flip a card and rate it "good" by pressing `kk`. 

In the process, this commit also fixes a small bug introduced in @tatsumoto-ren's fef2844 commit, where unchecking the `Spacebar also answers card` in `Preferences > Review` also disabled the ability to rate cards using the `enter/return` key.

## Potential Issues With This Commit

This commit renames the `onEnterKey` callback function in the `Reviewer` class to `flip_or_rate_card`, a name that better befits a function called by the `space`, `enter/return`, and `k` keys. I am not aware of any hooks Anki generates from this function, but this could pose a backwards compatability issue.


SpicyRicecaker:
Oops, just realized that I can remove the whole function I added by passing in `self.onEnterKey` along with `k`. Will update PR.

tatsumoto-ren:
I never intended to reveal the back side with <kbd>k</kbd> because that would introduce the same problem people had with pressing <kbd>space</kbd> twice, which is that you can accidentally press the key twice and have to undo. Or worse, you won't notice and grade a card incorrectly. Note that previously, when it was possible to rate a card "good" only with the <kbd>3</kbd> key, the key wasn't used to reveal the back side. To make the behavior more apparent to the users, I would consider adding a new checkbox in Preferences to support this functionality, e.g. "✅press `k` to flip card". 

I am guessing it is intended that there's no distinction between <kbd>space</kbd> and <kbd>enter</kbd>, so not sure if it makes sense to introduce that distinction now. If there are add-ons that wrap `onEnterKey()`, they will break, so I suggest keeping it for compatibility and creating a separate function for another action.

SpicyRicecaker:
@tatsumoto-ren, thanks for looking over my pr!

I agree with all of the points you made. The current `3` key that rates cards as "good" does not flip cards. Therefore, the `k` key that rates cards as "good" should not flip cards by default either. I'll implement `k` flipping a card as a feature to be enabled through preferences.

I also agree that the `onEnterKey` function should be kept for backwards compatability: I'll create a new wrapper function for the `k` key instead.

I also think that it's intentional that there's no distinction between `space` and `enter`. However, if this is the case, then
1. the `Spacebar also answers card` checkbox label should be renamed to `Spacebar/enter also answers card` to avoid possible confusion users may have with their `enter` key not answering their current card
2. all instances of `spacebar` (e.g. in `spacebar_rates_card`) in the source code introduced in fef2844 should be renamed to `enter_key` to reflect the name of the default `onEnterKey` callback that we have now. 

Because the function that toggled the the `Spacebar also answers card` checkbox was named `spacebar_rates_card` while being called inside the `onEnterKey` callback function, I thought that a distinction *was* being made between the `enter` key and `spacebar`. I think these changes will decrease the confusion that future contributors may face.

tatsumoto-ren:
Looks good!

dae:
> This makes the k key function differently than the space key currently does in the latest anki release, where space + space can be used to both flip the card and rate the card as "good".

I'm afraid I'm not following the logic here. My understanding is that `k` is equivalent to `3`. Whether the user uses their left hand and number keys, or right hand and hjkl, their thumb will sit on the spacebar, and they can both reveal the answer and rate it good by pressing space twice, unless the option in the preferences to disable that behaviour has been toggled. You can't press `3` to flip the card and it's not something I've seen users ask for frequently, so I'm not sure why we should be making `k` behave differently.

ftl/core/preferences.ftl
```diff
@@ -19,7 +19,8 @@ preferences-preferences = Preferences
 preferences-scheduling = Scheduling
 preferences-show-learning-cards-with-larger-steps = Show learning cards with larger steps before reviews
 preferences-show-next-review-time-above-answer = Show next review time above answer buttons
-preferences-spacebar-rates-card = Spacebar also answers card
+preferences-enter-key-rates-card = Press Spacebar/Enter to answer card
```
dae, line 22: This is technically more correct, but I wonder whether there are many users who actually prefer enter over space, and if not, whether the more succinct description would be better. The enter shortcut mainly exists for legacy - at one point in the past, enter would reveal the answer but not rate the card.

SpicyRicecaker:
Hey @dae, 
Thanks for looking over my PR. 

## Logic for `k`

I agree that the logic I used in my initial comment for allowing `k` to flip cards doesn't make any sense, and I apologize for that. 

I just think `kk` to flip and rate cards would be a nice feature to have for Anki. `k`, as opposed to `3`, is located on the home row, so my middle finger is already there. Tapping my middle finger twice is less clunky than tapping my thumbs twice, and is more comfortable for long-term reviews. Thanks to @tatsumoto-ren's suggestions, I gated this behind a checkbox in preferences, so it's not the default behavior. 

## Readability

I personally almost never use the enter key to review cards, but I was surprised while reviewing when the enter key didn't work. Maybe we could make the preference both technically correct and succint?

What do you think about `Spacebar (or enter) also answers card` as opposed to the current `Spacebar also answers card` option in Anki? It'd be technically correct for the price of fewer words.

dae:
> I just think kk to flip and rate cards would be a nice feature to have for Anki.

It seems like a reasonable feature, but the problem is that if Anki included every feature someone would find useful, we'd end up with an (even more) unmanageable amount of options, and the code would become even more difficult to maintain. I'm afraid I think this is probably better done in an add-on for now, and if the add-on proves popular enough in the future, then you'll be in a better position to argue that it should be included in Anki, as @tatsumoto-ren did with his hjkl change.

> What do you think about `Spacebar (or enter) also answers card`

I prefer that over `spacebar/enter`. I personally still slightly prefer what we currently have, but if you feel the current wording caused you legitimate confusion then I guess it's best changed.

tatsumoto-ren:
@SpicyRicecaker the Flexible Grading add-on could include the features mentioned here, until the time to merge them into Anki comes.

SpicyRicecaker:
I removed the `kk` flips card feature, but kept the `Spacebar (or enter) also answers card` checkbox preference rename (and internal refactoring). 

I'm still leaning toward adding the `(or enter)`, as 
1. the [docs](https://docs.ankiweb.net/studying.html?highlight=space#keyboard-shortcuts) still say that either `Enter` or `Space` will show the answer (which a new or existing user may read)
2. it seems there is [some interest](https://www.reddit.com/r/Anki/comments/pvwn84/why_didnt_anyone_tell_me_that_i_can_use_enter/) for using the `Enter` key during reviews, if someone is using a numpad for example. 

I could imagine a scenario where someone flips the checkbox option, forgets about it, then tries to use the `enter` keys for reviews occasionally or because they're using a numpad, only for it not to work.

dae:
I'm ok with the change in label, as only 3 translators have translated it so far. I don't understand why you've changed all references from spacebar to enter in the code though; that makes the change larger than it needs to be, and will require users to change the setting in the next update if they have previously customized it. Why not stick with what we currently have, which is the more common case anyway?

SpicyRicecaker:
In the source code, the `onEnterKey` function is the single callback that handles both `space` and `enter` key presses during reviews. However, we use a flag named `spacebar_rates_card` to check if we should allow `space` **or** `enter` to rate cards.

In my earlier comment to Ren, I explained that this difference between the flag and function name is confusing, since as a contributor it made me think that **only** `spacebar` should be blocked from rating a card (when the actual intent and result was that it blocked **both** `space` and `enter` from rating a card). 

I then proposed that we rename `spacebar` to `enter_key` in the source code, since we would then be checking the `enter_key_rates_card` flag inside the `onEnterKey` callback, making the naming consistent and increasing readability.

I completely didn't consider that it 
> will require users to change the setting in the next update if they have previously customized it

so I'll roll back the change.

dae:
Thanks for rolling it back. To get this change into the next rc, I've made the change manually. For future PRs, please add yourself to the CONTRIBUTORS file so that CI can pass.

You could potentially rename onEnterKey to make things clearer, but you would need to add an alias to the old name, and it would still break things if an add-on wrapped the old name alias, so I'm not sure it's worth it.

PR #2497, opened 2023-05-14
Implement non elevated installer

### LINKED ISSUE #1597: Can't install on standard user
Hi, I can't install Anki on the standard user.
Please make an installation for the current user/all users.
image
image
Thanks.

mgrottenthaler:
See also <https://forums.ankiweb.net/t/windows-install-non-privileged-to-appdata/29725/2>

1. Installation will now not anymore require elevation to admin (UAC)
2. Changed default install location to `%LOCALAPPDATA%\Programs\Anki`
3. Changed registry keys from HKLM to HKCU (so that non admins can write)
    * to do so: also adapted the macros from "fileassoc.nsh"
4. Implemented "upgrade"/"migrate" from elevated install to non elevated

Closes #1597


dae:
Thanks for looking into this Martin! Please add yourself to the CONTRIBUTORS file as part of this PR, and then I can give this a review.

mgrottenthaler:
Thanks! I did that, pipeline is successful now.

One thing that I forgot to mention up there: I am calling the original `uninstall.exe` if an old elevated version (basically checking if there is a key in HKLM) is found.

qt/bundle/win/anki.template.nsi
```diff
@@ -127,10 +130,43 @@ FunctionEnd
...
-  SetShellVarContext all
+  SetShellVarContext current
+
+  ; "Upgrade" from elevated anki
+  ReadRegStr $0 HKLM "Software\WOW6432Node\Anki" "Install_Dir64"
+  ${IF} $0 != ""
+      ; old value exists, we want to call uninstall with silent parameter
+      ; the wait does not work, I suspect this is because of the UAC prompt
+      ; it is still better to use the wait version in my opinion
+      ExecShellWait "open" "$0\uninstall.exe /S"
```
dae, line 168: Unfortunately this not only seems to not wait, it also seems to fail to uninstall the previous version, so the user ends up with two Anki icons on their desktop, one for the old version and one for the new one.

I experimented with using ExecWait and https://stackoverflow.com/questions/24595887/waiting-for-nsis-uninstaller-to-finish-in-nsis-installer-either-fails-or-the-uni, but didn't have any luck - maybe it's unable to request elevated permissions?

If you can't think of a way to solve this, maybe another approach we could use is to detect the existing registry key, and if found, abort the install, asking the user to uninstall their previous Anki version first? It's a bit more hassle for them, but is better than having multiple copies floating around.

mgrottenthaler: Are you sure it was not uninstalled? In my tests it worked. Uninstall is silent, so you will have to check the folders/regkeys.

`ExecWait` does indeed not work, just as you said, it doesn't work with elevated commands. `ExecShellWait` could execute the `uninstall.exe` though. Only the "wait" part didn't work as expected. I think it's because of the UAC prompt, which most probably already returns something, after you press Yes/No, and doesn't wait for the uninstaller to finish. Therefore install of the new anki and uninstalling of the old anki is done in parallel. They do not share any paths/regkeys, so this shouldn't lead to any problems.

But yes, I agree, it's not the cleanest solution. Therefore I suggest the following:

1. if regkey is found
    1. display message with something like "Old version of anki found. Please uninstall it first and execute this installer again. Starting uninstaller now..."
    2. execute uninstaller.exe without silent
    3. Abort installer.
2. User executes installer again the if at `1.` is not executed

This way, we can be sure that the uninstaller finished successfully.

mgrottenthaler: Another issue that should be addressed, before this is merged: this change might have an impact on package managers like chocolatey. Are there any others that you are aware of? The maintainers should most probably be informed.

mgrottenthaler: I implemented this in 08fd98c4b638346e716826ed03330a5daea37423

I works now like described above. Some possible changes would be:

* call `uninstall.exe` with the `/S` flag, the user would not see the uninstall wizard
* sleep for a certain amount of seconds then check if the uninstall is done (this seems a little bit hacky though)

dae: > Are you sure it was not uninstalled? In my tests it worked. Uninstall is silent, so you will have to check the folders/regkeys.

Yep, I tested again with the same results - no removed files/folders and reg entry unchanged. Your latest code seems to work correctly though, which is great. I think it's fine to ask the user to re-run the installer after uninstalling; it's not much trouble, and the alternative solutions are hacky.

> this change might have an impact on package managers like chocolatey

If they're just invoking the official installer, users could presumably work around any issues by removing the package and adding it again I guess?

qt/bundle/win/anki.template.nsi
```diff
@@ -127,10 +130,43 @@ FunctionEnd
...
 Section ""
 
-  SetShellVarContext all
+  SetShellVarContext current
+
+  ; "Upgrade" from elevated anki
+  ReadRegStr $0 HKLM "Software\WOW6432Node\Anki" "Install_Dir64"
+  ${IF} $0 != ""
+      ; old value exists, we want to inform the user that a manual uninstall is required first and then start the uninstall.exe
+      MessageBox MB_ICONEXCLAMATION|MB_OK "Cannot automatically upgrade old Anki version, please uninstall manually and then start this installer again.$\r$\nStarting uninstaller now..."
```
dae, line 166:
```suggestion
      MessageBox MB_ICONEXCLAMATION|MB_OK "A previous Anki version needs to be uninstalled first. After uninstallation completes, please run this installer again."
```

dae:
I've given this some testing and it seems to work well now - thanks Martin! Just a minor wording suggestion, and then I can merge this in.

qt/bundle/win/anki.template.nsi
mgrottenthaler: > Yep, I tested again with the same results - no removed files/folders and reg entry unchanged. Your latest code seems to work correctly though, which is great. I think it's fine to ask the user to re-run the installer after uninstalling; it's not much trouble, and the alternative solutions are hacky.

perfect! I commited your suggestion

> If they're just invoking the official installer, users could presumably work around any issues by removing the package and adding it again I guess?

Yes most probably your are right. I will still send them a message to inform them. I myself am a chocolatey user so I have some personal involvement in this ;)

dae:
Thanks Martin! :rocket:

PR #2501, opened 2023-05-16
Fix null error on window resizing

Aquafina-water-bottle:
This fixes the issue I reported here: https://forums.ankiweb.net/t/29185
Note that it is not actually a Windows only error; I was able to properly reproduce it on my standard Linux setup. I didn't realize I was on 2.1.60 until today, and when updating to 2.1.63, the error was indeed logged as expected.

Related screenshot from the AnkiWebView Inspector add-on:
image


dae:
Thanks for looking into this Austin!

PR #2502, opened 2023-05-16
Allow the user to configure answer keys

tatsumoto-ren:
This PR adds a group box in Preferences > Review which lets the user change answer keys. By default answer keys are set to numbers: [1,2,3,4].

screenshot-2023-05-16-09-08-38

Some translations are currently not implemented.

### DIFF
--- a/ftl/core/preferences.ftl
+++ b/ftl/core/preferences.ftl
@@ -72,6 +72,7 @@ preferences-import-export = Import/Export
 preferences-network-timeout = Network timeout
 preferences-reset-window-sizes = Reset Window Sizes
 preferences-reset-window-sizes-complete = Window sizes and locations have been reset.
+preferences-shortcut-placeholder = Enter an unused shortcut key, or leave empty to disable.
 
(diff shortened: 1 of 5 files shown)


dae:
The reason I suggested an option for now is that I was keen to get a small regression fix out in an update with minimal chances of regressions, and this has some more moving parts. It's not an unreasonable approach though; I will give it a bit more thought and get back to you.

tatsumoto-ren:
Some people like `hjkl`, some people like `jkl;` instead, so I think a request to make the keys configurable would arise eventually.

vorand:
Many users may benefit from this feature because the "Customize Keyboard Shortcuts" add-on, which many users depend on to configure jkl; as answer keys, is incompatible with version 2.1.63.

dae:
The code is well-written, and solves the conflicts issue. There are some other PRs that will benefit from at least a few days of testing as well, so I think you're right, it's probably best to jump straight to this implementation and skip the option stopgap. Thanks for the quick fix.

PR #2503, opened 2023-05-16
Add zoom out and zoom in shortcuts

tatsumoto-ren:
Added missing keyboard shortcuts for zooming in and out, which can be handy while reviewing cards.

### DIFF
--- a/qt/aqt/forms/main.ui
+++ b/qt/aqt/forms/main.ui
@@ -258,10 +258,16 @@
    <property name="text">
     <string>qt_accel_zoom_in</string>
    </property>
+   <property name="shortcut">
+    <string notr="true">Ctrl+=</string>
+   </property>
   </action>
   <action name="actionZoomOut">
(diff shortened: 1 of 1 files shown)


dae:
Previous discussion: https://github.com/ankitects/anki/pull/1668#issuecomment-1043191227

tatsumoto-ren:
> Previous discussion: #1668 (comment)

If I understand correctly, something else is conflicting with ctrl+plus and ctrl+minus?

dae:
Two issues as far as I understand it: one of the shortcuts conflicts with the subscript or superscript shortcut in the browser, and the standard shortcuts are hard/impossible to type on some keyboard layouts.

tatsumoto-ren:
Will close for now, as it seems there's an issue preventing the hotkeys from being available by default. As a workaround, the hotkeys can be enabled by installing one of the existing add-ons from AnkiWeb.

PR #2508, opened 2023-05-18
Fix page scrolling when input box loses focus

hikaru-y:
https://forums.ankiweb.net/t/back-side-unexpectedly-aligns-to-center-without-hr-function-enabled/30396/2


hikaru-y:
I've changed the approach a bit, because ab5238b did not address a case where the page content exceeded the width of the viewport.

dae:
Thanks Hikaru! Hopefully QtWebEngine will eventually fix this on their end.

PR #2509, opened 2023-05-19
Ensure there's no duplicate shortcuts after running state_shortcuts_will_change

tatsumoto-ren:
There are add-ons that `extend` the list of shortcuts but forget to check if the new keys are already occupied. Such add-ons will break every time Anki changes its default shortcuts. 

Some affected add-ons: 

* https://ankiweb.net/shared/info/850294128
* https://ankiweb.net/shared/info/1844908621 ([code](https://github.com/glutanimate/hint-hotkeys/blob/41054cdd509ec91acf95260e8a24ba7dab1702e7/src/hint_hotkeys/_addon.py#L74))

Instead of forcing the authors to release fixes, we can filter the list of shortcrust for them, after they modify it. If this PR is merged, any default shortcuts should always be overridden by add-ons. And Anki will be able to freely modify any of its shortcuts in the future without worrying about add-ons.

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -1126,6 +1126,9 @@ title="{}" {}>{}</button>""".format(
         gui_hooks.state_shortcuts_will_change(self.state, shortcuts)
         # legacy hook
         runHook(f"{self.state}StateShortcuts", shortcuts)
+        # remove duplicate shortcuts (possibly added by add-ons)
+        # by filtering them through a dictionary.
+        shortcuts = list(dict(shortcuts).items())
         self.stateShortcuts = self.applyShortcuts(shortcuts)
 
     def clearStateShortcuts(self) -> None:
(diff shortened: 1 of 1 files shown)


qt/aqt/main.py
```diff
@@ -1112,12 +1112,21 @@ def setupKeys(self) -> None:
         self.applyShortcuts(globalShortcuts)
         self.stateShortcuts: list[QShortcut] = []
 
+    def _normalize_shortcuts(
+        self, shortcuts: Sequence[tuple[str, Callable]]
+    ) -> Sequence[tuple[QKeySequence, Callable]]:
+        """
+        Remove duplicate shortcuts (possibly added by add-ons)
+        by normalizing them and filtering through a dictionary.
```
dae, line 1120:
```suggestion
        by normalizing them and filtering through a dictionary.
        The last duplicate shortcut wins, so add-ons will override
        standard shortcuts if they append to the shortcut list.
```

dae:
Seems reasonable, thanks.

PR #2510, opened 2023-05-20
Expose some collection methods already public in rslib backend

### LINKED ISSUE #2494: rslib: how can I update a note?
Hi, I'm trying to update some notes with a Rust program I'm writing. I've got adding notes working, but I'm still pretty new to Rust and am a little confused why [this update_notes function](https://github.com/ankitects/anki/blob/802f10ea13f3e154772f5b75255c09f933f3f30a/rslib/src/notes/mod.rs#L354) isn't public. Is there a better way to update a note, or should there be a public function here?

joelkoen:
Closes #2494

### DIFF
--- a/rslib/src/notes/mod.rs
+++ b/rslib/src/notes/mod.rs
@@ -350,12 +350,11 @@ impl Collection {
         self.set_current_notetype_id(note.notetype_id)
     }
 
-    #[cfg(test)]
-    pub(crate) fn update_note(&mut self, note: &mut Note) -> Result<OpOutput<()>> {
+    pub fn update_note(&mut self, note: &mut Note) -> Result<OpOutput<()>> {
         self.transact(Op::UpdateNote, |col| col.update_note_inner(note))
     }
 
-    pub(crate) fn update_notes_maybe_undoable(
+    pub fn update_notes_maybe_undoable(
         &mut self,
         notes: Vec<Note>,
         undoable: bool,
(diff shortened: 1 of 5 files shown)


rslib/src/notes/mod.rs
```diff
@@ -350,12 +350,11 @@ impl Collection {
...
     }
 
-    #[cfg(test)]
-    pub(crate) fn update_note(&mut self, note: &mut Note) -> Result<OpOutput<()>> {
+    pub fn update_note(&mut self, note: &mut Note) -> Result<OpOutput<()>> {
         self.transact(Op::UpdateNote, |col| col.update_note_inner(note))
     }
 
-    pub(crate) fn update_notes_maybe_undoable(
+    pub fn update_notes_maybe_undoable(
```
dae, line 357: Might be better to leave this one private unless there was a particular use-case you had in mind. The backend only exports this because some legacy Python/add-on code depends on the old system of deferring transactions, but that should not be an issue with projects using the Rust API directly.

dae (commented):
Thanks Joel! Just one comment.

rslib/src/notes/mod.rs
dae: To get the other changes into the next rc, I'll revert this line without waiting for feedback, but feedback is still welcome.

dae, line 357:
```suggestion
    pub(crate) fn update_notes_maybe_undoable(
```

PR #2513, opened 2023-05-23
Various changes to I/O handling

dae:
@krmanik @glutanimate would appreciate your thoughts

While working on the misaligned occlusions fix, I noticed a couple of issues:

- The 1MB limit on canvas size gives us an effective resolution of about 1300x765, which leads to the occlusions appearing somewhat blurry on a 4k monitor. 
- Because we're storing positions relative to that capped 1MB limit, we can't easily increase the limit, as occlusions would then be in the wrong position unless we scaled them appropriately.
- A similar issue exists with image resizes - if the user uses an external program to alter the size of their images (eg shrinking them while keeping the same dimensions), the occlusions will be misaligned.

This makes me think we would be better off storing the positions as a ratio of the image size, instead of using absolute dimensions. I seem to recall when we discussed this previously, I voiced concerns about pixel precision. I'm somewhat less concerned about that now, as we're already scaling positions to keep within the canvas limit, and larger images will tend to be scaled down to fit the screen anyway. I believe Mani also mentioned that it would be more work to use ratios, but now that we only offer 3 tools, it doesn't appear to be too complicated. I've mocked up an example with just the rectangle too so far. What do you think?


ts/image-occlusion/generate.ts
```diff
@@ -40,22 +37,34 @@ export function generate(hideInactive: boolean): { occlusionCloze: string; noteC
...
         }
     });
 
     occlusionCloze += clozeData;
     return { occlusionCloze, noteCount };
 }
 
-const getCloze = (object, index, relativePos, hideInactive): string => {
+function floatToDisplay(number: number): string {
+    return number.toFixed(6).replace(/^0+|0+$/, "");
```
glutanimate, line 51: The 0-tail-trimming did not seem to work for me here, but switching the RegEx to global seemed to do the trick

```suggestion
    return number.toFixed(6).replace(/^0+|0+$/g, "");
```

ts/image-occlusion/generate.ts
```diff
@@ -40,22 +37,34 @@ export function generate(hideInactive: boolean): { occlusionCloze: string; noteC
...
+}
+
+function xRatio(canvas: HTMLCanvasElement, x: number): string {
```
glutanimate, line 54: Incredibly minor and maybe not an actual improvement, but from a testing & DRY standpoint, would it perhaps make sense to combine the ratio functions for x and y into one (both here and when restoring to absolute values) and passing in the canvas width/height directly? I.e. a single `toPixelRatio` and `fromPixelRatio` function?

glutanimate:
:+1: on switching to relative coordinates, it seems like the sensible choice for all the reasons you mentioned. I would also agree that pixel precision should not be a major concern. I would imagine that `canvas`'  sub-pixel rendering for floating-point coords is another inherent factor of imprecision that's already present anyways.

Gave the branch a quick test and drawing and rendering rectangles seems to work without any issues. Left a couple of super minor suggestions on the changes.

Regarding the canvas size limit: Hadn't noticed that before, but you're right, I think we should definitely raise it if possible. Throwing a few high DPI screenshots at IO, the limited mask resolution is definitely noticeable, particularly when zooming in to draw shapes more precisely.

I don't remember, is there any particular reason we capped this at 1MP? If I understand correctly, Safari limits canvas sizes to 4096x4096, right?

ts/image-occlusion/generate.ts
dae: Nice catch - coming from a Python background, JS's regex handling always trips me pu.

dae:
Older iPhones (that are no longer supported by AnkiMobile) may have had smaller limits, but I think everything that is currently supported will work with 16M. Mani may have just been being conservative here. I might do a very brief benchmark to confirm bumping the number doesn't have a huge impact on loading time.

ts/image-occlusion/generate.ts
dae: I may be missing something, but I'm not sure how this would help. To go from pixels-within-canvas to ratio-within-canvas, we need to divide by either width or height depending on the axis. And same for the other direction - we need to know which axis we're operating on to determine the appropriate pixel.

glutanimate: Sorry, I think my comment was a bit ambiguous. I was mostly musing about whether it would maybe make sense to pass the width or height in directly into a couple of common relative <-> absolute conversion functions, i.e. along the lines of:

```ts
function toAxisRatio(part: number, base: number): string {
     return floatToDisplay(part / base);
}

function fromAxisRatio(ratio: string, base: number): string {
     return parseFloat(ratio) * base;
}

// and then e.g. 

clozeData += `:width=${toAxisRatio(obJson.width, canvas.width)}`;

const width = fromAxisRatio(cloze.dataset.width!, canvas.width);
```

Mostly coming out of a place of playing around with the functions in the console before trying the PR and thinking how, if we add tests for them, passing the width & height in directly could simplify things vs. constructing canvas inputs.

But yeah, a tiny improvement if any at all and I feel a bit bad about wasting your time on this

dae: I don't have particularly strong feelings here, but given that the current implementation saves about 5-6 characters per line, I slightly prefer it over passing in explicit width/height each time.

dae:
@glutanimate @krmanik I've made some further changes, since it will be difficult to make them once people start relying on this. Feedback welcome! Only thing remaining is bumping the the canvas size limit. One thing to investigate is whether we should cap to min(16M, max(viewport, image size)), so that the occlusions don't look blurry when placed on a low-res image.

dae:
Looks nice and sharp on Retina displays now. On a 4k/Linux machine here, devicePixelRatio is 1, and with a test image that's about 1024 pixels width, the occlusions don't look very sharp. Maybe we should be just focusing on viewport size? Will experiment some more.

dae:
Switching to the container size has improved things on my Linux machine, and it seems to work well for me across the devices I tested with.

dae:
Merging this in before a new rc, but feedback when you have a chance would still be welcome :-)

PR #2515, opened 2023-05-23
Make Anki's MathJax typesetting awaitable

nihil-admirari:
`MathJax.startup.promise` must be reassigned to make downstream code await not only MathJax's initial typesetting, but Anki's typesetting as well. See e.g. the following block in [MathJax's documentation](https://docs.mathjax.org/en/latest/web/typeset.html)

```js
function typeset(code) {
  MathJax.startup.promise = MathJax.startup.promise
    .then(() => MathJax.typesetPromise(code()))
    .catch((err) => console.log('Typeset failed: ' + err.message));
  return MathJax.startup.promise;
}
```

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -124,6 +124,7 @@ mgrottenthaler <github.com/mgrottenthaler>
 Austin Siew <github.com/Aquafina-water-bottle>
 Joel Koen <mail@joelkoen.com>
 Christopher Woggon <christopher.woggon@gmail.com>
+Nil Admirari <https://github.com/nihil-admirari>
 
 ********************
 
--- a/ts/reviewer/index.ts
+++ b/ts/reviewer/index.ts
@@ -147,14 +147,14 @@ export async function _updateQA(
     bridgeCommand("updateToolbar");
 
     // wait for mathjax to ready
-    await MathJax.startup.promise
+    await (MathJax.startup.promise = MathJax.startup.promise
         .then(() => {
             // clear MathJax buffers from previous typesets
             MathJax.typesetClear();
 
             return MathJax.typesetPromise([qa]);
         })
-        .catch(renderError("MathJax"));
+        .catch(renderError("MathJax")));
 
     qa.style.opacity = "1";
 


dae:
Would you mind explaining in more detail what you're trying to accomplish? Presumably you want to use this in your card template?

nihil-admirari:
`modern-cloze-overlapper` [waits on MathJax](https://github.com/nihil-admirari/modern-cloze-overlapper/blob/master/_cloze-overlapper.mjs#L289) before re-typesetting all the clozes, including clozes in MathJax.

And since you've asked what I'm doing: I've noticed that Anki apparently got native [Image Occlusion](https://github.com/ankitects/anki/commit/7686cb8de8a27a2d29eea542e343c8ebfa59c95c). Various cloze add-ons are no less popular. How about updating Anki's cloze filter with all the functionality of these add-ons, e.g.: `{{cloze before=1 after=1 only-context etc.::...}}`?

dae:
Mutating an existing promise provided by the library feels a bit unclean to me - it's not clear that awaiting startup.promise actually awaits rendering as well. Can't you just append an entry to onShownHook?

> {{cloze before=1 after=1 only-context etc.::...}}

I'm not familiar with the add-on. What are these supposed to do?

nihil-admirari:
> Mutating an existing promise provided by the library feels a bit unclean to me

Seems like that's the recommended way to go.

> Can't you just append an entry to onShownHook?

I was unable to find any docs on `onShownHook`. Can you please provide a link?

> I'm not familiar with the add-on. What are these supposed to do?

Original [Cloze Overlapper](https://github.com/glutanimate/cloze-overlapper) was rewritten in [pure JS](https://github.com/michalrus/anki-simple-cloze-overlapper) and then [rewritten again](https://github.com/nihil-admirari/modern-cloze-overlapper). The second repo mentioned has an animation showing what the add-on does: https://github.com/michalrus/anki-simple-cloze-overlapper/blob/main/screen-recording.gif.

nihil-admirari:
> Can't you just append an entry to onShownHook?

If you've meant Anki's [`onShownHook`](https://addon-docs.ankiweb.net/reviewer-javascript.html), then I simply can't use it: [modern-cloze-overlapper](https://github.com/nihil-admirari/modern-cloze-overlapper) is not an Anki add-on, it an ordinary JS module loaded from the media folder by the card's template. `onShownHook` is available, but appending to it does nothing at all.

dae:
onShownHook seems to work for me. The following in front template:

```
{{Front}}
<script>
onShownHook.push(() => { console.log("called after mathjax"); })
</script>
```

with the following patch:

```diff
diff --git a/ts/reviewer/index.ts b/ts/reviewer/index.ts
index afa10779c..8512c96c7 100644
--- a/ts/reviewer/index.ts
+++ b/ts/reviewer/index.ts
@@ -152,7 +152,9 @@ export async function _updateQA(
             // clear MathJax buffers from previous typesets
             MathJax.typesetClear();
 
-            return MathJax.typesetPromise([qa]);
+            return MathJax.typesetPromise([qa]).then(() => {
+                console.log("mathjax done");
+            });
         })
         .catch(renderError("MathJax"));
 
```

yields

```
JS info /_anki/js/reviewer.js:16963 mathjax done
JS info :2 called after mathjax
```

nihil-admirari:
> onShownHook seems to work for me.

I was calling it from `<script type="module">`. Plain `<script>` does work.

[`await _runHook(onShownHook)`](https://github.com/ankitects/anki/blob/main/ts/reviewer/index.ts#L161) is called after MathJax typesetting was awaited, which means that adding to `onShownHook` is indeed a viable alternative.

dae:
Please note that the different clients all behave differently at the moment. AnkiMobile currently doesn't export that hook, and I don't think AnkiDroid does either.

The next desktop release will allow you to import it from a new location, and I will update AnkiMobile so it also works at the new location. You'll need to use code like this:

```
require("anki/reviewer").onShownHook.push(...)
```

require is a symbol that is available at runtime.

Since older clients will be either missing require, anki/reviewer, or that hook, your code should be prepared to display an error when the template is used on an older client.

Also, you may want to consider whether onUpdateHook is more appropriate, which should let you update the text prior to MathJax being rendered.

nihil-admirari:
> I don't think AnkiDroid does either.

It does, I've already checked: https://github.com/ankidroid/Anki-Android/blob/main/AnkiDroid/src/main/assets/scripts/card.js#L249.

> require("anki/reviewer")

Maybe standard JS modules would be better? It's 2023 after all.

dae:
Thanks, I didn't notice Henrik had added that.

require() is used because the API exports are dynamically created and don't correspond to files on disk. We also bundle JS source together for releases, as lots of separate files is harder to manage, and does not perform well when there's latency between the frontend and backend.

nihil-admirari:
Does Anki Mobile support `onUpdateHook` now? https://github.com/nihil-admirari/modern-cloze-overlapper/issues/3 errors with
```
ReferenceError: Can't find variable: onUpdateHook
```

nihil-admirari:
Also [`.mjs` files are served as `application/octet-stream`](https://github.com/nihil-admirari/modern-cloze-overlapper/issues/3#issuecomment-1817705213), not as [`text/javascript`](https://html.spec.whatwg.org/multipage/scripting.html#scriptingLanguages:javascript-mime-type).

dae:
Sorry, exposing it via require() proved to be more complicated than first expected, but I thought I'd exposed it as a global. Will look into that, and fix the content type issue in the next update.

nihil-admirari:
Neither `require` nor `onUpdateHook` are available in AnkiWeb either. Worst of all: on AnkiWeb script runs before MathJax is fully initialised (`MathJax.startup.promise` is undefined), which makes it impossible to await MathJax initialisation at all. The latter may be a [Firefox special](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script), I haven't tested in Chrome or WebKit:

> Starting in Firefox 4, inserting `<script>` elements that have been created by calling `document.createElement("script")` no longer enforces execution in insertion order. This change lets Firefox properly abide by the specification. To make script-inserted external scripts execute in their insertion order, set `.async=false` on them.

dae:
anki.onUpdateHook will be available in the next AnkiMobile beta - please let me know if you'd like to try it. https://docs.ankimobile.net/beta-testing.html

I'm afraid getting this working with AnkiWeb is not a high priority at the moment.

nihil-admirari:
@AlphaStrioleno, would you like to test `modern-cloze-overlapper` without workarounds, i.e. an older commit: https://github.com/nihil-admirari/modern-cloze-overlapper/blob/5e036f6c6a44021d3eb49538cb89cefb92205fa5/_cloze-overlapper.mjs?

PR #2522, opened 2023-05-26
Making phrasing more clear in dialog which lets user choose a deck to…

chrissy0:
I accidentally pushed O and was faced with a message that didn't fully make sense to me. I suspected I was actually moving cards to a different deck. In reality, I was just asked which deck I would like to display options for. I have made this more clear for other users who might not know what pressing O does.


dae:
Thanks!

PR #2523, opened 2023-05-26
 Add add-on installation hooks 

abdnh:
A follow-up to #2518

This adds the `addons_manager_will_install_addon` and `addons_manager_did_install_addon` hooks. My main use case is to prevent update errors caused by the add-on holding a DB connection to some file in the add-on folder.  Example usage:
```python
def on_addon_manager_will_install_addon(manager, module):
    if module == manager.addonFromModule(__name__):
        db.close()


def on_addon_manager_did_install_addon(manager, module):
    if module == manager.addonFromModule(__name__):
        db.init()


gui_hooks.addon_manager_will_install_addon.append(
    on_addon_manager_will_install_addon
)

gui_hooks.addon_manager_did_install_addon.append(
    on_addon_manager_did_install_addon
)
```

Another common source of installation errors are open .pyd files coming from bundled dependencies. The AnkiHub add-on for example patches `AddonManager.deleteAddon` twice to work around similar errors. (I probably have a bunch of more add-ons suffering from this issue too).

dc066bb fixes a crash in .deleteAddon() when run from the update routine and OSError is thrown there.

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -452,10 +452,11 @@ class AddonManager:
         base = self.addonsFolder(module)
         if os.path.exists(base):
             self.backupUserFiles(module)
-            if not self.deleteAddon(module):
+            try:
+                self.deleteAddon(module)
+            except Exception:
                 self.restoreUserFiles(module)
-                return
-
+                raise
         os.mkdir(base)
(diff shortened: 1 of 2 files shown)


dae:
Thanks Abdo!

PR #2527, opened 2023-05-29
Default to current deck in csv import if settings allow it

dae:
Reuses defaults_for_adding(). In the future we might also want to update the last deck/notetype on successful completion, if entries weren't specified in the file.

https://forums.ankiweb.net/t/importing-new-notes-to-wrong-deck-in-anki-2-1-63/30598

@RumovZ mind taking a quick look to see if this looks fine to you?

### DIFF
--- a/rslib/src/import_export/text/csv/metadata.rs
+++ b/rslib/src/import_export/text/csv/metadata.rs
@@ -42,10 +42,11 @@ impl Collection {
         path: &str,
         delimiter: Option<Delimiter>,
         notetype_id: Option<NotetypeId>,
+        deck_id: Option<DeckId>,
         is_html: Option<bool>,
     ) -> Result<CsvMetadata> {
         let mut reader = open_file(path)?;
-        self.get_reader_metadata(&mut reader, delimiter, notetype_id, is_html)
+        self.get_reader_metadata(&mut reader, delimiter, notetype_id, deck_id, is_html)
     }
 
     fn get_reader_metadata(
@@ -53,6 +54,7 @@ impl Collection {
         mut reader: impl Read + Seek,
         delimiter: Option<Delimiter>,
         notetype_id: Option<NotetypeId>,
+        deck_id: Option<DeckId>,
         is_html: Option<bool>,
     ) -> Result<CsvMetadata> {
         let mut metadata = CsvMetadata::from_config(self);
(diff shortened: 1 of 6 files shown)


rslib/src/import_export/text/csv/metadata.rs
```diff
@@ -176,36 +177,20 @@ impl Collection {
...
+        metadata: &mut crate::pb::import_export::CsvMetadata,
         notetype_id: Option<NotetypeId>,
+        deck_id: Option<DeckId>,
     ) -> Result<()> {
-        if let Some(ntid) = notetype_id {
-            metadata.notetype = Some(CsvNotetype::new_global(ntid));
-        } else if metadata.notetype.is_none() {
-            metadata.notetype = Some(CsvNotetype::new_global(self.fallback_notetype_id()?));
+        let defaults = self.defaults_for_adding(DeckId(0))?;
+        if metadata.notetype.is_none() {
```
RumovZ, line 187: This should execute unconditionally if `notetype_id` is provided, or the user won't be able to change the notetype if one is specified in the imported file.

rslib/src/import_export/text/csv/metadata.rs
```diff
@@ -176,36 +177,20 @@ impl Collection {
...
+            metadata.notetype = Some(CsvNotetype::new_global(
+                notetype_id.unwrap_or(defaults.notetype_id),
+            ));
         }
-        Ok(())
-    }
-
-    fn maybe_set_fallback_deck(&mut self, metadata: &mut CsvMetadata) -> Result<()> {
         if metadata.deck.is_none() {
```
RumovZ, line 192: Same as above.

ts/import-csv/ImportCsvPage.svelte
```diff
@@ -56,17 +57,25 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
     );
-    $: getCsvMetadata(path, delimiter, undefined, isHtml).then((meta) => {
+    $: getCsvMetadata(path, delimiter, undefined, undefined, isHtml).then((meta) => {
         columnLabels = meta.columnLabels;
         preview = meta.preview;
     });
-    $: if (globalNotetype?.id !== lastNotetypeId || delimiter !== lastDelimeter) {
+    $: if (
+        globalNotetype?.id !== lastNotetypeId ||
+        lastDeckId !== deckId ||
```
RumovZ, line 66: Is it necessary to rerun `getCsvMetadata()` after the deck has changed? I think it doesn't have any influence on other settings or the preview, does it?

RumovZ (changes requested):

ts/import-csv/ImportCsvPage.svelte
dae: I think you're right; I incorrectly assumed that the deck in the returned metadata would be used when importing the file, so I thought it needed to be updated.

dae:
Thanks for the quick feedback Rumo, much appreciated. Do my changes address your concerns?

RumovZ (approved):
Looks good to me now. Good idea to add that documentation!

PR #2530, opened 2023-05-30
change window name when reviewing or deck screen

sudomain:
Caused by my [forum post](https://forums.ankiweb.net/t/feature-request-when-studying-a-deck-change-window-title-to-studying-deck-name/30736), I need Anki to have a different window title when studying a deck than when the deck overview screen is shown. Currently the title is "USERNAME - Anki" for both.

Possible further aesthetic enhancements would be:
1. having the deck name that's being studied in the title
2. number of cards and/or ETA left to study in the title
3. Having a different window title than "USERNAME - Anki" on the study/deck overview screen (not sure what the name is for this, but it's the screen that shows after selecting a deck but before pressing "study now")

I didn't attempt these things in case they were changes are too large (or too computationally expensive). I only needed a different window title to target Anki using an xdotool script.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -124,6 +124,7 @@ mgrottenthaler <github.com/mgrottenthaler>
 Austin Siew <github.com/Aquafina-water-bottle>
 Joel Koen <mail@joelkoen.com>
 Christopher Woggon <christopher.woggon@gmail.com>
+Sudomain <sudomain1@gmail.com>
 
 ********************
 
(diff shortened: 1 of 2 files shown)


dae:
I'm afraid this seems like a niche use case, and it could potentially break other usages of external tools that expect the profile name to be found in the title. I'd recommend you tackle this in an add-on instead, either updating that shortcut add-on to add support for the rebuild key, or by updating the title in the add-on.

sudomain:
Understood. Additionally, I found an add-on, [deck name in title](https://ankiweb.net/shared/info/699175524), which changes the window title. If that add-on had "window" in the name, it would've saved me a couple hours looking at Anki's code 😂. Thanks anyway

PR #2531, opened 2023-06-01
Fix toggleable editor button styles

abdnh:
`toggleEditorButton` was missing.


abdnh:
The `rightside` and `disables` args are also partially broken. It'd be nice to get `disables` working again. Will look into that in a future PR (hopefully it's a simple fix).

PR #2532, opened 2023-06-01
Disable add-on buttons when no field is focused by default

abdnh:
https://github.com/ankitects/anki/pull/2531#issuecomment-1572792476


dae:
Thanks Abdo!

Issue #2525, opened 2023-05-26
One-key shortcuts like 'd' do not work when Korean keyboard layout active


dae:
https://forums.ankiweb.net/t/about-hotkeys-in-other-languages/30038/3

chrissy0:
This issue can be closed, see #2533 and #2538.

PR #2533, opened 2023-06-02
Partially supporting 한글 (hangul, Korean characters)

chrissy0:
This pull request enables shortcuts while a Korean keyboard layout is active.

Issue: #2525

Some shortcuts don't work, I have commented them out and annotated them with a comment.

Best,
Chris


chrissy0:
Could you help me figure out why it says `Author christopher.woggon, at the domain gmail.com NOT found in list` even though I have added myself to the `CONTRIBUTORS` file? My previous pull request was successfully merged.

dae:
This feels like a bit of a hack, and it's going to make the code harder to maintain if we have a bunch of different languages listed there. But if it solves the problem and doesn't introduce regressions, I guess it's our best option for now. One thing I'd ask you do is move the shortcuts into a separate function, and adjust that list to include *self.korean_shortcuts() or similar, so they're kept in one place.

The contributors check goes on the Git commit author, not the contents of the file:

commit 70ce584d63144c15c83c216a4d2f83bed5a5f979
Author: chrissy0 <44864267+chrissy0@users.noreply.github.com>

That was likely created by some GitHub UI action.

If you edit the file (add a trailing space to your entry for example) and add the edit to this PR, it should resolve the issue.

chrissy0:
Thanks!

I think it would be a good idea to open an issue for the remaining, not yet working shortcuts.

Best,
Chris

dae:
Thanks Chris! Feel free to add an issue to track that problem.

PR #2536, opened 2023-06-05
Fix "Add-ons possibly involved" message never appearing on Windows

abdnh:


dae:
Thanks as always Abdo!

PR #2539, opened 2023-06-08
Automate schema 11 other duplicates clearing

RumovZ:
While working on the discussed notetype import improvements, I noticed an issue with the duplicate clearing and decided to try to get rid of it for good with a proc macro. What do you think?
I could also add it to the remaining schema 11 structs, so that if they get extended, too, in the future, there's no risk of forgetting that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,6 +97,7 @@ dependencies = [
  "csv",
  "difflib",
  "env_logger",
+ "field_names_derive",
  "flate2",
  "fluent",
  "fluent-bundle",
@@ -1172,6 +1173,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e51093e27b0797c359783294ca4f0a911c270184cb10f85783b118614a1501be"
 dependencies = [
  "instant",
+]
+
+[[package]]
+name = "field_names_derive"
+version = "0.0.0"
+dependencies = [
+ "quote",
+ "syn 2.0.12",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ members = [
   "rslib/i18n",
   "rslib/i18n_helpers",
   "rslib/linkchecker",
+  "rslib/field_names_derive",
   "pylib/rsbridge",
   "build/configure",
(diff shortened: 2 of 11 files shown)


dae:
Thanks for catching that mistake! Automating this seems like a good way to ensure it doesn't happen again. I'll circle back to this tomorrow; just a quick comment for now - do we need that proc macro, or could we rely on https://github.com/TedDriggs/field_names instead? It's nice how it stays out of the public API.

RumovZ:
I noticed, there is an issue: it won't work for fields renamed in serde. Do you mind if I rename all applicable schema 11 fields to the corresponding values in `#[serde(rename = ...)`?

dae:
It would be nice to keep support for renames, because otherwise there's a risk we'll forget that they can't be used, and introduce the same issue again.

Were you thinking of reading the serde rename tags in the procmacro? That might be a bit more complicated, as there may be lines like #[serde(default, rename=...)].

One alternative to a procmacro would be to serialize into a Value instead of a String, and then into a String from that, as that automatically removes duplicate entries. That adds a bit of extra overhead that I was originally trying to avoid, but I'm probably being overly-cautious here.

Another option would be to keep the explicit listing of fields, but check them in unit tests via Value. That would minimize runtime overhead, but mean adding a new field required updating the list (but we'd get a CI error if we forgot).

WDYT?

RumovZ:
> Were you thinking of reading the serde rename tags in the procmacro? That might be a bit more complicated, as there may be lines like #[serde(default, rename=...)].

Well, it would be mainly copy-paste from the serde_derive repo.

> One alternative to a procmacro would be to serialize into a Value instead of a String, and then into a String from that, as that automatically removes duplicate entries.

Wow, that's super simple! A quick benchmark does indicate that `to_string(to_value(noteype))` leads to a 300% slowdown, but we're only talking microseconds here. Especially since this code only runs on downgrade, that's probably acceptable, isn't it?

I don't think this can be used for automated testing, though, due to the many fields with `skip_serializing_if` attributes. In general, they won't cause any conflicts with `other`, unless explicitly mutated to appear in the serialization.

dae:
Good point about skip_serializing. Let's just go the simple Value option then, and save ourselves the extra code :-)

PR #2540, opened 2023-06-09
Fix two issues that can cause editor fields to behave incorrectly

### LINKED ISSUE #2521: "Show HTML by default" buggy after changing current notetype?
https://forums.ankiweb.net/t/bug-2-1-64-default-html-editor/30573/

hikaru-y:
Fixes #2521


dae (commented):
Thanks Hikaru! Just a minor naming nit.

dae:
Great, thanks Hikaru!

PR #2542, opened 2023-06-10
Automate schema 11 other duplicates clearing

RumovZ:
See #2539.


dae:
Always nice when a change makes things simpler - thanks Rumo!

PR #2547, opened 2023-06-13
Migrate to protobuf-es

dae:
Motivation:

- Protobuf-es has a nicer API: messages are represented as classes, and
fields which should exist are not marked as nullable.
- As it uses modules, only the proto messages we actually use get included
in our bundle output. Protobuf.js put everything in a namespace, which
prevented tree-shaking, and made it awkward to access inner messages.
- Code generation is faster - ./run after touching a proto file drops from about 8s to 6s on my machine. The tradeoff
is slower decoding/encoding (https://github.com/ankitects/anki/issues/2043), but that was mainly a concern for the
graphs page, and was unblocked by
https://github.com/ankitects/anki/commit/37151213cd9d431f449ba4b3bc4c0329a1d9af78

Approach/notes:

- We generate the new protobuf-es interface in addition to existing
protobuf.js interface, so we can migrate a module at a time, starting
with the graphs module.
- rslib:proto now generates RPC methods for TS in addition to the Python
interface. The input-arg-unrolling behaviour of the Python generation is
not required here, as we declare the input arg as a PlainMessage<T>, which
marks it as requiring all fields to be provided.
- i64 is represented as bigint in protobuf-es. We were using a patch to
protobuf.js to get it to output Javascript numbers instead of long.js
types, but now that our supported browser versions support bigint, it's
probably worth biting the bullet and migrating to bigint use. Our IDs
fit comfortably within MAX_SAFE_INTEGER, but that may not hold for future
fields we add.
- Oneofs are handled differently in protobuf-es, and are going to need
some refactoring.

Other notable changes:

- Added a --mkdir arg to our build runner, so we can create a dir easily
during the build on Windows.
- Simplified the preference handling code, by wrapping the preferences
in an outer store, instead of a separate store for each individual
preference. This means a change to one preference will trigger a redraw
of all components that depend on the preference store, but the redrawing
is cheap after moving the data processing to Rust, and it makes the code
easier to follow.
- Drop async(Reactive).ts in favour of more explicit handling with await
blocks/updating.
- Renamed add_inputs_to_group() -> add_dependency(), and fixed it not adding
dependencies to parent groups. Renamed add() -> add_action() for clarity.

Once the scheduling work is done, this closes https://github.com/ankitects/anki/issues/2043


dae:
@RumovZ Mind having a quick test of the import-csv page to confirm I've preserved your oneof handling correctly? I'll do some more testing of the various pages tomorrow.

Looks like protobuf-es have landed some speedups since I last looked: https://github.com/bufbuild/protobuf-es/pull/459 :tada:

dae:
Tree shaking needs a bit more investigation - even without it, the congrats.js bundle has dropped from about 175k to 105k, but adding `PURE` annotations drops it further to 84k.

RumovZ:
Looks good, as far as I can tell! Two notes below.

ts/import-csv/index.ts
```diff
@@ -32,22 +26,15 @@ const i18n = setupI18n({
...
-        [notetypes, decks, metadata] = await Promise.all([
-            gettingNotetypes,
-            gettingDecks,
-            gettingMetadata,
-            i18n,
-        ]);
-    } catch (err) {
-        alert(err);
-        throw (err);
-    }
```
RumovZ, line 50: I think the alert didn't work anyway. Was that why you've removed it? Handling this more gracefully is still on the todo list.

ts/import-csv/FieldMapper.svelte
```diff
@@ -3,27 +3,27 @@ Copyright: Ankitects Pty Ltd and contributors
...
-    import { getNotetypeFields } from "./lib";
     import MapperRow from "./MapperRow.svelte";
 
     export let columnOptions: ColumnOption[];
     export let tagsColumn: number;
-    export let globalNotetype: ImportExport.CsvMetadata.MappedNotetype | null;
+    export let globalNotetype: CsvMetadata_MappedNotetype | null;
 
-    let lastNotetypeId: number | undefined = -1;
+    let lastNotetypeId: bigint | undefined = -1n;
```
RumovZ, line 18: Does protobuf-es also properly support optionals so that we can move away from sentinel values?

ts/import-csv/index.ts
dae: The RPC methods automatically alert when an error occurs by default, and processing should stop after one of the promises fails, so I think we can probably get away without an outer catch.

ts/import-csv/FieldMapper.svelte
dae: For lists of optionals (eg change notetype screen), Protobuf doesn't allow for repeated optionals, so the optional would still need to be placed in a wrapper message that is repeated, eg the code would need to check 'if (x[5].val !== null)' instead of 'if (x[5] !== null)'. IIRC that's the main reason we are currently using sentinels in lists.

I don't recall why a sentinel was used in the code you quoted, and undefined probably would work instead.

ts/import-csv/index.ts
RumovZ: If we're talking about a popup dialog here, it doesn't work for me. If the file can't be parsed, there's nothing but an empty window.

ts/import-csv/FieldMapper.svelte
RumovZ: If I'm not mistaken, the old library always set missing optional numbers to 0, and there was no way to distinguish it from the actual value 0.

ts/import-csv/index.ts
dae: Do you mean on this branch? Not sure what's happening there; I was getting a pop-up in my testing. I've fixed it so instead of a generic error it relays the backend message now. Tested on Windows too to make sure it works there:
2023-06-14T22:14:09,805846314+10:00

dae:
I haven't noticed any breakages in my testing today, so let's give it a try on main. If you still have issues with the error message not popping up, please let me know.

ts/import-csv/FieldMapper.svelte
RumovZ: Ah, that could be it.

ts/import-csv/index.ts
RumovZ: I see, you're testing with invalid Unicode. That works here, too. I was testing with an empty file, but that seems to be a special case, because it's a frontend error.
The reason is the slicing in ImportCsvPage.svelte, line 67. But probably the dialog should abort earlier for empty files.

dae: Fixed :-) https://github.com/ankitects/anki/commit/de8f62f831402a997f4eba5f9a69b403ee2f004d

PR #2549, opened 2023-06-18
Refactor progress handling

dae:
Previously it was Backend's responsibility to store the last progress, and when calling routines in Collection, one had to construct and pass in a Fn, which wasn't the most ergonomic. This PR adds the last progress state to the collection, so that the routines no longer need a separate progress arg, and makes some other tweaks to improve ergonomics.

ThrottlingProgressHandler has been tweaked so that it now stores the current state, so that callers don't need to store it separately. When a long-running routine starts, it calls col.new_progress_handler(), which automatically initializes the data to defaults, and updates the shared UI state, so we no longer need to manually update the state at the start of an operation.

The backend shares the Arc<Mutex<>> with the collection, so it can get at the current state, and so we can update the state when importing a backup.

Other tweaks:

- The current Incrementor was awkward to use in the media check, which uses a single incrementing value across multiple method calls, so I've added a simpler alternative for such cases. The old incrementor method has been kept, but implemented directly on ThrottlingProgressHandler.
- The full sync code was passing the progress handler in a complicated way that may once have been required, but no longer is.
- On the Qt side, timers are now stopped before deletion, or they keep running for a few seconds.
- I left the ChangeTracker using a closure, as it's used for both importing and syncing.

### DIFF
--- a/qt/aqt/mediacheck.py
+++ b/qt/aqt/mediacheck.py
@@ -59,6 +59,7 @@ class MediaChecker:
 
     def _set_progress_enabled(self, enabled: bool) -> None:
         if self._progress_timer:
+            self._progress_timer.stop()
             self._progress_timer.deleteLater()
             self._progress_timer = None
         if enabled:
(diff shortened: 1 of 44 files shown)


dae:
Mind taking a quick look when you have a chance @RumovZ?

RumovZ:
Nice, seems to make things a lot simpler!

PR #2550, opened 2023-06-18
Fix two issues with MathJax editor

### LINKED ISSUE #1947: MathJax editing area loses focus when switching apps
If you're typing a MathJax equation and switch to another app to reference something, the typing area is gone when you switch back. Might be related to #1861?

https://forums.ankiweb.net/t/mathjax-live-preview-loses-focus-when-jumping-between-windows/21122

hikaru-y:
Fixes #1947

Also fixes an issue where the MathJax editor would not close when changing notes via shortcut (Ctrl+N/Ctrl+P).


dae:
I think the original on:blur handling was to work around an issue where the editor didn't disappear as a different note was selected in the editor, and that no longer seems to happen, so looks good to me. Thanks Hikaru :-)

PR #2551, opened 2023-06-18
Issue #2546 Exclude Fabric's canvas dependency

### LINKED ISSUE #2546: Figure out how to exclude Fabric's canvas dependency
From what I can tell, we don't need it, and it adds an extra 190MB to node_modules.

This might work: https://github.com/yarnpkg/yarn/issues/4611

kavelrao:
Using the fix from here: https://github.com/yarnpkg/yarn/issues/4611#issuecomment-1595182653 to avoid installing `canvas`.

Closes #2546


dae:
Nice - thanks Kavel!

PR #2558, opened 2023-06-26
Fix main window geometry/state not being restored in some cases

abdnh:
Introduced by 19b08eb280df18c4059606b3c24d3d0230db7c43

Report: https://forums.ankiweb.net/t/65-the-size-of-the-main-window-is-reset-every-time-it-is-closed-again/31364


dae:
Thanks for catching that Abdo!

chezfo:
resolved
2.1.66

PR #2566, opened 2023-07-01
Attempt to update to CodeMirror 6

dae:
CodeMirror 6's API has changed dramatically, so we'll need to rework our adaptor to work with it. With these work-in-progress changes, the text in the HTML editor updates when editing with the standard editor and vice-versa. But there is still work to do:

- The various config options need to be converted to the new Extension-based way of providing options.
- The remaining .on() events need to be handled somehow - only the change event has currently been reworked.
- The way changes are synced between the two editors may need rethinking. Currently, a change in one will be sent to the other, and if the other alters the representation, that results in the change being sent back again. This has led to the editor hanging in the past when neither side could reach agreement. After updating to CodeMirror 6, when you attempt to type a `<` in CodeMirror, it sends the bare `<` to the WYSIWYG editor, and the WYSIWYG editor sends back a `&lt` instead - presumably because the DOM is assuming that's what a bare `<` symbol was intended to be. We may need some way to ignore any incoming changes when we write out a change, or make the process one-way, with whichever editor is currently in focus taking precedence.

Any help would be appreciated!

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
         "@popperjs/core": "^2.11.8",
         "bootstrap": "^5.3.0",
         "bootstrap-icons": "^1.10.5",
-        "codemirror": "^5.63.1",
+        "codemirror": "^6.0.1",
         "css-browser-selector": "^0.6.5",
         "d3": "^7.0.0",
         "fabric": "^5.3.0",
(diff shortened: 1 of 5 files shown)


dae:

PR #2575, opened 2023-07-12
Align code blocks to left

### LINKED ISSUE #2573: Align code blocks to the left by default
The [\<pre>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre) HTML element represents preformatted text which is to be presented exactly as written in the HTML file. It is often used to represent multiple lines of code. For example:
```html
<pre>
def foo():
    print('bar')
</pre>
```

Currently, the `<pre>` element is centered by default, which results in unreadable code blocks ([HTML source for this card](https://pastebin.com/S0ThZyN4)):
image

The expected behaviour is for code blocks to be left aligned:
image

I believe this style covers everything:
```css
pre {
    text-align: left;
}
```
It only makes sense to align preformatted text to the left, so RTL text shouldn't be an issue.

List items have already been aligned in https://github.com/ankitects/anki/issues/1801. Now let's do the same for code blocks.

shunlog:
Fixes #2573

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -127,6 +127,7 @@ Christopher Woggon <christopher.woggon@gmail.com>
 Kavel Rao <github.com/kavelrao>
 Ben Yip <github.com/bennyyip>
 mmjang <752515918@qq.com>
+shunlog <github.com/shunlog>
 
 ********************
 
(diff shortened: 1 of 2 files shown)


dae:
Thanks @shunlog! Please rebase over the latest main to fix the conflict, then I'll be happy to merge this.

shunlog:
@dae done (for now)

dae:
Thanks!

Issue #2453, opened 2023-03-24
Bold button is confused by font-weight span


dae:
Paste the following in a field:

```
<span style="font-weight: 400;">test</span>
```

Select one of the characters inside and attempt to bold it. We end up with:

```
<b>te</b>s<b>t</b>
```

Original report: https://forums.ankiweb.net/t/applying-bold-inverts-selection/28641/3

SeojinK1m:
Can I take this issue?

dae:
Sure

mmjang:
Can we change the threshold `400` to `700`?

```
        const fontWeight = element.style.fontWeight;
        if (fontWeight === "bold" || Number(fontWeight) >= 400) {
```

https://github.com/ankitects/anki/blob/a2cd9ec34da353aa5f4d65002111ccd3966ab3a4/ts/editor/editor-toolbar/BoldButton.svelte#L28C30-L28C30

According this [mdn doc](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#common_weight_name_mapping), "bold" is mapped to `font-weight: 700`.

I did a little bit testing, and this change gives more reasonable behaviour.

From normal to bold:

<img width="741" alt="image" src="https://github.com/ankitects/anki/assets/5045203/b0f74d27-f7ea-486a-b8e0-85e71b49e121">

<img width="750" alt="image" src="https://github.com/ankitects/anki/assets/5045203/b36d14ef-84cc-4eda-8a1b-570e7f2a2458">

From bold to normal:

<img width="737" alt="image" src="https://github.com/ankitects/anki/assets/5045203/9164b7ef-85b5-40cb-bf0c-6eddc45e16d1">

<img width="633" alt="image" src="https://github.com/ankitects/anki/assets/5045203/b8b41b1d-16bf-46e5-85dd-9535638cd082">

dae:
Thanks for digging into it! If the docs say 700, happy to go with a check for 700 - I've pushed this change to the main branch.

PR #2579, opened 2023-07-14
Decide if element is bold by getComputedStyle (#2453)

mmjang:
Some testing:

image


dae:
Sorry @mmjang , I didn't notice you'd pushed a PR for this until after I pushed the 400->700 change. My change lacks the getComputedStyle() call in your PR, but that doesn't seem to be necessary to fix the problem. Can you think of an example where it would be required, or do you think we are ok now?

mmjang:
@dae `getComputedStyle` is more accurate. If the boldness style of an element is inherited from it's parent, then `element.style.fontWeight` may not work. Check this example:

image

dae:
Thanks, that's a good point. If we fix this for BoldButton.svelte, perhaps we should be applying similar changes for underline, italics, color, and so on?

mmjang:
> Thanks, that's a good point. If we fix this for BoldButton.svelte, perhaps we should be applying similar changes for underline, italics, color, and so on?

Yep, I think so.

dae:
Haven't forgotten about this, just haven't had time to circle back to it yet I'm afraid.

dae:
I've finally found time to add the italics case as well, but have left colors for now. Sorry this got blocked for so long.

dae:
Unfortunately I had to revert this, as it caused a regression: https://forums.ankiweb.net/t/anki-24-10-beta/49989/150

PR #2580, opened 2023-07-16
Fix addon update dialog showing twice

RumovZ:
723a3aaa5b889e64c4bcfef11a2335acaab8a91b is actually redundant with 122a6f860f8382c31804211ca38e0149287713d0, but I guess it's still more correct this way.

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -498,6 +498,8 @@ class AnkiQt(QMainWindow):
             self._refresh_after_sync()
             if onsuccess:
                 onsuccess()
+            if not self.safeMode:
+                self.maybe_check_for_addon_updates(self.setupAutoUpdate)
 
         self.maybe_auto_sync_on_open_close(_onsuccess)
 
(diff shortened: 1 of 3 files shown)


qt/aqt/main.py
```diff
@@ -960,6 +962,7 @@ def wrap_on_updates_installed(log: list[DownloadLogEntry]) -> None:
                 requested_by_user=False,
             )
             self.pm.set_last_addon_update_check(int_time())
+            self.pm.set_last_run_version()
```
dae, line 965: This might cause confusion down the line. We might start using the last run version for something other than add-on checks (eg, to upgrade the profile metadata), and might add some option in the future to disable the check for add-on updates, not realizing it will also prevent the last run version from being updated. Maybe it might be better to move this up into the onsuccess call, and pass a version_changed option into maybe_check_for_addon_updates(), or something like that?

RumovZ: Good point, though I'd also like to avoid setting and reading the value at different places. Then you'd always have to double-check whether you're reading the old or the new value.
What do you think of the new solution?

dae: Much nicer - now code that doesn't run immediately can also get at the previous version too. Thanks Rumo!

dae:
Re .ninja_deps, can't remember if I mentioned this already, but you might want to try tools/install-n2 now - I pushed a fix for the console issue on Windows.

RumovZ:
Works like a charm, thanks. 🙂

Issue #2564, opened 2023-07-01
Address accessibility warnings


dae:
Currently we have some accessibility warnings disabled when running svelte-check, as Svelte upgrades introduced a bunch of them at once, and we couldn't tackle them at the time. They're disabled here:

https://github.com/ankitects/anki/blob/baa631c6efca6aeba8e0dc08056bbc43b1057d30/build/ninja_gen/src/node.rs#L228

If you remove one of those lines and run `./ninja check:svelte`, you can see where they occur (and they're also already shown in the editor). It would be nice if we could address these, even a few at a time.

JayK1524:
I would like to work on this issue. Please guide me, how can I resolve this.

dae:
If you Google one of the excluded error types, there are pages such as https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-noninteractive-tabindex.md that provide suggestions on how things might be addressed. I'm no accessibility expert, so it will likely be a learning experience for both of us. Please do some Googling, and if you think you've got an approach for how to address these, feel free to discuss before you start work on the change.

derekdang:
I'm interested in working on this issue. I addressed the lint issues in ts/graphs/CardCounts.svelte to get an idea of what needs to be done. 

Added the tabIndex so the items that were clickable are now tabbable. The items that were clickable now open with Space/Enter.
I followed principles/guidelines found in 
https://www.w3.org/WAI/GL/wiki/Making_actions_keyboard_accessible_by_using_keyboard_event_handlers_with_WAI-ARIA_controls
https://usability.yale.edu/web-accessibility/articles/links
![accessibility demo](https://github.com/ankitects/anki/assets/21960193/a35a1fbd-8baa-4b30-822b-7a9c42fe2ff4)

https://gist.github.com/derekdang/dd231feb6ba558679bd80c642bc80bf8#file-cardcounts-svelte-L40-L53
https://gist.github.com/derekdang/dd231feb6ba558679bd80c642bc80bf8#file-cardcounts-svelte-L81-L84
The gist above highlights my changes.

adeleke5140:
I have experience with fixing a11y errors. Can I get on this?

dae:
Thank you both! Anyone's welcome to work on this, but please note down which things you're planning to fix, so people don't end up duplicating work. I'd also recommend starting small, so we can work through any potential design/style issues.

[...]

PR #2582, opened 2023-07-17
Address a11y warning for TabbedValue

derekdang:
Resolves the warnings for `TabbedValue.svelte` & `SettingTitle.svelte` as part of #2564.

Tabs for Daily Limits section in Deck Options are now `Button`s instead of `Span`s.
To verify, enable V3 Scheduler in Preferences and go to Deck Options for any Deck.

Also added explicit ignore for a11y warnings in SettingTitle. Clicking the Setting Title opens the contextual Help Content directly as a convenience. In the discussion in #2578, the user will be able to navigate to the HelpModal without a mouse.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -129,7 +129,7 @@ Ben Yip <github.com/bennyyip>
 mmjang <752515918@qq.com>
 shunlog <github.com/shunlog>
 3ter <github.com/3ter>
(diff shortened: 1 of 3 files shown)


derekdang:
After 
image

Before
image

dae:
Thanks Derek! I've pushed a minor tweak that makes the underlines straight again, but it otherwise looks good.

PR #2583, opened 2023-07-18
Add keyboard shortcut for legacy stats

Luc-Mcgrady:
Currently, shift clicking the stats button is the only way to open the legacy stats menu.
This adds pressing shift-t as an option to do the same thing.

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -1102,6 +1102,7 @@ title="{}" {}>{}</button>""".format(
             ("a", self.onAddCard),
             ("b", self.onBrowse),
             ("t", self.onStats),
+            ("shift+t", self.onStats),
             ("y", self.on_sync_button_clicked),
         ]
         self.applyShortcuts(globalShortcuts)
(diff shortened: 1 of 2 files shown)


qt/aqt/main.py
```diff
@@ -1102,6 +1102,7 @@ def setupKeys(self) -> None:
             ("a", self.onAddCard),
             ("b", self.onBrowse),
             ("t", self.onStats),
+            ("shift+t", self.onStats),
```
dae, line 1105:
```suggestion
            ("Shift+t", self.onStats),
```

dae (commented):
Looks ok, but please bear in mind that the old stats screen will be removed at one point, so in the medium term any add-ons that rely on this old screen will need to be updated.

Luc-mcgrady:
I don't think this will even break if the stats screen is removed because of the way its set up with the function being the same for both.  
Is it possible to modify the non legacy stats screen with an add-on? I don't think I've seen an add-on that's done it?

dae:
I just meant that if you rely on add-ons using that old screen, some future release will break them if they have not updated to the new screen. It is possible to inject new graphs in that screen I believe, but nobody's gotten around to making an example of how to do it that add-on authors can copy yet.

Thanks for updating the capitalization. Unfortunately the CONTRIBUTORS file has taken a conflict; if you could rebase your changes over the latest main branch, I'll get them merged in.

Luc-Mcgrady:
Should be good now 👍.

PR #2585, opened 2023-07-20
Fix a11y error in CardsCount Component

adeleke5140:
This PR aims replaces the `span` element with a semantically appropriate `button` element.
This would ensure that:

1. The button can be tabbed.
2. There is proper focus indicator.
3. The a11y warning don't show up.

One question I have is:
To fix this error, I had to uncomment each of this line:
```
  build.add_variable(
            "compiler_warnings",
            [
                "a11y-click-events-have-key-events", //here
                "a11y-no-noninteractive-tabindex", //here
                "a11y-no-static-element-interactions", //and here
            ]
            .iter()
            .map(|warning| format!("{}$:ignore", warning))
            .collect::<Vec<_>>()
            .join(","),
        );
```
Is that the best practice of fixing the errors?
Or is there a cleaner way?

### DIFF
--- a/ts/graphs/CardCounts.svelte
+++ b/ts/graphs/CardCounts.svelte
@@ -64,7 +64,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                         <td>
                             <span style="color: {d.colour};">■&nbsp;</span>
                             {#if $prefs.browserLinksSupported}
-                                <span class="search-link" on:click={() => dispatch('search', { query: d.query })}>{d.label}</span>
+                                <button class="search-link" on:click={() => dispatch('search', { query: d.query })}>{d.label}</button>
                             {:else}
                                 <span>{d.label}</span>
                             {/if}
@@ -124,8 +124,15 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         }
     }
 
-    .search-link:hover {
+    .search-link {
+        border: none;
+        background: transparent;
         cursor: pointer;
+        box-shadow: none;
+        padding: 4px;
+    }
+
+    .search-link:hover {
         color: var(--fg-link);
         text-decoration: underline;
     }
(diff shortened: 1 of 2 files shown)


ts/graphs/CardCounts.svelte
```diff
@@ -124,8 +124,15 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         }
     }
 
-    .search-link:hover {
+    .search-link {
+        border: none;
+        background: transparent;
         cursor: pointer;
+        box-shadow: none;
+        padding: 4px;
```
dae, line 132: Do we need the padding? The text looks a bit too spaced out now, compared to the other graphs around it.

dae (commented):
Thanks Kehinde/Kenny! (please let me know which you prefer). New approach looks good, with just a minor comment on spacing.

dae:
> Is that the best practice of fixing the errors?
Or is there a cleaner way?

That's the way I'd do it. Once those errors have been fixed in all files (in future PRs), we can remove them permanently from that file.

ts/graphs/CardCounts.svelte
adeleke5140: Kenny is fine.

Regarding the spacing, I added the padding so that the text has some breathing room when tabbed. It ensures it isn't squished with the focus outline.

I can fix the excessive spacing by modifying the margin value to match the existing design. Here is how it looks now.

Screenshot 2023-07-20 at 15 35 11

adeleke5140:
> > Is that the best practice of fixing the errors?
> > Or is there a cleaner way?
> 
> That's the way I'd do it. Once those errors have been fixed in all files (in future PRs), we can remove them permanently from that file.

Got it, thanks.

ts/graphs/CardCounts.svelte
dae: Looks good - thanks Kenny!

PR #2587, opened 2023-07-21
Deck Options accessibility changes

derekdang:
As part of the discussion in #2564, created ClickableBadge component for use cases where the Badge has on:click functionality. Initial refactoring done for areas where Badge is used in Deck Options.

aria-label for the Button in ClickableBadge is set to the tooltip so that Screen Readers will be able to provide information to users. The title can also be set where necessary but I opted not to set it for HelpModal since it is intuitive.

RevertButton and HelpModal both now make use of the ClickableBadge. Enabled tabbable on the resulting DropdownItem from RevertButton. 

There is an undesired effect when the Dropdown Item is in focus, the normal Badge component's border is also outlined.
image
I tried setting outline to none and some other approaches but haven't been able to fix it.


ftl/core/help.ftl
```diff
@@ -6,6 +6,9 @@
 # Tooltip for links to the manual
 help-open-manual-chapter = Open { $name } in the manual
 
+# Tooltip for info circle badge in deck options
+help-open-help-modal = Open help content for { $name }
```
dae, line 10: I wonder if this adds much over reusing the existing string above it?

dae:
Thanks for your continued work on this! A couple of minor issues I noticed:

- button styling is shown when hovering over the (i), which causes it to move. Perhaps we should remove the styling, or add appropriate padding/margins so the button doesn't move.
- The revert button is shown a few pixels lower than it was previously, which makes it more less aligned with the controls like switches to its left
- When tabbing through, the revert button comes before the control, despite it appearing on the right - not sure what's causing that.

I can't seem to reproduce the double highlight on a Qt 6.4 system here - when I trigger the revert button, the highlight disappears and a new one appears on the front.

dae:

PR #2590, opened 2023-07-25
Improve namespacing for debug console

RumovZ:
https://forums.ankiweb.net/t/debug-console-throws-errors-on-correct-code/32382

If we don't pass dicts for globals and locals to `exec()`, the ones from the surrounding code are used. But functions and list comprehensions have their own locals dict, and thus can't access other names declared in the REPL or locals like `mw`.
This patch fixes that by passing in a dict that is used both for globals and locals.


dae:
Thanks for the quick fix Rumo!

Issue #2588, opened 2023-07-24
Image occlusion doesn't scale when page zoomed in


dae:
Because a max-height is set on the occlusion container, zooming the page in with ctrl+mousewheel results in the image actually shrinking slightly, due to the change in margins.

If we use an explicit height, instead of capping the height relative to the viewport, then scaling works, but images do not fit the screen by default. 

```diff
diff --git a/ts/image-occlusion/review.scss b/ts/image-occlusion/review.scss
index 8e4065fd1..95da4ba5a 100644
--- a/ts/image-occlusion/review.scss
+++ b/ts/image-occlusion/review.scss
@@ -5,7 +5,8 @@
     transform: translate(-50%, 0);
     // allow for 20px margin on html element, or short windows can truncate
     // image
-    max-height: calc(95vh - 40px);
+    //max-height: calc(95vh - 40px);
+    height: 1000px;
 }
 
 #image-occlusion-container img {
```

If anyone has ideas on how we could both scale-to-fit by default (and when window is resized), but also support zooming, please let us know!

Heads up @krmanik @glutanimate

krmanik:
I have tested by start review and zoom in then go back to deck selection the zoom is correct but incorrect for reviewer (in io notetype). Again, restart review and zoom out and then again go back to deck selection and the zoom is again incorrect. Also removing CSS for all three class and checking zoom have different result than expected. The possible solution should be in JS implementation, and we may not need to change CSS of above class. Is this solution be used?

```diff
diff --git a/ts/image-occlusion/review.ts b/ts/image-occlusion/review.ts
index 580d46fe3..620fd6b16 100644
--- a/ts/image-occlusion/review.ts
+++ b/ts/image-occlusion/review.ts
@@ -11,6 +11,8 @@ import { Polygon } from "./shapes/polygon";
 import { Rectangle } from "./shapes/rectangle";
 import type { Size } from "./types";
 
+let scale = 1.0;
+
 export function setupImageCloze(): void {
     window.addEventListener("load", () => {
         window.addEventListener("resize", setupImageCloze);
@@ -32,6 +34,14 @@ function setupImageClozeInner(): void {
         return;
     }
 
+    container.addEventListener("wheel", (e) => {
+        e.preventDefault();
+        const step = 0.05;
+        const delta = Math.sign(e.deltaY) * step;
+        scale = Math.max(scale + delta, 0.1);
+        container.style.transform = `translate(-50%, 0) scale(${scale})`;
+    });
+
[...]

PR #2591, opened 2023-07-25
Introduce new Reviewer zoom mode for images

glutanimate:
Inspired by https://github.com/ankitects/anki/issues/2588, but hoping to address image zoom in a more general capacity as I think  this has been a long-standing issue for quite a few users (even predating native zoom, see e.g. the reviews on [zoom add-ons](https://ankiweb.net/shared/info/1956318463) from as far back as 2.0).

In its current iteration, this PR introduces a **new JS-based zoom system** that works in parallel to the Qt-native system. When the Reviewer is active, any user-triggered zoom actions can be switched over to this alternate mode by holding down <kbd>Shift</kbd>. In alternate mode, zooming is performed by scale-transforming `document.body`, preserving the layout sizing and thus allowing users to freely zoom into images. Quick comparison between the mode (native first, JS second):

https://github.com/ankitects/anki/assets/5459332/452dad48-3968-4f31-9eca-6e10c833b98a

The **downside** to this approach, and why I ended up implementing it as an optional mode, is that text does not reflow when the layout sizing is preserved, so long-ish paragraphs will end up extending past the viewport. This makes the mode ill-suited for use cases where users will want to adjust the zoom level in order to boost text readability – which I assume are quite common, e.g. when using Anki at larger screen distances.

I played around with this a lot, going back-and-forth between different approaches, and wasn't able to find a full-page zoom mode that works well for both scenarios – thus the two modes, even if not particularly ideal from a discoverability standpoint.

On the upside, this PR also introduces a couple of zoom **features** in the alternate mode which I think are quite neat to have:

- the zoom level is now communicated via a small info box in the top right
- zoom levels in alternate mode are persisted across Anki sessions – not sure if this is 100% desirable for everyone, could switch it back to just being persistent within a session

**Some thoughts on alternatives**

If we are happy to handle this on a less generic level, then I think it could be interesting to take [@krmanik's approach](https://github.com/ankitects/anki/issues/2588#issuecomment-1648289462) and extend it to all images, i.e. see if we can't equip all `img`s with a ctrl+scroll handler that would allow users to zoom into individual images. The problem I would see here is that we don't have full control over user-created note types, or at least not as much as we do with a default note type like IO, and so we could run into scenarios where card layout or styling interferes with image zooming or creates unexpected behavior of some sort.

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -1342,15 +1342,9 @@ title="{}" {}>{}</button>""".format(
         qconnect(m.actionPreferences.triggered, self.onPrefs)
 
         # View
-        qconnect(
-            m.actionZoomIn.triggered,
-            lambda: self.web.setZoomFactor(self.web.zoomFactor() + 0.1),
-        )
-        qconnect(
-            m.actionZoomOut.triggered,
-            lambda: self.web.setZoomFactor(self.web.zoomFactor() - 0.1),
-        )
-        qconnect(m.actionResetZoom.triggered, lambda: self.web.setZoomFactor(1))
+        qconnect(m.actionZoomIn.triggered, self.zoom_in)
+        qconnect(m.actionZoomOut.triggered, self.zoom_out)
+        qconnect(m.actionResetZoom.triggered, self.reset_zoom)
         # app-wide shortcut
         qconnect(m.actionFullScreen.triggered, self.on_toggle_full_screen)
         m.actionFullScreen.setShortcut(
@@ -1396,6 +1390,24 @@ title="{}" {}>{}</button>""".format(
     def show_menubar(self) -> None:
         self.form.menubar.setMaximumSize(QWIDGETSIZE_MAX, QWIDGETSIZE_MAX)
         self.form.menubar.setMinimumSize(0, 0)
+
+    def zoom_in(self) -> None:
+        if self.state != "review":
+            self.web.setZoomFactor(self.web.zoomFactor() + 0.1)
+        else:
+            self.reviewer.zoom_in()
+
+    def zoom_out(self) -> None:
+        if self.state != "review":
+            self.web.setZoomFactor(self.web.zoomFactor() - 0.1)
+        else:
+            self.reviewer.zoom_out()
+
+    def reset_zoom(self) -> None:
+        if self.state != "review":
+            self.web.setZoomFactor(1)
+        else:
+            self.reviewer.reset_zoom()
 
     # Auto update
     ##########################################################################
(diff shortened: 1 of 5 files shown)


glutanimate:
One idea on how we could tackle the discoverability issue slightly: Instead of having two menu options that act differently depending on whether <kbd>Shift</kbd> is depressed, we could add another set for the new zoom mode and relabel the old ones. I.e., I would imagine the menu to look something like this in that case:

```
Increase text size (Ctrl++)
Decrease text size (Ctrl+-)
Zoom in (Ctrl+Shift++)
Zoom out (Ctrl+Shift+-)
Reset zoom (Ctrl+0)
```

krmanik:
I have tested this PR and it works as expected but we should have similar zoom option as before so user will not confused with new changes to zoom option. Also I think we should switch the `Ctrl+Shift+(+/-)` for text and `Ctrl+(+/-)` for window zoom.

dae:
Initially thought it was broken, then realised I needed to be holding down ctrl as well. :-)

The scrolling in and out and the feedback on zoom level seems to work well.

For reference, AnkiMobile has separate options for zooming (pinch) and text size as well, with some differences:

- Zooming state is not persisted, and resets when the user answers the card. My thinking here was that the user is probably zooming into an image where the text is hard to read, and may not want all images shown that big. Zooming out also ensures text is not cut off on the left and right of the screen, so it seemed like a better default. WDYT?
- Text size adjustments work by dynamically rewriting all font references on the page to be relative (in rems), and then adjusting the body font. This causes the text to scale, while leaving images untouched (they don't confusingly shrink when zooming in due to margins). 

Only the text size is displayed as a menu option in the UI; zooming is only accessible via pinch or double-tap.

Not suggesting we just copy the AnkiMobile approach here, but throwing it out as a possible alternative for discussion.

Regarding shortcut keys, we held off on adding them in the past because they don't work in some keyboard layouts, and were conflicting with the editor:  https://github.com/ankitects/anki/pull/1668#issuecomment-1043191227

dae:
Here we have a user who doesn't want the zoom to change as they go through cards: https://forums.ankiweb.net/t/image-occlusion-improvement-on-anki-mobile/32908

krmanik:
@glutanimate 
If you are busy then I should complete this PR?

dae:
@glutanimate ping

glutanimate:
@dae I should have some time to revisit this over the next couple of weeks, thanks for the slight nudge :)

dae:
Another gentle nudge. No judgement - people in glass houses... ;-)

chrislongros:
@glutanimate Any progress? :)

dae:

PR #2202, opened 2022-11-21
Move away from Bazel

dae:
Bazel brought a lot of nice things to the table, such as rebuilds based on content changes instead of modification times, caching of build products, detection of incorrect build rules via a sandbox, and so on. Rewriting the build in Bazel was also an opportunity to improve on the Makefile-based build we had prior, which was pretty poor: most dependencies were external or not pinned, and the build graph was poorly defined and mostly serialized. It was not uncommon for fresh checkouts to fail due to floating dependencies, or for things to break when trying to switch to an older commit.

For day-to-day development, I think Bazel served us reasonably well - we could generally switch between branches while being confident that builds would be correct and reasonably fast, and not require full rebuilds (except on Windows, where the lack of a sandbox and the TS rules would cause build breakages when TS files were renamed/removed).

Bazel achieves that reliability by defining rules for each programming language that define how source files should be turned into outputs. For the rules to work with Bazel's sandboxing approach, they often have to reimplement or partially bypass the standard tools that each programming language provides. The Rust rules call Rust's compiler directly for example, instead of using Cargo, and the Python rules extract each PyPi package into a separate folder that gets added to sys.path.

These separate language rules allow proper declaration of inputs and outputs, and offer some advantages such as caching of build products and fine-grained dependency installation. But they also bring some downsides:

- The rules don't always support use-cases/platforms that the standard language tools do, meaning they need to be patched to be used. I've had to contribute a number of patches to the Rust, Python and JS rules to unblock various issues.
- The dependencies we use with each language sometimes make assumptions that do not hold in Bazel, meaning they either need to be pinned or patched, or the language rules need to be adjusted to accommodate them.

I was hopeful that after the initial setup work, things would be relatively smooth-sailing. Unfortunately, that has not proved to be the case. Things frequently broke when dependencies or the language rules were updated, and I began to get frustrated at the amount of Anki development time I was instead spending on build system upkeep. It's now about 2 years since switching to Bazel, and I think it's time to cut losses, and switch to something else that's a better fit.

The new build system is based on a small build tool called Ninja, and some custom Rust code in build/. This means that to build Anki, Bazel is no longer required, but Ninja and Rust need to be installed on your system. Python and Node toolchains can be automatically downloaded like in Bazel, though the build will use a system Python 3.9 if one is installed (and the same could be done for Node, though testing will be required to determine the minimum working version).

This new build system should result in faster builds in some cases:

[...]

PR #2593, opened 2023-07-27
[Development] Specify `extra` directory as a designated ignored folder

glutanimate:
Excludes `extra/` from version tracking, file formatters, and file checks, which should allow contributors to add their own files to the project tree without interfering with checks (e.g. in my case MWE code samples for PRs, etc.)

Also adds mypy caches to the exclusion rules, allowing contributors to use the VS Code mypy integration.

(follow-up to a [fairly old discussion](https://github.com/ankitects/anki/pull/2202#issuecomment-1328159393), picking it up now as I have been running into a lot of woes with dprint checks recently)

### DIFF
--- a/.dprint.json
+++ b/.dprint.json
@@ -29,7 +29,6 @@
         "qt/bundle/PyOxidizer",
         "target",
         ".mypy_cache",
-        ".pytest_cache",
         "extra"
     ],
     "plugins": [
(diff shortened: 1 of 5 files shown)


dae:
I'd forgotten I was the one that suggested "extra" as the name :sweat_smile: Thanks Aristotelis.

PR #2598, opened 2023-07-28
[IO API] Create Python entry points for creating and editing IO notes

glutanimate:
A first small batch of changes to create the API surface that the new version of IOE will need. More to follow as I continue with the rework – the bulk of the work will lie on the JS/Svelte side.

This batch targets the python editor API, slightly refactoring the current IO-related methods and creating entry-point methods for adding and editing image occlusions that abstract the Py → JS comms implementation details away.

CC @krmanik @dae

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -51,6 +51,7 @@ class AddCards(QMainWindow):
         add_close_shortcut(self)
         self._load_new_note()
         self.setupButtons()
+        self.col.add_image_occlusion_notetype()
         self.history: list[NoteId] = []
         self._last_added_note: Optional[Note] = None
         gui_hooks.operation_did_execute.append(self.on_operation_did_execute)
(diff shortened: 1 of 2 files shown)


dae:
Looks good in isolation, but unfortunately has conflicted with Mani's recent changes. Mani's code assumes that the file picker function would be called on both add and edit paths, and your code is assuming it's only called for new notes, so will presumably need tweaking.

glutanimate:
OK, updated and rebased against `main`.

NB: Merging this PR would re-enable IO, so let's defer it until 2.1.66 is released, or if the release is likely to take longer and @krmanik and I need to work on IO things in parallel, let's merge it into an IO feature branch perhaps.

dae:
Release is out :rocket:

dae:
Your recent PRs are failing when merged - eg https://buildkite.com/ankitects/anki-ci/builds/6166#018a2565-c300-4559-a029-fa20a10581d5. Would you mind doing a quick PR that adds an edit to CONTRIBUTORS from GitHub's interface?

glutanimate:
[Done](https://github.com/ankitects/anki/pull/2623)! Looks like GH switched my account over to the new ID-based noreply format.

Issue #2596, opened 2023-07-27
Image occlusion mask editor only loads for the first created note


glutanimate:
Follow-up to #2485 

With the latest `main` build, the IO mask editor disappears as soon as you attempt to create a second note:

https://github.com/ankitects/anki/assets/5459332/6ff85138-7bec-49b4-9bf8-20bd940471f6

The console log and stdout seem clean, did not see any error messages.

glutanimate:
Thanks for the quick fix! This works well when using the image selection button. However, now that I'm starting to tinker with programmatically launching the editor into IO mode, I'm finding that I'm running into the same issue as before.

My code is using the updated python API in #2598, but I can also reproduce the issue on main.

Essentially, I'm skipping the image selection step prompted for by the button and calling `setupMaskEditor` directly with a predetermined image. As a result, the `imageOcclusionMode` reset you added in 1d7559819ca3520898247585bfcac96904737bec is skipped over, which I assume explains why the problem resurfaces. FWIW, the behavior is exactly the same as the user-interaction mediated trigger above, i.e. calling setupMaskEditor works fine the first time before the mask editor has been invoked at least once in a session, but then fails subsequently when trying to switch back from e.g. a basic note type.

Given the fix above, the first thing I tried was adding the same reset to `setupMaskEditor`, i.e. setting `imageOcclusionMode = undefined` [at the top](https://github.com/ankitects/anki/blob/1d7559819ca3520898247585bfcac96904737bec/ts/editor/NoteEditor.svelte#L402), but unsurprisingly this didn't work. I assume the problem is that a re-render is not triggered quick enough, right? Changes applied at button onClick time have a lot more time to play out of course, given that the user moves on to select an image, etc.

With that in mind, do you think there's another way we could approach this perhaps? Still getting up to speed on Svelte, so I don't have a full overview of how the rendering lifecycle works.

glutanimate:
OK, nvm, I think #2599 should do the trick - writing things out always helps :sweat_smile:

dae:
> I assume the problem is that a re-render is not triggered quick enough, right? 

My guess as to what's happening here: if there's a delay between the mode being unset and set to a different value, the existing image occlusion component is removed from the DOM. If the mode gets changed without being first removed, the existing component is removed, and presumably it's in an inconsistent state then. If we need to force the existing component to be removed first, we could probably do so by awaiting tick() before setting a new non-null value.

PR #2599, opened 2023-07-28
Perform occlusion mode reset directly after adding notes

glutanimate:
Seems to fix the additional scenarios discussed in the #2596 comments

Does this look OK to you? Don't fully grok the implications of resetting the mode here, but if it works on image selection button click, I'd imagine that this is safe, too.

One caveat: `resetIOImageLoaded()` is called from Python and thus only pertinent on desktop. If #2596 also occurs on mobile we might have to solve this differently (though I wasn't able to reproduce this and would imagine that it's impossible as each mask editing session happens in a new web view context, right?)

### DIFF
--- a/ts/editor/NoteEditor.svelte
+++ b/ts/editor/NoteEditor.svelte
@@ -426,6 +426,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         if (page) {
             page.remove();
         }
+        imageOcclusionMode = undefined;
     }
     globalThis.resetIOImageLoaded = resetIOImageLoaded;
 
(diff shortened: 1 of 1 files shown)


glutanimate:
Likely superseded by #2600 as `imageOcclusionMode` resets are now performed on `saveNow`. Will need to confirm once both #2600 and #2598 are merged.

dae:
> Does this look OK to you?

It would probably be better to await tick(), as otherwise we may not be guaranteed that the previous instance had been removed yet, as Svelte does updates periodically and not immediately.

>  as each mask editing session happens in a new web view context, right?

Yep, that's correct.

glutanimate:
@dae thanks for the insights on `tick()`, makes a lot of sense!

Closing this as I can confirm the problem is fixed with #2600

PR #2600, opened 2023-07-29
follow-up pr for io button in note editor

krmanik:
In this PR the reset image and io mode (hide all & hide one) implemented. The button for mobile client for two mode moved to toolbar. The option to change mode in toolbar is helpful in add, edit and browse mode.
The reset image feature need help, I have tried to implement but the option is only available to desktop client and didn't find better solution for both mobile and desktop. In browse mode changing the note does not update mask editor so there is fix in this PR.

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -115,8 +115,8 @@ class AddCards(QMainWindow):
         self.addButton.setToolTip(shortcut(tr.adding_add_shortcut_ctrlandenter()))
 
         # add io button
-        self.io_add_button = bb.addButton(f"{tr.actions_add()} {downArrow()}", ar)
-        qconnect(self.io_add_button.clicked, self.onAddIo)
+        self.io_add_button = bb.addButton(f"{tr.actions_add()}", ar)
+        qconnect(self.io_add_button.clicked, self.add_io_note)
         self.io_add_button.setShortcut(QKeySequence("Ctrl+Shift+I"))
 
         # close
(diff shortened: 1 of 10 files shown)


glutanimate:
Cheers for the quick follow-ups, Mani! I ran into the browser issue the other day and was thinking about looking into that, but came here to see that you'd already fixed it. Nice! Can confirm that all works well here now.

On the **mode switch**:

This works well for me and I like that you integrated the dropdown into the toolbar. The icon indicator is a bit subtle compared to a text dropdown, so I could see users wondering where the option is located, but I assume people will get used to it fairly quickly. If you find the time, my only wishlist item there would be to persist the last chosen mode across multiple sessions, as I think users will find that more important now that the option is a bit more hidden and it takes more mouse traversal to get to the "Add" button.

On the **image switch**:

Thanks for looking into that as well! This works for me as well, but I did notice a number of smaller issues. You might already be aware of them, but just in case:

- Changing the image currently does not seem to work in AddCards. I'm not sure if this is a bug or by design, but generally speaking, I think that changing the image when creating occlusions for the first time is more or less equally useful as changing it when editing. So if it's not too much work, I would vote to also enable this in AddCards mode.
- Image changing currently does not seem to handle differences in aspect ratio that well. It seems like the new image is fit into the existing dimensions of the canvas, which can distort the image if it has a different aspect ratio than the previous one.
- Changing the image also seems to leave some frame artifacts in the mask editor (see screenshot below)
- Finally, when changing the image, the masks currently disappear for me until I click on the canvas again.

Screenshot of issue 2&3:

io-change-image-glitches

Also, one final observation: It seems like there's a small regression in this PR that's similar to #2485, where the mask editor will remain blank under some scenarios. The scenario I've seen this happen is as follows:

1. In AddCards, switch to Image Occlusion
2. Before adding the note, switch back to a different note type, e.g. Basic
3. Switch back to Image Occlusion

Expected result: We are back in the mask editor

Actual result: Editor is blank

dae:
I like the move of the all/one toggle to the toolbar!

While there are some issues that still need addressing, because both you and @glutanimate are working on the same code at the moment, I'll merge this in now to reduce the chances of conflicts.

krmanik:
Currently issues are image not reset on changing it and resize of window should resize image also. I am working on the fix.

PR #2601, opened 2023-07-31
Fixed MIME handling for non-standard umask(1) values

nullfs:
I set `umask` to 027, which apparently causes the `xdg-mime` commands to create unreadable entries in _/usr/share/mime_ when run as root. This has bricked my DE on at least one occasion so I think it would be reasonable to ensure that the contents of this directory remain readable


dae:
Sorry, a recursive chmod seems inappropriate in an install script. Your issue is probably better fixed either in xdg-mime or your system.

PR #2602, opened 2023-08-03
fix: reset image button in io note editor

krmanik:
The PR fixes the reset image option in io note editor.

### DIFF
--- a/ts/editor/NoteEditor.svelte
+++ b/ts/editor/NoteEditor.svelte
@@ -42,6 +42,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 <script lang="ts">
     import { bridgeCommand } from "@tslib/bridgecommand";
     import * as tr from "@tslib/ftl";
+    import { resetIOImage } from "image-occlusion/mask-editor";
     import { onMount, tick } from "svelte";
     import { get, writable } from "svelte/store";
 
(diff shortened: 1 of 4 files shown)


ts/editor/NoteEditor.svelte
```diff
@@ -411,6 +417,15 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
             }
         }
 
+        // add event on window resize if event exist then do not add again
+        if (!window.onresize) {
+            window.onresize = () => {
+                if (imageOcclusionMode?.kind === "add") {
+                    resetIOImage(imageOcclusionMode.imagePath);
```
dae, line 424: I think this might be done more cleanly with Svelte's onMount(), perhaps in MaskEditor.svelte?

ts/image-occlusion/MaskEditor.svelte
```diff
@@ -44,6 +48,12 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
             });
         }
     }
+
+    onMount(() => {
+        window.onresize = () => {
+            setCanvasZoomRatio(canvas, instance);
```
dae, line 54: A couple of tidy-ups please:

- addEventListener('resize', ...) would be better, as window.onresize could cause conflicts with other code down the line
- The listener should be removed when the component is unmounted, which you can do with onDestroy

dae:
Thanks Mani!

Issue #2608, opened 2023-08-13
Search and replace dialogue: item skipped by keyboard navigation


arkhandir:
While navigating by keyboard (`tab`), the first checkbox listing is skipped (not selected) in the search and replace dialogue, it is selected instead after the last checkbox listing.
Anki ⁨2.1.65 (aa9a734f)⁩ and Version ⁨2.1.66 (46915605)⁩
Python 3.9.15 Qt 6.5.0 PyQt 6.5.0
macOS 12.6.8 
![2023-08-13 20 02 38](https://github.com/ankitects/anki/assets/1059703/cd5b38ac-55de-4fad-9c0a-b36f9be68a63)

dae:
This probably just requires opening the .ui file in Designer and adjusting the tab order list.

PR #2609, opened 2023-08-14
fixed issue #2608

MiVHQW:
It was indeed the tabstop order. I aligned the order with the correct ui order.


MiVHQW:
I will open a new request with the addition of adding my name to the contributors file, so that the tests can pass.

PR #2610, opened 2023-08-14
fixed issue #2608 

MiVHQW:
It was indeed the tabstop order. I fixed the order in a way that it now correctly aligns with the ui order.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -133,6 +133,7 @@ Derek Dang <github.com/derekdang/>
 Luc Mcgrady <github.com/Luc-Mcgrady>
 Kehinde Adeleke <adelekekehinde06@gmail.com>
 Marko Juhanne <github.com/mjuhanne>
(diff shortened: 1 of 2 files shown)


dae:
I appreciate the contribution, but did you test this? The last two checkboxes are still tabbing out of order.

MiVHQW:
Actually not; sorry Damien for not testing this. Since I found the tabstob lines in exactly the wrong order shown in the screenshot in the .ui file for this window, the change seemed so trivial to me, that I did not test it.

Now I know, that there are no trivial changes :o I will close this request and try to find another solution and next time test it prior to submitting regardless of the triviality. It is interesting though, that this was not the solution. 🤔

MiVHQW:
I found the solution: It was a wrong name in one of the tabstops. As a result this tabstop was skipped. I submitted a new pull request after also testing it this time and it worked fine.

Issue #1895, opened 2022-06-02
Import/Export todos


RumovZ:
With basically all of the old code ported to Rust, it's time to turn to new features and improvements. This issues is supposed to track such todos, which have previously been scattered across various PRs and issues.
Some discussion is on #1018 and this list will be updated according to the progression over there.

General
---

- [ ] Custom exporters/importers.

Apkg
---

- [x] Check invalid ids on import (ids 1000x too large).
- [ ] Notetype remapping on import.
- [ ] Private fields.
- [ ] Private tags.
- [x] Update with older mtime. (https://github.com/ankitects/anki/pull/2612)
- [ ] Generate cards on import (#1836).
- [x]  Populate media DB on import (#1831).
- [x] Export filtered decks (without conversion to regular ones). (https://github.com/ankitects/anki/pull/1743#discussion_r840255927)
- [ ] Warn about extraneous exported cards and decks due to scattered siblings.
- [x] Allow importing without scheduling, if file contains some. (https://github.com/ankitects/anki/pull/2612)

JSON
---

More notes on https://github.com/ankitects/anki/pull/1850#issuecomment-1122156181.
-   [ ] Validation checks.
-   [ ] Error handling. (More specific errors, don't abort if failure results from target collection state, ...)
-  [ ] Support importing new cards (https://github.com/ankitects/anki/pull/1850/files#r869037548.
-  [ ] Port SuperMemo and Pauker importers to JSON, or figure out if we can map the legacy NoteImporter code to JSON automatically (or semi-automatically)
- [ ] Extract Mnemosyne, SuperMemo and Pauker importers into separate add-ons

CSV
---

-  [x] Preview for first few records.
-  [x] Optionally export notetype and deck columns. (Let user decide if by name or id?)
-  [x] Handle failures in `GetCsvMetadata`.
-  [x] Support note id/guid column. (just guid)
-  [ ] Report unmapped columns.
-  [x] Maybe add tooltips to import options. (#2655)
-  [x] Restrict duplicate check to current deck (https://github.com/ankitects/anki/pull/2372)

From https://forums.ankiweb.net/t/import-problems-suggestions/5856. Might also apply to JSON.
- [ ] Batch importing.
- [ ] Import media files from same folder.
- [ ] Import notetypes.
- [x] Create decks on import.

dae:
> Import deck options without scheduling information.

[...]

PR #2612, opened 2023-08-19
Merging Notetypes on Import

### LINKED ISSUE #2482: New importer creates new copy of remapped notetypes on each import
When we encounter a notetype schema conflict in .apkg import, we add the incoming notetype with a new id. This happens each time the file is imported, so if the user imports the file multiple times, they end up with a bunch of duplicate notetypes.

The way the original importing code handled this is to add remapped notetypes with an incremental id, and reuse an existing one with a matching schema. I briefly brought this up on https://github.com/ankitects/anki/pull/1743#issuecomment-1082569967, but it looks like we may have forgotten about it.

Ideally, we would reuse a previous remapped notetype if available, as if the user has a workflow that involves regularly importing updated versions of the shared deck, they'll end up with lots of different notetypes.

RumovZ:
Fixes #2482.

As you can see, the problem is with the help modal (and also the revert arrow, but I haven't spend much time looking into that). Would appreciate your help.

grafik


dae:
Modal issue turned out to be fairly simple thankfully - while one would logically expect HelpModal to be self-contained, it relies on global bootstrap styles.

RumovZ:
Thanks! I'm lost regarding those dependencies in the web layer. 🙂
Maybe it would make sense to add an `OptionsPage.svelte` component which would be an abstraction of `DeckOptionsPage.svelte`. Presumably, the number of options dialogs in the web layer will only grow.

rslib/src/notetype/fields.rs
```diff
@@ -38,6 +38,7 @@ impl NoteField {
             ord: None,
             name: name.into(),
             config: NoteFieldConfig {
+                id: Some(rand::random()),
```
RumovZ, line 41: I wonder if this is sufficient. If my math is correct, there's a 0.003% chance to have an id collision within a notetype with 500 fields ([Birthday Problem](https://en.wikipedia.org/wiki/Birthday_problem)). Is there an upper bound on the number of fields?
I wanted to avoid potential future problems with JS precision, but maybe I should still use an u64 and null the first 11 bits.

RumovZ:
There's also a more general issue I could use help with: If a job fails, I don't get any helpful error message. E.g. I'm currently struggling to add a field to a proto message, but the only output I get is this:
```
C:\code\anki [notetype-import ↑3 +1 ~5 -0 !]> ./run
    Finished release [optimized] target(s) in 0.27s
failed: pylib:rsbridge

Build failed.
```
It's the same for other jobs and whether I use `ninja check`, `ninja check:eslint:import-anki-package` or whatever. Is it the same for you?

abdnh:
> If a job fails, I don't get any helpful error message.

@RumovZ I struggled with this while working on my last PR but forgot to report it. Very frustrating.

dae:
This was a Windows-specific issue - please try updating n2 using the link above.

dae:
Sorry, pushing that fix to the main branch caused GitHub to think it "fixed" this PR, not the linked comment.

rslib/src/notetype/fields.rs
dae: I don't think there's a hard limit, but the editor might struggle if too many are used. In any case, I think we're probably better off going with the safer option. JS precision shouldn't be an issue with protobuf-es - i64s are stored as BigInts.

RumovZ:
> This was a Windows-specific issue - please try updating n2 using the link above.

Works perfectly again, thanks.

dae:
I tried busting the Rust cache to fix the CI error, but another PR ran first, and the issue is back. Will try updating to the latest Rust version, and if the issue still exists, we should probably report it upstream.

dae:
Let's see how things go after you rebase this. :-)

RumovZ:
Works again. 🤷

What do you think about also writing a little add-on for retroactively adding template and fields ids? This would allow deck authors to take full advantage of the merge feature with existing decks going forward.

dae:
Sorry for the delay. What do you think about doing this in Check Database instead? That way it solves the problem even for mobile-only users, and avoids users needing to download extra stuff.

RumovZ:
The problem is that if you add IDs to an imported notetype, they will conflict with the IDs the original creator may have added, making a correct merge impossible.
So we only want deck authors to add IDs to their shared notetypes.

dae:
:man_facepalming: that makes sense. Add-on does seem like the best option then :-)

RumovZ:
I sneaked in some points from https://github.com/ankitects/anki/issues/1895. Regarding error handling, should we expose a complete `BackendError` to TS callers, at least optionally? Then they could do things like link to the attached help page.

Ready for review now. I guess you want to merge soon in order to fix the link checks.

ftl/core/importing.ftl
```diff
@@ -30,13 +30,22 @@ importing-map-to = Map to { $val }
...
+    Warning: This will necessitate a full sync and may mark existing notes as modified.
 importing-mnemosyne-20-deck-db = Mnemosyne 2.0 Deck (*.db)
 importing-multicharacter-separators-are-not-supported-please = Multi-character separators are not supported. Please enter one character only.
 importing-notes-added-from-file = Notes added from file: { $val }
 importing-notes-found-in-file = Notes found in file: { $val }
 importing-notes-skipped-as-theyre-already-in = Notes skipped, as up-to-date copies are already in your collection: { $val }
 importing-notes-skipped-update-due-to-notetype = Notes not updated, as notetype has been modified since you first imported the notes: { $val }
 importing-notes-updated-as-file-had-newer = Notes updated, as file had newer version: { $val }
+importing-omit-scheduling = Omit scheduling
+importing-omit-scheduling-help = If checked, no scheduling information the file may contain will be imported.
```
dae, line 48:
```suggestion
importing-omit-scheduling-help = If enabled, Anki will remove any previous reviews that the deck sharer included, and all cards will appear as new cards.
```

ftl/core/importing.ftl
```diff
@@ -57,6 +66,17 @@ importing-unable-to-import-from-a-readonly = Unable to import from a read-only f
 importing-unknown-file-format = Unknown file format.
 importing-update-existing-notes-when-first-field = Update existing notes when first field matches
 importing-updated = Updated
+importing-update-if-newer = If newer
+importing-update-always = Always
+importing-update-never = Never
+importing-update-notes = Update notes
+importing-update-notes-help =
+    When to update a note with the imported equivalent. By default, this is only done
```
dae, line 74:
```suggestion
    When to update a note an existing note in your collection. By default, this is only done
```

ftl/core/importing.ftl
```diff
@@ -57,6 +66,17 @@ importing-unable-to-import-from-a-readonly = Unable to import from a read-only f
...
+    if the imported note was more recently modified.
+importing-update-notetypes = Update notetypes
+importing-update-notetypes-help =
+    When to update a notetype with the imported equivalent. By default, this is only done
```
dae, line 78:
```suggestion
    When to update an existing notetype in your collection. By default, this is only done
```

ftl/core/importing.ftl
```diff
@@ -57,6 +66,17 @@ importing-unable-to-import-from-a-readonly = Unable to import from a read-only f
 importing-unknown-file-format = Unknown file format.
 importing-update-existing-notes-when-first-field = Update existing notes when first field matches
 importing-updated = Updated
+importing-update-if-newer = If newer
+importing-update-always = Always
+importing-update-never = Never
+importing-update-notes = Update notes
+importing-update-notes-help =
+    When to update a note with the imported equivalent. By default, this is only done
+    if the imported note was more recently modified.
```
dae, line 75:
```suggestion
    if the matching imported note was more recently modified.
```

ftl/core/importing.ftl
```diff
@@ -57,6 +66,17 @@ importing-unable-to-import-from-a-readonly = Unable to import from a read-only f
...
+importing-update-notetypes = Update notetypes
+importing-update-notetypes-help =
+    When to update a notetype with the imported equivalent. By default, this is only done
+    if the imported notetype was more recently modified.
```
dae, line 79:
```suggestion
    if the matching imported notetype was more recently modified.
```

proto/anki/import_export.proto
```diff
@@ -45,8 +47,21 @@ message ExportCollectionPackageRequest {
   bool legacy = 3;
 }
 
+message ImportAnkiPackageOptions {
+  enum UpdateCondition {
```
dae, line 51: Minor and not a required change. From https://protobuf.dev/programming-guides/style/:

>  Prefer prefixing enum values instead of surrounding them in an enclosing message.

I used to think this was silly ("why should I have to repeat the prefix each time just because C++ is bad?). But I've come around to it, because the nesting approach is rather awkward, especially in Rust. If you have two Foo enums nested in messages Bar and Baz, you have to import bar::Foo and baz::Foo, and it gets confusing what you're referring to. If instead you define the enums at the top level and give them a suitable name (eg BarFoo and BazFoo), it's immediately obvious which one you're referring to, and you don't need to deal with imports from submodules.  There's no longer a need to deal with multiple enums with the same name. I've started to feel the same way about other nested messages as well - keeping them at the top level and giving them a more specific name tends to make things easier (eg ImportAnkiPackageUpdateCondition). The prefixes it requires can end up rather hideous though unfortunately. I wish there was a decent RPC scheme that had the language support Protobuf does, but had better first-class support for things like discriminated unions.

rslib/src/import_export/package/apkg/import/mod.rs
```diff
@@ -21,17 +21,26 @@ use zip::ZipArchive;
...
 use crate::import_export::package::Meta;
+use crate::import_export::package::UpdateCondition;
 use crate::import_export::ImportProgress;
 use crate::import_export::NoteLog;
 use crate::media::MediaManager;
 use crate::prelude::*;
 use crate::progress::ThrottlingProgressHandler;
 use crate::search::SearchNode;
 
+/// New template indices by old template indices for a given notetype.
```
dae, line 36:
```suggestion
/// A map of old to new template indices for a given notetype.
```

dae: (assuming I've understood the meaning of that comment correctly)

dae (commented):
There's a couple of files I haven't reviewed yet, and haven't done any testing so far, but everything I've seen so far looks good as usual :-) Will do my best to get this over the line tomorrow.

dae:
> should we expose a complete BackendError to TS callers

Any thoughts on how we might do that? We could perhaps have a 400 or 500 page that has a content type of application/binary be interpreted as protobuf error, though I haven't looked into whether that would be practical or not.

RumovZ:
Sorry if I'm being naive. I haven't looked into it deeply. I would have assumed we could just construct the message in `mediasrv.py` and `BackendError.fromBinary()` the response in `postProtoInner()`. Whether we look at the status code or just `try` would be a detail.

dae:
The issue I was thinking about is that we can't be guaranteed that when result.ok is false, that we'll get valid protobuf error data back in result.blob. For example, if we accidentally forget to implement a method, mediasrv.py will return a 404 code with "not found", which will not parse as the error, or if we've added this 'return the error as protobuf' code to only specific methods and not all of them. We'd also need to consider cases where there's a backend error vs some other error - for example set_scheduling_states() never calls into the backend, so if something went wrong in the call, we'd have to synthesize an appropriate backend error if the ts code always expected one.

dae:
Perhaps resetting to new should be the default? I tried it and it worked well.

I then modified a notetype in the source collection, exported, and tried importing again into another profile that had previously imported the content. When I imported with 'merge notetypes', it correctly applied the changed notes. :+1: 

I then reverted the target collection to a backup and tried again with 'merge notetypes' disabled, and it told me it couldn't update the note, as the notetype had changed. Theoretically, couldn't we at least update fields that exist in both the source and target collections, even if we're not making changes to the notetypes?

What was also somewhat surprising is if I then retried the import with 'merge notetypes' enabled again, it continued to tell me that the notetype had changed - so if the user doesn't have that setting enabled from the start, they end up in a state where they can never get the notes to update. I wonder if there'd be a way to avoid that?

Minor other issues:

- 'missing key' is shown when hovering over the import button
- 'always update' does exactly what it says, though users may perhaps expect it to skip notes that are identical in the source and target collection
- the distinction between merging and updating notetypes is a bit unclear. I presume the latter is referring to non-schema-change things like CSS and template tweaks? If so, might be worth explaining that in a bit more detail in the help pop-up.

dae:
Ah hah - I can recover from the 'never can update' state by deleting the unused extra notetype that got added when importing with 'merge notetypes' disabled. This won't be obvious to users, so I wonder if we could improve things somehow.

RumovZ:
> we can't be guaranteed that when result.ok is false, that we'll get valid protobuf error data back in result.blob.

Can't we just synthesize the error on the TS side when parsing the response as a BackendError fails?

> Perhaps resetting to new should be the default?

No objections, but then I'll change the name, so `false` continues to be the default.

> Theoretically, couldn't we at least update fields that exist in both the source and target collections, even if we're not making changes to the notetypes?

Yes, that should be possible. Should we maybe split that off into a different PR? Partial updates would be a whole new thing.

> if I then retried the import with 'merge notetypes' enabled again, it continued to tell me that the notetype had changed

I see, that's tricky. For reference:
1. Note A with notetype N is imported.
2. Later with `Merge notetypes` disabled, note B also with notetype N is imported, but the schema of N has changed, and a new version N* is created.  
3.  Updates for A and B are imported with `Merge notetypes` enabled. We find a notetype with a matching schema, N*, so no merging takes place. B can be updated, A cannot.

In 3., if instead we reused N, which has a matching id, but not schema, we could update A, but not B. So we need to merge the incoming notetype with *all* previously imported versions (N, N*, N**, ...).

> 'missing key' is shown when hovering over the import button

✔️ 

> always update' does exactly what it says, though users may perhaps expect it to skip notes that are identical in the source and target collection

✔️ 

> the distinction between merging and updating notetypes is a bit unclear.

I tried to explain it a bit more. Please feel free to improve it further.
Also, schema changes and conflict-causing changes are not exactly the same thing:
1. You may import a notetype with the sort field changed without conflict, but that will trigger a full sync, as it's a schema change.
5. Renaming fields and templates is not a schema change, but will cause conflicts on import (unless they have ids).

But maybe we can spare the users from these details.

<hr>

FYI, I won't have much time continuing work on this before Sunday.

dae:
> Can't we just synthesize the error on the TS side when parsing the response as a BackendError fails?

If by synthesize you mean create an error with whatever text we received in it, that works for me too. I just would not want to lose the text message if we can, as it will make debugging a bit trickier.

> No objections, but then I'll change the name, so false continues to be the default.

Thanks, that'd be great.

> Should we maybe split that off into a different PR?

Sounds good.

> I see, that's tricky. For reference:

In my case I'd changed the notetype but not added any new notes, so the newly-imported notetype was unused. Not sure how common that would be compared to cases where both notetype is changed and notes are added though. If you don't see any immediate solution for this, let's spin it of into a future task too.

> I tried to explain it a bit more

Looks good; thanks for that. No rush on timing - I've been a bit busy with FSRS and AnkiDroid work recently, and @glutanimate is still working on I/O related changes he wants to land before the next beta.

RumovZ:
I wasn't able to test or run after the last rebase due to a familiar error:
```
C:\code\anki [notetype-import ↓36 ↑55]> .\run.bat
    Finished release [optimized] target(s) in 0.28s
failed: build:configure
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DidNotExecute { cmdline: "\\rust\\debug\\configure.exe", source: Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." } }', build\runner\src\run.rs:86:30
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `tools\\..\out\rust\debug\configure.exe`
thread 'main' panicked at 'GenPythonProto must generate at least one output', C:\code\anki\build\ninja_gen\src\build.rs:261:13
error: process didn't exit successfully: `tools\\..\out\rust\debug\configure.exe` (exit code: 101)
thread 'main' panicked at 'assertion failed: status.expect(\"ninja\").success()', build\runner\src\build.rs:152:5
```
As `configure.exe` does in fact exist, I assume the issue is something different. Since it's also affecting the main branch and resolved by itself after a while in the past, I assume it's a cache thing. However, I've emptied `out`, `target` and `node_modules` to no avail. Do you have any other ideas?

dae:
Please update to the latest n2 again, using tools\install-n2

dae:
(https://github.com/ankitects/anki/commit/e07e60495bb6595f61d2bab6e4f32c986f2db60f)

dae:
In the latest commit, I noticed that the order fields are merged in is unpredictable, presumably due to the use of a hashmap for storing the notetypes. I presume this is ok, but thought I'd check with you first. Apart from that, assuming my other changes look good to you, this is ready to merge. Thanks for looking into the merging of previous notetypes - the new behavior is much less surprising. Combined with @abdnh's work on the import log screen, this should hopefully cut down on a lot of confusion about the importing process!

RumovZ:
Thanks, that fixed running, unfotunately I'm still having trouble with other jobs:
```
C:\code\anki [notetype-import ↑1 +0 ~2 -0 !]> .\tools\ninja.bat check
    Finished release [optimized] target(s) in 0.28s
failed: cargo-nextest
    Updating crates.io index
  Installing cargo-nextest v0.9.43
    Blocking waiting for file lock on package cache
    Updating crates.io index
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
   Compiling nextest-runner v0.30.0
error[E0308]: mismatched types
    --> C:\Users\lissy\.cargo\registry\src\index.crates.io-6f17d22bba15001f\nextest-runner-0.30.0\src\runner.rs:1476:32
     |
1476 |             job.assign_process(handle)?;
     |                 -------------- ^^^^^^ expected `winapi::ctypes::c_void`, found `c_void`
     |                 |
     |                 arguments to this method are incorrect
     |
     = note: `c_void` and `winapi::ctypes::c_void` have similar names, but are actually distinct types
note: `c_void` is defined in crate `core`
    --> C:\Users\lissy\.rustup\toolchains\1.72-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\core\src\ffi\mod.rs:213:1
     |
213  | pub enum c_void {
     | ^^^^^^^^^^^^^^^
note: `winapi::ctypes::c_void` is defined in crate `winapi`
    --> C:\Users\lissy\.cargo\registry\src\index.crates.io-6f17d22bba15001f\winapi-0.3.9\src\lib.rs:38:5
     |
38   |     pub enum c_void {}
     |     ^^^^^^^^^^^^^^^
note: method defined here
    --> C:\Users\lissy\.cargo\registry\src\index.crates.io-6f17d22bba15001f\win32job-1.0.2\src\job.rs:99:12
     |
99   |     pub fn assign_process(&self, proc_handle: HANDLE) -> Result<(), JobError> {
     |            ^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0308`.
error: could not compile `nextest-runner` (lib) due to previous error
error: failed to compile `cargo-nextest v0.9.43`, intermediate artifacts can be found at `C:\code\anki\tools\\..\out\rust`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ReturnedError { cmdline: "cargo install --color always cargo-nextest --version 0.9.43 --root out", code: Some(101) }', build\runner\src\run.rs:86:30

Build failed.
```

I'm a bit confused about the new test. I don't see why the nested transaction would trigger a panic there, when it doesn't in the test from 2e112be. And if I revert 1e7e5e4, it still passes. (Not saying, we should; just want to understand where the panic is coming from.)

Got rid of the notetype map. It was usually called with `.values()` anyway, and even where not, a vec should usually be faster.

Regarding 71db356, I realize that makes it a bit harder for translators, but it would also be very confusing if the setting wasn't translated consistently.
Actually, we could write a helper to check if any translation has violated the template syntax by comparing their ASTs.  🙂

dae:
Re the build error, please try updating to the latest commit on main.

I'm not sure what happened with the test - I did roll back the fix to check the test failed, but maybe I got confused by the intermittent failures. I've rewritten the test and confirmed it does exercise that path now.

Template change sounds good. Comparing the ASTs might be tricky, as a translator may have introduced plural handling into a string that doesn't handle it in English (eg when translating: "Count: $count")

dae:
Happy for me to merge this in now? Thanks for all your work on this Rumo.

abdnh:
Some issues I noticed:

1. Backslashes are not handled correctly when importing by double clicking on an apkg file.
image

<details>
<summary>Fix</summary>

```diff
diff --git a/qt/aqt/import_export/import_dialog.py b/qt/aqt/import_export/import_dialog.py
index 2318b47d3..6a6e00d81 100644
--- a/qt/aqt/import_export/import_dialog.py
+++ b/qt/aqt/import_export/import_dialog.py
@@ -2,6 +2,7 @@
 # License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
 from __future__ import annotations
+import json
 
 import aqt
 import aqt.deckconf
@@ -48,9 +49,9 @@ class ImportDialog(QDialog):
         self.setLayout(layout)
         restoreGeom(self, self.TITLE, default_size=(800, 800))
 
-        escaped_path = path.replace("'", r"\'")
+        escaped_path = json.dumps(path.replace("'", r"\'"))
         self.web.evalWithCallback(
-            f"anki.{self.SETUP_FUNCTION_NAME}('{escaped_path}');",
+            f"anki.{self.SETUP_FUNCTION_NAME}({escaped_path});",
             lambda _: self.web.setFocus(),
         )
         self.setWindowTitle(tr.decks_import_file())
```

</details>

2. Tooltip styles are missing.
image

<details>
<summary>Fix</summary>

```diff
diff --git a/ts/import-anki-package/import-anki-package-base.scss b/ts/import-anki-package/import-anki-package-base.scss
index 7c6cf484b..efcc43878 100644
--- a/ts/import-anki-package/import-anki-package-base.scss
+++ b/ts/import-anki-package/import-anki-package-base.scss
@@ -11,6 +11,7 @@
 @import "bootstrap/scss/modal";
 @import "bootstrap/scss/carousel";
 @import "sass/bootstrap-forms";
+@import "sass/bootstrap-tooltip";

 .night-mode {
     @include bootstrap-dark.night-mode;
```

</details>

3. This is up to preference, but just noting it down: the import log used to extend to the full width of the page but this is no longer the case for the apkg import page due to `width: min(100vw, 70em)`. The CSV dialog doesn't have this style though (I think I removed that in #2568). I guess the new screen and the CSV dialog looks better with constrained width, but for the import log page it can leave a lot of space that could be used by the details table.

dae:
Thanks, I've pushed the first two fixes. For 3, WDYT Rumo?

RumovZ:
Thank, @abdnh and @dae.
I agree that the import log should look consistent. Will make a PR.
I would also like to rework the CSV import dialog so that it looks like this one and has help utilities.

PR #2614, opened 2023-08-20
Change “Delete Note” shortcut to cmd+backspace

montyevans:
Per this [thread](https://forums.ankiweb.net/t/change-add-cmd-delete-as-delete-note-shortcut/32965)


dae:
Hmm. This will mean that people on Windows/Linux now need to use a new shortcut too. Perhaps we should only change the shortcut on Macs? There's an 'isMac' variable you can check for.

montyevans:
Hmmn. I think there's a SCSS mixin called `isMac`, but the only reference I can find in the Python code is [this](https://github.com/ankitects/anki/blob/dcf91fcc509fa377c1907caf9e0b4640c9e992d4/qt/aqt/utils.py#L1281) comment, but it might be out of date? I can't see an `isMac` variable among the global variables referenced there.

In any case, I think this change is probably net helpful on Windows / Linux too? I'm was a Windows user until last year, and I remember backspace much more common in shortcuts / keybindings than delete (i.e. forward delete). I generally understand  backspace to be the "default" delete, regardless of OS; in cases where deleting _forwards_ is useful / meaningful, then the OSes diverge on how to treat that. 

So I'd softly vote to keep this as is, but defer to you for cross-OS norms: if you have a pointer to the `isMac` variable that would be sweet!

montyevans:
Fair enough - just updated the PR accordingly. Thanks for the `is_mac` pointer: silly of me not to check :)

dae:
Thanks Monty!

PR #2618, opened 2023-08-23
Shift output suppression into n2

dae:
After updating with tools/install-n2, you should now be able to see
the last line of long-running commands like cargo invocations.


ahmedyarub:
@dae building on Windows is failing due to unexpected variables in ninja script (hide_success and hide_last_line). Removing these variables and their definitions solves the problem for me.

dae:
Thanks for letting me know; I've pushed a fix.

PR #2626, opened 2023-08-28
Allow searching for missing custom data properties

abdnh:
This extends `prop:cdn` to support searching for missing properties using something like `prop:cdn:r=none`.

Context: https://forums.ankiweb.net/t/more-flags-add-flags-with-custom-names-and-colors/33661/4

### DIFF
--- a/ftl/core/search.ftl
+++ b/ftl/core/search.ftl
@@ -27,7 +27,6 @@ search-invalid-whole-number = expected a whole number in "`{ $context }`", but f
 search-invalid-positive-whole-number = expected a positive whole number in "`{ $context }`", but found "`{ $provided }`".
 search-invalid-negative-whole-number = expected a whole number less than or equal to 0 in "`{ $context }`", but found "`{ $provided }`".
 search-invalid-answer-button = expected an answer button between 1-4 in "`{ $context }`", but found "`{ $provided }`".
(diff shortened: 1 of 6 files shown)


dae:
Hmm. If we introduce a way to search for strings as well in the future (https://github.com/ankitects/anki/issues/2627), this would presumably need to be done there too. What do you think about having some separate query like has-cd:foo? Hopefully it would make things a bit simpler. Apologies for only thinking about this after you'd already done the work. :-(

abdnh:
> What do you think about having some separate query like has-cd:foo?

Sounds reasonable. Will look into that.

abdnh:
I implemented this using yet another SQL function. Should we maybe add a general `extract_custom_data` function and remove `extract_custom_data_number`?

dae:
Code changes look good.

I'm ok with using a single extract_custom_data function that just returns strings. It would lead to possibly confusing ordering on values that were supposed to be numeric but weren't, but that's probably not a big deal. I think we'll need to keep the cdn search though, as we'll get different results with numbers and strings:

sqlite> select 1 < 3;
1 < 3 = 1
sqlite> select '1' < 3;
'1' < 3 = 0

Maybe cdn could be rewritten to do the equivalent of this?

sqlite> select cast('1' as float) < 3;
cast('1' as float) < 3 = 1

PR #2630, opened 2023-08-29
Support AVIF

nihil-admirari:
New image format, supported everywhere: https://caniuse.com/avif.

Currently, pictures can be manually copied to media folder, but attach file dialog does not support AVIF.


nihil-admirari:
> failures:
>   
>  check_links

seems unrelated to these changes.

dae:
Thanks Nil!

PR #2632, opened 2023-09-02
Implemented support to audio_did_seek_relative

evandrocoan:
I kind missed this feature on the first pull-request: https://github.com/ankitects/anki/pull/2340

Addons example usage: https://github.com/evandrocoan/AnkiAddons/commit/eca627a5f532bb55bd36d949c1a90ec49cccfa42

### DIFF
--- a/qt/tools/genhooks_gui.py
+++ b/qt/tools/genhooks_gui.py
@@ -218,12 +218,16 @@ hooks = [
     Hook(
         name="audio_did_pause_or_unpause",
         args=["webview: aqt.webview.AnkiWebView"],
-        doc="""Called when the audio is paused or unpaused.""",
+        doc="""Called when the audio is paused or unpaused.
+        This hook is triggered by the action in Anki's More menu or the related key bind.
+        The webview is provided in case you wish to use this hook with web-based audio.""",
     ),
     Hook(
         name="audio_did_seek_relative",
         args=["webview: aqt.webview.AnkiWebView", "seek_seconds: int"],
-        doc="""Called when the audio is sought forward (positive seek) or backwards (negative seek).""",
+        doc="""Called when the audio is sought forward (positive seek) or backwards (negative seek).
+        This hook is triggered by the action in Anki's More menu or the related key bind.
+        The webview is provided in case you wish to use this hook with web-based audio.""",
     ),
     # Debug
     ###################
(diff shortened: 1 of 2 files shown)


abdnh:
Maybe it makes sense to add the calls to seek_relative() in sound.py instead of reviewer.py? I also suggest using the `av_player_` prefix to match the other hooks.

evandrocoan:
> Maybe it makes sense to add the calls to seek_relative() in sound.py instead of reviewer.py?

It would require changing its interface to receive a webview object only to use it on this hook. I would say it is not friendly to mix them because av_player uses a native system player, while this hook uses the chromium webview (another component, although both play audio).

> I also suggest using the av_player_ prefix to match the other hooks.

I would not use this prefix because it is not an av_player hook, it is a hook for chromium webview. But instead of naming the hook `webview_audio_did_seek_relative`, I left the `webview` part out. We already have the `audio_did_pause_or_unpause` and `audio_will_replay` hooks  without this `webview_` prefix as this one, so we could keep it out for now.

dae:
You may be using it for controlling audio in the webview, but that's not the only way it could be used, so it's not inherently a webview-based hook. If this were the only hook, I'd suggest you remove the webview argument to make it clearer. But since it's already included in the hook above it, I guess it's better to be consistent. It might be worth adding a comment in those hooks like the following? "This hook is triggered by the action in Anki's More menu. The webview is provided in case you wish to use this hook with web-based audio."

evandrocoan:
I added the notes.

qt/tools/genhooks_gui.py
```diff
@@ -218,7 +218,16 @@ def custom_link_handler(url):
     Hook(
         name="audio_did_pause_or_unpause",
         args=["webview: aqt.webview.AnkiWebView"],
-        doc="""Called when the audio is paused or unpaused.""",
+        doc="""Called when the audio is paused or unpaused.
+        This hook is triggered by the action in Anki's More menu or the related key bind.
```
dae, line 222:
```suggestion
        This hook is triggered by the action in Anki's More menu or the related key binding.
```

qt/tools/genhooks_gui.py
```diff
@@ -218,7 +218,16 @@ def custom_link_handler(url):
...
+        The webview is provided in case you wish to use this hook with web-based audio.""",
+    ),
+    Hook(
+        name="audio_did_seek_relative",
+        args=["webview: aqt.webview.AnkiWebView", "seek_seconds: int"],
+        doc="""Called when the audio is sought forward (positive seek) or backwards (negative seek).
+        This hook is triggered by the action in Anki's More menu or the related key bind.
```
dae, line 229:
```suggestion
        This hook is triggered by the action in Anki's More menu or the related key binding.
```

dae:
Thanks!

PR #2640, opened 2023-09-06
Change Anki's version scheme; bump to 23.09

dae:
@glutanimate @abdnh could you confirm this is still getting your add-on update info correctly?


dae:
Thanks Abdo, the latest changes should fix that.

PR #2643, opened 2023-09-07
Expose video driver options for Qt6

### LINKED ISSUE #2625: Consider exposing gfx driver options in the GUI for Qt6
With Qt5, if the user selects a graphics driver that doesn't work on their system, we're able to detect it in msgHandler() and automatically select a different driver, so the user is not stuck being unable to open Anki. This no longer worked in Qt6, so the option was removed from the GUI, as we don't want users to make an easy change in the GUI and then be locked out.

A number of GUI versions have elapsed since then, and it may be time to check that this is still the case. For example, if vulkan is selected and fails on your systems as in https://forums.ankiweb.net/t/2-1-66-card-not-revealing/33603/5, does Qt let us know in the message handler?

If not, we should at least document the options described in that thread, as currently we only cover software/no software in the manual.

Some extra background: as of Qt 6.5.1, Qt switched from using OpenGL by default, to using DirectX11. I don't believe dx11 was an option when we originally switched to Qt6 - or if it was, I wasn't aware of it.

@RumovZ @abdnh if either of you wouldn't mind checking this when you have a chance, that'd be appreciated.

abdnh:
This exposes the video driver preference option for Qt6, and allows switching between the new [graphics backends](https://doc.qt.io/qt-6/topics-graphics.html).

## Changes

- In addition to the existing OpenGL and Software drivers, the following drivers are available depending on the platform: Metal, Vulkan, Direct3D 11.
- Anki now defaults to Direct3D on Windows, Metal on macOS, and Vulkan on other platforms.

Closes #2625


qt/aqt/preferences.py
```diff
@@ -404,15 +401,18 @@ def update_video_driver(self) -> None:
...
+                return tr.preferences_video_driver_software_mac()
+            else:
+                return tr.preferences_video_driver_software_other()
+        elif driver == VideoDriver.OpenGL:
+            if is_mac:
+                return tr.preferences_video_driver_opengl_mac()
+            else:
+                return tr.preferences_video_driver_opengl_other()
+
+    return driver.name
```
abdnh, line 418: @dae should we add strings for the new defaults (with a "recommended" label)?

qt/aqt/__init__.py
```diff
@@ -455,8 +463,9 @@ def msgHandler(category: Any, ctx: Any, msg: Any) -> None:
     if driver == VideoDriver.OpenGL:
         # Leaving QT_OPENGL unset appears to sometimes produce different results
         # to explicitly setting it to 'auto'; the former seems to be more compatible.
-        pass
-    else:
+        if qtmajor > 5:
+            QQuickWindow.setGraphicsApi(QSGRendererInterface.GraphicsApi.OpenGL)
```
abdnh, line 467: While QQuickWindow is used here, this should affect QWidgets too. You can confirm that by checking `mw.windowHandle().surfaceType()`.

dae:
Looking good! Some comments:

- I think OpenGL should be the default on Linux, as that's the Qt default as far as I know
- vulkan should be excluded from Macs; when I tried it, Anki crashed on startup without Qt ever printing a message

qt/aqt/preferences.py
dae: What if we just used English strings like "Metal", and added a fluent key so we can add "default" to whatever is the platform default - e.g. "Metal (default)"? "default" might be a better choice than "recommended", because what's most suitable depends on the user's system.

qt/aqt/__init__.py
abdnh: Setting `QSG_RHI_BACKEND` here didn't work too BTW. It only works if I set it before all Qt imports.

dae:
Thanks Abdo!

PR #2646, opened 2023-09-08
Add ability to tab to DuplicateLink

SteelColossus:
The change that allows the DuplicateLink to be tabbed to is the removal of the `tabindex`. The other changes are to make the link only as wide as the text, otherwise the tab outline appears around the whole `<span>`, which would make it look weird.

Here is a screenshot of what the link looks like when it's been tabbed to after the change:

image

This issue was mentioned previously in https://github.com/ankitects/anki/issues/2564#issuecomment-1672324190.


dae:
Thanks for this!

PR #2647, opened 2023-09-09
Refactor media sync handling

dae:
- The media USN is now returned in sync/meta, which avoids an extra round-trip.
- Media syncing is now automatically started by the syncing code at the end of a normal or full sync, which avoids it competing for bandwidth and resources, and avoids duplicate invalid login messages when the auth token is invalid.
- Added a new media_sync_progress() method to both check if media is syncing, and get access to the latest progress.
- Updated the sync log screen to only show the latest line, like AnkiMobile.
- Show media sync errors in a pop-up, so they don't get missed. Use a non-modal pop-up to avoid potential conflicts with other modals.

### DIFF
--- a/proto/anki/collection.proto
+++ b/proto/anki/collection.proto
@@ -8,6 +8,7 @@ option java_multiple_files = true;
 package anki.collection;
 
 import "anki/generic.proto";
+import "anki/sync.proto";
 
 service CollectionService {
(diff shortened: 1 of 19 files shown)


dae:
@abdnh mind giving this a quick look-over?

abdnh:
Looking good to me.

dae:
Thanks Abdo!

PR #2648, opened 2023-09-09
Prevent main styling from leaking into import log

RumovZ:
https://github.com/ankitects/anki/pull/2612#issuecomment-1712320801


abdnh:
The outer horizontal margin is gone. Maybe we also need some margin at the top.

image

RumovZ:
I will address this in a new PR to unify the two import dialogs.

Adding some whitespace to the log page sounds good, but I'm not much help when it comes to styling.

PR #2654, opened 2023-09-12
Integrate FSRS into Anki

### LINKED ISSUE #2443: Integrate FSRS into Anki as an optional feature
Currently users need to paste code into the custom scheduling section of the deck options in order to use FSRS. It may make sense to integrate this code directly into Anki, to make it easier to use.

According to Jarrett, the scheduling code does not change frequently: https://github.com/ankitects/anki/pull/2421#issuecomment-1457925875

Users may wish to apply different weights to different decks. The existing custom scheduling code relies on deck name matches (prefix matches?) in order to determine the weights. We could perhaps store the weights in the deck presets - this would allow a deck and its subdecks to share the same weights. Alternatively the weights could be stored in individual decks, which is the most flexible, but will lead to extra storage. The weights do not appear to be particularly large: https://github.com/ankitects/anki/pull/2421#issuecomment-1460046666

This would be an optional feature, ~and would require the add-on/weights trained externally for now. If it is successful, it may make sense to look into a Rust-based training method in the future, so that users don't need to rely on external sites/programs~.

@L-M-Sherlock @RumovZ any thoughts?

dae:
Feel free to have a play with this, but I recommend you make backups and keep an eye on the card stats, as it's received minimal testing so far.

A couple of notes:

- Difficulty has been remapped to a 0-1 value in Anki, and is shown in percentages in the browser/stats, as that implicitly tells the users what the min and max values can be, and is consistent with retrievability.
- S/D aren't stored in custom data, so they won't conflict with fsrs4anki & the helper. The values can be accessed with card.fsrs_memory_state in Python, and normal.review.fsrsMemoryState in custom scheduling.
- I have not yet had time to look into any of the points raised on https://github.com/open-spaced-repetition/fsrs-rs/issues/63
- Anki has separate paths for when cards are overdue or due in the future (filtered decks). Was it correct to treat them the same for FSRS, or should we special-case these too?

Feedback welcome - I'm sure there are things I've missed!

Closes #2443


L-M-Sherlock:
> * Anki has separate paths for when cards are overdue or due in the future (filtered decks). Was it correct to treat them the same for FSRS, or should we special-case these too?

The FSRS scheduler deal with them in the same way. Here is related code: https://github.com/open-spaced-repetition/fsrs4anki/blob/5a34201f4d5b3a9cc1a5da4511b6bc32b7c6e909/fsrs4anki_scheduler.js#L278-L290

```javascript
function is_review() {
  if (states.current.normal?.review !== undefined) {
    if (states.current.normal?.review !== null) {
      return true;
    }
  }
  if (states.current.filtered?.rescheduling?.originalState !== undefined) {
    if (Object.hasOwn(states.current.filtered?.rescheduling?.originalState, 'review')) {
      return true;
    }
  }
  return false;
}
```

L-M-Sherlock:
image

For new card, it shouldn't have stability. We need to skip new cards when updating parameters.

L-M-Sherlock:
image


The Difficulty should not be negative.

dae:
Memory state issue fixed by https://github.com/ankitects/anki/pull/2654/commits/69e6c6742d4d769370e5056605a9cb70699423ee

L-M-Sherlock:
image

This card still hasn't been updated.

dae:
A single learning step is excluded in training. Are you saying that it should not be excluded when calculating the memory state?

L-M-Sherlock:
Yes. I exclude single learning step in training, because it doesn't have review history for the model. But in calculating the memory state, singel learning step is enough to predict the memory state.

rslib/src/scheduler/fsrs/weights.rs
```diff
@@ -95,40 +105,63 @@ fn anki_to_fsrs(revlogs: Vec<RevlogEntry>, next_day_at: TimestampSecs) -> Vec<FS
...
-    entries.drain(..index_to_keep);
-
-    // we ignore cards that don't start in the learning state
-    if let Some(entry) = entries.first() {
-        if entry.review_kind != RevlogReviewKind::Learning {
-            return None;
+/// Transform the revlog history for a card into a list of FSRSItems. FSRS
+/// expects multiple items for a given card when training - for revlog
+/// `[1,2,3]`, we create FSRSItems corresponding to `[1,2]` and `[1,2,3]`.
+fn single_card_revlog_to_items(
```
L-M-Sherlock, line 129: We can pass in an arg to control whether to send out [[1], [1,2], [1,2,3]] or [[1,2], [1,2,3]].

L-M-Sherlock: For calculating, we need to send out [1] for revlog [1].

dae:
Does the latest commit do what you expect?

dae:
Sorry, I mean https://github.com/ankitects/anki/pull/2654/commits/3d1af8c474f69bd732fc3e907f5db15256361aa7

L-M-Sherlock:
image

The weights are NaN after e1b555520cb5058db45f26e167dc5eee65278cb3. But I test the collection in fsrs-rs and get a correct result.

dae:
A question about updating memory state when the user has learning steps. single_card_revlog_to_items() ignores all reviews after the first one that day. Do we need to worry about that when updating memory state during learning? It would be nice if it did not matter, as that would mean we'd have to start checking the revlog history during review.

L-M-Sherlock:
Ideally, FSRS only update the memory state in the first learning and reviews whose elapsed days are greater or equal to 1 day. That's why I don't recommend using interday learning steps. In the custom scheduling of FSRS. For more details, you can check the script:

https://github.com/open-spaced-repetition/fsrs4anki/blob/5a34201f4d5b3a9cc1a5da4511b6bc32b7c6e909/fsrs4anki_scheduler.js#L108-L110

https://github.com/open-spaced-repetition/fsrs4anki/blob/5a34201f4d5b3a9cc1a5da4511b6bc32b7c6e909/fsrs4anki_scheduler.js#L145-L152

dae:
Could we solve this by assigning the initial memory state when cards graduate learning, instead of on their first learning step?

L-M-Sherlock:
I think it would be nice. Even if the user use interday learning steps, we could pass the revlog history for calculating the memory state when the card graduate learning.

L-M-Sherlock:
image

I press `easy` for this new card, but its stability is still 0.

dae:
I'm a bit confused as to why you're getting a 0 stability - I haven't yet implemented memory assignment on graduation, so fsrsMemoryState should be None until it gets reviewed again. Do you still have custom scheduling active? Does it happen when add-ons are disabled?

L-M-Sherlock:
I disable the helper add-on and now it is normal.

dae:
Might be a bug with saving the card in Python. Will look into it.

dae:
I've just used the latest review for now, but we could potentially gather the revlog history for learning cards when fsrs is enabled, and use memory_state() instead in the future.

dae:
Ok, you can now do BURN_LOG=$(pwd)/extra ./run to get the training log and checkpoints

dae:
I've been thinking a bit more about gathering the revlog history for learning cards.

It would mean that the starting S/D match what the user gets when they enable FSRS, which is good - having cards start at a different difficulty when reviewed normally vs FSRS enabled / weights changed is a bit confusing.

On the other hand, it's somewhat inconsistent with the standard Anki scheduler. Learning steps were first introduced so that users could fail a card repeatedly until they first memorize it, without having later reviews penalized by those early failures. I know FSRS doesn't have the same issues with low ease levels that SM2 does, but I'm curious - have you explored what sort of difference it makes to FSRS's performance when all learning steps except the last one (where rating must be 3 or 4 in order to graduate) are excluded from training? If excluding the other learning steps makes a non-trivial regression in performance, then that's a strong argument for fetching them during learning.

L-M-Sherlock:
Do you mean to pick the rating of the last review during the learning stage as the first rating for initializing the memory state? I haven't tested it. FSRS uses the first rating for the new cards to initialize the memory staste because in my first research accepted by KDD, the first rating has a huge impact to the initial stability. Of course, the research is based on a language learning app, so the pattern would be different in other subjects. But if we only consider the last rating in the learning stage, the initial stability will only has two options: one for good, and one for easy.

According to the data collected from various Anki users, The initial stability has strong correlation to the first rating.

dae:
Ok, sounds like we're best off gathering the revlogs. Thanks :-)

dae:
What do you think of the new approach? For the minimum viable implementation, can you think of anything that still must be implemented before we merge this in?

L-M-Sherlock:
Nice. I have tested in my local device. It's the time to merge.

L-M-Sherlock (approved):

PR #2656, opened 2023-09-13
fix single_card_revlog_to_items

L-M-Sherlock:
```rust
fn keep_first_revlog_same_date(
    entries: Vec<RevlogEntry>,
    next_day_starts_at: i64,
    timezone: Tz,
) -> Vec<RevlogEntry> {
    let mut entries = entries.clone();
    let mut unique_dates = std::collections::HashSet::new();
    entries.retain(|entry| {
        let date = convert_to_date(entry.id, next_day_starts_at, timezone);
        unique_dates.insert(date)
    });
    entries
}
```
https://github.com/open-spaced-repetition/fsrs-rs/blob/3389062ff6ae23e4543d8872c0db354a73d1e40a/src/convertor_tests.rs#L128-L140

If there are multiple reviews (no matter the review kind) done on one day, keep only the first review.

### DIFF
--- a/rslib/src/scheduler/fsrs/weights.rs
+++ b/rslib/src/scheduler/fsrs/weights.rs
@@ -358,7 +358,6 @@ mod tests {
         );
     }
 
-
     #[test]
     fn single_learning_step_skipped_when_training() {
         assert_eq!(
--- a/rslib/src/scheduler/fsrs/weights.rs
+++ b/rslib/src/scheduler/fsrs/weights.rs
@@ -135,12 +135,14 @@ fn single_card_revlog_to_items(
     next_day_at: TimestampSecs,
     training: bool,
 ) -> Option<Vec<FSRSItem>> {
-    let last_learn_entry = entries
-        .iter()
-        .enumerate()
-        .rev()
-        .find(|(_idx, e)| e.review_kind == RevlogReviewKind::Learning)
-        .map(|(idx, _)| idx);
+    let mut last_learn_entry = None;
+    for (index, entry) in entries.iter().enumerate().rev() {
+        if entry.review_kind == RevlogReviewKind::Learning {
+            last_learn_entry = Some(index);
+        } else if last_learn_entry != None {
+            break;
+        }
+    }
     let first_relearn = entries
         .iter()
         .enumerate()
@@ -149,7 +151,7 @@ fn single_card_revlog_to_items(
     if let Some(idx) = last_learn_entry.or(first_relearn) {
         // start from the (re)learning step
         if idx > 0 {
-            entries.drain(..idx - 1);
+            entries.drain(..idx);
         }
     } else {
         // we ignore cards that don't have any learning steps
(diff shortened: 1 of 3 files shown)


L-M-Sherlock:
The convertor still has some problems. I'm fixing them.

> This breaks the (Learn, Learn) case you presented earlier.

Where did I present it?

L-M-Sherlock:
I have another question:

Does review in advance generate a revlog with `RevlogReviewKind::Filtered`?

dae:
https://github.com/ankitects/anki/pull/2654#issuecomment-1715222532

Re your other question, in v3, the behaviour changed:

    /// Old Anki versions called this "Cram" or "Early", and assigned it when
    /// reviewing cards ahead. It is now only used for filtered decks with
    /// rescheduling disabled.
    Filtered = 3,

L-M-Sherlock:
> Re your other question, in v3, the behaviour changed:

OK. `Early` shouldn't be filtered out. I need to keep them.

L-M-Sherlock:
```rust
        .iter()
        .enumerate()
        .rev()
        .map(|(idx, _)| idx);
```

`find()` is short-circuiting. It will return 2 for (learn, learn, learn). But the correct output is 0. So I replace it with:

```rust

    let mut last_learn_entry = None;
        if entry.review_kind == RevlogReviewKind::Learning {
            last_learn_entry = Some(index);
        } else if last_learn_entry != None {
            break;
        }
    }
```

rslib/src/scheduler/fsrs/weights.rs
```diff
@@ -161,16 +163,12 @@ fn single_card_revlog_to_items(
     entries.retain(|entry| {
         let manually_rescheduled =
             entry.review_kind == RevlogReviewKind::Manual || entry.button_chosen == 0;
-        let cram = entry.review_kind == RevlogReviewKind::Filtered;
+        let cram = entry.review_kind == RevlogReviewKind::Filtered && entry.ease_factor == 0;
```
L-M-Sherlock, line 166: @dae, I add an extra condition to keep `early` reviews, because it affected the schedule.

rslib/src/scheduler/fsrs/weights.rs
```diff
@@ -149,7 +151,7 @@ fn single_card_revlog_to_items(
     if let Some(idx) = last_learn_entry.or(first_relearn) {
         // start from the (re)learning step
         if idx > 0 {
-            entries.drain(..idx - 1);
+            entries.drain(..idx);
```
L-M-Sherlock, line 154: @dae, ..idx doesn't include idx, so it's unnecessary to subtract one.

dae:
Re short-circuiting, sorry, I misunderstood what the original code was doing. Thanks for making these changes Jarrett, I'll merge them in. Regarding the (learn, learn) case, if training=false then the first learning step won't be skipped, so I guess things are ok now :+1:

Issue #2628, opened 2023-08-29
Reproduce and investigate possible race condition in task/undo handling


dae:
https://forums.ankiweb.net/t/target-undo-op-not-found/33622/18

mjuhanne:
This may be caused when trying to do _col.merge_undo_entries()_ with 30 or more items. 

For example:
```
def add_notes_blocking(col, notes, deck_id, checkpoint):

    print("**** Adding %d notes to checkpoint %s" % (len(notes), checkpoint))
    pos = col.add_custom_undo_entry(checkpoint)

    for note in notes:
        col.add_note(note, deck_id)

    return col.merge_undo_entries(pos)

def add_notes_op(*, notes, deck_id, checkpoint) -> CollectionOp[OpChanges]:
    return CollectionOp(aqt.mw, 
        lambda col, notes=notes,deck_id=deck_id,checkpoint=checkpoint: 
            add_notes_blocking(col, notes, deck_id, checkpoint)
    )

```
If I try to feed this with 30 or more notes, "target undo op not found" error appears.

Then I try to circumvent this by splitting the notes into 29 note batches, each having its own checkpoint (ugly solution, I know):

```
def add_notes(notes, deck_id, checkpoint, success_func):

    maximum_notes_per_checkpoint = 29
    notes_per_checkpoint = [ notes[i:i + maximum_notes_per_checkpoint] 
        for i in range(0, len(notes), maximum_notes_per_checkpoint)
    ]

    checkpoint_number = 0
    for notes_in_checkpoint in notes_per_checkpoint:
        checkpoint_name = checkpoint + '_' + str(checkpoint_number+1)
        print("Creating checkpoint %s" % checkpoint_name)

        op = add_notes_op(notes=notes_in_checkpoint, deck_id=deck_id, checkpoint=checkpoint_name )
        if checkpoint_number == len(notes_per_checkpoint)-1:
            # run success_func only after last batch (checkpoint) is processed
            op.success(success_func)
        op.run_in_background()
        checkpoint_number += 1
```

This may work for 35-40 notes but when adding 300 notes with concurrent 5 background processes I get a bunch of 'target undo op not found' messages again.

Is there a way to do this elegantly and without errors?

dae:
[...]

PR #2659, opened 2023-09-15
Add a backend method to add notes in bulk

abdnh:
Using col.add_note() in a CollectionOp where custom undo entries are involved can easily trigger the "target undo op not found" error due to the undo queue limit. Currently, one has to resort to workarounds such as in the example in https://github.com/ankitects/anki/issues/2628#issuecomment-1703827442

This PR adds the col.add_notes() method to add a list of notes in bulk, registering a single undo entry.

### DIFF
--- a/proto/anki/notes.proto
+++ b/proto/anki/notes.proto
@@ -64,7 +64,7 @@ message AddNoteResponse {
 }
 
 message AddNotesRequest {
-  repeated AddNoteRequest note_and_deck_id = 1;
+  repeated AddNoteRequest requests = 1;
 }
 
 message AddNotesResponse {
--- a/proto/anki/notes.proto
+++ b/proto/anki/notes.proto
@@ -14,6 +14,7 @@ import "anki/cards.proto";
 
 service NotesService {
   rpc NewNote(notetypes.NotetypeId) returns (Note);
+  rpc AddNote(AddNoteRequest) returns (AddNoteResponse);
   rpc AddNotes(AddNotesRequest) returns (AddNotesResponse);
   rpc DefaultsForAdding(DefaultsForAddingRequest) returns (DeckAndNotetype);
   rpc DefaultDeckForNotetype(notetypes.NotetypeId) returns (decks.DeckId);
@@ -52,9 +53,18 @@ message Note {
   repeated string fields = 7;
 }
 
-message AddNotesRequest {
-  repeated Note notes = 1;
+message AddNoteRequest {
+  Note note = 1;
   int64 deck_id = 2;
+}
+
+message AddNoteResponse {
+  collection.OpChanges changes = 1;
+  int64 note_id = 2;
+}
+
+message AddNotesRequest {
+  repeated AddNoteRequest note_and_deck_id = 1;
 }
 
 message AddNotesResponse {
--- a/proto/anki/notes.proto
+++ b/proto/anki/notes.proto
@@ -14,7 +14,7 @@ import "anki/cards.proto";
 
 service NotesService {
   rpc NewNote(notetypes.NotetypeId) returns (Note);
-  rpc AddNote(AddNoteRequest) returns (AddNoteResponse);
+  rpc AddNotes(AddNotesRequest) returns (AddNotesResponse);
   rpc DefaultsForAdding(DefaultsForAddingRequest) returns (DeckAndNotetype);
   rpc DefaultDeckForNotetype(notetypes.NotetypeId) returns (decks.DeckId);
   rpc UpdateNotes(UpdateNotesRequest) returns (collection.OpChanges);
@@ -52,14 +52,14 @@ message Note {
   repeated string fields = 7;
 }
 
-message AddNoteRequest {
-  Note note = 1;
+message AddNotesRequest {
+  repeated Note notes = 1;
   int64 deck_id = 2;
 }
 
-message AddNoteResponse {
+message AddNotesResponse {
   collection.OpChanges changes = 1;
-  int64 note_id = 2;
+  repeated int64 nids = 2;
 }
 
 message UpdateNotesRequest {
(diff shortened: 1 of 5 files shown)


abdnh:
Hmm, isn't Pylint's `invalid-name` warning disabled? Any idea why it's being reported?

dae:
Not sure what's going on there, but renaming it to idx is probably the fastest solution :-)

dae:
How about using `repeated AddNoteRequest note_and_deck_id = 1` instead of passing a list of notes but a single deck id? That is more flexible, and should require fewer code changes.

rslib/src/notes/mod.rs
```diff
@@ -333,21 +337,52 @@ impl Collection {
...
     ) -> Result<()> {
-        self.canonify_note_tags(note, ctx.usn)?;
-        note.prepare_for_update(ctx.notetype, normalize_text)?;
-        note.set_modified(ctx.usn);
-        self.add_note_only_undoable(note)?;
-        self.generate_cards_for_new_note(ctx, note, did)?;
-        self.set_last_deck_for_notetype(note.notetype_id, did)?;
-        self.set_last_notetype_for_deck(did, note.notetype_id)?;
-        self.set_current_notetype_id(note.notetype_id)
+        let mut notetypes: HashMap<NotetypeId, Arc<Notetype>> = HashMap::new();
```
dae, line 347: This is unnecessary; notetypes are already cached by get_notetype()

rslib/src/notes/mod.rs
```diff
@@ -333,21 +337,52 @@ impl Collection {
...
+        for note in notes.iter() {
+            if !notetypes.contains_key(&note.notetype_id) {
+                let nt = self
+                    .get_notetype(note.notetype_id)?
+                    .or_invalid("missing note type")?;
+                notetypes.insert(note.notetype_id, nt);
+            }
+        }
+        let mut contexts: HashMap<NotetypeId, CardGenContext<&Notetype>> = HashMap::new();
```
dae, line 356: These are fairly cheap to create; I wouldn't bother caching them.

rslib/src/notes/mod.rs
```diff
@@ -333,21 +337,52 @@ impl Collection {
...
+            let ctx = contexts.get(&note.notetype_id).unwrap();
+            self.generate_cards_for_new_note(ctx, note, did)?;
+        }
+        for notetype_id in notetypes.into_keys() {
+            self.set_last_deck_for_notetype(notetype_id, did)?;
+            self.set_last_notetype_for_deck(did, notetype_id)?;
+        }
+        if let Some(last_note) = notes.last() {
+            self.set_current_notetype_id(last_note.notetype_id)?;
+        }
```
dae, line 383: These are quite cheap too. It's great that you're aiming to be efficient here, but in this particular case I think there probably wouldn't be a big difference between this code and just doing the equivalent of `for note_and_deck_id in note_and_deck_ids { old_code() }`. But happy to be proved wrong with a benchmark :-)

rslib/src/notes/mod.rs
```diff
@@ -333,21 +337,52 @@ impl Collection {
...
-        self.canonify_note_tags(note, ctx.usn)?;
-        note.prepare_for_update(ctx.notetype, normalize_text)?;
-        note.set_modified(ctx.usn);
-        self.add_note_only_undoable(note)?;
-        self.generate_cards_for_new_note(ctx, note, did)?;
-        self.set_last_deck_for_notetype(note.notetype_id, did)?;
-        self.set_last_notetype_for_deck(did, note.notetype_id)?;
-        self.set_current_notetype_id(note.notetype_id)
+        let mut notetypes: HashMap<NotetypeId, Arc<Notetype>> = HashMap::new();
+        for note in notes.iter() {
```
dae, line 348: You can do `in &notes` and `in &mut notes` instead of calling iter(mut)

rslib/src/notes/mod.rs
abdnh: Yes. I did a quick test and there's no big difference (~0.2s when adding 10K notes).

abdnh:
> How about using `repeated AddNoteRequest note_and_deck_id = 1` instead of passing a list of notes but a single deck id? That is more flexible, and should require fewer code changes.

Should this be exposed as `def add_notes(self, notes: Sequence[Note], dids: Sequence[DeckId]) -> OpChanges:`? I'd find this slightly inconvenient when I want to add all notes to the same deck (the most common case, I think). What about exposing a different `add_notes_to_decks` function that has this signature?

dae:
What about a public export of AddNoteRequest = AddNoteRequest at the top of collection.py, and changing the signature to add_notes(self, requests: Iterable[AddNoteRequest]). If you have a list of notes, it's just a case of `(AddNoteRequest(note=n, deck_id) for n in notes)`. If you wanted to make things even easier, you could accept [Note | AddNoteRequest] and allow passing in a deck_id as an optional second arg to be used whenever a bare note is provided.

abdnh:
> What about a public export of AddNoteRequest = AddNoteRequest at the top of collection.py, and changing the signature to add_notes(self, requests: Iterable[AddNoteRequest]). If you have a list of notes, it's just a case of (AddNoteRequest(note=n, deck_id) for n in notes). 

But the caller would need to call note._to_backend_note() in this case, right? Should it be marked as public?

dae:
Good point - in that case I'd recommend a dataclass that takes a native note, which we then convert in the function.

proto/anki/notes.proto
```diff
@@ -62,6 +63,15 @@ message AddNoteResponse {
   int64 note_id = 2;
 }
 
+message AddNotesRequest {
+  repeated AddNoteRequest note_and_deck_id = 1;
```
dae, line 67:
```suggestion
  repeated AddNoteRequest note_and_deck_ids = 1;
```

Protobuf convention

dae: or we could just call it requests

dae (commented):
Thanks Abdo - just one minor naming tweak, and it should be good to go.

PR #2660, opened 2023-09-15
Allow full screen mode on new Qt6 drivers

abdnh:


ftl/core/actions.ftl
```diff
@@ -55,7 +55,7 @@ actions-previous-card-info = Previous Card Info
 # By convention, the name of a menu action is suffixed with "..." if additional
 # input is required before it can be performed. E.g. "Export..." vs. "Delete".
 actions-with-ellipsis = { $action }...
-actions-currently-unsupported = This action is currently not supported on your system.
+actions-fullscreen-unsupported = Full screen mode is not supported for your video driver. Try switching to a different one from the preferences screen.
```
dae, line 58: It might be worth changing the key to -unsupported2, so that translators are notified that it needs re-translation. Otherwise looks good - thanks!

dae: Duh, how did I not notice you'd already changed it. Ignore me :-)

Issue #2556, opened 2023-06-23
Unable to save Fields dialog if all existing fields are deleted


dae:
https://forums.ankiweb.net/t/bug-unrecoverable-state-when-modifying-card-type/31440

dae:
> However I’m not sure what would be the desired solution.

Me either :-)

> A potential solution that I can think of is to throw an error when the user tries to delete the last remaining field which appears on the front of a card.

The updating of the template is deferred until saving time, so that would not be easy. Maybe update_templates_for_renamed_and_removed_fields() could check if the template has no field references after calling rename_and_remove_fields(), and add a reference to the first field if the template's currently empty? An added complication is that it would need to be a cloze reference for a cloze notetype.

swektr:
Finally decided to give this issue a shot.  I wanted to get more familiar with Rust first but then I forgot about this issue.  Anyways I've made a draft pull request at #2663

PR #2663, opened 2023-09-17
Fix unable to save field dialog if certain fields are deleted

swektr:
Implemented solution suggested in issue #2556
I decided to use the already available `ParsedTemplate::all_referenced_field_names` and `ParsedTemplate::all_referenced_cloze_field_names` to check for field referenceless and/or clozeless sides.  However, I wonder if it might be better to define and use something like the following to avoid the overhead of allocating a new HashSet: 
```rust
// UNTESTED
impl ParsedTemplate {
    pub(crate) fn contains_cloze_replacement(&self) -> bool {
        self.0.iter().any(|node| matches!(
            node, 
            ParsedNode::Replacement {key:_, filters} if filters.iter().any(|f| f=="cloze")
        ))
    }

    pub(crate) fn contains_field_replacement(&self) -> bool {
        self.0.iter().any(|node| matches!(node, ParsedNode::Replacement {key:_,filters:_}))
    }
}
```

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -139,6 +139,7 @@ Nil Admirari <https://github.com/nihil-admirari>
 Michael Winkworth <github.com/SteelColossus>
 Mateusz Wojewoda <kawa1.11@o2.pl>
 Jarrett Ye <jarrett.ye@outlook.com>
+Sam Waechter <github.com/swektr>
 
 ********************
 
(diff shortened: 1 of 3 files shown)


swektr:
I figured out why the build / check was failing.  I forgot to add cargo to my PATH and also forgot to run ./ninja format

dae:
Thanks for looking into this! If possible, it would be great if we had a unit test that confirms this is working.

I don't have a strong preference on HashSet vs the code above - updating notetypes is not a particularly hot path, but the code above is more elegant.

swektr:
Haven't gotten a chance to do play around with tests in Rust yet but I can give that a shot.  Where should I add the test? I noticed that notetype/template.rs has a test section at the bottom but notetype/mod.rs does not, I just add one there right?

dae:
Yep, please add one to the same file if there isn't one already.

swektr:
Added the tests and ended up using the new functions above.

dae:
Looks great - thank you for this!

PR #2670, opened 2023-09-20
Skip template checks in Fields screen

### LINKED ISSUE #2605: Need to handle old collections with colon in field names
https://forums.ankiweb.net/t/front-template-has-a-problem-found-audio-french-but-there-is-no-field-called-french/32790

While we have code to strip out the colons when saving changes already, saving fails because the template checks fail. Simplest solution may be to bypass the template checks when saving from the Fields screen, which means exposing skip_checks in the update_notetype_legacy method.

abdnh:
Closes #2605

(This probably makes #2663 redundant/less useful).


dae:
Thanks Abdo! I think #2663 still has some value, as that way the user doesn't need to figure out how to fix the problem themselves. But had this landed first, it might have been harder to justify the work on that. :-)

PR #2673, opened 2023-09-21
Add an option to stop the timer on answer

### LINKED ISSUE #2665: Add an option to stop the card timer when answer revealed
This would only affect the on-screen display - we still need the full time with question+answer when recording entries in the review log.

https://forums.ankiweb.net/t/stop-timer-after-showing-answer/29937

abdnh:
This adds a preference setting to stop the displayed timer when the card is answered.

Closes #2665


dae (commented):
Thanks Abdo! Just one minor change required: the tab order in the preferences is not correct

PR #2674, opened 2023-09-21
Add "Anki 2.1.50+ Collection Package" to exporting.ftl

hnvy:
Hello,

I couldn't find this `Anki 2.1.50+ Collection Package` string on https://i18n.ankiweb.net, so I hope I have taken the correct steps to rectify this. Not a significant PR, I know.

Thank you for your hard work, as always.

HR

### DIFF
--- a/ftl/core/exporting.ftl
+++ b/ftl/core/exporting.ftl
@@ -1,6 +1,7 @@
 exporting-all-decks = All Decks
 exporting-anki-20-deck = Anki 2.0 Deck
 exporting-anki-collection-package = Anki Collection Package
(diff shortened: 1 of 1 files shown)


dae:
This string is only used when the legacy export screen is enabled in the preferences or by an add-on, and the plan is to remove it in the future, so I'm not sure it's worth asking translators to translate it. The string in the new exporter is translatable ("Support older Anki versions")

hnvy:
> the plan is to remove it in the future, so I'm not sure it's worth asking translators to translate it

Thank you for your reply.

Understood, I'll close this PR then.



HR

PR #2675, opened 2023-09-23
Support Qt.TextFormat.MarkdownText in aqt.utils

abdnh:


dae:
Didn't know about that one!

PR #2683, opened 2023-09-25
Match IO extensions in a case-insensitive manner

abdnh:
I happened to test with an image with the `.JPG` extension. The image was showing when adding the note or reviewing, but not when editing.


dae:
Whoops - thanks!

PR #2684, opened 2023-09-25
More sensible FSRS progress indicator precision

meliache:
Its very helpful having a sub-permille precision in a progress indicator, percent-precision or at most a tenth of a percent should be sufficient for any indicator.

But in particular the compute-retention progress has 10 steps, i.e. the progress increases in 10%-intervals (10%, 20%, ...), it *cannot* have sub-decimal progress-percentages, see the [fsrs-rs `optimal_retention` function](https://github.com/open-spaced-repetition/fsrs-rs/blob/2d5b19b4941599b1fbc7a49dbbe89774f975e4ed/src/optimal_retention.rs#L365-L368). So there integer percents should be enough, everything else is misleading.

The compute-weights progress is currently (as of beta-2) not showing up at all. Maybe if the bug is fixed it can show sub-percent percentages, so for know I changed that to 0.1% precision. But I think integer percentages should be fine here as well, so upon request I can fix that.

**Before:**  

**After:**  
image

Also see my [comment on this issue in the beta forum thread](https://forums.ankiweb.net/t/anki-23-10-beta/34912/39).

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -140,6 +140,7 @@ Michael Winkworth <github.com/SteelColossus>
 Mateusz Wojewoda <kawa1.11@o2.pl>
 Jarrett Ye <jarrett.ye@outlook.com>
 Sam Waechter <github.com/swektr>
+Michael Eliachevitch <m.eliachevitch@posteo.de>
 
 ********************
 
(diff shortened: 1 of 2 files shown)


dae:
Looks good - thanks!

PR #2685, opened 2023-09-26
Button to set desired FSRS retention to optimal/calculated

meliache:
Temporarily save the calculated optimal retention and display it with a button that sets the desired retention above to this value.Don't show button until attention had been calculated. Disable button when optimal and desired attention are equal.

I find this nicer than the current alert-popup solution, as it avoids a popup and gives a choice to the user to accept the calculated retention or not, while also persisting the calculated retention on the screen for a bit.
| Before computation | After computation | After clicking once |
|--------|--------|--------|
| image  |  image | image |

#### TODO
- [X] What's still missing is that the `optimalRetention` variable is global and persists when I change presets. When changing presets the variable should reset to `undefined`, which would also makes the button disappear. Ideally it should also disappear when changing the FSRS parameters. So probably it should be made part of some deck options state and subscribe to some events. But with that I might need some help. Also I thought whether that variable should go into the deck options schema but tbh it's not something we want to persist between sessions, users should recalculate it.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -140,6 +140,7 @@ Michael Winkworth <github.com/SteelColossus>
 Mateusz Wojewoda <kawa1.11@o2.pl>
 Jarrett Ye <jarrett.ye@outlook.com>
 Sam Waechter <github.com/swektr>
+Michael Eliachevitch <m.eliachevitch@posteo.de>
 
 ********************
 
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -334,7 +334,7 @@ deck-config-get-params = Get Params
 deck-config-fsrs-on-all-clients =
     Please ensure all of your Anki clients are Anki(Mobile) 23.10+ or AnkiDroid 2.17+. FSRS will
     not work correctly if one of your clients is older.
-deck-config-your-optimal-retention = Your optimal retention is { $num }.
+deck-config-set-optimal-retention = Set optimal retention to { $num }.
 
 ## NO NEED TO TRANSLATE. This text is no longer used by Anki, and will be removed in the future.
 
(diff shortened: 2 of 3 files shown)


dae:
I'm a bit conflicted by this change - it is more convenient, but I do worry that users are going to shoot themselves in the foot by blindly selecting the suggested retention without understanding the implications. If they have set a low number of minutes per day for example, the optimizer is going to suggest 0.75, which may be demotivating when they forget a lot.

meliache:
Thanks for the fixes. I am kind of new to both typescript and also svelte so this is really appreciated. To be honest I viewed this PR as a  suggestion of how I imagined the user interface for accepting the calculated retention might look, but still not sure it's *the best* way.

E.g. the computed percentage could be displayed in a different widget (or as simple text) seperate from the set/accept button, to keep the button smaller. I just don't know which svelte element to use and for the button implementation copy-paste was enough.

Also I cold imagine some tighter spacial grouping with the desired retentions but not sure how to do that. And as I mentioed in the forums it would be nice to have some signalling when the min or max values were returned.

But in any case I think it's good to have a confirmation stepeafter calculation and also it to be as part of the UI and not a pop-up.

meliache:
> I'm a bit conflicted by this change - it is more convenient, but I do worry that users are going to shoot themselves in the foot by blindly selecting the suggested retention without understanding the implications. If they have set a low number of minutes per day for example, the optimizer is going to suggest 0.75, which may be demotivating when they forget a lot.

I didn't see this comment when I wrote myprevious reply. I was thinking this suggestion to be between the beta1, which just set the config without asking and the tooltipp in the beta2.

Instead of the button I would also be fine with just showing a copy-able text field with the simulation result, but I have personally a slight aversion of using alert popup for this. And in any case an in-app popup and documentation for how to interpret the settings and the result would be nice.

dae:
I will merge this in for now, and we can iterate later.

meliache:
Thanks, fine with me :)

meliache:
And btw, I'm really happy to be able to finally contribute to Anki. I think this is because the FSRS inclusion is exciting enough to try the betas and building from source, but the initial betas were unpolished enough that there were low-hanging fruit what to improve. Let's hope this is not my last contribution, now that I understand the codebase better and have a working build-process.

PR #2686, opened 2023-09-27
Implement "stop timer on answer" as a preset option

abdnh:
For context:
https://github.com/ankitects/anki/issues/2665
https://forums.ankiweb.net/t/anki-23-10-beta/34912/35


dae:
Thanks Abdo!

PR #2689, opened 2023-09-27
 Allow creating an image occlusion from the clipboard

### LINKED ISSUE #2664: Allow creating an image occlusion from the clipboard
Previously mentioned on #2631

abdnh:
Closes #2664

### DIFF
--- a/ftl/core/notetypes.ftl
+++ b/ftl/core/notetypes.ftl
@@ -54,3 +54,4 @@ notetype-error-no-image-to-show = No image to show.
 notetypes-no-occlusion-created = You must make at least one occlusion.
 notetypes-no-occlusion-created2 = Unable to add. Either you have not added any occlusions, or the first field is empty.
 notetypes-io-select-image = Select Image
+notetypes-io-select-image-from-clipboard = Select Image from clipboard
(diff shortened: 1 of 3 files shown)


ftl/core/notetypes.ftl
```diff
@@ -54,3 +54,4 @@ notetype-error-no-image-to-show = No image to show.
 notetypes-no-occlusion-created = You must make at least one occlusion.
 notetypes-no-occlusion-created2 = Unable to add. Either you have not added any occlusions, or the first field is empty.
 notetypes-io-select-image = Select Image
+notetypes-io-select-image-from-clipboard = Select Image from Clipboard
```
glutanimate, line 57: (minor) Would suggest changing this to `Paste Image from Clipboard` to signify that there is no selection screen.

glutanimate (commented):
Working well here, thanks Abdo! One general thought:

With these changes we'd have two different implementations of pasting images into the editor. IMHO it might make sense to unify them, even if that means a slightly larger diff footprint and more (though still conservative) refactoring.

dae:
LGTM - thanks as always Abdo!

PR #2698, opened 2023-10-02
fix normal function

krmanik:
I have checked the issue. The issues are in normalization function, every time a new point added in polygon the points normalized with previous value which decrease the value till it get undefined. I have implemented this fix. But other issue is that when n number of polygon added in canvas in edit mode, only n-1 number of polygon saved. I tried to find the issue but still working on it.

### DIFF
--- a/ts/image-occlusion/shapes/position.ts
+++ b/ts/image-occlusion/shapes/position.ts
@@ -5,11 +5,17 @@ import type { Size } from "../types";
 
 /** Position normalized to 0-1 range, e.g. 150px in a 600x300px canvas is 0.25 */
 export function xToNormalized(size: Size, x: number): number {
+    if (x >= 0 && x <= 1) {
+        return x;
+    }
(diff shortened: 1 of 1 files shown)


dae:
Hmm, I'm not sure this is the right approach. If a point happens to be on pixel 1, it's going to be normalized to 1, and moved all the way to the right/bottom, won't it?

dae:
Maybe makeNormal/makeAbsolute should be changed to asNormal()/asAbsolute(), and have them return the modified values, instead of mutating the object?

krmanik:
I guess I need help on this issue. Sorry.

dae:
No problem. You're welcome to this one @abdnh if it interests you; if not, I'll try to look into this ~mid next week.

abdnh:
I assume this was fixed in #2712?

dae:
I think so, yep - thanks!

PR #2699, opened 2023-10-02
Describe meaning of FSRS stability statistic as 90% drop in retrievability

meliache:
The previous description that the stability can be combined with desired retention to calculate the intervals is very vague and gives no intuition what a given stability means. But the fact of it being a decay constant for a retrievability drop from 100% to 90% is in my opinion easy to understand on an intuitive level. Then you know that it should look very similar to the interval statistic if you choose a retention of 90% (only difference due to fuzz, siblings etc), if your desired retention is 80% then you can expect the intervals to be larger than stability, and if your desired retention is 95% then the intervals will be smaller than the stability.

In my commit I found it important to say that it's the interval for retrievability drop _from 100% to 90%_ instead of shortening it to just "retrievatility drop to 90%", because the latter could be misunderstood as the time it will take from the current moment in time (and current retrievatility) to drop to 90%, but as far as I understand it's a decay constant (like a half time) and thus always refers to the 100%&ndash;90% interval.

The meaning of "retrievability" itself is not obvious but explained in the dedicated statistic below.

Note that I'm not an expert of FSRS but what I wrote is my understanding from reading descriptions online. 

Only downside with my suggested description is that it's a bit long for a subtitle and creates a line wrap on my screen. I assume a more in-depth description of the FSRS parameters will eventually happen in the manual and in-app settings tooltips, so the statistics tooltips don't need to explain everything. But if the subtitle is too long I would rather remove the second sentence, i.e. the original subtitle.


dae:
@L-M-Sherlock could you confirm if that's correct? When I look at a card with a 1 year stability that is due today, the current retrievability appears to be whatever desired retention is set to at the time.

L-M-Sherlock:
> When I look at a card with a 1 year stability that is due today, the current retrievability appears to be whatever desired retention is set to at the time.

Yeah. Ideally, its retrievability is equal to your desired retention. If your desired retention is 90%, the interval of this card should be 1 year.

dae:
So to confirm, the proposed change here is incorrect?

L-M-Sherlock:
The desired retention doesn't affect the stability. If the DR of user is 80%, for a card with 1 year stability, the interval will be about 2 years.

L-M-Sherlock:
So the committed change is correct.

L-M-Sherlock:
So the committed change is correct.

dae:
Ok, thanks for clarifying. 

@meliache I would like to shorten the subtitle - how about we cut out the second sentence? For the first sentence, what do you think about this instead?

"Predicted delay that you have a 90% chance of remembering."

dae:
As I'm about to bake a new beta, I'll tentatively go ahead with that wording, but if you disagree, happy to discuss/change it.

meliache:
I'm fine with the change in e5257045710cd23583120e99bad49af4079d999b, thanks 👍

PR #2703, opened 2023-10-03
Update rust-csv

martinetd:
Upstream rust-csv 1.3.0 includes our patch and can be used instead of the ankitects repo.

This fixes `cargo vendor`

EDIT: as said in other issue feel free to drop this and commit yourself if that's less work, I don't care about authorship


dae:
Looks good - thanks!

PR #2705, opened 2023-10-04
Add text tool to image occlusion

### LINKED ISSUE #2651: Add text tool to image occlusion?
https://forums.ankiweb.net/t/feature-request-editing-in-image-occlusion/34052

http://fabricjs.com/fabric-intro-part-2#text

abdnh:
This adds an image occlusion tool to insert text. (See #2651)

## Issues

This is a draft because there are still some issues to be tackled:
- [x] Text objects are always grouped.
- [x] Text objects are hidden on the back side, just like masks.
- [x] Resizing doesn't work properly.
- [ ] Because text objects are not masks (at least, that's how I understand this feature will be used), they need to be stored somehow independently of clozes. I used a hack here where text objects are stored as the "zeroth cloze", which is always active. E.g. `{{c0::image-occlusion:text:left=.3208:top=.2655:fill=rgb(0,0,0):oi=1:text=some text}}`. I'm sure someone has a more elegant solution.

@glutanimate @krmanik tagging the IO guys for feedback.
@krmanik let us know if your implementation is more mature and I'll close this PR.

Closes #2651


ethanwater (commented):
not a contributor, was just browsing and this looks like a nice draft. the use of zeroth is pretty cool even if it's just for a draft.

abdnh:
This should be ready for review. One additional note though:

- [ ] The black text color can be problematic for some images. Should we make it customizable via a CSS variable?

dae:
Thanks Abdo, I'll take a look (currently visiting family, so it may take a few days). Regarding the text, CSS customizations might be awkward as users won't want to have to change I/O notetype depending on the color of their images. Maybe the text tool should automatically include a rectangle behind the text in a color that makes the text readable, or show the text with an inverse outline?

ts/image-occlusion/shapes/from-cloze.ts
```diff
@@ -9,11 +9,14 @@ import type { Shape, ShapeOrShapes } from "./base";
...
 import { Rectangle } from "./rectangle";
+import { Text } from "./text";
 
 /** Given a cloze field with text like the following, extract the shapes from it:
  * {{c1::image-occlusion:rect:left=10.0:top=20:width=30:height=10:fill=#ffe34d}}
  */
-export function extractShapesFromClozedField(clozeStr: string): ShapeOrShapes[] {
+export function extractShapesFromClozedField(
+    clozeStr: string,
+): ShapeOrShapes[] {
```
dae, line 19: Not a big deal, but noting for the future: these changes in formatting make the review a little more difficult - I'm guessing you may have formatted with prettier? You should be able to configure VSCode to use dprint instead once the dprint extension (in .vscode.dist's recommendations) is installed.

ts/image-occlusion/shapes/from-cloze.ts
```diff
@@ -34,7 +37,14 @@ export function extractShapesFromClozedField(clozeStr: string): ShapeOrShapes[]
     const output: ShapeOrShapes[] = [];
 
     for (const index in clozeList) {
-        if (clozeList[index].length > 1) {
+        if (index === "c0") {
+            clozeList[index].forEach((cloze) => {
+                let shape: Shape | null = null;
+                if ((shape = extractShapeFromClozeText(cloze))) {
```
dae, line 43: I wish TypeScript provided a more ergonomic way of doing this!

ts/image-occlusion/shapes/from-cloze.ts
```diff
@@ -89,7 +106,12 @@ export function extractShapesFromRenderedClozes(selector: string): Shape[] {
 
 function extractShapeFromRenderedCloze(cloze: HTMLDivElement): Shape | null {
     const type = cloze.dataset.shape!;
-    if (type !== "rect" && type !== "ellipse" && type !== "polygon") {
+    if (
+        type !== "rect"
+        && type !== "ellipse"
+        && type !== "polygon"
+        && type !== "text"
+    ) {
```
dae, line 114: Might be worth factoring this out into a separate isValidType(), which could be a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)

ts/image-occlusion/shapes/from-cloze.ts
```diff
@@ -101,18 +123,29 @@ function extractShapeFromRenderedCloze(cloze: HTMLDivElement): Shape | null {
         rx: cloze.dataset.rx,
         ry: cloze.dataset.ry,
         points: cloze.dataset.points,
+        text: cloze.dataset.text,
+        sx: cloze.dataset.sx,
+        sy: cloze.dataset.sy,
```
dae, line 128: If we locked the aspect ratio, we'd only need to store one dimension, too.

dae:
First impressions - works quite well! Couple of thoughts:

- Using c0 was clever.
- Perhaps it would make sense to lock the aspect ratio when resizing text objects, so the text doesn't get distorted?

dae:
Code was nice and neat!

ts/image-occlusion/shapes/from-cloze.ts
abdnh: Yes, I have prettier set as the default. I'll change that. I made sure to run `ninja format` before committing, but apparently running prettier before dprint sometimes results in dprint behaving differently.

dae: Some formatters try to preserve manual formatting, so they'll e.g. wrap long lines, but won't unwrap short lines that could be on a single line. Personally I think that's a bad idea as it leads to problems like this, but it's not something we can change unfortunately :-) IIRC, Apple's Swift formatter behaves like this too.

abdnh:
> Perhaps it would make sense to lock the aspect ratio when resizing text objects, so the text doesn't get distorted?

This is implemented now. Fabric used to have a per-object `lockUniScaling` property for that, which would have made things simpler, but they removed it: https://github.com/fabricjs/fabric.js/issues/7378

I used the workaround suggested in that issue. (after hitting some walls with different approaches)

glutanimate:
Nice work, Abdo, thanks! A few observations and suggestions from my side as well:

- The text doesn't seem to respond to viewport size changes at the moment, or perhaps not as expected, which can lead to labels not being covered:

    https://github.com/ankitects/anki/assets/5459332/d96228fc-d6a9-42f7-a111-5e5dc5db7ecb


- Drawing a mask over a text label will oftentimes cause both to be shifted to the canvas origin:

     https://github.com/ankitects/anki/assets/5459332/fd7cf97a-396c-4b9e-b29a-8bbd2e5952b1

     Generated field content (note the negative coords):
     
     ```
     {{c0::image-occlusion:text:left=-0.0948:top=-0.0641:fill=rgb(0,0,0):oi=1:text=Test:scale=1.}}
     {{c2::image-occlusion:rect:left=-0.2298:top=-0.1435:oi=1:width=.4576:height=.2848}}
     ```

- The aspect ratio lock works nicely when adding notes, but does not seem to be active when editing:


    https://github.com/ankitects/anki/assets/5459332/74c6c9d2-034e-4f39-91e2-700f0282ddd4




- Using `c0` for text is clever! However, it has some minor ramifications which we'll want to take into account/address:

     + It looks like the `c0` markers count towards the current cloze index, meaning that cards that contain text will skip a cloze index (e.g. c0 → c2 → c3). Nothing major, but probably good to address.
     + If the text contains a colon, it will cause the label to be truncated. Other edge cases with cloze/mustache syntax might also exist, so I would recommend that we look into escaping some characters.

- I would recommend that we change the font family on the text to `arial`, both to be in sync with Anki's defaults and due to serifed fonts rendering fairly blurrily in the canvas/fabric context.

abdnh:
Thanks for your valuable feedback @glutanimate!
I pushed fixes for some issues and will look into the rest soon.

> Drawing a mask over a text label will oftentimes cause both to be shifted to the canvas origin:

This also occurs if you place a mask over an existing mask, so it's probably not related to this PR.

> The text doesn't seem to respond to viewport size changes at the moment, or perhaps not as expected, which can lead to labels not being covered:

Should be most fixed now. I guess we also need to set the font size dynamically too, but I've not looked into that yet.

dae:
Thanks for your work on this Abdo. I ran into a couple of issues when I tested it out:

- I can't seem to move the text to the right of the screen, but can move it to the left:

https://github.com/ankitects/anki/assets/180542/7ebe2a20-8e3f-4ac5-8bbc-b44d4e6c9f2c

It seems to be triggered by entering in some longer text, then shrinking it down - maybe the code that determines if the shape is outside the boundary is not handling the scale factor?

- Also notice how the text appears to be a bit fuzzy, and difficult to read against the darker background? I presume you picked an outline over a backing rectangle because it was simpler to implement, but perhaps it would be worth reconsidering this - I think it's going to lead to more legible text that way, especially at smaller sizes. And to do that, we're presumably going to need the rectangle coordinates inside the cloze, which will be hard to change later.

- Regarding escaping of :, is that something our cloze parsing logic actually supports? If not, one hacky but easier way to fix this would be to replace : with a similiar unicode char instead, when saving out the occlusions.

abdnh:
> Regarding escaping of :, is that something our cloze parsing logic actually supports?

Yes, that works without problems there.

> If not, one hacky but easier way to fix this would be to replace : with a similiar unicode char instead, when saving out the occlusions.

a84fa00f63f36922963e79fa77e13c86dfa1e116 handles escaping colons with a nom parser. Maybe not worth it if only for escaping colons, but I thought might prove necessary in the future if more properties are added, and removing the duplicate parsing logic from the TS side is a nice bonus.

> I presume you picked an outline over a backing rectangle because it was simpler to implement, but perhaps it would be worth reconsidering this - I think it's going to lead to more legible text that way, especially at smaller sizes. And to do that, we're presumably going to need the rectangle coordinates inside the cloze, which will be hard to change later.

A background color is drawn now, with no additional properties needed.

dae:
Much nicer - thanks Abdo!

glutanimate:
Thanks again for your work on this and all the fixes and improvements @abdnh!

One remaining suggestion: Would recommend that we change the background color to a non mask color. Created a follow-up for this in #2726.

PR #2709, opened 2023-10-06
Invalid sorting

### LINKED ISSUE #2701: When user selects a field that can't be sorted on, they get stuck on it
https://forums.ankiweb.net/t/anki-23-10-beta/34912/104

When they attempt to select a different field, it tries to refresh the current search (via call_after_note_saved) and fails, and the config remains set to the invalid value. Maybe we should a) have a new protobuf method to set the desired sort order, that can error out if it's invalid, and b) in the case of an invalid config, fall back to the default sort order instead of throwing an error?

@RumovZ if you think you'll have time to tackle this in the next week please do (asking you first since this is code you wrote); if you're busy, please let us know and I'll see if @abdnh would like to tackle it.

In the short term, users can work around it with

`mw.col.set_config("noteSortType", "noteFld")`

RumovZ:
Fixes #2701. Regarding my comment there, it turns out that 3 is already done by the collection method `_build_sort_mode()`.


.vscode.dist/launch.json
```diff
@@ -0,0 +1,24 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Run",
```
dae, line 8: Since these are Windows-specific, maybe they should be called "Run (Windows)" and "Build (Windows)"?

dae (commented):
Thanks Rumo, just a minor nitpick.

.vscode.dist/launch.json
RumovZ: Makes sense. I tried to make it work on other OSes now, but didn't test it. I hope it at least provides a good starting point for developers on these platforms.

.vscode.dist/launch.json
```diff
@@ -8,10 +8,13 @@
...
             "type": "python",
             "request": "launch",
-            "python": "${workspaceFolder}/out/pyenv/scripts/python.exe",
             "program": "tools/run.py",
             "console": "integratedTerminal",
             "cwd": "${workspaceFolder}",
+            "python": "${workspaceFolder}/out/pyenv/bin/python",
+            "windows": {
+                "python": "${workspaceFolder}/out/pyenv/scripts/python.exe"
+            },
```
dae, line 17: Ah, that's a nicer approach. Thanks!

PR #2624, opened 2023-08-24
[PoC] Simplify I/O mode

hikaru-y:

PR #2711, opened 2023-10-08
Propagate editor UI state transitions to add-ons

glutanimate:
Signal changes to the editor interface state via a Python hook, allowing add-on subscribers to adjust their UI to state transitions.

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -90,6 +90,18 @@ class EditorMode(Enum):
     BROWSER = 2
 
 
+class EditorState(Enum):
+    """
+    Current input state of the editing UI.
+    """
+
+    INITIAL = -1
+    FIELDS = 0
+    IO_PICKER = 1
+    IO_MASKS = 2
+    IO_FIELDS = 3
+
+
 class Editor:
     """The screen that embeds an editing widget should listen for changes via
     the `operation_did_execute` hook, and call set_note() when the editor needs
@@ -124,6 +136,7 @@ class Editor:
         self.last_field_index: int | None = None
         # current card, for card layout
         self.card: Card | None = None
+        self.state: EditorState | None = None
         self._init_links()
         self.setupOuter()
         self.add_webview()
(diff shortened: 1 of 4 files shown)


ts/editor/NoteEditor.svelte
```diff
@@ -526,6 +565,13 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
             ...oldEditorAdapter,
         });
 
+        editorState = getEditorState(
```
glutanimate, line 568: Not sure if the repetition here can be avoided somehow?

ts/editor/NoteEditor.svelte
```diff
@@ -485,6 +486,44 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         return false;
     }
 
+    // Signal editor UI state changes to add-ons
+
+    let editorState: EditorState = EditorState.Initial;
+    let lastEditorState: EditorState = editorState;
+
+    function getEditorState(
+        ioMaskEditorVisible: boolean,
```
glutanimate, line 495: Essentially following the current render logic here. We'll probably be able to simplify this in the future (#2624)

ts/editor/NoteEditor.svelte
```diff
@@ -485,6 +486,44 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+        } else if (!ioMaskEditorVisible && isImageOcclusion) {
+            return EditorState.ImageOcclusionFields;
+        }
+        return EditorState.Fields;
+    }
+
+    function signalEditorState(newState: EditorState) {
+        tick().then(() => {
+            globalThis.editorState = newState;
+            bridgeCommand(`editorState:${newState}:${lastEditorState}`);
```
glutanimate, line 513: In the future we might want to add a proper API/hook for JS subscribers here.  Did not pursue this further due to lack of time and as this solution works well for IOE for now.

qt/aqt/editor.py
```diff
@@ -124,6 +136,7 @@ def __init__(
         self.last_field_index: int | None = None
         # current card, for card layout
         self.card: Card | None = None
+        self.state: EditorState | None = None
```
dae, line 139: Do we need both None and INITIAL states here?

dae (commented):
Looks fine, just a quick question.

qt/aqt/editor.py
glutanimate: No, you're right, don't see much use in starting with `None`. Changed!

PR #2712, opened 2023-10-08
Fix polygon tool not working

### LINKED ISSUE #2679: Creating a second polygon clears out the previous one
https://forums.ankiweb.net/t/anki-23-10-beta-1/34912/15

hikaru-y:
Fixes #2679

Also,
- Fixes an issue where polygon objects were not converted correctly to cloze strings
  - Currently, the polygon tool is not working at all because the `points` property of all polygon objects is converted to `points=.0000`, not only when editing but also when adding.
- Fixes an issue where added polygons are sometimes not saved to the DB when switching notes or closing the editor


dae:
Thank you very much for these Hikaru!

Issue #2717, opened 2023-10-09
Show profile name when deleting profiles


soliviantar:
From: https://forums.ankiweb.net/t/show-profile-name-when-deleting-profiles/35483

When a user goes to File > Switch profile… they can delete a profile, but the pop-up window only says:

`All cards, notes, and media for this profile will be deleted. Are you sure?`

and it might be blocking the profile selection that will be deleted (see attached image, the profile that will be deleted is not shown anywhere). To prevent mistakes, I think it would be useful to change the message to:

`All cards, notes, and media for the profile [NAME OF PROFILE] will be deleted. Are you sure?`

![delete profiles](https://github.com/ankitects/anki/assets/25809253/e1494d1d-82ae-4187-9e06-8b77d3147602)

adeleke5140:
This looks quite fun to tackle, @dae would it be possible to add the hacktoberfest label to i?

chandraiyengar:
Hey, this is my first issue so i might have missed some things. I've made that PR^, let me know if there's anything to change/add

dae:
Closed by #2722

PR #2722, opened 2023-10-10
Add profile name in profile deletion confirmation message

chandraiyengar:
For #2717 This PR will add the profile name in the english confirmation message when deleting a profile

### DIFF
--- a/ftl/qt/qt-misc.ftl
+++ b/ftl/qt/qt-misc.ftl
@@ -1,6 +1,6 @@
 qt-misc-addon-will-be-installed-when-a = Add-on will be installed when a profile is opened.
 qt-misc-addons = Add-ons
-qt-misc-all-cards-notes-and-media-for = All cards, notes, and media for this profile will be deleted. Are you sure?
+qt-misc-all-cards-notes-and-media-for = All cards, notes, and media for the profile "{ $name }" will be deleted. Are you sure?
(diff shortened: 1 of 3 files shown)


ftl/qt/qt-misc.ftl
```diff
@@ -1,6 +1,6 @@
 qt-misc-addon-will-be-installed-when-a = Add-on will be installed when a profile is opened.
 qt-misc-addons = Add-ons
-qt-misc-all-cards-notes-and-media-for = All cards, notes, and media for this profile will be deleted. Are you sure?
+qt-misc-all-cards-notes-and-media-for = All cards, notes, and media for the profile "{ $name }" will be deleted. Are you sure?
```
dae, line 3: Please leave the old translation and add a new one, so that translators get notified that an update is required.

```suggestion
qt-misc-all-cards-notes-and-media-for2 = All cards, notes, and media for the profile "{ $name }" will be deleted. Are you sure?
```

dae (commented):
Thank you for contributing!

ftl/qt/qt-misc.ftl
abdnh: You don't have to worry about other languages. They are managed in a [separate repo](https://github.com/ankitects/anki-desktop-ftl) that's synced here regularly.

dae:
Thank you both!

PR #2726, opened 2023-10-12
Change IO text background color to white

glutanimate:
I would recommend that we use a neutral background color for labels. Using the mask color could be confusing and make occlusions more difficult to parse at a glance.


dae:
Thanks!

PR #2727, opened 2023-10-12
Remove v1/v2 support from the backend

abdnh:


rslib/src/card/mod.rs
```diff
@@ -184,8 +183,8 @@ impl Card {
     }
 
     /// Caller must ensure provided deck exists and is not filtered.
-    fn set_deck(&mut self, deck: DeckId, sched: SchedulerVersion) {
-        self.remove_from_filtered_deck_restoring_queue(sched);
+    fn set_deck(&mut self, deck: DeckId) {
+        self.remove_from_filtered_deck_restoring_queue();
```
dae, line 187: pub fn set_deck(&mut self, cards: &[CardId], deck_id: DeckId) -> Result<OpOutput<usize>> {

should check sched version and return an error

rslib/src/import_export/package/apkg/import/cards.rs
```diff
@@ -136,7 +132,7 @@ impl CardContext<'_> {
         self.remap_template_index(card);
         card.shift_collection_relative_dates(self.collection_delta);
         if !keep_filtered {
-            card.maybe_remove_from_filtered_deck(self.scheduler_version);
+            card.maybe_remove_from_filtered_deck();
```
dae, line 135: import_cards_and_revlog() should check sched version / return error

pylib/anki/collection.py
```diff
@@ -1453,13 +1381,7 @@ def flush(self) -> None:
...
 _Collection = Collection
 
-
-@dataclass
-class _ReviewsUndo:
-    entries: list[LegacyReviewUndo] = field(default_factory=list)
-
-
-_UndoInfo = Union[_ReviewsUndo, LegacyCheckpoint, None]
+_UndoInfo = Union[LegacyCheckpoint, None]
```
dae, line 1384: Not required for this PR, but I'm wondering whether it might be worth considering dropping support for checkpoints as well in a follow-up. They can cause the user to lose a few minutes of work in a crash, and once they're removed, we can also deprecated the non-undo-supporting methods like card.flush(). WDYT?

dae (commented):
Thanks Abdo, looks good - just a couple of extra places where we should probably be returning an error.

dae:
Also, it looks like there's still some state.v3Scheduler references in the deck options that we could clean up if we restricted that page to the v3 scheduler.

pylib/anki/collection.py
abdnh: I think it's reasonable to do that now a lot of legacy code is being removed. I'll look into that in a follow-up PR.

qt/aqt/deckoptions.py
```diff
@@ -106,7 +99,7 @@ def display_options_for_deck_id(deck_id: DeckId) -> None:
 
 def display_options_for_deck(deck: DeckDict) -> None:
     if not deck["dyn"]:
-        if KeyboardModifiersPressed().shift or aqt.mw.col.sched_ver() == 1:
+        if not aqt.mw.col.v3_scheduler():
```
dae: As some add-ons only expose their functionality in the old deck options screen, we'll need to keep the shift check.

Issue #2724, opened 2023-10-11
"Preview" window size gets reset every time (starting from 2.1.66)


CrAZONE0:
Starting from Anki 2.1.66 every time you open the "Preview" window the size will get reset. Addons have nothing to do with the issue. Happens on the latest beta as well (23.10beta4)

Platform: Windows 11

dae:
Original report: https://forums.ankiweb.net/t/anki-doesnt-remember-the-window-size-of-preview/35565

@abdnh can you reproduce this on Windows?

PR #2728, opened 2023-10-12
Fix previewer geometry not being restored sometimes

abdnh:
I was able to reproduce #2724 inconsistently. Maximizing/minimizing the window also fixed it for me. I found that moving the restoreGeom() call down fixes it as far as I can tell. I have no idea why this works.


dae:
Me either! :man_shrugging:

PR #2682, opened 2023-09-25
Fix wrong mask position after merging groups

abdnh:
Closes #2644

Ref: https://github.com/fabricjs/fabric.js/issues/4206

PR #2729, opened 2023-10-12
Fix selected shapes shifting to canvas origin

### LINKED ISSUE #2725: Overlaying one shape over another causes both to be shifted to the canvas origin
Split off from https://github.com/ankitects/anki/pull/2705#issuecomment-1752147723:

> - Drawing a mask over a text label will oftentimes cause both to be shifted to the canvas origin:
> 
>      https://github.com/ankitects/anki/assets/5459332/fd7cf97a-396c-4b9e-b29a-8bbd2e5952b1
> 
>      Generated field content (note the negative coords):
>      
>      ```
>      {{c0::image-occlusion:text:left=-0.0948:top=-0.0641:fill=rgb(0,0,0):oi=1:text=Test:scale=1.}}
>      {{c2::image-occlusion:rect:left=-0.2298:top=-0.1435:oi=1:width=.4576:height=.2848}}
>      ```

In practice users will likely most often run into this when adding text, but this seems to affect all shapes and could e.g. also be triggered by two closely adjacent shapes overlaying each other slightly (and inadvertently).

hikaru-y:
Fixes #2725

The problem in #2725 is caused by the fact that when newly added shape overlaps another, they become selected. In fact, in addition to the #2725 case, the problem of shifting to the canvas origin also occurs when the user clicks on the **Add** button with multiple non-overlapping shapes selected with the mouse.

Multiple objects in the `activeSelection` seem to have coordinates relative to the selected objects, just as objects in a `fabric.Group` have coordinates relative to the `Group`. I suppose we could recalculate the coordinates in the same way as #2682, but I took this approach because I thought it would be easier to simply deselect them before converting them to cloze text, unlike in the case of `Group`.

---

96f5befe58ac3a025f5c3fa5b8d6f8abb78358a8 is not required to fix #2725, but I believe the preferred behavior is for only the newly added shape to be in the selected state.

### DIFF
--- a/ts/image-occlusion/shapes/to-cloze.ts
+++ b/ts/image-occlusion/shapes/to-cloze.ts
@@ -36,6 +36,10 @@ export function exportShapesToClozeDeletions(occludeInactive: boolean): {
  */
 function baseShapesFromFabric(occludeInactive: boolean): ShapeOrShapes[] {
     const canvas = globalThis.canvas as Canvas;
+
+    // Prevents multiple shapes in 'activeSelection' from shifting to the canvas origin
+    canvas.discardActiveObject();
+
     makeMaskTransparent(canvas, false);
     const objects = canvas.getObjects() as FabricObject[];
     return objects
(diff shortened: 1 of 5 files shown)


dae:
Hikaru swoops in with more I/O fixes! Much appreciated. :-)

PR #2732, opened 2023-10-13
Turn edit window into a QMainWindow

### LINKED ISSUE #2672: Cmd+, preferences shortcut conflicts with bullet list shortcut
https://forums.ankiweb.net/t/shortcut-for-bullets-in-the-edit-window-opens-preferences-instead/34522

One way we could solve this is to turn the edit window into a main window like add/browse, so it gets its own menu without preferences listed there.

abdnh:
Closes #2672

I didn't notice any undesirable side effects of this. I only tested on Windows though.


dae:
Thanks Abdo, I can confirm this has resolved the menu issue on macOS.

PR #2734, opened 2023-10-14
Reduce shaking of bottom bar

### LINKED ISSUE #2668: Top/bottom bar "shaking" when hiding/showing
https://forums.ankiweb.net/t/small-bug-wobbly-bottom-bar/34497/

abdnh:
Closes #2668

The shaking should be mostly gone now. The animation is still not smooth, but this is probably a Qt issue.

### DIFF
--- a/qt/aqt/toolbar.py
+++ b/qt/aqt/toolbar.py
@@ -202,7 +202,7 @@ class BottomWebView(ToolbarWebView):
     def animate_height(self, height: int) -> None:
         self.web_height = height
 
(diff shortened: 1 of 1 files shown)


dae:
Thanks Abdo!

maxhark:
I have been having this issue on my m1 mac since like a month ago, never happened in almost a year before. I tried updating Anki and trying every renderer, as well as the reduce motion setting. The bottom bar still either shrinks (or blinks if reduce motion) almost half the time I hover over it.

PR #2735, opened 2023-10-14
Another approach to preventing selected shapes from shifting

hikaru-y:
Another approach to fixing #2725 .

I've found that the approach in #2729 is flawed. While it correctly converts multiple shapes that are selected to cloze text, it also causes an issue where the selection is cleared when it should not be. For example, if you select multiple shapes and release the mouse while moving them around, the selection will be cleared.

We should take the same approach as in #2682.


dae:
Looks good - thanks Hikaru!

Issue #1764, opened 2022-04-01
Improving dynamic TS package ergonomics


dae:
Re https://github.com/ankitects/anki/pull/1626#issuecomment-1024622000, I've been hunting for a way to make this a bit less cumbersome & more robust than manually declaring the exports. What if instead, we  export the module's interface in the same place as we register it, ensuring that the interface matches what we're exporting? Eg

```diff
diff --git a/ts/editor/NoteEditor.svelte b/ts/editor/NoteEditor.svelte
index 957b07616..51747d084 100644
--- a/ts/editor/NoteEditor.svelte
+++ b/ts/editor/NoteEditor.svelte
@@ -17,8 +17,8 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     }
 
     import { registerPackage } from "../lib/runtime-require";
-    import contextProperty from "../sveltelib/context-property";
-    import lifecycleHooks from "../sveltelib/lifecycle-hooks";
+    import contextProperty, { ContextProperty } from "../sveltelib/context-property";
+    import lifecycleHooks, { LifecycleHooks } from "../sveltelib/lifecycle-hooks";
 
     const key = Symbol("noteEditor");
     const [context, setContextProperty] = contextProperty<NoteEditorAPI>(key);
@@ -26,11 +26,18 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     export { context };
 
-    registerPackage("anki/NoteEditor", {
+    /** anki/NoteEditor */
+    export interface NoteEditorPackage {
+        context: ContextProperty<NoteEditorAPI>;
+        lifecycle: LifecycleHooks<NoteEditorAPI>;
+        instances: NoteEditorAPI[];
+    }
+    const noteEditorPackage = {
         context,
         lifecycle,
         instances,
-    });
+    } as NoteEditorPackage;
+    registerPackage("anki/NoteEditor", noteEditorPackage as any);
 </script>
 
 <script lang="ts">
```

Add-ons could then use require() with a cast to get a typed module:

```typescript
import type { NoteEditorPackage } from "@anki/editor/NoteEditor.svelte";
const NoteEditor = require("anki/NoteEditor") as NoteEditorPackage;
```

I spent some time trying to make this work in the context of a 'declare module' so we could keep the standard TS-style import, but couldn't figure out a way to spread the interface into the module definition except by doing so manually:

```
declare module "anki/NoteEditor" {
[...]

PR #2739, opened 2023-10-15
Add APIs for IO card rendering

glutanimate:
Extends IO with a number of **JS APIs** that allow altering **IO card rendering**:

- Add `onWillDrawShapes` and `onDidDrawShapes` callback parameters to IO setup function, allowing consumers to modify rendering both before and after the occlusion masks are drawn
- Add cloze ordinal to `Shape` objects
- Allow consumers to draw their own shapes directly by exposing `drawShape` and shape classes

**Refactoring changes** as a result of these additions:

- Create a new namespace for all public IO APIs available at card render time under `globalThis.anki.imageOcclusion`
- Accordingly, change the image occlusion setup call from `anki.setupImagecloze()` to `anki.imageOcclusion.setup()` while retaining backwards  compatibility with old templates
- Update (de)normalization of shapes to be non-mutating. Mutating shapes would likely be a recipe for trouble when combined with IO API use by external consumers.
- Consistently use "image occlusion" terminology for all identifiers, compared to the previous mix of "image occlusion" and "image cloze"

-----

I was a bit unsure on the best way to add these APIs, walking through multiple different implementations and e.g. also experimenting with an extended hook system, but I eventually ended up settling on this template-centric approach with callback parameters.

(FWIW, [pushed the extended hook system here](https://github.com/ankitects/anki/compare/main...glutanimate:anki:poc-extended-reviewer-hooks), in case it could still come in handy in the future)

One point still has me a bit confused, however. We currently expose globals in the reviewer context through a number of means:

- For somewhat older private APIs used in Python → JS communication (e.g. `_drawMark`), we [`export` the corresponding symbols ](https://github.com/ankitects/anki/blob/main/ts/reviewer/index.ts#L220)in `reviewer/index.ts`
- For newer public APIs like `mutateNextCardStates` and `setupImageCloze`, we register them to a [`globalThis.anki` object](https://github.com/ankitects/anki/blob/main/ts/reviewer/index.ts#L16-L17). We do so in two places, `reviewer/index.ts` (for desktop) and [`reviewer/reviewer_extras.ts`](https://github.com/ankitects/anki/blob/main/ts/reviewer/reviewer_extras.ts) (I assume for AnkiMobile, AnkiWeb and maybe AnkiDroid?)
- Finally we have public template APIs like `onUpdateHook` that are both exposed via `export`s and [`registerPackage`](https://github.com/ankitects/anki/blob/main/ts/reviewer/index.ts#L266C32-L274), but do not seem to be exposed in `reviewer/reviewer_extras.ts`, while still being available on mobile

Which option, or combination thereof would you recommend using here? Given that we were already using the `anki` global for `setupImageCloze`, that's what I ended up going with. But I was wondering if we have started migrating more towards the `registerPackage` approach (or started and then stopped, given `mutateNextCardStates` and `setupImageCloze`?).

### DIFF
--- a/ts/image-occlusion/review.ts
+++ b/ts/image-occlusion/review.ts
@@ -4,41 +4,51 @@
 import * as tr from "@tslib/ftl";
 
 import { optimumPixelSizeForCanvas } from "./canvas-scale";
-import type { Shape } from "./shapes";
+import { Shape } from "./shapes";
 import { Ellipse, extractShapesFromRenderedClozes, Polygon, Rectangle, Text } from "./shapes";
 import { TEXT_BACKGROUND_COLOR, TEXT_FONT_FAMILY, TEXT_PADDING } from "./tools/lib";
 import type { Size } from "./types";
 
-export type ImageClozeShapes = {
+export type ImageOcclusionShapes = {
     active: Shape[];
     inactive: Shape[];
 };
 
 export type DrawShapesFilter = (
-    shapes: ImageClozeShapes,
+    shapes: ImageOcclusionShapes,
     properties: ShapeProperties,
     context: CanvasRenderingContext2D,
-) => ImageClozeShapes;
+) => ImageOcclusionShapes;
 
 export type DrawShapesCallback = (
-    shapes: ImageClozeShapes,
+    shapes: ImageOcclusionShapes,
     properties: ShapeProperties,
     context: CanvasRenderingContext2D,
 ) => void;
 
-export interface SetupImageClozeOptions {
+export interface SetupImageOcclusionOptions {
     onWillDrawShapes?: DrawShapesFilter;
     onDidDrawShapes?: DrawShapesCallback;
 }
 
-export function setupImageCloze(setupOptions?: SetupImageClozeOptions): void {
+export const imageOcclusionAPI = {
+    setup: setupImageOcclusion,
+    drawShape,
+    Ellipse,
+    Polygon,
+    Rectangle,
+    Shape,
+    Text,
+};
+
+function setupImageOcclusion(setupOptions?: SetupImageOcclusionOptions): void {
     window.addEventListener("load", () => {
-        window.addEventListener("resize", () => setupImageCloze(setupOptions));
+        window.addEventListener("resize", () => setupImageOcclusion(setupOptions));
     });
-    window.requestAnimationFrame(() => setupImageClozeInner(setupOptions));
+    window.requestAnimationFrame(() => setupImageOcclusionInner(setupOptions));
 }
 
-function setupImageClozeInner(setupOptions?: SetupImageClozeOptions): void {
+function setupImageOcclusionInner(setupOptions?: SetupImageOcclusionOptions): void {
     const canvas = document.querySelector(
         "#image-occlusion-canvas",
     ) as HTMLCanvasElement | null;
--- a/ts/image-occlusion/review.ts
+++ b/ts/image-occlusion/review.ts
@@ -105,37 +105,46 @@ function drawShapes(
     }
 
     for (const shape of activeShapes) {
-        drawShape(
+        drawShape({
             context,
             size,
             shape,
-            properties.activeShapeColor,
-            properties.activeBorder.color,
-            properties.activeBorder.width,
-        );
+            fill: properties.activeShapeColor,
+            stroke: properties.activeBorder.color,
+            strokeWidth: properties.activeBorder.width,
+        });
     }
     for (const shape of inactiveShapes.filter((s) => s.occludeInactive)) {
-        drawShape(
+        drawShape({
             context,
             size,
             shape,
-            properties.inActiveShapeColor,
-            properties.inActiveBorder.color,
-            properties.inActiveBorder.width,
-        );
+            fill: properties.inActiveShapeColor,
+            stroke: properties.inActiveBorder.color,
+            strokeWidth: properties.inActiveBorder.width,
+        });
     }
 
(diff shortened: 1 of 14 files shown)


ts/image-occlusion/review.ts
```diff
@@ -56,47 +82,74 @@ function setupImageClozeInner(): void {
...
+    const context: CanvasRenderingContext2D = canvas.getContext("2d")!;
     const size = canvas;
-    for (const shape of extractShapesFromRenderedClozes(".cloze")) {
-        drawShape(ctx, size, shape, properties, true);
+
+    let activeShapes = extractShapesFromRenderedClozes(".cloze");
+    let inactiveShapes = extractShapesFromRenderedClozes(".cloze-inactive");
+    let properties = getShapeProperties();
+
+    const processed = onWillDrawShapes?.({ activeShapes, inactiveShapes, properties }, context);
```
glutanimate, line 100: Using a regular callback and expecting API consumers to mutate the `DrawShapesData` object would also be possible here. Pondered a bit between that und the current filter approach, but eventually went with the latter to be more explicit about the capabilities of this API vs. e.g. `onDidDrawShapes` where mutating the input would have no effect.

ts/image-occlusion/review.ts
```diff
@@ -56,47 +82,74 @@ function setupImageClozeInner(): void {
...
-    const { color, border } = active
-        ? {
-            color: properties.activeShapeColor,
-            border: properties.activeBorder,
-        }
-        : {
-            color: properties.inActiveShapeColor,
-            border: properties.inActiveBorder,
-        };
+function drawShape({
```
glutanimate, line 140: Refactored `drawShape` to be more generally usable by API consumers and shift image occlusion implementation details into its callers.

ts/image-occlusion/shapes/ellipse.ts
```diff
@@ -28,20 +28,28 @@ export class Ellipse extends Shape {
...
-    makeNormal(size: Size): void {
-        super.makeNormal(size);
-        this.rx = xToNormalized(size, this.rx);
-        this.ry = yToNormalized(size, this.ry);
+    toNormal(size: Size): Ellipse {
+        return new Ellipse({
+            ...this,
+            rx: xToNormalized(size, this.rx),
+            ry: yToNormalized(size, this.ry),
+            left: xToNormalized(size, this.left),
```
glutanimate, line 40: Might be able to shave off the `left` and `top` normalization in this and other `Shape` child classes by more clever use of method inheritance

dae:
Apart from that, I'm ok with the changes you have here, but a regression in the text tool will need to be fixed first. Depending on the size of the review window, the text aspect ratio is getting messed up, and at some sizes the text will disappear completely.
2023-10-17T10:14:45,761953419+10:00
2023-10-17T10:15:21,906738824+10:00

Re exporting:

- mobile currently has its own reviewer implementation that exposes onUpdateHook. It imports reviewer_extras.
- the motivation for registerPackage() was allowing Svelte components authored by add-on authors to be integrated into our pages, as if they bundle their own copy of Svelte, things break. I'd suggest avoiding registerPackage() for now - Svelte 5's runes may make it no longer necessary, or we may end up avoiding dynamic components and exposing add-ons is a generic JS way instead in the future.

glutanimate:
> regression in the text tool will need to be fixed first. Depending on the size of the review window, the text aspect ratio is getting messed up, and at some sizes the text will disappear completely.

Huh, that's a weird one, good catch! But it looks like it's not specific to this PR – I can also reproduce with main (0e2453243961fabbb2c08ba71f2f2648b2b17285):

https://github.com/ankitects/anki/assets/5459332/0c5c602a-5449-4c29-9a87-c1b2fd815288

So I would propose we split it off into a separate issue.

ts/image-occlusion/shapes/index.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright: Ankitects Pty Ltd and contributors
+// License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+
+export { Shape } from "./base";
+export { Ellipse } from "./ellipse";
+export { extractShapesFromRenderedClozes } from "./from-cloze";
+export { Polygon } from "./polygon";
+export { Rectangle } from "./rectangle";
+export { Text } from "./text";
```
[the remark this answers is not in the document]
glutanimate: I can definitely see that. I would also agree that the convenience factor in internal use really isn't that high. 

However,  given that third-party code consuming these APIs will likely also want to have access to their corresponding types, I'm wondering how we should best handle this here.

E.g., with the current state I can just import the various `Shape` types from this barrel file without having to worry about the internal structure of the `shapes` package (annotating shape types with e.g. `typeof globalThis.anki.imageOcclusion.Shape` doesn't work without some major TS acrobatics which are likely a bad practice of their own).

Perhaps just bundling type exports in barrel files like these could be an option? Or splitting out type exports in a different way like d.ts files?

FWIW, I don't feel particularly strong about this from a stylistic standpoint, just want to make sure that we go with as defensive a solution as possible, so that using `anki` types is not too brittle long-term.

dae:
Sorry, I did briefly test this on main before reporting it; not sure why I didn't spot the issue then.

ts/image-occlusion/shapes/index.ts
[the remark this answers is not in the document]
dae: If you're consuming the types then it's arguably public API, and I remove my objection.

I'd like to return to the question of the best way to expose these to add-on authors in the future - some previous discussion about it can be found at https://github.com/ankitects/anki/issues/1764.

PR #2740, opened 2023-10-15
Prevent media syncing from blocking collection tasks

### LINKED ISSUE #2736: Don't block collection tasks when syncing media
@abdnh I just noticed https://community.ankihub.net/t/a-progress-dialog-is-shown-for-background-operations-such-as-media-syncing-in-anki-23-10-beta/8314 - the change in behavior was not intentional. We can probably resolve this by using without_collection in start_monitoring(), and then use self.mw.backend instead of self.mw.col.

abdnh:
Closes #2736

> then use self.mw.backend instead of self.mw.col.

That doesn't seem to be necessary?

### DIFF
--- a/qt/aqt/mediasync.py
+++ b/qt/aqt/mediasync.py
@@ -61,7 +61,9 @@ class MediaSyncer:
 
                 time.sleep(0.25)
 
-        self.mw.taskman.run_in_background(monitor, self._on_finished)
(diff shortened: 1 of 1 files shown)


dae:
Thanks Abdo. .backend would better reflect that the collection isn't being locked, but you're right that it's not strictly necessary.

PR #2741, opened 2023-10-16
Rename Analyze to Evaluate in FSRS Deck Options

user1823:
I know that these words are synonyms, but synonyms often have subtle differences in meaning. Here, the word "Evaluate" fits the function of the button better.


meliache:
Just the opinion from another beta tester: Both names are tbh equally fine with me. Ideally it should be obvious that the button doesn't _do_ or change anything but is just for showing some statistics. And it should be clear that the button is just optional for nerds. Not sure how to best show that in the UX besides the name. And of course I hope that eventually we'll get better documentation of the two figures-of-merit (log-loss and RSME) that are shown.

Also a bit off-topic, but do we really need that button? The calculation of the figures of merit is so fast that we could automatically calculate the two numbers every time the weights change and display them under the weights field. Though that also might confuse users, the status-quo has the advantage of hiding this nerd-information.

dae:
> The calculation of the figures of merit is so fast

It can take multiple seconds on larger collections / slower devices - I think it's better triggered manually.

PR #2742, opened 2023-10-17
Drop support for checkpoints

abdnh:
I hope this is what you have in mind. I didn't notice any issues testing things such as syncing & backups and with some add-ons enabled (except of course some add-ons breaking because they are using removed methods), but DB stuff are tricky.

Is the `db.transact()` API you suggested intended for add-ons that access the database directly?

### DIFF
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -125,6 +125,7 @@ class Card(DeprecatedNamesMixin):
             desired_retention=self.desired_retention,
         )
 
+    @deprecated(info="please use col.update_card()")
     def flush(self) -> None:
         hooks.card_will_flush(self)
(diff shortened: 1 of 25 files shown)


dae:
Thanks Abdo! .transact() applies to both direct DB access and backend methods - the backend methods use savepoints which are like an inner transaction, so they can be grouped together with an outer transaction if add-on authors want to ensure multiple changes are applied atomically.

abdnh:
@RisingOrange db.begin() is used in AnkiHub and can be safely removed as far as I can tell.

RisingOrange:
> @RisingOrange db.begin() is used in AnkiHub and can be safely removed as far as I can tell.

Thanks!

Issue #2692, opened 2023-09-29
Odd behavior when user users an odd number of colons


dae:
We should probably be stripping colons out of the the segments when converting to the machine format.

https://forums.ankiweb.net/t/syncing-not-working-missing-deck-other-issues/35084

geldbogen:
Hey I would like to work on this issue. Could you maybe give me a resource, where it is documented how Anki converts the database to the "machine format"? 
That would be very helpful :)

dae:
The relevant code is in decks/name.rs. Internally, Anki uses \x1f instead of '::' to separate deck levels.

dae:
How did you go? No problem if you need a bit more time, but if you don't think you'll be able to work on it, please let us know so someone else can pick it up.

geldbogen:
Sorry for my late reply, I don't have any knowledge in Rust, therefore I don't think I can solve this issue. However currently I am reading into it, and if next week or so nobody came up with a solution yet I might be able to contribute something :)

swektr:
Hey, took a look at this and I have a solution for either case, but should 1. ALL colons get stripped or 2. just leading/trailing colons? For instance maybe “Biology 8:30am” is intentional but likely not “:Biology”.

dae:
We should probably only be doing leading/trailing colons - single ones in the middle of a name won't cause any ambiguity when it comes to splitting.

PR #2748, opened 2023-10-19
Trim extra colons on conversion to NativeDeckName

swektr:
This fixes issue #2692. Leading/trailing colons are remove along with white space so 
` : : :foo:bar: : : `" becomes "`foo:bar`, 
`:::` becomes `blank\x1fblank`


dae:
Thank you!

Issue #2541, opened 2023-06-10
Silent uninstall of Anki no longer possible on Windows


superusercode:
Documented in https://github.com/microsoft/winget-pkgs/issues/109046#issuecomment-1578281688 , using the nullsoft silent flags does not hide the anki uninstall prompt and prevents updating anki through winget.

> The installer itself requests for an uninstall once it detects a previous version in the system but the uninstall is not unattended/silent. Even if we include UpgradeBehavior: uninstallPrevious in the manifest, it does not resolve the issue. You can confirm this by using the --uninstall-previous experimental arg in the current preview version of WinGet, which has the same effect as UpgradeBehavior: uninstallPrevious in the manifest.
> Are there are any command line switches that can bypass the interactive prompts during the uninstall? If so we can add them in the manifest and make the uninstall silent. But even if that's resolved, it seems the uninstaller returns a success code even though the package isn't uninstalled completely that signals WinGet to proceed with the installation process. I believe the issue should be raised up to the publisher for a fix.
> ![image](https://github.com/ankitects/anki/assets/60588434/85f88550-470b-474d-b18c-f5bba13fa6b0)

dae:
This only needs to happen once (we're migrating from system-wide install to single-user install), and the uninstaller program is from the old version, so it's not easy for us to retroactively update it.

PR #2757, opened 2023-10-20
Make it possible to silent uninstall with WinGet

kelciour:
The related issue probably can be closed: https://github.com/ankitects/anki/issues/2541

Reference:

> https://github.com/microsoft/winget-pkgs/issues/49601#issuecomment-1054511535
> 
> Currently winget is able to silently uninstall nullsoft and inno setup application that register `QuietUninstallString` in Registry and Notepad++ here does not do that, this can either be fixed by requesting developer to add it or once when the feature at [microsoft/winget-cli#1885](https://github.com/microsoft/winget-cli/issues/1885) implemented.

https://github.com/notepad-plus-plus/notepad-plus-plus/commit/c19033c91711bc5c35bc63b804c5d31756ee58a4

> https://github.com/microsoft/winget-cli/issues/1368#issuecomment-898698659
> 
> I believe Winget simply executes the uninstaller listed in the registry. I don't think it actually remembers the installer type nor has the ability to modify the command line of the uninstaller.
> 
> Ask the Krita developers to also register the silent uninstall command:
> 
> ```
> WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\${KRITA_UNINSTALL_REGKEY}" "QuietUninstallString" '"$INSTDIR\uninstall.exe" /S'
> ```

https://github.com/KDE/krita/commit/a56d6aad87f78a5a04f0dcb656f124e048a77738


dae:
Nice find!

PR #2758, opened 2023-10-21
Add APIs for IO mask editing

glutanimate:
Extends IO with a number of **JS and Python APIs** that facilitate interacting with the **mask editor** during adding or editing notes:

- Add `editor_mask_editor_did_load_image` hook that signals full initialization and image loading to Python
- Expose methods for querying, adding, and deleting shapes via `globalThis.maskEditor`

### DIFF
--- a/ts/image-occlusion/MaskEditor.svelte
+++ b/ts/image-occlusion/MaskEditor.svelte
@@ -23,6 +23,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         setupMaskEditor,
         setupMaskEditorForEdit,
     } from "./mask-editor";
+    import { MaskEditorAPI } from "./tools/api";
     import Toolbar from "./Toolbar.svelte";
 
     export let mode: IOMode;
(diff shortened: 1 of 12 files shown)


ts/image-occlusion/MaskEditor.svelte
```diff
@@ -32,8 +34,17 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
 
+    $: {
+        globalThis.maskEditor = canvas ? new MaskEditorAPI(canvas) : null;
+    }
+
     const dispatch = createEventDispatcher();
 
+    function onImageLoaded() {
+        const data = mode.kind === "add" ? mode.imagePath : mode.noteId;
+        bridgeCommand(`ioImageLoaded:${JSON.stringify(data)}`);
```
krmanik, line 45: This line may not work on Mobile client because this is specific desktop client.

glutanimate: Thanks for taking a look, Mani!

Would this raise an exception in this case, or is `bridgeCommand` just a no-op on mobile? I was assuming the latter, but don't actually know for sure.

krmanik: This will be the error.
```
Uncaught ReferenceError: bridgeCommand is not defined.
```

glutanimate: Thanks! I moved the `bridgeCommand` calls to `NoteEditor`, which should fix the mobile compat hopefully.

dae: Good catch Mani!

PR #2760, opened 2023-10-21
Indicate current occlusion type in pop-up menu

hikaru-y:
This adds an indicator to the popup menu to show the current occlusion type.

https://forums.ankiweb.net/t/anki-23-10-beta-5-6/35677/46

io-indicator_2
---
This also does the following:
- #### Refactor
  - Simplifies handling of occlusion types in the editor code.
- #### Change
  - Now updates the occlusions field every time a mask or text is updated, not only in editing mode but also in adding mode, so that IO cards can be previewed correctly in the card layout screen.
- #### Fix
  - Fixes a11y warnings in Toolbar.svelte.
    a11y (at 59d89dd) 

---
edit: Changed the 1st screenshot.

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -373,7 +373,7 @@ class AddCards(QMainWindow):
         self.ifCanClose(doClose)
 
     def add_io_note(self) -> None:
-        self.editor.web.eval("setOcclusionFieldInner()")
+        self.editor.web.eval("updateOcclusionsField();")
         self.add_current_note()
         self.editor.web.eval("resetIOImageLoaded()")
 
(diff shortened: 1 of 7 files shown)


dae:
Sorry Hikaru, this took a merge conflict. The changes look good though - the code is simpler, and the indicator is a nice touch.

glutanimate:
Thanks, Hikaru!

In terms of resolving the merge conflict: I would say let's keep the just added `MaskEditorAPI.getShapes` in sync with `baseShapesFromFabric`'s signature and also drop `occludeInactive` there.

dae:
Thanks Hikaru!

PR #2761, opened 2023-10-21
Change export notes' parent window in card browser

kelciour:
I'd like to suggest the following change.

## Before

image

image

## After

image


dae:
Makes sense - thanks!

PR #2762, opened 2023-10-21
Fix modified polygon not being saved when editing IO note

hikaru-y:
This PR fixes an issue where modified polygons are not saved when editing an IO note.

#### Steps to reproduce
1. Add an IO note containing a polygon
1. Open the note in _edit current_ or _browse_
1. Select the polygon and transform or move it
1. Close the editor or switch to another note
1. Open the note again

#### Result
- The polygon has been saved as it was before the modification.


dae:
Thank you!

PR #2764, opened 2023-10-22
Allow disabling occlusion mask border

glutanimate:
Previously, setting the border width to zero via the template CSS properties would have no effect.

### DIFF
--- a/ts/image-occlusion/review.ts
+++ b/ts/image-occlusion/review.ts
@@ -153,7 +153,9 @@ function drawShape({
     if (shape instanceof Rectangle) {
         ctx.fillRect(shape.left, shape.top, shape.width, shape.height);
         // ctx stroke methods will draw a visible stroke, even if the width is 0
-        if (strokeWidth) ctx.strokeRect(shape.left, shape.top, shape.width, shape.height);
+        if (strokeWidth) {
+            ctx.strokeRect(shape.left, shape.top, shape.width, shape.height);
+        }
     } else if (shape instanceof Ellipse) {
         const adjustedLeft = shape.left + shape.rx;
         const adjustedTop = shape.top + shape.ry;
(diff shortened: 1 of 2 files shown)


ts/image-occlusion/review.ts
```diff
@@ -152,7 +152,8 @@ function drawShape({
     ctx.lineWidth = strokeWidth;
     if (shape instanceof Rectangle) {
         ctx.fillRect(shape.left, shape.top, shape.width, shape.height);
-        ctx.strokeRect(shape.left, shape.top, shape.width, shape.height);
+        // ctx stroke methods will draw a visible stroke, even if the width is 0
+        if (strokeWidth) ctx.strokeRect(shape.left, shape.top, shape.width, shape.height);
```
dae, line 156: Please use {} even for single line ifs

glutanimate: Felt weird to me as well, but `dprint ` seems to prefer this format. At least, `./ninja format` does remove the curlies if I add them. Curiously enough, every formatter seems to handle this differently: VS Code's native formatter preserves the curly braces, and `prettier` preserves the statement if you start _without_ braces, but converts it to a multi-line statement if you start _with_ braces.

Seems like we could set the behavior with `dprint`'s `useBraces` setting:

> useBraces
> If braces should be used or not.
> 
> "maintain" - Uses braces if they're used. Doesn't use braces if they're not used.
> **"whenNotSingleLine" - Uses braces when the body is on a different line. (Default)**
> "always" - Forces the use of braces. Will add them if they aren't used.
> "preferNone" - Forces no braces when the header is one line and body is one line. Otherwise forces braces.

WDYT?

dae: You can prevent it from removing the braces if you add a newline after the { before formatting, but changing the option to 'always' looks like it would make things easier.

glutanimate: :+1:  changed the statements and set the dprint option

dae:
Thanks!

PR #2765, opened 2023-10-22
Add auto-advance options to deck preset

### LINKED ISSUE #2645: Add auto-advance options to deck preset
The mobile clients have auto-advance features, so we should probably add this to the desktop as well, and move the option into the deck preset: https://forums.ankiweb.net/t/auto-advance-per-deck/33994

abdnh:
Closes #2645

### DIFF
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -590,6 +590,7 @@ class Reviewer:
             ("5", self.on_pause_audio),
             ("6", self.on_seek_backward),
             ("7", self.on_seek_forward),
+            ("Shift+A", self.toggle_auto_advance),
             *self.korean_shortcuts(),
         ]
 
@@ -968,7 +969,7 @@ timerStopped = false;
             [tr.studying_replay_own_voice(), "V", self.onReplayRecorded],
             [
                 tr.actions_auto_advance(),
-                "",
+                "Shift+A",
                 self.toggle_auto_advance,
                 dict(checked=self.auto_advance_enabled),
             ],
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -185,6 +185,7 @@ class Reviewer:
     def cleanup(self) -> None:
         gui_hooks.reviewer_will_end()
         self.card = None
+        self.auto_advance_enabled = False
 
     def refresh_if_needed(self) -> None:
         if self._refresh_needed is RefreshNeeded.QUEUES:
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -434,6 +434,8 @@ class Reviewer:
         this_timer = None
 
         def on_show_question_timeout() -> None:
+            if self.card is None:
+                return
             if self._show_question_timer == this_timer:
                 try:
                     answer_action = list(AnswerAction)[conf["answerAction"]]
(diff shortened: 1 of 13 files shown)


qt/aqt/reviewer.py
```diff
@@ -364,6 +388,28 @@ def _showQuestion(self) -> None:
         # user hook
         gui_hooks.reviewer_did_show_question(c)
 
+        conf = self.mw.col.decks.config_dict_for_deck_id(self.card.current_deck_id())
+        timer = None
+
+        if conf["secondsToShowAnswer"]:
+            timer = self._show_answer_timer = self.mw.progress.timer(
```
dae, line 395: Any particular reason you didn't use single_shot() here?

dae (commented):
Looking good. Sorry, I forgot to mention one thing about how this should work: there should be a menu action that toggles auto-advance, so it doesn't happen until the user explicitly starts it. Its value shouldn't be saved, so it should always start off. The primary reason for this is that the congratulations page will switch back to review when the day rolls over, and users won't want to wake up to find all their cards have already been answered. :-)

qt/aqt/reviewer.py
abdnh: This is to compare the timers so that we make sure a previous card's timer doesn't get to execute on the next card.

abdnh:
> users won't want to wake up to find all their cards have already been answered. :-)

That's a feature I'd love to use 😂

qt/aqt/reviewer.py
```diff
@@ -883,6 +966,12 @@ def _contextMenu(self) -> list[Any]:
             [tr.studying_audio_and5s(), "7", self.on_seek_forward],
             [tr.studying_record_own_voice(), "Shift+V", self.onRecordVoice],
             [tr.studying_replay_own_voice(), "V", self.onReplayRecorded],
+            [
+                tr.actions_auto_advance(),
+                "",
```
abdnh, line 971: I tried `Shift+A` but it didn't work for me. Any suggestions?

dae: It's a bit confusing; it needs to be updated in multiple locations.

```diff
diff --git a/qt/aqt/reviewer.py b/qt/aqt/reviewer.py
index 856b5dd85..7df436318 100644
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -590,6 +590,7 @@ class Reviewer:
             ("5", self.on_pause_audio),
             ("6", self.on_seek_backward),
             ("7", self.on_seek_forward),
+            ("Shift+A", self.toggle_auto_advance),
             *self.korean_shortcuts(),
         ]
 
@@ -968,7 +969,7 @@ timerStopped = false;
             [tr.studying_replay_own_voice(), "V", self.onReplayRecorded],
             [
                 tr.actions_auto_advance(),
-                "",
+                "Shift+A",
                 self.toggle_auto_advance,
                 dict(checked=self.auto_advance_enabled),
             ],
````

dae:
Couple of remaining issues:

- toggling auto-advance to on doesn't start it until you manually show answer
- it remains on when you go back to the overview and then back into review
- it might be nice if we automatically stop it when the user opens add/edit/browse

dae:
(in case it wasn't clear, I'm planning to skip over this for the 23.10 release, so there's no time pressure on this)

dae:
Thanks Abdo. I might need to hold off on this for a week or two, so we can get a hotfix update out first.

PR #2766, opened 2023-10-22
Fix IO groups breaking upon editing

### LINKED ISSUE #2759: Grouped shapes lose their grouped status upon editing
When editing a note with grouped shapes, the shapes seem to lose their grouping in the masks editor. Performing changes that prompt the note to be saved causes all groups to be dissolved.

**Steps to reproduce**:

1. Create an IO note with grouped shapes
2. Edit the note while e.g. moving one shape slightly
3. Close the editor

**Expected results**

- Shapes continue to be grouped during editing
- Shapes continue to be grouped after editing

**Actual results**

- Shapes are not grouped during editing
- Shapes are not grouped after editing

https://github.com/ankitects/anki/assets/5459332/effe9663-2ef7-486d-a5a2-3ce2488882a1

abdnh:
Closes #2759

### DIFF
--- a/proto/anki/image_occlusion.proto
+++ b/proto/anki/image_occlusion.proto
@@ -59,9 +59,13 @@ message GetImageOcclusionNoteResponse {
     string value = 2;
   }
 
-  message ImageOcclusion {
+  message ImageOcclusionShape {
     string shape = 1;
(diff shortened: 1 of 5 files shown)


dae:
Thank you for the quick fix!

PR #2767, opened 2023-10-22
Fix scaling of text objects

### LINKED ISSUE #2752: Text labels have altered aspect ratio or disappear as window is resized
Previously reported on https://github.com/ankitects/anki/pull/2739#issuecomment-1765472055

@abdnh mind looking into this when you have a chance?

abdnh:
Closes #2752


dae:
This seems to be an improvement, but unfortunately hasn't fully solved the problem. I created a tall (1000x5000) blank image and added an occlusion. In the review screen, it gets distorted:

2023-10-23T09:45:13,971199507+10:00
2023-10-23T09:45:02,049799594+10:00

Here's the image I tested with:
Untitled

abdnh:
Should be fixed now.

dae:
Nice!

PR #2773, opened 2023-10-24
Fix shape selection malfunction after creating shapes in succession

### LINKED ISSUE #2769: Shape selection seems to stick to newly created shapes, sometimes producing artifacts
https://forums.ankiweb.net/t/anki-23-10-release-candidate/35967/40

Can partially reproduce this here. It seems like the selection / resize controls currently stick to newly drawn shapes. Have not been able to reproduce the [Windows-solitaire-like](https://i.kym-cdn.com/entries/icons/original/000/030/063/solitaire.jpg) artifacting yet.

Steps to reproduce the selection issue:

1. Draw a shape
2. Draw a second shape
3. Select the first shape

**Expected result**: First shape is selected

**Actual result**: First shape is momentarily selected before selection jumps back to second shape

hikaru-y:
Fixes  #2769 

Also, disables rotation when duplicating a shape.

### DIFF
--- a/ts/image-occlusion/tools/lib.ts
+++ b/ts/image-occlusion/tools/lib.ts
@@ -120,6 +120,7 @@ const pasteItem = (canvas: fabric.Canvas): void => {
             top: clonedObj.top + 10,
             evented: true,
         });
+        disableRotation(clonedObj);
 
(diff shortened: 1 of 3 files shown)


dae:
Thanks Hikaru!

PR #2775, opened 2023-10-25
Fix Linux install script CWD check

virinci:
This PR fixes:
1. A minor bug that results in unexpected behavior when the Linux `install.sh` script is executed from a directory that has any file named `install.sh`.

This can be tested as:
```shell
$ touch install.sh
$ ./anki/install.sh  # this proceeds to try and install all the files
```
where the actual install script is located in the sub-directory named `anki`.

The solution for this matches the CWD against the parent directory of the install script.

2. The second fix in the `cp` command was suggested by `shellcheck` in order to avoid unexpected behavior if any of the files being copied has a leading hyphen in its name.


dae:
Feels like a pretty niche case :-) But thanks.

PR #2787, opened 2023-10-29
Resolve a11y for tag options buttons

grepgrok:
This is in working on #2564 and focused on the three items under `ts/tag-editor/tag-options-button`. To break them down:

* `TagAddButton`: nothing much needs to happen here
    * there is no need for any keyevents to occur here as you cannot (and should not need to) focus on it.
* `TagOptionsButton`: we can just ignore the warnings
    * it complains about there being `on:click` on a static element except this is explicitly done to ensure the container does nothing (i.e. non-interactable element is successfully non-interactable but the checker doesn't realize this)
* `TagsSelectedButton`: can now `Shift-Tab` to focus back to this and press `Enter` or `Space` to show the options
    * I debated adding keyboard controls to switch between these options but decided against it as each is clearly given the option of an intuitive keyboard option anyway
    * I also vertically centered the button in the row because it was annoying me that it was slightly off

In accordance with standard ARIA, both `TagAddButton` and `TagsSelectedButton` would ideally be `<button>` but when I tried to do that, it screwed with the styling and I couldn't figure out how to fix that. The current change merely maintains styling while resolving a11y warnings.


grepgrok:
also, is there some way I can run the checks locally before submitting a pull request?

dae:
Thanks Ben! I'm a bit swamped with the 23.10 release, but will try to return to this in the next few days. You can run local checks with ./check.

ts/change-notetype/Alert.svelte
```diff
@@ -25,11 +25,27 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
     $: icon = collapsed ? plusIcon : minusIcon;
+
+    function onKeyDown(event: KeyboardEvent) {
+        switch (event.code) {
+            case "Enter":
+            case "Space":
+                collapsed = !collapsed;
+                break;
+        }
+    }
```
dae, line 36: It might be worth factoring this out into a helper function so we can use it like onEnterOrSpace(() => (collapsed = !collapsed))

grepgrok: You previously mentioned something similar to that [in the original #2564 issue](https://github.com/ankitects/anki/issues/2564#issuecomment-1634072182). Given how common those buttons being synonymous are in the the ARIA APG, I would agree that that would be good as a universal function. Also, it would be helpful for consistency in "select the thing I am focusing on" (I might prefer to call it something like onSelect() on onAriaSelect()). Though identifying what in the code base that would actually work on I think would take a bit and be best suited to a different issue or PR. I'll look into it.

ts/change-notetype/Alert.svelte
```diff
@@ -38,14 +54,14 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     {/if}
     {unusedMsg}
     {#if collapsed}
-        <div>
+        <div id="unused-list-collapsed">
```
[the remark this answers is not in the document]
grepgrok: No; I believe the ARIA APG would classify this component as the disclosure pattern which optionally suggests adding an aria-controls property which would take these ids. I was initially trying to get it together but decided against as the collapsed items are very close in the DOM.

grepgrok:
Thanks for the feedback. I was trying to figure out what components were and evidently forgot to remove all my logging. Took care of it now though.

As for the onEnterOrSpace() helper function, I'm not sure what parts of the code need that (or frankly how to do it, though I'm sure a little research can solve that). Also, I would argue that the underlying idea of "select this thing" should be brought under a greater issue of custom keyboard bindings and shortcuts and generally separating abstract actions like "select" from realized actions like "press enter." Though, I'm sure that requires some funky implementations for how to actually do that well (I personally think Visual Studio Code has a very nice and robust system for reference, but no idea how it works) and the team probably has other issues to handle right now.

Sorry, bit of rant. TLDR; I can put together an onEnterOrSpace() helper but I think a more thorough, deeper look into keyboard binding customization is in order.

grepgrok:
First off, I'm so sorry for the abysmally messy commit history, I'm new to... basically all this git and github stuff and I have no idea how to make it look less like a mess.

I added the onEnterOrSpace helper so it can be referenced pretty much exactly as you suggested. I think a more semantic name like "onKeyboardSelect" may help identify it as something to be used in more places. I'll try to look through the components to see what is doing the semantic role it fills and should be using it.

ts/change-notetype/tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "include": ["*"],
     "references": [
         { "path": "../lib" },
+        { "path": "../lib/keys" },
```
dae, line 6: This section lists typescript projects, not individual files, so this line is not required.

ts/lib/keys.ts
```diff
@@ -120,3 +120,14 @@ export function isArrowDown(event: KeyboardEvent): boolean {
 
     return isApplePlatform() && metaPressed(event) && event.code === "KeyN";
 }
+
+export function onEnterOrSpace(callback: () => void): (event: KeyboardEvent) => void {
+    return function _onEnterOrSpace(event: KeyboardEvent): void {
```
dae, line 125:
```suggestion
    return (event: KeyboardEvent) => {
```

dae (commented):
No worries about the commits; they'll all get squashed into one on merge anyway.

I have a minor preference for the current name, as I'm not sure I'd understand what "onKeyboardSelect" was trying to communicate if I didn't already know.

dae:
Hi Ben,

Unfortunately this took a conflict - would you mind updating CONTRIBUTORS again? Happy to merge after that.

grepgrok:
I have no idea why this is failing

dae:
Sorry, this was due to a stale file on the CI server. All good now :+1:

PR #2788, opened 2023-10-29
Allow creation of empty filtered decks

### LINKED ISSUE #2753: Add a 'create even if empty' option to filtered deck screen
Making it an option instead of a confirmation prompt avoids the need to run the search twice. The option would need to be hidden when the filtered deck already exists.

https://forums.ankiweb.net/t/i-want-to-create-filtered-decks-even-if-they-dont-have-cards-yet/35902

Gustaf-C:
Closes #2753.

It was specified in the issue that the option should be hidden when the deck already exists. I have however allowed it to be shown again if any of the parameters are changed. My thinking behind this is that if someone would like to change an already existing filtered deck they should be allowed to do that even if it would be empty.

Another thought, should this be implemented into the custom study screen as well? Could be that someone wanted to create a template for filtered decks. In that case I could probably implement that too.


dae:
Thanks Gustaf, I'll get this reviewed in the next few days. We can probably skip custom study, as I haven't heard anyone asking for this functionality in that screen AFAIR.

qt/aqt/filtered_deck.py
```diff
@@ -55,22 +55,25 @@ def __init__(
...
         # set on successful query
         self.deck: FilteredDeckForUpdate
 
         QueryOp(
             parent=self.mw,
             op=lambda col: col.sched.get_or_create_filtered_deck(deck_id=deck_id),
-            success=self.load_deck_and_show,
+            success=lambda deck: (
+                self.load_deck_and_show(deck),  # type: ignore
+                self._after_load_deck_setup(deck_id),  # type: ignore
```
dae, line 68: You shouldn't need the ignore statements here - if you're getting an error about it not being able to infer the type, breaking the lambda out into a nested def should fix it.

qt/aqt/filtered_deck.py
```diff
@@ -100,6 +108,17 @@ def _initial_dialog_setup(self) -> None:
...
+    def _after_load_deck_setup(self, deck_id) -> None:
+        form = self.form
+        re_enable = self._re_enable_allow_empty_toggle
+        if deck_id != 0:
+            qconnect(form.search.textChanged, re_enable)
+            qconnect(form.order.currentTextChanged, re_enable)
+            qconnect(form.limit.valueChanged, re_enable)
+            qconnect(form.search_2.textChanged, re_enable)
+            qconnect(form.order_2.currentTextChanged, re_enable)
+            qconnect(form.limit_2.valueChanged, re_enable)
```
dae, line 120: This seems more complicated than necessary. Why not just show the option all the time then, and avoid the extra code that checks if the deck is new? It should also be changed to say "Create/update this deck even if ...".

Gustaf-C: I agree that it is too complicated, but since you asked for it to be hidden on old decks I thought I'd include it and let you make the call. I'll it change to always show and update the string.

qt/aqt/filtered_deck.py
Gustaf-C: I also found the error weird, but I had to do it this way to avoid the update running after the setup, which would always show the option.

Gustaf-C:
Update: Always show option, update string.


When running ./ninja check i get the following error, which fails the build, doesn't seem to impact the CI check though:

error: equal expressions as operands to `==`
  --> build/runner/src/bundle/artifacts.rs:56:8
   |
56 |     if env!("TARGET") == "x86_64-apple-darwin" {
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#eq_op
   = note: `#[deny(clippy::eq_op)]` on by default

ftl/core/decks.ftl
```diff
@@ -2,6 +2,7 @@ decks-add-new-deck-ctrlandn = Add New Deck (Ctrl+N)
 decks-build = Build
 decks-cards-selected-by = cards selected by
 decks-create-deck = Create Deck
+decks_create_even_if_empty = Create/update this deck even if it's empty
```
dae, line 5:
```suggestion
decks_create_even_if_empty = Create/update this deck even if empty
```

Slightly shorter description is more likely to fit on mobile screens

dae (commented):
Thanks @Gustaf-C, LGTM. I'm not sure what would cause that clippy error - do you have some version of Rust installed on a Linux system? If you're using rustup, clippy should follow the Rust version in ./rust-toolchain.toml

PR #2804, opened 2023-11-04
Allow im-/exporting with or without deck configs

### LINKED ISSUE #2777: Consider an option to export/import deck preset values
Currently imports and exports include/ignore deck configs based on whether scheduling is included or not. Some users (e.g. teachers) want the ability to export their custom deck settings without including their own scheduling.

RumovZ:
Closes #2777.


RumovZ:
Just realized that we only import new configs. Should I implement updating existing ones? Otherwise, we should probably at least mention that limitation in the help string.

dae:
Thanks Rumo, I'll circle back to this after 23.10.1 is out the door. Let's skip updating unless someone makes a strong case for it, as I imagine it will cause people to think Anki is at fault when they update a deck and suddenly they're getting things like a different number of new cards a day.

RumovZ:
New cards per day may already change, though, because the deck itself is getting updated (things like `NormalDeck.new_limit`).

dae:
Don't we handle that in prepare_deck()?

dae:
(though we still need to de-couple 'contains scheduling' and 'include scheduling' - can't remember if I previously pointed https://github.com/ankitects/anki/commit/ceb8a4ac2025bed600848b179ef989867fad1787 out to you)

dae:
Ah, just skimmed through your changes (hadn't looked at them yet) - I see you're now doing that based on include_config. Would it make sense to revert that part to being based on scheduling, and make a rule that deck authors can only adjust the limits in their template, and not on individual decks? We could think of the option as 'include preset' rather than 'include config' to make it clearer, and it would ensure we don't accidentally import the deck's current review count for the day if the exporter included scheduling and the importer didn't. It would also make it easier for importers to override the default daily limits if we implemented updating in the future.

RumovZ:
> revert that part to being based on scheduling, and make a rule that deck authors can only adjust the limits in their template, and not on individual decks

I was also considering that, but couldn't decide and opted for the (slightly) simpler alternative. I will change it.

> ensure we don't accidentally import the deck's current review count for the day if the exporter included scheduling and the importer didn't.

Funny you mention it. I see I've accidentally dropped the reset of `deck.common`. 😅

dae:
Thanks Rumo!

Issue #2808, opened 2023-11-05
Image occlusion buttons look odd in RTL mode


dae:
https://forums.ankiweb.net/t/anki-23-10-1-release-candidate/36631/4

dae:
Handled in #2809

Issue #2800, opened 2023-11-03
Image occlusion editor does not render SVGs


ghost:
- **anki version:** anki-23.10-mac-apple-qt6.dmg
- this is for anki's new image occlusion note type, not the image occlusion add-on

# reproduce
1. click add new note
2. select "Image Occlusion" note type
3. click "Select Image"
4. select a `.svg` image
5. image frame is presented in the image occlusion editor, but the image itself is not

# additional details
- png images render fine
- the image doesn't render in the edit window of search/browse either
- the image does render in preview or when learning/reviewing the note
- bug remains after disabling all add-ons

dae:
Fixed in #2809

Issue #2774, opened 2023-10-25
Various minor I/O issues


dae:
https://forums.ankiweb.net/t/anki-23-10-release-candidate/35967/53

The rotation issue has since been fixed (by removing it everywhere, as we don't store rotation data in the cloze)

hnvy:
Hello,

One more issue related to undo and redo (sorry, I couldn't edit my original post, so I have just placed the GIF here).

Harvey

![BH](https://github.com/ankitects/anki/assets/84394475/325f2bf3-0345-4307-8ec3-372248f207b7)

dae:
More issues here: https://forums.ankiweb.net/t/anki-23-10-release-candidate/35967/98

BEST8OY:
## Inaccessible shapes under grouped shapes
1- Create three shapes next to each other
2- Group the first and the last one
3- Now you can't access the second shape unless you move the grouped shapes first

Expertium:
Suggestion: when “Toggle Translucency” is enabled, creating another mask disables it, and moving an already existing mask also disables it. I think that’s inconvenient, and instead translucency should stay enabled until the user manually disables it.

krmanik:
I will tackle some of the I/O issues.

dae:
Very minor - resize markers blend into some backgrounds: https://forums.ankiweb.net/t/unable-to-see-boundary-marker-in-image-occlusion/36772

dae:
Another very minor one: Image field is not hidden until user selects an image.

dae:
A few of these have been fixed in #2809

BEST8OY:
https://forums.ankiweb.net/t/anki-23-12-beta/37771/28

krmanik:
I will further implement IO fixes later, may be end of the month.

abdnh:
@BEST8OY 
> https://forums.ankiweb.net/t/anki-23-12-beta/37771/29

Not clear to me what's the issue here?

> https://forums.ankiweb.net/t/image-occlusion-blue-edge-boxes-are-too-big-when-the-selection-is-small/37929

Except of this (which @krmanik said will tackle), the rest of the issues reported here all seem to have been fixed.

BEST8OY:
> Not clear to me what's the issue here?

Sorry for the very late respone.
I'm not sure either.
1- It's fixed already somehow by other PRs.
2- I mistook the whole thing and there was no bug to begin with!

krmanik:
There are issues and features request from users.

[...]

Issue #2770, opened 2023-10-23
Provide shortcut keys in I/O editor


dae:
Also, allow to delete occlusion blocks (what do you call the yellow rectangles that hides what is being tested?) by tapping the `Delete` button in the keyboard would be nice

_Originally posted by @BrayanDSO in https://github.com/ankitects/anki/issues/2768#issuecomment-1774163617_

BrayanDSO:
A nice shortcut would be to capture a image bitmap or image file pasted by pressing `Ctrl+V`

BEST8OY:
I would like to see some of these too if possible.
Ctrl+Z for undo
holding Alt and clicking on a box to delete that box.
Holding Ctrl and clicking to select multiple box because the select tool is not really flexible if you have a lot of boxes in an image

dae:
You can already select by shift+clicking. I think the delete key is going to be more obvious to users than using alt.

BrayanDSO:
Also, allow to move IO masks with the keyboard arrow keys

dae:
Some suggested shortcuts: https://forums.ankiweb.net/t/image-occlusion-enhanced-keybinds-image-occlusion/36610

BEST8OY:
> You can already select by shift+clicking. I think the delete key is going to be more obvious to users than using alt.

@dae 
In Windows holding Shift sometime produce a notice in windows which says you've pressed Shift key more than 5 times and it's really annoying in long run.
Please evaluate the possibility of switching to Ctrl+Clicking for selecting

dae:
Solved with #2809

PR #2809, opened 2023-11-06
fixes: remove unfinished shapes, remove selectable and make shapes remain inside canvas

### LINKED ISSUE #2588: Image occlusion doesn't scale when page zoomed in

krmanik:
IO fixes

#2588
#2770
#2774
#2800
#2808


krmanik:
Click through transparent area have following warning.
```
Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true. See: https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently
```
I have tried this for both upper and lower canvas, but it does not solve the warning so maybe the issues is in fabric.js.
```
canvas.getContext("2d", {willReadFrequently: true});
```

dae:
Thanks a lot for looking into these Mani! I'll give them a proper review in the next few days after 23.10.1 is out the door.

BEST8OY:
Not sure if I'm allowed to ask for this here, sorry in advance if I am.
A shortcut for "Toggle Masks" in reviewer would be handy.
Something like "M"

krmanik:
> Not sure if I'm allowed to ask for this here, sorry in advance if I'm not. A shortcut for "Toggle Masks" in reviewer would be handy. Something like "M"

Add key H for toggle mask because, M is already assigned to menu.

ts/image-occlusion/mask-editor.ts
```diff
@@ -120,8 +120,12 @@ function initCanvas(onChange: () => void): fabric.Canvas {
...
 }
 
-const getImageData = (imageData): string => {
+const getImageData = (imageData, path): string => {
     const b64encoded = protoBase64.enc(imageData);
+    const extension = path.split(".").pop();
+    if (extension === "svg") {
+        return "data:image/svg+xml;base64," + b64encoded;
+    }
     return "data:image/png;base64," + b64encoded;
```
dae, line 129: Ok, so we've been lying about the content type :-) While the browser tolerates this for things like .jpg images, we should probably be trying to return the correct content type for jpg/webp/gif as well.

ts/image-occlusion/review.ts
```diff
@@ -111,6 +111,12 @@ function setupImageOcclusionInner(setupOptions?: SetupImageOcclusionOptions): vo
         } else {
             button.style.display = "none";
         }
+
+        window.addEventListener("keydown", (event) => {
+            if (event.key === "h") {
+                toggleMasks();
```
dae, line 117: Some users use hjkl for reviewing, which this will break. Maybe we could use a capital M instead?

dae:
Sorry for the delay here Mani - I've just gotten around to testing this. There seems to have been a regression in the positioning of images in the editor. I open the editor, select an image, and the image is offset to the right, with the right side cut off:

2023-11-13T10:47:06,814411483+10:00

If I use the zoom tool, I can move the image so it appears correctly, but when I resize the window it moves back to the right.

It appears to have regressed in d44de202943272eaebb6650991cfd844a6746c9d.

Other notes:

- Zooming seems to work well in the editor, but I would say we haven't solved #2588 until the review screen supports zooming too. Minor: when making the window bigger, the mask borders get blurry.
- The shortcuts are a nice addition, but currently they conflict with the text tool - for example, try typing a 'p' into a text area
- Very minor: rectangles and ellipses are moved inside bounds after you let go of the mouse, whereas polygons and groups are never allowed to leave the bounds. If it's very easy to make the former behave like the latter, maybe it would be worth doing that to be consistent?
- Other fixes look good!

ts/image-occlusion/mask-editor.ts
krmanik: Yeah, I have added more mime type.

ts/image-occlusion/review.ts
krmanik: It can be done.

krmanik:
Fixing these issues added more issues, I will push the changes considering above suggested review in next commit latest by weekend. :-)

dae:
No problem Mani, please let me know when you're ready for another review.

dae:
Hi Mani,

There's one issue that I think we should probably solve before merging this in, as it will break the text tool for people using git otherwise. I see you've pushed https://github.com/ankitects/anki/pull/2809/commits/c68944454997664b273767095feba2de0dc2fef5, which has presumably fixed editing, but I hadn't actually noticed that issue. What I meant in my earlier comment is that when the text tool is used (below polygon), it's not possible to type text into it because the shortcuts are still active.

Re https://github.com/ankitects/anki/pull/2809/commits/5a2742766727a997f6631e7d9360408308c54263, we don't include the shortcuts inside translation strings - instead they are stored separately, so we can modify them depending on the platform, and change them without breaking translations. For example, the bold button:

```
    const keyCombination = "Control+B";
...
    <IconButton
        tooltip="{tr.editingBoldText()} ({getPlatformString(keyCombination)})"
```

krmanik:
I have updated the shortcut implementation as suggested.

dae:
Thanks Mani, I'll merge this in now and we can make any further tweaks in follow-up PRs.

One possible way to work around the RemoveFormatButton issue would be to only enable that shortcut when the IO mask editor is not visible. That said, I think users coming from the existing add-on may miss the existing shortcuts, that don't require the control key. Maybe an alternative solution would be to use the text:editing:entered and text:editing:exited events to set a editingTextElement flag, and only respond to those shortcut keys when the flag is not set?

krmanik:
I will match the keyboard shortcut and use the text event in next PR.

PR #2815, opened 2023-11-08
Change tooltip strings for image occlusion center alignments

snowtimeglass:
The current tooltips probably don't match the icons respectively.
  image
  image

This commit will:

- Fix the swapping of the following two tooltips: "Center Vertically" and "Center Horizontally".

- Incidentally, add a name to CONTRIBUTORS


dae:
I think you're right; I got confused because I see 'vertically' and think the vertical bar should match. How about we use the wording Figma uses to make it a little clearer?

https://help.figma.com/hc/en-us/articles/360039956914-Adjust-alignment-rotation-and-position

snowtimeglass:
>I got confused because I see 'vertically' and think the vertical bar should match.

Yes, I also feel the same confusion.

>How about we use the wording Figma uses to make it a little clearer?

I agree with the idea. I also think the wording Figma uses is clearer. 

Would you like me to change the following strings?
https://github.com/ankitects/anki/blob/d491622414bf03007d413c94c6c0cc1856bad5cd/ftl/core/editing.ftl#L84

https://github.com/ankitects/anki/blob/d491622414bf03007d413c94c6c0cc1856bad5cd/ftl/core/editing.ftl#L87

dae:
Yes please.

dae:
Thank you!

snowtimeglass:
Thank you, too!

PR #2820, opened 2023-11-09
Show due date of buried/suspended cards in card info

### LINKED ISSUE #2813: Show due date of buried/suspended cards in card info
https://forums.ankiweb.net/t/show-due-date-of-buried-and-suspended-cards-in-card-info/36770

I think the reason we didn't do this in the past was because we were going via the queue instead of the card type, but that should no longer be an issue with v2+ IIRC.

abdnh:
Closes #2813

I feel this can be simplified but not sure how.

### DIFF
--- a/rslib/src/stats/card.rs
+++ b/rslib/src/stats/card.rs
@@ -81,7 +81,7 @@ impl Collection {
             }
             CardType::Review | CardType::Learn | CardType::Relearn => (
                 {
-                    if card.due <= 1_000_000_000 {
+                    if due <= 1_000_000_000 {
                         let days_remaining = due - (self.timing_today()?.days_elapsed as i32);
                         let mut due = TimestampSecs::now();
                         due.0 += (days_remaining as i64) * 86_400;
--- a/rslib/src/stats/card.rs
+++ b/rslib/src/stats/card.rs
@@ -70,23 +70,34 @@ impl Collection {
         } else {
             card.due
         };
-        Ok(match card.queue {
-            CardQueue::New => (None, Some(due)),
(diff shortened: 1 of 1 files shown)


dae:
I wonder, could we just rely on ctype? We'd need to guess at whether it's an intra or interday learning card based on if  the value is over a certain number, like in restore_queue_from_type(), and would need to handle the new/preview case still.

rslib/src/stats/card.rs
```diff
@@ -70,27 +70,31 @@ impl Collection {
...
+            }
+            CardType::Review | CardType::Learn | CardType::Relearn => (
                 {
-                    if card.ctype == CardType::New {
-                        // new preview card not answered yet
-                        None
-                    } else {
+                    if card.ctype == CardType::Review
+                        || (matches!(card.ctype, CardType::Learn | CardType::Relearn)
+                            && card.due <= 1_000_000_000)
```
dae, line 86: We could probably just check card.due and omit the extra ctype checks?

rslib/src/stats/card.rs
```diff
@@ -70,27 +70,28 @@ impl Collection {
...
+                    (None, Some(due))
+                }
+            }
+            CardType::Review | CardType::Learn | CardType::Relearn => (
                 {
-                    if card.ctype == CardType::New {
-                        // new preview card not answered yet
-                        None
-                    } else {
+                    if card.due <= 1_000_000_000 {
```
dae, line 84:
```suggestion
                    if due <= 1_000_000_000 {
```

dae:
Thanks Abdo!

PR #2824, opened 2023-11-11
Allow applying limits of inactive parents

### LINKED ISSUE #2818: Consider providing a way to have parent decks impose a limit on reviews
Not sure what the best course of action is here - changing the current behavior will bring back the old problems for everyone, and is potentially disruptive. We could perhaps add another preset option that controls whether the clicked-on deck takes limits from the root or from itself. It's been a while since I looked into the counts traversal code, so we'll need to look into how simple or complicated it would be to add this in. @RumovZ any thoughts?

RumovZ:
Closes #2818.

Turned out to be relatively simple – or at least it seems that way before any bugs are uncovered. 🙃 Quite a few paths to test by now, but at least we don't have to worry about v2 anymore!

It should even be possible to implement this on the preset level. But that could get really confusing if you have different settings in a line of decks, so probably best to skip it.

There is one anomaly, though:
grafik
There is a review card in `a` and a new card in `a::b`. All limits are set to 1. The new card is not queued in `a` due to the *review* limit. However, it is queued in `a::b` because there is no review card to exhaust the limit.
I think this surprising behaviour is unavoidable as long as we allow applying parent limits and the review limit affecting new cards at the same time.


dae:
Thanks Rumo, it seems to work well. I've made some minor tweaks to the wording, but happy to discuss alternatives if you disagree.

PR #2825, opened 2023-11-11
fix: refresh model list when legacy addon adds notetype

kieranlblack:
When legacy addons add note types, they automatically add the note type to the collection models. When this was detected, the dialog box presenting the list of models was not being updated due to an early return in the code. This commit adds a list refresh along this path of execution to ensure the gui is updated with the newly added model.


dae:
Thanks!

PR #2827, opened 2023-11-13
Persist FSRS weights search in preset

### LINKED ISSUE #2814: Persist FSRS weights search in preset
https://forums.ankiweb.net/t/save-optimize-fsrs-weights-search-to-preset/36802

This is complicated by the fact that we need to change 'preset:foo' at editing time, as the user may rename their preset. We should have a blank entry by default, update the placeholder to show 'preset:foo', and replace the blank text with that text at save time. If the text is non-empty, then we'd use that instead.

abdnh:
Closes #2814

> and replace the blank text with that text at save time.

If we leave the blank text at save time, the default search will continue to work normally if the user renames their preset. So I wonder whether we should do this?

### DIFF
--- a/proto/anki/deck_config.proto
+++ b/proto/anki/deck_config.proto
@@ -154,6 +154,7 @@ message DeckConfig {
     float desired_retention = 37;
     bool reschedule_fsrs_cards = 39;
     float sm2_retention = 40;
+    string weight_search = 45;
 
     bytes other = 255;
   }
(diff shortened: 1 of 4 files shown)


dae:
Sorry for not being clear. That's what I meant - keeping the field blank, but replacing the blank field text with the actual search when we call computeFsrsWeights() (which we're already doing)

PR #2828, opened 2023-11-14
Fix filtered deck ordering of intraday learning cards and reviews

### LINKED ISSUE #2812: Filtered deck ordering surprising when a mix of intraday learning cards and reviews
https://forums.ankiweb.net/t/bug-order-due-filtered-deck-pulls-all-reviews-before-intraday-learning/36752

intraday cards use a timestamp like 1662096551; reviews used a days-since-collection-creation stamp like 130, leading to the reviews being shown first. Fixing this would require mapping the latter to the former prior to sorting.

abdnh:
Closes #2812

### DIFF
--- a/rslib/src/scheduler/filtered/mod.rs
+++ b/rslib/src/scheduler/filtered/mod.rs
@@ -121,7 +121,7 @@ impl Collection {
                 format!("({})", term.search)
             }
         );
-        let order = order_and_limit_for_search(term, ctx.today);
+        let order = order_and_limit_for_search(term, ctx.today, self.storage.creation_stamp()?.0);
 
         for mut card in self.all_cards_for_search_in_order(&search, SortMode::Custom(order))? {
             let original = card.clone();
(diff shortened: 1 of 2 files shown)


rslib/src/scheduler/filtered/mod.rs
```diff
@@ -121,7 +121,7 @@ impl Collection {
                 format!("({})", term.search)
             }
         );
-        let order = order_and_limit_for_search(term, ctx.today);
+        let order = order_and_limit_for_search(term, ctx.today, self.storage.creation_stamp()?.0);
```
abdnh, line 124: I believe this is only 100% accurate in V1, and we need to get this from the SchedTimingToday calculations in timing.rs instead?

dae: Yes, the current approach won't handle the rollover correctly. Please use .days_elapsed instead (eg (due -  {days_elapsed}) * 86400 + {current_timestamp})

dae:
Thanks Abdo!

PR #2829, opened 2023-11-15
Fix misalignment of IO masks in Qt5 when image is vertically long

hikaru-y:
An alternative to c478689.

- This fixes an issue in Qt5 where the IO masks are misaligned if the image is vertically long.
- Also, as in Qt6, the IO image is now constrained to fit in the viewport.


dae:
Thanks Hikaru! I like how this improves things for Qt5 while not risking regressions in Qt6.

PR #2831, opened 2023-11-16
Silently auto-convert non-string Inputs to strings in showText()

DerDemystifier:
The `showText()` function is explicitly typed to accept `txt` as a string (`str`). However, usage patterns indicate that users often pass non-string types, likely expecting automatic conversion due to the function's name and purpose. 

This tiny edit introduces an auto-conversion of the `txt` parameter to a string, enhancing the function's usability. It removes the necessity for manual string conversion by users, thereby ensuring handling of various input types and preventing potential runtime errors or unexpected behaviors.

### DIFF
--- a/qt/aqt/utils.py
+++ b/qt/aqt/utils.py
@@ -327,6 +327,7 @@ def showText(
     layout = QVBoxLayout(diag)
     diag.setLayout(layout)
     text: QPlainTextEdit | QTextBrowser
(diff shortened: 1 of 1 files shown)


dae:
I'm sorry, I think it's reasonable for showText() to expect to receive text. If you find yourself frequently needing to display non-text values, you could write your own helper function.

PR #2832, opened 2023-11-17
Clarify learning steps recommendation

user1823:
The previous text could suggest that 1d was OK, but it is actually not recommended.

I am not sure if I should change the name of the string too.


dae:
Personally, I think 1+ days already implies that 1 day isn't ok, so I don't think there's a need to make a new string which would break the existing translations, but I am happy to merge in the tweaked English wording.

PR #2833, opened 2023-11-17
Fix switching of add & history buttons when selecting IO notetype

### LINKED ISSUE #2826: I/O switches add and history buttons
https://forums.ankiweb.net/t/bug-switching-to-io-note-type-swaps-add-history-buttons/37015

akashreddy03:
Fixes #2826


dae:
Thanks Akash! This also has the nice side benefit of making the ctrl+enter shortcut combo work :-)

PR #2835, opened 2023-11-18
Expose ImportAnkiPackageOptions

abdnh:


dae:
I was about to add this before replying to a forum post, then realised you'd beat me to it :-)

abdnh:
I also added it after testing the new importer API while replying to https://forums.ankiweb.net/t/working-with-an-ankideck-using-python/37311 but didn't suggest the new API anyway after finding a workaround for the legacy importer.

PR #2836, opened 2023-11-18
 Load MathJax's text-chtml-full component in reviewer/previewer

### LINKED ISSUE #1773: Unify exposed MathJax add-ons
We appear to be exposing mathtools in the editor but not reviewer; we should be consistent.

https://forums.ankiweb.net/t/anki-2-1-50-beta-6-release-candidate/18181/99

abdnh:
Loading the tex-chtml-full component in the review/preview screens results in the same TeX extensions as in the editor being loaded (where tex-svg-full is used).

image

Closes #1773


dae:
I used the inspector to profile the speed difference, and it brings the ~85ms script load to ~100ms on my (fast) system, which is not ideal, but not terrible, either. Auto-loading might be a way to avoid the extra cost in the future, but this will at least move things in the right direction. Something else that might be looking into as a follow-up: the editor is using text-svg-full, and we're using text-chtml-full in the reviewer. Maybe we could switch the editor to chtml as well?

abdnh:
> the editor is using text-svg-full, and we're using text-chtml-full in the reviewer. Maybe we could switch the editor to chtml as well?

Looks like it's not very simple. One issue is that the CSS doesn't get loaded automatically.

dae:
Thanks for looking into it. It looks like the opposite - switching the reviewer to svg - would work, but as it may possibly cause other regressions, probably best to leave things as they are for now.

PR #2837, opened 2023-11-20
Add windows batch file to install n2

mikehardy:
the setup-ninja github action is unmaintained, n2 appears to work for all operating systems now, would be nice to have n2-installer for all operating systems for use in github actions

I copied the `@echo off` from the other batch files
I don't think there is anything needed with regards to working directory manipulation

Could be this easy?

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -150,6 +150,7 @@ virinci <github.com/virinci>
 snowtimeglass <snowtimeglass@gmail.com>
 Ben Olson <github.com/grepgrok>
 Akash Reddy <github.com/akashreddy03>
+Mike Hardy <github@mikehardy.net>
 
 ********************
 
--- a/tools/install-n2.bat
+++ b/tools/install-n2.bat
@@ -1,0 +1,3 @@
+@echo off
+
+cargo install --git https://github.com/ankitects/n2.git --rev 3b725cf9c321efb90496dd2458cf5c1abbef4dba


mikehardy:
Would be interesting to modify this (or whatever sets things up prior to this) to make use of it but I don't speak buildkite:

https://github.com/ankitects/anki/blob/main/.buildkite/windows/entrypoint.bat

dae:
Hi Mike,

Was there a reason the current recommended 'bash tools/install-n2' does not work for you? msys (and thus bash) is already a requirement on Windows, and if possible, I'd rather not have to update two files each time I need to update the version pin.

Anki's Mac+Windows CI environment doesn't get re-created on each run, so it's able to make use of a system-installed n2, but you could certainly run the install-n2 script as part of ankidroid-backend's workflow if you wanted to, and that's what our Linux environment currently does. It would add a few minutes to an uncached build though.

mikehardy:
That (just running the bash file on windows) is obvious but had not occurred to me, I was thinking to windows-y and forgot cross-platform stuff is possible.

Indeed it appears I can "bash" the thing in 3 different ways https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#shells

I'll play around with that and I'm certain I can get it to work

mikehardy:
final update, just doing `bash ./anki/tools/install-n2` worked fine - of course it did - so that install-n2 step is now unified / no os-specific matrix step and only one script here. Clear benefits from clear thinking, thanks

PR #100, opened 2015-03-14
Recognize a couple new media formats.

flimzy:
Adding 'webp' was necessary for Anki to recognize webp images
(https://developers.google.com/speed/webp/?csw=1) as images rather than
audio.  The addition of the two audio formats I prefer was just incidental,
as it appears Anki assumes audio files if a file is not an image file.

PR #2840, opened 2023-11-22
Fix sorting of due timestamps in the browser

abdnh:
https://forums.ankiweb.net/t/bug-report-about-due-sorting/37269

### DIFF
--- a/rslib/src/search/mod.rs
+++ b/rslib/src/search/mod.rs
@@ -363,7 +363,7 @@ fn card_order_from_sort_column(column: Column, timing: SchedTimingToday) -> Cow<
         )
         .into(),
         Column::Deck => "(select pos from sort_order where did = c.did) asc".into(),
-        Column::Due => format!("c.type asc, (case when c.due > 1000000000 then due else (due - {}) * 86400 + {} end) asc", timing.days_elapsed, TimestampSecs::now().0).into(),
+        Column::Due => format!("(case when c.due > 1000000000 then due else (due - {}) * 86400 + {} end) asc", timing.days_elapsed, TimestampSecs::now().0).into(),
         Column::Ease => format!("c.type = {} asc, c.factor asc", CardType::New as i8).into(),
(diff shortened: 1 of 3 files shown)


abdnh:
(just noticed note_due_order.sql will need fixing too)

rslib/src/search/mod.rs
```diff
@@ -363,7 +363,7 @@ fn card_order_from_sort_column(column: Column, timing: SchedTimingToday) -> Cow<
         )
         .into(),
         Column::Deck => "(select pos from sort_order where did = c.did) asc".into(),
-        Column::Due => "c.type asc, c.due asc".into(),
+        Column::Due => format!("c.type asc, (case when c.due > 1000000000 then due else (due - {}) * 86400 + {} end) asc", timing.days_elapsed, TimestampSecs::now().0).into(),
```
dae, line 366: The original motivation for sorting by c.type first was so that new cards and reviews don't end up getting mixed together if you have a card due on day 100 and a new card with #100. Because we're converting review dates to timestamps now, perhaps we can drop the initial type sorting here and below?

abdnh: Makes sense. One minor thing that might cause confusion is how new cards will get mixed with other types.
image

dae: Hmm, I didn't expect that - I thought that since the due dates are being converted to timestamps, they'd all come after the new numbers. Any idea why that's happening?

abdnh: We're treating new card positions as days and converting them like we do for review cards. The cards with due date 2025-07-22 in my screenshot have the same due value (2683).

abdnh: Adding a check to avoid converting new card positions should work. Will push a fix.

dae: Ah, that makes sense - I feel silly for having overlooked that :-) Thanks Abdo.

PR #2845, opened 2023-11-24
ignore filtered revlog in time of last review

L-M-Sherlock:
https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/260#issuecomment-1825102502

### DIFF
--- a/rslib/src/storage/revlog/time_of_last_review.sql
+++ b/rslib/src/storage/revlog/time_of_last_review.sql
@@ -2,6 +2,9 @@ SELECT id / 1000
 FROM revlog
 WHERE cid = $1
   AND ease BETWEEN 1 AND 4
-  AND (type != 3 OR factor != 0)
+  AND (
+    type != 3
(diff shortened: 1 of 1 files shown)


rslib/src/storage/revlog/time_of_last_review.sql
```diff
@@ -2,5 +2,9 @@ SELECT id / 1000
 FROM revlog
 WHERE cid = $1
   AND ease BETWEEN 1 AND 4
+  AND (
+    type != 3
+    OR factor != 0
+  )
```
dae, line 8: If the user reviews cards early, factor will be non-zero, so the revlog entry won't be ignored. That's fine, but I just wanted to check that's what you intended here.

L-M-Sherlock: Yeah. Reviews cards early should be included in the review history if the user doesn't disable `Reschedule cards based on my answers in this deck`.

PR #2847, opened 2023-11-24
fix: shortcut update and shape corner transparent

krmanik:
Fix:
- remove transparent corner after undo/redo
- shortcut update, match with IOE


dae:
Thanks Mani!

PR #2852, opened 2023-11-28
add DSR into special fields

L-M-Sherlock:
In the custom scheduling version of FSRS, it supports to display memory state in card's front.

image

This PR allows users to add DSR variables into their card template.

image

It would also allow users to display different fields under conditions of DSR.

But I don't know how to display them when FSRS is disabled. Help wanted.

### DIFF
--- a/rslib/src/notetype/mod.rs
+++ b/rslib/src/notetype/mod.rs
@@ -77,6 +77,9 @@ lazy_static! {
         "Subdeck",
         "Tags",
         "Type",
+        "Difficulty",
+        "Stability",
+        "Retrievability"
     ]);
 }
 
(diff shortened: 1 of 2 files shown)


dae:
I'm afraid I don't want to add any more special fields, as they require extra computation even if users don't use them. Until a feature comes in the future where the special fields can be lazily-calculating, you'll either need to rely on the card info screen, or implement this in an add-on: https://github.com/ankitects/anki-addons/blob/main/demos/field_filter/__init__.py

L-M-Sherlock:
> or implement this in an add-on: https://github.com/ankitects/anki-addons/blob/main/demos/field_filter/__init__.py

Thanks for this example. I will implement it in my helper add-on.

PR #2854, opened 2023-11-29
Replace css_browser_selector with our own implementation

### LINKED ISSUE #2846: Write our own css_browser_selector implemetation
css_browser_selector is an old, unmaintained library. It can't be easily included in our js bundles, and I've just discovered the CC-BY license is considered non-free by Debian (and perhaps other distros). As the functionality is fairly trivial, we should write our own implementation instead, without referencing the original source. We should only need to do some checks against the user agent, such as:

```js
(function() {
    const ua = navigator.userAgent.toLowerCase();

    function addBodyClass(className) {
        document.body.classList.add(className);
    }

    if (/ipad/.test(ua)) {
        addBodyClass('ipad');
    } else if (/iphone/.test(ua)) {
        addBodyClass('iphone');
    } else if (/android/.test(ua)) {
        addBodyClass('android');
    }

    if (/ipad|iphone|ipod/.test(ua)) {
        addBodyClass('ios');
    }
})();
```

We'll need to support the options listed on https://docs.ankiweb.net/templates/styling.html#platform-specific-css, and should probably support .chrome/.safari/.firefox as well.

@abdnh another one for you if you're interested

abdnh:
Closes #2846

The classes are actually added to the root element rather than body. css_browser_selector adds the following classes in my Windows device for example: `webkit chrome chrome112 win win8 js non-retina orientation_landscape maxw_1920`. Only `win` and `chrome` are supported by this basic implementation.

MDN reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent

### DIFF
--- a/ts/reviewer/browser_selector.ts
+++ b/ts/reviewer/browser_selector.ts
@@ -26,9 +26,7 @@ export function addBrowserClasses() {
 
     if (test(/ipad|iphone|ipod|android/)) {
         addClass("mobile");
-    }
-
-    if (test(/linux/)) {
+    } else if (test(/linux/)) {
         addClass("linux");
     } else if (test(/windows/)) {
         addClass("win");


dae:
Thanks Abdo! Always nice when we can drop a trivial dependency :-) I don't recall seeing any of those other classes in use in card templates, so there should hopefully be minimal impact to existing templates.

PR #2856, opened 2023-11-30
Prefer image pastes over HTML in the editor

abdnh:
Issue: https://forums.ankiweb.net/t/bug-some-images-cannot-be-pasted-into-anki-s-editor/37799

GoldenDict loads the clipboard with both HTML and image data.

Sidenote: The internal argument of editor_will_process_mime the hook is always False after the last changes to internal paste handling.

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -1490,16 +1490,9 @@ class EditorWebView(AnkiWebView):
         # print("urls", mime.urls())
         # print("text", mime.text())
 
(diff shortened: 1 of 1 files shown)


dae:
I was a bit concerned about unintended regressions similar to https://forums.ankiweb.net/t/powerpoint-copy-and-paste-test-as-text-not-image/4003, but hopefully this will be ok, and can't know if we don't try. Thanks Abdo!

PR #2857, opened 2023-11-30
Fix some RTL issues in help modals

abdnh:
https://forums.ankiweb.net/t/anki-23-12-beta/37771/17

`:dir()` is not yet available in stable Chrome/Chromium: https://developer.mozilla.org/en-US/docs/Web/CSS/:dir#browser_compatibility


dae:
Sorry, I missed that :dir was not yet supported. OK translation also appreciated - thanks Abdo!

PR #2859, opened 2023-12-01
Do not add a fallback field reference if front side contains conditionals

abdnh:
Issue: https://forums.ankiweb.net/t/anki-23-12-beta/37771/22


dae:
Thanks!

PR #2860, opened 2023-12-01
Fix auto advance state not being restored after focus change in some cases

abdnh:
Probably fixes https://forums.ankiweb.net/t/anki-23-12-beta/37771/23

Using the FocusOut event proved to be problematic, because it's also received when the reviewer's context menu is opened but the focusChanged signal was not emitted in this case, resulting in auto advance staying disabled.


dae:
Thanks Abdo! I've also logged the TTS issue on https://github.com/ankitects/anki/issues/2868

PR #2861, opened 2023-12-02
Fix image overlay not closing when clicking outside on Qt6.6

hikaru-y:
https://forums.ankiweb.net/t/anki-23-12-beta/37771/32


dae:
Thanks Hikaru!

PR #2862, opened 2023-12-02
Include elapsed_secs in learning card state

### LINKED ISSUE #1557: including elapsed_secs in learning card state
Below is a start on this, but it has some problems that will need fixing/thought:

- elapsed calculation does not take into account cards shown during the learn ahead limit
- elapsed calcualtion does not handle interday learning cards, and assumes that due is a timestamp. In the interday learning case, we may need to synthesize a value based on self.timing.next_day_at, after checking self.card.queue.

Original report by @ijgnd on https://forums.ankiweb.net/t/v3-scheduler-custom-js-scheduling-include-elapsed-seconds-for-learning-cards/15845

```diff
diff --git a/proto/anki/scheduler.proto b/proto/anki/scheduler.proto
index 7876e3554..e8756d061 100644
--- a/proto/anki/scheduler.proto
+++ b/proto/anki/scheduler.proto
@@ -46,6 +46,7 @@ message SchedulingState {
   message Learning {
     uint32 remaining_steps = 1;
     uint32 scheduled_secs = 2;
+    uint32 elapsed_secs = 3;
   }
   message Review {
     uint32 scheduled_days = 1;
diff --git a/rslib/src/backend/scheduler/states/learning.rs b/rslib/src/backend/scheduler/states/learning.rs
index 55f56e3d1..7e9d392e8 100644
--- a/rslib/src/backend/scheduler/states/learning.rs
+++ b/rslib/src/backend/scheduler/states/learning.rs
@@ -8,6 +8,7 @@ impl From<pb::scheduling_state::Learning> for LearnState {
         LearnState {
             remaining_steps: state.remaining_steps,
             scheduled_secs: state.scheduled_secs,
+            elapsed_secs: state.elapsed_secs,
         }
     }
 }
@@ -17,6 +18,7 @@ impl From<LearnState> for pb::scheduling_state::Learning {
         pb::scheduling_state::Learning {
             remaining_steps: state.remaining_steps,
             scheduled_secs: state.scheduled_secs,
+            elapsed_secs: state.elapsed_secs,
         }
     }
 }
diff --git a/rslib/src/scheduler/answering/current.rs b/rslib/src/scheduler/answering/current.rs
index 9db8601f8..d760ee8b2 100644
--- a/rslib/src/scheduler/answering/current.rs
+++ b/rslib/src/scheduler/answering/current.rs
@@ -62,13 +62,20 @@ impl CardStateUpdater {
         let ease_factor = self.card.ease_factor();
         let remaining_steps = self.card.remaining_steps();
 
+        let elapsed_secs = |scheduled_secs: u32| -> u32 {
+            let scheduled_at = due as i64 - scheduled_secs as i64;
+            (self.timing.now.0 - scheduled_at).max(0) as u32
+        };
+
         match self.card.ctype {
             CardType::New => NormalState::New(NewState {
                 position: due.max(0) as u32,
             }),
             CardType::Learn => {
+                let scheduled_secs = self.learn_steps().current_delay_secs(remaining_steps);
                 LearnState {
-                    scheduled_secs: self.learn_steps().current_delay_secs(remaining_steps),
+                    scheduled_secs,
+                    elapsed_secs: elapsed_secs(scheduled_secs),
                     remaining_steps,
                 }
             }
@@ -82,18 +89,22 @@ impl CardStateUpdater {
                 leeched: false,
             }
             .into(),
-            CardType::Relearn => RelearnState {
-                learning: LearnState {
-                    scheduled_secs: self.relearn_steps().current_delay_secs(remaining_steps),
-                    remaining_steps,
-                },
-                review: ReviewState {
-                    scheduled_days: interval,
-                    elapsed_days: interval,
-                    ease_factor,
-                    lapses,
-                    leeched: false,
-                },
+            CardType::Relearn => {
+                let scheduled_secs = self.relearn_steps().current_delay_secs(remaining_steps);
+                RelearnState {
+                    learning: LearnState {
+                        scheduled_secs,
+                        elapsed_secs: elapsed_secs(scheduled_secs),
+                        remaining_steps,
+                    },
+                    review: ReviewState {
+                        scheduled_days: interval,
+                        elapsed_days: interval,
+                        ease_factor,
+                        lapses,
+                        leeched: false,
+                    },
+                }
             }
             .into(),
         }
diff --git a/rslib/src/scheduler/answering/mod.rs b/rslib/src/scheduler/answering/mod.rs
index 1bce3ffef..851be883c 100644
--- a/rslib/src/scheduler/answering/mod.rs
+++ b/rslib/src/scheduler/answering/mod.rs
@@ -9,7 +9,6 @@ mod review;
 mod revlog;
 
 use rand::{prelude::*, rngs::StdRng};
-
 use revlog::RevlogEntryPartial;
 
 use super::{
@@ -53,6 +52,7 @@ struct CardStateUpdater {
     timing: SchedTimingToday,
     now: TimestampSecs,
     fuzz_seed: Option<u64>,
+    learn_ahead_secs: u32,
 }
 
 impl CardStateUpdater {
@@ -337,6 +337,7 @@ impl Collection {
             .get_deck(card.deck_id)?
             .ok_or(AnkiError::NotFound)?;
         let config = self.home_deck_config(deck.config_id(), card.original_deck_id)?;
+        let learn_ahead_secs = self.learn_ahead_secs();
         Ok(CardStateUpdater {
             fuzz_seed: get_fuzz_seed(&card),
             card,
@@ -344,6 +345,7 @@ impl Collection {
             config,
             timing,
             now: TimestampSecs::now(),
+            learn_ahead_secs,
         })
     }
 
diff --git a/rslib/src/scheduler/states/learning.rs b/rslib/src/scheduler/states/learning.rs
index 57223bc36..579712268 100644
--- a/rslib/src/scheduler/states/learning.rs
+++ b/rslib/src/scheduler/states/learning.rs
@@ -8,6 +8,7 @@ use crate::revlog::RevlogReviewKind;
 pub struct LearnState {
     pub remaining_steps: u32,
     pub scheduled_secs: u32,
+    pub elapsed_secs: u32,
 }
 
 impl LearnState {
@@ -33,6 +34,7 @@ impl LearnState {
         LearnState {
             remaining_steps: ctx.steps.remaining_for_failed(),
             scheduled_secs: ctx.steps.again_delay_secs_learn(),
+            elapsed_secs: 0,
         }
     }
 
@@ -40,6 +42,7 @@ impl LearnState {
         if let Some(hard_delay) = ctx.steps.hard_delay_secs(self.remaining_steps) {
             LearnState {
                 scheduled_secs: hard_delay,
+                elapsed_secs: 0,
                 ..self
             }
             .into()
@@ -59,6 +62,7 @@ impl LearnState {
             LearnState {
                 remaining_steps: ctx.steps.remaining_for_good(self.remaining_steps),
                 scheduled_secs: good_delay,
+                elapsed_secs: 0,
             }
             .into()
         } else {
diff --git a/rslib/src/scheduler/states/normal.rs b/rslib/src/scheduler/states/normal.rs
index 7043b24dc..f05264359 100644
--- a/rslib/src/scheduler/states/normal.rs
+++ b/rslib/src/scheduler/states/normal.rs
@@ -40,6 +40,7 @@ impl NormalState {
                 // New state acts like answering a failed learning card
                 let next_states = LearnState {
                     remaining_steps: ctx.steps.remaining_for_failed(),
+                    elapsed_secs: 0,
                     scheduled_secs: 0,
                 }
                 .next_states(ctx);
diff --git a/rslib/src/scheduler/states/relearning.rs b/rslib/src/scheduler/states/relearning.rs
index a295717db..42156ff9c 100644
--- a/rslib/src/scheduler/states/relearning.rs
+++ b/rslib/src/scheduler/states/relearning.rs
@@ -37,6 +37,7 @@ impl RelearnState {
                 learning: LearnState {
                     remaining_steps: ctx.relearn_steps.remaining_for_failed(),
                     scheduled_secs: again_delay,
+                    elapsed_secs: 0,
                 },
                 review: ReviewState {
                     scheduled_days: self.review.failing_review_interval(ctx),
@@ -79,6 +80,7 @@ impl RelearnState {
             RelearnState {
                 learning: LearnState {
                     scheduled_secs: good_delay,
+                    elapsed_secs: 0,
                     remaining_steps: ctx
                         .relearn_steps
                         .remaining_for_good(self.learning.remaining_steps),
diff --git a/rslib/src/scheduler/states/review.rs b/rslib/src/scheduler/states/review.rs
index 11dad02c6..93c3843c3 100644
--- a/rslib/src/scheduler/states/review.rs
+++ b/rslib/src/scheduler/states/review.rs
@@ -86,6 +86,7 @@ impl ReviewState {
                 learning: LearnState {
                     remaining_steps: ctx.relearn_steps.remaining_for_failed(),
                     scheduled_secs: again_delay,
+                    elapsed_secs: 0,
                 },
                 review: again_review,
             }
```

Gustaf-C:
Closes #1557.

I decided to implement this where `elapsed_time` has a positive value in seconds for same day learning cards, and a negative value in days for interday learning cards. The other option I considered was to set it as the elapsed days in seconds for interday learning cards, but I figured that could give some weird results when cards that are "usually" same day learning cards are reviewed close to the rollover and get scheduled for the next day.

Since the calculation needs the current time, the pre-calculated state will differ from the answered state if a second passes between when a card is shown and when it is answered. To get around this simply set the `elapsed_times` to be equal for the two states, which feels a bit janky but should be fine since it shouldn't be needed after the next states have been calculated. Correct me if I'm wrong here.

The tests work but I am not completely satisfied with how they are done since they change the due value to trick the `elapsed_time` calculation, and they have a potential fail point where the tests succeeding depends on them not taking too long to run so the timestamp is unchanged. I have given it a one second leeway now which should avoid bad luck since the time sensitive part only takes ~3ms even on my old system, but if the tests were to slow down considerably they would fail due to this. However I don't know how I would fake the system clock for a proper test without these issues, but maybe they are acceptable as is?

### DIFF
--- a/rslib/src/scheduler/answering/current.rs
+++ b/rslib/src/scheduler/answering/current.rs
@@ -1,11 +1,11 @@
 // Copyright: Ankitects Pty Ltd and contributors
 // License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
+use super::get_fuzz_seed_for_id_and_reps;
 use super::CardStateUpdater;
 use crate::card::CardQueue;
 use crate::card::CardType;
 use crate::decks::DeckKind;
-use crate::scheduler::answering::get_fuzz_seed;
 use crate::scheduler::states::CardState;
 use crate::scheduler::states::LearnState;
 use crate::scheduler::states::NewState;
@@ -66,24 +66,25 @@ impl CardStateUpdater {
         let ease_factor = self.card.ease_factor();
         let remaining_steps = self.card.remaining_steps();
         let memory_state = self.card.memory_state;
-        let elapsed_time = |last_ivl: u32| {
+        let elapsed_secs = |last_ivl: u32| {
             match self.card.queue {
                 CardQueue::Learn => {
                     // Decrease reps by 1 to get correct seed for fuzz.
                     // If the fuzz calculation changes, this will break.
-                    let mut card_copy = self.card.clone();
-                    card_copy.reps -= 1;
-                    let last_ivl_with_fuzz =
-                        self.learning_ivl_with_fuzz(get_fuzz_seed(&card_copy), last_ivl);
+                    let last_ivl_with_fuzz = self.learning_ivl_with_fuzz(
+                        get_fuzz_seed_for_id_and_reps(self.card.id, self.card.reps - 1),
+                        last_ivl,
+                    );
                     let last_answered_time = due as i64 - last_ivl_with_fuzz as i64;
-                    (self.now.0 - last_answered_time) as i32
+                    (self.now.0 - last_answered_time) as u32
                 }
                 CardQueue::DayLearn => {
                     let days_since_col_creation = self.timing.days_elapsed as i32;
                     // Need .max(1) for same day learning cards pushed to the next day.
-                    let last_ivl_as_days = (last_ivl / 86400).max(1) as i32;
+                    // 86_400 is the number of seconds in a day.
+                    let last_ivl_as_days = (last_ivl / 86_400).max(1) as i32;
                     let elapsed_days = days_since_col_creation - due + last_ivl_as_days;
-                    -elapsed_days
+                    (elapsed_days * 86_400) as u32
                 }
                 _ => 0, // Not used for other card queues.
             }
@@ -98,7 +99,7 @@ impl CardStateUpdater {
                 LearnState {
                     scheduled_secs: last_ivl,
                     remaining_steps,
-                    elapsed_time: elapsed_time(last_ivl),
+                    elapsed_secs: elapsed_secs(last_ivl),
                     memory_state,
                 }
             }
@@ -118,7 +119,7 @@ impl CardStateUpdater {
                 RelearnState {
                     learning: LearnState {
                         scheduled_secs: last_ivl,
-                        elapsed_time: elapsed_time(last_ivl),
+                        elapsed_secs: elapsed_secs(last_ivl),
                         remaining_steps,
                         memory_state,
                     },
--- a/rslib/src/scheduler/answering/current.rs
+++ b/rslib/src/scheduler/answering/current.rs
@@ -2,8 +2,10 @@
 // License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
 use super::CardStateUpdater;
+use crate::card::CardQueue;
 use crate::card::CardType;
 use crate::decks::DeckKind;
+use crate::scheduler::answering::get_fuzz_seed;
 use crate::scheduler::states::CardState;
 use crate::scheduler::states::LearnState;
 use crate::scheduler::states::NewState;
@@ -64,15 +66,39 @@ impl CardStateUpdater {
         let ease_factor = self.card.ease_factor();
         let remaining_steps = self.card.remaining_steps();
         let memory_state = self.card.memory_state;
+        let elapsed_time = |last_ivl: u32| {
+            match self.card.queue {
+                CardQueue::Learn => {
+                    // Decrease reps by 1 to get correct seed for fuzz.
+                    // If the fuzz calculation changes, this will break.
+                    let mut card_copy = self.card.clone();
+                    card_copy.reps -= 1;
+                    let last_ivl_with_fuzz =
+                        self.learning_ivl_with_fuzz(get_fuzz_seed(&card_copy), last_ivl);
+                    let last_answered_time = due as i64 - last_ivl_with_fuzz as i64;
+                    (self.now.0 - last_answered_time) as i32
+                }
+                CardQueue::DayLearn => {
+                    let days_since_col_creation = self.timing.days_elapsed as i32;
+                    // Need .max(1) for same day learning cards pushed to the next day.
+                    let last_ivl_as_days = (last_ivl / 86400).max(1) as i32;
+                    let elapsed_days = days_since_col_creation - due + last_ivl_as_days;
+                    -elapsed_days
+                }
+                _ => 0, // Not used for other card queues.
+            }
+        };
 
         match self.card.ctype {
             CardType::New => NormalState::New(NewState {
                 position: due.max(0) as u32,
             }),
             CardType::Learn => {
+                let last_ivl = self.learn_steps().current_delay_secs(remaining_steps);
                 LearnState {
-                    scheduled_secs: self.learn_steps().current_delay_secs(remaining_steps),
+                    scheduled_secs: last_ivl,
                     remaining_steps,
+                    elapsed_time: elapsed_time(last_ivl),
                     memory_state,
                 }
             }
@@ -87,20 +113,24 @@ impl CardStateUpdater {
                 memory_state,
             }
             .into(),
-            CardType::Relearn => RelearnState {
-                learning: LearnState {
-                    scheduled_secs: self.relearn_steps().current_delay_secs(remaining_steps),
-                    remaining_steps,
-                    memory_state,
-                },
-                review: ReviewState {
-                    scheduled_days: interval,
-                    elapsed_days: interval,
-                    ease_factor,
-                    lapses,
-                    leeched: false,
-                    memory_state,
-                },
+            CardType::Relearn => {
+                let last_ivl = self.relearn_steps().current_delay_secs(remaining_steps);
+                RelearnState {
+                    learning: LearnState {
+                        scheduled_secs: last_ivl,
+                        elapsed_time: elapsed_time(last_ivl),
+                        remaining_steps,
+                        memory_state,
+                    },
+                    review: ReviewState {
+                        scheduled_days: interval,
+                        elapsed_days: interval,
+                        ease_factor,
+                        lapses,
+                        leeched: false,
+                        memory_state,
+                    },
+                }
             }
             .into(),
         }
--- a/rslib/src/scheduler/answering/mod.rs
+++ b/rslib/src/scheduler/answering/mod.rs
@@ -269,12 +269,17 @@ impl Collection {
 
         let mut updater = self.card_state_updater(card)?;
         answer.cap_answer_secs(updater.config.inner.cap_answer_time_to_secs);
-        let current_state = updater.current_card_state();
-        require!(
-            current_state == answer.current_state,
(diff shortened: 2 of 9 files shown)


rslib/src/scheduler/answering/current.rs
```diff
@@ -64,15 +66,39 @@ impl CardStateUpdater {
...
         let memory_state = self.card.memory_state;
+        let elapsed_time = |last_ivl: u32| {
+            match self.card.queue {
+                CardQueue::Learn => {
+                    // Decrease reps by 1 to get correct seed for fuzz.
+                    // If the fuzz calculation changes, this will break.
+                    let mut card_copy = self.card.clone();
+                    card_copy.reps -= 1;
+                    let last_ivl_with_fuzz =
+                        self.learning_ivl_with_fuzz(get_fuzz_seed(&card_copy), last_ivl);
```
dae, line 77: It would be nice to factor this out into get_fuzz_seed_for_id_and_reps, so we don't need to clone the card.

rslib/src/scheduler/answering/mod.rs
```diff
@@ -269,12 +269,17 @@ impl Collection {
...
-        require!(
-            current_state == answer.current_state,
-            "card was modified: {current_state:#?} {:#?}",
-            answer.current_state,
-        );
+        let mut current_state = updater.current_card_state();
+        // If the states aren't equal, it's probably because some time has passed.
+        // Try to fix this by setting elapsed_time equal.
+        if answer.current_state != current_state {
+            self.set_elapsed_time_equal(answer, &mut current_state);
```
dae, line 276: We probably don't need the outer check, since set_elapsed_time_equal() will only apply to (re)learning cards.

rslib/src/scheduler/answering/mod.rs
```diff
@@ -417,6 +422,33 @@ impl Collection {
         self.add_tags_to_notes_inner(&[nid], "leech")?;
         Ok(())
     }
+
+    fn set_elapsed_time_equal(&self, answer: &mut CardAnswer, current_state: &mut CardState) {
```
dae, line 426: Since this only changes `current`, answer should not be mutable. Also, a brief docstring explaining the reason for this function would be good.

rslib/src/scheduler/answering/mod.rs
```diff
@@ -670,4 +702,85 @@ mod test {
...
+        let now = TimestampSecs::now();
+        let timing = col.timing_for_timestamp(now)?;
+        let col_age = timing.days_elapsed as i32;
+        let shift_due_time = col_age - 3; // Three days late
+        elapsed_time_helper(&mut col, shift_due_time, post_answer, expected_elapsed_time)?;
+
+        Ok(())
+    }
+
+    fn elapsed_time_helper(
```
dae, line 758: A name that better describes what it does (e.g. assert_elapsed_times_are_approx_equal) would be clearer.

dae:
I'm not too fond of +seconds/-days I'm afraid:

- While there's some precedent in the way revlogs are handled, they are actually -seconds/+days
- scheduled_secs is always seconds, and it feels like it might be better to be consistent
- keeping a single unit makes things a little less complicated for people using these data structures

> To get around this simply set the elapsed_times to be equal for the two states, which feels a bit janky but should be fine since it shouldn't be needed after the next states have been calculated.

Yeah, I think that's fine.

> The tests work but I am not completely satisfied with how they are done since they change the due value to trick the elapsed_time calculation

You could probably simplify this by creating/modifying a card in memory, and having the tests compare output from .normal_study_state(), instead of using the whole card answering flow.

Gustaf-C:
Thanks for the review.

About the representation of the elapsed_time, I agree that just using one unit is easier to handle. Assuming we go with my second option instead, i.e. elapsed_days * (seconds in a day), the situation I think might get misleading is the following:

- Study card 1 15 minutes before rollover, scheduled for 5 minutes before rollover
- Study card 2 9 minutes before rollover, scheduled for 1 minute after rollover -> gets pushed to DayLearn

If we then study both cards some time after the rollover, let's say 3 hours it will show:
- Card 1 elapsed_time: ~3 hours (in seconds): 600 secs
- Card 2 elapsed_time: 24 hours (in seconds), scheduled_secs: 600 secs

Despite card 1 having a longer "actual" elapsed_time, what is shown will be the opposite and there is (afaik) no way to determine that the elapsed time for card 2 isn't actually exact through the custom scheduling interface.

Now admittedly this is a bit of an edge case, and I don't know if it would ever really matter. I'm just a bit weary of discarding that information without thinking about it. One possible option I see to address this is to include a bool, maybe `interday_learn`, to show that the card was an interday learning card. That would give both benefits, only one unit to think about but the information is retained if someone would like to use it.

rslib/src/scheduler/answering/mod.rs
Gustaf-C: I figured it might save a few instructions depending on how the compiler figures it out, but yeah it's probably unnecessarcy.

rslib/src/scheduler/answering/mod.rs
Gustaf-C: It simplified the function somewhat but I know that it's bad form. I'll switch it around so we update the answer_state to match the current_state instead of the other way around. That also eliminates the need to mark the current state as mutable.

dae:
I think there'll always be corner cases around the rollover time, which is why we recommend users set it to a time when they're likely to be in the middle of a sleep.

I think I might be missing something re: a separate interday_learn flag. Can't it be inferred based on any elapsed_secs >= 86400?

rslib/src/scheduler/states/learning.rs
```diff
@@ -13,6 +13,9 @@ use crate::revlog::RevlogReviewKind;
 pub struct LearnState {
     pub remaining_steps: u32,
     pub scheduled_secs: u32,
+    // Positive: Same day learning, elapsed time in seconds
+    // Negative: Interday learning, elapsed time in days
+    pub elapsed_secs: u32,
```
dae, line 18: Stale comment

dae (commented):
Thanks Gustaf, it looks good. I'm not too concerned about the day rollover issue - I suspect it won't really matter in practice. Just one minor issue I noticed:

Gustaf-C:
Yeah it as a real edge case and the actual impact is probably small.

dae:
Thanks!

PR #2863, opened 2023-12-02
Ignore rescheduled cards in old stats page

user1823:
Fixes https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/320

Rescheduled cards are not actual reviews. So, including them in Cards Studied Today is inappropriate.

### DIFF
--- a/pylib/anki/stats.py
+++ b/pylib/anki/stats.py
@@ -746,9 +746,9 @@ select count(), avg(ivl), max(ivl) from cards where did in %s and queue = {QUEUE
                 "id > %d" % ((self.col.sched.day_cutoff - (days * 86400)) * 1000)
             )
         if lims:
-            lim = "where type != {REVLOG_RESCHED} and " + " and ".join(lims)
+            lim = "and " + " and ".join(lims)
         else:
-            lim = "where type != {REVLOG_RESCHED}"
+            lim = ""
         ease4repl = "ease"
         return self.col.db.all(
             f"""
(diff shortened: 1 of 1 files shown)


user1823:
I am not sure what's the issue with the last commit. Help would be appreciated.

pylib/anki/stats.py
```diff
@@ -746,9 +746,9 @@ def _eases(self) -> Any:
                 "id > %d" % ((self.col.sched.day_cutoff - (days * 86400)) * 1000)
             )
         if lims:
-            lim = "where " + " and ".join(lims)
+            lim = "where type != {REVLOG_RESCHED} and " + " and ".join(lims)
```
Gustaf-C, line 749: These should be f-strings:
```suggestion
            lim = f"where type != {REVLOG_RESCHED} and " + " and ".join(lims)
```

pylib/anki/stats.py
```diff
@@ -746,9 +746,9 @@ def _eases(self) -> Any:
...
         else:
-            lim = ""
+            lim = "where type != {REVLOG_RESCHED}"
```
Gustaf-C, line 751:
```suggestion
            lim = f"where type != {REVLOG_RESCHED}"
```

L-M-Sherlock, line 751: It should be f"where type != {REVLOG_RESCHED}"

pylib/anki/stats.py
L-M-Sherlock, line 749: It should be f"where type != {REVLOG_RESCHED} and "

L-M-Sherlock (changes requested):

dae:
Thank you both.

PR #2869, opened 2023-12-04
Fix Mathjax can only be pasted once

### LINKED ISSUE #2691: Mathjax can be pasted at most once
Running on f3bb845961973bcfab34acfdc4d314294285ee74, no add-on

How to reproduce:
* Open the new card window
* create a mathjax entry, let's say $a$
* select the mathjax element
* click somewhere in some field so not to select anything
* paste twice (ctrl+v)

Expected result
* see two mathjax entries $a$

Actual result
* the second html of the second paste is `Mathjax`

akashreddy03:
Fixes #2691  

When pasting Mathjax without selecting any text before and after, it creates an extra trailing space (empty FrameElement) which hinders consequent pastes and just keeps on adding space instead of the content. This PR removes that extra FrameElement after the paste.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -149,7 +149,7 @@ Gustaf Carefall <https://github.com/Gustaf-C>
 virinci <github.com/virinci>
 snowtimeglass <snowtimeglass@gmail.com>
 Ben Olson <github.com/grepgrok>
-Akash Reddy <github.com/akashreddy03>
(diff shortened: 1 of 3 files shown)


dae:
Thanks for looking into this, Akash!

Arthur-Milchior:
Thank you very much @akashreddy03. REALLY appreciate this fix!

PR #2870, opened 2023-12-04
skip revlog with time = 0 when calc recall cost

L-M-Sherlock:
https://github.com/open-spaced-repetition/fsrs-rs/issues/126

### DIFF
--- a/rslib/src/scheduler/fsrs/retention.rs
+++ b/rslib/src/scheduler/fsrs/retention.rs
@@ -149,8 +149,9 @@ impl Collection {
                         && r.taken_millis > 0
                 })
                 .map(|r| r.taken_millis);
-            if let Ok(length) = revlogs_filter.try_len() {
-                revlogs_filter.sum::<u32>() as f64 / length as f64 / 1000.0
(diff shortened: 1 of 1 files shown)


rslib/src/scheduler/fsrs/retention.rs
```diff
@@ -139,10 +143,14 @@ impl Collection {
...
-                .filter(|r| r.review_kind == RevlogReviewKind::Learning && r.button_chosen >= 1)
+                .filter(|r| {
+                    r.review_kind == RevlogReviewKind::Learning
+                        && r.button_chosen >= 1
+                        && r.taken_millis > 0
+                })
                 .map(|r| r.taken_millis);
-            if total_first > 0.0 {
-                revlogs_filter.sum::<u32>() as f64 / total_first / 1000.0
+            if let Ok(length) = revlogs_filter.try_len() {
```
dae, line 152: I don't think this will work, as Rust won't know the length without executing the iterator. You'll need revlogs_filter.clone().count().

dae:
Thanks!

PR #2873, opened 2023-12-05
Fix/keep two decimals for stability

L-M-Sherlock:
Reason: some users reported that the rescheduling results are inconsistent between built-in FSRS and the helper add-on. For example, if the memory stability is 8.48, the built-in FSRS will schedule it with 8 days interval. However, it is stored as 8.5 in database. The helper will round it to 9 days. It would also induce the inconsistency during the fuzzing.

### DIFF
--- a/rslib/src/storage/card/data.rs
+++ b/rslib/src/storage/card/data.rs
@@ -163,7 +163,7 @@ mod test {
         };
         assert_eq!(
             data.convert_to_json().unwrap(),
-            r#"{"s":123.5,"d":1.235,"dr":0.99}"#
+            r#"{"s":123.46,"d":1.235,"dr":0.99}"#
         );
     }
 }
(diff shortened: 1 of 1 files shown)


L-M-Sherlock:
@dae, I find that the `round_to_places` couldn't solve the problem completely. For example, if the stability is `20.49875`, `round_to_places(s, 2)` is 20.5, which still causes the inconsistency.

dae:
Can't you adjust the helper to use the same precision/rounding as Anki instead?

L-M-Sherlock:
> Can't you adjust the helper to use the same precision/rounding as Anki instead?

Helper reads the value from database. But the value has been rounded before it is stored in database.

L-M-Sherlock:
For example, if the stability is 20.49875 calculated by built-in FSRS, it will be stored as 20.5. Then, the stability is incorrect in the card info because 20.5 is rounded to 21.

image

dae:
Ok, I see. We could potentially solve this by rounding somewhere in fsrs-rs, but since we'll hopefully implement the helper's functions natively in Rust in the future, perhaps it would be more efficient to document this limitation, and wait until then?

L-M-Sherlock:
I'm OK. It doesn't affect too many cards.

PR #2874, opened 2023-12-05
Fix TTS not working properly with auto advance

### LINKED ISSUE #2868: auto advance and tts conflict?
https://forums.ankiweb.net/t/anki-23-12-beta/37771/60

abdnh:
Closes #2868

This was ultimately caused by av_player_did_end_playing being called twice for TTS tags. I can't say I understand how it exactly resulted in that weird behavior.

### DIFF
--- a/qt/aqt/tts.py
+++ b/qt/aqt/tts.py
@@ -257,10 +257,8 @@ class MacTTSFilePlayer(MacTTSPlayer):
         # inject file into the top of the audio queue
         from aqt.sound import av_player
 
+        av_player.current_player = None
         av_player.insert_file(self.tmppath)
-
-        # then tell player to advance, which will cause the file to be played
-        cb()
 
 
 # Windows support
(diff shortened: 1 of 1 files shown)


qt/aqt/tts.py
```diff
@@ -606,13 +604,10 @@ def _on_done(self, ret: Future, cb: OnDoneCallback) -> None:
             if exception := ret.exception():
                 print(str(exception))
                 tooltip(tr.errors_windows_tts_runtime_error())
-                cb()
```
dae, line 609: It's not safe to remove this - if you inject an error into _play() above, you'll see that if cb is not called, no further sounds can play, as Anki never notices that the previous play request has failed/finished.

qt/aqt/tts.py
```diff
@@ -257,11 +257,9 @@ def _on_done(self, ret: Future, cb: OnDoneCallback) -> None:
         # inject file into the top of the audio queue
         from aqt.sound import av_player
 
+        av_player.current_player = None
         av_player.insert_file(self.tmppath)
 
-        # then tell player to advance, which will cause the file to be played
-        cb()
-
```
dae, line 264: Just a heads up that MacTTSFilePlayer is not the default player on macOS. When I manually adjust the bottom of sound.py to use this player instead, things work correctly with this change though. :-)

dae:
I've given this a test on Windows and macOS and it appears to work well - thank you!

Issue #2876, opened 2023-12-06
transforms to groups of objects are not persisted


dae:
- When items are made into a group, and the group resized, the changes are not persisted
- When 2+ items are selected but not made into a group, resizing the selection results in glitchy behavior

https://forums.ankiweb.net/t/cannot-change-image-occlusion-mask-dimensions/38078

dae:
@abdnh interested in looking into this?

abdnh:
Will have a look

PR #2878, opened 2023-12-08
Persist mask group scaling

abdnh:
Handles the first issue in #2876

Will look into the other one in a separate PR.


dae:
Nice - thanks for these!

PR #2879, opened 2023-12-08
More fixes for auto advance state

abdnh:
https://forums.ankiweb.net/t/anki-23-12-beta/37771/79

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -11,7 +11,7 @@ import signal
 import weakref
 from argparse import Namespace
 from concurrent.futures import Future
-from typing import Any, Literal, Sequence, TypeVar, cast
+from typing import Any, Literal, Optional, Sequence, TypeVar, cast
 
 import anki
 import anki.cards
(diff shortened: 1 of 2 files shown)


dae:
I'm having a bit of trouble following the logic here - could you help me understand it better?

- Why do we need to start auto-advance when the main window receives focus and is in review mode? Couldn't the toggle set reviewer.auto_advance_enabled directly instead, instead of waiting for the focus-in event?
-  Having a tristate stored in main.py and then another bool in reviewer makes it a bit difficult to follow. It would be nicer if the advancing logic was all stored in reviewer.py, with main.py just informing the reviewer that focus has changed (eg by adding reviewer.focus_was_lost(), and if necessary, .focus_was_gained())

dae:
I noticed a separate issue when making some tweaks today: the answer side was being shown only briefly before it was going to the next card. I suspect this is related to the 'wait for audio' feature, as the cards I was testing with didn't have any audio on them.

abdnh:
> Why do we need to start auto-advance when the main window receives focus and is in review mode? 

Assuming you mean the auto_advance_if_enabled() call specifically, my thinking is that, without it, auto advance won't resume until the next question/answer is revealed if the timer timed out.


Will look into the other issues (probably tomorrow).

abdnh:
Looks like the whole focus handling logic is unnecessary

dae:
Thanks Abdo!

PR #2886, opened 2023-12-12
Change link to manual in help screen from icon to text

Gustaf-C:
As suggested in the forums.

https://forums.ankiweb.net/t/anki-23-12-beta/37771/98

Screenshot 2023-12-12 at 21 08 30

Front end stuff is very much a "try and see what works" thing for me, so I'm sure this could be done better :)


dae:
Looks good - thanks!

PR #2888, opened 2023-12-13
Correct the description of FSRS Difficulty in Stats

user1823:
The current description of Difficulty is misleading; it describes (1 − retrievability) better.


dae:
I find the suggested wording unclear. How about something like the following?

The higher the difficulty, the slower intervals will increase.

user1823:
I applied your suggestion, with one change: replaced intervals with stability.

PR #2889, opened 2023-12-14
More precise percent of days studied

user1823:
https://forums.ankiweb.net/t/anki-23-12-beta/37771/101

I know that this is not a very necessary change and it makes the code longer. But, I feel that this is good to have. Obviously, you can disagree with me.

The proposed code works even when a single day is missed in over 50 years.


user1823:
The test failed due to formatting issue. That can be corrected if you agree with the proposed solution.

chrislongros (approved):
I agree that the percentage should be more precise in > 99% ranges.

dae:
I am ok with the proposed change if you can fix the code so that the tests pass.

user1823:
I need some help.

The error message says:

> Error formatting /state/build/lin-ci/ankitects/anki-ci/ts/graphs/reviews.ts. Message: Expected ',', got ')' at /state/build/lin-ci/ankitects/anki-ci/ts/graphs/reviews.ts:382:46
>  
>            if (studiedPercent < 99.5) {            
> Had 1 error(s) formatting. Failed with code Some(1): /state/build/node_modules/.bin/dprint check

But, I don't think that there should be a comma after 99.5 in the above line.

dae:
While I appreciate you making an effort to submit a PR instead of just a feature request, if you're unable to test and run them locally before submitting, there's not much of a difference unfortunately. Your syntax does not appear to be valid (it would be valid in Rust), and you appear to be returning a number in the 100 case instead of a string.

I asked ChatGPT to convert the code for you: 

```typescript
percent: (() => {
    if (studiedPercent < 99.5) {
      return localizedNumber(studiedPercent);
    } else if (studiedPercent < 99.95) {
      return localizedNumber(studiedPercent, 1);
    } else if (studiedPercent < 100) {
      return localizedNumber(studiedPercent, 2);
    } else {
      return '100';
    }
  })()
```

user1823:
Ok, I accept that fixing this is not my cup of tea. If any other dev is willing to complete this, it would be great.

Also, sorry for the inconvenience caused. In the future, I will be careful to submit PRs only when I am almost 100% sure about the code working as intended.

Luc-Mcgrady:
Sometimes the tests fail for no reason. This, to me, appears to be one of those times. 
Try and merge the base branch or force push to re-run the tests and see what happens.

user1823:
@Luc-Mcgrady, you were absolutely right; the tests have passed now.

PR #2893, opened 2023-12-16
Fix macOS not registering some copy actions when Anki is not in focus

### LINKED ISSUE #2843: Raycast copy and paste problem
The objective of this issue tracker is to gather users which are affected by the issue
---

**App involved**: [Raycast (Floating Notes)](https://www.raycast.com/)

[**Forum post related**](https://forums.ankiweb.net/t/copy-paste-issue/37168?u=namelessgo)

**Workaround**: Because Anki relies on clipboard change events, it seems that switching to another application, fixes the issue

Example of workaround: Copy "A"in anki, then "B" in floating note (Raycast), then switch application and paste on Anki, it would fix the issue and paste the correct "B" instead of "AB"

Steps to reproduce:
1. Copy text (a) from field in Anki.
2. Copy another text (b) from another app.
3. Paste text in Anki (this renders text a)
4. Paste text in Anki again (this renders text b).

65f01f5eb813305c774d36083971285a73065a89
Credits for reporting, providing the steps to reproduce and GIF: herrsilen

msisovic:
Fixes #2843.

The `QClipboard.dataChanged` signal doesn't seem to work properly on macOS. In the [official Qt documentation](https://doc.qt.io/qt-6/qclipboard.html#dataChanged) it says: 

> With Qt version 4.3 or higher, clipboard changes made by other applications will only be detected when the application is activated.

I however found the signal triggers only if you unminimize the app, not just when you switch back focus to it. This looks like incorrect behavior to me.

That's why I propose a solution here where we remove reliance on that signal, and just check if the last known `QMimeData` clipboard content is equal to the current one on every paste, and see if a change happened that way.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -153,6 +153,7 @@ Akash Reddy <akashreddy2003@gmail.com>
 Lucio Sauer <watermanpaint@posteo.net>
 Gustavo Sales <gustavosmendes14@gmail.com>
 Shawn M Moore <https://github.com/sartak>
+Marko Sisovic <msisovic13@gmail.com>
 
 ********************
 
--- a/qt/aqt/about.py
+++ b/qt/aqt/about.py
@@ -188,6 +188,7 @@ def show(mw: aqt.AnkiQt) -> QDialog:
             "Jarrett Ye",
             "Gustavo Sales",
             "Akash Reddy",
+            "Marko Sisovic",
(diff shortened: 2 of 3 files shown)


msisovic:
@EricWong1994 The issues seem to be connected. I can't tell when it will be fixed though, I am at somewhat of a dead end with it right now.

PR #2894, opened 2023-12-16
Do not close web pages when Esc is pressed and a modal is open

### LINKED ISSUE #2300: Esc closes deck options window instead of closing pop-ups
Running on Anki 2.1.55 (also tested on the latest commit and the problem persists).

Steps to reproduce:
- Open the deck options window
- Open the Add/Clone/Rename Preset pop-up
- Press esc

Expected behavior:
- Pop-up closes

Actual behavior:
- Deck options window closes

The same problem affects the presets dropdown list (on the same window), and I don't think that's the intended behavior since pressing esc on the card browser's search history dropdown list (as an example) only closes the list.

abdnh:
Closes #2300

### DIFF
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -295,7 +295,7 @@ class AnkiWebView(QWebEngineView):
         self.eval(
             """
         document.addEventListener("keydown", function(evt) {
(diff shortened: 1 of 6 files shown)


dae:
Thank you!

PR #2896, opened 2023-12-19
Fix opening card info window closes it when already open

### LINKED ISSUE #2892: Card info should perhaps update instead of toggle?
https://forums.ankiweb.net/t/have-to-select-info-twice-when-the-card-info-window-is-already-open-first-time-just-closes-the-window/38396

akashreddy03:
Fixes #2892  

Since we're already updating card info when the row changes in the browse window, this fix just brings the card info window to the front instead of closing when it is already open.


dae:
Thanks Akash!

PR #2899, opened 2023-12-21
Offer to save changes when card layouts screen is closed

### LINKED ISSUE #2282: improve confirmation screen when closing add dialog
Yes/no buttons would be clearer as eg. "Discard" and "Keep Editing", and the prompt will probably need to be changed to something like "Discard current input?". This will probably require adding some new translation strings.

https://forums.ankiweb.net/t/saving-changes-to-evaluation-exit/25546

abdnh:


dae:
Thanks Abdo, quality work as always. One clarification: we need to keep #2282 around, as that was originally about changing the wording of the dialog that's shown when closing the Add screen.

PR #2900, opened 2023-12-22
Fix macOS clipboard issues

### LINKED ISSUE #2843: Raycast copy and paste problem

msisovic:
Fixes #2843.

Also fixes [this issue](https://forums.ankiweb.net/t/copy-paste-of-screenshots-dont-work-once-i-copy-the-text-on-the-anki-app/34839), which actually wasn't properly fixed.

The issue is that `QClipboard.dataChanged` doesn't always properly fire on macOS when performing copy operations outiside of the app. I fixed this by adding just-in-time checking if the clipboard changed before pasting.

The ideal solution would be getting Qt to work properly, but until then this solution should be 100% working.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -153,6 +153,7 @@ Akash Reddy <akashreddy2003@gmail.com>
 Lucio Sauer <watermanpaint@posteo.net>
 Gustavo Sales <gustavosmendes14@gmail.com>
 Shawn M Moore <https://github.com/sartak>
+Marko Sisovic <msisovic13@gmail.com>
 
 ********************
 
--- a/qt/aqt/about.py
+++ b/qt/aqt/about.py
@@ -188,6 +188,7 @@ def show(mw: aqt.AnkiQt) -> QDialog:
             "Jarrett Ye",
             "Gustavo Sales",
             "Akash Reddy",
+            "Marko Sisovic",
         )
     )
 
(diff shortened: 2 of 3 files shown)


dae:
Thanks for looking into this, Marko!

EricWong1994:
I check this in the [23.12rc1](https://github.com/ankitects/anki/releases/tag/23.12rc1) version. 
It has been fixed.  Thanks a lot
❤️

dae:
Thanks for letting us know!

PR #2901, opened 2023-12-22
Update to FSRS-4.5

L-M-Sherlock:
I make a minimum modification that could work as usual.


dae:
Thanks Jarrett. I was not sure if simply setting the limits to the maximum would be sufficient.

PR #2902, opened 2023-12-22
Change the description of stability.ftl

Expertium:


dae:
While I think what you've written here is pretty clear (especially to someone already familiar with the DSR values used by FSRS), as previously mentioned, I think subtitles need to strike a balance between clarity and brevity to avoid cluttering the screen, and this one feels a bit verbose. Mentioning "in days" is not particularly useful as the unit is already visible in the hovering tooltip, and "likely to remember" feels simpler to the new user than talking about retrievability decreasing.

Also, for future PRs, please ensure the tests pass.

Expertium:
I'm sorry for being rude, but "The delay at which you're 90% likely to remember." sounds like it was written by a sentient bag of cocaine. If you don't like my suggestion - fine, but please change that sentence to something that makes at least a little bit of sense. @user1823, maybe you have a good idea?

PR #2903, opened 2023-12-22
Improve confirmation message in Add screen

### LINKED ISSUE #2282: improve confirmation screen when closing add dialog

abdnh:
Now should close #2282 😃


dae:
Thanks Abdo :-)

PR #2907, opened 2023-12-24
Fix new cards from parent decks being gathered

### LINKED ISSUE #2906: Cards from parent decks are being included
Have reproduced. https://forums.ankiweb.net/t/anki-23-12-beta/37771/165

@RumovZ when you're not busy, would you mind looking into this? It might be worth tackling before the importing issue, as it's likely to affect more people. Hopefully it doesn't require major changes to our current approach.

RumovZ:
Fixes #2906.


dae:
Quick fix much appreciated :pray:

PR #2909, opened 2023-12-25
Skip tests failing around rollover hour

### LINKED ISSUE #2511: Flaky v3 unit test around 3:44am GMT
Seen at around 3:44am GMT:

thread 'deckconfig::update::test::should_keep_at_least_one_remaining_relearning_step' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidInput { source: InvalidInputError { message: "bug: card modified without updating queue: id:1684727050997 card:1684727051 entry:1684727050", source: None, backtrace: None } }', rslib/src/scheduler/answering/mod.rs:400:61

I think some other tests will also flake at around 3:55am. We should run the tests locally while forcing a particular time to debug these issues.

abdnh:
Closes #2511

I've not looked much into what causes the tests to fail.

### DIFF
--- a/rslib/src/import_export/package/apkg/import/notes.rs
+++ b/rslib/src/import_export/package/apkg/import/notes.rs
@@ -604,6 +604,7 @@ impl Notetype {
 mod test {
     use anki_proto::import_export::ExportAnkiPackageOptions;
     use anki_proto::import_export::ImportAnkiPackageOptions;
+    use chrono::Timelike;
     use tempfile::TempDir;
 
     use super::*;
@@ -611,6 +612,7 @@ mod test {
     use crate::import_export::package::media::SafeMediaEntry;
     use crate::notetype::CardTemplate;
     use crate::notetype::NoteField;
+    use crate::skip_if_rollover_hour;
 
     #[derive(Default)]
     struct ImportBuilder {
@@ -955,6 +957,7 @@ mod test {
 
     #[test]
     fn reimport_with_merge_enabled_should_handle_duplicates() -> Result<()> {
+        skip_if_rollover_hour!();
         // import from src to dst
         let mut src = Collection::new();
         NoteAdder::basic(&mut src)
--- a/rslib/src/import_export/package/apkg/import/notes.rs
+++ b/rslib/src/import_export/package/apkg/import/notes.rs
@@ -604,7 +604,6 @@ impl Notetype {
 mod test {
     use anki_proto::import_export::ExportAnkiPackageOptions;
     use anki_proto::import_export::ImportAnkiPackageOptions;
-    use chrono::Timelike;
     use tempfile::TempDir;
 
     use super::*;
@@ -612,7 +611,6 @@ mod test {
     use crate::import_export::package::media::SafeMediaEntry;
     use crate::notetype::CardTemplate;
     use crate::notetype::NoteField;
-    use crate::skip_if_rollover_hour;
 
     #[derive(Default)]
     struct ImportBuilder {
@@ -957,7 +955,6 @@ mod test {
 
     #[test]
     fn reimport_with_merge_enabled_should_handle_duplicates() -> Result<()> {
-        skip_if_rollover_hour!();
         // import from src to dst
         let mut src = Collection::new();
         NoteAdder::basic(&mut src)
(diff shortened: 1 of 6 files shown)


rslib/src/import_export/package/apkg/import/notes.rs
```diff
@@ -955,6 +957,7 @@ mod test {
 
     #[test]
     fn reimport_with_merge_enabled_should_handle_duplicates() -> Result<()> {
```
abdnh, line 959: This is a weird one. The last assert is failing.

dae:
My apologies Abdo, I think I may have sent you on a wild goose chase here. We already have code that attempts to shift the time away from the rollover period (while keeping the tests running, which helps ensure we don't let bugs through accidentally at certain hours of the day). I'd assumed that code was running and this was some issue that would need to be handled differently. But it turns out that the build scripts were just not enabling the timeshift. This should be all that's required:

```rust
diff --git a/build/ninja_gen/src/cargo.rs b/build/ninja_gen/src/cargo.rs
index 33e7778ce..645203170 100644
--- a/build/ninja_gen/src/cargo.rs
+++ b/build/ninja_gen/src/cargo.rs
@@ -153,6 +153,7 @@ impl BuildAction for CargoTest {
     fn files(&mut self, build: &mut impl FilesHandle) {
         build.add_inputs("", &self.inputs);
         build.add_inputs("", inputs![":cargo-nextest"]);
+        build.add_env_var("ANKI_TEST_MODE", "1");
         build.add_output_stamp("tests/cargo_test");
     }
 
```

That will alter the behavior of elapsed(), which should fix the failing tests.

Out of curiosity, I only see 3 failing tests at around 3:55, and I see you've excluded quite a few other tests in this PR. Were you seeing errors on your machine? After applying the above patch to main, do you still see any errors around that time?

abdnh:
I noticed the shifting code in elapsed(), but didn't think of checking whether it's enabled for Rust tests. Instead, I forced the function to return a particular time always (rather than shifting) and checked failing tests. 

> Out of curiosity, I only see 3 failing tests at around 3:55, and I see you've excluded quite a few other tests in this PR. Were you seeing errors on your machine? 

I noticed some were breaking at different points in time around the rollover hour. But testing more, it turned out this has nothing to do with the rollover hour, as some were randomly breaking when I set different times of the day 🤦‍♂️ Probably something is wrong with my modification to elapsed() (I wasn't setting the nanoseconds part before - but the syncing tests at least are still failing with this though):
```diff
diff --git a/rslib/src/timestamp.rs b/rslib/src/timestamp.rs
index a020d706d..b4952275a 100644
--- a/rslib/src/timestamp.rs
+++ b/rslib/src/timestamp.rs
@@ -4,6 +4,8 @@
 use std::time;

 use chrono::prelude::*;
+use chrono::Duration;
+use lazy_static::lazy_static;

 use crate::define_newtype;
 use crate::prelude::*;
@@ -96,6 +98,10 @@ impl TimestampMillis {
 }

 fn elapsed() -> time::Duration {
+    lazy_static! {
+        static ref START_TIME: time::SystemTime = time::SystemTime::now();
+    }
+
     if *crate::PYTHON_UNIT_TESTS {
         // shift clock around rollover time to accommodate Python tests that make bad
         // assumptions. we should update the tests in the future and remove this     
@@ -109,8 +115,18 @@ fn elapsed() -> time::Duration {
         }
         elap
     } else {
-        time::SystemTime::now()
-            .duration_since(time::SystemTime::UNIX_EPOCH)
-            .unwrap()
+        let shifted_now = Local::now().with_hour(3).unwrap().with_minute(55).unwrap();
+        let date = shifted_now
+            + Duration::from_std(
+                (time::SystemTime::now() - START_TIME.elapsed().unwrap())
+                    .elapsed()
+                    .unwrap(),
+            )
+            .unwrap();
+
+        time::Duration::new(
+            date.timestamp() as u64,
+            date.timestamp_nanos_opt().unwrap() as u32,
+        )
     }
 }
```

dae:
Just to confirm, with the current PR (without the changes you quoted), the sync tests aren't failing? If so, I guess we're ok? :-)

abdnh:
Yes, they are not failing (Just confirmed that by shifting the time to the rollover hour)

PR #2910, opened 2023-12-26
Improve description of stability in Stats

user1823:
Supersedes and closes https://github.com/ankitects/anki/pull/2902


Expertium:
Maybe change "delay" to "interval"? I think it's easier 
to understand. Also, Dae said 
> "likely to remember" feels simpler to the new user than talking about retrievability decreasing.

So I'm not sure if he will approve.

dae:
I don't have strong feelings about delay/interval, but delay matches the interval graph above it. This retains brevity, and is good enough - I've reached my bikeshedding quota for a while :-)

PR #2912, opened 2023-12-27
Cloze notetype merges

### LINKED ISSUE #2904: Merging cloze notetypes can break them
Regarding the report of basiskarten in https://forums.ankiweb.net/t/anki-23-12-beta/37771/121, my working hypothesis is that the user changed the notetype of previously imported notes and that's why the various different notetypes got merged. If they didn't, it would indicate a bug. basiskarten is investigating this.

Definitely an issue is the merging of cloze notetypes. I hadn't considered those at all in the initial implementation. I suggest the following solution:
1. For cloze notetypes, fields are merged as usual, but templates are not, because the (first) template of each notetype is assumed to be identical, regardless of id.
2. If an attempt is made to merge a cloze with a regular notetype, the import aborts with an error. I don't see any way to merge in a somewhat reasonable fashion in this case.

Considering this issue has the potential to break notetypes, do you want to include the fix in 23.12? I would then try to submit it in the next couple of days.

RumovZ:
https://forums.ankiweb.net/t/anki-23-12-beta/37771/121

Closes #2904.


dae:
Thanks for this Rumo :-)

Issue #2915, opened 2023-12-28
Shortcuts are conflicting with text input in image occlusion on mobile


dae:
We're already testing for `!$textEditingState` which works in the desktop editor, but we perhaps need to be adjusting that state as the user navigates between the two tabs on mobile. Original report:

https://forums.ankiweb.net/t/bugs-report-image-occlusion-problems/38868

@abdnh if you're not busy with holidays/family, would you mind looking into this when you have a chance? It should be a pretty easy fix, and you can test the mobile client's page via http://localhost:40000/_anki/pages/image-occlusion.html#test-/home/dae/foo.jpg, replacing the path at the end with a valid path.

abdnh:
I guess this can be closed.

dae:
Thanks Abdo!

PR #2916, opened 2023-12-28
fix: enable keyboard in mobile io note editor

krmanik:
https://forums.ankiweb.net/t/bugs-report-image-occlusion-problems/38868
#2915


dae:
Thanks for the very quick fix Mani! Do you think it would make sense to reuse textEditingState instead of introducing an extra variable? If there's a reason why they need to be separate, that's fine.

krmanik:
Text editing state store value for canvas text, but in note editor we have html text area so both are different. So, it needs to store the state for note editor for mobile io also.

```
    canvas.on("text:editing:entered", function() {
        textEditingState.set(true);
    });

    canvas.on("text:editing:exited", function() {
        textEditingState.set(false);
    });
```

dae:
Great, thanks again Mani!

PR #2918, opened 2023-12-28
Allow > inside HTML attributes

### LINKED ISSUE #2913: Media regexes get confused by angle brackets in img tags
https://forums.ankiweb.net/t/images-incorrectly-marked-as-unused-when-checking-media-due-to-special-characters-in-img-tag/38863

Moving from our current regex approach to a HTML parser will likely have a performance hit and increase complexity, so perhaps in the short term we could strip the brackets out of the alt tags when pasting.

vaxr:
Fixes #2913

Extends the `HTML_MEDIA_TAGS` regex to allow for `>` in non-matched attributes. The regex might not deal with things like nested escaped quotes, but neither does the original one, and I didn't want to complicate it even further.

Hope this helps. I'm new here, so sorry if I missed something.

### DIFF
--- a/rslib/src/text.rs
+++ b/rslib/src/text.rs
@@ -110,9 +110,9 @@ lazy_static! {
             (?:
                 [^>]
             |
-                "(?:[^"]+?)"
+                "[^"]+?"
             |
-                '(?:[^']+?)'
+                '[^']+?'
             )+
 
             # capture `src` or `data` attribute
--- a/rslib/src/text.rs
+++ b/rslib/src/text.rs
@@ -104,7 +104,19 @@ lazy_static! {
     pub static ref HTML_MEDIA_TAGS: Regex = Regex::new(
         r#"(?xsi)
             # the start of the image, audio, or object tag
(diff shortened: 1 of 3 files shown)


rslib/src/text.rs
```diff
@@ -104,7 +104,19 @@ lazy_static! {
...
-            <\b(?:img|audio|video|object)\b[^>]+\b(?:src|data)\b=
+            <\b(?:img|audio|video|object)\b
+
+            # any non-`>`, except inside `"` or `'`
+            (?:
+                [^>]
+            |
+                "(?:[^"]+?)"
+            |
+                '(?:[^']+?)'
```
dae, line 115: Do we need the inner (?: ... ) in these two clauses?

dae (commented):
Thank you for contributing! Just one clarification:

rslib/src/text.rs
vaxr: You're right, we don't need them. Removed.

dae:
Thanks Viktor!

vaxr:
@meliache Thanks for bringing this up! I'll take a look.

PR #2919, opened 2023-12-30
add padding to control corner to shapes

krmanik:
Add padding to control corner, so shapes can be selected, and control corner should be outside of the shape.


dae:
Nice - thanks Mani!

PR #2920, opened 2023-12-30
Use our translation of 'Discard'

abdnh:
https://forums.ankiweb.net/t/some-rtl-and-translate-issues/38892/9


qt/aqt/utils.py
```diff
@@ -164,6 +164,8 @@ def __init__(
                 b = self.addButton(button, QMessageBox.ButtonRole.ActionRole)
             elif isinstance(button, QMessageBox.StandardButton):
                 b = self.addButton(button)
+                if button == QMessageBox.StandardButton.Discard:
```
dae, line 167:
```suggestion
                # a translator has complained the default Qt translation is inappropriate, so we override it
                if button == QMessageBox.StandardButton.Discard:
```

PR #2922, opened 2024-01-01
FSRS - Ignore revlogs before date while optimizing

Luc-Mcgrady:
Allows users to ignore revlogs generated before a given date while reviewing.
Useful for ignoring bad reviews, a feature of the non-inbuilt optimizer.

image

Implemented for Optimize, Evaluate and Optimize all presets.


proto/anki/scheduler.proto
```diff
@@ -338,6 +338,7 @@ message RepositionDefaultsResponse {
 message ComputeFsrsWeightsRequest {
   /// The search used to gather cards for training
   string search = 1;
+  int64 ignoreBefore = 2;
```
dae, line 341: This should be snake_case.

proto/anki/deck_config.proto
```diff
@@ -152,6 +152,7 @@ message DeckConfig {
 
     // for fsrs
     float desired_retention = 37;
+    string ignore_before_date = 46;
```
dae, line 155: Please rename this and the references below to ignore_revlogs_before_date / ignore_revlogs_before_secs to make it clearer.

rslib/src/revlog/mod.rs
```diff
@@ -59,6 +59,16 @@ pub struct RevlogEntry {
...
+pub trait RemoveBefore {
+    fn remove_before(self, ms: i64) -> Self;
+}
+
+impl RemoveBefore for Vec<RevlogEntry> {
+    fn remove_before(self, ms: i64) -> Self {
+        self.into_iter().filter(|revlog| revlog.id.0 > ms).collect()
+    }
+}
+
```
dae, line 71: I'd prefer this were a standalone function instead of a method in a trait, as we only use it twice. It would also be nice if it avoided doing anything when ms is 0, as otherwise the we wastefully rebuild the vec without doing any filtering.

ts/deck-options/DateInput.svelte
```diff
@@ -0,0 +1,30 @@
...
+</script>
+
+<div>
+    <ConfigInput>
+        <Row --cols={2}>
+            <Col>
+                <slot />
+            </Col>
+            <Col>
+                <input bind:value={date} type="date" />
```
dae, line 20: This doesn't look great on Safari, especially when not set (will attach photo in follow-up). You can test with `ANKI_API_HOST=0.0.0.0 ./run`, then surfing to `http://localhost:40000/_anki/pages/deck-options.html#test` on a mobile device.

rslib/src/deckconfig/update.rs
```diff
@@ -332,7 +334,19 @@ impl Collection {
...
-            match self.compute_weights(&search, idx as u32 + 1, config_len) {
+
+            let ignore_before = match &config.inner.ignore_before_date {
+                s if s.is_empty() => 0,
+                s => {
+                    NaiveDate::parse_from_str(s.as_str(), "%Y-%m-%d")
+                        .or_else(|err| invalid_input!(err, "Error parsing date: {s}"))?
+                        .and_time(NaiveTime::from_hms_milli_opt(0, 0, 0, 0).unwrap())
+                        .timestamp()
+                        * 1000
```
dae, line 345: .as_millis()

rslib/src/scheduler/fsrs/weights.rs
```diff
@@ -110,14 +113,20 @@ impl Collection {
         Ok(())
     }
 
-    pub fn evaluate_weights(&mut self, weights: &Weights, search: &str) -> Result<ModelEvaluation> {
+    pub fn evaluate_weights(
+        &mut self,
+        weights: &Weights,
+        search: &str,
+        ignore_before: i64,
```
dae, line 120: TimestampMillis

rslib/src/scheduler/fsrs/weights.rs
```diff
@@ -33,12 +34,14 @@ impl Collection {
     pub fn compute_weights(
         &mut self,
         search: &str,
+        ignore_before: i64,
```
dae, line 37: TimestampMillis

ftl/core/deck-config.ftl
```diff
@@ -398,6 +399,10 @@ deck-config-reschedule-cards-warning =
 
     Use this option sparingly, as it will add a review entry to each of your cards, and
     increase the size of your collection.
+deck-config-ignore-before-tooltip = 
+    This setting makes the optimizer ignore reviews that were made before 00:00 on a given date.
```
dae, line 403:
```suggestion
    If set, reviews before the provided date will be ignored when optimizing & evaluating FSRS parameters.
```

ftl/core/deck-config.ftl
```diff
@@ -398,6 +399,10 @@ deck-config-reschedule-cards-warning =
...
+
+    This can be used if old reviews no long reflect the difficulty of the deck, for example if significant changes were made to the cards.
```
dae, line 405:
```suggestion
    This can be useful if you imported someone else's scheduling data, or have changed the way you use the answer buttons.
```

ftl/core/deck-config.ftl
```diff
@@ -398,6 +399,10 @@ deck-config-reschedule-cards-warning =
 
     Use this option sparingly, as it will add a review entry to each of your cards, and
     increase the size of your collection.
+deck-config-ignore-before-tooltip = 
+    This setting makes the optimizer ignore reviews that were made before 00:00 on a given date.
+
```
dae, line 404:
```suggestion
```

dae (commented):
Thanks Luc. Some feedback:

ts/deck-options/DateInput.svelte
dae: IMG_1948

Luc-Mcgrady: I'm afraid I can't test this as I don't have an Iphone :(

dae:
I will see if it can be customized. It might take a while until I have a chance to.

rslib/src/scheduler/fsrs/weights.rs
```diff
@@ -110,16 +123,23 @@ impl Collection {
...
         let mut anki_progress = self.new_progress_handler::<ComputeWeightsProgress>();
         let guard = self.search_cards_into_table(search, SortMode::NoOrder)?;
-        let revlogs = guard
+        let revlogs: Vec<RevlogEntry> = guard
             .col
             .storage
             .get_revlog_entries_for_searched_cards_in_card_order()?;
-        anki_progress.state.fsrs_items = revlogs.len() as u32;
-        let items = fsrs_items_for_training(revlogs, timing.next_day_at);
+        let filtered_revlogs = remove_revlogs_before(revlogs, ignore_revlogs_before_ms);
```
user1823, line 139: I am not sure if the revlogs should also be filtered when evaluating the weights.
- In the Python version, if the card was first reviewed before the revlog_start_date, then all of its reviews are filtered out (because of the code that filters out incomplete revlogs).
- In the Rust version, cards with incomplete revlogs are treated better because their evaluation corresponds with how the cards will actually be scheduled.
But due to this difference, after the proposed change, the cards which were first reviewed before the "start date" will have their memory states calculated from SM-2 variables during evaluation and using normal formulas during scheduling.

Luc-Mcgrady: >In the Python version, if the card was first reviewed before the revlog_start_date, then all of its reviews are filtered out (because of the code that filters out incomplete revlogs).

Python code in question:
https://github.com/open-spaced-repetition/fsrs-optimizer/blob/951981130b53dd023f6986127e7898b815e5c0d5/src/fsrs_optimizer/fsrs_optimizer.py#L663-L666        
```py
df = df.groupby("card_id").filter(
            lambda group: group["review_time"].min()
            > time.mktime(datetime.strptime(revlog_start_date, "%Y-%m-%d").timetuple())
            * 1000
        )
```

That does seem to be the case. That would explain why my deck optimized using the external and internal optimizer still differed when using this even with the same ignore date (Something I chalked up to differences in the rust and python implementations). As I see it my way is superior because otherwise cards that are reviewed before the ignore before date aren't completely ignored by the optimizer which ruins the idea of giving the card a "second chance". This would limit the cards the optimizer uses to the ones only added after the date. 

> In the Rust version, cards with incomplete revlogs are treated better than the Python version because their evaluation corresponds with how the cards will actually be scheduled.
But this strength of the Rust version also introduces new complications when making changes. After the proposed change, the cards which were first reviewed before the "start date" will have their memory states calculated from SM-2 variables during evaluation and using normal formulas during scheduling.

This change does not effect scheduling, it merely effects the "Evaluate" button on the deck config menu. I don't believe The Evaluate button is affected by the card states if that is what you're saying?

user1823: > As I see it my way is superior because otherwise cards that are reviewed before the ignore before date aren't completely ignored by the optimizer which ruins the idea of giving the card a "second chance". 

The Rust optimizer also ignores such cards. The difference is in Evaluation.

> This change does not effect scheduling, it merely effects the "Evaluate" button on the deck config menu.

Yes, your change doesn't affect the scheduling. Before and after your change, the cards that were introduced before the start date will have their memory states calculated using normal FSRS formulas (if they have complete revlogs). 

However, after your change, during the evaluation, the memory states of such cards are calculated using SM-2 variables (which the scheduler uses when it encounters incomplete revlogs).

> I don't believe The Evaluate button is affected by the card's memory states if that is what you're saying?

It is not affected by the current memory states. But, it is affected by the memory states generated at each step of the simulation of the review history that happens when the user clicks `Evaluate`.

Luc-Mcgrady: I made it so that it adds a fake learning review to the start of every truncated review which should prevent sm2 being used and it seems to have made no difference.

```rs
// weights.rs:29 
fn remove_revlogs_before(revlogs: Vec<RevlogEntry>, ms: TimestampMillis) -> Vec<RevlogEntry> {

    // If there is no minimum time or the minimum time is lower than every review's review time.
    if ms.0 == 0 || revlogs.iter().map(|revlog| revlog.id.0).min().unwrap_or(-1) > ms.0 {
        revlogs
    } else {
        // Insert a fake revlog at the start so single_card_revlog_to_items doesn't use sm2.
        let filtered_revlogs = revlogs.into_iter().filter(|revlog| revlog.id.0 > ms.0);

        let mut out = vec![RevlogEntry {
            id: ms.into(),
            cid: 0.into(),
            usn: 0.into(),
            button_chosen: 1,
            interval: -1,
            last_interval: -1,
            ease_factor: 250,
            taken_millis: 0,
            review_kind: RevlogReviewKind::Learning,
        }];

        out.extend(filtered_revlogs);
        out
    }
}
```

With change:
image

Without change:
image

Unless I misunderstood the detection code or something else:

https://github.com/ankitects/anki/blob/3982e0c8fe691d3e84deceb7555f2d2fd0a8fbe6/rslib/src/scheduler/fsrs/weights.rs#L177-L188

I don't think this affects the reviews at all.
This was still worth looking into though. Thanks.

dae: @L-M-Sherlock @expertium can you also confirm you're happy with the approach this PR takes?

L-M-Sherlock: > That does seem to be the case. That would explain why my deck optimized using the external and internal optimizer still differed when using this even with the same ignore date (Something I chalked up to differences in the rust and python implementations). As I see it my way is superior because otherwise cards that are reviewed before the ignore before date aren't completely ignored by the optimizer which ruins the idea of giving the card a "second chance". This would limit the cards the optimizer uses to the ones only added after the date.

I don't agree. What if the user use `hard` as `again` before the ignore date? These reviews will be used as history features in your implementation. They will pollute the dataset and make FSRS confused.

Luc-Mcgrady: I'll change it to the python method if you want but I'm still confused about the mechanics of this.
In a revlog where | represents the ignore before date:

```
Original revlog:              1 3 3 2 2 2 | 1 1 3 3 1 3
Current implementation:                   | 1 1 3 3 1 3
```
I want to know exactly where the pollution comes from here. Did I miss some place where the SM-2 values are substituted in or something? What would I need to change to remove this pollution?
Using the python implementation this entire revlog would be ignored which means this feature would be rendered useless for people who didn't add any cards after the ignore before because it would ignore every card, right?

L-M-Sherlock: Sorry, I misunderstood the current implementation. The current problem is the incomplete revlogs. In FSRS, the first rating is treated in different way than subsequent ratings. The current implementation will let FSRS use review data in pretraining process, which is incorrect.

user1823: In my opinion, the best solution is to make the evaluation exactly match the scheduler behaviour. This is because the main purpose of the evaluation is to determine the effectiveness of scheduling, which would be most accurate when both treat the revlogs in the same way.

To do this, the revlogs before the start date should NOT be removed during evaluation.

> Using the python implementation this entire revlog would be ignored which means this feature would be rendered useless for people who didn't add any cards after the ignore before because it would ignore every card, right?

The optimizer will still completely ignore those cards, which means that the issue will persist. However, incomplete revlogs are not usable for training. Refer to https://github.com/open-spaced-repetition/fsrs-rs/issues/114 for a more detailed description.

user1823: > The current implementation will let FSRS use review data in pretraining process, which is incorrect.

@L-M-Sherlock, I don't think so because the first entry in the splitted revlog won't be of LEARN type. However, I may be interpreting the code incorrectly. So, you can double-check.

Luc-Mcgrady: > In FSRS, the first rating is treated in different way than subsequent ratings

So while filtering the revlogs, The first revlog should be preserved even if its before the date?

> I hope that this was for testing purpose only and you don't want to include it in the PR.

Yep just testing.

> In my opinion, the best solution is to make the evaluation exactly match the scheduler behavior. This is because the main purpose of the evaluation is to determine the effectiveness of scheduling, which would be most accurate when both treat the revlogs in the same way.

Would a better alternative solution be to make it so that the scheduler ignores these cards too therefore having the scheduler treat both the same way? I apparently forgot that revlogs were used during scheduling not just optimization earlier 😖.

user1823: > Would a better alternative solution be to make it so that the scheduler ignores these cards too 

I don't think that there is a major downside to this approach except that we will have to use memory_state_from_sm2 (which is not very reliable) in a larger number of situations. But, if the past review history is very different from the current one, this can be a reasonable compromise (which the user has to make by choosing a revlog start date).

However, I still want to know the opinion of @L-M-Sherlock on this one.

Luc-Mcgrady: Alright, I programmed it to affect scheduling too. If this is a bad change I'll revert it and just remove the evaluation filter code.

dae:
I'm not particularly fond of the complexity this change is introducing for a relatively rare case. As an alternative approach, what if we provided an add-on that removed old revlogs before a certain date?

user1823:
> what if we provided an add-on that removed old revlogs before a certain date?

I think that this would be the best approach, especially if the add-on could merge the old revlogs from a backup file into the current collection. This would be needed if the user wants to undo the change after some time. In this regard, the current approach (the one implemented in this PR) is safer because the user can simply blank out the start date if they want to undo the change.

rslib/src/scheduler/fsrs/weights.rs
L-M-Sherlock: > I don't think so because the first entry in the splitted revlog won't be of LEARN type.

FSRS itself doesn't have any filters except the outlier filter. Other filters are implemented in Anki.

Luc-Mcgrady:
The addon [already exists](https://ankiweb.net/shared/info/1345711295) but is broken for me. Another negative of this is it would affect the stats menu too which would suck IMO.

It is purely anecdotal but a few people do have this problem https://www.reddit.com/r/Anki/comments/18vd2th/how_would_you_rate_your_experience_using_fsrs_so/kfqf5pp/.

Expertium:
I don't think it should be done with an add-on. That would mean that we are releasing a half-baked feature that can only be properly used on PC (since add-ons only work on PC).

L-M-Sherlock:
If we plan to merge this PR, I suggest to add an unit test.

rslib/src/scheduler/fsrs/memory_state.rs
```diff
@@ -43,14 +44,18 @@ impl Collection {
     /// memory state should be removed.
     pub(crate) fn update_memory_state(
         &mut self,
-        entries: Vec<(Option<UpdateMemoryStateRequest>, SearchNode)>,
+        entries: Vec<(
+            Option<UpdateMemoryStateRequest>,
+            SearchNode,
+            TimestampMillis,
+        )>,
```
dae, line 51: It would be a bit cleaner to define a struct for this, as tuples get unwieldy as they grow in size, and that will let us label the TimestampMillis value.

L-M-Sherlock:
I find out the an incorrect case:

```
rating:           again | good good good
review kind: learn | learn learn review
```

If the first revlog is created before the ignore date, the second revlog will be considered the first one, leading `good` to be mistakenly treated as the first rating.

L-M-Sherlock:
Yeah. So I suggest to only include cards whose first revlog is created after the ignore date.

user1823:
> Yeah. So I suggest to only include cards whose first revlog is created after the ignore date.

There is one complexity in implementing this idea. What if the user has reset the card and the first revlog entry after the reset is after the ignore date?

L-M-Sherlock:
> What if the user has reset the card and the first revlog entry after the reset is after the ignore date?

Maybe just apply the filter after detecting the last learn entry:

https://github.com/ankitects/anki/blob/cf5f716037f39e981765b2c3c06c23f003e44f64/rslib/src/scheduler/fsrs/weights.rs#L181-L198

user1823:
> Maybe just apply the filter after detecting the last learn entry:

Yes, that's a good idea.

So, the intended flow is:
- Step 1: Remove all revlogs upto the first Learn entry in the last group of Learn entries. (code linked above)
- Step 2: Remove revlogs before the start_date.
- Step 3: Exclude the cards that had revlogs before the start date even after the filtering in Step 1 from training. For these cards, use `memory_state_from_sm2` at the first revlog after the start_date during evaluation and calculation of memory states irrespective of the review kind (except Manual ratings).

Luc-Mcgrady:
> Maybe just apply the filter after detecting the last learn entry:

Right! I've moved the filtering to the fsrs_items_for_training function after the single_card_revlog_to_items and therefore the complete revlog check.

> There is one complexity in implementing this idea. What if the user has reset the card and the first revlog entry after the reset is after the ignore date?



I haven't dealt with any first revlog issues yet. 

So the intention is:
```
L = learn, R = review

L L L R R R | R R R ->
            | R R R ->
Ignored completely / SM2

L L L R R R | L L R R R ->
            | L L R R R

L L L R R R | R L L R R R ->
            |   L L R R R
```
?
That seems a pretty good way of doing it.
> If we plan to merge this PR, I suggest to add an unit test.

Absolutely. I'm not sure what such a unit test would look like but one is needed.

I have probably miss-timed making this PR as I have uni assignment deadlines looming and have spent more time on this than I probably should have.

rslib/src/scheduler/fsrs/weights.rs
```diff
@@ -216,7 +248,8 @@ pub(crate) fn single_card_revlog_to_items(
         let manually_rescheduled =
             entry.review_kind == RevlogReviewKind::Manual || entry.button_chosen == 0;
         let cram = entry.review_kind == RevlogReviewKind::Filtered && entry.ease_factor == 0;
-        if manually_rescheduled || cram {
+        let before_date = entry.cid.0 < review_revlogs_before.into();
```
L-M-Sherlock, line 251: `cid` is the timestamp of created time of the card. It is unequal to the first review time.

user1823:
I have edited my previous comment (https://github.com/ankitects/anki/pull/2922#issuecomment-1884359704) to make it more clear how to implement this. The flow is the same; I have just described it in a more clear way.

Expertium:
> * During evaluation and calculation of memory states, for these cards, use `memory_state_from_sm2` at the first revlog after the start_date irrespective

I think that it's better to calculate memory states normally, just using the new parameters. Ignoring reviews before the starting date should only affect the optimization procedure.

rslib/src/scheduler/fsrs/weights.rs
Luc-Mcgrady: Well spotted!

L-M-Sherlock: I find out a new problem. When training, we ignore cards that don't have any learning steps. What if the learning steps are removed in line 251?

L-M-Sherlock:
> I find out a new problem. When training, we ignore cards that don't have any learning steps. What if the learning steps are removed in line 251?

I figure it out. Once we find out the first Learn entry in the last group of Learn entries for a card in training mode, we compare its review date with the ignore date. If it is earlier than the ignore date, we remove this card's review history.

https://github.com/ankitects/anki/blob/862dfd639de44adb453146206b4ed04a3a5f173d/rslib/src/scheduler/fsrs/weights.rs#L186-L198

Luc-Mcgrady:
So if we had a card like
```
L = learn, R = review
L L R R R | R R R R
```
We'd ignore it 
but if we had a card like
```
L L R R R | R R L L R R
```
it would become
```
          | R R L L R R
```
Which the code that's already implemented 

https://github.com/ankitects/anki/blob/862dfd639de44adb453146206b4ed04a3a5f173d/rslib/src/scheduler/fsrs/weights.rs#L213-L217

would shrink to
```
L L R R
```
?
That seems like it would work to me.

L-M-Sherlock:
What if we had a card like

```
L = learn, R = review
L L | L R R R R R R R
```

user1823:
>  if we had a card like
> 
> ```
> L L R R R | R R L L R R
> ```
> 
> it would become
> 
> ```
>           | R R L L R R
> ```
> 
> Which the code that's already implemented would shrink to
> 
> ```
> L L R R
> ```
> 
> ? That seems like it would work to me.

This approach works for this card. But, I suggest inverting the order in which the two filters apply. My suggestion would make more sense when you consider the following case.

> What if we had a card like
> 
> ```
> L = learn, R = review
> L L | L R R R R R R R
> ```

The following suggestion contained in [my previous comment](https://github.com/ankitects/anki/pull/2922#issuecomment-1884359704) is useful here:

> Exclude the cards that have their cid in `x` from training. During evaluation and calculation of memory states, for these cards, use `memory_state_from_sm2` at the first revlog after the start_date irrespective of the review kind (except Manual ratings).

So, exclude this card from training and use `memory_state_from_sm2` at the first entry (L) after the `start_date`.

L-M-Sherlock:
> The following suggestion contained in [my previous comment](https://github.com/ankitects/anki/pull/2922#issuecomment-1884359704) is useful here:

All of my words are not related to the scheduling. For scheduling, I think we still calculate the memory state from entire review history (excluding the reviews before the last first learning entry.

user1823:
My thinking:
If the user was not correctly using the buttons in the past, using the parameters from recent reviews on the past review history will not yield the correct memory states.

Also, because in SM-2, only the interval and Ease matter, the user would have tried to prevent these values from becoming too weird.

So, amongst these options (revlogs vs interval & ease), I think that interval & ease will yield more accurate results for cards with skewed rating history.

About evaluation, I previously made the following comment:

> In my opinion, the best solution is to make the evaluation exactly match the scheduler behaviour. This is because the main purpose of the evaluation is to determine the effectiveness of scheduling, which would be most accurate when both treat the revlogs in the same way.

Expertium:
> If the user was not correctly using the buttons in the past, using the parameters from recent reviews on the past review history will not yield the correct memory states.

But using "memory_state_from_sm2" is even less accurate, so we have to choose the lesser of the two evils.
Though you do make a solid argument here:
> In my opinion, the best solution is to make the evaluation exactly match the scheduler behaviour. This is because the main purpose of the evaluation is to determine the effectiveness of scheduling, which would be most accurate when both treat the revlogs in the same way.

If memory states are calculated differently during evulation and during deployment, then evaluation will produce unrealistic estimates of log-loss and RMSE.

user1823:
> But using memory_state_from_sm2 is even less accurate, so we have to choose the lesser of the two evils.

Yes, neither of the two approaches is completely reliable. However, I still think that memory_state_from_sm2 would yield better results in the case of cards with skewed review history.

But, I think that we need another opinion here.

> This might be the only "pro" of using memory_state_from_sm2 during deployment.

If we ultimately decide to use the complete revlog for inferring memory states, doing the same in the evaluation (while still filtering the revlogs during training) is easy. So, this is not actually a "pro".

Luc-Mcgrady:
I implemented it Sherlock's way. 
```rs
    // Ignore the entire card if the first learning step of the last group of
    // learning steps is before the ignore_revlogs_before date
    if let Some(idx) = last_learn_entry {
        if entries[idx].id.0 < ignore_revlogs_before.0 {
            return None;
        }
    }
```
If there's something critically wrong with this that I'm not seeing feel free to let me know.

L-M-Sherlock:
I think it's time to merge this PR. @dae

user1823:
>      // Ignore the entire card if the first learning step of the last group of
>     // learning steps is before the ignore_revlogs_before date

This can only happen when training. What happens when inferring memory states?

dae:
@L-M-Sherlock I have not had a chance to look into the iOS issue yet, and I'd like to be sure you all agree on the correct way to solve the revlog issues before it gets merged.

Luc-Mcgrady:
Something I still don't like about this is the total removal of cards which have learn steps before the learn date.

Say we have a deck with cards which have reviews of all exactly

```
| = ignore before

L L L L R R R | R R R
1 1 3 3 2 2 2 | 1 1 3
```

This deck couldn't use the ignore before because it would simply ignore all their cards. Would it be too much to ask people to hit "forget" on all their cards to generate more learning steps if changing their review habits? If that's the method we're doing that fact should probably be specified in the advanced help menu to avoid confusion in such a situation.

> This can only happen when training. What happens when inferring memory states?

I'll fix it so that it uses the first revlog as the ignore point if there's no learning steps.

Expertium:
> Would it be too much to ask people to hit "forget" on all their cards to generate more learning steps if changing their review habits?

Ideally, we want this feature to be self-sufficient, in the sense that the user doesn't have to do anything _other_ than selecting a date.

user1823:
> Would changing it so that it uses the first revlog as the ignore point if there's no learning steps work?

I don't understand what you mean by this.

My point is that you can ignore all the revlogs when training but not when scheduling (otherwise how will you calculate the intervals?)

So, the approach used by you is desirable during training.

But when inferring memory states, we have to choose one of the two options:
- keep the entire revlog (i.e. don't filter anything)
- use my suggestion (https://github.com/ankitects/anki/pull/2922#issuecomment-1884359704).

The following comment explains why I think that the second approach is better.
https://github.com/ankitects/anki/pull/2922#issuecomment-1900687504

But, before you make the changes, I request @L-M-Sherlock to give his opinion on which is better (revlog or memory_states_from_sm2) in case of cards with skewed rating history.

Expertium:
> For scheduling, I think we still calculate the memory state from entire review history (excluding the reviews before the first Learn entry in the last sequence of Learn entries.

This is what LMSherlock said last time.

user1823:
> This is what LMSherlock said last time.

But, I had not explained my point till then. Let's wait for him to respond.

Luc-Mcgrady:
> > Would changing it so that it uses the first revlog as the ignore point if there's no learning steps work?
> 
> I don't understand what you mean by this.
> 
> My point is that you can ignore all the revlogs when training but not when scheduling (otherwise how will you calculate the intervals?)

What I thought you meant is the fact that incomplete revlogs will be kept regardless of when it was reviewed because there will be no learning step to compare against (Which is probably another problem).

Yes that would be another problem if there were no learning steps, another reason why people using this feature would be forced to hit the forget button on cards if they want to use it properly.

user1823:
> What I thought you meant is the fact that incomplete revlogs will be kept regardless of when it was reviewed because there will be no learning step to compare against

This is not a problem because the current (main branch) code removes all revlogs of cards that don't have any learn steps during training. Obviously, this holds true only if the changes made in this PR haven't broken that part of the code.

L-M-Sherlock:
My opinion here is the code has become more and more complicated. @user1823's method seems better but the current method is simple.

Expertium:
To be honest, I don't fully understand all the details here, but [this solution by user1823](https://github.com/ankitects/anki/pull/2922#issuecomment-1884359704) sounds good to me. I would take complex code that produces accurate parameters and memory states over simple code that makes FSRS inaccurate. Also, complex code is a problem for a handful of devs, inaccurate FSRS would be a problem for everyone who uses it.

Luc-Mcgrady:
Would it help if in the case there were no learn steps after the ignore date, the first relearning step was used in its place?

```
L = Learning, R = Review, r = Relearning,

L L R R | R R r r R R r r R R ->
        |     r r R R r r R R

L L R R | R R r r R R L L R R ->
        |             L L R R

L L R R | R R R R R R ->
Ignored
```

This would prevent people from having to reset their cards. 

@user1823's solution would be to have it remove the cards while training but truncate the revlogs while calculating memory states if I'm understanding it correctly? Would it not confuse FSRS if it trained from a different data than it uses?

user1823:
> if in the case there were no learn steps after the ignore date, the first relearning step was used in its place?

Post-lapse stability ≠ initial stability.
So, this approach is against the basic theory of FSRS.

If the user has introduced a sufficient number of new cards after the start_date, then there is no problem. But if not, there are two options:
- use the default parameters
- use your approach (it's inherently inaccurate, but we don't have many choices)

However, at the same time, implementing your approach while preventing adverse effects on users who have introduced sufficient number of new cards after the start_date sounds like going into the area where code complexity increases greatly with minimal benefits.

Still, if you want to implement it and other developers don't have any issue with it, there should be safeguards to ensure that relearn is treated as learn only when
- the total number of reviews after the start_date is large enough (say, more than 10,000); and
- the number of first learn entries after the start_date is low enough (say, less than 200).

> @user1823's solution would be to have it remove the cards while training but truncate the revlogs while calculating memory states if I'm understanding it correctly? 

Yes, you got it right.

> Would it not confuse FSRS if it trained from different data than it uses for memory states?

Firstly, the current version (main branch) deals with the incomplete revlogs in the same way (remove the card during training, use the incomplete revlogs when calculating memory states).

As you said, it is not ideal. But, we don't have any other choice.
- Incomplete revlogs can't be used when training. Using them for training would yield rubbish parameters.
- We need some way to calculate memory states for all cards during scheduling. Otherwise, we can't calculate the intervals.

user1823:
You might think what's the benefit of doing all this if the user has to use the default parameters in the end. Let me explain:
- The users who have sufficient new cards introduced after the start_date will be able to train the parameters. So, the situation would be much better for them than the current one.
- The users who haven't introduced sufficient new cards will still benefit from this, even if they have to use the default parameters.
   The explanation is quite simple: The scheduler will now ignore their past revlogs where they used the buttons incorrectly. So, the memory states would be calculated more accurately than before.

Also, in the future, L-M-Sherlock may offer different sets of parameters for different types of materials/ learners, etc. I am not promising this, but I feel that it can happen at some point of time.
If/when this happens, the affected users can get even more accurate memory states. So, the downside of not being able to train their own weights would not be that great.

Expertium:
> * So, the memory states would be calculated more accurately than before.

That is assuming that calculating memory states from SM2 is accurate, and I doubt it.

user1823:
> Ideally, we should somehow test whether calculating memory states from SM2 is better than using the full review history with improper button usage.

It is easy to test once my suggestion (https://github.com/ankitects/anki/pull/2922#issuecomment-1884359704) is implemented. Just evaluate with start_date unfilled and filled using the same weights in both cases.

If it turns out that for **some** users, using the complete revlog is better than SM-2, they can train after filling the start_date and clear the start_date before clicking `Save` so that the complete revlog is used to calculate the memory states.

Expertium:
> If it turns out that for some users, using the complete revlog is better than SM-2, they can train after filling the start_date and clear the start_date before clicking Save so that the complete revlog is used to calculate the memory states.

That's a bit complicated, though. @L-M-Sherlock so what will be your final decision? Should [user1823's](https://github.com/ankitects/anki/pull/2922#issuecomment-1884359704) suggestion be implemented or not? We've spent a lot of time discussing this, and I think it's time to make the final call.

bonedoc1:
>   The explanation is quite simple: The scheduler will now ignore their past revlogs where they used the buttons incorrectly. So, the memory states would be calculated more accurately than before.
> 
Just want to add my two cents. As one of the (many it seems) people whose intervals become way too long, this seems like it  would be a fix for that issue. if this is caused by the "hard" button, maybe i used it incorrectly when i initially started Anki a few years ago. I truly can't recall. But I definitely have not used it incorrectly over the past 9 months when i started to do anki religiously again. So if "ignore reviews before ***" was implemented and this fixed the very long intervals issue, that would be fantastic. A lot of people like myself use a large deck (e.g. AnKing) for years so if the reviews you did in the beginning of the deck weren't done properly, it sucks that detrimentally effects your current/future optimization. Seems like people are constantly posting about this issue. My true retention on "deck life" is 98.5% (which seems like the cause of the issue and may be my own fault) and the intervals get pushed out to 4+ years. However, over the past month I've been trying to solve this issue myself and my true retention is 89.7%. I'm happy to beta test this feature out on my deck or do whatever I can to help.

Sorry if this isnt the appropriate place for this, I just wanted to add my thoughts. You guys have obviously much more knowledge on this topic and it's very much appreciated that you're all working very hard on this. Thanks!

L-M-Sherlock:
> so what will be your final decision

I leave it to @Luc-Mcgrady if he is available to implement the complex version.

ts/deck-options/DateInput.svelte
magicgoose: perhaps it could be just a text input accepting a simple format like today is 2024-01-29

Expertium: Text input is too prone to user error

ts/deck-options/FsrsOptions.svelte
```diff
@@ -86,6 +87,14 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         }
     }
 
+    function getignoreRevlogsBeforeMs() {
```
dae, line 90:
```suggestion
    function getIgnoreRevlogsBeforeMs() {
```

dae:
Ok, regarding the iPhone styling, the following makes it look better:

```
width: 100%;
-webkit-appearance: none;
height: 1.5em;
```

The other issue is that it's just blank until a date is selected. One way we could work around this is to set a 1970-1-1 date by default instead.

rslib/src/scheduler/fsrs/weights.rs
```diff
@@ -299,10 +336,14 @@ pub(crate) mod tests {
 
     const NEXT_DAY_AT: TimestampSecs = TimestampSecs(86400 * 100);
 
+    pub(crate) fn days_ago_ms(days_ago: i64) -> TimestampMillis {
```
dae, line 339: If this and the method below don't need to be used in other files, the pub(crate) is unnecessary.

dae (commented):
Some minor points below. Once they and the conflicts are addressed, and the iOS styling fixed, I presume you guys are happy to have this merged in?

user1823:
> The other issue is that it's just blank until a date is selected. One way we could work around this is to set a 1970-1-1 date by default instead.

I don't see any issue with the box being blank until the user selects a date. It simply means that the option is disabled. But I am fine with any of the two approaches.

> I presume you guys are happy to have this merged in?

Based on the comment history, I presume that this PR is removing old revlogs during training but not during scheduling and evaluation. I am fine with this approach, but I think that it would have been better to remove the old revlogs during scheduling and evaluation too.

If @Luc-Mcgrady doesn't want to or doesn't have the time to implement my idea (https://github.com/ankitects/anki/pull/2922#issuecomment-1884359704), you can merge this after the highlighted minor issues and conflicts are addressed.

user1823:
> @user1823 Is this what your idea would be?

If the card has entries like `L L | L R R`, the proposed code would schedule the card incorrectly because the learn rating after the start_date would be taken as the first rating.

I have proposed the sequence (https://github.com/ankitects/anki/pull/2922#issuecomment-1884359704) after giving it a lot of thought. Variations from that sequence would likely result in unexpected behaviour.

dae:
> I don't see any issue with the box being blank until the user selects a date. It simply means that the option is disabled. But I am fine with any of the two approaches.

On iOS it's just a blank grey rectangle until a date has been set, which looks more like a rendering error than a tappable button, so showing 1970-01-01 would be preferable.

user1823:
This seems to be a good solution.

If you can skip removing the whole learning group and instead use the learn revlogs (after ignore_date) for converting to memory states, we would accomplish what I wanted to. 

This can possibly done by using `revlogs_complete = false;` or `last_learn_entry = None;`.

In effect, my idea will cause the following conversion (but not explicitly)
```
L | L L R R ->
  | R R R R
```

If it is not easy to do, I approve your idea (https://github.com/ankitects/anki/pull/2922#issuecomment-1937022539).

Expertium:
@L-M-Sherlock @user1823 I know that this is a bit too late at this point, but I have an idea for a radically different approach: when calculating the logloss, just assign 0 weight to reviews that happened before a certain date.
We'll have to modify how logloss is calculated. It would be like this:
`loss = -(y * torch.log(y_pred) + (1 - y) * torch.log(1 - y_pred)) * weight`
Weight is 1 if the review happened after the date, and 0 otherwise. This is much more elegant than the current solution. No need to use SM-2, no need to create sophisticated filters. Just make it so that reviews before a certain date do not affect the loss function.

user1823:
> This is much more elegant than the current solution.

Though I don't understand the code well, I don't think that it would be that simple to implement as it seems. It would still need a change in several lines of code. It will also require a change in the fsrs-rs code.

> Just make it so that reviews before a certain date do not affect the loss function.

After the suggested change, they won't affect the loss directly but they will still affect it indirectly (incorrect memory states in earlier reviews → incorrect memory states in later reviews → high loss → optimizer tries to change the parameters → more correct memory states in earlier reviews → more correct memory states in later reviews → smaller loss)

Overall, I don't think that we will benefit much (if at all) by switching to that approach.

Luc-Mcgrady:
I added the default date, It makes the date-picker open in 1970 making the date harder to select but that should be fine.

@user1823 I've made some significant changes to the way your method is implemented and would like to know I haven't broken/forgot anything in the process.

user1823:
>  I've made some significant changes to the way your method is implemented

Is it different from what you proposed in https://github.com/ankitects/anki/pull/2922#issuecomment-1937022539? I am unable to see the difference.

Luc-Mcgrady:
> I am unable to see the difference.

I made the changes in https://github.com/ankitects/anki/pull/2922/commits/4c3894eaef512bb0af447249dd2b182ff5a3ecaf.

```rs
 if training {
        // While training ignore the entire card if the first learning step of the last
        // group of learning steps is before the ignore_revlogs_before date
        if let Some(idx) = first_of_last_learn_entries {
            if entries[idx].id.0 < ignore_revlogs_before.0 {
                return None;
            }
        }
    } else {
        // While reviewing if the first learning step is before the ignore date,
        // ignore every review before and including the last learning step
        if let Some(idx) = first_of_last_learn_entries {
            if entries[idx].id.0 < ignore_revlogs_before.0 && idx < entries.len() - 1 {
                    .iter()
                    .enumerate()
                    .rev()
                    .map(|(idx, _)| idx);

                entries.drain(..(last_learn_entry? + 1));
                revlogs_complete = false;
                first_of_last_learn_entries = None;
            }
        }
    }
```

Functionally it should be the same.

user1823:
I can't verify whether the code works as described in the code comments or not. But, if it works as described, it looks good to me.

dae:
Thanks for all the work you've put into this Luc. Once the merge conflicts are addressed, I'll merge this in.

user1823:
@bonedoc1, you can test this feature in the beta release: https://forums.ankiweb.net/t/anki-24-04-beta/41792

PR #2924, opened 2024-01-01
Initialize Browser.aspect_ratio earlier to fix AttributeError

abdnh:
https://forums.ankiweb.net/t/weirdness-when-esc-pressed-in-full-screen-browse-window/39008


dae:
Thanks Abdo!

PR #2928, opened 2024-01-03
Fix toolbar button shifting on hover

BlueGreenMagick:
Fixes toolbar buttons (Get Shared, Create Deck, Import File) shifting on hover because border width changes from 0.5px to 1px.

Fixed by universally increasing button border width to 1px. I'm not sure if `border: 0.5px` line is even applied anywhere other than the toolbar buttons as all other buttons borders seems to be modified using class selectors.


dae:
Thanks!

PR #2929, opened 2024-01-04
Ignore some extra chars in no-combining search

### LINKED ISSUE #2926: Our no-combining search does not ignore all diacritics
https://forums.ankiweb.net/t/search-using-nc-does-not-find-polish-diacritic-letter-l-or-croatian-diacritic-letter-d/39034

abdnh:
Closes #2926

I just took the translation list in  [WordPress page](https://developer.wordpress.org/reference/functions/remove_accents/) and kept only letters that don't change after NFKD normalization.

### DIFF
--- a/rslib/src/text.rs
+++ b/rslib/src/text.rs
@@ -380,22 +380,61 @@ pub(crate) fn ensure_string_in_nfc(s: &mut String) {
     }
 }
 
+static EXTRA_NO_COMBINING_REPLACEMENTS: phf::Map<char, &str> = phf::phf_map! {
+'€'  =>  "E",
+'Æ'  =>  "AE",
+'Ð'  =>  "D",
+'Ø'  =>  "O",
+'Þ'  =>  "TH",
+'ß'  =>  "s",
+'æ'  =>  "ae",
+'ð'  =>  "d",
+'ø'  =>  "o",
+'þ'  =>  "th",
+'Đ'  =>  "D",
+'đ'  =>  "d",
+'Ħ'  =>  "H",
+'ħ'  =>  "h",
+'ı'  =>  "i",
+'ĸ'  =>  "k",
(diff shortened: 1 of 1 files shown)


rslib/src/text.rs
```diff
@@ -380,22 +380,61 @@ pub(crate) fn ensure_string_in_nfc(s: &mut String) {
...
 
     // we need to create a new string without the combining marks
-    s.chars()
+    let mut s = s
+        .chars()
         .nfkd()
         .filter(|c| !is_combining_mark(*c))
-        .collect::<String>()
-        .into()
+        .collect::<String>();
```
dae, line 431: Use of phf is nice! I think the step below could be done more efficiently though, by using .map() here to transform the unwanted characters, instead of modifying the string below. You should be able to look up each character instead of having to iterate over the phf, too.

dae: (might be more efficient to do this in a 'for char in ... { output.push(char)  or output.push_str(repl) }' than using flat_map)

rslib/src/text.rs
```diff
@@ -424,17 +424,16 @@ pub(crate) fn without_combining(s: &str) -> Cow<str> {
...
     // we need to create a new string without the combining marks
-    let mut s = s
-        .chars()
-        .nfkd()
-        .filter(|c| !is_combining_mark(*c))
-        .collect::<String>();
-
-    for (k, v) in EXTRA_NO_COMBINING_REPLACEMENTS.entries() {
-        s = s.replace(|c| c == *k, v);
+    let mut out = String::new();
```
dae, line 427: Minor optimization that should avoid allocations in the common case:

```suggestion
    let mut out = String::with_capacity(s.len());
```

dae:
Thanks Abdo!

PR #2930, opened 2024-01-06
Flip indent/outdent buttons for RTL

### LINKED ISSUE #2917: Some minor RTL issues
https://forums.ankiweb.net/t/some-rtl-and-translate-issues/38892

abdnh:
Closes #2917

The ideal solution would be to make the buttons match the computed direction of the list, but it's probably not worth bothering with that.


dae:
Works for me; thanks Abdo.

PR #2931, opened 2024-01-06
Fix undo handling of group and some other IO tools

abdnh:
This should fix issues 6 and 7 reported by @hnvy in [#2774](https://github.com/ankitects/anki/issues/2774#issuecomment-1780133665)

### DIFF
--- a/ts/image-occlusion/tools/tool-undo-redo.ts
+++ b/ts/image-occlusion/tools/tool-undo-redo.ts
@@ -19,7 +19,7 @@ type UndoState = {
     redoable: boolean;
 };
 
-const shapeType = ["rect", "ellipse", "i-text"];
+const shapeType = ["rect", "ellipse", "i-text", "group"];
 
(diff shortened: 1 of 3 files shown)


ts/image-occlusion/tools/tool-undo-redo.ts
```diff
@@ -49,7 +49,12 @@ class UndoStack {
     setCanvas(canvas: fabric.Canvas): void {
         this.canvas = canvas;
         this.canvas.on("object:modified", (opts) => this.maybePush(opts));
-        this.canvas.on("object:removed", (opts) => this.maybePush(opts));
+        this.canvas.on("object:removed", (opts) => {
+            // `destroyed` is a custom property set on groups in the ungrouping routine to avoid adding a spurious undo entry
```
dae, line 53: The comment here was a good addition

dae (commented):
Thanks for looking into this Abdo!

PR #2932, opened 2024-01-08
Simplifying IO image centre terminology

hnvy:
Hello,

I hope this finds you well.

I was teaching the new Anki IO features to my mentees (some of which are not native English speakers, using English Anki), and I wanted to pass this feedback to you. For some reason, the students found the terminology to be confusing.

Some students expected the terminology to be switched (their thinking was that the "vertical" is referring to the "Y-axis of the **_canvas_**")

image

To be honest, I too found it difficult to wrap my head around "Align Vertical/Horizontal Center". Perhaps this type of terminology is easy for experts?

I have also updated the shortcuts.

I thought to do a PR since it was simple enough to change.

If there is anything else that needs doing, please let me know.

I used Microsoft PowerPoint as a source, since it is probably one of the most used tools out there. Please see the GIF below:

PowerPoint-Alignment


dae:
A related previous change: https://github.com/ankitects/anki/pull/2815

Personally, I'm not sure the new wording is much clearer :-). "Align center" is aligning the middle point of each shape in a vertical direction, and vice versa, so you could argue these are inverted as well. I guess it's intended to convey "align shapes along the middle of the canvas"? 

I don't have a strong attachment to the current wording, and can see how it could be confusing too, but if we change this again, I'd like it to be the last time :-) As alternatives to align center, what about "align middle vertically" or "align vertically", and align middle -> align horizontally?

@snowtimeglass any thoughts?

hnvy:
Oh! I see this has been discussed before.

> Personally, I'm not sure the new wording is much clearer

Well, the reason I chose it is because PowerPoint (which is probably used more than Figma?) was using it. But, I do agree.

Interestingly, other tools also use PowerPoint terminology: https://www.biorender.com/learn/4-ways-to-align-objects-in-biorender and https://help.venngage.com/hc/en-us/articles/11017636940429-Aligning-elements-on-the-canvas

So, my thinking was: since this is popular, users will be acquainted with it already (despite its imperfection).

Another one I thought of was to make it clear which axis is Anki referring to: "align to vertical/horizontal axis of canvas"?

> if we change this again, I'd like it to be the last time

I can close this PR! It's not a major issue. The icons speak for themselves. My aim was to reduce workload, not increase it!

dae:
> Oh! I see this has been discussed before. I can close this PR!

I'm happy to entertain suggestions, just hoping we can come up with a change that won't need changing again :-)

PR #2933, opened 2024-01-09
Use card.reps - 1 when calculating fuzz

dae:
https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/343#issuecomment-1879584562

https://forums.ankiweb.net/t/reschedule-is-inconsistent-with-normal-schedule-in-fuzz/39363


L-M-Sherlock:
Yeah! I test it and it works well.

PR #2936, opened 2024-01-10
Fix importing of Mnemosyne due dates

dae:
https://forums.ankiweb.net/t/please-help-anki-making-me-restudy-all-5k-imported-cards/39309

### DIFF
--- a/rslib/src/import_export/text/import.rs
+++ b/rslib/src/import_export/text/import.rs
@@ -26,6 +26,7 @@ use crate::notetype::CardTemplate;
 use crate::notetype::NoteField;
 use crate::prelude::*;
 use crate::progress::ThrottlingProgressHandler;
+use crate::scheduler::timing::SchedTimingToday;
 use crate::text::strip_html_preserving_media_filenames;
 
 impl ForeignData {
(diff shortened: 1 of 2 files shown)


dae:
@RumovZ look ok to you?

rslib/src/import_export/text/import.rs
```diff
@@ -625,10 +632,8 @@ impl ForeignCard {
         }
     }
 
-    fn native_due(self, today: u32) -> i32 {
-        let remaining_secs = self.interval as i64 - TimestampSecs::now().0;
-        let remaining_days = remaining_secs / (60 * 60 * 24);
-        0.max(remaining_days as i32 + today as i32)
+    fn native_due(self, timing: &SchedTimingToday) -> i32 {
+        (self.due - timing.next_day_at.0 as i32) / 86_400
```
RumovZ, line 636: Don't we still need to adjust by `timing.days_elapsed`? And wouldn't it be more correct to round up?
```suggestion
        (self.due - timing.next_day_at.0 as i32 + 86_400 - 1) / 86_400 + timing.days_elapsed as i32
```
(Code not checked.)

dae: :man_facepalming: thanks Rumo, silly oversights on my part.

PR #2940, opened 2024-01-10
Fix note id mix-up when resolving notetype conflicts

RumovZ:


dae:
Thanks for the quick fix Rumo!

PR #2941, opened 2024-01-11
Make tags and tag suggestions able to be styled by add-ons

glutanimate:
I would like to extend [Colorful Tags](https://github.com/glutanimate/colorful-tags) with the ability to also render user-selected tag colors in the editor. Right now this is not really possible as there is no stable & performant way to target individual tags and suggestion components.

This PR introduces two new data attributes, `data-suggestion` and `data-tag-name` that are applied to the corresponding elements, making them very easy to target via CSS or JS.

A couple of thoughts:

- I was thinking of perhaps renaming the attributes to use a common `data-addon` prefix, e.g. `data-addon-tag` and `data-addon-suggestion`. That might be a more expressive way of highlighting that these are consumed by add-ons while also making it easier to grep for add-ons that might be accessing the attributes in the future. WDYT?
- Do you think it would be prudent to escape some characters in the tag content? I wasn't able to find any Svelte-specific recommendations here. Given that Svelte seems to render the attributes double-quoted, it seems like replacing potentially ambiguous `&` and `"` would be the [spec-conformant way](https://html.spec.whatwg.org/multipage/syntax.html#syntax-attributes). However, it doesn't look like we're doing this anywhere else and in some places we already are setting some attributes to values that could contain both `"`  and `&` (e.g. `button.title` on `Tag.svelte`).

### DIFF
--- a/ts/tag-editor/AutocompleteItem.svelte
+++ b/ts/tag-editor/AutocompleteItem.svelte
@@ -22,10 +22,10 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 <div
     bind:this={buttonRef}
     tabindex="-1"
-    data-suggestion={suggestion}
     class="autocomplete-item"
     class:selected
     class:active
+    data-addon-suggestion={suggestion}
     on:mousedown|preventDefault
     on:mouseup
     on:mouseenter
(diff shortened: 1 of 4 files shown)


dae:
- data-addon-* seems reasonable, but I don't have a strong preference
- I'm guessing that because Svelte sets the attributes via Javascript instead of raw HTML, ampersands/etc probably won't break, so I don't have much of a preference here either.

glutanimate:
Cheers, I switched to `data-addon-*` and, having confirmed that `&` and `"` work, will leave the attribute handling as is then to remain consistent with other code.

Issue #2923, opened 2024-01-01
Adjust _agent_name() to look like Chrome


dae:
This will increase compatibility when downloading images, e.g. https://forums.ankiweb.net/t/errors-when-pasting-certain-images/38732/3. Currently we fall back to remote links when a download fails, but they're likely to break in the future, and they are slower.

vaxr:
I have done a little digging using the two sample images from the forum thread. They fail for different reasons, neither of which looks like it would be fixed by another user agent:

## Image 1
<https://o.quizlet.com/LbvqR76u3UVrFNFPpF7gqA.jpg>

This image downloads fine using `curl` as long as any user agent is specified. The Anki user agent works fine. I have tried to replicate the failing (403) Anki call using `requests`, but couldn't get it to fail:

```python
# Python 3.9
import requests

s = requests.Session()
url='https://o.quizlet.com/LbvqR76u3UVrFNFPpF7gqA.jpg'
headers = {'User-Agent': 'Anki 23.12.1'}

response = s.get(url, stream=True, headers=headers, timeout=30, verify=True)

print(response.status_code)
print(response.headers['Content-Type'])
# --> 200
# --> image/jpeg
```

Replicating Anki's exact request headers in curl didn't fail either.

## Image 2
<https://dr282zn36sxxg.cloudfront.net/datastreams/f-d%3A05e1465a992f7f2754fb257864af8398f6e7ae94ecc4af9ffebf00fa%2BIMAGE_THUMB_POSTCARD_TINY%2BIMAGE_THUMB_POSTCARD_TINY.1>

Before the image URL is fetched, urllib unquotes it:

https://github.com/ankitects/anki/blob/4ccb814fb3da5cf291b3962c1eddced07ecee4f4/qt/aqt/editor.py#L912-L916

Resulting in a [different URL that the server doesn't recognize](https://dr282zn36sxxg.cloudfront.net/datastreams/f-d:05e1465a992f7f2754fb257864af8398f6e7ae94ecc4af9ffebf00fa+IMAGE_THUMB_POSTCARD_TINY+IMAGE_THUMB_POSTCARD_TINY.1).

dae:
Thanks for digging into these. For image 1, when I test with the debug console:

- I get a 403 error
- When user-agent is not set, or is set to a Chrome one, I still get a 403

Maybe they are filtering based on IP, or some other header?

For image 2, hmm, that's troublesome - it would be nice to fix, if we can do so in a way that doesn't break other links.

[...]

PR #2943, opened 2024-01-12
Don't urllib.unquote when using requests

vaxr:
See #2923

Please review carefully before merging. I have only tested this manually and superficially.

### Motivation
- The `urllib.parse.unquote(url)` line causes problems with downloading some images, e.g. Image 2 from the linked issue. In this case because the server expects a quoted URL.

### Rationale
- The comment explains that the unquoting is because of limitations in `urllib`.
- Probably more recently, remote downloading is done through `requests` instead of `urllib`, while `urllib` remains in use for local downloads.
- `requests` probably doesn't have any such limiation. At least it works fine on first glance.
- The solution is to pass `url` unmodified to `requests`, and unquoted to `urllib`.
- The unquote call is consequently only performed inside the local download block.
- There is already a second unquote call down the line before writing to file system, so it should only make a difference for the download step.
- Checking for `startswith("file://")` before unquoting should be fine, because that part should never be quoted.

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -911,21 +911,18 @@ require("anki/ui").loaded.then(() => require("anki/NoteEditor").instances[0].too
 
     def _retrieveURL(self, url: str) -> str | None:
         "Download file into media folder and return local filename or None."
-        # urllib doesn't understand percent-escaped utf8, but requires things like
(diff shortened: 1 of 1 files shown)


dae:
Your approach looks sound. Thanks for looking into this!

PR #2948, opened 2024-01-14
 feat: support fld:in: fld:not:/fld:ex: fld:re:

qxo:
fld:in: means include in the field ie: fld:in:vote,god or fld:in:vote|god or fld:in:vote;god
    fld:not:/fld:ex: means exclude in the field ie: fld:not:vote,god  or fld:ex:vote|god or fld:not:vote;god
    fld:re: means regexp way match, ie: fld:re:^v[a-z]+

### DIFF
--- a/proto/anki/search.proto
+++ b/proto/anki/search.proto
@@ -102,6 +102,7 @@ message SearchNode {
     uint32 introduced_in_days = 19;
     Field field = 20;
     string literal_text = 21;
+    string fld = 22;
   }
 }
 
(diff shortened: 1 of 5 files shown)


abdnh:
- Note that you can use the `fieldName:text` syntax to do something similar, e.g `*:foo OR *:bar`, but that applies to all fields, not only sort fields as in your implementation.
- I think using `fld` as a name for this can be confusing.

qxo:
> * Note that you can use the `fieldName:text` syntax to do something similar, e.g `*:foo OR *:bar`, but that applies to all fields, not only sort fields as in your implementation.
> * I think using `fld` as a name for this can be confusing.

rename `fld`  to `sfld`  ?

dae:
I've never seen someone request this before, so I'm not sure it would be used enough to justify inclusion I'm afraid.

PR #2950, opened 2024-01-14
Default to 'Bury Card' action consistently

abdnh:
I apparently couldn't make my up mind about what the default should be, so I used 'Answer Good' in some places.


dae:
Been there before :-) Thanks Abdo.

PR #2953, opened 2024-01-15
Fix MathJax editor not opening when navigating into blocks

### LINKED ISSUE #2914: Can't navigate into MathJax block elements with keyboard
https://forums.ankiweb.net/t/the-cursor-doesnt-enter-into-mathjax-blocks-when-editing/38883

2.1.66 behaved the same - this doesn't appear to be a regression.

abdnh:
Closes #2914

This was added in #1547 and appears to be intentional but removing it doesn't seem to cause any issues.


dae:
Thanks Abdo!

Issue #2952, opened 2024-01-15
Future due graph misses cards due on the last day


dae:
https://forums.ankiweb.net/t/lifetime-prognosis-doesnt-show-card-furthest-into-the-future/39649/

PR #2955, opened 2024-01-18
Fixed future-due graph not showing last cards due on last day

49Indium:
This is a small change to solve #2952.

The D3 `bin().thresholds` function was given an array, with the last value being the last day. However, the last bin collects everything smaller than (but not equal) to this last value (see the documentation [here](https://d3js.org/d3-array/bin#bin_thresholds)). Instead of calling `ticks` to provide the spacing, this patch provides the desired number of bars directly, which `thresholds` uses to make that number of equally sized bins.

I'm new to contributing, so please let me know if I've missed anything.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -155,6 +155,7 @@ Gustavo Sales <gustavosmendes14@gmail.com>
 Shawn M Moore <https://github.com/sartak>
 Marko Sisovic <msisovic13@gmail.com>
 Viktor Ricci <ricci@primateer.de>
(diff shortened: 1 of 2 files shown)


dae:
Thanks for looking into this! Unfortunately, this approach appears to be broken. On the deck I tested it with, when I set the future due graph to "all time", I get "3 days ago, 11 cards due". With the main branch, it's 2 cards instead, and clicking on it reflects that in the browser. My guess is that the bins are larger than shown.

dae:
Please create a new PR if you're able to address the mentioned issue.

PR #2958, opened 2024-01-22
Fix flicker when hovering over deck list

### LINKED ISSUE #2908: Avoid animating hover changes in webviews
Currently we animate hover transitions in the default interface, and these animations perform quite poorly on some video drivers. This makes the interface feel laggy/clunky, and is a bad out-of-box experience. Compare the standard hover:

https://github.com/ankitects/anki/assets/180542/7839197c-8ab3-482a-bb75-9952de7e47ee

With the hover performance in minimalist mode:

https://github.com/ankitects/anki/assets/180542/9ae67bf4-d253-4739-bc38-0391a3869710

We should remove the animation, but keep the color change. 

f10a29a93dc2dc66cabc91bec58d489204f53841 is semi-related, though in that case I removed the hover change altogether, instead of just removing the animation (the transition directive in the css)

abdnh:
Closes #2908

Removing `backdrop-filter` as @jcznk suggested did the trick. The most popular add-on to change the background, [Custom Background Image and Gear Icon](https://ankiweb.net/shared/info/1210908941), removes it anyway (https://github.com/AnKing-VIP/Custom-background-image-and-gear-icon/issues/55).


dae:
Thanks Abdo!

PR #2966, opened 2024-01-25
Make image occlusion's option buttons tabbable

### LINKED ISSUE #2954: Make it easier to select images for occlusion
https://forums.ankiweb.net/t/add-keyboard-shortcut-to-image-occlusion-add-card-to-select-either-of-the-two-options/39695

We could use explicit shortcuts, or we could focus the Select Image button so the user can use space or tab then space.

PLameiras:
Closes #2954

Speeds up the creation of image occlusion cards by enabling the selection of 'Select Image' or 'Paste Image from Clipboard' using the tab and space keys.


dae:
Thank you for contributing! I think this could probably be simplified to `tabbable={true}` though, without the extra variable.

dae:
Thanks Pedro!

PR #2967, opened 2024-01-27
Fix notetype conflict resolution

RumovZ:
While investigating a bug reported by Thomas (basiskarten) which led to fields and templates being incorrectly mapped in the merged notetype, I've also discovered another one which unfortunately could corrupt the user's collection.
It is only triggered when importing with merging enabled and one of the updated notes had  its notetype changed in either collection, and then undoing the import, so hopefully it hasn't caused any damage, yet.


dae:
Thanks for catching this Rumo! Thankfully there haven't been any reports of such issues at this stage.

PR #2971, opened 2024-01-28
Increase parent limits in custom study when 'limits start from top' is enabled

### LINKED ISSUE #2965: Increase parent limits in custom study when 'limits start from top' is enabled
https://forums.ankiweb.net/t/anki-23-12-1-small-bug-global-setting-limits-start-from-top-prevents-increasing-amount-of-new-cards-via-custom-study/40103

RumovZ:
I'm not updating `extend_new` and `extend_review` for the parents. Those were only relevant in v2, right? However, they still appear to be in use for the custom study default values.

Closes #2965.


dae:
Thanks Rumo! extend_new/review are used for the defaults, and updated in custom_study_inner().

PR #2977, opened 2024-01-29
update FSRS to v0.2.0

L-M-Sherlock:
@dae, I don't know how to deal with the translation. So I leave it for you.


rslib/src/scheduler/fsrs/weights.rs
```diff
@@ -39,12 +39,12 @@ impl Collection {
         let mut anki_progress = self.new_progress_handler::<ComputeWeightsProgress>();
         let timing = self.timing_today()?;
         let revlogs = self.revlog_for_srs(search)?;
-        if revlogs.len() < 1000 {
+        if revlogs.len() < 400 {
```
user1823, line 42: Here, 400 is the number of distinct cards and not the number of revlog entries. Right?

L-M-Sherlock: It's the number of revlog entries.

user1823: What if the user has just 100 cards with 4 entries each? Assuming two button usage and 90% retention, the number of first ratings in Again and Good will be 10 and 90. I don't think that they are sufficient to generate reasonable S0 values.

If we assume that all the buttons are used, the situation is even worse.

L-M-Sherlock: FSRS has L1 regularization for initial stability fitting. When the sample size is small, FSRS doesn't allow the initial stability deviate from the default values too much.

dae:
Thanks Jarrett! Please see my pushed commit for future reference.

PR #2979, opened 2024-01-29
Fix incorrect name in lang.py

bicolino34:
Ukrainian language started with an incorrect symbol Y (U+0059) while it should start with У (U+0423). It is now fixed to the correct title "**Українська мова**" (previously **Yкраїнська мова**)


dae:
Thanks! As the tests weren't passing, I've manually made this change.

Issue #2980, opened 2024-01-30
Freehand annotations in image occlusion


dae:
Associated forum thread: https://forums.ankiweb.net/t/urgent-feature-request-drawing-and-highlighting-in-image-occlusion-mode/40309

It requests the ability to add annotations to images, such as:

- freehand drawing of arrows & text
- highlighting portions of the image

The latter could perhaps be done more efficiently with semi-transparent rectangles, but the former will probably require serializing the canvas to a pixel-based image, perhaps using the same dimensions as the original image, and overlaid above the original image.

A semi-related issue that may make sense to consider at the same time: we have [another user](https://forums.ankiweb.net/t/labels-for-image-occlusion/40181) wanting to use various shapes as permanent additions (like the text tool), and wanting to add text labels as occlusions. 

Another semi-related issue is that currently the cloze deletion # for shapes is lost when they are edited: #2975 

@krmanik @glutanimate and others, do you have any thoughts on how these features should fit together / be implemented?

krmanik:
My implementation idea will be adding a layer above image. A button to switch between annotation mode. When annotation mode is on user have access to all shapes, but the shapes are not saved in occlusion instead as image because it available to notes in current occlusion. The layer includes freeform drawing, text and other shapes. When occlusion created, the occlusion mask added to occlusion field of note type and the annotation image generated and saved in collection media. In mask editor one more canvas added for creating annotation related data.

dae:
Another thing to consider is how the editing case will work. If we bake the shapes into the overlay image, the user will not be able to manipulate them if they later want to edit the note, and we'd need something like an eraser tool as well. It may be somewhat confusing from a UI-perspective.

Another option would be to use vector shapes/paths for the overlay as well, and not bake anything into pixel form. Freeform drawing would make use of path drawings, and the user would be able to resize/delete the path/other shapes later.

If such an approach is practical, would we even need an overlay? We could just draw the shapes like we do with the text tool instead.

In terms of layering, does it make sense to have such shapes above the image? What if the user wants to e.g. add some text to the image, and then put a mask over that text? Perhaps it would make more sense to have the order inverted, or give the user the ability to raise/lower shapes above each other?

krmanik:
> Another option would be to use vector shapes/paths for the overlay as well, and not bake anything into pixel form. Freeform drawing would make use of path drawings, and the user would be able to resize/delete the path/other shapes later.

[...]

PR #2983, opened 2024-01-31
feat: freeform drawing and annotation support to image occlusion

krmanik:
Added support for freeform drawing and annotation. The freeform data is stored in image field as svg. When editing the svg data parsed as canvas object then again saved as svg in image field. The freedraw image is second layer hidden below occlusion and above original image.

When path tool selected freedraw enabled and these data are stored as string in image field. When annotation tool, in topbar selected, all the shapes are drawn as annotation with semi-transparent color with `isAnnotation` attribute added to fabric object for that shape. Also `Path` class added to support type, it is not added to cloze occlusion field. The data are stored in image field.

I have tried to add annotation support for shapes other than path using cloze ordinal 0, it works when added but after editing some card removed (also tried `ia=1` is annotation) . @dae @abdnh, I need help on this issues.


glutanimate:
Thanks, Mani, this is looking quite nice already!

Some initial feedback, just focused on editor UX for now, while we're still discussing other parts of the implementation in #2980:

- **Tool properties**: I would recommend that we use a similar interaction pattern to Notability here and only show the properties pop-up upon clicking/tapping on the tool for the second time. I.e.: Only show the pop-up if the tool is already active and the user clicks/taps on it for a second time. That way users can quickly switch between tools without invoking the pop-up on every switch. Changing the color/brush is likely something they won't do as often as switching between tools
- **Color indicator**: I would suggest that we move this directly below the tool icon, similar to how we do it in the note editor for the text background color, i.e.: Screenshot_20240212_005344
      Right now the indicator blends too much into the button border, IMHO, and I think it would be easier to parse if there was a clear visual separation between the indicator and button border. An alternative option might be to color the tool icon itself in the selected color, but visibility in dark mode might become a problem in that case.
- **Color visibility**: The white option in the properties pop-up is quite hard to see at the moment. Perhaps we could boost visibility a bit by drawing a thin grey border around it.
- **Support for grouping**: Grouped paths are currently dropped when saving the image. I think that grouping is something that users will likely also expect to be able to do for annotations as it e.g. allows to combine multiple lines of handwritten text together and easily move it around. One case we need to guard against with this, though, is users attempting to group annotations together with masks. It would be good to prevent this in the first place.
- **Support for undoing**: I'm sure you already had this on your scope, but it would be nice to also extend the undo/redo system to annotations.

krmanik:
The other UX will be fixed in next commit, but undo redo very slow on path tool, I will check again for it.

PR #2986, opened 2024-02-02
Fix IO text size in review screen

### LINKED ISSUE #2867: I/O text is different size in review screen
Text in the review screen starts at the same top/left point, but as it's smaller than during editing, the bottom or right part of the backing box may not cover something that it did in the editing screen.

Original report: https://forums.ankiweb.net/t/text-misplaced-on-native-image-occlusion/37932

abdnh:
Closes #2867

Seems to fix it.

## Before
image

## After
image


abdnh:
Spoke too soon...
image

abdnh:
Fabric.js has its font rendering logic. We probably have to copy it to make appearance match exactly.

We are not handling mutli-line text too.
image

dae:
Ah, that would explain the differences. It would be nice to avoid a canvas/fabric.js dependency if possible. It might be worth putting this on the back burner for now, and seeing how the other changes in #2980 pan out first.

glutanimate:
Isn't the underlying issue that we aren't scaling the font size to the rendered canvas size? I'm scratching my head a bit on why I didn't notice the extent of this until now, but it definitely appears to be the case both in the editor and reviewer (this is on 23.12.1):

https://github.com/ankitects/anki/assets/5459332/4adbd65c-173d-4d11-be02-39c8a9170568

Storing a relative font size for `Text` shapes and then processing it as part of `toNormal` and `toAbsolute` might solve all positioning issues.

As for rendering multi-line text, we could [build a simple solution to split the text at newlines](https://stackoverflow.com/questions/5026961/html5-canvas-ctx-filltext-wont-do-line-breaks), or use [Canvas-Txt](https://github.com/geongeorge/Canvas-Txt/) which seems well supported, has no dependencies and would come in at 1-2kB.

abdnh:
> Isn't the underlying issue that we aren't scaling the font size to the rendered canvas size? 

I noted that in the original PR (https://github.com/ankitects/anki/pull/2705#issuecomment-1753466620) but incorrectly decided it was not necessary. 2dd83723d218893a37ca2155cfdaf294d2c853b7 stores the relative font size, but it doesn't seem to completely solve the issue.

dae:
Regarding text wrapping, could disabling it in fabric.js be an alternative solution? If the user has to manually enter linebreaks where they want them, then it should be consistent across the two screens.

abdnh:
> If the user has to manually enter linebreaks where they want them, then it should be consistent across the two screens.

.fillText() doesn't support multi-line text at all.

dae:
Oof, that's unfortunate. I guess we could ignore newlines in Fabric forcing the user to create multiple text annotations if they wanted multiple lines, though that's not very user-friendly.

krmanik:
If \n is stored in text then we can split string and draw the text in loop with given height.

dae:
Yes, that's true, though it would mean the text would be left-aligned and we'd need to ensure fabric behaved the same way.

abdnh:
Fun fact: fabric is already included in the review screen, but it doesn't appear to be working.

dae:
Hah! I'm guessing that's a transitive dependency from shapes.ts. Presumably we're not paying a runtime cost of creating a fabric object instance though?

abdnh:
I get this error when I try to create a Canvas object using `fabric.Canvas("image-occlusion-canvas")`:
```
reviewer.js:8784 Uncaught TypeError: Cannot read properties of undefined (reading 'apply')
    at Object.klass [as Canvas] (reviewer.js:8784:29)
    at <anonymous>:1:21
```

Randyblo7 (commented):
Yeah

abdnh:
I rebased this PR to only include relative font size change (5c2451344359ae28296a75213bdb9dc9c28393d9).

Fabric is now loading without errors in the reviewer (maybe after recent packaging changes). I'll look into changing review.ts to use it.

glutanimate:
> Fabric is now loading without errors in the reviewer (maybe after recent packaging changes). I'll look into changing review.ts to use it.

Hmm, not too fond of that personally. I really like how performant and minimal the current implementation is, and I would be wary of marrying card rendering to fabric.js. IMHO it feels a bit overkill to address the issue at hand. Consider that both missing font scaling and multi-line text support *should* have been fairly obvious problems. Yet it seems like very few users if any actually ran into them.

Generally speaking, I don't think that our text rendering needs to be pixel-perfect. 5c24513 seems to solve the positioning issues for the most part, only leaving minor differences in rendering. I would personally be happy with that for now, unless users see pixel differences as a problem.

As for multi-line support:

When I was playing around with this a couple of months ago, it felt quite doable for us to just extend current text drawing with newline splits. How about something like this?

```patch
diff --git a/ts/routes/image-occlusion/review.ts b/ts/routes/image-occlusion/review.ts
index 38329bc79..749716403 100644
--- a/ts/routes/image-occlusion/review.ts
+++ b/ts/routes/image-occlusion/review.ts
@@ -294,20 +294,46 @@ function drawShape({
         ctx.font = `${shape.fontSize}px ${TEXT_FONT_FAMILY}`;
         ctx.textBaseline = "top";
         ctx.scale(shape.scaleX, shape.scaleY);
-        const textMetrics = ctx.measureText(shape.text);
+
+        const lines = shape.text.split("\n");
+
+        const baseMetrics = ctx.measureText("M");
+        const fontHeight = baseMetrics.actualBoundingBoxAscent + baseMetrics.actualBoundingBoxDescent;
+        const lineHeight = 1.5 * fontHeight;
+
+        const linePositions: { text: string; x: number; y: number; width: number; height: number }[] = [];
+
+        let maxWidth = 0;
+        let totalHeight = 0;
+
+        for (let i = 0; i < lines.length; i++) {
+            const textMetrics = ctx.measureText(lines[i]);
+            linePositions.push({
+                text: lines[i],
+                x: shape.left / shape.scaleX,
+                y: shape.top / shape.scaleY + i * lineHeight,
+                width: textMetrics.width,
+                height: lineHeight,
+            });
+            if (textMetrics.width > maxWidth) {
+                maxWidth = textMetrics.width;
+            }
+            totalHeight += lineHeight;
+        }
+
         ctx.fillStyle = TEXT_BACKGROUND_COLOR;
         ctx.fillRect(
             shape.left / shape.scaleX,
             shape.top / shape.scaleY,
-            textMetrics.width + TEXT_PADDING,
-            textMetrics.actualBoundingBoxDescent + TEXT_PADDING,
+            maxWidth + TEXT_PADDING,
+            totalHeight + TEXT_PADDING,
         );
         ctx.fillStyle = "#000";
-        ctx.fillText(
-            shape.text,
-            shape.left / shape.scaleX,
-            shape.top / shape.scaleY,
-        );
+
+        for (const line of linePositions) {
+            ctx.fillText(line.text, line.x, line.y);
+        }
+
         ctx.restore();
     }
 }

```

I just put this together crudely and quickly as a POC, but it works quite well:

Screenshot_20240408_010603

Both text alignment and line height can be specified as parameters to fabric's text constructors, so we would just have to make sure to use the same defaults on both sides.

dae:
I'm inclined to agree with @glutanimate here - if the differences aren't so major that users are complaining frequently, keeping Fabric out of the reviewer means we have one less moving part that can go wrong, and less JS that needs to be loaded during reviewer startup.

abdnh:
@dae this should be ready to go.

dae:
Seems to work well - thank you both!

PR #2987, opened 2024-02-02
Maintain original IO cloze order in editing mode

### LINKED ISSUE #2975: Image occlusions are reordered when modified
https://forums.ankiweb.net/t/image-occlusion-23-12-1-cloze-order-changes-when-editing-masks/40338

abdnh:
Closes #2975

### DIFF
--- a/proto/anki/image_occlusion.proto
+++ b/proto/anki/image_occlusion.proto
@@ -66,6 +66,7 @@ message GetImageOcclusionNoteResponse {
 
   message ImageOcclusion {
     repeated ImageOcclusionShape shapes = 1;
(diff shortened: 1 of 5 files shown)


dae:
Thanks for looking into this Abdo. Unfortunately, it looks like the current approach may have a bug. I created an image occlusion with 3 rectangles in a vertical row. When I preview them, I see the top, then middle, then bottom. Then I edited the occlusion and slightly moved each shape. When I previewed them, the order had changed - the top shape was no longer the first.

abdnh:
Should be fixed now.

dae:
Thanks Abdo!

PR #2988, opened 2024-02-03
Do not reset original_due when card stays in filtered deck

### LINKED ISSUE #2938: original_due is reset when card answered in filtered deck
apply_normal_study_state() sets original_due to 0 when a card remains in a filtered deck. The v1 scheduler used original_due for relearning cards as well, which may have been why we were trying to zero it out, but this would be better done as the card leaves the filtered deck instead.

Original report about a review card transitioning to relearning: https://forums.ankiweb.net/t/error-in-future-due-chart-for-filter-deck/39460

abdnh:
Closes #2938

### DIFF
--- a/rslib/src/scheduler/answering/mod.rs
+++ b/rslib/src/scheduler/answering/mod.rs
@@ -148,7 +148,10 @@ impl CardStateUpdater {
                 match filtered {
                     FilteredState::Preview(next) => self.apply_preview_state(current, next),
                     FilteredState::Rescheduling(next) => {
-                        self.apply_normal_study_state(current, next.original_state)
+                        let revlog = self.apply_normal_study_state(current, next.original_state);
+                        self.card.original_due = self.card.due;
+
+                        revlog
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


rslib/src/scheduler/answering/mod.rs
```diff
@@ -148,7 +148,10 @@ impl CardStateUpdater {
                 match filtered {
                     FilteredState::Preview(next) => self.apply_preview_state(current, next),
                     FilteredState::Rescheduling(next) => {
-                        self.apply_normal_study_state(current, next.original_state)
+                        let revlog = self.apply_normal_study_state(current, next.original_state);
+                        self.card.original_due = self.card.due;
```
dae, line 152: I suspect I'm missing something - why is this line required? I would have thought this would already be handled when the card is moved into the filtered deck.

abdnh: [This assert](https://github.com/ankitects/anki/blob/312d396505a4806b76b74dfda6d276ec595e17ed/pylib/tests/test_schedv3.py#L718) fails without it. That's because remove_from_filtered_deck_restoring_queue() is restoring the original due date before the card was moved to the filtered deck, regardless of whether rescheduling is enabled.

dae: Ah, thanks, that makes sense.

Issue #2842, opened 2023-11-23
Editor buttons bugs and weird behaviour


Gustaf-C:
All of these where tested on macOS 11.7.10, Anki version 23.12 (24e59124), Qt6, no addons:

1. Using any of the bold, italic or underline buttons and deleting all typed text will deselect the corresponding buttons, but when text is typed they will be immediately reselected and the text will use the corresponding style.
2. Toggling the sub- or superscript button should ideally de-toggle the other one immediately. Currently the button status doesn't reflect which function is actually active until you type something.
3. Deleting all characters while sub- or superscript is active and then typing will deselect the corresponding button, but when typing something again it will set the font size but not "location". The button remains inactive.
4. Clicking on either ordered or unordered list twice will deactivate the bold, italic, underline, sub- and superscript, list and alignment buttons until something other than one of these is clicked.
5. Toggling the ordered or unordered list buttons should ideally de-toggle the other one immediately. Currently the button status doesn't reflect which function is actually active until you type something.
6. The align buttons have inconsistent corner rounding.
7. Ideally the align actually used should be selected when opening the alignment menu for the first time.
8. When switching align style the other align buttons should be deselected.
9. Once an align button has been toggled it can't be un-toggled. The actual behaviour does change though. (This is fixed if point 8 is fixed.)
10. When clicking on the currently selected field, bold, italic, sub- and superscript will be deselected. Both unordered and ordered list will however stay active when doing the same thing (I think them staying active is preferable).
11. Selecting any of bold, italic, sub- and superscript and then selecting any of the lists or changing text alignment will show the selected style as still active, however it will immediately inactivate when something is typed and the text will not use any style.

PR #2989, opened 2024-02-04
Fix corner rounding of alignment buttons in editor

lucasscharenbroch:
This resolves (6) from #2842

Before:
before
After:
after


dae:
Thanks for looking into this Lucas!

It looks like your change has introduced a minor problem: the top right (one before paperclip) and bottom left icon are no longer rounded at the end/start.

lucasscharenbroch:
Thanks Dae; I'm surprised I missed those.

Here's the updated screenshot.
after_that

dae:
Thank you Lucas! I've confirmed this looks correct in RTL languages as well.

PR #2992, opened 2024-02-07
Fix maybe_coalesce_note_undo_entry()

### LINKED ISSUE #2974: maybe_coalesce_note_undo_entry() is not working
It's failing because the current and previous ops store [note change, collection modified], not just [note change], so we can't just use .last().

https://forums.ankiweb.net/t/feature-request-undo-update-note/40374

abdnh:
Closes #2974

### DIFF
--- a/rslib/src/notes/undo.rs
+++ b/rslib/src/notes/undo.rs
@@ -88,8 +88,7 @@ impl Collection {
             .iter()
             .rev()
             .take_while(is_col_modified_change)
-            .collect_vec()
-            .len()
+            .count()
             + 1;
         let current_op_change = current_op
             .changes
--- a/rslib/src/notes/undo.rs
+++ b/rslib/src/notes/undo.rs
@@ -74,31 +74,13 @@ impl Collection {
         let Some(current_op) = self.current_undo_op() else {
             return;
         };
-        let is_col_modified_change = |change: &&UndoableChange| match change {
-            UndoableChange::Collection(col_change) => {
-                matches!(col_change, UndoableCollectionChange::Modified(_))
-            }
-            _ => false,
-        };
-
-        let changes_to_pop = current_op
-            .changes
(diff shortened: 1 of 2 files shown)


rslib/src/notes/undo.rs
```diff
@@ -64,20 +67,47 @@ impl Collection {
...
+            _ => false,
+        };
+
+        let changes_to_pop = current_op
+            .changes
+            .iter()
+            .rev()
+            .take_while(is_col_modified_change)
+            .collect_vec()
+            .len()
```
dae, line 92: This can be done more efficiently with .count() instead of .collect_vec().len()

rslib/src/notes/undo.rs
```diff
@@ -64,20 +67,47 @@ impl Collection {
...
+            + 1;
+        let current_op_change = current_op
+            .changes
+            .iter()
+            .rev()
+            .find(|c| !is_col_modified_change(c));
```
dae, line 98: While very unlikely, this could theoretically drop the entire undo entry if the operation changed multiple notes, and the last change of the two operations was for the same note. Perhaps it would make more sense to compare .first() instead? This coalescing is only intended to happen in the single-note update case.

abdnh: Hmm, will it drop the entire entry though? The code is only popping UndoableCollectionChange::Modified changes and a single UndoableNoteChange::Updated at the end of the vec. I changed that anyway.

dae: Sorry, you're right - I misread the code. Thinking about it more, it might make the code a bit simpler to explicitly match/clear the entire list of changes. I've pushed a tweak that does this - does it look ok to you? Apologies for the changing requirements :-)

abdnh: Looking better to me!

PR #2994, opened 2024-02-08
Add log-in button to preferences screen

### LINKED ISSUE #2990: Add tip to prefs screen when syncing not set up
Maybe something like "Syncing not set up. Use the sync button in the main screen to start syncing."?

https://forums.ankiweb.net/t/loggout-from-anki-to-login-again-after-profile-was-deleted/40683

lucasscharenbroch:
This changes the *AnkiWeb Account* box on the preferences screen when the user is not logged in.

Before:
before
After:
after

Resolves #2990


qt/aqt/forms/preferences.ui
```diff
@@ -798,6 +798,22 @@
...
+          <item>
+           <widget class="QPushButton" name="syncLogin">
+            <property name="sizePolicy">
+             <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
+              <horstretch>0</horstretch>
+              <verstretch>0</verstretch>
+             </sizepolicy>
+            </property>
+            <property name="text">
+             <string notr="true">LOGIN</string>
```
dae, line 810: Instead of setting the translation in the Python code, you can name it here instead, eg sync_log_in_button.

dae:
Hi Lucas,

The change itself looks well written. A couple of comments on adding a log-in button to that screen:

- There's text immediately above it that suggests the user sync from the main screen. We might want to show one or the other but not both.
2024-02-08T16:10:58,507598677+10:00
- I was a bit apprehensive when I noticed you added a login button, as I've always found AnkiDroid's logging-in-doesn't-initiate-a-sync to be confusing. But I've gone back and checked, and I think the main issue there is the login screen is shown when the user taps on the sync button, so my expectation is that a sync will happen. If the user logs in from the preferences screen, I guess that's not going to be as much of a problem. But it may still be worth displaying a dialog box after a successful login, to remind the user they'll need to close Anki or click on the sync button to sync for the first time. What do you think?

@david-allison @mikehardy I probably should have mentioned this in the past

david-allison:
> I was a bit apprehensive when I noticed you added a login button, as I've always found AnkiDroid's logging-in-doesn't-initiate-a-sync to be confusing. But I've gone back and checked, and I think the main issue there is the login screen is shown when the user taps on the sync button, so my expectation is that a sync will happen. If the user logs in from the preferences screen, I guess that's not going to be as much of a problem. But it may still be worth displaying a dialog box after a successful login, to remind the user they'll need to close Anki or click on the sync button to sync for the first time. What do you think?

As of 2.17, logging in via the onboarding process initiates a (download) sync. 

I'd support the above as a UX enhancement

dae:
Your code looks good, but I'm afraid I'm not so fond of syncing being started while the preferences screen is open, as changes from other devices will be lost if the user saves their preferences after syncing. Could we either change the message to "Save changes and sync now?" + automatically save+close the dialog first, or change it from a prompt to a suggestion to the user to sync after saving?

dae:
Thanks for your work on this Lucas, I think it's a nice addition.

PR #2996, opened 2024-02-08
Keep previous FSRS parameters if they get worse when optimizing

### LINKED ISSUE #2972: Keep previous FSRS parameters if they get worse when optimizing
https://forums.ankiweb.net/t/measures-to-prevent-new-fsrs-parameters-from-being-worse-than-the-old-ones/38957

abdnh:
Closes #2972

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1794,9 +1794,9 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "0.2.0"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4938928321ed55a54cd8b90f0aa9dad79ee223aed6462cc37e83eb80b8bddb5a"
+checksum = "eece5d2325704da8667c3a796e64a3c0b046100ff758ba88f91d0dee80d6deb9"
 dependencies = [
  "burn",
  "itertools 0.12.0",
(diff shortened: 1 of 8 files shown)


user1823:
If the older parameters are better, I think that we should show a message like "You already have optimal parameters."

https://github.com/ankitects/anki/issues/2972#issuecomment-1918738285

dae:
Thanks Abdo, looks good.

@user1823 I'm not sure that's a good idea - there's a fairly high chance the user will sometimes get worse results, and such a message may make the user think there's something wrong and such behavior is not expected. @L-M-Sherlock any thoughts?

(merging as-is for now, but we can change this in a follow-up if need be)

L-M-Sherlock:
It's not a matter of the user, so I don't want to display the message. Even if we display this message, the user has nothing could do to deal with that.

user1823:
Personally, I would expect something to change if I press a button and see a progress happening. If nothing changes, I would wonder "Did the process not go to completion?"

But if both of you think that we should not add the message, then it's OK.

Expertium:
> such a message may make the user think there's something wrong and such behavior is not expected.

I don't think that "You already have optimal parameters" sounds worrying. At least I wouldn't think that something went wrong. But I would definitely suspect that something's wrong if I clicked "Optimize" and the parameters didn't change. So I'm in favor of the message. We can make a poll on ankiforums.

PR #3003, opened 2024-02-10
Improve typing `deck_due_tree()`

nix6839:
I have improved the typing for `mw.col.sched.deck_due_tree()` by using overloading to specify that it will return `None` only when `top_deck_id` is received as a `DeckId` type.


dae:
Thanks!

dae:
(sorry, the CONTRIBUTORS file has a conflict - would you mind rebasing over main?)

nix6839:
I've rebased it. Thank you so much for creating such an awesome app!

PR #3007, opened 2024-02-12
Fix circular import in legacy importer

abdnh:
https://forums.ankiweb.net/t/bug-report-python-anki-module-circular-import/41026


dae:
Thanks for the proactive fix!

PR #3008, opened 2024-02-12
Fix deck names differing in case being duplicated in CSV import

### LINKED ISSUE #2951: Deck names are duplicated when a CSV import uses a different case for the deck name
https://forums.ankiweb.net/t/importing-with-deck-column-create-many-decks-unstead-of-one/39566

abdnh:
Closes #2951

The unicase solution was trivial to implement.


dae:
Always nice when things are straightforward! Thanks Abdo.

PR #3009, opened 2024-02-13
Fix visual behavior of modal buttons in editor

lucasscharenbroch:
This resolves (8) and (5) from #2842.

Clicking on the justification/[un]ordered-list buttons now causes the corresponding variants to be updated (visually), which fixes the problem in the screenshot.

screenshot

(Both lists and all 4 justification methods all enabled at once (technically only one of each is enabled, but they all appear to be enabled, because the buttons haven't been updated)).


dae:
Thanks Lucas, it's nice that this reuses the existing state handling routines.

PR #3010, opened 2024-02-13
Tweak type-in-answer filter hack

### LINKED ISSUE #3004: Inconsistency between previewer and reviewer
For a custom card template, the ```<hr>``` between the Question and the Answer is not appearing in the reviewer though it appears in the previewer. I understand that Anki can't support every possible customization of the templates. But since the preview is correct, the reviewer is also expected to display the card correctly.

Reviewer:
image

Previewer:
image

Card: [bug.apkg.zip](https://github.com/ankitects/anki/files/14229293/bug.apkg.zip) (rename to remove .zip at the end)

Edit:
The issue is about the use of quotes in the id attribute (i.e., `<hr id=answer>` vs `<hr id="answer">`)

For a basic card, the horizontal line displays whether the quotes are used or not. But for some reason, for this card template, it is not being displayed in the reviewer when quotes are not used.

To avoid confusion, it is probably better to make things consistent.

lucasscharenbroch:
Fixes #3004

### DIFF
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -721,6 +721,7 @@ class Reviewer:
     def typeAnsAnswerFilter(self, buf: str) -> str:
         if not self.typeCorrect:
             return re.sub(self.typeAnsPat, "", buf)
+        orig = buf
         origSize = len(buf)
         buf = buf.replace("<hr id=answer>", "")
         hadHR = len(buf) != origSize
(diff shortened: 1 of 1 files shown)


dae:
Thanks Lucas. I hadn't realised this was only happening when `type:` is only included on the front side. It is indeed quite a corner case, but hopefully this fix will avoid future user surprises.

PR #3012, opened 2024-02-15
Refactor text attribute button components

lucasscharenbroch:
The code for the `BoldButton`, `ItalicButton`, `UnderlineButton`, `SuperscriptButton`, and `SubscriptButton` is almost entirely identical: I've factored out the common code into a new component: `TextAttributeButton`.

I also removed the `clearer` function from UnderlineButton.svelte - this appears to have been left behind from an older implementation of `Surrounder`.


dae:
Thanks Lucas - much nicer!

PR #3013, opened 2024-02-16
Make a deep copy in addcards/set_note

### LINKED ISSUE #3011: Copy note+discard results in original note being changed
https://forums.ankiweb.net/t/bug-creating-copy-of-note-and-then-discarding-will-destructively-edit-the-original-note/40976

lucasscharenbroch:
Fixes #3011


dae:
Thanks for looking into this!

PR #3014, opened 2024-02-19
Show message if current FSRS parameters are optimal

abdnh:
https://forums.ankiweb.net/t/poll-display-you-already-have-optimal-parameters-if-the-previous-parameters-are-kept/41024

image

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -427,6 +427,7 @@ deck-config-percent-of-reviews =
     }
 deck-config-optimizing-preset = Optimizing preset { $current_count }/{ $total_count }...
 deck-config-fsrs-must-be-enabled = FSRS must be enabled first.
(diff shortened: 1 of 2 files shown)


dae:
One minor bug: when the parameters are empty/the defaults, hitting optimize shows the message even though it filled in the data. 

Do you think maybe the tooltip would be better done as an alert()? It might make it a bit clearer that something has popped up, as it can be a bit tricky to spot if the user has a bunch of other warnings already shown on that screen.

abdnh:
> Do you think maybe the tooltip would be better done as an alert()? It might make it a bit clearer that something has popped up, as it can be a bit tricky to spot if the user has a bunch of other warnings already shown on that screen.

I used that at first but thought the message doesn't need to be that conspicuous. But since most people voted for displaying a message, maybe less people will complain about alert().

dae:
Thanks Abdo!

Issue #2999, opened 2024-02-09
Revlog minimum for FSRS optimizing is checked prior to thinning


dae:
We require at least 400 review logs to calculate custom parameters, but that figure is based on the raw available revlogs, not the available count after we do things like filter out manual rescheduling. Ideally, we'd fail if the post-filtering count is below 400. Accomplishing this efficiently will probably require tallying up the FSRSItems and their reviews after they are created.

Associated report: https://forums.ankiweb.net/t/old-reviews-before-forget-used-in-fsrs-optimisation/40819/17

PR #3019, opened 2024-02-21
Check required review count for FSRS after filtering

abdnh:
Closes https://github.com/ankitects/anki/issues/2999

### DIFF
--- a/rslib/src/scheduler/fsrs/retention.rs
+++ b/rslib/src/scheduler/fsrs/retention.rs
@@ -75,12 +75,6 @@ impl Collection {
             .col
             .storage
             .get_revlog_entries_for_searched_cards_in_card_order()?;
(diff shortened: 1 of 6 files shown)


rslib/src/scheduler/fsrs/retention.rs
```diff
@@ -75,7 +75,7 @@ impl Collection {
             .col
             .storage
             .get_revlog_entries_for_searched_cards_in_card_order()?;
-        if revlogs.len() < 1000 {
+        if revlogs.len() < 400 {
```
abdnh, line 78: I assume this was overlooked when the minimum review requirement was relaxed.

dae:
Apologies Abdo, this took a conflict with #2922. Otherwise looks good.

PR #3021, opened 2024-02-23
Fix regex skipping over all src except the last

vaxr:
See #2918

Adapting HTML_MEDIA_TAGS to allow for `>` inside '' and "" led to multiple images inside a field sometimes being disregarded and marked as "unused". This seems to have been caused by a missing lazy (?) quantifier for the regex part skipping over non-`>`.

@meliache Thanks for finding this! If you can, please check if this resolves the bug from your end.

### DIFF
--- a/rslib/src/media/check.rs
+++ b/rslib/src/media/check.rs
@@ -818,4 +818,26 @@ Unused: unused.jpg
 
         Ok(())
     }
+    #[test]
+    fn multiple_images() -> Result<()> {
+        let (_dir, _mgr, mut col) = common_setup()?;
+        let mut checker = col.media_checker()?;
+
+        let field = "<img alt='foo' src='foo-ss.jpg'><img alt='bar' src='bar-ss.jpg'>";
+        let seen = normalize_and_maybe_rename_files_helper(&mut checker, field);
+        assert!(seen.contains("foo-ss.jpg"));
+        assert!(seen.contains("bar-ss.jpg"));
+
+        let field = "<img alt=\"foo\" src=\"foo-dd.jpg\"><img alt=\"bar\" src=\"bar-dd.jpg\">";
+        let seen = normalize_and_maybe_rename_files_helper(&mut checker, field);
+        assert!(seen.contains("foo-dd.jpg"));
+        assert!(seen.contains("bar-dd.jpg"));
+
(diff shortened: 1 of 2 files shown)


vaxr:
Not sure why the build failed. Seems to be unrelated? Tests run locally.

rslib/src/media/check.rs
```diff
@@ -816,6 +816,28 @@ Unused: unused.jpg
...
+        let field = "<img alt=\"foo\" src=\"foo-dd.jpg\"><img alt=\"bar\" src=\"bar-dd.jpg\">";
+        let seen = normalize_and_maybe_rename_files_helper(&mut checker, field);
+        assert!(seen.contains("foo-dd.jpg"));
+        assert!(seen.contains("bar-dd.jpg"));
+
+        let field = "<img alt='foo' src='foo-sd.jpg'><img alt=\"bar\" src=\"bar-sd.jpg\">";
+        let seen = normalize_and_maybe_rename_files_helper(&mut checker, field);
+        assert!(seen.contains("foo-sd.jpg"));
+        assert!(seen.contains("bar-sd.jpg"));
+
```
meliache, line 840: Would it make sense to add some tests for where there's whitespace, a newline or text-characters between the image tags? Also maybe tests for when there are more than two images? For the particular bug that I saw that didn't matter, the bug was always the same. But in my mind ideally, the tests should cover more than the bug that we just fixed, but also unforeseen future bugs. On the other hand that would add a lot of boilerplate code. I'm haven't done much Anki development, so not sure what the testing guidelines here are...

vaxr: For the issue at hand, the tests added are the minimum retention test addressing this particular fix. There are three because she bug only showed up on subsequent usage of single or double quotes. That's all a retention test should do as far as I'm concerned.

I agree it would be great to have more tests in general. How detailed the tests should be is a matter of debate I suppose. Surely there are lower hanging fruits.

meliache:
> please check if this resolves the bug from your end.

Compiling finished, I can't reproduce the bug on this PR, thanks :+1:

rslib/src/media/check.rs
meliache: Thanks for explaining, that's fine with me.

dae:
Thank you @meliache for spotting the problem, and thank you @vaxr for the quick fix, complete with tests!

PR #3024, opened 2024-02-24
clamp interval before fuzz_bounds

### LINKED ISSUE #3015: Fuzz doesn’t work with short maximum interval
> I find that the maximum interval is applied after fuzz_bounds. So if the original interval is longer than maximum interval, the fuzz intervals will be longer than maximum interval, too. The current implementation will clamp the fuzz lntervals with maximum interval, so they will be equal to maximum interval. Should we apply maximum interval to the original interval before fuzz?
> 

Originally reported on https://forums.ankiweb.net/t/fuzz-doesnt-work-with-short-maximum-interval/41274

L-M-Sherlock:
close #3015

original:

```
constrained_fuzz_bounds(500.0, 2, 365) = (
    365,
    365,
)
```

current:

```
constrained_fuzz_bounds(500.0, 2, 365) = (
    345,
    365,
)
```

### DIFF
--- a/rslib/src/scheduler/states/fuzz.rs
+++ b/rslib/src/scheduler/states/fuzz.rs
@@ -165,8 +165,8 @@ mod test {
         assert_lower_middle_upper!(20.1, 3, 1000, 17, 20, 23);
 
         // respect limits and preserve uniform distribution of valid intervals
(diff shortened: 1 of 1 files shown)


dae:
Thanks @L-M-Sherlock!

PR #3025, opened 2024-02-24
Reset memory state after 'Set Due'

### LINKED ISSUE #2998: Assigned memory state after 'set due date'+optimize incorrect
If the user uses an interval like '0!' to set a card's due date, then optimizes the deck to assign new memory state, the card will receive the wrong stability.

Reported on https://forums.ankiweb.net/t/old-reviews-before-forget-used-in-fsrs-optimisation/40819/17?u=dae

abdnh:
Closes #2998

### DIFF
--- a/rslib/src/scheduler/reviews.rs
+++ b/rslib/src/scheduler/reviews.rs
@@ -54,6 +54,7 @@ impl Card {
             // if it's already set
             self.ease_factor = ease_factor;
         }
+        self.memory_state = None;
     }
 }
 


abdnh:
This didn't fix the issue... Will have another try.

abdnh:
So manual rescheduling reviews are ignored here: https://github.com/ankitects/anki/blob/68b374e65f012f12f6ba266aea293cfaa9616389/rslib/src/scheduler/fsrs/weights.rs#L310

But even if that isn't the case, it's not clear how to handle button_chosen=0 where FSRS only knows about 1-4.

dae:
I think we may need to adjust this area here:

https://github.com/ankitects/anki/blob/68b374e65f012f12f6ba266aea293cfaa9616389/rslib/src/scheduler/fsrs/memory_state.rs#L310

or

https://github.com/ankitects/anki/blob/68b374e65f012f12f6ba266aea293cfaa9616389/rslib/src/scheduler/fsrs/weights.rs#L245

Some previous discussion about the latter:  https://github.com/ankitects/anki/pull/3002#issuecomment-1937464827 that we shouldn't include set due date here, but wouldn't doing so cause revlogs

We probably can't just adjust the latter to return revlogs_complete=false, as that path also stops further iterating, and the value may be reverted by the check below for manual on the first step.

@user1823 do you have some thoughts on what the correct or at least acceptable behavior would be in this case? Currently we're returning revlogs_complete=true and using the existing revlogs (minus the final manual entry) to infer the memory state - so there's no inferring of sm2 state from the card. Since we ignore the final manual entry, the correct interval is not picked up. Maybe we could force sm2 state to be derived from the card when the final revlog entry is 'set due date'?

user1823:
Assuming that people use 0! in Set Due Date as a replacement of Forget, I think that the DSR should be reset.

However, solely relying on the Set Due Date being the last revlog entry is not going to work because after the user performs more revlog, it would no longer be the last entry.

> Maybe we could force sm2 state to be derived from the card when the final revlog entry is 'set due date'?

This would screw up the DSR of ALL the cards because of the entry added by "Reschedule cards on change".

Currently, I have only one idea but it feels more like a hack than a solution.

The idea:
If the interval of Set Due Date is less than 10 days AND is less than 20% of the previous interval, treat the entry as a Forget entry and use SM-2 on the next entry.

dae:
I'm afraid I don't like the suggestion at the end - such logic is hard to explain to users, and a real pain when troubleshooting "why did Anki do this?" questions.

> This would screw up the DSR of ALL the cards because of the entry added by "Reschedule cards on change".

I think you're right - unfortunately I don't think we have a way of distinguishing between those two cases. I think it had originally been my intention to use the logged ease factor to distinguish, but we log the FSRS difficulty when the cards have existing memory state :-(

I don't have any bright ideas at the moment. More suggestions welcome :-)

PR #3027, opened 2024-02-24
Show total review count when optimizing FSRS parameters

### LINKED ISSUE #3000: Unfiltered revlog count not clear?
Associated report: https://forums.ankiweb.net/t/old-reviews-before-forget-used-in-fsrs-optimisation/40819/16

We could perhaps switch to the post-filtered count after the approach mentioned in #2999 is implemented, though we'd only be able to show it when 100% completion is reached.

abdnh:
Closes #3000

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -424,14 +424,10 @@ deck-config-a-100-day-interval =
         [one] A 100 day interval will become { $days } day.
        *[other] A 100 day interval will become { $days } days.
     }
-deck-config-percent-of-items =  
-    { $count ->
-        [one] { $pct }% of { $count } item
-       *[other] { $pct }% of { $count } items
-    }
-deck-config-reviews = { $reviews ->
-        [one] { $reviews } review
-       *[other] { $reviews } reviews
+deck-config-percent-of-reviews =  
+    { $reviews ->
+        [one] { $pct }% of { $reviews } review
+       *[other] { $pct }% of { $reviews } reviews
     }
 deck-config-optimizing-preset = Optimizing preset { $current_count }/{ $total_count }...
 deck-config-fsrs-must-be-enabled = FSRS must be enabled first.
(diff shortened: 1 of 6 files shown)


ftl/core/deck-config.ftl
```diff
@@ -424,10 +424,14 @@ deck-config-a-100-day-interval =
         [one] A 100 day interval will become { $days } day.
        *[other] A 100 day interval will become { $days } days.
     }
-deck-config-percent-of-reviews =  
-    { $reviews ->
-        [one] { $pct }% of { $reviews } review
-       *[other] { $pct }% of { $reviews } reviews
+deck-config-percent-of-items =  
+    { $count ->
+        [one] { $pct }% of { $count } item
```
dae, line 429: Let's call them cards rather than items, because that's what they actually are. In hindsight, FSRSCard might have been a better name for them.

dae: No, ignore me. I always forget that we create multiple FSRSItems for a card.

user1823: What exactly is an FSRS item, by the way? I think that it is better to document it somewhere because the users will begin asking it that after this is merged.

dae:
Thanks Abdo.

I agree that FSRS items is a bit of a vague term, and it might be nice to avoid exposing them to end-users. Could we just display the post-filtered review count instead? When optimizing the review count is returned by fsrs_items_for_training, and in the computing memory state case, only one fsrs item per card is returned, so it should be safe to tally the resulting reviews in the FSRS items. WDYK?

ftl/core/deck-config.ftl
abdnh: This is how it's documented in fsrs-rs: https://github.com/open-spaced-repetition/fsrs-rs/blob/e34265a5de7a9fbb138ea8b600deae44cf05a129/src/dataset.rs#L16

dae:
Thanks, that's nicer :-)

PR #3028, opened 2024-02-24
Make backend.get_scheduling_states accessible via Collection

janAte1:
An add-on I was working on needed to use backend.get_scheduling_states, which made anki give a warning telling me to use a corresponding public method on Collection instead, but there wasn't any, so I'm adding it

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -161,6 +161,7 @@ Kai Knoblich <kai@FreeBSD.org>
 Lucas Scharenbroch <lucasscharenbroch@gmail.com>
 Antonio Cavallo <a.cavallo@cavallinux.eu>
 Han Yeong-woo <han@yeongwoo.dev>
+Antoni Paduch <antoni.paduch@gmail.com>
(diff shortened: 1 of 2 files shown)


dae:
You probably don't want to do that, as it will ignore custom scheduling. The states are also accessible behind a hidden reviewer._v3.states - could you use that instead?

janAte1:
Thank you for your reply.
reviewer._v3.states can only get the states of the card in the front of the review queue.
What I want to do is create a card and then recreate it's review history from data exported from another application by repeatedly calling sched.answer_card. 
To do that I need the states of the added card, which seem to only be accessible trough backend.get_scheduling_states.
(Writing data directly to the database would work, but I want the cards' states to be set based on the imported reviews)

dae:
I'm not sure that's going to work well, as answer_card() is intended to be used only for user reviews, and depends on the system time, and will give an error if the card is not at the top of the queue. You'll probably need to manually write the review entries / update the card state instead.

PR #3029, opened 2024-02-25
fix progress manager window race condition

kieranlblack:
The progress manager was indicating windows were in a clean state before actually ensuring they were in a clean state. The update to mark things as clean now occurs after cleanup has occurred.

This would generally be fine since all UI code is run from the main thread but we actually allow for events to be handled while in the midst of performing cleanup on the open `ProgressDialog`. See:

https://github.com/ankitects/anki/blob/68b374e65f012f12f6ba266aea293cfaa9616389/qt/aqt/progress.py#L264-L272

Something must've changed recently in PyQt6 causing things to run with different timings exposing this race condition.

I think this is what was causing the reports on the forums of the `Processing...` windows which would not close (and also the one in the chinese support addon issues):

- https://forums.ankiweb.net/t/anki-stuck-on-processing/41311
- https://forums.ankiweb.net/t/stuck-when-editing-a-card/39540
- https://forums.ankiweb.net/t/bug-pasting-card-leads-to-never-ending-processing-pop-up-window/38455

Basically a second window coming along and messing with the internal state before it was actually closed resulting in things being left orphaned.

Close Gustaf-C/anki-chinese-support-3/issues/60
Close Gustaf-C/anki-chinese-support-3/issues/61

### DIFF
--- a/qt/aqt/progress.py
+++ b/qt/aqt/progress.py
@@ -223,9 +223,11 @@ class ProgressManager:
             self._win.form.progressBar.setValue(self._counter)
 
     def finish(self) -> None:
-        self._levels -= 1
-        self._levels = max(0, self._levels)
-        if self._levels == 0:
+        # we must latch the levels update and perform it after cleanup has occured or we expose ourselves to a race
+        # condition where a second progress could see levels == 0 and wrongly assume everything is in a clean state
+        next_levels = self._levels - 1
+        next_levels = max(0, next_levels)
+        if next_levels == 0:
             if self._win:
                 self._closeWin()
             if self._busy_cursor_timer:
@@ -239,6 +241,7 @@ class ProgressManager:
             self._backend_timer.stop()
             self._backend_timer.deleteLater()
             self._backend_timer = None
+        self._levels = next_levels
 
     def clear(self) -> None:
         "Restore the interface after an error."


dae:
Thanks for looking into this Kieran. Clamping the values to 0 feels a bit like a bandaid rather than a root-fix (moving all state behind a mutex or ensuring no reentrancy+main thread access only), but if you've tested this and confirmed it addresses the issue, better to have a bandaid than no fix at all!

(processEvents() can cause reentrancy issues/crashes, so eventually I'd like to see its usage completely removed from the codebase)

kieranlblack:
@dae I was just thinking about this and I believe even though it seems to fix the being reported, the code I submitted still isn't fully free of race conditions due to the reentrancy of `processEvents()`. Consider the following scenario:

1. `finish()` is called with `_levels = 1`
2. `_closeWin()` is called and yields control via `processEvents()`
3. `start()` is called and increments `_levels` to 2
4. `processEvents()` returns
5. `_win` is cancelled and set to `None`
6. `_levels` is set to 0

So we end up in a state where `_levels = 0` yet there is still a `ProgressDialog` active. It is a similar issue to as before just wandered into in a different way. Here we would never end up calling `cancel()` or `isdeleted()` on a `ProgressDialog` which is `None` but still have the issue of getting orphaned windows.

The reentrancy is somewhat tricky, I suppose we need to check the `_levels` "semaphore" on gaining control again. We can make the metaphorical bandaid slightly larger or remove it like you were saying. Some ideas:
- we move the window presented check into the start `finish()` and out of `_closeWin()` and do it first before updating any internal state, this way after finishing the check we are guaranteed that the value we read for `_levels` won't change
- we have `_closeWin()` check `_levels` on reentry, if it isn't 1 then we return False and can bail out in `finish()` after setting `_levels` appropriately
- we use `taskman` to wait until the window has been shown for a while and then on the done callback check all the other logic from the main thread, this removes the usage of `processEvents()`

Do you feel one of these is significantly better than the others?

PR #3030, opened 2024-02-25
fix `processEvents()` reentrancy bug in progress manager window handling

kieranlblack:
My fix in #3029 was not fully correct, this PR should be a corrected version though.

> @dae I was just thinking about this and I believe even though it seems to fix the being reported, the code I submitted still isn't fully free of race conditions due to the reentrancy of `processEvents()`. Consider the following scenario:
> 
> 1. `finish()` is called with `_levels = 1`
> 2. `_closeWin()` is called and yields control via `processEvents()`
> 3. `start()` is called and increments `_levels` to 2
> 4. `processEvents()` returns
> 5. `_win` is cancelled and set to `None`
> 6. `_levels` is set to 0
> 
> So we end up in a state where `_levels = 0` yet there is still a `ProgressDialog` active. It is a similar issue to as before just wandered into in a different way. Here we would never end up calling `cancel()` or `isdeleted()` on a `ProgressDialog` which is `None` but still have the issue of getting orphaned windows.
> 
> The reentrancy is somewhat tricky, I suppose we need to check the `_levels` "semaphore" on gaining control again. We can make the metaphorical bandaid slightly larger or remove it like you were saying. Some ideas:
> 
> * we move the window presented check into the start `finish()` and out of `_closeWin()` and do it first before updating any internal state, this way after finishing the check we are guaranteed that the value we read for `_levels` won't change
> * we have `_closeWin()` check `_levels` on reentry, if it isn't 1 then we return False and can bail out in `finish()` after setting `_levels` appropriately
> * we use `taskman` to wait until the window has been shown for a while and then on the done callback check all the other logic from the main thread, this removes the usage of `processEvents()`
> 
> Do you feel one of these is significantly better than the others?

I implemented the third option from the list of ideas as it removed the `processEvents()` call but am more than happy to pivot to something else entirely if you want! Just lmk.

I tested this by playing around with the app for a little bit and inserting artificial delays in an addon. The path where the window was never shown and where we need to delay for half a second both seem to be working. No infinite Processing... windows got stuck open.

### DIFF
--- a/qt/aqt/progress.py
+++ b/qt/aqt/progress.py
@@ -252,21 +252,21 @@ class ProgressManager:
         # currently shown then we need to give the window system a half-second to
         # present the window before we close it again - fixes progress window getting
         # stuck, especially on ubuntu 16.10+
+        elapsed_time = time.monotonic() - self._shown
+        time_to_wait = 0.5 - elapsed_time
         # NOTE: we must not yield control if the window is not shown since we don't want
         # to expose ourselves to the possibility of something showing the window in the
         # meantime
-        if self._shown:
-            elapsed_time = time.monotonic() - self._shown
-            if (time_to_wait := 0.5 - elapsed_time) > 0:
-                self.mw.taskman.run_in_background(
-                    lambda time_to_wait=time_to_wait: time.sleep(time_to_wait),
-                    do_window_cleanup,
-                    uses_collection=False,
-                )
-            else:
-                do_window_cleanup()
-        else:
+        if (not self._shown) or (time_to_wait <= 0):
             do_window_cleanup()
+        else:
+            # NOTE: we can't use self.single_shot here since it won't call the callback
+            # until _levels = 0, but if we are in this branch then _levels > 0
+            self.mw.taskman.run_in_background(
+                lambda time_to_wait=time_to_wait: time.sleep(time_to_wait),
+                do_window_cleanup,
+                uses_collection=False,
+            )
 
     def clear(self) -> None:
         "Restore the interface after an error."
--- a/qt/aqt/progress.py
+++ b/qt/aqt/progress.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import time
+from concurrent.futures import Future
 from dataclasses import dataclass
 
 import aqt.forms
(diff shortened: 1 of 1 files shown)


qt/aqt/progress.py
```diff
@@ -223,25 +224,49 @@ def update(
...
+        # currently shown then we need to give the window system a half-second to
+        # present the window before we close it again - fixes progress window getting
+        # stuck, especially on ubuntu 16.10+
+        # NOTE: we must not yield control if the window is not shown since we don't want
+        # to expose ourselves to the possibility of something showing the window in the
+        # meantime
+        if self._shown:
+            elapsed_time = time.monotonic() - self._shown
+            if (time_to_wait := 0.5 - elapsed_time) > 0:
+                self.mw.taskman.run_in_background(
```
dae, line 261: Using sleep() in a background thread could be done more simply with self.single_shot()

qt/aqt/progress.py
```diff
@@ -223,25 +224,49 @@ def update(
...
+                    lambda time_to_wait=time_to_wait: time.sleep(time_to_wait),
+                    do_window_cleanup,
+                    uses_collection=False,
+                )
+            else:
+                do_window_cleanup()
+        else:
+            do_window_cleanup()
```
dae, line 269: This logic could be slightly simplified by calculating the time to sleep, and ensuring it's >= 0. That way the two else paths just becomes delays of 0, meaning they'll get run by Qt in the next main loop invocation.

dae (commented):
Removing processEvents() is probably the best path forward. Thank you for looking into this!

qt/aqt/progress.py
kieranlblack: Oh nice that is perfect!

qt/aqt/progress.py
kieranlblack: Hmm this one is a little tricky. If `self._shown == 0` then we don't want something to sneak in and make it shown before we get invoked at the back of the timer callback queue, so we have to handle it then and there. We could remove the inner `else` if we max the time to sleep with 0 but that I feel like that obscures the intent of the code.

Unless maybe I am missing your meaning?

dae: One alternative would be to set time_to_wait at the top, and then immediately cleanup if not _shown or delay <= 0. It's a minor point in any case, so no strong preference on my end. :-)

kieranlblack: Okay I see! Thanks for pointing that out. I also have no strong preference and won't read it as many times as you will in the future, so I'll reduce the nesting with the or.

qt/aqt/progress.py
kieranlblack: @dae I just looked at this a little more and we can't actually use `single_shot` here since it wraps the callback and won't call it until `_levels` is 0 which is the whole point of our callback :/ I'll stick to the threadpool for now and add a comment about it.

dae: Sorry to lead you down the wrong path there -  I'd forgotten about that check.

PR #3031, opened 2024-02-25
fix text input custom style wiping out base user style

kieranlblack:
While trying to inject custom style for the chinese support addon I came across this issue, see Gustaf-C/anki-chinese-support-3#32.

In short, if an addon applies custom css to their own tag via `.textContent` after the `userBase` css has been applied, it wipes out most / all of the `userBase` custom tag. If the custom css from the addon is blessed by the microtasks queue and is applied before the `userBase` stuff is applied there are no issues. A somewhat interesting race condition.

Behaviour without the patch in this PR:

Expected behaviour (behaviour with the PR patch):

Find the related post on the forum about the issue here:
https://forums.ankiweb.net/t/customstyles-strange-behaviour/37432


dae:
I'm not very familiar with this part of the code, but it seems like a reasonable change - thanks Kieran!

PR #3034, opened 2024-02-25
Localize type-in preview strings

### LINKED ISSUE #3032: Type-in-answer preview has hardcoded strings
As shown in the video below, the type-in-answer preview example strings are hardcoded in English even if the app is in another language:

https://github.com/ankitects/anki/assets/69634269/e524ab51-70b6-4ed4-b4b0-fa3a566c1d1a

Relevant code:

https://github.com/ankitects/anki/blob/d4a52b75480cd69724a8d14c7744b032d91072dd/qt/aqt/clayout.py#L587

BrayanDSO:
* Fixes #3032

also corrects the provided/expected strings comparison:

image

### DIFF
--- a/ftl/core/card-templates.ftl
+++ b/ftl/core/card-templates.ftl
@@ -61,4 +61,7 @@ card-templates-restore-to-default = Restore to Default
 card-templates-restore-to-default-confirmation = This will reset all fields and templates in this notetype to their default
     values, removing any extra fields/templates and their content, and any custom styling. Do you wish to proceed?
 card-templates-restored-to-default = Notetype has been restored to its original state.
-
+# Sample text to be used as the answer provided by the user in a "type in the answer" card preview
+card-templates-type-in-provided = example
+# Sample text to be used as the expected answer in a "type in the answer" card preview
+card-templates-type-in-expected = sample
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -584,7 +584,10 @@ class CardLayout(QDialog):
         hadHR = origLen != len(txt)
 
         def answerRepl(match: Match) -> str:
-            res = self.mw.col.compare_answer("example", "sample")
+            res = self.mw.col.compare_answer(
+                tr.card_templates_type_in_expected(),
+                tr.card_templates_type_in_provided(),
+            )
             if hadHR:
                 res = f"<hr id=answer>{res}"
             return res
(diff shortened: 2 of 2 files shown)


ftl/core/card-templates.ftl
```diff
@@ -61,4 +61,5 @@ card-templates-restore-to-default = Restore to Default
 card-templates-restore-to-default-confirmation = This will reset all fields and templates in this notetype to their default
     values, removing any extra fields/templates and their content, and any custom styling. Do you wish to proceed?
 card-templates-restored-to-default = Notetype has been restored to its original state.
-
+card-templates-example = example
+card-templates-sample = sample
```
abdnh, line 65: Since the example is intended to show how comparison works, can we add an explanation for translators so that they pick a good example in their languages, instead of literally translating the strings?

BrayanDSO: Super reasonable and I had thought the same, but I don't know exactly how to do it yet here with Fluent, since I never went to Pontoon to see how things are displayed there.

From [Fluent docs](https://projectfluent.org/fluent/guide/comments.html), I think that I should add a comment, but still not sure

BrayanDSO: Okay, found the [correct docs](https://translating.ankiweb.net/anki/developers). Will change it

BrayanDSO: Added them. Thanks for the push!

dae:
I'm a bit concerned that even with translation hints (which was a good suggestion), translators might struggle to come up with two distinct words/phrases that produce an easily-understandable typing comparison.  In the past, Anki used "example" and "exomple", and more than one person filed a bug report thinking that we'd accidentally misspelled it! It doesn't really even need to be English. If you feel that using English words there is unfair to other language speakers, we could use something like "12345" and "12245" instead. WDYT?

BrayanDSO:
Looks like this now:

image


image

dae:
Maybe reverse the two? Otherwise I think that looks ok, and bypasses all the translation issues. What do you think?

BrayanDSO:
When I read `123` at the type-in box in the question, I can associate that it is an example, but not so much with `124`.

Maybe something extra could be added into the question box to make that clearer. like `example: 123` and then compare `123` to `124` normally in the answer, but I'm really not sure of that

dae:
We can't easily inject text outside of the typing area - it would need fragile logic that looks for the input/code sections. I think we're probably either going to need to stick to numbers, or the existing hard-coded English, unless someone has another suggestion.

BrayanDSO:
I meant inside the type box. 

```
1234 (example)
```

or
```
example: 1234
```

then in the answer
```
1234
     I
    V
1235
```

dae:
I was wrong about injecting the text - we're looking for `[type:` already, which should be pretty unambiguous. So we could inject that text into the typing area. But don't you think it would be clearer to include it on a previous line, so it's more obvious what text was 'typed'?

BrayanDSO:
some options:

image

image

image

dae:
I think the second is the clearest. What do you think about changing the wording to something like "Example typed text:"?

PR #3038, opened 2024-02-27
Preserve HTML formatting inside clozes

### LINKED ISSUE #2991: Manipulate the DOM when creating cloze deletions?
The current approach to cloze deletions just inserts `{{cN:` and `}}` around the selected text, which can result in malformed HTML when the text inside the cloze is hidden, such as `<div>{{c1:something</div>else}}`. Perhaps we could solve this with a more sophisticated approach that manipulates the DOM, such as closing/re-opening the active tags around the cloze markers?

It's a long-standing issue; the latest report is https://forums.ankiweb.net/t/minor-bug-and-buttons-break-html-formatting/40695

abdnh:
Closes #2991


dae:
Thanks Abdo. I've given this some testing, and it seems to perform better in some simple cases like when a styling tag terminates halfway through a cloze. It's a bit of a mixed bag with some other common patterns however. For example, with `<ul><li>one</li><li>two</li><li>three</li></ul>`, attempting to cloze the first two lines results in some odd behavior:

```
<ul><li>{{c1::</li><li>one</li><li>two</li>}}<br><li>three</li></ul>
```

This results in the cloze appearing on the same line as 'three', and still results in the first `li` tag missing a closing tag when the {{c1...}} part is hidden.

Clozing of lists was one of the main motivations for this change, so it unfortunately doesn't feel practical to merge in its current state. I'm not sure what the best path forward would be - we could potentially try to tweak the algorithm, or add special cases for lists/other common structures, or put this back in the too-hard basket for now. What do you think?

abdnh:
> attempting to cloze the first two lines results in some odd behavior:

Unless I'm missing something, clozing the first two lines results in this HTML in this PR:
```html
<ul><li></li>{{c1::<li>one</li><li>two</li>}}<li></li><li>three</li></ul>
```


> This results in the cloze appearing on the same line as 'three', and still results in the first li tag missing a closing tag when the {{c1...}} part is hidden.

I'm not seeing this behavior with the same code.
image
image

abdnh:
I pushed a change that fixes empty `li` tags and handle partially covered tags. The logic is messy but it appears to work.

One big drawback with the whole range manipulation approach though is that undo no longer works.

dae:
If I select `ld</b>and<i>it` in the following and create a cloze:

```
<b>bold</b>and<i>italic</i>
```

it gives:

```
<b>bo{{c1::ld</b>and<i>it}}alic</i>
```

For the list case, I enter the following:

```
<ul>
  <li>one</li>
  <li>two</li>
  <li>three</li>
</ul>
```

Then I click before 'o' in one', and shift+down twice, then create a cloze. I get:

```
<ul>
  {{c1::<li>one</li>
  <li>two</li>
  <li>}}three</li>
</ul>
```

That results in [...]three for the question, and the answer has one, two and also three highlighted, due to the HTML turning into:

```
<ul>
  <span class="cloze" data-ordinal="1"><li>one</li>
  <li>two</li>
  <li></span>three</li>
</ul>
```

Perhaps such issues are going to be unavoidable if we keep doing the cloze replacements via simple text substitution?

dae:
> One big drawback with the whole range manipulation approach though is that undo no longer works.

Interestingly, if you click in the HTML editor area and ctrl+z, that does appear to work :-)

BlueGreenMagick:
> Then I click before 'o' in one', and shift+down twice, then create a cloze. I get:
> 
> ```
> <ul>
>   {{c1::<li>one</li>
>   <li>two</li>
>   <li>}}three</li>
> </ul>
> ```


I think this particular case is fixable by checking if range start is at end of range startContainer, (e.g. `{{c1::</li>`)or range end is at start of range endContainer (e.g. `<li>}}`)

dae:
I'm a bit fearful that if we can't find some generic approach that works in most/all situations, it may not be practical to implement this change. If we have to start special-casing multiple different situations, it's going to struggle to work in all cases, and we risk certain targeted changes breaking other targeted changes. Happy to be proved wrong though!

abdnh:
I'm also wary of adding more special cases, but for lists at least, this approach seems to perform better than what we have.


Regarding the `<b>bold</b>and<i>italic</i>` example, the existing approach based on execCommand() results in this:
```html
<b>bo</b>{{c1::<b>ld</b>and<i>it</i>}}<i>alic</i>
```

So we'll need a special case for `<b>` and similar tags. Will look into whether this is practical to implement.

dae:
This PR seems to handle the two examples I posted above well now.  I'm still a bit worried about other potential regressions, but I guess we won't know about them unless we try this. Happy for me to merge this in before the next rc?

dae:
(alternatively we could wait until the next beta round)

dae:
@abdnh Based on the forum comments, I'm thinking we'll need to roll this back for rc3/release. Do you agree?

abdnh:
The issue reported by @jcznk is a one-line fix. I was not able to reproduce the first issue reported by the other user yet.

abdnh:
That's reasonable.

dae:
Thanks Abdo. I'll hold off on the new PR you sent through (thanks!), and will merge it in once this gets unreverted.

dae:
(this is still pending until we can get a release with a longer beta period)

badlydrawnrob:
Just a follow-on from #2991 and [this forum post](https://forums.ankiweb.net/t/minor-bug-and-buttons-break-html-formatting/40695) it seems to be happening mainly for me when you're selecting to the end of a line (which contains the surrounding tag, in my case something like `<span id="cb1-3">`) and pressing the button for a cloze tag.

Doing that places the cloze tag _outside_ it's original parent wrapper tag. It doesn't seem to break formatting so badly if you don't select to the end of the line. Obviously, it depends where you select the text:

```html
<span id="cb1-3"><a href="#cb1-3" aria-hidden="true" tabindex="-1"></a><span class="kw">import</span> <span class="dt">Html</span> <span class="kw">exposing</span> (<span class="dt">Html</span><span class="op">,</span> <span class="fu">button</span><span class="op">,</span> <span class="fu">div</span><span class="op">,</span> <span class="fu">text</span><span class="op">,</span> <span class="fu">img</span>)</span>
```

Becomes

```html
<span id="cb1-3"><a href="#cb1-3" aria-hidden="true" tabindex="-1"></a><span class="kw">import</span> <span class="dt">Html</span> <span class="kw">exposing</span> ({{c1::<span class="dt">Html</span><span class="op">,</span> <span class="fu">button</span><span class="op">,</span> <span class="fu">div</span><span class="op">,</span> <span class="fu">text</span><span class="op">,</span> <span class="fu">im</span>}}<span class="fu">g</span>)</span>
```

So it actually seems to have duplicated the `<span class="fu">` tag where you'd expect, but it hasn't pulled it outside of it's parent wrapper `<span id="cb1-3">` like it would've done if you'd selected to the end of the line (the last `)` parenthesis)

dae:
Current status of this PR, if memory serves correctly: this change was deployed in the 24.06rcs and reverted again due to another discovered regression.

PR #3048, opened 2024-03-03
Makes strings in the Compute Optimal Retention section translatable

tekinosman:


dae:
These were left untranslated, because the feature is listed as experimental, and may change. @L-M-Sherlock, do you think we should be translating these strings yet?

L-M-Sherlock:
It's still experimental. And I plan to refactor it in the feature. Some strings will change.

tekinosman:
My apologies for this oversight. Let's close it.

dae:
No problem, thank you for contributing in any case!

PR #3049, opened 2024-03-03
Use `Card %d` for IO template names

### LINKED ISSUE #3043: Image Occlusion `Cards...` preview uses `Cloze %d` as the cards names
As an example, that leads to a problem in portuguese, since the current translation for cloze is literally `Words omission`

image

BrayanDSO:
* Fixes #3043

image

image

### DIFF
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -293,7 +293,7 @@ class CardLayout(QDialog):
         qconnect(widg.returnPressed, self.on_search_next)
 
     def setup_cloze_number_box(self) -> None:
-        names = (tr.card_templates_cloze(val=n) for n in self.cloze_numbers)
+        names = (tr.card_templates_card(val=n) for n in self.cloze_numbers)
         self.pform.cloze_number_combo.addItems(names)
         try:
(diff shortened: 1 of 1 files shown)


dae:
My thinking in #3043 is that we'd use 'card x' for both I/O notetypes and normal clozes, so we don't need to check what type of card it is.

You should be able to fix the formatting issues with './ninja format'

BrayanDSO:
AFAIU, that box in the right of `Options` is only shown in `Cloze` types

'./ninja format'. I tried but it didn't work nor I got any error messages, and I ran out of time. I'll try to debug it later

dae:
I/O notetypes are cloze notetypes. Your code is showing something different for each of them; I'm suggesting we just use the same string in both cases.

dae:
You can use `./ninja check -v` to see the commands that are run, and can run them manually if you want to see what they output.

dae:
No worries - thank you!

abdnh:
> Actually, I was just using Powershell instead of Bash. Fixed it now

You can still use PowerShell if you have Bash on your path by the way.

PR #3050, opened 2024-03-04
Use system certificate store

### LINKED ISSUE #3016: System certs are not picked up
I figured out my SSL issues with Anki. On my computer, I have Netskope installed which intercepts HTTPS/SSL. I had to add the Netskope self-signed root certificate to the `cacert.pem` file of the `certifi` package that ships with the Anki app.

I used the following command to figure out what the relevant certificates are:

``` 
openssl s_client -showcerts -connect texttospeech.googleapis.com:443
```

And it gave me all certificates in the certificate chain as part of the output. 

I then added the self-signed Netskope certificate to `/Applications/Anki.app/Contents/Resources/certifi/cacert.pem`. This is on macOS. I hope this helps others running into similar issues.

_Originally posted by @skadauke in https://github.com/ankitects/anki/issues/1965#issuecomment-1953031027_

abdnh:
Closes #3016

I confirmed this has an effect by emptying cacert.pem. The file still has to exist because it's read by certifi/requests.


dae:
Thanks for digging into the packaging aspect as well!

Issue #3037, opened 2024-02-27
Native IO Long Images are blurry in add/edit view


dae:
> I use long images for IO and while editing them I find the images as blurry (although they appear to be normal on the review screen) and the cloze boxes too big.

Originally reported on https://forums.ankiweb.net/t/native-io-long-images-are-blurry-in-edit-view/41486

Attaching a sample image that can be used to confirm the issue.

I think this may be an issue with panzoom/Safari's available quality with the underlying zoom transformations. If instead of zooming in with a pinch gesture in the main area, and you instead pinch on the sidebar so the whole document zooms in, the image is more legible.

<img width="830" alt="tall" src="https://github.com/ankitects/anki/assets/180542/b744223e-ae9a-4aba-828e-57d6e9453bc5">

krmanik:
I will look into this, maybe we need to move away from panzoom and use panzoom of fabric.js.

dae:
Thanks Mani. Hopefully it's something that can be worked around, and not a fundamental issue with Safari.

dae:
Should be fixed in #3052

PR #3052, opened 2024-03-04
fix: blur in io, remove panzoom and use fabricjs for panzoom

krmanik:
I have updated gradually so there are no atomic commits here, but I will explain what going on in this PR.

The panzoom library removed, the same functionality is implemented using 
https://codepen.io/amsunny/pen/XWGLxye
http://fabricjs.com/docs/fabric.Canvas.html#viewportTransform
http://fabricjs.com/using-transformations

The IO image is set as background image for canvas, it is easier for zoom/pan using canvas only.
A bounding rectangular area added with same width and height of background image, all shapes are allowed to draw inside bounding area only.

I have used large image for testing.

https://github.com/ankitects/anki/assets/12841290/9f487f34-7c46-42a1-815a-74be7b04ff38

### DIFF
--- a/ts/image-occlusion/mask-editor.ts
+++ b/ts/image-occlusion/mask-editor.ts
@@ -33,11 +33,11 @@ export const setupMaskEditor = async (
     // get image width and height
     const image = document.getElementById("image") as HTMLImageElement;
     image.src = getImageData(imageData.data!, path);
-    image.onload = async function() {
+    image.onload = function() {
         const size = optimumCssSizeForCanvas({ width: image.width, height: image.height }, containerSize());
         setCanvasSize(canvas);
         onImageLoaded({ path });
-        await setupBoundingBox(canvas, size);
+        setupBoundingBox(canvas, size);
         undoStack.reset();
     };
 
@@ -200,7 +200,7 @@ export async function resetIOImage(path: string, onImageLoaded: (event: ImageLoa
         image.height = size.height;
         setCanvasSize(canvas);
         onImageLoaded({ path });
-        await setupBoundingBox(canvas, size);
+        setupBoundingBox(canvas, size);
     };
 }
 globalThis.resetIOImage = resetIOImage;
--- a/ts/image-occlusion/mask-editor.ts
+++ b/ts/image-occlusion/mask-editor.ts
@@ -5,22 +5,16 @@ import { protoBase64 } from "@bufbuild/protobuf";
 import { getImageForOcclusion, getImageOcclusionNote } from "@tslib/backend";
 import * as tr from "@tslib/ftl";
 import { fabric } from "fabric";
-import type { PanZoom } from "panzoom";
 import { get } from "svelte/store";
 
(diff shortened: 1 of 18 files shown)


krmanik:
The pinch to zoom is not supported in fabricjs and author of fabricjs suggested to use [HammerJS](https://github.com/hammerjs/hammer.js) in this issue [#8849](https://www.github.com/fabricjs/fabric.js/issues/8849#issuecomment-1925940607). I have used the library for pinch to zoom in mobile client only. Tested on device with port forwarding in chrome.

https://github.com/ankitects/anki/assets/12841290/4185ac42-a738-42a2-8267-68962c9fa700

dae:
Thanks Mani. I've given this a brief test with the large image from https://github.com/ankitects/anki/issues/3037 on iOS. It works pretty well, and appears to correctly preserve the image clarity, but can get a bit laggy when zooming in/out at times, which I guess is something we can't avoid.

Would it be practical to allow the two-finger pinch gesture without requiring the user to first enter zooming mode? Interestingly on the desktop, a regular pinch to zoom shows the image in lower quality in this PR, because it's zooming the entire document. But when you zoom properly using the zoom tool, there's no image quality loss.

I've briefly skimmed the code; I'll give it a proper review in the next few days.

ts/image-occlusion/mask-editor.ts
```diff
@@ -129,10 +118,49 @@ function initCanvas(onChange: () => void): fabric.Canvas {
...
+            width: size.width,
+            height: size.height,
+            hasBorders: false,
+            hasControls: false,
+            lockMovementX: true,
+            lockMovementY: true,
+            evented: false,
+            stroke: "red",
+        });
+        globalThis.boundingBox = boundingBox;
```
dae, line 137: We should avoid polluting the global namespace where possible, especially with generic names like this. You've already added a getBoundingBox() function - could we make it use a file-private variable instead of a global?

dae (commented):
I've looked over the code a bit more carefully. Only one minor comment:

ts/image-occlusion/mask-editor.ts
krmanik: I will remove global namespace, got busy but in next commit the issues will be resolved.

krmanik:
Now, zooming can be performed in draw mode on the mobile client (panning is currently not supported). Previously, I had set the background image to the canvas, which resulted in the canvas being slow. In the current implementation, the image tag element is used instead of the background image, fixing the issues. The values are calculated based on the position of the bounding box in the canvas area and available container area.

krmanik:
Now panning is also supported. The canvas lagging, blur, zoom and pan, all issues fixed in the current PR along with support for touch based devices.

https://github.com/ankitects/anki/assets/12841290/ad2add6f-fa84-47b0-8fa8-27f8de32214a

dae:
Thanks Mani, this is great - performs better, and much more ergonomic. This also improves clarify when pinch-zooming on a Mac as well. Great work!

PR #3053, opened 2024-03-04
Fix alert not appearing when FSRS parameters input is focused

abdnh:
https://forums.ankiweb.net/t/anki-24-04-beta/41792/27


dae:
Thanks for investigating this, Abdo!

PR #3055, opened 2024-03-06
Show preset and original deck in card info

### LINKED ISSUE #3020: show preset in card info
> could we add a feature request for displaying preset in card information, seems like a good enhancement

Originally reported on https://forums.ankiweb.net/t/issues-with-anking/41344/8

abdnh:
Closes #3020

### DIFF
--- a/ts/card-info/CardStats.svelte
+++ b/ts/card-info/CardStats.svelte
@@ -106,7 +106,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         statsRows.push({ label: tr2.cardStatsCardTemplate(), value: stats.cardType });
         statsRows.push({ label: tr2.cardStatsNoteType(), value: stats.notetype });
         let deck: string;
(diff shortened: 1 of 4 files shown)


ts/card-info/CardStats.svelte
```diff
@@ -105,7 +105,14 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
         statsRows.push({ label: tr2.cardStatsCardTemplate(), value: stats.cardType });
         statsRows.push({ label: tr2.cardStatsNoteType(), value: stats.notetype });
-        statsRows.push({ label: tr2.cardStatsDeckName(), value: stats.deck });
+        let deck: string;
+        if (stats.deck !== stats.originalDeck) {
+            deck = `${stats.deck} (${stats.originalDeck})`;
+        } else {
+            deck = stats.deck;
+        }
```
dae, line 113: Making originalDeck optional would be slightly more elegant than comparing the two deck names, but it's not particularly important - let me know if you want this merged without that change.

PR #3059, opened 2024-03-07
fix: cloze_numbers_in_string treating `{{c0::}}` as valid

BrayanDSO:
[cloze_numbers_in_string treating](https://github.com/ankitects/anki/blob/9228c87b3af7e8ecd611aa4b221d7a2012c56b32/rslib/src/cloze.rs#L411-L415) and consequently [cloze_numbers_in_note](https://github.com/ankitects/anki/blob/9228c87b3af7e8ecd611aa4b221d7a2012c56b32/rslib/src/notes/service.rs#L127) considered `{{c0::}}` valid, although they shouldn't


dae:
@BrayanDSO mind elaborating on what issue the current code was causing?

@abdnh Will this change cause problems for the text annotations in I/O?

BrayanDSO:
> @BrayanDSO mind elaborating on what issue the current code was causing?

Note.cloze_numbers_in_fields, which is an available method for Add-on creators, has `0` among the returned result if `{{c0::}}` is present in the card, although it shoudn't.

https://github.com/ankitects/anki/blob/6ac60f8032c125f9178797aba029a2ead17a2587/pylib/anki/notes.py#L137-L138

Also, it is a current cause of an issue in AnkiDroid if `{{c0::}}` is present when a card is previewed because of the faulty return value of  [cloze_numbers_in_note](https://github.com/ankitects/anki/blob/9228c87b3af7e8ecd611aa4b221d7a2012c56b32/rslib/src/notes/service.rs#L127)

dae:
I'm not sure there's a "right" answer here - I suspect you could make a case for 0 being excluded here, or at a different layer instead. If it doesn't break anything, I don't have any objections. Will be interested to hear from @abdnh.

rslib/src/cloze.rs
```diff
@@ -417,7 +417,7 @@ pub fn cloze_numbers_in_string(html: &str) -> HashSet<u16> {
 fn add_cloze_numbers_in_text_with_clozes(nodes: &[TextOrCloze], set: &mut HashSet<u16>) {
     for node in nodes {
         if let TextOrCloze::Cloze(cloze) = node {
-            if !(cloze.image_occlusion().is_some() && cloze.ordinal == 0) {
+            if !(cloze.image_occlusion().is_some() || cloze.ordinal == 0) {
```
abdnh, line 420: We can remove the check for image occlusion. I only added it to avoid affecting the behavior of normal clozes in any way (#2705):

```suggestion
            if cloze.ordinal != 0 {
```

abdnh (approved):

abdnh:
> @abdnh Will this change cause problems for the text annotations in I/O?

It won't cause problems.

PR #3060, opened 2024-03-08
Added new translator name

Foxy-null:
I tried polishing the Japanese in Anki a bit. I would be happy if my name were included as a contributor.


dae:
Thank you for your contributions!

PR #3061, opened 2024-03-08
alert when the resp.weights is empty

L-M-Sherlock:
When the current weights are empty and the new weights are worse than the default weights, the previous implementation doesn't pop up the `alert(tr.deckConfigFsrsParamsOptimal())` because `$config.fsrsWeights.length` is 0.

By the way, I also remove the condition `weights.fsrs_items >= 1000` when computing all weights.


dae:
Thanks Jarrett!

PR #3065, opened 2024-03-10
refactor optimal retention to minimize workload per memorization & fix progress goes backwards

L-M-Sherlock:
I will update the dependency after the following PR is merged:
- https://github.com/open-spaced-repetition/fsrs-rs/pull/165

Help wanted:
- brief introduction of this feature.
- notice user that desired retention < optimal retention is worse than desired retention > optimal retention because the former means less memorization & more workload and the latter means more memorization & more workload.
- notice user that the setting of Maximum interval would affect the result.

@Expertium, could you help me?

Edit:

This PR also contains:
- https://github.com/open-spaced-repetition/fsrs-rs/pull/166
- https://github.com/open-spaced-repetition/fsrs-rs/pull/167

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1792,9 +1792,7 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "0.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7c7e6a1986cc2b7a64445d84e2c453ecd8d95dcf90b797205c54573697e10b17"
+version = "0.5.1"
 dependencies = [
  "burn",
  "itertools 0.12.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1792,7 +1792,9 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "0.5.1"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a8346889b88a7cad403872851849c5208e51263931c521bb706c5ca90d907cc2"
 dependencies = [
  "burn",
  "itertools 0.12.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1792,9 +1792,9 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "0.5.2"
+version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


Expertium:
Perhaps just add a short text description under "Compute optimal retention (experimental)"? The problem is how to make it short and concise.

> Setting a retention lower than optimal is not recommended. Also, changing the maximum interval will impact the result.

L-M-Sherlock:
> EDIT: on second thought, max. interval should be a setting under "Compute optimal retention (experimental)", for 2 reasons:

These two settings are in the same section and their positions have been very near.

image

Expertium:
Ok, then ust add a text message.

dae:
> notice user that desired retention < optimal retention is worse than desired retention > optimal retention because the former means less memorization & more workload and the latter means more memorization & more workload.

To confirm I understand you correctly, do you mean:

- If the computed optimal retention (OR) is < desired retention (DR), we should say something like "Your desired retention is below optimal. Increasing it is recommended"?
- If OR >= DR, we show no message?

> The problem is that having a setting that affects the result of "Compute optimal retention" outside of, well, "Compute optimal retention", is weird.

It's already using a bunch of other information from the collection such as pass rates and the configured parameters. Do we really need to treat max interval differently?

L-M-Sherlock:
> * If the computed optimal retention (OR) is < desired retention (DR), we should say something like "Your desired retention is below optimal. Increasing it is recommended"?

Do you mean DR < OR? If so, it's what I mean.

dae:
A question for release planning: are there many other changes you'd like to make to FSRS before 24.04 releases? Would a stable release in 2-3 weeks after one more rc work, or would you like more time?

L-M-Sherlock:
There will not be many more changes, except that currently I'm trying to optimize the performance of optimal_retention, but there has been no progress so far.

dae:
Thanks guys!

Expertium:
Too bad I thought about this so late, but still: it's not obvious at all what "optimal" means, so I think there should be a short sentence saying something like "This value of retention minimizes the amount of studying per day" or something along those lines.

L-M-Sherlock:
> it's not obvious at all what "optimal" means

I'm writing a wiki page for it. Could you help me? Here is the link: https://github.com/open-spaced-repetition/fsrs4anki/wiki/The-optimal-retention

Expertium:
I meant that it should be clarified within Anki itself.

Expertium:
@user1823 here's another article that you can help us with, if you want to: https://github.com/open-spaced-repetition/fsrs4anki/wiki/The-Optimal-Retention

PR #3066, opened 2024-03-10
fix: ensure proper drawing of shape tools after pan/zoom

krmanik:
After pan/zoom the shape tools are not working, so the issue is fixed in this PR. Also removed bottom toolbar for more editing area in mobile devices.

In Android/iOS, the JS function can be called to add notes.

```js
const mode = imageOcclusion.mode;
imageOcclusion.addNote(mode);
```
Additional Context
https://github.com/ankidroid/Anki-Android/issues/15838

### DIFF
--- a/ts/image-occlusion/ImageOcclusionPage.svelte
+++ b/ts/image-occlusion/ImageOcclusionPage.svelte
@@ -6,11 +6,11 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import * as tr from "@tslib/ftl";
 
     import Container from "../components/Container.svelte";
+    import { addBrowserClasses } from "../reviewer/browser_selector";
     import { addOrUpdateNote } from "./add-or-update-note";
     import type { IOMode } from "./lib";
     import MasksEditor from "./MaskEditor.svelte";
     import Notes from "./Notes.svelte";
-    import StickyFooter from "./StickyFooter.svelte";
     import { hideAllGuessOne, textEditingState } from "./store";
 
     export let mode: IOMode;
@@ -18,6 +18,10 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     async function addNote(): Promise<void> {
         addOrUpdateNote(mode, $hideAllGuessOne);
     }
+    globalThis.imageOcclusion = {
+        mode,
+        addNote,
+    };
 
     const items = [
         { label: tr.notetypesOcclusionMask(), value: 1 },
@@ -29,6 +33,8 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         textEditingState.set(tabValue === 2);
         activeTabValue = tabValue;
     };
+
+    addBrowserClasses();
 </script>
 
 <Container class="image-occlusion">
@@ -51,8 +57,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     <div hidden={activeTabValue != 2}>
         <Notes />
     </div>
-
-    <StickyFooter {mode} {addNote} />
 </Container>
 
 <style lang="scss">
--- a/ts/image-occlusion/ImageOcclusionPage.svelte
+++ b/ts/image-occlusion/ImageOcclusionPage.svelte
@@ -6,22 +6,12 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import * as tr from "@tslib/ftl";
 
     import Container from "../components/Container.svelte";
-    import { addBrowserClasses } from "../reviewer/browser_selector";
-    import { addOrUpdateNote } from "./add-or-update-note";
     import type { IOMode } from "./lib";
     import MasksEditor from "./MaskEditor.svelte";
     import Notes from "./Notes.svelte";
-    import { hideAllGuessOne, textEditingState } from "./store";
+    import { textEditingState } from "./store";
 
     export let mode: IOMode;
-
-    async function addNote(): Promise<void> {
-        addOrUpdateNote(mode, $hideAllGuessOne);
-    }
-    globalThis.imageOcclusion = {
-        mode,
-        addNote,
-    };
 
     const items = [
         { label: tr.notetypesOcclusionMask(), value: 1 },
@@ -33,8 +23,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         textEditingState.set(tabValue === 2);
         activeTabValue = tabValue;
     };
-
-    addBrowserClasses();
 </script>
 
 <Container class="image-occlusion">
--- a/ts/image-occlusion/index.ts
+++ b/ts/image-occlusion/index.ts
@@ -4,10 +4,15 @@
 import "./image-occlusion-base.scss";
 
 import { ModuleName, setupI18n } from "@tslib/i18n";
+import { get } from "svelte/store";
 
 import { checkNightMode } from "../lib/nightmode";
+import { addOrUpdateNote } from "./add-or-update-note";
 import ImageOcclusionPage from "./ImageOcclusionPage.svelte";
 import type { IOMode } from "./lib";
+import { hideAllGuessOne } from "./store";
+
+globalThis.anki = globalThis.anki || {};
 
 const i18n = setupI18n({
     modules: [
@@ -24,6 +29,16 @@ const i18n = setupI18n({
 export async function setupImageOcclusion(mode: IOMode, target = document.body): Promise<ImageOcclusionPage> {
     checkNightMode();
     await i18n;
+
+    async function addNote(): Promise<void> {
+        addOrUpdateNote(mode, get(hideAllGuessOne));
+    }
+
+    // for adding note from mobile devices
+    globalThis.anki.imageOcclusion = {
+        mode,
+        addNote,
+    };
 
     return new ImageOcclusionPage({
         target: target,
--- a/ts/image-occlusion/tools/tool-zoom.ts
+++ b/ts/image-occlusion/tools/tool-zoom.ts
@@ -156,26 +156,50 @@ const onDrag = (canvas, opt) => {
     redraw(canvas);
 };
 
-export const onDragX = (canvas: fabric.Canvas, opt: WheelEvent) => {
-    const delta = opt.deltaX;
+export const onWheelDrag = (canvas: fabric.Canvas, event: WheelEvent) => {
+    const deltaX = event.deltaX;
+    const deltaY = event.deltaY;
     const vpt = canvas.viewportTransform;
-    canvas.lastPosX = opt.clientX;
+    canvas.lastPosX = event.clientX;
+    canvas.lastPosY = event.clientY;
 
-    vpt[4] -= delta;
-    canvas.lastPosX -= delta;
+    vpt[4] -= deltaX;
+    vpt[5] -= deltaY;
+
+    canvas.lastPosX -= deltaX;
+    canvas.lastPosY -= deltaY;
     canvas.setViewportTransform(vpt);
     constrainBoundsAroundBgImage(canvas);
     redraw(canvas);
 };
 
-export const onDragY = (canvas: fabric.Canvas, opt: WheelEvent) => {
-    const delta = opt.deltaY;
-    const vpt = canvas.viewportTransform;
-    canvas.lastPosY = opt.clientY;
+// for mac trackpad
+export const onGesture = (event) => {
+    event.preventDefault();
 
-    vpt[5] -= delta;
-    canvas.lastPosY -= delta;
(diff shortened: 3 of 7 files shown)


ts/image-occlusion/tools/tool-zoom.ts
```diff
@@ -205,8 +208,13 @@ const fitCanvasVptScale = (canvas: fabric.Canvas) => {
...
     const w1 = boundingBox.width;
-    const h2 = innerHeight - 79;
     const w2 = innerWidth - 42;
-
+    let h2 = window.innerHeight;
+    h2 = isMobile() ? h2 - 48 : h2 - 79;
     return Math.min(w2 / w1, h2 / h1);
 };
+
+const isMobile = () => {
```
dae, line 246: platform.ts:isDesktop()

ts/image-occlusion/ImageOcclusionPage.svelte
```diff
@@ -6,18 +6,22 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
     import Notes from "./Notes.svelte";
-    import StickyFooter from "./StickyFooter.svelte";
     import { hideAllGuessOne, textEditingState } from "./store";
 
     export let mode: IOMode;
 
     async function addNote(): Promise<void> {
         addOrUpdateNote(mode, $hideAllGuessOne);
     }
+    globalThis.imageOcclusion = {
```
dae, line 21: Please avoid polluting the global namespace. If you need to make something globally available, putting it in the top-level 'anki' would be better.

dae (commented):
Thanks for the quick fixes Mani. Some minor issues below.

Also, ctrl+scroll for zooming on desktop is nice. I wonder whether it's possible to also support pinch to zoom on Macs, that tend to use a trackpad? I think ctrl+scrollwheel is the most common on other platforms, but on a Mac users probably expect to pinch instead.

krmanik:
> I wonder whether it's possible to also support pinch to zoom on Macs, that tend to use a trackpad?

It may be possible, I will check it now.
https://developer.apple.com/documentation/webkitjs/gestureevent
https://www.github.com/hammerjs/hammer.js/issues/968

dae:
(If a solution is not quickly found, I'm happy to merge this PR without the Mac pinch-zoom changes)

ts/image-occlusion/tools/tool-zoom.ts
```diff
@@ -156,26 +156,50 @@ const onDrag = (canvas, opt) => {
...
+    canvas.lastPosY -= deltaY;
     canvas.setViewportTransform(vpt);
     constrainBoundsAroundBgImage(canvas);
     redraw(canvas);
 };
 
-export const onDragY = (canvas: fabric.Canvas, opt: WheelEvent) => {
-    const delta = opt.deltaY;
+// for mac trackpad
+export const onGesture = (event) => {
```
krmanik, line 177: I have implemented this solution but currently I do not have access to hardware to test this for trackpad, I request you test it, it new changes required, it may be done in this method.

krmanik:
Now supported feature is scroll using mouse wheel up-down and using alt + mouse wheel in left right.
The touchpad (not trackpad, as not tested so don't know) can be used scroll and zoom without any key.
Remove zoom tool because it is redundant, because now panning can be done by CTRL + mouse drag, for touch screen device two touch point can be used to zoom and pan.

ts/image-occlusion/index.ts
```diff
@@ -25,6 +30,16 @@ export async function setupImageOcclusion(mode: IOMode, target = document.body):
...
 
+    async function addNote(): Promise<void> {
+        addOrUpdateNote(mode, get(hideAllGuessOne));
+    }
+
+    // for adding note from mobile devices
+    globalThis.anki.imageOcclusion = {
+        mode,
+        addNote,
+    };
```
krmanik, line 41: For mobile device a button at top corner needs to add then this global method needs to call for adding/updating notes.

dae:
Unfortunately this approach doesn't work. It looks like the gesture events are WebKit-specific. We use QtWebEngine even on Mac, so the ongesture* events don't fire.

The change in your latest commit does cause two fingers to scroll, as onWheelDrag is being called, but currently the panning speed is very fast and difficult to control. It seems a very quick two-finger flick will cause wheel events to fire for about 2 seconds, so the image quickly scrolls off the screen. I tried capping the delta to 1 or 5 so the movement is slower, but then it looks laggy when the image keeps moving for 1-2 seconds after you release your fingers from the trackpad.

I also tested this code on iOS, and unfortunately it doesn't work there either - the image quickly disappears when you attempt to pinch or two-finger drag.

So unfortunately it seems like this approach will probably not work, but I do appreciate you looking into it. Could we roll back the macOS/iOS changes to the commit before 'drag with touchpad'? If the changes made there are necessary for other platforms, a platform-specific check would be fine.

krmanik:
OK, I'll look into this again, specifically for mac and iOS, and push a fix or revert changes by tomorrow.

dae:
Thanks Mani. I'll get a new beta out within a few days of merging this in.

krmanik:
The issue in mac is solved in last commit, it is required to add `preventDefault` and `passive: false`. It works but there is some lag but when AnkiWebView Inspector is opened there is no lag, as it can be seen in attached demo below.
For iOS we do not need gesture event, it is working better without any changes, so I have removed gesture event implementation.

macOS demo

https://github.com/ankitects/anki/assets/12841290/a0b8522b-b3b8-4f20-b811-b4d3fbf47a20

iOS demo

https://github.com/ankitects/anki/assets/12841290/ff43ed1e-08ab-400f-a9eb-94bca1ea9751

glutanimate:
Thanks for the updates, Mani! Following up on the desktop issues in #3068:

- Fitting the image to the editor area on resize works perfectly now, nice work!
- Scrolling also works well now, both horizontally and vertically. However, it seems like triggering zoom via the mouse wheel will cause the mouse wheel to be stuck in zoom/pan mode and disable scrolling for the rest of the session.
- I know you mentioned that you to had to switch to Shift+Mouse wheel due to an issue with assigning Ctrl. However, if you think there is a way to make Ctrl work still, I think it would be worth the effort to look into. Ctrl+wheel will be a lot more intuitive to users as it's used in so many other places as the default for zooming (even Anki itself).

    What was the issue your ran into with Ctrl and panning, specifically?
- Along the same lines, do you think it would be feasible to add scrollbars back under the new zoom/pan implementation? I would say this is another interaction pattern users would expect and likely be missing quite a bit. Consider that, for users who primarily use their mouse, this is likely the main way they pan through zoomed-in content. Scrollbars also serve as a nice indicator for the current zoom state.

Also, thanks for the removal of the red border, appreciate it!

krmanik:
> What was the issue your ran into with Ctrl and panning, specifically?

In mac, CTRL+Click open up context menu which is conflicting, there are event like `contextmenu` or following, I have tried but didn't get better result. So, I have used shift for mac and control for pc.

```js
document.addEventListener("contextmenu",function(event){

});

// or

document.addEventListener("mousedown",function(event){
    if( event.ctrlKey || event.button == 2 ) {
    }
});
```
https://stackoverflow.com/questions/14725824/any-way-to-detect-ctrl-click-in-javascript-for-osx-browsers-no-jquery

> Along the same lines, do you think it would be feasible to add scrollbars back under the new zoom/pan implementation?

I have tried to fix this also using css overflow, it seems we have to implement other approach, which I will fix the scrollbar in follow up PR.

dae:
>  It works but there is some lag but when AnkiWebView Inspector is opened there is no lag

The difference is quite noticeable - I wonder what's causing that? I played around briefly with trying to throttle setViewportTransform() etc or call it in an animation frame, but it didn't seem to make a big difference. I also noticed that diagonals appear to be a bigger problem - if you move directly up/down or directly left/right on the trackpad, it's pretty smooth, but if you move on a diagonal, it gets quick choppy. Weird!

Understand if we can't fix it - it's still more convenient than the separate panning mode.

> However, it seems like triggering zoom via the mouse wheel will cause the mouse wheel to be stuck in zoom/pan mode and disable scrolling for the rest of the session.

I'm away from a computer with a mouse at the moment so can't test this. It sounds like this may need fixing before the PR is merged?

krmanik:
> However, it seems like triggering zoom via the mouse wheel will cause the mouse wheel to be stuck in zoom/pan mode and disable scrolling for the rest of the session.

I just got access to my system and tested for mouse wheel and it works as expected, @glutanimate please try again, it has been fixed in latest commit.

For the scrollbar, I have explored the fabric.js issues and found that this is an active issue [fabricjs #3164](https://www.github.com/fabricjs/fabric.js/issues/3164). Nevertheless, I will fix this issue in the next pull request.

dae:
Apparently diagonals not working is expected for the wheel event: https://stackoverflow.com/questions/56825793/javascript-wheel-event-locks-to-horizontal-or-vertical-movement-and-does-not-all

Current state of the PR seems ok on my system - once @glutanimate feels this is ready I'll merge it in.

krmanik:
But when Ankiweb inspector addon opened there is no issues.

glutanimate (approved):
All good on my side as well, everything is working well :+1: 

Splitting the scrollbar work off into a follow-up makes sense to me.

-----

Some more thoughts regarding a potential follow-up:

- Regarding **panning**: Using different bindings across different platforms is reasonable workaround, but could be confusing to users and I think we should avoid it if we can.

    Another approach could be to either go with holding the middle mouse button or right mouse button to drag. This seems to be a common pattern in drawing apps (e.g. Inkscape, Krita, various websites). Another common assignment seems to be Space + LMB drag (Krita, Illustrator).

   RMB-dragging seems like it could be annoying and cause ambiguity with context menu bindings. My personal preference would be to go the [Krita way](https://docs.krita.org/en/reference_manual/tools/pan.html) and implement both MMB drag and Space + LMB drag. That should give users an option they will likely be familiar with, regardless of which other app they came from - and also one that works for all mouses.

     Finally, a nice to have would be to have the mouse pointer switch to a `grabbing` state while panning is active.

- Regarding **zooming**: If the panning situation is resolved, I would recommend to universally go with Ctrl+Wheel on all platforms

- Regarding **scrolling**: Similarly, I would suggest to switch the horizontal scroll binding to Shift+Wheel, as that seems to be the more common convention (Inkscape, Chrome, etc.).

dae:
Thank you both!

krmanik:
I will implement as per suggestion in next PR. 
Thanks @glutanimate

dae:
In terms of release timing, I was thinking of making another rc tomorrow. If the next PR is close I can wait an extra day or two; if you think it will take longer, I might make another rc after it's ready.

krmanik:
The scrollbar will take some time but the way to pan/zoom suggested by glutanimate will be completed by tomorrow.

krmanik:
I have just tried to implement this patch but it seems Space+Touchpad have issue now, I think it is better to use Ctrl+Wheel, because it solves two issues, Ctrl+Wheel = zoom in, zoom out and Ctrl+Mouse move = panning. The glutanimate suggestion is good but I have complicated implementation, it will complicate more. Also, horizontal scroll can be done by ALT+Wheel.

Following patch works great with external mouse but it is causing issue with Space+Touchpad (on windows).
For me, the way forward would be solving macOS context menu issue, middle button for panning and using Shift + Wheel for horizontal scroll in all platforms.

```diff
diff --git a/.gitignore b/.gitignore
index 4c43afd06..2523b743e 100644
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ node_modules
 .ninja_deps
 /extra
 yarn-error.log
+svelte-kit/
\ No newline at end of file
diff --git a/ts/image-occlusion/Toolbar.svelte b/ts/image-occlusion/Toolbar.svelte
index 64e251d98..463fc3f6e 100644
--- a/ts/image-occlusion/Toolbar.svelte
+++ b/ts/image-occlusion/Toolbar.svelte
@@ -55,31 +55,31 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     let dbclicked = false;
     let move = false;
     let wheel = false;
-    const controlKey = isApplePlatform() ? "Shift" : "Control";
+    const controlKey = " ";
 
     onMount(() => {
         window.addEventListener("mousedown", (event) => {
-            if (event.ctrlKey) {
+            if (event.key === " ") {
                 clicked = true;
             }
         });
         window.addEventListener("mouseup", (event) => {
-            if (event.ctrlKey) {
+            if (event.key === " ") {
                 clicked = false;
             }
         });
         window.addEventListener("mousemove", (event) => {
-            if (event.ctrlKey) {
+            if (event.key === " ") {
                 move = true;
             }
         });
         window.addEventListener("wheel", (event) => {
-            if (event.ctrlKey) {
+            if (event.key === " ") {
                 wheel = true;
             }
         });
         window.addEventListener("dblclick", (event) => {
-            if (event.ctrlKey) {
+            if (event.key === " ") {
                 dbclicked = true;
             }
         });

```

Also, typescript tells that mouse event does not have key but it works,
image

PR #3067, opened 2024-03-10
Adding a note to the import options to indicate that leech and marked tags will be removed if disabled

arbyste:
When importing a deck with "Import any learning progress" disabled, "leech" and "marked" tags are removed from the imported notes. This pull request adds this information to the info tooltip.


dae:
Thank you!

PR #3072, opened 2024-03-12
speed up computing optimal retention

L-M-Sherlock:
I promise this PR is the last one related FSRS-rs before the release of Anki 24.04.

- https://github.com/open-spaced-repetition/fsrs-rs/pull/168

Before this PR, it costs ~47s to compute optimal retention when `Days to simulate` is 3650. Now it is ~36s. It's nearly 24% faster than the previous one.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1792,9 +1792,9 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "0.5.3"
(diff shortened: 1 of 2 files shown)


dae:
No need to promise; if you notice other problems that need fixing before 24.04 goes out, that's fine. I'll aim to get another build out within a week.

PR #3077, opened 2024-03-17
Switch to SvelteKit

dae:
Our current build process uses Svelte directly, and bundles each page (e.g.
deck options, card info, etc) into a separate "page" (.html, .js and .css file).
This PR switches us over to SvelteKit, and builds all of the pages into a
single bundle, in a single go.

There are a number of trade-offs with his approach. The positives first:

- We now rely on SvelteKit's routing instead of manually initializing pages.
So instead of loading deck-options.html and then calling anki.setupDeckOptions()
or similiar, we instead surf to `/deck-options/<deck id>`, and mediasrv.py
takes care of serving the single .html/.js entrypoint. SvelteKit then
displays the page. This opens the door to navigating from one page to another
without flashing/slow-loading, which we will hopefully take more advantage of
in future Anki updates.
- Hot module reload (HMR) is now supported when developing, so the edit/build/test
cycle is faster. Ensure Anki's running, run './yarn dev' in a new terminal,
then surf to e.g. http://localhost:5173/graphs.
- Vite comes with its own testing framework, vitest, which was mostly a drop-in
replacement for Jest. It's faster, and has far fewer dependencies. It also works
properly with live checking now: ./yarn vitest to watch for changes.
- TypeScript checking seems to be improved, and there were a number of errors
that needed working around.

There are a number of downsides, too:

- Building Anki requires doing a production SvelteKit build, which is slow:
something like 13s at the moment, compared to the sub-4s that rebuilding
an individual ts page takes. And because everything is built at once,
any change inside ts/ is going to trigger that 13s rebuild. This slowdown
won't be felt when working on the pages with HMR, but will be felt when
switching betweens PRs/bisecting.
- svelte-check and eslint are now run across the whole codebase at once
too. They're a bit slower too, though they waste less CPU cycles parsing
the same shared libraries multiple times, which may help when running
multiple tests at once.
- The mobile clients will need to be updated to pull files in from the new
location, and use them instead of the old individual pages.

The significantly-slower build step gives me pause, but this change also
brings a lot of nice stuff. 

@abdnh @glutanimate @RumovZ @krmanik and others, what do you think?

(drafted because the PR still needs polish, and I'd like to try solve the issue where
the sveltekit folder is written outside of out/)


abdnh:
> Hot module reload (HMR) is now supported when developing, so the edit/build/test
cycle is faster. Ensure Anki's running, run './yarn dev' in a new terminal,
then surf to e.g. http://localhost:5173/graphs.

This is great! Can we make this work for refreshing the in-app pages?

RumovZ:
Seems like a reasonable trade-off and a move towards more modern web technologies like client-side routing. And having worked more with Rust than the frontend, a 13 seconds build step doesn't particularly scare me. :)

glutanimate:
Really like the improvements. Builds unfortunately seem to be heavily CPU-bound – on a machine with an older i7, the previous 3s become 30s. If we could get HMR to work within the context of Anki, that would alleviate things greatly for parallel add-on/Anki API development (which in most cases will have to happen in-app).

dae:
Ok, you can now define HMR=1, and ./run will a) not build SvelteKit and b) use a running dev instance to display pages inside Anki, so they'll auto-refresh. You'll be responsible for starting `./yarn dev` yourself (at least for now - some fancy tmux setup might be nice in the future).

Would someone on Windows mind confirming that files are importing correctly in Windows? I didn't check the paths are being handled correctly in the URL.

abdnh:
> Would someone on Windows mind confirming that files are importing correctly in Windows? I didn't check the paths are being handled correctly in the URL.

Importing is working for me.

krmanik:
This is really good to have this in one place. I have checked for import on windows and it run correctly.

I have also tried for moving `.svelte-kit` to `out` dir by exploring almost all similar to this in svelte kit repo, I also not able figured it out. I keep experimenting, there may be some solution that may work.

asukaminato0721:
> ... and eslint are now run across the whole codebase at once
> too. They're a bit slower too

maybe try <https://oxc-project.github.io/docs/guide/usage/linter.html> ?

dae:
Until it gets all the features eslint has (including the svelte plugin for eslint), that unfortunately won't help much (except as a quicker first-pass).

abdnh:
`tools\ninja.bat format:dprint` is failing for me after this PR with the following error:
```
n2: error: out\build.ninja:405: input node_modules\@sveltejs\kit\src\types\synthetic\+dynamic+private.md missing
```

Not a big deal since I have format on save enabled, but maybe someone has an idea about it.

krmanik:
May be,

https://github.com/ankitects/anki/pull/3109/commits/b04ef665decc56b4c4e435770e0bff10742e8ba9

abdnh:
That's it. Thanks @krmanik!

PR #3079, opened 2024-03-17
fix: custom study today's new/review card limit

fernando-gap:
* fix: "Increase today's new/review card limit: don't increase/decrease more than what's enough.
* add: `new_count` and `new_review` fields to `scheduler.proto`, so it is available by a custom study request by default.

Increasing/decreasing the spin of "increase today's new/review limit by" by more than the current available cards (e. g. 99999) consecutively will cause the review limit of the deck to be too large or too negative, causing the next custom study review/new limits to not correctly adjust for smaller cards when decreasing or increasing.

It's essentially possible to create an integer overflow/underflow as per the lack of checking in `review_limit` at `limits.rs`.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -165,6 +165,7 @@ Jean Khawand <jk@jeankhawand.com>
 Pedro Schreiber <schreiber.mmb@gmail.com>
 Foxy_null <https://github.com/Foxy-null>
 Arbyste <arbyste@outlook.com>
+Nandoka <nandoka1957@gmail.com>
 ********************
 
 The text of the 3 clause BSD license follows:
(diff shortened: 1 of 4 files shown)


dae:
I appreciate you taking the time to contribute, but I'm not sure about this one. When you say "underflow", do you mean that the number goes below zero, or do you mean that it wraps around? The former is expected - one reason that negative numbers have been allowed in custom study is that it allows users to undo a positive change they made previously. 

The extend limits feature is also likely to be retired in the future, so it's better handled with the today-only limit that was introduced in v3. For this reason, I'm reluctant to touch working code (and this Python-based solution will mean separate implementations for the mobile clients).

rileygreen022:
Any chance you could add this?

https://forums.ankiweb.net/t/ability-to-exclude-recently-again-cards-when-reviewing-ahead/42748

dae:
Please keep feature requests in the forum, not here.

PR #3080, opened 2024-03-18
better pan zoom with space for drag, shift for horizontal scroll and ctrl for zoom

krmanik:
> Some more thoughts regarding a potential follow-up:
>
> Regarding panning: Using different bindings across different platforms is reasonable workaround, but could be confusing to users and I think we should avoid it if we can.
>
> Another approach could be to either go with holding the middle mouse button or right mouse button to drag. This seems to be a common pattern in drawing apps (e.g. Inkscape, Krita, various websites). Another common assignment seems to be Space + LMB drag (Krita, Illustrator).
>
> RMB-dragging seems like it could be annoying and cause ambiguity with context menu bindings. My personal preference would be to go the [Krita way](https://docs.krita.org/en/reference_manual/tools/pan.html) and implement both MMB drag and Space + LMB drag. That should give users an option they will likely be familiar with, regardless of which other app they came from - and also one that works for all mouses.
>
> Finally, a nice to have would be to have the mouse pointer switch to a grabbing state while panning is active.
>
> Regarding zooming: If the panning situation is resolved, I would recommend to universally go with Ctrl+Wheel on all platforms
>
> Regarding scrolling: Similarly, I would suggest to switch the horizontal scroll binding to Shift+Wheel, as that seems to be the more common convention (Inkscape, Chrome, etc.).

https://github.com/ankitects/anki/pull/3066#pullrequestreview-1940796645

### DIFF
--- a/ts/image-occlusion/Toolbar.svelte
+++ b/ts/image-occlusion/Toolbar.svelte
@@ -32,7 +32,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import { drawCursor } from "./tools/tool-cursor";
     import { removeUnfinishedPolygon } from "./tools/tool-polygon";
     import { undoRedoTools, undoStack } from "./tools/tool-undo-redo";
-    import { disableZoom, enableZoom, onWheelDrag } from "./tools/tool-zoom";
+    import { disablePan, disableZoom, enablePan, enableZoom, onWheelDrag, onWheelDragX } from "./tools/tool-zoom";
 
     export let canvas;
     export let iconSize;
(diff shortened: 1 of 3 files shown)


ts/image-occlusion/Toolbar.svelte
```diff
@@ -51,75 +51,87 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
+    const spaceKey = " ";
+    const controlKey = "Control";
+    const shiftKey = "Shift";
 
     onMount(() => {
         window.addEventListener("mousedown", (event) => {
-            if (event.ctrlKey) {
-                clicked = true;
+            if (event.key === spaceKey) {
+                spaceClicked = true;
```
krmanik, line 67: The  `keydown` event can be chained here so key error will be fixed.

ts/image-occlusion/Toolbar.svelte
```diff
@@ -51,75 +51,87 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
             }
-        });
-        window.addEventListener("mouseup", (event) => {
-            if (event.ctrlKey) {
-                clicked = false;
-            }
-        });
-        window.addEventListener("mousemove", (event) => {
-            if (event.ctrlKey) {
+            if (event.which === 2) {
```
krmanik, line 69: This is deprecated which I think look for other solution.

krmanik:
It is ready for review now, I have implemented exactly @glutanimate suggested except panning using middle mouse button implementation (because zero width/height shapes added to canvas when using middle mouse button). I have tested the current PR in Android, iOS, windows and macOS.

Now controls are,
- Spacebar + Mouse move = Panning
- Ctrl + Mouse wheel = Zoom in/out
- Shift + Mouse wheel = Horizontal scroll

dae:
Thanks for your work on this Mani. It seems to work well in my tests, so I'll merge this in for the next rc.

PR #3082, opened 2024-03-20
Remove unneeded inputs for PythonEnvironmentStub and add documentation for offline build

antecrescent:
The `python_binary` and `pyenv_folder` targets, used to create a fully-fledged venv, aren't needed if `NO_VENV` is set and `setup_venv_stub` is called.
Consequently, the `setup_python` function, which only adds a dependency on the former target, can be called conditionally. This allows users of `NO_VENV` to unset the `PYTHON_BINARY` environment variable altogether.

I also (mostly) copied the commit text of 42cc2c913ce53b5057ce890edb0b4b09d73c8dfe to the documentation section for package managers, since it seemed important enough to me :)


antecrescent:
I don't really understand, how this build failure is related to my changes.

dae:
Thank you! I've rerun the tests - you can also do so by closing+reopening this PR. Please fix the minor formatting error, then I'm happy to merge it in.

dae:
You marked this as a draft - you don't want me to merge it yet?

antecrescent:
I must admit that I didn't test this properly. With these changes, `NO_VENV=1 ./ninja wheels` panics when running the `configure` crate, because `build_bundle` tries to setup a regular Python venv, but since `setup_python` has never run, the build is missing a rule for`:python_binary`.

My first idea was to simply add a guard clause like
```
if env::var("NO_VENV").is_err() {
        build_bundle(build)?;
}
```
This would make sense, given that the bundling process expects an actual Python venv to be present, which the user tries to prevent in the first place.
On the other hand, `NO_VENV` currently only controls the creation of the "main" venv at `out/pyenv` and I don't want to disturb/overcomplicate the official bundling process.

I fear, that I'm gravely overthinking the issue at hand and I'd be happy about your input!

In any case, I'd like to test my changes a bit more thoroughly before marking them as "ready for review/merge"!

dae:
Maybe you could use a similar approach to this:

https://github.com/ankitects/anki/pull/3077/commits/318ef31ba6f0342d998d59e8a8cefa68e71eedee

That said, the bundle action is only used for the official builds, and I'll never be setting NO_VENV, so this may not affect anyone in practice.

antecrescent:
TLDR: I ended up digging deeper and solving the problem of (un)needed inputs by simplifying the offline build process through elimination of `NO_VENV`. Is this okay?

While playing around with several approaches to masking/altering `build_bundles`, when `NO_VENV` is set, I noticed the tight coupling of `NO_VENV` and `OFFLINE_BUILD`.
Together, they add support for offline builds, but I looked into boiling them down into one, since:
- `OFFLINE_BUILD` implies `NO_VENV`, the regular Python venv needs network access for  `pip_sync` to work and in some cases, the offline build depends on system site-packages.
- I'm not sure if `NO_VENV` without `OFFLINE_BUILD` is a use case worth supporting. I only came up with one scenario:
    - Pypi's Python wheels may clash with system packages, which may frustrate or even deter new developers. In the case of Gentoo Linux, Pypi's PyQt6 wheels are often incompatible with the underlying Qt6 suite, which can be heavily customized at build-time.

        On the other hand, compatibility can be achieved through setting the correct compile flags and development in a containerized environment is always an option.

(Ping @knobix, if you have some insights :) )

I chose to eliminate `NO_VENV` and keep `OFFLINE_BUILD`, because the latter makes it more obvious, what is intended to happen, when one sets it.
Additionally, I chose to guard `build_bundles` behind `!OFFLINE_BUILD` because I did not want to bother trying to provide an offline variant, when it may not affect anyone in practice and it seemed like the easiest (correct) solution.

dae:
Thanks, will wait to see what @knobix thinks.

docs/linux.md
```diff
@@ -84,6 +84,57 @@ PYTHON_BINARY, NODE_BINARY, YARN_BINARY and/or PROTOC_BINARY to use locally-inst
...
+   ```
+
+5. Build Anki:
+
+   ```
+   /path/to/cargo build --package runner --release
+   OFFLINE_BUILD=1 \
+   out/rust/release/runner build wheels
+   ```
+
```
knobix, line 137: Maybe it makes sense to add here an additional point, e.g. 

6. Build documentation with Sphinx (optional)
```
out/rust/release/runner build python:sphinx
```

knobix:
Thank you, @antecrescent , for these improvements! I actually had it on my list to write a little documentation, which I can now tick off. :)

I was able to build Anki successfully with these improvements and I only have one small suggestion (see inline comment for `linux.md`) regarding the documentation, otherwise it looks good to me.

docs/linux.md
antecrescent, line 137: How about something like this instead?
```suggestion
You are now ready to build wheels and Sphinx documentation fully
offline.
```
I think, I should try not to duplicate instructions on how to build wheels and Sphinx documentation. Both of these are already describe in `docs/development.md` and this file is mentioned in the very next section.
Also, if anything, step 5 should probably read `OFFLINE_BUILD=1 ./ninja ...` since no other documented installation command exposes the build systems internals.

antecrescent: How about something like this instead?
```
You are now ready to build wheels and Sphinx documentation fully
offline.
```
I think, I should try not to duplicate instructions on how to build wheels and Sphinx documentation. Both of these are already describe in `docs/development.md` and this file is mentioned in the very next section.
Also, if anything, step 5 should probably read `OFFLINE_BUILD=1 ./ninja ...` since no other documented installation command exposes the build systems internals.

knobix: Makes sense and your suggestion looks fine for me.

antecrescent: Thanks for your input!

dae:
Thank you both!

antecrescent:
Thank you!

PR #3087, opened 2024-03-22
use rmse in comparing parameters

L-M-Sherlock:
The FSRS has update its metric in this PR:
- https://github.com/open-spaced-repetition/fsrs-rs/pull/160

So it's unnecessary to use the universal metric here.

### DIFF
--- a/rslib/src/scheduler/fsrs/weights.rs
+++ b/rslib/src/scheduler/fsrs/weights.rs
@@ -93,10 +93,12 @@ impl Collection {
             }
         });
         let fsrs = FSRS::new(Some(current_weights))?;
+        let current_rmse = fsrs.evaluate(items.clone(), |_| true)?.rmse_bins;
         let mut weights =
             fsrs.compute_parameters(items.clone(), fsrs_items < 1000, Some(progress2))?;
-        let metrics = fsrs.universal_metrics(items, &weights, |_| true)?;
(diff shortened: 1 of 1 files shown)


user1823:
But, is the new RMSE actually better than the universal metric for comparing the two sets of parameters?

L-M-Sherlock:
The new RMSE is less likely to cheat and still easy to understand:
https://github.com/open-spaced-repetition/fsrs4anki/wiki/The-Metric

And keeping the consistency between evaluation and comparing parameters could reduce the confusion of users.

Expertium:
> But, is the new RMSE actually better than the universal metric for comparing the two sets of parameters?

The way I see it, both the new RMSE and UM are uncheatable, so we could use either. But there are two reasons to use RMSE:
1) It's more intuitive
2) It's confusing if we use different metrics "externally" (what is shown to the user) and "internally" (how two sets of parameters are compared)

Also, I had the following "bug", which is likely due to UM:
1) Optimize parameters
2) Evaluate parameters
3) Reset them to default
4) Optimize them again and obtain new parameters
5) Evaluate the new parameters

New parameters after the reset result in lower RMSE and lower log loss.

user1823:
Ok, so let's use the new RMSE

dae:
Thanks guys.

Issue #3078, opened 2024-03-17
Forget→Reset


dae:
> The mobile clients (both AnkiMobile and AnkiDroid) use the word “Reset” while the Desktop client uses the word “Forget”. I feel that the terms should be unified across platforms to decrease confusion.

Originally reported on https://forums.ankiweb.net/t/unify-the-terms-for-reset-card/42508

vasll:
I looked into this and the quick approach would be to change these key values in the .ftl language files:
- at `ftl\core\actions.ftl`: `actions-forget-card = Forget card` into `Reset card`
- at `ftl\qt\qt-accel.ftl`: `qt-accel-forget = &Forget` into `&Reset`

But code-wise I don't know if it would still have sense to call this feature `forget`. To "logically" change the `forget` function to be called a `reset` function across the codebase many different files should be refactored. I have already locally tried this across and it compiles & passes the tests. Few .py, .rs. ui files were refactored

One last thing before sending the PR: I modified files in two subrepos (the .ftl files), but I committed the change only to my local `main` repo. Do I also have to create forks for those subrepos and commit there? It says _"when your PR is accepted, the strings will be automatically added here."_ but since my files from the subrepos are still uncommitted I'm not sure how that works

dae:
Only those two files need modifying. You can keep the key name the same, as we don't want to force a retranslation for the other languages. I will sync the changes to the other repos once this gets merged in.

Have you checked that &R is not used by another menu item already?

vasll:
> Have you checked that &R is not used by another menu item already?

&R is already used in `qt-accel-redo` and `qt-accel-reset-zoom`, so there would be 3 items with the same qt accelerator key, what would you suggest to do? For the string "Reset" each character is at least used 2 times as an accelerator key already

dae:
If I recall correctly, they only need to be unique for menu items in the same top-level menu - so one in the the Edit menu and one in the Cards menu shouldn't be a problem, for example.

vasll:
At the top-level only the `Reset` options uses the R key so it's fine to use `&Reset` as a string in `qt-accel-forget`. I'm new to qt so I didn't know, thanks for the patience

dae:
No problem, sorry for not being clear the first time! A PR would be appreciated.

dottedmag:
This FAQ answer still refers to the action as "Forget": https://faqs.ankiweb.net/resetting-progress-in-a-deck.html

dae:
I've pushed an update, thanks.

PR #3088, opened 2024-03-22
Refactored "Forget" to "Reset" (#3078)

vasll:
Solves issue #3078 

The respective *.ftl file keys `actions-forget-card` and `qt-accel-forget` will keep the same name to prevent having to force a retranslation for other languages, only the string value was changed


dae:
Thank you!

PR #3089, opened 2024-03-23
Reset original id when cloning notetypes

RumovZ:
https://forums.ankiweb.net/t/merge-note-types-on-import-bug/42686

The fix is trivial, but I'm unsure what to do about collections that contain those bad original ids now. 😕 
Do you have an idea? Should merged notetypes be listed so users at least can revert the import? Should there even be an option to reset a given original id?

### DIFF
--- a/pylib/anki/models.py
+++ b/pylib/anki/models.py
@@ -259,6 +259,7 @@ class ModelManager(DeprecatedNamesMixin):
             self.col.tr.notetypes_copy(val=cloned["name"])
         )
         cloned["id"] = 0
+        cloned["originalId"] = None
         if add:
             self.add(cloned)
         return cloned


dae:
I'm not sure to be honest. With only one report of a user encountering this problem so far, I don't know how prevalent it's likely to be at this point. 

> Should merged notetypes be listed so users at least can revert the import?

Sorry, I'm not sure what you mean by this - where would they be listed? Where would such an option be?

dae:
(thank you for the quick fix, by the way!)

RumovZ:
An overview of any merged notetypes could be shown on the import log page. But I'm having a hard time imagining how an option to reset original ids could look like. It wouldn't be helpful on the import log page when merging has already taken place.

I also hope it doesn't affect that many users. If you're not too concerned, I'm fine with just keeping an eye on it for now, too.

dae:
Let's take a wait and see approach. Appreciate the suggestions though!

PR #3092, opened 2024-03-24
Fix summary string for notes failed to be imported

RumovZ:
https://forums.ankiweb.net/t/bug-csv-import-error-bc-merge-notetype-disabled/40644

This makes the summary string more generic, and moves the part concerning notetype merging into the tooltip for actually affected notes.


dae:
Thanks!

PR #3093, opened 2024-03-24
Disable ligatures in type-in highlighting

### LINKED ISSUE #3033: Ligatures misunderstood by text comparator
[Ligature Problem.apkg.zip](https://github.com/ankitects/anki/files/14397407/Ligature.Problem.apkg.zip)

If a typed answer containing an incorrect ligature is submitted, the text comparator only sees the correct part of the ligature and ignores the incorrect part, highlighting the incorrect answer in green. This has been observed with incorrect "fi" being matched as green with back "f", but presumably occurs with other ligatures as well.

To see an example, import the attached deck, which contains one card with the front:

> The back of this card is "f". Type "fi" to see that the "i" is highlighted in green even though it is an error.

and back:

> f

Follow the instructions on the card, and you will see the problem.

Danika_Dakika on the forum has already found a solution. One just needs to add the following to the card styling (under .card):

>  font-variant-ligatures: no-common-ligatures;

However, it would probably good if the problem didn't arise in the first place.

abdnh:
Closes #3033

### DIFF
--- a/ts/reviewer/reviewer.scss
+++ b/ts/reviewer/reviewer.scss
@@ -75,6 +75,7 @@ pre {
 
 code#typeans {
     white-space: pre-wrap;
+    font-variant-ligatures: none;
 }
 
 .typeGood {


dae:
Thanks Abdo!

PR #3095, opened 2024-03-25
Accept custom root CA in the backend

voczi:
Allow a custom root certificate to be pinned when creating the backend.

### DIFF
--- a/rslib/src/backend/mod.rs
+++ b/rslib/src/backend/mod.rs
@@ -23,6 +23,7 @@ use std::thread::JoinHandle;
 use futures::future::AbortHandle;
 use once_cell::sync::OnceCell;
 use prost::Message;
+use reqwest::Certificate;
 use reqwest::Client;
 use tokio::runtime;
 use tokio::runtime::Runtime;
@@ -56,7 +57,7 @@ pub struct BackendInner {
     state: Mutex<BackendState>,
     backup_task: Mutex<Option<JoinHandle<Result<()>>>>,
     media_sync_task: Mutex<Option<JoinHandle<Result<()>>>>,
-    web_client: OnceCell<Client>,
+    web_client: Mutex<Option<Client>>,
 }
 
 #[derive(Default)]
@@ -91,7 +92,7 @@ impl Backend {
             state: Mutex::new(BackendState::default()),
             backup_task: Mutex::new(None),
             media_sync_task: Mutex::new(None),
-            web_client: OnceCell::new(),
+            web_client: Mutex::new(None),
         }))
     }
 
@@ -137,10 +138,38 @@ impl Backend {
             .clone()
     }
 
-    fn web_client(&self) -> &Client {
-        // currently limited to http1, as nginx doesn't support http2 proxies
-        self.web_client
-            .get_or_init(|| Client::builder().http1_only().build().unwrap())
+    fn set_custom_cert(&self, cert_str: String) -> Result<()> {
+        if let Ok(certificate) = Certificate::from_pem(cert_str.as_bytes()) {
+            let mut client_mutex = self.web_client.lock();
+
+            if let Ok(ref mut web_client) = client_mutex {
+                let _ = (**web_client).insert(
+                    Client::builder()
+                        .use_rustls_tls()
+                        .add_root_certificate(certificate)
+                        .http1_only()
+                        .build()
+                        .unwrap(),
+                );
+                return Ok(());
+            } else {
+                return Err(AnkiError::Interrupted);
+            }
+        }
+
+        Err(AnkiError::InvalidCertificateFormat)
+    }
+
+    fn web_client(&self) -> Result<Client> {
+        let mut client_mutex = self.web_client.lock();
+
+        if let Ok(ref mut web_client) = client_mutex {
+            return Ok((**web_client)
+                .get_or_insert(Client::builder().http1_only().build().unwrap())
+                .clone());
+        } else {
+            return Err(AnkiError::Interrupted);
+        }
     }
 
     fn db_command(&self, input: &[u8]) -> Result<Vec<u8>> {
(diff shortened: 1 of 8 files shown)


rslib/src/backend/mod.rs
```diff
@@ -72,12 +74,21 @@ pub fn init_backend(init_msg: &[u8]) -> result::Result<Backend, String> {
         };
 
     let tr = I18n::new(&input.preferred_langs);
+    let cert_str = input.custom_cert;
 
-    Ok(Backend::new(tr, input.server))
+    if cert_str.is_some() {
+        let cert = Certificate::from_pem(cert_str.unwrap().as_bytes());
+
+        if cert.is_ok() {
```
dae, line 82: Better to crash if user provided an invalid cert, instead of silently ignoring it.

dae (commented):
I'd prefer not to change the Backend constructor. Could you make this work with an env var instead? AnkiDroid can then ensure it's set at startup, and the env var colud be checked in web_client().

voczi:
> Could you make this work with an env var instead?

Yeah, good idea. I should be able to leverage JNI and create a new Rust method that sets this as an env var.

rslib/src/backend/mod.rs
dae: (to clarify: if the user is directly setting the env var, crashing makes sense. On mobile clients, some validation on the frontend would make sense)

dae:
No need for JNI - it can be done from Kotlin. https://github.com/ankidroid/Anki-Android/blob/85abd410bb517e527dbf4d4e8dd49d109ab007ab/AnkiDroid/src/main/java/com/ichi2/anki/AnkiDroidApp.kt#L93

voczi:
> No need for JNI

What are your thoughts on exposing a JNI method for this, but for the purpose of validating the input so the frontend knows if the certificate isn't properly formatted?

dae:
JNI methods are not very ergonomic to write/maintain, which is why rsdroid's API consists of only a few different methods. There is merit to doing the checking in Rust though. Maybe doing a check/unwrap would be sufficient? Any panics should be caught in rslib_bridge, and can be mapped to a Kotlin error in panic_to_backend_error(). 

If you're not so worried about validating at run time, and just when the config is changed, then another option would be to add a backend method to something like sync.proto that validates a certificate and returns true/false. The frontend clients could use that to check the value is valid before they persist it.

voczi:
I tried another approach. Have a look whenever you have time.

dae:
- I don't understand why you're using try_lock instead of lock, and introducing a WebClientConflict error. The inner client can be cheaply cloned, so lock+clone should not block for long.
- Instead of web_client() returning an option, it could return an error, and then all the calling locations could just bubble it up.

voczi:
Ok, I've done some more changes now.

dae:
Please remove the backend conflict case. You should .lock().unwrap() the mutex; it will only be invalid if it's been poisoned.

rslib/src/backend/mod.rs
```diff
@@ -137,10 +138,40 @@ impl Backend {
...
     }
 
-    fn web_client(&self) -> &Client {
-        // currently limited to http1, as nginx doesn't support http2 proxies
-        self.web_client
-            .get_or_init(|| Client::builder().http1_only().build().unwrap())
+    fn set_custom_cert(&self, cert_str: String) -> Result<()> {
+        let cert_conv = Certificate::from_pem(cert_str.as_bytes());
+
+        if cert_conv.is_ok() {
```
dae, line 144: Instead of a check + unwrap, you should use if let. There are a few similar cases that also need updating.

voczi:
Should look better now.

rslib/src/backend/mod.rs
```diff
@@ -137,10 +139,39 @@ impl Backend {
...
+                    Client::builder()
+                        .use_rustls_tls()
+                        .add_root_certificate(certificate)
+                        .http1_only()
+                        .build()
+                        .unwrap(),
+                );
+                return Ok(());
+            } else {
+                return Err(AnkiError::Interrupted);
```
voczi, line 158: Not sure if I should be returning this, or anything at all, here. 🤔

rslib/src/backend/sync.rs
```diff
@@ -154,6 +155,15 @@ impl crate::services::BackendSyncService for Backend {
         )?;
         Ok(())
     }
+
+    fn set_custom_certificate(&self, _input: generic::String) -> Result<generic::Bool> {
+        #[cfg(feature = "rustls")]
+        return Ok(generic::Bool::from(
+            self.set_custom_cert(_input.val).is_ok(),
+        ));
+        #[cfg(not(feature = "rustls"))]
```
voczi, line 164: Could there be a cleaner way of doing this? Would be nice if I could exclude the whole function but that won't work since it's declared in sync.proto.

dae:
This PR still needs work:

- You are checking the return value of web_client() in multiple places, instead of just bubbling up the error
- Interrupted is not the correct error. That error should only be used when a long-running operation is cancelled by the user.
- The return value of .lock() should be unwrapped, not checked. It will only be an error if the code panics.

rslib/src/backend/sync.rs
dae: The only alternative I can think of is defining a separate version of the function. I don't have a preference either way.

rslib/src/backend/mod.rs
```diff
@@ -137,10 +139,39 @@ impl Backend {
...
+    fn web_client(&self) -> Result<Client> {
+        let mut client_mutex = self.web_client.lock();
+
+        if let Ok(ref mut web_client) = client_mutex {
+            return Ok((**web_client)
+                .get_or_insert(Client::builder().http1_only().build().unwrap())
+                .clone());
+        }
+
+        Err(AnkiError::Interrupted)
```
dae, line 174: Does this even need a fallible return value? The only thing being checked is the return value of lock(), which should be unwrapped. If you make this just return the client, then you shouldn't even need to bubble the error up.

Also, you're creating a client even if there is an existing one, and then discarding the value. The build() call should only be made if the option is empty.

PR #3098, opened 2024-03-28
highlight io shapes in answer side

krmanik:
Highlight io shapes in answer side, in anki manual it needs to mention that color should include alpha values for transparent.
https://forums.ankiweb.net/t/image-occlusion-back-template-to-have-translucent-boxes

</img>


dae:
Thanks Mani, I'll return to this after 24.04 is out the door.

Issue #3096, opened 2024-03-26
The update interval of timer display


dae:
> Sometimes I use the timer (deck option - timer - show answer timer) to monitor how much time I spend in answering a card. However, from time to time I feel that the timer is not very consistent, i.e. it may display different numbers for the same amount of time spent. This is especially obvious when enabling auto advance and “stop timer on answer”: e.g. if I set the “time to reveal answer” as 3.4 seconds, when the answer is revealed the timer may stop at either 00:03 or 00:04.

Originally reported on https://forums.ankiweb.net/t/the-update-interval-of-timer-display/42849

Two issues by the looks of it:

- in reviewer-bottom we use a 1 second timer that's not aligned to when a card is presented
- the start time is rounded to integer seconds in reviewer.py:780

We don't want to increase the update frequency too much, as it will waste CPU cycles, but maybe 0.5 or 0.3 would be enough?

AntoineQ1:
Hello !
I'd love to contribute to Anki ; could I suggest the necessary code changes in a PR ?
Sincerely,

dae:
Yes, a PR would be welcome - thank you for the offer.

PR #3100, opened 2024-03-28
Improve timer interval display (#3096)

AntoineQ1:
**To improve the displayed timer accuracy, I carried out the following changes :**

1. The frequency of the timer has been updated from 1.0 second to 0.3 second
2. The timer is reset with each new card

_Resolves #3096_

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -168,6 +168,7 @@ Arbyste <arbyste@outlook.com>
 Vasll <github.com/vasll>
 laalsaas <laalsaas@systemli.org>
 ijqq <ijqq@protonmail.ch>
+AntoineQ1 <https://github.com/AntoineQ1>
 ********************
 
 The text of the 3 clause BSD license follows:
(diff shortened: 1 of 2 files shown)


dae:
Thank you!

Minor thought: how about 333 as the delay, so that the timer ends up getting updated as close to a second as possible?

AntoineQ1:
You're welcome :)

Great idea, I just updated the code accordingly !

dae:
Sorry, I feel silly for not thinking of this earlier, but do we even need to increase the timer resolution? Since you're starting the timer as the question is shown now, perhaps we could simply refresh every 1000ms after that?

AntoineQ1:
No problem ! Indeed, after some tests, it seems it's not necessary.
Also, it helped me discover that the timer didn't reset to 0 immediatly when a new card was displayed (I didn't see it before because it was only 0.3 / 0.333 seconds later), so I added a time update just after the new question is shown to account for that.

dae:
Great - thanks again!

PR #3101, opened 2024-03-29
rename sm2 retention to historical retention

L-M-Sherlock:
fix https://forums.ankiweb.net/t/anki-24-04-beta-rc/41792/124?u=l.m.sherlock

@dae, @Expertium, what do you think of the new naming and tooltip?

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -378,10 +378,18 @@ deck-config-desired-retention-tooltip =
     values will greatly increase your workload, and lower values can be demoralizing when you forget
     a lot of material.
 deck-config-historical-retention-tooltip =
-    If your actual retention before switching to FSRS or setting the ignore reviews date was significantly
-    different from 0.9, adjusting this value will allow Anki to better estimate your memory state when it
-    encounters cards that are missing or ignoring review logs. Since review logs typically won't be missing
-    unless you explicitly deleted them to free up space, most users will not need to adjust this.
+    When some of your review history is missing, FSRS needs to fill in the gaps. By default, it will
+    assume that when you did those old reviews, you remembered 90% of the material. If your old retention
+    was appreciably higher or lower than 90%, adjusting this option will allow FSRS to better approximate
+    the missing reviews.
+
+    Your review history may be incomplete for two reasons:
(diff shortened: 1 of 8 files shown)


rslib/src/deckconfig/schema11.rs
```diff
@@ -86,7 +86,7 @@ pub struct DeckConfSchema11 {
     #[serde(default = "wait_for_audio_default")]
     wait_for_audio: bool,
     #[serde(default)]
-    sm2_retention: f32,
+    historical_retention: f32,
```
dae, line 89: You might want to keep it as sm2_retention in this file, and just add `/// doccomment` that it is historical_retention. Otherwise, users who have downgraded their proflie, and then upgrade with a new Anki version, will lose their existing settings.

ftl/core/deck-config.ftl
```diff
@@ -377,11 +377,11 @@ deck-config-desired-retention-tooltip =
     less frequently, and you will forget more of them. Be conservative when adjusting this - higher
     values will greatly increase your workload, and lower values can be demoralizing when you forget
     a lot of material.
-deck-config-sm2-retention-tooltip =
-    If your actual retention before switching to FSRS was significantly different from 0.9, adjusting
-    this value will allow Anki to better estimate your memory state when it encounters cards that are
-    missing review logs. Since review logs typically won't be missing unless you explicitly deleted them
-    to free up space, most users will not need to adjust this.
+deck-config-historical-retention-tooltip =
```
dae, line 380: Suggested wording:

When some of your review history is missing, FSRS needs to fill in the gaps. By default, it will assume that when you did those old reviews, you remembered 90% of the material. If your old retention was appreciably higher or lower than 90%, adjusting this option will allow FSRS to better approximate the missing reviews.

Your review history may be incomplete for two reasons:

1) Because you've used the 'ignore reviews before' option.
2) Because you previously deleted review logs to free up space, or imported material from a different SRS program.

The latter is quite rare, so unless you've used the former option, you probably don't need to adjust this setting.

L-M-Sherlock: Done.

dae:
Thanks Jarrett!

PR #3104, opened 2024-03-29
Use "importing-added" string in importing detail table

snowtimeglass:
Supplement to the task of the commit 184ad80 (Clone "added" string for importing)

Currently, for the "Added" as an importing status in the table in the "Import File" window, `adding-added` string seems to be used, as before.

This commit is intended to fix it, that is, to use `importing-added` string for the status text.


dae:
Thanks!

PR #3105, opened 2024-03-29
Fix mpv loadfile syntax change

kelciour:
- https://forums.ankiweb.net/t/audio-error-on-arch-linux/42975/1
- https://github.com/mpv-player/mpv/pull/13406


dae:
Thanks for this quick fix, Kelciour!

PR #3106, opened 2024-03-31
Update to FSRS-rs v0.6.1

### LINKED ISSUE #3094: FSRS - Research suggests the minimum limit can be reduced to 16
Hey everyone,

We currently have a minimum of 1000 reviews for FSRS to be optimised, which I believe is now changing to 400. After some research and discussions with @L-M-Sherlock it seems we can lower this as far as 16, or remove it completely.

I've forked and updated srs-benchmark which contains the benchmarking code for this:
https://github.com/RlzHi/srs-benchmark/blob/main/notebook/metric_over_size.ipynb
download

The dry run is FSRS 4.5 without optimisations, and after approximately 16 reviews optimising FSRS performs better than the default parameters.

L-M-Sherlock:
related update:
- https://github.com/open-spaced-repetition/fsrs-rs/pull/172
- https://github.com/open-spaced-repetition/fsrs-rs/pull/173

close #3094

@user1823, @Expertium, @RlzHi, please review this PR.


user1823 (approved):

L-M-Sherlock:
These tooltips should also update. @Expertium, @user1823, any suggestion?

https://github.com/ankitects/anki/blob/3cd8c5700b90f3cf9afbbee078df1bb9346f86a0/ftl/core/deck-config.ftl#L393-L396

https://github.com/ankitects/anki/blob/3cd8c5700b90f3cf9afbbee078df1bb9346f86a0/ftl/core/deck-config.ftl#L413-L418

Expertium:
I suggest just replacing 1000 with 8. Also, the same needs to be done with the "insufficient review history" error message.

L-M-Sherlock:
> I suggest just replacing 1000 with 8. Also, the same needs to be done with the "insufficient review history" error message.

The user still could click `optimize` to get the default parameters when the number of reviews is lower than 8. It's more important to explain which reviews are counted. Assuming a new user does 1000 reviews in a deck of new cards, the `optimize` button still pop up the `insufficient data` error because those reviews are the first review of each cards. The first four parameters are pretrained in the second review of each card, and the rest params are trained in the third and following reviews of each card.

dae:
Maybe one option would be to stop showing an error, and just fall back on the default parameters instead?

dae:
If we don't change anything, we can keep showing the message that the parameters are currently optimal.

Expertium:
@Dae, do you plan to release 24.04.1 after you merge this pull request?

dae:
.1 will be a quick update with only some bugfixes, so I wasn't planning to include this.  There may be a .2 if we have a longer testing period to try it out.

dae:
Thanks @L-M-Sherlock. This will go into the main branch, but probably not into the 24.04.1 branch.

PR #3108, opened 2024-04-01
Corrected minor non-idiomatic snippet of code

jthulhu:
Creating an array with a single iterator just for the purpose of iterating over it can be achieved more clearly by using the standard library function `once` which serves exactly the same purpose.


dae:
Thank you!

PR #3109, opened 2024-04-01
fix select all and change ordinal in edit mode in io

krmanik:
The edit mode needs careful review, I have come up with the solution, but it may need improvement.
@abdnh pinging for review.

Closes https://github.com/ankitects/anki/pull/3109

### DIFF
--- a/ts/routes/image-occlusion/shapes/to-cloze.ts
+++ b/ts/routes/image-occlusion/shapes/to-cloze.ts
@@ -3,8 +3,8 @@
 
 import { fabric } from "fabric";
 import { cloneDeep } from "lodash-es";
-
 import { get } from "svelte/store";
+
 import { groupItemsModified } from "../store";
 import { getBoundingBox } from "../tools/lib";
 import type { Size } from "../types";
--- a/ts/routes/image-occlusion/shapes/to-cloze.ts
+++ b/ts/routes/image-occlusion/shapes/to-cloze.ts
@@ -42,16 +42,9 @@ export function exportShapesToClozeDeletions(occludeInactive: boolean): {
 
     let nextOrdinal = maxOrdinal + 1;
 
-    for (let i = 0; i < shapes.length; i++) {
-        const shapeOrShapes = shapes[i];
-
-        // shapes with width or height less than 5 are not valid
+    shapes.map((shapeOrShapes) => {
         if (shapeOrShapes === null) {
-            continue;
-        }
-        // if shape is Rect and fill is transparent, skip it
-        if (shapeOrShapes instanceof Rectangle && shapeOrShapes.fill === "transparent") {
-            continue;
+            return;
         }
 
         // Maintain existing ordinal in editing mode
@@ -89,7 +82,7 @@ export function exportShapesToClozeDeletions(occludeInactive: boolean): {
         if (!(shapeOrShapes instanceof Text)) {
             noteCount++;
         }
-    }
+    });
     return { clozes, noteCount };
 }
 
@@ -113,6 +106,8 @@ export function baseShapesFromFabric(): ShapeOrShapes[] {
             const parent = selectionContainingMultipleObjects?.contains(object)
                 ? selectionContainingMultipleObjects
                 : undefined;
+            // shapes with width or height less than 5 are not valid
+            // if shape is Rect and fill is transparent, skip it
             if (object.width! < 5 || object.height! < 5 || object.fill == "transparent") {
                 return null;
             }
--- a/ts/routes/image-occlusion/shapes/to-cloze.ts
+++ b/ts/routes/image-occlusion/shapes/to-cloze.ts
@@ -3,7 +3,9 @@
 
 import { fabric } from "fabric";
 import { cloneDeep } from "lodash-es";
+import { get } from "svelte/store";
 
+import { shapeUngroupState } from "../store";
 import { getBoundingBox } from "../tools/lib";
 import type { Size } from "../types";
 import type { Shape, ShapeOrShapes } from "./base";
@@ -166,7 +168,7 @@ function shapeOrShapesToCloze(
 
     // Maintain existing ordinal in editing mode
     let ordinal = shapeOrShapes.ordinal;
-    if (ordinal === undefined || Number.isNaN(ordinal)) {
+    if (ordinal === undefined || Number.isNaN(ordinal) || get(shapeUngroupState)) {
         if (type === "text") {
             ordinal = 0;
         } else {
--- a/ts/routes/image-occlusion/shapes/to-cloze.ts
+++ b/ts/routes/image-occlusion/shapes/to-cloze.ts
@@ -4,6 +4,8 @@
 import { fabric } from "fabric";
 import { cloneDeep } from "lodash-es";
 
+import { get } from "svelte/store";
+import { groupItemsModified } from "../store";
 import { getBoundingBox } from "../tools/lib";
 import type { Size } from "../types";
 import type { Shape, ShapeOrShapes } from "./base";
@@ -123,11 +125,6 @@ function fabricObjectToBaseShapeOrShapes(
         size = { width: 0, height: 0 };
     }
 
-    // @ts-expect-error not defined
-    if (cloned.modified) {
-        shape.modified = true;
-    }
-
     shape = shape.toNormal(size);
     return shape;
 }
@@ -171,13 +168,7 @@ function shapeOrShapesToCloze(
 
     // Maintain existing ordinal in editing mode
     let ordinal = shapeOrShapes.ordinal;
-
-    if (shapeOrShapes.modified) {
-        ordinal = index + 1;
-        shapeOrShapes.ordinal = ordinal;
-    }
-
-    if (ordinal === undefined) {
+    if (ordinal === undefined || get(groupItemsModified)) {
         if (type === "text") {
             ordinal = 0;
         } else {
--- a/ts/routes/image-occlusion/shapes/to-cloze.ts
+++ b/ts/routes/image-occlusion/shapes/to-cloze.ts
@@ -19,17 +19,41 @@ export function exportShapesToClozeDeletions(occludeInactive: boolean): {
     const shapes = baseShapesFromFabric();
 
     let clozes = "";
-    let index = 0;
-    let nextOrdinal = 1;
-    shapes.forEach((shapeOrShapes) => {
+    let noteCount = 0;
+
+    // take out all ordinal values from shapes
+    const ordinalList = shapes.map((shape) => {
+        if (Array.isArray(shape)) {
+            return shape[0].ordinal;
+        } else {
+            return shape.ordinal;
+        }
+    });
+
+    const filterOrdinalList: number[] = ordinalList.flatMap(v => typeof v === "number" ? [v] : []);
+    const maxOrdinal = Math.max(...filterOrdinalList, 0);
+
+    const missingOrdinals: number[] = [];
+    for (let i = 1; i <= maxOrdinal; i++) {
+        if (!ordinalList.includes(i)) {
+            missingOrdinals.push(i);
+        }
+    }
+
+    let nextOrdinal = maxOrdinal + 1;
+
+    for (let i = 0; i < shapes.length; i++) {
+        const shapeOrShapes = shapes[i];
+
         // shapes with width or height less than 5 are not valid
         if (shapeOrShapes === null) {
-            return;
+            continue;
         }
         // if shape is Rect and fill is transparent, skip it
         if (shapeOrShapes instanceof Rectangle && shapeOrShapes.fill === "transparent") {
-            return;
+            continue;
         }
+
         // Maintain existing ordinal in editing mode
         let ordinal: number | undefined;
         if (Array.isArray(shapeOrShapes)) {
@@ -37,32 +61,36 @@ export function exportShapesToClozeDeletions(occludeInactive: boolean): {
         } else {
             ordinal = shapeOrShapes.ordinal;
         }
-        if (ordinal === undefined || Number.isNaN(ordinal)) {
+
+        if (ordinal === undefined) {
+            // if ordinal is undefined, assign a missing ordinal if available
             if (shapeOrShapes instanceof Text) {
                 ordinal = 0;
+            } else if (missingOrdinals.length > 0) {
+                ordinal = missingOrdinals.shift() as number;
             } else {
                 ordinal = nextOrdinal;
+                nextOrdinal++;
             }
+
             if (Array.isArray(shapeOrShapes)) {
                 shapeOrShapes.forEach((shape) => (shape.ordinal = ordinal));
             } else {
                 shapeOrShapes.ordinal = ordinal;
             }
         }
-        nextOrdinal = ordinal + 1;
 
         clozes += shapeOrShapesToCloze(
             shapeOrShapes,
-            index,
             ordinal,
             occludeInactive,
         );
-        if (!(shapeOrShapes instanceof Text)) {
-            index++;
-        }
-    });
 
-    return { clozes, noteCount: index };
+        if (!(shapeOrShapes instanceof Text)) {
+            noteCount++;
+        }
+    }
+    return { clozes, noteCount };
 }
 
 /** Gather all Fabric shapes, and convert them into BaseShapes or
@@ -77,6 +105,7 @@ export function baseShapesFromFabric(): ShapeOrShapes[] {
         : null;
     const objects = canvas.getObjects() as fabric.Object[];
     const boundingBox = getBoundingBox();
+    // filter transparent rectangles
     return objects
         .map((object) => {
             // If the object is in the active selection containing multiple objects,
@@ -84,7 +113,7 @@ export function baseShapesFromFabric(): ShapeOrShapes[] {
             const parent = selectionContainingMultipleObjects?.contains(object)
                 ? selectionContainingMultipleObjects
                 : undefined;
-            if (object.width! < 5 || object.height! < 5) {
+            if (object.width! < 5 || object.height! < 5 || object.fill == "transparent") {
                 return null;
             }
             return fabricObjectToBaseShapeOrShapes(
@@ -158,7 +187,6 @@ function fabricObjectToBaseShapeOrShapes(
  {{c1::image-occlusion:rect:top=.1:left=.23:width=.4:height=.5}} */
 function shapeOrShapesToCloze(
     shapeOrShapes: ShapeOrShapes,
-    index: number,
     ordinal: number,
     occludeInactive: boolean,
 ): string {
@@ -171,7 +199,7 @@ function shapeOrShapesToCloze(
     let type: string;
     if (Array.isArray(shapeOrShapes)) {
         return shapeOrShapes
-            .map((shape) => shapeOrShapesToCloze(shape, index, ordinal, occludeInactive))
+            .map((shape) => shapeOrShapesToCloze(shape, ordinal, occludeInactive))
             .join("");
     } else if (shapeOrShapes instanceof Rectangle) {
         type = "rect";
--- a/ts/routes/image-occlusion/shapes/to-cloze.ts
+++ b/ts/routes/image-occlusion/shapes/to-cloze.ts
@@ -3,9 +3,7 @@
 
 import { fabric } from "fabric";
 import { cloneDeep } from "lodash-es";
-import { get } from "svelte/store";
 
-import { groupItemsModified } from "../store";
 import { getBoundingBox } from "../tools/lib";
 import type { Size } from "../types";
 import type { Shape, ShapeOrShapes } from "./base";
@@ -168,7 +166,7 @@ function shapeOrShapesToCloze(
 
     // Maintain existing ordinal in editing mode
     let ordinal = shapeOrShapes.ordinal;
-    if (ordinal === undefined || get(groupItemsModified)) {
+    if (ordinal === undefined || Number.isNaN(ordinal)) {
         if (type === "text") {
             ordinal = 0;
         } else {
--- a/ts/routes/image-occlusion/shapes/to-cloze.ts
+++ b/ts/routes/image-occlusion/shapes/to-cloze.ts
@@ -123,6 +123,11 @@ function fabricObjectToBaseShapeOrShapes(
         size = { width: 0, height: 0 };
     }
 
+    // @ts-expect-error not defined
+    if (cloned.modified) {
+        shape.modified = true;
+    }
+
     shape = shape.toNormal(size);
     return shape;
 }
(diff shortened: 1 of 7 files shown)


dae:
I gave this a quick test, and it seems to fix the problem for me in both edit mode and the add screen.

krmanik:
Initial test works, but after adding some notes with group mode then again in edit mode, ungroup last shapes then group some other shapes. The ordinal seems to change and causing error something like No cloze found on number.

krmanik:
The solution after changing group item can be fixed by introducing a new method which checks for ordinal number in cloze strings and reorder, the issue also fixed by `Tools -> Empty Cards`. The complicated solution may be reduced by just using just `ordinal = 0`, but some users may want to have same ordinal in edit mode for other shapes.

```
    let ordinal = 0;   // <-----
    if (ordinal === undefined) {
        if (type === "text") {
            ordinal = 0;
        } else {
            ordinal = index + 1;
        }
        shapeOrShapes.ordinal = ordinal;
    }
```

dae:
Sorry Mani, could I trouble you to explain the issue in more detail? If a user reduces the number of occlusions, I'd expect that to break one or more of their cards. It would be nice if we could display a useful message during review in that case, like we do when there's an empty cloze. But I don't think we need to be cleaning up the cards until the user uses Tools>Empty Cards.

krmanik:
The ordinal is undefined for all shapes when group/ungroup tools used, so that the cloze number should be in order.

dae:
Hmm, this doesn't seem to work - if I add two rectangles to a new card, then edit the card and select all+group, the two shapes stay disconnected when the editor is closed.

krmanik:
Now it will work as expected, but the users still have to use `Tools -> Empty Cards`.

krmanik:
> I've pushed a fix for the dprint error. Tests aren't currently passing, but seems to work well when I tried the grouping operation again with the latest code.

Thanks, I have rebased it to main.

ts/routes/image-occlusion/shapes/to-cloze.ts
```diff
@@ -166,7 +168,7 @@ function shapeOrShapesToCloze(
 
     // Maintain existing ordinal in editing mode
     let ordinal = shapeOrShapes.ordinal;
-    if (ordinal === undefined) {
+    if (ordinal === undefined || get(groupItemsModified)) {
```
abdnh, line 171: This can break the ordinals of ungrouped masks when a grouping operation is done on other masks in edit mode.

Can we instead update the ordinals of grouped objects when they are grouped?
For example, If we group objects with the ordinals 2 and 4, all objects in the group will be assigned a single ordinal (maybe the smallest one, so 2).

krmanik: I don't think so, because let's say a user have created group cloze with more than two shapes as single group cloze, then during review he edited the shape to ungroup to individual shapes. Now total cards increased from one to total number of shapes. Now again during review he split it into two group, now ordinal must start from beginning to have consistent result, (Card 1 and Card 2)

dae: I don't think we can solve this for all situations, especially if the user is closing and opening the editor between changes. I'd assumed that if we give new ordinals to shapes when they're removed from a group, that would be sufficient?

krmanik: The new ordinal to removed shapes may not also solve this, I have manually assigned cloze number to removed shape by editing clozes in Occlusion field and result to that Card was something like this. It makes all shapes as occlusion for front as well as back.

The other solution should be when shapes are merged, we can merge cards also, but we cannot know which card associated with which shape. So, I can say that we should warning message when users performing group/ungroup of shapes.

image

dae: We could show a warning when the group/ungroup tools are used on an existing note in the future, but I'm not too concerned about it at the moment as long as it's not completely breaking.

dae: Just to check we're on the same page - would you like this PR merged in without the changes that Abdo suggested?

krmanik: I have thought of implementation idea for Abdo suggestion, I have implemented it in next commit, and it seems, it also solves the empty card issues. But it still needs to some review.

ts/routes/image-occlusion/tools/lib.ts
```diff
@@ -88,12 +93,18 @@ export const unGroupShapes = (canvas: fabric.Canvas): void => {
     group.destroyed = true;
     canvas.remove(group);
 
+    // @ts-expect-error not defined
+    const maxOrdinal = Math.max(...items.map((item) => item.ordinal));
```
krmanik, line 97: Max should be from all shapes not from group, need changes.

ts/routes/image-occlusion/shapes/to-cloze.ts
krmanik: Users need to do `Tools -> Empty Cards`, I have tried to solve but once cards added, the ordinal decreases but Cards remains same, so it needs to be fixed separately.

ts/routes/image-occlusion/tools/lib.ts
```diff
@@ -63,12 +63,19 @@ export const groupShapes = (canvas: fabric.Canvas): void => {
 
     const activeObject = canvas.getActiveObject() as fabric.ActiveSelection;
     const items = activeObject.getObjects();
+
+    // @ts-expect-error not defined
+    const minOrdinal = Math.min(...items.map((item) => item.ordinal));
```
dae, line 68: Something that might be nice in the future (not in this PR) is to tell TypeScript about the additional properties we've added to these objects, so we don't need to ignore errors/cast them to any each time we access them. I think it may be possible in a .d.ts file?

ts/routes/image-occlusion/shapes/to-cloze.ts
dae: Sorry, I'm not quite following - what do you think needs to be done differently?

dae:
Thanks Mani, this seems to work well. @abdnh, any thoughts/comments? If not, I'll merge this in tomorrow and get a 24.04.1rc out.

ts/routes/image-occlusion/shapes/to-cloze.ts
krmanik: Nothing, I was implementing different solution when ungroup shapes. In case of ungroup shapes, find all the ordinal from shapes excluding groups shapes then start loop from beginning and put the ordinal to shapes in order. But I implemented simplest solution, use max ordinal (increment if required).

ts/routes/image-occlusion/tools/lib.ts
krmanik: This can be done, I will look into this.

ts/routes/image-occlusion/tools/lib.ts
```diff
@@ -86,8 +93,12 @@ export const unGroupShapes = (canvas: fabric.Canvas): void => {
...
     canvas.remove(group);
 
-    items.forEach((item) => {
-        item.set({ opacity: get(opacityStateStore) ? 0.4 : 1 });
+    // @ts-expect-error not defined
+    const maxOrdinal = Math.max(canvas.getObjects().map((item) => item.ordinal));
+
+    items.forEach((item, index) => {
+        // @ts-expect-error not defined
+        item.set({ opacity: get(opacityStateStore) ? 0.4 : 1, ordinal: maxOrdinal + index + 1 });
```
abdnh, line 101: This is not working as expected. It can result in two unrelated masks getting the same ordinal, causing them to be grouped. To reproduce:
1. Add 3 masks
2. Group two of the masks
3. Ungroup
4. Reload the note (e.g. by selecting another browser note then back to the note)
5. Notice one mask from the original group is grouped with the third.

krmanik:
I was trying to implement this, but it works sometimes and sometimes does not, but the simplest solution will be just use svelte store only for ungroup shapes which works now.

```diff
diff --git a/ts/routes/image-occlusion/tools/lib.ts b/ts/routes/image-occlusion/tools/lib.ts
index f7d00846c..38fccd703 100644
--- a/ts/routes/image-occlusion/tools/lib.ts
+++ b/ts/routes/image-occlusion/tools/lib.ts
@@ -93,18 +93,53 @@ export const unGroupShapes = (canvas: fabric.Canvas): void => {
     group.destroyed = true;
     canvas.remove(group);
 
-    // @ts-expect-error not defined
-    const maxOrdinal = Math.max(canvas.getObjects().map((item) => item.ordinal));
+    // take ordinals for all shape excluding the ones in the group
+    const excludedObjectOrdinals = getCanvasObjects(canvas)
+        .filter((obj) => !items.includes(obj))
+        .map((item) => item.ordinal)
+        .filter((ordinal) => ordinal !== undefined || !Number.isNaN(ordinal));
+
+    console.log("excludedObjectOrdinals", excludedObjectOrdinals);
+
+    const totalShapes = getCanvasObjects(canvas).length;
+
+    console.log("totalShapes", totalShapes);
+
+    const newOrdinal = [];
+    for (let i = 1; i < totalShapes; i++) {
+        if (!excludedObjectOrdinals.includes(i)) {
+            newOrdinal.push(i);
+        }
+    }
+
+    console.log("newOrdinal", newOrdinal);
 
     items.forEach((item, index) => {
         // @ts-expect-error not defined
-        item.set({ opacity: get(opacityStateStore) ? 0.4 : 1, ordinal: maxOrdinal + index + 1 });
+        item.set({ opacity: get(opacityStateStore) ? 0.4 : 1, ordinal: newOrdinal[index] });
         canvas.add(item);
     });
 
     redraw(canvas);
 };
 
+const getCanvasObjects = (canvas) => {
+    const objectList = [];
+    canvas.forEachObject(function(object) {
+        if (object.type === 'group') {
+            const group = object;
+            group.forEachObject(function(obj) {
+                objectList.push(obj);
+            });
+        } else {
+            objectList.push(object);
+        }
+    });
+
+    // filter out the transparent bounding box from the selection
+    return objectList.filter((obj) => obj.fill !== "transparent");
+}
+
 const copyItem = (canvas: fabric.Canvas): void => {
     const activeObject = canvas.getActiveObject();
     if (!activeObject) {

```

ts/routes/image-occlusion/tools/lib.ts
krmanik: I have updated the implementation, it should work now, but order cannot be maintained in case of ungrouping of shapes.

dae:
@abdnh any other thoughts, or happy for this to be merged in?

ts/routes/image-occlusion/tools/lib.ts
krmanik: No problem, we can go with this.

abdnh: @dae While the issue I reported is fixed, this is re-introducing the original problem fixed by #2987 when there's an ungroup operation. Unrelated masks could get reordered.

This appears to work better:
```diff
diff --git a/ts/routes/image-occlusion/shapes/to-cloze.ts b/ts/routes/image-occlusion/shapes/to-cloze.ts
index 01d8b62c9..fbdd03f07 100644
--- a/ts/routes/image-occlusion/shapes/to-cloze.ts
+++ b/ts/routes/image-occlusion/shapes/to-cloze.ts
@@ -3,9 +3,7 @@

 import { fabric } from "fabric";
 import { cloneDeep } from "lodash-es";
-import { get } from "svelte/store";

-import { shapeUngroupState } from "../store";
 import { getBoundingBox } from "../tools/lib";
 import type { Size } from "../types";
 import type { Shape, ShapeOrShapes } from "./base";
@@ -22,6 +20,7 @@ export function exportShapesToClozeDeletions(occludeInactive: boolean): {

     let clozes = "";
     let index = 0;
+    let nextOrdinal = 1;
     shapes.forEach((shapeOrShapes) => {
         // shapes with width or height less than 5 are not valid
         if (shapeOrShapes === null) {
@@ -31,7 +30,33 @@ export function exportShapesToClozeDeletions(occludeInactive: boolean): {
         if (shapeOrShapes instanceof Rectangle && shapeOrShapes.fill === "transparent") {
             return;
         }
-        clozes += shapeOrShapesToCloze(shapeOrShapes, index, occludeInactive);
+        // Maintain existing ordinal in editing mode
+        let ordinal: number | undefined;
+        if (Array.isArray(shapeOrShapes)) {
+            ordinal = shapeOrShapes[0].ordinal;
+        } else {
+            ordinal = shapeOrShapes.ordinal;
+        }
+        if (ordinal === undefined) {
+            if (shapeOrShapes instanceof Text) {
+                ordinal = 0;
+            } else {
+                ordinal = nextOrdinal;
+            }
+            if (Array.isArray(shapeOrShapes)) {
+                shapeOrShapes.forEach((shape) => (shape.ordinal = ordinal));
+            } else {
+                shapeOrShapes.ordinal = ordinal;
+            }
+        }
+        nextOrdinal = ordinal + 1;
+
+        clozes += shapeOrShapesToCloze(
+            shapeOrShapes,
+            index,
+            ordinal,
+            occludeInactive,
+        );
         if (!(shapeOrShapes instanceof Text)) {
             index++;
         }
@@ -134,6 +159,7 @@ function fabricObjectToBaseShapeOrShapes(
 function shapeOrShapesToCloze(
     shapeOrShapes: ShapeOrShapes,
     index: number,
+    ordinal: number,
     occludeInactive: boolean,
 ): string {
     let text = "";
@@ -145,7 +171,7 @@ function shapeOrShapesToCloze(
     let type: string;
     if (Array.isArray(shapeOrShapes)) {
         return shapeOrShapes
-            .map((shape) => shapeOrShapesToCloze(shape, index, occludeInactive))
+            .map((shape) => shapeOrShapesToCloze(shape, index, ordinal, occludeInactive))
             .join("");
     } else if (shapeOrShapes instanceof Rectangle) {
         type = "rect";
@@ -166,16 +192,6 @@ function shapeOrShapesToCloze(
         addKeyValue("oi", "1");
     }

-    // Maintain existing ordinal in editing mode
-    let ordinal = shapeOrShapes.ordinal;
-    if (ordinal === undefined || Number.isNaN(ordinal) || get(shapeUngroupState)) {
-        if (type === "text") {
-            ordinal = 0;
-        } else {
-            ordinal = index + 1;
-        }
-        shapeOrShapes.ordinal = ordinal;
-    }
     text = `{{c${ordinal}::image-occlusion:${type}${text}}}<br>`;

     return text;
diff --git a/ts/routes/image-occlusion/store.ts b/ts/routes/image-occlusion/store.ts
index b1653f4fa..5847250dd 100644
--- a/ts/routes/image-occlusion/store.ts
+++ b/ts/routes/image-occlusion/store.ts
@@ -17,5 +17,3 @@ export const ioImageLoadedStore = writable(false);
 export const opacityStateStore = writable(false);
 // store state of text editing
 export const textEditingState = writable(false);
-// store state of shape ungroup
-export const shapeUngroupState = writable(false);
diff --git a/ts/routes/image-occlusion/tools/lib.ts b/ts/routes/image-occlusion/tools/lib.ts
index 9a6738b10..bc0606cdc 100644
--- a/ts/routes/image-occlusion/tools/lib.ts
+++ b/ts/routes/image-occlusion/tools/lib.ts
@@ -4,7 +4,7 @@
 import { fabric } from "fabric";
 import { get } from "svelte/store";

-import { opacityStateStore, shapeUngroupState } from "../store";
+import { opacityStateStore } from "../store";

 export const SHAPE_MASK_COLOR = "#ffeba2";
 export const BORDER_COLOR = "#212121";
@@ -94,11 +94,14 @@ export const unGroupShapes = (canvas: fabric.Canvas): void => {
     canvas.remove(group);

     items.forEach((item) => {
-        item.set({ opacity: get(opacityStateStore) ? 0.4 : 1 });
+        item.set({
+            opacity: get(opacityStateStore) ? 0.4 : 1,
+            // @ts-expect-error not defined
+            ordinal: undefined,
+        });
         canvas.add(item);
     });

-    shapeUngroupState.set(true);
     redraw(canvas);
 };

```

What do you think @dae @krmanik ?

krmanik: This does not seem to work, create more 4 shapes, group shape, so group 1 has two shapes and group 2 has two shapes, click on add button, the error is (You have a cloze notetype but have not made any cloze deletions. Proceed?)
I think we have spent enough time solving this, we should continue without thinking about order as it is not breaking changes.

This updated patch based on above on latest main commit,

<details>
<summary>Code</summary>

```diff
diff --git a/ts/routes/image-occlusion/shapes/to-cloze.ts b/ts/routes/image-occlusion/shapes/to-cloze.ts
index 01d8b62c9..fbdd03f07 100644
--- a/ts/routes/image-occlusion/shapes/to-cloze.ts
+++ b/ts/routes/image-occlusion/shapes/to-cloze.ts
@@ -3,9 +3,7 @@
 
 import { fabric } from "fabric";
 import { cloneDeep } from "lodash-es";
-import { get } from "svelte/store";
 
-import { shapeUngroupState } from "../store";
 import { getBoundingBox } from "../tools/lib";
 import type { Size } from "../types";
 import type { Shape, ShapeOrShapes } from "./base";
@@ -22,6 +20,7 @@ export function exportShapesToClozeDeletions(occludeInactive: boolean): {
 
     let clozes = "";
     let index = 0;
+    let nextOrdinal = 1;
     shapes.forEach((shapeOrShapes) => {
         // shapes with width or height less than 5 are not valid
         if (shapeOrShapes === null) {
@@ -31,7 +30,33 @@ export function exportShapesToClozeDeletions(occludeInactive: boolean): {
         if (shapeOrShapes instanceof Rectangle && shapeOrShapes.fill === "transparent") {
             return;
         }
-        clozes += shapeOrShapesToCloze(shapeOrShapes, index, occludeInactive);
+        // Maintain existing ordinal in editing mode
+        let ordinal: number | undefined;
+        if (Array.isArray(shapeOrShapes)) {
+            ordinal = shapeOrShapes[0].ordinal;
+        } else {
+            ordinal = shapeOrShapes.ordinal;
+        }
+        if (ordinal === undefined) {
+            if (shapeOrShapes instanceof Text) {
+                ordinal = 0;
+            } else {
+                ordinal = nextOrdinal;
+            }
+            if (Array.isArray(shapeOrShapes)) {
+                shapeOrShapes.forEach((shape) => (shape.ordinal = ordinal));
+            } else {
+                shapeOrShapes.ordinal = ordinal;
+            }
+        }
+        nextOrdinal = ordinal + 1;
+
+        clozes += shapeOrShapesToCloze(
+            shapeOrShapes,
+            index,
+            ordinal,
+            occludeInactive,
+        );
         if (!(shapeOrShapes instanceof Text)) {
             index++;
         }
@@ -134,6 +159,7 @@ function fabricObjectToBaseShapeOrShapes(
 function shapeOrShapesToCloze(
     shapeOrShapes: ShapeOrShapes,
     index: number,
+    ordinal: number,
     occludeInactive: boolean,
 ): string {
     let text = "";
@@ -145,7 +171,7 @@ function shapeOrShapesToCloze(
     let type: string;
     if (Array.isArray(shapeOrShapes)) {
         return shapeOrShapes
-            .map((shape) => shapeOrShapesToCloze(shape, index, occludeInactive))
+            .map((shape) => shapeOrShapesToCloze(shape, index, ordinal, occludeInactive))
             .join("");
     } else if (shapeOrShapes instanceof Rectangle) {
         type = "rect";
@@ -166,16 +192,6 @@ function shapeOrShapesToCloze(
         addKeyValue("oi", "1");
     }
 
-    // Maintain existing ordinal in editing mode
-    let ordinal = shapeOrShapes.ordinal;
-    if (ordinal === undefined || Number.isNaN(ordinal) || get(shapeUngroupState)) {
-        if (type === "text") {
-            ordinal = 0;
-        } else {
-            ordinal = index + 1;
-        }
-        shapeOrShapes.ordinal = ordinal;
-    }
     text = `{{c${ordinal}::image-occlusion:${type}${text}}}<br>`;
 
     return text;
diff --git a/ts/routes/image-occlusion/store.ts b/ts/routes/image-occlusion/store.ts
index b1653f4fa..5847250dd 100644
--- a/ts/routes/image-occlusion/store.ts
+++ b/ts/routes/image-occlusion/store.ts
@@ -17,5 +17,3 @@ export const ioImageLoadedStore = writable(false);
 export const opacityStateStore = writable(false);
 // store state of text editing
 export const textEditingState = writable(false);
-// store state of shape ungroup
-export const shapeUngroupState = writable(false);
diff --git a/ts/routes/image-occlusion/tools/lib.ts b/ts/routes/image-occlusion/tools/lib.ts
index 9a6738b10..ea60ad3d3 100644
--- a/ts/routes/image-occlusion/tools/lib.ts
+++ b/ts/routes/image-occlusion/tools/lib.ts
@@ -4,7 +4,7 @@
 import { fabric } from "fabric";
 import { get } from "svelte/store";
 
-import { opacityStateStore, shapeUngroupState } from "../store";
+import { opacityStateStore } from "../store";
 
 export const SHAPE_MASK_COLOR = "#ffeba2";
 export const BORDER_COLOR = "#212121";
@@ -94,11 +94,14 @@ export const unGroupShapes = (canvas: fabric.Canvas): void => {
     canvas.remove(group);
 
     items.forEach((item) => {
-        item.set({ opacity: get(opacityStateStore) ? 0.4 : 1 });
+        item.set({
+            opacity: get(opacityStateStore) ? 0.4 : 1,
+            // @ts-expect-error not defined
+            ordinal: undefined,
+         });
         canvas.add(item);
     });
 
-    shapeUngroupState.set(true);
     redraw(canvas);
 };
 

```

</details>

dae: Nice catch @abdnh! And thanks for the quick fix @krmanik.

Unfortunately I ran into another issue. I created three rectangles and merged them into a group, then created three more, and left them ungrouped. I added the note. Then I edited it, and ungrouped the three rectangles. Two surprises:

- The group's ordinal was not reused, so the first card is now empty and needs to be cleaned up with tools>empty cards
- One of the shapes in the group was given the same ordinal as one of the shapes that was never in a group, so it ended up in a new group.

ts/routes/image-occlusion/shapes/to-cloze.ts
```diff
@@ -21,6 +21,9 @@ export function exportShapesToClozeDeletions(occludeInactive: boolean): {
     let clozes = "";
     let index = 0;
     let nextOrdinal = 1;
+
+    console.log(shapes);
```
krmanik, line 25: Adding this console log seems to fix the issue, so is the issue race condition?

dae: I'm afraid for me, the same issue still happens when I update to the latest commit which has this console.log call. I tried 3 times, and it happened on the 1st and 2nd time, but not the 3rd. If you try a few more times, does it ever happen? Could it be something other than a race condition, such as unpredictable dictionary ordering?

krmanik: I didn't understand the underlying issues, so I have changed the implementation, it works as expected. The general idea is takeout ordinal which does not assign to any shape from a list of ordinal and assign to shape.

krmanik:
The issue is resolved now in latest commit.

dae:
Great, thanks Mani. Will review this tomorrow. Abdo, if you have a chance
before then, your input would be appreciated.

dae:
Thanks Mani, this seems to work well, and if there are any issues with the code, I didn't notice them. ^_^; I'm going to merge this in now so I can get a beta out today, but if you have any concerns @abdnh, please let us know.

abdnh:
Looking good to me. Didn't notice any issues either.

PR #3114, opened 2024-04-04
Ignore except raised by pip_system_certs not found

### LINKED ISSUE #3113: Anki gives ModuleNotFoundError without pip_system_certs installed
As of version 24.04, if `pip_system_certs` module is not installed on the system, running Anki will error like this:

```shell
$ anki
  File "/usr/bin/anki", line 17, in <module>
    import aqt
  File "/usr/lib/python3.11/site-packages/aqt/__init__.py", line 9, in <module>
    import pip_system_certs.wrapt_requests
ModuleNotFoundError: No module named 'pip_system_certs'
```

This dependency is introduced by #3050, to solve #3016, and the original bug report is on MacOS with a niche network setup. This dependency does not benefit normal Linux users using Anki with their system packages at all, as most Linux distros have already [configured](https://gitlab.archlinux.org/archlinux/packaging/packages/python-requests/-/blob/main/certs.patch) their python-requests module to use system certificate instead, also most users don't have this niche network setup.

This `pip_system_certs` module is not packaged officially for most Linux distros e.g. Fedora, Ubuntu, Debian, Arch. It has only [12 stars on GitLab](https://github.com/conan-io/docs/issues/3123#issuecomment-1479767544). And will break existing binary installation process. It will be better if we can catch and ignore the `ModuleNotFoundError` thus make this dependency an optional one.

escape0707:
Close #3113

This dependency usually doesn't benefit Linux distros with requests library configured to use system certificate already. And is not packaged by most distros. Making it optional will make most Linux users' installation much easier.


escape0707:
Maybe `requirements.aqt.in` should be changed too, if all Linux distros don't need this dependency? I'm not sure about that, tho.

dae:
Thank you for the change. I think this is fine, but we should probably print a message to stdout in the exception case - something like "pip_system_certs not installed; custom SSL certs may not work".

Distros probably need to make up their own mind about the .in/.txt files - either patching it out or using their own dependency list.

escape0707:
> we should probably print a message to stdout in the exception case - something like "pip_system_certs not installed; custom SSL certs may not work".

Great advice, I'll change the code to print that asap.

> Distros probably need to make up their own mind about the .in/.txt files - either patching it out or using their own dependency list.

I think maybe packagers who actually make packages out of source code building should have greater say than me. So I'll leave the judge to them.

dae:
Thank you!

PR #3121, opened 2024-04-07
Fix IO mask position slightly off in edit mode

abdnh:
I was looking into #2986 again and noticed all shapes suffer from a positioning issue (probably introduced after the recent zoom changes).

## How to reproduce

1.  Add a new note with a mask
image
2. Edit the note and notice the position of the mask
image

## Reference

http://fabricjs.com/docs/fabric.Object.html#getBoundingRect


dae:
Thanks Abdo!

PR #3129, opened 2024-04-09
Update deck-config.ftl to clarify what optimal retention means

### LINKED ISSUE #3118: Clarify what optimal retention means
> seems like renaming “Optimal retention” to “Minimum recommended retention” is considered a good idea, so I would suggest renaming it.

Originally reported on https://forums.ankiweb.net/t/clarify-what-optimal-retention-means/42803/35

AntoineQ1:
Renaming “Optimal retention” to “Minimum recommended retention".
_Currently, only the text has been changed, not the variables (e.g. "deck-config-compute-optimal-retention")._

Resolves #3118.

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -343,7 +343,7 @@ deck-config-must-have-400-reviews =
 # Numbers that control how aggressively the FSRS algorithm schedules cards
 deck-config-weights = FSRS parameters
 deck-config-compute-optimal-weights = Optimize FSRS parameters
-deck-config-compute-optimal-retention = Compute minimum recommended retention
+deck-config-compute-minimum-recommended-retention = Compute minimum recommended retention
 deck-config-optimize-button = Optimize
 deck-config-compute-button = Compute
 deck-config-ignore-before = Ignore reviews before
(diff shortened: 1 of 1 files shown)


ftl/core/deck-config.ftl
```diff
@@ -357,7 +357,7 @@ deck-config-get-params = Get Params
 deck-config-fsrs-on-all-clients =
     Please ensure all of your Anki clients are Anki(Mobile) 23.10+ or AnkiDroid 2.17+. FSRS will
     not work correctly if one of your clients is older.
-deck-config-predicted-optimal-retention = Predicted optimal retention: { $num }
+deck-config-predicted-optimal-retention = Predicted minimum recommended retention: { $num }
```
dae, line 360: How about we drop 'predicted', so "Minimum recommended retention: ..."?

dae (commented):
These changes change the meaning of the text, so we should use new keys (e.g.  instead of only changing the text, so that translators get notified. Please keep the existing strings, but move them to the bottom of the file, so we can clean them up in the future.

AntoineQ1:
Are further changes needed, or is this enough?

dae:
Thank you!

PR #3134, opened 2024-04-11
Improve typing of custom fabric.Object properties

abdnh:


dae:
Thanks as always Abdo. Under what circumstances does the current bug get triggered? Wondering whether I should rebuild 24.04.1 with this fix.

dae:
(the typing improvements are a nice bonus, by the way :-)

abdnh:
> Under what circumstances does the current bug get triggered?

Not really sure actually :) I recall seeing the issue while testing #3109 a couple of days ago but then I can no longer reproduce it now. I took a look now at Fabric's source to see how canvas.getObjects() is implemented: it simply copies canvas._objects, while canvas.add() appends to the array, so the issue shouldn't occur and I must have been imagining things. I'll revert this fix and keep the typing change.

dae:
Thanks Abdo, that's welcome news - building the releases is currently rather involved, so was nice not to have to do it again for 24.04.1!

PR #3135, opened 2024-04-11
Svg icon

### LINKED ISSUE #3127: SVG icon styling is broken
Probably by #3077.

I found out that instead of `@mdi/svg` you can use `@mdi/js`, which contains the raw path data, and then wrap the icons in `ts/lib/components/icons.ts` with svg instead of img elements. This way, CSS styling like `fill` is correctly applied.

However, then I noticed that the icons from `ts/routes/image-occlusion/icons.ts` are already rendered as svgs, despite seemingly using the exact same logic. Is there some preprocessing magic going on?

RumovZ:
Did you mean something like this in #3127?
The diff worked at first, but at some point I also had to `yarn add --dev svgo`.

I couldn't find a way to satisfy both eslint and dprint. Is the simple-import-sort plugin strictly necessary?

I moved all icon imports into one module to make it easier to use a consistent set of icons across Anki.

Closes #3127


RumovZ:
@dae, do you have a suggestion how I should deal with the formatting issue? Sorry for the ping, but I have made a few other styling fixes that I would like to push after this is out the door.

dae:
Sorry, I lost track of this - please feel free to ping me any time I don't reply after a few days. If dprint is already sorting the imports for us, let's ditch the plugin - it's not the first time it's caused trouble.

RumovZ:
Great! Then this is ready for merge from my side.

In the future when eslint is dropped, Icon.svelte and the URL imports in icons.ts can be removed.

dae:
Thanks for fixing this, Rumo!

PR #3136, opened 2024-04-11
Add keyboard shortcuts for empty cards and toggle mask on image occlusion

### LINKED ISSUE #3122: Keyboard shortcuts for: tools -> empty cards…, image occlusion -> toggle mask editor
> It would be nice to have keyboard shortcuts for both these functions which I find myself using quite often. Thanks for your work on Anki!

Originally reported on https://forums.ankiweb.net/t/keyboard-shortcuts-for-tools-empty-cards-image-occlusion-toggle-mask-editor/43215/4

escribapetrus:
# Add keyboard shortcuts for empty cards and toggle mask on image occlusion

This change adds two shortcuts:
- empty cards (Ctrl+Shift+0)
- toggle mask (Ctrl+Shift+M)

Please let me know if there is a better way to set these shortcuts.  🍄 

Closes #3122


escribapetrus:
It's weird that the email information is breaking the CI. Not only have I made my email public, but also I was already on the CONTRIBUTORS file.

dae:
When you changed that file, you must have made a change via the GitHub UI, so the listed email is a users.noreply.github.com one. Add a space to your existing entry using your local machine and push the change, and it should fix it.

Please try to find something other than [. It likely will cause problems on some keyboard layouts, where [ is accessed with/without the shift key.

dae:
(ctrl+alt is another option)

escribapetrus:
@dae Thank you. Fixed the error on CONTRIBUTORS with your tip.

How do you feel about `Ctrl+Shift+0` for empty cards?

dae:
I don't know if that would cause problems. AZERTY keyboard layouts apparently require shift to be held down to access numbers, so I'm not sure ctrl+shift+0 could be distinguished from ctrl+0, which we're already using.

How about ctrl+alt+e?

dae:
Thank you!

Issue #3041, opened 2024-02-28
About section, macOS


okkainen:
I suggest:

1. using an icon with a transparent background 
2. Introduce margins to the buttons:
apply a left margin to the "Copy Debug Info" button.
apply a right margin to the "OK" button.

<img width="855" alt="image" src="https://github.com/ankitects/anki/assets/37247954/8bf0fa90-8ac1-4dee-9c6a-03b5278c53d7">

2.1. Additionally, I'd suggest adding a bottom margin to both buttons to ensure consistency (in terms of bottom margin) with the bottom buttons in other sections, as illustrated in the screenshot below:
<img width="760" alt="image" src="https://github.com/ankitects/anki/assets/37247954/33219292-2aaf-473c-bc93-caef4a1caaf6">

nbybhen:
Hey, I'm willing to work on this issue. Seems like a great first-issue to enter into the with!

dae:
Sure, please go ahead.

nbybhen:
I've taken time to sift through the files, and am still unsure as to where the contributions for the About page would be. I was under the assumption that it would be displayed via TS, but I was mistaken. Mind giving any pointers?

dae:
Grep for one of the names, and it will reveal the file.

ShreyasKallingal:
I think this should be marked as closed. Was about to start working on it when I realized it was fixed by #3137.

jamesnelmore:
Is a transparent icon still something that is wanted? I think at the very least we should round the corners.

dae:
It's not a high priority, but if it's something you feel like improving, I'd be happy to accept a PR.

marcusl07:
Very minor issue, but the OK button has two little black lines going through it. 

<img width="373" alt="Image" src="https://github.com/user-attachments/assets/13fba461-d686-48df-aaf7-e4afdc92ef9e" />

graves:
> > OK button has two little black lines going through it.
> 
> It doesn't for me (on latest Anki build from source; on debian sid). Maybe you're experiencing graphics related issues? Try to change the video driver to something else and check if it persists: https://docs.ankiweb.net/troubleshooting.html?#troubleshooting.

Can confirm.

```
Version ⁨25.02 (63c2a09e)⁩
Python 3.9.18 Qt 6.6.2 PyQt 6.6.1
```

Just cloned the `main` branch and built on Apple Silicon. The OK button looks good!

[...]

PR #3137, opened 2024-04-12
Add Margins on the Buttons on the About Page

Ianwu0812:
It's my first Pull Request in Anki, if you see any problem, please let me know.

This change is want to fix the opening Issue, which is https://github.com/ankitects/anki/issues/3041.

The problem mentioned in the Issue is about the buttons on the About window, does not have margin.

I added the margin on the Button in the About window to fix this Issue.


dae:
Thank you Wu - looks good!

PR #3141, opened 2024-04-14
Return current weights if fsrs items is zero & handle error in evaluation and optimal retention

L-M-Sherlock:
Related discussion: https://forums.ankiweb.net/t/anki-24-04-2-beta/43562/8?u=l.m.sherlock

I don't know why the error isn't handled like Anki 24.04.1:

image

So I handle it in the `.svelte` file. I'm not sure it is a right way to handle it. Please feel free to modify it.


dae:
Thanks Jarrett. I've pushed a more general fix for the unreported errors.

PR #3142, opened 2024-04-14
Ignore errors in do_window_cleanup to fix shutdown issue

### LINKED ISSUE #3126: Anki 24.04 does not close properly
> I recently updated to latest anki version 24.04 on my Fedora 39 (Linux).
> In this version when I start anki and close it multiple times without any changes made to the decks or settings.
> At one point it closes with an instant effect and then If I start to run it again I cannot do it because it says Already running in terminal window. For which I have to go to system monitor and Kill it from there to start it again.
> It was not happening in earlier version 23.12.
> I just want it to close properly so I do not have to go in system monitor to kill the process.

Originally reported on https://forums.ankiweb.net/t/anki-24-04-does-not-close-properly/43185/15

Probably connected to 7784321b90be4ed0b9a7cc157cd88295722cb826. Maybe we could swallow errors when accessing the timers, or have some 'shutting down' flag that causes the errors to be ignored?

abdnh:
Closes #3126

I was able to reproduce it by changing time_to_wait too (and having auto sync enabled).

### DIFF
--- a/qt/aqt/progress.py
+++ b/qt/aqt/progress.py
@@ -232,20 +232,23 @@ class ProgressManager:
 
             next_levels = self._levels - 1
             next_levels = max(0, next_levels)
-            if next_levels == 0:
-                if self._win:
-                    self._closeWin()
-                if self._busy_cursor_timer:
-                    self._busy_cursor_timer.stop()
-                    self._busy_cursor_timer = None
-                self._restore_cursor()
-                if self._show_timer:
-                    self._show_timer.stop()
-                    self._show_timer = None
-            if self._backend_timer:
-                self._backend_timer.stop()
-                self._backend_timer.deleteLater()
-                self._backend_timer = None
+            try:
+                if next_levels == 0:
+                    if self._win:
+                        self._closeWin()
+                    if self._busy_cursor_timer:
+                        self._busy_cursor_timer.stop()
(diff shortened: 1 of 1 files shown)


voczi:
Figured out you could just do it like this:
`if self._x_timer and not sip.isdeleted(self._x_timer):`
Repeat for every timer in that method.

Docs for the single_shot method did also mention this.

abdnh:
Yes, I thought of that, but preferred to go for less code.

qt/aqt/progress.py
```diff
@@ -232,20 +232,23 @@ def do_window_cleanup(future: Future | None = None):
...
+                    self._restore_cursor()
+                    if self._show_timer:
+                        self._show_timer.stop()
+                        self._show_timer = None
+                if self._backend_timer:
+                    self._backend_timer.stop()
+                    self._backend_timer.deleteLater()
+                    self._backend_timer = None
+            except RuntimeError as exc:
+                print(f"do_window_cleanup error ignored: {exc}")
```
dae, line 251:
```suggestion
                # during shutdown, the timers may have already been deleted by Qt
                print(f"do_window_cleanup error ignored: {exc}")
```

dae:
Great, thanks Abdo!

PR #3143, opened 2024-04-14
Avoid locale.getdefaultlocale() if possible

### LINKED ISSUE #3115: DeprecationWarning: 'locale.getdefaultlocale' is deprecated
When running Anki with PYTHONDEVMODE=1 I see:

```
[...]/lib/python3.11/site-packages/anki/lang.py:185: DeprecationWarning: 'locale.getdefaultlocale' is deprecated and slated for removal in Python 3.15. Use setlocale(), getencoding() and getlocale() instead.
  (sys_lang, enc) = locale.getdefaultlocale()
```

abdnh:
Closes #3115

locale.getlocale() returns "English_United States" as the language code on my system: https://bugs.python.org/issue38805

This PR falls back to locale.getlocale() when locale.getdefaultlocale() is not available.

### DIFF
--- a/pylib/anki/lang.py
+++ b/pylib/anki/lang.py
@@ -183,6 +183,8 @@ def get_def_lang(lang: str | None = None) -> tuple[int, str]:
     or English if not available."""
     try:
         (sys_lang, enc) = locale.getdefaultlocale()
+    except AttributeError:
+        (sys_lang, enc) = locale.getlocale()
     except:
         # fails on osx
         sys_lang = "en_US"
--- a/pylib/anki/lang.py
+++ b/pylib/anki/lang.py
@@ -183,6 +183,7 @@ def get_def_lang(lang: str | None = None) -> tuple[int, str]:
     """Return lang converted to name used on disk and its index, defaulting to system language
     or English if not available."""
     try:
+        # getdefaultlocale() is deprecated since Python 3.11 but we need to keep using it due to a bug: https://bugs.python.org/issue38805
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", DeprecationWarning)
             (sys_lang, enc) = locale.getdefaultlocale()
--- a/pylib/anki/lang.py
+++ b/pylib/anki/lang.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import locale
 import re
+import warnings
 import weakref
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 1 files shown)


dae:
I don't mind hiding the warning for now to stop users complaining about it, but perhaps the fallback to the new API will do more harm that good, if it's returning the locale in a different format? Isn't that effectively the same as falling back to en_US on Windows?

A brief comment about how the new API differs might be worth adding, so we don't have to rediscover this in the future when we update to 3.15 and find startup is crashing.

abdnh:
My hope was that the issue will be fixed before Anki switches to 3.15.

> find startup is crashing.

But the code is catching AttributeError specifically for that reason? I'll add a comment.

pylib/anki/lang.py
```diff
@@ -182,7 +183,12 @@ def get_def_lang(lang: str | None = None) -> tuple[int, str]:
     """Return lang converted to name used on disk and its index, defaulting to system language
     or English if not available."""
     try:
-        (sys_lang, enc) = locale.getdefaultlocale()
+        # getdefaultlocale() is deprecated since Python 3.11 but we need to keep using it due to a bug: https://bugs.python.org/issue38805
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            (sys_lang, enc) = locale.getdefaultlocale()
+    except AttributeError:
+        (sys_lang, enc) = locale.getlocale()
```
dae, line 191:
```suggestion
        # this will return a different format on Windows (e.g. Italian_Italy), resulting in us falling back to en_US
        # further below
        (sys_lang, enc) = locale.getlocale()
```

pylib/anki/lang.py
```diff
@@ -182,7 +183,12 @@ def get_def_lang(lang: str | None = None) -> tuple[int, str]:
     """Return lang converted to name used on disk and its index, defaulting to system language
     or English if not available."""
     try:
-        (sys_lang, enc) = locale.getdefaultlocale()
+        # getdefaultlocale() is deprecated since Python 3.11 but we need to keep using it due to a bug: https://bugs.python.org/issue38805
```
dae, line 186:
```suggestion
        # getdefaultlocale() is deprecated since Python 3.11, but we need to keep using it as getlocale() behaves differently: https://bugs.python.org/issue38805
```

dae (commented):
I've added some extra comments to clarify what I meant. Happy for me to merge in?

PR #3144, opened 2024-04-14
HMR on Windows

RumovZ:
With 9f40f0c7034594d283a7a01b64ec9a3e48f85d28, I got HMR working in the browser. But if I understood you correctly in #3077, it should also work in Qt. There I only get `ERR_CONNECTION_REFUSED` with `HMR=1`. Did it work for you, @abdnh?
The console shows nothing. Turning off the firewall didn't make a difference.

### DIFF
--- a/.vscode.dist/launch.json
+++ b/.vscode.dist/launch.json
@@ -6,9 +6,13 @@
     "configurations": [
         {
             "name": "Run",
-            "type": "python",
+            "type": "debugpy",
             "request": "launch",
             "program": "tools/run.py",
+            "args": [
+                // "-p",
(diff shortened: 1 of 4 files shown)


dae:
Thanks Rumo! I've pushed a fix for using it in Qt.

dae:
Does fiddling with `set QTWEBENGINE_CHROMIUM_FLAGS=--remote-allow-origins=http://localhost:8080` make a difference, e.g. using 127.0.0.1 instead?

abdnh:
It's caused by VLAN software as I suspected.

PR #3150, opened 2024-04-18
Fix spacebar causing stutter when editing cards

### LINKED ISSUE #3131: Pressing the spacebar will stutter when editing cards
See the description/videos on the link below.

Originally reported on https://forums.ankiweb.net/t/pressing-the-spacebar-will-stutter-when-modifying-the-card/43385

I haven't attempted to reproduce this yet. First step will be trying to reproduce, then bisecting to discover what caused this.

abdnh:
Some image occlusion event handlers were getting triggered globally in the editor. The keydown handler is registered inside a mousemove handler, which is resulting in countless handlers.

Closes #3131


dae:
Nice - thanks for getting to the bottom of this!

PR #3153, opened 2024-04-20
Fix auto sync not working on Windows shutdown

### LINKED ISSUE #2577: Changes not being synced when shutting down PC
Anki automatically syncs cards/changes when closing the program, which works just fine, when ... closing the program.

However, when the user decides to not properly do a sync or close the program, but instead just shuts down the whole PC without manually closing the program first, no final sync is happening.

This is quite easy to reproduce:
1. Sync a second device using AnkiWeb
2. Make sure they are on the same version (sync both repeatedly until no changes are done)
3. Do a change on the first device, do not sync manually
4. Shut down the PC without doing another sync
5. Sync the second device, see that no changes have happened
6. Only when booting the first device and manually doing a sync again, the second device can receive the change from step 3.

Expected behaviour:
Just as when manually closing the program, it should do a final sync on shutting down the whole device, instead of just cutting off and keeping the changes locally-only.

I'm not sure which OSes this actually happens with, I just tested with Windows - feel free to comment/add labels if other systems are affected as well.

abdnh:
Closes #2577

## Solution

This handles the [WM_QUERYENDSESSION](https://learn.microsoft.com/en-us/windows/win32/shutdown/wm-queryendsession) message to block shutdown until syncing is done.

## Reproduction

Can be tested more easily using this tool: https://github.com/stefankueng/sendmessage

### DIFF
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -321,20 +321,19 @@ class AnkiApp(QApplication):
     def _set_windows_shutdown_block_reason(self, reason: str) -> None:
         if is_win:
             import ctypes
-            import ctypes.wintypes
+            from ctypes import windll, wintypes  # type: ignore
 
-            ctypes.windll.user32.ShutdownBlockReasonCreate(
-                ctypes.wintypes.HWND.from_param(int(mw.effectiveWinId())),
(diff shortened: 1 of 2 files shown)


dae:
Great work ! Thanks for this Abdo.

PR #3160, opened 2024-04-22
fix io draw in mobile client

krmanik:
Adding this to `onTouchstart` fix the issue.

```
const canvas = globalThis.canvas;
```
https://github.com/ankitects/anki/commit/c62e2d8df0f8bbdfedfb6d71bbf09d1ca353253f


dae:
Thanks so much for the quick fix, Mani!

PR #3162, opened 2024-04-23
Add number truncation before back-end translation

### LINKED ISSUE #3102: “1 hours today”
> I’ve noticed that if you study for exactly one hour, you’re told “Studied n cards in 1 hours today (_m_s/card).”
> This is an incredibly minor point, but ideally it would say “1 hour” instead of “1 hours” (but be plural again for “1.1 hours”, “1.2 hours”, etc.).

Originally reported on https://forums.ankiweb.net/t/1-hours-today/42956

It's caused by non-integer numbers that are rounded to an integer in format_number_values(), eg 1.0005 becomes 1.00 which ends up being to 1. I'm not sure we can easily fix this, as I think the plural choice is happening in fluent's code, not ours. Converting the fluent numbers to integers if they round to an integer might work?

lucasscharenbroch:
Fixes #3102

When the [front-end `translate`](https://github.com/ankitects/anki/blob/839fd7bc8bcafd73aac0259d2efdc273486c6a98/ts/lib/generated/ftl-helpers.ts#L17) is called, [all numbers are truncated to 2 decimal places](https://github.com/ankitects/anki/blob/839fd7bc8bcafd73aac0259d2efdc273486c6a98/ts/lib/generated/ftl-helpers.ts#L23).
The same was not true for the back-end `translate`, which seems to be the cause of this corner-case.

It appears that the Rust implementation of Fluent doesn't automatically do the truncation (before the comparison to `[one]` in FTL) according to the FluentNumberOptions (`maximum_fraction_digits`), while the JS version does. This is why my code explicitly rounds the number, where the TS only sets `maximumFractionDigits`.

### DIFF
--- a/rslib/i18n/src/lib.rs
+++ b/rslib/i18n/src/lib.rs
@@ -482,6 +482,10 @@ mod test {
             "You have 1 hat."
         );
         assert_eq!(
+            tr.translate("plural", Some(tr_args!["hats"=>1.001])),
+            "You have 1 hat."
+        );
+        assert_eq!(
             tr.translate("plural", Some(tr_args!["hats"=>1.1])),
             "You have 1.1 hats."
         );
--- a/rslib/i18n/src/lib.rs
+++ b/rslib/i18n/src/lib.rs
@@ -22,13 +22,46 @@ type FluentBundle<T> = FluentBundleOrig<T, intl_memoizer::concurrent::IntlLangMe
 
 pub use fluent::fluent_args as tr_args;
 
-pub trait Number: Into<FluentNumber> {}
-impl Number for i32 {}
-impl Number for i64 {}
-impl Number for u32 {}
-impl Number for f32 {}
-impl Number for u64 {}
-impl Number for usize {}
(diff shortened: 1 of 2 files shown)


dae:
Thanks for looking into this Lucas!

There are a couple of things that would be nice to have:

- Instead of rebuilding the args into a new container, we could apply the transformation directly in the generated code. The strings.rs file we generate creates functions like this:

```
    /// {$count} cards added.
    #[inline]
    pub fn importing_cards_added<'a>(&'a self, count: impl Number) -> Cow<'a, str> {

        let mut args = FluentArgs::new();
        args.set("count", count.into());

        self.translate("importing-cards-added", Some(args))
    }
```

For numbers, we could call a function to modify them before calling .set(), and avoid the extra work.

- A small unit test that verifies things work correctly (i.e. if the fix is removed, the unit test starts failing) would be good. You could pick a random string from the generated code to test with.

lucasscharenbroch:
Thanks, Dae!

I've made changes accordingly.

PR #3148, opened 2024-04-17
update optimal retention and parameters tooltip

L-M-Sherlock:
I removed the words about the 1000 review limitation from the tooltip of FSRS parameters, and mentioned the 400 review limitation in the tooltip of optimal retention.

PR #3165, opened 2024-04-24
Show review count with FSRS evaluation results

### LINKED ISSUE #3070: Shift revlog count from progress to return value
Sorry for barging in, but I suggest adding the review count to the "Evaluate" window.
image

_Originally posted by @Expertium in https://github.com/ankitects/anki/issues/3051#issuecomment-1987231996_

abdnh:
Closes #3070


dae:
Thanks Abdo! It might be nice to do something similar for the Optimize button, to solve https://github.com/ankitects/anki/pull/3148#issuecomment-2069698296. Since we don't have a pop-up for that one, maybe we could just display it underneath, like we do for progress. What do you think?

PR #3166, opened 2024-04-24
Use key instead of code in type-in field

### LINKED ISSUE #3069: Could the webview code use KeyboardEvent.key instead of KeyboardEvent.code?
> I attempted to upgrade from Anki 2.0.47 to 24.04beta1 but I was unable to answer any cards with a type-in-answer field like {{type:field}}. This issue occurs because I use a custom keyboard layout that swaps the enter key with the caps lock key, because I find it more ergonomic on European keyboards. While this setup works well with all other programs, the new Anki version failed to recognize my custom caps lock enter. I suspect this is because Anki is using raw keycodes instead of keysyms.

Originally reported on https://forums.ankiweb.net/t/could-the-webview-code-use-keyboardevent-key-instead-of-keyboardevent-code/42226

abdnh:
Closes #3069

This only switches to KeyboardEvent.key when Enter is handled to avoid any potential issues.

### DIFF
--- a/ts/lib/sveltelib/input-handler.ts
+++ b/ts/lib/sveltelib/input-handler.ts
@@ -116,7 +116,7 @@ function useInputHandler(): [InputHandlerAPI, SetupInputHandlerAction] {
             specialKey.dispatch({ event, action: "caretRight" });
         } else if (isArrowLeft(event)) {
             specialKey.dispatch({ event, action: "caretLeft" });
-        } else if (event.code === "Enter" || event.code === "NumpadEnter") {
+        } else if (event.key === "Enter") {
             specialKey.dispatch({ event, action: "enter" });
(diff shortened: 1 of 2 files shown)


dae:
Were you able to confirm this works with numpad enter too? My regular keyboards lack a numpad, but I have one lying around somewhere I can plug in, if you don't have one.

abdnh:
> Were you able to confirm this works with numpad enter too?

My keyboard doesn't have numpad too, but I assume it should work since no other value is listed for Enter here: https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#whitespace_keys

dae:
I double-checked with a numeric keyboard just in case, and can confirm it works fine. Thanks :-)

PR #3167, opened 2024-04-24
Improve and unify web dialog styling

### LINKED ISSUE #3156: Insufficient padding in import page
As visible in the screenshot below, there isn't enough padding, especially on the left side.

image

Anki version 24.04.2 (a1fa865b)⁩

RumovZ:
This makes the deck options, change notetype and import dialogs look more consistent, and introduces various other improvements (and possibly subjective tweaks).

Incidentally fixes  #3156.

@abdnh, heads-up because I think you've worked on these dialogs. Hope it's agreeable.


ts/lib/components/ButtonToolbar.svelte
```diff
@@ -39,11 +39,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     .button-toolbar {
         flex-wrap: var(--buttons-wrap);
         padding-left: 0.15rem;
-
-        :global(.button-group) {
-            /* TODO replace with gap once available */
-            margin-right: 0.15rem;
-            margin-bottom: 0.15rem;
-        }
+        gap: 1rem;
```
RumovZ, line 42: That comment was from 2021, and `gap` is supposedly [well supported since April 2021](https://caniuse.com/flexbox-gap). Not sure which client was the blocker at the time, but hopefully it's fine now.
(The `margin` solution was broken.)

ts/routes/import-page/ImportLogPage.svelte
```diff
@@ -7,47 +7,46 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-    $: foundNotes = result?.log?.foundNotes ?? 0;
-    let closeButton: HTMLElement;
+    $: summaries = getSummaries(response.log!);
+    $: foundNotes = response.log?.foundNotes ?? 0;
+
+    const gutterBlockSize = 0.5;
+    const computedStyle = getComputedStyle(document.documentElement);
+    const rootFontSize = parseInt(computedStyle.fontSize);
+    // Container padding + Row padding + Row margin + TitledContainer padding
+    const bottomOffset = (3 * gutterBlockSize + 0.75) * rootFontSize;
```
RumovZ, line 25: Not entirely happy with this fragile solution, but couldn't think of a viable alternative.

ts/lib/components/ButtonToolbar.svelte
dae: Qt5 is Chromium 77, so this solution won't work with the Qt5 build unfortunately. The new spacing is also a bit too big I think - it used to be smaller.

dae:
Thanks for this Rumo - I'll give it a proper review within a few days.

ts/lib/components/ButtonToolbar.svelte
RumovZ: Pity, I'll revert to using `margin`.

In my opinion, there are many places where a bit more spacing wouldn't hurt, but since that's more or less a matter of taste, I tried to hold off from such changes.
However, in this case there wasn't any margin at all. Does it look different on a Mac?

Before:
grafik

Now:
grafik

abdnh:
Nice work as usual @RumovZ! I like the improvements to the import log screen. Looks much nicer with these small tweaks!

ts/lib/components/ButtonToolbar.svelte
dae: I agree the latter screenshot looks better. I was looking at the editor when I made my previous comment - the buttons feel a bit widely-spaced now:
2024-05-01T17:42:57,625111958+10:00

It was perhaps a bit too narrow before:
2024-05-01T17:45:23,849219010+10:00

So maybe somewhere in between might be nice :-)

dae:
Sorry for the delay. I'll merge this in so I don't hold it up any further, and we can address the editor spacing in a follow-up.

ts/lib/components/ButtonToolbar.svelte
RumovZ: Whoops, that was unintended. I only meant to change the 3 pages mentioned above. Thanks for noticing! I'll push a follow-up.

PR #3171, opened 2024-04-27
update to fsrs-rs v0.6.2

L-M-Sherlock:
It improves the accuracy of FSRS (mainly in the pretraining stage). I'm not sure whether it could catch up the 24.04.2?


dae:
Thanks Jarrett. Currently the main branch is a bit unpolished due to the Svelte changes, and we're in May now, so 24.04.2 will probably become 24.05 and take some more time.

PR #3174, opened 2024-04-27
Do not clear unused parent tags

abdnh:
A user reported to me that sidebar tags are getting collapsed after clearing unused tags.


rslib/src/storage/note/mod.rs
```diff
@@ -217,16 +219,19 @@ impl super::SqliteStorage {
             .map_err(Into::into)
     }
 
-    pub(crate) fn all_tags_in_notes(&self) -> Result<HashSet<String>> {
+    pub(crate) fn all_tags_in_notes(&self) -> Result<HashSet<UniCase<String>>> {
```
dae, line 222: Minor nitpick: you'd typically expect the storage layer to return the exact data that's stored on disk, with a higher-level (e.g. clear_unused_tags_inner()) taking care of synthesizing implied values. Let's at least document that it's a bit unusual:

```suggestion
    /// All tags referenced by notes, and any parent tags as well.
    pub(crate) fn all_tags_in_notes(&self) -> Result<HashSet<UniCase<String>>> {
```

dae:
Thanks, just a minor comment.

PR #3181, opened 2024-04-29
Use median in calculating cost and remove outliers

### LINKED ISSUE #3172: (FSRS) Use median in calculating cost and remove outliers
Corresponding PR in Python optimizer:
- https://github.com/open-spaced-repetition/fsrs-optimizer/pull/109

Changes:
- Use median instead of mean to calculate average costs
- Remove entries where time = 0 ms
- Remove entries where time > 1200000 ms (20 min)

The second change was probably already made in
- https://github.com/ankitects/anki/pull/2870.

L-M-Sherlock:
Close #3172


rslib/src/scheduler/fsrs/retention.rs
```diff
@@ -135,15 +135,22 @@ impl Collection {
...
+                    let mut group_vec = group.into_iter().map(|r| r.taken_millis).collect_vec();
+                    group_vec.sort_unstable();
+                    let median_millis = if group_vec.len() % 2 == 0 {
+                        let mid = group_vec.len() / 2;
+                        (group_vec[mid - 1] + group_vec[mid]) as f64 / 2.0
+                    } else {
+                        group_vec[group_vec.len() / 2] as f64
+                    };
+                    let median_secs = median_millis / 1000.0;
+                    arr[button_chosen as usize - 1] = median_secs;
```
dae, line 153: The median calculation is repeated 3 times - it would be nicer if this was factored out into a separate function, and that would also make it easy to add a simple unit test that makes sure it handles 0, 1 and 2 element vecs correctly.

L-M-Sherlock: I have refactored the code just now.

dae: Thanks!

PR #3182, opened 2024-04-29
Fix '366 of 365 days studied'

### LINKED ISSUE #2961: 366 of 365 days studied

abdnh:
Closes #2961

I think the issue here is that .nice() is modifying the domain to `[-365, 5]` when `xMin = -364`, so the graph is including an additional day.

I couldn't confirm whether @iuliachiriac's investigation in #3044 applies to the new graphs screen.

Here's the code I used to insert revlog entries for testing:
```python
from aqt import mw

cid = 1714400365037
for i in range(1, 367):
    id = 1000 * (mw.col.sched.day_cutoff - 86400 * i + 50)
    mw.col.db.execute(
        "INSERT OR REPLACE INTO revlog (id,cid,usn,ease,ivl,lastIvl,factor,time,type) VALUES (?, ?, -1, 3, 1.5, 1.5, 2500, 100, 0)",
        id,
        cid,
    )
```


dae:
Code to generate the entries was a nice touch!

It looks like I originally added .nice() to fix an issue, so I'm a little concerned this will cause a regression. Sadly, I wasn't very specific about the issue at the time, but suspect it's one of the ones mentioned on/before Jul 6: https://forums.ankiweb.net/t/anki-2-1-28-beta/629

cfc33815da285244b3045c08734314b96f5068d6

abdnh:
Probably it's this issue: https://forums.ankiweb.net/t/anki-2-1-28-beta/629/9
I guess it's only relevant for GraphRange.AllTime, so I pushed a change that applies nicing there.

dae:
Good thinking. Thanks!

PR #3184, opened 2024-04-30
fix polygon tool draw

krmanik:
> > Also, I found that the polygon  tool of image occlusion doesn’t work in in anki 24.04.1.
> 
> Originally reported on https://forums.ankiweb.net/t/is-this-bug-or-the-new-modification-in-anki-desktop/44072
>


dae:
Thanks as always, Mani!

PR #3186, opened 2024-05-01
Clean up #3167

RumovZ:
https://github.com/ankitects/anki/pull/3167#discussion_r1585985319


dae:
Thanks for the quick fix Rumo! Editor spacing looks good IMHO.

PR #3188, opened 2024-05-02
add template filter "cloze-show"

jakeprobst:
This is a feature request dressed up as a PR.

I have found that what I like about clozes is how they make creating multiple cards from a single sentence very easy.
What I don't like is how it hides the word I would like to focus on.

So I added a template filter where instead of hiding the word, it highlights it.
cloze-show1
cloze-show2
cloze-show3

Long ago in old anki I had written an addon to do exactly this: https://ankiweb.net/shared/info/1224266113. But with how things are now it seems if I want this functionality it needs to be in anki proper.

I am very open to changing the template tag name `cloze-show` to something more fitting, but nothing has come to mind.

(I have been out of the anki-game for many years and am coming back and trying to get feature parity with how I used to do things)

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -173,6 +173,7 @@ jthulhu <https://github.com/jthulhu>
 Escape0707 <tothesong@gmail.com>
 Loudwig <https://github.com/Loudwig>
 Wu Yi-Wei <https://github.com/Ianwu0812>
+jake <https://github.com/jakeprobst>
 
 ********************
 
--- a/rslib/src/cloze.rs
+++ b/rslib/src/cloze.rs
@@ -461,13 +461,10 @@ pub(crate) fn cloze_only_filter<'a>(text: &'a str, context: &RenderContext) -> C
 }
 
 pub(crate) fn cloze_show<'a>(text: &'a str, context: &RenderContext) -> Cow<'a, str> {
-    strip_html_inside_mathjax(
-        reveal_cloze_text(text, context.card_ord + 1, false).as_ref(),
-    )
-    .into_owned()
-    .into()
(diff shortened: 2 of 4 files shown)


dae:
Are you aware this can be achieved with just JS? The HTML output in the review screen includes the original text in a data attribute, and there are existing card templates out there that take advantage of this to change what is shown.

jakeprobst:
From what I could tell only the `[...]` was getting sent over from the backend to frontend. I did initially look online for more frontend-based solutions so I wouldn't have to modifiy anki directly and I could not find any.

After a some reading and digging, I did come up with this:
```
{{cloze:main}}
<script>
function reveal_cloze() {
	let cloze = document.getElementsByClassName("cloze").item(0);
	cloze.innerHTML = cloze.getAttribute('data-cloze');
}
reveal_cloze();
</script>
``` which does work as desired, but this required some real legwork to figure out.

Personally I think that something like a `cloze-show` filter (and accompanying documentaion) would help the discoverability of a feature like this.


But it works for me now either way, and if you don't think this is needed feel free to close this PR.

dae:
I'm a bit torn. Your change is fairly small, and doesn't represent a big maintenance burden. But it's also not really how cloze deletions were intended to be used, and I don't recall anyone requesting it before. That makes me lean towards a "not right now, but maybe in the future"

brishtibheja:
Can you not just look at the back quickly to know what was clozed?

jakeprobst:
That seems like a horrible workflow. The back also contains additional information I would like to avoid seeing on the front.
Doesn't matter though, the js I posted above has been working great. The only real issue is discoverability of such a feature if anyone wanted it.

PR #3194, opened 2024-05-11
Use SplashScreen flag for all aqt tooltips

voczi:
Prevents tooltip from showing when parent is out of focus. Noticeable at startup with the collection sync tooltip, but also with all other tooltips. WindowType reference: https://doc.qt.io/qt-6/qt.html#WindowType-enum

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -173,6 +173,7 @@ jthulhu <https://github.com/jthulhu>
 Escape0707 <tothesong@gmail.com>
 Loudwig <https://github.com/Loudwig>
 Wu Yi-Wei <https://github.com/Ianwu0812>
+Voczi <dev@voczi.com>
 
 ********************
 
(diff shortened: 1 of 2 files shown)


voczi:
Only thing is that the tool tip doesn't follow the parent window if you decide to move it. Not sure how big of an issue that would be.

dae:
Seems reasonable, thanks. Let's give it a try.

abdnh:
This is causing focus to move to the tooltip until hidden. Slightly annoying when adding cards quickly, but not a big deal.

voczi:
> This is causing focus to move to the tooltip until hidden. Slightly annoying when adding cards quickly, but not a big deal.

Eh? The tooltip for an added card is 500ms if I understood the code correctly. Are you adding cards with Power Automate or something?

abdnh:
I just noticed this while adding cards quickly to test a PR.

voczi:
@dae Would really appreciate it if you pinged here with the problem causing the revert. I shouldn't have to manually check the PR.

PR #3197, opened 2024-05-12
Do not show media auto sync errors

### LINKED ISSUE #3130: Media auto sync: "connection closed before message completed⁩"
> Hi,
> i use Anki on MacVentura 13.6.3 Intel.
> Anki Version: Version ⁨24.04 (429bc9e1)⁩
> and i get this error from time to time. (when i “unsleep” the notebook).
> then i click the error message several times and its gone until the next time.
> A network error occurred.
> Error details: ⁨error sending request for url (): connection closed before message completed⁩

Originally reported on https://forums.ankiweb.net/t/a-network-error-occurred-error-details-error-sending-request-for-url-connection-closed-before-message-completed/43294

I'd thought that we might want to show errors to the user when automatically syncing, but perhaps we should just dump them to the console instead. If there's some persistent error, users should end up seeing it when they sync manually or close/open the program.

abdnh:
Closes #3130

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -1071,16 +1071,13 @@ title="{}" {}>{}</button>""".format(
         else:
             after_sync(False)
 
-    def maybe_auto_sync_media(self) -> None:
-        if self.can_auto_sync():
-            return
-        # media_syncer takes care of media syncing preference check
-        self.media_syncer.start(True)
-
     def can_auto_sync(self) -> bool:
+        "True if syncing on startup/shutdown enabled."
+        return self._can_sync_unattended() and self.pm.auto_syncing_enabled()
+
+    def _can_sync_unattended(self) -> bool:
         return (
-            self.pm.auto_syncing_enabled()
-            and bool(self.pm.sync_auth())
+            bool(self.pm.sync_auth())
             and not self.safeMode
             and not self.restoring_backup
         )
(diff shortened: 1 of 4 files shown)


qt/aqt/main.py
```diff
@@ -1075,7 +1075,7 @@ def maybe_auto_sync_media(self) -> None:
         if self.can_auto_sync():
```
abdnh, line 1075: Looks like an inverted check bug?

dae:
Sorry, I think the terminology here may be confusing.

The "autosync" setting is about syncing on program open/close. The issue on #3130 was about errors popping up when the "periodically sync media" option is enabled - the one run by on_autosync_timer(). We want to show errors for the former, but not for the latter.

dae:
...so confusing that I appear to have confused myself. ^_^; I now see that you're only passing in True in _maybe_auto_sync_media. So it looks like this is already ok? It might be nice to rename is_autosync to is_periodic_sync to hopefully avoid confusion in the future.

qt/aqt/main.py
dae: You're right, but it's doubly-wrong - it's looking at the auto-sync setting instead of the periodic sync setting. I'll refactor.

dae:
Thanks Abdo!

PR #3198, opened 2024-05-12
Fix/exclude suspended cards when optimize all presets

L-M-Sherlock:
fix https://forums.ankiweb.net/t/fsrs-parameters-get-worse-when-optimizing-all-presets/44669/4?u=l.m.sherlock


dae:
Thanks Jarrett!

PR #3199, opened 2024-05-15
Fix/ignore revlogs when there is not learn entry after the last forget entry

L-M-Sherlock:
fix https://forums.ankiweb.net/t/fsrs-parameters-get-worse-when-optimizing-all-presets/44669/9?u=l.m.sherlock


dae:
Thanks Jarrett!

PR #3204, opened 2024-05-18
User-led media importation managment

bphd:
More often than not, medias are lacking in collection.media, and those seems to be imported ONLY if the associated card is imported too. Thing is that "always" parameter skips the card like would "if newer" and "never" (to correct). But even without that, sometimes we want to keep our own card, even if newer or older, without taking one from the deck, we just want to re-import the medias because we lost them (more often than not like I've said). I've only found language variable to modify, but not the occurences in the code


dae:
While I appreciate you making an attempt, a PR with only string changes is not really helpful. Since this is a feature request, it would be better posted in the suggestions section of the forum.

PR #3206, opened 2024-05-21
Fix future due search missing cards on last day

### LINKED ISSUE #2952: Future due graph misses cards due on the last day

abdnh:
Closes #2952


dae:
Thanks Abdo. We still have some weirdness with a very small bar at the end that we'll need to look into, but that's not this PR's fault.

PR #3208, opened 2024-05-22
do not recreate io page when toggle button used

krmanik:
> On 24.6 rc1, in the Add window, if you toggle the mask editor (either via the button or Ctrl+Shift+M), all the masks will be deleted.

https://forums.ankiweb.net/t/anki-24-06-release-candidate/44926/40


abdnh:
Good catch! Maybe the issue can be solved with an additional bool @krmanik?

krmanik:
Additional bool can be used but in edit mode when changing notetype will recreate the page, which we do not want.

PR #3209, opened 2024-05-23
Reschedule → Reposition

user1823:
After repositioning a card, Anki used to display "Undo Reschedule", not "Undo Reposition".


dae:
Thanks!

Issue #2469, opened 2023-04-12
Selection range not adjusted when navigating between rows via preview window


dae:
https://forums.ankiweb.net/t/bug-arrow-keys-in-preview-break-shift-click-selection-in-card-browser/29282

RRomreoJr:
Hello, I wanted to know if I could try and fix this issue. I've been looking into it out of curiosity.

PR #3213, opened 2024-05-28
Fixes shift click selection after programmatic selection in most cases, Issue #2469

RRomeroJr:
Proposed fix for issue #2469
Original thread:
https://forums.ankiweb.net/t/bug-arrow-keys-in-preview-break-shift-click-selection-in-card-browser/29282

Some things I noticed while looking into this.

- Changing the selection programmatically followed by shift-click selecting is what causes the issue. 
This would happen when:
    - `<` /  `>` buttons were clicked in the preview window
    - `Ctrl + N` or `Ctrl+ P` table shortcuts were used in the browser window
    - `Home` or `End` table shortcuts were used in the browser window
   
   Then..
   - `Shift + Left Click` selecting in the table view
- Some quick googling showed that this was a common issue with Qt, and I reproduced it in a small mockup program I made. I'm not sure if it is a bug with Qt itself, or if there is some better/ more proper way to select items in a TableView from code.

What I added feels like a work-around more than a proper fix, but, as far as I can tell, it fixes all the mentioned scenarios.

**However**, if the user holds `Shift` while clicking the `<` or `>`  buttons in the preview window the same bug reappears. I haven't been able to figure out why this still happens.

### DIFF
--- a/qt/aqt/browser/table/table.py
+++ b/qt/aqt/browser/table/table.py
@@ -608,6 +608,7 @@ class Table:
                 direction,
                 self.browser.mw.app.keyboardModifiers(),
             )
+        # Setting current like this avoids a bug with shift-click selection
         self._view.setCurrentIndex(index)
         self._view.selectionModel().select(
             index,
--- a/qt/aqt/browser/table/table.py
+++ b/qt/aqt/browser/table/table.py
@@ -608,7 +608,8 @@ class Table:
                 direction,
                 self.browser.mw.app.keyboardModifiers(),
             )
-        self._view.selectionModel().setCurrentIndex(
+        self._view.setCurrentIndex(index)
+        self._view.selectionModel().select(
             index,
             QItemSelectionModel.SelectionFlag.Clear
             | QItemSelectionModel.SelectionFlag.Select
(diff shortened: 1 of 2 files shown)


RRomeroJr:
Sorry about making 2 pull requests. In hindsight I could've fixed the other one.

If this is acceptable, I'd like to ask one question. I'd like to try to make an automated test for this behavior, but I'm not sure were to start looking implement it. If possible, any advice on where to look would be appreciated.

dae:
Thanks for looking into this! Just one minor request: please link to the issue in the comment, to make it a little easier to find the related discussion.

> What I added feels like a work-around more than a proper fix, but, as far as I can tell, it fixes all the mentioned scenarios.

Even if it's not perfect, if it's less bad than what we currently have, that's a win.

RRomeroJr:
Added link up top 👍

qt/aqt/browser/table/table.py
```diff
@@ -608,7 +608,9 @@ def _move_current(
                 direction,
                 self.browser.mw.app.keyboardModifiers(),
             )
-        self._view.selectionModel().setCurrentIndex(
+        # Setting current like this avoids a bug with shift-click selection
```
dae, line 611:
```suggestion
        # Setting current like this avoids a bug with shift-click selection
        # https://github.com/ankitects/anki/issues/2469
```

dae:
Thank you!

PR #3218, opened 2024-05-29
Add a preference to toggle LaTeX generation

abdnh:
I added the option to the Editing section but it doesn't exactly fit there. Any suggestions?


dae:
I moved it to Review, but Editing works too - there's not a strong argument for one over the other.

Thanks for these quick fixes!

PR #3219, opened 2024-05-29
Pass -- to mpv/mplayer before filename

abdnh:


dae:
Thanks Abdo. I'm not sure if string concatenation will work in this case - it certainly doesn't work on non-Windows platforms at least.

PR #3220, opened 2024-05-30
Set 'text/plain' as default type of Flask's Response

abdnh:


dae:
Thanks Abdo! I've gone with a slightly different approach that limits the change to the 404 page, and adds a comment so we don't accidentally revert it in the future.

abdnh:
Paths such as _addons/ will still suffer from the issue though?

dae:
Damn, you're right. I should have looked more carefully before making the change! I'd prefer to be explicit about the return value instead of setting the default, so I might extract this into a helper function - I'll post back here when I've made the change, as a second pair of eyes would be good :-)

dae:
Look ok to you? https://github.com/ankitects/anki/commit/5082d9ae5c6259128af1ddccce7e675f3ef563dc

abdnh:
Looking good. That was the solution I went for first, but decided to set a default instead to be 100% safe.

PR #3222, opened 2024-06-01
Avoid askUser() in sync dialogs

abdnh:
The AnkiHub add-on triggers an AnkiWeb sync before its sync process if a full sync is required and cancels the process if the user didn't resolve the conflict. We got [reports](https://community.ankihub.net/t/anking-step-deck-wont-install-due-to-ankhihub-sync-cancelled/229383) from some Mac users that no confirmation dialog is shown and we suspect the issue has something to due with askUser() being blocking. A few months ago we had a similar issue with StudyDeck not appearing and we solved it by passing a callback to avoid .exec().

### DIFF
--- a/qt/aqt/sync.py
+++ b/qt/aqt/sync.py
@@ -28,7 +28,6 @@ from aqt.qt import (
 )
 from aqt.utils import (
     ask_user_dialog,
-    askUser,
     disable_help_button,
     showText,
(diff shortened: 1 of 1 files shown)


dae:
I'll be happy to see a day when we don't have any more .exec() calls in the code!

PR #3226, opened 2024-06-07
fix sveltekit respecting yarn env var

AlexBocken:
With the replacement of Svelte with SvelteKit I've noticed how anki can no longer build successfully when using an the env var `YARN_BINARY`. 

Investigating a bit, I’ve noticed how in `build/ninja_gen/src/node.rs`, the function in question has `./yarn` hardcoded in as opposed to reading in the env var if appropriate.

This PR reads-in the yarn variable for `SveltekitBuild` in the same manner as it already does for the other functions.


dae:
Thanks. I seem to recall there was a reason why it was changed, but since the build seems to work fine now without ./yarn, I'm not sure what the reason was.

PR #3228, opened 2024-06-08
Fix image occlusion errors in previewer and card template editor

glutanimate:
A quick follow-up fix for IO rendering in the previewer and card layout screens, following the example of 71311f201e8c5eb6b89539b1283f655a5eec7fb8


dae:
Appreciate the help!

PR #3231, opened 2024-06-13
Make SvelteCheck and ViteTest respect `YARN_BINARY`

antecrescent:
This is kind of a follow up to https://github.com/ankitects/anki/pull/3226. It allows me to run the tests fully offline in the Gentoo Linux package.
I removed the `./yarn{,.bat}` files in the build system along with all references to them, since they seem to be no longer needed.

Checks pass locally on Linux and I hope that Windows CI passes, too, since I didn't set up a Windows VM to test this.

### DIFF
--- a/build/ninja_gen/src/node.rs
+++ b/build/ninja_gen/src/node.rs
@@ -214,16 +214,13 @@ pub struct SvelteCheck {
 
 impl BuildAction for SvelteCheck {
     fn command(&self) -> &str {
-        if cfg!(windows) {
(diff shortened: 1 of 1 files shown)


dae:
No objections to the other changes, but ./yarn(.bat) is useful for when the user does not have node installed globally. It allows running things like './yarn dev' without having to manually put out/extracted/node on the path first.

antecrescent:
Sorry for the inactivity, I now implemented your feedback!

dae:
Thanks!

PR #3236, opened 2024-06-18
Fix/Relearning cards' Intervals don't update after changing Desired Retention

### LINKED ISSUE #3234: Intervals don't update after changing Desired Retention
My `desired retention` was previously set at 77 percent which I then update to 90 percent. I don't use `reschedule cards at change`. Now the intervals shown on top of `hard`, `good` and `easy` buttons should change. Note that I'm speaking of the intervals that each grade will assign to the cards and not the already assigned intervals. They are shown to you by default in the review screen.

Here's where I think the anomaly lies — the same intervals shown for one of my relearning cards doesn't change at all. The said card has gone through the default steps `1m 10m` and `FSRS` determines now what happens at the last step. It feels as if the change in desired retention isn't being taken into account.

*( originally reported here in FSRS's repo: https://github.com/open-spaced-repetition/fsrs4anki/issues/657 )*

___

**Addendum:**

The deck for ease of reproducibility —

[漢字 Writing__Level 10-4__4. Level 6-20240617133543.zip](https://github.com/user-attachments/files/15871976/Writing__Level.10-4__4.Level.6-20240617133543.zip)

(rename `.zip` —> `.apkg`)

cid:`1564439362436`

L-M-Sherlock:
fix #3234

### DIFF
--- a/rslib/src/scheduler/states/relearning.rs
+++ b/rslib/src/scheduler/states/relearning.rs
@@ -53,7 +53,17 @@ impl RelearnState {
             }
             .into()
         } else {
-            self.review.into()
+            if let Some(states) = &ctx.fsrs_next_states {
+                let (minimum, maximum) = ctx.min_and_max_review_intervals(1);
+                let interval = states.again.interval;
+                ReviewState {
+                    scheduled_days: ctx.with_review_fuzz(interval as f32, minimum, maximum),
+                    ..self.review
+                }
+                .into()
+            } else {
+                self.review.into()
+            }
         }
     }
 
@@ -77,7 +87,17 @@ impl RelearnState {
             }
             .into()
         } else {
-            self.review.into()
+            if let Some(states) = &ctx.fsrs_next_states {
+                let (minimum, maximum) = ctx.min_and_max_review_intervals(1);
+                let interval = states.hard.interval;
+                ReviewState {
+                    scheduled_days: ctx.with_review_fuzz(interval as f32, minimum, maximum),
+                    ..self.review
+                }
+                .into()
+            } else {
+                self.review.into()
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


user1823:
Though I am not 100% sure, I think that your fix forces the recalculation of the interval every time a relearning card leaves the RELEARN state. If it is so, I think that the following fix is more efficient.

My idea is that Anki should clear the stored value of the interval of relearning cards (like https://github.com/ankitects/anki/commit/d9f5487815e57d420ef248389eea4413c911bc3d) when certain operations (listed below) are performed and then recalculate the interval during review if it is missing (because it was cleared).

The stored value of interval of relearning cards should be cleared when:
- the desired retention is changed
- the card is moved into different deck

I am not sure if we need to treat LEARN cards in the same way.

L-M-Sherlock:
> My idea is that Anki should clear the stored value of the interval of relearning cards

This implementation would be complex.

> Though I am not 100% sure, I think that your fix forces the recalculation of the interval every time a relearning card leaves the RELEARN state.

It's necessary if we update to FSRS-5 because it considers the short-term schedule.

rslib/src/scheduler/states/relearning.rs
```diff
@@ -52,6 +52,14 @@ impl RelearnState {
                 },
             }
             .into()
+        } else if let Some(states) = &ctx.fsrs_next_states {
```
user1823, line 55: Does this work if there are no memory states at that point of time (e.g. the deck was changed)?

dae: The state is recalculated prior to that point.

user1823:
Not 100% sure if this is a bug, but still worth looking into:

A relearning card on AnkiDroid (24.06.2 backend) shows the intervals `20m 30m 6d 7d`


The same card on Anki 24.06.3 rc1 shows the intervals `20m 30m 7d 7d`


The change of `Good` interval from `6d` to `7d` may be caused by something like https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/343 (the reps count has increased by 1 because of the last `Again` rating)

But, should the `Easy` button have the same interval as the `Good` button?

Card Info:


Card: [A.apkg.zip](https://github.com/user-attachments/files/15957108/A.apkg.zip) (rename to remove .zip)

L-M-Sherlock:
Seems I should set the interval as states.easy.interval plus one?

https://github.com/ankitects/anki/blob/0d8d816a0c424b1e033af7d4d1331816bbb61426/rslib/src/scheduler/states/relearning.rs#L135-L149

user1823:
> Seems I should set the interval as states.easy.interval plus one?

Maybe, yes

PR #3238, opened 2024-06-19
Add option to automatically refresh card browser

voczi:
Add option to automatically refresh card browser whenever a new note is added to the collection. Off by default. Discussed here: https://forums.ankiweb.net/t/browser-auto-refresh/45866
Suggestions for potential improvements are appreciated.

### DIFF
--- a/ftl/core/preferences.ftl
+++ b/ftl/core/preferences.ftl
@@ -41,6 +41,8 @@ preferences-theme-follow-system = Follow System
 preferences-theme-light = Light
 preferences-theme-dark = Dark
 preferences-v3-scheduler = V3 scheduler
+preferences-auto-refresh = Automatically refresh browser
+preferences-auto-refresh-tooltip = Refresh the card browser whenever a new note is added. Might cause performance issues.
 preferences-ignore-accents-in-search = Ignore accents in search (slower)
 preferences-backup-explanation =
     Anki periodically backs up your collection. After backups are more than 2 days old,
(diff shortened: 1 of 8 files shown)


dae:
Your code looks fine, but I'm afraid I'm not sure if merging this is a good idea:

- It's yet another option
- It uses a generic 'auto refresh' label, but only applies to the specific case of manually adding a single card. Cue incoming bug reports from users that it doesn't do anything when other actions like deletions are performed.
- When cards are added via the add screen, you can already locate them via the History button
- An existing hook exists to make this trivially done by an add-on

voczi:
> When cards are added via the add screen, you can already locate them via the History button

I know. It's just annoying when you're using window snapping, with the card browser and the note adder on each side, then it doesn't refresh the browser when you add a new note.

> It uses a generic 'auto refresh' label, but only applies to the specific case of manually adding a single card. Cue incoming bug reports from users that it doesn't do anything when other actions like deletions are performed.

I could change the label and/or tooltip. But if anyone comes with a bug report for the feature working as it should, then we could explain it there and then. Maybe we could turn it into a feature request rather than a bug report.

m10d:
> * When cards are added via the add screen, you can already locate them via the History button

I'm not able to figure out what you mean by this - I add a card/note, and looking for anything that will "single click" to refresh my "filter"

I put filter in quotes because while I recognize it's a generic filter under the hood, I'm typically looking at just one deck, so from the user's perspective I'm not really filtering anything.

Also, regarding the `history` I'm hoping you can point me to - ideally I could just sort by "most recently created".  I'm not sure if notes have this, but for everything already tracked, it seems helpful?  Doing this would also be a huge help adding tags for some academic workflows - eg one chapter at a time, linearly through.  It would be less error prone, at least for me, to just set the tag for the last X minutes based on what I was working on, then forget to update it in the `add` window as I move on to the next chapter. 

Thanks!

PR #3241, opened 2024-06-19
Fix remaining accessibility warnings

### LINKED ISSUE #2564: Address accessibility warnings

abdnh:
Closes #2564

### DIFF
--- a/ts/editor/PlainTextBadge.svelte
+++ b/ts/editor/PlainTextBadge.svelte
@@ -5,6 +5,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 <script lang="ts">
     import * as tr from "@generated/ftl";
     import { getPlatformString, registerShortcut } from "@tslib/shortcuts";
+    import { onEnterOrSpace } from "@tslib/keys";
     import { createEventDispatcher, onDestroy } from "svelte";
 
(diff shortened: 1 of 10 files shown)


ts/editor/PlainTextBadge.svelte
```diff
@@ -40,6 +41,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     class:visible={show || !animated}
     class:highlighted={!off}
     on:click|stopPropagation={toggle}
+    on:keydown={onEnterOrSpace(() => toggle())}
```
abdnh, line 44: `on:keydown={undefined}` will work as well for the purpose of fixing the warning.

dae:
Thanks Abdo, it's great to have these down to 0. Will merge this in once the next patch update is out.

PR #3246, opened 2024-06-20
Remove threshold of compute minimum recommended retention

L-M-Sherlock:
Related issue:
- https://github.com/open-spaced-repetition/fsrs-optimizer/issues/116

Related PR:
- https://github.com/open-spaced-repetition/fsrs-optimizer/pull/117


user1823:
Though I appreciate the continuous improvements, I think that we should try to avoid having to change the strings so frequently because it creates extra work for translators and causes some strings to remain untranslated if they were updated close to a release.

One idea is to split the strings so that the translators don't have to retranslate the whole string every time a small part is changed.

dae:
I did ask this in the past :-(

https://github.com/ankitects/anki/pull/3148#discussion_r1574880705

As a general rule, we should try to avoid placing specific values in the language strings, to reduce the chances of this happening again in the future.

@user1823 are you objecting to this being merged as-is?

user1823:
> @user1823 are you objecting to this being merged as-is?

I feel that it would be better if we minimised the effect of such changes on the translators. But, if you are fine with merging it, I am no one to object.

dae:
> I feel that it would be better if we minimised the effect of such changes on the translators.

I agree, but do you have some concrete suggestions for how this PR could be done differently?

user1823:
As far as this string is concerned, the only part that I feel can change in the future is "This tool assumes that you’re starting with 0 learned cards, and will attempt to find the desired retention value that will lead to the most material learnt, in the least amount of time."

If we change this to "This tool will attempt to find the desired retention value that will lead to the most material learnt, in the least amount of time.", we retain the meaning and make the string more resilient.

If, for any reason, we need to re-add a minimum limit in the future, we can then add a new string that says "This feature requires a minimum of {x} reviews." This way, we won't affect the current string.

dae:
> This tool assumes that you’re starting with 0 learned cards

I thought that was still the case - did I miss us changing that? Or are you suggesting we change the string even though we do still assume a start of 0?

user1823:
> Or are you suggesting we change the string even though we do still assume a start of 0?

This one. Telling that we assume a start of 0 is not useful in my opinion and this is something that might change in the future.

L-M-Sherlock:
> This one. Telling that we assume a start of 0 is not useful in my opinion and this is something that might change in the future.

In my plan, it will be considered in a built-in simulator.

dae:
Thank you both!

PR #3253, opened 2024-06-23
Fix image occlusion flicker caused by setupI18n()

### LINKED ISSUE #3239: [24.06.2] Image Occlusion regression (flickering)
Version ⁨24.06.2 (33a92379)⁩
Python 3.9.18 Qt 6.6.2 PyQt 6.6.1
Windows 10, no add-ons.

Also reproducible on 24.06.1. Does not occur on 24.06 rc2.

Each time the Front and/or Back side of an IO card is loaded, the image flickers. The following gifs are recorded on fresh installs of Anki, using the default IO note type. The same problem arises when using any custom note type containing the "anki.imageOcclusion.setup()" script.

On 24.06rc2 (no flickering)
On 24 06rc2

On 24.06.2 (flickering when loading the Front and or Back side of a card)
On 24 06 2

abdnh:
Closes #3239

### DIFF
--- a/ts/routes/image-occlusion/review.ts
+++ b/ts/routes/image-occlusion/review.ts
@@ -44,11 +44,6 @@ interface SetupImageOcclusionOptions {
 
 async function setupImageOcclusion(setupOptions?: SetupImageOcclusionOptions): Promise<void> {
     await waitForImage();
-    await setupI18n({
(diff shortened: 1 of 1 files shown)


dae:
Thanks for the quick fix - appreciated :-)

PR #3255, opened 2024-06-24
A11Y: Improve the accessibility of the Preferences dialog

RastislavKish:
Hello,

I'm a blind developer, who would like to use Anki. :) The app does have some serious a11y problems at the moment, but after having looked through the codebase, I believe it should be possible to do something about most of them and make the app a really pleasant and engaging experience for screenreader users.

I have started implementing the necessary fixes, and would like to upstream as many of them as possible. I think it would be best to address the improvements gradually in several thematic PRs rather than one big single a11y refactor, so they can be easily reviewed and discussed case by case.

This PR enhances the Preferences screen. Specifically, there are two kind of issues addressed:

* Configure buddy widget for labels.
* Move unit labels of spinboxes into spinbox suffix property.

The former creates a semantic link between components such as textboxes, comboboxes etc. and the labels used to describe them, for example the "Language:" label and the language selector. This allows screenreaders to present input widget functionality during keyboard navigation. It's a purely semantic change that doesn't have any visual implications.

The latter changes how Anki displays spinbox units (for example, Network timeout: 60 s). QSpinBox has a dedicated suffix property intended for this purpose, which exposes the unit to screenreaders together with the value. Anki seems to use frequently an alternative approach, placing a separate label next to the spinbox, losing the semantics.

I have removed these unit labels and moved their content into the suffix properties of their respective spinboxes. There is likely a minor visual disproportion caused by the removal of the widgets that will need to be addressed by someone sighted. But primarily I wonder if Anki is okay with this way of representing units from stylistical standpoint, as this is going to be one of the more frequent a11y fixes from what I've seen so far.

Thanks for all the amazing work, really impressively structured codebase, and keeping the project open-source!

## Reproduction steps

1. Launch a screenreader (Orca on Linux, NVDA on Windows (not tested))
2. Launch Anki
3. Open Preferences
4. use Tab and arrow keys to move around the interface. Namely, make sure to cross the language combobox and the spinboxes on the Review tab.

## Before the PR

The screenreader would read "English (United States) ComboBox" instead of "Language: English (United States) ComboBox" and "20 SpinBox" instead of "Learn ahead limit: 20 mins SpinBox", to name some examples.

## After the PR

All input components of the Preferences dialog shall be presented with their respective description labels, SpinBoxes shall present their units.

## Additional notes

Removing the unit labels may have left some gaps in the interface which may need to be fixed stylistically.

## Environment

Tested on Ubuntu Mate 22.04, Orca 44.2 compiled from gnome-44 branch of the [Orca repo](https://gitlab.gnome.org/GNOME/orca).

Another test carried out with Orca main branch (47.alpha at the time of testing) on Ubuntu Mate 24.04. Under this configuration, the value of spinboxes is not read at all by the screenreader, however this seems like a bug in either Orca or QT, not Anki.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -177,6 +177,7 @@ RRomeroJr <117.rromero@gmail.com>
 Xidorn Quan <me@upsuper.org>
 Alexander Bocken <alexander@bocken.org>
 James Elmore <email@jameselmore.org>
+Rastislav Kish <rastislav.kish@protonmail.com>
 
 ********************
 
--- a/qt/aqt/forms/preferences.ui
+++ b/qt/aqt/forms/preferences.ui
@@ -44,6 +44,9 @@
             <property name="text">
              <string>preferences_language</string>
             </property>
+            <property name="buddy">
+             <cstring>lang</cstring>
+            </property>
            </widget>
           </item>
           <item row="0" column="1">
@@ -63,6 +66,9 @@
            <widget class="QLabel" name="video_driver_label">
             <property name="text">
              <string>preferences_video_driver</string>
+            </property>
+            <property name="buddy">
+             <cstring>video_driver</cstring>
             </property>
            </widget>
           </item>
@@ -116,6 +122,9 @@
             <property name="text">
              <string>preferences_style</string>
             </property>
+            <property name="buddy">
+             <cstring>styleComboBox</cstring>
+            </property>
            </widget>
           </item>
           <item row="0" column="0">
@@ -123,12 +132,18 @@
             <property name="text">
              <string>preferences_theme</string>
             </property>
+            <property name="buddy">
+             <cstring>theme</cstring>
+            </property>
            </widget>
           </item>
           <item row="2" column="0">
            <widget class="QLabel" name="uiSizeLabel">
             <property name="text">
              <string>preferences_user_interface_size</string>
+            </property>
+            <property name="buddy">
+             <cstring>uiScale</cstring>
             </property>
            </widget>
           </item>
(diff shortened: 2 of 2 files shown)


dae:
Thanks for taking the time to contribute! None of the regular developers have a strong background in a11y, so the help would be appreciated.

Qt's suffix handling is a little clunky, but may be worth it for the improved accessibility.

One issue I noticed is that the suffix looks strange without a leading space, eg

        self.form.network_timeout.setSuffix(" "+self.form.network_timeout.suffix())

RastislavKish:
Thanks for having a look at this!

One particular problem with QT are its accessibility quirks, for example, on Linux the flat review browsing mode of Orca tends to break a lot in Anki due to QT a11y bugs, meaning I can't as a user access pretty much anything out of the scope of standard keyboard navigation. did not test on Windows, but I would be surprised if situation was very different over there.

Unless I figure out what exactly is causing these anomalies and if there's a way to design the interface so QT would fulfil its a11y duties correctly, having the suffixes assigned seems like the only way to present the units to the user other than requiring them to do trial/error or searching up the specific preference in the documentation.

Padding with spaces sounds like a good idea, I will implement this.

In the long term, I think the styling may not be a big issue. As far as I understand, Anki is undergoing a UI migration to Svelte, which has excellent accessibility properties and shall provide everything necessary.

RastislavKish:
I believe the styling should be in place. One space is used for now, more can be added if necessary.

Currently, the styled spinboxes are listed out explicitly in the code. I believe it should be possible to do this automatically for all spinboxes with a suffix of this dialog, but this would affect even those which their authors may prefer not to style like the uiScale spinbox.

PR #3256, opened 2024-06-25
Fix FSRS easy interval being same as good interval in relearning cards

user1823:
https://github.com/ankitects/anki/pull/3236#issuecomment-2187787774

Before https://github.com/ankitects/anki/pull/3236, the Easy interval for relearning cards was Good interval + 1. This PR restores that behaviour.

~This is intended to be a temporary change. When FSRS will consider short-term reviews (in V5), the Easy interval will automatically become larger than the Good interval.~

### DIFF
--- a/rslib/src/scheduler/states/relearning.rs
+++ b/rslib/src/scheduler/states/relearning.rs
@@ -134,14 +134,16 @@ impl RelearnState {
 
     fn answer_easy(self, ctx: &StateContext) -> ReviewState {
         let scheduled_days = if let Some(states) = &ctx.fsrs_next_states {
-            let (minimum, maximum) = ctx.min_and_max_review_intervals(1);
+            let (mut minimum, maximum) = ctx.min_and_max_review_intervals(1);
+            let good = ctx.with_review_fuzz(states.good.interval as f32, minimum, maximum);
+            minimum = good + 1;
             let interval = states.easy.interval;
             ctx.with_review_fuzz(interval as f32, minimum, maximum)
         } else {
-            self.review.scheduled_days
+            self.review.scheduled_days + 1
         };
         ReviewState {
-            scheduled_days: scheduled_days + 1,
+            scheduled_days,
             elapsed_days: 0,
             memory_state: ctx.fsrs_next_states.as_ref().map(|s| s.easy.memory.into()),
             ..self.review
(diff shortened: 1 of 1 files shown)


user1823:
@L-M-Sherlock, please review the PR.

user1823:
Reason for second commit:
Adding 1 to states.easy.interval won't work because fuzz is applied after that.

L-M-Sherlock (approved):
LGTM

user1823:
The code is now future-proof. In FSRS v5, this code will prevent the fuzz from causing the Easy interval to become equal to or less than the Good interval (when the non-fuzzed intervals are close enough).

L-M-Sherlock (approved):

rslib/src/scheduler/states/relearning.rs
```diff
@@ -134,7 +134,9 @@ impl RelearnState {
 
     fn answer_easy(self, ctx: &StateContext) -> ReviewState {
         let scheduled_days = if let Some(states) = &ctx.fsrs_next_states {
-            let (minimum, maximum) = ctx.min_and_max_review_intervals(1);
+            let (mut minimum, maximum) = ctx.min_and_max_review_intervals(1);
+            let good = ctx.with_review_fuzz(states.good.interval as f32, minimum, maximum);
+            minimum = good + 1;
```
user1823, line 139: I have copied this code from learning.rs but I think that we need the following to deal with the case where `good` = `maximum`.

```rs
maximum = maximum.max(minimum);
```

dae:
Thank you both!

rslib/src/scheduler/states/relearning.rs
```diff
@@ -135,8 +135,11 @@ impl RelearnState {
     fn answer_easy(self, ctx: &StateContext) -> ReviewState {
         let scheduled_days = if let Some(states) = &ctx.fsrs_next_states {
             let (mut minimum, maximum) = ctx.min_and_max_review_intervals(1);
+            // ensure 1 greater than good where possible
             let good = ctx.with_review_fuzz(states.good.interval as f32, minimum, maximum);
-            minimum = good + 1;
+            if minimum < maximum {
```
user1823, line 140: Imo, there was no need to add this extra condition because constrained_fuzz_bounds already deals with this case.

https://github.com/ankitects/anki/blob/60b25535efaf23baa6a8a00e3509d9b43a4cdc06/rslib/src/scheduler/states/fuzz.rs#L79

dae: Thanks, good catch.

PR #3258, opened 2024-06-25
 Fix deck config input's background color regression

abdnh:
Report: https://forums.ankiweb.net/t/anki-24-06-3-rc/46403/2
Cause: https://github.com/ankitects/anki/pull/3243


dae:
Thank you for the quick fix, Abdo!

PR #3265, opened 2024-06-27
Restore $deckOptions

### LINKED ISSUE #3146: Restore ability for add-ons to extend the deck options screen
https://forums.ankiweb.net/t/anki-24-04-2-beta/43562/12

abdnh:
Closes #3146

An ugly solution until we figure out a better API.


abdnh:
Not directly related, but are those setup functions in index.ts files such setupDeckOptions still used?

dae:
Nope - the index.ts entrypoints became obsolete with the move to SvelteKit, so we should remove them when we have a chance.

Thanks for the fix!

PR #3272, opened 2024-06-30
Fix profile manager used before checked for None

Luc-Mcgrady:
If pm is None it will throw here:
https://github.com/ankitects/anki/blob/d678e39350a2d243242a69f4e22f5192b04398f2/qt/aqt/__init__.py#L674-L677
instead of giving a descriptive error message as it should.

[Bugged reddit user](https://www.reddit.com/r/Anki/comments/1dr7r14/traceback_error_aqt_line_556_in_run_i_have_a_mac/)


dae:
Thank you!

PR #3273, opened 2024-07-01
Update base images and introduce health endpoint

jeankhawand:
sync-server: introduce `/health` endpoint to check if the service is reachable.

bump(alpine): bump alpine base image from `3.19` to `3.20`

bump(rust): bump rust-alpine build image from `1.76` to `1.79`


dae:
Thanks!

PR #3275, opened 2024-07-03
Apply fuzz to SM2 lapse interval and respect max ivl

user1823:
Imo, there is no reason for not applying fuzz to SM2 lapse intervals


user1823:
The test failure is not related to my changes, IMO.

dae:
Thanks!

PR #3282, opened 2024-07-07
Fix import loop (circular import)

davidculley:
I wanted to write an unit test but couldn't because I encountered an import loop (meaning: before you can finish importing A you must import B, but you can import B only after you finished importing A, so the **import fails** because it would result in an **infinite loop**).

This pull request fixed that problem.

Here's an **explainer** of the problem and why this pull request is necessary:
https://www.youtube.com/watch?v=UnKa_t-M_kM


dae:
The fact that it doesn't work out of the box is unfortunate. It used to - this broke when pylib switched to an implicit namespace, which means we can't run any code when the base module is imported. A simple workaround is for calling code to import anki.collection first. The other imports should then import correctly.

I'm not sure about this PR. Importing a submodule as a separate symbol is not something our codebase currently typically does, and having to prefix types like NoteId is not ideal.

dae:
A no for now I'm afraid.

PR #3283, opened 2024-07-07
Update type annotation syntax

davidculley:
- Use `something | None` rather than `Optional[something]`
- Use `tuple` rather than `Tuple`, etc.
- Replace old `typing` types with newer `collections.abc` types

This pull request might look huge but it's only ever two or so lines per file that changed. It's only the mentioned three types of changes.


dae:
Thanks!

davidculley:
I'm glad I could help. 🙂

PR #3285, opened 2024-07-08
Fix error if regex can't find the filename

davidculley:
Ensured that the returned match object is not None.


abdnh:
Doesn't this just make the error less descriptive?

davidculley:
Fixed it.

Now it prints an error message more helpful than a non-descriptive `'NoneType' object has no attribute 'group'` if the string didn't match the regular expression. This also makes the type checker happy.

It also no longer uses exceptions to control the flow.

dae:
Is this correcting an error you were seeing in the wild? I would not expect AnkiWeb to be be sending a 200 return code if the header is missing.

davidculley:
No, this is fixing a type error reported by Pyright.

Since I use type checkers and linters quite extensively, I'd like them to be happy with the code so that the critical problems can't go unnoticed.

If the code is overengineered, I'd like to see this instead:

```python
match = re.match(
    "attachment; filename=(.+)", resp.headers["content-disposition"]
)
assert match is not None
fname = match.group(1)
```

dae:
Happy to see typechecking issues fixed. When we expect something never to be None, I feel an assert better conveys that than an if statement, and it adds less noise to the codebase.

dae:
Thanks!

PR #3286, opened 2024-07-08
Fix 'NoneType object is not subscriptable' error

davidculley:
You can't access the first item of a list if you happen to get `None` instead of a list.

This pull request fixes a

```
TypeError: 'NoneType' object is not subscriptable
```

by ensuring the returned value is indeed a list and not `None`.


dae:
Is this an error you were seeing in the wild?

davidculley:
No, my intention was to fix the type errors that Pyright reports.

Since I use type checkers and linters quite extensively, I'd like them to be happy with the code so that the critical problems can't go unnoticed.

After some more research, it seems that when you download an add-on from `https://ankiweb.net/shared/download/{number}`, you are redirected to `'https://ankiweb.net/svc/shared/download-addon/{number}`. And that latter URL contains `t=`, `minpt=`, `maxpt=` and `bidx=` between the `?` and `#`.

I now realize the safeguard I implemented is useful only if any of these four keys `t`, `minpt`, `maxpt` or `bidx` does not return a value. I can simplify the code if it's overengineered.

PR #3287, opened 2024-07-08
Simplify method to get addon directories

davidculley:
- simplify logic to get only directories
- use Paths rather than strings for path operations
- use list comprehensions
- simplify check if environment variable is set
- reverse list in-place rather than creating a copy of it


dae:
I'm afraid I think https://github.com/ankitects/anki/pull/3311#issuecomment-2241499283 applies here.

PR #3288, opened 2024-07-08
Update Tooltip

brishtibheja:
updated the tooltips as described here: https://forums.ankiweb.net/t/update-tooltips/46849


dae:
Thank you for contributing!

Because the meaning has changed and we want the translators to update the strings too, we need to leave the existing ones alone, and add new ones with a -2 suffix, then update the code to point to the new strings. Please also update the CONTRIBUTORS file so that automatic tests pass.

brishtibheja:
> add new ones with a -2 suffix

I did but one of them already had a -2 suffix. I left it as it is. I added the previous strings to the end of the file. If you need to change something, please do it.

Expertium:
Continuing the conversation I had with Sorata in Discord, `deck-config-historical-retention-tooltip` should be changed. It makes it sound like review history is **permanently deleted** when "Ignore reviews before" is used, which is not true. It's just omitted.

PR #3290, opened 2024-07-08
Improve exception handling

davidculley:
Replaced `except:` with `except Exception:`.

The bare `except:` has many unintentional effects and is almost never what you want.

See this explainer video: https://youtu.be/zrVfY9SuO64

Also, instead of using `except:` as a fallback for what wasn't caught by `except Exception:`, specify explicitly the exceptions that would be caught by that clause, as there are only three options left if an exception is not an `Exception`, according to the [hierarchy of exceptions](https://docs.python.org/3/library/exceptions.html#exception-hierarchy).


davidculley:
Can these two exception clauses be consolidated?

https://github.com/ankitects/anki/blob/486cc8a2abba7858e0e8b69e9768d3be0a47a4dd/qt/aqt/profiles.py#L282

https://github.com/ankitects/anki/blob/486cc8a2abba7858e0e8b69e9768d3be0a47a4dd/qt/aqt/profiles.py#L288

qt/aqt/profiles.py
```diff
@@ -285,7 +285,7 @@ def rename(self, name: str) -> None:
                 showWarning(tr.profiles_anki_could_not_rename_your_profile())
             else:
                 raise
-        except:
+        except (SystemExit, KeyboardInterrupt, GeneratorExit):
```
dae, line 288:
```suggestion
        except BaseException:
```

PR #3291, opened 2024-07-09
Update `apt-get` to `apt`

davidculley:
`apt-get` is more modern than `apt`.


dae:
This adds extra noise to the output, as now we get 'WARNING: apt does not have a stable CLI interface. Use with caution in scripts.'

Issue #2629, opened 2023-08-29
Something (probably Svelte 4 upgrade) has broken exported components


dae:
https://forums.ankiweb.net/t/anki-2-1-66-beta/32463/27

Perhaps caused by 'svelte/internal' no longer being publicly exported in Svelte 4. We'll need to investigate whether the current approach is even feasible anymore. Even if we're able to fix this, I think we're going to need a less fragile approach moving forward.

dae:
Let's focus on a short term fix for now. I think this broke when svelte/internal was changed to svelte in export-runtime.ts. We should start by updating the example add-on to svelte4, and switching it to the new import location to see if that fixes things.

abdnh:
This seems to work as a short term solution:
```diff
diff --git a/ts/sveltelib/export-runtime.ts b/ts/sveltelib/export-runtime.ts
index 0744db3d9..5f9de74b3 100644
--- a/ts/sveltelib/export-runtime.ts
+++ b/ts/sveltelib/export-runtime.ts
@@ -7,7 +7,15 @@

 import { registerPackageRaw } from "@tslib/runtime-require";
 import * as svelteRuntime from "svelte";
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import * as svelteInternal from "svelte/internal";
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import * as svelteDiscloseVersion from "svelte/internal/disclose-version";
 import * as svelteStore from "svelte/store";

 registerPackageRaw("svelte", svelteRuntime);
 registerPackageRaw("svelte/store", svelteStore);
+registerPackageRaw("svelte/internal", svelteInternal);
+registerPackageRaw("svelte/internal/disclose-version", svelteDiscloseVersion);
```

I tested the example add-on with this patch and Svelte 4.

dae:
@abdnh thanks!

@luc-languagetools after updating your add-on to Svelte 4, can you confirm running from the latest Anki HEAD fixes your problem?

dae:
Assuming this is fixed until we hear otherwise.

luc-languagetools:
I actually took this opportunity to move away from svelte components. I can iterate on new features faster if it's python only, because of regression testing (pyqt-test + unit tests). I'll probably get back into svelte at some point later for some other features, but will probably be for standalone components (similar to what Anki does with Rust + Svelte).

[...]

Issue #3280, opened 2024-07-06
Can't upgrade past Vite 5.0


dae:
Vite 5.1+ appears to fail when we use './ninja sveltekit' - it chokes on references to items like 'document'. Not sure at this point whether it's going to require adjustments to our code, or whether it's an issue on Vite's/Svelte's end. Presumably it relates to the separate web and SSR builds that Vite does, and the fact that we reference window/document at the top level of some files.

PR #3292, opened 2024-07-10
Update to Svelte 5

abdnh:
## Issues

- `svelte/internal` was removed, so exported components are [broken again](https://github.com/ankitects/anki/issues/2629).
- New warnings


dae:
From a brief play:

- deck options wasn't loading; I've pushed a fix
- hovering over a graph spams this to the console:
Uncaught TypeError: Cannot read properties of undefined (reading 'html')

It appears to be related to https://svelte-5-preview.vercel.app/docs/breaking-changes#components-are-no-longer-classes and getOrCreateTooltip(). And relatedly, there are a few cases in the code where we're calling .$set() which we'll presumably need to update to use stores instead.

dae:
Unfortunately this failed due to https://forums.ankiweb.net/t/is-it-possible-to-unpin-the-dprint-package/37065. It also broke builds in other PRs, because for some reason yarn didn't revert the binary back to an old version afterwards. Was the upgrade strictly necessary, or was it to e.g. maintain consistency with a local dprint you have installed on your machine? We can consider bumping the minimum Linux version if it's a blocker.

dae:
On second thought, I think we've waited long enough for the glibc upgrade. I'll push a commit to main today that updates us to the latest version.

dae:
Sorry, screwed up the merge by accidentally rebasing first. Fixing...

ts/editor/rich-text-input/RichTextInput.svelte
```diff
@@ -22,7 +21,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         inputHandler: InputHandlerAPI;
         /** The API exposed by the editable component */
         editable: ContentEditableAPI;
-        customStyles: Promise<CustomStyles>;
+        customStyles: Promise<Record<string, any>>;
```
abdnh, line 24: Couldn't figure out the right way to type components in Svelte 5.

ts/editable/mathjax-element.ts
```diff
@@ -5,6 +5,7 @@ import { on } from "@tslib/events";
 
 import { placeCaretAfter, placeCaretBefore } from "$lib/domlib/place-caret";
 
+import { createClassComponent } from "svelte/legacy";
```
abdnh, line 8: I used createClassComponent because $state() doesn't work here for some reason (tried adding a .svelte suffix to the filename)

ts/editor/rich-text-input/RichTextInput.svelte
dae: No worries, I would have done the same. :-)

dae:
Are you aware of any remaining blocking issues apart from the internal Svelte exports? I'd like to get this in before building a new beta, if it's nearly there.

abdnh:
Not aware of any other blockers.

abdnh:
Regarding #3280, it turned out the references in routes/tmp are the source of the error so I disabled them for now.

dae:
This seems to have roughly halved the Svelte 4 build time, which is is a great win. :tada:

PR #3294, opened 2024-07-10
Refactor `_addButton` method of editor

davidculley:
- Condensed and simplified the various if-else-construct.
- Disentangled the logic of a particularly confusing if-else-construct.
- Transformed variables to snake_case (see PEP8) and gave them clearer names.
- Fixed an import warning (`import os`).
- Fixed some f-strings.


qt/aqt/editor.py
```diff
@@ -287,7 +288,7 @@ def _addButton(
         self,
         icon: str | None,
         cmd: str,
-        tip: str = "",
+        title_attribute: str = "",
```
dae, line 291: This is part of the public API, so changing it could break add-ons.

davidculley: Right, I didn't take add-ons into account. I checked only the code base of Anki itself.

Lesson learned, won't happen again. 🙂

dae:
Thanks!

davidculley:
I'm glad I could help!

PR #3295, opened 2024-07-11
Simplify `lastCard` method in reviewer

davidculley:
[Disentangled](https://en.wikipedia.org/wiki/De_Morgan's_laws) the logic and made the code more readable.


dae:
Another case of https://github.com/ankitects/anki/pull/3311#issuecomment-2241499283 I'm afraid. This code will end up getting reimplemented in Svelte at one point, so is probably not worth touching at the moment.

PR #3296, opened 2024-07-11
Fix improper imports

davidculley:
Fix a bunch of import warnings that boil down to using star imports (`from something import *`).

See https://docs.astral.sh/ruff/rules/undefined-local-with-import-star-usage/

### DIFF
--- a/pylib/anki/cards.py
+++ b/pylib/anki/cards.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import pprint
 import time
+from typing import NewType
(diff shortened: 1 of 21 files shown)


dae:
Changes look good, but this has taken conflicts I'm afraid.

davidculley:
Resolved the conflict.

dae:
(sorry, I meant to strip the co-author annotation)

davidculley:
It's alright.

PR #3297, opened 2024-07-11
Add `@deprecated` to deprecated methods

davidculley:
Show proper deprecation information rather than commenting the code.


davidculley:
The check for this pull request never finished. There's neither a green checkmark nor a red x. What's up with that?

dae:
@glutanimate you wrote 'softly' there - do you approve of these changes? https://github.com/ankitects/anki/pull/2969/commits/e598464665c89b69013a5b82a49add82fe9df741

glutanimate:
@dae I was hesitant to explicitly mark it as deprecated because `addonFromModule()` calls are so ubiquitous in add-ons, and `AddonManager` itself still uses it (plus we haven't deprecated any other `AddonManager` methods, yet).

I would honestly favor just consolidating the implementations, so that `addonFromModule` calls `addon_from_module`, and so that any internal calls in `AddonManager` use `addon_from_module`.

Any time we deprecate an add-on API, it adds a maintenance cost to all add-on consumers as they have to either break compatibility with previous versions or add shims to conditionally consume either API. IMHO in this case the maintenance benefit on Anki's side does not seem that large – we would continue to be inconsistent on PEP8 and instance vs static methods. Since the APIs here are largely add-on focused, it seems justifiable to keep `addonFromModule` around without a deprecation notice. I think some similar examples to this also exist in other modules where pure PEP8 refactors are only softly deprecated (e.g. in `collection`, `models`) – the pattern there is to add `@deprecated`, but comment it out.

Generally speaking, I think it would be nice to refactor `AddonManager` at some point to make it less reliant on a running Anki session (i.e. reduce or eliminate dependencies on `mw` and `pm` instances), as that would have tangible benefits for testing add-ons. We could then also introduce a full PEP8 API while continuing to support `camelCase` equivalents. At that point adding `deprecated` messages might make more sense, both because there would now be a more concrete benefit for add-on authors to switch to the new APIs and because they could update all consumed methods at once.

PR #3298, opened 2024-07-12
Feat/FSRS-5

### LINKED ISSUE #3270: Good interval of relearning card sometimes increases after pressing Again
After https://github.com/ankitects/anki/pull/3236, the interval of a relearning card is recalculated when it leaves the relearning stage. The fuzz applied depends upon the value of `reps`, which in turn depends on the number of times the card was rated `Again` or `Hard` during the relearning phase.

Let me use an example to explain the problem better.

- For testing purpose, I pressed `Again` on a review card with an 11 month interval.
- When the card was shown again (as a relearning card), the `Good` interval was shown as 1.1 month.
- I pressed `Again` again and then the `Good` interval became 1.2 month. This increase in interval on pressing `Again` is not expected.
- I pressed `Again` again and then the `Good` interval became 1 month.
- I pressed `Again` again and then the `Good` interval became 1.3 month. 

Because the value of `reps` is increasing continuously, the fuzz is changing, leading to unexpected results.

@L-M-Sherlock, please have a look. I know that this issue will be solved in FSRS v5 but we need a stopgap measure till then.

One idea is that we should store the value of reps in a new variable when the card enters the relearning stage and use this variable to calculate fuzz instead of using the actual value of reps.

Theoretically, the same problem can occur with learning cards too. But that requires w[0] to be sufficiently large, which makes it rare.

L-M-Sherlock:
TODO:

- [x] refactor [`get_optimal_retention_parameters`](https://github.com/ankitects/anki/blob/9d8782c31ca87b1de1a3d7e60d5bc4c2590479a1/rslib/src/scheduler/fsrs/retention.rs#L77-L226) to adapt the `SimulatorConfig` of FSRS-5 [Python Code](https://github.com/open-spaced-repetition/fsrs-optimizer/blob/89ac5cdee4845216d04a491c53d71820d3b644cf/src/fsrs_optimizer/fsrs_optimizer.py#L572-L654)
- [x] upgrade user's old parameters of FSRS-4.5 to new parameters of FSRS-5

fix #3270

### DIFF
--- a/.deny.toml
+++ b/.deny.toml
@@ -23,6 +23,7 @@ allow = [
   "CC0-1.0",
   "Unlicense",
   "Zlib",
+  "Unicode-3.0",
 ]
 confidence-threshold = 0.8
 # eg { allow = ["Zlib"], name = "adler32", version = "*" },
--- a/.buildkite/linux/entrypoint
+++ b/.buildkite/linux/entrypoint
@@ -22,7 +22,7 @@ echo "--- Ensure libs importable"
 SKIP_RUN=1 ./run
 
 echo "--- Check Rust libs"
-cargo install cargo-deny --version 0.14.12
+cargo install cargo-deny --version 0.14.24
 cargo deny check
 
 echo "--- Cleanup"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1901,9 +1901,9 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "1.1.0"
+version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ad4417a8209cde62c3b39619ce6f9757f1ce930dd2cc52234bd05657eed5de82"
+checksum = "7d10da26a5610c8626cb6ce767b354eb74ff9c7853405953a6cb6da7f619632a"
 dependencies = [
  "burn",
  "itertools 0.12.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1901,9 +1901,7 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "0.6.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "70cec685337af48789e58cea6ef59ee9f01289d1083428b03fe14e76b98c817c"
+version = "1.0.0"
 dependencies = [
  "burn",
  "itertools 0.12.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1901,9 +1901,9 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "1.1.1"
+version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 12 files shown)


asukaminato0721:
[get_optimal_retention_parameters](https://github.com/ankitects/anki/blob/9d8782c31ca87b1de1a3d7e60d5bc4c2590479a1/rslib/src/scheduler/fsrs/retention.rs#L77)  ？

L-M-Sherlock:
@dae, I have some problems with the CI. Could you help me?

.deny.toml
```diff
@@ -23,6 +23,7 @@ allow = [
   "CC0-1.0",
   "Unlicense",
   "Zlib",
+  "Unicode-3.0",
```
L-M-Sherlock, line 26: @dae, I added this line to solve the problem. Is it acceptable?

L-M-Sherlock: Oops, I get a new error:


error[custom]: unknown term
--
  | ┌─ /state/build/lin-ci/ankitects/anki-ci/.deny.toml:26:4
  | │
  | 26 │   "Unicode-3.0",
  | │    ^^^^^^^^^^^
  |  
  | 2024-07-15 12:39:03 [ERROR] failed to deserialize config from '/state/build/lin-ci/ankitects/anki-ci/.deny.toml'

asukaminato0721: try

```
[Unicode License v3](https://spdx.org/licenses/Unicode-3.0.html)
```

ref https://github.com/EmbarkStudios/cargo-deny/blob/main/deny.template.toml#L89C7-L89C33

and https://spdx.org/licenses/

asukaminato0721: Confusing, <https://github.com/EmbarkStudios/cargo-deny/blob/6344cc566621410a0865632b4ef0e82a20408676/deny.template.toml#L89C7-L89C33> said it's spdx license, but `"Unicode-3.0"` is spdx format.

asukaminato0721: oh, this may be the problem. <https://github.com/EmbarkStudios/spdx/pull/65/files#diff-94815ebc8dd748656c1a4414adabdbbc8eeefcf0e1b1fc069223bbdf8bf08974R1843 it's new added.>

asukaminato0721: well, 
the 3.0 available from spdx 0.10.4
cargo deny uses "0.10", so I guess here is the problem.

<https://github.com/EmbarkStudios/cargo-deny/blob/6344cc566621410a0865632b4ef0e82a20408676/Cargo.toml#L103>

asukaminato0721: I guess install the latest cargo-deny could fix this. Which is now `0.14.24`.

asukaminato0721: lol

L-M-Sherlock:
@dae, this PR is ready for review.

VenroyDEV:
hyped for fsrs5

.deny.toml
dae: Yes, all good now - thank you both.

dae:
Sorry for the delay. Looks good to me - thank you for all your work on this!

PR #3299, opened 2024-07-12
Fix unused parent tags getting cleared

abdnh:
#3174 didn't fix the issue.


dae:
Sorry, didn't mean to include myself in the commit.

PR #3301, opened 2024-07-12
refactor: use type hint for `Callable[[], None]`

davidculley:
Beginners might not know what this construct means.

A type alias helps them understand the code and self-documents the code. The more people understand the code, the more can contribute.


glutanimate:
Sorry, I appreciate the sentiment, but I think we should expect people to understand basic language features. This change would add a lot of noise to the code and conversely actually make things harder to read for contributors familiar with Python types.

davidculley:
There we are of different opinions. 🙂 The types `int` or `list[str]` are definitely basic features of the languages, but the meaning of `Callable[[], None]` is not immediately obvious and definitely on another level than `list[str]`.

It also requires typing the characters `[` and `]` four times in total every time you want to use it, and these characters are not easy to type, at least not on non-American keyboards.

You also have to be careful to write exactly the same numbers of closing brackets as opening brackets and not make a mistake.

Whereas it's very easy to simply begin writing `Function` and then press Tab for auto-complete.

As for the noise, it adds exactly one additional line to a file (if you don't count (invisible) whitespace changes or other style changes).

But to merge it or not is not my decision to make. I made the suggestion, if it won't get merged, it's fine by me too.

glutanimate:
Just to qualify my comment a bit: I agree with you that `Callable[[], None]` is not the most intuitive syntax out there, and I think that this is one case where e.g. typescript types are much easier to understand at a glance. However, for better or for worse, it is the standard way that callables of this form are annotated in typed Python code, and I don't think Anki should go its own way here.

Consider also that this might encourage add-on authors to import those types from Anki, introducing more coupling.

dae:
Thanks for weighing in @glutanimate. I'm afraid I agree with his thoughts here, @davidculley.

davidculley:
I understand. 👍

PR #3305, opened 2024-07-15
Remove check for small IO rects

### LINKED ISSUE #3267: Bug: Anki Deletes most/all of my IO clozes when modifying occlusion positions
From the video, it looks like the grouping operation may be responsible.

Originally reported on https://forums.ankiweb.net/t/bug-anki-deletes-most-all-of-my-io-clozes-when-modifying-occlusion-positions/46524

abdnh:
Closes #3267

@krmanik Regarding your comment
> The issue is that on clicking on canvas it add empty width/height shapes, so to exclude it, we need to check the size, the size can be decreased also.

I actually can't reproduce this. No shapes are added if I click on the canvas. Maybe something has changed since then?

### DIFF
--- a/ts/routes/image-occlusion/shapes/to-cloze.ts
+++ b/ts/routes/image-occlusion/shapes/to-cloze.ts
@@ -106,9 +106,8 @@ export function baseShapesFromFabric(): ShapeOrShapes[] {
             const parent = selectionContainingMultipleObjects?.contains(object)
                 ? selectionContainingMultipleObjects
                 : undefined;
-            // shapes with width or height less than 5 are not valid
             // if shape is Rect and fill is transparent, skip it
-            if (object.width! < 5 || object.height! < 5 || object.fill == "transparent") {
+            if (object.fill == "transparent") {
                 return null;
             }
             return fabricObjectToBaseShapeOrShapes(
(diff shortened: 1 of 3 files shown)


krmanik:
Sorry for late reply, recently got little bit busy.

Okay, Then it is good to remove, can you search in vscode for the `object.width! < 5 || object.height! < 5` it may exist in other file also?

dae:
I'm afraid this reintroduces the problem that those checks were presumably originally added to prevent. If you open a new I/O with this PR, and click a few times, an empty shape is created each time. If you switch to the select tool, the empty shapes can be seen. Perhaps we should either a) change the threshold instead of removing it, or b) leave things as they are for now (maybe not many users are affected? the shapes have to be pretty small)

Screenshot 2024-07-21 at 3 08 07 PM

krmanik:
I agree
Sorry dae, I didn’t get time to test before commenting.

dae:
Not a problem! I almost missed the issue myself, as they're not immediately obvious.

abdnh:
> Perhaps we should either a) change the threshold instead of removing it, or b) leave things as they are for now (maybe not many users are affected? the shapes have to be pretty small)

OK. I vote for leaving things as they are. That sounds reasonable.

PR #3306, opened 2024-07-15
replace showWarning with show_warning

tatsumoto-ren:
When there's no internet connection, `showWarning()` causes Anki to get stuck when an add-on is performing a long operation using `QueryOp().with_progress().run_in_background()`. I replaced `showWarning()` with `show_warning()` to avoid the block.


btrkeks:
This seems to fix the deadlock I experienced

PR #3307, opened 2024-07-16
Refactor `ephemeral_card` of notes

davidculley:
Don't use Boolean expressions to implement default values.


dae:
(please see https://github.com/ankitects/anki/pull/3311#issuecomment-2241499283)

PR #3308, opened 2024-07-16
refactor: use list comprehension

davidculley:


dae:
Approving this one as you've already done the work and it's trivial, but it's another case of https://github.com/ankitects/anki/pull/3311#issuecomment-2241499283

PR #3310, opened 2024-07-17
Create `notes_were_updated` hook

Aleks-Ya:
Add gen hook `notes_were_updated` which is invoked when a user updates notes from the UI or **by using an add-on**.

### Use case
I am developing the ["Note Size"](https://ankiweb.net/shared/info/1188705668) add-on which displays a "Size" column in the Browser. I have to cache note sizes; otherwise, the Browser will be slow when the user searches for many notes.
The cache entries need to be updated when individual notes are updated: manually from the UI or programmatically by add-ons (e.g. generating sounds by TTS, loading images from the Internet, compressing images by add-ons, etc.).
Without this hook, users claim that they need to close the Browser to see the updated note size.

Using existing hooks can't achieve the same result.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -181,6 +181,7 @@ James Elmore <email@jameselmore.org>
 Ian Samir Yep Manzano <https://github.com/isym444>
 David Culley <6276049+davidculley@users.noreply.github.com>
 Rastislav Kish <rastislav.kish@protonmail.com>
+Aleksey Yablokov <alex_ya@mailbox.org>
 
 ********************
 
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -494,9 +494,11 @@ class Collection(DeprecatedNamesMixin):
         self, notes: Sequence[Note], skip_undo_entry: bool = False
     ) -> OpChanges:
         """Save note changes to database."""
-        return self._backend.update_notes(
+        changes: OpChanges = self._backend.update_notes(
             notes=[n._to_backend_note() for n in notes], skip_undo_entry=skip_undo_entry
         )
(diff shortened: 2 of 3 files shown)


dae:
I'm afraid I don't think this is a good idea, because it's not going to work in all cases. This hook would only apply when the update_note() method is called, so it's not going to work when notes are changed by bulk actions in the browse screen for example, syncing, etc.

Aleks-Ya:
@dae , thanks for reviewing

File `collection.py` has 2 methods for updating notes:

2. `update_notes` (bulk action). It invokes the new hook `notes_were_updated`.
1. `update_note`. It just invokes `update_notes` with single note and therefore also invokes `notes_were_updated`.

The hook is invoked in both cases, isn't?

Please, correct me if I'm missing something

dae:
The hook would only fire for Python code that calls that routine. So col.update_note(note) would trigger it, but other calls like col.find_and_replace() would not.

Aleks-Ya:
@dae , I see, but what options do we have? I didn't find any hooks in the backend (`rslib` folder). So, is adding a Python hook the only option? I can also invoke the `notes_were_updated` hook for `find_and_replace`, `update_image_occlusion_note` and other functions you tell. Can it be the solution?

dae:
Sorry, I don't think it's practical to invoke that hook for every operation that modifies notes, and it still wouldn't help with things like syncing, that happen without the Python layer. This is why I think adding the hook is a bad idea, because users will expect it to work in all circumstances.

Aleks-Ya:
Looks like the target behavior (react when other addons update notes) can be achieved using gen hook `note_will_flush`.
Closing this PR.

PR #3314, opened 2024-07-21
Ignore reviews before -> Ignore cards reviewed before.ftl

Expertium:
First of all, this should make it more clear what this feature does. As it is right now, it's misleading.
Second of all, I added an extra sentence to the tooltip. I recently found a user who encountered an edge case - if ALL cards have been reviewed before the selected date, and NO NEW cards will be introduced, then the optimizer will always have zero data to work with. I added a clarification about that.
And also, I think "reviewed" is better than "introduced", as "introduced" may make the user think that this is about the creation date of the card.


Expertium:
@user1823 this is related to my recent pull request in the fsrs4anki repo, where I added a similar warning to the FSRS tutorial

user1823:
You didn't respond to the concern that I expressed there. Let's continue the discussion in that PR.

ftl/core/deck-config.ftl
```diff
@@ -411,8 +411,9 @@ deck-config-reschedule-cards-warning =
     Use this option sparingly, as it will add a review entry to each of your cards, and
     increase the size of your collection.
 deck-config-ignore-before-tooltip-2 = 
-    If set, cards introduced before the provided date will be ignored when optimizing FSRS parameters.
+    If set, cards reviewed before the provided date will be ignored when optimizing FSRS parameters.
     This can be useful if you imported someone else's scheduling data, or have changed the way you use the answer buttons.
+    If all of your cards have been reviewed before the selected date and you do not plan to add new cards, it's better to use Reset.
```
dae, line 416:
```suggestion
    If all of your cards have been reviewed before the selected date and you do not plan to add new cards, it's better to reset the cards to new.
```

How about that?

Expertium: Me and user1823 are still discussing what's the best course of action in this case (all cards reviewed before the selected date, no new cards will be added), but yeah, I don't mind this correction.

user1823 (changes requested):
I don't recommend merging this until we figure out what advise we want to give in such cases. See https://github.com/open-spaced-repetition/fsrs4anki/pull/671

brishtibheja:
@dae This feature isn't easily explainable to the users. I think unless this is reworked to made more intuitive this should be in the Advanced section. Or at least at the end of the FSRS section.

dae:
No objections from me to moving it to the advanced section.

Expertium:
No objections from me either.

Expertium:
I removed that line with the advice. LMSherlock decided to add a new feature to the Helper add-on that will allow users to replace all instances of "Hard" with "Again". As for users who did not misuse "Hard" AND still wish to use "Ignore reviews before" for whatever reason AND have reviewed all cards before the selected date AND do not plan to introduce new cards, LMSherlock believes that this is a very rare edge case, and we shouldn't be concerned about it.
So I suggest renaming this feature to "Ignore cards reviewed before" and moving it under Advanced.

brishtibheja:
The number of people who will want to use 'Ignore reviews before' will certainly decrease from that. More reason not to have it in FSRS category. But we will want to still document that (in the tutorial and in the Anki Manual).

@dae Can you move this feature to Advanced?

brishtibheja:
Is this PR ready to merge @Expertium?

Expertium:
Yes, but I'm not sure if it will work, since moving "Ignore cards reviewed before" under Advanced is in a different PR

brishtibheja:
Oh, so that's why you commented there. I think it should work because in the code we use `title: tr.deckConfigIgnoreBefore()` and not the string itself.

dae:
@Expertium could you please update CONTRIBUTORS again? Happy to merge this in once the tests pass.

Expertium:
I have exactly 0 clue how checks work and why changing a string somehow causes them to fail.
By "update CONTRIBUTORS" do you mean put my name there? Ok.

dae:
Thanks!

PR #3320, opened 2024-07-25
Rename review limit

brishtibheja:
I changed `Maximum reviews/day` —> `Maximum cards/day` and `New cards ignore review limit` —> `New cards ignore maximum limit`.

As for tooltips, I think we should remove the text under 'deck-config-limit-interday-bound-by-reviews'. Instead, intraday learning cards should be mentioned as an exception to maximum card limit. I am not sure where do I put this or should I delete the aforementioned part.


ftl/core/deck-config.ftl
```diff
@@ -38,7 +38,7 @@ deck-config-tab-description =
     - `Preset`: The limit is shared with all decks using this preset.
     - `This deck`: The limit is specific to this deck.
     - `Today only`: Make a temporary change to this deck's limit.
-deck-config-new-cards-ignore-review-limit = New cards ignore review limit
+deck-config-new-cards-ignore-review-limit = New cards ignore maximum limit
```
dae, line 41:
```suggestion
deck-config-new-cards-ignore-review-limit = New cards ignore maximum
```

No strong feelings here, but what do you think about this instead?

brishtibheja: The app will be used by people with varied levels of product knowledge and English ability. I think 'limit' is clearer.

PR #3321, opened 2024-07-25
chore(mypy): update Python version assumed by mypy

davidculley:
Since we use the `|` operator, we use at least Python 3.10, but mypy (and probably some other things too) still assumes we use Python 3.9.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -1,5 +1,5 @@
 [mypy]
-python_version = 3.9
+python_version = 3.10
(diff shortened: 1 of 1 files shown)


davidculley:
Never mind, I [found out](https://github.com/ankitects/anki/pull/3323#issuecomment-2250694636) that we do still use Python 3.9.18 (and the `|` operator works only because of `from __future__ import annotations` at the top of each file).

PR #3323, opened 2024-07-25
Update type annotations to use `|` operator

davidculley:
Updated `Optional` and `Union` type annotations to the `|` operator.

### DIFF
--- a/pylib/anki/importing/anki2.py
+++ b/pylib/anki/importing/anki2.py
@@ -31,7 +31,7 @@ class MediaMapInvalid(Exception):
 
 class Anki2Importer(Importer):
     needMapper = False
-    deckPrefix: Optional[str] = None
+    deckPrefix: str | None = None
     allowUpdate = True
     src: Collection
     dst: Collection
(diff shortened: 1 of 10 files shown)


davidculley:
This pull request fails because the `|` operator requires Python 3.10 or newer, but the Python version installed (using Pyenv) is 3.9.18.

I think we need to update these two files:

- https://github.com/ankitects/anki/blob/main/build/ninja_gen/src/python.rs
- https://github.com/ankitects/anki/blob/main/qt/bundle/pyoxidizer.bzl

The Python releases can be found [here](https://github.com/indygreg/python-build-standalone/releases).

1. Do we want to update Python from 3.9 to something newer?
2. If yes, can anyone do this for me please?

dae:
While you are welcome to use CI for testing if you wish, you may find ./check on your local machine gives you a faster feedback cycle. I find it useful to put it into a git hook, so it happens automatically on push, and aborts the push if the check fails.

davidculley:
I'm completely unfamiliar with the toolchain Anki uses and with Rust, but I'm slowly figuring things out. 🙂 Got the PR to pass! Thank you for your help!

Such a git hook sounds very useful. (I already use [`pre-commit`](https://github.com/pre-commit/pre-commit).) Do you mind sharing the script?

dae:
I have this at the top of pre-push:

```
set -e
if [ "$SKIP_CI" != "1" ]; then
    ./check
fi
exit
```

It's not perfect - it runs when I do things like try to push a tag deletion as well. But works well for the common case.

PR #3326, opened 2024-07-26
Updated tooltips

brishtibheja:
https://github.com/ankitects/anki/pull/3320 Did the tooltips changes. @dae can you change the appropriate code?
I also reused deck-config-limit-interday-bound-by-reviews for the intraday learning card exception.

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -15,12 +15,12 @@ deck-config-title = Deck Options
 ## Daily limits section
 
 deck-config-daily-limits = Daily Limits
-deck-config-new-limit-tooltip =
+deck-config-new-limit-tooltip-2 =
     The maximum number of new cards to introduce in a day, if new cards are available.
     Because new material will increase your short-term review workload, this should typically
-    be at least 10x smaller than your review limit.
-deck-config-review-limit-tooltip =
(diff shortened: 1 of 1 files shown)


dae:
We change the key name when it's important that translators update their translations too. Is that the case here? I thought this was more about making the English wording clearer without fundamentally altering the meaning.

ftl/core/deck-config.ftl
```diff
@@ -18,31 +18,30 @@ deck-config-daily-limits = Daily Limits
...
     The selected deck's limits control the total cards that will be shown.
 deck-config-limit-new-bound-by-reviews =
     The review limit affects the new limit. For example, if your review limit is
     set to 200, and you have 190 reviews waiting, a maximum of 10 new cards will
     be introduced. If your review limit has been reached, no new cards will be
     shown.
 deck-config-limit-interday-bound-by-reviews =
-    The review limit also affects interday learning cards. When applying the limit,
-    interday learning cards are fetched first, then reviews.
+    The maximum limit doesn't affect (re)learning cards with less than 1d interval.
```
dae, line 35:
```suggestion
    The limit does not apply to intraday (re)learning cards.
```

or

```suggestion
    The limit does not apply to (re)learning cards due on that day.
```

Issue #3151, opened 2024-04-19
Add maximize buttons to Note Types editor


dae:
> Some things do not fit, so the window needs to be resized, but there is no maximize button. Examples of what does not fit:
> Tools  → Manage Note Types:
> horizontally and vertically
> Fields
> vertically
> Options
> the default texts fit

Originally reported on https://forums.ankiweb.net/t/add-maximize-buttons-to-note-types-editor/43803

balavinaithirthan:
What are the files that are related to this issue?

dae:
models.py and models.ui

cdonat2:
@dae is this issue still relevant? If so, i could try to take a look if thats ok.

dae:
It's still open, but I won't be able to guide you through it I'm afraid.

cdonat2:
Thanks for letting me know. If there is any other issue that has a higher
priority I would be happy to contribute. In any case I will check out this
issue in detail as well.

Is there by any chance a person to contact that could recommend an issue as
a starting point for contributing?

dae:
If you have questions, you're welcome to ask on forums.ankiweb.net, and someone may be able to answer you.

cdonat2:
Thanks will do.

ChazWinterYT:
Close this issue if it has been addressed, please.

PR #3328, opened 2024-07-27
Add maximize hint for models.py (#3151)

cdonat2:
Ran ./ninja check successfully. This Pull Request adds maximize and minimize hints to Note Types editor also when running on Windows, but as this is my first Pull Request please take a look and if there is anything that can be improved or if this Pull Request is not required at all please let me know.


dae:
Thanks Christian!

PR #3329, opened 2024-07-28
Update to FSRS-rs v1.1.4 

L-M-Sherlock:
Related PR:
- https://github.com/open-spaced-repetition/fsrs-rs/pull/203
- https://github.com/open-spaced-repetition/fsrs-rs/pull/204
- https://github.com/open-spaced-repetition/fsrs-rs/pull/205

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,4 +1,5 @@
 [settings]
+py_version=39
 known_first_party=anki,aqt,tests
 profile=black
 extend_skip=qt/bundle
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,0 +1,2 @@
+target-version = "py39"
+extend-exclude = ["qt/bundle"]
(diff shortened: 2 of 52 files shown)


dae:
Thanks as always Jarrett! I'll try to get to this tomorrow.

Regarding the upgrade to Rust 1.80, in the future, I'd prefer FSRS maintain a 'minimum Rust version' policy that allows us to go a few Rust versions back. There are some times when a Rust upgrade will break things (for example, dropping support for older OS versions), and if FSRS makes those changes before Anki is ready to, we won't be able to update to it.

L-M-Sherlock:
OK. Should I revert the version for FSRS-rs for this PR?

dae:
It's probably fine this time, just a comment for the future.

dae:
Sorry for the delay!

voczi:
Just a heads up: https://github.com/ankidroid/Anki-Android-Backend will need a similar toolchain update when the dependency is updated there (to rust ver 1.80)

PR #3330, opened 2024-07-28
Update about.py

asukaminato0721:
one of the collaborators of https://github.com/open-spaced-repetition/fsrs-rs


dae:
Thank you for your contributions!

PR #3344, opened 2024-08-05
style: fix line endings

davidculley:
- Added missing newline character to a file's last line (if necessary) to make the files POSIX-compliant.
- (Removed superfluous empty lines at the file ending.)


dae:
- If we are to make these changes, we should also update our checks to catch these problems, or they will creep back in
- We shouldn't be changing auto-generated files like licenses.json or .gitmodules

davidculley:
I could add a git hook (executed by [pre-commit](https://pre-commit.com/)) that can then be run by the same script that already runs mypy, pylint, etc. if you want me to.

(That's what I already use on my machine to automatically fix these types of issues.)

Contributors wouldn't have to use `pre-commit` locally (but they can if they want, of course) but the CI job would tell them when they forgot a newline at the line-ending.

dae:
The check needs to work with ./ninja check, and across all platforms. The simplest cross platform solution is probably to add another check to the minilints program.

davidculley:
I'm unsure where to put that as I'm unfamiliar with Anki's toolchain.

A quick search for why the CI job fails seems to lead to the function [`strip_line_ending`](https://github.com/ankitects/anki/blob/872ecf481ab33b179726284b754d46ca95deda67/rslib/src/import_export/text/csv/metadata.rs#L544), which seems to want to strip the line endings I'm proposing to add.

As I'm unfamiliar with Rust too, I might need help here. No pressure though, I only opened this pull request because I thought it would be a trivial one.

dae:
If our SQL formatting is not adding a trailing newline, sql_format.ts can be adjusted.

dae:
Please re-open if you'd like to continue work on this.

PR #3345, opened 2024-08-06
Added name to contributors list as well as addressed issue #3343

### LINKED ISSUE #3343: Unused functions in ts/lib/tslib/dom.ts incorrectly use HTMLParamElement
In ts/lib/tslib/dom.ts:

https://github.com/ankitects/anki/blob/bab83ffce02129426028e25a848d98ad1f1f5626/ts/lib/tslib/dom.ts#L145-L148

These two functions contain an obvious mistake: `HTMLParamElement` should be `HTMLParagraphElement`. These functions are only used by the two exported functions below:

https://github.com/ankitects/anki/blob/bab83ffce02129426028e25a848d98ad1f1f5626/ts/lib/tslib/dom.ts#L151-L152
which do not seem to be used anywhere, so these functions can likely be removed.

DillBal:
Ran a ninja check and the build compiled successfully.

Closes #3343

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -184,6 +184,7 @@ Rastislav Kish <rastislav.kish@protonmail.com>
 Expertium <https://github.com/Expertium>
 Christian Donat <https://github.com/cdonat2>
 Asuka Minato <https://asukaminato.eu.org>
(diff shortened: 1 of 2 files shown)


dae:
Thank you!

PR #3346, opened 2024-08-07
Add option for toggling update checks

voczi:
See: https://forums.ankiweb.net/t/how-to-disable-update-check-on-windows/47401/12

### DIFF
--- a/ftl/core/preferences.ftl
+++ b/ftl/core/preferences.ftl
@@ -41,6 +41,7 @@ preferences-theme-follow-system = Follow System
 preferences-theme-light = Light
 preferences-theme-dark = Dark
 preferences-v3-scheduler = V3 scheduler
+preferences-check-for-updates = Check for program updates
 preferences-ignore-accents-in-search = Ignore accents in search (slower)
 preferences-backup-explanation =
     Anki periodically backs up your collection. After backups are more than 2 days old,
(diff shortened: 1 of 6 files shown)


voczi:
on a second thought, maybe it would be more appropriate to put this setting in the network tab

ftl/core/preferences.ftl
```diff
@@ -41,6 +41,7 @@ preferences-theme-follow-system = Follow System
 preferences-theme-light = Light
 preferences-theme-dark = Dark
 preferences-v3-scheduler = V3 scheduler
+preferences-enable-update-check = Enable checking for program updates
```
dae, line 44:
```suggestion
preferences-enable-update-check = Check for program updates
```

qt/aqt/main.py
```diff
@@ -1431,7 +1431,8 @@ def show_menubar(self) -> None:
     def setup_auto_update(self, _log: list[DownloadLogEntry]) -> None:
         from aqt.update import check_for_update
 
-        check_for_update()
+        if aqt.mw.pm.enable_update_check():
```
dae, line 1434: `update_check_enabled` or `check_for_updates` for the function / key name would be a bit clearer/more consistent

dae (commented):
'Appearance' is probably a little better than the 'Syncing' tab - either under general or distractions, perhaps?

voczi:
Should be all good now.

dae:
Thank you!

Issue #3317, opened 2024-07-22
Enable strict_optional in mypy config


dae:
Our mypy configuration uses `no_strict_optional = true`. This tells mypy to assume nullable variables will not be none, so code like the following will pass a check, and fail at runtime instead:

```
foo: int | None = None
foo + 5
```

If we adjust the mypy config, we can get it to throw errors instead:

```diff
diff --git a/.mypy.ini b/.mypy.ini
index ef32ed7a2..1207b6aa3 100644
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -22,6 +22,7 @@ exclude = (qt/bundle/PyOxidizer|pylib/anki/_vendor)
 
 [mypy-anki.*]
 disallow_untyped_defs = True
+no_strict_optional = false
 [mypy-anki.importing.*]
 disallow_untyped_defs = False
 [mypy-anki.exporting]
```

There's quite a few in pylib that need correcting, e.g.

```
pylib/anki/importing/anki2.py:316: error: Item "None" of "Optional[DBProxy]" has no attribute "execute"  [union-attr]
pylib/anki/importing/csvfile.py:114: error: Value of type "Union[str, list[str], None]" is not indexable  [index]
pylib/anki/importing/csvfile.py:116: error: Value of type "Union[str, list[str], None]" is not indexable  [index]
pylib/anki/importing/csvfile.py:118: error: Value of type "Union[str, list[str], None]" is not indexable  [index]
pylib/anki/importing/csvfile.py:122: error: Argument 1 to "reader" has incompatible type "Union[str, list[str], None]"; expected "Iterable[str]"  [arg-type]
pylib/anki/importing/apkg.py:62: error: Item "None" of "Optional[ZipFile]" has no attribute "read"  [union-attr]
Found 181 errors in 24 files (checked 336 source files)
```

The first error is caused by doing 'col.db.execute()' when col.db is DBProxy | None. Adding an `assert col.db` will tell mypy that we are confident db will not be null in that case, and the error will go away. In other cases where we're not 100% sure if the value will be valid, the routine should check for null / exit early / etc.

The work on this can be done incrementally, as you can enable no_strict_optional for individual submodules at a time (e.g. pylib.importing).

[...]

PR #3355, opened 2024-08-13
mypy: enable option `strict_optional` for 'pylib' (#3317)

davidculley:
Fixed all errors within the directory `pylib` by simply adding assertions. These can be replaced later on by better solutions. But we can now set the option `strict_optional` to `true` for the entire directory `pylib`. 

Why this option should be `true`: https://mypy.readthedocs.io/en/stable/config_file.html#confval-strict_optional

### DIFF
--- a/.buildkite/linux/docker/Dockerfile.amd64
+++ b/.buildkite/linux/docker/Dockerfile.amd64
@@ -1,4 +1,4 @@
-FROM debian:10-slim
+FROM debian:11-slim
 
 ARG DEBIAN_FRONTEND="noninteractive"
 
@@ -52,13 +52,6 @@ RUN mkdir -p /etc/buildkite-agent/hooks && chown -R user /etc/buildkite-agent
 
 COPY buildkite.cfg /etc/buildkite-agent/buildkite-agent.cfg
 COPY environment /etc/buildkite-agent/hooks/environment
-
-# Available in Debian 11 as ninja-build, but we're building with Debian 10
-RUN curl -LO https://github.com/ninja-build/ninja/releases/download/v1.11.1/ninja-linux.zip \
-    && unzip ninja-linux.zip \
-    && chmod +x ninja \
-    && mv ninja /usr/bin \
-    && rm ninja-linux.zip
 
 RUN mkdir /state/rust && chown user /state/rust
 
--- a/.dprint.json
+++ b/.dprint.json
@@ -31,14 +31,15 @@
         "target",
         ".mypy_cache",
         "extra",
-        "ts/.svelte-kit"
+        "ts/.svelte-kit",
+        "ts/vite.config.ts.timestamp*"
     ],
     "plugins": [
-        "https://plugins.dprint.dev/typescript-0.85.1.wasm",
-        "https://plugins.dprint.dev/json-0.17.4.wasm",
-        "https://plugins.dprint.dev/markdown-0.15.3.wasm",
-        "https://plugins.dprint.dev/toml-0.5.4.wasm",
-        "https://plugins.dprint.dev/prettier-0.13.0.json@dc5d12b7c1bf1a4683eff317c2c87350e75a5a3dfcc127f3d5628931bfb534b1",
-        "https://plugins.dprint.dev/disrupted/css-0.2.2.wasm"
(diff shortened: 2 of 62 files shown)


dae:
```diff
diff --git a/pylib/anki/models.py b/pylib/anki/models.py
index 0d7024933..d3dd5c886 100644
--- a/pylib/anki/models.py
+++ b/pylib/anki/models.py
@@ -134,9 +134,7 @@ class ModelManager(DeprecatedNamesMixin):
 
     def current(self, for_deck: bool = True) -> NotetypeDict:
         "Get current model. In new code, prefer col.defaults_for_adding()"
-        model_id = self.col.decks.current().get("mid")
-        assert model_id is not None
-        notetype = self.get(model_id)
+        notetype = self.get(self.col.decks.current().get("mid"))
         if not for_deck or not notetype:
             notetype = self.get(self.col.conf["curModel"])
         if notetype:
@@ -154,7 +152,7 @@ class ModelManager(DeprecatedNamesMixin):
         except NotFoundError:
             return None
 
-    def get(self, id: NotetypeId) -> NotetypeDict | None:
+    def get(self, id: NotetypeId | None) -> NotetypeDict | None:
         "Get model with ID, or None."
         # deal with various legacy input types
         if id is None:
diff --git a/pylib/anki/tags.py b/pylib/anki/tags.py
index af1f2c6f9..ce6335991 100644
--- a/pylib/anki/tags.py
+++ b/pylib/anki/tags.py
@@ -121,10 +121,9 @@ class TagManager(DeprecatedNamesMixin):
     def rem_from_str(self, deltags: str, tags: str) -> str:
         "Delete tags if they exist."
 
-        def wildcard(pat: str, repl: str) -> Match:
+        def wildcard(pat: str, repl: str) -> Match | None:
             pat = re.escape(pat).replace("\\*", ".*")
             return_value = re.match(f"^{pat}$", repl, re.IGNORECASE)
-            assert return_value is not None
             return return_value
 
         current_tags = self.split(tags)
```

davidculley:
@dae Thanks a lot for the super helpful fix. 🙂

dae:
Making the code generator insert them would be tricky. I suggest

```diff
diff --git a/.mypy.ini b/.mypy.ini
index 96dae8669..f15a768c2 100644
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -26,6 +26,8 @@ disallow_untyped_defs = True
 strict_optional = true
 [mypy-aqt.*]
 strict_optional = false
+[mypy-_aqt.*]
+strict_optional = false
 [mypy-anki.importing.*]
 disallow_untyped_defs = False
 [mypy-anki.exporting]
```

dae:
Please note that Anki currently fails to start, but you've added an assert in a location where the value can be none. We'll need to do a quick pass over each of the asserts that have been added to double-check the values should never be None, as there may be other instances where the types were not correctly indicating that the value was nullable.

davidculley:
> I suggest

Good idea. Thanks for helping me out. 👍

> We'll need to do a quick pass over each of the asserts that have been added to double-check the values should never be None

Absolutely. I have no idea if any of the assertions I added are actually correct. I don't understand the codebase well enough to know that. My plan was to first make all mypy errors disappear, then to make the pipeline succeed if necessary, and then go over all assertions I added and see if they're actually correct. If any assertion is wrong, replace it with a real solution and repeat the step of making the pipeline succeed. Kind of like an "assertion-driven development". 😅

So with your help, I got the pipeline to succeed now. 🙂 Now we can look which assertions need to be replaced.

davidculley:
To make the code changes easier to review for you, I moved some changes into their separate pull request (#3364).

The assertions become easier to review if you first merge #3364. 🙂

dae:
I'd appreciate it if you could make an attempt to review these yourself first - Anki wouldn't even start until the latest commit I pushed, and when I started reviewing the changes, I quickly encountered changes that don't appear to be correct, such as

```
    def get_image_for_occlusion(self, path: str | None) -> GetImageForOcclusionResponse:
        assert path is not None
```

dae:
Do you plan to return to this? I can keep it around if you're currently busy, as it would be a nice change.

PR #3356, opened 2024-08-13
Rename “Change Deck” to "Select Deck" (#3266)

### LINKED ISSUE #3266: Rename “Change Deck” to "Select Deck"
Currently, the undo action for moving a card to a different deck, and selecting a different deck, use the same "Change Deck" string. To better distinguish between the two, we could add a new string for Op::SetCurrentDeck, such as "Select Deck" instead.

Originally reported on https://forums.ankiweb.net/t/rename-change-deck-to-move-cards/46286/4

bpnguyen107:
Closes #3266


dae:
Thanks Ben!

PR #3358, opened 2024-08-14
Add hook to browser for refresh on new card

taylorobyen:
This change connects a new hook in the Browser class to automatically refresh upon the creation of a new card. I often have the card browser open when adding cards to my decks and always found it strange I had to reclick the deck in the browser to see my new additions. So I thought this would be a nice QoL improvement.

Since the `add_cards_did_add_note` hook returns an extra value I added a helper method of `on_new_note` to the Browser class to call the `search` method. The `search` method seemed to be the best way to update the contents of the browser, but if there is a better way I am happy to hook into a different method. I look forward to your feedback.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -188,7 +188,7 @@ Asuka Minato <https://asukaminato.eu.org>
 Dillon Baldwin <https://github.com/DillBal>
 Voczi <https://github.com/voczi>
 Ben Nguyen <105088397+bpnguyen107@users.noreply.github.com>  
-Taylor Obyen <https://github.com/taylorobyen>
+Taylor Obyen <https://github.com/taylorobyen> 
 ********************
 
 The text of the 3 clause BSD license follows:
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -186,6 +186,7 @@ Christian Donat <https://github.com/cdonat2>
 Asuka Minato <https://asukaminato.eu.org>
 Dillon Baldwin <https://github.com/DillBal>
 Voczi <https://github.com/voczi>
+Taylor Obyen <https://github.com/taylorobyen>
 ********************
 
 The text of the 3 clause BSD license follows:
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -3,31 +3,28 @@
 For info on contributing things other than code, such as translations, decks
 and add-ons, please see https://docs.ankiweb.net/contrib
 
-With most users now on 2.1, the past 2 years have been focused on paying down some
-of the technical debt that Anki's codebase has built up over the years, and making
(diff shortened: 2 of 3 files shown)


voczi:
https://github.com/ankitects/anki/pull/3238#issuecomment-2183969016

dae:
I'm afraid I'm not fond of this change, for the reasons mentioned in the link @voczi helpfully provided. But thank you for taking the time to contribute!

taylorobyen:
@dae I read through your discussion with @voczi and it seemed the first couple points were in relation to having this be an optional feature. I agree with you that this shouldn't be another option, it should just be included by default. I believe a forum post to discuss this issue further may be in order to gauge how the community feels about changing this behavior. As for the points in the previous PR that relate to this I address below:

- When cards are added via the add screen, you can already locate them via the History button
The History window having this feature implemented points more towards that the Browser window should also behave this way to have consistent behavior when adding cards. Even the front page with your decks listed automatically refresh when adding cards to the deck. Naturally, it would improve consistency across the application by having the Browser window auto refresh as well. 

- An existing hook exists to make this trivially done by an add-on
It technically could be done via an add-on, but that doesn't mean we shouldn't improve the Anki out of the box behavior.

It also seemed like in your response to @voczi that you were concerned with inconsistent refresh behavior if on card additions it refreshed, but on card deletes it doesn't. I completely agree with this and think it would be beneficial to refresh consistently. I wasn't aware there was a way to delete cards outside of the browser window. If you provide me a high level of what actions should refresh the Browser I can look into adding additional hooks to ensure consistency. 

Thank you for getting back to me so quickly!

dae:
- Searching on each card add is potentially slow, and could slow down the adding process.
- Refreshing the list on deletion is not desirable IMHO, as again it is slow, and it makes it harder to notice when cards have (accidentally) been deleted.

taylorobyen:
@dae Is there a better way to refresh instead of doing a search each time? Since we know what the user's current filter is perhaps, we can just load the card into the table without needing to do a complete search which should be significantly more efficient. If you point me in the right direction I should be able to revise to meet that requirement.

As for deletes we would definitely want to avoid refreshing when deleted from the Browser window, so the users have the opportunity to rollback. An undo option in the future might be good for this as well.

dae:
The sorting and filtering is done at the database level, so I'm afraid it's not practical to know whether a row should be inserted, and where it should be inserted.

taylorobyen:
Addressed performance concerns by blocking database queries when searching to refresh on new card additions. This instead relies on the cached data to retrieve the cards.

dae:
I'm sorry, I don't think this addresses the concerns I raised above, as it still runs the search again. I apologise if my earlier comments were not clear and have resulted in wasted work - this is not a change I would like to make.

PR #3362, opened 2024-08-16
Right click context menu on images not useful

### LINKED ISSUE #3337: Right click context menu on images not useful
> The Right Click > Copy / Right Click > Paste and Ctrl+C / Ctrl+X in the card editor only works when an image is selected by dragging (appears blue), not when it is selected by clicking as if it will be resized, which is misleading as right clicking still gives the full Copy/Cut/Paste options.

Originally reported on https://forums.ankiweb.net/t/in-editor-copying-cutting-image-currently-unintuitive/47684

bpnguyen107:
Closes #3337 

- I added a "Copy image" action to the editor context menu that shows up when right-clicking on an image. This allows the user to choose what to do when the image is selected together with text if they want to copy the image or copy all the selected content. I think this also matches the behavior of the context menus in browsers like Edge or Chrome.
- I also added hiding the "Cut" and "Copy" actions in the context menu if nothing is selected so the user isn't misled that cutting the image does anything without selecting it first as the forum post said.

Video:
https://github.com/user-attachments/assets/d466f97f-50e5-4ab7-b620-317cd854414d

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -186,7 +186,7 @@ Christian Donat <https://github.com/cdonat2>
 Asuka Minato <https://asukaminato.eu.org>
 Dillon Baldwin <https://github.com/DillBal>
 Voczi <https://github.com/voczi>
-Ben Nguyen <105088397+bpnguyen107@users.noreply.github.com> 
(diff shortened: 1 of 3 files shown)


dae:
Looks great - thanks Ben!

PR #3364, opened 2024-08-17
python: add missing type annotations for None values

davidculley:
Make the use of the type hint `Optional` compliant with the 2018 update of [PEP 484](https://peps.python.org/pep-0484/). (Make the use of `Optional` explicit rather than merely `implicit`.) Otherwise `pyright` won't recognize `Optional` types.

---

Quoting from [PEP 484](https://peps.python.org/pep-0484/):

> A past version of this PEP allowed type checkers to assume an optional type when the default value is `None`, as in this code:

```python
def handle_employee(e: Employee = None): ...
```

> This would have been treated as equivalent to:

```python
def handle_employee(e: Optional[Employee] = None) -> None: ...
```

> This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.

### DIFF
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -153,7 +153,7 @@ class Reviewer:
         self._answeredIds: list[CardId] = []
         self._recordedAudio: str | None = None
         self.typeCorrect: str | None = None  # web init happens before this is set
-        self.state: Literal["question", "answer", "transition", None] | None = None
+        self.state: Literal["question", "answer", "transition"] | None = None
         self._refresh_needed: RefreshNeeded | None = None
         self._v3: V3CardInfo | None = None
         self._state_mutation_key = str(random.randint(0, 2**64 - 1))
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -152,8 +152,8 @@ class Reviewer:
         self.previous_card: Card | None = None
         self._answeredIds: list[CardId] = []
         self._recordedAudio: str | None = None
-        self.typeCorrect: str = None  # web init happens before this is set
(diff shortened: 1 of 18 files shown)


qt/aqt/reviewer.py
```diff
@@ -152,8 +152,8 @@ def __init__(self, mw: AnkiQt) -> None:
         self.previous_card: Card | None = None
         self._answeredIds: list[CardId] = []
         self._recordedAudio: str | None = None
-        self.typeCorrect: str = None  # web init happens before this is set
-        self.state: Literal["question", "answer", "transition", None] = None
+        self.typeCorrect: str | None = None  # web init happens before this is set
+        self.state: Literal["question", "answer", "transition", None] | None = None
```
dae, line 156: This looks redundant. Was it a required change? I'll shift it outside the literal for now, in case it was required to satisfy a type checker, as either works for me.

```suggestion
        self.state: Literal["question", "answer", "transition"] | None = None
```

qt/aqt/switch.py
```diff
@@ -1,5 +1,7 @@
 # Copyright: Ankitects Pty Ltd and contributors
 # License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
+from __future__ import annotations
```
dae, line 3: Very minor: I prefer a newline after the comments, but I'm aware we don't do that everywhere already.

dae:
Thanks, splitting these up is indeed helpful.

qt/aqt/reviewer.py
davidculley: Good catch! My change seems indeed to have been redundant. Thanks for cleaning it up.

PR #3365, opened 2024-08-17
mypy: fix type checking error

davidculley:
To have an easier time in my other pull requests, when figuring out which type errors were introduced by me and which already existed, I'm fixing some of the type errors on the current main branch.

This pull request fixes this mypy error:

> error: Argument 1 to "_answerCard" of "Reviewer" has incompatible type "int"; expected "Literal[1, 2, 3, 4]"  [arg-type]

The function `_answerCard` wants only numbers 1, 2, 3 or 4 (that is, a variable of type `Literal[1, 2, 3, 4]`. But we're passing a variable of type `int` which can hold many more numbers than these mentioned four.

I'm basically just

- casting `int` to `Literal[1, 2, 3, 4]` (which by itself would already suffice)
- adding a check that the given number is indeed one of the four numbers, so that we don't mute mypy only to miss runtime errors when there actually is one because mypy stayed silent since we muted it.

### DIFF
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -596,14 +596,11 @@ class Reviewer:
             for ease in aqt.mw.pm.default_answer_keys:
                 key = aqt.mw.pm.get_answer_key(ease)
                 if key:
-                    if ease in (1, 2, 3, 4):
-                        ease = cast(Literal[1, 2, 3, 4], ease)
-                        answer_card_according_to_pressed_shortkey = functools.partial(
-                            self._answerCard, ease
-                        )
-                        yield (key, answer_card_according_to_pressed_shortkey)
-                    else:
-                        raise ValueError("Invalid value")
+                    ease = cast(Literal[1, 2, 3, 4], ease)
+                    answer_card_according_to_pressed_shortkey = functools.partial(
+                        self._answerCard, ease
+                    )
+                    yield (key, answer_card_according_to_pressed_shortkey)
 
         return [
             ("e", self.mw.onEditCurrent),
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -591,6 +591,20 @@ class Reviewer:
     def _shortcutKeys(
         self,
     ) -> Sequence[tuple[str, Callable] | tuple[Qt.Key, Callable]]:
+
+        def generate_default_answer_keys():
+            for ease in aqt.mw.pm.default_answer_keys:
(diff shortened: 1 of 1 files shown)


qt/aqt/reviewer.py
```diff
@@ -591,6 +591,20 @@ def korean_shortcuts(
     def _shortcutKeys(
         self,
     ) -> Sequence[tuple[str, Callable] | tuple[Qt.Key, Callable]]:
+
+        def generate_default_answer_keys():
+            for ease in aqt.mw.pm.default_answer_keys:
+                key = aqt.mw.pm.get_answer_key(ease)
+                if key:
```
dae, line 598: Walrus operator would make this more concise

qt/aqt/reviewer.py
```diff
@@ -591,6 +591,20 @@ def korean_shortcuts(
...
+                    if ease in (1, 2, 3, 4):
```
dae, line 599: We control default_answer_keys, so there should be no need to check this.

qt/aqt/reviewer.py
davidculley: Should `key` only be not `None` or does it also need to be different from the empty string?

In other words: I would like to change `if key:` to `if key is not None:`, provided this doesn't change its meaning.

qt/aqt/reviewer.py
davidculley: I'll remove the check then. 👍

qt/aqt/reviewer.py
dae: The user can enter an empty string in the preferences, so we also need to guard against that.

dae:
🚀

qt/aqt/reviewer.py
davidculley: Thank you for fixing the code. I wasn't sure about that.

qt/aqt/reviewer.py
```diff
@@ -591,6 +591,18 @@ def korean_shortcuts(
     def _shortcutKeys(
         self,
     ) -> Sequence[tuple[str, Callable] | tuple[Qt.Key, Callable]]:
+
+        def generate_default_answer_keys() -> (
+            Generator[tuple[str, partial], None, None]
+        ):
+            for ease in aqt.mw.pm.default_answer_keys:
+                key = aqt.mw.pm.get_answer_key(ease)
+                if key is None:
```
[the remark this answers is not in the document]
davidculley: Thank you.

PR #3366, opened 2024-08-17
Add distroless Dockerfile and implement internal health check

jeankhawand:
- rslib(http_server): add `is_running()` method
- rslib(sync): introduce `--healthcheck` argument for health probe in distroless
- doc(syncserver): add table comparing Dockerfile and Dockerfile.distroless
- Expand cross-platform support with distroless
- add `Dockerfile.distroless`

- Dockerfile: bump rust `1.79` to `1.80.1`
- Dockerfile: bump alpine `3.20` to `3.20.2`

Note: Implemented an internal health check because distroless images do not include curl, which is used to reduce image size and attack surface. For more details, see https://blog.sixeyed.com/docker-healthchecks-why-not-to-use-curl-or-iwr/ https://github.com/GoogleContainerTools/distroless


rslib/src/sync/http_server/mod.rs
```diff
@@ -219,7 +219,10 @@ impl SimpleServer {
             }
         }
     }
-
+    pub fn is_running() -> bool {
+        let addr = "127.0.0.1:8080";
```
jeankhawand: Should not hardcode host and port.

jeankhawand: will use `let config = envy::prefixed("SYNC_")
            .from_env::<SyncServerConfig>().unwrap();` to grab host and port

jeankhawand (commented):
image
rebuild image with change, now should be good to go!

rslib/src/sync/http_server/mod.rs
jeankhawand: image

docs/syncserver/README.md
```diff
@@ -10,13 +10,27 @@ the build products and runtime dependencies from the rest of your system.
...
+| **Image Size**           | ❌ Larger image size                                       | ✅ Smaller image size                                     |
+| **Customization**        | ✅ Easier to customize with additional packages            | ❌ Limited customization options                          |
+| **Attack Surface**       | ❌ Larger attack surface due to more installed packages    | ✅ Reduced attack surface                                 |
+| **Libraries**            | ✅ More libraries available                                | ❌ Limited libraries                                      |
+| **Start-up Time**        | ❌ Slower start-up time due to larger image size           | ✅ Faster start-up time                                   |
+| **Tool Compatibility**   | ✅ Compatible with more tools and libraries                | ❌ Compatibility limitations with certain tools           |
+| **Maintenance**          | ❌ Higher maintenance due to larger image and dependencies | ✅ Lower maintenance with minimal base image              |
+| **Architecture Support** | amd64, arm64                                               | amd64, arm64, arm, s390x, ppc64le                         |
```
dae, line 25: This table is a great addition for people who aren't sure which one to pick. My only question is: have you tested that arm, s390x and ppc64le work? I would be somewhat surprised if all our dependencies work with them.

rslib/sync/main.rs
```diff
@@ -1,14 +1,28 @@
...
+            return;
+        }
+    }
     if env::var("RUST_LOG").is_err() {
         env::set_var("RUST_LOG", "anki=info")
     }
     set_global_logger(None).unwrap();
     println!("{}", SimpleServer::run());
 }
+fn run_health_check() {
```
dae, line 22: Newline would be nice above this.

dae (commented):
Thanks Jean, this looks like a nice addition. Just a couple of minor things:

docs/syncserver/README.md
jeankhawand: Apologies for missing this earlier! I was experimenting with multi-architecture builds over the weekend and encountered some unexpected issues, especially with the `arm` architecture. While the build is working fine for `amd64` and ` arm64` I ran into the following error when attempting to build:

```bash
Copy code
4.397 error: failed to get `percent-encoding-iri` as a dependency of package `anki v0.0.0 (/app/rslib)`
4.397 
4.397 Caused by:
4.398   failed to load source for dependency `percent-encoding-iri`
```
will remove this line for now.

dae:
Thanks Jean!

dae:
A user has reported that this isn't currently working correctly: https://forums.ankiweb.net/t/healtcheck-endpoint-on-anki-sync-server/52582

alametti:
> A user has reported that this isn't currently working correctly: https://forums.ankiweb.net/t/healtcheck-endpoint-on-anki-sync-server/52582

That would be me, @jeankhawand let me know if any information is missing

jeankhawand:
Hello @alametti 

Thanks for bringing this to my attention. I’ve identified the root cause of the issue. Starting with Docker Engine 26+, localhost resolves to the IPv6 address ::1, reference: https://github.com/docker/for-mac/issues/7269 and https://docs.docker.com/engine/release-notes/26.0/#bug-fixes-and-enhancements-2

As a workaround, you can disable IPv6 using the sysctl setting `net.ipv6.conf.all.disable_ipv6: 1`. This will resolve the unhealthy reporting issue.

I’ll work on raising a PR to address this by explicitly using IPv4 for the health check.

PR #3368, opened 2024-08-18
don't ignore buried cards in future due graph

### LINKED ISSUE #1984: Include buried cards in future due graph
Would need to exclude them from backlog/today, but include them from tomorrow onwards.

jakeprobst:
the future due graph undercounts cards by a notable amount when using "Bury Review Siblings".

before
future_due_current

after
future_due_modified

but more importantly it makes the load balancer look better :leaves: 

Closes #1984

### DIFF
--- a/rslib/src/stats/graphs/future_due.rs
+++ b/rslib/src/stats/graphs/future_due.rs
@@ -6,6 +6,7 @@ use std::collections::HashMap;
 use anki_proto::stats::graphs_response::FutureDue;
 
 use super::GraphsContext;
+use crate::card::CardQueue;
 use crate::scheduler::timing::is_unix_epoch_timestamp;
 
 impl GraphsContext {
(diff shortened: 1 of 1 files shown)


dae:
Thanks Jake!

dae:
Re-reading the issue I linked, should we perhaps be checking for due <= 0 instead?

jakeprobst:
changing it to `due <= 0` does not seem to make a difference in how the chart ends up. Though the number for today in the graph is not entirely accurate as it still misses some cards. For example for today the graph says I have 197 due but I actually have 208 due. Checking `Backlog` does show additional cards that would explain the difference.

To fix the thing you are thinking of I think, this would need to see if `Backlog` is checked and then count negative days as zero.

user1823:
> changing it to due <= 0 does not seem to make a difference in how the chart ends up.

This may be because you don't have any cards that are due in the past and are currently buried.

PR #3378, opened 2024-08-22
Fix field focus lost when pressing alt in the editor

abdnh:
https://forums.ankiweb.net/t/bug-field-loses-focus-when-switching-to-katakana-input-mode/47767

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -58,6 +58,7 @@ class AddCards(QMainWindow):
         gui_hooks.operation_did_execute.append(self.on_operation_did_execute)
         restoreGeom(self, "add")
         gui_hooks.add_cards_did_init(self)
+        self.setMenuBar(None)
         self.show()
 
     def set_note(self, note: Note, deck_id: DeckId | None = None) -> None:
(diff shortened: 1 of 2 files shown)


dae:
Thanks for figuring this out Abdo!

tequa:
@abdnh @dae Hi, so this commit causes some unexpected behavior at least on mac. This behavior specifically does not occur when I remove the "self.setMenuBar(None)" line in `addcards.py` 

https://forums.ankiweb.net/t/bug-report-using-certain-shortcuts-in-the-card-editor-triggers-the-main-window/55152

Steps to reproduce
```
1. Add a card (press A)
2. press “command + comma” to add a bullet point list, it should work just fine. Clear the card
3. Now just click on the main Anki window again to focus on it and bring it to the foreground
4. Now go back to the card editor. Press “command + comma” again
5. Instead of adding a bullet point list, it opens the preferences in the main window
```

abdnh:
The issue was partially fixed in https://github.com/ankitects/anki/commit/71ae5a6b673bc8943f048fed906751febfe585bd recently (not in a stable release yet).

@dae I assume we need to do the same change for editcurrent.py

PR #3380, opened 2024-08-23
.DS_Store file shown as an unused media file

### LINKED ISSUE #3341: .DS_Store file shown as an unused media file
> recently on my iMac when I use the Tools command “check media” the .DS_Store file is shown as an unused file. If I delete it, it is regenerated by MacOS. In my opinion it should not be shown. Is it shown intentionally or is this a bug?

We should perhaps exclude it from the check, like we already do in `files_in_trash()`.

Curiously, when I went to reproduce this, I could not get the `.DS_Store` file to be generated on my machine.

Originally reported on https://forums.ankiweb.net/t/ds-store-file-shown-as-an-unused-media-file/47751

bpnguyen107:
Closes #3341 

I also updated the media_check unit test to cover this

### DIFF
--- a/.buildkite/linux/docker/Dockerfile.amd64
+++ b/.buildkite/linux/docker/Dockerfile.amd64
@@ -1,4 +1,4 @@
-FROM debian:10-slim
+FROM debian:11-slim
 
 ARG DEBIAN_FRONTEND="noninteractive"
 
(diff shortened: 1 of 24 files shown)


dae:
Thanks Ben! The test update was a nice touch.

PR #3382, opened 2024-08-23
Fix Svelte exports not working in Deck Options

abdnh:
#2629 came back recently (after the switch to SvelteKit I guess).

This broke the Exam Notifier add-on: https://github.com/AnKing-VIP/exam-notifier/issues/26


dae:
Thanks Abdo. This may break again once we land #3292, but it's a simple fix for now.

PR #3383, opened 2024-08-24
Create a nice website for the developer's documentation

davidculley:
Added a minimal proof of concept that compiles the `.md` files in `/docs/` into a beautiful website (using MkDocs) and then hosts that website using GitHub Pages.

The website should become available at the following URL after merging this pull request:

https://ankitects.github.io/anki/

Browsing that website should provide a better experience than the `.md` files rendered by GitHub.

I can continue work on this website if it is considered useful.

For a preview of how this could look like, see (for example) the [documentation of FastAPI](https://fastapi.tiangolo.com/learn/).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,28 @@
+name: ci
+on:
+  push:
+    branches:
+      - main
+permissions:
+  contents: write
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Configure Git Credentials
+        run: |
+          git config user.name github-actions[bot]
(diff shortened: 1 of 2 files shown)


brishtibheja:
I personally like mkdocs, but mdbook is already being used for the manual/faqs so that is probably a better option for this website. (I'll also mention that a subdomain of ankiweb.net can be used here, although this one's fine IMO.)

brishtibheja:
I am not a developer really so I cannot tell you about that. I do know there are alternative renderers ([the manual](https://rust-lang.github.io/mdbook)). Perhaps dae will know better about mdbook.

dae:
I'm afraid this feels like adding more maintenance burden for not a lot of gain. The files in docs/ are not large, and it's expected that developers will typically be reading them from a checked out commit.

PR #3386, opened 2024-08-26
Empty cards become undoable

Arthur-Milchior:
If there was a reason for this operation not to be undoable, I can't easily guess it. My main hyposhesis was that the number of deleted card may be too big. But I realized that deleting a deck is undoable and may delete as many note.

As you may know, I realized that only the undoable operations triggered notification in AnkiDroid that we may have to update the UI. And while I just wanted to trigger more notifications, some reviewers thought it would be nicer if the operation were returning a OpChanges. So here it's done. If you would please consider merging it.

I decided to introduce a new string because the closest strings I could find currently are "Empty cards..." and the trailing commas don't seem nice in "undo". And the title, which we may not be able to reuse in all language


Arthur-Milchior:
I guess the failures are:
> Arthur, at the domain Milchior.fr
> ....
> Author arthur, at the domain milchior.fr NOT found in list

I don't understand this failure. I guess it's a question of comparing upper/lower case. Except that in the contribtuors file, I'm listed as `Arthur Milchior <arthur@milchior.fr>`, and I have no idea where the upper case in `Milchior.fr` comes from. I'd appreciate help here to resolve the contributor failure please.

david-allison:
@Arthur-Milchior This should be resolved if you make a whitespace change to `CONTRIBUTORS`

----

* Your original commit to `CONTRIBUTORS` used the uppercase `A` and `M`
  * https://github.com/ankitects/anki/commit/2c0de4948b73bd8cbcb1fb83da57c4f00a8800a4.patch 
* CI checks against the email in the commit to `CONTRIBUTORS`, rather than the text in the file
  * https://github.com/ankitects/anki/blob/a179da382759b9c7f8265faf37efdcbb47d8ee08/tools/minilints/src/main.rs#L151 
* Technically, the local segment of an email is case sensitive, so `arthur@` and `Arthur@` can lead to separate mailboxes

Arthur-Milchior:
I changed the configuartion in git locally. Hope it'll solve it.

david-allison:
It was working previously, the build failure was due to a `rustfmt` error

dae:
Thanks Arthur! I can't recall any reason why this was deliberately omitted - I think it was probably just not a high priority when undo support was originally added to the codebase.

PR #3390, opened 2024-08-28
docs: explain how to make mypy work with Protobuf

davidculley:
Add documentation for how to make `mypy` work with Protobuf without having to build the entire project or installing Rust.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -189,6 +189,10 @@ For Python:
 ./ninja python:sphinx && open out/python/sphinx/html/py-modindex.html
 ```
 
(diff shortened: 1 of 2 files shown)


dae:
I'm afraid I don't want to encourage this David. You recently submitted a PR where Anki wouldn't even start, because you're not testing a full build. That's partly the fault of CI for not catching it, but it's also bad practice to be submitting changes without confirming a proper build locally. Merging this in will also mean another file that either needs to be updated as changes are made to the codebase, or confusion when it inevitably falls out of sync with the codebase and build scripts.

PR #3396, opened 2024-09-01
Change Wayland app_id to net.ankiweb.Anki

ingemarberg:
Anki's icon is not displayed when Anki is running on KDE Plasma 6 on Wayland. Instead of displaying the correct icon, Plasma instead defaults to a generic Wayland icon. This was originally [reported on the forums](https://forums.ankiweb.net/t/missing-icon-on-wayland-for-flatpak/46243) and was initially thought to be a [Flatpak bug](https://github.com/flathub/net.ankiweb.Anki/issues/154). However, after some investigation, I believe that this is related to Plasma instead. Plasma expects the Wayland app_id to be set according to the reverse domain standard, which it currently is not (the current value is `anki.desktop`). I've been able to confirm the same behaviour for both the git main repo and the official Anki 24.06.3 Qt6 build running on Plasma 6.1.4 as well. GNOME seems to be more forgiving with the Wayland app_id, which is probably why this problem only occurs on Plasma.

This PR fixes this by setting the Wayland app_id to `net.ankiweb.Anki` and renaming `anki.desktop`. I've also updated the install and uninstall scripts with the renamed .desktop file, as well as the changed StartupWMClass value. Please see the [KDE Community Wiki](https://community.kde.org/Guidelines_and_HOWTOs/Wayland_Porting_Notes#Application_Icon) for details. However, I'm not sure if this is best way to handle the renamed .desktop file. Would it be a good idea to add a check to the install script in order to delete the old `anki.desktop` file if it exists? Will my small change to the uninstall script be enough?

The following warning would be resolved as well:
`Qt warning: QGuiApplication::setDesktopFileName: the specified desktop file name ends with .desktop. For compatibility reasons, the .desktop suffix will be removed. Please specify a desktop file name without .desktop suffix`

I've tested this on Plasma 6.1.4 running on both X11 and Wayland, and on GNOME 46. The only problem I've ran into is the window decorations when building from source on GNOME. I can see the same behaviour when building from source without making any changes, so perhaps this is to be expected in this case?

Official Anki 24.06.3 Qt6 build
image

Building from source
image

Oh, and one last thing. I've made one [minor contribution](https://github.com/ankitects/anki/pull/2433) before and forgot to add myself to `qt/aqt/about.py`. Would it be appropriate to do so with this PR?

### DIFF
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -667,7 +667,7 @@ def _run(argv: list[str] | None = None, exec: bool = True) -> AnkiApp | None:
 
     # create the app
     QCoreApplication.setApplicationName("Anki")
-    QGuiApplication.setDesktopFileName("anki.desktop")
+    QGuiApplication.setDesktopFileName("net.ankiweb.Anki")
     app = AnkiApp(argv)
     if app.secondInstance():
         # we've signaled the primary instance, so we should close
--- a/qt/aqt/about.py
+++ b/qt/aqt/about.py
@@ -199,6 +199,7 @@ def show(mw: aqt.AnkiQt) -> QDialog:
             "Ian Samir Yep Manzano",
             "Asuka Minato",
             "Eros Cardoso",
+            "Ingemar Berg",
         )
     )
 
--- a/qt/bundle/lin/install.sh
+++ b/qt/bundle/lin/install.sh
@@ -23,13 +23,13 @@ mkdir -p "$PREFIX"/share/applications
 mkdir -p "$PREFIX"/share/man/man1
 cd "$PREFIX"/share/anki && (\
 mv -Z anki.xpm anki.png "$PREFIX"/share/pixmaps/;\
-mv -Z anki.desktop "$PREFIX"/share/applications/;\
+mv -Z net.ankiweb.Anki.desktop "$PREFIX"/share/applications/;\
 mv -Z anki.1 "$PREFIX"/share/man/man1/)
 
 xdg-mime install anki.xml --novendor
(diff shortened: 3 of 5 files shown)


ingemarberg:
I did some more testing on a Fedora 40 VM running KDE Plasma 6.0.3. The official Anki 24.06.3 Qt6 build works as expected and the correct icon is displayed immediately, even without changing the Wayland app_id. I was testing on OpenSUSE Tumbleweed before. Is this still worth doing or should I close the PR?

ingemarberg:
After giving this some more thought, I'm going to mark this as a draft for now. I'll update the title and do some more testing, as well as reverting the StartupWMClass value since it wouldn't be affected by this.

dae:
You're welcome to add yourself to about.py either here or in a separate PR. Thank you for the detailed testing and report here. Have I understood you correctly that the issue doesn't exist in a Plasma 6.0.x install on one distro, but does on a 6.1.x install on a different distro? If so, I wonder whether this will become a problem as people upgrade anyway?

If we do rename the file, our install and uninstall scripts will presumably also need to ensure the old one is removed, as I imagine it might lead to duplicate menu items otherwise.

ingemarberg:
Sorry for the slow reply. I haven't had time to work on this in the last two weeks. However, I hope to be able to finish this PR and submit it for review this Sunday.

Thanks for the heads-up on the install script. I'll make sure to add a check to rename the existing .desktop file if it exists.

ingemarberg:
I'll close this as I spoke too soon about the requirements for Plasma on Wayland. See #3419 for further discussion. I'm sorry if I wasted anyone's time.

PR #3398, opened 2024-09-01
Add comment about the usage of the input field in the statistics page (#3394)

### LINKED ISSUE #3394: Request to add a comment for the input field in the statistics page
This is a request to add a comment for the input field in the statistics page stating that it is used as a public API in other clients, something along the lines of:

_"this form is an external API and should not be changed - other clients e.g. AnkiDroid programmatically update this form"_

In Ankidroid we are using this input field to set the current deck(for stats displaying) only for the statistics screen as some users have asked. See relevant discussion at https://github.com/ankidroid/Anki-Android/pull/15648#pullrequestreview-2244547928 . Tried to propose a PR but I don't know svelte.
Thank you!

lukesbart:
Resolve #3394 which requested that a comment be added to describe the usage of the input form in the statistics page in external apps.


dae:
Thanks for contributing! I've posted a comment on the original issue: https://github.com/ankidroid/Anki-Android/pull/15648#discussion_r1752566907

ts/routes/graphs/RangeBox.svelte
```diff
@@ -83,6 +83,8 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
             {collection}
         </label>
 
+        <!-- This form is an external API and should not be changed -
+	other clients e.g. AnkiDroid programmatically update this form -->
```
mikehardy, line 87:
```suggestion
        <!-- This form is an external API and care should be taken when changed -
	other clients e.g. AnkiDroid programmatically update this form by id -->
```

ts/routes/graphs/RangeBox.svelte
```diff
@@ -83,6 +83,8 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
         <input
             type="text"
```
mikehardy, line 89: @dae - I looked on svelte and I *think* this is how we add an id? And if so, I believe it's a reasonable id?
This would unblock the cross-repo PR thing we have going here, I'm just making a best effort to move it forward :-)

```suggestion
            type="text"
            id="searchText"
```

mikehardy:
I unblocked our PR despite the brittle "twiddle first text box" current state as it appears we *will* have some form of resolution here but it'll take a while for whatever it is to flow downstream.

Just for completeness, cross-link --> https://github.com/ankidroid/Anki-Android/issues/17075

dae:
I've tweaked the id name to be less likely to conflict

PR #3399, opened 2024-09-02
Updated error message to provide help for import if 'Legacy Import/Export' is enabled

gregorylearns:
I updated to the latest version but error still persisted. Figured out that it was the Legacy Importing that was the culprit

Enhanced the error message from "Unable to read file. It probably requires a newer version of Anki to import." 
  to include a suggestion for users to try unchecking 'Legacy import/export Handling' under Preferences > Editing > 
  Import/Export if they encounter the issue.

I also did not do the _Test must pass_ check under the `contributing.md` since i just changed a string.

First time contributing to the project! Very very grateful to the team!!!


dae:
Thanks for contributing Gregory!

PR #3410, opened 2024-09-08
If deck options are modified, ask before closing

Arthur-Milchior:
This imitates the way the note editor behaves. If a user assumes by error that chanhges are automatically saved, it ensures they won't lose them.

Also, this will eventually allows to have the same feature on AnkiDroid. While, currently, we always ask the user whether they want to close the deck options, even when there are no modification, which seems to regularly frustate users (including myself).

I'm new to Svelte, please let me know whether there is a better way to obtain the information from Svelte state that I missed.

Note that I ensured that only a boolean can be obtained. I didn't cause the whole state to be accessible. May be useful for some add-ons, I guess, but risks breaking too much things.

Regarding the deckoptions.py, I tried to imitate addcards.py way to check whether the add card view can be closed. Reusing the same function and variable name when possible.

### DIFF
--- a/qt/aqt/deckoptions.py
+++ b/qt/aqt/deckoptions.py
@@ -109,7 +109,7 @@ class DeckOptionsDialog(QDialog):
         """Calls `callback` with the information of whether any deck options are modified."""
         if self.ready:
             self.web.evalWithCallback(
-                "globalThis.anki.deckOptionsPendingChanges()", callback
+                "anki.deckOptionsPendingChanges()", callback
             )
         else:
             callback(False)
--- a/qt/aqt/deckoptions.py
+++ b/qt/aqt/deckoptions.py
@@ -14,6 +14,7 @@ from aqt.qt import *
 from aqt.utils import (
     KeyboardModifiersPressed,
     addCloseShortcut,
+    ask_user_dialog,
     disable_help_button,
     restoreGeom,
     saveGeom,
@@ -33,6 +34,8 @@ class DeckOptionsDialog(QDialog):
         self.mw = mw
         self._deck = deck
         self._setup_ui()
+        self._close_event_has_cleaned_up = False
+        self._ready = False
 
     def _setup_ui(self) -> None:
         self.setWindowModality(Qt.WindowModality.ApplicationModal)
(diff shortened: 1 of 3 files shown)


qt/aqt/deckoptions.py
```diff
@@ -57,15 +60,58 @@ def _setup_ui(self) -> None:
...
+
+        def callbackWithUserChoice(choice: int):
+            if choice == 0:
+                # The user accepted to discard current input.
+                self._close()
+
+        def callbackWithDataInformation(hasModifiedData: bool):
+            if hasModifiedData:
+                ask_user_dialog(
+                    tr.adding_discard_current_input(),
```
dae, line 89: card_templates_discard_changes() would be clearer

qt/aqt/deckoptions.py
```diff
@@ -57,15 +60,58 @@ def _setup_ui(self) -> None:
...
         self.mw.col.set_wants_abort()
         self.web.cleanup()
         self.web = None
         saveGeom(self, self.TITLE)
         QDialog.reject(self)
 
+    def hasModifiedData(self, callback: Callable[[bool], None]):
+        """Calls `callback` with the information of whether any deck options are modified."""
+        if self.ready:
+            self.web.evalWithCallback("globalThis.isStateModified()", callback)
```
dae, line 111: This is not a great name - it's not specific to the deck options screen, and may conflict with other usages. We also don't want to be polluting the global namespace like this. This is better placed until a prefix for Anki, and for the deck options, like 'anki.deckOptionsPendingChanges()'

qt/aqt/deckoptions.py
```diff
@@ -57,15 +60,58 @@ def _setup_ui(self) -> None:
...
+            return
+        evt.ignore()
+        self.ifCanClose()
+
+    def _close(self):
+        """Close. Ensure the closeEvent is not ignored."""
+        self._close_event_has_cleaned_up = True
+        self.close()
+
+    def ifCanClose(self):
```
dae, line 78: We can't avoid camelCase for the Qt methods, but our own methods like this one and the ones embedded inside it, are better done with snake_case.

dae (commented):
Mostly looks good:

qt/aqt/deckoptions.py
Arthur-Milchior: How do I create such a prefix?
If I try to directly do 
```
anki.deckOptionsPendingChanges = () => {
        return data.state.isModified();
    };
```

I got an error message telling me that `anki` is not defined.

My understanding is that I can't just declare it in the svelte file otherwise it would not be accessible from python. If I try to declare `export val anki` in `+page.ts` i've got the same problem. This is why I thought using the `globalThis` would be the only way to have a value in svelte that could escape svelte.

Arthur-Milchior:
@dae in case you missed it, I answered with a question under your comment above. I fear technically I'm currently stucked, surely because I'm a beginner in svelte

Arthur-Milchior:
In the meantime, I at least changed it so that there is less polution, and the code builds

qt/aqt/deckoptions.py
dae: An example from another file:

```typescript
    globalThis.anki = globalThis.anki || {};
    globalThis.anki.imageOcclusion = {
        mode,
        save,
    };
```

Arthur-Milchior: oh okay, `anki` is a field of `globalThis`. Nice

Arthur-Milchior:
Then I guess it can finally be merged. Hopefully

qt/aqt/deckoptions.py
```diff
@@ -57,15 +60,60 @@ def _setup_ui(self) -> None:
...
         self.web.cleanup()
         self.web = None
         saveGeom(self, self.TITLE)
         QDialog.reject(self)
 
+    def has_modified_data(self, callback: Callable[[bool], None]):
+        """Calls `callback` with the information of whether any deck options are modified."""
+        if self.ready:
+            self.web.evalWithCallback(
+                "globalThis.anki.deckOptionsPendingChanges()", callback
```
dae, line 112:
```suggestion
                "anki.deckOptionsPendingChanges()", callback
```

globalThis is useful for bypassing typechecking; we don't gain anything by using it in the Python code

dae:
Thanks Arthur!

PR #3411, opened 2024-09-09
change links to new knowledge base

brishtibheja:
Found these links in reviewer were outdated so merely replaced them here with ones. I couldn't find a FAQ entry for "front of the card is blank" though. Maybe we need to create one.


brishtibheja:
Hmm... did I do it wrong? Check is failing.

dae:
The formatting needed updating after the line length changed.

PR #3412, opened 2024-09-10
Add an option to show image from editor in folder

user1823:
This feature is quite handy, especially when you want to see an enlarged view of the image.

For e.g., see https://forums.ankiweb.net/t/copying-images-out-of-cards-is-hard/20272/6

Credits: @abdnh's Reveal in File Manager add-on (https://github.com/abdnh/anki-misc/tree/master/reveal_in_file_manager)

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -1586,35 +1586,34 @@ class EditorWebView(AnkiWebView):
 
     def contextMenuEvent(self, evt: QContextMenuEvent) -> None:
         m = QMenu(self)
-        self._maybe_add_cut_action(m)
-        self._maybe_add_copy_action(m)
+        if self.hasSelection():
+            self._add_cut_action(m)
+            self._add_copy_action(m)
         a = m.addAction(tr.editing_paste())
         qconnect(a.triggered, self.onPaste)
-        self._maybe_add_image_menu(AnkiWebView(self), m)
+        if self._opened_context_menu_on_image():
+            self._add_image_menu(AnkiWebView(self), m)
         gui_hooks.editor_will_show_context_menu(self, m)
         m.popup(QCursor.pos())
 
-    def _maybe_add_cut_action(self, menu: QMenu) -> None:
-        if self.hasSelection():
-            a = menu.addAction(tr.editing_cut())
-            qconnect(a.triggered, self.onCut)
+    def _add_cut_action(self, menu: QMenu) -> None:
+        a = menu.addAction(tr.editing_cut())
+        qconnect(a.triggered, self.onCut)
 
-    def _maybe_add_copy_action(self, menu: QMenu) -> None:
-        if self.hasSelection():
-            a = menu.addAction(tr.actions_copy())
-            qconnect(a.triggered, self.onCopy)
+    def _add_copy_action(self, menu: QMenu) -> None:
+        a = menu.addAction(tr.actions_copy())
+        qconnect(a.triggered, self.onCopy)
 
-    def _maybe_add_image_menu(self, webview: AnkiWebView, menu: QMenu) -> None:
-        if self._opened_context_menu_on_image():
-            a = menu.addAction(tr.editing_copy_image())
-            qconnect(a.triggered, self.on_copy_image)
+    def _add_image_menu(self, webview: AnkiWebView, menu: QMenu) -> None:
+        a = menu.addAction(tr.editing_copy_image())
+        qconnect(a.triggered, self.on_copy_image)
 
-            if is_win or is_mac:
-                url = webview.lastContextMenuRequest().mediaUrl()
-                file_name = url.fileName()
-                path = os.path.join(mw.col.media.dir(), file_name)
-                a = menu.addAction(tr.editing_show_in_folder())
-                qconnect(a.triggered, lambda: showinFolder(path))
+        if is_win or is_mac:
+            url = webview.lastContextMenuRequest().mediaUrl()
+            file_name = url.fileName()
+            path = os.path.join(mw.col.media.dir(), file_name)
+            a = menu.addAction(tr.editing_show_in_folder())
+            qconnect(a.triggered, lambda: showinFolder(path))
 
 
 # QFont returns "Kozuka Gothic Pro L" but WebEngine expects "Kozuka Gothic Pro Light"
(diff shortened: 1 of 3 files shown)


user1823:
Though I haven't tested this, I think that the automated test failure is not related to my changes.

abdnh:
It would be nice to open the file manager with the file selected. Here are relevant examples from add-ons:

https://github.com/ijgnd/anki21__OpenInExternalEditor_Rename_Duplicate_for_Image_Audio_Video/blob/master/src/showInFilemanager.py

https://github.com/abdnh/anki-misc/blob/master/reveal_in_file_manager/__init__.py

user1823:
> It would be nice to open the file manager with the file selected. 

What is the use case?

abdnh:
It potentially allows for more operations on the file (e.g. copy, opening in some editor, etc) but probably just opening the image is easier to implement, as you don't need to handle different file managers on Linux AFAIK.

user1823:
> It potentially allows for more operations on the file (e.g. copy, opening in some editor, etc)

Ability to copy the image has already been added in https://github.com/ankitects/anki/pull/3362

Adding the ability to open the image in editors doesn't seem important enough to justify the complexity. Plus, if we open the file manager instead of the file itself, opening the image will require an additional click.

glutanimate:
This looks like code extracted almost 1:1 from Image Occlusion Enhanced, including the lack of type annotations on `open_image`, which would be nice to improve here. Although this can be OK for more trivial snippets like this ([opinions differ](https://opensource.stackexchange.com/a/5061) on what is trivial), it would still be nice to credit add-on authors in these cases. Even features enabled by seemingly trivial code can take a while to ideate, develop, and test, so I think it's only fair to maintain that continuity of credits going from add-on feature to native feature.

Also I'm not sure if the implementation needs to be as complex as it was here. I wrote this ≈7 years ago, and at that point the OS-specific calls might have been to work around Qt bugs which are no longer present today. In [other places in Anki](https://github.com/ankitects/anki/blob/af115c7fdab5ae48e33fb8332f8793e2e7d6d5ae/qt/aqt/utils.py#L883-L884) we just delegate picking the correct command to Qt's desktop services API, so I would suggest checking if that is sufficient here as well. If you find that the direct calls are still needed, then it could be advisable not to use `subprocess` directly as the add-on did, but rather use existing utilities in Anki like [`call`](https://github.com/ankitects/anki/blob/34809f2520674b33dc288c8e138cc4c48958e284/pylib/anki/utils.py#L211). Similarly, it would probably be good to use the `no_bundled_libs` context for the `QDesktopServices.openUrl` call.

qt/aqt/editor.py
```diff
@@ -1621,11 +1609,12 @@ def _maybe_add_image_menu(self, menu: QMenu) -> None:
             a = menu.addAction(tr.editing_copy_image())
             qconnect(a.triggered, self.on_copy_image)
 
+            # inspired by Image Occlusion Enhanced add-on
```
glutanimate, line 1612: Sorry, my bad, I think I should have expressed myself better: I just meant leaving a small note in the PR in reference to the add-on :) I think including credits like this in the code is a nice gesture, but if we do it for every change inspired by add-ons, especially very small and trivial ones, the codebase would get very noisy very quickly. Perhaps something like this would _maybe_ make sense for much larger pieces of code (like full modules/packages). Though there I think that add-on authors would moreso be happy to just have a small footnote in changelogs, if the contribution/inspiration is large enough

user1823:
This is ready for a second review now.

> It's an extra step compared to a separate option to open in the default editor, but it's more flexible if we can't have both.

@dae, does this mean that adding both options would be acceptable?

> In other places in Anki, we just delegate picking the correct command to Qt's desktop services API, so I would suggest checking if that is sufficient here as well.

It seems that Qt doesn't provide any option for selecting the desired file in the file manager. So, I have used direct calls. As per your suggestion, I have used use [call](https://github.com/ankitects/anki/blob/34809f2520674b33dc288c8e138cc4c48958e284/pylib/anki/utils.py#L211) rather than using `subprocess` directly.

user1823:
Felt like learning the basics of git. Squashed the commits in this PR just to see how it works.

qt/aqt/utils.py
```diff
@@ -885,6 +886,33 @@ def openFolder(path: str) -> None:
...
+        tell application "Finder"
+            activate
+            select POSIX file '{path}'
+        end tell
+        """
+        call(osascript_to_args(script))
+    else:
+        # Just open the file in any other platform
+        with no_bundled_libs():
+            QDesktopServices.openUrl(QUrl(f"file://{path}"))
```
dae, line 903: Have you copied+pasted this code from somewhere?  From a quick read of the Qt docs, it looks like openUrl() should do what we want on all platforms?

qt/aqt/editor.py
```diff
@@ -1585,28 +1586,34 @@ def _processImage(self, mime: QMimeData, extended: bool = False) -> str | None:
...
 
-    def _maybe_add_copy_image_action(self, menu: QMenu) -> None:
-        if self._opened_context_menu_on_image():
-            a = menu.addAction(tr.editing_copy_image())
-            qconnect(a.triggered, self.on_copy_image)
+    def _add_image_menu(self, webview: AnkiWebView, menu: QMenu) -> None:
+        a = menu.addAction(tr.editing_copy_image())
+        qconnect(a.triggered, self.on_copy_image)
+
+        if is_win or is_mac:
```
dae, line 1611: Any feature we add should work across all OSes - is there a reason why non-Mac/Windows has been excluded here?

qt/aqt/utils.py
user1823: For local files, openUrl() opens the file in the default viewer if it can. But, we want it to highlight the file even when there is a viewer available.

qt/aqt/editor.py
user1823: It would definitely be better to make a feature work on all platforms. But, I have no idea how to make it work with all the different file managers in Linux. Even if this doesn't work on Linux, it is still an improvement especially when you consider that Windows + Mac users comprise most of the Anki's userbase.

qt/aqt/utils.py
dae: Sorry, I was under the mistaken impression that it would select the file, not open it directly.

qt/aqt/editor.py
dae: "We can't do it on Linux easily" is a reasonable answer for now :-) Thanks for clarifying.

user1823:
>> It's an extra step compared to a separate option to open in the default editor, but it's more flexible if we can't have both.
>
> @dae, does this mean that adding both options would be acceptable?

Would you accept another PR adding an "Open image" option?

dae:
Yes, if you feel it's necessary.

PR #3413, opened 2024-09-10
Fix pasting from the primary selection

### LINKED ISSUE #3249: The “reuse internal” clipboard feature cannot handle primary selections
> This is a bug report.
> Steps to reproduce:
> 
> Run Anki-24.04.1 on X11 Linux.
> Copy text A from a field using Ctrl+C to put that text into the clipboard buffer.
> Select some other text B to put that text into the primary selection.
> Middle-click with the mouse to paste the primary selection.
> Text A from the clipboard buffer is actually pasted instead of text B from the primary selection.

Originally reported on https://forums.ankiweb.net/t/the-reuse-internal-clipboard-feature-cannot-handle-primary-selections/46115

krischerven:
Fixes #3249


dae:
Thanks Kris!

PR #3415, opened 2024-09-12
typeanswer: minor cleanups

twwn:
Minor edits for slightly better readability/disambiguation.

Only the "skip DiffContext" has a little functional change: Pulling ahead the "nothing typed" check, so the case no longer unnecessarily runs any of the comparison code.


twwn:
I am not sure why Github's check forced the final fold() change, it seems detrimental and was not requested by my local checks.

rslib/src/typeanswer.rs
```diff
@@ -5,113 +5,160 @@ use std::borrow::Cow;
...
 
-    fn to_tokens(&self) -> DiffOutput {
+    fn to_tokens(&self) -> DiffTokens {
         let mut matcher = SequenceMatcher::new(&self.provided, &self.expected);
-        let opcodes = matcher.get_opcodes();
-        let mut provided = vec![];
-        let mut expected = vec![];
-        for opcode in opcodes {
+        let mut provided_tokens = Vec::new();
+        let mut expected_tokens = Vec::new();
```
dae, line 83: The class already includes Tokens in the name, so using _tokens for the property names feels a bit redundant.

rslib/src/typeanswer.rs
```diff
@@ -5,113 +5,160 @@ use std::borrow::Cow;
...
+            let encoded_text = htmlescape::encode_minimal(&isolated_text);
+            let class = token.to_class();
+            acc.push_str(&format!("<span class={class}>{encoded_text}</span>"));
+            acc
+        })
+    }
+
+    /// Prefixes a leading mark character with a non-breaking space to prevent
+    /// it from joining the previous token.
+    fn isolate_leading_mark(text: &str) -> Cow<str> {
```
dae, line 145: This method only depends on its inputs and not on self, so is better implemented as a free function. Same for the other methods that don't use self.

rslib/src/typeanswer.rs
twwn: `provided` and `expected` were used repeatedly as identifier for different variables in that same file. So I figured the addition worth it for clarity.

rslib/src/typeanswer.rs
twwn: That might be my familiarity with C++ and its tenants of *"beware namespace pollution"* and encapsulation. But this is Rust, are free functions a convention? The `fn new` "constructors" are also free functions?

Generally I find code easier to read if it's structured following the execution flow as close as possible—the better example for this question here being `fn get_slice`. I also found this way cleaner to rewrite `impl Diff` as shared `trait` in PR #3422, for instance `fn render_expected_tokens` there needs `&self` (in the NonCombining variant) but `fn render_tokens` doesn't.

twwn: On second thought, as far as this PR is concerned, turning them back into free functions requires barely more than shifting a line.

But: How shall I do it for the eponymous associated functions in the follow-up PR's `trait Diff`'s `impl`? Leave them as-is—it's a shared `trait`, after all? Or append `_noncombining` to the new variants? Since they're not used in the in the trait's default implementations, I can do either here.

twwn: Done, with the follow-up PR adjusted as these reverts to free functions require.

(Plus two more commits there as suggestions.)

dae:
Thank you for your patience!

PR #3416, opened 2024-09-12
Add a delete option to the card right-click menu in the browser

krischerven:
This would be a duplicate of the option under Notes, but it would be more intuitive for newcomers to not have to go into a nested submenu to find an option that you would normally expect to find in the main right-click menu. It's also more efficient for non-hotkey users.


dae:
Sorry, I think it's better if users understand the note/card distinction, and this just muddies the waters.

PR #3419, opened 2024-09-16
Fix Qt desktop file name warning

krischerven:
When running Anki from the terminal, you receive the following warning:

> Qt warning: QGuiApplication::setDesktopFileName: the specified desktop file name ends with .desktop. For compatibility reasons, the .desktop suffix will be removed. Please specify a desktop file name without .desktop suffix

This PR removes that warning.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -118,7 +118,7 @@ yellowjello <github.com/yellowjello>
 Ingemar Berg <github.com/ingemarberg>
 Ben Kerman <ben@kermanic.org>
 Euan Kemp <euank@euank.com>
-Kieran Black <kieranlblack@gmail.com> 
+Kieran Black <kieranlblack@gmail.com>
 XeR <github.com/XeR>
 mgrottenthaler <github.com/mgrottenthaler>
 Austin Siew <github.com/Aquafina-water-bottle>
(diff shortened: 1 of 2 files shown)


dae:
Related: #3396. @ingemarberg, how would you like to proceed?

ingemarberg:
Both PRs will take care of this warning. This comes down to whether we want to rename the .desktop file according to the reverse domain standard or not. If we decide to keep the `anki.desktop` file, this is a good way to get rid of the warning. If not, the other PR will also get take care of it.

dae:
If keeping the .desktop file is a valid alternative provided we have setDesktopFileName(), that seems like it might be the least disruptive approach, as then we don't need to worry about removing old files on upgrade. Happy for me to merge this instead, or have I missed something?

ingemarberg:
I agree that would be the easiest way to deal with it. In fact, it seems that I spoke too soon about the requirements for Plasma on Wayland. I did some more testing on Plasma 6.1.4 on Fedora 40 and everything works just fine. After doing some research, I'm pretty sure that the .desktop file on my normal system was the cause of the missing icon in #3396. This was probably a Flatpak issue after all and not related to Plasma.

I'm not sure why the KDE wiki lists this as a hard requirement for Wayland, but maybe I just misunderstood. Sorry for wasting your time @dae. I'll close my PR now.

dae:
No problem, appreciate you taking the time. And thank you @krischerven!

PR #3422, opened 2024-09-16
 typeanswer: [type:nc] – ignores combining characters

twwn:
Building on my previous [PR's cleanup](https://github.com/ankitects/anki/pull/3415), adds a comparison variant to `[type]` which ignores when combining characters of the expected field are missing from the typed input. It still shows these characters in the "expected" line for reference.

It's useful for languages with e.g. diacritics that are required for reference (such as in dictionaries), but rarely actually learned or used in everyday writing. Among these languages: Arabic, Hebrew, Persian, Urdu.

**Previously:**
pre

**Now:**
after
[This also removes the need to have both variants of words/sentences present as separate fields, to show them redundantly, etc.]

The bool `combining` controls it as new final parameter of both relevant `compare_answer` functions. In Python, it's set to true by default.

Use on the note templates: `[type:nc:field]`


dae:
Thanks twwn, I'll try to review this soon. (For my own reference, previous PR: https://github.com/ankitects/anki/pull/680)

twwn:
The minor commits have links to relevant Stack Overflow threads in their messages.

The commit *"simplify by using nfkd throughout"* does so by removing the use of NFC in the original case as well. Though that required adjusting two testcases, this only reflects internal processing: I compared both before/after and the rendered end results are the same.

dae:
If you could bring the branch up to date with main (either here or in a new PR if that's easier), that'll make it easier for me to review.

twwn:
Done.

The small commits are just suggestions. I've split the NFKD commit further: The second one is only there if want the extra function gone, but since that requires an additional bool, I myself would rather not actually merge it.

dae:
Thanks, I'll get back to this again soon. In case you didn't see it, there's been a report about issues with the recent changes: https://forums.ankiweb.net/t/anki-24-10-beta/49989/22

twwn:
That report, it's since `prepare_expected` (which strips the html tags) is currently no longer run for the two "new" cases: when nothing is typed, or the *typed=expected* one.

I can…

* ensure that HTML tags actually get evaluated rather than just dumped as text (that's definitively my omission)
* add `prepare_expected` back in

… each option for both cases, or mix them. The original behavior is restored with `prepare_expected` for both cases.

It's a moot argument if changes are not to break anything, but IMHO it's really bad card design to clutter the field that you type with additional tags or even media.

That's why I feel like the most natural way (if that word makes sense here) is to add `prepare_expected` back in for the *typed=expected* case but leave in & instead evaluate the HTML for the *nothing-typed* one. As in: `type::` during review becomes an offer—if the user skips it, Anki would now treat the field as if the template included it directly.

dae:
> instead evaluate the HTML for the nothing-typed one.

So HTML tags like bold would appear when nothing is typed, and not appear otherwise? I fear that inconsistency might lead to some additional "bug" reports.

> it's really bad card design to clutter the field that you type with additional tags or even media.

If you're saying that you think users should put those things in a different field that's not being typed in, I fully agree, though sadly it seems to be quite common to put things like sound tags in the same field.

twwn:
>So HTML tags like bold would appear when nothing is typed, and not appear otherwise?

Yes, it would break with the old behavior. But it's just a suggestion, I entirely understand if it's not deemed good/worthwhile.

I guess I'll just restore `prepare_expected` for both cases for now, it's easier to do, anyway. It'll take a while since I really need to sleep 😃.

dae:
Sounds good, and no worries; it'll likely be a few days before I can build another beta anyway.

dae:
Thank you for the atomic commits and multiple PRs, that has made my life a lot easier. :-) Regarding the NFKD change, I'm a bit worried this might have unintended side effects on other inputs, but hopefully if that's the case it'll become apparent due to beta testing.

The micro optimisation was probably unnecessary here, but good practice for hotter code paths. I've only reverted it because it caused a conflict for the commit above it.

dae:
Am I correct in assuming that https://forums.ankiweb.net/t/anki-24-10-beta/49989/143 is caused by the NFKD simplification? If that change has caused us to output a mix of NFC (note text) and NFD, that seems suboptimal. Thoughts?

twwn:
I _would_ have said: Although this deck's issue could be from unsupported Javascript, we can easily return to NFC for the default comparison.

But the deck with the issue is weird: It not at all use `type::Vocab` for the typing field, but `type:sort_id`? So it's comparing not to "ようがん" here but "7330". That's not an issue of NFD/NFC, and its Javascript doesn't do anything that would trip over this, anyway.

I'll look into it more, maybe ask on the forum.

twwn:
I _would_ have said: Although this deck's issue could be from unsupported Javascript, we can easily return to NFC for the default comparison.

But the deck with the issue is weird: It doesn't not at all use `type::Vocab` for the typing field, but `type:sort_id`? So it's comparing not to "ようがん" here but "7330". That's not an issue of NFD/NFC, doubly unsupported, and its Javascript doesn't do anything that would trip over the latter, anyway. That suggested hotfix even normalizes to NFD, not to the previous NFC.

Shall I look into this more? Since it seems to be a serious issue of unsupported spaghetti code.

twwn:
Alright, I dunno why returning to NFC didn't work out the first time when I wrote the prior comment's versions. I've now recompiled a fresh checkout and it actually does seem to fix it.

So I'd for now propose to use NFC instead throughout:

* Anki NFC's all field content, anyway? Is that's what you meant by "output a mix"? In that case, it should indeed be NFC.
* I don't actually recall if there was an algorithmic reason I picked NFKD way back when I originally wrote the noncombining comparison. From testing a few words right now, NFC seems to work fine as well.

So I'll revert (into the open PR) and just keep testing. Best case: It just works for both comparisons. Worst case: NFKD returns only for the NC comparison.

_Unrelated:_ I asked the protobuf maintainers about an Anki startup crash. [They say the bug](https://github.com/protocolbuffers/protobuf/issues/17797) is within Anki.

twwn:
(PR #3482)

twwn:
I haven't observed the crash with the binary builds (rc2/rc3), they may yet be unaffected. But even on my distro build, the crash occurs frequently but not consistently, as if caused by some race condition.

I wish the error message gave an actual pointer as to where to look, but it's left me clueless.

Anyway, Arch and Anki seem to be using the same protobuf version? If I understand correctly, the only difference would be that the former [builds from the general tarball](https://gitlab.archlinux.org/archlinux/packaging/packages/protobuf/-/blob/main/PKGBUILD?ref_type=heads), and the latter only pulls the python bindings from PyPi.

dae:
IIRC, Arch's `anki-bin` uses our Python wheels, and `anki` builds them on the local machine. Do both exhibit the same problem?

twwn:
Yes. I just tried with `anki` (my PKGBUILD is identical except I'm using the git repo  @ c1a2b03 + `-ffat-lto-objects`), then replaced it with `anki-bin` (altered to install 24.10b3), and repeated that switch once more. Exact same crash.

The former invokes `python` directly, the latter `python3`, that's the only meaningless difference in the output.

dae:
@twwn Is は/ぱ expected to work? It seems not to at the moment: https://forums.ankiweb.net/t/bug-with-type-nc/52893

twwn:
That's due to the return to NFC.

Although semantically it doesn't make sense to use `nc:` for は/ば due to them being different sounds, I guess NFKD does behave more as users will expect. So I'll open a PR that restores it, as I wrote in the [Oct 8 comment](https://github.com/ankitects/anki/pull/3422#issuecomment-2400669438).

(Actually, my use of `nc:` as abbreviation is coincidence. I was oblivious to the `nc:` search option. So I didn't check whether it behaves the same, either.)

monk-time:
Works like a charm, thank you so much! This feature is beyond crucial for Hebrew, now that I've tested it I can't imagine how I used to live without it.

PR #3423, opened 2024-09-16
Consider cards already reviewed today when calculating new/review ratio

### LINKED ISSUE #3374: Consider cards already reviewed that day when calculating new/review ratio?
Originally reported on https://forums.ankiweb.net/t/dont-rebuild-queue-on-selecting-another-deck/48349/10

krischerven:
Fixes #3374


dae:
Sorry, I don't think this is important enough to justify a schema change. My thinking was that we might be able to use the existing counts stored in the select deck, and add them to the pending count. If that won't do, I'm not sure fixing this is worth new tables, or extracting the info from the revlog.

krischerven:
> My thinking was that we might be able to use the existing counts stored in the select deck, and add them to the pending count.

I'll look into that

dae:
Thank you! Please open a new PR when you're ready.

Issue #3147, opened 2024-04-17
Integration of the Helper add-on stats


dae:
> I think that since LMSherlock and Dae are planning to integrate FSRS Helper add-on features, it makes sense to implement the stats as well.

Originally reported on https://forums.ankiweb.net/t/integration-of-the-helper-add-on-stats/43643

L-M-Sherlock:
I'm considering it. The first question is, the current stats doesn't have any table. Is it acceptable to insert a table into the stats? Do you mind it will break the style of current design?

![image](https://github.com/user-attachments/assets/6e26594b-1c9b-4e25-b12c-d6868477187a)

Expertium:
I'm concerned with whether it will be displayed correctly on mobile devices.

dae:
It'll require some scrolling - you can see it if you resize the desktop browser to fit 3 columns. 
![2024-09-20T22:01:47,990699719+10:00](https://github.com/user-attachments/assets/1f28f0f7-bf31-4136-97fb-318579e26859)

salmanuc:
[...]

PR #3425, opened 2024-09-18
Feat/true retention stats

L-M-Sherlock:
Related issue:
- #3147

Preview:

image


dae:
- If this is not expected to change soon, perhaps we should be making it translatable from the start?
- The summary at the bottom is a bit long. Could we move it to the top, and make it simpler? E.g. 'Pass rate of cards with an interval >= 1 day'?

dae:
Also, the green text doesn't fit so well with the rest of Anki's interface. Maybe standard black would be a better fit, with perhaps some bolding on key elements?

L-M-Sherlock:
> * If this is not expected to change soon, perhaps we should be making it translatable from the start?

Good point! I will make it translatable later.

> * The summary at the bottom is a bit long. Could we move it to the top, and make it simpler? E.g. 'Pass rate of cards with an interval >= 1 day'?

The summary is written by @user1823 in https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/200. I'd like to ask for his opinion.

> Also, the green text doesn't fit so well with the rest of Anki's interface. Maybe standard black would be a better fit, with perhaps some bolding on key elements?

OK.

user1823:
> 'Pass rate of cards with an interval >= 1 day'?

Looks good.

L-M-Sherlock:
image

Expertium:
"7 days ago" and 30days ago" implies that it's retention for ONE specific day. "Last week" and Last month" is better. "Last year" would be nice, though this is already a ton of numbers.

L-M-Sherlock:
OK

image

I also removed the Card Counts part because it's confusing and irrelevant to true retention.

Expertium:
Will you add these as well?
image
I really hope so, I like knowledge acquisition rate, and I think a lot of people will like it too

ftl/core/statistics.ftl
```diff
@@ -86,6 +86,18 @@ statistics-counts-learning-cards = Learning
 statistics-counts-relearning-cards = Relearning
 statistics-counts-title = Card Counts
 statistics-counts-separate-suspended-buried-cards = Separate suspended/buried cards
+statistics-true-retention-title = True Retention
+statistics-true-retention-subtitle = Pass rate of cards with an interval >= 1 day.
```
user1823, line 90:
```suggestion
statistics-true-retention-subtitle = Pass rate of cards with an interval ≥ 1 day.
```

L-M-Sherlock:
> I really hope so, I like knowledge acquisition rate, and I think a lot of people will like it too

The stats has included some of them:

image

image

Expertium:
Some, yes, but not all

L-M-Sherlock:
Fine. I added the Estimated total knowledge (cards):

image

Daily load and retention by notes require more works to do, so I will consider it in another PR.

Expertium:
I think these stats need their own dedicated section, with explanations copied from the add-on.

dae:
Thanks Jarrett.

Some minor issues I noticed that I won't block on:

- collecting the results into a hashmap first is not the most efficient, but it doesn't really matter in this case
- total knowledge is probably more useful as an integer
- in some cases, the start of the table is cut off as the graphs window is resized

dae:
(further changes welcome in follow-up PRs)

PR #3431, opened 2024-09-20
Add "open image" option to editor

user1823:


user1823:
The name of the function "openFolder" is misleading because it can also be used to open files. Do you have any suggestion that is clear yet succint?

dae:
`openFolder` is existing code that may be used by add-ons, so simplest to just leave the name alone.

The code at present is incorrect, and will print an error message when there's no media URL on the clipboard.

qt/aqt/editor.py
```diff
@@ -1622,10 +1623,13 @@ def _add_image_menu(self, webview: AnkiWebView, menu: QMenu) -> None:
         a = menu.addAction(tr.editing_copy_image())
         qconnect(a.triggered, self.on_copy_image)
 
+        url = webview.lastContextMenuRequest().mediaUrl()
```
user1823, line 1626: > The code at present is incorrect, and will print an error message when there's no media URL on the clipboard.

Due to this code?

I can't think of a case where there won't be a media URL because this code runs only when the user right-clicks on an image.

https://github.com/ankitects/anki/blob/81c30bf99e14da7b8d39083583f9c6a853cf4658/qt/aqt/editor.py#L1609

dae: lastContextMenuRequest() is returning None there. Happens on two different machines I tested with.

user1823: Because this PR is built on top of the code added by @bpnguyen107 in https://github.com/ankitects/anki/pull/3362, I would like to ask him if he has any suggestion.

bpnguyen107 (changes requested):
Hey @user1823, I was able to replicate @dae's error. I believe the suggestions I left you would fix those errors. Your open image feature still doesn't work though. 

Also, I would suggest manually testing all your code changes to Anki using ./run if you aren't already. It allows you to catch these bugs!

bpnguyen107:
Also, I think I found a regression while looking into this pull request. Pasting an image into the editor no longer works using either Ctrl+V or Right-Click -> Paste. I am guessing it was introduced in https://github.com/ankitects/anki/pull/3413 since that has to do with pasting. @dae

user1823:
> Also, I suggest manually testing all your code changes to Anki using ./run if you aren't already. It allows you to catch these bugs!

Sorry. But, that is too much for me because I am not really a developer.

Also, because my previous PR (https://github.com/ankitects/anki/pull/3412) also includes `url = webview.lastContextMenuRequest().mediaUrl()`, I wonder why it is a problem in this PR but not in that PR. Or does the same problem exist in that PR too and it was just left undiscovered?

bpnguyen107:
I don't know. Did you test your previous PR?

user1823:
The error should be fixed now (thanks Ben). I would appreciate if someone could test whether the feature works correctly.

user1823:
Well, I tested this by creating an add-on that patches the function in Anki's latest beta. The open image function seems to work correctly after these changes, though the show in folder function doesn't seem to work.

user1823:
With the latest change, the "show in folder" option also works. However, I am now directly using subprocess rather using call.

user1823:
> I'd also expect the previous usage to work. Does the original call not work for you

It didn't work. Probably, there is a trivial error but I couldn't find it.

user1823:
This works for me. Thanks!

dae:
Damn, I'd intended to merge this in for beta 2. Oh well.

PR #3432, opened 2024-09-20
Bug with “A 100 day interval will become X days.”

### LINKED ISSUE #3430: Bug with “A 100 day interval will become X days.”
We can fix this by truncating the precision to 2 decimal places when calculating the value.

> Steps to reproduce:
> 
> Choose a value of desired retention. It will have two digits after the decimal place
> Try adding a third digit (1, 2, 3 or 4, otherwise the value will be rounded up) after the decimal place, which will immediately disappear
> Notice that X in “A 100 day interval will become X days.” changes
> 
> In the gif, X varies between 66 and 62 days, despite the fact the third digit after the decimal place automatically disappears after I enter it.
> 
> Version ⁨24.06.3 (d678e393)⁩

Originally reported on https://forums.ankiweb.net/t/bug-with-a-100-day-interval-will-become-x-days/49629

bpnguyen107:
Fixes #3430

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -187,7 +187,7 @@ Christian Donat <https://github.com/cdonat2>
 Asuka Minato <https://asukaminato.eu.org>
 Dillon Baldwin <https://github.com/DillBal>
 Voczi <https://github.com/voczi>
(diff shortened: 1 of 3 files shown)


dae:
Thanks Ben!

Issue #3427, opened 2024-09-20
Add 'Revert to Backup' option to File menu


dae:
> Currently, in order to access the backups, the user has to go to File → Switch Profile, which is very unintuitive. I mean, do you, the person reading this, think, “Oh, Switch Profile, yeah, that’s probably where the backups are”?
> My suggestion is to put it between Create Backup and Exit
> 

We probably want a confirmation dialog that shows the time the backup was created (or a relative time like '2 hours ago')

Originally reported on https://forums.ankiweb.net/t/move-open-backup/49488

PR #3434, opened 2024-09-21
Add revert to backup option to file menu

taylorobyen:
This addresses #3427 

`Load Backup` button to File menu on the main window
image
When loading a backup the relative time it was taken is displayed
image

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -191,7 +191,7 @@ Ben Nguyen <105088397+bpnguyen107@users.noreply.github.com>
 Themis Demetriades <themis100@outlook.com>
 Luke Bartholomew <lukesbart@icloud.com>
 Gregory Abrasaldo <degeemon@gmail.com>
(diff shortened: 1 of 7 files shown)


dae:
Sorry Taylor, I appear to have made a thinko. When I suggested mentioning the previous backup time, it was because I was assuming we'd be automatically selecting the most-recent backup. Giving the user a list of backups is more flexible, and it makes the time shown somewhat redundant. I was originally going to ask you to move the "x minutes ago" code into Rust so it can be shared across the clients, but since it's a bit redundant, the fastest way to get the PR over the line would be to change the message to show the backup filename instead of a relative time. Does that work for you?

taylorobyen:
No worries Damien, I've switched it to be filename as requested. 👍

dae (commented):
(will put this through myself; was a pending suggestion that I forgot to send)

dae:
Thanks Taylor!

PR #3441, opened 2024-09-26
Fix/Reschedule doesn't work on cards in filtered deck

### LINKED ISSUE #3439: [Bug] Reschedule doesn’t work on cards in filtered deck
> If I use Reschedule cards on change, while some cards are in a filtered deck, rescheduling doesn’t happen for cards in the filtered deck.

Originally reported on https://forums.ankiweb.net/t/bug-reschedule-doesnt-work-on-cards-in-fiktered-deck/49896

L-M-Sherlock:
fix #3439

A similar function may also need to be fixed:

https://github.com/ankitects/anki/blob/f3b0afcc624c6fb2ff259cf34e819cc2c651da67/rslib/src/search/sqlwriter.rs#L484-L495

### DIFF
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -60,10 +60,11 @@ pub enum SearchNode {
     EditedInDays(u32),
     CardTemplate(TemplateKind),
     Deck(String),
-    /// Matches cards in a list of decks (original_deck_id is not checked).
+    /// Matches cards in a list of deck ids. Cards are matched even if they are
+    /// in a filtered deck.
     DeckIdsWithoutChildren(String),
     /// Matches cards in a deck or its children (original_deck_id is not
-    /// checked).
+    /// checked, so filtered cards are not matched).
     DeckIdWithChildren(DeckId),
     IntroducedInDays(u32),
     NotetypeId(NotetypeId),
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -161,7 +161,12 @@ impl SqlWriter<'_> {
                 write!(self.sql, "n.mid = {}", ntid).unwrap();
             }
             SearchNode::DeckIdsWithoutChildren(dids) => {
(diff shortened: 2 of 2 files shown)


dae:
These are currently documented as not matching cards in filtered decks:

```
    /// Matches cards in a list of decks (original_deck_id is not checked).
    DeckIdsWithoutChildren(String),
    /// Matches cards in a deck or its children (original_deck_id is not
    /// checked).
    DeckIdWithChildren(DeckId),
```

I've done a brief scan over the existing usages, and didn't notice any cases where the change is likely to break things, but it's not how they were originally intended to behave, so I get a bit nervous about changing them. Any thoughts on this change @RumovZ?

RumovZ:
Wouldn't that at least break (or alter) user searches containing `did:...`?

Looks like you've added `DeckIdsWithoutChildren` to avoid bugs by forcing explicitness. As it seems the original deck id is a similar source of issues, maybe it would be best to make that explicit, too:
```Rust
DeckIds {
    ids: String,
    with_children: bool,
    with_filtered: bool,
}
```

brishtibheja:
> Wouldn't that at least break (or alter) user searches containing `did:...`?

That isn't documented anywhere, so do users use this search? I think most wouldn't be aware.

RumovZ:
It's probably not used much, but might still be useful to add-ons and power users.
In any case, I don't see an argument for breaking any compatibility when it's so easy to support both use cases.

dae:
On making with_filtered explicit: my first thought was "good idea, why didn't I think of that?". :-) Then I noticed a few things:

- One of our current two cases uses a number, not a string of numbers, so merging gets a little more complicated
- The current behavior of not including filtered decks in a did:xxx search is inconsistent with a deck:xxx search
- I can't think of many cases where one would want to exclude the filtered cards, and in those cases, it's easy to exclude them from the search. Whereas we provide no existing way to include filtered cards in the search when searching via deck id.

Given the above, I feel like changing the docs might be the simplest way forward. Merging without feedback to keep things moving forward, but happy to discuss/change/revert if you feel strongly.

PR #3443, opened 2024-09-26
Fix occlusion drift again

### LINKED ISSUE #3370: Image occlusion masks don't stay in the same place

taylorobyen:
Fixes #3370

Addressed bugs preventing occlusions from being added as a result of my previous fix. 

I'm opening this as a draft for now, because I'm getting some strange behavior with the occlusions appearing on images they aren't supposed to. I'm able to recreate this on the main branch as well as the latest official version of Anki, so it doesn't appear to be my changes causing it. I will follow up once I have more information.

### DIFF
--- a/ts/routes/image-occlusion/shapes/to-cloze.ts
+++ b/ts/routes/image-occlusion/shapes/to-cloze.ts
@@ -4,7 +4,7 @@
 import { fabric } from "fabric";
 import { cloneDeep } from "lodash-es";
 
-import { getBoundingBox } from "../tools/lib";
+import { getBoundingBoxSize } from "../tools/lib";
 import type { Size } from "../types";
 import type { Shape, ShapeOrShapes } from "./base";
 import { Ellipse } from "./ellipse";
@@ -97,7 +97,7 @@ export function baseShapesFromFabric(): ShapeOrShapes[] {
         ? activeObject
         : null;
     const objects = canvas.getObjects() as fabric.Object[];
-    const boundingBox = getBoundingBox();
+    const boundingBox = getBoundingBoxSize();
     // filter transparent rectangles
     return objects
         .map((object) => {
@@ -168,10 +168,6 @@ function fabricObjectToBaseShapeOrShapes(
         );
         shape.left = newPosition.x;
         shape.top = newPosition.y;
-    }
-
-    if (size == undefined) {
-        size = { width: 0, height: 0 };
     }
 
     shape = shape.toNormal(size);
--- a/ts/routes/image-occlusion/store.ts
+++ b/ts/routes/image-occlusion/store.ts
@@ -17,3 +17,19 @@ export const ioImageLoadedStore = writable(false);
 export const opacityStateStore = writable(false);
 // store state of text editing
 export const textEditingState = writable(false);
+// Stores if any objects have been modified on the canvas
+export const saveNeededStore = writable(false);
+
+/**
+ * Saves all shape information as clozes to backend
+ */
+export function saveChanges() {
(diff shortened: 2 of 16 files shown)


dae:
Ouch, that's no good. If it's reproducible, would you mind bisecting to figure out the commit that caused it, and providing steps to reproduce?

ts/routes/image-occlusion/shapes/to-cloze.ts
```diff
@@ -170,10 +170,6 @@ function fabricObjectToBaseShapeOrShapes(
         shape.top = newPosition.y;
     }
 
-    if (size == undefined) {
-        size = { width: 0, height: 0 };
-    }
```
dae, line 175: What motivated this change? Perhaps related to the Svelte 5 update?

taylorobyen: Size should never be undefined now; I've moved it higher up in the logic:
`export const getBoundingBoxSize = (): Size => {
    const boundingBoxSize = getBoundingBox()?.getBoundingRect(true);
    if (boundingBoxSize) {
        return {width: boundingBoxSize.width, height: boundingBoxSize.height};
    }
    return {width: 0, height: 0};
}`

taylorobyen:
Latest 2 commits fix #3449, upon testing I don't see any new regressions/bugs as a result of my changes. The entire architecture was a bit of a mess, so I refactored it a bit to be more streamlined. It seems like the issue was caused from ` $: $changeSignal, onChange();` in `MaskEditor.svelete` being called immediately upon note change which would overwrite the selected notes shapes with the previously selected notes shapes. 

 However, I have found more unrelated bugs that I will report. Once I knock out these next couple bugs I'll see if I can identify when #3449 was introduced.

taylorobyen:
Fixed #3451, #3452, and #3456. Also, I made it so on creation of text labels it enters edit mode on the label and switches the tool back to the cursor. I found having to click onto the label to edit it after creation tedious and then clicking off the label to deselect I would always unintentionally create additional labels.  

I've done some more testing and now everything seems to work as intended.

dae (commented):
Thanks for looking into these issues Taylor!

@krmanik I know you're busy recently, but I wanted to give you a brief chance to review this PR, as it's mainly adjusting code I think you wrote in the past. If I don't hear from you in 48 hours, I'll merge this in, assuming there are no other pending changes.

ts/routes/image-occlusion/store.ts
```diff
@@ -17,3 +17,19 @@ export const ioImageLoadedStore = writable(false);
...
+export function saveChanges() {
+    saveNeededStore.set(true);
+}
+
+/**
+ * Clears the pending save operation after completion
+ */
+export function saveCompleted() {
+    saveNeededStore.set(false);
+}
```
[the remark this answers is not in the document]
taylorobyen: I've removed the helper functions and have instead accessed the store directly for all calls.

dae:
Thanks Taylor! It's a bit earlier than 48 hours, but I presume Mani doesn't have anything to add.

PR #3447, opened 2024-09-27
lazy_static → once_cell → stabilized versions

twwn:
Two cleanup commits:

1. Replace the remaining lazy_static use with once_cell, effectively its replacement. The former is also unmaintained.
2. Replace once_cell with stabilized LazyCell / LazyLock as much as possible.

See 2.'s commit message for details and the one remaining use.


twwn:
The repo check fails over…

>MSRV (Minimum Supported Rust Version) is `1.65.0`

… but [`rust-toolchain.toml`](https://github.com/ankitects/anki/blob/main/rust-toolchain.toml) says it's 1.81?

PR #3454, opened 2024-09-28
Fix error when closing deck options

voczi:
Addresses https://forums.ankiweb.net/t/anki-24-10-beta/49989/12
Repro consists of closing the deck options page really fast after opening it. Or just by having a slower PC, maybe?


Arthur-Milchior:
Okay, forgetting the _ was really absurd. Sorry.

dae:
I didn't catch it either. Thanks @voczi!

PR #3459, opened 2024-09-30
Fix Mnemosyne importer chokes on due dates >= 2038

### LINKED ISSUE #3428: Mnemosyne importer chokes on due dates >= 2038
> Hello
> Just starting to use Anki. I tried to import cards from Mnemosyne but I got the error message :
> Error: 500: JsonError { info: “invalid value: integer 2172355200, expected i32 at line 1 column 50592” }
> What can I do ?
> Thanks

next_rep with a value like 2172355200 fails

Originally reported on https://forums.ankiweb.net/t/error-importing-mnemosyne-cards-to-anki/49420/6

bpnguyen107:
I believe this fixes #3428. It seems to be a [well-known problem](https://en.wikipedia.org/wiki/Year_2038_problem).

I haven't tested it yet though. Can you share the problematic .db with me so I can see if this fixes the problem @dae? Even if it doesn't, we probably want to keep this change thinking ahead towards 2038... It'd probably be a good idea to make sure anywhere else Unix time is used in the codebase is also at least 64 bits.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -187,7 +187,7 @@ Christian Donat <https://github.com/cdonat2>
 Asuka Minato <https://asukaminato.eu.org>
 Dillon Baldwin <https://github.com/DillBal>
 Voczi <https://github.com/voczi>
-Ben Nguyen <105088397+bpnguyen107@users.noreply.github.com>  
+Ben Nguyen <105088397+bpnguyen107@users.noreply.github.com>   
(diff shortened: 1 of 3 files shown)


dae:
Thanks Ben!

The main outstanding issue I'm aware of is that card.mod is an i32. It's not a pressing issue, as we don't store 1+ day intervals in seconds, but we'll need to deal with it at one point.

PR #3464, opened 2024-10-01
typeanswer: hoist strip_expected

twwn:
One final cleanup to the existing code: Given `DiffTrait` exclusively works on the output of `prepare_expected` (renamed `strip_expected`), hoist it into a single call within `compare_answer`.

---

This will also make it trivial to implement an "offer" option for *type* (user types nothing → Anki includes field unstripped): `{{type:offer:Field}}`

But, question: When a field replacement involves 2+ ':', prior to even reaching `template_filters.rs`, Anki seems to "sanitize" that replacement? The field gets treated as if as directly included: E.g., `[[type:offer:nc:Field]]` never reaches `apply_filters`—its input instead already includes the field's HTML? Could you tell me where this sanitization happens, please?

This also plays into another idea: A new parameterizable filter that sanitizes a field for safe use as CSS identifier(s), maybe via the *[cssparser](https://docs.rs/cssparser/latest/cssparser/)* crate, with basic slice notation. In particular, so `{{Tags}}` with hierarchical tags can be used, since those unfortunately where implemented with a `::` delimiter (rather than, say, `__`, which would be legal in CSS). This would aid with styling specialized card types.

E.g, `{{css:Tags}}` would turn *"1st::2nd::3rd foo"* into *"1st 2nd 3rd foo"*, `{{css:0-1:Tags}}` into *"1st 2nd foo"*, `{{css:-1:Tags}}` into *"3rd foo"*, etc. Maybe with another parameter when another replacement is desired, etc.

Would this be sth. worthwhile?

### DIFF
--- a/rslib/src/typeanswer.rs
+++ b/rslib/src/typeanswer.rs
@@ -34,12 +34,12 @@ macro_rules! format_typeans {
 
 // Public API
 pub fn compare_answer(expected: &str, typed: &str, combining: bool) -> String {
-    if typed.is_empty() {
-        format_typeans!(htmlescape::encode_minimal(&prepare_expected(expected)))
-    } else if combining {
-        Diff::new(expected, typed).to_html()
-    } else {
-        DiffNonCombining::new(expected, typed).to_html()
+    let stripped = strip_expected(expected);
+
+    match typed.is_empty() {
+        true => format_typeans!(htmlescape::encode_minimal(&stripped)),
+        false if combining => Diff::new(&stripped, typed).to_html(),
+        false => DiffNonCombining::new(&stripped, typed).to_html(),
     }
 }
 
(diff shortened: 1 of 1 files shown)


dae:
I'm afraid I'm not very fond of either of the suggestions:

- I still think it's confusing to behave inconsistently when no input has been provided. You might think users would be deliberately opting in to this behaviour, but deck authors will also use it if we make it available.
- This feels like a fairly niche use case, and users can already apply complex transforms with JavaScript

aleksejrs:
I haven't tested, but it seems like this broke https://github.com/scott2000/answerset because of 3 vs 4 arguments.

dae:
> This add-on is implemented as a monkey patch

> Monkey patching is useful in the testing stage, and while waiting for new hooks to be integrated into Anki. But please don’t rely on it long term, as monkey patching is very fragile, and will tend to break as Anki is updated in the future.

Issue #3338, opened 2024-08-01
Toggling list items currently loses editor focus when empty item


dae:
> The Ctrl+, and Ctrl+. operation to toggle unordered/ordered list items will allow more text to be typed after the list if the caret is on a list item with text, but if it is on an empty list item, the editor loses the ability to be typed in until it is clicked again, while still retaining the blue outline.

Originally reported on https://forums.ankiweb.net/t/toggling-list-items-currently-loses-editor-focus-when-empty-item/47685

lukesbart:
I am currently working on this issue. I found where this call happens and am trying to figure out why the element loses focus.

PR #3465, opened 2024-10-01
Fix editor loses focus when toggling list and copy selects wrong target deck for filtered decks

### LINKED ISSUE #3391: "Create Copy" selects the wrong target deck for cards in filtered decks
I am using Anki version 24.06.3. As far as I remember, this behavior has been consistent since the feature was first introduced. The issue doesn't seem to occur when using "Create Copy" on a note that is not in a filtered deck, <s>or when opening the Add window from the Reviewer (either by using the Add button or the "A" keyboard shortcut), whether studying "normal" or filtered decks</s>.

**Steps to reproduce:**

1. Load a card that is currently in a filtered deck (either using the Reviewer or the Browser).
   <p></p>
2. Use "Create Copy" (Ctrl+Alt+E).
   <p></p>
3. **Expected behavior:** When the Add window opens, Anki automatically selects the correct Target Deck.  
   **Actual behavior:** Anki selects the wrong deck.
   <p></p>

singurty:
Fixes #3391 and #3338

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -194,6 +194,7 @@ Gregory Abrasaldo <degeemon@gmail.com>
 Taylor Obyen <162023405+taylorobyen@users.noreply.github.com>
 Kris Cherven <krischerven@gmail.com>
 twwn <github.com/twwn>
+Shirish Pokhrel <singurty@gmail.com>
 
 ********************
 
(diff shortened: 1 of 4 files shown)


dae:
Thanks for contributing Shirish!

I've given this a test, and noticed it doesn't work correctly when the cursor is not at the bottom. For example, if you have a list of 3 elements, and the cursor is on the first, creating a list and then removing it again will move the cursor to the end. And since the current behaviour does seem to work correctly in some circumstances, this seems like a step in the wrong direction. So I'll cut that part out, and merge the other part of your PR.

Issue #3336, opened 2024-07-31
Expose original position in card info and browser column


dae:
Previous discussion:

- https://forums.ankiweb.net/t/position-info-is-shown-even-when-the-card-is-not-new/38087/6
- https://forums.ankiweb.net/t/how-to-display-original-due-numbers-while-learning-a-deck/47592

Originally reported on https://forums.ankiweb.net/t/how-to-display-original-due-numbers-while-learning-a-deck/47592/4

PR #3471, opened 2024-10-03
Add card position column and always show position in card info

taylorobyen:
Implements #3336

- New position column that displays the original position of the card, supported with cards and notes sortable
image
- Original position now always show in card info
image

### DIFF
--- a/rslib/src/search/note_original_position_order.sql
+++ b/rslib/src/search/note_original_position_order.sql
@@ -7,4 +7,4 @@ INSERT INTO sort_order (nid)
 SELECT nid
 FROM cards
 GROUP BY nid
-ORDER BY COALESCE(odue, due);
+ORDER BY COALESCE(extract_original_position(data), due);
--- a/rslib/src/search/note_original_position_order.sql
+++ b/rslib/src/search/note_original_position_order.sql
@@ -7,4 +7,10 @@ INSERT INTO sort_order (nid)
 SELECT nid
 FROM cards
 GROUP BY nid
-ORDER BY COALESCE(extract_original_position(data), due);
+ORDER BY COALESCE(
+    extract_original_position(data),
+    CASE
+      WHEN type == 0 THEN due
+      ELSE 0
+    END
+  );
--- a/rslib/src/search/note_original_position_order.sql
+++ b/rslib/src/search/note_original_position_order.sql
@@ -1,0 +1,10 @@
+DROP TABLE IF EXISTS sort_order;
+CREATE TEMPORARY TABLE sort_order (
+  pos integer PRIMARY KEY,
+  nid integer NOT NULL UNIQUE
+);
+INSERT INTO sort_order (nid)
+SELECT nid
+FROM cards
+GROUP BY nid
+ORDER BY COALESCE(odue, due);
--- a/rslib/src/storage/sqlite.rs
+++ b/rslib/src/storage/sqlite.rs
@@ -70,6 +70,7 @@ fn open_or_create_collection_db(path: &Path) -> Result<Connection> {
     add_regexp_tags_function(&db)?;
     add_without_combining_function(&db)?;
     add_fnvhash_function(&db)?;
+    add_get_original_position_function(&db)?;
     add_extract_custom_data_function(&db)?;
     add_extract_fsrs_variable(&db)?;
     add_extract_fsrs_retrievability(&db)?;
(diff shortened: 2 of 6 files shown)


rslib/src/storage/sqlite.rs
```diff
@@ -205,6 +206,36 @@ fn add_regexp_tags_function(db: &Connection) -> rusqlite::Result<()> {
     )
 }
 
+/// eg. get_original_position(c.due, c.data) -> i64
+/// Check if c.data contains the position, if not it's still in c.due
+fn add_get_original_position_function(db: &Connection) -> rusqlite::Result<()> {
```
dae, line 211: Similar routines above all use extract_ in the name; using it here too would be consistent.

I also I think this routine would be better off focusing on returning position | null. The caller can coalesce the returned null value to card.due if they want to fall back on it. That said, do we really want to be doing that? Can't we just leave cards without an original position unsorted?

rslib/src/search/note_original_position_order.sql
```diff
@@ -0,0 +1,10 @@
+DROP TABLE IF EXISTS sort_order;
+CREATE TEMPORARY TABLE sort_order (
+  pos integer PRIMARY KEY,
+  nid integer NOT NULL UNIQUE
+);
+INSERT INTO sort_order (nid)
+SELECT nid
+FROM cards
+GROUP BY nid
+ORDER BY COALESCE(odue, due);
```
dae, line 10: This looks incorrect? It's sorting cards based on their due date, not the original position.

dae:
> Original position now always show in card info

How does the user know if the card has an original position or not? Instead of falling back on due if position is unavailable, wouldn't it be clearer to show them separately when they differ?

rslib/src/storage/sqlite.rs
taylorobyen: >The caller can coalesce the returned null value to card.due if they want to fall back on it. That said, do we really want to be doing that? Can't we just leave cards without an original position unsorted?

Unless I'm misunderstanding something, every card will always have an original position. However, the column that stores it will change.
1. User makes card, original position is in the due column
2. User reviews card, due is replaced with a timestamp of next review and original position is pushed to the data column

It makes senses to fallback to the due column since the original position has to be there if it isn't in data column yet.

taylorobyen:
>How does the user know if the card has an original position or not? Instead of falling back on due if position is unavailable, wouldn't it be clearer to show them separately when they differ?

Based off my response to your other comment it makes the most sense to display it the same, regardless of the source. I'm assuming our end users don't care what column the original position is being sourced from as long as it is the actual original position. Now if you're aware of edge cases of the original position not existing in the card's data even after first review then I'll have to handle that, but in my testing I didn't see that occurring.

dae:
- "Position" is the order new cards will appear in (the due #). It's only relevant for new cards.
- "Original position" is assigned when new cards are answered for the first time. It exists solely to restore the due # if the cards are reset to new.
- Anki didn't used to assign an original position, so non-new cards that were answered with older Anki versions will not have an original position either.
- When an original position is missing, and the card is new, it's arguable whether the current position is also the original position. But if the card is not new, then falling back to card.due is not really valid.

The use case I'm imagining: the user may have a mix of new and non-new cards, and want to sort them in the order they originally were in, which means sorting based on coalesce(original, current).

taylorobyen:
> The use case I'm imagining: the user may have a mix of new and non-new cards, and want to sort them in the order they originally were in, which means sorting based on coalesce(original, current).

Isn't this what my current implementation does?

image

taylorobyen:
@brishtibheja Do you have any input on the implementation of this? I saw you were on the original thread. I have an idea of what Damien wants but any clarifying details would be helpful.

taylorobyen:
Added handling of cards without position data
image
image

brishtibheja:
@taylorobyen I think going forward, we should make a distinction between `position` and `due`. In your screenshots, I see due column showing the current position (due #) of the card (and position is showing original position). I have seen users getting confused over why due is not showing them a date. @dae thoughts?

Also, to remind everyone of the use case brought up in the forums, it was mainly focused on creating shared decks while studying the decks. In such a situation, `position` helps us reposition the cards in order while creating them.

Lets say I reset all my cards without original position info, what will `position` show?

taylorobyen:
So, making the distinction would mean an `Original Position` and a `Current Position` column with them also being displayed separately in the card info window? 
- Original Position: Value only shown if able to be parsed from data
- Current Position: Value only shown if it's a new card and this is derived from due

> Lets say I reset all my cards without original position info, what will position show?

A new position is assigned to the card and that position will be shown.

brishtibheja:
> making the distinction would mean an Original Position and a Current Position column

I simply meant not showing current `due #` in due column and instead show this under `position`. I think it is confusing for many reasons, one of them being we already use the word *due* to mean several different things. It makes sense to use due to only mean due date in the UI/documentation.

One thing I'll point out is that the reposition feature should always change the value in `position` and not `due #`. If it doesn't users can be left confused. Same with the two position based sort orders.

As for original due vs current due I think they only differ if you reset all of your cards without  `Restore original position`. I think it makes sense to just show the current position here but I'll defer to dae's wisdom.

dae:
Sorry to keep you waiting, and for not being clear above. Your changes LGTM - thank you!

@brishtibheja, I appreciate that some users get confused by due numbers for new cards. But if we take that out of the due column and only show it in the position column, that means the user needs both a due column and a position column to be able to sort all of their cards. We would either want to include position as a default column (taking up more screen space), or users would need to add the column themselves. I think the latter might end up confusing more users than the dual-purpose due column.

jchwenger:
Thanks a lot for this @taylorobyen, seeing this only now, but I had been [hoping for something like this](https://forums.ankiweb.net/t/how-to-display-original-due-numbers-while-learning-a-deck/47592), super nice!

PR #3474, opened 2024-10-04
Fix ease button alignment

taylorobyen:
Same fix to align the ease buttons, but included a z-index to ensure it's above the other elements. I was unable to recreate the issue as there were no steps to reproduce.

### DIFF
--- a/qt/aqt/data/web/css/reviewer-bottom.scss
+++ b/qt/aqt/data/web/css/reviewer-bottom.scss
@@ -23,11 +23,14 @@ body {
     position: relative;
 }
 
-#middle {
-    position: absolute;
-    left: 50%;
-    transform: translateX(-50%);
-    z-index: 5;
+#innertable > tbody > tr {
+    display: flex;
+    align-items: stretch;
+    justify-content: center;
+}
+
+#innertable > tbody > tr > td {
+    flex: 1;
 }
 
 button {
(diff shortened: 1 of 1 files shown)


dae:
I did not attempt to reproduce the inability to click, and just used the inspector to confirm the div was overlapping other elements.

I'm not a CSS wiz, so it might be wrong here, but simply adjusting the z-index feels pretty hacky, and I'm concerned it might be setting us up for more problems down the road. On the other hand, the answer buttons are likely going to be significantly refactored in the future, so a quick and dirty fix might be justified here. Anyone else want to chime in?

taylorobyen:
I came up with a better solution that yields the same results:

French:
image
image

Belarusian
image
image

dae:
Thank you for humoring me Taylor!

dae:
Sadly (and no fault of your own), this appears to have caused a different regression: https://forums.ankiweb.net/t/anki-24-10-beta/49989/258

Lots of people use add-ons that modify the bottom buttons, so I fear this may end up causing a fair bit of trouble, and I may need to revert it again :(

taylorobyen:
That's unfortunate, it seems like it'll be difficult to change anything UI related if small changes like this cause so many issues to 3rd party plugins. I'm curious how the Svelte transition of these pages will go.

dae:
A rewrite in Svelte will of course break things too, so we can't avoid it forever. But for now, this was a relatively minor bugfix, and living without it for now seems preferable to the breakages we'll cause.

taylorobyen:
Yeah that's fair enough.

PR #3475, opened 2024-10-05
NF: Improve typing of AnkiWebView action

Arthur-Milchior:
There are only two types of actions. The code does not seems to consider this to be extandable, and the changed variable is private, so I would expect it relatively acceptable to change the way it work. This would improve the typing.

If this is rejected, at least, the callback should be noted as optional, so that the IDE does not complain that `if cb` can not be false.

==========

I should note that `./ninja format` failed with 

```
failed: format:dprint
[dprint-plugin-prettier]
Shutting down due to error: failed to fill whole buffer
Shutting down due to error: failed to fill whole buffer
Error resolving plugin https://plugins.dprint.dev/prettier-0.46.1.json: Failed plugin schema verification. This may indicate you are using an old version of the dprint CLI or plugin and should upgrade.: The plugin schema version was 5, but expected 4.
Failed with code Some(1): /home/milchior/anki/out/node_modules/.bin/dprint fmt
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.07s
```

I don't understand why easily, but it seems that the py files still were formatted correctly

### DIFF
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -8,9 +8,9 @@ import json
 import os
 import re
 import sys
-from collections.abc import Callable, Sequence
+from collections.abc import Callable
 from enum import Enum
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any, Optional, cast
 
 import anki
 import anki.lang
@@ -285,7 +285,7 @@ class AnkiWebView(QWebEngineView):
         self.onBridgeCmd: Callable[[str], Any] = self.defaultOnBridgeCmd
 
         self._domDone = True
-        self._pendingActions: list[tuple[str, Sequence[Any]]] = []
+        self._pendingActions: list[Callable[[], None]] = []
         self.requiresCol = True
         self.setPage(self._page)
         self._disable_zoom = False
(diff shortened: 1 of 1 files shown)


dae:
Unsure about that error, but it looks like mypy is failing with a different one (maybe Action needs to be qualified with the class name?)

Arthur-Milchior:
After pulling a fresh copy of anki, ninja works again.

Arthur-Milchior:
Actually, just using a callable seems fine enough

dae:
That's more elegant - thanks!

glutanimate:
It looks like the change here [broke Image Occlusion Enhanced](https://forums.ankiweb.net/t/anki-24-10-beta/49989/202?u=glutanimate). Although I'm not a fan of holding back thoughtful code quality and typing improvements like this for the sake of add-ons, would it be possible to revert this change for now? IOE is on its slow way out, but there is still a large of proportion of Anki users using it, and since this change is largely DX-focused, I think it's justifiable to hold it back in order to avoid users not being able to upgrade due to IOE incompatibility.

FWIW, I created an issue on IOE's tracker to move away from the reliance on these APIs, but it will likely take a bit of time for me to look into this (not to mention for any change to actually reach users), so a revert or switch to a more conservative refactoring for now would be much appreciated :pray:  https://github.com/glutanimate/image-occlusion-enhanced/issues/316

PR #3476, opened 2024-10-05
NF: `assert mw` to improve typing

Arthur-Milchior:
My looking at some code, the IDE showed a lot of warning that we accessed a value that is not defined on None. Indeed `mw` could be None.

By adding those assert, the IDE know that `mw` is non null. If it were not the case, a null pointer exception would be called a few lines later anyway.

I did it only for `mw`, as it's a very special variable in anki and my goal was not to remove all ide's null warning.

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -238,6 +238,8 @@ class AddonManager:
     def loadAddons(self) -> None:
         from aqt import mw
 
+        assert mw
+
         broken: list[str] = []
         error_text = ""
         for addon in self.all_addon_meta():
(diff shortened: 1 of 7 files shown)


dae:
Some associated discussion: https://github.com/ankitects/anki/issues/3317
And a related PR: https://github.com/ankitects/anki/pull/3355

One potential alternative to adding these asserts is to change the typing of mw to claim it's non-optional (akin to lateinit in Kotlin). It's a lie, but the code that accesses mw is typically only running after the main window has been created. What do you think?

Arthur-Milchior:
I'm also perfectly fine with that.
another idea I had, similar to the way Kotlin works, was to have a function `require_mw`.  That returns the mw and fails if it's null. But that would be far too much work for what it is

Issue #3472, opened 2024-10-04
Deck option changes not detected until focus is lost


dae:
Originally reported on https://forums.ankiweb.net/t/anki-24-10-beta/49989/6

> When closing the Deck Options screen, the Discard changes? dialog is not shown if changes were made only by directly interacting with an input field (either using the keyboard or the mouse, but not through UI buttons).

Currently the Python code is expecting a synchronous return from deckOptionsPendingChanges(), so we can't just throw in an await tick(). We might need to switch to using a bridge command or similar instead.

dae:
@Arthur-Milchior heads up, in case you feel like tackling this

Arthur-Milchior:
That's frustrating. The state is updated only on `blur` which is async operation. So forcing `blur` from python is a bad idea. Unless, once blur is done, the js sends a pycmd to python to tell it to close. That can be done, but seems uselessly complex for such a bug.

Checking whether the currently active element had changed since its initialization does not seems easily doable, since this element is a basic HTML input element, which don't store its default value.

Updating the state each time the value change would be more costly. Probably acceptable on any device today, but that's not elegant. And worse, it would require to change a lot of input field so that instead of `on:blur` it actually listen to change in the input value

PR #3477, opened 2024-10-05
Deck options warn users if they are in a text field

Arthur-Milchior:
The state is updated only when the field is left. This means that `deckOptionsPendingChanges` return False when the UI is updated and the update was not propagated to the state.

This means any local change could be lost if the user press "esc".

With this commit, the user will get asked to confirm they want to leave even if they ended up not making any changes, as long as they are focused in an input field.

Here are the alternative considered.

1. The `save` function avoid this issue by first calling `blur` to remove the focus to the text field. This is a DOM operation, it's async, so `save` wait for one tick.

We can't use async operation, the expressions needs to return a value tha can be encoded as JSON, and a promise can't be returned to python.

2. having js explicitly send the answer with `pycmd`. The added complexity does not seems worthwhile.

3. reflecting all changes from the UI immediately to the state.
4. having a method that checks whether the current value is equal to the initial value of the field

Both case requires deep change to the UI. It's not clear to me that this bug is worth the added complexity.

5. Do nothing. After all, the current state is not that bad. Most change are still preserved by the recent improvement to the deck options. And if the previous state has been acceptable for a long time, current state is certainly acceptable too.

Fixed: #3472

### DIFF
--- a/ts/routes/deck-options/[deckId]/+page.svelte
+++ b/ts/routes/deck-options/[deckId]/+page.svelte
@@ -13,7 +13,14 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     globalThis.anki ||= {};
     globalThis.anki.deckOptionsPendingChanges = () => {
-        return data.state.isModified();
+        if (data.state.isModified()) {
+            return true;
+        }
+        // The state does not know of any modified data.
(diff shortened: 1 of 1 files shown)


dae:
Checking if the user is in a text field is a creative hack :-) I agree we're into diminishing returns here, but since Hikaru has put in the effort in #3478 to switch to a bridge command, I'll opt for that approach instead.

PR #3478, opened 2024-10-06
Fix issues with 'Discard changes' confirmation dialog

### LINKED ISSUE #3472: Deck option changes not detected until focus is lost

hikaru-y:
This is a follow-up to #3410.

Fixes #3472

Also fixes a few other issues with the 'Discard changes' confirmation dialog in the Deck Options window.

- The confirmation dialog should always appear when there are any pending changes, but it does not appear in the following cases:
  - When you have added or cloned a preset without making any changes to the existing presets
  - When you have changed the global options
  - When you have changed “This deck” or “Today only” in the “Daily Limits” section
- Once a user has changed any of the options (except the global options), the confirmation dialog appears, even if the user has changed everything back to the original values.
  - I think this is probably not the behavior most users would expect.
- The `parent` argument is not set when calling `ask_user_dialog()`, which causes a memory leak.

### DIFF
--- a/qt/aqt/deckoptions.py
+++ b/qt/aqt/deckoptions.py
@@ -92,6 +92,7 @@ class DeckOptionsDialog(QDialog):
                         QMessageBox.StandardButton.Discard,
                         (tr.adding_keep_editing(), QMessageBox.ButtonRole.RejectRole),
                     ],
+                    parent=self,
                 )
             else:
                 self._close()
--- a/qt/aqt/deckoptions.py
+++ b/qt/aqt/deckoptions.py
@@ -62,42 +62,38 @@ class DeckOptionsDialog(QDialog):
         if cmd == "deckOptionsReady":
             self._ready = True
             gui_hooks.deck_options_did_load(self)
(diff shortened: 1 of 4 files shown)


hikaru-y:
I noticed that #3477 had already been pushed just after I opened this PR. However, I will leave this PR open as it contains fixes for issues other than #3472.

dae:
Thanks Hikaru - always a pleasure to see your TypeScript contributions!

Arthur-Milchior:
Thanks Hikaru for correcting the feature I imperfectly implemented.
Plus, it's quite instructive. I don't usually do any js/ts, so I am just discovering that the type system with Required and Pick is quite more expressive than I thought.

PR #3479, opened 2024-10-06
Add "Show on leech tooltips" review preference

KolbyML:
# The problem
I have been using AnkiMobile for over a year now, and Anki longer then that. I find AnkiMobile makes it a lot easier to focus on repping my cards, whereas I get easily distracted on PC. On AnkiMobile I also disabled "Show remaining card count" which massively helped me not get distracted.

This brings me to my last major pain point using AnkiMobile. I have leech threshold set low, but every time `onLeech()` is called a Tooltip pops up which tells me `onLeech()` was called this distracts me in a few ways.

- The tooltip on AnkiMobile blocks visibility of my next card, so I have to wait for the tooltip to go away to see the important part of my next card which is covered by the tooltip, which becomes tedious when you see it every day.
- My main issue with the tooltip is that it gives me major confirmation basis as in I think when I am repping my cards, my leech threshold suspends a lot more cards then it actually does, which is very distracting. I think, wow, I am failing all my cards, but in reality, barely any cards are actually being suspended. So due to confirmation basis I think `onLeech` is being called for over 50% of cards, when reality it is only 5% of cards which is find reasonable. This causes me to hesitate repping my cards as I am scared of seeing the `onLeech()` tooltip again

# Solution

Add a review preference to disable the `onLeech()` tooltips from showing to remove the major distraction/(pain point) I have been having with anki for months now to where it is just getting more and more tedious slowly over time to the point I made this PR.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -196,6 +196,7 @@ Kris Cherven <krischerven@gmail.com>
 twwn <github.com/twwn>
 Shirish Pokhrel <singurty@gmail.com>
 Park Hyunwoo <phu54321@naver.com>
+KolbyML <31669092+KolbyML@users.noreply.github.com>
 
 ********************
 
--- a/ftl/core/preferences.ftl
+++ b/ftl/core/preferences.ftl
@@ -23,6 +23,7 @@ preferences-show-learning-cards-with-larger-steps = Show learning cards with lar
 preferences-show-next-review-time-above-answer = Show next review time above answer buttons
 preferences-spacebar-rates-card = Spacebar (or enter) also answers card
 preferences-show-play-buttons-on-cards-with = Show play buttons on cards with audio
+preferences-show-on-leech-tooltips = Show on leech tooltips
 preferences-show-remaining-card-count = Show remaining card count
 preferences-some-settings-will-take-effect-after = Some settings will take effect after you restart Anki.
 preferences-tab-synchronisation = Synchronization
--- a/proto/anki/config.proto
+++ b/proto/anki/config.proto
@@ -55,6 +55,7 @@ message ConfigKey {
     SHIFT_POSITION_OF_EXISTING_CARDS = 24;
     RENDER_LATEX = 25;
     LOAD_BALANCER_ENABLED = 26;
+    HIDE_ON_LEECH_TOOLTIPS = 27;
   }
   enum String {
     SET_DUE_BROWSER = 0;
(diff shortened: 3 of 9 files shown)


dae:
I appreciate you taking the time to include a PR with your suggestion, but as this is the first time I've seen such a request, I'm not sure there's enough demand here to justify its inclusion.

Ideally, we shouldn't be adding new options to solve pain points. What if the leech message were moved to the bottom of the screen instead, so it doesn't obscure the card content?

KolbyML:
> Ideally, we shouldn't be adding new options to solve pain points. What if the leech message were moved to the bottom of the screen instead, so it doesn't obscure the card content?

Of course, this is better than nothing. 

Tooltips could be considered in-app notifications, so I would like an option to remove notifications cause I am not that interested in them.

Of course, I don't have much leverage, and I won't be mad if the PR isn't merged. What you said is reasonable, and this isn't my project.

I think I might just try and figure out how to mod the AnkiMobile.ipa to remove tooltips or see how hard it would be to port the opensource Anki project to iOS itself. Of course, I might run out of motivation before I get a working product and give up, but I am still annoyed enough after over a year of tooltips to give it a shot.

dae:
I appreciate you being understanding. How about we meet halfway - I'll add a check for a hidden option, which is minimally disruptive. You can set with the debug console in the desktop:

```
mw.col.set_config("amLeechPopup", False)
```

Please let me know how it goes in the next beta: https://docs.ankimobile.net/beta-testing.html

KolbyML:
As long as I can disable the tooltip on AnkiMobile I don't mind how it is implemented. Thank you very much @dae I really appreciate it :D. Hopefully I can contribute something meanful to Anki in the future to repay you for your kindness 🙏.

I will close the issue in favor of the middle ground you proposed.

KolbyML:
> I appreciate you being understanding. How about we meet halfway - I'll add a check for a hidden option, which is minimally disruptive. You can set with the debug console in the desktop:
> 
> ```
> mw.col.set_config("amLeechPopup", False)
> ```
> 
> Please let me know how it goes in the next beta: https://docs.ankimobile.net/beta-testing.html

Do you want me to make a PR for this?

dae:
Sorry, I'm not sure what you're asking here. The change is already in the AnkiMobile beta. Were you able to confirm it's fixed for you?

KolbyML:
> Sorry, I'm not sure what you're asking here. The change is already in the AnkiMobile beta. Were you able to confirm it's fixed for you?

Oh my bad I missed it xD. I just tested it and it works thank you very much <3

PR #3480, opened 2024-10-06
Add .venv to gitignore

voczi:
Not sure what the typical Anki developer uses as a dev environment. But I like to have a venv for every Python project so I can easily get code completion and such, without having to worry about local system dependency conflicts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ node_modules
 /extra
 yarn-error.log
 ts/.svelte-kit
+/.venv


dae:
You can put anything you want in extra/ already, or you can use out/pyenv. I don't think we should be adding new ignored files if they're not used as part of Anki's build process / not super common.

voczi:
VS code python extension uses a default venv path set to source folder root + .venv

dae:
Ok, that does change my opinion somewhat. But presumably there's also a way to tell VS Code to look elsewhere for a given project though - perhaps we should have the Anki repo just defaulting to out/pyenv instead?

dae:
Happy to re-open if you'd like to continue the conversation.

PR #3483, opened 2024-10-07
Fix editor loses focus when toggling list if list item empty

cy7sh:
Fixes https://github.com/ankitects/anki/issues/3338

I tried to fix the issue with https://github.com/ankitects/anki/pull/3465. Can you please check if this works?


dae:
It does appear to work, but I'm a little worried about unintentional side-effects, as the focus handling change is not limited to the list buttons. Perhaps instead of changing focus(), we could do the blur/focus only when toggling list state, by e.g. making a new CommandIconButton for that case?

@hikaru-y @abdnh any thoughts?

abdnh:
This pattern is used in refocus() and it doesn't seem to be causing any issues.

dae:
Ok, let's see how we go.

PR #3486, opened 2024-10-09
Enable strict_optional in aqt/. and aqt/browser

bpnguyen107:
Related to #3317

I mostly added asserts to get rid of the mypy errors

I couldn't figure out how to get around the error from this line though, so I did type: ignore
https://github.com/bpnguyen107/anki/blob/b39de459b4942686d8301b2aff546560a3431ad5/qt/aqt/browser/table/table.py#L652

### DIFF
--- a/qt/aqt/browser/table/table.py
+++ b/qt/aqt/browser/table/table.py
@@ -77,7 +77,7 @@ class Table:
         return self._len_selection
 
     def has_current(self) -> bool:
-        return self._view.selectionModel().currentIndex().isValid()
+        return self._selection_model().currentIndex().isValid()
 
     def has_previous(self) -> bool:
         return self.has_current() and self._current().row() > 0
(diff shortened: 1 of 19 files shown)


qt/aqt/browser/table/table.py
```diff
@@ -358,11 +363,17 @@ def _update_font(self) -> None:
...
+        assert self._view is not None
+        vh = self._view.verticalHeader()
+        assert vh is not None
+        vh.setDefaultSectionSize(curmax + 6)
 
     def _setup_headers(self) -> None:
+        assert self._view is not None
         vh = self._view.verticalHeader()
-        hh = self._view.horizontalHeader()
+        assert vh is not None
```
bpnguyen107, line 375: Yeah, I originally had the asserts the other way, but I kept getting an AssertionError on these asserts for the vertical header. For some reason, the `vh` object was being evaluated as falsy. So I decided to change all the asserts to be `is not None` instead in case that behavior was the same for any other Qt objects.

PR #3489, opened 2024-10-11
Enable strict_optional for aqt/data, aqt/forms, aqt/import_export

bpnguyen107:
Related to #3317 

aqt/data and aqt/forms had no mypy errors

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -187,7 +187,7 @@ Christian Donat <https://github.com/cdonat2>
 Asuka Minato <https://asukaminato.eu.org>
 Dillon Baldwin <https://github.com/DillBal>
 Voczi <https://github.com/voczi>
-Ben Nguyen <105088397+bpnguyen107@users.noreply.github.com>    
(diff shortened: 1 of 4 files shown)


CONTRIBUTORS
```diff
@@ -187,7 +187,7 @@ Christian Donat <https://github.com/cdonat2>
 Asuka Minato <https://asukaminato.eu.org>
 Dillon Baldwin <https://github.com/DillBal>
 Voczi <https://github.com/voczi>
-Ben Nguyen <105088397+bpnguyen107@users.noreply.github.com>    
+Ben Nguyen <105088397+bpnguyen107@users.noreply.github.com>
```
dae, line 190: Are you finding you need to do this every time?

dae:
Thanks Ben!

CONTRIBUTORS
bpnguyen107: I realized the check kept failing because I never changed my email in git config to match that email 😅 It should be good moving forward!

PR #3492, opened 2024-10-12
Hide default deck when empty

### LINKED ISSUE #3334: Hide default deck by default
Perhaps a simpler solution is to:

- remove the exception that shows the default deck when there are no other decks
- hide it from the browser sidebar unless it's empty
- add a new, normal deck like 'my first deck' when new collections are created, or show 'no decks' until the user adds/imports a deck

Originally reported on https://forums.ankiweb.net/t/hide-default-deck-by-default/47663

taylorobyen:
Closes  #3334 

Changes
- Default deck hidden when empty and instead displays "No Decks"
image
- Deck preset counts ignore default deck if empty

I have yet to implement the proper translation for the "No Decks," popup, wanted to first get a pulse check if this was still the direction we wanted to go with this. 

I have a few additional items I would like to add that weren't specifically mentioned:
- When add menu clicked notify user to create a deck and open up the deck creation window instead
- When deck is created when currently 0 decks automatically select it as the current deck (more intuitive when opening browser than having a blank ghost deck selected)


brishtibheja:
Just voicing my opinion on this, I think it would be nicer to have a CTA like "Add a Deck" instead of "No Decks". There are the buttons on bottom saying the same, but IMO "No Decks" is a bit uninformative. Still looks great!

rslib/src/deckconfig/update.rs
```diff
@@ -103,6 +103,9 @@ impl Collection {
     fn get_deck_config_use_counts(&self) -> Result<HashMap<DeckConfigId, usize>> {
         let mut counts = HashMap::new();
         for deck in self.storage.get_all_decks()? {
+            if deck.id == DeckId(1) && self.storage.deck_is_empty(deck.id)? {
```
dae, line 106: Your approach is perfectly valid, but you can also do

```suggestion
            if deck.id.0 == 1 && self.storage.deck_is_empty(deck.id)? {
```

rslib/src/decks/tree.rs
```diff
@@ -256,8 +241,10 @@ impl Collection {
         let decks_map = self.storage.get_decks_map()?;
 
         add_collapsed_and_filtered(&mut tree, &decks_map, timestamp.is_none());
+
         if self.default_deck_is_empty()? {
-            hide_default_deck(&mut tree);
+            tree.children
+                .retain(|n| n.deck_id != 1 || !n.children.is_empty());
```
dae, line 247: The default deck may not be at the top level - you need to traverse the tree, like hide_default_deck() was doing.

dae:
The deck list is shared by all clients, so any CTA needs to be worded in a client-agnostic way, like "Add a deck to get started"

rslib/src/decks/tree.rs
taylorobyen: Add a check for the entire tree: [1dd6b36](https://github.com/ankitects/anki/pull/3492/commits/1dd6b3678303f8e33a6cddff376a1fac97dbeba8)

taylorobyen:
CTA is added:
image

taylorobyen:
When attempting to add cards, this popup appears to notify the user to create their first deck:
image

It then opens the create deck dialogue for them:
image

The add dialogue is then opened with the newly created deck selected:
image

dae:
This involved a few more UI tweaks than I was hoping for at the moment. There's plenty of room for improvement here still, but I don't think we should be tackling much on the UI end until we can get migrated over to shared Svelte code.

In the mean time,  some suggestions for simplifying this a bit:

Automatically opening the add deck dialog doesn't teach the user how to add a deck so they know how to add a second one in the future. It also doesn't teach them what a deck is, or what a good name would be. And the CTA pushes the user towards adding an empty deck, when the majority of users will be starting out with a shared deck instead.

Instead, I think we should:

1) change the CTA to 'Add a shared or empty deck to get started'
2) make it text, not a button, so the user goes looking for the buttons below
3) if the user clicks on 'add' before they've added any decks, instead of a pop-up, automatically add a 'my first deck' (a normal one, not id=1). I suggest is_deck_available() by moved into the Rust layer, and converted to ensure_deck_available(). It can then take care of the check and update in one go, and the three clients can then invoke it as they open the add screen.

WDYT?

dae:
My suggested CTA is intended to nudge the users towards the buttons at the bottom. My thinking was that if user instead clicks Add on top thinking that's what they need to do, they probably don't want to be thinking about a deck name. But instead of automatically adding one, simply repeating the CTA in a pop-up so the user must find the buttons could work too.

taylorobyen:
Will open this in the future if I find more time.

PR #3442, opened 2024-09-26
Feat/Easy Days

L-M-Sherlock:
close #3116

Help and suggestion wanted:
1. How to convert the interval into weekday when considering the `next day starts at`?
2. Where is a good place to store `easy_days_percentages`?
3. Should we make it configurable in preset-level? It's annoying to set easy days in preset-level if the user has a lot of preset.

---

2024-10-08 update: 

image

PR #3495, opened 2024-10-13
Fix 'Discard changes' dialog appearing even when no changes are made

### LINKED ISSUE #3491: Deck options appear changed from start
https://forums.ankiweb.net/t/anki-24-10-beta/49989/166

@hikaru-y it appears to be the absence of limits.newToday/reviewToday in the current preset vs the original. Curiously when I hit esc instead of using the x on the window, no confirmation pops up. Attaching a repro case provided by a user.

[collection-2024-10-12@10-58-10.zip](https://github.com/user-attachments/files/17349152/collection-2024-10-12%4010-58-10.zip)

hikaru-y:
Fixes #3491

> Attaching a repro case provided by a user.
> collection-2024-10-12@10-58-10.zip

I was unable to reproduce the issue using the provided collection or newly created profiles. However, I managed to reproduce it with some of my existing profiles, so I investigated the cause.

> it appears to be the absence of limits.newToday/reviewToday in the current preset vs the original.

Firstly, I don't think the absence of `limits.newToday/reviewToday` should be the cause of the issue.

When working on #3478, I recognized that some of the properties of `limits` could change during the initialization of the `DailyLimits` component. For `DailyLimits`, all missing properties in the orginal config default to `undefined`. Therefore, in #3478, properties with the value `undefined` are excluded when comparing orginal and current to prevent false positives.

After running `git bisect`, I found that the issue has been introduced in #3442.

The issue is caused by the fact that the `EasyDays` component, like `DailyLimits`, may modify the config object during its initialization.

https://github.com/ankitects/anki/blob/f00211df359f8214e9e97688776876a724f59266/ts/routes/deck-options/EasyDays.svelte#L18-L20

#### Approach in this PR

Retrieve the original configs asynchronously after all components are mounted, allowing child components to adjust the config during their initialization phase.

-----------

> Curiously when I hit esc instead of using the x on the window, no confirmation pops up.

From testing, it appears that this only happens if the user never interacts with the webview (e.g., clicking on the webview or pressing a key) from the time the window is opened until the `Esc` key is pressed. In this case, AnkiWebView's `onEsc()` is not called, and DeckOptionsDialog's `reject()` is called directly. As a result, `check_pending_changes()` is not executed.

I don't think this is something that needs to be addressed, because as a practical matter it's impossible for a user to change deck options without interacting with the webview at all, and the expected behavior is for the confirmation dialog not to pop up if no changes have been made.

Apart from the above, I found another issue while investigating.

Since #3410, we use `evt.accept()` instead of `closeEvent()` of the base class when closing the window, but it seems that `evt.accept()` does not trigger `reject()` afterwards. As a result, the cleanup process within `reject()` does not run. You can confirm this issue in Anki 24.10 beta (all of 1, 2, and 3) by the fact that if you close the window without saving the config, the geometry is not saved.

Approach in this PR
Call `closeEvent()` of the base class instead of `evt.accept()`.

### DIFF
--- a/qt/aqt/deckoptions.py
+++ b/qt/aqt/deckoptions.py
@@ -69,8 +69,7 @@ class DeckOptionsDialog(QDialog):
 
     def closeEvent(self, evt: QCloseEvent) -> None:
         if self._close_event_has_cleaned_up:
-            evt.accept()
-            return
+            return super().closeEvent(evt)
         evt.ignore()
         self.check_pending_changes()
 
--- a/ts/routes/deck-options/[deckId]/+page.svelte
+++ b/ts/routes/deck-options/[deckId]/+page.svelte
@@ -16,7 +16,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     globalThis.anki.deckOptionsPendingChanges = async (): Promise<void> => {
         await commitEditing();
         if (bridgeCommandsAvailable()) {
-            if (data.state.isModified()) {
+            if (await data.state.isModified()) {
                 bridgeCommand("confirmDiscardChanges");
             } else {
                 bridgeCommand("_close");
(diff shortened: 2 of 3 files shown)


dae:
Great work Hikaru - the new approach is cleaner, and your research and commit message is excellent. Apologies for leading you up the wrong path with the limited triaging I did prior to posting this :bow:

PR #3496, opened 2024-10-14
Fix/only let FSRS take over short-term schedule when steps are empty

L-M-Sherlock:
source: https://forums.ankiweb.net/t/fsrs-5-1d-scheduling-and-learning-steps/50242/72?u=l.m.sherlock

In previous implementation #3375, FSRS will take over the scheduling when the (re)learning steps run out, which is unintended.

However, someone wishes for the previous one: https://forums.ankiweb.net/t/fsrs-5-1d-scheduling-and-learning-steps/50242/77?u=l.m.sherlock

Should we add an extra option to support this case?

### DIFF
--- a/rslib/src/scheduler/fsrs/retention.rs
+++ b/rslib/src/scheduler/fsrs/retention.rs
@@ -63,7 +63,7 @@ impl Collection {
                         .is_ok()
                 },
             )?
-            .clamp(0.75, 0.95))
+            .clamp(0.7, 0.95))
     }
 
     pub fn get_optimal_retention_parameters(
--- a/rslib/src/scheduler/states/learning.rs
+++ b/rslib/src/scheduler/states/learning.rs
@@ -49,7 +49,10 @@ impl LearnState {
         } else {
             let (minimum, maximum) = ctx.min_and_max_review_intervals(1);
             let (interval, short_term) = if let Some(states) = &ctx.fsrs_next_states {
-                (states.again.interval, states.again.interval < 0.5)
+                (
+                    states.again.interval,
+                    ctx.steps.is_empty() && states.again.interval < 0.5,
+                )
             } else {
                 (ctx.graduating_interval_good as f32, false)
             };
@@ -91,7 +94,10 @@ impl LearnState {
         } else {
             let (minimum, maximum) = ctx.min_and_max_review_intervals(1);
             let (interval, short_term) = if let Some(states) = &ctx.fsrs_next_states {
-                (states.hard.interval, states.hard.interval < 0.5)
+                (
+                    states.hard.interval,
+                    ctx.steps.is_empty() && states.hard.interval < 0.5,
+                )
             } else {
                 (ctx.graduating_interval_good as f32, false)
             };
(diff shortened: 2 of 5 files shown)


user1823:
Another request:
If w[17] OR w[18] is zero, disable the FSRS short-term scheduler because FSRS won't be able to increase the stability with short-term reviews.

Inspired by https://forums.ankiweb.net/t/fsrs-5-1d-scheduling-and-learning-steps/50242/73

Expertium:
> Should we add an extra option to support this case?

Maybe make a simple command that can be added in the custom scheduling field?

> If w[17] OR w[18] is zero, disable the FSRS short-term scheduler because FSRS won't be able to increase the stability with short-term reviews.

Makes sense.

DerIshmaelite:
> Maybe make a simple command that can be added in the custom scheduling field?

But this would not apply to all presets, if you were to have different presets would it :(

DerIshmaelite:
> > If w[17] OR w[18] is zero, disable the FSRS short-term scheduler because FSRS won't be able to increase the stability with short-term reviews.
> 
> Makes sense.

What about the cases, where the parameters are still 0 even after optimizing, or the cases where no 0 is even added at all (remains at 17 params instead of 19).

image

DerIshmaelite:
> > or the cases where no 0 is even added at all (remains at 17 params instead of 19).
> 
> They are added "behind the scenes", without the user seeing it, which I didn't know at the time when you took this screenshot

But why is the scheduler on and having these 0 values exactly bad? Some decks have these 0 values and I wouldn't want the scheduler disabled.

Expertium:
If those parameters are zero, FSRS won't do anything useful with your same-day reviews data

user1823:
Because if w[17] or w[18] is 0, the interval will not increase in same-day reviews unless the next review (with the same interval) crosses the next day starts at.

DerIshmaelite:
> Because if w[17] or w[18] is 0, the interval will not increase in same-day reviews unless the next review (with the same interval) crosses the "next day starts at" time.

Ahhh so it stays frozen. This makes sense.

mlidbom:
I just want to chime in and say that I would also very much like for the current behavior to remain available somehow. I've only been using this for a short time but I already know I would just hate to see it go away. 

Here's why: For some decks I know from experience that I need a quite small, 10m, first (re)learning step, much smaller than the steps FSRS gives me, but after that it's just fabulous to have FSRS add more learning steps when it can tell that I need them. So I'd love for this behavior to remain available somehow.

I tried removing my steps, and my recall for these decks with just FSRS short term scheduling became much worse. It's not so much that I can't recall at all with the FSRS steps, as that I will fail a card that I cannot recall fast, and to get there I need the first small step it seems.

user1823:
> I just want to chime in and say that I would also very much like for the current behavior to remain available somehow.

This request is tracked in 
- https://github.com/ankitects/anki/issues/3497

L-M-Sherlock:
I suggest merging it firstly and then I will implement the feature in another PR.

mlidbom:
One small suggestion. If it is not much harder, a UI option in the addon would be a great addition to my mind. This keeps the complexity out of Anki proper, while still adding discoverability and simplicity for users of the addon.

PR #3500, opened 2024-10-15
Enable strict_optional for aqt/editor.py

bpnguyen107:
There are a little under 600 mypy errors in 48 files left in the aqt directory, so I'm going to try to work through them file by file.

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -34,7 +34,7 @@ from anki.collection import Config, SearchNode
 from anki.consts import MODEL_CLOZE
 from anki.hooks import runFilter
 from anki.httpclient import HttpClient
-from anki.models import NotetypeId, StockNotetype
+from anki.models import NotetypeDict, NotetypeId, StockNotetype
 from anki.notes import Note, NoteFieldsCheckResult, NoteId
 from anki.utils import checksum, is_lin, is_mac, is_win, namedtmp
 from aqt import AnkiQt, colors, gui_hooks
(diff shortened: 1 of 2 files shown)


qt/aqt/editor.py
```diff
@@ -242,38 +242,35 @@ def addButton(
...
-                QShortcut(  # type: ignore
-                    QKeySequence(keys),
-                    self.widget,
-                    activated=on_hotkey,
-                )
+            QShortcut(  # type: ignore
+                QKeySequence(keys),
+                self.widget,
+                activated=on_hotkey,
+            )
```
bpnguyen107, line 273: I moved this code out of the if-block because the mypy error was `qt/aqt/editor.py:245: error: Function "func" could always be true in boolean context  [truthy-function]`

I think an alternative could be changing the type in the function definition to `func: Callable[[Editor], None] | None`

qt/aqt/editor.py
```diff
@@ -696,7 +704,7 @@ def cleanup(self) -> None:
         # prevent any remaining evalWithCallback() events from firing after C++ object deleted
         if self.web:
             self.web.cleanup()
-            self.web = None
+            self.web = None  # type: ignore
```
bpnguyen107, line 707: I did `# type: ignore` here instead of changing self.web's type to `EditorWebView | None` so that I didn't have to put a bunch of `assert self.web is not None`. However, I can do the latter if that's more desirable.

qt/aqt/editor.py
dae: I'd make the same call :+1:

qt/aqt/editor.py
dae: In cases like this where the property ends up getting set as part of __init__() and unset only on cleanup, it's not very risky, and it avoids the extra assertion boilerplate. :+1:

dae (commented):
Thanks Ben, this all looks good.

PR #3501, opened 2024-10-15
Fully switch to File::set_times() from utime crate

nix6839:
set_times will fail for directories unless the file is opened in read-only mode.

Reference: https://github.com/rust-lang/rust/issues/123883


dae:
TIL. Thank you very much, @nix6839!

PR #3503, opened 2024-10-15
Fully switch to File::set_times() from utime crate

nix6839:
From #3501

Fixed failing on Windows.


dae:
Thanks for looking into this - it's nice we can drop the dependency now.

PR #3505, opened 2024-10-16
Feat/option to enable FSRS short-term scheduler when (re)learning steps run out && speed up features based on simulation

### LINKED ISSUE #3497: New option to enable FSRS short-term scheduler
https://forums.ankiweb.net/t/fsrs-5-1d-scheduling-and-learning-steps/50242/54

The FSRS short-term scheduler is very unlikely to schedule an interval less than 4 hours. If one has to clear all their (re)learning steps to enable the short-term scheduler, it is unlikely to be useful because the Again interval would be too long.

So, for the users who want to try the FSRS short-term scheduler, we have to find a solution that allows them to define their own steps and let FSRS determine if they need more same-day reviews or not.

As suggested in the Forums post, I think that a good solution would be to allow an interested user to enable the short-term scheduler through the debug console.

Or as suggested by Expertium, we can make a simple command that the user can enter into the custom scheduling field.

@L-M-Sherlock

L-M-Sherlock:
close #3497

To enable it, please execute `mw.col._set_enable_fsrs_short_term_with_steps(True)` in [Debug Console](https://docs.ankiweb.net/misc.html#debug-console).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1860,9 +1860,9 @@ dependencies = [
 
 [[package]]
 name = "fsrs"
-version = "1.3.2"
+version = "1.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c8db50c4140167e02aaf968837701c6ac887cf7d6d60e7de58019b5b793a4d25"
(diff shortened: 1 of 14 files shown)


dae:
Thanks Jarrett!

PR #3507, opened 2024-10-16
Feat/Estimated Total Knowledge By Note & Daily Load

### LINKED ISSUE #3446: Feat Daily load and retention by notes stats
Fine. I added the Estimated total knowledge (cards):

image

Daily load and retention by notes require more works to do, so I will consider it in another PR.

_Originally posted by @L-M-Sherlock in https://github.com/ankitects/anki/issues/3425#issuecomment-2365992087_

L-M-Sherlock:
close #3446

preview:

image

image

### DIFF
--- a/rslib/src/stats/graphs/future_due.rs
+++ b/rslib/src/stats/graphs/future_due.rs
@@ -33,7 +33,7 @@ impl GraphsContext {
             have_backlog |= due_day < 0;
             *due_by_day.entry(due_day).or_default() += 1;
             if matches!(c.queue, CardQueue::Review | CardQueue::DayLearn) {
-                daily_load += 1.0 / c.interval as f32;
+                daily_load += 1.0 / c.interval.max(1) as f32;
             } else {
                 daily_load += 1.0;
             }
--- a/rslib/src/stats/graphs/future_due.rs
+++ b/rslib/src/stats/graphs/future_due.rs
@@ -13,6 +13,7 @@ impl GraphsContext {
     pub(super) fn future_due(&self) -> FutureDue {
         let mut have_backlog = false;
         let mut due_by_day: HashMap<i32, u32> = Default::default();
+        let mut daily_load = 0.0;
         for c in &self.cards {
             if matches!(c.queue, CardQueue::New | CardQueue::Suspended) {
                 continue;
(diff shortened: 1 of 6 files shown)


rslib/src/stats/graphs/future_due.rs
```diff
@@ -31,10 +32,16 @@ impl GraphsContext {
             }
             have_backlog |= due_day < 0;
             *due_by_day.entry(due_day).or_default() += 1;
+            if matches!(c.queue, CardQueue::Review | CardQueue::DayLearn) {
+                daily_load += 1.0 / c.interval as f32;
```
dae, line 36:
```suggestion
                daily_load += 1.0 / c.interval.max(1) as f32;
```

We don't want to crash if a card has an interval of 0 set (which is invalid, but could happen with manual edits)

rslib/src/stats/graphs/retrievability.rs
```diff
@@ -19,25 +19,36 @@ impl GraphsContext {
             next_day_at: Default::default(),
         };
         let fsrs = FSRS::new(None).unwrap();
+        let mut note_retrievability: std::collections::HashMap<i64, (f32, u32)> =
```
dae, line 22:
```suggestion
        // note id -> (sum, count)
        let mut note_retrievability: std::collections::HashMap<i64, (f32, u32)> =
```

filipenanclarez:
I have problem here when language interface is Brazilian Portuguese.

The word break make this label confusing.

Please look the screen below (iPhone11):

IMG_0454

Can you check this @L-M-Sherlock ?

L-M-Sherlock:
> The word break make this label confusing.

This is because the text is too long on mobile, causing it to wrap automatically. I'm not sure how to fix this issue. Can anyone help?

filipenanclarez:
> > The word break make this label confusing.
> 
> This is because the text is too long on mobile, causing it to wrap automatically. I'm not sure how to fix this issue. Can anyone help?

Can we break the layout like this?

**Conhecimento Total Estimado:
3.227 cartões
403 notas**

L-M-Sherlock:
> Can we break the layout like this?
> 
> **Conhecimento Total Estimado: 3.227 cartões 403 notas**

The label and value are displayed in a table. It's hard to follow the layout without refactoring it.

PR #3508, opened 2024-10-16
Create a new kind of revlog entry for reschedule cards on change

### LINKED ISSUE #3357: Create a new revlog entry for "Reschedule cards on change"
Currently, the revlog entry for "Reschedule cards on change" is the same as that for "Set Due Date". This is the cause of several issues, including
- https://github.com/ankitects/anki/issues/2998
- https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/442
- https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/434

There must be other issues too that I am not remembering.

The solution is to use a separate revlog entry to distinguish "Reschedule cards on change" and "Set Due Date".

The first step is to decide the column in the revlog table whose value can be changed while minimizing add-on breakage (they may be making assumptions about the possible values).
Possible options: `type`, `ease`, `factor`

IMO, `type` is the only column where a change would appear to be "logically" correct. However, I am not sure if using `type = 5` for "Reschedule cards on change" would lead to significant add-on breakage.

L-M-Sherlock:
close #3357

### DIFF
--- a/rslib/src/scheduler/fsrs/memory_state.rs
+++ b/rslib/src/scheduler/fsrs/memory_state.rs
@@ -120,8 +120,8 @@ impl Collection {
                                         *due = (timing.days_elapsed as i32) - days_elapsed
                                             + card.interval as i32;
                                         // Add a manual revlog entry if the last entry wasn't manual
-                                        if !last_info.last_revlog_is_manual {
-                                            self.log_manually_scheduled_review(
+                                        if !last_info.last_revlog_is_rescheduled {
+                                            self.log_rescheduled_review(
                                                 &card,
                                                 original_interval,
                                                 usn,
@@ -237,7 +237,7 @@ struct LastRevlogInfo {
     last_reviewed_at: Option<TimestampSecs>,
     /// If true, the last action on this card was a reschedule, so we
     /// can avoid writing an extra revlog entry on another reschedule.
-    last_revlog_is_manual: bool,
+    last_revlog_is_rescheduled: bool,
 }
 
 /// Return a map of cards to info about last review/reschedule.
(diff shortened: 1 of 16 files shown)


rslib/src/scheduler/fsrs/memory_state.rs
```diff
@@ -249,18 +250,18 @@ fn get_last_revlog_info(revlogs: &[RevlogEntry]) -> HashMap<CardId, LastRevlogIn
...
         .for_each(|(card_id, group)| {
             let mut last_reviewed_at = None;
-            let mut last_revlog_is_manual = false;
+            let mut last_revlog_is_rescheduled = false;
             for e in group.into_iter() {
                 if e.button_chosen >= 1 {
                     last_reviewed_at = Some(e.id.as_secs());
                 }
-                last_revlog_is_manual = e.review_kind == RevlogReviewKind::Manual;
+                last_revlog_is_rescheduled = e.review_kind == RevlogReviewKind::Rescheduled;
```
dae, line 258: What about older cards that were rated in previous Anki versions. Should we be accepting either manual or rescheduled here?

rslib/src/stats/graphs/reviews.rs
```diff
@@ -38,7 +38,7 @@ impl GraphsContext {
                     count.filtered += 1;
                     time.filtered += review.taken_millis;
                 }
-                RevlogReviewKind::Manual => unreachable!(),
+                RevlogReviewKind::Manual | RevlogReviewKind::Rescheduled => unreachable!(),
```
dae, line 41: You forgot to exclude the rescheduled case at the top, so it's reachable.

rslib/src/scheduler/fsrs/memory_state.rs
L-M-Sherlock: My concern is the user may feel confused if they set due date  for a card and reschedule all cards, they will only see the first revlog entry generated by `set due date`.

rslib/src/stats/graphs/reviews.rs
L-M-Sherlock: aea26e9807f710e3f22aaca9c279d6e4540974ff

rslib/src/scheduler/fsrs/memory_state.rs
dae: This only affects card rescheduling when saving deck options, yes? Won't an upgrading user with previous set due date use see quite different results when rescheduling now, compared to what they were getting previously? Is that intentional?

PR #3512, opened 2024-10-20
Allow setting addons directory with "ANKI_ADDONS" env var

JuneStepp:
This is very helpful for packaging addons, as I've been working on for Nix.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -197,6 +197,7 @@ twwn <github.com/twwn>
 Cy Pokhrel <cy@cy7.sh>
 Park Hyunwoo <phu54321@naver.com>
 Tomas Fabrizio Orsi <torsi@fi.uba.ar>
+June Stepp <git@junestepp.me>
 
 ********************
 
--- a/qt/aqt/profiles.py
+++ b/qt/aqt/profiles.py
@@ -310,7 +310,9 @@ class ProfileManager:
         return path
 
     def addonFolder(self) -> str:
-        return self._ensureExists(os.path.join(self.base, "addons21"))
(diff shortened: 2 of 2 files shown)


dae:
Your goal is to package Anki add-ons as nix packages? I'm afraid that's not something I wish to encourage. Aside from the fact that this will give errors when users try to install add-ons from within Anki (since you presumably plan to point it at a read-only location), it's also likely to lead to errors for users when the nix packages fall out of sync with updates by the add-on authors.

JuneStepp:
That's very fair, but those are more of a Nix problem than an Anki one. The subset of Nix users who go out of their way to configure everything declaritvely are well aware of the problems and already tackle them with many other apps.

With regard to  your specific concerns:
> this will give errors when users try to install add-ons from within Anki 

A small Anki add-on is used to remind users when they've configured add-ons with Nix and send them to the correct documentation. 

> it's also likely to lead to errors for users when the nix packages fall out of sync with updates by the add-on authors.

Update scripts are used to automatically open a Nixpkgs PR for review the same day an add-on is updated.

Overall, I'd argue that this small change shouldn't cause any significant maintenance burden, but will be very helpful to both Nix and other users when want to keep add-ons separate from their profiles and collections data.

dae:
It's true that it's unlikely to be a burden in terms of code changes, but if you make the Anki packages in nix do this automatically, I'm almost certain it will cause user confusion, and people writing to us and add-on authors, which will be a burden.

The change you wish to make is in a part of the code that does not change often. Since this is a change that only nix will benefit from, it seems more appropriate for the nix packages to do it in a patch instead.

JuneStepp:
This sort of thing is never default behavior. Will still go ahead and close the PR, since you disagree on it having a wider user case, though. Thanks for your time!

PR #3516, opened 2024-10-21
Fix calculation of daily load

user1823:
The previous code caused buried cards to be counted as 1 instead of 1/ivl.
Also, moved the code above because we don't want to exclude the buried cards that are due today.

In the add-on, the only cards that are excluded in the calculation are new cards and suspended cards.

https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/42ec95d9578f35da0ed5908570e9513f93a00820/stats.py#L25

### DIFF
--- a/rslib/src/stats/graphs/future_due.rs
+++ b/rslib/src/stats/graphs/future_due.rs
@@ -26,17 +26,14 @@ impl GraphsContext {
                 due - (self.days_elapsed as i32)
             };
 
+            daily_load += 1.0 / c.interval.max(1) as f32;
+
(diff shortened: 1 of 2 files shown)


dae:
Happy with this @L-M-Sherlock?

user1823:
Do we need a round() somewhere? The final result shouldn't have a decimal place, but Jarrett's code didn't have a round() either.

L-M-Sherlock (approved):
LGTM

dae:
Thank you!

PR #3518, opened 2024-10-21
Fix/skip suspended cards in graphs context retrievability

### LINKED ISSUE #3517: Discrepancy between add-on and native "Estimated total knowledge"
@L-M-Sherlock

The native Estimated total knowledge is higher than the add-on.

Native Estimated total knowledge: 5,865 cards / 4,726 notes

Add-on Estimated total knowledge: 5839 cards / 4707 notes

Deck: [Deck.zip](https://github.com/user-attachments/files/17175446/04-24.zip) (Rename to .apkg)

(After importing the deck, enable FSRS with the default parameters. I don't know why the FSRS settings are not exported with the deck.)

L-M-Sherlock:
close #3517

image

### DIFF
--- a/rslib/src/browser_table.rs
+++ b/rslib/src/browser_table.rs
@@ -115,8 +115,8 @@ impl Card {
         } else if self.is_due_in_days() {
             Some(
                 TimestampSecs::now().adding_secs(
-                    ((self.original_or_current_due() - timing.days_elapsed as i32)
-                        .saturating_mul(86400)) as i64,
+                    (self.original_or_current_due() as i64 - timing.days_elapsed as i64)
+                        .saturating_mul(86400),
                 ),
             )
         } else {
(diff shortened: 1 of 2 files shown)


user1823:
> There still exists a minor discrepancy.

It is likely because the add-on counts all siblings, including the suspended ones.

brishtibheja:
It's not a good idea to do that here but not elsewhere in stats. Also, if you trust in the DSR values, you know some material of the suspended cards. That not being in total knowledge is weird.

I actually keep most of my cards (suspended) even after I've learned them to a very good level (for duplicate checking). Works for yomitan etc. If some users would want to look at total knowledge or other graphs with `is:suspended` cards excluded perhaps can add an option for them? Similar to what we have in "Card Counts".

dae:
First things first, let's make it consistent with the add-on. If you're proposing further changes to both, that's perhaps better discussed separately.

PR #3525, opened 2024-10-23
template_filters: make {{type:}} forward compatible

twwn:
Make Anki treat `{{type:unknown:field}}` field replacements as `{{type:field}}` instead of `{{field}}`.

This way users always get to type and especially: The front side will always render as intended.

The previous behavior is seriously inconvenient for further reasons, including:

* Updated clients aren't all released at the same time, causing an unavoidable mismatch period when a new variant is introduced. Now it causes minimum friction.
* Devs can't readily test new in-dev variants without breaking/inconveniencing their own mobile experience.
* Some users will have (potentially perpetually) outdated Anki clients and likely complain, wasting both dev & deck creator time.

### DIFF
--- a/rslib/src/template_filters.rs
+++ b/rslib/src/template_filters.rs
@@ -31,12 +31,12 @@ pub(crate) fn apply_filters<'a>(
     let mut text: Cow<str> = text.into();
 
     // type:cloze & type:nc are handled specially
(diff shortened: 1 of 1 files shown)


dae:
Seems reasonable, thanks.

Issue #3521, opened 2024-10-22
FSRS parameters reset to SM-2 after changing deck


jcznk:
I'm using Anki 24.10 beta 4. This also occurs in version 24.06.3 (Windows 10, qt6). Testing was done on a new profile.

After changing a card's deck, the FSRS parameters (`Stability`, `Difficulty`, `Retrievability`) are lost and replaced with SM-2 parameters (`Ease`). This seems to affect both the scheduling and sorting. For example, when sorting by `Difficulty Ascending`, these cards are always shown first since they no longer have a `Difficulty`.

In the example below, all reviews were performed using a filtered deck, but the same behavior seems to occur for "regular" review entries.

Parameters and intervals before changing the deck:  
<img src="https://github.com/user-attachments/assets/c5c9961a-0002-4299-9a27-295abaebc7af" height="400">

Parameters and intervals after changing the deck:  
<img src="https://github.com/user-attachments/assets/ebb66905-2ea4-4635-beb1-be26b4fac7ef" height="400">

dae:
That is by design. They are recalculated when the card is reviewed again.

jcznk:
May I ask the reasoning behind this design choice? From my understanding, it seems to negatively impact both the scheduling intervals and the sorting.

brishtibheja:
Ah, I have faced this before but I failed to reproduce it later.

I assume it was done so that memory states need not be recalculated everytime deck is changed? Can we not still keep the previous parameters though so that sorting isn't affected?

___

@jcznk In case you didn't thought of this, clicking optimise will fix this issue. Just make sure the cards aren't suspended.

user1823:
> I assume it was done so that memory states need not be recalculated everytime deck is changed? 

Yes, different decks can have different presets and therefore different parameters. So, the memory states from the previous deck need to be cleared.

> Can we not still keep the previous parameters though so that sorting isn't affected?

Do you mean memory states? We can't keep the previous ones. Otherwise, Anki has no way to know at the time of review if the memory states are calculated using the parameters of the new preset or the old preset.

> From my understanding, it seems to negatively impact both the scheduling intervals and the sorting.

Because the memory states are recalculated just before you answer a card, the scheduling is unaffected. However, stats and sorting are affected.

Recalculating the memory states just after moving the card should be possible. I don't know why Anki doesn't do that. Maybe performance concerns? @dae

jcznk:
[...]

PR #3530, opened 2024-10-26
Fix/buried new cards shouldn't be counted in daily load

### LINKED ISSUE #3529: Burying still affects daily load
@L-M-Sherlock

After https://github.com/ankitects/anki/pull/3516, I expected that burying cards will have no effect on the daily load.

However, even in 24.10 rc 1, burying cards increases the daily load.

Slightly unrelated, but there seems to a difference in rounding of the value between the add-on and native stats. After unburying all cards, Anki showed a daily load of 161 reviews/ day while the add-on showed a daily load of 162 reviews/day.
(With siblings of today's reviews buried, Anki showed a daily load of 166 reviews/ day.)

L-M-Sherlock:
fix #3529, https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/480

New cards don't have memory state, so I exclude them in that way. And it's consistent with FSRS helper.

### DIFF
--- a/rslib/src/stats/graphs/future_due.rs
+++ b/rslib/src/stats/graphs/future_due.rs
@@ -6,7 +6,7 @@ use std::collections::HashMap;
 use anki_proto::stats::graphs_response::FutureDue;
 
 use super::GraphsContext;
-use crate::card::CardQueue;
+use crate::card::{CardQueue, CardType};
 use crate::scheduler::timing::is_unix_epoch_timestamp;
 
 impl GraphsContext {
@@ -26,7 +26,7 @@ impl GraphsContext {
                 due - (self.days_elapsed as i32)
             };
 
-            if c.memory_state.is_some() {
+            if c.ctype != CardType::New {
                 daily_load += 1.0 / c.interval.max(1) as f32;
             }
 
--- a/rslib/src/stats/graphs/future_due.rs
+++ b/rslib/src/stats/graphs/future_due.rs
@@ -6,7 +6,8 @@ use std::collections::HashMap;
 use anki_proto::stats::graphs_response::FutureDue;
 
 use super::GraphsContext;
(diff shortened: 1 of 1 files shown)


user1823:
What if you exclude cards having `c.type = CARD_TYPE_NEW`? This way, daily load will be useful for SM-2 users too.

L-M-Sherlock:
> What if you exclude cards having `c.type = CARD_TYPE_NEW`? This way, daily load will be useful for SM-2 users too.

Because it will be inconsistent with the helper add-on. For details, please see:
- #3521

user1823:
But, in that case (memory states lost due to changing deck), excluding these cards from the daily load makes the calculation inaccurate.

So, I would say that it is better to change the calculation in the helper add-on than making the calculation inccurate in Anki too.

user1823:
If you think that changing the calculation in the add-on would be too difficult (because it is coupled with the other stats), just remove Daily Load from the add-on in 24.10+. Anki already has it and duplicating features is not necessary.

brishtibheja:
Is that necessary, a very tiny percentage of users use FSRS Helper who'll also notice discrepancies like this. The add-on feature is probably useful to those who wouldn't update to 24.10 version any time soon.

L-M-Sherlock:
> The add-on feature is probably useful to those who wouldn't update to 24.10 version any time soon.

The add-on could know which version the user is running. If the version is greater or equal to 24.10, just hide daily load. If not, display it.

user1823:
Because the buried new cards shouldn't be included in any of the stats in this file, it might make sense to exclude those cards in the very beginning, i.e., just after `for c in &self.cards {`.

This won't affect the other stats. But, it might slightly improve the performance and prevent similar bugs in the future.

dae:
Further tweaks to this part of the code might be better done as part of #3379

PR #3536, opened 2024-10-27
Remove legacy import option

### LINKED ISSUE #3493: Remove legacy importer option from GUI
> The error message “Unable to read file. It probably requires a newer version of Anki to import” but I do have the latest version and I have tried uninstalling and reinstalling the app, as well as restarting my computer.

Users sometimes accidentally enable this, and then get stuck. This option was only provided so that the 'special fields' add-on can continue to work without changes. But because it's causing problems, I think we should instead require the add-on to dynamically switch Anki to the old codepath, so that when the add-on is not installed, the new importer is always used.

@ankingmed is the add-on still in regular use, or is it considered deprecated now that AnkiHub exists?

Originally reported on https://forums.ankiweb.net/t/downloaded-anki-deck-wont-open-but-i-have-the-latest-version/50526

abdnh:
Closes #3493

### DIFF
--- a/qt/aqt/forms/preferences.ui
+++ b/qt/aqt/forms/preferences.ui
@@ -612,28 +612,6 @@
         </widget>
        </item>
        <item>
-        <widget class="QGroupBox" name="groupBox_4">
-         <property name="title">
-          <string>preferences_import_export</string>
-         </property>
-         <layout class="QVBoxLayout" name="verticalLayout_11">
-          <item>
-           <widget class="QCheckBox" name="legacy_import_export">
-            <property name="sizePolicy">
-             <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
-              <horstretch>0</horstretch>
-              <verstretch>0</verstretch>
-             </sizepolicy>
-            </property>
-            <property name="text">
-             <string notr="true">Legacy import/export handling</string>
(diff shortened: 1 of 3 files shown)


dae:
If a user has the legacy importer enabled and upgrades, isn't this going to leave them with it stuck on? Perhaps we should be making legacy_import_export() always return False as well, and then the add-on can override it?

abdnh:
> If a user has the legacy importer enabled and upgrades, isn't this going to leave them with it stuck on?

That crossed my mind, but thought maybe it's a rare case.

> Perhaps we should be making legacy_import_export() always return False as well, and then the add-on can override it?

The add-on no longer uses the config option. It patches the new importing path instead: https://github.com/johnpincock/SpecialFields/pull/39/files

Expertium:
https://www.reddit.com/r/Anki/comments/1hfr5e5/no_more_legacy_importexport_handling_in_2411/

> No more legacy import/export handling in 24.11?

> I noticed this since I use u/Glutanimate's Export Deck to HTML and Print to PDF and it relies on the legacy export options. Had to downgrade.

image

abdnh:
Add-ons that need legacy handling can be fixed with minor changes (see the Special Fields add-on)

PR #3537, opened 2024-10-27
Enable strict_optional for aqt/deckbrowser.py

bpnguyen107:
Is there any preference for doing a PR per file as I work through the rest of the aqt mypy modules? I figured it would be easier to review this way, but I can make the PRs bigger if that is better.


dae:
These type changes are fairly easy to review, so I don't mind if you do a number of files at once if that's easier. A single PR that changes everything at once is a bit daunting though, so a few of them would be good. :-)

PR #3541, opened 2024-10-28
Fix flaky tests

### LINKED ISSUE #3353: flaky tests: state_application, unicode_normalization
Hello,

We run tests as part of the nixos build and it would appear some tests are flaky:

```
---- scheduler::answering::test::state_application stdout ----
thread 'scheduler::answering::test::state_application' panicked at rslib/src/scheduler/answering/mod.rs:593:9:
assertion `left == right` failed
  left: Normal(Learning(LearnState { remaining_steps: 2, scheduled_secs: 60, elapsed_secs: 0, memory_state: None }))
 right: Normal(Learning(LearnState { remaining_steps: 2, scheduled_secs: 60, elapsed_secs: 1, memory_state: None }))


failures:
    scheduler::answering::test::state_application
```

```
---- media::check::test::unicode_normalization stdout ----
thread 'media::check::test::unicode_normalization' panicked at rslib/src/media/check.rs:723:13:
assertion `left == right` failed
  left: MediaCheckOutput { unused: ["ぱぱ.jpg"], missing: ["foo[.jpg", "normal.jpg"], missing_media_notes: [NoteId(1581236386334)], renamed: {"は\u{309a}は\u{309a}.jp>
 right: MediaCheckOutput { unused: [], missing: ["foo[.jpg", "normal.jpg"], missing_media_notes: [NoteId(1581236386334)], renamed: {"は\u{309a}は\u{309a}.jpg": "ぱぱ.>


failures:
    media::check::test::unicode_normalization
```

I could reproduce the `state_application` one, running it in a tight loop (grabbing the exact test command from `cargo test -vv` to avoid the cargo overhead) reproduces within 10-60s on my laptop (~1/250 - 1/2000 runs)

That can be worked around by masking the elapsed_time explicitly but it's not great, looking at the code `self.set_elapsed_secs_equal` might interfere with the test if it runs right at a second boundary?

Unfortunately I couldn't reproduce the second one through sheer running it in a loop, either other tests interfere with some global state? or it could be an environment problem, I'm not sure.
I also had a quick look and ぱぱ.jpg apparently shouldn't have been unused but I haven't gone far enough to figure what makes things unused and why it would be the case here...

abdnh:
Closes #3353

## state_application

Adding some set_elapsed_secs_equal() calls is a simple solution.

## unicode_normalization

I couldn't reproduce the issue here so unsure about this fix. But the issue does seem to be about filesystem metadata: `ぱぱ.jpg` is missing from MediaCheckOutput. This can only happen if it was not returned by check_media_folder() at all.

### DIFF
--- a/rslib/io/src/lib.rs
+++ b/rslib/io/src/lib.rs
@@ -8,6 +8,7 @@ use std::fs::FileTimes;
 use std::fs::OpenOptions;
 use std::io::Read;
 use std::io::Seek;
+use std::io::Write;
 use std::path::Component;
 use std::path::Path;
 use std::path::PathBuf;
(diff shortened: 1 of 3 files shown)


dae:
Thanks Abdo. We might need to sync the parent folder as well, but first let's see how this goes.

PR #3552, opened 2024-11-05
change a link

brishtibheja:
Is this used somewhere? Just to be sure, I updated this one too.


dae:
Good catch - it's not used anywhere. I'll just remove it instead.

Issue #3460, opened 2024-09-30
Add 'reverse relative overdueness'; consider making it the default


dae:
> Lastly, if you don’t feel like making this big change right now, at least add Reverse Relative Overdueness (aka Retrievability Descending), as it’s objectively the best. Might be worth naming it Least Overdue First, and renaming Relative Overdueness to Most Overdue First.

See the linked thread for wording discussion.

Originally reported on https://forums.ankiweb.net/t/improving-sort-orders/50081

Expertium:
I still think that long-term we should only have 3-4 sort orders + "Custom" that opens a new window where there would be lots of metrics (stability/difficulty/overdueness/interval length/due date/etc.) and an Ascending/Descending toggle, so that people with oddly specific preferences don't lose their favorite functionality.

brishtibheja:
> I still think that long-term we should only have 3-4 sort orders 

I think designing user experience for a wider (non-nerds) audience involves simplification. I'll agree with you. Also, the mandatory [UXmyth article](https://uxmyths.com/post/746610684/myth-21-people-can-tell-you-what-they-want).

> users may assume it works based on the number of days a card is overdue without regard to the card’s interval. (dae's comment)

In my experience, most users including me, Expertium, and some others still misunderstood `relative overdueness`. Isn't that fine?

I think we should be a bit more task-oriented. The namings here do not convey that `least overdue first` should be used with large backlogs and not `most overdue first`. I think most of the namings should convey what we at least expect the usage to be rather than what the sort is internally doing.

dae:
I'm skeptical we could come up with task-based names for 3-4 sort orders, let alone all the existing ones. Can you provide some examples of what you had in mind?

brishtibheja:
That would depend on what the users who requested the sort orders were trying to do. For example, depending on user goals `ascending intervals` can be renamed to `recently reviewed first` (though it doesn't work that way if you're going through a backlog which then would need addressing).

Some others:

`random` = `increase interleaving` (just an e.g.; that's how some see it but might confuse users more)

`reverse relative overdueness` = `optimise for backlog`

`difficulty ascending` = `easy cards first`

`difficulty descending` = `difficult cards first`

___

I also think it would be useful to remove `relative overdueness` altogether as it doesn't serve much purpose. IIRC this was mainly thought to be useful for backlogs but it actually harms the user with a backlog (imho).

dae:
> difficulty ascending = easy cards first

[...]

PR #3559, opened 2024-11-07
Add descending retrievability

dae:
A partial fix for #3460. Some notes:

- This only adds the extra option for now, and does not change the default yet.
- The 'day', 'deck then day' and 'day then deck' orders sort by day primarily because it was the most efficient in older versions of Anki, and that is no longer as much of a concern. When making the this new sort order the default in the future, we may want to drop 'day' and 'day then deck', and replace 'deck then day' with 'deck then most retrievable first'.
- In the end, I decided not to add the [special treatment of the current day](https://github.com/ankitects/anki/issues/3460#issuecomment-2407533104) for now, as I'd like to see how the simpler approach is received first.

What do you all think?


user1823:
Looks good enough to me.

Just to note, this seems to remove the Relative Overdueness option for SM-2 users.

abdnh (approved):
LGTM

abdnh:
> Just to note, this seems to remove the Relative Overdueness option for SM-2 users.

It's synonymous with ascending retrievability - the now order also works with SM-2.

user1823:
Sorry, I didn't look closely at the code.

rslib/src/storage/card/mod.rs
```diff
@@ -770,15 +773,18 @@ impl fmt::Display for ReviewOrderSubclause {
             ReviewOrderSubclause::EaseDescending => "factor desc",
             ReviewOrderSubclause::DifficultyAscending => "extract_fsrs_variable(data, 'd') asc",
             ReviewOrderSubclause::DifficultyDescending => "extract_fsrs_variable(data, 'd') desc",
-            ReviewOrderSubclause::RelativeOverdueness { today } => {
-                temp_string = format!("ivl / cast({today}-due+0.001 as real)", today = today);
+            ReviewOrderSubclause::RetrievabilitySm2 { today, order } => {
+                temp_string = format!(
+                    "ivl / cast({today}-due+0.001 as real) {order}",
```
user1823, line 778: This is fine. But, for a more accurate sorting, we can calculate the reciprocal and then sort in the opposite direction.

So, it would become something like cast({today}-due as real) / ivl

dae: No objections to a more accurate implementation, but I'm nervous to change working code so close to a release - perhaps this could be fixed in a follow-up.

PR #3571, opened 2024-11-11
Deck options without bridge

Arthur-Milchior:
Thanks to @BrayanDSO I learned we were not supposed to have BridgeCommand anymore. So here I remove them in the deck options

### DIFF
--- a/qt/aqt/deckoptions.py
+++ b/qt/aqt/deckoptions.py
@@ -46,7 +46,6 @@ class DeckOptionsDialog(QDialog):
         addCloseShortcut(self)
 
         self.web = AnkiWebView(kind=AnkiWebViewKind.DECK_OPTIONS)
-        self.web.set_bridge_command(self._on_bridge_cmd, self)
         self.web.load_sveltekit_page(f"deck-options/{self._deck['id']}")
         layout = QVBoxLayout()
         layout.setContentsMargins(0, 0, 0, 0)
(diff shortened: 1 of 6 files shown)


BrayanDSO:
Reference: https://github.com/ankidroid/Anki-Android/issues/14361#issuecomment-1701946364

abdnh (approved):

dae:
Sorry for the delayed review. A couple of questions:

- If 'requireClose' and 'confirmDiscardChanges' are intended only to be used by the deck options as the comment implies, I think they need a more specific name.
- If they're intended to be used in other contexts in the future too, I think closeCurrentWindow might be a clearer name
- Did you consider using a simple JavaScript confirm() instead?

Arthur-Milchior:
Regarding js's `confirm`. Do you want it to also be used in Anki?
If yes, I can try.
I would not have done it without your permission, because it'd be a user-facing change that does not seems to solve any actual user issue. More specifically,  if I did that, the confirmation request window would be different from the one the user usually see when getting asked questions about the remaining of anki.

By the way, I guess the same problem would occur in ankidroid, with `confirm` we'd get a browser confirmation view. While the user is used to the Material dialog.

Regarding `requireClose` and `confirmDiscardChanges`, right now, it's only used by the deck options. I know anki less than you, and I've no idea whether any similar request may eventually be used.
I'd have thought maybe it'd be useful in the NoteEditor, given that it already show the same discard questions. But I see the "close" button is a PyQT button so I don't expect we'd actually need it. 

Anyway, even if we end up with duplication because this feature is used in two views, that does not seems that bad.

dae:
We already use confirm() when the user removes a preset in the deck options, so there's precedent. I believe all platforms have the ability to intercept the confirm() request, and present it as they wish. The desktop does this with javaScriptConfirm()

https://stackoverflow.com/questions/2726377/how-to-handle-a-webview-confirm-dialog

As those methods will only be used for the deck options for now, I think adding 'deckOptions' to them might be a bit clearer.

Arthur-Milchior:
And now it uses `Confirm`. It simplifies the code. Thanks.

qt/aqt/deckoptions.py
```diff
@@ -58,56 +57,31 @@ def _setup_ui(self) -> None:
...
-        elif cmd == "_close":
-            self._close()
+    def ready(self):
+        self._ready = True
+        gui_hooks.deck_options_did_load(self)
 
     def closeEvent(self, evt: QCloseEvent | None) -> None:
-        if self._close_event_has_cleaned_up:
+        if self._close_event_has_cleaned_up or not self._ready:
+            print(f"{self._ready=}, {self._close_event_has_cleaned_up=}")
```
abdnh, line 66:
```suggestion

```

qt/aqt/deckoptions.py
```diff
@@ -58,56 +57,31 @@ def _setup_ui(self) -> None:
...
             return super().closeEvent(evt)
         assert evt is not None
+        print("request js to handle it")
```
abdnh, line 69:
```suggestion

```

qt/aqt/deckoptions.py
```diff
@@ -58,56 +57,31 @@ def _setup_ui(self) -> None:
...
 
-    def _on_bridge_cmd(self, cmd: str) -> None:
-        if cmd == "deckOptionsReady":
-            self._ready = True
-            gui_hooks.deck_options_did_load(self)
-        elif cmd == "confirmDiscardChanges":
-            self.confirm_discard_changes()
-        elif cmd == "_close":
-            self._close()
+    def ready(self):
```
abdnh, line 60: Nitpick: a verbal form is clearer: `set_ready()`

abdnh (approved):

abdnh:
Pylint check is failing due to an unused `ask_user_dialog` import.

dae:
Thanks for seeing this through Arthur!

iamllama:
> Does anyone have problem with discarding changes of presets?
Linux - Using linux archive in github release
1- Open preset
2- Modify the settings
3- Try to close the preset settings page
4- Ask for discard and click yes
5- It doesn’t close the preset window
> https://forums.ankiweb.net/t/anki-25-01-beta/54490/57

Was able to reproduce this, seems like [this call](https://github.com/ankitects/anki/blame/b65fa693dacbd92c0bdd2a8c3800d0827432d108/qt/aqt/mediasrv.py#L590) to `aqt.mw.app.activeWindow` returns `None` sometimes(?) and nothing ends up happening when yes is clicked

abdnh:
No luck reproducing this yet. @iamllama does it make a difference whether you use your keyboard or mouse to close the window?

iamllama:
I can't reproduce it at all on win 10, with kb or mouse

On a ubuntu 24 vm it happens when trying to close with both kb and mouse. But if I keep trying for 2-6 times (random) it works eventually

abdnh:
Oh I see. I only tested on Windows 10.

PR #3572, opened 2024-11-11
NF: explaining the bridge between languages

Arthur-Milchior:
It was, honestly, quite confusing. And if I got it wrong, I guess this code review will be the time for me to learn what was wrong.

### DIFF
--- a/docs/language_bridge.md
+++ b/docs/language_bridge.md
@@ -1,0 +1,84 @@
+Anki's codebase uses three layers.
+
+1. The web frontend, created in Svelte and typescript,
+2. The Python layer and
+3. The core Rust layer.
+
+Each layer can can makes RPC (Remote Procedure Call) to the layers below it. While it should be avoided, Python can also invoke Typescript functions. The Rust layers never make calls to the other layers. Note that it can make RPC to AnkiWeb and other servers, which is out of scope of this document.
+
+In this document we'll provide examples of bridge between languages, explaining:
+
+- where the RPC is declared,
+- where it is called (with the appropriate imports) and
+- where it is implemented.
+
+Imitating those examples should allow you to make call and create new RPCs.
+
+## Declaring RPCs
+
+Let's consider the method `NewDeck` of `DecksServices`. It's declared in [decks.proto](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/proto/anki/decks.proto#L14) as `rpc NewDeck(generic.Empty) returns (Deck);`. This means this methods takes no argument (technically, an argument containing no information), and returns a [`Deck`](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/proto/anki/decks.proto#L54).
+
+Read [protobuf](./protobuf.md) to learn more about how those input and output types are defined.
+
+If the RPC implementation is in Python, it should be declared in the service [frontend.proto](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/proto/anki/frontend.proto#L24C3-L24C66)'s `FrontendService`. RPCs declared in any other services are implemented in Rust.
+
+## Making a Remote Procedure Call
+
+In this section we'll consider how to make Remote Procedure Call (RPC) from languages used in Anki. Languages used for AnkiDroid and AnkiMobile are out of scope of this document.
+
+### Making a RPC from Python
+
+Python can invoke the `NewDeck` method with [`col._backend.new_deck()`](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/pylib/anki/decks.py#L168). This python method takes no argument and returns a `Deck` value.
+
+However, most Python code should not call this method directly. Instead it should call [`col.decks.new_deck()`](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/pylib/anki/decks.py#L166). Generally speaking, all back-end functions called from Python should be called through a helper method defined in `pylib/anki/`. The `_backend` part is an implementation detail that most callers should ignore. This is especially important because add-ons should expect a relatively stable API independent of the implementation details of the RPC.
+
+### Invoking method from TypeScript
+
+Let's consider the method [`rpc GetCsvMetadata(CsvMetadataRequest) returns (CsvMetadata);`](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/proto/anki/import_export.proto#L20) from `ImportExportService`..
+
+It's used in the TypeScript class [`ImportCsvState`](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/ts/routes/import-csv/lib.ts#L102), as an asynchronous function. It's argument is a single javascript object, whose keys are as in [`CsvMetadataRequest`](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/proto/anki/import_export.proto#L138) and it returns a `CsvMetadata`.
+
+The method was imported with `import { getCsvMetadata } from "@generated/backend";` and the types were imported with `import type { CsvMetadata } from "@generated/anki/import_export_pb";`. Note that it was not necessary to import the input type given that it's simply an untyped javascript object.
+
+## Implementation
+
+Let's now look at implementations of those RPCs.
+
+### Implementation in Rust
+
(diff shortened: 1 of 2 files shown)


Arthur-Milchior:
Appart from the sentence you quoted, is there anything you believe needs more polishing? I would have hoped everything is understandable.
I'm happy you confirm everything seems to be correct, it took me some time to understand what I wrote down. Hope it'll help other devs. And great that my "as far as I understand" ensured you confirmed the understanding

docs/protobuf.md
```diff
@@ -116,3 +130,56 @@ Inside the `pb` module you will find all generated Rust types and their implemen
   a valid variant, so the Rust code needs to deal with a lot of `Option`s. As we
   don't expect other parts of Anki to send invalid messages, using an `InvalidInput`
   error or `unwrap_or_default()` is usually fine.
+
+# Communication between languages
+
+Currently, Rust does not send message to Python or Ts. That is, the
```
dae, line 136: 'send' is somewhat misleading here, since the request does not return until the call has completed and a return value is provided - I'd use things like "call" or "invoke" a method instead.

I would lede with something like:

Anki's codebase uses three layers:

1) The core Rust layer
2) The Python layer
3) The web frontend/TypeScript layer

Each layer only makes RPC calls to the layer below it. TypeScript code makes calls to the Python-level HTTP server. Python code makes calls to the underlying Rust layer. The Rust layer never makes RPC calls.

docs/protobuf.md
```diff
@@ -1,5 +1,19 @@
 # Protocol Buffers
 
+ProtoBuf is a format used both to save data in storage and transmit
+data between services. You can think of it as similar to JSON with
+schemas, given that you can use basic types, list and records. Except
+that it's usually represented in an efficient byteform and not in a
+human readable way.
+
+In this document we'll first explain how to encode and decode protobuf
+messages in the three languages used in anki. Typescript, Python and
```
dae, line 10: You don't actually talk about encoding/decoding below - you talk about invoking methods.

dae (commented):
This still feels pretty rough at the moment I'm afraid. Maybe someone else would like to take a stab at rewriting it?

docs/protobuf.md
```diff
@@ -116,3 +130,56 @@ Inside the `pb` module you will find all generated Rust types and their implemen
...
+# Communication between languages
+
+Currently, Rust does not send message to Python or Ts. That is, the
+back-end do not send message to the front-end.  Instead, the front end
+can request data, that Rust returns as part of a RPC initiated by the
+front end. Those RPC are discussed below.
+
+## Sending a message to typescript.
+
+This is the only case that don't use protobuf. Any Python code, having
```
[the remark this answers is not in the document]
Arthur-Milchior: Data between all languages goes through ProtoBuf with a single exception: calling javascript from Python.
The remaining of the document state that everytime you want to do anything cross language you should define a RPC in protobuf.
I think it's important to note that there is an exception, and that this is the only case that does not behave similarly to the other one. Otherwise, the new contributors may be looking for the way to create RPC to call TS from python and lose time not finding them

Arthur-Milchior:
In order to take your feedback into account I:
* splitted the bridge document from the protobuf document. This ensure that explaining how to call TS from python is not out of place
* Rewrote the document using declaration, then calls, then definition of the RPCs. I used current examples (with permalink) so that a contributor can look at the code and imitate it.

docs/language_bridge.md
```diff
@@ -0,0 +1,89 @@
...
+
+Imitating those examples should allow you to make call and create new RPCs.
+
+## Declaring RPCs
+
+Let's consider the method `NewDeck` of `DecksServices`. It's declared in [decks.proto](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/proto/anki/decks.proto#L14) as `rpc NewDeck(generic.Empty) returns (Deck);`. This means this methods takes no argument (technically, an argument containing no information), and returns a [`Deck`](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/proto/anki/decks.proto#L54).
+
+Read [protobuf](./protobuf.md) to learn more about how those input and output types are defined.
+
+If the RPC implementation is in Python, it should be declared in [frontend.proto](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/proto/anki/frontend.proto#L24C3-L24C66)'s `FrontendService`. Otherwise, it can be declared in any other service. Obviously, the service should be consistent with what the method do.
```
dae, line 23: While you mention it below, you should probably explicitly state here that all RPCs are implement at the Rust layer, except when they are placed in frontend.proto.

docs/language_bridge.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Making RPC
```
dae, line 25:
```suggestion
## Making a Remote Procedure Call
```

docs/language_bridge.md
```diff
@@ -0,0 +1,89 @@
...
+
+The method NewDeck is implemented in Rust's [DecksService](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/rslib/src/decks/service.rs#L21) as `fn new_deck(&mut self) -> error::Result<anki_proto::decks::Deck>`. It should be noted that the method name was changed from Pascal case to snake case, and the rps's argument of type `generic.Empty` is ignored.
+
+### Implementation in Python
+
+Let's consider the implementation of the method [DeckOptionsRequireClose](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/qt/aqt/mediasrv.py#L578). It's defined as `def deck_options_require_close() -> bytes:`. In this case, there should be a returned value. However, it'll be ignored, so returning `b""` is perfectly fine.
+
+It must be noted that:
+
+- the RPC is not received on main. In order to do any work with window, we should call `aqt.mw.taskman.run_on_main`.
```
dae, line 57:
```suggestion
- the incoming HTTP request is not processed on the main thread. In order to do any work with the GUI, we should call `aqt.mw.taskman.run_on_main`.
```

docs/language_bridge.md
```diff
@@ -0,0 +1,89 @@
...
+- at the time the RPC is received, we have no guarantee that the main view's active window is a DeckOptionsDialog. So this should be checked before doing anything else. If it's not the case, the call should be ignored.
+- While MediaServer handles all request from typescript, it's not the appropriate place for most implementation. So it's expected that, once it has determined that the active view has the expected type, this active view actually handle the request.
```
dae, line 59: The last two points seem somewhat specific to this particular RPC, and I'm not sure they're adding much.

docs/language_bridge.md
```diff
@@ -0,0 +1,89 @@
...
+
+```js
+globalThis.anki || = {};
+globalThis.anki.methodName = async (): Promise<void>=>{body}
+```
+
+Note that if the function is asynchronous, you can't directly send the
+result to a callback. Instead your function will have to call a post
+method that will be sent to Python or Rust. Which leads us to the next
+section.
```
dae, line 87: What next section?

dae (commented):
The new structure is better, thank you. There's still a number of minor grammatical mistakes. It's 2025, and we shouldn't need to go back and forth on these - please use an LLM to fix your wording.

docs/protobuf.md
dae: RPC call is tautological, but pretty common in my experience. But my point was more that Python→Rust and web→Python are 'calls' in the sense that they block until they receive a reply (assuming the TS promise is awaited). As you know, web.eval(WithCallback) is more akin to injecting some text into the JavaScript console, and it's not possible for the Python code to block until the reply is received.

docs/language_bridge.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Making RPC
+
+In this section we'll consider how to make RPC from languages used in Anki. Languages used for AnkiDroid and AnkiMobile are out of scope of this document.
+
+### Making a RPC from Python
+
+Python can invoke the `NewDeck` method with [`col._backend.new_deck()`](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/pylib/anki/decks.py#L168). This python method takes no argument and returns a `Deck` value.
+
+However, most of Python code should not call this method directly. Instead it should call [`col.decks.new_deck()`](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/pylib/anki/decks.py#L166). Generally speaking, all back-ends functions called from Python should be called through an helper method defined in pylib/anki/. The _backend part is an implementation details that most callers should ignore. This is especially important because add-ons should expect a relatively stable API independently on the implementation details of the RPC.
```
phwoo, line 33:
```suggestion
However, most Python code should not call this method directly. Instead it should call [`col.decks.new_deck()`](https://github.com/ankitects/anki/blob/acaeee91fa853e4a7a78dcddbb832d009ec3529a/pylib/anki/decks.py#L166). Generally speaking, all back-end functions called from Python should be called through a helper method defined in `pylib/anki/`. The `_backend` part is an implementation detail that most callers should ignore. This is especially important because add-ons should expect a relatively stable API independent of the implementation details of the RPC.

PR #3573, opened 2024-11-11
Enable strict_optional for aqt/deckchooser, about, webview

bpnguyen107:


qt/aqt/utils.py
```diff
@@ -275,7 +275,7 @@ def showCritical(
 
 
 def showInfo(
-    text: str,
```
abdnh, line 278: Instead of changing the typing of the utils here, I think it makes more sense to add some ifs to the offending code in webview.py.

bpnguyen107: That makes sense, thanks Abdo.

PR #3575, opened 2024-11-13
Update contributors list and two-way sync of CONTRIBUTORS.md and about.py

gregorylearns:
I apologize if this is not that much of a contribution to the code itself. I noticed some differences in the two files `CONTRIBUTORS.md` and `about.py`. The files are not in sync and some people who are in about.py were not in CONTRIBUTORS.md and vice versa.

I also did not do the Test must pass check under the CONTRIBUTING.md since i just changed a few strings.

Thanks again!


brishtibheja:
That's because there are contributors who haven't contributed code but did translations etc.

Also, not everyone who contributed may want their name to be on Anki's About screen (or they might have been contributing from different aliases). I think both should files should be separated.

gregorylearns:
Oh right, I agree as well. I'll close the pull request. thanks!

PR #3576, opened 2024-11-13
Exclude new cards from Future Due stats

user1823:
https://github.com/ankitects/anki/pull/3530#issuecomment-2439924619

Before https://github.com/ankitects/anki/commit/7ea573b004c976d20d95d706893993c1b6d79f74, they were excluded anyway.


dae:
Thank you!

PR #3582, opened 2024-11-17
Fix/inconsistent retrievability calculations between normal/filtered decks and display/sorting

L-M-Sherlock:
source: https://forums.ankiweb.net/t/anki-24-10-release-candidate/51191/120?u=l.m.sherlock

This PR addresses two inconsistencies in retrievability calculations:

1. Fixes the discrepancy in retrievability values between normal decks and filtered decks for the same card by using the original due date (odue) when card belongs to filtered deck.
2. Fixes retrievability calculations between display strings and review sorting to ensure consistent behavior

Changes:
- Modified due date handling to consider original due date (odue) when card belongs to filtered deck.
- Removed unnecessary max(0) operations from due date calculations

These changes ensure that retrievability values are calculated consistently across all contexts where they are used.

### DIFF
--- a/rslib/src/browser_table.rs
+++ b/rslib/src/browser_table.rs
@@ -111,7 +111,7 @@ impl Card {
     /// Returns the card's due date as a timestamp if it has one.
     fn due_time(&self, timing: &SchedTimingToday) -> Option<TimestampSecs> {
         if self.queue == CardQueue::Learn {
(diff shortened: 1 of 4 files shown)


dae:
Thanks Jarrett - the effort you put in to address not only issues directly inside FSRS but also in Anki's integration of it is greatly appreciated. :-)

PR #3599, opened 2024-11-26
Fix memory state inference from historical review logs

L-M-Sherlock:
This PR addresses two issues related to memory state inference when processing historical review logs:

1. When "Ignore cards reviewed before" is enabled, the code now properly respects the cutoff date when determining the first review entry. Previously, it would consider review entries before the cutoff date, leading to incorrect memory state calculations.

2. Improves the accuracy of memory state inference by properly handling ease factors from different spaced repetition algorithms. When a review entry has an ease factor ≤ 1.1, it's identified as being generated by FSRS, and the difficulty is calculated using the appropriate scale for that algorithm. This ensures correct conversion of historical difficulty values regardless of which algorithm generated them.

cc @dae @Luc-Mcgrady


L-M-Sherlock:
The CI problem:

https://rustsec.org/advisories/RUSTSEC-2024-0399

Luc-Mcgrady (approved):

rslib/src/scheduler/fsrs/memory_state.rs
```diff
@@ -307,11 +308,15 @@ pub(crate) fn single_card_revlog_to_item(
...
             // the revlog has been truncated, but not fully
-            let starting_state = fsrs.memory_state_from_sm2(
+            let mut starting_state = fsrs.memory_state_from_sm2(
                 first_review.ease_factor,
                 first_review.interval,
                 historical_retention,
             )?;
+            // if the ease factor is less than 1.1, the revlog entry is generated by FSRS
+            if first_review.ease_factor <= 1.1 {
+                starting_state.difficulty = (first_review.ease_factor - 0.1) * 9.0 + 1.0;
```
user1823, line 318: Why is this `ease_factor - 0.1`? Isn't the formula `D = (ease factor) * 9 + 1`?

user1823 (commented):
Looks good. Just one question

rslib/src/scheduler/fsrs/memory_state.rs
dae:
```
    /// Returns the difficulty normalized to a 0.1-1.1 range,
    /// which is used in revlog entries.
    pub(crate) fn difficulty_shifted(&self) -> f32 {
        self.difficulty() + 0.1
    }
```

dae:
Thanks Jarrett. Good timing, as I was about to build an rc!

rslib/src/scheduler/fsrs/memory_state.rs
user1823: So, the difficulty is normalized twice, first from 1-10 to 0-1 and then to 0.1-1.1. Interesting!

user1823:
By the way, I noticed just now that we have two functions with very similar names - `single_card_revlog_to_item(` and `single_card_revlog_to_items(`. It might be good idea to rename one of them in a follow up PR.

dae:
Sigh, GitHub is playing up. 

"Merge attempt failed
Merge already in progress"

I'll try to manually merge this instead.

@user1823 The key difference between those two routines is whether they produce a single output or multiple outputs, so I think the name is appropriate. What are you suggesting instead?

user1823:
> What are you suggesting instead?

I will have to study the code more deeply to give you a suggestion. But, currently, only the letter `s` differentiates the two functions. In this state, it would be very easy to introduce bugs by writing one in place of the other.

dae:
Merged in https://github.com/ankitects/anki/commit/6874440cdad97185c13cb37f717bcd787fe2c445

PR #3602, opened 2024-11-29
Fix deck button not clickable in stats screen for smaller screens

### LINKED ISSUE #3595: Can't Select Deck in Stats Screen (Mobile)
## Steps

1. Open stats screen
2. Select `collection`
3. Try switching back to `deck`

### Expected behaviour

switches back to `deck`

### Actual behaviour

Clicking there doesn't work most of the times. To switch back to `deck` it sometimes take 3-4 clicks. This doesn't happen anywhere else in the same screen.

___

Was asked to report it here. Originally posted in https://github.com/ankidroid/Anki-Android/issues/16459.

Sawansunar56:
fix #3595 

Changed the z-index of the spinner element in the stats screen, which caused the spinner to take the click events if the spinner and the radio buttons(deck, collection) were on top of each other, due to insufficient space. I just made it so that it goes behind  if the element is not loading and in front if it is.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -197,6 +197,7 @@ twwn <github.com/twwn>
 Cy Pokhrel <cy@cy7.sh>
 Park Hyunwoo <phu54321@naver.com>
 Tomas Fabrizio Orsi <torsi@fi.uba.ar>
+Sawan Sunar <sawansunar24072002@gmail.com>
 
 ********************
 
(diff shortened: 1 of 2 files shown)


mikehardy:
@Sawansunar56 can you add a quick screen capture or image showing the UI affects of this, and explain in your PR description above (with an edit...) exactly what was done and why? That would help review a lot

Thanks!

Sawansunar56:
https://github.com/user-attachments/assets/bc7f3b9f-1a86-49d4-b76b-41f0c95e0064

NOTE: I increased the opacity for this instance to just show the change.

Sawansunar56:
> Nice - thanks for the description and the video, makes it clear

I appreciate the input as well.

abdnh (approved):
Looking good - Thank you @Sawansunar56 !

PR #3603, opened 2024-12-01
Remove infinite spinning icon for reset parameters dropdown and add a question mark

### LINKED ISSUE #3531: Improve feedback of resetting parameters [deck options]
> https://www.reddit.com/r/Anki/comments/1g9qh7f/resetting_fsrs_parameters_to_default_not_working/ 2
> This user says that it looks like resetting parameters is stuck in an endless loop (it’s not).
> Some time ago someone on Discord also thought that resetting parameters is forever stuck (I can’t find screenshots, unfortunately).
> It seems that the feedback could be improved.
> Maybe just make it so that the circle-arrow symbol needs to be clicked only once, and there is no pop-up that requires a second click?

Originally reported on https://forums.ankiweb.net/t/improve-feedback-of-resetting-parameters-deck-options/51036/9

Removing the confirmation step is the easiest on our end. I'm not sure it's needed - in the rare cases where a user misclicks, they can either avoid saving, or save and then undo if they want to check what the old value was. But a normal pop-up dialog with `confirm()` would be ok too.

bpnguyen107:
Fixes #3531 

Before:
image

After:
image

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -315,7 +315,7 @@ deck-config-confirm-remove-name = Remove { $name }?
 deck-config-save-button = Save
 deck-config-save-to-all-subdecks = Save to All Subdecks
 deck-config-save-and-optimize = Optimize All Presets
-deck-config-revert-button-tooltip = Restore this setting to its default value.
+deck-config-revert-button-tooltip = Restore this setting to its default value?
 
 ## These strings are shown via the Description button at the bottom of the
 ## overview screen.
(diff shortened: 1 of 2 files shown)


abdnh:
I personally prefer the first two options listed by @Expertium [here](https://github.com/ankitects/anki/issues/3531#issuecomment-2439588200):

> Here are a few options:
> 
> 1.  Remove the confirmation entirely
> 2. Make it a more prominent pop-up, like the "Discard changes" one
> 3. Keep the current implementation, but don't make the arrow animated, just make it static
> 
> 

But this looks OK.

abdnh (approved):

bpnguyen107:
I also thought removing the popup entirely seemed the most clear when I was working on this, but I saw some users wanted to keep the popup

dae:
Thanks guys! I personally don't feel we need a confirmation either, but I gather some users feel differently.

PR #3604, opened 2024-12-02
Apply gradient effect to forgetting curve

OuOu2021:
Preview: 
image
image

a card with desired retention set to 0.95:
image
the same card, but with retention set to 0.70
image

Descriptions:
1. The reason for using tomato color instead of red is to reduce the obvious and confusing purple color between red and steel blue.
2. The center of steel blue color stands for the desired retention of the card.

-----------------
2024.12.5 update: 
1. Add a reference line for desired retention.
2. Add y axis title.
3. Display the desired retention in the tooltip when the mouse is near the reference line.
output
4. I18n: make vertical axis title translatable & improve style
image


chrislongros:
It looks beautiful

Expertium:
I guess I'm the minority - I don't like the color gradient here.

YukiNagat0:
I personally like the color gradient.

> I guess I'm the minority - I don't like the color gradient here. But if everyone else is ok with it, may I at least offer a suggestion? Make desired retention a horizontal line.

The idea about the horizontal line for the desired retention seems very nice! Something like a dashed line (maybe with the same `steelblue` color) will be great!

Expertium:
Also, add "Probability of recall" to the Y axis, please

dae:
I ran into one issue: when changing the graph duration, multiple desired retention lines end up getting rendered:
Screenshot 2024-12-09 at 2 55 55 pm

Once that and the CONTRIBUTORS conflict is fixed, happy to merge. Thank you for your contribution!

ftl/core/card-stats.ftl
```diff
@@ -35,6 +35,8 @@ card-stats-fsrs-forgetting-curve-first-week = First Week
 card-stats-fsrs-forgetting-curve-first-month = First Month
 card-stats-fsrs-forgetting-curve-first-year = First Year
 card-stats-fsrs-forgetting-curve-all-time = All Time
+card-stats-fsrs-forgetting-curve-probability-of-recalling = Probability Of Recalling
```
dae, line 38:
```suggestion
card-stats-fsrs-forgetting-curve-probability-of-recalling = Probability of Recall
```

Expertium:
Maybe add "Desired retention" above the horizontal line to make it more obvious that it represents, well, desired retention?

OuOu2021:
> Maybe add "Desired retention" above the horizontal line to make it more obvious that it represents, well, desired retention?

I think the desired retention is clear enough in the tooltip so there is no need to add additional text into the chart, which may affect the readability.

dae:
Thank you!

OuOu2021:
> Thank you!

I'm glad I could contribute to a project that I use every day. Thank you for your guidance and support!

sommerluk:
The string “Probability of Recall” is now showing up in the translation system. I wonder what is the exact meaning. I guess it is the same as “Retrievability”, which is used also in various other contexts in Anki, and also in the very same card statistics dialog, above in the list of data. If so, may I suggest renaming it to “Retrievability” for greater consistency?

PR #3605, opened 2024-12-03
Slider easy day gui

Luc-Mcgrady:
Related forum post: https://forums.ankiweb.net/t/easy-days-ui-improvement-different-button-size/52665/10

image

### DIFF
--- a/ts/routes/deck-options/EasyDays.svelte
+++ b/ts/routes/deck-options/EasyDays.svelte
@@ -36,6 +36,10 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     ];
 </script>
 
+<datalist id="easy_day_steplist">
+    <option>0.5</option>
+</datalist>
+
 <TitledContainer title={tr.deckConfigEasyDaysTitle()}>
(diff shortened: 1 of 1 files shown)


Expertium:
I think it's confusing that right now the titles go Normal -> Reduced -> Minimum, but after the update it will be Minimum -> Reduced -> Normal. But the benefit of making the order more clear (Minimum < Reduced) probably outweighs the immediate confusion, in the long run at least.

Expertium:
Alternative: https://forums.ankiweb.net/t/easy-days-ui-improvement-different-button-size/52665/20?u=expertium
image

I'm curious which one people prefer

Expertium:
Also, I still think that the order should be Normal -> Reduced -> Minimum and not Minimum <- Reduced <- Normal

Expertium:
https://forms.gle/4BG1F3qhuahjkuC48
Feel free to participate, everyone!

abdnh (approved):

dae:
Thanks everyone. Sadly the vote stacking does not surprise me - happens a fair bit on AnkiWeb too.

Issue #3242, opened 2024-06-19
Optimizing FSRS parameters shows 0 reviews


user1823:
<img width=500 src="https://github.com/ankitects/anki/assets/92206575/7f9099a7-e711-49a7-8005-50c3ca2804d2">

However, sometimes, it does show the correct count too.
<img width=500 src="https://github.com/ankitects/anki/assets/92206575/6404d7e1-ac4d-437f-aef3-4ab244dcaefe">

Test file: [Test.apkg.zip](https://github.com/user-attachments/files/15902832/Test.apkg.zip) (Rename to remove .zip)

In this file, try to optimize the parameters of the deck B (subdeck of A) to reproduce the issue.

user1823:
I imported the apkg into a fresh profile and I was able to reproduce it a few times. The issue seems to appear randomly. 

However, I have reasons to think that the issue is more likely to be reproduced when the optimization finishes too fast. 

For e.g., if I start a screen recording, the CPU usage increases and the laptop becomes slightly slower and I am almost never able to reproduce the issue.

benasocj:
I experience the same issue.

In the following scenarios the parameters have always been optimized already. For this reason the "...appear to be optimal" popup is always shown (which is expected).

Following scenarios I have come across:

### 1. Only "0 reviews" shown

Popup is shown:
![screenshot_count_zero1](https://github.com/user-attachments/assets/34f46d21-a82b-4f5b-a973-6745971841bf)
After I close the popup:
![screenshot_count_zero2](https://github.com/user-attachments/assets/a512116a-7b7d-4771-a103-d2ef1e62f2d2)

### 2. Only "1,896 reviews" shown

Popup is shown:
![screenshot_count_only1](https://github.com/user-attachments/assets/2a355ae8-f82b-48ce-b4c3-a761958038d0)
After I close the popup:
![screenshot_count_only2](https://github.com/user-attachments/assets/a437a409-2102-4ddb-af2c-255360fb897f)

### 3. "1,896 reviews" shown along with random percentage

Popup is shown, however with "1,896 reviews" and a random percentage already shown in the background (sometimes 60%, 58.6% or 49.3%):
![screenshot_percent_and_count1](https://github.com/user-attachments/assets/7a8087de-939a-4930-bcd8-511b10ef11cd)
![screenshot_percent_and_count2](https://github.com/user-attachments/assets/57a0bda4-942e-4c8b-a3fd-2142f956b084)
[...]

PR #3610, opened 2024-12-04
Add "No reviews to optimize" message

Luc-Mcgrady:
Gives a different message when there are no reviews as opposed to when it has already optimal reviews.

image

Related forum post: https://forums.ankiweb.net/t/change-the-parameters-appear-to-be-optimal-message-of-there-are-0-reviews/52729

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -452,7 +452,7 @@ deck-config-percent-of-reviews =
 deck-config-optimizing-preset = Optimizing preset { $current_count }/{ $total_count }...
 deck-config-fsrs-must-be-enabled = FSRS must be enabled first.
 deck-config-fsrs-params-optimal = The FSRS parameters currently appear to be optimal.
-deck-config-fsrs-params-no-reviews = No reviews found. Please check that this preset is assigned to all decks you want to optimize and try again.
+deck-config-fsrs-params-no-reviews = No reviews found. Please check that this preset is assigned to all decks you want to optimize (including subdecks) and try again.
 
 deck-config-wait-for-audio = Wait for audio
 deck-config-show-reminder = Show Reminder
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -452,6 +452,7 @@ deck-config-percent-of-reviews =
 deck-config-optimizing-preset = Optimizing preset { $current_count }/{ $total_count }...
 deck-config-fsrs-must-be-enabled = FSRS must be enabled first.
 deck-config-fsrs-params-optimal = The FSRS parameters currently appear to be optimal.
+deck-config-fsrs-params-failed = Failed to optimize, Ensure there are sufficient reviews in cards with this preset.
 
 deck-config-wait-for-audio = Wait for audio
 deck-config-show-reminder = Show Reminder
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -452,7 +452,7 @@ deck-config-percent-of-reviews =
 deck-config-optimizing-preset = Optimizing preset { $current_count }/{ $total_count }...
 deck-config-fsrs-must-be-enabled = FSRS must be enabled first.
 deck-config-fsrs-params-optimal = The FSRS parameters currently appear to be optimal.
-deck-config-fsrs-params-failed = Failed to optimize, Ensure there are sufficient reviews in cards with this preset.
+deck-config-fsrs-params-no-reviews = There are no reviews on cards with this preset. Please ensure that this preset is assigned to all decks you want to optimize and try again.
 
 deck-config-wait-for-audio = Wait for audio
 deck-config-show-reminder = Show Reminder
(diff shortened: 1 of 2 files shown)


user1823:
This doesn't seem to be correct.

If the default parameters are better than the optimized parameters, you will get the new error (failed optimization). But, I think that the "already optimal" error is more appropriate in that situation.

A proper fix would be to show the new error when the number of reviews is zero.

Expertium:
I agree with user1823. The popup should be shown only if there are 0 reviews.

ts/routes/deck-options/FsrsOptions.svelte
```diff
@@ -148,7 +148,13 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                         ignoreRevlogsBeforeMs: getIgnoreRevlogsBeforeMs(),
                         currentParams: params,
                     });
-                    if (
+
+                    if (!resp.fsrsItems) {
+                        setTimeout(
+                            () => alert(tr.deckConfigFsrsParamsNoReviews()),
+                            200,
+                        );
```
Luc-Mcgrady, line 156:
```suggestion
                        );
                        $config.fsrsParams5 = defaults.fsrsParams5
```
I'm wondering if it should also clear any parameters that are already in the box if the user has 0 reviews on that preset.

Expertium:
Nice, thank you. I suggest some minor wording changes:
"No reviews found. Please check that this preset is assigned to all decks you want to optimize and try again."

ts/routes/deck-options/FsrsOptions.svelte
```diff
@@ -148,7 +148,13 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                         ignoreRevlogsBeforeMs: getIgnoreRevlogsBeforeMs(),
                         currentParams: params,
                     });
-                    if (
+
+                    if (!resp.fsrsItems) {
```
user1823, line 152: Is checking the number of fsrsItems in Svelte reliable (vs checking them in Rust)? Asking just because there was an issue in the past where Anki would report 0 reviews even where the preset had some.

See
- https://github.com/ankitects/anki/issues/3242
- https://github.com/ankitects/anki/pull/3591

user1823 (commented):
Left a question

ts/routes/deck-options/FsrsOptions.svelte
user1823: Not sure on this one. But, why should Anki do something which is not required?

ts/routes/deck-options/FsrsOptions.svelte
Luc-Mcgrady: I've made it so that it reporting no reviews wont prevent it from setting new parameters.

ts/routes/deck-options/FsrsOptions.svelte
Luc-Mcgrady: I'm not convinced on this myself, If a user optimises a preset with no reviews but still has parameters because they cloned the preset for example I feel like "optimise" should set it to the optimal, which in the case of an empty preset would be the default.

But yeah I think you're right.

user1823 (approved):
LGTM

abdnh:
Do we need another message for this? Maybe we can just edit `deck-config-not-enough-history` to include the last part:

> Insufficient review history to perform this operation. Please check that this preset is assigned to all decks you want to optimize and try again.

Expertium:
That's the thing - in this Reddit post it didn't trigger.

https://www.reddit.com/r/Anki/comments/1h6lyi6/what_this_popup_say/

image

Notice that it says "0 reviews", but the message doesn't say "Insufficient review history"

user1823:
But, this doesn't mean that we need to create a new string. We can use that string for this case too.

Currently, that string is used when the number of revlogs is non-zero but fsrs-rs returns an error.

Luc-Mcgrady:
We could re-use it but we couldn't change it to include warnings about the presets as it wouldn't make sense to mention presets in the context it is currently used.

https://github.com/ankitects/anki/blob/d897b4588e062df43a65ce03e4382315cf464268/rslib/src/scheduler/fsrs/params.rs#L132-L136
https://github.com/ankitects/anki/blob/d897b4588e062df43a65ce03e4382315cf464268/rslib/src/error/mod.rs#L179

I'd still say its probably better to keep them separate with the extra hint about what might be going wrong. Maybe even include an extra one in case the user accidentally changed the default search. 
image

ftl/core/deck-config.ftl
```diff
@@ -452,6 +452,7 @@ deck-config-percent-of-reviews =
 deck-config-optimizing-preset = Optimizing preset { $current_count }/{ $total_count }...
 deck-config-fsrs-must-be-enabled = FSRS must be enabled first.
 deck-config-fsrs-params-optimal = The FSRS parameters currently appear to be optimal.
+deck-config-fsrs-params-no-reviews = No reviews found. Please check that this preset is assigned to all decks you want to optimize and try again.
```
brishtibheja, line 455: what about this:

```suggestion
deck-config-fsrs-params-no-reviews = No reviews found. Please check that this preset is assigned to all decks and their subdecks you want to optimize and try again.
```

Luc-Mcgrady:
```suggestion
deck-config-fsrs-params-no-reviews = No reviews found. Please check that this preset is assigned to all decks you want to optimize (including subdecks) and try again.
```

dae:
Thanks all!

PR #3613, opened 2024-12-06
Fix reuse internal clipboard for primary selection pasting

### LINKED ISSUE #3249: The “reuse internal” clipboard feature cannot handle primary selections

bpnguyen107:
Should fix #3249. I couldn't test if it works because I don't use X11 though. @tholin maybe you can see if this fixes your issue?

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -1438,21 +1438,25 @@ class EditorWebView(AnkiWebView):
         self._store_field_content_on_next_clipboard_change = True
         self._internal_field_text_for_paste = None
 
-    def _on_clipboard_change(self) -> None:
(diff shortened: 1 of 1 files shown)


dae:
Thanks Ben! I've confirmed this does the right thing on a Linux machine.

PR #3618, opened 2024-12-06
Snap: keep LD_LIBRARY_PATH when in snap environment

boukendesho:
The mpv player in the snap's  isolated environment relies on ld_library_path to work properly.


dae:
Thanks Hideo!

PR #3621, opened 2024-12-08
Remove use of window.location.href in CardInfoDialog

iamllama:
This PR proposes to replace the card info dialog's navigation method from setting `window.location.href` via `eval` to using `AnkiWebView.load_sveltekit_page`

This lets addons, like Anki ReColor and Anki Redesign, style the dialog consistently, even when the card is changed/updated, via `gui_hooks.webview_did_inject_style_into_page`

### DIFF
--- a/qt/aqt/browser/card_info.py
+++ b/qt/aqt/browser/card_info.py
@@ -67,11 +67,17 @@ class CardInfoDialog(QDialog):
         self.setLayout(layout)
 
     def update_card(self, card_id: CardId | None) -> None:
+        from aqt.theme import theme_manager
+
         try:
             self.mw.col.get_card(card_id)
         except NotFoundError:
             card_id = None
 
+        if theme_manager.night_mode:
+            extra = "#night"
+        else:
+            extra = ""
         assert self.web is not None
         self.web.eval(
             f"""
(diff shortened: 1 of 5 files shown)


dae:
This requires a full page reload each time the card is changed, which is less efficient. Can you figure out a way to solve this problem in a less drastic way, such as another hook that is fired after the URL is updated?

iamllama:
> This requires a full page reload each time the card is changed, which is less efficient

From testing with ankiwebview inspector, it seems that both `window.location.href` and `load_sveltekit_page` cause a full page reload

Here's the network waterfall when using `window.location.href` and changing to another note in the browser:
image

When using `load_sveltekit_page` instead:
image
NB. the background isn't grey here because the hook is called

iamllama:
The only way I can think of to avoid full page reloads is to somehow expose [goto](https://svelte.dev/docs/kit/$app-navigation#goto) to the global scope, where it's then accessible via qt's `eval`

I've tried that by tacking it onto `window` and it does reactively update `CardInfo` without re-rendering the entire page (and resetting styles), but it might not be idiomatic. Thoughts?

dae:
Thank you for the detailed investigation! Agree with you that exposing goto directly is a bit ugly. The postMessage solution seems acceptable for now, though we may need to revisit it once we start including multiple "pages" within a single document.

Could I trouble you to resolve the merge conflicts?

iamllama:
Rebased!

dae:
Thank you!

PR #3342, opened 2024-08-04
Update to Qt 6.7

dae:
6.7.1+ adds a new subwheel, that we'll need to remove again come Qt 6.8

PR #3622, opened 2024-12-09
[Qt 6.7] Set ForceDarkMode attribute in AnkiWebView

louwers:
This PR allows using [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) in styles. For example:

```css
p {
 color: red;
}

@media (prefers-color-scheme: dark) {
  p {
   color: blue;
  }
}
```

I know the `night-mode` is already added to the HTML root, but since there is now a CSS feature for this I think it is preferred. Also, it will work with external CSS style sheets that are dark mode aware.

Note that this depends on #3342 since `QWebEngineSettings.ForceDarkMode` was [introduced in Qt 6.7](https://doc.qt.io/qtforpython-6/PySide6/QtWebEngineCore/QWebEngineSettings.html#PySide6.QtWebEngineCore.QWebEngineSettings.WebAttribute
), hence this is marked a Draft for now.

https://github.com/user-attachments/assets/137dbf24-2002-48df-91a8-99313cdcd04c

### DIFF
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -289,10 +289,6 @@ class AnkiWebView(QWebEngineView):
         self.set_kind(kind)
         if title:
             self.set_title(title)
-        self.settings().setAttribute(
-            QWebEngineSettings.WebAttribute.ForceDarkMode,
-            theme_manager.get_night_mode(),
-        )
         self._page = AnkiWebPage(self._onBridgeCmd)
         # reduce flicker
         self._page.setBackgroundColor(theme_manager.qcolor(colors.CANVAS))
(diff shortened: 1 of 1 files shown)


dae:
It might be worth adding a version check and doing this conditionally - even if we stop providing Qt5 builds, there may be Linux distros who continue to ship Anki with a Qt5 or earlier Qt6 version.

louwers:
Good idea, will add the version check.

louwers:
Moved to `on_theme_did_change` and setting the setting on the page so that it updates dynamically.

Added `hasattr` check so this will start working for Qt 6.7 builds.

dae:
Thanks Bart - appreciated!

PR #3623, opened 2024-12-09
Fix deck options button twitching on hover

BlueGreenMagick:
In deck options page, hovering on buttons shifts layout in 2 places:
1. Hovering on the tab buttons ('This Deck', 'Today Only') in 'Daily Limits' section makes the button text twitch.
2. In the help modal, hovering on the tab buttons makes the entire page layout shift. (The issue is especially pronounced in the 'Audio' help modal)

The issue is caused by a 1px border being added to the button on hover, where there is no border. This PR fixes the 2 issues.


dae:
Thanks! ❤️

PR #3627, opened 2024-12-09
typeanswer: [type:nc] - use nfkd again

twwn:
as by discussion [here](https://github.com/ankitects/anki/pull/3422#issuecomment-2527192240)


dae:
Thanks @twwn, appreciated.

PR #3629, opened 2024-12-10
Remove hardcoded note/card colours from switch.py

iamllama:
Currently, the ACCENT_CARD and ACCENT_NOTE colours are hardcoded into `Switch.__init__` as default args.

> Default parameter values are evaluated from left to right when the function definition is executed. This means that the expression is evaluated once, when the function is defined, and that the same “pre-computed” value is used for each call.
https://docs.python.org/3/reference/compound_stmts.html#function-definitions

This means that it doesn't use the current colour values after addons (recolor, redesign etc.) have potentially changed them

This PR proposes to have `Switch` look up the colours when it's initialised, and not just when it's first defined

EDIT: not sure why it's failing on rust crate advisories


abdnh (approved):

dae:
Thank you!

Issue #3608, opened 2024-12-04
Two Image Occlusion issues


jcznk:
Tested and reproduced on both 24.06.3 and 24.11, Windows 10, Qt6. Both issues are related to the Image Occlusion editor (Add window).

- Add window: If you toggle the mask editor and navigate to the fields section, you cannot scroll using the mousepad or the mouse scroll wheel. If you attempt to, the "mask editor" will be scrolled instead.
Short demo:
https://github.com/user-attachments/assets/25db67e6-3d57-46dd-aa7a-7384b49d372e
(Steps: 1. toggle the mask editor, 2. scroll with the mouse wheel, and nothing seems to happen, 3. when I go back to the mask editor, the image has been scrolled up.)

- Add window: If you navigate to the fields section, add one or more tags, return to the mask editor, and then go back to the fields section, the tag(s) will "disappear." If you add the note after the tags have "disappeared", they will still be included in the newly created note. However, if you re-focus the tag bar before adding, the tags will be lost.

- Also:

  1. Add an Image Occlusion note with one or more tags.
  2. After adding the note, don't close the Add window. Select or paste an image and create some masks.
  3. Go to the fields section, delete the tag(s) from the previous note, and add one or more new tags.
  4. Return to the mask editor.
  5. Go back to the fields section — the editor will display the tags from the previous note instead of the correct ones.
  6. As before: if you simply add the new note, the correct tags will be included. However, if you re-focus the tag bar before adding, the tags from the previous note will be added instead.

Issue #3581, opened 2024-11-16
Cannot put image occlusion mask right at the edge of the image


abdnh:
> Windows 10, Anki version ⁨24.11 (97e6b3d3)⁩, release candidate.
> 
> Steps to reproduce:
> 
> Zoom in as much as possible when making/editing an IO card.
> Try to make an occlusion such that its edge is touching the edge of the image itself.
> The edge of the occlusion doesn’t touch the edge of the image, there is a gap.

Originally reported on https://forums.ankiweb.net/t/io-bug-cant-put-an-occlusion-right-at-the-edge-of-an-image/51982

abdnh:
Might be related to this: https://github.com/ankitects/anki/issues/2588

Though as noted in the forums, the behavior is different on stable.

abdnh:
Fixed in #3630

PR #3630, opened 2024-12-11
Allow object to move to right edge and bottom edge and allow scroll of note fields when not using IO

krmanik:
> Add window: If you toggle the mask editor and navigate to the fields section, you cannot scroll using the mousepad or the mouse scroll wheel. If you attempt to, the "mask editor" will be scrolled instead.

https://github.com/ankitects/anki/issues/3608

> Zoom in as much as possible when making/editing an IO card.
Try to make an occlusion such that its edge is touching the edge of the image itself.
The edge of the occlusion doesn’t touch the edge of the image, there is a gap.

https://github.com/ankitects/anki/issues/3581

### DIFF
--- a/ts/routes/image-occlusion/tools/lib.ts
+++ b/ts/routes/image-occlusion/tools/lib.ts
@@ -288,9 +288,9 @@ export const makeShapesRemainInCanvas = (canvas: fabric.Canvas, boundingBox: fab
         const left = obj.left!;
 
         const topBound = boundingBox.top!;
-        const bottomBound = topBound + boundingBox.height!;
+        const bottomBound = topBound + boundingBox.height! + 5;
         const leftBound = boundingBox.left!;
-        const rightBound = leftBound + boundingBox.width!;
+        const rightBound = leftBound + boundingBox.width! + 5;
 
         obj.left = Math.min(Math.max(left, leftBound), rightBound - objWidth);
         obj.top = Math.min(Math.max(top, topBound), bottomBound - objHeight);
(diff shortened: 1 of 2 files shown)


ts/routes/image-occlusion/tools/lib.ts
```diff
@@ -288,9 +288,9 @@ export const makeShapesRemainInCanvas = (canvas: fabric.Canvas, boundingBox: fab
         const left = obj.left!;
 
         const topBound = boundingBox.top!;
-        const bottomBound = topBound + boundingBox.height!;
+        const bottomBound = topBound + boundingBox.height! + 5;
```
abdnh, line 291: Any idea why this is needed?

abdnh (approved):

ts/routes/image-occlusion/tools/lib.ts
krmanik: The width and height less than the value is not drawn, so using that value fixes edge problems.

dae: I believe the root issue is the user can tap on the screen / make a very small gesture, and end up creating a rectangle only a pixel or two in width. It then doesn't get rendered on the screen, so the user ends up with an invisible box that still counts as an occlusion. Does that sound correct @krmanik? Or am I thinking of another issue?

If my memory is correct here, maybe another solution would be refusing to create the shapes if they're below a certain size?

krmanik: This is before and after screen recording, I think above is different issues, but not sure.

**Before**
<video src="https://github.com/user-attachments/assets/9ab9afd5-b723-4046-9fcc-d537ea75f56e">

**After**
<video src="https://github.com/user-attachments/assets/a223025c-0e49-4c41-8518-2d452c293257">

abdnh: @dae That's a different issue (can't remember whether it was fixed)

PR #3631, opened 2024-12-13
Fix CardInfoPlaceholder not showing when card id is invalid

iamllama:
When the card info dialog is opened during study (or in the browser), and when study ends (or the browser has 0 results), the card info dialog's card id is set to `None` and shows the error thrown when parsing it as a BigInt.

image

With this pr, the error is caught and an optional is returned by load that's handled by CardInfo, showing the placeholder instead of the thrown error

image

Split out from #3621


brishtibheja:
How about using gray text colour like [in the stats screen]() instead of brackets?

abdnh (approved):

dae:
Happy to accept a follow-up commit that tweaks the formatting - let's merge this in now, since it's a bug-fix.

PR #3633, opened 2024-12-14
Qt 6.8.1

dae:
Bumps minimum glibc to 2.35, and minimum macOS to 12

I need to do some more testing of the packaged builds before this is ready to go.


dae:
Let's give this a try in the next build.

Issue #2921, opened 2023-12-31
Incorrect stability of card with incomplete revlog


user1823:
In Anki 23.12.1 (as well as 23.10.1), the cards with incomplete revlog have incorrect stability.

For example, the following card has a stability of 2.13 months but I expected it to be 9.44 months (for SM2 retention = 0.92).

<img width="533" alt="image" src="https://github.com/ankitects/anki/assets/92206575/465e9630-d3a8-4d2b-b10f-cc93b995cc07">

The card is attached for easy reproduction: [Bug.apkg.zip](https://github.com/ankitects/anki/files/13800708/Bug.apkg.zip) (rename the file to remove .zip at end)

The stability is not corrected if I change the weights in Deck Options and click "Save" to recalculate memory states. However, if I review the card on AnkiDroid and then reschedule it using helper add-on's auto-reschedule after sync function, the stability gets corrected.

Example of such card:
<img width="560" alt="image" src="https://github.com/ankitects/anki/assets/92206575/c7d42e8f-b5aa-4add-b27b-07ef15653240">

dae:
The review log is only used when the are 2+ usable entries. For 0 or 1 entries, stability is inferred from the card state.

user1823:
> For 0 or 1 entries, stability is inferred from the card state.

This can cause issues in the 1 entry case. If the SM2 retention and desired retention are not the same, the stability will change each time the memory states are recalculated.

So, I think that the revlog should be used even in the 1 entry case.

dae:
I don't disagree, but it's not a trivial change, as set_memory_state() expects to only receive an existing memory state if there are 2+ entries, as the later entries are applied to the initial learning state.

user1823:
> but it's not a trivial change

I understand. But, a solution must be found. I am sure that I am not the only person affected by this issue.

user1823:
Fixed in https://github.com/ankitects/anki/pull/3639

PR #3639, opened 2024-12-15
Fix/fallback to non-manual entry when first_of_last_learn_entries non found

### LINKED ISSUE #3634: Set due date can confuse FSRS's calculations
> After updating to Anki 24.11, and re-optimising my parameters, some of my cards have had their intervals increased by an extremely large amount of time, for no apparent reason. For example, in the image below, you can see a card that before I rescheduled had an interval of 27 days. However, after updating to Anki 24.11 (and so FSRS 5), and rescheduling my cards, the interview changed to 6.67 years. I most definitely do not know this card, and a select few other cards that have had similar changes to their intervals.

Originally reported on https://forums.ankiweb.net/t/fsrs-5-increased-interval-by-5-years/53003/7

L-M-Sherlock:
fix #3634

### DIFF
--- a/rslib/src/scheduler/answering/mod.rs
+++ b/rslib/src/scheduler/answering/mod.rs
@@ -32,7 +32,7 @@ use crate::deckconfig::DeckConfig;
 use crate::deckconfig::LeechAction;
 use crate::decks::Deck;
 use crate::prelude::*;
-use crate::scheduler::fsrs::memory_state::single_card_revlog_to_item;
+use crate::scheduler::fsrs::memory_state::fsrs_item_for_memory_state;
 use crate::scheduler::states::PreviewState;
 use crate::search::SearchNode;
 
@@ -437,7 +437,7 @@ impl Collection {
                 // and will need its initial memory state to be calculated based on review
                 // history.
                 let revlog = self.revlog_for_srs(SearchNode::CardIds(card.id.to_string()))?;
-                let item = single_card_revlog_to_item(
+                let item = fsrs_item_for_memory_state(
                     &fsrs,
                     revlog,
                     timing.next_day_at,
--- a/rslib/src/scheduler/fsrs/memory_state.rs
+++ b/rslib/src/scheduler/fsrs/memory_state.rs
@@ -386,8 +386,8 @@ mod tests {
                 difficulty: 5.7909784,
             }),
         );
-        // but if there's only a single revlog entry, we'll fall back on current card
-        // state
+        // but if there's only a single revlog entry, we'll fall back on the first
+        // non-manual entry
         let item = single_card_revlog_to_item(
             &fsrs,
             vec![RevlogEntry {
@@ -398,17 +398,15 @@ mod tests {
             TimestampSecs::now(),
             0.9,
             0.into(),
-        )?;
-        assert!(item.is_none());
-        card.interval = 123;
-        card.ease_factor = 2000;
-        card.ctype = CardType::Review;
-        card.set_memory_state(&fsrs, item, 0.9)?;
+        )?
+        .unwrap();
+        assert!(item.item.reviews.is_empty());
+        card.set_memory_state(&fsrs, Some(item), 0.9)?;
         assert_int_eq(
             card.memory_state,
             Some(FsrsMemoryState {
-                stability: 122.99994,
-                difficulty: 7.334526,
+                stability: 99.999954,
+                difficulty: 5.840841,
             }),
         );
         Ok(())
--- a/rslib/src/scheduler/fsrs/memory_state.rs
+++ b/rslib/src/scheduler/fsrs/memory_state.rs
@@ -290,14 +290,14 @@ pub(crate) fn fsrs_item_for_memory_state(
                 item,
                 starting_state: None,
             }))
-        } else if let Some(first_non_manual_entry) = output.filtered_revlogs.first() {
+        } else if let Some(first_user_grade) = output.filtered_revlogs.first() {
             // the revlog has been truncated, but not fully
             let first_review = FirstReview {
-                interval: first_non_manual_entry.interval.max(1) as f32,
-                ease_factor: if first_non_manual_entry.ease_factor == 0 {
+                interval: first_user_grade.interval.max(1) as f32,
+                ease_factor: if first_user_grade.ease_factor == 0 {
                     2500
                 } else {
-                    first_non_manual_entry.ease_factor
+                    first_user_grade.ease_factor
                 } as f32
                     / 1000.0,
             };
@@ -384,8 +384,8 @@ mod tests {
                 difficulty: 5.7909784,
             }),
         );
-        // but if there's only a single revlog entry, we'll fall back on the first
-        // non-manual entry
+        // cards with a single review-type entry also get memory states from revlog
+        // rather than card states
         let item = fsrs_item_for_memory_state(
             &fsrs,
             vec![RevlogEntry {
--- a/rslib/src/scheduler/fsrs/memory_state.rs
+++ b/rslib/src/scheduler/fsrs/memory_state.rs
@@ -284,20 +284,7 @@ pub(crate) fn single_card_revlog_to_item(
         interval: f32,
         ease_factor: f32,
     }
-    let first_review = entries
-        .iter()
-        // ignore manual and rescheduled revlogs and revlogs before the cutoff
-        .find(|e| e.button_chosen > 0 && e.id.0 >= ignore_revlogs_before.0)
-        .map(|e| FirstReview {
-            interval: e.interval.max(1) as f32,
-            ease_factor: if e.ease_factor == 0 {
-                2500
-            } else {
-                e.ease_factor
-            } as f32
-                / 1000.0,
-        });
-    if let Some((mut items, revlogs_complete, _)) =
+    if let Some((mut items, revlogs_complete, _, filtered_entries)) =
         single_card_revlog_to_items(entries, next_day_at, false, ignore_revlogs_before)
     {
         let mut item = items.pop().unwrap();
@@ -306,7 +293,16 @@ pub(crate) fn single_card_revlog_to_item(
                 item,
                 starting_state: None,
             }))
-        } else if let Some(first_review) = first_review {
+        } else if let Some(first_non_manual_entry) = filtered_entries.first() {
+            let first_review = FirstReview {
+                interval: first_non_manual_entry.interval.max(1) as f32,
+                ease_factor: if first_non_manual_entry.ease_factor == 0 {
+                    2500
+                } else {
+                    first_non_manual_entry.ease_factor
+                } as f32
+                    / 1000.0,
+            };
             // the revlog has been truncated, but not fully
             let mut starting_state = fsrs.memory_state_from_sm2(
                 first_review.ease_factor,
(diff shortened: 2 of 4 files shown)


user1823:
Just to be sure, `non_manual_entries.is_some()` is true only if there are non-manual entries after the Reset entry, right?

Did you try this on a card that has no entries after a reset entry?

L-M-Sherlock:
Sure:

image

L-M-Sherlock:
Wait a minute... The `single_card_revlog_to_items` is too complex than my thoughts... The current implementation uses the first non-manual entry instead of the first non-manual entry after the `Reset` entry. I will fix it.

L-M-Sherlock:
@dae, sorry for bothering you, but I don't understand this snippet of code:

https://github.com/ankitects/anki/blob/d90050600316cd58d25bc7f61fc762c3f704c101/rslib/src/scheduler/fsrs/params.rs#L289-L300

What's the aim of it?

dae:
Introduced in https://github.com/ankitects/anki/commit/db93939ded947d74bb25bd8552a2b2356a096509. 

I believe the goal was to treat them as complete even if no learning entries have been found, if they started with a reschedule?

user1823:
Dae added that code to implement my suggestion in https://github.com/open-spaced-repetition/fsrs4anki/issues/540#issuecomment-1848833376

The issue: https://github.com/open-spaced-repetition/fsrs4anki/issues/557

If the user uses Set Due Date to introduce a card rather than by normal way, the cards were getting very high difficulty.

L-M-Sherlock:
My PR will infer the memory state from the first non-manual entry aftet `Reset` and set revlogs_complete to false. But this snippet of code will reverse my operation in some cases.

user1823:
In https://github.com/open-spaced-repetition/fsrs4anki/issues/675, I suggested removing this snippet. So, I think that you can remove it.

L-M-Sherlock:
OK, here is the next problem:

https://github.com/ankitects/anki/blob/d90050600316cd58d25bc7f61fc762c3f704c101/rslib/src/scheduler/fsrs/memory_state.rs#L287-L299

I think `first_review` should be extracted from the items returned by `single_card_revlog_to_items` instead of the original review entries.

user1823:
> I think first_review should be extracted from the items returned by single_card_revlog_to_items instead of the original review entries.

Your suggestion seems reasonable to me.

L-M-Sherlock:
So this unit test should be removed.

https://github.com/ankitects/anki/blob/d90050600316cd58d25bc7f61fc762c3f704c101/rslib/src/scheduler/fsrs/memory_state.rs#L397-L421

L-M-Sherlock:
OK, the next problem:

https://github.com/L-M-Sherlock/anki/blob/35fa731b75b476ea730e7c2f38725d353143faef/rslib/src/scheduler/fsrs/params.rs#L328-L332

https://github.com/L-M-Sherlock/anki/blob/35fa731b75b476ea730e7c2f38725d353143faef/rslib/src/scheduler/fsrs/params.rs#L266-L272

If we find a relearning entry, should we set the revlogs_complete to true?

user1823:
> If we find a relearning entry, should we set the revlogs_complete to true?

What did the previous code do? I will have to search through previous discussions for reasoning for/against this. Unfortunately, I won't be able to devote much time today and tomorrow.

user1823:
On a gross overview, the changes made till now look good to me.

L-M-Sherlock:
Here is the previous code:

https://github.com/ankitects/anki/blob/d90050600316cd58d25bc7f61fc762c3f704c101/rslib/src/scheduler/fsrs/params.rs#L327-L338

It was introduced in 90b0c6d84072ec0b11a43bd063a40cd1b784bdb0

user1823:
If the previous code set it to true, then let's keep it that way. There must be some reason why it was done like that.

L-M-Sherlock:
> If the previous code set it to true, then let's keep it that way. There must be some reason why it was done like that.

In previous code, `revlogs_complete` is true only when Anki finds the `first_of_last_learn_entries` or the first entry is Manual or Rescheduled.

L-M-Sherlock:
If `revlogs_complete` is false, Anki will infer the memory state from the relearn entry. If it's true, the rating of the relearn entry will be treated as the first rating.

user1823:
Does this PR also fix https://github.com/ankitects/anki/issues/2921?

user1823:
A minor suggestion:

Bring this comment
```
// the revlog has been truncated, but not fully
```
between the following two lines in params.rs
```rs
        } else if let Some(first_non_manual_entry) = filtered_entries.first() {
            let first_review = FirstReview {
```

L-M-Sherlock:
> Does this PR also fix #2921?

Yes. The `starting_state` is converted from the first review of that affected card.

L-M-Sherlock:
> But, what does [90b0c6d](https://github.com/ankitects/anki/commit/90b0c6d84072ec0b11a43bd063a40cd1b784bdb0) do?

If the `first_of_last_learn_entries` is non-found, Anki will try to find the `first_relearn_entry`. If it's found, Anki will remove all entries before this entry. In the commit version, the `first_relearn_entry` was treated as the first learning entry. Its rating was used to generate the initial stability and difficulty.

user1823:
> Its rating was used to generate the initial stability and difficulty.

Using normal formula or using memory_state_from_sm2?

L-M-Sherlock:
> Using normal formula or using memory_state_from_sm2?

Using normal formula. But it was changed in db93939ded947d74bb25bd8552a2b2356a096509. This commit introduced `revlogs_complete`. And `revlogs_complete` is true only when `last_learn_entry` is found. If `revlogs_complete` is false, Anki will use `memory_state_from_sm2`.

user1823:
> But it was changed in https://github.com/ankitects/anki/commit/db93939ded947d74bb25bd8552a2b2356a096509.

Well, in that case, using `memory_state_from_sm2` was a bug. But, this change was made 1 year ago and no one has complained. So, let's keep `revlogs_complete` as False in that case for now.

user1823:
Next issue:

In the following code, if there are no learn entries, no review is ignored.

https://github.com/ankitects/anki/blob/f91dec44eed7051248f16115b8d575c17b5a238d/rslib/src/scheduler/fsrs/params.rs#L310-L327

Solution: (pseudocode)

```rs
 } else { 
     // While reviewing if the first learning step is before the ignore date, 
     // ignore every review before and including the last learning step 
     if let Some(idx) = first_of_last_learn_entries { 
         if entries[idx].id.0 < ignore_revlogs_before.0 && idx < entries.len() - 1 { 
                 .iter() 
                 .enumerate() 
                 .rev() 
                 .map(|(idx, _)| idx); 
  
             entries.drain(..(last_learn_entry? + 1)); 
             revlogs_complete = false; 
             first_of_last_learn_entries = None; 
         } 
     } else if first_of_last_learn_entries.is_none() {
           delete entries with id < ignore_revlogs_before
         } 
 } 

```

L-M-Sherlock:
Wait a minute:

In the current pull request, if we identify both a relearn entry and a non-manual entry, with the non-manual entry preceding the relearn entry, we infer the memory state from the relearn entry rather than the non-manual entry.

Is it expected?

user1823:
> we infer the memory state from the relearn entry rather than the non-manual entry. Is it expected?

No

L-M-Sherlock:
OK, so we need to remove the `first_relearn_entries` and related code.

user1823:
I think that we should remove `or(first_relearn_entries)` from the following. WDYT?

```rs
    if let Some(idx) = first_of_last_learn_entries.or(first_relearn_entries) {
        // start from the (re)learning step
        if idx > 0 {
            entries.drain(..idx);
        }
```

L-M-Sherlock:
> In the following code, if there are no learn entries, no review is ignored.

OK, we have a new problem:

user1823:
> We need to break this loop when the entry is before the ignore date:

No,

What if we have something like 
```
L L | L L R
```
If we break the loop, we can't distinguish this case from something like ` | L L R R `. The current code distinguishes between these two cases.

Here, | is the ignore date.

user1823:
Another thing:

In this PR, you added
```rs
    } else if let Some(idx) = non_manual_entries {
        // if there are no learning entries but there are non-manual entries,
        // we ignore all entries before the first non-manual entry
        if idx > 0 {
            entries.drain(..idx);
        }
```

Is this required? The code immediately after this is already removing all the manual entries.

user1823:
> What if the non-manual entry is before the ignore date?

Ignore that entry. That's what I wrote in my pseudocode.

user1823:
> This code is used to remove entries before Reset.

I assume that by "this code", you mean the code you added.

Then what does the `break` do in the following code?

```rs
            // If we find a `Learn` entry after the `Reset` entry, we should
            // ignore the entries before the `Reset` entry
            if first_of_last_learn_entries.is_some() {
                revlogs_complete = true;
                break;
            // If we find a non-manual entry after the `Reset` entry, we should
            // ignore the entries before the `Reset` entry
            } else if non_manual_entries.is_some() {
                revlogs_complete = false;
                break;
```

user1823:
Ok, I guess that I am now understanding the purpose of this code.

Even if the loop breaks, the entries are still there. If we don't remove them, `first_non_manual_entry` in memory_state.rs will be incorrect. Right?

If I am correct, I suggest moving that code just after the loop ends, i.e., just before

```rs
    if training {
        // While training ignore the entire card if the first learning step of the last
```

user1823:
Next issue/suggestion:

We should also exclude cram revlogs (entry.review_kind == RevlogReviewKind::Filtered && entry.ease_factor == 0) here.

```rs
        if matches!(entry.button_chosen, 1..=4) {
            non_manual_entries = Some(index);
        }
```

This is the last suggestion from me for today. I will take another look tomorrow to ensure that no issue is remaining and then we can request dae for a final review.

By the way, thanks for this PR. You have squashed tons of bugs with this.

L-M-Sherlock:
I refactored a lot of codes.

# Cram revlogs are skipped

```rust

        if entry.review_kind == RevlogReviewKind::Filtered && entry.ease_factor == 0 {
            continue;
        }
```
# Only search the first non-manual revlog entry after ignore date

```rust
        if matches!(entry.button_chosen, 1..=4) && entry.id.0 > ignore_revlogs_before.0 {
            non_manual_entries = Some(index);
        }
```

# fallback to non-manual revlog entry if the first of last learn entires is before ignore date

I referred to your idea: https://github.com/ankitects/anki/pull/2922#issuecomment-1937037115

```
        // While reviewing if the first learning step is before the ignore date,
        // fallback to non_manual_entries
        if let Some(idx) = first_of_last_learn_entries {
            if entries[idx].id.0 < ignore_revlogs_before.0 && idx < entries.len() - 1 {
                revlogs_complete = false;
                first_of_last_learn_entries = None;
            }
        }
```

dae:
Sorry, I forgot we iterate on the entries in reverse, which I think has made my renaming incorrect. Reviewing...

dae:
Ok, sorry about the delay. Tests past after the changes in the last commit, but I'm not 100% confident the behaviour is identical - can you spot any problems?

L-M-Sherlock:
For this case, the `revlog_complete` should be false.

```
L | L R
```

user1823:
The last commit (https://github.com/ankitects/anki/pull/3639/commits/7fd9a7789efe8e9633bac66f526ab8456974c3aa) has introduced a functional change (I think) and it isn't desirable.

If a card has revlog entries like `L L | L R R`, this card
- is used for training
- gets it memory states using normal FSRS formulas (instead of memory_states_from_sm2) at the L entry just after the cutoff.

where | = cutoff date

user1823:
Apart from the above mentioned functional issue, the code looks good to me.

However, I have some suggestions regarding wording:

memory_state.rs
- rename `first_non_manual_entry` to `first_user_graded_entry` (or something like that)
- change this comment
```
        // but if there's only a single revlog entry, we'll fall back on the first
        // non-manual entry
```
to 
```
        // cards with a single review-type entry also get memory states from revlog rather than card states
```

params.rs
- change this comment
```
        // While reviewing, if the first learning step is before the ignore date,
        // we won't start from the learning step.
```
to 
```
        // While reviewing, if the first learning step is before the ignore date,
        // infer memory states from sm2 at the revlog entry just after.
```
- change this comment
```
        // if there are no learning entries, but the user has reviewed the card,
        // we ignore all entries before the first grade
```
to 
```
        // ignore all entries before the first user-graded rating matched by the loop
```

PR #3642, opened 2024-12-16
Fix Fsrs simulator input problems

Luc-Mcgrady:
fixes that new cards/deck size can not be set to 0.
fixes that the default values are tied to the default preset rather than the current one.

Forum post: https://forums.ankiweb.net/t/bug-cant-set-additional-new-cards-to-simulate-to-0-after-changing-it/53267


rbrownwsws:
It looks like you need to merge / rebase on `main` to get the checks to pass.

cargo-deny is upset about the `idna` vuln that was fixed in d014f7237789671d5c9d88abffd7709afacf2fb4

dae:
Thank you both!

rbrownwsws:
This was all @Luc-Mcgrady 's work. I just noticed the reason the checks were failing.

dae:
I know, but pointing that out is also a useful contribution :-)

PR #3645, opened 2024-12-17
Fix/dataPoint index is off by one day in simulator & remove moving average

L-M-Sherlock:
source: https://forums.ankiweb.net/t/the-fsrs-simulator-graph-tooltips-are-wrong/53268#p-150754-h-1-the-number-of-reviews-in-the-tooltips-seem-to-be-off-by-one-day-1


dae:
Thanks Jarrett, and sorry the bugfixes are taking up your time. I'm thinking of putting out a new beta in early Jan so these fixes don't wait too long.

BYT3M3:
It appears that this fix has traded in one off-by-one for another one:
image
As can be seen on this screenshot, the tooltip is correct now, but the tick for the horizontal axis is off.

PR #3656, opened 2024-12-20
Fix potential race condition when refreshing the webview during review

Lolle2000la:
This happens especially when the `refresh_if_needed` gets called in fast succession, allowing `mw.fade_in_webview` to get called concurrently. This causes the webview to, under certain conditions, present a previous or corrupted frame since the underlying call to is `web.eval` not thread safe. This will ensure that the renders happen in order.

More context: Sometimes on different devices, when checking the back of a card, the back does not get shown. Either nothing changes or the previous cards back gets displayed, though manually causing a re-render by selecting text will fix this. This happens especially often with complicated cards, i.e. those that load scrips, display a lot of contents, use more css etc.

I have searched for the cause of this and saw that the affected code seemed to have circumstances where the `web.eval`-function gets called concurrently, causing this bug. The lock around the `refresh_if_needed` fixes this, ensuring the correct order of frames. I am not sure if this fixes the root issue (most likely not), but it does fix the error with the wrong order/false presentation of frames.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -201,6 +201,7 @@ Dongjin Ouyang <1113117424@qq.com>
 Sawan Sunar <sawansunar24072002@gmail.com>
 hideo aoyama <https://github.com/boukendesho>
 Ross Brown <rbrownwsws@googlemail.com>
+Luca Auer <lolle2000.la@gmail.com>
 🦙 <github.com/iamllama>
 
 ********************
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -10,6 +10,7 @@ from collections.abc import Callable, Generator, Sequence
 from dataclasses import dataclass
 from enum import Enum, auto
 from functools import partial
+from threading import Lock
 from typing import Any, Literal, Match, Union, cast
 
 import aqt
@@ -155,6 +156,7 @@ class Reviewer:
         self._combining: bool = True
         self.typeCorrect: str | None = None  # web init happens before this is set
         self.state: Literal["question", "answer", "transition"] | None = None
+        self._refresh_lock = Lock()
         self._refresh_needed: RefreshNeeded | None = None
         self._v3: V3CardInfo | None = None
         self._state_mutation_key = str(random.randint(0, 2**64 - 1))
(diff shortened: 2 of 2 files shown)


iamllama:
> allowing mw.fade_in_webview to get called concurrently

I'm not seeing how that happens, since `refresh_if_needed` gets called in the main thread and `fade_in_webview` is its only caller, unless i'm mistaken

> the underlying call to is web.eval not thread safe

The js actions are run sequentially in the main thread as weill i think, so perhaps qt is doing some reordering on its own 🤔

dae:
Previous users who have reported similar problems have found that changing their video driver resolved the issue for them. Can you confirm whether the issue you're seeing goes away with a different driver but without this patch?

Lolle2000la:
> Previous users who have reported similar problems have found that changing their video driver resolved the issue for them. Can you confirm whether the issue you're seeing goes away with a different driver but without this patch?

Thank you for your reply! I have observed this on my PC with a Nvidia GPU and on my laptop with Intel integrated graphics. I had also tried other render methods like Vulkan, but the Bug still appeared.

Lolle2000la:
> > allowing mw.fade_in_webview to get called concurrently
> 
> I'm not seeing how that happens, since `refresh_if_needed` gets called in the main thread and `fade_in_webview` is its only caller, unless i'm mistaken
> 
> > the underlying call to is web.eval not thread safe
> 
> The js actions are run sequentially in the main thread as weill i think, so perhaps qt is doing some reordering on its own 🤔

I am definitely not an expert on QT, so if you think this shouldn't fix the bug, then I will try to reinvestigate.

What it seems like is that the call is not thread safe, and it can happen that the method gets called to force a refresh when something changes (i.e. next card). When it happens multiple times it seems to start rendering multiple frames.

Putting a lock around the body of the method seemed to fix the issue, but I am somewhat sure that there is still more to this. I just didn't observe the bug happening anymore with the lock on my laptop and PC.

dae:
So you find it happens even when set to Software display mode?

As @iamllama has pointed out, UI updates should already be serialized, as they're all supposed to be run on the main thread. Furthermore, Javascript is sent asynchronously to the webview, meaning the lock may well have been released before the webview gets around to processing the opacity change. While I'm not doubting the lock fixes the problem for you, I suspect it's more about a change in timing on your machine, rather than a robust fix.

If you suspect a threading race condition, I think a better first step might be to add some checks into the code to verify that the code is actually being called on a non-UI thread.

Lolle2000la:
That definitely sounds like like a very reasonable assumption. Like you say, I might just shift some timings with the lock that just happen to work. I will close this pull request for now and re-investigate. The `fade_in_webview` and `fade_out_webview` seem like reasonable candidates to look into for now, so I will look into that.

PR #3657, opened 2024-12-21
Place the forgetting curve above the revlog

OuOu2021:
## Description
The length of the revlog is variable, which leads to uncertainty about the position of the forgetting curve. Therefore, it may be a better choice to place the forgetting curve above the revlog.


iamllama:
Would be nice to discuss this in the forums first honestly, the revlog info seems more important imo

OuOu2021:
@iamllama I'm sorry. It really should have been discussed first. Please check out https://forums.ankiweb.net/t/redesign-card-info/53438

PR #3658, opened 2024-12-22
Translation comments for actiony-all-selected and actiony-any-selected

sommerluk:
It's sometimes difficult to translate Anki, especially if strings are short and there is a lack of context:

https://forums.ankiweb.net/t/actions-all-selected-actions-any-selected/53469

More detailed comments could help. This PR introduces two of them.

Remarks:

- I have read the documentation about Fluent language. If I understand correctly, there is no automatic text extraction like in "gettext". And it is enough to just edit manually the flt files. Is that correct?
- I had no possibility to actually test this PR.
- Changing only a comment and not the string itself does not trigger a re-translation, does it?


dae:
Thank you for contributing! You are correct that this is the only file that needs changing, and a comment won't trigger a retranslation.

PR #3660, opened 2024-12-24
Sort FSRSItems by RevlogId for training

L-M-Sherlock:
According to my recent [benchmarking result](https://github.com/open-spaced-repetition/srs-benchmark?tab=readme-ov-file#weighted-by-the-number-of-reviews), recency weighting could improve FSRS-5 ~4.8% relatively. This enhancement is comparable to the improvements seen in FSRS-5 over FSRS-4.5, or FSRS-4.5 over FSRS v4.

It requires a minor change in Anki's side: the items for training should be sorted in chronological order.

Of course, I can expose the weight assigning in the API, but it will introduce breaking changes and make the interface too complicated.

Relative PR:

- https://github.com/open-spaced-repetition/fsrs-rs/pull/260/


dae:
Nice gain!

PR #3662, opened 2024-12-24
Add support for html comments to template syntax

iamllama:
From https://forums.ankiweb.net/t/conditional-brackets-are-still-recognized-inside-html-comments/2207 and https://forums.ankiweb.net/t/is-it-possible-to-add-comment-in-template-that-is-not-displayed-in-review-mode/19962

This PR proposes adding support for html comments (`<!-- -->`) to the template syntax, e.g.
```
{{#A}}
    <!-- {{B}} -->
    {{C}}
    <!--
    {{#Image}}
        {{#Image}}
        {{#Caption}}
    {{/Image}}
    -->
{{\A}}
```
where brackets inside closed comment blocks are treated as text

Just as styling allows comments in it, it can be handy to keep old/useful snippets around without having to make another note/card type or stash it somewhere else. Since templates are html, html comments are used instead of something like `/* */`

Couldn't find a suitable parser combination that could pass current tests *and* avoid allocations, hence the mess 😅
Extra relevant tests have been added as well. Unclosed comments aren't treated as errors, since ankidroid still lacks a wysiwyg field editor (desktop enforces closed comments).

Breakage:

- Legacy mode is left alone, as i assumed it to be under feature-freeze

- The following would no longer cause card generation, despite currently not rendering anything:
```
<!--
{{#A}}
    {{#B}}
{{/A}}
-->
```
This seems like the most concerning one, but it can be fixed by moving the comments inside the {{#A}} scope instead of around

- This would no longer be valid:
```
<!-- {{#A}} -->
    {{#B}}
{{/A}}
```

Given the breakages and the lack of issues requesting this for quite some time, this might not be well received, so i'd just like to put this out here to gather feedback. Happy holidays!

### DIFF
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -157,7 +157,8 @@ class CardLayout(QDialog):
         )
 
     def _fieldsOnTemplate(self, fmt: str) -> str:
-        matches = re.findall("{{[^#/}]+?}}", fmt)
+        fmt_without_comments = re.sub("<!--.*?-->", "", fmt)
+        matches = re.findall("{{[^#/}]+?}}", fmt_without_comments)
         chars_allowed = 30
         field_names: list[str] = []
         for m in matches:
--- a/rslib/src/template.rs
+++ b/rslib/src/template.rs
@@ -13,6 +13,7 @@ use nom::branch::alt;
 use nom::bytes::complete::tag;
 use nom::bytes::complete::take_until;
 use nom::combinator::map;
+use nom::combinator::recognize;
 use nom::combinator::rest;
 use nom::combinator::verify;
 use nom::sequence::delimited;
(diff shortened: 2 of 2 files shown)


brishtibheja:
> Given the breakages and the lack of issues requesting this for quite some time, this might not be well received, so i'd just like to put it out here to gather feedback.

I personally don't think there would be a lot of people with templates that depend on the current behaviour. I like the change 👍

dae:
Thank you for contributing! Sorry it's taken a while to get back to you.

I'm a bit torn on this one. Treating HTML comments like template syntax seems a bit hacky, but at least HTML comments are not nestable, so we at least don't need a full HTML parser to handle it.

text_token_inner feels a bit complicated, but I'm not sure yet whether that complexity is avoidable - I'll have a further look into this hopefully tomorrow.

dae:
Thanks llama, I feel that's a lot clearer than before, and I like how you factored out `take_until_handlebar_start_or_tag()` as well.

While looking through these changes, I was initially confused by the behaviour I was seeing, due to `next_token()` only looking for the start of a handlebar clause, instead of a full one. It's not the first time it's confused me, and it got me thinking about whether we could simplify `next_token()` to use the full `handlebar_token()` / `comment_token()` matchers instead, and make it marginally more efficient at the same time.

What do you think about the approach I've just pushed? A notable user-visible difference is that an unclosed '{{' is no longer an error, though I'm not sure that's a positive thing.

iamllama:
LGTM! It's a lot more readable and general now without hardcoded tags and separate legacy handling 👍 Was worried about the perfomance of iterating per char vs `take_until`'s use of `str::find`, but without simd i doubt there's a noticeable difference if any

dae:
Ok, let's do it. Thank you for helping to improve the code!

PR #3665, opened 2024-12-24
Don't Exclude Suspended Cards from Retreivability Graph

brishtibheja:
As per https://forums.ankiweb.net/t/include-suspended-cards-in-retreivability-graph/52757/10?u=sorata

Partially revert ankitects/anki#3518


user1823:
Only the following part needs to be removed:

```rs
            if card.queue == CardQueue::Suspended {
                continue;
            }
```

Other changes are not related.

brishtibheja:
Does it look good now?

user1823 (approved):
LGTM.

PR #3667, opened 2024-12-25
Fix/last date of existing card should not be positive in simulation

L-M-Sherlock:
Due to [Set Due Date doesn’t update the interval of card](https://forums.ankiweb.net/t/set-due-date-doesnt-update-the-interval-of-card/52646), the current converting method sometime set the last due date of existing cards to the future, which is invalid to the simulator. For example: https://github.com/open-spaced-repetition/fsrs-rs/pull/265#discussion_r1896721981

This PR forces `last_date` less or equal to zero to avoid the problem.

@Luc-Mcgrady, would you mind taking a look at this PR?


Luc-Mcgrady (approved):
:+1:

PR #3674, opened 2024-12-27
Ensure data is stored in a volume in anki-sync-server Docker image

omarkohl:
Otherwise data would be lost by default when removing (or re-creating) a
container.

It would be possible to expose the default directory (e.g.
/home/anki/.syncserver) but it would be different for the two Dockerfiles and
less convenient for users of the Docker container to specify such a long path
when naming their volumes.

Setting the permissions is necessary since anki will be running with 'anki'
user permissions inside the container.


jeankhawand (approved):
Please rebase your changes

dae:
Thanks!

PR #3675, opened 2024-12-27
avoid warning by setting SYNC_PORT as ARG in Dockerfile

omarkohl:
>    1 warning found (use docker --debug to expand):
>    UndefinedVar: Usage of undefined variable '$SYNC_PORT'


abdnh:
I've not tested this, but it's looking good to me. Thank you!

PR #3676, opened 2024-12-27
Simulator legend tooltip

Luc-Mcgrady:
Shows some stats about the simulations when the legend is hovered over.

Mobile example video: https://github.com/user-attachments/assets/ec4e49d9-d616-4a51-9def-b59cfda305dc

Forum link: https://forums.ankiweb.net/t/suggestion-have-average-review-stat-for-each-sim/53449/8

### DIFF
--- a/ts/routes/graphs/simulator.ts
+++ b/ts/routes/graphs/simulator.ts
@@ -18,7 +18,7 @@ import {
 
 import * as tr from "@generated/ftl";
 import { timeSpan } from "@tslib/time";
-import { sum, sumBy } from "lodash-es";
+import { sumBy } from "lodash-es";
 import type { GraphBounds, TableDatum } from "./graph-helpers";
 import { setDataAvailable } from "./graph-helpers";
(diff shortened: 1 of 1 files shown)


brishtibheja:
Thanks Luc! Nitpicking here: shouldn't the time stats unit also have `/day` in it? In this case, `mins/day`.

Luc-Mcgrady:
Getting the i18n of /day working with the seconds/minutes/hours seemed like a huge pain.
I'm assuming here I can't just do something like
```ts
timeSpan(count) + // {count} seconds|minutes|hours
tr.perDay() // /day
```
And that we'd need 3 more translations of 
```
statisticsElapsedTimeSecondsPerDay
statisticsElapsedTimeMinutesPerDay
statisticsElapsedTimeHoursPerDay
```

dae:
You can do something like

statistics-per-day = {time}/day

That allows translators to reorder the words if necessary, eg their language equiv. of 'per day: {time}', etc.

PR #3677, opened 2024-12-28
Tweak how the True Retention stats table displays numbers

rbrownwsws:
- Always show fractional parts of numbers even if they are 0 (91.0% not 91%).
  > This makes the numbers line up nicer across rows.
- Show "N/A" for percentages instead of 0% when there are 0 total reviews.
  > I feel showing 0% in this case can be a bit misleading.
  >
  > e.g.
  > You have not reviewed any mature cards today.
  > You look at the "all" table and see it shows 0% for mature today.
  > You might assume that you saw mature cards but failed all of them.

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tf>
</thead>

<tbody>
<tr>
<td>

</td>

<td>

</td>
</tr>
</tbody>
</table>

PR #3679, opened 2024-12-28
Add percentage to FSRS spinner

Arthur-Milchior:
This commit add a percentage option in SpinBox and SpinBoxFloatRow, set to False by default.

If it's true, a percent symbol is added at the end of the line before the increase/decrease button.

While the value is represented as a percentage without decimal places, the internal representation is not changed. Which mean that a multiplier must used to compute the string value, indicate to the input field the min, max and step, and when updating the result.

Currently result is as this 
image

I was able to obtain 
image
 but not to let it be very close to the number, nor to put it outside of the box on the left of the reset value.

If anyone knows their CSS well, that could be helpful here. At least the data part is correctly ported

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -32,8 +32,8 @@ AMBOSS MD Inc. <https://www.amboss.com/>
 Aristotelis P.  <https://glutanimate.com/contact>
 Erez Volk <erez.volk@gmail.com>
 zjosua <zjosua@hotmail.com>
-Arthur Milchior <arthur@milchior.fr>
 Yngve Hoiseth <yngve@hoiseth.net>
+Arthur Milchior <arthur@milchior.fr>
 Ijgnd
 Yoonchae Lee <bluegreenmagick@gmail.com>
 Evandro Coan <github.com/evandrocoan>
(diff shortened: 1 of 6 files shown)


Expertium:
I had in mind either this
image
Or this
image

The first one is probably more intuitive.

Arthur-Milchior:
As mentioned on discord if you know how to make the css, I'd love it.
Sadly I don't understand it enough to correct it

abdnh:
> The first one is probably more intuitive

First one looks tricky but the second should be easy to implement, and I think looks better than the current implementation.

abdnh (approved):

iamllama:
Would localisation need to be handled?

image
from https://forums.ankiweb.net/t/make-space-width-space-usage-in-front-of-units-consistent/53798/13

abdnh:
Related to localization: #3677

It's tricky here though because we have an input.

Expertium:
@dae I'd like you to add this to 25.01 milestones, please.

dae:
I don't object to the move to a percentage, but I find the current formatting to be pretty ugly I'm afraid. Until a CSS guru comes along with a simple+good looking solution, a simpler solution like removing the displayed '%' symbol from the right (while still showing 90 instead of 0.9) would unblock this.

PR #3680, opened 2024-12-30
Fix/issue 3659 wrong window closes command w

### LINKED ISSUE #3659: Wrong window closes on command-w
> The wrong window sometimes closes when using the close window keyboard shortcut in Anki for Mac Version ⁨24.11 (87ccd24e)⁩.
> Repro steps:
> 
> Open the Anki application.
> In the main (User 1) window, click Add (to open the Add window) and then click Browse (to open the Browse window).
> Click the title bar of the Add window, to focus it.
> Press command-w.
> 
> Expected: the Add window should close, since it’s focused.
> Actual: the Browse window closes.

Originally reported on https://forums.ankiweb.net/t/wrong-window-closes-on-command-w/53194

This happens because the Browse screen is a main window and has its own menubar, complete with Close action. The add window is not, so the browse menubar is retained. We could fix this by making close act on the currently-active window, or removing it from the menubar.

tampueroc:
Fixes #3659
Prevents closing the browser window unless it’s the active window.

### DIFF
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -385,7 +385,6 @@ class Browser(QMainWindow):
         else:
             f.actionClose.setVisible(True)
 
-
     def _editor_web_view(self) -> EditorWebView:
         assert self.editor is not None
         editor_web_view = self.editor.web
@@ -399,13 +398,12 @@ class Browser(QMainWindow):
             evt.accept()
             return
 
+        assert self.editor is not None
         if current_window() == self:
             self.editor.call_after_note_saved(self._closeWindow)
             evt.accept()
         else:
             evt.ignore()
-
-        assert self.editor is not None
 
     def _closeWindow(self) -> None:
         assert self.editor is not None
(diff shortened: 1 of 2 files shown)


qt/aqt/browser/browser.py
```diff
@@ -380,6 +380,11 @@ def set_flag_func(desired_flag: int) -> Callable:
         add_ellipsis_to_action_label(f.actionCopy)
         add_ellipsis_to_action_label(f.action_forget)
 
+        if isinstance(self, AddCards):
```
abdnh, line 383: `self` can't be AddCards here, so this doesn't make sense.

qt/aqt/browser/browser.py
```diff
@@ -394,9 +399,11 @@ def closeEvent(self, evt: QCloseEvent | None) -> None:
             return
 
         assert self.editor is not None
-
-        self.editor.call_after_note_saved(self._closeWindow)
-        evt.ignore()
+        if current_window() == self:
```
abdnh, line 402: Same as above.

abdnh (changes requested):

dae:
Closing to reduce clutter; happy to reopen if/when you return to this.

Issue #3647, opened 2024-12-17
Provide a way to exit full-screen mode using the mouse


jcznk:
Anki 24.11 (Qt6) Windows 11

Deck Browser and Reviewer: Full-screen mode can be enabled via View > Toggle Full Screen using the mouse. However, there appears to be no way to exit full-screen mode without using the keyboard.

In the Card Browser, this is not an issue because the top bar (Edit, View, Notes, Cards, Go, Help) remains visible even in full-screen mode.

Suggestion: In the Deck Browser and Reviewer, either keep the top bar visible during full-screen mode or display it when hovering over the top of the screen with the mouse.

jcznk:
I suggested modifying the behavior of the top bar for two main reasons:

1. To improve accessibility. I worry that some users might accidentally enable full-screen mode while clicking around in the top bar and then struggle to exit it if they are unaware of the F11 shortcut. To address this, it would be helpful to retain the UI element they used to enter full-screen mode, making it easier for them to "find their way out".
2. To make the Deck Browser and Reviewer behave more consistently with the Card Browser.

That said, I also like your right-click context menu suggestion and would personally be fine with either option or both.

By the way, I noticed that the top menubar does appear in the Reviewer while in full-screen mode, but only when hovering over the top/bottom bar with "Hide the top/bottom bar during review" set to 'Yes' in Preferences.

KartikSharma0:
I have implemented your suggestions on keeping the menu bar while in full screen.

YukiNagat0:
Doesn't the menu bar already spawn when you hover to the top of the window in full screen in Deck Browser/Reviewer?
The code @KartikSharma0 changed in his PR has been around for ~3 years and there has been no negative feedback on it as far as I can remember. Everyone knows the F11 shortcut. If even users didn't know it, they would probably try hovering over the top of the window.

This issue addresses a problem that literally no one has experienced. The current implementation seems the most obvious and correct.

jcznk:
> Doesn't the menu bar already spawn when you hover to the top of the window in full screen in Deck Browser/Reviewer?

On my copy of Anki (24.11 for Windows 11, Qt6), this only works if `Hide top/bottom bar during review` is set to "Yes" in the Preferences. Otherwise, nothing happens when hovering at the top of the window. I didn't initially notice this discrepancy as I personally do not use the setting.

[...]

PR #3682, opened 2025-01-01
Show menu bar in full screen

KartikSharma0:
Fix: #3647 
EDIT: The menu bar will be visible to users in full screen when the cursor hovers near the top of the screen.
1. Added a mouse movement event to detect changes in the cursor's position.
2. When the cursor's height (y value) is near 1 (top of the screen) the menu bar becomes visible. 

~1. Removed ``hide_menubar`` function when full screen is toggled on Windows and non-mac systems.~
~3. Added the ``show_menubar`` function in the ``on_toggle_full_screen`` function without conditional statement.~

Added my name in CONTRIBUTORS as this is my first PR for this project.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -203,6 +203,7 @@ hideo aoyama <https://github.com/boukendesho>
 Ross Brown <rbrownwsws@googlemail.com>
 🦙 <github.com/iamllama>
 Lukas Sommer <sommerluk@gmail.com>
+Kartik Sharma <github.com/KartikSharma0>
 
 ********************
 
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -163,6 +163,15 @@ class MainWebView(AnkiWebView):
             self.mw.toolbarWeb.hide_timer.start()
             self.mw.bottomWeb.hide_timer.start()
             return True
+        
+        if evt.type() == QEvent.Type.MouseMove:
+            if self.mw.fullscreen and not is_mac:
+                cursor_pos = self.mapFromGlobal(QCursor.pos())
+                if cursor_pos.y() < 1:
+                    self.mw.show_menubar()
+                else:
+                    self.mw.hide_menubar()
+            return True
 
         return False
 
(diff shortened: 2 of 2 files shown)


abdnh:
I can imagine some people complaining about this change. But right now, the menu bar actually re-appears in full-screen mode when you exit the review screen. Can we maybe fix that and show the menu bar only on hover instead?

YukiNagat0:
> I can imagine some people complaining about this change. But right now, the menu bar actually re-appears in full-screen mode when you exit the review screen. Can we maybe fix that and show the menu bar only on hover instead?

As discussed in the original issue (https://github.com/ankitects/anki/issues/3647#issuecomment-2567175657 and subsequent comments), permanently displaying the `menu bar` is not a valid solution. After further discussion with the issue's author, we agreed that the `menu bar` should always follow a `show on hover` behavior, regardless of the `Hide top bar` setting. This will require refactoring the `Hide top/bottom bar` settings code to ensure it does not interfere with the `menu bar`'s behavior.

Therefore, the changes introduced in this PR are not aligned with the desired direction.

KartikSharma0:
@abdnh @YukiNagat0  I've implemented the change for showing the menu bar when hovering near the top of the screen as discussed. 
I also have a question as someone new to the repository: is there a file for declaring constant variables? For example, in my code I check if the y coordinate of the cursor position is less than a value to detect when the cursor is hovering at the top of the screen. I would like to declare the y coordinate value as a constant variable.

YukiNagat0:
@KartikSharma0, I apologize for the quibble, but I believe the `show on hover` behavior is already implemented. However, it only works correctly when the `Hide top/bottom bar` options are enabled. You can verify this yourself by turning on these options and hovering at the top in full-screen mode from the main menu.
image

The flaw in the current (master branch) implementation is that the `menubar` (`File Edit View ...`) and `toolbar` (`Decks Add Browse ...`) are treated as a single entity. Specifically, the `TopWebView.show/hide` methods already include `show/hide_menubar` calls.

The goal is to decouple the `menubar` and `toolbar` behaviors so that the `menubar` always exhibits the `show on hover` behavior, regardless of the `Hide top bar` setting. This can be achieved without introducing new logic (such as checking the cursor position, as suggested in your case).

KartikSharma0:
Thanks for the thorough explanation. I will look into this issue from the approach you mentioned.

KartikSharma0:
@YukiNagat0 I appreciate your patience in explaining the desired functionality. I have attached a video showing the menu bar behaviour in full screen after the most recent commits in this PR. I believe it accomplishes what you have described, please take a look and let me know what you think.

[Anki- show menu bar on hover.webm](https://github.com/user-attachments/assets/d2a66ad4-64d9-4b23-9c08-7a43a3e06d15)

PR #3685, opened 2025-01-02
Remove broken editor image menu when on qt5

iamllama:
Relevant post: https://forums.ankiweb.net/t/right-click-paste-error/53835

When on qt5, right-clicking in the editor throws. This is due to [`lastContextMenuRequest`](https://doc.qt.io/qt-6/qwebengineview.html#lastContextMenuRequest) only having been introduced in qt 6.2.

Until such time as someone manages to find an appropriate workaround for qt5, this pr proposes to remove that feature when on qt5 to make the editor's context menu functional again at least.

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -1472,6 +1472,8 @@ class EditorWebView(AnkiWebView):
         self.triggerPageAction(QWebEnginePage.WebAction.CopyImageToClipboard)
 
     def _opened_context_menu_on_image(self) -> bool:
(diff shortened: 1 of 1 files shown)


abdnh:
Here's a reference if someone is interested in contributing a Qt5 workaround: https://github.com/glutanimate/image-occlusion-enhanced/blob/33711026fbbfd0950fcfaee88ce776ab5e395f9b/src/image_occlusion_enhanced/main.py#L164

PR #3687, opened 2025-01-04
Allow choosing filtered decks in stats

### LINKED ISSUE #3600: Deck selector in stats does not show filtered decks
Reported on https://forums.ankiweb.net/t/anki-24-10-11-release-candidate/51191/151

Maybe DeckChooser should support the `dyn` argument to pass to StudyDeck.

iamllama:
Fixes #3600

`StudyDeck` is already able to include filtered decks. As suggested, this change adds an optional passthru param to `DeckChooser` that lets `NewDeckStats` tell `StudyDeck` to include them

Since it's already possible to open the stats window when a filtered deck is selected, this behaviour seems sensible

Also fixes the bug where if the stats window is opened while a filtered deck is selected, the deck chooser button shows `Default` instead

### DIFF
--- a/qt/aqt/deckchooser.py
+++ b/qt/aqt/deckchooser.py
@@ -21,10 +21,12 @@ class DeckChooser(QHBoxLayout):
         label: bool = True,
         starting_deck_id: DeckId | None = None,
         on_deck_changed: Callable[[int], None] | None = None,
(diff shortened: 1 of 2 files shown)


abdnh:
Thank you!

PR #3688, opened 2025-01-04
Set Due Date: Set interval to actual elapsed days when FSRS is enabled

user1823:
https://github.com/ankidroid/Anki-Android/issues/17707


user1823:
Looks like this is not as easy as I thought it would be. Sorry for the clutter.
If someone knows how to get this to completion, please feel free to do so.

PR #3690, opened 2025-01-05
Fix csv columns potentially being numbered wrongly in the header when exporting

iamllama:
Originally reported by @brishtibheja in https://forums.ankiweb.net/t/bug-guid-column-is-marked-as-deck-column/53924

With the way [csv columns are currently numbered](https://github.com/ankitects/anki/blob/f1a5808d83e809b41d1299d2245ca9f8a7ddb5b7/rslib/src/import_export/text/csv/export.rs#L219) when exporting, a column selection that isn't a contiguous subsequence of `guid, notetype, deck, tags*` causes them to be numbered wrongly in the csv output.

By contiguous, e.g.
Fine: `guid, notetype, deck` or `notetype, deck` or `deck` etc.
Not fine: `guid, deck` or `notetype, tags` or `guid, notetype, tags` etc.

*the field(s) before `tags` don't affect this

### DIFF
--- a/rslib/src/import_export/text/csv/export.rs
+++ b/rslib/src/import_export/text/csv/export.rs
@@ -222,13 +222,23 @@ impl NoteContext {
     }
 
     fn deck_column(&self) -> Option<usize> {
-        self.with_deck
-            .then(|| 1 + self.notetype_column().unwrap_or_default())
+        self.with_deck.then(|| {
+            1 + self
+                .notetype_column()
+                .or_else(|| self.guid_column())
+                .unwrap_or_default()
+        })
     }
 
     fn tags_column(&self) -> Option<usize> {
-        self.with_tags
-            .then(|| 1 + self.deck_column().unwrap_or_default() + self.field_columns)
+        self.with_tags.then(|| {
+            1 + self
+                .deck_column()
+                .or_else(|| self.notetype_column())
+                .or_else(|| self.guid_column())
+                .unwrap_or_default()
+                + self.field_columns
+        })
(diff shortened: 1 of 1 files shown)


rslib/src/import_export/text/csv/export.rs
```diff
@@ -222,13 +222,23 @@ impl NoteContext {
...
     }
 
     fn tags_column(&self) -> Option<usize> {
-        self.with_tags
-            .then(|| 1 + self.deck_column().unwrap_or_default() + self.field_columns)
+        self.with_tags.then(|| {
+            1 + self
+                .deck_column()
+                .or_else(|| self.notetype_column())
+                .or_else(|| self.guid_column())
```
dae, line 238: I think I'm missing something - why fall back on notetype/guid here when deck_column() does that already?

iamllama: It's for when the deck column isn't exported, in which case `deck_column` returns `None` and `tags_column` then has to fall back to the next earliest column(s)

Otherwise, the tags column would be 1 + 0 + the number of fields, potentially overlapping with the guid and notetype columns if they exist

This is the same for all the columns, they have to keep falling back to the next earliest non-None column

dae: 🤦  thanks :-)

RumovZ:
Could have been handled better when I originally wrote it, so we're not reliant on getting the column order right in all the various places. But I think it's reasonable to just fix the current logic.

PR #3691, opened 2025-01-06
Fix EasyDays not applying defaults correctly in deck options

iamllama:
Steps to reproduce:
1) Open deck options and switch to a preset without easy day settings (if not already on one)
2) Switch to another such preset (repeat as many times as you wish)
3) Modify the easy day settings in any one of those presets
4) Note that the rest of those presets all reflect the same changes

`default.easyDaysPercentages` needs to be cloned, otherwise all the presets with applied default easy day settings share the same copy


dae:
Thanks!

PR #3692, opened 2025-01-06
Added colour to simulator tooltip

Luc-Mcgrady:
image

You can break this by spamming the button but that should be fixed in #3670 
Forum post: https://forums.ankiweb.net/t/simulator-have-coloured-icons-in-tooltip/54044


dae:
Thanks!

Issue #3589, opened 2024-11-18
Merge note types not updating until after restart?


dae:
I can't reproduce the issue mentioned in the initial report, but I am seeing the note is skipped on the second import, so the third field doesn't get filled in. Is that expected @RumovZ?

Originally reported on https://forums.ankiweb.net/t/merge-note-types-not-updating-until-after-restart/52053

[example.zip](https://github.com/user-attachments/files/17803009/example.zip)

dae:
Related? https://forums.ankiweb.net/t/update-notes-if-newer-only-works-one-time/52064

iamllama:
> https://forums.ankiweb.net/t/update-notes-if-newer-only-works-one-time/52064
> but I am seeing the note is skipped on the second import, so the third field doesn't get filled in

This happens because the imported note's mtime is set to **now** (i.e. after the second export(!)) when calling `Collection::update_note_inner_without_cards` in `NoteContext::update_note`:
https://github.com/ankitects/anki/blob/f1a5808d83e809b41d1299d2245ca9f8a7ddb5b7/rslib/src/import_export/package/apkg/import/notes.rs#L466-L470

When importing the second export, the mtimes are checked here:
https://github.com/ankitects/anki/blob/f1a5808d83e809b41d1299d2245ca9f8a7ddb5b7/rslib/src/import_export/package/apkg/import/notes.rs#L521-L525

BUT the second export's note mtime is now less than the existing note's mtime (set above), and isn't considered to be newer.

This could perhaps be fixed by having `NoteContext::update_note` patch the existing note to have the imported note's mtime instead of setting it to now

But i'm not sure if that's a good idea, since if the note is imported a long time after it was initially exported, it'd be hard to tell from its mtime that it was freshly modified. Unless that's acceptable?

PR #3693, opened 2025-01-07
Fix newer notes incorrectly being skipped when importing successive exports

iamllama:
> Related? https://forums.ankiweb.net/t/update-notes-if-newer-only-works-one-time/52064

Fixes the "skipped notes when importing" issue in #3589

Let's say someone exports a deck X as an apkg (1), modifies it, exports it again (2), modifies it, exports it again (3)
And on a new profile: imports apkg 1, imports apkg 2, and tries to import apkg 3

Currently, when **existing** notes are updated while importing an apkg (2), the mtime is set to the *time of the import*.

This means that the notes in apkg 3 are now considered to be older than existing notes, and are skipped when trying to import it.

This is incorrect behaviour, since the user didn't manually modify any notes after each import.

With this change, the incoming note's mtime is preserved to allow importing successive exports

### DIFF
--- a/rslib/src/import_export/package/apkg/import/notes.rs
+++ b/rslib/src/import_export/package/apkg/import/notes.rs
@@ -15,6 +15,7 @@ use crate::import_export::package::UpdateCondition;
 use crate::import_export::ImportError;
 use crate::import_export::ImportProgress;
 use crate::import_export::NoteLog;
+use crate::notes::UpdateNoteInnerWithoutCardsArgs;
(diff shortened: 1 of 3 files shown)


dae:
@RumovZ any thoughts? I can't recall if the mtime bump was something deliberate and required, or just a byproduct of the existing update_note() implementation

@iamllama changes look good on my end, and thanks for tackling that clippy todo while you were there!

RumovZ:
I also don't recall, but probably using the incoming mtime is more appropriate.

Issue #2563, opened 2023-07-01
Consider lazy-loading MathJax and CodeMirror in editor


dae:
MathJax appears to account for about 1/3rd of our (2.9MB large!) editor bundle, and codemirror appears to be the second-largest contributor. If we were able to lazy-load these as required, we may be able to improve editor startup time for users who don't use MathJax or the HTML editor by default. We're also using the full MathJax bundle - maybe there's a way to lazy-load the required parts? A good starting point would be to just strip them out and compare editor loading time before and after, to see what sort of performance gains there are.

nihil-admirari:
AnkiDroid simply [searches for \\(, \\), \\[ or \\]](https://github.com/ankidroid/Anki-Android/blob/d22e277bd99cffe6f04c3dd6340ca67caea893e1/AnkiDroid/src/main/java/com/ichi2/anki/cardviewer/CardHtml.kt#L90) in card's template before deciding to load MathJax.

The majority of MathJax extensions are lazy-loaded by default unless [`autoload` is disabled](https://docs.mathjax.org/en/latest/input/tex/extensions/autoload.html). [`mchem`](https://docs.mathjax.org/en/latest/input/tex/extensions/mhchem.html) should be safe to remove from https://github.com/ankitects/anki/blob/78b4a391cc9bf3a127fe3a11984bff75b909b136/ts/mathjax/index.ts#L6 `noerrors` and `mathtools` however, are not autoloaded.

KurtPrice:
@dae Should this issue be closed? The last PR was merged and @nihil-admirari 's comment seems to address the concerns.

abdnh:
I guess initializing MathJax conditionally can still be useful to speed up the editor if it's never used. CodeMirror will also need to be investigated.

BrayanDSO:
What about lazy-loading MathJax in the reviewer?

user1823:
I implemented this. My code works, i.e., MathJax is not loaded when not required (confirmed by absence of `JS warning /_anki/js/editor.js:439 No version information available for component [tex]/mathtools` messages in the console) and gets loaded when required. But, I didn't notice any meaningful performance improvements.

I didn't notice any meaningful perf improvement even after completely stripping out mathjax.js.

Edit: This doesn't seem to be the correct way to lazy load MathJax. mathjax.js looks like just a configuration file. The real mathjax is still loaded unconditionally by editor.js

```diff
diff --git a/qt/aqt/editor.py b/qt/aqt/editor.py
[...]

PR #3705, opened 2025-01-07
Issue 2563 - remove mhchem

Delizald:
As per issue [2563](https://github.com/ankitects/anki/issues/2563) looks like it is safe to remove "mhchem" from the packages that MathJax is loading.


dae:
Thanks!

PR #3714, opened 2025-01-11
Introduce PUID and PGID env variables to syncserver Dockerfile

omarkohl:
PUID and PGID are optional env variables to specify the user and group id of the user that the anki-sync-server process should run with.

This gives more flexibility for solving permission problems with volumes and is a common pattern for Docker images (e.g. see here: https://docs.linuxserver.io/general/understanding-puid-and-pgid/)

The anki-sync-server process will write any files with the permissions of the user it's running with, which can be a problem when you need to access those files from outside the container or when they are being written into a bind mount that is owned by a particular user on the host system.

To be able to implement this the entrypoint.sh needs to run as root (since it needs to create a user and change file permissions). anki-sync-server then needs to be started with the user 'anki', which is why the new dependency 'su-exec' is required. The user 'anki' and group 'anki-group' can no longer be created at image build time because then their ids would be fixed.

Also update the build instructions to require building the Docker image inside the directory where the Dockerfile resides since the build now needs to copy the entrypoint.sh and it seems wrong the specify the path docs/syncserver/entrypoint.sh inside the Dockerfile.

### DIFF
--- a/docs/syncserver/Dockerfile
+++ b/docs/syncserver/Dockerfile
@@ -13,16 +13,14 @@ FROM alpine:3.21.0
 
 ARG SYNC_PORT=8080
 
-RUN adduser -D -h /home/anki anki
-
-RUN mkdir -p /anki_data && chown -R anki /anki_data
+# Default PUID and PGID values (can be overridden at runtime). Use these to
+# ensure the files on the volume have the permissions you need.
+ENV PUID=1000
+ENV PGID=1000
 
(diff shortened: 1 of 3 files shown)


dae:
I don't have much experience with using docker across heterogenous environments, so I'll defer to @jeankhawand here, but it seems ok at first glance from my end.

jeankhawand:
@dae  LGTM!

dae:
Thank you both!

PR #3715, opened 2025-01-11
Include SYNC_USER1 in syncserver Dockerfile as documentation

omarkohl:
**Note:**  I recommend reviewing only the second commit of this PR since the first one is a separate PR #3714 . I will rebase things as needed. (a.k.a. stacked diffs).

Graphical Docker tools frequently display the exposed ENV variables in the UI so it's user-friendly to display this to show the user it can (and should) be set.

### DIFF
--- a/docs/syncserver/Dockerfile
+++ b/docs/syncserver/Dockerfile
@@ -13,16 +13,14 @@ FROM alpine:3.21.0
 
 ARG SYNC_PORT=8080
 
-RUN adduser -D -h /home/anki anki
-
-RUN mkdir -p /anki_data && chown -R anki /anki_data
+# Default PUID and PGID values (can be overridden at runtime). Use these to
+# ensure the files on the volume have the permissions you need.
+ENV PUID=1000
+ENV PGID=1000
 
 COPY --from=builder /anki-server/bin/anki-sync-server /usr/local/bin/anki-sync-server
 
-
-RUN apk update && apk add --no-cache bash && rm -rf /var/cache/apk/*
-
-USER anki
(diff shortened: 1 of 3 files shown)


dae:
I'm not really keen on the idea of having a default password set, as it's not great security practice. If you must include something in the dockerfile, perhaps we should have code on the server end that refuses to run when the default password hasn't been changed.

@jeankhawand heads up

omarkohl:
I agree that there is risk associated with setting a default password.

Preventing the server from starting when the default password is used would be worse than discarding this PR in my opinion, since it would require users for whom the container is not starting to be able to read and correctly interpret the log output of the container, which is even more of a hassle than reading the documentation to figure out how to set a user and password.

The reason why I tend to think that in this case the default password is fine:
* The main use case for this container is private / at home usage. It seems unlikely (though not impossible!) that someone would run it completely unprotected in a public network.
* A user has to type in the password into their Anki devices before being able to sync. The password is "veryinsecure!" for a reason, namely to make them reconsider changing that password.

dae:
I still have a mild preference for not using a default password, but as it's more about the principle than a serious risk in this case, I don't feel strongly enough about this to the block the change. Just waiting on @jeankhawand's feedback on these changes now.

jeankhawand:
@dae Exactly. Setting a default password introduces significant security risks. As @omarkohl intent is to make this image more or less “ephemeral.” However, setting a default password undermines that goal and exposes users to vulnerabilities, especially if the image is reused or misconfigured.

I strongly prefer requiring users to set `SYNC_USER1` when running the container. This approach eliminates the risk of default credentials being exploited, enforces better security practices, and ensures that the container configuration remains intentional and explicit.

omarkohl:
Fair enough

PR #3716, opened 2025-01-11
Ignore SYNC_PORT and SYNC_BASE in syncserver Dockerfile

omarkohl:
**Note**: I recommend reviewing only the last commit of this PR since the first one is a separate PR #3714 and the second one also is #3715  . I will rebase things as needed. (a.k.a. stacked diffs).

Hardcode them to:

    SYNC_PORT=8080
    SYNC_BASE=/anki_data

If these env variables are passed into the container with different values,
they are ignored.

The reasons is if the user modifies SYNC_BASE they risk data loss since
anki-sync-server will no longer write data into the volume. If they change
SYNC_PORT they need to also change it when mapping this internal port to the
external port of the container, which could be confusing plus it has no benefit
to allow this since it's always possible to change the external port even if
the internal port is fixed to 8080 (e.g. `-p 1234:8080`).

In both cases there is no benefit to making these values configurable and there
are risks associated.

Unfortunately there is no easy way of implementing this for the
Dockerfile.distroless so it's up to the user not to modify these values.

### DIFF
--- a/docs/syncserver/Dockerfile
+++ b/docs/syncserver/Dockerfile
@@ -11,8 +11,6 @@ anki-sync-server
 
 FROM alpine:3.21.0
 
-ARG SYNC_PORT=8080
-
 # Default PUID and PGID values (can be overridden at runtime). Use these to
 # ensure the files on the volume have the permissions you need.
 ENV PUID=1000
@@ -22,11 +20,8 @@ COPY --from=builder /anki-server/bin/anki-sync-server /usr/local/bin/anki-sync-s
 
 RUN apk update && apk add --no-cache bash su-exec && rm -rf /var/cache/apk/*
 
-ENV SYNC_PORT=${SYNC_PORT}
 
-ENV SYNC_BASE=/anki_data
-
-EXPOSE ${SYNC_PORT}
+EXPOSE 8080
 
 COPY entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
@@ -37,7 +32,7 @@ CMD ["anki-sync-server"]
 # This health check will work for Anki versions 24.08.x and newer.
 # For older versions, it may incorrectly report an unhealthy status, which should not be the case.
 HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
-    CMD wget -qO- http://127.0.0.1:${SYNC_PORT}/health || exit 1
+    CMD wget -qO- http://127.0.0.1:8080/health || exit 1
 
 VOLUME /anki_data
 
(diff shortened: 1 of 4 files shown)


dae:
Seems reasonable. @jeankhawand?

jeankhawand:
@dae LGTM! thanks @omarkohl I am wondering for users that will migrate to this newer build they will for sure lose their existing decks... In this case it's advised for them when opening anki app to sync their local to AnkiWeb otherwise they will loose their decks.

omarkohl:
> Thanks, @omarkohl. I’m wondering about existing users migrating to this newer build—they will likely lose their existing decks. In this case, I believe Anki will check for diffs and display a prompt asking users whether to sync locally or fetch changes from AnkiWeb. Therefore, it’s recommended that users sync their local data to AnkiWeb.

Before my commit b1898202 from a little over a week ago no data was stored in a volume, so removing the container and starting a new one based on a new image would lead to data loss. That is why I created that commit, because otherwise data loss for users trying to upgrade their image was almost unavoidable (probably the only exception is if they manually created a volume for `/home/anki/.syncserver`). Upgrading the docker container was basically not possible., Now it is possible, even though upgrading from before b1898202 to after it **remains** a little tricky. Or am I missing something?

Asking users to force sync from device to server is probably the easiest solution, yes. Are you suggesting to include that in the README or what were you trying to say?

jeankhawand:
> > Thanks, @omarkohl. I’m wondering about existing users migrating to this newer build—they will likely lose their existing decks. In this case, I believe Anki will check for diffs and display a prompt asking users whether to sync locally or fetch changes from AnkiWeb. Therefore, it’s recommended that users sync their local data to AnkiWeb.
> 
> Before my commit b1898202 from a little over a week ago no data was stored in a volume, so removing the container and starting a new one based on a new image would lead to data loss. That is why I created that commit, because otherwise data loss for users trying to upgrade their image was almost unavoidable (probably the only exception is if they manually created a volume for `/home/anki/.syncserver`). Upgrading the docker container was basically not possible., Now it is possible, even though upgrading from before b1898202 to after it **remains** a little tricky. Or am I missing something?
> 
> Asking users to force sync from device to server is probably the easiest solution, yes. Are you suggesting to include that in the README or what were you trying to say?

Could you please mention it in Readme so the users can be aware about that and once they migrate to newer image they must sync their local. For Distroless it will be a bit tricky cause there is no tty session dunno if `docker cp` will work so they can copy data from default to `/anki_data`

omarkohl:
done

dae:
Thanks Omar, happy to merge this in once the conflicts are addressed.

PR #3717, opened 2025-01-11
Clear memory states during bulk action if item is None

user1823:
Prevents issues like https://forums.ankiweb.net/t/suggestion-copy-card-debug-info-button/54206/10 and https://github.com/ankitects/anki/issues/3634

If item is None, the previous code uses the card state to infer memory states, which leads to a vicious cycle that keeps changing the memory states at every reschedule if desired retention and historical retention are not the same.

@L-M-Sherlock, WDYT?

### DIFF
--- a/rslib/src/scheduler/fsrs/memory_state.rs
+++ b/rslib/src/scheduler/fsrs/memory_state.rs
@@ -163,11 +163,20 @@ impl Collection {
             historical_retention,
             ignore_revlogs_before_ms_from_config(&config)?,
         )?;
-        card.set_memory_state(&fsrs, item, historical_retention)?;
-        Ok(ComputeMemoryStateResponse {
-            state: card.memory_state.map(Into::into),
-            desired_retention,
-        })
+        if item.is_some() {
+            card.set_memory_state(&fsrs, item, historical_retention)?;
+            Ok(ComputeMemoryStateResponse {
+                state: card.memory_state.map(Into::into),
+                desired_retention,
+            })
+        } else {
+            card.memory_state = None;
+            card.desired_retention = None;
+            Ok(ComputeMemoryStateResponse {
+                state: None,
+                desired_retention,
+            })
+        }
     }
 }
 
--- a/rslib/src/scheduler/fsrs/memory_state.rs
+++ b/rslib/src/scheduler/fsrs/memory_state.rs
@@ -85,8 +85,8 @@ impl Collection {
                 progress.update(true, |state| state.current_cards = idx as u32 + 1)?;
                 let mut card = self.storage.get_card(card_id)?.or_not_found(card_id)?;
                 let original = card.clone();
-                if let Some(req) = &req {
-                    card.set_memory_state(&fsrs, item, historical_retention.unwrap())?;
+                if let (Some(req), Some(item)) = (&req, item) {
+                    card.set_memory_state(&fsrs, Some(item), historical_retention.unwrap())?;
                     card.desired_retention = desired_retention;
                     // if rescheduling
                     if let Some(reviews) = &last_revlog_info {
--- a/rslib/src/scheduler/fsrs/params.rs
+++ b/rslib/src/scheduler/fsrs/params.rs
@@ -427,7 +427,7 @@ fn revlog_entry_to_proto(e: RevlogEntry) -> anki_proto::stats::RevlogEntry {
 pub(crate) mod tests {
     use super::*;
 
-    const NEXT_DAY_AT: TimestampSecs = TimestampSecs(86400 * 100);
+    const NEXT_DAY_AT: TimestampSecs = TimestampSecs(86400 * 1000);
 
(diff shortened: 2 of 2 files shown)


user1823:
I just noticed that the FSRS Helper add-on uses `compute_memory_state` for updating the memory states while Anki uses `update_memory_state` for bulk actions. So, my patch won't work when rescheduling cards using the add-on (by selecting cards in the browser and hitting `Update memory cards & reschedule`). Any suggestions for that?

Basically, the goal is that the card should get memory state when item is None only during reviewing. In all other conditions, the memory state should not be calculated if item is None.

L-M-Sherlock:
Does the problem mentioned in https://forums.ankiweb.net/t/suggestion-copy-card-debug-info-button/54206/10 still exist after #3634? @Luc-Mcgrady, could you help me confirm it?

Luc-Mcgrady:
I'm not sure what fixed it, but it seems fixed now even on the main branch?

Main branch + this branch:
image

Last release:
image

user1823:
What happens if you
- use different values for desired retention and historical retention,
- set the ignore cards reviewed before to a future date,
- make a minor modification to the parameters,
- enable Reschedule cards on change, and then
- save the deck options?

With and without this change

Luc-Mcgrady:
Future ignore-reviews-before :
- without (release)  HR = 0.9 DR = 0.83:
image
- without (main branch)  HR = 0.9 DR = 0.83:
image
- with:
- - HR = 0.9 DR = 0.83
image
- - HR = 0.5 DR = 0.7 ( same memory state :white_check_mark: )
image
- without (main branch (again))  HR = 0.5 DR = 0.7 :
image

I'm sorry I missed the point of the PR and just checked if it fixed my card :sweat_smile: 
LGTM

user1823:
Something doesn't look right. With the changes in this PR, I expected the card to have **no memory states** and interval to be the **same as that before rescheduling**.

Just to confirm, did you use Anki's "Reschedule cards on change" and not the helper add-on's rescheduling feature?

user1823:
> If ignore reviews before is in the future then I doubt having my specific card would matter for this?

You are right, it doesn't matter. But, I don't know how to build from source code. So, some help in testing would be appreciated.

Also see edits to my comment above.

Luc-Mcgrady:
```rs
 &items[0] = (
    CardId(
        1662864375170,
    ),
    Some(
        FsrsItemForMemoryState {
            item: FSRSItem {
                reviews: [
                    FSRSReview {
                        rating: 2,
                        delta_t: 0,
                    },
                    // ...
                    FSRSReview {
                        rating: 3,
                        delta_t: 63,
                    },
                ],
            },
            starting_state: Some(
                MemoryState {
                    stability: 0.9999996,
                    difficulty: 2.0842972,
                },
            ),
        },
    ),
)
[rslib/src/scheduler/fsrs/memory_state.rs:84:13] ignore_before = TimestampMillis(
    1741824000000,
)
```

ignore reviews before doesn't empty the revlogs while not re-training.

Is not being able to build from source something I can help with?

user1823:
> ignore reviews before doesn't empty the revlogs while not re-training.

That seems to be a bug (provided that the ignore before date is in future).

> Is not being able to build from source something I can help with?

I never really tried to set that up. Thanks for the offer, but let me try first.

user1823:
The latest changes try to fix the issue you highlighted (ignore reviews before doesn't empty the revlogs while not re-training if ignore before date is set to a future date.)

The test failure is not related to my changes.

Can you please test this again? I didn't succeed building Anki from source code. It was showing some network error. I will try again later.

Luc-Mcgrady:
image
Seems to work now :+1:

user1823:
> The test failure is not related to my changes.

It actually might be, but I really don't understand why. In this test, how can my changes cause the item to be None?

https://github.com/ankitects/anki/blob/39e293b27d36318e00131fd10144755eec8d1922/rslib/src/scheduler/fsrs/memory_state.rs#L387-L411

Luc-Mcgrady:
```
[rslib/src/scheduler/fsrs/params.rs:281:9] user_graded = true
[rslib/src/scheduler/fsrs/params.rs:281:9] within_cutoff = false
thread 'scheduler::fsrs::memory_state::tests::bypassed_learning_is_handled' panicked at rslib/src/scheduler/fsrs/memory_state.rs:400:10:
called `Option::unwrap()` on a `None` value
```
within_cutoff is false :thinking:

user1823:
> because the revlogs ID is 0

But, how can that be? The following should set it to be 100 days before today.
revlog(RevlogReviewKind::Review, 100)

Luc-Mcgrady:
```rs
// params.rs line 432
    const NEXT_DAY_AT: TimestampSecs = TimestampSecs(86400 * 100);
```

the revlogs id's are based on 100 days in the future, just put an extra 0 on there

user1823:
Thanks Luc for all the help in debugging and testing.

@dae, this is ready for a final review now.

dae:
Thank you both!

PR #3718, opened 2025-01-11
Fix "Note Types" dialog moving down each time it is opened

rbrownwsws:
Fixes this bug:

https://forums.ankiweb.net/t/bug-note-types-window-moves-down-every-time-it-is-opened/54248

Setting window flags seem to make the window move down by the title bar height, so you have to restore geometry after setting them.

N.B. There may be similar bugs elsewhere, I have not checked all dialogs.


dae:
Thank you!

PR #3719, opened 2025-01-11
"Copy template as markdown" button.

Luc-Mcgrady:
image

Example copied text:
**Front Template**

```html
{{Front}}
```

**Back Template**

```html
{{FrontSide}}

<hr id=answer>

{{Back}}
```

**Styling**

```css
.card {
    font-family: arial;
    font-size: 20px;
    text-align: center;
    color: black;
    background-color: white;
}
```

related(?) forum post: https://forums.ankiweb.net/t/suggestion-copy-card-debug-info-button/54206/9


Danika-Dakika:
Sorry for hijacking your post, but I'm really excited about this now! As far as placement -- 

Is there a reason you put it between the existing "Deck Override" and "Browser Appearance" options, as opposed to after them? 

Or could this be something that would be used frequently enough that it might merit a button in the window? [I'm raising that as a genuine question -- not a demand.] "Copy as Markdown" might look too long there, but "Copy" might be misleading if someone thought they could copy it for another purpose (like copying a template/type to another note type). Maybe "Copy for Help" and pairing it with the "Help" button on the left?
image

Luc-Mcgrady:
There's no reason I put it there it was just convenient. I'd assumed this would be a bit of a niche feature so tucked it away a little bit. That and I was trying to emulate Anki-droid's example as closely as possible.

dae:
Looks like a useful addition, thank you!

I think it's probably going to contribute to clutter down the bottom, so the dropdown does seem like it might be the best place. Regarding the wording, how about "Copy to Clipboard" or similar? 'Markdown' feels like it's more technical than it needs to be, as users wanting support don't need to be familiar with that technology.

Luc-Mcgrady:
I would prefer to call it "Copy to Clipboard" but noticed it was called "Copy as Markdown" in ankidroid so decieded to keep it consistant.

If i'm going to split hairs, something else i'd like to change from the ankidroid implmentation is i'd prefer:

`##`
## Front Template 

to

`**`
**Front Template**

dae:
As this is a debug tool, consistency is probably not of the upmost importance. Maybe @ankidroid would be interested in tweaking their wording, but they can of course keep it the same if they'd prefer.

Likewise for the formatting - no preferences on my end.

ftl/core/card-templates.ftl
```diff
@@ -37,6 +37,7 @@ card-templates-card = Card { $val }
 card-templates-card-types-for = Card Types for { $val }
 card-templates-cloze = Cloze { $val }
 card-templates-deck-override = Deck Override...
+card-templates-copy-info = Copy Info to Clipboard
```
dae, line 40:
```suggestion
card-templates-copy-info = Copy to Clipboard
```

dae: Is 'info' really adding much here?

jcznk:
It seems this button is not added to Cloze and Image Occlusion note types (Anki 25.02 Windows 11). Not sure if this is intended.

Luc-Mcgrady:
Oops!

PR #3710, opened 2025-01-09
Fix menubar in fullscreen

YukiNagat0:
Fixes: https://github.com/ankitects/anki/issues/3647

As alternative to: https://github.com/ankitects/anki/pull/3682

Demo: 

https://github.com/user-attachments/assets/416fd503-73a0-4ab3-81fa-84f04a92ba93

PR #3721, opened 2025-01-12
Refactor/Small refactoring on fullscreen and comments for clarification

YukiNagat0:
Follow-up to https://github.com/ankitects/anki/pull/3710.

Fixes the quirk with the `_reviewState` method, that I have described in the https://github.com/ankitects/anki/pull/3710#issuecomment-2579704897


dae:
Thanks Yuki!

PR #3722, opened 2025-01-12
Unpin svelte and update sveltekit-svg

wackbyte:
Warnings from enums were fixed in sveltejs/svelte#14192 (and then the warning was later entirely removed in sveltejs/svelte#14663)


dae:
Thank you for this!

PR #3723, opened 2025-01-12
Reduce use of type casting

wackbyte:
replaces some redundant and brittle `as` casts with more straightforward forms


dae:
In a few of these cases I feel like it's more personal preference than anything else, but on the whole it looks like a nice improvement - thank you again!

Issue #3720, opened 2025-01-12
Use platform-native button ordering


dae:
> I’m over 80, so I now rely on Anki as my major alternate hippopotamus (-campus ?)
> I would be most grateful if someone could write an app to interchange the “Add” and “Close” buttons on the “Add” dialog (which is presented for adding/editing a card).
> In the current configuration, I spend 98.37% of my Anki time reading the “Discard current input?” dialog.
> I might not be alone:
> "AI Overview
> Learn more
> Primary & Secondary Action Buttons | by Nick Babich | UX Planet
> The desired action button in a dialog box should generally be placed on the right side. This is because it follows the natural reading pattern from left to right, and aligns with the user’s expectation of progressing forward."
> Thanks

Currently the discard action is on the left, which is correct for macOS, but on Windows the opposite is standard. I'm not sure if there's a standard on Linux.

Originally reported on https://forums.ankiweb.net/t/add-on-required-to-interchange-the-add-and-close-buttons/54196

PR #3725, opened 2025-01-12
Use platform-native button layout in dialogs and messageboxes

iamllama:
Relevant to #3720

Qt 5 and 6 provide a `button-layout` [stylesheet prop](https://doc.qt.io/qt-6/stylesheet-reference.html#button-layout) to specify which platform's layout should be used in `QDialogButtonBox`, which is also implicitly used in messageboxes. This pr proposes to use that prop

On win, this is a no-op (i suspect the same is true on macs). On linux (my machine at least), it seems that the qt fusion theme's `QStyle::SH_DialogButtonLayout` style hint defaults to 0, i.e the win layout. According to [this doc comment](https://github.com/qt/qtbase/blob/6.8.1/src/widgets/widgets/qdialogbuttonbox.cpp#L574), it doesn't seem to be consistently set automatically.

EDIT: It's correctly set to 3 (gnome) when in a ubuntu vm, meaning it's a no-op there as well. This pr's only real outward-facing change might be on enforcing the gnome layout on linux systems that don't set it, since qt already uses the platform-native layout in other cases

On linux, I opted to default to gnome's layout instead of kde's to avoid the problem of desktop env detection (unless someone knows how 👀)

Comparing the addcards and fields dialogs on linux:

Before (same as win):
image
image

After:
image
image

Unfortunately, i don't think this will help the [forum post](https://forums.ankiweb.net/t/add-on-required-to-interchange-the-add-and-close-buttons/54196)'s OP in any way.

Here's how qt6 decides platform-native button ordering: https://github.com/qt/qtbase/blob/6.8.1/src/gui/kernel/qplatformdialoghelper.cpp#L58, tldr:
```
Win:   Accept - Destructive - Action - Reject - Help
Mac:   Help - Action - Destructive - Reject - Accept
Gnome: Help - Action - Destructive - Reject - Accept
Kde:   Help - Action - Accept - Destructive - Reject
```
where the "Add" button is `Action`, "Discard" is `Destructive`, "Close" and "Keep editing" are `Reject`

In all cases, "Discard" comes before "Keep editing", so that's already in line with all native behaviours, unless we rethink what roles to assign those buttons (should "Discard" be given the `Destructive` role in a modal asking for confirmation to discard?)

Likewise, "Add" is always to left of "Close"

### DIFF
--- a/qt/aqt/stylesheets.py
+++ b/qt/aqt/stylesheets.py
@@ -1,7 +1,7 @@
 # Copyright: Ankitects Pty Ltd and contributors
 # License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
-from anki.utils import is_win, is_mac
+from anki.utils import is_mac, is_win
 from aqt import colors, props
 from aqt.theme import ThemeManager
 
--- a/qt/aqt/stylesheets.py
+++ b/qt/aqt/stylesheets.py
@@ -1,7 +1,7 @@
 # Copyright: Ankitects Pty Ltd and contributors
 # License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
-from anki.utils import is_win
+from anki.utils import is_win, is_mac
 from aqt import colors, props
 from aqt.theme import ThemeManager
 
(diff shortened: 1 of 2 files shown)


dae:
I'm not sure what the best way to solve the macOS issue is. I've always assumed the Accept role is only intended for when closing the dialog, but if that's not the case, we could argue that the user is "accepting" the current card.

PR #3726, opened 2025-01-13
set media folder mtime on sync

jhhr:
Related to https://forums.ankiweb.net/t/edited-media-files-auto-sync/54293/2?u=jhhr

Since it's so easy to enable this, wouldn't it make sense to make this default sync behaviour? I couldn't find one right now, but I recall there's been a thread or two on the forums with users wondering why their edits to media files didn't get synced. So, users probably expect syncing to work this way.


dae:
It's deliberate, and when we tried reverting the mtime check, it caused regressions.

35cbde65d7c5cf17ce0c75e73648ba348f5a4dc3

jhhr:
Ok, shame. I guess it'll have to stay an addon then.

PR #3730, opened 2025-01-16
Fix "Create copy" for IO notes

### LINKED ISSUE #3247: Using “Create copy…” in Image Occlusion card doesn’t copy the image
> I wanted to create a copy of an “Image Occlusion” card. I used the “Create Copy…” button, but the window “Add” didn’t include the image.

Originally reported on https://forums.ankiweb.net/t/using-create-copy-in-image-occlusion-card-doesnt-copy-the-image/46037

iamllama:
Closes #3247

~~The mask editor is now provided with the copied note's image if it exists (which should almost always be the case)~~

The copied note's image and occlusions now carry over to the new note, along with its tags, occlusion mode, misc. fields etc


abdnh (approved):
Looking good.

Should we maybe also copy masks?

dae:
Thank you both!

PR #3733, opened 2025-01-17
Add support for copied image files when adding IO notes

### LINKED ISSUE #2772: Allow getting a copied file with the `Paste image from Clipboard` I/O option
Apparently, `Paste Image from Clipboard` only works if there is a image bitmap in my clipboard. But if the user has a image file copied, it doesn't work.

So, as a QOL feature, it would be nice to be able to get the copied file by just pressing the button instead of being forced to use the `Select Image` button

iamllama:
Closes #2772

Adds support for images copied from file managers (specifically, the `text/uri-list` mime format) when clicking the "Paste Image from Clipboard" button

Chose to leave doPaste and doDrop alone to avoid interfering with the misc. (header etc.) fields' existing copy-paste behaviour

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -1100,6 +1100,13 @@ require("anki/ui").loaded.then(() => require("anki/NoteEditor").instances[0].too
 
         self.parentWindow.activateWindow()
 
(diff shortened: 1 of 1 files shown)


dae:
Thank you!

PR #3736, opened 2025-01-18
Revert Editor.set_note's signature change with an alternative approach for #3730

iamllama:
Reported in https://discord.com/channels/368267295601983490/368267296046841857/1330000729582211185

Changing the signature of `Editor.set_note` in #3730 broke the [Cloze (Hide All)](https://ankiweb.net/shared/info/1709973686) addon, and it might not be the only one :/

This pr proposes adding a field to `Note` to keep track of the original note it was copied from (if any), which doesn't get [passed to the backend](https://github.com/ankitects/anki/blob/7ba143bad8cfd60bf865c1274c1fc3e6c3d3886a/pylib/anki/notes.py#L72), and using that instead of drilling it through to `Editor.loadNote`

### DIFF
--- a/pylib/anki/notes.py
+++ b/pylib/anki/notes.py
@@ -32,6 +32,8 @@ class Note(DeprecatedNamesMixin):
     data = ""
     id: NoteId
     mid: NotetypeId
+    # used when creating a copy of an existing note
+    orig_note_id: NoteId | None = None
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


pylib/anki/notes.py
```diff
@@ -32,6 +32,8 @@ class Note(DeprecatedNamesMixin):
     data = ""
     id: NoteId
     mid: NotetypeId
+    # used when creating a copy of an existing note
+    orig_note_id: NoteId | None = None
```
dae, line 36: Adding a property here for an editor-specific feature feels a bit unclean. What about storing it in the editor instance instead?

dae:
Thanks!

PR #3742, opened 2025-01-18
Fix/re-optimize FSRS if short-term param is weird

L-M-Sherlock:
Previous discussion: https://github.com/open-spaced-repetition/fsrs4anki/issues/708#issuecomment-2586399698

Current problem: I don't know how to refresh the progress for the second optimization.

### DIFF
--- a/proto/anki/scheduler.proto
+++ b/proto/anki/scheduler.proto
@@ -346,6 +346,7 @@ message ComputeFsrsParamsRequest {
   string search = 1;
   repeated float current_params = 2;
   int64 ignore_revlogs_before_ms = 3;
+  uint32 num_of_relearning_steps = 4;
 }
 
 message ComputeFsrsParamsResponse {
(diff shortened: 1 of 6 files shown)


user1823:
I just realized that there are two issues with this approach:
- Because PLS depends upon the previous stability and previous difficulty, it is possible that the weird behaviour occurs for some values of S and D but not for others. Here, we are checking for one particular set of S and D values.
- Here, we are reading the number of relearning steps from the preset. But, what if the user has decreased the number of steps and the older revlogs show the weird behaviour? Such cases will be missed by this approach.

Another issue (easily fixable):
We want to count only the same-day relearning steps. Multiday steps won't cause the weird behaviour.

L-M-Sherlock:
> * Because PLS depends upon the previous stability and previous difficulty, it is possible that the weird behaviour occurs for some values of S and D but not for others. Here, we are checking for one particular set of S and D values.

If the weird behavior doesn't occur for S = 1 and D = 1, it will not occur for S > 1 and D > 1 because S_f / S becomes less.

dae:
Look ok @L-M-Sherlock?

L-M-Sherlock:
debug

@dae, the progress always flickers once I click `Optimize`.

Expertium:
I'm late, but I have a question - have you guys tested whether it works correctly with "Optimize all"?

L-M-Sherlock:
> I'm late, but I have a question - have you guys tested whether it works correctly with "Optimize all"?

Yep, I have tested it.

PR #3744, opened 2025-01-18
FIX lang selection resetting to en_US for some langs

GithubAnon0000:
Fixes https://forums.ankiweb.net/t/anki-25-01-beta/54490/17?u=anon_0000.

# Issue
Set a hand full of certain languages in the preferences screen and see that the translations have been applied after reboot. The language selection in preferences wrongly shows en_US though, not the current active language. If you wanted to switch to `en_US` in this case, then you'd have to first switch to a working language (like de_DE) and then switch to en_US.

I tested every single language manually and found that only 4 were not working as expected (see changes I want to commit).

# Solution
`anki/qt/aqt/preferences.py` has the functions `setup_language()` and `current_lang_index()`. I noticed that it defaults to en_US, if the language is not in `compatMap` and it couldn’t return the index of the current language. No idea if this code is faulty but I headed over to `anki/pylib/anki/lang.py` afterwards.

Here, in `compatMap`, I added e.g. `"la": "la_LA"`. I knew the code since I could get it with `print("––– lang is ", lang)` in `preferences.py` (`current_lang_index()` retrieves `la` for latin).

After adding those code changes from my PR, the problem for those selected languages had gone away.

No idea if that's best practices though or if something else should be fixed instead.


dae:
Thank you!

PR #3745, opened 2025-01-18
Add translation strings for True Retention table

rbrownwsws:
Based on this discussion on the forum: https://forums.ankiweb.net/t/anki-25-01-beta/54490/5

I reused translation strings from the Cards Counts chart in the redesign of the True Retention table as it meant less translation work was required.

Unfortunately in some languages these translations are quite long, which can make the table overflow and need scroll bars if the card is small:

image

This PR adds new translation strings specific to the True Retention table so you can do something like this:

```
statistics-counts-young-cards = Junge Karten
statistics-true-retention-young = Jung
```

so you can be verbose where you have plenty of space, and terse where you do not.

An unfortunate downside of this is it requires new translations to be added for all languages, which will often be identical to the ones from Card Counts and mean less maintained languages may not get a translation for a while.

**This should probably not be merged without input from people more involved in the translation of Anki.**

### DIFF
--- a/ftl/core/statistics.ftl
+++ b/ftl/core/statistics.ftl
@@ -103,14 +103,25 @@ statistics-counts-separate-suspended-buried-cards = Separate suspended/buried ca
 ## comparison to the "desired retention" parameter of FSRS, which forecasts
 ## future retention. True Retention is the percentage of all reviewed cards
 ## that were marked as "Hard," "Good," or "Easy" within a specific time period.
+##
+## Most of these strings are used as column / row headings in a table.
+## (Excluding -title and -subtitle)
+## It is important to keep these translations short so that they do not make
+## the table too large to display on a single stats card.
+##
+## N.B. Stats cards may be very small on mobile devices and when the Stats
+##      window is certain sizes.
 
 statistics-true-retention-title = True Retention
 statistics-true-retention-subtitle = Pass rate of cards with an interval ≥ 1 day.
 statistics-true-retention-range = Range
 statistics-true-retention-pass = Pass
 statistics-true-retention-fail = Fail
+statistics-true-retention-total = { statistics-counts-total-cards }
 statistics-true-retention-count = Count
 statistics-true-retention-retention = Retention
+statistics-true-retention-young = { statistics-counts-young-cards }
+statistics-true-retention-mature = { statistics-counts-mature-cards }
 statistics-true-retention-all = All
 statistics-true-retention-today = Today
 statistics-true-retention-yesterday = Yesterday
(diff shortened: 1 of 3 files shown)


Danika-Dakika:
I'm not worried about new strings being added -- as long as you try your best to re-use available strings [whenever possible](https://translating.ankiweb.net/anki/developers.html). 

Someday we'll finish TR -- but I don't think these strings will delay that! 

Since you know these strings have space limitations, if you can figure out how to add a group/resource comment for translators, that would be great. 

image

[Translators already get the [string character-length for comparison](https://translating.ankiweb.net/anki/core.html#simple-replacements).]

rbrownwsws:
@Danika-Dakika

> Since you know these strings have space limitations, if you can figure out how to add a group/resource comment for translators, that would be great.

I've already tried to add comments, hopefully I did it correctly.

I added this to the existing group comment.

**ftl/core/statistics.ftl:**

```diff
  ## True Retention represents your actual retention rate from past reviews, in
  ## comparison to the "desired retention" parameter of FSRS, which forecasts
  ## future retention. True Retention is the percentage of all reviewed cards
  ## that were marked as "Hard," "Good," or "Easy" within a specific time period.
+ ##
+ ## Most of these strings are used as column / row headings in a table.
+ ## (Excluding -title and -subtitle)
+ ## It is important to keep these translations short so that they do not make
+ ## the table too large to display on a single stats card.
+ ##
+ ## N.B. Stats cards may be very small on mobile devices and when the Stats
+ ##      window is certain sizes.
```

I also added comments like this to hint that they should be similar to the Card Counts values:

```diff
+ # This will usually be the same as statistics-counts-young-cards
+ statistics-true-retention-young = Young
```

snowtimeglass:
Looks nice to me, too. I thank you because the new string for "Total" here would also be useful in Japanese for expressing the meaning of "Total" in this situation, which is distinguished from simple sum of numbers. 

----

> An unfortunate downside of this is it requires new translations to be added for all languages, which will often be identical to the ones from Card Counts and mean less maintained languages may not get a translation for a while.


I guess the solution in the following case would work for this case, as well.
 
- Inquiry about adding one more string for “Added”
  https://forums.ankiweb.net/t/inquiry-about-adding-one-more-string-for-added/41998

  - `Clone "added" string for importing`
    https://github.com/ankitects/anki/commit/184ad8068ee2b12abb9163ddbe023c97e2331e84　


I mean, the translations for the existing strings would be copied (cloned) into the new strings in Pontoon. And then, you, I, and others would rewrite the translations as necessary.

dae:
Thank you! I will copy the existing translations into the new strings after merging this to avoid creating work for translators.

dae:
https://github.com/ankitects/anki-core-i18n/commit/5c28baf8e68542135ff7f5e05a7f613a46424793

PR #3747, opened 2025-01-19
More accurate sorting by R

user1823:


user1823:
The `+0.001` is for the case when elapsed_days = scheduled_days. Adding 0.001 ensures cards with shorter intervals show up before the cards with longer intervals. An alternative is to use a second order (ivl) when the cards have the same relative overdueness, but that probably would be an overkill.

user1823:
Based on my experience with current stable and latest beta, it seems that when the deck has a mix of cards with and without memory states (e.g. due to moving some cards between decks), the cards without memory states show up first.

After this change, Anki should be able to merge the two queues to generate a unified list ordered by relative overdueness, irrespective of whether the card has memory states or not. But, I don't know how to merge those queues. So, please make that change yourself or create an issue for the same.

Expertium:
> when the deck has a mix of cards with and without memory states (e.g. due to moving some cards between decks)

Remind me, why aren't memory states recalculated when moving cards between decks?

user1823:
> Remind me, why aren't memory states recalculated when moving cards between decks?

According to dae, it is because of performance concerns. Moving cards is a fairly common action. Recalculating memory states every time would make Anki feel slower.

rslib/src/storage/card/mod.rs
```diff
@@ -774,8 +774,9 @@ impl fmt::Display for ReviewOrderSubclause {
             ReviewOrderSubclause::DifficultyAscending => "extract_fsrs_variable(data, 'd') asc",
             ReviewOrderSubclause::DifficultyDescending => "extract_fsrs_variable(data, 'd') desc",
             ReviewOrderSubclause::RetrievabilitySm2 { today, order } => {
+                // - (elapsed days+0.001)/(scheduled interval)
```
dae, line 777:
```suggestion
```

user1823: No hard feelings, but I think that a comment makes understanding the formula easier.

dae:
> After this change, Anki should be able to merge the two queues to generate a unified list ordered by relative overdueness, irrespective of whether the card has memory states or not. But, I don't know how to merge those queues. So, please make that change yourself or create an issue for the same.

There's no separate "queue" - the cards without memory state don't have the memory values that `add_extract_fsrs_relative_retrievability()`, so it returns null, which ends up being sorted to the top. To have those cards sorted properly, we'd either need their memory state, or some other calculation that approximates an appropriate position.

user1823:
> some other calculation that approximates an appropriate position.

After the changes I have made, the Sm2 retrievability should be roughly the same as the FSRS retrievability. So, we can use the Sm2 retrievability as the approximate for those cards.

rslib/src/storage/card/mod.rs
dae: Oh, I assumed that was stale code and not a deliberate comment. Could you move it under the format!( line then please?

dae:
Thanks!

PR #3748, opened 2025-01-19
Fix numeric deck and notetype names being treated as ids when importing csv

iamllama:
Reported in https://forums.ankiweb.net/t/import-problem-deck-field-is-ignored-since-version-24-11

Fixes numeric deck and notetype names being treated as ids in `#deck:...`, or in the specified deck/notetype column

`DeckIdsByNameOrId::get`, `Collection::deck_id_by_name_or_id` and `Collection::notetype_by_name_or_id` are modified to fallback to treating ids as names if said deck/notetype id is nonexistent

Modifying `Collection.deck_id_by_name_or_id` may seem redundant, as the metadata deck [goes through DeckIdsByNameOrId::get anyway](https://github.com/ankitects/anki/blob/1be94a8b0461defcbf2a2d39e5a1f316446b0169/rslib/src/import_export/text/import.rs#L162), but since its used by the import options page, it's kept for consistency when dealing with `#deck:<numeric>`

The only breakage I can think of is when an exporter exports notes from a deck with a name that's the same as one of the importer's decks' ids, in which case the deck with that id would be used instead of making a new deck with the id as the name. But this seems highly pathological, and is identical to existing behaviour regardless.

### DIFF
--- a/rslib/src/import_export/text/csv/metadata.rs
+++ b/rslib/src/import_export/text/csv/metadata.rs
@@ -688,6 +688,35 @@ pub(in crate::import_export) mod test {
     }
 
     #[test]
+    fn should_fallback_to_parsing_deck_ids_as_deck_names() {
+        let mut col = Collection::new();
+        let numeric_deck_id = col.get_or_create_normal_deck("123456789").unwrap().id.0;
+        let numeric_deck_2_id = col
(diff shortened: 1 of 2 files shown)


abdnh (approved):

dae:
Another good change - thanks again!

PR #3750, opened 2025-01-19
Auto generate contributors list (for about dialog)

GithubAnon0000:
https://forums.ankiweb.net/t/names-in-about-window-do-not-match-names-in-contributing/54570?u=anon_0000

This python script automatically parses the CONTRIBUTORS file and formats it, so that about.py can use it.

about.py now includes about_contributors.py, which was generated by that python script.

_Note:_
1. There are a lot of print functions commented out which I used while developing. Didn't know whether I should remove those.
2. Pycharm only shows 4 "typos" for the word "allusers". No other warnings, erros, ect.
3. I cannot test `./ninja check` with the latest main, since it refuses to build for me (I used the branch from current stable as a base). I also cannot run in with the stable branch, since my vm has less than 50mb space left. If github doesn't auto check it, I'll do so once I increase my vms disk capacity.
4. Linting is probably wrong, but no idea how to lint according to `.pylintrc`.
5. Maybe the copyright notice at the top is incorrect? Not sure if there's a certain style I must follow.
6. @sommerluk appears two times in the list of contributors (once with noreply email and once with another email). Should one of them be removed @sommerluk ?

I did build this with the current stable branch and checked that it works. So probably no issues with the beta either.

### DIFF
--- a/tools/generate_contributors_list.py
+++ b/tools/generate_contributors_list.py
@@ -7,7 +7,7 @@ import re
 
 # Get path to input file (CONTRIBUTORS).
 input_file  = "../CONTRIBUTORS"
-# Get path to output file (CONTRIBUTORS_formatted).
+# Get path to output file (about_contributors.py).
 output_file = "../qt/aqt/about_contributors.py"
 # Template file which is used to create the `allusers` function in about.py
 tmp_allusers_template = """#This list has been generated from CONTRIBUTORS file.


user1823:
Related:
- https://github.com/ankitects/anki/pull/3575

user1823:
I think so, yes.
But maybe, there should be a comment in one or both of these files to indicate that they are not intended to be identical so that other people won't create such PRs.

Issue #3635, opened 2024-12-14
Force use of Yarn Classic to build Anki


dae:
> I had quite a trouble trying to build Anki, and the cause was because Yarn Modern (Berry) was being used instead of Yarn Classic (1.x).
> At first, I didn’t realize that wrong Yarn version was the cause. The build fails while building canvas@npm:1.0.0 package, with the message that 'cairo/cairo.h' file not found.

Originally reported on https://forums.ankiweb.net/t/use-yarn-classic-to-build-anki/52919/4

Possibly solution: https://forums.ankiweb.net/t/use-yarn-classic-to-build-anki/52919/5?u=dae

mikehardy:
Just for full cross-linking purposes, I succeeded with the proposed solution of pinning the yarn version for corepack-enabled systems and posted PR #3752 

I motion that this issue should be re-titled though - the title is prescriptive when it is actually just a workaround, and I believe it should be more like "Web build doesn't work with modern yarn", with my PR just being a temporary fix.

dae:
Closed by #3772

PR #3752, opened 2025-01-20
build: specify the canvas transitive node dependency build failure workaround in future-compatible way

mikehardy:
without this, the canvas dependency has compilation issues finding the cairo.h include file, on modern yarn even when it is present on the system

I originally tried  BlueGreenMagick's idea https://forums.ankiweb.net/t/use-yarn-classic-to-build-anki/52919/5

But I didn't like it, because it wasn't a fix-forward type solution. It just sticks you on old yarn which is slow and no longer maintained.

Fixes
- #3635

Tested locally on a macOS machine that has modern yarn (v4.6.0 currently) installed with corepack enabled and I was able to build despite seeing exactly what was described in 3635 prior

Tested locally on same macOS machine but after adding `"packageManager": "yarn@1.22.22",` to package.json to force old yarn and make sure it still worked.

I assume CI uses old yarn as well so will show if it works or not

In Anki-Android-Backend repo, related issue is/was:
- ankidroid/Anki-Android-Backend/pull/457

Related work here
- #2551


package.json
```diff
@@ -67,7 +67,6 @@
         "@popperjs/core": "^2.11.8",
         "bootstrap": "^5.3.0",
         "bootstrap-icons": "^1.10.5",
-        "canvas": "npm:empty-npm-package",
```
mikehardy, line 70: this isn't needed if resolutions forces it to resolve some way / any way

package.json
```diff
@@ -80,7 +79,7 @@
         "mathjax": "^3.1.2"
     },
     "resolutions": {
-        "canvas": "npm:empty-npm-package",
+        "canvas": "npm:empty-npm-package@1.0.0",
```
mikehardy, line 82: the clue that this might work was in the yarn berry error message that npmjs.com didn't have the tag "empty-npm-package"

I realized it wasn't resolving the parts as we expected - when I added the version to the specification it correctly resolved "empty-npm-package" at it's current version  ref (https://www.npmjs.com/package/empty-npm-package?activeTab=versions) and canvas no longer tried to build and broke on my system - in both yarn versions

dae:
Thanks for digging into this Mike! If that was the only impediment to a yarn upgrade, I'll look into updating our pinned version in a follow-up commit.

Just a minor conflict to fix - I'd do it for you, but that would break the contributor detection.

dae:
Nevermind, I'll handle this, but I think you may need to update CONTRIBUTORS again on your next contribution.

PR #3754, opened 2025-01-20
Cleanup av_player on profile close

kelciour:
To reproduce the issue, a profile needs to be opened and closed several times.

The list of players will keep growing in size and include the list of players that were previously instantiated but were shut down on profile close a bit later.

```
from aqt.sound import av_player

for i, p in enumerate(av_player.players, 1):
  print(i, p)
```

**Before**

image

**After**

image


YukiNagat0:
`self.players.clear()` will be more "pythonic" way to do it.

kelciour:
Thank you! It's been changed.

abdnh (approved):

Issue #3702, opened 2025-01-07
Clicking ‘Add’ on the main screen with the ‘Add’ window already open should update it to the currently selected deck


dae:
> This is a small QoL suggestion. For those of us using multiple decks and adding to each, I sometimes forget the ‘add’ menu is already open and assume that the add screen has opened to the currently selected deck (as it is when opened for the first time).
> The behavior of setting the target deck to the currently selected one should be consistent regardless of whether the add window is already open. I don’t envisage any drawbacks to this change. Benefits would be not unintentionally losing new cards in different decks, or having to manually move them every time you make this mistake.

Originally reported on https://forums.ankiweb.net/t/clicking-add-on-the-main-screen-with-the-add-window-already-open-should-update-it-to-the-currently-selected-deck/53936

PR #3755, opened 2025-01-21
Change Add button to update Add window to current deck (#3702)

lukesbart:
Added new event handler to check current deck against deck in add window and update if necessary


iamllama:
Implementing [`reopen`](https://github.com/ankitects/anki/blob/64ca90934bc26ddf7125913abc9dd9de8cb30c2b/qt/aqt/__init__.py#L147) for `AddCards` is a cleaner way to go about this imo

PR #3756, opened 2025-01-21
Allow reopening the add cards window

### LINKED ISSUE #3702: Clicking ‘Add’ on the main screen with the ‘Add’ window already open should update it to the currently selected deck

iamllama:
Closes #3702 as an alternative to #3755

With this change, reopening the add cards window behaves the same* as if the window was manually closed and reopened, i.e selecting both the default deck and notetype for adding

*The contents of some filled-in fields can be lost when switching to a notetype with fewer fields

For context on why keeping the same behaviour across opening and reopening is important, please see https://github.com/ankitects/anki/pull/3755#issuecomment-2603598451

Happy to close this pr if it turns out to not be needed


abdnh:
I think it makes sense to respect the preference here 👍

PR #3759, opened 2025-01-22
Remove unfinished polygon when undoing and redoing in IO mode

### LINKED ISSUE #3741: IO: Undo Doesn't Work for Polygon Tool
## Steps

1. Create a rectangular mask.
2. Use the polygon tool to put some points on the image.
3. Use undo from toolbar.

## Behaviour

Using undo removes the rectangular mask. I expected it to simply remove the unfinished polygon from my image.

iamllama:
Fixes #3741. Also fixes the issue where if you try to draw a polygon after the steps in 3741, the points of the previously unfinished polygon are incorrectly carried over

The issue is that the polygon tool maintains its own state (points, lines, active shape) independently of the undo stack, which isn't updated until after the polygon is completed. When undoing, the canvas is reset to an earlier snapshot that doesn't include any of the unfinished polygon's shapes, but the polygon tool's internal state isn't reset

The fix proposed here is to treat removing an unfinished polygon (and resetting the polygon tool's state) an an undo step in its own right, without actually pushing to or popping from the undo stack

To clarify, if the stack is empty when drawing a polygon, it can't be undone and has to be finished first. If the stack isn't empty while drawing a polygon, undoing will simply remove the unfinshed polygon without popping the stack

When redoing, the polygon tool's state has to be cleared as well, because the incoming canvas snapshot won't include the unfinished polygon, but it isn't treated a discrete step so the user can actually see the effects of the redo without having to press it again.

This is admittedly an interim hacky fix in want of a more "complete" solution. Such a solution might allow undoing each point of an unfinished polygon, perhaps by pushing each to the undo stack. But this might require considering how to draw a unfinished polygon from points in the stack, possibly without tracking state in the polygon tool, and how to remove said polygon/points from the canvas and stack in the event of a tool change or pinch zoom. Which is going to require a lot more thinking on my (and/or someone else's) part 😅 And from my testing, the changes proposed in this pr would make the polygon tool play nice with undo/redo in the meantime


abdnh:
A minor issue is that the undo button won't recognize an unfinished polygon if it's the first shape.
image

PR #3760, opened 2025-01-22
Fix: Simulator Ignores New Cards Already Introduced

### LINKED ISSUE #3708: Simulator Ignores New Cards Already Introduced
* In a fresh deck, add 10 new cards.
* Set `new card/day` to 5.
* Finish learning the 5 new cards introduced today.
* Now, run the simulator and check the number of reviews at `day 0` and `day 1`

The 5 `is:new` cards you’ve got left should be simulated at `day 1` because you finished your new cards for today. Here, I instead see them being simulated at `day 0`.

*Originally posted [on the forums](https://forums.ankiweb.net/t/simualtor-doesnt-consider-cards-already-introduced-today/54067)*

Luc-Mcgrady:
fixes #3708

Where you have reviewed 4 cards today:
image

Is there a better way than using `introduced:1` which doesn't include cards which have been forgotten and then re-introduced?

### DIFF
--- a/rslib/src/scheduler/fsrs/simulator.rs
+++ b/rslib/src/scheduler/fsrs/simulator.rs
@@ -26,21 +26,21 @@ impl Collection {
         let days_elapsed = self.timing_today().unwrap().days_elapsed as i32;
         let mut converted_cards = cards
             .into_iter()
-            .filter(|c| c.queue != CardQueue::Suspended && c.queue != CardQueue::PreviewRepeat)
+            .filter(|c| c.queue != CardQueue::Suspended && c.queue != CardQueue::PreviewRepeat && c.queue != CardQueue::New)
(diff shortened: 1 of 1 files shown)


brishtibheja:
> doesn't include cards which have been forgotten and then re-introduced?

can you clarify what you mean?

also, does this handle the case where learn count is lower than `introduced:1`?

Luc-Mcgrady:
If you have a card which has previously been introduced but then you hit "reset" on it, its `introduced` will still be when it was introduced the very first time. (unless its changed since I last checked)

> also, does this handle the case where learn count is lower than introduced:1?

Seems to :+1: 
image

dae:
Any objections to me merging this in @L-M-Sherlock?

L-M-Sherlock (approved):
LGTM

dae:
Thanks Luc!

PR #3762, opened 2025-01-23
add scripting field to card template

jakeprobst:
scripting

### DIFF
--- a/ftl/core/card-templates.ftl
+++ b/ftl/core/card-templates.ftl
@@ -9,6 +9,7 @@ card-templates-card-type = Card Type:
 card-templates-front-template = Front Template
 card-templates-back-template = Back Template
 card-templates-template-styling = Styling
+card-templates-template-scripting = Scripting
 card-templates-front-preview = Front Preview
 card-templates-back-preview = Back Preview
 card-templates-preview-box = Preview
--- a/proto/anki/card_rendering.proto
+++ b/proto/anki/card_rendering.proto
@@ -127,6 +127,7 @@ message RenderCardResponse {
   repeated RenderedTemplateNode answer_nodes = 2;
   string css = 3;
   bool latex_svg = 4;
+  string js = 5;
 }
 
 message RenderedTemplateNode {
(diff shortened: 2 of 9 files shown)


jakeprobst:
In those cases you can just define two functions and call them in the corresponding template.

And to slip a slope just for fun: why not have front and back styling too? :leaves:

jakeprobst:
yep!

dae:
I'm afraid I'm not keen on this change. JS in card templates is difficult to update, and hard to support, and I think we may be better off moving in the opposite direction, moving the majority of styling and JS out of card templates, and into some sort of add-on/component system for templates, so deck authors and users can opt-in to particular styles or functionality, instead of pasting in code that is at risk of going stale.

user1823:
I can't comment on the overall cost vs benefit of this change. 

But, this change makes using JS in an individual notetype easier. Currently, if I want to create or edit a JS function in my card template, I have to do that in multiple places — the Front templates and the Back templates of all the cards generated from that notetype (in my case, Forward and Reverse cards). After this change, I will have to make the change only at a single place.

This PR won't help if the user wants to make the same change in multiple notetypes, though.

> into some sort of add-on/component system for templates

This would be great, but based on the discussions in the Forums, this doesn't seem to be coming anytime soon.

jakeprobst:
My own [use case](https://github.com/ankitects/anki/pull/3188) that led to this PR is similar to what user1823 is saying. I have a note type that has 15+ cards each with mostly-the-same javascript on them, if I want to make a change I need to edit each one individually and it is quite tedious.

It is not like this PR is suddenly making scripting possible. A quick look at the forums and I found a variety of threads with scripting suggestions. Anki's ability to arbitrarily script the cards to add new features is a great feature. It just makes it more convenient to do. You could even argue that this makes updating outdated scripts easier to manage as the scripts are in one place, away from the templating, rather than mixed in and spread across a bunch of cards in the note.

> discussions in the Forums
can you post the link? I couldn't find a topic that seemed relevant to this.

dae:
On the positive side, it saves you having to surround the code in script tags, and is a bit neater. On the negative side, it completely breaks on older Anki clients, and moves us further in the wrong direction. I'm afraid I don't think it's worth introducing breaking backwards-incompatible changes for a mild QoL improvement that will be temporary. For now, I'd recommend you either look into referencing a file in the media folder instead, or use one of the existing add-ons that can automatically update your templates with the contents of external files.

user1823:
> can you post the link? I couldn't find a topic that seemed relevant to this.

https://forums.ankiweb.net/t/cross-platform-js-addons-for-the-reviewer/46082

PR #3771, opened 2025-01-26
Fall back to SM2 relative order when memory state missing

### LINKED ISSUE #3770: When sorting by retrievability, fall back to SM2 values when memory state is missing
In `add_extract_fsrs_relative_retrievability()` instead of returning None when no memory state is found, we may be able to return the [sm2 value](https://github.com/ankitects/anki/pull/3747/files#diff-02fcdde1afadf6a1e79ed8285b7f879d04754f190175f001ee7fad038afca3dbR779) instead.

Related: #3747

dae:
Closes #3770

@user1823 would you mind giving this a test?

### DIFF
--- a/rslib/src/storage/sqlite.rs
+++ b/rslib/src/storage/sqlite.rs
@@ -346,51 +346,53 @@ fn add_extract_fsrs_relative_retrievability(db: &Connection) -> rusqlite::Result
         move |ctx| {
             assert_eq!(ctx.len(), 5, "called with unexpected number of arguments");
 
-            let Ok(card_data) = ctx.get_raw(0).as_str() else {
+            let Ok(due) = ctx.get_raw(1).as_i64() else {
                 return Ok(None);
             };
-            if card_data.is_empty() {
+            let Ok(interval) = ctx.get_raw(3).as_i64() else {
                 return Ok(None);
-            }
-            let card_data = &CardData::from_str(card_data);
-            let Ok(due) = ctx.get_raw(1).as_i64() else {
+            };
+            let Ok(next_day_at) = ctx.get_raw(4).as_i64() else {
                 return Ok(None);
             };
             let days_elapsed = if due > 365_000 {
                 // (re)learning
-                let Ok(next_day_at) = ctx.get_raw(4).as_i64() else {
-                    return Ok(None);
-                };
                 next_day_at.saturating_sub(due) as u32 / 86_400
             } else {
                 let Ok(days_elapsed) = ctx.get_raw(2).as_i64() else {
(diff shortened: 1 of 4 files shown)


rslib/src/storage/sqlite.rs
```diff
@@ -346,51 +346,53 @@ fn add_extract_fsrs_relative_retrievability(db: &Connection) -> rusqlite::Result
...
+                            // power should be the reciprocal of the value of DECAY in FSRS-rs,
+                            // which is currently -0.5
+                            -(current_retrievability.powi(-2) - 1.)
+                                / (desired_retrievability.powi(-2) - 1.),
+                        ));
+                    }
+                }
+            }
 
+            // FSRS data missing; fall back to SM2 ordering
```
user1823, line 393: Sorry, I couldn't build Anki from source but the code looks good to me. However, don't we need an `else` just above this line?

user1823 (commented):
Sorry, I couldn't build Anki from source but the code looks good to me. Left one question.

rslib/src/storage/sqlite.rs
```diff
@@ -613,3 +615,36 @@ impl Display for SqlSortOrder {
...
+
+        // value from fsrs
+        let mut pos: Option<f64>;
+        pos = col.storage.db_scalar(&sql).unwrap();
+        assert_eq!(pos, Some(0.0));
+        // erasing the memory state should not result in None output
+        col.storage.db.execute("update cards set data=''", [])?;
+        pos = col.storage.db_scalar(&sql).unwrap();
+        assert!(pos.is_some());
+        // but it won't match the fsrs value
```
user1823, line 646: The values will match if you disable fuzz and rounding, i.e., when interval is exactly equal to the value calculated from stability and desired retention. Not sure how that affects this test, but I thought that this might be worth mentioning.

rslib/src/storage/sqlite.rs
dae: Nope, if any of the nested conditions fail, we want to fall back on the outer scope's handling.

rslib/src/storage/sqlite.rs
user1823: Well, the divisor doesn't need to be the same for two fractions to be equal. 😅

E.g. 1/2 = 2000/4000

In your example above, the fraction is equal to 0 in both the cases. It's just that the +0.001 part caused the SM2 value to miss becoming 0.

dae: 🤦 that was silly. But as you said, they still won't match due to the 0.001.

PR #3774, opened 2025-01-26
Use the term Retrievability consistently

sommerluk:
#3604 introduced the string “Probability of Recall” as description of the y-axis in the card stats diagram. I suppose it means the same as “Retrievability”, which is used yet in various other contexts in Anki, and also in the very same card statistics dialog, above in the list of data. Therefore, I suggest changing it to “Retrievability” for greater consistency.


dae:
Will defer to @Expertium @user1823 @brishtibheja here

brishtibheja:
If I were to choose, I'll go with "retreivability" here; but I've brought up completely removing the axis titles in the beta thread. Hope that is what we do in the future.

dae:
No objections on my end to removing the label - it would make it consistent with the rest of our graphs.

PR #3776, opened 2025-01-27
Add drag/drop support when adding IO images

iamllama:
Relevant forum topic: https://forums.ankiweb.net/t/image-occlusion-not-working-jan-2025

With this pr, images can now be now dragged and dropped onto the IO picker screen instead of having to be selected or copied and pasted. Tested on ubuntu 24 and win 10

NB: this suffers from the bug brought up in and fixed by:
- #3775

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -1559,7 +1559,7 @@ class EditorWebView(AnkiWebView):
             and (html := self._processUrls(mime, allowed_suffixes=pics))
             and (path := self.editor.extract_img_path_from_html(html))
         ):
-            self.editor.setup_mask_editor_for_new_note(path)
+            self.editor.setup_mask_editor(path)
             return
 
         evt_pos = evt.position()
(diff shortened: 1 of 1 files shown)


qt/aqt/editor.py
```diff
@@ -1553,6 +1553,15 @@ def dropEvent(self, evt: QDropEvent | None) -> None:
         extended = self._wantsExtendedPaste()
         mime = evt.mimeData()
         assert mime is not None
+
+        if (
+            self.editor.state is EditorState.IO_PICKER
+            and (html := self._processUrls(mime, allowed_suffixes=pics))
+            and (path := self.editor.extract_img_path_from_html(html))
+        ):
+            self.editor.setup_mask_editor_for_new_note(path)
```
abdnh, line 1562: To make it work when "Reset Image Occlusion" is used:

```suggestion
            self.editor.setup_mask_editor(path)
```

abdnh (changes requested):

abdnh (approved):

PR #3777, opened 2025-01-27
Added: SpinnerBox percentage symbol.

Luc-Mcgrady:
Adds the trailing % as referenced in #3679

Displays a percentage directly following the percentage values in the deck config.

percentage

Supports i18n:
```deck-config-percent-input = %{ $pct }```
image
```deck-config-percent-input = { $pct } %```
image

Known issue: The percentage sign flies off the right hand side of the box if the user fills the input to the extent where it gets horizontal scroll.

Sorry if doing it using invisible characters is a little hacky, I couldn't think of a better way.

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -449,6 +449,7 @@ deck-config-percent-of-reviews =
         [one] { $pct }% of { $reviews } review
        *[other] { $pct }% of { $reviews } reviews
     }
+deck-config-percent-input = { $pct }%
 deck-config-optimizing-preset = Optimizing preset { $current_count }/{ $total_count }...
 deck-config-fsrs-must-be-enabled = FSRS must be enabled first.
 deck-config-fsrs-params-optimal = The FSRS parameters currently appear to be optimal.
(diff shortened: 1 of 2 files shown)


YukiNagat0:
I will suggest the most obvious thing, but maybe just change the `Desired retention` and `Historical retention` strings to `Desired retention (%)` and `Historical retention (%)`? In this case, it will not be necessary to invent a bicycle for the input field. And overall, it will be more customary/usual UI decision, because it is very rare to see the input field, that have UI elements (percent sign in this case) in it.

dae:
Adjusting the label would have been sufficient, but Luc's done the work at this point, so it's a harder call.

On the one hand, it's the most readable for users, and we do have precedent (e.g. user interface size in preferences), as Qt provides such a widget already. On the other hand, this introduces more things that can go wrong. For example, the layout is different on iOS, where I found I had to change the 1px top to 3.5px instead to get it to align with rest of the text.

I don't have strong feelings. Anyone else want to weigh in?

Luc-Mcgrady:
That should hopefully do it. I don't own an apple device so can't test it. I am hopeful as my solution is also used here:
https://github.com/ankitects/anki/blob/de7a693465ca302e457a4767c7f213c76478f0ee/ts/routes/deck-options/CardStateCustomizer.svelte#L46

Arthur-Milchior:
Thank you very much for completing the work I tried to do. I can't imagine it takes so many line of codes for such a simple symbol!

PR #3778, opened 2025-01-27
Added "preset" to "Optimize all" and "Optimize current"

Expertium:
I think this one really calls for a poll.
Dae, I highly suggest you to ask multiple users whether "Optimize current" and "Optimize all" are clear enough or need "preset". I'm pretty sure most people will find it confusing without "preset".


sunrucong:
I say include "preset"!

dae:
Fair enough. My thinking at the time was that if the user is in doubt, it should be obvious what 'all' means when they see the progress dialog. But that assumes the user will experiment with the UI to learn more, which is not a safe assumption.

PR #3779, opened 2025-01-28
Add "Show image in folder" and "Open image" context menu actions for IO

iamllama:
Relevant forum post: https://forums.ankiweb.net/t/unable-to-find-image-source-in-note-type-image-occlusion/54940

There's currently no easy way to get at an IO note's image on disk, since the field is hidden the only practical means are through code or by squinting at the `Question` or `Answer` columns in the browser

The "show image in folder" (win/mac) and "open image" context menu actions for images were added in
- https://github.com/ankitects/anki/pull/3412
- https://github.com/ankitects/anki/pull/3431

This pr proposes to extend them to IO, by showing those actions when right clicking on the mask editor, allowing users to open the image file that's used in the current IO note, in their media folder

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -144,6 +144,8 @@ class Editor:
         # current card, for card layout
         self.card: Card | None = None
         self.state: EditorState = EditorState.INITIAL
+        # used for the io mask editor's context menu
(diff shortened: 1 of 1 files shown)


dae:
Thank you! ❤️

PR #3781, opened 2025-01-28
Fix discard changes randomly being a noop on certain linux systems

iamllama:
> Does anyone have problem with discarding changes of presets?
Linux - Using linux archive in github release
1- Open preset
2- Modify the settings
3- Try to close the preset settings page
4- Ask for discard and click yes
5- It doesn’t close the preset window
https://forums.ankiweb.net/t/anki-25-01-beta/54490/57
https://github.com/ankitects/anki/pull/3571#issuecomment-2615433625

Was able to reproduce this on an ubuntu 24 vm and wsl (ubuntu 24), but not on win 10

According to the docs for `QApplication::activeWindow`
> Returns the application top-level window that has the keyboard input focus, or nullptr if no application window has the focus

The latter seems to be happening on my 2 linux systems, and on least one other person's. Not sure what the common thread is, whether it's a ubuntu issue or possibly a wayland one

The fix proposed here is to explicitly restore the active window at the end of `askUser`, if it was unset

Can confirm that this makes "discard changes" work reliably, and on win 10 the branch isn't taken

EDIT: this issue becomes much rarer when switching to the software video driver

~~Setting as draft because the issue just reappeared out of nowhere again~~ Turns out there was a difference in behaviour between the deprecated [QApplication.setActiveWindow](https://doc.qt.io/qt-6/qapplication-obsolete.html#setActiveWindow) and the recommended alternative [QWidget.activateWindow](https://doc.qt.io/qt-6/qwidget.html#activateWindow). Using the former seems to have fixed the issue for good

### DIFF
--- a/qt/aqt/mediasrv.py
+++ b/qt/aqt/mediasrv.py
@@ -581,7 +581,9 @@ def deck_options_require_close() -> bytes:
         if isinstance(window, DeckOptionsDialog):
             window.require_close()
 
-    aqt.mw.taskman.run_on_main(handle_on_main)
+    # on certain linux systems, askUser's QMessageBox.question unsets the active window
+    # so we wait for the next event loop before querying the next current active window
+    aqt.mw.taskman.run_on_main(lambda: QTimer.singleShot(0, handle_on_main))
     return b""
 
 
--- a/qt/aqt/utils.py
+++ b/qt/aqt/utils.py
@@ -429,6 +429,11 @@ def askUser(
             openHelp(help)
         else:
             break
+    # https://doc.qt.io/qt-6/qapplication.html#activeWindow
+    # on some systems, qt unsets the active window when this messagebox causes the previous window to lose focus
+    # to handle that case, this hack sets the previously active window (parent) as the current active window anew
+    if parent and not parent.isActiveWindow():
+        parent.activateWindow()
     return r == QMessageBox.StandardButton.Yes
 
 
--- a/qt/aqt/utils.py
+++ b/qt/aqt/utils.py
@@ -432,8 +432,10 @@ def askUser(
     # https://doc.qt.io/qt-6/qapplication.html#activeWindow
     # on some systems, qt unsets the active window when this messagebox causes the previous window to lose focus
     # to handle that case, this hack sets the previously active window (parent) as the current active window anew
+    # QApplication.setActiveWindow was deprecated in 6.5, but its recommended alternative (QWidget.activateWindow)
+    # doesn't work, for some reason
     if parent and not parent.isActiveWindow():
-        parent.activateWindow()
+        aqt.mw.app.setActiveWindow(parent)
     return r == QMessageBox.StandardButton.Yes
 
 
--- a/qt/aqt/utils.py
+++ b/qt/aqt/utils.py
@@ -429,13 +429,6 @@ def askUser(
             openHelp(help)
         else:
             break
-    # https://doc.qt.io/qt-6/qapplication.html#activeWindow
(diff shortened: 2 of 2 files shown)


dae:
Thanks for digging into this! I'm ok with a hack if there's no alternative, but this will likely bite us at some point in the future when the old routine is dropped, so it would be worth making sure we've exhausted alternatives first.

I noticed there's a .requestActivate() - presumably that doesn't help at all?

javaScriptConfirm() invokes askUser() without a parent, so presumably on your system, activeWindow() at the top of askUser returns the window, but returns None after the modal has closed? Does this happen even if you add, say, a 1 second delay between the closing of the dialog and then check for the active window?

iamllama:
> I noticed there's a .requestActivate() - presumably that doesn't help at all?

wasn't able to try this unfortunately, it's only on defined on QWindow, whereas the deck options dialog is a QDialog

> Does this happen even if you add, say, a 1 second delay between the closing of the dialog and then check for the active window?

this did the trick! `QMessageBox.question` immediately unset the active window on my linux systems (not win), but on the next event loop iteration the deck options dialog was made active again. no idea why this is the case, qt unhelpfully disclaims "platform-specific behaviours", but am glad to be rid of the deprecated function 😅

there other other places in mediasrv where this pattern is used, but they don't seem to use messageboxes beforehand, so i doubt fixing them is currently necessary 🤔

dae:
I presume what's going on here is that Qt receives a message from the window manager about the change in window focus, and that message may only arrive after our code runs normally. A fixed delay may still fail if the WM is slow enough, but it's a  simple fix, so let's start from there and see how things go. Thank you!

BEST8OY:
@dae @iamllama 
It's not been fixed for me yet.

iamllama:
Could be that waiting till the next event loop really was too optimistic then

@BEST8OY if you're able to build and run anki, does using a longer delay fix it for you?

If not, could you make a test addon (https://addon-docs.ankiweb.net/a-basic-addon.html) with the following snippet and check if the deck options window now closes 1 second after discarding changes?
```py
from aqt import mw, mediasrv
from aqt.deckoptions import DeckOptionsDialog
from aqt.qt import QTimer

def patch() -> bytes:
    def handle_on_main() -> None:
        window = mw.app.activeWindow()
        if isinstance(window, DeckOptionsDialog):
            window.require_close()

    mw.taskman.run_on_main(lambda: QTimer.singleShot(1000, handle_on_main))
    return b""

mediasrv.post_handlers["deckOptionsRequireClose"] = patch

```

BEST8OY:
> If not, could you make a test addon (https://addon-docs.ankiweb.net/a-basic-addon.html) with the following snippet and check if the deck options window now closes 1 second after discarding changes?

Did it the addon way and it got fixed.
But I could still feel that 1 second delay so I tried different numbers, 0.75s doesn't work but 0.8s or 800 works just fine.

BEST8OY:
@iamllama Should I keep using the fix as addon until it's properly fixed?

iamllama:
> @iamllama Should I keep using the fix as addon until it's properly fixed?

Without further reports on this from others and without more info on your system, it's hard to identify a more basal fix

I doubt delaying closing by >0.8s will go over well with the majority who don't seem to have the issue or are fine with the current fix, but will defer to @dae on that (sorry for the ping!)

BEST8OY:
just to add some information about this
I use hyprland and this Anki tar.zst package built by this repo
When I open the option window of a deck, it's a floating window by default and the discard option doesn't work

But when I turn the option window into a non-floating window (hyprland) the discard option works just fine.
The discard option also works if I turn that non-floating window into a floating one again.

I'm not sure to say if it's a hyprland problem or not!

PR #3786, opened 2025-02-02
Cap preview cols to prevent stall when csv is parsed with the wrong delimiter

### LINKED ISSUE #3588: Bug with importer for files containing quotes
> If a wrong delimiter is guessed by the importer and the file contains quotes like this .txt file with commas as delimiter:
> Row1,
> Row2,"|""|"
> Row3,"|""|"
> ...
> 
> The “new” importer will create a single row for Row2 and Row3 including extra columns.
> This means the entire file will be displayed in the sample:
> image1056×314 13.2 KB
> For a large enough file this results in the importer loading forever (the loading time grows exponentially with the file size).

Originally reported on https://forums.ankiweb.net/t/bug-with-importer-for-files-containing-quotes/51732/2

iamllama:
Resolves #3588

> Sure, [here is an example](https://www.dropbox.com/scl/fi/yorxjlwzszwd6j7a0e9u2/Cities-of-Your-Country-First-4000.csv?rlkey=yvtuelr9xl9a0raf3dx0p93rz&st=xf6wywo7&dl=0) which takes 25 seconds to load for my pc.

Looking at what passes for the [csv spec](https://datatracker.ietf.org/doc/html/rfc4180#page-2), the example is considered **invalid** csv when pipe-delimited. But python's csv.reader parses it in the same way as the crate. I can't speak to how they handle quirks, but my guess is that the last pipe in each row having one unclosed quote after it causes the record terminator to be quoted and ignored, and the file ends up as being 1 header and 1 record long

The stall doesn't seem to be happening on the backend (which parses and responds in ~100ms), but on the webpage itself. Profiling on firefox confirms that it's choking on rendering ~40000 columns in the preview table

This pr proposes to cap the preview table's columns at 1000, a number small enough that it should render reasonably quickly (~1s, prev ~20s), and ludicrously large enough that exceeding it almost certainly means wrong delimiter and/or invalid csv (who has notetypes with >1000 fields?)

This'll allow the user to quickly recover from a wrong guess by the delimiter heuristic, rather than have to wait ~20 seconds for the page to load

A warning is shown when columns have been clipped:
image

EDIT: failing on an openssl advisory 👀

### DIFF
--- a/ftl/core/importing.ftl
+++ b/ftl/core/importing.ftl
@@ -67,6 +67,12 @@ importing-packaged-anki-deckcollection-apkg-colpkg-zip = Packaged Anki Deck/Coll
 importing-pauker-18-lesson-paugz = Pauker 1.8 Lesson (*.pau.gz)
 # the '|' character
 importing-pipe = Pipe
+# Warning displayed when the csv import preview table is clipped (some columns were hidden)
+importing-preview-truncated =
+    { $count ->
+        [one] Only the first column is shown. If this doesn't seem right, try changing the field separator.
+        *[other] Only the first { $count } columns are shown. If this doesn't seem right, try changing the field separator.
+    }
 importing-rows-had-num1d-fields-expected-num2d = '{ $row }' had { $found } fields, expected { $expected }
 importing-selected-file-was-not-in-utf8 = Selected file was not in UTF-8 format. Please see the importing section of the manual.
 importing-semicolon = Semicolon
--- a/ftl/core/importing.ftl
+++ b/ftl/core/importing.ftl
@@ -68,9 +68,9 @@ importing-pauker-18-lesson-paugz = Pauker 1.8 Lesson (*.pau.gz)
 # the '|' character
 importing-pipe = Pipe
 # Warning displayed when the csv import preview table is clipped (some columns were hidden)
+# $count is intended to be a large number (1000 and above)
 importing-preview-truncated =
     { $count ->
-        [one] Only the first column is shown. If this doesn't seem right, try changing the field separator.
         *[other] Only the first { $count } columns are shown. If this doesn't seem right, try changing the field separator.
     }
 importing-rows-had-num1d-fields-expected-num2d = '{ $row }' had { $found } fields, expected { $expected }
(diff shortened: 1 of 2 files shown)


ftl/core/importing.ftl
```diff
@@ -67,6 +67,12 @@ importing-packaged-anki-deckcollection-apkg-colpkg-zip = Packaged Anki Deck/Coll
 importing-pauker-18-lesson-paugz = Pauker 1.8 Lesson (*.pau.gz)
 # the '|' character
 importing-pipe = Pipe
+# Warning displayed when the csv import preview table is clipped (some columns were hidden)
+importing-preview-truncated =
+    { $count ->
+        [one] Only the first column is shown. If this doesn't seem right, try changing the field separator.
```
dae, line 73: When first reading this, I assumed you were repurposing this message to also warn the user when only a single column in shown, as it's not bad advice then either.

If this was just for the translation system, you could probably even drop the [one] case, and mention in the translation comment that $count will only ever be 1000 or a similarly-large number, which would allow translators to skip zero/one/few cases.

On the other hand, if you were thinking of using this in the 1 case as well in the future, I'd recommend rewording it so it can make use of $count, as hard-coding a number instead of using a variable in the one case has tripped up translators in the past.

iamllama:
> Is your thinking with the underscored _tr similar to this?

Yep, it'd make svelte treat `<tr>` as a component otherwise. Might be cleaner to just import the one translation, but i figured having `tr.` in there would be more greppable 🤔

> If this was just for the translation system, you could probably even drop the [one] case, and mention in the translation comment that $count will only ever be 1000 or a similarly-large number, which would allow translators to skip zero/one/few cases.

Included the `[one]` case for completeness' sake, dropping it seems better than changing it to something like "Only the first 1 column is shown", which sounds off 😅. Added a comment to clarify that as suggested

dae:
> Yep, it'd make svelte treat <tr> as a component otherwise. Might be cleaner to just import the one translation, but i figured having tr. in there would be more greppable 🤔

Would you mind elaborating on what you mean by that? I changed the _tr to tr in my local checkout, and it still passes tests/works correctly.

iamllama:
Well that's embarrassing. I tried it earlier in the morning and got a `component_name_lowercase` svelte warning, along with the <tr> tags silently being removed from the rendered table, but now it seems fine? I must've done something wrong back then, sorry about that!

dae:
I've experienced that many a time too - don't sweat it :-)

Thank you for this fix!

PR #3788, opened 2025-02-04
Fix/disable FSRS short-term scheduler if w[17] or w[18] is zero

### LINKED ISSUE #3782: Disable the FSRS short-term scheduler if w[17] or w[18] is zero
If w[17] OR w[18] is zero, disable the FSRS short-term scheduler because FSRS won't be able to increase the stability with short-term reviews.

Inspired by https://forums.ankiweb.net/t/fsrs-5-1d-scheduling-and-learning-steps/50242/73

_Originally posted by @user1823 in https://github.com/ankitects/anki/issues/3496#issuecomment-2410867005_

This is even more important after https://github.com/ankitects/anki/pull/3742 because the optimizer will make w[17] and w[18] equal to zero for more users now.

L-M-Sherlock:
close #3782

### DIFF
--- a/rslib/src/scheduler/states/review.rs
+++ b/rslib/src/scheduler/states/review.rs
@@ -124,7 +124,7 @@ impl ReviewState {
                 review: again_review,
             }
             .into()
-        } else if ctx.fsrs_enable_short_term
+        } else if ctx.fsrs_allow_short_term
             && (ctx.fsrs_short_term_with_steps_enabled || ctx.relearn_steps.is_empty())
             && scheduled_days < 0.5
         {
--- a/rslib/src/scheduler/states/review.rs
+++ b/rslib/src/scheduler/states/review.rs
@@ -124,8 +124,8 @@ impl ReviewState {
                 review: again_review,
             }
             .into()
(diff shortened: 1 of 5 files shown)


rslib/src/scheduler/states/review.rs
```diff
@@ -124,7 +124,8 @@ impl ReviewState {
                 review: again_review,
             }
             .into()
-        } else if (ctx.fsrs_short_term_with_steps_enabled || ctx.relearn_steps.is_empty())
+        } else if (ctx.fsrs_enable_short_term && ctx.fsrs_short_term_with_steps_enabled
```
user1823, line 127: Parenthesis seem to be missing here. It should be 

```rs
        } else if (ctx.fsrs_enable_short_term && (ctx.fsrs_short_term_with_steps_enabled
            || ctx.relearn_steps.is_empty()))
```

user1823:
Using the name `fsrs_allow_short_term` instead of `fsrs_enable_short_term` might be less confusing.

rslib/src/scheduler/states/review.rs
L-M-Sherlock: fixed

dae:
Thank you both!

Issue #3062, opened 2024-03-09
Using 'cut' on a hierarchical tag replaces double colons (::)


jcznk:
Tested on Anki 24.01 beta1, no add-ons, Windows 10 qt6. IIRC, it also used to happen on previous versions.

Steps: 

1. go to the Editor/Browser and select a note that has a hierarchical tag (or add one).
<img src="https://github.com/ankitects/anki/assets/60730312/93a924a8-ec4f-4694-963b-402bacd7feba" width="400">

2. Click on the tag bar next to the hierarchical tag you want to cut, and move the cursor so that the tag is displayed as plain text.
<img src="https://github.com/ankitects/anki/assets/60730312/751fcca0-489e-423c-8305-6082514d3c6c" width="400">

3. Select the tag and then 'cut' it (e.g., using Ctrl+X), and then paste it. The double colon (::) is replaced by a ratio symbol (∷).
E.g., #::ExampleParentTag::ExampleChildTag becomes #∷ExampleParentTag∷ExampleChildTag. This can be inconvenient if you want to use the tag for Browser searches or Find & Replace, as you have to fix it manually.

The problem does not seem to occur if you copy the tag instead of cutting it.
(But, if I copy the text of the tag when the tag is *not* rendered as plain text, the double column does get replaced with a ratio symbol, and a few divs and brs are also added: `<div><br></div><div><div>#∷ExampleParentTag∷ExampleChildTag</div></div><div><br></div>`)

jcznk:
Also, the "tag input" bar does not seem fully distinguish between double colons (::) and the ratio symbol (∷). When I paste `#∷ExampleParentTag∷ExampleChildTag` in the tag bar, the ratio symbols are replaced with double colons. However, attempting to paste the ratio symbol `∷` alone, outside of hierarchical tags, into the tag bar does not result in any input being accepted.

PR #3791, opened 2025-02-04
Fix ∷ in tags not being converted to :: when cut

iamllama:
Resolves part of #3062

The tag editor displays `::` as `∷` to simplify(?) hierarchical tag editing (https://github.com/ankitects/anki/pull/1264#issuecomment-915254400)

The conversion back to `::` is handled when copying, but not when cutting, which this pr aims to rectify

Omitting `setPosition` moves the caret to the end
Omitting `dispatch` means not opening autocomplete or updating it if already open (existing behaviour)

Didn't notice any regressions with this, but the editor code in general is very confusing and i might've missed something 👀

### DIFF
--- a/ts/lib/tag-editor/TagInput.svelte
+++ b/ts/lib/tag-editor/TagInput.svelte
@@ -213,6 +213,18 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         );
     }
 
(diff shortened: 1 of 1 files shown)


dae:
>  but the editor code in general is very confusing 

Tell me about it!

Really appreciate the fix, and your great PR summaries. ❤️

PR #3792, opened 2025-02-05
Update Deck Options strings to clarify Timers

Danika-Dakika:
This is a companion to ankitects/anki-manual#248 . 

Clarifying the naming and descriptions of the Deck Options/Timer features to match the updates in the manual. Per the discussion there, I've made changes to the existing strings directly, rather than create new strings that require translation.

CONTRIBUTORS/about.py -- 
I know that they are not meant to match necessarily, but I added myself to both. I can strike one or the other if that's overdoing it.


Danika-Dakika:
Checks failed -- https://buildkite.com/ankitects/anki-ci/builds/9665#0194d868-19f0-46b0-8230-c2a8ca8f4bf2/55-87
I see how this happened. The jumplink in the manual hasn't been updated yet. This can wait until that happens, or I can revert that change and rely on the legacy jumplink-anchor.

dae:
You definitely belong in both files! Thanks Danika :-)

PR #3794, opened 2025-02-06
Fix "discard changes" appearing even after resetting IO note

iamllama:
In the add window, when resetting image occlusion, the occlusions field is cleared* (by way of `canvas.clear`), but the image field isn't. This causes "discard changes" to appear when trying to close the window, even though there's seemingly nothing to discard

The fix proposed is to clear it when resetting in add mode. Edit mode is left alone since the note is still editable afterwards and new masks can be added over the same image

*which is why `saveFieldNow` is needed again to avoid a race


dae:
Just in time! Comment explaining why was a nice touch. Thanks as always. :-)

PR #3795, opened 2025-02-06
Update reviewer.py to prevent custom scheduler js from commenting out py code

GithubAnon0000:
Fixes: https://forums.ankiweb.net/t/custom-scheduler-comment-bug/55483?u=anon_0000

This commit adds a new line to the js code, if the last line of the code is  a comment. Tested with:
```js
console.log(JSON.stringify(states, null, 4));
console.log(JSON.stringify(states, null, 2));
//console.log(JSON.stringify(states, null, 2));
```

### DIFF
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -289,13 +289,6 @@ class Reviewer:
 
         if js := self._state_mutation_js:
             self._states_mutated = False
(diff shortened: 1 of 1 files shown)


YukiNagat0:
The conditional can be simplified to:
```python
if js.rsplit("\n", 1)[-1].startswith("//"):
    js += "\n"
```

PR #3811, opened 2025-02-12
Fix: Cloze cards missing "copy card info" button

Luc-Mcgrady:
https://github.com/ankitects/anki/pull/3719#issuecomment-2650875582


brishtibheja:
(gotta name this correctly: cloze, not close)

Issue #3787, opened 2025-02-03
Consider using uv in our build system


dae:
It promises considerable performance improvements over things like pip and pip-sync, and seems to have nicer ergonomics.

dae:
Some reasons why we might not want to:
- No dependabot yet: https://github.com/dependabot/dependabot-core/issues/10478
- uv's creator is apparently relatively new and venture funded, and there's a risk they'll close source or abandon the project in the future

JeppeKlitgaard:
I have been using `uv` for all recent projects and it is developing very rapidly and seems to have a lot of momentum. Perhaps my favourite feature is that it will, like rustup, install the correct tooling when running (based on a `.python-version` file along with dependencies listed in `pyproject.toml`). I am getting set up for building anki locally and `uv` would have made this a bit more ergonomic.

andrewsanchez:
The dependabot issue has been resolved and `uv` has seen rapid and widespread adoption. This blogpost which I also shared in #3081  is very informative: https://gregoryszorc.com/blog/2024/12/03/transferring-python-build-standalone-stewardship-to-astral/

I've recently migrated large projects to `uv`. I saw huge performance boots and DX improvements and would happily lend a hand here!

Issue #3809, opened 2025-02-12
Create symlink to enable bash scripts on Windows MSYS2


JeppeKlitgaard:
I am building and running Anki on Windows using MSYS2, which generally works really well.

One significant papercut that I have run into, however, is that the `pyenv` directory with binaries is different on Windows (`out/pyenv/scripts`) as opposed to `out/pyenv/bin`. The bash scripts generally assume the Linux path and fail in this manner:

![Image](https://github.com/user-attachments/assets/00b06ab1-9ade-4655-a01c-45c2c6afe8a7)

Since MSYS2 bash automatically appends the .exe if it is missing and an executable exists, this can be solved simply by creating a symlink from `out/pyenv/bin` to `out/pyenv/scripts` on Windows. I wanted to make a PR for this myself, but I am finding the build tooling to be a bit difficult to get a grasp on.

JeppeKlitgaard:
I have gotten sufficiently frustrated with trying to build it directly on Windows and am now developing in WSL instead. 

This issue can be closed, though it might be worthwhile discussing whether to explicitly recommend developing on Ubuntu in WSL for Windows users now that GUIs are supported in WSL.

dae:
Aside from some fairly minor corner cases like some of the scripts in tools\ needing to be invoked with bash, building should work fine on Windows without WSL. It's how the official builds are built, and the core developers @abdnh and @RumovZ are both building on Windows without WSL AFAIAA. What problems did you hit?

No objections to the symlink if it doesn't introduce regressions. It can likely be done in setup_pyenv(), using junction::create, which already is used in another file.

dae:
(please note that you're expected to invoke things like run(.bat) via cmd.exe or a PS session - if you're using bash, that may cause it to try run the UNIX ./run instead)

PR #3812, opened 2025-02-12
build: script to run qt-designer

JeppeKlitgaard:
This adds a shell script to run the Qt-Designer tool, which is conveniently shipped with the `pyqt6-tools` package.

Since there currently is no Qt 6.6 version available on PyPi (see https://github.com/altendky/pyqt-tools/issues/132), it is best to keep the venv of this tool separate from the rest of the development environment. Currently `pyqt6-tools` seems to have intermittent maintenance, meaning it will also be likely that it will require a specific Python version to be runnable.

With this in mind, the `pyqt6-tools` tool is managed separately with `uv`. It was mentioned in #3787 that `uv` could possible be added to the build toolchain, so I figured this tiny corner could perhaps be the first spot for it.

I completely understand if addition of `uv` at this stage is not desirable, in which case I will close this PR.


dae:
This is sufficiently isolated and unessential that I'm not worried about the uvx usage even if the answer to #3787 ends up being "no". Unfortunately though, it's of limited usefulness at the moment:

- It's broken on Windows until #3809 
- It won't build on modern Macs, as there are no ARM64 wheels, and uvx fails to install when running under Rosetta

I'd be happy to merge if the former can be addressed (I'll follow up on that ticket later), and the script can print a message and exit when run on a Mac.

dae:
Closing for now

PR #3814, opened 2025-02-13
Fix card layout editor crash when editing while reviewing 

### LINKED ISSUE #3808: Wrong card number when opening template editor
I am testing on Anki 25.02 for Windows. Also experienced the behavior on previous versions.

Steps to reproduce:

1. Load a card in the Reviewer.
2. Open the Editor (press `E` or click `Edit`).
3. Without closing the Editor, return to the Reviewer.
4. Continue reviewing cards.
5. Bring up the Editor again → it correctly updates to show the new card.
5. Open the Card Layout Editor (`Ctrl+L` or `Cards...`).

→ Expected behavior: the Template Editor should load the card type corresponding to the card you "re-opened" the Editor for (e.g., card type 2).
→ Actual Behavior: the Template Editor instead loads the card type corresponding to the first card the Editor was opened for (e.g., card type 4). If the current note type doesn't have that card type number, the Template Editor throws an error and refuses to open.

```
  File "aqt.webview", line 667, in handler
  File "aqt.editor", line 647, in <lambda>
  File "aqt.editor", line 381, in _onCardLayout
  File "aqt.clayout", line 91, in __init__
  File "aqt.clayout", line 104, in redraw_everything
  File "aqt.clayout", line 111, in update_current_ordinal_and_redraw
  File "aqt.clayout", line 486, in fill_fields_from_template
  File "aqt.clayout", line 483, in current_template
IndexError: list index out of range

```

iamllama:
Closes #3808

When reviewing, the editor's card and note are set on initial-open, but on reopen, only the note is set. This causes the old (initial) card's ord to be used when opening the card layout editor, which can at best cause the wrong card to be selected and shown, and at worse, throw if the current note doesn't have that many cards/templates

The fix proposed is to set the editor's card when reopening

### DIFF
--- a/qt/aqt/editcurrent.py
+++ b/qt/aqt/editcurrent.py
@@ -67,6 +67,7 @@ class EditCurrent(QMainWindow):
 
     def reopen(self, mw: aqt.AnkiQt) -> None:
         if card := self.mw.reviewer.card:
+            self.editor.card = card
             self.editor.set_note(card.note())
 
     def closeEvent(self, evt: QCloseEvent | None) -> None:


dae:
Thanks for digging into this!

PR #3815, opened 2025-02-13
Feat/support load balance and easy days in rescheduling

L-M-Sherlock:
It supports load balance and easy days natively in rescheduling. For the sake of performance, I didn't reuse the `LoadBalancer` and `LoadBalancerContext`.

Preview:

25.02:

image

The PR:

image

The PR with easy days:

imageimage


jakeprobst:
> For the sake of performance, I didn't reuse the `LoadBalancer` and `LoadBalancerContext`.
this feels like a setup for future sadness, how bad was the perf hit that made this the preferable choice?

L-M-Sherlock:
> this feels like a setup for future sadness, how bad was the perf hit that made this the preferable choice?

This function is I/O bound:

https://github.com/ankitects/anki/blob/038d85b1d9e1896e93a3e4a26f600c79ddc33611/rslib/src/scheduler/states/load_balancer.rs#L120-L128

For rescheduling, the `today` is the last review date, so I have to re-create load balancer for each card. It's very slow. It would take 1 minute to reschedule thousands of cards.

dae:
For future maintenance, we definitely want as much code shared between the paths as possible - whether that's by reusing old code and doing things like mutating an existing context for each card instead of creating a new one, or by introducing this new code and having the old code use it.

YukiNagat0:
Does this change respect the `mw.col.load_balancer_enabled` option?

L-M-Sherlock:
> Does this change respect the `mw.col.load_balancer_enabled` option?

It does now.

> For future maintenance, we definitely want as much code shared between the paths as possible - whether that's by reusing old code and doing things like mutating an existing context for each card instead of creating a new one (which may not be practical here), or by introducing this new code and having the old code use it.

I'm reusing some code snippet now.

L-M-Sherlock:
In the latest commit, if the user changes the easy days setting and enable "reschedule cards on change", the rescheduling will be executed without modifying the desired retention or FSRS parameters.

dae:
Thanks Jarrett, this should make a lot of people happy. As this is largely based on @jakeprobst's code, I'll defer to him on the final approval, but no issues from my end.

One thought: it won't be obvious to users that reschedule on change affects easy days as it's in a different section. One optional improvement would be to display a yellow warning when easy days are changed and reschedule is off - something like:

"Existing reviews will not be rescheduled unless '{ deck-config-reschedule-cards-on-change }' is enabled in the FSRS options."

jakeprobst:
looks good to me!

L-M-Sherlock:
In the 51f43f5, I added this warning:

image

It works well with previous warning, too.

image

user1823:
> Existing reviews will not be rescheduled unless 'Reschedule cards on change' is enabled in the FSRS options.

What if the user doesn't have FSRS enabled? IMO, this warning should be hidden in that case.

user1823:
Also somewhat related:
Currently, the FSRS helper add-on has its own implementation of load balancer and easy days. Does it make sense to expose the Rust code to add-ons so that the Helper add-on could use it directly instead of re-implementing it in Python?

jakeprobst:
>  Currently, the FSRS helper add-on has its own implementation of load balancer and easy days. Does it make sense to expose the Rust code to add-ons so that the Helper add-on could use it directly instead of re-implementing it in Python?

My initial thought is this is probably really difficult to do, unfortunately

dae:
> What if the user doesn't have FSRS enabled? IMO, this warning should be hidden in that case.

I thought about it, but hiding it gives the user no clue that it's not retroactive in the SM2 case.

dae:
Thanks everyone!

dae:
Sorry, I try to remove my attribution when making such trivial changes, but I forgot again. :-(

PR #3816, opened 2025-02-13
x86_64-unknown-linux-gnu has issues with 1.84.0 toolchain for some reason

jakeprobst:
There might be some issues with 1.84.0 on linux systems? Attempting to do anything with cargo leads to `error: Missing manifest in toolchain '1.84.0-x86_64-unknown-linux-gnu'`

bumping to 1.84.1 appears to have fixed the issue. I believe that since it is a minor version bump that this should not cause any sort of issues?

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
 # older versions may fail to compile; newer versions may fail the clippy tests
-channel = "1.84.0"
+channel = "1.84.1"


dae:
Builds fine on Linux CI and my local Linux machines - seems like some sort of caching issue on your local machine? While 
I appreciate you taking the time to contribute this, without any other reports of people hitting it, I'm not sure it's fair to require others tracking the main branch to download a Rust update to address what is probably a local issue. 1.85 is due out in about 5 days in any case, and I do plan to bump the version then, as it's the first release which includes the new 2024 edition.

PR #3817, opened 2025-02-13
add hook for day rollover

jakeprobst:
I was looking in anki to see how day rollover is managed so I can run code around when it triggers for an addon I am working on and saw that it was trivial to add a hook for it! I get that I can hook the `main_window_did_init` and then duplicate this logic in my addon, but this seems like a nice thing to have a hook for.

The use case for me is for it to be used in conjunction with cases I'd hook `profile_did_open` for, but to also handle cases where I left anki on overnight and would like the logic to still run without having to close and reopen anki.

My feelings on this aren't particuarly strong so feel free to close this PR if this is not desired.


dae:
The hook seems useful, my only request is that you follow the naming convention (e.g day_did_rollover)

dae:
Thanks!

Issue #3807, opened 2025-02-11
No Gap Between Simulator Buttons


brishtibheja:
The "simulate" and "clear last simulation" button have no gap between them when vertically spaced under width constraints.

Forums: https://forums.ankiweb.net/t/anki-25-01-02-beta-rc/54490/83?u=sorata

PR #3822, opened 2025-02-15
Update FsrsOptions.svelte to add margin / gap between simulator buttons

GithubAnon0000:
Fixes: #3807 

After the change it looks like this:
anki


dae:
Thanks!

PR #3825, opened 2025-02-16
Use tilted (filled and unfilled) sticky icons in the cards editor

GithubAnon0000:
This is a follow-up on #3821.

Fixes: https://forums.ankiweb.net/t/if-the-user-completely-erases-a-sticky-field-that-should-automatically-unstickify-it/38177/16?u=anon_0000.

# Changes made
I took the icons from @sprvlcn here: https://forums.ankiweb.net/t/if-the-user-completely-erases-a-sticky-field-that-should-automatically-unstickify-it/38177/14?u=anon_0000.

ts/lib/icons/contain-plus.svg is on one single line, which is why I put the new icon code in one line too.

The hollow icon was hardly readable though by default, which is why I changed the fill to currentColor.

# Pictues
The first field is checked / active / sticky, the second one is not checked / is inactive / is not sticky.

### Light
anki

### Dark
anki

### Light
anki

### Dark
anki

### Light
anki

### Dark
anki

# Others
This needs approval from @dae who wanted to see the change in situ.


dae:
LGTM - thank you!

PR #3826, opened 2025-02-16
Use default flag name when flag is renamed to empty string

### LINKED ISSUE #3818: Browser: can't restore default flag names
Testing on Windows 11, version 25.02. Anki has always behaved this way, as far as I remember.

In the Card Browser, you can rename each flag using the right-click context menu. However, once you rename a flag, there seems to be no way to restore the default name. You can manually enter a name identical to the original, but this won’t use Anki’s default string, meaning it won’t update if there are changes or if you switch the interface language.

A couple of ideas:

1. If a custom name is being used, add a "Restore Original Name" option to the right-click context menu.
2. Revert to the original string if the user clicks "Rename", deletes the custom name, and then confirms the change.

Image

mumtazrifai:
Fixes #3818

Right click on a flag in the browser sidebar > Rename > Delete all of the text > Hit Enter to confirm.
After doing this, Anki will go back to using the default name of the flag, including using the correct translation if the user switches languages.

anki-flags


YukiNagat0:
Python Style Guide suggestion:
https://github.com/ankitects/anki/blob/89b0748be1d981b38e588bad910376bde7aaffd4/qt/aqt/browser/sidebar/tree.py#L456
Change the `if new_name == ""` to `if not new_name `

dae:
Thank you for contributing!

This could be done in a simpler way. Instead of storing a translation, simply delete the entry for that key from the dictionary, so that the default translation is used instead. This would change less code, and ensure the text remains correct when the user switches interface language.

mumtazrifai:
I already do delete the dictionary entry from the `"flagLabels"` config entry here:

https://github.com/ankitects/anki/blob/8380131b4434810030869cbd0146da592f48e739/qt/aqt/flags.py#L67-L74

When I tested it, the flag label did switch to the correct translation after I changed the language, which indicates to me that it is using the default value and not a stored value in the config.

The reason why I also store the translations in `default_flag_names` is to make sure that the displayed name of the flag is updated after the function is called. Though after taking another look it seems like that's unnecessary and I can just call `require_refresh()` to accomplish the same thing.

I hope I correctly understood what you meant! If not please clarify further.

PR #3827, opened 2025-02-17
Debounce mathjax rendering to avoid stalling

### LINKED ISSUE #3803: Implements debouncing in mathjax
When I'm writting mathjax, the editor freeze entirely, which makes it hard to use. I think it should be debounced. That is, when there is some change in the mathjax editor, wait, let's say, one second without any changes before sending it to be displayed. 

This would ensure that at most one rendering request is done per second, instead of asking to render all intermediate steps, most of them not being proper mathjax.

Even if I can blind type and do relatively few typos, having the screen frozen sometime for 10 seconds is frustrating simply because when I type an expression such as
```
[F(\alpha_1,\dots,\alpha_g,\alpha_{g+1}):F]= [F(\alpha_1,\dots,\alpha_g,\alpha_{g+1}):F(\alpha_1,\dots,\alpha_g)][F(\alpha_1,\dots,\alpha_g):F]= [(F(\alpha_1,\dots,\alpha_g))(\alpha_{g+1}):F(\alpha_1,\dots,\alpha_g)][F(\alpha_1,\dots,\alpha_g):F] \le [F(\alpha_{g+1}):F]\prod_{a=1}^g\deg\alpha_a=\alpha_{g+1}\prod_{a=1}^g\deg\alpha_i=\prod_{a=1}^{g+1}\deg\alpha_i
```

I want to do copy/paste a lot instead of retyping everything. And I can't copy anything when the UI is frozen.

iamllama:
Closes #3803

Typing in the mathjax editor causes it to rerender, which can be slow when done per character, given a large enough expression

This pr proposes to debounce rendering to svg by 500ms to avoid stalls


ts/editable/Mathjax.svelte
```diff
@@ -25,13 +25,17 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import { pageTheme } from "$lib/sveltelib/theme";
 
     import { convertMathjax, unescapeSomeEntities } from "./mathjax";
+    import { ChangeTimer } from "../editor/change-timer";
```
dae, line 28: While not a problem at the moment as we have a single compilation unit, this is introducing a circular dependency. change-timer would be better moved to editable or something like sveltelib.

dae (commented):
Thanks! Just one minor point:

dae:
Thanks :-)

PR #3828, opened 2025-02-18
Cache rendered mathjax to avoid stalling when editing plaintext

iamllama:
#3827 deals with stalling when editing in the richtext editor's mathjax overlay editor. But when the plaintext editor is changed, all the mathjax custom element components are recreated* and thus rerendered on each change.

This can be much slower since now all the mathjax expressions in the field are re-rendered even if none of them were changed

The fix proposed here is to add and use lru-cache (an existing indirect dep) to cache the rendered svg. A max cache size of 10 is chosen to limit memory use, and should account for most normal usecases with <10 mathjax exprs per field

*this seems to have been noted already in https://github.com/ankitects/anki/pull/1502#issuecomment-976590300. it's also why the debouncing within the component has no effect here

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
         "jquery": "^3.5.1",
         "jquery-ui-dist": "^1.12.1",
         "lodash-es": "^4.17.21",
+        "lru-cache": "^10.2.0",
         "marked": "^5.1.0",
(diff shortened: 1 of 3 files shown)


dae:
LGTM - thank you!

PR #3829, opened 2025-02-18
Feat/support load balance and easy days in FSRS simulator

L-M-Sherlock:
### Key Changes
- Updated FSRS dependency to use a version that supports post-scheduling hooks
- Added `post_scheduling_fn` to simulator that applies load balance and easy days
- Added easy days percentages to simulator proto
- Refactored easy days parsing logic for reusability

### TODO
1. Refactor simulator daily stats chart
   - Current chart becomes unreadable with easy days due to high volatility of the number of reviews per day

2. Review post_scheduling_fn implementation
   - Check for potential code duplication with existing scheduling logic

3. Consider moving simulator to dedicated page
   - Simulator could benefit from more screen space

### Related Issues
- Depends on FSRS-rs PR: https://github.com/open-spaced-repetition/fsrs-rs/pull/283
- UI improvements needed for better visualization of easy days impact

Preview:

image

image


YukiNagat0:
At first glance, same as https://github.com/ankitects/anki/pull/3815 - no support for the `mw.col.load_balancer_enabled` option (disabled `load_balancer`)

dae:
> Updated FSRS dependency to use a version that supports post-scheduling hooks

Nice, that is a cleaner solution.

> Consider moving simulator to dedicated page

I can't think of a better place to put it for now. What about instead of embedding it in the page, you have a clickable "Simulator" link that displays it as as a larger pop-over on the page?

L-M-Sherlock:
> no support for the `mw.col.load_balancer_enabled` option (disabled `load_balancer`)

Fixed in 59d5ef0

> What about instead of embedding it in the page, you have a clickable "Simulator" link that displays it as as a larger pop-over on the page?

Sounds good. But I need help. I don't know how to develop that.

dae:
An example of how to do that is TextInputModal.svelte, which is used in ConfigSelector.svelte

Associated docs: https://getbootstrap.com/docs/5.0/components/modal/

L-M-Sherlock:
Now the simulator is moved to a larger pop-over on the page:

image

And I add a button to allow users to save these settings to the preset:
- Desired retention
- New cards/day
- Maximum reviews/day
- Maximum interval
- New cards ignore review limit

L-M-Sherlock:
Now it looks nice when you want to compare Easy Days with Non Easy Days in one year.

image

YukiNagat0:
Regarding the **readibility** issue, I believe that introducing a `Smooth` toggle would be the most appropriate solution. This allows users the flexibility to choose between a smoothed visualization and the exact number of reviews for a given date, catering to different use cases.

As for the `window size` of the moving average, I personally favor making it configurable, with a default value calculated using the formula: `window_size = ceil(days_to_simulate / 180)` or `/ 360`, so only graphs with a larger learning span are affected by smoothing if the user doesn’t input an exact desired window size.

ts/routes/deck-options/SimulatorModal.svelte
```diff
@@ -0,0 +1,299 @@
...
+    export let computing: boolean;
+    export let simulating: boolean;
+    export let openHelpModal: (key: string) => void;
+    export let onPresetChange: () => void;
+
+    let config = state.currentConfig;
+    let simulateSubgraph: SimulateSubgraph = SimulateSubgraph.count;
+    let tableData: TableDatum[] = [];
+    const bounds = defaultGraphBounds();
+    bounds.marginLeft += 8;
```
Luc-Mcgrady, line 38:
```suggestion
    
    const default_bounds = defaultGraphBounds();
    let bounds = defaultGraphBounds();
    bounds.marginLeft += 8;

    window.addEventListener("resize", function () {
        bounds.width = window.innerWidth - bounds.marginLeft - bounds.marginRight
        bounds.height = window.innerWidth * (default_bounds.height / default_bounds.width)
    });

```
Looks better while full screen in.

Before:
image

After:
image

If you would prefer just the line to change
```js
    svg.append("g")
        .attr("class", "lines")
        .attr("fill", "none")
        .attr("stroke-width", 1.5)
        .attr("stroke-linejoin", "round")
        .attr("stroke-linecap", "round")
        .selectAll("path")
        .data(Array.from(groups.entries()))
        .join("path")
        .attr("vector-effect", "non-scaling-stroke") // < ------ This
        .attr("stroke", (d, i) => color[i % color.length])
        .attr("d", d => line()(d[1].map(p => [p[0], p[1]])))
        .attr("data-group", d => d[0]);
```

image

L-M-Sherlock:
I added the Smooth toggle:

image

image

L-M-Sherlock:
But I don't know why the ticks become larger when I zoom out the window:

image

image

ts/routes/deck-options/SimulatorModal.svelte
```diff
@@ -0,0 +1,383 @@
...
+
+<style>
+    .modal {
+        background-color: rgba(0, 0, 0, 0.5);
+    }
+
+    :global(.modal-xl) {
+        max-width: 100vw;
+    }
+
```
Luc-Mcgrady, line 375:
```suggestion

    .modal :global(.tick) :global(.text) {
        font-size: 10px !important;
    }

```

You can override the font-size changes for just the simulator graph with this.
Though as you say, I suspect there's a purpose in it being the way it is.

L-M-Sherlock: @dae, do you know the purpose of this code?

https://github.com/ankitects/anki/blob/8e13e6bfc159a0d73bbec146c75c16613f9a597e/ts/routes/graphs/Graph.svelte#L39-L58

Expertium:
We should add a tooltip to clarify that the "Smooth" toggle only affects the graph and has nothing to do with actual scheduling.

dae:
The graph is looking pretty bad on mobile/small devices at the moment. I seem to recall it looking a bit better when I checked on this PR the other day, so it might be a regression?

Screenshot 2025-02-21 at 5 30 01 pm

Going through a review now.

One other initial thought was on how extra options have been added, presumably so that the user can configure them directly from that screen, and so it becomes apparent that those options also alter the simulation. But it's not complete, as easy days is not listed there. 

Screenshot 2025-02-21 at 5 29 24 pm

Note how the graph is not actually visible at this screen size - the user has to scroll down.

Is it worth repeating those options there? (yes is an acceptable answer if you think so, I'm just brainstorming here). Are there any where it might make sense to show them in a more concise, read-only form? E.g. something like the following in small grey text: "Other influencing settings: 20 new cards/day, 200 reviews/day, 2 easy days, max interval 36500, ..."?

user1823:
> Is it worth repeating those options there? 

IMO, yes. The user might want to play with these values to see the effect on their future workload without wanting to change the actual settings. Also, if the user has to change the values scattered throughout the deck options page, the simulator would become difficult to use.

ts/routes/deck-options/SimulatorModal.svelte
```diff
@@ -0,0 +1,387 @@
...
+                        $config.reviewOrder = simulateFsrsRequest.reviewOrder;
+                        onPresetChange();
+                    }}
+                >
+                    <!-- {tr.deckConfigApplyChanges()} -->
+                    {"Save to Preset Options"}
+                </button>
+
+                {#if simulating}
+                    {tr.qtMiscProcessing()}
```
dae, line 321: Didn't catch this one in the past - our graphs shouldn't refer to any Qt strings, as they're supposed to only be for the Qt UI. I'll dupe it into the core repo.

YukiNagat0:
Regarding the easy days: As user1823 pointed out, users might want to experiment with settings in the simulator. However, when it comes to easy days, users are unlikely to adjust them frequently. They already know which days are busy for them, so there’s no need to display these settings again on the simulator page.

Objectively, the simulator is a tool for exploration - users can play with it to find their ideal settings, which is why the adjustable options are present on this page. The "Save to Preset Options" button is there just to save users from having to remember the values they set in the pop-up to transfer to the main settings page . Since users don’t need to recall which days they’re busy, there’s no reason to duplicate the easy days settings on the simulator page.

Something like that. Personally, I like the current set of the adjustable settings. For me, `Save to Preset Options` button is redundant: I can remember all 4 values: `DR`, `New cards/day`, `Max interval`, `Sort order`. But I understand, that someone can find this button useful..

ts/routes/deck-options/SimulatorModal.svelte
```diff
@@ -0,0 +1,387 @@
...
+    export let simulating: boolean;
+    export let openHelpModal: (key: string) => void;
+    export let onPresetChange: () => void;
+
+    const config = state.currentConfig;
+    let simulateSubgraph: SimulateSubgraph = SimulateSubgraph.count;
+    let tableData: TableDatum[] = [];
+    const fsrs = state.fsrs;
+
+    const default_bounds = defaultGraphBounds();
```
dae, line 45: camelCase, if this part needs to be preserved

ts/routes/deck-options/SimulatorModal.svelte
```diff
@@ -0,0 +1,387 @@
...
+    bounds.marginLeft += 8;
+
+    function updateBounds() {
+        bounds.width = window.innerWidth - bounds.marginLeft - bounds.marginRight;
+        bounds.height =
+            window.innerWidth * (default_bounds.height / default_bounds.width);
+    }
+
+    window.addEventListener("resize", updateBounds);
+    updateBounds();
```
dae, line 56: Commenting out these two lines considerably improves the graph appearance. What is this dynamic resizing trying to accomplish?

ts/routes/deck-options/SimulatorModal.svelte
dae: Yep, the text can become unreadably small at the smaller widths/breakpoints without those lines:

Screenshot 2025-02-21 at 5 58 37 pm
Screenshot 2025-02-21 at 5 59 06 pm

ts/routes/deck-options/SimulatorModal.svelte
Luc-Mcgrady: https://github.com/ankitects/anki/pull/3829#discussion_r1961831696

ts/routes/deck-options/SimulatorModal.svelte
dae: Ok, you can now use `tr.actionsProcessing` after rebasing over main.

ts/routes/deck-options/SimulatorModal.svelte
dae: Ah, ok. I'd recommend we find another way to do that such as your other suggestion, as the graph code expects to have a consistent internal bounds that gets scaled to the page size.

Expertium:
> They already know which days are busy for them, so there’s no need to display these settings again on the simulator page.

But we need to convey that Easy Days affect the simulation. Plus, as you said, the simulator is a tool for exploration, so I'd say we should add Easy Days sliders to the simulator.

YukiNagat0:
> > They already know which days are busy for them, so there’s no need to display these settings again on the simulator page.
> 
> But we need to convey that Easy Days affect the simulation. Plus, as you said, the simulator is a tool for exploration, so I'd say we should add Easy Days sliders to the simulator.

It is already obvious, if `Days to simulate` <= 365 for any number of easy days (default is 365). Or with disabled `Smooth` option.

I will suggest to set the `Smooth` option to `False` by default.

ts/routes/deck-options/SimulatorModal.svelte
L-M-Sherlock: I removed this code.

dae:
Reviewing now. Are you planning to release the updated fsrs-rs on crates.io? If so, it would be nice to point to that instead of a git commit.

rslib/src/scheduler/fsrs/simulator.rs
```diff
@@ -44,12 +146,43 @@ impl Collection {
...
                 due: ((introduced_today_count + i) / req.new_limit as usize) as f32,
+                interval: f32::NEG_INFINITY,
             });
             converted_cards.extend(new_cards);
         }
+        let deck_size = converted_cards.len();
         let p = self.get_optimal_retention_parameters(revlogs)?;
+
+        let easy_days_percentages = parse_easy_days_percentages(req.easy_days_percentages)?;
+        let next_day_at = self.timing_today().unwrap().next_day_at;
```
dae, line 157: Here too

rslib/src/scheduler/states/load_balancer.rs
```diff
@@ -389,7 +390,7 @@ pub fn select_weighted_interval(
     Some(intervals_and_weights[selected_interval_index].0)
 }
 
-fn interval_to_weekday(interval: u32, next_day_at: TimestampSecs) -> usize {
+pub fn interval_to_weekday(interval: u32, next_day_at: TimestampSecs) -> usize {
```
dae, line 393: Personally, I prefer to use pub(crate) in such circumstances, even if the entire file is already pub(crate), as it keeps things explicit, and won't break if the module is made public in the future for some reason.

ts/routes/deck-options/FsrsOptions.svelte
```diff
@@ -64,7 +51,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     let computingParams = false;
     let checkingParams = false;
     let computingRetention = false;
-    let simulating = false;
+    const simulating = false;
```
dae, line 54: Why define this here, instead of passing it in directly as an argument below? In fact, why pass it as an argument at all, instead of just defining it in SimulatorModal?

dae (commented):
Just a couple of minor points and then good to go on my end 👍

dae:
> Are you planning to release the updated fsrs-rs on crates.io?

Assuming that's a no for now, so I'll merge this in.

dae:
Thank you for your work on this as always!

L-M-Sherlock:
> Assuming that's a no for now, so I'll merge this in.

Yeah, because @Luc-Mcgrady added more features into the simulator:
- https://github.com/open-spaced-repetition/fsrs-rs/pull/293

So the simulator is still not stable now. I plan to update fsrs-rs on crates.io in the next month.

PR #3832, opened 2025-02-20
Increase font size in mathjax preview for accessibility

GithubAnon0000:
Apparently no font size should be lower than 12px, see https://www.boia.org/blog/accessibility-tips-let-users-control-font-size.

With the current 55%, I get a computed font size of 8.25px though. Considering the text shows the helpful message "Press ⁨Enter⁩ to accept, ⁨Shift+Enter⁩ for new line.", I think we should add a minimum font size.

For reference. This is the 55% size:
anki

And this is the size with the change from this PR:
anki


dae:
You're right, it's too small. Thanks!

PR #3837, opened 2025-02-27
Feat: Simulator suspend after lapse count

Luc-Mcgrady:
Adds the "Suspend after X leeches" feature to the simulator.

Limits for examples:
(1: no limit, 2: 14, 3: 7, 4: 100)
image
image

### DIFF
--- a/.vscode.dist/settings.json
+++ b/.vscode.dist/settings.json
@@ -2,7 +2,7 @@
     "editor.formatOnSave": true,
     "[python]": {
         "editor.codeActionsOnSave": {
-            "source.organizeImports": true
+            "source.organizeImports": "explicit"
         }
     },
     "files.watcherExclude": {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2099,7 +2099,7 @@ dependencies = [
 [[package]]
 name = "fsrs"
 version = "3.0.0"
-source = "git+https://github.com/open-spaced-repetition/fsrs-rs.git?rev=f7a746306b5d98712705912d8e0a8846cfb113b2#f7a746306b5d98712705912d8e0a8846cfb113b2"
+source = "git+https://github.com/open-spaced-repetition/fsrs-rs.git?rev=22f8e453c120f5bc5996f86558a559c6b7abfc49#22f8e453c120f5bc5996f86558a559c6b7abfc49"
 dependencies = [
  "burn",
  "itertools 0.12.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2099,7 +2099,7 @@ dependencies = [
 [[package]]
 name = "fsrs"
 version = "3.0.0"
-source = "git+https://github.com/open-spaced-repetition/fsrs-rs.git?rev=034ca9dabe1d429f157b31fa2a97d1156e770816#034ca9dabe1d429f157b31fa2a97d1156e770816"
+source = "git+https://github.com/open-spaced-repetition/fsrs-rs.git?rev=f7a746306b5d98712705912d8e0a8846cfb113b2#f7a746306b5d98712705912d8e0a8846cfb113b2"
 dependencies = [
  "burn",
  "itertools 0.12.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2099,7 +2099,7 @@ dependencies = [
 [[package]]
 name = "fsrs"
 version = "3.0.0"
-source = "git+https://github.com/open-spaced-repetition/fsrs-rs.git?rev=96520531415e032781adfe212f8a5eed216006be#96520531415e032781adfe212f8a5eed216006be"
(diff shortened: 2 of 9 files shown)


user1823:
As the number of settings that influence the simulation continue to grow, I am beginning to like dae's suggestion here:

> Are there any where it might make sense to show them in a more concise, read-only form? E.g. something like the following in small grey text: "Other influencing settings: 20 new cards/day, 200 reviews/day, 2 easy days, max interval 36500, ..."?

_Originally posted by @dae in https://github.com/ankitects/anki/issues/3829#issuecomment-2674187714_

Though it's important to include settings like limits per day and max interval in the simulator page, we can put the minor settings like easy days and leech actions in small grey text at the bottom.

Luc-Mcgrady:
How about something as simple as an "Advanced settings" drop down?
image
image

user1823:
Works for me 👍

YukiNagat0:
Maybe move `Max interval` to main list, instead of advanced drop down?

L-M-Sherlock:
I haven't seen the dropdown of "Easy Days". Did you push the commit?

Luc-Mcgrady:
I included that as a placeholder for the image and then forgot to implement it 😅.

Luc-Mcgrady:
Should be there now.
image
Also made the header sticky so that you can exit out without scrolling up.

Expertium:
Will there be tooltips in this window?

Luc-Mcgrady:
Peek 2025-03-01 12-55
Tooltips for which part? There is the help modal (which opens behind the simulator modal (?)).

L-M-Sherlock (approved):

dae:
Thanks for your work on this @Luc-Mcgrady. Some minor issues/thoughts I noticed:
- the graph is truncated by default in my window - the x axis labels aren't visible
- if we're going to have a 'save to preset options' button, it should work on all settings like max interval as well
- nested disclosures (easy days in advanced) gets a bit confusing. Do we need both?
Screenshot 2025-03-10 at 2 08 14 pm

Luc-Mcgrady:
> the graph is truncated by default in my window - the x axis labels aren't visible

Is scrolling down an issue? I can't see scroll bars on your image like I can on mine.
image

> if we're going to have a 'save to preset options' button, it should work on all settings like max interval as well

I can't take credit for that feature but I believe it already does. 
https://github.com/ankitects/anki/blame/cad6e0b0bfd0b0b4c4777db20678bd066ad21f39/ts/routes/deck-options/SimulatorModal.svelte#L296

> nested disclosures (easy days in advanced) gets a bit confusing. Do we need both

I'll keep both but move easy eays out of the advanced section.

dae:
> Is scrolling down an issue? I can't see scroll bars on your image like I can on mine.

Scrollbars are hidden on a Mac by default. I've made some tweaks to the layout that seems to fix the starting height for me - does that work for you? Otherwise good to go on my end.

> I can't take credit for that feature but I believe it already does.

Sorry, I tested it earlier and it didn't work, but that's because I tried increasing the limit and it was already at its max.

YukiNagat0:
**Issues Encountered**:

1. **Unexpected Scroll Behavior**
With latest commit (https://github.com/ankitects/anki/pull/3837/commits/f37206a5ef63c83584e7bd7f206ba4f4efe7936a) it is harder to observe, but you can scroll the `Options for Deck` window while interacting with the `FSRS simulator`. The attached recording demonstrates this behavior, where the secondary scrollbar (from the `Options` window) moves independently.

https://github.com/user-attachments/assets/dde4ca86-8447-43e2-95bb-1388d1ea345e

2. **Graphical Artifacts on Windows**
When using `Direct3D` or `Vulkan` video drivers artifacts on Windows you can sometimes observe video artifacts - it is not a new thing, but it was never addressed. However, I've noticed that specifically on the `FSRS simulator` pop-over that artifacts are significantly worse: appear very often and cover all the screen. These artifacts are absent when switching to `OpenGL` or `Software` video drivers .

https://github.com/user-attachments/assets/0e87f649-e037-473a-a77b-5ae30f55c1d5

dae:
Some CSS wizard may be able to resolve the former issue. The latter issue is with that video driver, and not something we really have the power to control AFAIAA.

YukiNagat0:
> The latter issue is with that video driver, and not something we really have the power to control AFAIAA.

Speaking of video drivers, do you know if it has become possible to add fullscreen support for `OpenGL`? On my machine `OpenGL` seems to be the most stable, but what keeps me from using it is the fact that it does not support fullscreen.

Luc-Mcgrady:
> does that work for you? Otherwise good to go on my end.

LGTM! :rocket:

dae:
I'd be surprised if Qt had fixed the full screen issues. You can check yourself by removing the check inside Anki.

PR #3838, opened 2025-03-02
feat: add title to some of the sync dialogs

BrayanDSO:
Context: I use Windows 11 and [komorebi](https://github.com/LGUG2Z/komorebi) as my window manager. It can differentiate windows by executable, window class, and window title. The issue is that many Anki windows spawn with `Anki` as title and change it later, so the program can't differentiate them when they spawn. Also, there are dialogs/windows without a proper title, like the sync conflict dialog.

So I propose changing the title of some of the sync dialogs. Despite of my original reason, I think that the titles should be improved anyway.

The changes are described in the commits

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -98,7 +98,7 @@ gnnoh <gerongfenh@gmail.com>
 Sachin Govind <sachin.govind.too@gmail.com>
 Bruce Harris <github.com/bruceharris>
 Patric Cunha <patricc@agap2.pt>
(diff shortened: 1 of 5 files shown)


BrayanDSO:
> 69634269+BrayanDSO, at the domain users.noreply.github.com
> Author 69634269+brayandso, at the domain users.noreply.github.com NOT found in list

Apparently the script needs to handle different casing. Or just use the GitHub ID instead of trying to get the user name as well.

PR #3839, opened 2025-03-02
fix: Task manager exception handling

RisingOrange:
This fixes some issues with `aqt.taskman.TaskManager`:
- `TaskManager.run_in_background()` can raise an exception if there was an exception in a (possibly unrelated) background task that was just completed. When this happens, the new background task is not run and the exception bubbles up to the code that called `run_in_background()`.
  - This can lead to confusing issues. For example, we encountered an [issue](https://github.com/AnkiHubSoftware/ankihub_addon/pull/1098) where the central exception handler of the AnkiHub add-on was raising an exception after trying to start a background task (for uploading logs). As a fix, we wrapped the `run_in_background()` calls in the exception handler with `run_on_main()`. We encountered similar issues with this before.
- When there are multiple closures in `TaskManager._closures`, and one of the closures raises an exception, the other closures are not called at all

## Proposed changes
- In `TaskManager._on_closures_pending()`, catch exceptions and re-raise them using `QTimer.singleShot` with a delay of 0 instead of allowing them to bubble up immediately.
  - Effects:
    - All closures are run
    - The caller of `run_in_background()` is not interrupted by exceptions frogm closures the and the new background task is scheduled successfully
    - All exceptions from closures are raised

## Risks
- Maybe some code relies on the current taskman exception-handling behavior?
  - If making this change is too risky, it could be adjusted. However, it would be good to prevent exceptions from bubbling up to the caller of `run_in_background()`, or at least document that this can happen.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -213,6 +213,7 @@ GithubAnon0000 <GithubAnon0000@users.noreply.github.com>
 Mike Hardy <github@mikehardy.net>
 Danika_Dakika <https://github.com/Danika-Dakika>
 Mumtaz Hajjo Alrifai <mumtazrifai@protonmail.com>
+Jakub Fidler <jakub.fidler@protonmail.com>
 
 ********************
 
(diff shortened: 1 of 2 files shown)


dae:
Seems like a reasonable approach. Thanks for digging into this!

PR #3840, opened 2025-03-04
Feat/grade now

### LINKED ISSUE #3694: Provide a 'grade now' feature
Useful to e.g. push cards forward after they've been reviewed externally, and because FSRS would prefer every rescheduling have a grade associated with it.

The scheduling code currently assumes that a card will only be answered when it's at the top of the scheduling queue. The queue is potentially tens of thousands of cards big, and it's not ideal if we have to search through the entire list on every card answer to locate the matched card and remove it from the queue (in `queue/mod.rs`).

Perhaps one way we could avoid unwanted extra processing is to refactor the card answering so that there is a from_queue flag. If false, the entire queue would be searched, and there would be no error generated if the card was not found in the queue (when e.g. the user has rated a card that's not due yet).

Presumably we should ignore suspended cards in such a mode, or throw an error when attempting to reschedule them?

L-M-Sherlock:
This PR adds a new "Grade Now" feature to the browser, allowing users to grade cards directly without going through the review process. This is useful for quickly adjusting card states in bulk.

## Changes

1. Added a new action in the browser UI with keyboard shortcut Ctrl+Shift+G
2. Added backend support for grading cards outside the normal review flow
3. Added proper queue management for graded cards
4. Added translation strings and UI elements

## Implementation Details

The implementation includes:

- A new `GradeNow` operation in the backend
- A dialog in the browser that shows the four standard grading buttons (Again, Hard, Good, Easy)
- Proper handling of queue updates after grading
- Support for grading multiple selected cards at once

## Screenshots

image

image

image

close #3694


Expertium:
I think "Set Due Date" should be removed then

user1823:
> I think "Set Due Date" should be removed then

No, both have their own use cases.
- Grade now is used when the user has recalled the material externally - successfully or unsuccessfully .
- Set due date is used when the user doesn't want to answer the card at that time (e.g. got passively exposed to the information). 

SuperMemo also has both the options: https://supermemopedia.com/wiki/Ctrl%2BJ_vs._Ctrl%2BShift%2BR

Expertium:
I thought the whole point of adding "Grade Now" was to remove "Set Due Date" since FSRS can't properly work with it?

JSchoreels:
Would it be feasible to make it accessible through APIs like AnkiConnect so third party tools like Yomitan could be used to mark a word as "Forgotten" ? For example, if I'm reading a book, I hover a word I forgot, if I see it's already in Anki and has reviews, I could mark it as "Again" meaning I forgot it, and thus my external review would be counted by the scheduler as a review

user1823:
> if I'm reading a book, I hover a word I forgot, if I see it's already in Anki and has reviews, I could mark it as "Again" meaning I forgot it, and thus my external review would be counted by the scheduler as a review

Yeah, that's the purpose of adding the Grade now feature.

However, I can't say whether connecting this to third party tools would be easily possible or not.

YukiNagat0:
> However, I can't say whether connecting this to third party tools would be easily possible or not.

All that’s needed is to expose the `grade_now` API (in the programming interface sense, not a web-API). This would allow tools like `Anki Connect` or other Anki add-ons to interact with the feature directly through Python code.

YukiNagat0:
**upd**: It’s worth noting that `Anki Connect` already provides a [Graphical Actions](https://git.sr.ht/~foosoft/anki-connect#graphical-actions) API, which allows users to view selected cards. This means `Anki Connect's` existing API is already capable of handling the task.

For instance, `Yomitan` demonstrates this functionality by enabling users to quickly view duplicates (cards already in their collection) for a specific word. This same mechanism could be adapted to `Grade Now` a card:

1. Scan the word you want to Grade Now (via pop-up or Search Page).
2. Click the view-note button.
3. Grade the card directly in Anki using the built-in Grade Now interface.

YukiNagat0:
The advantage of leveraging the [Graphical Actions](https://git.sr.ht/~foosoft/anki-connect#graphical-actions) API is that developers of end-apps (e.g., `Yomitan`) wouldn’t need to create their own custom interface for displaying grades or handling grading. Instead, they can seamlessly redirect users to `Anki`’s native interface, where users can perform a variety of actions on the card: grade it, edit it, delete it, and so on.

brishtibheja:
> I thought the whole point of adding "Grade Now" was to remove "Set Due Date" since FSRS can't properly work with it?
@Expertium I think people should be allowed to grade the cards when they're using SDD. Or, allow setting a due date when using GN. That was also the initial idea with this feature if I remember correctly, i.e. reschedules will stop having a manual `0` entry making it easier for FSRS to schedule those cards.

L-M-Sherlock:
To solve the set due date problem, we may require the user to grade before set due date.

brishtibheja:
I don't think it'd be confusing per se. Just that the issues with SDD will still be there for some users. There's a good way to solve those by having users use GN along with SDD, but if they're seperated like this users won't know they've got to do this.

Danika-Dakika:
[I question whether a PR for one feature is a useful place to discuss removing/altering another feature, but I'll respond here, since the discussion is already underway.]

There's nothing confusing about having Set Due Date and Grade Now as 2 separate features -- they do completely different things. It's great if Grade Now can address some SDD use-cases, specific to _New_ cards, that were challenges for FSRS, but it certainly doesn't cover all of the uses. For instance, Grade Now doesn't allow a user to set a card to be due on a particular date/range. 

My primary use case for SDD is when I've made major changes to a note, even adding New siblings that will be introduced soon, and I don't want to wait months to study the existing Review siblings again. For that, Grade Now would be useless to me. I don't want to _skip_ studying those cards and grade them today -- I want to _actually_ study them, just sooner. 

SDD is also a reasonable technique for addressing a missed-day/vacation backlog: scatter those overdue cards across the next X days. There are better techniques -- but it's a fast and simple alternative for folks who don't want to deal with Filtered decks, and it's better **by far** than the "postpone your whole collection X days"-type add-ons.

rslib/src/scheduler/answering/mod.rs
```diff
@@ -363,14 +371,22 @@ impl Collection {
...
+                matches!(
+                    answer.new_state,
+                    CardState::Filtered(FilteredState::Preview(PreviewState {
+                        finished: true,
+                        ..
+                    }))
+                ),
+            )
+        } else {
+            self.update_queues_after_grading_card(&card, timing)
```
dae, line 388: This is not very efficient - removing a single card requires a search through the queues, and the user may have selected hundreds or thousands of cards at a time. Especially for larger selections, this is probably going to perform a lot better if the seen card ids are collected into a HashSet, and then the queues are filtered with .retain() after all cards have been processed.

Alternatively we could discards the queues and have them rebuilt, which is what is currently done when using SDD.

dae (commented):
Thanks Jarrett, I've left some comments about the technical side of things below.

I too find myself wondering how exactly we should integrate SDD and GC into Anki. Would some sort of unified interface where the user could specify grade and/or due date make sense? Are there situations where it would make sense for FSRS users to use SDD without grading the card?

user1823:
> Are there situations where it would make sense for FSRS users to use SDD without grading the card?

Currently, I use SDD in only one situation and in that situation, grading the card would be harmful.

I use SDD when I am in the reviewer and encounter a card that I don't want to answer for the next few days (depending on its stability). Basically, I use SDD as a "prolonged bury" feature.

Why I don't want to answer it?
It's because I saw a very related card that day or 1-2 days ago, which is artificially boosting my memory of the answer. If I grade the card, it will be pushed way far in the future and I will likely forget it by then. By using SDD, I "bury" the card for 10-15 days, assuming that the effect of the passive exposure would have vanished till then and I can test myself fairly.

What would happen if I rate it `Good` and then use SDD?
Experience suggests that if I press Good on a card after 1 year and then press Again after 10 days, FSRS gets terribly confused (because FSRS doesn't know that I rated Good only because I read the fact 5 minutes ago).
In fact, in the past, I have had to delete these faulty `Good` ratings from my collection in order to make FSRS work in a reasonable manner. See https://github.com/open-spaced-repetition/fsrs4anki/issues/572#issuecomment-1872799384

rslib/src/scheduler/answering/mod.rs
```diff
@@ -309,17 +309,25 @@ impl Collection {
     /// Answer card, writing its new state to the database.
     /// Provided [CardAnswer] has its answer time capped to deck preset.
     pub fn answer_card(&mut self, answer: &mut CardAnswer) -> Result<OpOutput<()>> {
-        self.transact(Op::AnswerCard, |col| col.answer_card_inner(answer))
+        self.transact(Op::AnswerCard, |col| col.answer_card_inner(answer, true))
```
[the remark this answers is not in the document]
L-M-Sherlock: I moved it into `CardAnswer` in c1730d8.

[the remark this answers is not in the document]
L-M-Sherlock: But it fails the ci test.

rslib/src/scheduler/answering/mod.rs
```diff
@@ -309,17 +309,25 @@ impl Collection {
...
             .storage
             .get_card(answer.card_id)?
             .or_not_found(answer.card_id)?;
         let original = card.clone();
         let usn = self.usn()?;
 
+        if card.queue == CardQueue::Suspended {
+            invalid_input!("Can't answer suspended cards");
+        }
+
```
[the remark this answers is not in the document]
L-M-Sherlock: Done in c1730d8

rslib/src/scheduler/answering/mod.rs
L-M-Sherlock: I implemented the first suggestion in 91e1d6f.

rslib/src/scheduler/queue/mod.rs
```diff
@@ -242,6 +243,78 @@ impl Collection {
...
+                // Requeue any learning cards that need to be requeued
+                for &card_id in processed_card_ids {
+                    if let Some(card) = self.storage.get_card(card_id)? {
+                        if let Some(_requeued_learning) =
+                            queues.maybe_requeue_learning_card(&card, timing)
+                        {
+                            // We don't need to save undo information here since
+                            // this is a batch operation
+                            // and grade now can be undone as a normal op
+                        }
```
dae, line 310: Looking at this more carefully, I find myself wondering if you've actually accidentally taken the second approach I mentioned? `maybe_clear_study_queues_after_op()` does not cover the grade now case, so at the end of the operation, so the next queue operation after this one causes the queues to be rebuilt. If I'm right about that, all this code could be dropped.

dae (commented):
Getting there! A couple more technical concerns below.

On the GUI end, I agree with @iamllama that a dropdown may be a better choice - the current pop-up feels a bit clunky, and can't be navigated easy with a keyboard. But that can be deferred to a follow-up PR by someone with more of a GUI focus.

Or not. I still also find myself wondering whether it might make sense to present these two features with a unified interface, allowing one to grade & optionally set due date or vice versa, but I don't have any great ideas for how it could be done at the moment.

rslib/src/scheduler/queue/mod.rs
L-M-Sherlock: I didn't know `maybe_clear_study_queues_after_op()` before you mentioned it :-). I will check whether the code is useless.

L-M-Sherlock: OK, I guess this info means I can remove the function `update_queues_for_processed_cards` safely?

```
[rslib/src/scheduler/queue/mod.rs:213:9] &op = OpChanges {
    op: GradeNow,
    changes: StateChanges {
        card: true,
        note: false,
        deck: true,
        tag: false,
        notetype: false,
        config: false,
        deck_config: false,
        mtime: true,
    },
}
[rslib/src/scheduler/queue/mod.rs:214:9] &op.requires_study_queue_rebuild() = true
```

Issue #3802, opened 2025-02-08
Text cursor misplacement in Preferences > Review > Scheduler


jcznk:
Testing on Anki 25.02rc1, Windows 11. This issue also occurred in the most recent previous versions, or rather, it has always occurred since the text string was moved inside the input field, which IIRC was done for accessibility reasons.

Context: Preferences > Review > Scheduler > Next day starts at / Learn ahead limit / Timebox time limit.

Currently, the text cursor can be placed in any of the following three positions:
1. right before the numerical value
2. right after the numerical value
3. right after the text string

When the cursor is placed after the text string, you cannot edit the numerical value by typing; you can only change it using the up/down arrows, either via the keyboard or the UI element.
 → My suggestion is to prevent the text cursor from ever being placed after the text string.

![Image](https://github.com/user-attachments/assets/165699ff-fabb-4ff5-936a-fefc241b6b3b)

soleuniverse101:
I was able to make it so that the user can't select the suffix text, nor place their cursor after it. I linked a PR to the issue : https://github.com/ankitects/anki/pull/3842

PR #3842, opened 2025-03-05
feat: prevent selection of suffix text in preferences spinboxes

soleuniverse101:
(see issue #3802)

It was done by adding to the spinboxes an event filter that checks whether the selection or cursor are in a valid state.

I hope that I correctly followed the contribution guidelines regarding type hints. I had to cast some of the objects to pass the ninja mypy tests.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -98,7 +98,7 @@ gnnoh <gerongfenh@gmail.com>
 Sachin Govind <sachin.govind.too@gmail.com>
 Bruce Harris <github.com/bruceharris>
 Patric Cunha <patricc@agap2.pt>
-Brayan Oliveira <github.com/BrayanDSO>
+Brayan Oliveira <github.com/brayandso>
 Luka Warren <github.com/lukawarren>
 wisherhxl <wisherhxl@gmail.com>
 dobefore <1432338032@qq.com>
--- a/qt/aqt/progress.py
+++ b/qt/aqt/progress.py
@@ -139,6 +139,7 @@ class ProgressManager:
         label: str | None = None,
         parent: QWidget | None = None,
         immediate: bool = False,
+        title: str = "Anki",
     ) -> ProgressDialog | None:
         self._levels += 1
         if self._levels > 1:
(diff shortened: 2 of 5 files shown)


dae:
Thank you for taking the time to contribute! I'm a bit nervous about unintended regressions here - it's a fairly unimportant issue, and any change can introduce unintended breakage. For example, we now get these errors when closing Anki:

Error in sys.excepthook:
TypeError: 'NoneType' object is not callable

Original exception was:
  File "/Users/dae/Work/desktop/anki/qt/aqt/preferences.py", line 56, in eventFilter
    assert isinstance(spinbox, QSpinBox)
AssertionError

soleuniverse101:
> Thank you for taking the time to contribute! I'm a bit nervous about unintended regressions here - it's a fairly unimportant issue, and any change can introduce unintended breakage. For example, we now get these errors when closing Anki:
> 
> Error in sys.excepthook: TypeError: 'NoneType' object is not callable
> 
> Original exception was: Traceback (most recent call last): File "/Users/dae/Work/desktop/anki/qt/aqt/preferences.py", line 56, in eventFilter assert isinstance(spinbox, QSpinBox) AssertionError

Sorry, I'm not completely familiar how Qt's types work and I think I misinterpreted how a certain function is called. I corrected the error you mentioned by properly testing the types of the objects I use in the event filter function. I wrongly assumed that I could cast them instead...

Thank you for pointing that out! I know that it's not a priority fix, I just wanted to start contributing to Anki with something small. I understand your concerns about introducing unnecessary change.

dae:
I've spotted another potential problem - I'm not sure Python's .len() and Qt's position code are going to agree when the QLineEdit contains non-Latin text comprised of multiple codepoints. I'm afraid I'm going to err on the side of caution and say no to this particularly change for now, but I do really appreciate you taking the time to contribute, so please don't give up on contributing!

soleuniverse101:
> I've spotted another potential problem - I'm not sure Python's .len() and Qt's position code are going to agree when the QLineEdit contains non-Latin text comprised of multiple codepoints. I'm afraid I'm going to err on the side of caution and say no to this particularly change for now, but I do really appreciate you taking the time to contribute, so please don't give up on contributing!

Sure, thank you for your feedback!

PR #3844, opened 2025-03-06
Prevent accidental dragging of audio playback buttons and hint links

beyondcompute:
Currently if a user drags slightly (happens to me often unintentionally) upon clicking an audio play button or a hint link, the action does not happen.

#### Some videos that demonstrate the issue below.

_I first click the button successfully but then twice I click it with dragging slightly after `mousedown` (this video also contains an audio track):_

https://github.com/user-attachments/assets/6ac02a5c-a459-4bcc-89f6-ea22d0812ca8

_An example with a hint link:_

https://github.com/user-attachments/assets/147f142d-39da-40fd-b3a9-99debae33903

#### After the attempted fix:

https://github.com/user-attachments/assets/8470fd6f-26df-4c38-bde3-e769ff88e370

hint_inspection

-------

Hopefully that will be good from the accessibility standpoint.

Thank you!

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -207,6 +207,7 @@ Luca Auer <lolle2000.la@gmail.com>
 Niclas Heinz <nheinz@hpost.net>
 Omar Kohl <omarkohl@posteo.net>
 David Elizalde <david.elizalde.r.a@gmail.com>
+beyondcompute <beyondcompute@gmail.com>
 Yuki <https://github.com/YukiNagat0>
 wackbyte <wackbyte@protonmail.com>
 GithubAnon0000 <GithubAnon0000@users.noreply.github.com>
--- a/qt/aqt/sound.py
+++ b/qt/aqt/sound.py
@@ -845,7 +845,7 @@ def av_refs_to_play_icons(text: str) -> str:
 
     def repl(match: re.Match) -> str:
         return f"""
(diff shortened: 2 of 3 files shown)


beyondcompute:
@dae Thank you for the review! 🙏

Would setting `draggable="false"` for audio playback button at least be acceptable?

beyondcompute:
@dae I’ve modified the PR, please let me know if that would be acceptable. 😊

Also, could I ask a question about writing an addon (please ignore if you don’t have time, etc.)? Is it possible these days to inject own HTML while rendering a template? I would just like to wrap `{{Word}}` from the deck’s template into a custom `<span>` element or something like that.

I saw, `card_will_render` hook was [removed here](https://github.com/ankitects/anki/commit/2a0b480103a37fea7e33737f9916351d31b6477b). And now there is the [`field_filter`](https://github.com/ankitects/anki/blob/9b5da546be49f37c8d6c286e09c86074b2f0c278/pylib/anki/template.py#L324) but that only seems to be called for custom filters. Is it possible to modify the rendering otherwise or is that against Anki’s philosophy and we are encouraged to modify deck’s templates instead?

Thank you! ⭐

dae:
I think that's a good compromise, and much less likely to break things - thank you!

I think the comments at the top of https://github.com/ankitects/anki-addons/blob/main/demos/field_filter/__init__.py may address your question about field_filter.

PR #3847, opened 2025-03-07
Added: Max interval too low warning.

Luc-Mcgrady:
Peek 2025-03-07 16-15

If the max interval is set to a value below 180 days (6 months) (gif is outdated), display a warning.
If the max interval is below 50 days, display a more prominent warning.

I'm not sure what the best thresholds for this are so change them as you wish.

Inspired by: https://www.reddit.com/r/Anki/comments/1j4t3zd/i_need_help_with_my_fsrs_i_think/


YukiNagat0:
How about lowering the threshold for the warning to 180 days? 
In the Japanese learning community, this is a commonly recommended interval, and I personally use 180 days as my `Maximum interval` as well.
As for the `alert-danger` threshold, I think 50 days works well.

Issue #3845, opened 2025-03-07
FSRS "evaluate" message has only partial i18n support


sommerluk:
**How to reproduce:**

In the deck options, FSRS, click on the button "Evalute". You get a message dialog like that:

![Image](https://github.com/user-attachments/assets/58be0f2d-9713-40a4-b833-37741678b381)

**Actual behaviour:**

The first part of the string "Log loss: 0.4109, RMSE(bins): 4,73%." is not internationalized: It does not show up in Pontoon for translation.

**Expected behaviour:**

All elements of the string show up in Pontoon for translation.

PR #3848, opened 2025-03-07
feat: add translation string to FSRS evaluation message

soleuniverse101:
(see issue #3845)

I think that translating those terms gives better insight into the FSRS algorithm, however they are not easy to translate since they're mathematical/algorithmic terms


soleuniverse101:
I'm not sure how I'm supposed to make the ckeck pass

dae:
I think there's still some debate among the FSRS devs about what exactly we should be showing there, and if we should be showing the information at all. I'd like for them to make a decision on that before we start asking translators to translate the message.

The errors were unrelated to your changes, and should go away when this PR is rebased over the latest main.

L-M-Sherlock:
These values have been hidden and only appeared when the old evaluate functionality is toggled by advanced users.

#4149

dae:
I thank you for contributing, and am sorry about the wasted effort.

PR #3850, opened 2025-03-09
Automatically select decks created in the deck browser

marcus8448:
When a new deck is created in the deck browser, it is now automatically selected so that cards can be added immediately without having to search for the deck in the list.


marcus8448:
I'm not sure about implementing this on the Rust layer, as I don't think new decks should be selected in all contexts.

I believe this change makes sense on the deck browser screen, as when I create a deck I normally want to add to it immediately. However, if I were to be studying a deck and decide to add cards to a new child deck (for example), I would not expect the selection to change.

I could make a new RPC call to add & select on the Rust side (if that helps), but that would still need changes to the other clients. Thoughts?

dae:
What if the user happens to be studying while the deck browser screen is open? I suspect if we try and make it selective, we're just going to end up causing confusion. And I suspect some people won't like this change, if we don't make it optional. Perhaps this is best discussed over on the forums first, to gauge demand and solicit ideas?

Issue #3830, opened 2025-02-18
Shift editor control from Python to TypeScript


dae:
Currently, the editor is driven by editor.py. When the user selects a note,
or changes a notetype/deck in the add screen, our Python code sends a bunch of
JavaScript to the webview in loadNote(). This presents a number of problems:

- It's messy - we have web implementation details leaking into the Python code
- It's prone to race conditions due to the asynchronous message passing, and
so our code has hacks like sending the current note ID back to Python to ensure
they match.
- It's a necessary step to being able to integrate the editor into a future
implemented-in-Svelte browse screen.
- And most importantly, it makes integrating the editor into the mobile clients
impractical, since they would need to duplicate all the logic that we currently
have in the Python code.

I propose we shift the majority of the logic into the web end. For the editcurrent
and browser cases, the Python-visible API would consist of something very simple
like loadNoteId(bigint | null). That would cause the JS side to request all required
information via mediasrv, instead of editor.py being responsible for pushing note
data & metadata with eval().

For the addcards case, the editor would need to be wrapped in another Svelte component
that provides selectors for deck and notetype, as we currently implement them in Qt.

Some challenges we'll face:

- Add-ons currently rely on Python hooks and monkey patching to alter editor behaviour,
and it may not be possible to preserve existing functionality in a backwards-compatible
way.
- Copy/paste and drag & drop might be tricky. Are the JS APIs sufficient for us to handle
this (mostly) on the JS end? Or will we need to continue handling this separately for each
client, and sending messages to the webview?

Back when Henrik was actively involved, he made decent progress on some of these changes.
Unfortunately, I felt the changes were too disruptive at the time, and the PRs ended
up withering.

- Loading via mediasrv: https://github.com/ankitects/anki/pull/1691
- Notetype/deck selector: https://github.com/ankitects/anki/pull/1881

They've likely developed a bunch of conflicts since then, but they may still be useful
as a reference.

Since this is a fairly large change, I think we'll want to start with a basic proof of concept
to feel out the add-on and pasting issues (e.g. ignoring the addcards case for now) first.

rlanday:
[...]

PR #3851, opened 2025-03-09
Feat: Created decks inherit parent's config

Luc-Mcgrady:
Peek 2025-03-08 23-15

Unsure if this should be a toggle-able setting or not.

### DIFF
--- a/rslib/src/decks/addupdate.rs
+++ b/rslib/src/decks/addupdate.rs
@@ -105,11 +105,7 @@ impl Collection {
         deck.name = NativeDeckName::from_native_str(machine_name);
         let parent_deck = self.first_existing_parent(machine_name, 0);
         if let Ok(Some(parent_deck)) = parent_deck {
-            if let DeckKind::Normal(parent_deck) = parent_deck.kind {
-                if let DeckKind::Normal(deck) = &mut deck.kind {
-                    deck.config_id = parent_deck.config_id;
-                }
-            }
+            deck.maybe_inherit_parent_config(&parent_deck);
         }
         deck.set_modified(usn);
         self.add_deck_undoable(&mut deck)
@@ -132,11 +128,7 @@ impl Collection {
                 parent_deck.name,
                 &child_split[parent_count..].join("\x1f")
             ));
-            if let DeckKind::Normal(parent_deck) = parent_deck.kind {
-                if let DeckKind::Normal(deck) = &mut deck.kind {
-                    deck.config_id = parent_deck.config_id;
-                }
-            }
+            deck.maybe_inherit_parent_config(&parent_deck);
             if need_create {
                 self.create_missing_parents(&deck.name, usn)?;
             }
--- a/rslib/src/decks/addupdate.rs
+++ b/rslib/src/decks/addupdate.rs
@@ -103,6 +103,14 @@ impl Collection {
     fn add_parent_deck(&mut self, machine_name: &str, usn: Usn) -> Result<()> {
         let mut deck = Deck::new_normal();
         deck.name = NativeDeckName::from_native_str(machine_name);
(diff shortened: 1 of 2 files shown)


iamllama:
If an existing deck with its own preset and params gets reparented (i.e `jp` -> `NEW::jp` or `langs::jp`), should it really inherit the `Default` preset or the parent's? I feel this might catch users off guard when they review without double-checking the config

Luc-Mcgrady:
Personally, I have a `Group` preset with a 9999 learn limit and no cards of its own.
Say I have a collection like
```
Group1 (group)
|- Deck1 (actual preset)
Group2 (group)
```
Should re-parenting Deck1 to Group2 give it the group preset?
Maybe a "Do you want to change preset to match the parent deck?" popup would work here?

YukiNagat0:
I thought this PR addresses only the case when the new deck is created with `old_deck::new_deck` syntax, where `old_deck` has `old_config` and `new_deck` inherits the `old_config`.
The changing behavior (drag existing deck into some other deck to create nested structure) needs to be preserved: dragging the deck (changing it's place in hierarchy) should not change it's config.

Luc-Mcgrady:
> I thought this PR addresses only the case when the new deck is created with old_deck::new_deck syntax, where old_deck has old_config and new_deck inherits the old_config

That is correct.

iamllama:
> Maybe a "Do you want to change preset to match the parent deck?" popup would be worth it here?

On one hand, renaming decks isnt that common an op that a popup would be obstructive. But on the other, it's hard to see users wanting to switch away from their deck's existing preset, key issue being its own optimised params (that the `Group2` preset doesn't necessarily have)

> The changing behavior (dragging existing deck into some other deck to create nested structure) needs to be preserved: dragging the deck (changing it's place in hierarchy) should not change it's config.

I think having *new* decks inherit their parent's preset is a good change, and i agree with yuki in that existing decks should be left alone by the impl

Luc-Mcgrady:
Oh sorry I misunderstood. I also don't think changing the preset while re-naming the decks is a good change.

Right now, only the newly created parent deck will inherit from default.

https://github.com/user-attachments/assets/2338e83e-5c4a-47ad-a59a-39aa05342ade

So you're saying that `langs` should have the preset `jp`? Or were you just worried that `jp` would be reset to `Default`?

Luc-Mcgrady:
Ok I see the problem now, If langs already exists then jp currently inherits from langs.

dae:
I'm not really fond of this approach I'm afraid - the parent handling code is now more complicated as it can't focus on only decks. Rather than making this magical/operate at such a low level, perhaps it could be better done as some explicit 'add child deck' action that optionally allows one to inherent from the parent preset?

user1823:
> some explicit 'add child deck' action

That would be a nice improvement. AnkiDroid has had a "create subdeck" option for ages.

Luc-Mcgrady:
The main problem I seek to solve with this PR is that people may inadvertently create "Default" decks and then forget to set them to the correct preset. So long as "Create Deck" exists in its current form, this problem will continue to occour when people use it.

That being said, a "Create Subdeck" option would work for me 👍 (So long as I remember to use it and can access it from the "Add Card" deck picker somehow). 
It may be mildly confusing to people what exactly the different functionality between "Create Subdeck" and "Create Deck" would be. Maybe instead of just "Create Subdeck" it could be "Copy Preset For New Deck" (or some more consisely worded version)?

dae:
The deck selection window in the add screen has an 'add' button. That dialog could potentially have a 'use parent preset' checkbox when a deck is a child for example. Please note that work done in this area is likely to be short-lived, due to #3830

Luc-Mcgrady:
> Please note that if done now, work done in this area is likely to be short-lived, due to https://github.com/ankitects/anki/issues/3830

I'll wait then :+1:

PR #3856, opened 2025-03-11
Fix cargo ignoring lockfile when building syncserver image

### LINKED ISSUE #3854: [Bug] Docker files for syncserver broke
This precise command used to work a couple days ago, but something has changed (without any code changing) that broke it.

To reproduce, from the root of the repo, execute:

```
cd docs/syncserver
docker build -f Dockerfile --no-cache --build-arg ANKI_VERSION=25.02 -t anki-sy
nc-server .
```

A couple days ago, this would complete successfully. However, now, with the exact same code, it results in an error:

```
error: failed to compile `anki-sync-server v0.0.0 (https://github.com/ankitects/anki.git?tag=25.02#038d85b1)`, intermediate artifacts can be found at `/tmp/cargo-installPIAt8G`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  failed to parse manifest at `/usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/priority-queue-2.2.2/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.83.0 (5ffbef321 2024-10-29)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
The command '/bin/sh -c cargo install --git https://github.com/ankitects/anki.git --tag ${ANKI_VERSION} --root /anki-server  anki-sync-server' returned a non-zero code: 101
```

The obvious fix would be to upgrade Cargo from 1.83.0 to 1.85.0. Doing so makes the build get further, but it will fail to a different (non-trivial) error.

iamllama:
Fixes #3854

Was able to reproduce. fsrs-rs' indirect dep priority-queue bumped its edition from 2021 to 2024 in a patch version recently, and although anki's lockfile has it pinned to 2.1.1, cargo [ignores](https://doc.rust-lang.org/cargo/commands/cargo-install.html#dealing-with-the-lockfile) the lockfile when installing if the `--locked` flag isn't passed to it

When building the sync server image, `2.1.1` in fsrs-rs' Cargo.toml is [equivalent](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements) to `^2.1.1`, so cargo pulls 2.2.2, which it can't handle yet (on 1.83)

The fix proposed is to pass the `--locked` flag to cargo in the dockerfile

### DIFF
--- a/docs/syncserver/Dockerfile
+++ b/docs/syncserver/Dockerfile
@@ -7,6 +7,7 @@ RUN apk update && apk add --no-cache build-base protobuf && rm -rf /var/cache/ap
 RUN cargo install --git https://github.com/ankitects/anki.git \
 --tag ${ANKI_VERSION} \
 --root /anki-server  \
+--locked \
 anki-sync-server
 
 FROM alpine:3.21.0
--- a/docs/syncserver/Dockerfile.distroless
+++ b/docs/syncserver/Dockerfile.distroless
@@ -7,6 +7,7 @@ RUN apt-get update && apt-get install -y build-essential protobuf-compiler && ap
 RUN cargo install --git https://github.com/ankitects/anki.git \
 --tag ${ANKI_VERSION} \
 --root /anki-server  \
+--locked \
 anki-sync-server
 
 FROM gcr.io/distroless/cc-debian12


SimonBaars:
Tested this solution, it works! Thanks for looking into this.

Btw, be sure to also update the `Dockerfile.distroless` (it's in the same folder).

dae:
Thanks!

PR #3858, opened 2025-03-12
Fix broken window decorations on unpackaged GNOME instances

krischerven:
Currently, there are no window decorations when testing on GNOME systems. This is inconvenient and also causes a segfault when the main window is closed.


dae:
Happy to merge once tests pass.

PR #3859, opened 2025-03-12
Fix superfluous message when a deck is dragged to its parent

krischerven:
Currently, decks can be drag-and-dropped to their parent, which does nothing (from the user's perspective), but claims to rename the deck. This fixes that behavior.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -218,6 +218,7 @@ Thomas Graves <fate@hey.com>
 Jakub Fidler <jakub.fidler@protonmail.com>
 Valerie Enfys <val@unidentified.systems>
 Julien Chol <https://github.com/chel-ou>
+ikkz <ylei.mk@gmail.com>
 
 ********************
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -192,7 +192,7 @@ Themis Demetriades <themis100@outlook.com>
 Luke Bartholomew <lukesbart@icloud.com>
 Gregory Abrasaldo <degeemon@gmail.com>
 Taylor Obyen <162023405+taylorobyen@users.noreply.github.com>
-Kris Cherven  <krischerven@gmail.com>
+Kris Cherven <krischerven@gmail.com>
 twwn <github.com/twwn>
 Cy Pokhrel <cy@cy7.sh>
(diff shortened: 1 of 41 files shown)


dae:
The correct way to fix this would be to have reparent_decks() not include such decks in the returned change count.

dae:
I'm afraid I think this is still needlessly complicated. I suspect all that is required is

```diff
diff --git a/rslib/src/decks/reparent.rs b/rslib/src/decks/reparent.rs
index a9fcd318f..bfb3fee05 100644
--- a/rslib/src/decks/reparent.rs
+++ b/rslib/src/decks/reparent.rs
@@ -36,6 +36,9 @@ impl Collection {
         for deck in deck_ids {
             if let Some(mut deck) = self.storage.get_deck(*deck)? {
                 if let Some(new_name) = deck.name.reparented_name(target_name) {
+                    if new_name == deck.name {
+                        continue;
+                    }
                     count += 1;
                     let orig = deck.clone();
 ```

Also, I'm not sure we should be hiding the message in that case. The user has performed an action - wouldn't it be better to show them '0 decks changed' (i.e. "I received your request, but it doesn't make sense"), rather than not showing anything, which leaves the user wondering whether Anki received the user's request or not.

dae:
Tags probably need similar treatment as well.

krischerven:
> Also, I'm not sure we should be hiding the message in that case. The user has performed an action - wouldn't it be better to show them '0 decks changed' (i.e. "I received your request, but it doesn't make sense"), rather than not showing anything, which leaves the user wondering whether Anki received the user's request or not.

Yeah, I suppose that's fair.

krischerven:
@dae I implemented your solution. What needs to be changed wrt tags? It seems that I can't reparent them in the first place.

dae:
Thanks, looks good so far. Tags can be reparented in the same way is decks in the browser, by switching the sidebar to select mode, then dragging.

dae:
Sorry to drag this out, but I think this could be done a bit more efficiently by adjusting reparented_name() to return None if the old and new names match. Then the map doesn't need to be iterated again, and the changes are minimized.

krischerven:
Thanks, I didn't realize we were filtering out the `None`s there!

PR #3860, opened 2025-03-12
Refactor: Make Load Balancer Optional Throughout Codebase

YukiNagat0:
This PR refactors the Load Balancer implementation to use `Option<LoadBalancer>` across the codebase, ensuring the Load Balancer is only instantiated when enabled. Previously, the Load Balancer was constructed even when disabled, incurring unnecessary overhead.

Before this PR, only the `LoadBalancerContext` was `Optional`, so the main `LoadBalancer` object was still constructed on every `build_queues` call, even with LB turned off (using the `mw.col.load_balancer_enabled = False` command).
This PR makes all `LoadBalancer` objects optional.

**Key Changes**:
1. Optional `LoadBalancer`:
   - Replaced direct `LoadBalancer` usage with `Option<LoadBalancer>` in all relevant components (e.g., `CardQueues`, `QueueBuilder`).
   - `LoadBalancer` initialization now depends on and is in sync with the `load_balancer_enabled` configuration flag.
2. "Sync" Fix via `ClearStudyQueues`
   - Added a `ClearStudyQueues` RPC and corresponding backend method to reset study queues when toggling the Load Balancer. This prevents desynchronization between the `load_balancer_enabled` flag and the `card_queues.load_balancer` state.
   - Example scenario that can cause desync:
      1. Open Anki and disable LB (`mw.col.load_balancer_enabled = False`)
      2. Click on any deck - it generates `card_queues` with an absent `load_balancer` (`None`)
      3. Go back to the Decks browser and turn on LB (`mw.col.load_balancer_enabled = True`)
      4. Click on the same deck.
      5. As a result, we have a situation where LB is turned on, but `load_balancer` is absent in the `card_queues` because the queues weren't reloaded.
   - **Fix**: Toggling the Load Balancer now forcibly clears queues, ensuring that LB and `load_balancer` are in sync. The new `ClearStudyQueues` RPC is invoked automatically when changing `load_balancer_enabled` (via `collection.py`).

### DIFF
--- a/ftl/core/actions.ftl
+++ b/ftl/core/actions.ftl
@@ -50,6 +50,7 @@ actions-select = Select
 actions-shortcut-key = Shortcut key: { $val }
 actions-suspend-card = Suspend Card
 actions-set-due-date = Set Due Date
+actions-toggle-load-balancer = Toggle Load Balancer
 actions-grade-now = Grade Now
 actions-answer-card = Answer Card
 actions-unbury-unsuspend = Unbury/Unsuspend
--- a/proto/anki/collection.proto
+++ b/proto/anki/collection.proto
@@ -19,7 +19,7 @@ service CollectionService {
   rpc MergeUndoEntries(generic.UInt32) returns (OpChanges);
   rpc LatestProgress(generic.Empty) returns (Progress);
   rpc SetWantsAbort(generic.Empty) returns (generic.Empty);
-  rpc SetLoadBalancerEnabled(generic.Bool) returns (generic.Empty);
+  rpc SetLoadBalancerEnabled(generic.Bool) returns (OpChanges);
 }
 
 // Implicitly includes any of the above methods that are not listed in the
--- a/pylib/anki/collection.py
+++ b/pylib/anki/collection.py
@@ -982,14 +982,14 @@ class Collection(DeprecatedNamesMixin):
         )
         return self.set_config(key, value, undoable=undoable)
 
-    def _get_enable_load_balancer(self) -> bool:
+    def _get_load_balancer_enabled(self) -> bool:
         return self.get_config_bool(Config.Bool.LOAD_BALANCER_ENABLED)
 
-    def _set_enable_load_balancer(self, value: bool) -> None:
+    def _set_load_balancer_enabled(self, value: bool) -> None:
         self._backend.set_load_balancer_enabled(value)
 
     load_balancer_enabled = property(
-        fget=_get_enable_load_balancer, fset=_set_enable_load_balancer
+        fget=_get_load_balancer_enabled, fset=_set_load_balancer_enabled
     )
 
     def _get_enable_fsrs_short_term_with_steps(self) -> bool:
--- a/rslib/src/collection/service.rs
+++ b/rslib/src/collection/service.rs
@@ -3,8 +3,9 @@
 use anki_proto::generic;
 
 use crate::collection::Collection;
-use crate::config::BoolKey;
 use crate::error;
+use crate::prelude::BoolKey;
+use crate::prelude::Op;
 use crate::progress::progress_to_proto;
 
 impl crate::services::CollectionService for Collection {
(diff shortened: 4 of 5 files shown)


dae:
> Previously, the Load Balancer was constructed even when disabled, incurring unnecessary overhead.

Did you actually benchmark it? I suspect your reading of a config variable is actually orders of magnitude slower than the trivial LoadBalancer initialisation.

For the second point, the queue is the Rust code's responsibility, and I don't think we should be exposing internal access to it. You can either do some operation that implicitly clears the queues (like `col.db.execute("update col set mod=mod")`), or define a public method in the Rust layer to toggle the setting on and off.

YukiNagat0:
> I suspect your reading of a config variable is actually orders of magnitude slower than the trivial LoadBalancer initialisation.

If you're referring to the `self.get_config_bool(BoolKey::LoadBalancerEnabled)` checks preceding `load_balancer.add_card`/`remove_card` calls, this can indeed be addressed by removing the configuration checks entirely. The existing `if let Some(load_balancer) = &mut card_queues.load_balancer` check is sufficient, as the configuration variable and `card_queues.load_balancer` state are synchronized after the `self._backend.clear_study_queues()` call.

> For the second point, the queue is the Rust code's responsibility, and I don't think we should be exposing internal access to it.

To avoid exposing internal queue state, how about the next sollution:
1. Add this method in `queue/mod.rs`:
```rust
impl Collection {
    pub(crate) fn clear_load_balancer(&mut self) {
        if let Some(card_queues) = self.state.card_queues.as_mut() {
            card_queues.load_balancer = None
        }
    }
```
2. Expose this via a dedicated RPC endpoint and call it from `collection.py` in the same fashion as `self._backend.clear_study_queues`

YukiNagat0:
^ Removed the `self.get_config_bool(BoolKey::LoadBalancerEnabled)` checks and changed the `&`/&mut` to `.as_ref()`/`.as_mut()`

YukiNagat0:
Now we are cooking! After https://github.com/ankitects/anki/pull/3860/commits/785f88a1a8f2e2d01730f0f539879e9d4fb80b98 this PR **improves** performance in comparison with main:
- Execution speed is faster because `LoadBalancerEnabled` is no longer checked on every `get_scheduling_states` call (Benefits **all users** regardless of load balancer configuration)
- Reduced memory usage for users with **disabled LB** - `load_balancer` allocation is eliminated

The final step is to agree on how to synchronize the config variable and the `LoadBalancer`. I believe the `ClearStudyQueues` RPC is the most reliable/effective way to synchronize these. That said, I acknowledge this is a rather direct/"straight" approach.

dae:
I'm afraid I think exposing a ClearStudyQueues RPC is the wrong approach here. I still think this too obscure an option to expose in the GUI, but I won't object to you adding it to GetPreferences/SetPreferences if you wish to do that instead. The GUI would ignore the extra field, but it could be updated in code. If you don't want to go that route, a routine that sets the option on or off is going to be cleanest.

YukiNagat0:
> If you don't want to go that route, a routine that sets the option on or off is going to be cleanest.

I'm not quite sure I understand your suggestion. Could you clarify what you mean by a "routine that sets the option on or off"? As mentioned earlier, the only way to reconstruct the `load_balancer` object is to clear the entire card_queues object (forcing reconstruction when needed).

dae:
The Rust code should be resetting the queues as the config option is changed. You can do this either by adding it to the preferences set/get code, or by creating a backend method set_load_balancer_enabled() that sets the config var and clears the queues at the same time. The caller in Python shouldn't need to know about the queues.

YukiNagat0:
Not sure about the return type of the `set_load_balancer_enabled` and the `SetLoadBalancerEnabled`.

dae:
You can make it undoable provided the queue resetting is done by adding that new operation to requires_study_queue_rebuild(), so queues are also reset when it's undone. This would be done instead of explicitly clearing them in the routine you've added. The return type should be the same as set_config_bool().

YukiNagat0:
I considered using `crate::prelude::*`, but doing so would require removing both `use crate::collection::Collection;` and `use crate::error;`, as well as replacing all instances of `error::Result<*>` with `Result<*>`.

dae:
Thanks for your patience Yuki - looks good. For future reference, such import changes are fine, but please do them in a separate commit (in the same PR) to make things easier to review.

YukiNagat0:
Thank you for your thoughtful review and feedback on this PR! Your insights were extremely helpful.

YukiNagat0:
> But is this string exposed somewhere in the ui?

It appears when you undo the command:
1. Open the console and disable LB: `mw.col.load_balancer_enabled = False`
2. Close the console
3. Press `Ctrl+Z` for undo action
4. In the bottom-left corner the message should appear: `Toggle Load Balancer undone`

YukiNagat0:
idk,
image

image

Seems like it appears only after you rebuild the queues (e.g. clicking on any deck from main screen)

dae:
The undo menus won't be updated automatically after running a command directly in the debug console. I overlooked the translation when merging in the PR - had I noticed, I would have suggested removing it, since it's a little silly to ask translators to translate something most people won't even see.

PR #3861, opened 2025-03-12
tools/minilints: Fix contributor detection

krischerven:
Contributor detection seems to fail with email addresses that do not end in `@users.noreply.github.com` due to a discrepancy between the email in the commit history and the email in the CONTRIBUTORS file. Right now, the email I have in CONTRIBUTORS that passed previous CI pipelines fails on any new PR.

I'm not sure if this is the right solution, but it fixes CI

### DIFF
--- a/tools/minilints/src/main.rs
+++ b/tools/minilints/src/main.rs
@@ -153,9 +153,9 @@ impl LintContext {
                 .args(["rev-parse", "--show-toplevel"])
                 .output()?
                 .stdout,
-        )?;
-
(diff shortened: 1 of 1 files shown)


dae:
The solution is documented in the CONTRIBUTORS file. Your PR changes reads the contents of the file instead of git history, which is not how it's supposed to work, and will break the check for other contributors.

krischerven:
Thanks, I missed that. Didn't realize that was how I edited CONTRIBUTORS before :+1:

PR #3862, opened 2025-03-12
Loosen csv metadata parsing

### LINKED ISSUE #3853: [Bug] File headers not working correctly with .csv files
> # Steps to reproduce
> 1. Open LibreOffice Calc and paste the following into it:
> ```
> #separator:Comma		
> This is a question.	And this is the answer.	Test::CSV
> This is a question 2.	And this is the answer 2.	Test::CSV
> This is a question 3.	And this is the answer 3.	Test::CSV
> ```
> It will look like this:
> anki|452x125
> 
> 2.  Safe the file as *test.csv* and make it comma separated:
> anki|571x377
> 3. Open Anki.
> 4. Import test.csv and notice that anki thinks the file is colon separated:
> anki|484x500
> 
> # Additional info
> 
> The *test.csv* looks like this:
> ```
> #separator:Comma,,
> This is a question.,And this is the answer.,Test::CSV
> This is a question 2.,And this is the answer 2.,Test::CSV
> This is a question 3.,And this is the answer 3.,Test::CSV
> ```
> 
> Removing the trailing commas in the first line (without modifying the tags *Test::CSV*) results in this:
> ```
> #separator:Comma
> This is a question.,And this is the answer.,Test::CSV
> This is a question 2.,And this is the answer 2.,Test::CSV
> This is a question 3.,And this is the answer 3.,Test::CSV
> ```
> And:
> anki|482x500
> As you can see Anki now correctly identifies the separator and the field greys out as well.
> 
> # Expected Behavior
> Anki should be able to handle the .csv file headers correctly. I shouldn't have to manually alter the first few lines every time I want to import a .csv.
> 
> # Actual Behavior
> Anki doesn't correctly apply the file headers, without prior manual modifications.
> 
> # Version
> ```
> Anki 25.02 (8fc822a6) (src) (ao)
> Platform: Linux-6.12.12-amd64-x86_64-with-glibc2.40
> 
> ===Add-ons (active)===
> (add-on provided name [Add-on folder, installed at, version, is config changed])
> AnkiWebView Inspector ['31746032', 2023-06-27T21:26, 'None', '']
> 
> ===IDs of active AnkiWeb add-ons===
> 31746032
> 
> ===Add-ons (inactive)===
> (add-on provided name [Add-on folder, installed at, version, is config changed])
> Image Occlusion Enhanced ['1374772155', 2022-04-09T09:15, 'None', '']
> Review Heatmap ['1771074083', 2022-06-30T03:43, 'None', '']
> Study Time Stats ['1247171202', 2024-02-24T17:59, 'None', '']
> 
> ```

Originally reported here: https://forums.ankiweb.net/t/bug-file-headers-not-working-correctly-with-csv-files/56243?u=anon_0000

iamllama:
Resolves #3853

Anki's current csv delimiter heuristic works by looking for potential delimiters in ascending expected frequency (`\t`, `|`, `;`, `:`, `,`, ` `) in the first 8kb. While a wrong guess isn't necessarily a problem since the delimiter can be changed in the import options page, hierarchical tags (::) throw a wrench in this, as pointed out by @GithubAnon0000. And it's a common enough usecase that anki maybe shouldn't penalise. I imagine this was also foreseen by Rumov when he left a todo 3 years ago

This pr proposes pulling in and using the ~~csv-sniffer~~ crate, which is able to correctly deduce the delimiter for the samples in #3588 and #3853 in a multiplatform manner

EDIT: replaced [csv-sniffer](https://github.com/jblondin/csv-sniffer) with [qsv-sniffer](https://github.com/jqnatividad/qsv-sniffer), a fork that fixes some of the former's issues. However, not only is it more bloated than csv-sniffer, it still (unsurprisingly) has samples that it fails on: https://github.com/jblondin/csv-sniffer/issues/18. Not sure if this is worth pursuing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,7 +129,6 @@ dependencies = [
  "prost",
  "prost-reflect",
  "pulldown-cmark 0.9.6",
- "qsv-sniffer",
  "rand",
  "regex",
  "reqwest 0.12.8",
@@ -917,12 +916,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bytecount"
-version = "0.6.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5ce89b21cab1437276d2650d57e971f9d548a2d9037cc231abdc0562b97498ce"
-
-[[package]]
 name = "bytemuck"
 version = "1.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1160,7 +1153,7 @@ checksum = "3538270d33cc669650c4b093848450d380def10c331d38c768e34cac80576e6e"
 dependencies = [
  "serde",
  "termcolor",
- "unicode-width 0.1.14",
+ "unicode-width",
 ]
 
 [[package]]
@@ -2390,7 +2383,7 @@ version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "14dbbfd5c71d70241ecf9e6f13737f7b5ce823821063188d7e46c41d371eebd5"
 dependencies = [
- "unicode-width 0.1.14",
+ "unicode-width",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


iamllama:
I've reverted the crate addition/usage, with another suggested method: trimming potential delimiters from non-freeform metadata lines

> I wonder if we could perhaps hackily solve this by stripping off trailing commas from comment lines, as a concession to spreadsheet users?

Instead of just removing commas, it makes use of the fact that the currently supported delimiters are all non-ascii-alphanumeric. This would allow samples like `#separator:,,,`, `#separator:Pipe|||`, `#html:true,,,` and `#tags column:8,,,` to be parsed instead of being ignored, as demonstrated in the test

All the freeform metadata values (`#deck`, `#notetype`, `#tags`, `#columns`) are left alone to avoid breakage, at the expense of inconsistency with the other lines. `#match scope` and `#if match` are left alone because their constrained values can include ` ` or `+`, which unfortunately aren't compatible with this method

dae:
One problem with such selective/"magical" approaches is that they can cause confusion. A user may find a simple metadata line in a spreadsheet works, then wonder why other ones don't work correctly when they go to use them in the future. But indiscriminately stripping non-alphanumeric characters from all metadata lines is going to confuse/inconvenience some users too, so it's not an easy call.

Given that your approach is a subset of the other one and doesn't preclude us doing more later if we feel we need to, it seems like the more conservative choice. Thank you @iamllama!

PR #3863, opened 2025-03-13
Show "and others" at the end of the contributor list in the About dialog

krischerven:
Right now there are 150 contributors (most of the names in the CONTRIBUTORS file) who do not show up in the contributor list in the About dialog, compared to 132 who do. This is a little misleading, so if we don't want to show every contributor by default (for privacy reasons?), we should consider indicating that there are more contributors than just the listed ones.


dae:
I have no objection to the change in principle, but it needs to be translatable, and tests must pass.

krischerven:
@dae I've implemented the recommended changes. Do translations need to be added before merge?

ftl/qt/about.ftl
```diff
@@ -8,3 +8,4 @@ about-if-you-have-contributed-and-are = If you have contributed and are not on t
 about-version = Version { $val }
 about-visit-website = <a href='{ $val }'>Visit website</a>
 about-written-by-damien-elmes-with-patches = Written by Damien Elmes, with patches, translation,    testing and design from:<p>{ $cont }
+about-and-others = and others
```
dae, line 11:
```suggestion
# appended to the end of the contributor list in the about screen
about-and-others = and others
```

dae:
Nope, that's fine. The translation is not following best practice by joining the strings in Python, but we're already imposing an English-like list, so this is the expedient choice.

PR #3865, opened 2025-03-16
Return a copy of note type in ModelManager.get()

abdnh:
## Proposed Changes

Make ModelManager.get() return a deep copy of the fetched note type.

## Rationale

Any modifications done by one caller will be reflected to other callers due to caching. This is undocumented and can be unexpected in some cases (e.g. https://github.com/AnkiHubSoftware/ankihub_addon/pull/1090#discussion_r1959702689)

## Sidenote

@dae do you have a recent zip of all add-ons? (probably not very useful here but would be nice anyway).

### DIFF
--- a/pylib/anki/models.py
+++ b/pylib/anki/models.py
@@ -151,7 +151,10 @@ class ModelManager(DeprecatedNamesMixin):
             return None
 
     def get(self, id: NotetypeId) -> NotetypeDict | None:
-        "Get model with ID, or None."
+        """Get model with ID, or None.
+
+        This returns a reference to a cached dict. Copy the returned model before modifying it if you're not calling .update_dict() afterward.
+        """
         # deal with various legacy input types
         if id is None:
             return None
--- a/pylib/anki/models.py
+++ b/pylib/anki/models.py
@@ -165,7 +165,7 @@ class ModelManager(DeprecatedNamesMixin):
                 self._update_cache(notetype)
             except NotFoundError:
                 return None
-        return notetype
+        return copy.deepcopy(notetype)
 
     def all(self) -> list[NotetypeDict]:
         "Get all models."
--- a/pylib/anki/models.py
+++ b/pylib/anki/models.py
@@ -165,7 +165,7 @@ class ModelManager(DeprecatedNamesMixin):
                 self._update_cache(notetype)
             except NotFoundError:
                 return None
-        return copy.deepcopy(notetype)
+        return notetype
 
     def all(self) -> list[NotetypeDict]:
         "Get all models."
--- a/pylib/tests/test_decks.py
+++ b/pylib/tests/test_decks.py
@@ -48,7 +48,9 @@ def test_remove():
     deck1 = col.decks.id("deck1")
     note = col.newNote()
     note["Front"] = "1"
-    note.note_type()["did"] = deck1
+    note_type = note.note_type()
+    note_type["did"] = deck1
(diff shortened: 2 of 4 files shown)


dae:
I don't have an up-to-date list handy, sorry. I was thinking of generating one the next time we start doing some more invasive changes, but I can generate one earlier if you need it for some reason.

My original plan as part of the Rust port was to expose the backend's get_deck and get_notetype() methods, so we're returning Protobuf instead of JSON, and then encourage the ecosystem to migrate over to that, and it would also indirectly fix this issue. The mobile clients (particularly AnkiMobile) are already taking advantage of these methods, but I never got around to exposing them on the Python end, as the current dict usage is pervasive, and relies on the cache, so it would be a bit of a project.

As you're no doubt aware, this is a potentially breaking change - add-ons that are silly enough to do something like:

```
mm.get(id)["foo"] = bar
mm.save(mm.get(id))
```

will break. But it's perhaps not going to be a common issue.

The other potential downside is we have the overhead of the clone operation on every cache access. Notetypes typically aren't that large, but some users embed megabytes of HTML/JS across lots of templates.

The alternative would be to document the footgun but not fix it. I don't have strong feelings either way. Any thoughts @glutanimate?

glutanimate:
Hey, I would agree with your assessment @dae. I don't think that relying on the cache for updating the note type is something you see often these days, and grepping through my local add-on folder, the only example I found was this one in 
[Add Unique Identifier to all Notes](https://ankiweb.net/shared/info/623971671):

```python
mw.col.models.addField(self.note.model(), mw.col.models.newField(config["NoteIdFieldName"]))
mw.col.models.save(self.note.model(), updateReqs=True)
```

where the code works because both `self.note.model()` calls (each calling `models.get()`) return the same reference. Grepping for these types of patterns never perfectly captures all uses of course, but it does not seem like something that would be common.

What was common in the past was updating the note type dict queried via e.g. `models.get(...)` and then just doing `models.save()` without any arguments, but that has been a noop for a while now, which I think gives us some safety in assuming that most add-ons will properly modify and then commit their reference of the note type dict.

So from an add-on compatibility standpoint, I personally would not have any large concerns against this, and it's always nice to avoid unexpected behavior in places like this that are likely to be touched by a lot of add-ons.

On the performance side of things, I'm having a harder time reasoning about the impact without doing some local testing. But my gut feeling would be that it could be good to check, given that we must have deemed caching necessary at some point. I am not sure to what degree Anki's code still relies on the presence of the cache, but there definitely also are add-ons like AMBOSS that iterate over all note types at multiple trigger points during Anki's runtime. It might be the same for AnkiHub?

dae:
I did a brief benchmark with one of Arthur's 1.5MB notetypes (https://milchior.fr/anki/Clean_no_media_april_23.colpkg):

About 10ms with the current code, about 2700ms with the new code.

```
import time
t = time.time()
for i in range(1000):
  mw.col.models.by_name("Song")
print((time.time() - t) * 1000, "ms")
```

Most notetypes are much smaller than this, and not fetched thousands of times during the course of review. It may be a noticeable slowdown for any add-ons that are fetching notetypes in a loop and relying on the cache though.

dae:
How would you like to proceed @abdnh?

abdnh:
Thank you for your input @dae @glutanimate. I think just adding a note about this behavior is enough (will push a change). In the common case, add-ons just call .get(), do their modifications then call .update()/.update_dict(), so this is not an issue there.

PR #3867, opened 2025-03-16
Feat/simplified relearning steps logic with updated FSRS training API

L-M-Sherlock:
related PR:
- https://github.com/open-spaced-repetition/fsrs-rs/pull/297

### DIFF
--- a/rslib/src/scheduler/fsrs/params.rs
+++ b/rslib/src/scheduler/fsrs/params.rs
@@ -106,14 +106,41 @@ impl Collection {
 
         let (progress, progress_thread) = create_progress_thread()?;
         let fsrs = FSRS::new(None)?;
-        let mut params = fsrs.compute_parameters(items.clone(), Some(progress.clone()), true, Some(num_of_relearning_steps))?;
+        let mut params = fsrs.compute_parameters(
+            items.clone(),
+            Some(progress.clone()),
+            true,
+            Some(num_of_relearning_steps),
+        )?;
         progress_thread.join().ok();
         if let Ok(fsrs) = FSRS::new(Some(current_params)) {
             let current_rmse = fsrs.evaluate(items.clone(), |_| true)?.rmse_bins;
             let optimized_fsrs = FSRS::new(Some(&params))?;
             let optimized_rmse = optimized_fsrs.evaluate(items.clone(), |_| true)?.rmse_bins;
-            if num_of_relearning_steps <= 1 && current_rmse <= optimized_rmse {
-                params = current_params.to_vec();
+            if current_rmse <= optimized_rmse {
+                if num_of_relearning_steps <= 1 {
+                    params = current_params.to_vec();
+                } else {
+                    let current_fsrs = FSRS::new(Some(&current_params))?;
+                    let memory_state = MemoryState {
+                        stability: 1.0,
+                        difficulty: 1.0,
+                    };
+
+                    let s_fail = current_fsrs.next_states(Some(memory_state), 0.9, 2)?.again;
+                    let mut s_short_term = s_fail.memory;
+
+                    for _ in 0..num_of_relearning_steps {
+                        s_short_term = current_fsrs
+                            .next_states(Some(s_short_term), 0.9, 0)?
+                            .good
+                            .memory;
+                    }
+
+                    if s_short_term.stability < memory_state.stability {
+                        params = current_params.to_vec();
+                    }
+                }
             }
         }
 
--- a/rslib/src/scheduler/fsrs/params.rs
+++ b/rslib/src/scheduler/fsrs/params.rs
@@ -14,8 +14,10 @@ use anki_proto::stats::DeckEntry;
 use chrono::NaiveDate;
 use chrono::NaiveTime;
 use fsrs::CombinedProgressState;
+use fsrs::ComputeParametersInput;
 use fsrs::FSRSItem;
 use fsrs::FSRSReview;
+use fsrs::MemoryState;
 use fsrs::ModelEvaluation;
 use fsrs::FSRS;
 use itertools::Itertools;
@@ -106,12 +108,12 @@ impl Collection {
 
         let (progress, progress_thread) = create_progress_thread()?;
         let fsrs = FSRS::new(None)?;
-        let mut params = fsrs.compute_parameters(
-            items.clone(),
-            Some(progress.clone()),
-            true,
-            Some(num_of_relearning_steps),
-        )?;
+        let mut params = fsrs.compute_parameters(ComputeParametersInput {
+            train_set: items.clone(),
+            progress: Some(progress.clone()),
+            enable_short_term: true,
+            num_relearning_steps: Some(num_of_relearning_steps),
+        })?;
         progress_thread.join().ok();
         if let Ok(fsrs) = FSRS::new(Some(current_params)) {
             let current_rmse = fsrs.evaluate(items.clone(), |_| true)?.rmse_bins;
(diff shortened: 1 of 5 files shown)


rslib/src/scheduler/fsrs/params.rs
```diff
@@ -107,36 +106,15 @@ impl Collection {
...
         let fsrs = FSRS::new(None)?;
-        let mut params = fsrs.compute_parameters(items.clone(), Some(progress.clone()), true)?;
+        let mut params = fsrs.compute_parameters(items.clone(), Some(progress.clone()), true, Some(num_of_relearning_steps))?;
         progress_thread.join().ok();
         if let Ok(fsrs) = FSRS::new(Some(current_params)) {
             let current_rmse = fsrs.evaluate(items.clone(), |_| true)?.rmse_bins;
             let optimized_fsrs = FSRS::new(Some(&params))?;
             let optimized_rmse = optimized_fsrs.evaluate(items.clone(), |_| true)?.rmse_bins;
-            if current_rmse <= optimized_rmse {
+            if num_of_relearning_steps <= 1 && current_rmse <= optimized_rmse {
```
user1823, line 115: I realize that you have added this so that a user who already has problematic parameters can save the new parameters produced after this PR.

However, this also means that after they have got correct parameters and they optimize again after a few days/weeks, they can get worse parameters because the RMSE comparison is effectively disabled for them.

So, by adding `num_of_relearning_steps <= 1 &&` here, you are replacing a one-time issue with a recurrent issue. IMO, a better solution is to revert this line back and advise a user who complains about S increasing after a failure to clear their FSRS parameters field and optimize again.

user1823 (changes requested):
Looks good except for one issue.

rslib/src/scheduler/fsrs/params.rs
user1823: Considering that the issue is difficult to detect, manual intervention (as I suggested above) is likely to be difficult. So, here's another solution. (Please check the `let current_fsrs = … ` part)

```rs
if current_rmse <= optimized_rmse {
    if num_of_relearning_steps <= 1 {
        params = current_params.to_vec();
    } else {
        let current_fsrs = FSRS::new(Some(&current_params))?;
        let memory_state = MemoryState {
            stability: 1.0,
            difficulty: 1.0,
        };
        
        let s_fail = current_fsrs
            .next_states(Some(memory_state), 0.9, 2)?
            .again;
        let mut s_short_term = s_fail.memory;
        
        for _ in 0..num_of_relearning_steps {
            s_short_term = current_fsrs
                .next_states(Some(s_short_term), 0.9, 0)?
                .good
                .memory;
        }

        if s_short_term.stability < memory_state.stability {
            params = current_params.to_vec();
        }
    }
}
```

The idea is that if current parameters have better RMSE and
- R steps ≤ 1, then we simply keep the current parameters,
- if R steps > 1, then we check if the weird behaviour occurs with the current parameters. If it doesn't, we keep the current parameters. If it does, we keep the new parameters (even though they have worse RMSE).

user1823: Wait a minute...
With the updated FSRS, the weird behaviour won't occur even if the user is still using the old parameters because the parameter_clipper will clamp the parameters, right? So, we don't need this check and can simply revert to the old one, right?
```rs
if current_rmse <= optimized_rmse {
   params = current_params.to_vec();
}

Comment (L-M-Sherlock) [reply rslib/src/scheduler/fsrs/params.rs:115]: The extra clipper is only applied when num_relearning_steps > 1.

You can check the details here: https://github.com/open-spaced-repetition/fsrs-rs/pull/297/files

Comment (user1823) [reply rslib/src/scheduler/fsrs/params.rs:115]: Yes, but you didn't answer my question. Why do we need to change this line?

Comment (L-M-Sherlock) [reply rslib/src/scheduler/fsrs/params.rs:115]: Your previous guess is correct. If the `num_relearning_steps` is greater than one, the PR's design will always replace the old param with the new param to avoid sticking with problematic parameters.

Comment (user1823) [reply rslib/src/scheduler/fsrs/params.rs:115]: But the old parameters won't cause the problem after the FSRS update because the scheduler will clamp the PLS, right? (Assuming that the clamping works in the scheduler too, not just the optimizer)

Comment (L-M-Sherlock) [reply rslib/src/scheduler/fsrs/params.rs:115]: Nope. The clipper is only applied during training.

Comment (L-M-Sherlock) [reply rslib/src/scheduler/fsrs/params.rs:115]: The model doesn't know the `num_relearning_steps`. Only the optimizer knows the `num_relearning_steps`. So the extra clipper only works during the optimization.

Comment (user1823) [reply rslib/src/scheduler/fsrs/params.rs:115]: Ok then, what do you think about my suggestion in https://github.com/ankitects/anki/pull/3867#discussion_r1997570236

Comment (L-M-Sherlock) [reply rslib/src/scheduler/fsrs/params.rs]: LGTM, I updated the PR.

**Comment by Expertium** [comment]
I'm getting a bit lost, can you guys explain how FSRS will behave once this and https://github.com/open-spaced-repetition/fsrs-rs/pull/297 are merged?

**Comment by L-M-Sherlock** [comment]
If you have two or more relearning steps, FSRS optimizer will clamp w[17] and w[18] to ensure the post-lapse stability after relearning will not exceed the pre-lapse stability.

**Comment by L-M-Sherlock** [comment]
@dae, I get this error when I try to format the code:

```
$ ./ninja format
    Finished `release` profile [optimized] target(s) in 0.15s
[2/2; 1 active; 0.087s] format:rust
FAILED: /Users/jarrettye/Codes/anki/out/tests/cargo_format.fmt 
/Users/jarrettye/Codes/anki/out/rust/release/runner run --stamp=/Users/jarrettye/Codes/anki/out/tests/cargo_format.fmt --cwd=cargo/format cargo fmt  --all
`cargo metadata` exited with an error: error: failed to load manifest for workspace member `/Users/jarrettye/Codes/anki/pylib/rsbridge`

Caused by:
  failed to load manifest for dependency `anki`

Caused by:
  failed to load manifest for dependency `fsrs`

Caused by:
  failed to parse manifest at `/Users/jarrettye/Codes/open-spaced-repetition/fsrs-rs/Cargo.toml`

Caused by:
  failed to parse the `edition` key

Caused by:
  this version of Cargo is older than the `2024` edition, and only supports `2015`, `2018`, and `2021` editions.

This utility formats all bin and lib files of the current crate using rustfmt.

Usage: cargo fmt [OPTIONS] [-- <rustfmt_options>...]

Arguments:
  [rustfmt_options]...  Options passed to rustfmt

Options:
  -q, --quiet
          No output printed to stdout
  -v, --verbose
          Use verbose output
      --version
          Print rustfmt version and exit
  -p, --package <package>...
          Specify package to format
      --manifest-path <manifest-path>
          Specify path to Cargo.toml
      --message-format <message-format>
          Specify message-format: short|json|human
      --all
          Format all packages, and also their local path-based dependencies
      --check
          Run rustfmt in check mode
  -h, --help
          Print help
Failed with code Some(1): cargo fmt --all
ninja: build stopped: subcommand failed.
```

Maybe this channel is too old for `edition2024`:

https://github.com/ankitects/anki/blob/d8c83ac075c3587da853a1aeef16300ff27fc357/cargo/format/rust-toolchain.toml#L1-L4

This nightly channel works for me:

```
channel = "nightly-2025-03-20"
```

dae:
Thanks, please feel free to make that change as part of your PR.

dae:
🚀

PR #3868, opened 2025-03-16
Control autoplay order from card layout

XaviACLM:
## Changes

Modified `aqt.sound.play_clicked_audio` such that pycmd calls to play can take multiple arguments, e.g. `pycmd('play:q:2,4,17')` will play the third, fifth and eighteenth audio in a card's question side in succession.

You can connect this to a button fine, but the intended use is to control autoplay from within a script block in card html (s.t. modifications to autoplay can be embedded into decks). For this reason `aqt.reviewer.Reviewer`, `aqt.clayout.CardLayout`, and `aqt.browser.previewer.Previewer` were also modified, extending their bridge to cmd with a new command 'autoplay' - this does the same as 'play', but only if the current card (deck) has autoplay enabled.

## Rationale

The purpose of this is to allow finer control of playback from script blocks, in particular such that the new behaviour can be exported with the deck. This is a rather minoritary issue (indeed, using js at all in your cards is recommended against), but it has come up a couple of times in the forums.

I imagine this pr may be rejected on the grounds of pycmd not being intended as a public interface - if so, I would appreciate some insight on what might be a better way to present this functionality to the user. On this subject I note that it is easy to imagine some less invasive options to enable control of autoplay from the layout, e.g. a static hidden html element to describe autoplay order. This is fine and probably a decent idea, but this pr is oriented towards cards where the autoplay order may depend on state from a script block - in this setting, a message must be passed from js to python in _some_ way, albeit I'm not sure if there are better options than pycmd. If a more pared-down approach to this were preferable, it'd also be possible to add a bridge command that ended up going to some function enqueue_clicked_audio -> AVPlayer.append_tags, though this loses the ability to condition itself on whether card autoplay is enabled.

## Notes

Fine control of autoplay is strictly speaking already possible, but if you want to do this while also having the standard play buttons, as far as I'm aware the only option is to include every audio twice in the fields - once as a [sound:...], for the buttons, and once as just the audio file, to through javascript generate a series of hidden <audio> elements with their ending events chained in sequence.

The changes in CardLayout do not interact with self.have_autoplayed because pycmd autoplay is meant to override non-pycmd autoplay (and moreover will always execute _after_ the non-pycmd autoplay call).

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -218,6 +218,7 @@ Thomas Graves <fate@hey.com>
 Jakub Fidler <jakub.fidler@protonmail.com>
 Valerie Enfys <val@unidentified.systems>
 Julien Chol <https://github.com/chel-ou>
+Xavi Arnal <xavi.aclm@gmail.com>
 
 ********************
 
--- a/qt/aqt/browser/previewer.py
+++ b/qt/aqt/browser/previewer.py
@@ -157,8 +157,14 @@ class Previewer(QDialog):
         if cmd.startswith("play:"):
             card = self.card()
             assert card is not None
-
             play_clicked_audio(cmd, card)
+
+        elif cmd.startswith("autoplay:"):
+            card = self.card()
+            assert card is not None
+            if card.autoplay():
+                self._web.setPlaybackRequiresGesture(False)
+                play_clicked_audio(cmd, card)
 
     def _update_flag_and_mark_icons(self, card: Card | None) -> None:
         if card:
--- a/qt/aqt/browser/previewer.py
+++ b/qt/aqt/browser/previewer.py
@@ -162,6 +162,7 @@ class Previewer(QDialog):
         elif cmd.startswith("autoplay:"):
             card = self.card()
             assert card is not None
+            assert self._web is not None
             if card.autoplay():
                 self._web.setPlaybackRequiresGesture(False)
                 play_clicked_audio(cmd, card)
--- a/qt/aqt/clayout.py
+++ b/qt/aqt/clayout.py
@@ -435,6 +435,10 @@ class CardLayout(QDialog):
     def _on_bridge_cmd(self, cmd: str) -> Any:
         if cmd.startswith("play:"):
             play_clicked_audio(cmd, self.rendered_card)
+        if cmd.startswith("autoplay:"):
+            if self.rendered_card.autoplay():
+                self.preview_web.setPlaybackRequiresGesture(False)
+                play_clicked_audio(cmd, self.rendered_card)
 
     def note_has_empty_field(self) -> bool:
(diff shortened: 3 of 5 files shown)


abdnh:
Sounds will still be played in the normal order though when the card is rendered. Is that OK for your use case?

XaviACLM:
I understand you mean unless overriden by a script block - sure, that makes sense to me.

abdnh:
I mean sounds are played in order in _showQuestion/_showAnswer. This cannot be overriden in JS, because your script will start running after that.

XaviACLM:
The script will run immediately after the standard autoplay starts playing - and calls to pycmd-play (or autoplay) end up at AVPlayer.play_tags (as opposed to .enqueue_tags), which interrupts any ongoing playback with the new tags. You can see this is true by using pycmd('play:q:0') in a script block in any card with more than one audio on the question side; only the first audio will play (though sometimes with a half-second delay).

XaviACLM:
A related note. I didn't add any firing of the av_player_will_play_tags hook because I noticed it was never fired on bridge calls, but I'm not sure if this was intentional. Perhaps it would be a good idea to add the required calls upon both pycmd play and autoplay? Please advise and I'll commit a quick fix.

abdnh:
> only the first audio will play (though sometimes with a half-second delay).

Yes. it's a bit annoying, but not a big deal.

>  I didn't add any firing of the av_player_will_play_tags hook because I noticed it was never fired on bridge calls, but I'm not sure if this was intentional.

The hook was added as a general version of reviewer_will_play_question_sounds, so this is intended: https://github.com/ankitects/anki/commit/7c543eeb2fc2fa18b0e1dce850aeaa6b65db2129

dae:
I'm not fond of this approach I'm afraid. Updating the audio queue after the card has been rendered/sounds started is hacky, and introducing bridge commands that are only generated by third-party code (and thus not tested in our codebase) makes maintenance harder.

Being able to selectively enable autoplay is a mentioned a somewhat niche feature, but there do seem to be some users who want it, so we'll likely want an official solution at one point. A cleaner solution might be to first introduce a new tag for sounds/media that supports additional properties and not just the filename, perhaps similar to our [anki:tts] stuff. But supporting that across the various clients is a bit of work, and might be better done as part of a reworking of the (auto)play code into the JS layer, so we can do things like embed videos inside the card, and have them play sequentially. I don't see us having time to work on this in the near future unfortunately, and due to patent/codec licensing concerns, it may not be practical while we're using Qt's community Chromium builds.

In terms of what you can do now, some users who want behaviour like this have rolled their own JS-based audio playing, bypassing Anki's [sound:] tags completely (or only using them as a way to mark the media as used).

XaviACLM:
Thanks for the insight, I'll look into your suggestions, Closing.

PR #3869, opened 2025-03-16
Add-ons Dialog: disable View Config/Page/Files buttons when clicking them would not lead to useful result

beyondcompute:
#### Why

I was getting confused while working with add-on configs, files, etc. often clicking the buttons without having any add-ons selected, and getting the `please_select_a_single_addon_first` pop-up. Perhaps this could improve the usability of the dialog, if buttons were only clickable when clicking them would lead to a useful action.

#### How

1. Added `itemSelectionChanged` handler because it is also triggered on empty selection, etc.
2. Left `currentRowChanged` handler to preserve the hook’s behavior. But moved the button-disabling logic to the new handler.
3. Upon selection state change, disabling the 3 buttons, enabling “View Files” for empty selection, and enabling all 3 buttons when only 1 add-on is selected (still checking if the add-on has Page and Config).

Note that now it would theoretically be possible to remove the [`onlyOneSelected`](https://github.com/ankitects/anki/blob/14dc979e448dbf1c18c63fca64a37cce66689503/qt/aqt/addons.py#L924) function. 📝

I hope, my code is not too sloppy (and I am sorry for doing multiple things, like adding the Cmd+W shortcut, in the same PR). Big thanks to the Anki team for all their effort! 🙌

#### Before:

_Note that when multiple add-ons are selected, the state of Page and Config buttons depends on which add-on is the first among the selected ones._

https://github.com/user-attachments/assets/24fa4af3-3e43-48f7-8749-dffa25cdad02

#### After:

https://github.com/user-attachments/assets/a8f5cf07-1752-4dfc-a02c-55601b4bae32

### DIFF
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -948,7 +948,7 @@ class AddonsDialog(QDialog):
             openLink(page)
 
     def onViewFiles(self) -> None:
-        # if nothing selected, open top level folder
+        # if nothing selected, open top-level folder
         selected = self.selectedAddons()
         if not selected:
             openFolder(self.mgr.addonsFolder())
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -43,7 +43,6 @@ from aqt.utils import (
     askUser,
     disable_help_button,
     getFile,
-    is_win,
     openFolder,
     openLink,
     restoreGeom,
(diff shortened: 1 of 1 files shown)


beyondcompute:
Not sure why the CI is failing. I added the email to CONTRIBUTORS in 339bf436d.

It’s mentioned like this in the check though:

```
…
beyondcompute, at the domain users.noreply.github.com
…
…
Author beyondcompute, at the domain gmail.com NOT found in list
```

PR #3870, opened 2025-03-16
NF: sligthly optimize cards.py

Arthur-Milchior:
As AnkiDroid wants to be similar to Anki, instead of making AnkiDroid slightly less efficient, I prefer to slightly improve Anki.

AnkiDroid related PR:
https://github.com/ankidroid/Anki-Android/pull/18112.


abdnh (approved):
Doesn't matter because the note type is cached

PR #3872, opened 2025-03-21
Fix/index out of bounds exception in Card Stats Data

L-M-Sherlock:
This PR fixes an issue where viewing card statistics for some cards would cause an index out of bounds panic in the Rust backend.

This bug is introduced in:
- #3866

## Problem

The panic happens when there are review logs that don't have corresponding memory states, typically when the removed review logs appear at the end of the review history.

For example:

image

## Solution

The solution involves two key changes:

1. In `rslib/src/stats/card.rs`:
   - Added boundary checks for the memory states array
   - Properly handle cases when the removed review log appears at different positions
   - Make the memory state optional to handle cases where no appropriate state exists

2. In `ts/routes/card-info/forgetting-curve.ts`:
   - Update the filtering logic to break when encountering entries with undefined memory states

## Code Changes Explained

The main issue in the code is that when processing card review logs, some log entries might not have corresponding memory states. The original code assumed that each log entry has a corresponding memory state, but in reality, this mapping might be incomplete, especially when log entries are filtered or removed.

In the Rust part, we've added handling for several edge cases:
1. When the removed review log is at the end of the log, we use the last memory state
2. When the removed review log is at the beginning of the log, we set the memory state to None (as there's no prior state available)
3. When the removed review log is in the middle of the log, we use the memory state for the previous review entry

In the TypeScript part for forgetting curve, we updated the filtering logic to detect and handle entries that don't have defined memory states.

For example:

image

These changes eliminate the boundary check issues and enable the system to correctly handle card statistics data in all scenarios.

### DIFF
--- a/rslib/src/stats/card.rs
+++ b/rslib/src/stats/card.rs
@@ -152,14 +152,23 @@ impl Collection {
             let mut revlog_index = 0;
             for entry in revlog {
                 let mut stats_entry = stats_revlog_entry(&entry);
-                let memory_state: FsrsMemoryState =
-                    if entry.id == item.filtered_revlogs[revlog_index].id {
-                        revlog_index += 1;
-                        memory_states[revlog_index - 1].into()
-                    } else {
-                        memory_states[revlog_index].into()
-                    };
-                stats_entry.memory_state = Some(memory_state.into());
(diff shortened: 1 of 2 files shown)


dae:
Thanks for catching this!

PR #3874, opened 2025-03-21
Add help modal to TR table

GithubAnon0000:
This adds additional info as clarification, which was taken from the anki manual.

anki

Also see:
https://forums.ankiweb.net/t/let-s-remove-the-answer-buttons-chart-from-stats/56170/24?u=anon_0000

**Edit:** This has been converted into a help modal instead. See https://github.com/ankitects/anki/pull/3874#issuecomment-2748808590.


YukiNagat0:
Is it really a necessary change? As you have noted in the thread, the [True Retention Table](https://docs.ankiweb.net/stats.html?highlight=true%2520retention#true-retention-table) is already documented in the Anki Manual, and the term "True Retention" itself inherently suggests it represents _actual_/_real_ retention metrics, in contrast to the "Desired Retention". I believe adjusting the graph order (https://github.com/ankitects/anki/pull/3873) would sufficiently address the core concern about clarity without introducing redundant explanations (which would also require translators’ work for this string).

Expertium:
> only one user has expressed uncertainty about how to check their True Retention:

Well, guess it's time for another survey. I'll make it tomorrow

YukiNagat0:
> Well, guess it's time for another survey. I'll make it tomorrow

I don't think that these surveys are representing reality: they are testing only the small Anki community on reddit.

Expertium:
> they are testing only the small Anki community on reddit.

Yes, the participants are people from r/Anki: a subreddit for Anki enthusiasts, for people who care about Anki enough to comment and post. Which means an average r/Anki regular is more familiar with Anki than a completely randomly selected Anki user. Which means that a randomly selected Anki user is **more likely, not less likely** to struggle with Anki. If r/Anki users struggle, it's a safe bet that people who aren't Anki enthusiasts struggle even more.

Expertium:
Alright, I posted it on r/Anki and in the Anki Discord server. Link to the summary of results: https://docs.google.com/forms/d/1pvIxywJDzZ_-DddyYg9eA0TiVMW749MsrfK_F_2NuMA/viewanalytics

Expertium:
@GithubAnon0000 can you make a hint that shows the text when you click on the question mark?
Like this:
True Retention question mark

YukiNagat0:
Please, don't remove anything.
Also, judging by poll results I don't think that anything should be changed: rearranging the order is sufficient.

Danika-Dakika:
> Which means that a randomly selected Anki user is **more likely, not less likely** to struggle with Anki. If r/Anki regulars struggle, it's a safe bet that people who aren't Anki enthusiasts struggle even more.

I've mentioned before that I don't think these are well-placed assumptions. You can't assume that a randomly selected user understands less -- it's just as likely they understand _more_, and they aren't showing up in these help-venues because they don't have any problem understanding. Plus, there are plenty of "enthusiasts" who read about Anki every day and _still_ don't understand it very well (and give poor advice because of it).

Danika-Dakika:
I appreciate the idea of this change, but it still seems unnecessary to me. Part of the value of the Stats page is how much it is focused on images, not words -- on presenting the data, not interpreting it. We've had Answer Buttons as the "check your retention" spot for many, many years without needing to import text like that from the manual. I think the True Retention table can stand on its own too. 

I also think the text that is already there --
image
-- is enough to explain-away any calculation differences between TR and Answer Buttons. 

[As an aside on the text of the PR: 
Saying that your DR "should be close to your TR" is backwards, right? DR is fixed, so it would be the other way around. Unless the message is that you should *set* your DR to match your TR? 

But perhaps more importantly: 
Saying that your retention "should be" anything is going to lead to "why isn't my retention doing that?" questions, even more than we already get. 🤦🏽 I know @Expertium is tired of answering those them because they are now summarily dismissed with a flow-chart. I don't think those of us who are still responding to those questions can handle an uptick!]

Expertium:
image
It seems like a significant portion of users do get confused, so we should probably address that.

image
"Don't change anything" is the most popular one. Though it's worth mentioning that people seem to be much more in favor of removing Answer Buttons than True Retention.

image
Text isn't welcomed, but perhaps we could turn it into a hint, if it's technically feasible.

YukiNagat0:
Last time I checked the results they were different:
"Do you think adding this text is a good change" was 50% for No.

Expertium:
> There a huge but though: People already fail to read that info in the deck option (e.g. where you set your DR). Do you really think this is going to do any good? I must say I doubt it.

It's better than nothing, at least.

YukiNagat0:
Results were falsificated!

Expertium:
...you have to be kidding me

YukiNagat0:
No you! Why you didn't stop the poll after you posted the results???

Expertium:
Why? We don't have to close the survey or make a decision right now. We can keep looking at results as more and more people participate.

YukiNagat0:
It's wrong to continue survey after you post the results!

Expertium:
Again - why? That's like saying "refreshing the page is wrong"

YukiNagat0:
BIAS. Some participants have already seen the results before starting the survey.

Expertium:
> BIAS. Some participants have already seen the results before starting the survey.

Participants can view the summary only **after** submitting their response. Unless you mean that they saw it here, on Github.

YukiNagat0:
> Participants can view the summary only **after** submitting their response. Unless you mean that they saw it here, on Github.

THEY CAN'T PARTICIPATE AGAIN AFTER FINISHING THE SURVEY.
I am talking about people who saw you post with results and started the survey after seeing the results

YukiNagat0:
You are talking mess.
Results on moment of publishing the results were following:
- Confusion about "TR" and "answer buttons": 70% for "No"
- "Do you think adding this text is a good change" was 50% for No.

Expertium:
(I'm just going to ignore Yuki from this point onwards, this is getting ridiculous)

@GithubAnon0000 @Danika-Dakika 
1) Regarding the question mark hint: if it's too difficult to implement because of some front-end stuff, fine. Let's just not add text then, especially since people who voted "Yes" are in the minority.
>Saying that your retention "should be" anything is going to lead to "why isn't my retention doing that?" questions
Feel free to suggest better wording, though I'm not sure if it's _that_ much of a concern.

2) Regarding removing % from "Answer Buttons", I think it's a good change. It would make it harder for users to get confused. If you guys insist that it's not a good change, alright.

YukiNagat0:
You are getting ridiculous! All that you have been doing since 2024 I just making Anki worse!
Just stop. Nobody thinks that things you are proposing are worth it or even just sane.

Expertium:
If someone clicks the question mark hint icon, it can help them. If someone doesn't click the icon, the result will be the same as if the icon didn't exist at all. So it's a net positive, albeit probably not a big one.

Luc-Mcgrady:
It's because `slot="tooltip"` doesn't exist on `Graph` but instead exists within `TitledContainer`.

```svelte
<TitledContainer {title}>
    {subtitle} <!-- You would have to style this -->
    <HelpModal
        title={tr.statisticsTrueRetentionTitle()}
        url={HelpPage.DeckOptions.fsrs}
        slot="tooltip"
        {helpSections}
        on:mount={(e) => {
            modal = e.detail.modal;
            carousel = e.detail.carousel;
        }}
    />
// ...
```
image

A better option may be to use an absolute positioned div inside Graph rather than slots.

Expertium:
Nice, thank you.
Regarding wording, maybe "If you are using FSRS, your true retention is expected to be close to your desired retention"? I replaced "should" with "is expected to".

Expertium:
I mean, we can't just _not_ mention the two (DR and TR), cause then what's the point of the hint in the first place? But I don't see a way to make wording better. Maybe Danika can do it.

Luc-Mcgrady:
> Is using TitledContainer an issue? Because the styling part is easy to solve:
No real problem but it just isn't a particularly nice way of doing things in case changes are made to the `Graph` element

Expertium:
I think it's good, but maybe @Danika-Dakika wants to change the wording of the hint for TR

Luc-Mcgrady:
> Is there a consensus whether this PR (as it currently is) should be closed or kept open for review by dae et al

My sole consensus?
I believe if an actual reviewer has a problem they will make themselves heard.

If it were up to me I might make the title clickable to open the modal like the deck options. Has that been suggested? It would still run into the same implementation problems.

Danika-Dakika:
As long as DR and TR switched places, I can live with the "should" being behind a tooltip and in the manual. In my experience, folks **who read** those sorts of things tend to be a lot better at asking reasonable questions about them, so I'm weirdly not as worried about the language.

Expertium:
Let's go with "is expected to" instead then.

> If it were up to me I might make the title clickable to open the modal like the deck options.

IMO a question mark hint is harder to miss.

Luc-Mcgrady:
I've figured out a nice solution!
Add an extra slot to Graphs.svelte

```svelte
    <TitledContainer class="d-flex flex-column" {title}>
        <slot slot="tooltip" name="tooltip"></slot>
        <div class="graph d-flex flex-grow-1 flex-column justify-content-center">
            {#if subtitle}
                <div class="subtitle">{subtitle}</div>
            {/if}
            <slot />
        </div>
    </TitledContainer>
```

Then you can get away with.
```svelte
<Graph {title} {subtitle}>
    <HelpModal
        title={tr.statisticsTrueRetentionTitle()}
        url={HelpPage.DeckOptions.fsrs}
        slot="tooltip"
        {helpSections}
        on:mount={(e) => {
            modal = e.detail.modal;
            carousel = e.detail.carousel;
        }}
    />
```

> So, from what I gather this PR should be closed

I'd advise you simply add the changes you want into this PR so you can keep the conversation history and everything in one place.

Expertium:
> I'd advise you simply add the changes you want into this PR so you can keep the conversation history and everything in one place.

Yep.

iamllama:
> I'm not sure what on:mount did but the modal works fine without it

It's for when multiple subheadings open the same modal, e.g in deck options. Not necessary right now, but I recommend you add it back to keep things consistent with everywhere else `HelpModal` is currently used

Luc-Mcgrady (commented):
I can provide a full diff if this doesn't work?

Never mind it has the hover cursor on all titles :sweat_smile:

ts/vite.config.ts.timestamp-1745185171287-f031d0eb609c1.mjs
Luc-Mcgrady, line 1: You accidentally committed this.

Luc-Mcgrady:
```
git rm ts/vite.config.ts.timestamp-1745185171287-f031d0eb609c1.mjs
```

Luc-Mcgrady: Yep it's fine.

dae:
Looks good - thank you! Regarding your issues with ./check, removing the .venv folder you've placed there should fix it.

dae:
Probably created with a tool like `uv`.

PR #3877, opened 2025-03-22
Fix panic when clearing today limits on the day collection was made

iamllama:
Fixes #3875 

See https://github.com/ankitects/anki/issues/3875#issuecomment-2745211077 for context on the problem

This pr proposes to fix the panic by extending the hack to account for the edge case when the collection was created on the day itself, rather than changing the deck schema to account for this one day 

No new inconsistency is introduced, and it having gone unnoticed till now by a dev suggests that users are unlikely to complain about the last today limit not being preserved on day 0 🤔

PR #3878, opened 2025-03-23
Fix deck day limits incorrectly being carried over when importing

iamllama:
For [ergonomics](https://github.com/ankitects/anki/pull/1955#discussion_r922768204), day limits aren't cleared* when not used. Instead, a hack is used whereby its set as if it were in the past and thus not applicable on the current day

This is problematic when importing because the current day depends on the collection's age, and so an imported deck's "past" day limit might fall within the importing collection's "future" if the importing col is newer

This means that the imported deck's today limit might turn itself on in the future, even if it was considered to be a past value when the deck was exported. Or, more commonly, when importing a deck that has a valid today limit in the orig col, it isn't applied because the importing col has a different today.

This pr proposes to fix that by ~~shifting~~ correcting the day limit dates ~~back~~ accordingly** when importing, or removing them if the situation described in #3877 would occur. ~~Sorry in advance if i've screwed up the math, it's a bit confusing~~ EDIT: used a simpler approach that should also be more consistent

*changed in #3877, but the problem described here remains
**analogous to ~~shifting back~~ correcting imported cards' due dates

### DIFF
--- a/rslib/src/decks/mod.rs
+++ b/rslib/src/decks/mod.rs
@@ -20,6 +20,7 @@ use std::sync::Arc;
 pub use anki_proto::decks::deck::filtered::search_term::Order as FilteredSearchOrder;
 pub use anki_proto::decks::deck::filtered::SearchTerm as FilteredSearchTerm;
 pub use anki_proto::decks::deck::kind_container::Kind as DeckKind;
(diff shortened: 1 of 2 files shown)


dae:
Thanks @iamllama! The added test is appreciated too. :-)

PR #3879, opened 2025-03-24
Cloze button get disabled outside of cloze field

Arthur-Milchior:
If the user tries to add a cloze in a field that is not a cloze type, the user gets an alert message.

I could have removed the buttons or disabled them, as it's done when the note type is not a cloze deletion, but I think it's better to avoid changes in the buttons. And anyway, the alert would still be useful if the user tried to use the shortcut to add the cloze.

In order to do this, I added a back-end method (that I expect we may reuse in ankidroid) to get the index of the fields used in cloze. This set is sent to the note editor, which propagates it where needed.

In mathjax, the cloze symbol is removed when the selected field is not a cloze field

I would appreciate help, as the alert message is not properly displayed, instead it shows

> missing key: adding-cloze-outside-cloze-field

### DIFF
--- a/proto/anki/notetypes.proto
+++ b/proto/anki/notetypes.proto
@@ -33,6 +33,7 @@ service NotetypesService {
   rpc GetFieldNames(NotetypeId) returns (generic.StringList);
   rpc RestoreNotetypeToStock(RestoreNotetypeToStockRequest)
       returns (collection.OpChanges);
+  rpc GetClozeFieldOrds(NotetypeId) returns (GetClozeFieldOrdsResponse);
 }
 
 // Implicitly includes any of the above methods that are not listed in the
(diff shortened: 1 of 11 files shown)


Arthur-Milchior:
I also realized that it was buggy after a note type change. This is repaired. I now disabled the buttons when the focus is in a field that is not a focus field. To be consistent with the button, I also ensured the shortcut does nothing, instead of displaying an alert.

I'm not claiming that this is adding a ton of value to be clear. Especially not for power users. But I introduced a friend to anki and saw him trying various things, and hence trying to add a cloze while he was in "back extra". I found strange that there was no warning sooner to let them know it can't be done, and wanted to correct that. I expect he was not the only one who didn't understand the limitation at first.

And while I admit that the message already appear to tell the user that cloze can't be used in back extra, I think that disabling the button is good practice. Not only does it give feedback to the new user that the available actions are not the same in all fields, but it ensures that, while trying the app, they won't try to add a cloze where it should not be.

I first thought you were referring to the warning that appears if you try to add the note. Which I found to arrive quite late. 
I missed the warning that arrive very soon after entering the cloze, because of #3909.

dae:
Please fix the tests and then I'll merge it in.

Arthur-Milchior:
Done

dae:
Thanks Arthur.

PR #3880, opened 2025-03-25
Comment for translators for statistics-total

sommerluk:
I hope statistics-total is not used in another context with a different meaning?

### DIFF
--- a/ftl/core/statistics.ftl
+++ b/ftl/core/statistics.ftl
@@ -254,9 +254,9 @@ statistics-elapsed-time-years = { $amount }y
 ##
 
 statistics-average-for-days-studied = Average for days studied
-# This does NOT refer to the total number of all existing cards. Instead, it
-# specifies the number of cards within the time frame currently selected by
-# the user via the radio buttons positioned above the diagram.
+# This term is used in a variety of contexts to refers to the total amount of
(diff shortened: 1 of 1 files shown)


dae:
Let's not hope, let's check.

```
% rg statisticsTotal
ts/routes/graphs/reviews.ts
397:        { label: tr.statisticsTotal(), value: totalString },

ts/routes/graphs/true-retention.ts
101:            return tr.statisticsTotal();

ts/routes/graphs/added.ts
91:            label: tr.statisticsTotal(),

ts/routes/graphs/future-due.ts
131:            label: tr.statisticsTotal(),
dae@dair:Work/desktop/anki% rg statistics_total
zsh: exit 1     rg statistics_total
```

The true retention graph is a bit different. How about something like the following wording instead?

# This is used in a variety of graphs to refer to the total amount of something (cards, mature cards, etc) for a given period.

If translators need to use different wording for different contexts, we'll need to split it into multiple strings.

dae:
Sorry, I must have had an old branch checked out when I did the check. 🤦 
It looks like statisticsTotal is also used in the FSRS simulator in the deck options:

```
ts/routes/graphs/simulator.ts
232:                ${tr.statisticsTotal()}: ${formatY(total)}`,
```

sommerluk:
Okay. I've adapted the wording.

PR #3881, opened 2025-03-25
Move smooth button above simulate graph.

Luc-Mcgrady:
@Expertium wanted to make it clearer to the user that the smooth button had no effect on the simulation.

I am personally ambivalent to this change, especially if it runs into issues with scrolling on mac.

image

### DIFF
--- a/ts/routes/deck-options/SimulatorModal.svelte
+++ b/ts/routes/deck-options/SimulatorModal.svelte
@@ -292,14 +292,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                         </SettingTitle>
                     </SwitchRow>
 
-                    <SwitchRow bind:value={smooth} defaultValue={true}>
-                        <SettingTitle
-                            on:click={() => openHelpModal("simulateFsrsReview")}
-                        >
-                            {"Smooth Graph"}
-                        </SettingTitle>
-                    </SwitchRow>
-
                     <SwitchRow
                         bind:value={suspendLeeches}
                         defaultValue={$config.leechAction ==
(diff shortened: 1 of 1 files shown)


YukiNagat0:
Please read the Discord conversation for full context: https://discord.com/channels/368267295601983490/1282005522513530952/1354043532352159817
No one except Expertium considers this change necessary.

My opinion:
- `Advanced settings` is the appropriate location for the `Smooth graph` toggle
- The purpose of `Smooth graph` is already clear; users are unlikely to assume it affects actual deck settings/options
- After this PR, the area containing `Simulate`, `Clear last simulation`, `Save to Preset options` and `Smooth graph` becomes overly cluttered with buttons. This increases the risk of misclicks.

I believe no changes are needed.

Ultimately, the decision is up to Dae: if there’s concern about potential misunderstandings, I suggest only renaming the toggle to `Smooth the simulated graph` while keeping it in `Advanced settings`.

Expertium:
> Probably because I don't even understand why the graph isn't a simple line but looks like this un-smoothed

If you have different days configured differently in Easy Days, that explains it. If all days are the same...idk

YukiNagat0:
> One of these is not like the others

It doesn't justify moving it from `Advanced settings`: `Advanced settings` are not meant to be tied to deck options. It is just a place where all the other settings are stored.

The main point it that moving it from `Advanced settings` will clutter the UI - I don't want to see this toggle when all groups are minimized.
(Also, in the current state of this PR, it will also cause misclicks, because toggle is on the left)

YukiNagat0:
> Maybe there is some sort of string that we could use?
> 
> Something like "Smooth Graph (doesn't affect simulation)".

`Smooth the simulated graph` or `Smooth the resulted graph`?

dae:
I agree with the others who think the advanced section is the most appropriate place for this. I don't have strong feelings on the wording, but think it shouldn't be made more wordy than any of the existing options.

Luc-Mcgrady:
@iamllama https://github.com/ankitects/anki/pull/3882#issuecomment-2753047515
>  it concerns the "smooth graph" change as well

image

Centring it works fine :+1: .

dae:
Are you planning to close this, revise it, or are you waiting for something from me?

Luc-Mcgrady:
Oops 😄.

Expertium:
@dae if you think this is a bad change, I highly disagree.

> All these settings except for Smooth Graph affect scheduling and are real settings that you can find in deck options. So grouping Smooth Graph - which only affects the plotting - with real settings seems like a bad UI to me.

User's shouldn't have to play "pick the odd one out"

PR #3882, opened 2025-03-25
Center align rows in FSRS simulator

GithubAnon0000:
I confused two rows with each other in the simulator because the option name and option toggle / textfields are quite some distance apart on my screen. My eyes followed the wrong imaginary line between name and toggle.

This PR centers the row strings. That way, the option name and option toggle / textfields are better aligned. The change is small but helps me a tiny bit.

Before:
anki

After:
anki

Yes, one can barely see the difference. I'd prefer having a better distinction between rows or some way to make it more obvious which row elements belong to each other. But I don't know how to achieve that, while still having some aesthetics.

### DIFF
--- a/ts/lib/components/Row.svelte
+++ b/ts/lib/components/Row.svelte
@@ -17,5 +17,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         flex-flow: row wrap;
         align-content: stretch;
         padding: var(--gutter-block, 0) 0;
+        align-items: center;
     }
 </style>


iamllama:
My two cents: at this point it might be better to seriously reconsider the simulator ui as a whole in the forum rather than splitting the discussion on each individual aspect of the simulator over individual prs

pinging @Luc-Mcgrady because it concerns the "smooth graph" change as well

dae:
Agreed that this might be better discussed in more general terms, but for now this seems like an uncontentious change, if people were to notice at all.

PR #3885, opened 2025-03-26
Fix #3335: Implement shared menu bar for macOS

### LINKED ISSUE #3335: Make Edit>Undo available on macOS in the add cards screen
Due to https://github.com/ankitects/anki/pull/1486, users need to navigate to the main window before they can access Undo via the Edit menu. Perhaps we should add an Edit menu to the add cards screen as well, with undo/redo as an option.

Originally reported on https://forums.ankiweb.net/t/make-undo-button-and-or-ctrl-z-universally-available-especially-lacking-in-card-creation/47665

ynonp:
This commit addresses issue #3335 by implementing a shared menu bar for macOS:

- Create a shared menu bar in the main window that can be used by child windows
- Change Add Cards dialog to use the shared menu bar on macOS

As menus in Qt also control keyboard shortcuts we can also use cmd+z now to undo from the add cards dialog

### DIFF
--- a/qt/aqt/addcards.py
+++ b/qt/aqt/addcards.py
@@ -57,7 +57,9 @@ class AddCards(QMainWindow):
         gui_hooks.operation_did_execute.append(self.on_operation_did_execute)
         restoreGeom(self, "add")
         gui_hooks.add_cards_did_init(self)
-        if not is_mac:
+        if is_mac:
+            self.setMenuBar(mw.shared_menubar)
+        else:
             self.setMenuBar(None)
         self.show()
 
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -16,6 +16,8 @@ from collections.abc import Callable, Sequence
 from concurrent.futures import Future
 from typing import Any, Literal, TypeVar, cast
 
+import PyQt6.QtWidgets as QtWidgets
+
 import anki
 import anki.cards
 import anki.sound
@@ -92,7 +94,7 @@ from aqt.utils import (
     showWarning,
     tooltip,
     tr,
-)
+    )
 from aqt.webview import AnkiWebView, AnkiWebViewKind
 
 install_pylib_legacy()
@@ -983,6 +985,24 @@ title="{}" {}>{}</button>""".format(
                 webview.force_load_hack()
 
         gui_hooks.card_review_webview_did_init(self.web, AnkiWebViewKind.MAIN)
+
+        if is_mac:
+            designed_menubar = self.menuBar()
+            self.shared_menubar = QtWidgets.QMenuBar(None)
+
+            # Copy all menus and actions
+            for action in designed_menubar.actions():
+                if action.menu():  # If it's a menu
+                    new_menu = self.shared_menubar.addMenu(action.text())
+                    # Copy all actions in this menu
+                    for sub_action in action.menu().actions():
+                        new_menu.addAction(sub_action)
+                else:  # If it's a direct action
+                    self.shared_menubar.addAction(action)
+
+            self.setMenuBar(self.shared_menubar)
+
+        self.setMenuBar(self.shared_menubar)
 
     def closeAllWindows(self, onsuccess: Callable) -> None:
         aqt.dialogs.closeAll(onsuccess)


dae:
Thank you for taking the time to contribute! I'm a bit concerned about possible regressions here, particularly around add-ons. Any thoughts @glutanimate?

It also appears to be broken at the moment - the add window won't open again after it's been opened once.

ynonp:
Hi and thanks for the feedback - I fixed the "Add" twice problem that was an oversight on my part. I don't have a Windows or Linux machines to verify but at least on my Mac I can now close and re-open the add cards window without a problem.

IMHO as a Mac user it feels more natural to have the same menubar throughout the application. This is also the recommendation in the Qt docs - see this paragraph from the QMenuBar docs:
https://doc.qt.io/qt-6/qmenubar.html#qmenubar-as-a-global-menu-bar

Alternatively we can set the parent of AddCards window to main and not set a menu there and then Qt uses automatic menu inheritance rules to show the main window's menu even when switching to the secondary window. This is described in the list here:
https://doc.qt.io/qt-6/macos-issues.html#menu-bar

I'm not sure about creating a separate menu just with the undo/redo action because it means to clone the actions objects and then you need to update their enabled status in both menus (i.e. I think we'll need to implement something like `update_undo_actions` in `addcards.py` also and call it whenever the main's update_undo_actions is called). Or did you have something else in mind?

glutanimate:
Hey, thanks for the contribution. I can see how it would be nice to conform to OS-specific design guidelines more. I do have some concerns about this and #3335 in general, however:

#### From an add-on compatibility standpoint

I think a change like this could affect add-ons in two main ways:

- **Keyboard shortcut interference**: Because of built-in assignments, and assignments in other add-ons, there already is a limited shortcut assignment space for add-ons in each Anki context. By bringing the mw actions into the AddCards context, that space would narrow even further and likely produce a number of add-on/add-on and add-on/built-in conflicts.
- **UX confusion on action context**: Existing actions registered by add-ons were designed with the main window context in mind, so there is likely going to be a number of cases where - from a UX standpoint - what an action does and in what context it operates can be ambiguous if they appear across contexts other than the main window.
  + In the current implementation, the PR actually does not include add-on actions in the AddCards context (due to toolbar cloning happening before add-on init), but I'm not sure if this is intentional. If it is, I think instead of the UX issue above we would have a different one on our hands, where users are confused why they cannot access an add-on provided action, depending on which window is in focus.
    Screenshot 2025-04-11 at 18 32 43 Screenshot 2025-04-11 at 18 32 35


#### UX of sharing actions in general

I think it's hard to avoid some inconsistencies with shared actions, e.g. user questions like: Why does keyboard shortcut X work in this context, but not Y (e.g. "Undo" vs. "Create filtered deck")? Why does action A change something in the main window, but not in AddCards (e.g. "Reset Zoom")? Why does pressing this menu item draw up a window in front of AddCards while this other menu item throws me into a window below it (e.g. "Study deck" vs "Create filtered deck"?


#### UX of sharing "undo" in particular

This is more a point against #3335 in general, not so much your PR, but personally I'm not that fond of the shared undo stack we currently have and would be worried about us extending it into even more contexts. I just think that it sets users up for accidents too easily. I know that personally I've found myself tripping over accidentally deleting a note or undoing an edit way too many times when going through reviews, and if we extend the undo stack to the editor, I would imagine that this would create scenarios like this even more frequently: Consider that, in the text editing context, we already have an undo stack that would continue to live separately from the shared undo stack. So depending on whether users are currently in a field or clicked outside of it, they could e.g. either end up deleting a typo or deleting the last note they added / last review they did.

With undo systems, I think it's important that the undo/redo commands correspond with whatever mental model a user has of their recent interactions with an application. In a multi-window/multi-context application like Anki, I would expect that mental model to be context-specific rather than global. So if a user performs an action in one context (e.g. editing a note in the card browser window), and then firing an undo from a different context (e.g. reviewing cards) undoes that action, IMHO that does not feel like expected behavior, in particular when those actions can have entirely different "weights" attached to them (e.g. undoing a single review vs. undoing a batch modification to the collection).

I'm not sure if this is something that came up in the discussion about the undo changes back in the day. I know that the rework was a nice improvement otherwise and made more actions undoable, but I really believe that the global undo stack we ended up with is not ideal and I actually see its behavior as a significant gotcha for new users.

I had a surprisingly hard time finding discussions about global vs local undo stacks from a UX standpoint, but I found the answers [this question on UX SE](https://ux.stackexchange.com/questions/14058) to make some good points for instance.

dae:
> IMHO as a Mac user it feels more natural to have the same menubar throughout the application. 

I appreciate this, but we're a cross platform app with a lot of legacy baggage :-) 

> Alternatively we can set the parent of AddCards window to main and not set a menu there

IIRC, we avoid giving the add cards screen a parent so it's treated like a top-level window, which alters the way it appears in e.g. the taskbar, as users expressed a preference for this.

> I'm not sure about creating a separate menu just with the undo/redo action because it means to clone the actions objects

Not proposing we do it dynamically - just manually add the entries to addcards's existing menu, and hook them up like the browser window does.

> So if a user performs an action in one context (e.g. editing a note in the card browser window), and then fires an undo from a different context (e.g. reviewing cards), IMHO undoing the previous-context action does not feel like expected behavior

Multiple undo stacks means dealing with merge conflicts, which is complicated, as you have potentially incompatible changes (e.g. user deletes a field from a notetype, then tries to undo edits to a note).

Issue #3189, opened 2024-05-05
Nested clozes cause exponentially growing html


naipofo:
Nested clozes generate some weird html, cause they render the inner cloze and then escape it to put in the data-cloze attribute. This results in weird html in attribute:
```html
<span class="cloze" data-cloze="<span class=&quot;cloze&quot; data-cloze=&quot;&quot; data-ordinal=&quot;1&quot;>[...]</span>" data-ordinal="1">[...]</span>
```
and that html contains nested html escapes that grow exponentially. This huge html chunk slowes down `extract_av_tags` to a crawl and crashes anki.

This starts to cause crashes with ~11 nested clozes. Some shared decks on ankiweb have those and a user might make them by mistake, so there should be some kind of safeguard.

Originally reported on https://forums.ankiweb.net/t/memory-leak-when-search-specific-kanji-japanese-hieroglyph/44394/2

dae:
Thanks for digging into this! It sounds like we may need to both limit the levels of nesting, and also place bounds on extract_av_tags if possible.

Danika-Dakika:
@abdnh @dae Was #3264 intended to break existing clozes? This is [causing some trouble for users](https://forums.ankiweb.net/t/nested-clozes-doesnt-work-for-more-than-3-nests/52661?u=danika_dakika).

abdnh:
Yes, the change affects existing clozes.

aiaiads:
This really has causes some problems for me. 3 nests limit is too low

aiaiads:
I can't believe you changed this feature so simply without any big discussion. I use anki mostly with the fill in the blanks template, so I have a lot of cards that are affected.

PR #3886, opened 2025-03-26
Improve performance of card rendering parser

iamllama:
> This huge html chunk slowes down extract_av_tags to a crawl and crashes anki.
#3189

Looked into why this was happening, seems to be the `recognize(manyX(pair(not(<parser>)), anychar)))` pattern, which calls `<parser>` on every single character. This was fine in 2021, but after #1968 and #2141, cards with deeply nested clozes commonly result in extremely bloated* inputs that it can't handle fast enough

The speedup is mainly realised by making use of memchr to skip ahead to occurences of `[`, a prefix of tags. Also, when nested within `recognize`, `manyX` needlessly allocates a vec, populates it per character and then throws it away. This pr removes that from `text_node` and `closing_parser` as well

Benchmarking with `rslib/bench.sh`, using the [deeply nested malformed cloze example](https://forums.ankiweb.net/t/memory-leak-when-search-specific-kanji-japanese-hieroglyph/44394/2)**, before and after:
```:
anki_tag_parse          time:   [85.267 µs 85.434 µs 85.645 µs]

anki_tag_parse          time:   [345.20 ns 346.54 ns 348.72 ns]
                        change: [-99.596% -99.594% -99.591%] (p = 0.00 < 0.05)
                        Performance has improved.
```
Where this appears to shine: was able to raise the nested cloze limit from 8 to 10 and open up the preview of the abovementioned malformed example without anki complaining of a blocked main thread (pre-pr: ~3.5s stall on my machine, a >94% speedup)

I can't see any breakage resulting from this. Only the internal representation was changed, i.e. consecutive text nodes are now possible, which shouldn't result in changes to the syntax this parser admits. Tested with decks that use sound and tts tags

*the mentioned malformed example gets blown up to 3mb when passed to the parser
**before (nested) cloze metadata gets added to it

### DIFF
--- a/rslib/src/card_rendering/mod.rs
+++ b/rslib/src/card_rendering/mod.rs
@@ -33,24 +33,24 @@ pub fn prettify_av_tags<S: Into<String> + AsRef<str>>(txt: S) -> String {
 }
 
 /// Parse `txt` into [CardNodes] and return the result,
-/// or [None] if it is only a text node.
+/// or [None] if it only contains text nodes.
 fn nodes_or_text_only(txt: &str) -> Option<CardNodes> {
     let nodes = CardNodes::parse(txt);
-    match nodes.0[..] {
-        [Node::Text(_)] => None,
-        _ => Some(nodes),
-    }
+    (!nodes.text_only).then_some(nodes)
 }
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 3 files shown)


dae:
Great work in digging into this! I can't think of any breakages either.

PR #3894, opened 2025-03-31
style: use consistent input styles in deck options

ikkz:
### Before:
截屏2025-03-31 21 06 43

### After
截屏2025-03-31 21 05 55
截屏2025-03-31 21 06 09


dae:
Thank you!

PR #3898, opened 2025-04-05
Changed minimum value of DR

Expertium:
And also of historical retention, it was too limited for some reason.

In [FSRS-6](https://github.com/open-spaced-repetition/fsrs-optimizer/pull/169/commits/c1f2402c2dc8952da4ff8d33a07319de571755a4) the curve will become flatter. Again. This is the last time this happens, we promise 😅.
Anyway, flatter curve means longer intervals. So in order to make sure that users don't run away from FSRS after being scared by long intervals, I think it's worth it to increase minimum DR. With the new curve, the intervals at DR=75.5% roughly correspond to intervals at DR=70% with the old curve, so I raised the minimum to 75%. This way the intervals in the new version of Anki with FSRS-6 won't be longer than before.

Slightly unrelated, but Dae, I want you to take a look at this idea regarding the UI and how we convey what DR does: https://forums.ankiweb.net/t/desired-retention-ui-overhaul/57678/33?u=expertium

And the proposal to change _default_ DR to 95%: https://forums.ankiweb.net/t/thoughts-on-changing-default-desired-retention-to-95/58273?u=expertium

@L-M-Sherlock

### DIFF
--- a/rslib/src/deckconfig/mod.rs
+++ b/rslib/src/deckconfig/mod.rs
@@ -293,14 +293,14 @@ pub(crate) fn ensure_deck_config_values_valid(config: &mut DeckConfigInner) {
     ensure_f32_valid(
         &mut config.desired_retention,
         default.desired_retention,
-        0.7,
+        0.75,
         0.99,
     );
     ensure_f32_valid(
         &mut config.historical_retention,
         default.historical_retention,
-        0.7,
-        0.97,
+        0.5,
+        0.99,
     )
 }
 
--- a/ts/routes/deck-options/FsrsOptions.svelte
+++ b/ts/routes/deck-options/FsrsOptions.svelte
@@ -103,9 +103,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     }
 
     function getRetentionWarningClass(retention: number): string {
-        if (retention < 0.7 || retention > 0.97) {
+        if (retention < 0.75 || retention > 0.97) {
             return "alert-danger";
-        } else if (retention < 0.8 || retention > 0.95) {
+        } else if (retention < 0.85 || retention > 0.95) {
             return "alert-warning";
         } else {
             return "alert-info";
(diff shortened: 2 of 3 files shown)


Expertium:
No, it's not a flat +x% increase
You can compare the forgetting curves here: https://www.desmos.com/calculator/8tauss2btt
Above 90% not much will change
image
S=100, green is FSRS-6, purple is FSRS-4.5 and 5. Y axis is probability of recall, X axis is time
The difference is mostly in how the curve behaves when t>>S, in other words, when the elapsed time is much greater than memory stability

rslib/src/deckconfig/mod.rs
```diff
@@ -293,14 +293,14 @@ pub(crate) fn ensure_deck_config_values_valid(config: &mut DeckConfigInner) {
...
-        0.7,
+        0.75,
         0.99,
     );
     ensure_f32_valid(
         &mut config.historical_retention,
         default.historical_retention,
-        0.7,
-        0.97,
+        0.5,
```
Luc-Mcgrady, line 302:
```suggestion
        0.75,
```
?

L-M-Sherlock:
I don't recommend change the minimum value of DR. It's enough to show a red danger alert. The problem is, the red danger alert would never appear because the retention is always greater or equal to 0.7.

user1823:
> With the new curve, the intervals at DR=75.5% roughly correspond to intervals at DR=70% with the old curve, so I raised the minimum to 75%. This way the intervals in the new version of Anki with FSRS-6 won't be longer than before.

This is a misinterpretation.

If the true retention was 70% earlier, then the new curve would produce the same intervals if the DR is set to 70% because the new version of FSRS will calculate a smaller stability for the same review history.

Expertium:
> I don't recommend change the minimum value of DR. It's enough to show a red danger alert. The problem is, the red danger alert would never appear because the retention is always greater or equal to 0.7.

I changed it though.
image

user1823:
> I was analyzing how curves behave when S is the same.

That's not the proper way to compare them because S is not fixed but it depends upon the algorithm.

FSRS6 can even produce smaller intervals than FSRS5. Let's say that a user's DR was 70% but their TR with FSRS5 was 66%. Now, because FSRS6 is more accurate, it would produce intervals that result in a TR that matches the DR more closely. And how do you increase the TR? Yeah, you guessed it right, by shortening the intervals. So, FSRS6 would produce smaller intervals than FSRS5 for this user.

PR #3406, opened 2024-09-07
Fix graph averages

abdnh:
Closes #3269

PR #3900, opened 2025-04-06
Replace some means in Stats with medians

user1823:
Median is better suited than mean for describing a skewed distribution.

Some other averages were not replaced because mean is more suitable for describing populations when the purpose is to estimate a population total.

@Expertium, you might be interested.

### DIFF
--- a/ts/routes/graphs/ease.ts
+++ b/ts/routes/graphs/ease.ts
@@ -106,7 +106,7 @@ export function prepareData(
     const xTickFormat = (num: number): string => localizedNumber(num, 0) + "%";
     const tableData = [
         {
-            label: tr.statisticsMedianEase(),
+            label: tr.statisticsAverageEase(),
             value: xTickFormat(data.average),
         },
     ];
--- a/ts/routes/graphs/ease.ts
+++ b/ts/routes/graphs/ease.ts
@@ -106,8 +106,8 @@ export function prepareData(
     const xTickFormat = (num: number): string => localizedNumber(num, 0) + "%";
     const tableData = [
         {
-            label: tr.statisticsAverageEase(),
-            value: xTickFormat(sum(Array.from(allEases.entries()).map(([k, v]) => (k + 2.5) * v)) / total),
+            label: tr.statisticsMedianEase(),
+            value: xTickFormat(data.average),
         },
     ];
 
--- a/ts/routes/graphs/ease.ts
+++ b/ts/routes/graphs/ease.ts
@@ -106,7 +106,7 @@ export function prepareData(
     const xTickFormat = (num: number): string => localizedNumber(num, 0) + "%";
     const tableData = [
         {
-            label: tr.statisticsAverageEase(),
+            label: tr.statisticsMedianEase(),
             value: xTickFormat(data.average),
         },
     ];
--- a/ftl/core/statistics.ftl
+++ b/ftl/core/statistics.ftl
@@ -148,7 +148,7 @@ statistics-card-ease-title = Card Ease
 statistics-card-difficulty-title = Card Difficulty
 statistics-card-stability-title = Card Stability
 statistics-card-stability-subtitle = The delay at which retrievability falls to 90%.
-statistics-average-stability = Average stability
+statistics-median-stability = Median stability
 statistics-card-retrievability-title = Card Retrievability
 statistics-card-ease-subtitle = The lower the ease, the more frequently a card will appear.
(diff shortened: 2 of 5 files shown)


ts/routes/graphs/ease.ts
```diff
@@ -106,8 +106,8 @@ export function prepareData(
     const xTickFormat = (num: number): string => localizedNumber(num, 0) + "%";
     const tableData = [
         {
-            label: tr.statisticsAverageEase(),
-            value: xTickFormat(sum(Array.from(allEases.entries()).map(([k, v]) => (k + 2.5) * v)) / total),
```
user1823, line 110: This was most likely omitted by mistake in https://github.com/ankitects/anki/pull/3406 by @abdnh.

Expertium:
I think having means for some stats and medians for other stats is confusing. And means are more intuitive to most people and less likely to make people think "Oh, it's some complicated statistical mumbo-jumbo". We could display both, but that would be too much information.
I suppose I'm more against this change than in favor of it.

user1823:
> And means are more intuitive to most people and less likely to make people think "Oh, it's some complicated statistical mumbo-jumbo". 

If that's the only argument against the change, we can keep the current label (average) but display the median value instead. Even if the average user can't understand the difference between a mean and a median, we should try to make sure that the Stats page allows them to get a better estimate of the "typical value".

<hr>

To maintainers: The CI failure is not related to my changes.

Expertium:
I endorse being dishonest about stats even less than I endorse having means in one place and medians in another, or having stats that only 1% of users understand.
I think there just isn't a good enough reason to change the way things are right now.
1) Having means for some stats and medians for other stats is confusing
2) Having both means and medians is too much information that most users are unlikely to find useful
3) Calling medians means is not only dishonest, but could lead to *even more* confusion if someone actually decides to check whether the math adds up

user1823:
> Calling medians means is not only dishonest

I am not proposing to call it a mean. Though the term "average" most commonly refers to a mean, it can also refer to a median.

From https://en.m.wikipedia.org/wiki/Average:

> Depending on the context, the most representative statistic to be taken as the average might be another measure of central tendency, such as the mid-range, median, mode or geometric mean. For example, the average personal income is often given as the median – the number below which are 50% of personal incomes and above which are 50% of personal incomes – because the mean would be higher by including personal incomes from a few billionaires.

Expertium:
Fine

Expertium:
Actually, the more I think about it, the more I feel like it doesn't solve anything

1) As I said, using the arithmetic mean in some places and the median in other places is confusing.
2) I struggle to imagine a situation where this change would benefit the average user. @user1823 What kind of benefit are you expecting?

user1823:
> I struggle to imagine a situation where this change would benefit the average user.

Many of the stats displayed by Anki don't really benefit the average user. So, if we go by this logic, we can actually remove many of the stats and most people won't bother.

But, I think that if we are displaying something, we should try to present it in the most statistically sound way possible.

brishtibheja:
I agree with @expertium on this. Even if English uses one single word to refer to different concepts, not all cultures are the same. For example, in JP different words are used for 「mean」,「median」. Ref: https://ja.statisticseasily.com/平均値と中央値/

ChatGPT says Russian, Chinese, Korean also generally use different words but I can't verify it myself.

Expertium:
Yeah, we don't refer to the median as "average" in Russian. Well, at least people who know stats don't.

dae:
I get the impression that 'average' is much more commonly used to refer to the mean in English as well. If we're going to present the median, I think it's better we label it accurately.

@Expertium's arguments about consistency and simplicity are reasonable - I suspect there's going to be a percentage of users who can understand 'average ...' un-aided, but may not be familiar with medians. That said, they can always learn - whereas if we choose a suboptimal way of presenting data for the sake of simplicity, people who understand medians (and care about the difference) have no (easy) way of getting the data they need.

So I guess the question comes down to: "is the median a sufficiently-better metric in certain cases to justify the extra potential confusion?" . For non-uniformally-distributed things like difficulty, it would seem like the median is the better choice?

user1823:
> "is the median a sufficiently-better metric in certain cases to justify the extra potential confusion?"

I feel so, but I may be biased.

> but may not be familiar with medians. That said, they can always learn

In addition, because the median will be displayed just below the graph and would have a value close to the middle of the distribution, guessing the meaning won't be too difficult for most users, I feel.

> If we're going to present the median, I think it's better we label it accurately.

So, should I revert the last commit?

PR #3901, opened 2025-04-06
Improve natural unit conversion for a time b/w 360 to 365 days

user1823:
Previously, 363 days would be converted to 12.1 months, which is quite confusing because
- a user would think that if the value is more than 12 months, why it isn't displayed in years
- the value is actually less than a year, which is counterintuitive as 12.1 m suggests a value more than a year.

### DIFF
--- a/ts/lib/tslib/time.ts
+++ b/ts/lib/tslib/time.ts
@@ -7,7 +7,7 @@ export const SECOND = 1.0;
 export const MINUTE = 60.0 * SECOND;
 export const HOUR = 60.0 * MINUTE;
 export const DAY = 24.0 * HOUR;
-export const MONTH = 30.0 * DAY;
+export const MONTH = 30.417 * DAY; // 365/12 ≈ 30.417
 export const YEAR = 365.0 * DAY;
 
 export enum TimespanUnit {
@@ -46,7 +46,7 @@ export function naturalUnit(secs: number): TimespanUnit {
         return TimespanUnit.Hours;
     } else if (secs < MONTH) {
         return TimespanUnit.Days;
-    } else if (secs < 12.0 * MONTH) {
+    } else if (secs < YEAR) {
         return TimespanUnit.Months;
     } else {
         return TimespanUnit.Years;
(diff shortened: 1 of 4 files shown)


ts/lib/tslib/time.ts
```diff
@@ -46,7 +46,7 @@ export function naturalUnit(secs: number): TimespanUnit {
         return TimespanUnit.Hours;
     } else if (secs < MONTH) {
         return TimespanUnit.Days;
-    } else if (secs < YEAR) {
+    } else if (secs < 12.0 * MONTH) {
```
iamllama, line 49: Why is it necessary to keep the old behaviour (i.e a year being 360 days) here?

user1823: If we make it 365 days here, a time like 363 days would be displayed as 12.1 months, which is confusing. I am adding another test in the corresponding Rust code that would fail if the part corresponding to this is changed.

user1823:
I updated the PR to use the average duration of month (as you suggested) and updated the relevant tests.

dae:
Thank you!

david-allison:
@user1823 The `Reviews - Time` section of the statistics** can be measured in months.

Will this change mean that power users will 'lose' progress if their 'total studied' time is > 1 month?

https://github.com/ankitects/anki/blob/615bbf95a1f20f3e7477a97b3265f6d4f19c5fdd/ts/routes/graphs/reviews.ts#L118

> Just in general, using "month" in this sense feels strange to me as a native English speaker. I'd prefer "hours" or maybe "days".

user1823:
@david-allison, I suppose that you mean the reviews section of the statistics, not settings.

Yes, you are right. Unfortunately, this change has the side effect of slightly decreasing the reported total study time if that's more than a month.

Also, as you said, I too feel that using "months" to describe study time is weird. This has been discussed on the forums too and Damien would accept a PR that prevents the study time from being shown in a unit greater than "hours".

https://forums.ankiweb.net/t/reviews-graph-units-of-total-time-studied-suggestion/61237/14

Edit: I opened a PR
- https://github.com/ankitects/anki/pull/4086

Issue #1503, opened 2021-11-23
HTML editor on the right instead of below field?


dae:
I think I raised this point before, but can't find the original location. I still think we should consider showing the HTML editor on the right instead. On a field that has a lot of content (eg, is larger than the entire editing area), clicking on HTML editor button does nothing except change the icon, as the HTML editor is hidden/shown outside of the visible area. Even if it does fit on the screen, the user may have to scroll up and down to see how changes made to one of the editors affect changes in the other editor, reducing the usefulness of showing them both on the screen at once.

hgiesel:
Would you want to make this configurable? I'd imagine that there are going to be people who prefer the vertical layout. Especially if their fields don't have that much content, and you don't have a big screen.

Also, whereas a vertical layout doesn't need a resize handle (there's a natural height), for a horizontal layout I think a resize handle that can sit in-between both inputs would make sense, kinda like the one that sits between browser and its sidebar.

dae:
Yep, I think you're right that we'd need a resize handle. Maybe a simpler solution for now would be to scroll to the top of the area when it's toggled on and it's out of view? Not something we need to rush in before the first beta, and we can always wait a while and see what the feedback is like once more users are trying it out.

dae:
Giving this a milestone just as a reminder to consider it after 2.1.50 has received widespread use, not because it needs implementing. If it turns out nobody minds the current behaviour, we can just close this. :-)

dae:
Nobody has asked for this; closing.

PR #3903, opened 2025-04-07
Cleanup old TODO

GithubAnon0000:
This Todo is no longer needed, since #1503 has been closed.


dae:
I think it may have been commented out as we had to maintain compatibility with older browser versions at the time. But as its author is no longer active here, removing it is probably for the best 👍

PR #3905, opened 2025-04-07
Add Cmd+W close shortcut (on Mac) to more dialogs

beyondcompute:
As I’ve been editing decks more often, I noticed that editor dialogs don’t respond to Cmd+W keyboard shortcut. Here’s an attempt to add the shortcut to “Edit Current”, “Fields” and “Cards” editor dialogs. And also to several modals.

UI of “Edit Current” dialog with several input fields

### DIFF
--- a/qt/aqt/editcurrent.py
+++ b/qt/aqt/editcurrent.py
@@ -9,7 +9,7 @@ from anki.collection import OpChanges
 from anki.errors import NotFoundError
 from aqt import gui_hooks
 from aqt.qt import *
-from aqt.utils import restoreGeom, saveGeom, tr
+from aqt.utils import add_close_shortcut, restoreGeom, saveGeom, tr
 
 
 class EditCurrent(QMainWindow):
@@ -36,6 +36,7 @@ class EditCurrent(QMainWindow):
         close_button = self.form.buttonBox.button(QDialogButtonBox.StandardButton.Close)
         assert close_button is not None
         close_button.setShortcut(QKeySequence("Ctrl+Return"))
+        add_close_shortcut(self)
         # qt5.14+ doesn't handle numpad enter on Windows
         self.compat_add_shorcut = QShortcut(QKeySequence("Ctrl+Enter"), self)
         qconnect(self.compat_add_shorcut.activated, close_button.click)
--- a/qt/aqt/addons.py
+++ b/qt/aqt/addons.py
@@ -40,7 +40,7 @@ from aqt import gui_hooks
 from aqt.log import ADDON_LOGGER_PREFIX, find_addon_logger, get_addon_logs_folder
 from aqt.qt import *
 from aqt.utils import (
-    addCloseShortcut,
+    add_close_shortcut,
     askUser,
     disable_help_button,
     getFile,
@@ -829,7 +829,7 @@ class AddonsDialog(QDialog):
         self.setAcceptDrops(True)
         self.redrawAddons()
         restoreGeom(self, "addons")
-        addCloseShortcut(self)
+        add_close_shortcut(self)
         gui_hooks.addons_dialog_will_show(self)
         self._onAddonSelectionChanged()
         self.show()
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -58,6 +58,7 @@ from aqt.utils import (
     HelpPage,
     KeyboardModifiersPressed,
     add_ellipsis_to_action_label,
+    addCloseShortcut,
     current_window,
     ensure_editor_saved,
     getTag,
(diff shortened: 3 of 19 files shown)


beyondcompute:
> ctrl+w to close a window would be nice to have on other OSs as well, not just for mac in my opinion.

Actually that sounds more than reasonable!

I am not very knowledgeable about other operating systems and their UI patterns (and it would be a bit harder for me to test) but perhaps we could un-Mac-ify this one (and remove explicit `is_mac` checks elsewhere):

https://github.com/ankitects/anki/blob/ccab18b7ba624d888f3d881e14f04c830e3eaa44/qt/aqt/utils.py#L991

qt/aqt/editcurrent.py
```diff
@@ -9,7 +9,7 @@
 from anki.errors import NotFoundError
 from aqt import gui_hooks
 from aqt.qt import *
-from aqt.utils import restoreGeom, saveGeom, tr
+from aqt.utils import add_close_shortcut, restoreGeom, saveGeom, tr
```
beyondcompute, line 12: A bit of a confusing naming perhaps. We have `addCloseShortcut` for `QDialog` instances specifically. And `add_close_shortcut ` for more general `QWidget` (which would include [`QMainWindow`](https://doc.qt.io/qt-6/qmainwindow.html)).

beyondcompute:
> I probably can't really help with the programming site of things, but I can test your proposed solution on debian linux (sid).

Here’s the branch: https://github.com/beyondcompute/anki/tree/make-ctrl_w-keyboard-shorcut-work-on-win-and-linux . (It includes changes for this PR as well. 🤷‍♂️)

I hope, someone who is really familiar with Anki and its multiple views and dialogs can test those thoroughly. Also, we’d obviously need a Windows tester. 🪟😳

beyondcompute:
Some investigation notes. 📝
1. Places where we use `addCloseShortcut` seem to work
2. In Browse tab’s window we do this: https://github.com/ankitects/anki/blob/ccab18b7ba624d888f3d881e14f04c830e3eaa44/qt/aqt/browser/browser.py#L291-L292 I am not sure, is it better to show it on all systems or should I just add the shortcut “manually” (`add_close_shortcut` or whatever) and keep the action hidden (whatever that means 🤔)
3. There’s also that weird behavior: `Cmd+W` in a sub-window (like Preview) of the Browse view closes both the “child” Preview window and the “parent” Browse window. I tried to address that here: https://github.com/ankitects/anki/pull/3913
4. I am adding more closing shortcuts in this PR (overall, there are many small dialogs for renaming things, etc… 🤔)

beyondcompute:
@GithubAnon0000 I’ve added the shortcut to more dialogs and updated the branch. But the Browse sub-dialogs still will not work (see the investigation notes above).

Not sure if any of this will get merged. 😄

Also something like https://github.com/ankitects/anki/pull/3913 perhaps would be nice to have.

dae:
Thank you! Happy to add the shortcut to any screens that are missing it, provided we don't introduce regressions. Some notes:

- I don't think we should be adding the shortcut to the other platforms. It may conflict with something else, and it's not a standard shortcut on those platforms. Across all platforms, you can already use Esc to close most Anki windows, and those other platforms tend to have some other shortcut like Alt+F4 that is provided by the desktop environment instead of the app itself.
- The browser change is superfluous - cmd+w already works there.
- Agreed on it being confusing that the camelCase version of the function behaves differently. Digging through the history, the setattr() line appears to have been added to make the typechecker happy, and appears it can be safely removed. Furthermore, a brief test of switching from the camelCase to the snake_case version for the deck options seems to indicate they're interchangeable. I wonder if we could remove one?

beyondcompute:
@dae change in `browser.py` is for the `grade_now` modal (or not sure that I am understanding correctly about the change in the browser) and I see the behavior difference between `main` and this. 😳

derivativeoflog7:
> > * I don't think we should be adding the shortcut to the other platforms. It may conflict with something else, and it's not a standard shortcut on those platforms. Across all platforms, you can already use Esc to close most Anki windows, and those other platforms tend to have some other shortcut like Alt+F4 that is provided by the desktop environment instead of the app itself.
> 
> Is that the default for windows? In linux Ctrl+w is pretty much normal (I never even heard of Alt+F4).
> 
> Maybe the `is_mac` code could be updated to `is_mac or is_linux` (or `!is_windows`)?
> 
> **Edit:** Alt+F4 does exist on linux (at least on gnome). It closes the entire window though, unlike Ctrl+w. That's especially noticable when apps have tabs.

In Linux (and Windows), Ctrl+W is usually reserved for application with tabbed views (which may cause the whole application to be closed when the last open tab is closed, depending on how it's implemented)

dae:
Merging this in for now, but a global approach like mentioned on https://github.com/ankitects/anki/pull/3913#pullrequestreview-2760169590 is probably the better longer-term solution.

PR #3907, opened 2025-04-08
Add a warning to "Ignore cards reviewed before"

Expertium:
image

Me and other users believe that currently "Ignore cards reviewed before" doesn't do a very good job of, well, conveying that it ignores **cards**, not just a part of their review history.


Expertium:
@Luc-Mcgrady I have no idea if what I did is correct, so please check
Ideally, I'd like to display the warning only if the chosen date is after the date of the first ever review the user did in Anki, but I don't know how to do that, so I just made it a constant warning that never disappears.

cdepillabout:
For what it's worth, I was the one that made the above comment.  As you can see, I was quite surprised that entire cards were getting ignored (even reviews done on those cards AFTER the date specified).

Here's a link to the conversation on Discord for more context: https://discord.com/channels/368267295601983490/1282005522513530952/1359003159317713056

Expertium:
Yeah, I'm just hoping that @Luc-Mcgrady will make a PR himself, since I suck at coding and barely figured out how to do this PR

PR #3904, opened 2025-04-07
Add an option to disable middle click to paste on Linux

derivativeoflog7:
Add an option to disable middle click to paste on Linux as a boolean in profileConf
I made sure to test if the setting works and saves properly, and can also confirm it won't cause issues when downgrading

PR #3908, opened 2025-04-08
Add an option to disable middle click to paste on Linux

derivativeoflog7:
Add an option to disable middle click to paste on Linux as a boolean in the collection config
This is in contrast to the previous PR (#3904) where it was added as a profile setting
I made sure to test if the setting works and saves properly, and can also confirm it won't cause issues when downgrading

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -219,6 +219,7 @@ Jakub Fidler <jakub.fidler@protonmail.com>
 Valerie Enfys <val@unidentified.systems>
 Julien Chol <https://github.com/chel-ou>
 ikkz <ylei.mk@gmail.com>
+derivativeoflog7 <https://github.com/derivativeoflog7>
 
 ********************
 
--- a/ftl/core/preferences.ftl
+++ b/ftl/core/preferences.ftl
@@ -13,6 +13,7 @@ preferences-media-is-not-backed-up = Media is not backed up. Please create a per
 preferences-on-next-sync-force-changes-in = On next sync, force changes in one direction
 preferences-paste-clipboard-images-as-png = Paste clipboard images as PNG
 preferences-paste-without-shift-key-strips-formatting = Paste without shift key strips formatting
+preferences-middle-click-paste = Middle click to paste
 preferences-generate-latex-images-automatically = Generate LaTeX images (security risk)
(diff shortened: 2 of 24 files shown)


dae:
@abdnh's suggestion to put this in the collection/Rust level is good general advice, but in this particular case, I'm afraid I'm not sure there is enough demand to justify its inclusion. I don't recall anyone ever requesting such a feature, and as middle-click paste is common across apps, wouldn't it make more sense to be addressing this system-wide, instead of adding such an option to each app? You may think adding another option in the preferences is not a big deal, but imagine how things would end up if we said yes to every niche option.

derivativeoflog7:
> I don't recall anyone ever requesting such a feature
I implemented it because not only I don't ever use MCTP, it's actively disrupting my workflow when using Anki on my laptop as it frequently misfires when I use touchpad gestures (and in general, I'm not the only one to hate MCTP on Linux) 

> wouldn't it make more sense to be addressing this system-wide
But the way it's implemented here, it's the code in Anki itself that listens for middle clicks in AnkiWebView and then calls the paste function when fired
In general, while some(?) compositors do have an option to disable it (at least on Wayland*), it's not uncommon to see applications re-implementing it (like Anki does) and then providing a toggle for the feature; and additionally I can't seem to find evidence online that suggests applications have a way to tell if it's enabled from the compositor or not

*KDE Plasma doesn't show any option to disable MCTP when running on X11

derivativeoflog7:
> a hidden profile option

That would simply mean reverting to the previous implementation (#3904), correct?


> that is toggled via a debug command tends to be the preferred route

That sounds reasonable and I understand your point, I didn't even think of that option. Would that require additional code, or is there already a general way to toggle profile settings via debug commands?

derivativeoflog7:
I took a look at the debug console, and mw is accessible from there, but you tell me what to do
Should I close this PR and reopen the old one after undoing adding the checkbox (and string)?

dae:
The PR should only need a few lines changed, like this other hidden option: 2772e358e10d2f5219055a00845a5a78e5f663ca

You can then enable the hidden option via the debug console via something like `mw.pm.profile['newOption'] = True`

derivativeoflog7:
@dae  I'm reopening #3904

PR #3910, opened 2025-04-09
Feat/Ignored before card count

Luc-Mcgrady:
Continuation of #3907

https://github.com/user-attachments/assets/a51fad1d-defa-4b1a-919a-56058f04355f

Displays the number of cards that will be ignored by ignore cards reviewed before to hopefully abate some confusion.

### DIFF
--- a/rslib/src/storage/card/get_ignored_before_count.sql
+++ b/rslib/src/storage/card/get_ignored_before_count.sql
@@ -2,3 +2,4 @@ SELECT DISTINCT COUNT(cid)
 FROM revlog
 WHERE id > ?
   AND type == 0
+  AND cid IN search_cids
--- a/rslib/src/storage/card/get_ignored_before_count.sql
+++ b/rslib/src/storage/card/get_ignored_before_count.sql
@@ -1,4 +1,4 @@
-SELECT DISTINCT COUNT(cid)
+SELECT COUNT(DISTINCT cid)
 FROM revlog
 WHERE id > ?
   AND type == 0
--- a/rslib/src/storage/card/get_ignored_before_count.sql
+++ b/rslib/src/storage/card/get_ignored_before_count.sql
@@ -1,0 +1,4 @@
+SELECT DISTINCT COUNT(cid)
+FROM revlog
+WHERE id > ?
+  AND type == 0
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -372,7 +372,7 @@ deck-config-good-above-easy = The easy interval should be at least as long as th
 deck-config-relearning-steps-above-minimum-interval = The minimum lapse interval should be at least as long as your final relearning step.
 deck-config-maximum-answer-secs-above-recommended = Anki can schedule your reviews more efficiently when you keep each question short.
 deck-config-too-short-maximum-interval = A maximum interval less than 6 months is not recommended.
-deck-config-ignore-before-info = { $included }/{ $totalCards } Cards will be considered while calculating parameters.
+deck-config-ignore-before-info = { $included }/{ $totalCards } cards will be used to optimize the FSRS parameters.
 
 ## Selecting a deck
 
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -372,7 +372,7 @@ deck-config-good-above-easy = The easy interval should be at least as long as th
 deck-config-relearning-steps-above-minimum-interval = The minimum lapse interval should be at least as long as your final relearning step.
 deck-config-maximum-answer-secs-above-recommended = Anki can schedule your reviews more efficiently when you keep each question short.
 deck-config-too-short-maximum-interval = A maximum interval less than 6 months is not recommended.
-deck-config-ignore-before-info = { $included }/{ $totalCards } cards will be used to optimize the FSRS parameters.
+deck-config-ignore-before-info = (Approximately) { $included }/{ $totalCards } cards will be used to optimize the FSRS parameters.
 
 ## Selecting a deck
 
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -372,6 +372,7 @@ deck-config-good-above-easy = The easy interval should be at least as long as th
 deck-config-relearning-steps-above-minimum-interval = The minimum lapse interval should be at least as long as your final relearning step.
 deck-config-maximum-answer-secs-above-recommended = Anki can schedule your reviews more efficiently when you keep each question short.
 deck-config-too-short-maximum-interval = A maximum interval less than 6 months is not recommended.
+deck-config-ignore-before-info = { $included }/{ $totalCards } Cards will be considered while calculating parameters.
 
 ## Selecting a deck
 
--- a/proto/anki/deck_config.proto
+++ b/proto/anki/deck_config.proto
@@ -23,7 +23,8 @@ service DeckConfigService {
   rpc GetDeckConfigsForUpdate(decks.DeckId) returns (DeckConfigsForUpdate);
   rpc UpdateDeckConfigs(UpdateDeckConfigsRequest)
       returns (collection.OpChanges);
-  rpc GetIgnoredBeforeCount(IgnoredBeforeSearch) returns (IgnoredBeforeCount);
+  rpc GetIgnoredBeforeCount(GetIgnoredBeforeCountRequest)
+      returns (GetIgnoredBeforeCountResponse);
 }
 
 // Implicitly includes any of the above methods that are not listed in the
@@ -34,12 +35,12 @@ message DeckConfigId {
   int64 dcid = 1;
 }
 
-message IgnoredBeforeSearch {
+message GetIgnoredBeforeCountRequest {
   string ignore_revlogs_before_date = 1;
   string search = 2;
 }
 
-message IgnoredBeforeCount {
+message GetIgnoredBeforeCountResponse {
   uint64 included = 1;
   uint64 total = 2;
 }
--- a/proto/anki/deck_config.proto
+++ b/proto/anki/deck_config.proto
@@ -23,7 +23,7 @@ service DeckConfigService {
   rpc GetDeckConfigsForUpdate(decks.DeckId) returns (DeckConfigsForUpdate);
   rpc UpdateDeckConfigs(UpdateDeckConfigsRequest)
       returns (collection.OpChanges);
-  rpc GetIgnoredBeforeCount(TimestampMilliseconds) returns (CardCount);
+  rpc GetIgnoredBeforeCount(IgnoredBeforeSearch) returns (IgnoredBeforeCount);
 }
 
 // Implicitly includes any of the above methods that are not listed in the
@@ -34,12 +34,14 @@ message DeckConfigId {
   int64 dcid = 1;
 }
 
-message TimestampMilliseconds {
+message IgnoredBeforeSearch {
   int64 ms = 1;
+  string search = 2;
 }
 
-message CardCount {
-  int64 count = 1;
+message IgnoredBeforeCount {
+  uint64 included = 1;
+  uint64 total = 2;
 }
 
 message DeckConfig {
--- a/proto/anki/deck_config.proto
+++ b/proto/anki/deck_config.proto
@@ -23,6 +23,7 @@ service DeckConfigService {
   rpc GetDeckConfigsForUpdate(decks.DeckId) returns (DeckConfigsForUpdate);
   rpc UpdateDeckConfigs(UpdateDeckConfigsRequest)
       returns (collection.OpChanges);
+  rpc GetIgnoredBeforeCount(TimestampMilliseconds) returns (CardCount);
 }
 
 // Implicitly includes any of the above methods that are not listed in the
@@ -31,6 +32,14 @@ service BackendDeckConfigService {}
 
 message DeckConfigId {
   int64 dcid = 1;
+}
+
+message TimestampMilliseconds {
+  int64 ms = 1;
+}
+
+message CardCount {
+  int64 count = 1;
 }
 
 message DeckConfig {
--- a/proto/anki/deck_config.proto
+++ b/proto/anki/deck_config.proto
@@ -35,7 +35,7 @@ message DeckConfigId {
 }
 
 message IgnoredBeforeSearch {
-  int64 ms = 1;
+  string ignore_revlogs_before_date = 1;
   string search = 2;
 }
 
--- a/qt/aqt/mediasrv.py
+++ b/qt/aqt/mediasrv.py
@@ -652,6 +652,8 @@ exposed_backend_list = [
     "evaluate_params",
     "get_optimal_retention_parameters",
     "simulate_fsrs_review",
+    # DeckConfigService
+    "get_ignored_before_count"
 ]
 
 
--- a/qt/aqt/mediasrv.py
+++ b/qt/aqt/mediasrv.py
@@ -653,7 +653,7 @@ exposed_backend_list = [
     "get_optimal_retention_parameters",
     "simulate_fsrs_review",
     # DeckConfigService
-    "get_ignored_before_count"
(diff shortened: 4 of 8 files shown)


Luc-Mcgrady:
@user1823 Would you mind reviewing this for me?

rslib/src/storage/card/get_ignored_before_count.sql
```diff
@@ -0,0 +1,5 @@
+SELECT COUNT(DISTINCT cid)
+FROM revlog
+WHERE id > ?
+  AND type == 0
+  AND cid IN search_cids
```
Luc-Mcgrady, line 5: This improperly counts `L  L | L R` cards. But I think a fix would be complicated.

user1823: Can you try this SQL?

```sql
WITH ordered AS (
  SELECT
    cid,
    id,
    type,
    LAG(type) OVER (PARTITION BY cid ORDER BY id) AS prev_type
  FROM revlog
),
groups AS (
  -- mark the start of a new group when type = 0 and previous row is not 0 (or there is no previous row)
  SELECT
    cid,
    id,
    type,
    CASE 
      WHEN type = 0 AND (prev_type IS NULL OR prev_type <> 0) THEN 1 
      ELSE 0 
    END AS group_start
  FROM ordered
),
group_assign AS (
  -- assign a group number by accumulating the group_start flag
  SELECT
    cid,
    id,
    type,
    SUM(group_start) OVER (PARTITION BY cid ORDER BY id ROWS UNBOUNDED PRECEDING) AS grp
  FROM groups
  WHERE type = 0
),
group_bounds AS (
  -- for each contiguous group (grp) for a given cid, compute the first (minimum) id
  SELECT
    cid,
    grp,
    MIN(id) AS first_in_group
  FROM group_assign
  GROUP BY cid, grp
),
target AS (
  -- the last contiguous group is the one with the largest first_in_group for each cid
  SELECT
    cid,
    MAX(first_in_group) AS first_in_last_group
  FROM group_bounds
  GROUP BY cid
)
SELECT COUNT(*)
FROM target
WHERE first_in_last_group < ?;
```

user1823 (commented):
Looks great. Left two suggestions.

rslib/src/storage/card/get_ignored_before_count.sql
user1823: If it is that slow, we might want to just use the simple SQL you included in the PR. But, would you give a last chance to the following optimized SQL to see if its performance is acceptable?

```sql
SELECT COUNT(*)
FROM (
  SELECT cid,
         MAX(first_id) AS first_in_last_group
  FROM (
    SELECT cid,
           id,
           CASE WHEN type = 0 AND prev_type <> 0 THEN 1 ELSE 0 END AS group_start,
           SUM(CASE WHEN type = 0 AND (prev_type IS NULL OR prev_type <> 0) THEN 1 ELSE 0 END)
             OVER (PARTITION BY cid ORDER BY id) AS grp,
           FIRST_VALUE(id) OVER (PARTITION BY cid,
                                       SUM(CASE WHEN type = 0 AND (prev_type IS NULL OR prev_type <> 0) THEN 1 ELSE 0 END)
                                       OVER (PARTITION BY cid ORDER BY id) ORDER BY id) AS first_id
    FROM (
      SELECT cid,
             id,
             type,
             LAG(type) OVER (PARTITION BY cid ORDER BY id) AS prev_type
      FROM revlog
      WHERE cid IN (search_cids)
    ) t
  ) sub
  GROUP BY cid
) target
WHERE first_in_last_group > ?;
```

Luc-Mcgrady: > But, would you give a last chance to the following optimized SQL to see if its performance is acceptable?

No problem!
image
```
Error: 500: DbError { info: "SqlInputError { error: Error { code: Unknown, extended_code: 1 }, msg: \"no such column: search_cids\", sql: \"SELECT COUNT(*)\\nFROM (\\n  SELECT cid,\\n         MAX(first_id) AS first_in_last_group\\n  FROM (\\n    SELECT cid,\\n           id,\\n           CASE WHEN type = 0 AND prev_type <> 0 THEN 1 ELSE 0 END AS group_start,\\n           SUM(CASE WHEN type = 0 AND (prev_type IS NULL OR prev_type <> 0) THEN 1 ELSE 0 END)\\n             OVER (PARTITION BY cid ORDER BY id) AS grp,\\n           FIRST_VALUE(id) OVER (PARTITION BY cid,\\n                                       SUM(CASE WHEN type = 0 AND (prev_type IS NULL OR prev_type <> 0) THEN 1 ELSE 0 END)\\n                                       OVER (PARTITION BY cid ORDER BY id) ORDER BY id) AS first_id\\n    FROM (\\n      SELECT cid,\\n             id,\\n             type,\\n             LAG(type) OVER (PARTITION BY cid ORDER BY id) AS prev_type\\n      FROM revlog\\n      WHERE cid IN (search_cids)\\n    ) t\\n  ) sub\\n  GROUP BY cid\\n) target\\nWHERE first_in_last_group > ?;\", offset: 807 }", kind: Other }
```
This error I can see is caused by `(search_cids)` having brackets but even fixing that
```
Error: 500: DbError { info: "SqlInputError { error: Error { code: Unknown, extended_code: 1 }, msg: \"misuse of window function SUM()\", sql: \"SELECT COUNT(*)\\nFROM (\\n  SELECT cid,\\n         MAX(first_id) AS first_in_last_group\\n  FROM (\\n    SELECT cid,\\n           id,\\n           CASE WHEN type = 0 AND prev_type <> 0 THEN 1 ELSE 0 END AS group_start,\\n           SUM(CASE WHEN type = 0 AND (prev_type IS NULL OR prev_type <> 0) THEN 1 ELSE 0 END)\\n             OVER (PARTITION BY cid ORDER BY id) AS grp,\\n           FIRST_VALUE(id) OVER (PARTITION BY cid,\\n                                       SUM(CASE WHEN type = 0 AND (prev_type IS NULL OR prev_type <> 0) THEN 1 ELSE 0 END)\\n                                       OVER (PARTITION BY cid ORDER BY id) ORDER BY id) AS first_id\\n    FROM (\\n      SELECT cid,\\n             id,\\n             type,\\n             LAG(type) OVER (PARTITION BY cid ORDER BY id) AS prev_type\\n      FROM revlog\\n      WHERE cid IN search_cids\\n    ) t\\n  ) sub\\n  GROUP BY cid\\n) target\\nWHERE first_in_last_group > ?;\", offset: 448 }", kind: Other }
```

user1823: To solve this, we would need a separate subquery (or a common table expression), which would have performance issues. So, just leave it and use your own simple query.

user1823: By the way, I just realized that we can make this faster by creating a temporary table with pre-computed values of first_in_last_group and then compare it to the newer values of ignore_before_date as the setting is being changed. This way, we won't need to compute the first_in_last_group every time a new date is selected.

sql/update_first_in_last_group.sql
```sql
-- Create and populate a temporary table with the first_in_last_group values.
CREATE TEMP TABLE IF NOT EXISTS temp_first_in_last_group (
    cid INTEGER PRIMARY KEY,
    first_in_last_group INTEGER
);

-- Refresh the temporary table.
WITH ordered AS (
  SELECT
    cid,
    id,
    type,
    LAG(type) OVER (PARTITION BY cid ORDER BY id) AS prev_type
  FROM revlog
  WHERE cid IN (search_cids)
),
groups AS (
  SELECT
    cid,
    id,
    type,
    CASE 
      WHEN type = 0 AND (prev_type IS NULL OR prev_type <> 0) THEN 1 
      ELSE 0 
    END AS group_start
  FROM ordered
),
group_assign AS (
  SELECT
    cid,
    id,
    type,
    SUM(group_start) OVER (PARTITION BY cid ORDER BY id ROWS UNBOUNDED PRECEDING) AS grp
  FROM groups
  WHERE type = 0
),
group_bounds AS (
  SELECT
    cid,
    grp,
    MIN(id) AS first_in_group
  FROM group_assign
  GROUP BY cid, grp
),
target AS (
  SELECT
    cid,
    MAX(first_in_group) AS first_in_last_group
  FROM group_bounds
  GROUP BY cid
)
-- Populate (or replace) the cache table.
INSERT OR REPLACE INTO temp_first_in_last_group (cid, first_in_last_group)
SELECT cid, first_in_last_group
FROM target;
```

The above SQL will be executed when the user changes the ignore_before setting for the first time within a session. On subsequent changes of the setting done before closing the Deck Options page, the above SQL won't be re-run.

Then, the count query becomes much simpler:
```sql
SELECT COUNT(*)
FROM temp_first_in_last_group
WHERE first_in_last_group > ?;
```

Luc-Mcgrady: I made a couple of adjustments but caching the `first_in_last_group` values works great!

https://github.com/user-attachments/assets/00054673-94ee-4c4d-a24d-be30c411d341

user1823: The above one seems to work great, but my dream revealed an even more efficient query.

```sql
WITH candidates AS (
  SELECT
    r.*,
    CASE 
      WHEN type = 0 AND 
           (
             LAG(type) OVER (PARTITION BY cid ORDER BY id) IS NULL OR
             LAG(type) OVER (PARTITION BY cid ORDER BY id) <> 0
           )
      THEN 1 
      ELSE 0 
    END AS boundary_flag
  FROM revlog
  WHERE cid IN search_cids
)
SELECT
  cid,
  MAX(id) AS first_in_last_group
FROM candidates
WHERE boundary_flag = 1
GROUP BY cid
```

Luc-Mcgrady: https://github.com/user-attachments/assets/cf58b91e-215d-4dd6-9f09-29e1bb0c9fff

(With caching disabled)

Just eyeballing it it appears slightly slower.

user1823: So, it looks like performance of SQL is not directly related to the number of CTEs used. Interesting 🤔 
Anyways, just forget about this suggestion, then.

dae:
I like the improved info for users, but am not very fond of the current approach:

- It runs as the page is loaded, leading to slower loads on large collections
- On change, it's invoking the backend method multiple times for each change. And if the user wants to hold down a key to scroll through dates, it's updating on every change.
- It's using a global cache that isn't cleaned up, and thus leaks memory
- It adds a bunch of extra code that partially duplicates our existing filtering logic, and we'd need to maintain them in lock-step moving forward.
- The user's chosen search also alters how many cards will be shown. This does update the notice when the search is changed, but as they're in different sections, this is not obvious.

What if instead, the optimization routines kept track of included/excluded cards, and returned them at the end of optimization? A message could then be shown as optimization completes indicating the number of cards that were included/excluded. It would mean we can re-use existing codepaths, and the performance impact over what we currently have would be negligible. We already show the review count, so this info could be shown next to it.

WDYT?

Expertium:
So if I understand you correctly, you want to show the number of included and total cards after optimization? I see two problems with that:
1) It's backwards - users set the date first and then run the optimization, in that order
2) More importantly, it means that if someone wants to see how changing the date affects the number of included cards, he will have to do a lot of optimizations and a lot of scrolling from the FSRS section to the Advanced section and back

Luc-Mcgrady:
> - It runs as the page is loaded, leading to slower loads on large collections.

Does this matter if its loaded asynchronously? I don't think it would block the user from interacting with the page while it is loading so it shouldn't matter right?

> - On change, it's invoking the back-end method multiple times for each change. And if the user wants to hold down a key to scroll through dates, it's updating on every change. 
> - It's using a global cache that isn't cleaned up, and thus leaks memory

Another option I was considering would be to instead of calling the back-end several times, have the list of `get_cards_first_in_last_group_of` passed to the front-end and then use JavaScript to calculate the number of values below the ignored_before. This would fix the memory leak although I imagine it would make it run slower. Good idea?

> - It adds a bunch of extra code that partially duplicates our existing filtering logic, and we'd need to maintain them in lock-step moving forward.

I'm personally happy with the value being an approximation rather than the exact correct value. Unless there are radical changes to the way ignore_reviews_before works, I can't foresee the SQL becoming terribly inaccurate even if it is ignored while future changes are made.

> - The user's chosen search also alters how many cards will be shown. This does update the notice when the search is changed, but as they're in different sections, this is not obvious.

I'm sorry, I can't think of anything that can be done about this :smile:.

>What if instead, the optimization routines kept track of included/excluded cards, and returned them at the end of optimization? A message could then be shown as optimization completes indicating the number of cards that were included/excluded. It would mean we can re-use existing codepaths, and the performance impact over what we currently have would be negligible. We already show the review count, so this info could be shown next to it.
>
>WDYT?

I could see that working as a compromise. A 
```
100 reviews 
9938/10000 cards ignored by ignored before
```
like message would probably still prevent people from misconstruing how ignore-reviews-before works. I do personally much prefer the message as it currently exists though.

Luc-Mcgrady:
> Why is calling the backend multiple times an issue? Is that bad coding practise?
> On change, it's invoking the back-end method multiple times for each change. And if the user wants to hold down a key to scroll through dates, it's updating on every change.

Oops I miss-read. The problem here is that the back-end is being called multiple times every time the value is changed once.

user1823:
Even with this SQL, we are not being 100% accurate (some cards may have incomplete review history, and thus ignored) and the heavy SQL query is introducing a host of issues (as pointed out by dae). So, I was thinking that we should just switch back to the simple SQL that Luc used initially and then show a message like the following after rounding off the results:

```
~2,000 / ~10,000 cards will be used to optimize the FSRS parameters.
```

> The user's chosen search also alters how many cards will be shown ... but as they're in different sections, this is not obvious.

Is this important? The goal of the PR is just to demonstrate the effect of the ignore cards before option.

Luc-Mcgrady:
I've reverted to a version which has a more approximate but faster SQL (https://github.com/ankitects/anki/pull/3910#discussion_r2035405783). I've also fixed the problem with multiple calls to the back-end. I hope this will be more acceptable to you :+1:.

dae:
I like that you've fixed the duplicate calls, and simplified the SQL. If we have to maintain two separate paths, then having one that only tries to be a rough approximation and is not expect to be identical is going to be easier to maintain. 

I tested this on a reasonably large collection with a beefy machine from this year, and each call is about 35ms. On slower devices and more extreme collection sizes, it's definitely going to be a noticeable delay. I don't think users should be paying that cost when they don't need it, such as:
- When opening the stats window. Unless they're specifically adjusting that option, they likely don't need to see the card count each time. You asked about doing it asynchronously, and that would be an improvement, but I think it would make more sense to wait until the user tries to change the value (or we add an explicit 'check' button).
- When they make edits to the search or limit, the info update should be delayed and only run after, say, 1 second of inactivity, or things may get laggy for users trying to type something in and having it calculated on every keystroke/repeated key.

Luc-Mcgrady:
> You asked about doing it asynchronously, and that would be an improvement, but I think it would make more sense to wait until the user tries to change the value (or we add an explicit 'check' button).

Oh I meant that I thought it was already done asynchronously. Does it cause unresponsiveness when you open the deck config?

> When opening the options window. Unless they're specifically adjusting that option, they likely don't need to see the card count each time.

Personally, I'm a little concerned that people that have set this option to something wrong before this update will then fail to notice it. 8ef5d6dc0520445f9424e98dc30953a74d23100c

> When they make edits to the search or limit, the info update should be delayed and only run after, say, 1 second of inactivity, or things may get laggy for users trying to type something in and having it calculated on every keystroke/repeated key.

Done! 3fd6e82cc84b97c519b603aa2a47fb1ec7322832

dae:
Thank you for the quick fixes! A gentle reminder about camelCase being the convention.

It doesn't cause a noticeable startup delay on my system, but we have a responsibility not to burn users' CPU cycles unnecessarily. I appreciate your point about helping existing users notice though, and as a compromise, I've temporarily disabled that logic - we can enable it again in a future update.

PR #3912, opened 2025-04-09
Add a warning when there is a cloze in "back extra" and "text" is empty 

### LINKED ISSUE #3909: Cloze note type shouldn't warin if there is a cloze in "back extra" when "text" is empty
How to reproduce:

* open the add note with the type "cloze"
* leave the "text" field empty
* enter a cloze in the "back extra" field

Expected:
* The warning stating that the cloze should not be in this field

Actual:
* nothing

Note that if you add some content to the "text" field, then the warning appear. Which is not very intuitive, given that this content is not related to the warning

rreemmii-dev:
Adds a warning if there is a cloze in the "back extra" field and the "text" field is left empty.
The warning is the same as when the "text" field has content.

Fixes #3909

### DIFF
--- a/rslib/src/notes/mod.rs
+++ b/rslib/src/notes/mod.rs
@@ -607,14 +607,12 @@ impl Collection {
                 let stripped = strip_html_preserving_media_filenames(&field1);
                 if stripped.trim().is_empty() {
                     NoteFieldsState::Empty
+                } else if cloze_state != NoteFieldsState::Normal {
+                    cloze_state
+                } else if self.is_duplicate(&stripped, note)? {
+                    NoteFieldsState::Duplicate
                 } else {
-                    if cloze_state != NoteFieldsState::Normal {
-                        cloze_state
-                    } else if self.is_duplicate(&stripped, note)? {
-                        NoteFieldsState::Duplicate
-                    } else {
-                        NoteFieldsState::Normal
-                    }
+                    NoteFieldsState::Normal
                 }
             } else {
                 NoteFieldsState::Empty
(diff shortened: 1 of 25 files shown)


dae:
For the user to be able to add a card, they need both:

- a non-blank first field
- clozes only in cloze fields

Your PR breaks the first check, allowing empty cards to be added.

rreemmii-dev:
I fixed it!

dae:
Thank you!

rslib/src/notes/mod.rs
```diff
@@ -594,27 +594,29 @@ impl Collection {
...
-            } else {
-                text.into()
-            };
-            let stripped = strip_html_preserving_media_filenames(&field1);
-            if stripped.trim().is_empty() {
-                NoteFieldsState::Empty
-            } else {
-                let cloze_state = self.field_cloze_check(note)?;
-                if cloze_state != NoteFieldsState::Normal {
+        Ok({
```
Arthur-Milchior, line 597: Thanks @rreemmii-dev for solving my issue.
I'd like to note, even if it's of little importance, that the documentation is not accurate anymore.
Specifically, the "Then for cloze notetypes". Given that the case of cloze is now considered first.

rreemmii-dev: Sorry for the delay, I'll look at it.
Previous to my change, the docs said "Check if the note's first field is empty or a duplicate", but the code checking for duplicate was far from the beginning of the function. I should change this part of the docs too, shouldn't I?

rreemmii-dev: I opened a PR with new documentation, feel free to suggest other changes if necessary:
https://github.com/ankitects/anki/pull/3944

PR #3915, opened 2025-04-11
Fix autoplay not being stopped on editor close if it interrupted another

iamllama:
Found an edge case while testing drag/dropping that i had unfortunately left unresolved in #3666: if an autoplay interrupts another, closing the editor doesn't currently stop it

It's caused by the call to `_on_play_finished` corresponding to an interrupted autoplay happening **after** the latest autoplay's call to `play_file_if_caller`. The fix proposed is add and use a flag to track when such interruptions occur

### DIFF
--- a/qt/aqt/sound.py
+++ b/qt/aqt/sound.py
@@ -142,7 +142,7 @@ class AVPlayer:
     interrupt_current_audio = True
     # caller key for the current playback (optional)
     current_caller = None
-    # whether the last call to play_file_if_caller interrupted another
+    # whether the last call to play_file_with_caller interrupted another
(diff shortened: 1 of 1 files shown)


qt/aqt/sound.py
```diff
@@ -142,6 +142,8 @@ class AVPlayer:
     interrupt_current_audio = True
     # caller key for the current playback (optional)
     current_caller = None
+    # whether the last call to play_file_if_caller interrupted another
```
dae, line 145:
```suggestion
    # whether the last call to play_file_with_caller interrupted another
```

dae:
Thanks!

PR #3916, opened 2025-04-12
Fix drag/drop breaking when editor is zoomed

iamllama:
Zooming the editor in/out causes the cursor coords qt provides to not match up with the [viewport's](https://developer.mozilla.org/en-US/docs/Web/API/Document/elementsFromPoint) when drag/dropping

Rather than disabling editor zoom (the browser explicitly provides menubar items for this), the fix proposed is to factor in zoom by scaling


dae:
Nice one!

PR #3917, opened 2025-04-12
service to state which card would be generated given note type id and field content

Arthur-Milchior:
Introduce get_non_empty_card_indexes

This method takes a note type id and the content of fields. It returns
the order of the cards that would be non-empty according to this note.

To be more specific, for cloze type, it return any cloze
number (except 0), even if it does not appear in a cloze field, in
order to be consistent with the way anki generates note.

For normal type, for each template, it returns whether the card would
be generated, and also whether the template has an error.

The goal here is to allow ankidroid to display in the note editor's
previewer which card will be generated. Which is done currently by
checking whether the "empty card" message appear in the previewer,
which is not ideal.

Please let me know whether you'd accept this service. In which case I'll add a few tests with the default note type. I preferred not to spend more time adding tests until I know whether it'd be useful or not

### DIFF
--- a/rslib/src/cloze.rs
+++ b/rslib/src/cloze.rs
@@ -442,6 +442,15 @@ pub fn add_cloze_numbers_in_string(field: &str, set: &mut HashSet<u16>) {
     add_cloze_numbers_in_text_with_clozes(&parse_text_with_clozes(field), set)
 }
 
+/// The set of cloze numbers as they appear in any of the field from `fields`.
+pub fn cloze_number_in_fields(fields: impl IntoIterator<Item: AsRef<str>>) -> HashSet<u16> {
+    let mut set = HashSet::with_capacity(4);
+    for field in fields {
+        add_cloze_numbers_in_string(field.as_ref(), &mut set);
+    }
+    set
+}
+
 fn strip_html_inside_mathjax(text: &str) -> Cow<str> {
     MATHJAX.replace_all(text, |caps: &Captures| -> String {
         format!(
--- a/rslib/src/cloze.rs
+++ b/rslib/src/cloze.rs
@@ -43,6 +43,7 @@ mod mathjax_caps {
 
 #[derive(Debug)]
 enum Token<'a> {
+    // The parameter is the cloze number as is appears in the field content.
     OpenCloze(u16),
     Text(&'a str),
     CloseCloze,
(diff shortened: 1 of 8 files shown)


user1823:
Wasn't this issue already solved by https://github.com/ankitects/anki/pull/3890?

Arthur-Milchior:
It seems so. I looked for this march 16 when I opened the bug in ankidroid. I had no idea that somebody solved it in the meantime. Wonderful.

Arthur-Milchior:
Also, we'll need to check. But given that we have always a small cost to pay to send request to the back-end, a method that returns all data may be better than having to send one request by template, especially as long as it does not provide a way to send multiple request as a batch.

But i'll see whether it makes sense when we can actually test the feature

rslib/src/cloze.rs
```diff
@@ -409,12 +411,14 @@ pub fn expand_clozes_to_reveal_latex(text: &str) -> String {
     buf
 }
 
+// Whether `text` contains any cloze with strictly positive number.
```
dae, line 414:
```suggestion
// Whether `text` contains any cloze number above 0
```

rslib/src/cloze.rs
```diff
@@ -409,12 +411,14 @@ pub fn expand_clozes_to_reveal_latex(text: &str) -> String {
...
 pub(crate) fn contains_cloze(text: &str) -> bool {
     parse_text_with_clozes(text)
         .iter()
         .any(|node| matches!(node, TextOrCloze::Cloze(e) if e.ordinal != 0))
 }
 
+/// Returns the set of cloze number as the appear in the fields's content.
```
dae, line 421:
```suggestion
/// Returns the set of cloze number as they appear in the fields's content.
```

rslib/src/cloze.rs
```diff
@@ -432,11 +436,21 @@ fn add_cloze_numbers_in_text_with_clozes(nodes: &[TextOrCloze], set: &mut HashSe
     }
 }
 
+/// Add to `set` the cloze numbers as they appear in `field`.
 #[allow(clippy::implicit_hasher)]
 pub fn add_cloze_numbers_in_string(field: &str, set: &mut HashSet<u16>) {
     add_cloze_numbers_in_text_with_clozes(&parse_text_with_clozes(field), set)
 }
 
+/// The set of cloze numbers as they appear in any of the field from `fields`.
```
dae, line 445:
```suggestion
/// The set of cloze numbers as they appear in any of the fields from `fields`.
```

dae:
> a method that returns all data may be better than having to send one request by template

If it proves necessary in the future, we can always add it then.

PR #3919, opened 2025-04-12
Remove use of createClassComponent in mathjax-element.ts

iamllama:
[createClassComponent](https://svelte.dev/docs/svelte/svelte-legacy#createClassComponent) was made available in svelte 5 as a way to ease migration, and is intended to be replaced by [mount](https://svelte.dev/docs/svelte/v5-migration-guide#Components-are-no-longer-classes). This pr proposes to do just that
 
What's odd is that i had to bump esbuild-svelte to 0.9.x. It introduced support for `.svelte.{ts,js}` files in [0.9.0](https://github.com/EMH333/esbuild-svelte/releases/tag/v0.9.0), which is [needed(?)](https://svelte.dev/docs/svelte/svelte-js-files) for `$state` use outside of svelte files

However, there are already 2 existing `.svelte.ts` files, `tooltip-utils.svelte.ts` and `add-or-update-note.svelte.ts`, which seemed to work fine before. My best guess is that the svelte compiler was able to pick up on their use of `$state` by svelte(kit) files, but not `mathjax-element.ts`'s since it's only trivially referenced in svelte files (`mathjaxConfig`)

Testing didn't reveal regressions with mathjax, the stats tooltips or the mobile mask editor's save tooltip


dae:
The move away from the old legacy code is very welcome! I did spot an issue though: the cmd+m m shortcut to add a new MathJax eqn in the add screen seems to be broken

iamllama:
Sorry for not catching that 😅 Pushed a fix that correctly opens up the overlay editor when using the mathjax toolbar buttons now, surprised this discrepancy between createClassComponent and mount wasn't documented (or perhaps i overlooked it)

dae:
Not a problem! Really appreciate you moving us in the right direction here. :-)

PR #3922, opened 2025-04-13
add min-height to fields

GithubAnon0000:
FIX: https://forums.ankiweb.net/t/having-empty-html-in-the-add-cards-fields-causes-the-field-to-shrink-and-be-unusable/58955?u=anon_0000

### DIFF
--- a/ts/editable/editable-base.scss
+++ b/ts/editable/editable-base.scss
@@ -2,6 +2,7 @@
 
 * {
     max-width: 100%;
+    min-height: 30px;
 }
 
 p {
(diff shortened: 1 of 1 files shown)


ts/editable/editable-base.scss
```diff
@@ -2,6 +2,7 @@
 
 * {
     max-width: 100%;
+    min-height: 30px;
```
iamllama, line 5:
```suggestion
    min-height: 10px;
```

Some templates, e.g. [crop theft](https://github.com/Kuuuube/crop-theft), intentionally set small font sizes for fields, to save on space without collapsing
image
10px would be a compromise between that and still being able to click inside the richtext editor

dae:
Thank you both!

PR #3923, opened 2025-04-13
Fix error when middle clicking in editor on systems w/o global mouse selection

iamllama:
On my system, middle clicking in the editor throws:
```
  File "/anki/qt/aqt/webview.py", line 354, in eventFilter
    self.onMiddleClickPaste()
  File "/anki/qt/aqt/editor.py", line 1554, in onMiddleClickPaste
    self._onPaste(QClipboard.Mode.Selection)
  File "/anki/qt/aqt/editor.py", line 1537, in _onPaste
    self._on_clipboard_change(mode)
  File "/anki/qt/aqt/editor.py", line 1492, in _on_clipboard_change
    elif self._internal_field_text_for_paste != self._get_clipboard_html_for_field(
  File "/anki/qt/aqt/editor.py", line 1501, in _get_clipboard_html_for_field
    assert mime is not None
AssertionError
```
> Returns a pointer to a QMimeData representation of the current clipboard data (can be nullptr if the given mode is not supported by the platform).
https://doc.qt.io/qt-6/qclipboard.html#mimeData

The asserts after calls to QClipboard.mimeData with a non-default mode don't always hold, as the docs suggest. The fix proposed is to replace them with guard clauses. Tested in an ubuntu vm to ensure middle-clicking still works when global mouse selection is supported

Regarding the existing middle-click option pr, this still has merit imo, as users shouldn't have to see errors to then dig into how to disable it via the debug console if their systems don't support it in the first place

Also took the opportunity to update about.py, seeing as this'd be my 50th pr 😄

### DIFF
--- a/qt/aqt/about.py
+++ b/qt/aqt/about.py
@@ -219,6 +219,7 @@ def show(mw: aqt.AnkiQt) -> QDialog:
             "Brayan Oliveira",
             "Market345",
             "Yuki",
+            "🦙 (siid)",
         )
     )
 
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -1497,8 +1497,8 @@ class EditorWebView(AnkiWebView):
 
     def _get_clipboard_html_for_field(self, mode: QClipboard.Mode) -> str | None:
         clip = self._clipboard()
-        mime = clip.mimeData(mode)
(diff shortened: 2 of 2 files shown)


dae:
Thank you for all of your contributions @iamllama - they're always appreciated :-)

Issue #3871, opened 2025-03-20
Migrate the reviewing screen to Svelte and sandbox if possible


dae:
Once we're done with #3830, I'd like for us to start on a new reviewing screen that can be shared across the clients. Posting this here to provide a bit of a roadmap, and encourage discussion. Feedback welcome!

The current review screen uses two webviews - the 'main' and 'bottom' webview, and we update the two in lockstep (e.g. displaying buttons when the answer is revealed). These changes are coordinated by the logic in reviewer.py. The code is awkward to maintain, as it relies on legacy practices (creating HTML/JS from Python strings, and sending messages back and forth across the Python/TS boundary). The mobile clients have had to implement their own reviewer logic, meaning changes to things like the answer buttons have to be done multiple times. This has also lead to inconsistencies, most notable of which is AnkiDroid reloads the webview on each card transition.

## Sandboxing

Another issue with the current approach is that it loads the user-provided content on the same origin as the program interface. This allows shared deck JS to manipulate Anki's interface, and make requests to our internal webserver. We've had to block certain methods from being accessed by the review screen due to security issues, and that's only partly solved the problem. It's become a FAQ: https://github.com/ankitects/anki/security/policy#faq

A potentially better solution for isolating user-provided content would be to load it inside an iframe with a different origin. (e.g. something like `<iframe sandbox="allow-scripts ...">`). This would prevent card content from accidentally or deliberately styling/modifying the interface outside of the card, or making any requests via our internal API. I've briefly experimented with this many years ago, but gave up at the time, as there were some extra issues to deal with (e.g. where keystroke events arrive when the outer vs inner frame has focus, sizing the iframe to fit properly, how to use a different origin, etc.) Nothing that looked insurmountable, but it got put in the too-hard basket at the time.

As far as I'm aware, there are a few ways we could give the iframe a different origin, limiting its ability to access our internal API:

- Hosting another webserver on another port. Cons: requires running multiple webservers, and possibly not as secure (some browsers have special rules for different ports on localhost I think?)
- Using 'localhost' and '127.0.0.1' for the different origins. Cons: I think we currently avoid 'localhost', as there are rare misconfigured Windows machines that lack the hostname
- Initializing the iframe with a blob: URL instead of using our server. Cons: the page needs to be bundled into a single file and base64 encoded, which would be awkward, and slow if it pulls in megabytes of dependencies.

[...]

PR #3925, opened 2025-04-15
Harden access to internal API

dae:
I received a report recently that our existing mitigations in mediasrv.py are not sufficient, meaning that a malicious shared deck could potentially access our internal API, and thus arbitrary data on the system. As users have an expectation that shared decks are safe to download, we'll need to get a fix for this out promptly. This is the least disruptive/invasive solution I could come up with:

- Stripping scripts from editor field content may end up breaking some decks that have such content embedded in fields. My thinking here is that such decks are quite rare, and placing such content in fields has never been encouraged.
- Switching to a custom URL scheme would have required a lot more changes, and [requires a newer Qt](https://github.com/ankitects/anki/issues/3871#issuecomment-2803806785).

@abdnh @glutanimate @iamllama feedback/testing would be appreciated. If nobody objects or has better ideas in the next ~24 hours, I'll merge this into main, and aim to get a bugfix build out with only this change about a day later.

### DIFF
--- a/qt/aqt/mediasrv.py
+++ b/qt/aqt/mediasrv.py
@@ -7,7 +7,9 @@ import enum
 import logging
 import mimetypes
 import os
+import random
 import re
+import string
 import sys
 import threading
 import traceback
@@ -698,7 +700,6 @@ def _extract_collection_post_request(path: str) -> DynamicRequest | NotFound:
 def _check_dynamic_request_permissions():
     if request.method == "GET":
         return
-    context = _extract_page_context()
 
     def warn() -> None:
         show_warning(
@@ -710,24 +711,16 @@ def _check_dynamic_request_permissions():
         aqt.mw.taskman.run_on_main(warn)
         abort(403)
 
-    if context in [
-        PageContext.NON_LEGACY_PAGE,
-        PageContext.EDITOR,
-        PageContext.ADDON_PAGE,
-        PageContext.DECK_OPTIONS,
-    ]:
-        pass
-    elif context == PageContext.REVIEWER and request.path in (
+    # does page have access to entire API?
+    if _have_api_access():
+        return
+
+    # whitelisted API endpoints for reviewer/previewer
+    if request.path in (
         "/_anki/getSchedulingStatesWithContext",
         "/_anki/setSchedulingStates",
         "/_anki/i18nResources",
     ):
-        # reviewer is only allowed to access custom study methods
-        pass
-    elif (
-        context == PageContext.PREVIEWER or context == PageContext.CARD_LAYOUT
-    ) and request.path == "/_anki/i18nResources":
-        # previewers are only allowed to access i18n resources
         pass
     else:
         # other legacy pages may contain third-party JS, so we do not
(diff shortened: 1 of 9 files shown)


qt/aqt/webview.py
```diff
@@ -28,35 +28,88 @@
...
         self.open_links_externally = True
 
-    def _setupBridge(self) -> None:
-        class Bridge(QObject):
-            def __init__(self, bridge_handler: Callable[[str], Any]) -> None:
-                super().__init__()
-                self.onCmd = bridge_handler
+    def _profileForPage(self, kind: AnkiWebViewKind) -> QWebEngineProfile:
+        have_api_access = kind in (
+            AnkiWebViewKind.DECK_OPTIONS,
```
iamllama, line 96:
```suggestion
            AnkiWebViewKind.MAIN,
            AnkiWebViewKind.DECK_OPTIONS,
```
The congrats screen fails to load otherwise:

image

abdnh:
One way I found to possibly gain API access in the reviewer is through a simple image tag, e.g. ``. example.com can simply read the `Authorization` header when the image is viewed in the editor and store it, then template scripts can request the token from the website and use it to fool Anki. Checking `info.requestUrl()` in `interceptRequest()` should guard against this.

qt/aqt/webview.py
iamllama: > AnkiWebViewKind.DECK_OPTIONS,

`IMPORT_LOG`, `IMPORT_CSV` and `IMPORT_ANKI_PACKAGE` should be whitelisted as well, for their respective import dialogs to work

abdnh (approved):

glutanimate (approved):
LGTM also from a general add-on compatibility standpoint

dae:
Thanks again all 🙏

dae:
https://github.com/ankitects/anki/releases/tag/25.02.1

qt/aqt/mediasrv.py
```diff
@@ -752,23 +746,11 @@ def legacy_page_data() -> Response:
...
-        query_params = parse_qs(referer.query)
-        query_id = query_params.get("id")
-        if not query_id:
-            return PageContext.UNKNOWN
-        id = int(query_id[0])
-        page_context = aqt.mw.mediaServer.get_page_context(id)
-        return page_context if page_context else PageContext.UNKNOWN
-    else:
-        return PageContext.UNKNOWN
+_APIKEY = "".join(random.choices(string.ascii_letters + string.digits, k=32))
```
nihil-admirari, line 749: Sorry for intruding, but it better be [`secrets.token_urlsafe`](https://docs.python.org/3/library/secrets.html#secrets.token_urlsafe). `random` is more or less predictable:
- https://blog.securityevaluators.com/hacking-the-javascript-lottery-80cc437e3b7f
- https://blog.securityevaluators.com/xorshift128-backward-ff3365dc0c17
- https://lemire.me/blog/2017/08/22/cracking-random-number-generators-xoroshiro128/
- https://github.com/deut-erium/RNGeesus

dae: Sorry, I lost track of this. This change will be in the next patch release.

PR #3928, opened 2025-04-17
Fix flashing when opening the stats, empty cards or find dupes dialogs

iamllama:
There's a slight visual regression (sorry!) after #3925 whereby the aforementioned dialogs flash more(?) after opening

`AnkiWebView.set_kind` now potentially [sets](https://github.com/ankitects/anki/blob/1a68c9f5d5bcc197b641fe7405e5d9a4823928f3/qt/aqt/webview.py#L401) a new page, ~~and it flashes despite setting its background colour [immediately after](https://github.com/ankitects/anki/blob/1a68c9f5d5bcc197b641fe7405e5d9a4823928f3/qt/aqt/webview.py#L365)~~. `set_kind` needs to be called for those dialogs because their forms are codegen'd from qt ui designer files, which, as far as i can tell from the [schema](https://doc.qt.io/qt-6.9/designer-ui-file-format.html), don't seem to allow passing in custom arguments directly to widget constructors

~~The fix proposed here is to replace the webview widgets in those ui files with subclasses that have the correct kind filled in, to avoid calling `set_kind` after init~~

~~After this, only the legacy deck stats would still call it (and use the hack in `set_kind`), as it shares the same form (and webview widget) with the "new" deck stats~~

EDIT: I've overthought this by overlooking that the page's background colour wasn't set again after `set_kind` sets a new page, adding it seems to have gotten rid of the flashes

### DIFF
--- a/qt/aqt/browser/find_duplicates.py
+++ b/qt/aqt/browser/find_duplicates.py
@@ -14,6 +14,7 @@ from anki.collection import SearchNode
 from anki.notes import NoteId
 from aqt.qt import *
 from aqt.qt import sip
+from aqt.webview import AnkiWebViewKind
 
 from ..operations import QueryOp
 from ..operations.tag import add_tags_to_notes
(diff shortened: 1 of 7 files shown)


dae:
Simple solutions are always nice! Thank you both for the quick fix.

PR #3933, opened 2025-04-19
Fix `AnkiWebPage` not being initialized for default web view kinds (e.g. in add-ons)

glutanimate:
#3925 and in particular the `domDone` race condition workaround in 24bca15fd3d9dc386916509eb2d4862d1184e709 unfortunately break add-ons that depend on default-kind `AnkiWebViews` initializing with a corresponding `AnkiWebPage`. This affects any add-ons that rely on `bridgeCommand` in add-on managed web views (for instance Contanki, [as reported on the forums](https://forums.ankiweb.net/t/add-ons-and-25-02-1-default-ankiwebview-api-access-workarounds/59309)).

Because in testing #3925 we happened to primarily test add-ons that either manage page creation manually or do not use `bridgeCommand`, we unfortunately missed this incompatibility.

This PR uses (with some minor alterations) @iamllama's web view subclasses proposed in the original version of #3928 to completely avoid setting the web view kind after web view initialization. Additionally, we fully drop `set_kind()` from the AnkiWebView API to discourage its use in either Anki-native code or add-ons.

(I mentioned in the corresponding commit that I found one add-on that was using `set_kind()`, but it turns out that this is no longer the case in current versions of the add-on, so dropping this method should not introduce any add-on breakages)

@iamllama I hope picking your commits here is OK. I initially looked into whether Qt designer supports some way to supply arguments to custom widgets, but quickly found that this is not possible ­­­­- or at least not in a way that would allow us to change things at widget construction time. So I thought back to your initial solution for the Qt forms web view flashing and picked that up.

### DIFF
--- a/qt/aqt/stats.py
+++ b/qt/aqt/stats.py
@@ -25,7 +25,7 @@ from aqt.utils import (
     tooltip,
     tr,
 )
-from aqt.webview import AnkiWebViewKind
+from aqt.webview import LegacyStatsWebView
 
 
 class NewDeckStats(QDialog):
@@ -153,6 +153,9 @@ class DeckStats(QDialog):
         self.name = "deckStats"
         self.period = 0
         self.form = aqt.forms.stats.Ui_Dialog()
+        # Hack: Switch out web views dynamically to avoid maintaining multiple
+        # Qt forms for different versions of the stats dialog.
+        self.form.web = LegacyStatsWebView(self.mw)
         self.oldPos = None
         self.wholeCollection = False
         self.setMinimumWidth(700)
@@ -231,7 +234,6 @@ class DeckStats(QDialog):
         stats = self.mw.col.stats()
         stats.wholeCollection = self.wholeCollection
         self.report = stats.report(type=self.period)
-        self.form.web.set_kind(AnkiWebViewKind.LEGACY_DECK_STATS)
         self.form.web.stdHtml(
             f"<html><body>{self.report}</body></html>",
             js=["js/vendor/jquery.min.js", "js/vendor/plot.js"],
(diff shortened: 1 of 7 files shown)


qt/aqt/stats.py
```diff
@@ -154,6 +153,9 @@ def __init__(self, mw: aqt.main.AnkiQt) -> None:
         self.name = "deckStats"
         self.period = 0
         self.form = aqt.forms.stats.Ui_Dialog()
+        # Hack: Switch out web views dynamically to avoid maintaining multiple
+        # Qt forms for different versions of the stats dialog.
+        self.form.web = LegacyStatsWebView(self.mw)
```
glutanimate, line 158: Constructing a web view twice is not ideal, but I couldn't think of a better solution without splitting Qt forms for the new and legacy stats dialogs in some way.

qt/aqt/webview.py
```diff
@@ -968,3 +961,53 @@ def on_body_classes_need_update(self) -> None:
     @deprecated(info="use theme_manager.qcolor() instead")
     def get_window_bg_color(self, night_mode: bool | None = None) -> QColor:
         return theme_manager.qcolor(colors.CANVAS)
+
+
+# Pre-configured classes for use in Qt Designer
```
glutanimate, line 966: Some of this might be YAGNI and over-engineered for the current use case, but given that `AnkiWebView` is the custom widget we use the most in Qt forms, I wanted to set things up in a way where - if we ever saw a need to supply other (new) construction time arguments - we could do so very easily.

iamllama:
> (I mentioned in the corresponding commit that I found one add-on that was using set_kind(), but it turns out that this is no longer the case in current versions of the add-on, so dropping this method should not introduce any add-on breakages)

It's possible that addons may have started using it after 25.02.1 (e.g. https://github.com/Kuuuube/kanjigrid/commit/208ff2b6fada473e0ad57367582ec18366f070e9), perhaps leaving it as-is with an added warning/deprecation notice would be preferable?

abdnh (approved):
Tested with a couple of add-ons.

dae:
Thanks all. I've created a new issue to track the related fallout: https://github.com/ankitects/anki/issues/3935

PR #3938, opened 2025-04-23
Persist tags when refetching csv metadata while importing

### LINKED ISSUE #3937: Tag disappears after typing tag then changing deck/note type in Text Import
> Problem
> 
> When I’m importing and I’ve already typed in a tag, changing the deck/note type seems to make the tag disappear after I imported it.
> Showcase
> 
> (i’m new user on the forum i can’t upload the vid i recorded on how to reproduce the steps)
> Steps to reproduce:
> 
> Open Anki and select Import File
> Select a file to import
> Fill in the tag blank
> Change the deck/note type
> Hit Import
> Expected: New card should have the typed in tag
> Reality: New card has no tags
> I did those steps with no add-ons enabled
> 
> Other info
> 
> Anki Version ⁨25.02.1 (234fa0c2)⁩
> System: Windows 11

Originally reported in https://forums.ankiweb.net/t/tag-disappears-after-typing-tag-then-changing-deck-note-type-in-import/59537

iamllama:
Fixes #3937

Tags are now persisted across metadata refetches

NB: refetching rereads the csv file, so if the `#tags:...` header line is edited across a refetch, the new global tags won't be reflected post-pr. But not only is this unlikely, the import dialog's tag inputs are already (pre-pr) broken in such a way that those changes won't be shown (but will still be applied). After this pr, what's reflected in the tags input will now consistently be what's applied to the imported notes

### DIFF
--- a/ts/routes/import-csv/lib.ts
+++ b/ts/routes/import-csv/lib.ts
@@ -99,6 +99,7 @@ export class ImportCsvState {
 
         const shouldRefetchMetadata = this.shouldRefetchMetadata(changed);
         if (shouldRefetchMetadata) {
+            const { globalTags, updatedTags } = changed;
             changed = await getCsvMetadata({
                 path: this.path,
                 delimiter: changed.delimiter,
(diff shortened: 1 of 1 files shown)


dae:
Nice!

PR #3940, opened 2025-04-23
Fix/Calculate missing memory states on simulate

Luc-Mcgrady:
Cards without memory states (likely due to being moved) are currently treated as new cards. This recalculates them.

Before:
image
After:
image

This does slow the simulation down with "ignore reviews before" because cards that will have no memory state even after calculation are processed every time.

### DIFF
--- a/rslib/src/scheduler/fsrs/simulator.rs
+++ b/rslib/src/scheduler/fsrs/simulator.rs
@@ -125,10 +125,13 @@ impl Collection {
             .get_revlog_entries_for_searched_cards_in_card_order()?;
         let mut cards = guard.col.storage.all_searched_cards()?;
         drop(guard);
-        for card in &mut cards {
-            if card.memory_state.is_none() {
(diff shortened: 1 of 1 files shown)


rslib/src/scheduler/fsrs/simulator.rs
```diff
@@ -123,8 +123,21 @@ impl Collection {
...
             .get_revlog_entries_for_searched_cards_in_card_order()?;
-        let cards = guard.col.storage.all_searched_cards()?;
+        let mut cards = guard.col.storage.all_searched_cards()?;
         drop(guard);
+        fn is_included_card(c: &Card) -> bool {
+            c.queue != CardQueue::Suspended
+                && c.queue != CardQueue::PreviewRepeat
+                && c.queue != CardQueue::New
+        }
+        for c in &mut cards {
```
dae, line 133:
```suggestion
        // calculate any missing memory state
        for c in &mut cards {
```

PR #3943, opened 2025-04-24
Fix fields with `\n` being ignored when searching all fields w/o regex

iamllama:
Reported in https://forums.ankiweb.net/t/potential-bug-newline-character-in-fields-html-disrupts-search-for-wildcard-fields/59617

`*:*somestring*` gets rewritten to `(?i)^.*somestring.*$`. But this doesn't match against a field with
```
<div>
  somestring
</div>
```
> Perhaps surprisingly, the default behavior for . is not to match any character, but rather, to match any character except for the line terminator (which is \n by default). When this mode is enabled, the behavior changes such that . truly matches any character.
https://docs.rs/regex/latest/regex/struct.RegexBuilder.html#method.dot_matches_new_line

The fix proposed is to add the `(?s)` flag that enables the abovementioned mode

### DIFF
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -555,7 +555,7 @@ impl SqlWriter<'_> {
     }
 
     fn write_all_fields(&mut self, val: &str) {
-        self.args.push(format!("(?i)^{}$", to_re(val)));
+        self.args.push(format!("(?is)^{}$", to_re(val)));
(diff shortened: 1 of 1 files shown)


iamllama:
This problem only happens because of the anchors ^ and $ that `write_all_fields` tacks on. Without them, newlines don't matter here as the match position isn't constrained anymore

`write_regex` and `write_single_field_regexp` do not add them automatically, so they aren't affected by this, unless a user explicitly uses anchors in a search. But adding them would perhaps make things more consistent too. Thoughts?

PR #3945, opened 2025-04-24
Add a way to pass information from browser_will_search to browser_did_search without having it going to the backend

JSchoreels:
Use case :

0. Dev Wants to implement a custom search filter, let's call it leeches:active, not related to a backend column, and not easily mappable to an existing field
1. User type : "deck:current **leeches:active**"" 
2. browser_will_search receive it, remove the "leeches:active", set it in context.properties
3. The backend execute the search "deck:current""
4. browser_did_search receive the result of the previous result, and also the "leeches:active" in context.properties to trigger the filtering

Why this solution ?
- If you let the leeches:active go to the backend, you'll get 0 result because it's not an existing property
- If you let the "browser_will_search" do the find itself, and have multiple anki addons hooking into it, it's basically a race between addons to know which one will have the right to do the find
- With multiple addon, the "leeches:active" in the search, might be passed in another addon search, which mean the addon will have 0 results.
- With multiple addon, if I set context.ids, I might disable others addon behaviour, while in this case, my use case is more to be able to do some kind of post-processing filtering.
- With this solution, every addon can add some info in the context.properties, that will stay hidden to the actual backend, and the browser_did_search hook will be able to trigger the corresponding behaviour (filtering, formatting) based on what the request was initially.
- Allowing this would enable more custom filter searches to be implemented externally without requiring the user to change the fields of his notes, or to have them tagged/flagged. It would enable addons devs to build experimental filters without impacting too much the backend side


dae:
Happy to merge this in, with one suggestion: maybe a name like 'addon_metadata' or something similar would better convey the purpose of the attribute than 'properties' does?

JSchoreels:
Make sense, I've adapted the property name !

dae:
Thanks!

PR #3946, opened 2025-04-25
Update to FSRS-rs v3.0.0 (fix historic memory state)

L-M-Sherlock:
It includes this patch: 
- https://github.com/open-spaced-repetition/fsrs-rs/pull/320

It fixed this bug:

> Currently, in Anki, A card with a review history completely ignored by ignore_reviews_before except for 1 review panics with
> 
> ```
> thread '<unnamed>' panicked at rslib/src/stats/card.rs:159:40:
> attempt to subtract with overflow
> ```
> 
> I presume this is the fix for that? We could also fix it in Anki with:
> 
> ```rust
>             let mut memory_states =
>                 fsrs.historical_memory_states(item.item, item.starting_state)?;
>             if let Some(starting_state) = item.starting_state {
>                 memory_states.insert(0, starting_state);
>             }
> ```
> 
> I'm not sure if it's best to fix it like this or let the programmer decide if they want to insert the starting state themselves manually like above?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2100,7 +2100,8 @@ dependencies = [
 [[package]]
 name = "fsrs"
 version = "3.0.0"
-source = "git+https://github.com/open-spaced-repetition/fsrs-rs.git?rev=c7717682997a8a6d53d97c7196281e745c5b3c8e#c7717682997a8a6d53d97c7196281e745c5b3c8e"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e7e39017be91629761c3a2802032eb38d226227e1a21433381da4310612199ee"
 dependencies = [
  "burn",
  "itertools 0.14.0",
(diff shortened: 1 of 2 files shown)


L-M-Sherlock:
@dae I plan to remove "Compute minimum recommended retention" because it always output 0.70 with FSRS-6. Maybe @Luc-Mcgrady's solution works: https://github.com/Luc-Mcgrady/anki/tree/cmmr-uses-simulate-config, but he hasn't submitted it.

dae:
No worries. @Luc-Mcgrady, should we remove it now or wait for you?

Luc-Mcgrady:
> @Luc-Mcgrady's solution works: https://github.com/Luc-Mcgrady/anki/tree/cmmr-uses-simulate-config, but he hasn't submitted it.

My testing wasn't very thorough but I can PR it.

If you remove it for the time being, and we then find some way to make it work, it can always be added back without too much fuss I imagine?

dae:
Yep

PR #3947, opened 2025-04-25
Feat/CMRR uses simulate config

Luc-Mcgrady:
Should hopefuly make CMRR more accurate by using the config provided by SimulatorConfig rather than a generic one.

image

Moves the CMRR button into the simulate modal.

cc @L-M-Sherlock @Expertium


Expertium:
[Forgetting curve integral.zip](https://github.com/user-attachments/files/19908739/Forgetting.curve.integral.zip)
Here's example code for calculating the average retrievability over time. I want you to try sum(R over the next 5 eyars) instead of sum(R now)
From my word file:

> 	Instead of using sum(R), we will use sum(avg_R(t1, t2)), the sum of average values of R over some future period of time. I suggest 5 years. Specifically, I mean “5 years from today, where today is whatever is the last day of the simulation”. In Python, it looks like this:
```python
sum_of_avg_r_over_a_year[today] = average_f_power_forgetting_curve(card_table[col["delta_t"]], card_table[col["delta_t"]] + 365*5,card_table[col["stability"]], DECAY).sum()
```

dae:
Have you guys reached a consensus on whether we're improving this or [removing it](https://github.com/ankitects/anki-manual/pull/372/files)?

Expertium:
Not yet. Another user suggested a different approach, so let's try that.
@Luc-Mcgrady instead of using sum(R), let's use sum(S*R). It brings the recommended value up, like we want, and it has an interpretation: sum of memory stabilities after accounting for the fact that you won't be able to recall all of your cards, and stabilities if forgotten cards will drop.

dae:
The "compute optimal retention" dropdown is not aligned with the dropdowns above, and looks a bit messy - would be nice to fix in a future update if this is going to stay. I noticed you dropped the progress handling, and it seems to run faster now, though I'm only seeing 0.7 in this test deck. Let's see how it goes for people.

Expertium:
> I'm only seeing 0.7 in this test deck

That's why Jarrett wanted to remove it. But another user came up with a better minimization criterion, so hopefully Luc will add that later.

Issue #3841, opened 2025-03-05
Make line-break possible in Stats/Answer buttons


sommerluk:
It would be great to make line-breaks possible in Stats/Answer buttons for the strings "Learning (1,23%)", "Young (1,23%)", "Mature (1,23%)". See here:

![Image](https://github.com/user-attachments/assets/e38ceb2f-b18c-4c78-a251-3a707c0bd65c)

Doing so would allow translators to use longer strings without that the letters of neighbour strings flow over each other.

soleuniverse101:
So you have an example of language which would need more space ? I was able to line wrap the text, but it only works on words separated by spaces. Do you think it would be necessary to cut words if they are too long (e.g. adding a hyphen at the end and finishing the word at the next line), or if the language does not use spaces (like Japanese) ?

soleuniverse101:
Was it the translation "Wiederholung (instabil)" for "Young" ? (https://i18n.ankiweb.net/de/core/core/templates/statistics.ftl/?search=statistics-counts-young&string=23480)

I tested it locally, here's the result :

![Image](https://github.com/user-attachments/assets/89679f04-eed4-4e8f-be0f-4a540a4dd1ec)

soleuniverse101:
Well as long as the words are not too long and are separated by spaces, the fix I have locally should work I think. Unfortunately, for languages where this would translate to long words without spaces, it's difficult to provide a solution because wrapping behavior can depend on the language (e.g. cut between syllables).

I'll link a PR with what I have at the moment

sommerluk:
![Image](https://github.com/user-attachments/assets/4de66b5f-4e3c-474c-9c5d-ff49f029d03b)

sommerluk:
I don't know Japanese. But for text in Latin scripts, in Anki the CSS "word-break" properry should be either "normal" or "break-word", and in my opinion the latter would be the better trade-of.

soleuniverse101:
> I don't know Japanese. But for text in Latin scripts, in Anki the CSS "word-break" properry should be either "normal" or "break-word", and in my opinion the latter would be the better trade-of.

Yes I know about it, unfortunately you can't use it here because the graph is created using SVG (through the d3 library). Apparently it's a known issue, not being able to wrap text automatically, so the solution I found is a crazy long function of someone with the same problem (https://gist.github.com/mbostock/7555321). It uses <tspan> elements inside the <text> element that contains the labels to manually shift their height when they are too long.

[...]

PR #3950, opened 2025-04-25
Line break before percentage

sommerluk:
This is not tested – I'm apparently unable to build Anki locally.

The idea is to introduce a line break in the statistics in the "Answer buttons" section, before the percent value. Currently, the horizontal space is very restricted, and the percentage value restricts it further. Putting it in a new line would give more horizontal space for translators. And I think it looks better even when the horizontal space is not required. While #3841 would not be actually solved by this pull request, it would be a good-enough solution for most translations.

PS: Which is the correct new line? Is \u000a okay?

### DIFF
--- a/ts/routes/graphs/buttons.ts
+++ b/ts/routes/graphs/buttons.ts
@@ -148,7 +148,7 @@ export function renderButtons(
                                     kind = tr.statisticsCountsMatureCards();
                                     break;
                             }
-                            return `${kind} \u200e(${totalCorrect(d).percent}%)`;
+                            return `${kind}\u000a\u200e(${totalCorrect(d).percent}%)`;
                         }) as any,
                     )
                     .tickSizeOuter(0),


dae:
This doesn't work. Please report your build issues on the forums, as you should be testing locally before submitting changes.

sommerluk:
Okay. It took some time, but I have a running build now. (Apparently, spaces in folder names in your path are problematic.)

@GithubAnon0000 Indeed the line break stuff does not work. I tested it now also, same result.

I've also read about `<tspan>` now, but introducing it literally into the string will render it literally. I'm not experienced in Typescript or SVG, and I've asked AI, which gave me code that manipulates the DOM directly, but I was not able to get it working.

Unfortunately, I fear I can't help find a solution in this case.

sommerluk:
Closing in favor of #3952

PR #3951, opened 2025-04-25
Fix/No memory state cards have revlogs in reverse order on card stats screen.

Luc-Mcgrady:
Currently cards without a memory state have their revlogs appear descending:
(bug, note the dates descending)
image


dae:
Thank you for the quick fix!

PR #3952, opened 2025-04-26
REMOVE percentages of x-axis in the answer buttons graph

GithubAnon0000:
Fixes: #3841 
Makes #3950 obsolete.

<details><summary>Old description</summary>
<p>

This seems to work, though I'm not sure if that's the right way to go about things. `d3` is like black magic to me, and being a novice at programming certainly doesn't help either.

I coded a line like this:
```ts
tspan2.setAttribute("dx", "-4em"); // i realized it works. It's probably a coincidence and a hack
```
The `-4em` part was retrieved by trial and error and is probably a hack. It does work perfectly fine on my machine, though.

Here's how the graph looks like with this PR:
anki

@sommerluk maybe you could build and try this?

</p>
</details> 

This PR breaks the percentage into a separate line (for the answer buttons graph only). This might improve translations, as longer strings can now be used correctly.

The transition has been removed from the graph. Not sure why that works, but @iamllama kindly pointed this out (see [forum post](https://forums.ankiweb.net/t/i-need-help-with-my-pr-splitting-the-x-axis-of-svg-d3-into-two-separate-lines-cuts-off-text-and-is-misaligned/59819/2?u=anon_0000)).

Here's what the graph with the current PR looks like:
_English_
anki
_German_
anki

### DIFF
--- a/ts/routes/graphs/buttons.ts
+++ b/ts/routes/graphs/buttons.ts
@@ -148,7 +148,6 @@ export function renderButtons(
                                     kind = tr.statisticsCountsMatureCards();
                                     break;
                             }
-                            console.log("Tick format for", d, "returns", kind);
                             return `${kind}`;
                         }) as any,
                     )
@@ -163,7 +162,6 @@ export function renderButtons(
                 .each(function(d) {
                     const current_text_element = this; 
                     const current_text_element_content = current_text_element.textContent;
-                    console.log("Current text content:", current_text_element_content);
 
                     current_text_element.textContent = "";
 
--- a/ts/routes/graphs/buttons.ts
+++ b/ts/routes/graphs/buttons.ts
@@ -156,30 +156,6 @@ export function renderButtons(
         )
         .attr("direction", "ltr");
 
-    svg.select<SVGGElement>(".x-ticks").selectAll<SVGTextElement, GroupKind>("text")
-        .each(function(this: SVGTextElement, d: GroupKind) {
-            if (!(this instanceof SVGElement)) {
-                return;
-            }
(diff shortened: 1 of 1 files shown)


sommerluk:
Sure. I do not really understand how the code works, but I can confirm it compiles. Here's the result:

grafik

The alignment is however different from your screenshot: Horizontally, it is right-aligned and not centred. And vertically, the lowest line is partially cut off.

dae:
Text still seems truncated to me:
Screenshot 2025-04-28 at 7 05 37 am

dae:
Perhaps an alternative solution is to move some of the text in those labels to a tooltip instead?

dae:
Yep, with the latest commit. Could be differences in fonts across platforms (I'm testing on macOS).

sommerluk:
The cutoff might be related to the size of the statistics page window. When the window is resized, we observe different cutoffs:

grafik

grafik

grafik

dae:
I think we'd have to completely rewrite the graph to be able to use d3plus, and it would mean it doesn't have the same look & feel as the others. That seems excessive just to address this issue. Why don't we just move the second line's content into a tooltip, or a text summary below, etc?

sommerluk:
Yes, and perhaps the best solution would be to simply remove the percentage value entirely from the x-axis.

1. I always found the percentage value at this position confusing, given that without context it's not obvious what it means.
2. The percentage value _is yet_ available in the tooltip of the bars (and there more understandable thanks to more context). So we would not even need to add it.

dae:
Re: no table, we might get some pushback from users who want to see the info at a glance. Let's see how we go.

Thank you for your patience!

PR #3953, opened 2025-04-26
add toggle for browser sidebar

JL710:
This PR enables the user to toggle the sidebar of the browser. 

The use case is for usage in split screen editing. If you have anki on a device like a laptop and editing notes while looking at another document, you often have anki only on one half of the screen. In that scenario the sidebar takes a lot of space. This PR enables you to toggle/hide it.


chrislongros:
Looks nice !!

JL710:
@GithubAnon0000 thanks for pointing that out. I did not know that.

dae:
Looks good. Thanks!

PR #3949, opened 2025-04-25
Update sorting by R for FSRS 6

user1823:

PR #3954, opened 2025-04-27
Fix/incorrect fallback of decay

L-M-Sherlock:
fix https://github.com/ankitects/anki/pull/3949#issuecomment-2832793960


JSchoreels:
Tested and look all good !

PR #3957, opened 2025-04-27
Fix mathjax images erroneously appearing enlarged

iamllama:
Reported in https://forums.ankiweb.net/t/anki-25-05-beta/59710/28

The `min-height` property added in https://github.com/ankitects/anki/commit/e7fbf159a67f8e73c21f871b8329f9275671aa80 wasn't specific enough and affected mathjax images. The fix proposed is to move it into the `anki-editable` scope (`min-height` is not inherited)


dae:
Thank you!

PR #3958, opened 2025-04-28
 Remove dead code in reviews_for_fsrs

mbrubeck:
The code removed here is never executed, because an earlier check for `first_of_last_learn_entries` would always break out of the loop before this check could succeed.  The code is also unnecessary because when `first_of_last_learn_entries` is set, `revlogs_complete` is already set to `true`.


mbrubeck:
Sorry, my original patch and description were not quite right.  Updated the PR with a better fix and description.

mbrubeck:
Heh, that’s actually the code change I originally submitted, but I mistakenly convinced myself it was wrong.  I’m happy to roll back and make that change instead.

user1823 (approved):

dae:
Thank you both!

PR #3959, opened 2025-04-28
Don't use intraday relearning step as starting memory state

mbrubeck:
When a filtered revlog history begins with a failed review that starts a sequence of relearning steps, use the last review in the sequence of relearning steps (rather than the first) for the starting memory state.

Bug report thread:

https://forums.ankiweb.net/t/fsrs-stability-when-first-non-ignored-revlog-has-a-relearning-interval/59894

### DIFF
--- a/rslib/src/scheduler/fsrs/params.rs
+++ b/rslib/src/scheduler/fsrs/params.rs
@@ -313,9 +313,12 @@ pub(crate) fn reviews_for_fsrs(
         if entry.review_kind == RevlogReviewKind::Filtered && entry.ease_factor == 0 {
             continue;
         }
+        // For incomplete review histories, initial memory state is based on the first
+        // user-graded review after the cutoff date with interval >= 1d.
         let within_cutoff = entry.id.0 > ignore_revlogs_before.0;
         let user_graded = matches!(entry.button_chosen, 1..=4);
-        if user_graded && within_cutoff {
+        let interday = entry.interval >= 1 || entry.interval <= -86400;
+        if user_graded && within_cutoff && interday {
             first_user_grade_idx = Some(index);
         }
 
@@ -477,6 +480,7 @@ pub(crate) mod tests {
             review_kind,
             id: days_ago_ms(days_ago).into(),
             button_chosen: 3,
+            interval: 1,
             ..Default::default()
         }
     }
@@ -709,6 +713,28 @@ pub(crate) mod tests {
         // L = learning step
         // L | L R
         assert_eq!(convert_ignore_before(revlogs, true, days_ago_ms(9)), None);
+    }
+
+    #[test]
+    fn skip_initial_relearning_steps() {
+        let revlogs = &[
+            revlog(RevlogReviewKind::Review, 10),
+            RevlogEntry {
+                button_chosen: 1, // Again
+                interval: -600,
+                ..revlog(RevlogReviewKind::Review, 8)
+            },
+            revlog(RevlogReviewKind::Relearning, 8),
+            revlog(RevlogReviewKind::Review, 6),
+        ];
+        // | = Ignore before
+        // A = Again
+        // X = Relearning
+        // R | A X R
+        assert_eq!(
+            convert_ignore_before(revlogs, false, days_ago_ms(9)),
+            fsrs_items!([review(0)], [review(0), review(2)])
+        );
     }
 
     #[test]


user1823:
I believe that we should instead use the first step that has an interval ≥ 1d.

Learning steps in the following code should also get the same treatment:

>         // While reviewing, if the first learning step is before the ignore date,
>         // we ignore it, and will fall back on SM2 info and the last user grade below.

That is, we should use `memory_state_from_sm2` on the first step with interval ≥ 1d after the ignore date.

user1823:
> Even if a relearning step is ≥ 1d, using it as the starting memory state will still throw out all stability data from the card history, since it will always use the relearning step as the starting memory state

You are right that with my suggestion, a user with multi-day steps would get their *initial* memory states from the steps they have configured, disregarding all the past history. But, this still seems to be a better choice than skipping all the steps because FSRS would then be able to use the ratings given during the relearning stage for refining its future predictions.

> This seems just as wrong with a 1d relearning step as with a 10m one.

The main disadvantage of using a 10 min step with memory_state_from_sm2 is that the card will get a very high difficulty. The memory_state_from_sm2 formula is not well suited for very small intervals.

> instead of the first review after the cutoff, we should get the SM2 state from the last review before the cutoff

This seems to be a nice idea. But, we would need to deal with some edge cases. One of them is the case where the user pressed Again before the cutoff and then the next rating was after the cutoff. Here, we can't use the interval of the Again rating because it would be equal to the first relearning step.

> For learning steps, I think we effectively already do this, right? 

Your arguments make sense.

mbrubeck:
Just pushed some minor cleanup to my test code.

I am happy to change this to skip initial < 1d reviews instead (as suggested by @user1823), but will give other reviewers a chance to weigh in before I make any substantive changes.

I'll experiment in another branch with my other suggested changes to the cutoff logic (i.e. using the last pre-cutoff review for the memory state), and can post that as a separate PR later if it works out well.

mbrubeck:
> I believe that we should instead use the first step that has an interval ≥ 1d.

I updated my patch to implement this suggestion.

rslib/src/scheduler/fsrs/params.rs
```diff
@@ -313,9 +313,11 @@ pub(crate) fn reviews_for_fsrs(
         if entry.review_kind == RevlogReviewKind::Filtered && entry.ease_factor == 0 {
             continue;
         }
+        // For incomplete review histories, initial memory state is based on the first
+        // user-graded review after the cutoff date with interval >= 1d.
         let within_cutoff = entry.id.0 > ignore_revlogs_before.0;
         let user_graded = matches!(entry.button_chosen, 1..=4);
-        if user_graded && within_cutoff {
+        if user_graded && within_cutoff && entry.interval >= 1 {
```
user1823, line 320: Will this be true for interday learning steps that were scheduled by the v2 scheduler?

- In the v3 scheduler, interval is stored in days for interday learning and relearning steps.
- In the v2 scheduler, it was stored in negative seconds. (e.g. as -86400 for a 1d learning step)

dae: What would we gain by doing so?

user1823: Is your comment directed at my request relating to check database? By doing so, we wouldn't need special handling for reviews that were scheduled by the v2 scheduler.

For example, see the special handling in https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/d9b0e0150d7ea26bfa1d46b1c8be61ec32bd0827/stats.py#L393

mbrubeck: Fixed to account for intervals < -86400.

dae: Given that our code already accounts for this case, and there are a lot of collections with the old format, I don't think we'd gain much in the short term. It would remove a footgun for PRs like this, but would force a one-way sync for all affected users. And if the old code paths were updated to remove the old case, they'd no longer work properly for some users until they check the DB, and we could accidentally break something when making the changes.

dae:
Thank you both!

rslib/src/scheduler/fsrs/params.rs
user1823: We can make check DB update the entries without updating the old code paths. This would ensure that they continue to work properly even if the users don't check the DB. Also, one-way sync is not required as check DB on each device would update the entries stored there.

This won't have any short term benefit, but after a long time, we would be able to simplify the codebase, assuming that most people would have used check DB atleast once till then.

Issue #3926, opened 2025-04-15
Improve feedback from "Evaluate" and potentially hide/move it (FSRS section)


Expertium:
![Image](https://github.com/user-attachments/assets/3d0bf50d-9540-44ce-9f33-e132b8a1f186)

Dae, despite what the screenshot above shows, I think we should disregard that poll and remove "Evaluate" anyway. David agrees, btw.

![Image](https://github.com/user-attachments/assets/68e30dc9-a996-4800-982c-e3f6573761fc)

1) "Evaluate" gives the user a bunch of numbers that 99.9% of users don't know how to interpret. A person on r/Anki even thought that log-loss means that his review history has been lost. And btw, Anki uses a different procedure (no splitting into train/test sets) compared to the [SRS benchmark repo](https://github.com/open-spaced-repetition/srs-benchmark), so Anki numbers aren't even directly comparable to benchmark numbers; the latter will be worse, on average.
2) The numbers don't provide any actionable information. They don't tell the user whether he is using the buttons correctly, they don't tell him whether there is a problem with how he does his reviews, or anything of that kind.
3) The numbers aren't even useful for *other* people - like me and Danika - who help users. The FSRS parameters field is at least somewhat useful: you can tell whether the user is using default parameters (which could be because he doesn't realize what "Optimize" is for) or optimized parameters, and sometimes you can spot abnormalities in optimized parameters. "Evaluate" doesn't help with any of that. Even in some rare cases where I can clearly tell that the metrics are really bad, they don't tell me what the cause is. If I see someone with a log-loss of 3.5 or something, I won't be able to tell what happened. Does this user bash the keyboard randomly when selecting the answer button? Did he import cards from somewhere else, and their history got mangled? Who knows.
I think over more than a year of helping on r/Anki, "Evaluate" came in handy, like, once.

(Just to be clear, I am NOT proposing removing the parameters field)

As much as I like deciding based on the opinion of the majority, I think in this case the opinion of the majority went astray. I'll conduct a survey on r/Anki to collect more feedback on "Evaluate".

Moving "Evaluate" somewhere else is nice in theory, but I can't think of a good implementation. If we put it in "Advanced", it will be awkward (scrolling back and forth between sections) and unclear that this button is related to FSRS in the first place, unless it says "Evaluate FSRS parameters" or something. Maybe we could collapse it, but again, I can't visualize a good implementation.

[...]

PR #3962, opened 2025-04-28
Feat/evaluate FSRS with time series split

L-M-Sherlock:
related:
- https://github.com/open-spaced-repetition/fsrs-rs/pull/326
- https://github.com/ankitects/anki/issues/3926#issuecomment-2833652356

### DIFF
--- a/rslib/src/scheduler/fsrs/params.rs
+++ b/rslib/src/scheduler/fsrs/params.rs
@@ -115,15 +115,14 @@ impl Collection {
             num_relearning_steps: Some(num_of_relearning_steps),
         })?;
         progress_thread.join().ok();
-        if let Ok(fsrs) = FSRS::new(Some(current_params)) {
-            let current_log_loss = fsrs.evaluate(items.clone(), |_| true)?.log_loss;
+        if let Ok(current_fsrs) = FSRS::new(Some(current_params)) {
+            let current_log_loss = current_fsrs.evaluate(items.clone(), |_| true)?.log_loss;
             let optimized_fsrs = FSRS::new(Some(&params))?;
             let optimized_log_loss = optimized_fsrs.evaluate(items.clone(), |_| true)?.log_loss;
             if current_log_loss <= optimized_log_loss {
                 if num_of_relearning_steps <= 1 {
                     params = current_params.to_vec();
                 } else {
-                    let current_fsrs = FSRS::new(Some(current_params))?;
                     let memory_state = MemoryState {
                         stability: 1.0,
                         difficulty: 1.0,
@@ -218,22 +217,24 @@ impl Collection {
 
     pub fn evaluate_params(
         &mut self,
-        params: &Params,
         search: &str,
         ignore_revlogs_before: TimestampMillis,
+        num_of_relearning_steps: usize,
     ) -> Result<ModelEvaluation> {
         let timing = self.timing_today()?;
-        let mut anki_progress = self.new_progress_handler::<ComputeParamsProgress>();
-        let guard = self.search_cards_into_table(search, SortMode::NoOrder)?;
-        let revlogs: Vec<RevlogEntry> = guard
-            .col
-            .storage
-            .get_revlog_entries_for_searched_cards_in_card_order()?;
+        let revlogs = self.revlog_for_srs(search)?;
         let (items, review_count) =
             fsrs_items_for_training(revlogs, timing.next_day_at, ignore_revlogs_before);
+        let mut anki_progress = self.new_progress_handler::<ComputeParamsProgress>();
         anki_progress.state.reviews = review_count as u32;
-        let fsrs = FSRS::new(Some(params))?;
-        Ok(fsrs.evaluate(items, |ip| {
+        let fsrs = FSRS::new(None)?;
+        let input = ComputeParametersInput {
+            train_set: items.clone(),
+            progress: None,
+            enable_short_term: true,
+            num_relearning_steps: Some(num_of_relearning_steps),
+        };
+        Ok(fsrs.evaluate_with_time_series_splits(input, |ip| {
             anki_progress
                 .update(false, |p| {
                     p.total_iterations = ip.total as u32;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2295,7 +2295,7 @@ dependencies = [
 [[package]]
 name = "fsrs"
 version = "4.0.0"
-source = "git+https://github.com/open-spaced-repetition/fsrs-rs.git?rev=092c20bac7d9239a991ae5b561556ad34c706c16#092c20bac7d9239a991ae5b561556ad34c706c16"
+source = "git+https://github.com/open-spaced-repetition/fsrs-rs.git?rev=b6b1ab87aaf8e25b72ad75ff763b6cd609f12640#b6b1ab87aaf8e25b72ad75ff763b6cd609f12640"
 dependencies = [
  "burn",
  "itertools 0.14.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ rev = "184b2ca50ed39ca43da13f0b830a463861adb9ca"
 [workspace.dependencies.fsrs]
 # version = "3.0.0"
 git = "https://github.com/open-spaced-repetition/fsrs-rs.git"
-rev = "092c20bac7d9239a991ae5b561556ad34c706c16"
+rev = "b6b1ab87aaf8e25b72ad75ff763b6cd609f12640"
 # path = "../open-spaced-repetition/fsrs-rs"
 
 [workspace.dependencies]
--- a/proto/anki/scheduler.proto
+++ b/proto/anki/scheduler.proto
@@ -435,9 +435,9 @@ message GetOptimalRetentionParametersResponse {
 }
 
 message EvaluateParamsRequest {
-  repeated float params = 1;
-  string search = 2;
-  int64 ignore_revlogs_before_ms = 3;
+  string search = 1;
+  int64 ignore_revlogs_before_ms = 2;
+  uint32 num_of_relearning_steps = 3;
 }
 
 message EvaluateParamsResponse {
--- a/rslib/src/scheduler/service/mod.rs
+++ b/rslib/src/scheduler/service/mod.rs
@@ -295,9 +295,9 @@ impl crate::services::SchedulerService for Collection {
         input: scheduler::EvaluateParamsRequest,
     ) -> Result<scheduler::EvaluateParamsResponse> {
         let ret = self.evaluate_params(
-            &input.params,
             &input.search,
             input.ignore_revlogs_before_ms.into(),
+            input.num_of_relearning_steps as usize,
         )?;
         Ok(scheduler::EvaluateParamsResponse {
             log_loss: ret.log_loss,
--- a/ts/routes/deck-options/FsrsOptions.svelte
+++ b/ts/routes/deck-options/FsrsOptions.svelte
@@ -180,9 +180,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                         ? $config.paramSearch
                         : defaultparamSearch;
                     const resp = await evaluateParams({
-                        params: fsrsParams($config),
                         search,
                         ignoreRevlogsBeforeMs: getIgnoreRevlogsBeforeMs(),
+                        numOfRelearningSteps: $config.relearnSteps.length,
                     });
                     if (computeParamsProgress) {
                         computeParamsProgress.current = computeParamsProgress.total;


user1823:
Just to note: After this change, we are no longer evaluating the parameters present in the field. We are now evaluating the ability of the FSRS optimizer to produce such parameters from the user data that can reliably predict the R of the reviews that were not used to train the parameters (i.e., the ability to generalize).

Expertium:
@L-M-Sherlock I'm running FSRS-6 with train = test (see https://github.com/open-spaced-repetition/srs-benchmark/pull/210)
Once I'm done, I'll have data that is comparable to what users see in Anki. So this PR won't be necessary, we can keep the current behavior of Evaluate where it evaluates a specific set of parameters rather than FSRS's ability to adapt. Then again, it's up to you. If you want to proceed with this PR, ok. Just saying that we can do it either way, and it's less confusing for the average user if we keep the current behavior.

dae:
The feature seems to be somewhat less useful if it's not considering the provided parameters, as it no longer offers any insight into how a given set of parameters may differ in performance from FSRS's calculated optimum parameters. I think users might prefer to get a figure that indicates how FSRS will perform with their given settings, rather than how FSRS might perform with ideal parameters?

Expertium:
@dae It's not quite that.
With Jarrett's implementation, we are not evaluating a specific set of parameters to begin with. Rather, we are evaluating how well FSRS generalizes to unseen data that it was not trained on. We are evaluating the *algorithm itself*, not *specific parameters*. It's quite different conceptually.

Regarding being useful, it has an advantage over the current implementation - it will discourage people from tweaking parameters to try to get better metrics. Of course, there will be a "The results of Evaluate do not depend on your current FSRS parameters" warning. Though, initially, people will be confused.

JSchoreels:
> We are evaluating the algorithm itself, not specific parameters. 

Hmm this is not true. You're evaluating the parameters of the optimization on a test set. You could evaluate any kind of parameters with the same procedure. It's basically

`
evaluate(optimize(training_set).parameters, test_set)
VS
evaluate(user_defined.parameters, test_set)
`

Maybe being sure we evaluate the user_defined.parameters on the test_set is not that easy, but depending on how we partition training_set and test_set, if we use some `card.id mod N` shouldn't be really difficult.

Expertium:
@1DWalker your opinion is welcome

Unfortunately, it seems like we will be discussing things for many hours and writing lots of comments without agreeing on anything. Again.

rslib/src/scheduler/fsrs/params.rs
```diff
@@ -218,22 +217,24 @@ impl Collection {
...
+        let revlogs = self.revlog_for_srs(search)?;
         let (items, review_count) =
             fsrs_items_for_training(revlogs, timing.next_day_at, ignore_revlogs_before);
+        let mut anki_progress = self.new_progress_handler::<ComputeParamsProgress>();
         anki_progress.state.reviews = review_count as u32;
-        let fsrs = FSRS::new(Some(params))?;
-        Ok(fsrs.evaluate(items, |ip| {
+        let fsrs = FSRS::new(None)?;
+        let input = ComputeParametersInput {
+            train_set: items.clone(),
```
JSchoreels, line 232: I see the `train_set` is `items.clone()`, but `items` since to still be defined by `let (items, review_count) =
            fsrs_items_for_training(revlogs, timing.next_day_at, ignore_revlogs_before);` which didn't change.

It feels like the training items would still be the same items than before ? I don't really see where the `training_set = whole_set \ test_set occurs ?

JSchoreels:
> The thing about the 5-way split approach is that we are evaluating 5 different sets of parameters obtained on 5 different splits to get a general idea of how well FSRS performs on unseen data **overall**, rather than with just one set of parameters on just one set of data, aka a **specific case**.

Could you show where those 5 different sets of parameters are generated ? I checked the PR code but I can't figure it out.

Also, I guess each of those 5 different sets of parameters are evaluated on their respective test set, right ?

So the `evaluate(parameters, test_set)` would still be compatible for that scenario, since each "parameters" could either be : One of those 5-split, the final parameters, or an user defined parameters.

But I don't seem to find any of that logic in the PR code ?

Expertium:
> Could you show where those 5 different sets of parameters are generated ? I checked the PR code but I can't figure it out.

https://github.com/open-spaced-repetition/fsrs-rs/pull/326/commits/0711eada7eed22ccae2cc2559c0b0c73c8fd72fc
image

It's technically not in this PR

> Also, I guess each of those 5 different sets of parameters are evaluated on their respective test set, right ?

Yes

image

rslib/src/scheduler/fsrs/params.rs
JSchoreels: Understood thanks to this comment https://github.com/ankitects/anki/pull/3962#issuecomment-2844712511

JSchoreels:
Ah I see, the diff in the cargo.toml is what will use this new logic.

Well, on my end I think it could be worth it to have a way to have that evaluate(parameters), even if it might mean run the user_defined.parameters on all of those test set and doing the same logic that is done for FSRS parameters (Taking the min ? an average of all those sub evaluate ?)

Of course, I also understand it can take precious time we might not have, but I think having some kind of easy way to see how bad some homemade tweak could hurt the predictions could be useful.

dae:
I'm not sure yet what the best option is here. I do appreciate how testing with the same training data is bad data science, but a feature that can only tell the user "this is how good/bad FSRS's optimized settings will perform on your collection' doesn't seem as useful, as it offers no insight into how the current parameters may differ from optimal, which makes it less useful for support, and may lead the user or support to believe that FSRS is performing better than it actually is, if their parameters are bad.

I'm not a data scientist, so the following idea may not work. But what if the training process used reproducible splits (eg card id % 10 < 2 for the test group), and evaluate used those same splits? Presumably that would give us the train/test separation while still factoring in the current params? Or are there downsides to such an approach that I've overlooked?

Expertium:
Yes, that was another proposed approach. It's somewhat inconvenient though, because @L-M-Sherlock will have to implement the card id % 10 < 2 approach both in Anki and in the benchmark, as opposed to implementing something *only* in Anki or *only* in the benchmark.
But yeah, I have no objections other than that.

user1823:
> if the training process used reproducible splits

If we use splits when optimizing the parameters, we are discarding some of the data, which will make the parameters slightly less accurate. (The more the data available for optimization, the better would be the outcome.)
I don't think it makes sense to make optimization worse for making evaluation more realistic.

dae:
That's a fair argument too, and perhaps an answer to "this is bad data science" → "the alternative is worse training results"

JSchoreels:
> ill make the parameters slightly less accurate. (The more the data available for optimization, the better would be the outcome.)
> I don't think it makes sense to make optimization worse for making evaluation more realistic.

Hmmm the point is that by sacrificing 20% of your review to become a "test set", you won't really lose 20% of accuracy. While, having no training/test split, might.

If let's say tomorrow Anki has a new scheduler based on Neural Networks, you never really know what kind of strange rules will be inferred by the model. Without a training/test split, you might get very bad rules like "If a review is done on 9PM and took 2.5s to answer, it will be a Again, but if it's 3s, it will be a good" kind of non-sense. 

But with a train/test splitted, you "catch" attempt from the optimizer to "cheat" the prediction by not giving him everything during training time.

I don't know how bad it could be for FSRS, but I know @1DWalker has been very vocal on how RMSE(bins) could be cheated and how even for FSRS this split is also crucial.

And even if it weren't that bad for FSRS, it still set a non-fair comparison basis for potential future scheduling, since they won't be able to cheat their way through (or on the opposite, if we don't split train/test for them, they'll just cheat and create as many rules as there is reviews for some users, so we'd get non-generalizable models)

user1823:
I completely agree that testing should be on the data on which the parameters were not trained.

However, the best way to do that is to make splits during evaluation (just like this PR already does) rather than also preventing the optimizer from accessing 20% of the data.

Expertium:
Possible train-test splits

Expertium:
Possible train-test splits

L-M-Sherlock:
> Hmmm the point is that by sacrificing 20% of your review to become a "test set", you won't really lose 20% of accuracy. While, having no training/test split, might, or even worse, completely destroy any accuracy on new data (see the following example).

I don't think so. If the time-series-split evaluation shows that FSRS could generalize well, there is not reason to prevent it from using all data to optimize.

dae:
I'm working with the following understanding. Please correct me if I've got anything wrong below:

- The majority of the FSRS regulars are in favour of this change, as it brings the same methodology/outputs as the benchmark
- The devs also don't want users fiddling around with their parameters and evaluating to try and get "better" results.
- @L-M-Sherlock stated earlier that these stats were originally added because the helper had them, and not because they are super useful
- @brishtibheja wonders whether we even need to show exact numbers instead of just a health check
- @Danika-Dakika has pointed out how the existing evaluate is useful as a tech support tool, and it becomes less useful if the params are ignored

Some thoughts based on the above:

Re support: having some way of telling that the settings are optimal is useful. Could we perhaps preserve some of the previous utility by having it include a line like "Your parameters are not currently optimal" when the configured ones don't exactly match the ones that would be generated by the evaluation, so the supporter knows when the params should be optimized?

Or, would it make sense to combine optimize preset+evaluate into a single button, so that optimizing both ensures the parameters are current, and tells the user if FSRS is a good/average/bad fit for their data/review style? (imaginging it would return something like RMSE under the hood, but the UI would convert that into something easy to understand)

Other ideas/feedback welcome too!

L-M-Sherlock:
> * stated earlier that these stats were originally added because the helper had them, and not because they are super useful

A minor mistake: the helper doesn't have them. The [Notebook version of Optimizer](https://github.com/open-spaced-repetition/fsrs4anki/blob/main/fsrs4anki_optimizer.ipynb) has it. And it also includes a series of charts and graphs to visualize the distribution of errors, which is helpful to detect the weakness of FSRS. The metrics alone is not very helpful.

Expertium:
> imaginging it would return something like RMSE under the hood, but the UI would convert that into something easy to understand

I was thinking of showing both numbers and a color bar thingy.

brishtibheja:
The numbers stop being useful with this change though.

Expertium:
Hm, I guess so, especially if I use a correction for n(reviews) and retention like I plan to, in which case the same RMSE can be either good or bad. Sure, let's not show numbers then.

YukiNagat0:
This PR and the whole idea to eliminate the `Evaluate` button looks like a complete mess to me. I really don’t want to participate in this discussion, but I’ll just write my opinion:

1. The `Evaluate` behavior should be preserved. I don’t care about the discrepancy between the benchmark and Anki’s implementation. I want to be able to see how well **_my weights_** perform. This is the whole purpose of the `Evaluate` button - why should I care about "evaluating the algorithm itself"?
2. I don’t understand where the idea that someone is playing around with their weights to get the best values comes from. Nobody is doing this. For example: I use the `Evaluate` button in the following way - optimize weights in Anki, write down the results, tweak some weights in a way I believe is more suitable for my studying behavior (e.g., changing the initial stability so the graduating interval becomes ~3 days instead of the 25 days that FSRS usually outputs), and then go back to Anki and evaluate my tweaked weights to know **_how much they worsen the metrics_**. Yes, I’m not trying to tweak the weights to get better metrics; instead, I want to know how much worse they evaluate compared to FSRS’s "ideal" parameters.
3. The `Health Check`/color bar idea is also messed up. I don’t understand the point of it. It serves zero purpose. The `Health Check` in other software is used to diagnose problems and resolve them. What user problem can the FSRS `Health Check` resolve? There is no way to improve it other than doing more Anki reviews and becoming more consistent in your answering behavior (being honest, using the same buttons, etc.).

Dae, you asked about the opinion of FSRS regulars - no, not everyone is in favor of this change. The idea that the `Evaluate` button outputs some obscure metrics unrelated to my parameters is gibberish. The `Evaluate` button should output the results of the user’s weights.

brishtibheja:
> Nobody is doing this. 

There are people who do this, no one is making it up. Even I did it at one point.

> What user problem can the FSRS Health Check resolve? There is no way to improve it other than doing more Anki reviews and becoming more consistent in your answering behavior (being honest, using the same buttons, etc.).

Agreed that it's not the best idea, but hopefully problems like "not optimising" or "bad grading behaviour" can be addressed by this.

Expertium:
1) https://docs.google.com/forms/d/1Uy8zr9QOS6u-oLVRwVCuQfyFUiSwKxt9pWEvlTGdn9k/viewanalytics
According to this survey of mine, only 15% of users know what values of log-loss and RMSE are good, and only 14% know how they are calculated.

2) https://docs.google.com/forms/d/17F4KM188RFHU9s-jyd9VDltk52ZgceTMY3qNhWeXYxE/viewanalytics
According to this survey, when presented with 3 options - keep Evaluate, remove it, or rework it into a "health check" - 73% of users voted for the health check as their **first preference**, meaning that they prefer it to removing Evaluate, and they also prefer it to keeping Evaluate unchanged.

In short, most people don't understand what Evaluate does and would prefer a rework.

L-M-Sherlock:
> * The `Evaluate` behavior should be preserved. I don’t care about the discrepancy between the benchmark and Anki’s implementation. I want to be able to see how well **_my weights_** perform. This is the whole purpose of the `Evaluate` button - why should I care about "evaluating the algorithm itself"?

Your weights are optimized by the built-in optimizer, right? With this PR, the **Evaluate** button could tell you how well the optimizer perform. It's unrelated to the evaluation of the algorithm itself, which is done in the benchmark.

The problem of the old **Evaluation** is, it evaluates the parameters on the same data they were trained on (or even slightly tweaked from), which often gives an overly optimistic score that doesn't reflect how well FSRS will actually predict recall for future reviews. 

The time-series split in the PR aims to give a more realistic estimate of FSRS's ability to generalize – essentially, how well the process of optimizing parameters works for your specific review history when predicting unseen data. The goal shifts from "How good are these specific numbers on past data?" to "How reliably can FSRS generate good parameters for my future reviews?".

YukiNagat0:
L-M-Sherlock, I was talking exactly about what you do: "How reliably can FSRS generate good parameters for my future reviews?"
I understand that the current implementation is just computing how weights preform on past data.
But why it is bad? I think that it is a completely valid way to evaluate how well weights perform.
What I want to say is the fact that if the user has many reviews and consistent with his answering behavior, then we can assume that past data indeed completely represents his memory patterns (no significant changes in the future). If it is the case, then it is completely fine to evaluate FSRS on past data

L-M-Sherlock:
> But why it is bad? I think that it is a completely valid way to evaluate how well weights perform.

We don't know whether it overfits the data if we use the same dataset in training and evaluating. If it's overfitting, the weights will perform worse than what the current evaluation button tells you. How weights preform on past data is weakly related to the future's performance. The PR's implementation could tell you how well the algorithm performs in the future because it's evaluated on "unseen" data (which is not used to train the parameters).

YukiNagat0:
> The PR's implementation could tell you how well the algorithm performs in the future because it's evaluated on "unseen" data (which is not used to train the parameters).

The problem is in the "algorithm performs". Why can not we evaluate how **_weights_** perform?
Maybe we can think about the solution that still evaluates user's weights and uses the unseen data?

brishtibheja:
afaik it's still "how weights perform" albeit more like "how we expect it to perform in the future".

Expertium:
Unrelated, but I was hoping that we would eventually switch from "weights" to "parameters" when talking about FSRS, for 2 reasons:
1) The word "weights" is confusing for a layman who isn't familiar with machine learning jargon.
2) It's not even technically correct since "weights" refers to neural networks, where a neuron receives input from multiple other neurons, each input value having its own coefficient, aka "weight".

I mentioned this all the way in late 2023 or early 2024, and unfortunately some people still say "weights" 😓

L-M-Sherlock:
Let me use a simple visualization to elaborate this problem:

image

The old evaluation tells us the metrics of green cell. The new evaluation tells us the metrics of blue cell. We care about the metrics of yellow cell.

The relation from blue cell to orange cell is similar to the relation from yellow cell to red cell.

dae:
@YukiNagat0 thank you for taking the time to post your thoughts. I do appreciate the argument that this is a loss in functionality, and my initial reaction was similar.

That said, the arguments for this are fairly convincing:

- Most users didn't understand the numbers or find the feature useful
- The only users who will miss this are those manually tweaking their parameters, and hoping for a "sanity check" of their changes
- The "sanity check" is of limited value, as it's not using a separate set of test data.

Regarding the health check, your points about it being actionable are reasonable. Perhaps it would make sense to simplify this+integrate it into the optimize routine? Not saying anything in the normal case, and providing some sort of message in the cases where SM2 may be a better fit?

Expertium:
I was thinking of not showing any text if health check says Good or Acceptable, and showing a list like this if it says Poor:
- Make sure that you are using the answer buttons honestly and consistently. Keep in mind that Hard is a passing grade, not a failing grade.
- Suspend or reformulate leeches.
- While splitting your material into different presets can be beneficial, you probably don't need more than 5-10 presets.
- Make sure you understand before you memorize.
etc.

There two problems. First, this will be a whole lot of text in a small pop-up. Second, the kind of person who has really bad habits is exactly the kind of person who won't read a lot of text.

dae:
Many people don't read carefully, but some will, and the message may also give a hint to tech support as to what the problem might be. We could also shift the hints into a web link instead of trying to show it all in a pop-up.

brishtibheja:
what if someone tries all the suggestions and evaluate still shows bad results? considering this, it's probably better to integrate the check into optimise and remove the seperate evaluate. 

> Second, the kind of person who has really bad habits is exactly the kind of person who won't read a lot of text.

when providing a warning message to users, we can then have checklist that you're supposed to fill before you can click OK. providing a link to docs isn't a good option IMO.

Expertium:
> what if someone tries all the suggestions and evaluate still shows bad results? considering this, it's probably better to integrate the check into optimise and remove the seperate evaluate.

How would that work with Optimize All Presets?

dae:
My thinking is that such a check would only be done when optimizing a specific preset, as it doesn't seem so useful for established users. But if we did have to add it to the bulk optimization case, such a warning could be displayed only once for the set, if a warning was triggered.

Expertium:
I still think that having a separate Evaluate button is better

dae:
To clarify my thinking here:

- I was thinking it would show nothing in the common case
- If the calculated RMSE/log loss is above some threshold the FSRS devs decide on, it would link users to a page offering some tips

In that context, a combined button makes sense to me: the user shouldn't need to an extra step to tell them if FSRS is a good fit (easily missed), and in the common case, the extra button wouldn't offer any utility.

But perhaps you're thinking of something else, like continuing to show the underlying numbers? If you are, what actionable info does that provide to the user?

Expertium:
There shouldn't be a "Your FSRS parameters appear to be optimal" message, and "Your parameters are poor" is also misleading. As I said earlier, we are not evaluating the current parameters, but rather, how well FSRS performs on unseen data, on average, based on different data and different parameters.
Something like "FSRS works well for you" or "FSRS can adapt to your review history well" would be better. Other than that, I think you're right, we don't need a bar if we're not showing numbers.

brishtibheja:
> what if someone tries all the suggestions and evaluate still shows bad results? considering this, it's probably better to integrate the check into optimise and remove the seperate evaluate.
https://github.com/ankitects/anki/pull/3962#issuecomment-2862475801

I want to clarify what I meant by this. If we keep a seperate evaluate, a scenario like this is possible:

1. Some user runs "Evaluate".
2. Evaluate shows a somewhat scary message that FSRS isn't working optimally for them and they need to do something. 
3. The user tries everything and a month later use "Evaluate" again. It still shows the same scary message again.

This is a situation where the user feels stuck in a rut. And it does the exact opposite of reassuring people of FSRS's effectiveness.

If we instead combine "Evaluate" with "Optimise" we can simply choose not to show anything to the user unless maybe RMSE gets worse. We can do the checklist thing I mentioned (nobody will click on a help link anyway). Anki decides when to show and what to show.

With "Evaluate", your only option is either:

1. Have UI elementsthat ensure the user is eternally aware of FSRS not being optimal for them.
2. Lie to the user.


>  It would pop-up for every preset, which would be annoying.
https://github.com/ankitects/anki/pull/3962#issuecomment-2868899681

That's not an issue if you mention the presets/decks that doesn't work in a single screen, just like other checks work in Anki.

> Second, the kind of person who has really bad habits is exactly the kind of person who won't read a lot of text.
https://github.com/ankitects/anki/pull/3962#issuecomment-2862372373

Sure, but is this a kind of person who's simultaneously:

1. Unikely to read a text that appears after optimisation.
2. More likely to click evaluate and read the text than click evaluate.

dae:
Guys, I do not feel like the PR is in a mergable state at the moment. Changing the way these stats are calculated fundamentally alters the feature, and while some ideas like a health check have been proposed as a replacement to what we had before, I don't think we have any consensus on how it would be implemented - or if it should be implemented at all. And until we have that, the only solution I see for now is to remove the simulate button after merging this in, as this change seems to make the whole feature less useful. But people will be upset if we do that, so I'd rather see us having a plan in place for what if anything we'll replace it with.

Expertium:
@dae I think this: https://github.com/ankitects/anki/pull/3962#issuecomment-2869168531 is pretty good. If we aren't showing metrics, then there's no need to show a slider on a scale either. We can just show text.

Expertium:
Integrating it into "Optimize" is much closer to removing it than to reworking it.

brishtibheja:
I think the UI is changed but the functionality remains.

dae:
I still feel an explicit 'health check' button is a poor choice. Concerned users need to go through extra steps to press it, and unconcerned users may have no idea there's a problem. I also still feel that showing a message to the user when everything's fine is unnecessary.

Since the new calculations only work with optimized parameters, showing the message when optimizing seems to make sense. Doing a health check with non-optimal params, and having them ignored is potentially confusing, and this approach avoids that. And ideally, the answer to 'how do I ensure FSRS is working well'? should just be 'click optimize'.

Regarding optimize 1 + optimize all - the health check could be supported in the latter as well, if there isn't a big performance impact - we could just note down any encountered problems and summarize them at the end. But if it doubles optimization time, that's going to be hard to justify to established users who have no need for the check. Needs a quick benchmark.

dae:
I'd like to get a new beta (and subsequent release) out fairly soon, and I've been holding off on building one until all the FSRS changes are in. If you hope to land this one in 25.06, a basic implementation that shows a warning to the user when they optimize a single preset and the RMSE/etc is above a threshold would be sufficient, with it optionally being added to 'optimize all' in the future.

Are there any other scheduling-related changes I should be waiting for?

Expertium:
Not quite scheduling related, but Luc will hopefully submit another PR with further refinements of the estimation of change in workload.

> a basic implementation that shows a warning to the user when they optimize a single preset and the RMSE/etc is above a threshold

We can do that.

Expertium:
Oh, and Luc is working on the Knowledge Over Time graph.

Regarding wording for the warning, I'm thinking of this:

> (current text with metrics and a mention of lower = better)
> 
> Your RMSE is unusually high, given your retention and the number of reviews. Recommendationss:
>  - Suspend or reformulate leeches.
>  - Use the answer buttons consistently. Keep in mind that "Hard" is a passing grade, not a failing grade.
>  - Understand before you memorize.

I'm explicitly saying "given your retention and the number of reviews" because I will not be using raw RMSE but RMSE adjusted for retention and n(reviews).

For Evaluate I think it makes sense to show the warning only if the parameters are non-default.

brishtibheja:
The strings can be improved later so it's not a huge deal but can we not say words like RMSE/log-loss? You also previously agreed that we don't show any metrics.

Expertium:
So then are we removing Evaluate completely and adding the warning to Optimize?
@dae

brishtibheja:
I think yes:

> basic implementation that shows a warning to the user when they **optimize** a single preset
https://github.com/ankitects/anki/pull/3962#issuecomment-2926903740

*(emphasis mine)*

Expertium:
Alright, then:

> FSRS doesn't fit your review history well. Recommendationss:
>  - Suspend or reformulate leeches.
>  - Use the answer buttons consistently. Keep in mind that "Hard" is a passing grade, not a failing grade.
>  - Understand before you memorize.

As for the cutoff of the value of normalized RMSE, I guess 95th percentile would be ok. In other words, ~95% of users would not see a warning.

dae:
Agreed that we shouldn't mention RMSE/etc. Your wording looks good, but some suggested tweaks:

Some potential alternatives for the first line:
- "Your memory is difficult to predict. Recommendations:"
- "Parameters optimized, but could be better. Recommendations:"

The former implies FSRS shouldn't be used; the latter makes it sound less dire, while still encouraging change.

And a new paragraph at the end to set expectations, something like:

"It may take weeks or months before parameters become optimal."

WDYT?

@Luc-Mcgrady no pressure/rush, but if you have a rough estimate of how long you'll need, that would be helpful for release planning.

Expertium:
> "Parameters optimized, but could be better. Recommendations:"

Not a fan, tbh. It makes it sound like the optimizer isn't doing a good job, but the optimizer is doing the best it can. I don't think "Your memory is difficult to predict" sounds dire, it sounds ok to me. And yeah, I agree that we shouldn't use wording that may make the user turn FSRS off.

dae:
How about "Your memory is difficult to predict optimally."?

(It still feels a bit more serious than the latter though)

Expertium:
That sounds odd IMO. I like it more without "optimally".

user1823:
> Since the new calculations only work with optimized parameters, showing the message when optimizing seems to make sense.

With this PR,
- optimization uses the complete review history
- evaluation makes 5 time-series splits, optimizes parameters on those splits and evaluates them on the next split (https://github.com/open-spaced-repetition/fsrs-rs/blob/cc4ae769e08f6fa65136b6980233b212dcba08c3/src/inference.rs#L623-L628)

So, the training of parameters during optimization and evaluation is not done in exactly the same manner. It should be possible to continue to the evaluate loop after the 4 split to obtain parameters that are trained on the entire review history. However, I am not sure how much slower it would be in comparison to the current optimize.

Expertium:
@user1823 I think Dae's idea is to NOT implement the change in this PR and just add a warning, without changes to the optimization procedure. And remove Evaluate

user1823:
> just add a warning

Warning based on the current method of evaluation? That sounds like a good and non-controversial idea. I feel that the new method of evaluation won't make it into 25.06 considering how many disagreements we currently have regarding this.

dae:
Sorry if I have not been clear. I'm not suggesting the evaluation change be dropped. I am saying that changing the evaluation procedure will mean a separate Evaluate button is no longer very useful, so this needs to be exposed separately like via the Optimize button. I am trying to provide you a path to get these changes merged in.

Expertium:
Ok, so just to avoid misunderstandings:
1) We don't change how *optimization* is done, we keep the current procedure and don't use the 5-way split (which is what this PR was initially about, before dozens of comments). This is not The Data Scientist Way, but eh, it's probably fine
2) We remove the "Evaluate" button
3) We calculate the metrics at the end of optimization
4) If they exceed a threshold, we show a warning. This way there is no "Evaluate" as a separate button, but health check is a thing

Dae, is this correct?

dae:
For 1, please see the post above. For 4, there's no separate 'health check' button, but we are checking for health when clicking on Optimize.

Expertium:
1) We **do** change how optimization is done, and this PR with the 5-way split gets merged
2) We remove the "Evaluate" button
3) We calculate the metrics at the end of optimization based on the 5-way split
4) If metrics exceed a threshold, we show a warning

There will be an "Optimize" button with a built-in health check, but no "Evaluate" button

Correct?

dae:
Yes. Where in 3, 'at the end of evaluation' is happening during Optimize.

>  I am saying that changing the evaluation procedure will mean a separate Evaluate button is no longer very useful,

And the argument to users who will miss it is that it was bad data science, and the old output couldn't be trusted anyway, right? Because some users will be (or already are) upset, and we need a reasonable answer when asked why the change was made.

Expertium:
> 'at the end of evaluation'

My bad, that was a mistake

Ok, got it

user1823:
If this is the plan, we would need some changes in FSRS-rs and a benchmark to compare the time taken for old "optimize" and new "optimize + health check". See https://github.com/ankitects/anki/pull/3962#issuecomment-2927068976

If the performance of the new "optimize + health check" turns out to be very poor, just adding a message after optimization based on current evaluation method and keeping the Evaluate button seems like a good option for this release. This would hopefully require only a single line of change to the current `main` branch code (excluding the text of the message).

> it was bad data science, and the old output couldn't be trusted anyway

That's true, but I want to highlight that Anki would still be using the old method internally to compare the parameters before and after optimization because we don't have a better way.

dae:
>  I want to highlight that Anki would still be using the old method internally to compare the parameters before and after optimization because we don't have a better way.

Sorry, I'm not following this. Are you saying we'd need both the old and new method? If so, what would we use the new one for?

brishtibheja:
> If so, what would we use the new one for?

For the health check (what is currently evaluate). Optimisation checks RMSE to determine whether to keep the old params and that one still uses the old method. Or is the plan to give up some data and use the new method?

dae:
And it's not "bad data science" in that case?

user1823:
Brishtibheja has summarized that correctly. I would expand on that slightly to make things clearer.

- During optimization, the new method would calculate the loss/RMSE on the 5 splits.
- After optimization, the old method would be used to calculate the loss on the whole data to know whether the newly generated parameters are better than the current parameters. (https://github.com/ankitects/anki/pull/3962/files#diff-00a5efbc608bd4165dd1ea67cc66bb31ec92ecfa4fec186ca3942fef27a2ca4eL118-L122)
- If we are keeping the optimized parameters, we would do a health check based on the loss/RMSE calculated in step 1.

I know this looks messy, but I don't think we have a better option.

> And it's not "bad data science" in that case?

It's bad data science. But, the alternative is to use 4/5th of the data for training and then comparing the parameters on the last 1/5th. However, this would make optimization worse.

Expertium:
> That's true, but I want to highlight that Anki would still be using the old method internally to compare the parameters before and after optimization because we don't have a better way.

Why? We can use parameters obtained on most of the data, minus the last split that is used only for testing.
image

Then we don't need the old method because we can get everything we need (metrics and parameters) from the new method

brishtibheja:
And if that test set only has like a 100 reviews?

Expertium:
You mean train set? If it has 100 reviews, well, then it has 100 reviews

dae:
Ok, I understand the argument that this is a tradeoff. But if the old approach is good enough for us to use in this case and must be kept around, it's hard to justify removing users' access to it. Maybe the evaluate button should be kept? All the other discussion about the optimize-based health check could still apply.

Expertium:
~~So then we don't merge this PR and keep the current optimization method?~~

man, I'm getting confused

user1823:
Dae's saying that we add a health check to the Optimize button (based on 5-split approach) and keep the Evaluate button that uses the old method.

dae:
Based on https://github.com/ankitects/anki/pull/3962#issuecomment-2927132029 I'm assuming that this calculation change would result in a more accurate result for the health check. Is that right? If so, and we're not breaking the existing Evaluate workflow, what's the argument for not merging it in?

Expertium:
> Dae's saying that we add a health check to the Optimize button (based on 5-split approach) and keep the Evaluate button that uses the old method.

Wait, a health check AND the current Evaluate? I don't see how that's a good idea

dae:
I agree Evaluate's not useful to most users, but some users want it. A compromise would be to move it into the Advanced section, or leave it exposed in the debug console but not the GUI.

brishtibheja:
debug console sounds fine but I doubt a lot of people would use it if its moved there... evaluate seems more like a nerdy fetish than something with a real use case.

Expertium:
https://github.com/ankitects/anki/pull/3962#issuecomment-2927132029

> During optimization, the new method would calculate the loss/RMSE on the 5 splits.
> After optimization, the old method would be used to calculate the loss on the whole data to know whether the newly generated parameters are better than the current parameters. (https://github.com/ankitects/anki/pull/3962/files#diff-00a5efbc608bd4165dd1ea67cc66bb31ec92ecfa4fec186ca3942fef27a2ca4eL118-L122)
> If we are keeping the optimized parameters, we would do a health check based on the loss/RMSE calculated in step 1.

If that's the plan, I don't see why we need to keep Evaluate

dae:
Because people are (ab)using it to benchmark their handmade changes to params. Even if you think it's a bad idea, removing the functionality will upset them, and when they inevitably want answers, it's hard to give them a good one if the code is still in use by Anki and just has been hidden away from them.

Expertium:
Look, I don't want to upset users either, but keeping Evaluate and adding a health check at the same time is too confusing.
We want some kind of solution that reduces the amount of confusion. Having both a health check and Evaluate is **strictly** more confusing than only having one of them.

Luc-Mcgrady:
Is this what people have in mind?

https://github.com/user-attachments/assets/49bf81b5-431c-40d4-8326-f1542e05a044

dae:
That looks like the evaluation is considerably longer than the actual optimization? If that's the case and it can't be improved, it's harder to justify doing this indiscriminately, and a stronger argument could be made for a separate "health check" button so users don't have to endure the longer wait when they're confident FSRS is performing well.

If we went down that path, perhaps one other way of dealing with the Evaluate button would be to stuff it into the health check, like:

```
["Your memory is predicted well" | "Your memory is difficult to predict"]

Stats for nerds:
RMSE: XXXX Log Loss: XXXX
```

Thoughts?

Expertium:
> ["Your memory is predicted well" | "Your memory is difficult to predict"]
> 
> Stats for nerds:
> RMSE: XXXX Log Loss: XXXX
> Thoughts?

Maybe collapse the numbers, so that the user has to deliberately click to see them? To be clear, I still think we should only keep one: either Evaluate (with numbers) or the health check (text only).

dae:
Collapsing seems reasonable. But you're saying even collapsed, we shouldn't have both? Why?

Expertium:
> We want some kind of solution that reduces the amount of confusion. Having both a health check and Evaluate is strictly more confusing than only having one of them.

dae:
I'm arguing it's all part of the same health check button, rather than two separate buttons.

Luc-Mcgrady:
> That looks like the evaluation is considerably longer than the actual optimization? If that's the case and it can't be improved, it's harder to justify doing this indiscriminately, and a stronger argument could be made for a separate "health check" button so users don't have to endure the longer wait when they're confident FSRS is performing well.

I can make it a pr for you to check if you want. Or you can find it here:
https://github.com/Luc-Mcgrady/anki/tree/evaluate-after-optimise
(.patch)
https://github.com/Luc-Mcgrady/anki/commit/b78d1e1d9b22e7dafb8cdfadea5ae19d096b321b.patch
(compare)
https://github.com/ankitects/anki/compare/main...Luc-Mcgrady:anki:evaluate-after-optimise?expand=1

Expertium:
> I'm arguing it's all part of the same health check button, rather than two separate buttons.

Alright, fine. Honestly, I'm too tired of debating Evaluate, so let's just get done with this.

dae:
What do others think?

Expertium:
Actually, no, wait. We can't show both the health check and metrics as calculated using the *current* evaluation method. Health check does not depend on the current parameters in the parameter field, but Evaluate as it is right now does.

EDIT: actually, even if both the text and the metrics are based on the new method, there is still a problem.

1) If we show the health check based on the **new** method and metrics calculated using the **current** method, users will ask why manually changing parameters doesn't change the text even though the metrics change.
2) If we show the health check based on the **new** method and the metrics calculated using the **new** method, users will ask why metrics don't depend on the parameters in the respective field.
3) If we show the health check based on the **current** method and the metrics calculated using the **current** method, then health check becomes significantly less useful or even useless, since it's now parameter-dependent.
4) If we show the health check based on the **current** method and the metrics calculated using the **new** method, then we get an utter mess: a useless (or at least less useful) health check and metrics that do not depend on the current parameters.

Ok, I thought about every possible combination and all of them have downsides.

brishtibheja:
Is adding a checkbox for optimise an option? So, there would be a toggle-able checkbox that says "do a health check when optimising" and anyone who doesn't need it can turn it off.

Expertium:
Give me a moment, I'll make a diagram

user1823:
I like the checkbox idea by @brishtibheja. This way, we add the health check to optimize rather than evaluate. We can keep the current Evaluate functionality but tuck it away in the Advanced section so that most users won't bother.

Also, in https://github.com/Luc-Mcgrady/anki/commit/b78d1e1d9b22e7dafb8cdfadea5ae19d096b321b, Anki is doing an optimization first and then doing an evaluation. We can make it faster by updating FSRS-rs to continue the evaluate loop in `fsrs.evaluate_with_time_series_splits()` after the 4 split to obtain parameters that are trained on the entire review history instead of using the `fsrs.compute_parameters()` function separately.

dae:
@Expertium IIRC, the old evaluate method was fairly quick, so I'm operating on the assumption that doing it in addition to the split evaluation wouldn't impact performance much. Number 1 was what I was thinking of, and I'm not sure it's such a big issue: we needn't publish the thresholds controlling the top message or exactly how it's implemented, so why should users expect a given stat change to alter the message? If it becomes a common question, we could add a brief explanation in the collapsed stat area.

The alternative is health check only and relegating the stats to a separate button or debug console. I'm not pushing for one at the moment, just spitballing.

@brishtibheja I don't have a strong preference, but for the sake of argument: the user would have to click twice in either case. A checkbox saves having to wait for the first operation to complete before the second one can be started, which is good. But it's probably not going to jump out at the user as much as blue button called "Health Check". Happy to go with consensus.

dae:
> We can make it faster

How much faster? If it adds, say, 20% to optimization time, the lack of a separate button is perhaps justifiable. If it adds 100%, that's hard to justify.

user1823:
> How much faster?

I am not sure. But a quick way to test that would be:

In Luc's implementation above,
- calculate the time taken by the `fsrs.compute_parameters()` part
- calculate the time taken by the `fsrs.evaluate_with_time_series_splits()` part

With my suggestion, the total time taken for "optimize + health check" should be about 20% more than the time taken by the `fsrs.evaluate_with_time_series_splits()` part currently.

So, if the `fsrs.compute_parameters()` part and the `fsrs.evaluate_with_time_series_splits()` part take approximately equal amount of time in Luc's implementation, we would be good to go.

Expertium:
@dae 

Evaluate metrics and text combination

...why did I do it in Photoshop instead of Excel? 😅

dae:
@user1823 judging by the video above, evaluate_with_time_series_splits is the much slower part, so there doesn't seem room to achieve optimize-only-like performance

@Expertium I'm not convinced the differing calculations of the top and bottom will be that confusing in practice, but I have thought of another reason why that may not be the best approach. People who are (ab)using this are presumably going through a cycle of tweak-check-tweak-check, and making them do it with the health check would make the check a lot slower than it was before (IIRC). While relegating this feature to the debug console will be more cumbersome for those users, if it means they can iterate/check much faster, it's probably still the lesser of two evils.

Expertium:
Unless my math is wrong, theoretically 5-way split is as fast as 2 * current optimization + 0.8 * evaluation

Current optimization: 5/5 = 1
Current evaluation: 5/5 = 1

5-way split:
Optimization: 1/5 + 2/5 + 3/5 + 4/5 = 2
Evaluation: 1/5 + 1/5 + 1/5 + 1/5 = 0.8

So you can calculate 2 * time(optimization) + 0.8 * time(evaluation) to get a rough estimate of how much time the 5-way split takes

dae:
So more than doubling our current optimization? I feel that's hard to justify doing every time.

dae:
So to summarise, how does this sound?

1) Evaluate button is removed, but the underlying routine is still exposed in our Python API. Providing a bit of example code that users can copy+paste to evaluate their tweaks would be a nice gesture for affected users.
2) Health check button or checkbox is added, which uses the new code in this PR. Which of the two to be determined by you guys. The health check just tells the user if things look good or not, with no specific stats.

brishtibheja:
Sounds good. And I think checkbox is better:

1. It can be turned on by default so that newbies don't miss out on it.
2. Checkbox mean we can later integrate this into "Optimize all preset".
3. Maybe some people would like health check to always run and checkbox is better for that (they won't need to click twice because the setting is saved).

dae:
Sounds reasonable to me. I gather you're suggesting we store this in the preset config, so the value persists over close+open?

brishtibheja:
yea right

dae:
Ok, sounds good.

One minor thought: adding a " (slow)" to the end of the checkbox might make it clearer to users why they might want to turn it off.

user1823:
I don't know how feasible it is but I have an idea to drastically improve the performance of evaluate_with_time_series_splits().

Currently, this function is doing 4 separate optimizations with increasing amounts of data.

To make it faster, we can make it do a single optimization on the whole data while "sampling" the parameters obtained after each split. This way, we will get the five sets of parameters by running the optimizer only once.

If this is possible to implement, it will hopefully be fast enough to be used with optimize every time, without the need of a separate button or checkbox.

Expertium:
I'm not sure what you mean. I would ping Jarrett, but he said that he will be busy in the next few days

Luc-Mcgrady:
I think he means that you use the previous optimiser when optimising the next split in order to skip running on the previously optimised FSRSItems?

image

Like this maybe? 

------

Quick question, Should the health check be run even if the parameters don't change / are already optimal. If so how should we handle both the messages appearing at the same time.

dae:
"already optimal" seems to contradict the other message, so perhaps we should just hide it in that case?

Expertium:
For Luc

Hmmm. I'm not sure if it's equivalent, we need @L-M-Sherlock for this.

> Quick question, Should the health check be run even if the parameters don't change / are already optimal.

Health check does not depend on the *current* parameters, so yes.

> If so how should we handle both the messages appearing at the same time.

I haven't thought about that. I guess display both messages, one after the other.

For Dae

> "already optimal" seems to contradict the other message, so perhaps we should just hide it in that case?

That's what happens when one thing depends on the current parameters and the other thing doesn’t.

dae:
Do you have a better proposal?

Expertium:
Tbh, no.

user1823:
@Luc-Mcgrady, your diagram is quite close to what I had in mind.

To be even more efficient, we don't need to stop the optimization process after each split. We can run the optimizer once through the whole data and as the parameters are being continuously changed by gradient descent, we can save the values that were achieved after each one-fifth part of the data. Then, run evaluation at the end using the corresponding parameters for each split.

But, I am not sure if this is even possible to do in the FSRS-rs crate, without any changes to burn-rs?

<hr>

About health check when the parameters don't change, I propose:
- Run the health check
- If the loss is poor, show the "difficult to predict' message. The user won't complain about the parameters not changing because we just told him that his memory is difficult to predict.
- If the loss is Ok, show the "already optimal" message.

dae:
I'll proceed as previously discussed for now - if the check later gets dramatically faster, we can reconsider combining it with the health check. Thanks to everyone who contributed here, and especially @Luc-Mcgrady for his hard work.

PR #3966, opened 2025-04-28
Check if self.card.reps>0 before substracing 1

JSchoreels:
Hello, today I used the "Learn Now" from AJT Card Management : https://ankiweb.net/shared/info/1021636467

And when it marked a card as "Learning" (but without reps or anything), it caused some PanicException when clicking on the deck :
`thread '<unnamed>' panicked at rslib/src/scheduler/answering/current.rs:74:69:
attempt to subtract with overflow`

Looking at the code, I was able to make the issue disappear by checking that self.card.reps is greater than 0 before substracting, and else return 0. So in case of overflow, it will just return 0

It's not per se an Anki issue but I think treating those values a bit more safely shouldn't hurt ?

This is the state of the card that cause the issue : [carddata.json](https://github.com/user-attachments/files/19941700/carddata.json). It misses the field "reviews", which could explain the panic ?

### DIFF
--- a/rslib/src/scheduler/answering/current.rs
+++ b/rslib/src/scheduler/answering/current.rs
@@ -71,14 +71,7 @@ impl CardStateUpdater {
                     // Decrease reps by 1 to get correct seed for fuzz.
                     // If the fuzz calculation changes, this will break.
                     let last_ivl_with_fuzz = self.learning_ivl_with_fuzz(
-                        get_fuzz_seed_for_id_and_reps(
-                            self.card.id,
-                            if self.card.reps > 0 {
-                                self.card.reps - 1
-                            } else {
-                                0
-                            },
-                        ),
+                        get_fuzz_seed_for_id_and_reps(self.card.id, self.card.reps.wrapping_sub(1)),
                         last_ivl,
                     );
                     let last_answered_time = due as i64 - last_ivl_with_fuzz as i64;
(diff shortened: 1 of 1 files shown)


rslib/src/scheduler/answering/current.rs
```diff
@@ -71,7 +71,7 @@ impl CardStateUpdater {
                     // Decrease reps by 1 to get correct seed for fuzz.
                     // If the fuzz calculation changes, this will break.
                     let last_ivl_with_fuzz = self.learning_ivl_with_fuzz(
-                        get_fuzz_seed_for_id_and_reps(self.card.id, self.card.reps - 1),
+                        get_fuzz_seed_for_id_and_reps(self.card.id, self.card.reps.wrapping_sub(1)),
```
Luc-Mcgrady, line 74:
```suggestion
                        get_fuzz_seed_for_id_and_reps(self.card.id, self.card.reps.saturating_sub(1)),
```
- https://doc.rust-lang.org/std/intrinsics/fn.wrapping_sub.html
- https://doc.rust-lang.org/std/intrinsics/fn.saturating_sub.html

```rs
fn main() {
    assert_eq!(0_u32.saturating_sub(1), 0);
    assert_eq!(0_u32.wrapping_sub(1), u32::MAX);
}
```

jakeprobst: I would argue that wrapping sub is preferable here. the context here is to generate a rng seed for a card. with saturating, 0 and 1 both lead to the same seed. and with wrapping 0 and 1 generate different seeds.

this is likely only raising an error because it was built in debug mode for some reason and so is panicing over underflows? so assuming release mode would just let the underflow happen, wrapping maintains the same functionality.

but ultimately its only for generating a seed that nothing important relies on so it doesn't actually matter.

JSchoreels: Yep, should have precised it but I started with teh saturating one and jake adviced me to switch since we're talking about having a seed value (so no biggies to let it go underflow, even better to keep different seed values for 0 and 1)

dae:
Thank you for the fix!

PR #3968, opened 2025-04-29
Allow custom AnkiWebViews to opt in for API access

abdnh:
This is needed for Card Info During Review: https://forums.ankiweb.net/t/card-info-during-review/86/20

## Alternative solutions

- Add `AnkiWebViewKind.CUSTOM_PAGE_WITH_API_ACCESS` to avoid modifying the signature of AnkiWebView.
- Allow `AnkiWebViewKind.DEFAULT` (might be risky).

### DIFF
--- a/qt/aqt/webview.py
+++ b/qt/aqt/webview.py
@@ -122,9 +122,10 @@ class AnkiWebPage(QWebEnginePage):
         self,
         onBridgeCmd: BridgeCommandHandler,
         kind: AnkiWebViewKind = AnkiWebViewKind.DEFAULT,
+        enable_api_access: bool = False,
         parent: QObject | None = None,
     ) -> None:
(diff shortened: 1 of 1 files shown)


dae:
The add-on didn't need any changes on Anki's end, as we can reuse the Kind that the integrated card stats pages use: https://github.com/ankitects/anki-addons/commit/ddd583364ccbb4b81eac61a3f3210dda2c897ac9

Did you have other add-ons in mind that need this change, where such a solution is not feasible?

abdnh:
> Did you have other add-ons in mind that need this change, where such a solution is not feasible?

No, not aware of any other add-on. I just thought it might be worth providing an explicit option to enable API access. Thinking of it more, adding a `AnkiWebViewKind.set_api_access_enabled()` method is probably the best option.

Will close this for now though.

PR #3970, opened 2025-04-29
add ellipsis to Grade Now

L-M-Sherlock:
> The context menu entry should display “Grade Now …” (with ellipsis). Ellipsis (three dots) are typically added programmatically to strings requiring additional user input before action execution. Other context menu options around “Grade Now” include ellipsis, so this should align for consistency.

source: https://forums.ankiweb.net/t/anki-25-05-beta/59710/37?u=l.m.sherlock

### DIFF
--- a/ftl/qt/qt-accel.ftl
+++ b/ftl/qt/qt-accel.ftl
@@ -30,7 +30,6 @@ qt-accel-tools = &Tools
 qt-accel-undo = &Undo
 qt-accel-redo = &Redo
 qt-accel-set-due-date = Set &Due Date...
-qt-accel-grade-now = &Grade Now...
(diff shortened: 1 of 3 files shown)


dae:
I propose a different approach. Since we'll presumably want to add these to the mobile clients in the future, I suggest the translation be moved into the rslib rtl and left without the dots (since mobile clients don't use dots). For the desktop, you can then make use of add_ellipsis_to_action_label()

dae:
Pre-Rust, all of our translations had ...s on them. Since we transitioned to Rust, for newer labels, I've been trying to put them in actions + make use of that routine so we can reuse them across the clients.

PR #3971, opened 2025-04-29
Fix/Layout shift on Card Counts button hover

Luc-Mcgrady:
Fixes:

https://github.com/user-attachments/assets/f30ce0e0-8f59-43c2-9efc-6e772477b03f


dae:
Thanks!

PR #3976, opened 2025-05-02
feat(scheduler): Allow bypassing queue logic via CardAnswer flag

xuanyue202:
## Problem

Currently, the backend scheduler assumes all card answers processed via the `answer_card` endpoint originate from the standard review queue (`from_queue` is implicitly true). This doesn't accurately represent scenarios like card previews or potentially some custom study sessions where answering a card should not affect the main queue counts or state.

## Solution

This PR introduces an optional boolean field `from_queue` to the `CardAnswer` protobuf message definition (`proto/anki/scheduler.proto`).

When `from_queue` is set to `false` by the client, the backend service (`rslib/src/scheduler/service/answering.rs`) will now set the internal `from_queue` flag to `false` for that specific answer. This allows the scheduler to correctly handle answers that should not modify the standard queue state.

If `from_queue` is omitted or set to `true`, the behavior remains unchanged (the answer is considered `from_queue = true`).

## Impact

This change provides more flexibility for clients interacting with the scheduler, enabling them to distinguish between answers that are part of the normal review flow and those that occur outside of it (e.g., previews), ensuring queue metrics remain accurate.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -225,6 +225,7 @@ babofitos <https://github.com/babofitos>
 Jonathan Schoreels <https://github.com/JSchoreels>
 JL710
 Matt Brubeck <mbrubeck@limpet.net>
+Yaoliang Chen <yaoliang.ch@gmail.com>
 
 ********************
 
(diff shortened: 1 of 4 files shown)


abdnh (approved):
LGTM

dae:
Reverting this due to https://github.com/ankitects/anki/pull/3983

If a separate method proves necessary in the future, we should address this issue at the same time.

PR #3977, opened 2025-05-02
Properly align label and radio input in the stats window

GithubAnon0000:
The way it currently looks is all over the place (especially visible for _deck_ and _all history_ ):
anki

After this PR the radio button circles and their labels are properly aligned:
anki

### DIFF
--- a/ts/routes/graphs/RangeBox.svelte
+++ b/ts/routes/graphs/RangeBox.svelte
@@ -112,6 +112,15 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 <div class="range-box-pad"></div>
 
 <style lang="scss">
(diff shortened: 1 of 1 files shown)


ts/routes/graphs/RangeBox.svelte
```diff
@@ -112,6 +112,15 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 <div class="range-box-pad"></div>
 
 <style lang="scss">
+    label {
+        display: inline-flex;
+        align-items: center;
+    }
+
+    input[type="radio"] {
+        margin-right: 0.3em;
```
abdnh, line 121: Please use `margin-inline-end` for RTL compatibility.
image

PR #3978, opened 2025-05-02
ADD focus and hover CSS rule to stats searchbar

GithubAnon0000:
# Problem
The stats search bar by default is very small and makes it difficult to properly search something in it. I read in the forums once or twice that users were writing their queries in a separate text editor, then pasted it into the search bar.

While this is bad, having a huge search bar by default likely isn't a great idea either.

# Solution
The solution is to expand the width of the search bar whenever it has focus (meaning the user types something into it). I also made it expand on hover (when the mouse hovers over it) to make it easier to read the query without having to click into the search bar again.

(Expanding height is possible but is of no use, since the text inside the search bar does not wrap anyways.)

# Visuals

https://github.com/user-attachments/assets/019e2262-2047-4337-80cd-7029b0258fa5

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -224,6 +224,7 @@ rreemmii-dev <https://github.com/rreemmii-dev>
 babofitos <https://github.com/babofitos>
 Jonathan Schoreels <https://github.com/JSchoreels>
 JL710
+Matt Brubeck <mbrubeck@limpet.net>
 
 ********************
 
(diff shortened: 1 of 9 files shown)


abdnh:
Ideally, the input will only expand to the right and maybe doesn't take whole available space. It feels clunky to me at the moment.

dae:
The clunkiness could perhaps be improved by animating the transition instead of immediately changing size. Another possible option would be to make the box read-only, and a click/tap on it opens up a modal text box like in the 'add preset' dialog.

GithubAnon0000:
[Going through old PRs and closing them. I won't have enough time to update the PRs or develope them further for probably a few months. In case someone wants to continue here, feel free to so.]

PR #3979, opened 2025-05-02
UPDATE answer button graph tooltip to include I) answer button name and II) description of what "correct" means

GithubAnon0000:
This is somewhat a followup to #3952 .

# Problems
1. Users might not know that "1" → "Again", "2" → "Hard"...
2. It's not obvious what "correct" means in the tooltip.

# Proposed solutions
1. We can just add the name of the button.
2. I added a very short description. Not 100% content with it, but I think this kind of info should certainly be prominently displayed. Adding it to the graphs description felt wrong though, as this info is only relevant once you read the tooltips.

# Visuals
_Before this PR:_
anki

_With this PR:_
anki

### DIFF
--- a/ftl/core/statistics.ftl
+++ b/ftl/core/statistics.ftl
@@ -229,7 +229,7 @@ statistics-stability-day-single =
 # hour range, eg "From 14:00-15:00"
 statistics-hours-range = From { $hourStart }:00~{ $hourEnd }:00
 statistics-hours-correct = { $correct }/{ $total } correct ({ $percent }%)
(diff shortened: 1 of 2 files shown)


dae:
- The English labels like 'Hard' may be incorrect for cards in learning on the v1 scheduler, but that's hopefully far enough behind us at this point that it's not a major concern.
- I find the tooltip too wide now. If we want keep it as a bit of text we tack on at the end, how about "(not 'Again')" as a more succinct alternative?

dae:
Agreed that we should avoid breaking existing translations where we can. Thank you for the contribution!

PR #3980, opened 2025-05-03
Fix/SQL retrievability underflow

Luc-Mcgrady:
Before, saturating_sub is called on an i64 which does not have a limit at 0. This would then underflow when converted to a u32.

```rs
[rslib/src/storage/sqlite.rs:328:17] days_elapsed.saturating_sub(review_day) = -1
[rslib/src/storage/sqlite.rs:337:13] days_elapsed = 4294967295
[rslib/src/storage/sqlite.rs:337:13] &ret = Ok(
    Some(
        2.892674e-5,
    ),
)
```

ref: https://discord.com/channels/368267295601983490/1367979269854204044/1367979269854204044


dae:
Nice!

PR #3981, opened 2025-05-03
CHANGE collection size too large error to add MB values and info about compressed vs. uncompressed.

GithubAnon0000:
fixes: https://forums.ankiweb.net/t/your-collection-size-is-large-suggestions/60084

I have added a conversion from bytes to MB and passed this to the error function instead of just the bytes. I also added the info that this is the uncompressed value, not the compressed value.

**Important:**
I don't know how to trigger this error without having an account and a large enough collection, so couldn't actually test it. I'm not sure whether it's okay to create a test ankiweb account for these kind of things and I'd not be sure how to create a collection large enough to be affected by this restriction anyways.

I'd love to know how I could test this!

`./check` does pass though.

**Edit:** I forgot to add translatable strings for "uncompressed". I'll do so later, but first I'd like to know how to test this window / make the window appear for testing purposes.

### DIFF
--- a/ftl/core/sync.ftl
+++ b/ftl/core/sync.ftl
@@ -53,7 +53,7 @@ sync-clock-off = Unable to sync - your clock is not set to the correct time.
 sync-upload-too-large =
     Your collection file is too large to send to AnkiWeb. You can reduce its size by removing any unwanted decks (optionally exporting them first), and then using Check Database to shrink the file size down.
     
-    ({ $details })
+    { $details } (uncompressed)
 sync-sign-in = Sign in
 sync-ankihub-dialog-heading = AnkiHub Login
 sync-ankihub-username-label = Username or Email:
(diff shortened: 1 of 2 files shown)


dae:
It can be triggered by attempting to do a one-way sync with a collection that's over the limit, or by changing the if line above your change to 'if true'. Please limit the decimal places to 1 or 2 though, or it's still hard to read.

dae:
I'm not sure adding 'uncompressed' twice to the message really helps with legibility either. I get that you're trying to answer people who say 'but my colpkg is smaller', but the distinction isn't really important for most people, as the way to resolve the issue is the same.

dae:
This won't break existing translations, as we haven't changed the key, so existing translations will continue to be used.

The documented limit is 250MB, but the code is currently only enforcing it at 300. No objections to a doco update if you're keen.

In any case, thank you for the PR!

PR #3986, opened 2025-05-05
Fix "Open Image" being broken for filenames with certain reserved characters

### LINKED ISSUE #3984: Broken image preview in image occlusion when name contains non-url safe characters
## What are the steps to reproduce this bug?
- Create a new Image Occlusion card, with the image filename having '#'  (or other unsafe character)
- Right click on the image
- Select `Open Image`

## Expected behaviour
- Image opens in default Image Preview app

## Result
- Nothing happens

## StackTrace
```
2025-05-04 21:28:14,799:DEBUG:aqt.mediasrv: POST /_anki/getImageForOcclusion
gio: file:///home/div/.local/share/Anki2/User%201/collection.media/609965f85823a7003548496f_#%23_BASIC%20MATHS%2001%20UMEED%20D_250330_012911_35.jpg.png: Error when getting information for file “/home/div/.local/share/Anki2/User 1/collection.media/609965f85823a7003548496f_”: No such file or directory
```

This error `“/home/div/.local/share/Anki2/User 1/collection.media/609965f85823a7003548496f_”: No such file or directory` is also affecting AnkiDroid

## Relevant commits/issues/PRs
https://github.com/ankitects/anki/pull/3377
https://github.com/ankidroid/Anki-Android/pull/18193
https://github.com/ankidroid/Anki-Android/issues/18173

iamllama:
Fixes #3984

> QUrl will automatically percent encode all characters that are not allowed in a URL and decode the percent-encoded sequences that represent an unreserved character (letters, digits, hyphens, underscores, dots and tildes). All other characters are left in their original forms.
https://doc.qt.io/qt-6/qurl.html#QUrl-3

This meant `#`s weren't escaped when constructing the `QUrl`, and caused the filename to be truncated. The fix proposed is to ~~quote the path beforehand~~ use [QUrl.fromLocalFile](https://doc.qt.io/qt-6/qurl.html#fromLocalFile) to construct the file url

No change is needed for windows, but i haven't been able to test on mac

### DIFF
--- a/qt/aqt/utils.py
+++ b/qt/aqt/utils.py
@@ -930,7 +930,7 @@ def openFolder(path: str) -> None:
         subprocess.run(["explorer", f"file://{path}"], check=False)
     else:
         with no_bundled_libs():
-            QDesktopServices.openUrl(QUrl(f"file://{path}"))
+            QDesktopServices.openUrl(QUrl.fromLocalFile(path))
 
 
 def show_in_folder(path: str) -> None:
@@ -947,7 +947,7 @@ def show_in_folder(path: str) -> None:
     else:
         # Just open the file in any other platform
         with no_bundled_libs():
-            QDesktopServices.openUrl(QUrl(f"file://{path}"))
+            QDesktopServices.openUrl(QUrl.fromLocalFile(path))
 
 
 def _show_in_folder_win32(path: str) -> None:


dae:
I've double-checked it's fine on a Mac, though I would have been very surprised if it wasn't. Thank you for clean fix :-)

PR #3987, opened 2025-05-07
Allow rotating IO masks

iamllama:
Relevant: https://forums.ankiweb.net/t/allow-rotating-io-masks/60336

image

NB1: editing on older clients can be destructive (rotation-wise only), as the `angle` IO cloze property would be stripped out
NB2: 6d767c4fffeedd9efead08029465138ac880eba7's message is mistakenly misleading, it's not meant to be a full revert of that commit

EDIT: the bounds checking leaves one case unresolved: a thin diagonal rect can have a bounding box large enough that it still remains in the canvas, but the actual rect itself is far enough away as to be hidden. we could check for non-intersection with the canvas, but im not sure where the mask should be moved to. thoughts?


dae:
Re the bounds checking, the seems like a rare-enough case that it's probably not worth the effort?

PR looks good. The only thought I had was about the way we're serializing the angle to four decimal places, which is probably more than we need. And once the user has made adjustments, even if they rotate back to what looks like 0, chances are it remains stuck on an angle like 0.05. We could perhaps address the former by normalizing the value to 10k steps, but I'm not sure if it's worth it for a few bytes savings per rotated occlusion.

iamllama:
Even with 10k steps its still difficult to get it back to 0, but the rounding error isn't noticeable and the bytes saved do add up 👍

dae:
I think it might automatically close if the branch becomes empty of new commits? Not sure though. In any case, thank you as always for the improvements, including going the extra step of normalizing those values. :-)

PR #3989, opened 2025-05-08
Make IO polygon markers centred and transparent

iamllama:
Before:
io-before

After:
io-after

ref: https://forums.ankiweb.net/t/io-polygon-tool-lines-alignment-with-the-dots/60459


dae:
The only issue I noticed turned out to be an existing one, where clicking on an existing polygon will close an open one. Thank you for all these I/O improvements :-)

PR #3990, opened 2025-05-09
Fix polygons closing when clicking existing masks while editing IO

iamllama:
> The only issue I noticed turned out to be an existing one, where clicking on an existing polygon will close an open one.
https://github.com/ankitects/anki/pull/3989#issuecomment-2863246099

The polygon tool checks whether a point was clicked by comparing against a mask's `id`. This prop is set when making new masks, but not when the masks are first hydrated from clozes. The fix proposed is to set the prop then as well

66eb527fc is for preventing new polygons from being selected/dragged while the polygon tool is active. This is made possible by 2b463a816 but doesn't seem intended given that existing polygons aren't selectable
~~EDIT: undoing and redoing don't preserve this ([previously noted](https://github.com/ankitects/anki/blob/5cc44b3f687aba2fce865f11904803a9470c9eb5/ts/routes/image-occlusion/tools/tool-undo-redo.ts#L92))~~ fixed

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -201,7 +201,7 @@ Dongjin Ouyang <1113117424@qq.com>
 Sawan Sunar <sawansunar24072002@gmail.com>
 hideo aoyama <https://github.com/boukendesho>
 Ross Brown <rbrownwsws@googlemail.com>
-🦙 <github.com/iamllama>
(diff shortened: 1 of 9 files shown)


dae:
You're on fire with these I/O fixes!

PR #3992, opened 2025-05-11
Fix/Leech suspended tooltip

Luc-Mcgrady:
https://github.com/ankitects/anki/blob/44ade0653e100fe98bf0702c42632798263b9dca/qt/aqt/reviewer.py#L556
Seems to set `self.card` to the state that was in
https://github.com/ankitects/anki/blob/44ade0653e100fe98bf0702c42632798263b9dca/qt/aqt/reviewer.py#L545-L549? This means that it will use the previous card's state if you pass `self.card` to `self.onLeech` (which wasn't attempted so I assume is a known issue)
To circumvent this, I store the suspended value before that function is called and pass that to `self.onLeech` as a boolean instead.


dae:
Thanks Luc.

Issue #3965, opened 2025-04-28
URL schemes are a potential security issue


dae:
We currently allow opening arbitrary URLs with external programs, not only http/https. This is a potential security issue, as an attacker could attempt to open a program that performs a destructive action.

This is quite a useful feature though - users use it for creating deep links into other apps like note taking apps and dictionary apps. So if we start blocking non-HTTP schemes by default, we'll need some way for users to opt out.

Potential options:

1) A global on/off toggle. This is the worst option, as once users enable it, any malicious content can sneak in going forward.
2) A notetype toggle. Users can approve trusted notetypes, while leaving new notetypes defaulting to off.
3) A global whitelist of valid schemes, e.g. 'dict1' allows any scheme starting with dict1://. More efficient if users use the same URL schemes across multiple notetypes.

I'm slightly leaning towards 3, as it's likely easier to implement than 2. Thoughts?

BrayanDSO:
3.

Also, if you click a non whitelisted link, the app ideally should warn that the URL is being blocked due to security reasons and provide a quick option to whitelist that scheme.

PR #3994, opened 2025-05-11
Add URL scheme whitelist

dae:
Attempts to address #3965. Once this is merged in, the plan is to build another bugfix-only 25.02 release.

### DIFF
--- a/.cursor/rules/i18n.md
+++ b/.cursor/rules/i18n.md
@@ -1,0 +1,7 @@
+- We use the fluent system+code generation for translation.
+- New strings should be added to rslib/core/. Ask for the appropriate file if you're not sure.
+- Assuming a string addons-you-have-count has been added to addons.ftl, that string is accessible in our different languages as follows:
+  - Python: from aqt.utils import tr; msg = tr.addons_you_have_count(count=3)
+  - TypeScript: import * as tr from "@generated/ftl"; tr.addonsYouHaveCount({count: 3})
+  - Rust: collection.tr.addons_you_have_count(3)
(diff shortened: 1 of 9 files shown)


BrayanDSO:
IMO, the UX could still be better.

image

Separating things by spaces is too basic. Seems like a lazy approach if the user is demanding enough. The entries should be separated and removeable with a click/tap.

image

And here you could add the scheme directly instead of making the user type it themselves. And open the link after that, instead of forcing the user to click it twice.

iamllama:
> Separating things by spaces is too basic. Seems like a lazy approach if the user is demanding enough.

Given that it's a one-off per scheme, and that its a somewhat common* practice to get users to type in to confirm assent on a potentially risky op, it doesn't seem too lazy

*https://ux.stackexchange.com/questions/130147/why-would-one-use-enter-the-name-of-the-project-to-confirm

soleuniverse101:
> IMO, the UX could still be better.
> 
> image
> 
> Separating things by spaces is too basic. Seems like a lazy approach if the user is demanding enough. The entries should be separated and removeable with a click/tap.
> 
> image
> 
> And here you could add the scheme directly instead of making the user type it themselves. And open the link after that, instead of forcing the user to click it twice.

Would adding buttons `Allow once` and `Always authorize scheme` be what you're suggesting ?

I think that's a good idea to prevent having to click twice on the link and having to manually modify something that resembles a config file.

Also `Always authorize scheme` (or whatever it's named) should then ask for a last confirmation with a message clearly stating (in red maybe) the security risk it represents.

dae:
Thanks all. #4002 improves things somewhat, so I'll merge that in after this one. Adding an 'allow once' button is a good way of allowing the user to experiment without having the scheme permanently whitelisted, without an obvious way to edit the allowed schemes.

PR #3995, opened 2025-05-11
Feat/90% desired retention warning

Luc-Mcgrady:
Keeps a warning active when the users desired retention 90% to try and hint at the correlation between desired retention and intervals.

I'm not positive on what the message should be, suggestions welcome.

c.c. @Expertium

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -475,6 +475,7 @@ deck-config-a-100-day-interval =
         [one] A 100 day interval will become { $days } day.
        *[other] A 100 day interval will become { $days } days.
     }
+deck-config-90-percent-desired-retention-warning = The higher the desired retention, the shorter the intervals between reviews.
 deck-config-percent-of-reviews =  
     { $reviews ->
         [one] { $pct }% of { $reviews } review
--- a/ts/routes/deck-options/FsrsOptions.svelte
+++ b/ts/routes/deck-options/FsrsOptions.svelte
@@ -75,7 +75,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
             (stability / factor) * (Math.pow(retention, 1 / decay) - 1),
         );
         if (days === 100) {
-            return "";
+            return tr.deckConfig90PercentDesiredRetentionWarning();
         }
         return tr.deckConfigA100DayInterval({ days });
     }


ftl/core/deck-config.ftl
```diff
@@ -475,6 +475,7 @@ deck-config-a-100-day-interval =
         [one] A 100 day interval will become { $days } day.
        *[other] A 100 day interval will become { $days } days.
     }
+deck-config-90-percent-desired-retention-warning = The higher the desired retention, the shorter the intervals between reviews.
```
Luc-Mcgrady, line 478:
```suggestion
deck-config-90-percent-desired-retention-warning =  Your intervals will be unaffected by the current desired retention.
```

Expertium: The intervals are always affected

brishtibheja: Maybe we save some space by not saying "desired retention"? "To have shorter intervals, increase the value from 90%."

Expertium: I don't have much faith in Anki users' ability to understand that "value" refers to DR.

dae:
Nothing else in the deck options screen shows an info 'warning' out of the box. This PR increases the amount of text the user sees by default, which I'm not really fond of - those boxes are intended to call the user's attention to things that may not be right.

That said, I do appreciate that devs want to make the option more discoverable, and showing something in the 90% case may be the best way to do that.

On the wording: it doesn't feel very accessible. If your goal is to appeal to the masses, 'retention', 'intervals' etc is not great. What about something like the following?

"To see cards more frequently, increase this above 90%"

Luc-Mcgrady:
Maybe both this message and(?) the `100 days will become x` message we currently have could be hidden until the user does something FSRS related. For example maybe show it if they:
- Enable FSRS for the first time
- Optimize
- Change desired retention or focus on the input

With the exception of if desired retention is set to a "warning" or "danger" value.

ftl/core/deck-config.ftl
Luc-Mcgrady, line 478:
```suggestion
deck-config-90-percent-desired-retention-warning = The higher this value, the more frequently cards will be shown to you.
```
?

Expertium:
> On the wording: it doesn't feel very accessible. If your goal is to appeal to the masses, 'retention', 'intervals' etc is not great.

But the setting is literally called "desired retention"

ftl/core/deck-config.ftl
Luc-Mcgrady: I mistakenly made this separate instead of continuing from here https://github.com/ankitects/anki/pull/3995#discussion_r2083509581
I'll commit this. If there are any more suggestions feel free.

brishtibheja:
would it be clear "this value" refers to desired retention and not the params below it? sure, many native speakers might understand "this value" is a single value but there might be some confusion. any thoughts on this?

Expertium:
@dae @Luc-Mcgrady how about only displaying the "The higher the desired retention, the shorter the intervals between reviews." message if the user has never changed DR from the default value of 90? Add another variable like `dr_changed` that will serve as a "flag": when the user changes DR, `dr_changed` will be set to True, and if it's True, then the message doesn't appear.
Basically, the message will be shown ONLY to people who have never changed DR and will stop appearing after that.
The "A 100 days interval will become X days" message should remain unchanged, as it is now.

Maltesaa:
There's already a page explaining what DR does inside the settings menu for people who care. I think this would just be unnecessary clutter for people who either don't care or already know.

If this is implemented I def agree that it should only be shown when the user enables FSRS for the first time, or if the user has never changed DR from the default.

dae:
I don't want to faff about with extra logic to show such a message only if the value has never been changed before, as we have to track that then. I'm not particularly fond of the current wording either, as I don't like how we switch between quantitive and qualitative messages for only 90%. If this is something that is really worth pursuing, and there's consensus that focusing on workload is more intuitive/useful than interval, then perhaps the "a 100 day interval will become 46 days" message could be reworked into something like "Reviews required: 217%"? Then the 100 case doesn't look silly.

Expertium:
Dae, if you want another idea, here: https://forums.ankiweb.net/t/desired-retention-ui-overhaul/57678/59?u=expertium
Anki desired retention Godo answer button sequence

Instead of showing intervals per buttons, we instead show a sequence of “Good” reviews and their intervals?
The advantage of this method is that it shows interval lengths after the first interval, giving the user a better understanding of what kind of intervals he will have.

But this is more complicated and will require more discussions than the simple fix in this PR. If you like this idea - great. If not - this PR offers the simplest possible solution.

Danika-Dakika:
> There is a problem - users don't understand that DR affects intervals
> ...
> So we have a problem and the simplest possible solution to that problem

You keep saying it's the "simplest solution" -- but that doesn't mean it's a **good** solution to that problem. I still don't see any indication that there's a segment of users who will understand this _better_ by having this warning **in the Deck Options screen** (vs. somewhere else), or that the folks who are looking for solutions will find this more actionable by having it there. In my experience, many users simply ignore anything in a colored box in that screen 😅 -- just like they ignore the ❔'s staring them in the face when they have _questions_ about something.

brishtibheja:
NNgroup's eye tracking study comes to mind. Users ignore UX elements that
resemble ads, so boxes like these.

It's still better than nothing though.

Expertium:
> NNgroup's eye tracking study comes to mind. Users ignore UX elements that resemble ads, so boxes like these.

Then we might as well give up, to be honest.

Expertium:
@Danika-Dakika I've said this before, and I'll say it again: both here and on the forums, people don't agree on anything and don't understand that having **a** solution is better than having **no** solution. With UI there are often many ways to approach a problem, and many *different* better alternatives to the current implementation. But if nobody agrees on anything, then *none* of the better alternatives will be implemented.

Is showing a blue box with "The higher the desired retention, the shorter the intervals between reviews." better than the current implementation? Yes.
Is showing fake answer buttons better than the current implementation? Yes.
Is showing a sequence of "Good" reviews and their intervals better than the current implementation? Yes.
Will we agree to implement any of the above? Evidently no.

brishtibheja:
> Btw, do you mean this one?
https://www.iso.org/files/live/sites/isoorg/files/styleguide/resources/How_People_Read_on_the_Web.pdf

Yes, probably. I read some shorter articles from their website about this.

> But if nobody agrees on anything, then none of the better alternatives will be implemented.

I think dae agreed to add this but alas, then the issue was complicated too much. https://github.com/ankitects/anki/pull/3995#issuecomment-2882699409

Danika-Dakika:
@Expertium 
> ... and don't understand that having **a** solution is better than having **no** solution.

But having **a** solution _is not_ necessarily better and some "solutions" are worse than making no change. I know you get frustrated when people won't sign-on to your proposals, because they are obvious improvements as far as you are concerned. But you cast those who disagree with you as people who "don't understand" instead of letting them be people who *disagree* with you. 

Both here and on the forums, there are plenty of things that _do_ get done -- often by consensus-building and compromise. They just might not be the things you pick to draw a line in the sand about. 

> Is showing a blue box with "The higher the desired retention, the shorter the intervals between reviews." better than the current implementation? Yes.
> Is showing fake answer buttons better than the current implementation? Yes.
> Is showing a sequence of "Good" reviews and their intervals better than the current implementation? Yes.

I disagree with all of those "yes"es (some more strongly than others!). It's not because I don't understand, it's because I don't think they are overall benefits to the UI, or that they will do anything to solve the problem you've set your mind on solving. Not every proposal is an improvement, and I, like many, would rather see things stay how they are now, until there's a better solution.

Luc-Mcgrady:
> Reviews required: 217% of default

Do you think maybe it could be beneficial to compare this to the previously set desired retention instead of default. e.g.


(just opened config, DR previously set to 80%)
```
80%
Your intervals|workload will remain unchanged.
```
(change the value)
```
85%
Your intervals will be x% shorter than before
``` 
> and there's consensus that focusing on workload is more intuitive/useful than interval

or the reciprocal:
```
Your workload will increase by approximately x%
```

dae:
- Showing a message with an x still requires state: you need to record that you've shown the message somewhere so you know not to show it again. But I misspoke earlier: the 'only when first enabling FSRS' approach could work, since we don't need to persist anything then, and just show the message when toggling. Perhaps that is the best compromise here?
- I don't mind the idea of the three good steps as a way of more clearly showing the effects of the multiplier, though as has been pointed out, it doesn't solve the "doesn't understand DR" problem that this PR was originally trying to solve. They could potentially be added in addition to this message, though I imagine it'd look more crowded if we're permanently showing the message.
- Increase relative to before, if we can do it quickly, seems like it might be intuitive? Do we have a consensus on intervals vs workload? My vote would be for latter.

dae:
Ah, looks like your latest push already handled the first point - thanks Luc. If nobody else has something to add, I'll merge this in ~tomorrow morning.

Expertium:
You can't just say `Your workload will increase by approximately { $percent }%.` without simulations.

Expertium:
Well, run the simulator and see if the results are close to the ballpark figure.

Expertium:
Also, I don't think that showing a decrease relative to the previous value of DR is intuitive. Relative to the default 90% is a lot more intuitive
I'm sorry, but I don't think this is a good change

Cargo.toml
```diff
@@ -36,8 +36,8 @@ rev = "184b2ca50ed39ca43da13f0b830a463861adb9ca"
 
 [workspace.dependencies.fsrs]
 # version = "3.0.0"
-git = "https://github.com/open-spaced-repetition/fsrs-rs.git"
-rev = "092c20bac7d9239a991ae5b561556ad34c706c16"
+git = "https://github.com/Luc-Mcgrady/fsrs-optimizer-burn.git"
+rev = "fe6af1d33a2edb33dd8a50d87742e60e6a51de01"
```
Luc-Mcgrady, line 40: Waiting for https://github.com/open-spaced-repetition/fsrs-rs/pull/331

Luc-Mcgrady:
https://github.com/user-attachments/assets/e12717a4-f266-4bfb-a47d-2ff4ffee1b18

Thanks to Jarret, that should hopefully show a more accurate value.

Cargo.toml
Expertium: Is the code using placeholder time values (23 and 7 seconds) or the user's actual answer times, the ones that the simulator uses?
I mean, it's not super important for the sake of an initial demonstration, just saying that it should use real answer times like the simulator

dae:
Precomputing one value shifts the delay from calculation time to startup, and makes everyone pay it. Such an optimization would be better off deferring the calculation until it's actually required, then caching it. I don't see it as a requirement for landing this PR though.

The duplicate boxes don't look fantastic, but they won't always look like that, and I don't consider it a blocker either.

Expertium:
Luc, record a new video, please. I want to see the rounding in practice. And don't click super fast, lol

Luc-Mcgrady:
https://github.com/user-attachments/assets/aaa9b80e-c787-49f1-882f-d0547ed8bc77

Expertium:
Hm, it shows 0.5% then 0.4%. Maybe make 1% the minimum value?

Expertium:
Alright, should be good. If rounding is still funky in some edge cases, it can be fixed in another PR later.

dae:
I've seen it do things like 20 → 15 → 20 as I decrease DR, which I presume is the extra Agains influencing things.

I managed to trigger this error once, but couldn't do so again:

```
2025-05-24 00:46:02,736:DEBUG:aqt.mediasrv: POST /_anki/evaluateParams
2025-05-24 00:46:02,738:DEBUG:aqt.mediasrv: POST /_anki/getRetentionWorkload
  File "/Users/dae/Work/desktop/anki/qt/aqt/mediasrv.py", line 691, in wrapped
    if data := handler():
  File "/Users/dae/Work/desktop/anki/qt/aqt/mediasrv.py", line 672, in <lambda>
    return lambda: getattr(aqt.mw.col._backend, f"{endpoint}_raw")(request.data)
  File "/Users/dae/Work/desktop/anki/out/pylib/anki/_backend_generated.py", line 1126, in evaluate_params_raw
    return self._run_command(13, 34, message)
  File "/Users/dae/Work/desktop/anki/pylib/anki/_backend.py", line 172, in _run_command
    raise backend_exception_to_pylib(err)
anki.errors.InvalidInput: Insufficient review history to perform this operation.

JS error /_app/immutable/chunks/DuYVE4GH.mjs:2 Uncaught (in promise) Error: 500: Insufficient review history to perform this operation.
2025-05-24 00:46:04,262:DEBUG:aqt.mediasrv: POST /_anki/latestProgress
```

Looks like a useful new feature - thanks to everyone who contributed. Happy to merge in once the associated PR lands in the  FSRS repo.

Expertium:
Now it will always be in multiples of 50. I suggest making it in multiples of 50% for >=100%, in multiples of 25% for >=10% and in multiples of 2% for >=1%

Luc-Mcgrady:
> I managed to trigger this error once ...

I believe that the only error that can be thrown from the function is `FSRSError::InvalidParameters` from `check_and_fill_parameters`. It's probably related to the `/_anki/evaluateParams` instead.

rslib/src/deckconfig/service.rs
```diff
@@ -96,6 +96,25 @@ impl crate::services::DeckConfigService for Collection {
...
     }
+
+    fn get_retention_workload(
+        &mut self,
+        input: anki_proto::deck_config::GetRetentionWorkloadRequest,
+    ) -> Result<anki_proto::deck_config::GetRetentionWorkloadResponse> {
+        const LEARN_SPAN: usize = 1000;
+
+        const PASS_COST: f32 = 7.;
+        const FAIL_COST: f32 = 23.;
```
L-M-Sherlock, line 107: The costs are hard-coded now. Is it intended?

Expertium: I assume it's just until you merge your PR. They definitely shouldn't be hard-coded in the final release.

Luc-Mcgrady: I think calculating the costs might be a source of slowdown so will leave it if its super significant.

Expertium: Considering that you need the user's entire history to calculate them - yeah, I wouldn't be surprised if it's slow. But at the same time, using hard-coded costs could make this function too inaccurate to be useful.
I'd like you to run it with fixed costs and with costs calculated from your collection and see how much slower the latter is.

ftl/core/deck-config.ftl
```diff
@@ -470,11 +470,12 @@ deck-config-compute-optimal-retention-tooltip4 =
...
     is not recommended, as it will lead to a higher workload, because of the high forgetting rate.
 deck-config-please-save-your-changes-first = Please save your changes first.
-deck-config-a-100-day-interval = 
-    { $days ->
-        [one] A 100 day interval will become { $days } day.
-       *[other] A 100 day interval will become { $days } days.
-    }
+deck-config-workload-percentage-increase = Your workload will increase by approximately { $percent }%.
+deck-config-workload-percentage-decrease = Your workload will decrease by approximately { $percent }%.
+deck-config-workload-percentage-unchanged = The higher this value, the more frequently cards will be shown to you.
```
user1823, line 475:
```suggestion
deck-config-workload-percentage-unchanged = Increase this value to review cards more frequently.
```

ftl/core/deck-config.ftl
```diff
@@ -470,11 +470,12 @@ deck-config-compute-optimal-retention-tooltip4 =
...
+deck-config-desired-retention-too-low = Your desired retention is very low, which can lead to very long intervals.
```
user1823, line 476:
```suggestion
deck-config-desired-retention-too-low = Such low desired retention can cause very long intervals and increased forgetting.
```

ftl/core/deck-config.ftl
```diff
@@ -470,11 +470,12 @@ deck-config-compute-optimal-retention-tooltip4 =
...
+deck-config-desired-retention-too-high = Your desired retention is very high, which can lead to very short intervals.
```
user1823, line 477:
```suggestion
deck-config-desired-retention-too-high = Such high desired retention can significantly increase workload with little added benefit.
```

user1823 (commented):
Left some suggestions. Rest LGTM.

rslib/src/deckconfig/service.rs
Expertium: How about this: compute them when the user changes DR for the first time since opening deck options and store it for as long as the deck options window is opened. This means that the first time the user changes DR it will be slow, but the consecutive runs will be faster.
For example, if user has DR at 90%, changing it to 91% will be slow, but changing it to 92%, 93%, etc. will be faster because costs are saved

ftl/core/deck-config.ftl
```diff
@@ -470,11 +470,12 @@ deck-config-compute-optimal-retention-tooltip4 =
     willing to invest more study time to achieve it. Setting your desired retention lower than the minimum
     is not recommended, as it will lead to a higher workload, because of the high forgetting rate.
 deck-config-please-save-your-changes-first = Please save your changes first.
-deck-config-a-100-day-interval = 
-    { $days ->
-        [one] A 100 day interval will become { $days } day.
-       *[other] A 100 day interval will become { $days } days.
-    }
+deck-config-workload-percentage-increase = Your workload will increase by approximately { $percent }%.
```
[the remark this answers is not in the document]
Luc-Mcgrady: Oh sorry, I must have missed your suggestion somehow.

The workload and number of reviews in this case are different things. Here workload will be using the review times of cards.
```
This would result in a ~{ $change }% lower workload than { %before }% desired retention.
```
seems to fit

Mock up:
image

Good with you?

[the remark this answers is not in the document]
user1823: Umm... This is slightly confusing. 450% higher workload means 5.5 times (not 4.5 times) the workload at 70%.

Maybe we should change the message to use ratios rather than % change:

```
This would result in ~{ $factor } times the  workload at { %before }% desired retention.
```

If you want to keep the % change, your mockup in the previous comment looks good to me.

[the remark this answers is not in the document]
user1823: For smaller changes, % change is more readable. For larger changes, a ratio is easier to grasp.

So, as a more radical approach, you can display % change if the value is less than 100%. If the change becomes 100% or more, switch to a ratio.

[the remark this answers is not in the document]
Expertium: I also think that using ratios is better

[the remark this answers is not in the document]
brishtibheja: +1 from me. As a maths fearing guy, percentages are scary and unintuitive.

ftl/core/deck-config.ftl
Luc-Mcgrady: "Increase this value to review cards more frequently" doesn't automatically imply it works the other way around.

 I also feel like using the imperative "Increase this value" isn't keeping in tone with the other warnings.

Random example:
```
deck-config-apply-all-parent-limits-tooltip =
    By default, the daily limits of a higher-level deck do not apply if you're studying from its subdeck.
    If this option is enabled, the limits will
    start from the top-level deck instead
```
vs (re-written in the same style)
```
deck-config-apply-all-parent-limits-tooltip =
    By default, the daily limits of a higher-level deck do not apply if you're studying from its subdeck.
    Enable this option if you want limits to start from the from the top-level deck instead.
```

ftl/core/deck-config.ftl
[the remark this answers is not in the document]
user1823: Is it really obvious that the % change is relative to the previously stored value? With the current text, the user might be confused "increased but relative to what?"

Plus, if the user is playing around with different values of DR, he might actually forget what the starting value was.

About % vs ratio, I too feel % is more intuitive but only in the -100 to +99% range. From +100% onwards, ratios become more intuitive.

ftl/core/deck-config.ftl
Luc-Mcgrady: "increased forgetting" sounds incorrect to me.
"and for you to forget cards more frequently"?. Also "Such" should be  "A" or "Such a".

ftl/core/deck-config.ftl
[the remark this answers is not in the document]
Luc-Mcgrady: I'll implement { $change } and { $previousDR }%
As for changing to factor at 100%, it seems like excess complexity where none is needed. Maybe a factor but with a percent instead? 
```
Your workload will be x% what it was at {dr}%
```
?

ftl/core/deck-config.ftl
user1823: The current wording looks somewhat repetitive and casual. What do you think of the following? This is my last suggestion for this tooltip.

```suggestion
deck-config-workload-percentage-unchanged = This value controls how often cards appear — higher means more often.
```

ftl/core/deck-config.ftl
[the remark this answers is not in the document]
user1823: Expressing ratio as % doesn't provide any benefit, IMO. Using % change OR ratio in decimals would be better.

Since people feel that switching between ratio and % change is unnecessary complexity, I would vote for using ratio (in decimals) for all cases.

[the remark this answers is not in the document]
Expertium: > Expressing ratio as % doesn't provide any benefit

Agreed.

> I would vote for using ratio (in decimals) for all cases.

Agreed as well.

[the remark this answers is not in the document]
Expertium: @dae sorry for frequent pings, but I think your feedback is welcome
(I don't want to make yet another poll)
Ratios or %?

ftl/core/deck-config.ftl
Luc-Mcgrady: I feel as if the current version is easier to understand. I'll unresolve this in case people think otherwise.

ftl/core/deck-config.ftl
Expertium: We already display the change in workload, but we don't mention intervals. So mentioning them here makes sense. Otherwise we're not mentioning them at all.

ftl/core/deck-config.ftl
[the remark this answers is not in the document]
dae: No strong preference - a ratio is fine too, though I think '2.2x' might be more widely understood for non-native speakers than '2.2 times'?

On the topic of being explicit about the current DR: it seems clear to me without it, but maybe that's only because I'm already aware how it behaves. I find the current wording a bit awkward though, especially without a comma. How about something like the following?

"Compared to $dr, your workload will increase by ..."

"If you save, your workload will increase by ..."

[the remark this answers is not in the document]
Expertium: "Compared to {dr}, your workload will increase/decrease by {ratio} times" sounds good IMO. Not sure if "{ratio}x" is better than "{ratio} times", I'll leave it up to you and Luc.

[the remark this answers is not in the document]
user1823: Does this look good?

"Your workload will be about 2.5x what it was at 80% retention."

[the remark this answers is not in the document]
dae: Not bad - it's natural English, though a bit complicated for a beginner. Perhaps we'd be better off without a full sentence?

Approximate workload: 2.5x (compared to 70% retention)

The bit in brackets could potentially be placed on the line below/made smaller/etc to make it less prominent.

rslib/src/storage/card/get_costs_for_retention.sql
```diff
@@ -0,0 +1,44 @@
+WITH searched_revlogs AS (
+  SELECT *
+  FROM revlog
```
user1823, line 3:
```suggestion
  FROM revlog
  WHERE ease > 0
```

This would ensure that we use 10k user-graded revlog entries for calculating the costs.

rslib/src/storage/card/get_costs_for_retention.sql
```diff
@@ -0,0 +1,44 @@
...
+lapse_count AS (
+  SELECT COUNT(time) AS lapse_count
+  FROM searched_revlogs
+  WHERE ease = 1
+    AND type = 1
+),
+fail_sum AS (
+  SELECT SUM(time) AS total_fail_time
+  FROM searched_revlogs
+  WHERE ease = 1
```
user1823, line 20:
```suggestion
  WHERE (ease = 1 AND type = 1)
```

The previous code would include the answer time of failures during Learn state too.

rslib/src/storage/card/get_costs_for_retention.sql
Luc-Mcgrady: > does this use all the revlog entries in the collection or only those of cards in the preset

https://github.com/ankitects/anki/pull/3995/commits/2bd9c3461f602202e65c579707ccaf5416461343#diff-5d2f9fadb535a77bdb3e0db4b7a5341767eb1a0d09b8ce2bc17e0772f4ac3a10L4

I deleted this line by accident :neutral_face:

rslib/src/storage/card/get_costs_for_retention.sql
```diff
@@ -0,0 +1,30 @@
+WITH searched_revlogs AS (
+  SELECT *
+  FROM revlog
+  WHERE cid in search_cids
+  ORDER BY id DESC
+  -- Use the last 10000 cards
```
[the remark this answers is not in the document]
Luc-Mcgrady: > Luc will most likely make a new function in FSRS-rs specifically for calculating answer times for this mini-simulator

I retract that. This SQL seems to work fine.

Cargo.toml
```diff
@@ -37,7 +37,7 @@ rev = "184b2ca50ed39ca43da13f0b830a463861adb9ca"
 [workspace.dependencies.fsrs]
 # version = "3.0.0"
 git = "https://github.com/open-spaced-repetition/fsrs-rs.git"
-rev = "092c20bac7d9239a991ae5b561556ad34c706c16"
+rev = "53784503103169ca6c0702ae2a6d5c84c3d42269"
```
dae, line 40: Please note that as this was squash merged, that commit may disappear - we need to use the rev from main instead.

dae (commented):
Nearly there

ts/routes/deck-options/FsrsOptions.svelte
```diff
@@ -237,12 +286,14 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
     max={0.99}
     percentage={true}
+    bind:focused={desiredRetentionFocused}
 >
     <SettingTitle on:click={() => openHelpModal("desiredRetention")}>
         {tr.deckConfigDesiredRetention()}
     </SettingTitle>
 </SpinBoxFloatRow>
 
+<Warning warning={desiredRetentionChangeInfo} className={"alert-info"} />
```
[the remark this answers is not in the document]
dae: Screenshot 2025-05-26 at 11 23 35 pm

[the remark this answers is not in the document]
Luc-Mcgrady: Looks great!
I've forced the height to 2 lines here to prevent the layout shift with the "previous dr message" ( 4ec35ce480426f8c978a2713af1fb107fce7a779). 
image
Worth it?

ts/routes/deck-options/FsrsOptions.svelte
```diff
@@ -383,5 +383,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         margin-bottom: 0.375rem;
     }
 
-    :global(.pre-wrap) { white-space: pre-wrap; }
+    :global(.two-line) {
+        white-space: pre-wrap;
+        min-height: calc(2ch + 30px);
```
user1823, line 388: Maybe something like
```suggestion
        min-height: calc(2ch + 30px);
        vertical-align: middle;
```

Luc-Mcgrady: Implemented here 
e2a95c722ed4155a1254a81d4bfd199f35f90e46 
image

dae (commented):
Looks good. Thank you for all your hard work Luc, and to everyone who contributed here.

dae:
I noticed this message appears after clicking the Optimize button. Was that intended?

Luc-Mcgrady:
Yep https://github.com/ankitects/anki/pull/3995/commits/03b42a97d86b3f7b522967aa91eb31d9c44b218e

user1823:
@Luc-Mcgrady, but why? It seems weird for a popup to appear in a place other than the one I am interacting with.

Also, would it make sense to show this popup for any user who has DR equal to the default (90%)?

Luc-Mcgrady:
> @Luc-Mcgrady, but why? It seems weird for a popup to appear in a place other than the one I am interacting with.

1. The parameters affect the workload calculation. 
2. The original purpose of this PR was to inform people desired retention existed and how it affected their reviewing. If they don't know it exists then they will never click on it to see the popup and optimizing seems like a good time to remind them.

> Also, would it make sense to show this popup for any user who has DR equal to the default (90%)?

That seems annoying for people who genuinely want to set their DR to 90%.

PR #3996, opened 2025-05-11
Clarify field separator being a guess when importing csv

iamllama:
Relevant: https://forums.ankiweb.net/t/csv-delimiter-detection-fails/60682 and https://github.com/ankitects/anki/issues/3853

This pr proposes to hint to users that the field separator guessed by anki is in fact a guess, and that it should be changed if said guess is wrong

#3786 added a warning to do just that if preview data is almost definitely deformed, but IMO its also useful to highlight in the general case

Its help text has also been changed to explain why the field separator cannot be modified sometimes (when the file has a `#separator:...`), which is not something i've seen the manual cover


dae:
Seems reasonable. Thanks!

PR #3997, opened 2025-05-12
Style the fsrs params input

iamllama:
Before:
image

After:
image

The styling for the date input (ignore before) is cleaned up as well


L-M-Sherlock:
Is it possible to make sure the text box always show all contents? It would be helpful when we debug based on users' screenshot.

iamllama:
[field-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing) would've been perfect for this, but its only available on chromium 123 and later

PR #3999, opened 2025-05-13
Resize fsrs params input to fit content

iamllama:
> Is it possible to make sure the text box always show all contents? It would be helpful when we debug based on users' screenshot.
https://github.com/ankitects/anki/pull/3997#issuecomment-2875008497

[field-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing) would've been perfect for this, but its only available on chromium 123 and later


dae:
CSS breakpoints and min-height might have been an alternative here. I try to avoid JS-based positioning where possible due to potential performance issues/jankiness, but the numbers are not updated frequently, so it probably doesn't matter. Thanks for splitting this out into a separate PR in any case :-)

PR #4002, opened 2025-05-14
Make URL schemes dialog more ergonomic

soleuniverse101:
PR to apply changes I suggested in the following comment https://github.com/ankitects/anki/pull/3994#issuecomment-2878028254

This makes adding a url scheme easier. Users can still change URL schemes in the preferences menu.

Here's the result :

image


dae:
Thank you Adnane! Regarding https://github.com/ankitects/anki/pull/3994#issuecomment-2878028254, I'm not sure if we need a second warning - maybe one is enough?

PR #4009, opened 2025-05-16
Fix searching for non-blank adjacent wildcard fields

### LINKED ISSUE #4006: Wildcard search FieldPrefix*:_* returns all notes if any note type has multiple FieldPrefix* fields
Tested on Anki 25.02.5, Windows 11. Also reproducible on the latest 25.05 beta. Not tested on other versions.

When performing a wildcard search using the pattern `FieldPrefix*:_*`, the `_*` (non-empty field) condition is ignored if any single note type contains multiple fields sharing that same prefix. 
In that case, the search seems to behave like `FieldPrefix*:*`, and matches all notes that have a field starting with the prefix, regardless of whether those fields are empty or non-empty.

Example: 

- suppose a note type has these two fields: `BookName` and `BookPage`.
- perform a Browser search: `Book*:_*`
-  → Expected: return notes where at least one field starting with `Book` is non-empty
-  → Actual: returns all notes from any note type that has a `Book*` field, even if the fields are empty

iamllama:
Fixes #4006

When searching, adjacent fields are lumped together for efficiency. But this breaks when searching for non-blank fields (`_*`) as the field separator (0x1f) is picked up by the resulting query

The fix proposed is to split up adjacent fields in this case

### DIFF
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -583,8 +583,10 @@ impl SqlWriter<'_> {
     }
 
     fn write_single_field(&mut self, field_name: &str, val: &str) -> Result<()> {
-        let field_indicies_by_notetype =
-            self.num_fields_and_fields_indices_by_notetype(field_name)?;
+        let field_indicies_by_notetype = self.num_fields_and_fields_indices_by_notetype(
+            field_name,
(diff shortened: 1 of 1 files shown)


dae:
You're fast becoming our resident bug-fixer!

PR #4011, opened 2025-05-16
upgrade esbuild versions

### LINKED ISSUE #3824: Update esbuild for latest CVE
I merged a dependabot alert today for esbuild, but it looks like we still have an older esbuild in our package list as well. We may need to update related dependencies like esbuild-sass-plugin as well, but when I briefly tried to do so, I ran into compilation issues, so it will need more investigation.

spiritualfather:
Fixes #3824


dae:
Thank you for this!

PR #4016, opened 2025-05-18
Restore richtext editor focus on mathjax overlay close event

### LINKED ISSUE #4014: If focus is in mathjax editor, browser navigation should behave as if the focus was in the text field
This is certainly a very niche issue I'm having. But I'm having a deck full of note whose n-th card is mathjax. I want to use ctrl+n/ctrl+p to navigate between notes and my cursor to remain in the same field. It works great without mathjax. But when I use the keyboard to navigate inside the mathjax editor, do my correction and then press ctrl+n, the next note is shown with no field focused.

The python part seems trivial. In `editor.py` add a new command `mathJaxFocus:idex` that set `currentField` and `last_field_index` to `int(idx)`. (I want a different command than `focus` so that it does not trigger the hook)

But the svelte part makes me very confused. 
Specifically, I see that `<MathjaxOverlay` is used in NoteEditor, but I don't understand how it receives any information. This tags receive no parameter at all, which confuses me extremely. I'd love to know when it receive some information, and then, from which field it receive it.

iamllama:
Closes #4014

Focus is currently restored when moving past the start/end in the overlay editor but not during the close event that's dispatched when pressing Esc or forcing a save, which this pr proposes to rectify

### DIFF
--- a/ts/editor/mathjax-overlay/MathjaxOverlay.svelte
+++ b/ts/editor/mathjax-overlay/MathjaxOverlay.svelte
@@ -232,7 +232,10 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                             placeHandle(true);
                             resetHandle();
                         }}
-                        on:close={resetHandle}
+                        on:close={() => {
+                            placeHandle(true);
+                            resetHandle();
+                        }}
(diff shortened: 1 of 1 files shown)


Arthur-Milchior:
Thank you very much. I would never have guessed it'd be so simple!

PR #4021, opened 2025-05-20
Fix IO masks not saving when scaled

### LINKED ISSUE #3988: IO: Masks that are scaled but not moved aren't saved
Steps to reproduce:

1) Open an existing IO note or add a new one
2) Create a rect/ellipse* mask
3) Scale it (by dragging a corner)
4) Notice** that it isn't saved
5) Scale it again, and move it or another mask
6) Notice that it is now saved

*polygons and text don't seem to be affected
**either by having the previewer open and waiting 600ms, or toggling the mask editor on/off, or switching to another note and back in the browser

iamllama:
Fixes #3988

Masks are potentially re-modified to be bounded and have normalised scale factors, but that doesn't retrigger a save, which this pr rectifies

### DIFF
--- a/ts/routes/image-occlusion/mask-editor.ts
+++ b/ts/routes/image-occlusion/mask-editor.ts
@@ -116,7 +116,6 @@ function initCanvas(): fabric.Canvas {
             modifiedPolygon(canvas, evt.target);
             undoStack.onObjectModified();
         }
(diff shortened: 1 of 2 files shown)


dae:
The cleanup was a nice touch, thank you!

Issue #4022, opened 2025-05-20
Upgrade from Python 3.9.18


andrewsanchez:
Anki currently uses Python 3.9.18. 3.9 reaches [EOL status 2025-10](https://devguide.python.org/versions/), so this is urgent-ish, I think.

Related issues:
- https://github.com/ankitects/anki/issues/3787
- https://github.com/ankitects/anki/issues/3081

dae:
Closed by #4074

PR #4023, opened 2025-05-20
Upgrade python

andrewsanchez:
Draft PR to play with upgrading Python.

Closes https://github.com/ankitects/anki/issues/4022

TO-DOs:
- [ ] Comprehensive search/replace for references to specific python versions (`3.9`)
  - [ ] Update any docs accordingly.

Nice-to-haves:
- [ ] See if there is room for improvement for automation or DX for upgrading Python versions in the future.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,7 +8,7 @@
 	shallow = true
 [submodule "qt/bundle/PyOxidizer"]
 	path = qt/bundle/PyOxidizer
-	url = https://github.com/ankitects/PyOxidizer.git
+	url = https://github.com/indygreg/PyOxidizer.git
 	shallow = true
 	update = none
 
--- a/build/ninja_gen/src/python.rs
+++ b/build/ninja_gen/src/python.rs
@@ -22,32 +22,32 @@ pub fn python_archive(platform: Platform) -> OnlineArchive {
     match platform {
         Platform::LinuxX64 => {
             OnlineArchive {
-                url: "https://github.com/indygreg/python-build-standalone/releases/download/20240107/cpython-3.9.18+20240107-x86_64_v2-unknown-linux-gnu-install_only.tar.gz",
-                sha256: "9426bca501ae0a257392b10719e2e20ff5fa5e22a3ce4599d6ad0b3139f86417",
+                url: "https://github.com/astral-sh/python-build-standalone/releases/download/20250517/cpython-3.13.3+20250517-x86_64_v2-unknown-linux-gnu-install_only_stripped.tar.gz",
+                sha256: "68f65b2e8098ea332ae06f2afe984f6532a42eaa9a6fc5393c4421840d8dfd70",
             }
         }
         Platform::LinuxArm => {
             OnlineArchive {
-                url: "https://github.com/indygreg/python-build-standalone/releases/download/20240107/cpython-3.9.18+20240107-aarch64-unknown-linux-gnu-install_only.tar.gz",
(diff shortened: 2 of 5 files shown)


.gitmodules
```diff
@@ -8,7 +8,7 @@
 	shallow = true
 [submodule "qt/bundle/PyOxidizer"]
 	path = qt/bundle/PyOxidizer
-	url = https://github.com/ankitects/PyOxidizer.git
+	url = https://github.com/indygreg/PyOxidizer.git
```
andrewsanchez, line 11: I pinned the submodule to the [latest release tag](https://github.com/indygreg/PyOxidizer/releases/tag/pyoxidizer%2F0.24.0).

build/ninja_gen/src/python.rs
```diff
@@ -22,32 +22,32 @@ pub fn python_archive(platform: Platform) -> OnlineArchive {
     match platform {
         Platform::LinuxX64 => {
             OnlineArchive {
-                url: "https://github.com/indygreg/python-build-standalone/releases/download/20240107/cpython-3.9.18+20240107-x86_64_v2-unknown-linux-gnu-install_only.tar.gz",
-                sha256: "9426bca501ae0a257392b10719e2e20ff5fa5e22a3ce4599d6ad0b3139f86417",
+                url: "https://github.com/astral-sh/python-build-standalone/releases/download/20250517/cpython-3.13.3+20250517-x86_64_v2-unknown-linux-gnu-install_only_stripped.tar.gz",
+                sha256: "68f65b2e8098ea332ae06f2afe984f6532a42eaa9a6fc5393c4421840d8dfd70",
```
andrewsanchez, line 26: Any reason not to use the `install_only_stripped` versions?

> An install_only_stripped archive is also available. This archive is equivalent to install_only, but without debug symbols, which results in a smaller download and on-disk footprint.

https://gregoryszorc.com/docs/python-build-standalone/main/running.html?highlight=stripped

abdnh:
Updating `pythonVersion` in pyproject.toml then deleting the pyenv folder made this work for me on Linux.

abdnh:
On Windows, I also had to update the pinned pywin32 version.

dae:
Did you test bundling with this, via './ninja bundle'? Unfortunately, I suspect this may not be simple at the moment. While the standalone Python builds are still actively maintained as they're used in other projects like uv, I was under the impression that changes would also be required to PyOxidizer, and it's not received love since the 3.10 days. When I try to bundle on a Mac, I get:

```
downloading https://github.com/astral-sh/python-build-standalone/releases/download/20250517/cpython-3.13.3+20250517-aarch64-apple-darwin-pgo+lto-full.tar.zst
error[PYOXIDIZER_BUILD]: reading tar.zst distribution data
    
    Caused by:
        0: reading tar distribution data
        1: expected version 7 standalone distribution; found version 8
       --> qt/bundle/pyoxidizer.bzl:127:14
        |
    127 |     policy = dist.make_python_packaging_policy()
        |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ resolve_distribution
```

And the same occurs with even 3.10, since presumably this is a change made to the standalone build generation, rather than a change in Python itself.

.gitmodules
dae: It is needed - we have some custom patches necessary to build the bundle, and they'd need to be rebased over the latest code.

dae:
Given the amount of bit rot and security issues at this point, I think we're best off not sinking any more effort into PyOxidizer. Andrew and I are currently exploring briefcase and uv as alternative options.

PR #4024, opened 2025-05-20
Fix DR not being in percent in the forgetting curve

GithubAnon0000:
Fixes: https://forums.ankiweb.net/t/anki-25-05-beta/59710/54?u=anon_0000

anki


Expertium:
I suggest only displaying the first two digits, there is no reason to display the extra zeros

Expertium:
Wait, is the screenshot slightly blurry or is `1` in `10` actually blurry?
Compare how 1 looks like in `11` and how it looks like in `10`.
image

dae:
Thank you both!

PR #4025, opened 2025-05-20
Use non breaking spaces for names on about page

Luc-Mcgrady:
Prevents line wrapping mid name:
image

like this:
image

### DIFF
--- a/qt/aqt/about.py
+++ b/qt/aqt/about.py
@@ -225,7 +225,7 @@ def show(mw: aqt.AnkiQt) -> QDialog:
         )
     )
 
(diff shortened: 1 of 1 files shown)


qt/aqt/about.py
```diff
@@ -225,6 +225,7 @@ def on_dialog_destroyed() -> None:
         )
     )
 
+    allusers = [user.replace(" ", "&nbsp") for user in allusers]
```
iamllama, line 228:
```suggestion
    allusers = [user.replace(" ", "&nbsp;") for user in allusers]
```

dae:
Thank you both!

Issue #3836, opened 2025-02-27
Use of contenteditable in the editor


dae:
> Thanks for starting this refactor. I use Anki every day and spent a significant amount of time trying to figure out how the editor works to try to fix some bugs, but really couldn’t make heads or tails of it. I thought this would be right up my alley since I used to work on text input and editing for Chrome for Android, but there’s quite a lot of complexity in the current Anki editor code. Here are my notes (you can search “Reproducible bug #1” and “bugs to fix” for specific things I was trying to fix):
> https://docs.google.com/document/d/1SxH6PU43Lfoxq3bdB2IWl06spdzg2ihUC34T70Y5oaQ/edit?usp=sharing
> 
> When I worked on Chrome, I believe the general wisdom was that trying to directly implement a WYSIWYG editor using contenteditable and execCommand() is fraught with difficulty. Here’s an opinion paper my former manager and one of my coworkers wrote together in 2017 (I guess that was a long time ago, but I don’t think much has really changed):
> https://docs.google.com/document/d/10qltJUVg1-Rlnbjc6RH8WnngpJptMEj-tyrvIZBPSfY/edit?tab=t.0#heading=h.5x0d5h95i329
> 
> Basically, if you try to directly manipulate the web content, you’re going to have to deal with all the complexity in Chrome’s text editing code (which was forked from WebKit, which was in turn forked from KHTML, so some of this code is over 20 years old and probably no one really understands how it works anymore), and you’re going to be chasing bugs ad infinitum. The recommended way of making an editor that works correctly is for the editor to maintain its own editing state which the editor’s own code maps to HTML. This is actually fairly complex, so the editing experts generally recommend using a third-party editing library. See the below email from the W3C public-editing-tf mailing list:
> 
> https://lists.w3.org/Archives/Public/public-editing-tf/2024Sep/0000.html
> ExecCommand IS NOT a viable way of creating a richtext editor for almost
> any usecase and hasn't been for the past 10-15 years. Many years of
> developer lives have been wasted attempting to do so, which is why we
> added a huge warning to the beginning of the document trying to ensure
> that no one else wastes times trying to do that.
>
> We have spent the past decade creating the alternative you are asking for.
> Check out EditContext and Input Events. Input Events is already shipped in
> all major browsers.
>
> However, these provide quite foundational APIs and a lot of work has to be
> done in JavaScript to create a viable editor. For most users who need a
> custom JavaScript editor, I would personally recommend using one of the
[...]

Issue #2474, opened 2023-04-16
Main Window Rework: How should I deprecate unused modules?


kleinerpirat:
I have refactored a major part of Anki's front-end and merged a couple of formerly external windows into the main window. In the process, many files like `deckbrowser.py`, `overview.py`, `studydeck.py`, `customstudy.py`, the toolbar classes etc. are not needed anymore and I'd like to delete them. The webview is populated from a new file called `mainpage.py`.

What is the best practice to minimize add-on breakages due to failed imports? I saw the methods in `anki._legacy`, but these are for deprecating single functions and redirecting camelCase to snake_case.

It would be wonderful if someone gave me an example, say for `aqt.deckbrowser.DeckBrowser`. The new class is `aqt.mainpage.MainPage`. Not all methods of the former class are included in the new one. I'm ready to write redirection shims for depended-upon functions, but I don't know how to start.

(If you don't think it's worth the effort to minimize breakages for a 300 file diff - because some might be unpreventable -, I'd be happy too.)

*Edit:* Here's a little sneak peek:

<img src="https://user-images.githubusercontent.com/62722460/232347963-fdfb6d77-589a-4657-aa37-c6afcf79c112.png" width="48%"> <img src="https://user-images.githubusercontent.com/62722460/232348066-df0c1880-09f3-4b1a-9df4-bd391b47c797.png" width="48%">
![image](https://user-images.githubusercontent.com/62722460/232348170-919fa100-c3af-442a-9d31-9f2a7784532f.png)

kleinerpirat:
@glutanimate friendly ping because I know you're an expert on add-on compatibility.

dae:
The least disruptive approach is to have a period where both the old and new implementations are available, and allow the active one to be switched with an env var or flag + a restart. That will allow us to start with it off by default so that beta testers can opt in and easily revert if any showstopper problems are found, and will allow users who depend on old add-ons to continue to use the old path while we allow time for add-ons to get updated. Doing this will likely require a combination of files that check the flag to decide which routine to call, and some files may need two different versions for the two paths when they are radically different.

kleinerpirat:
When I started this work I was hoping to avoid this, but now it seems like the easiest solution. I just hope we don't end up in a situation like the one we got with the stats screen.

kleinerpirat:
Alright, I separated the two paths. Two files are radically different and need to coexist with a new version using the `NEXT` suffix:

- `main` -> `mainNEXT`
[...]

PR #4029, opened 2025-05-21
Shift editor control to TypeScript

### LINKED ISSUE #3830: Shift editor control from Python to TypeScript

abdnh:
Closes https://github.com/ankitects/anki/issues/3830


phu54321:
Some tricky case you might wanna consider?

https://www.gdcdental.com/files/product/Westcott-Curved-%23-S35.jpg

because `Editor._retrieveURL` correctly downloads the URL but unquotes `%23` to `#` upon return, and just assigning `Westcott-Curved-#-S35.jpg` to `img['src']` won't work, as `#` act as a hash tag. It seems like it's the same on your svelte code too?

abdnh:
> It seems like it's the same on your svelte code too?

Yes, it has the same issue. We can look into it once we're done porting & testing the existing copy & paste behavior.

abdnh:
@dae This is ready for some testing. Pinging @glutanimate @iamllama or anyone interested in giving feedback.

## Known issues

- [x] Sticky fields not being sticky.
- [x] Misaligned sticky badge.
- [x] Fields cleared when switching note type in Add screen.
- [x] Incomplete pre-add check (`noteCanBeAdded()`)
- [x] Undo operations for backend calls moved to TypeScript.
- [x] Toggling MathJax preview doesn't take effect until editor is reopened: https://github.com/ankitects/anki/pull/2014
- [x] "Paste Image from Clipboard" is less reliable - it doesn't recognize files copied from Explorer on Windows but does work with images copied directly from image viewers or the web.
- [x] `navigator.clipboard.read()` is a new API, so it needs testing on different platforms.
- [ ] editor.parentWindow.activateWindow() removed from select_image_and_occlude and similar. I assume this is needed for macOS?
- [x] Svelte implementations for Deck and Notetype selectors.
- [ ] The elephant in the room: Add-on compatibility.

GithubAnon0000 (commented):
I just build it and tried it. So far, it seems to work quite well (ignoring the issues you already stated). Other issues I found:
1. It was requested that the field text becomes selectable and, ideally, that buttons won't be part of the selection when pressing <kbd>Ctrl+A</kbd>. The first one is an easy fix and I opened https://github.com/ankitects/anki/pull/4122/ to solve it; for the second one, I have no idea. I initially thought your PR was about the review screen (no idea how that happened), so I opened above PR. Maybe it's better if I close it and you change it in your code?
2. The _Eraser_ dropdown menu has `hr`s after every entry. This is a bit too much and could be removed or designed differently.
3. Same as number 2, but for the settings wheel next to _bold_ button.
4. The *Tags* field is cut off a bit at the bottom (maybe an issue with responsive design? Try to reduce the window size in different ways, and you'll see that it's not always cutoff by a lot, and sometimes cut off at the side as well).
5. Related to 4: the scrollbar is cut off as well.
6. The tags don't stay sticky after adding a card.
7. After adding a card, the _History_ button in the bottom right corner doesn't update, meaning it stays grayed out.
8. If you open the _add_ dialog, then immediately click _close_, you'll get a popup asking to discard current input. There is no input though.
9. The right click menu, where _paste_ resides, needs styling to be consistent with ankis ui.
10. I never used this before, so am not sure whether this PR introduced it. If you click a deck to study, then click the *Edit* button during review on a card with an image, then right click said image in the editor and click on _Show in Folder_, the file is opened in the systems webbrowser instead (I would expect it to open the file manager). It does the same like the other option, _Open Image_.
11. Related to 10: If you copy the image via right click, _Copy Image_, the image isn't added to the clipboard. It does work, though, if you <kbd>Ctrl+a</kbd> on the image field and then <kbd>Ctrl+V</kbd> (but not via right-click paste menu).


That's all I could find.

abdnh:
Thank you for testing @GithubAnon0000. Regarding some issues:

- (1) I think let's discuss & handle that in your PR, since the changes here are unlikely to touch that part.
- (4) This might be fixed after #4115.
- (5) Hmm, can you see this after pulling the latest changes?
- (10) Let me know if you can reproduce it in the stable version. I suspect it's a system-specific issue. What's your OS?
- (11) What's the format of the image?

user1823:
Regarding 10, it was me who added the "Show in Folder" option to Anki and this option was not supposed to be shown in Linux as I don't know how to deal with the different File Managers there. I made it hidden on Linux in the Python code. Is it not hidden on Linux in the TS/Svelte code?

GithubAnon0000:
# Answering abdnh

> (1) I think let's discuss & handle that in your PR, since the changes here are unlikely to touch that part.

Got it.

> (4) This might be fixed after

Indeed. On 0241113e2cde8565821d8d09cd487162285af8ba this issue is solved.



> (5) Hmm, can you see this after pulling the latest changes?

Yes. I zoomed into the view to make it more apparent, but it's cut off without zoom as well:
anki
(Please ignore the weird stripes. Anki never flawlessly worked for me, no matter which graphics driver I choose.)



> * (10) Let me know if you can reproduce it in the stable version. I suspect it's a system-specific issue. What's your OS?

It doesn't happen in the stable version, as the option doesn't exist there. Probably because of what user1823 said:


> "Show in Folder" option to Anki and this option was not supposed to be shown in Linux

I'm indeed on linux (debian sid).

> (11) What's the format of the image?

I'm not entirely sure since I don't quite remember which cards I had tested with. Testing it now, however, reveals that it works with .png but doesn't with .svg files. **Edit:** .jpeg / .jpg doesn't work either.

# Answering user1823 
I don't know how you implemented it. But if you can call a command, then `xdg-open /path/to/directory` will work regardless of desktop environment / regardless of the default file manager.

abdnh:
@user1823 

> Regarding 10, it was me who added the "Show in Folder" option to Anki and this option was not supposed to be shown in Linux as I don't know how to deal with the different File Managers there. I made it hidden on Linux in the Python code. Is it not hidden on Linux in the TS/Svelte code?

Oh, yes. I forgot to hide it on Linux!

user1823:
> But if you can call a command, then xdg-open /path/to/directory will work regardless of desktop environment / regardless of the default file manager.

But, I can't test this. Maybe you can try your suggestion and submit a PR to fix it in the Python code? After looking through the changes in this PR, I realized that if the Python code is made to work on Linux, no additional changes will be required here.

The relevant Python code is here:
https://github.com/ankitects/anki/blob/630bdd31893d5113b81cd60612cef3ebf146d5b4/qt/aqt/editor.py#L1738-L1741
https://github.com/ankitects/anki/blob/630bdd31893d5113b81cd60612cef3ebf146d5b4/qt/aqt/utils.py#L927-L941

Edit: Based on a quick Google search, it seems that xdg-open will open the image, not highlight it in the file manager. So, it won't help here.

GithubAnon0000:
I changed the utils.py file of this PR locally to the following:
```py
def show_in_folder(path: str) -> None:
    if is_win:
        _show_in_folder_win32(path)
    elif is_mac:
        script = f"""
        tell application "Finder"
            activate
            select POSIX file "{path}"
        end tell
        """
        call(osascript_to_args(script))
    else:
        subprocess.run(["xdg-open", path])
        # Just open the file in any other platform
        #with no_bundled_libs():
        #    QDesktopServices.openUrl(QUrl.fromLocalFile(path))
```

It opens my file manager (nautilus / gnome files) and does highlight the correct image.

I cannot suggest this change in this PR, since github won't let me. I'll open a separate PR.

iamllama:
Thanks for this, it's going to make working on the editor much faster! Noticed a few additional issues as well:
1. The add window now flashes like the stats window, maybe setting the webview bg colour to canvas would help here
2. The "Copy", "Cut" and "Paste" context menu actions don't seem to work on win 10 or wsl for me
3. On dark mode, the context menu's text is white
4. "Discard changes" shows when trying to close after adding a note
5. "Create copy" seems to be broken as well, but differently depending on the notetype
    - Cloze notes: clicking in and out of a field shows an error. On the console it prints `JS error /_app/immutable/chunks/BjIIu02o.mjs:2 Uncaught (in promise) Error` (no AnkiWebView Inspector :/)
    - IO notes: the editor loads and shows the masks/image, but is otherwise not functional (can't change tool)
    - Other types: fields are just blank
6. ~~When adding cloze notes without any clozes, the `You have a cloze notetype but have not made any cloze deletions. Proceed?` dialog doesn't appear anymore. Clicking on the `Add` button just does nothing until a cloze is added~~
7. ~~"The first field is empty" dialog doesn't show either (related to 6.)~~
8. When in the browser with a selected row and holding "Down", an error dialog is sometimes shown. When closed (or when the "Down" key is released if no error), the editor appears to rapidly cycle through each note before stopping at the last row when instead it should be debounced


Haven't been able to go through all the commits yet, will get to it when possible!

GithubAnon0000:
tatsumoto-ren posted the following in [#3830](https://github.com/ankitects/anki/issues/3830#issuecomment-3006354480):

> Rewriting the editor in pure Qt will achieve the following:
> 
> […]
> * Will work fast unlike the Electron-like experience we have today.
> 
> […]

<hr>

Is there a way to benchmark this and check which version (current implementation or implementation from this PR) is faster? I'd imagine the performance shouldn't be too bad with svelte, typescript, … considering the views that would be opened are low complexity and not many things run in parallel anyways.

tatsumoto-ren:
>the performance shouldn't be too bad

If you compare webview windows with pure-Qt windows in Anki, you'll notice that Qt windows appear instantly, while webviews can take nearly half a second, or even close to a full second, to load. The worst example is the "deck options" dialog, which used to perform much better in the past and had a more minimalist design.

screenshot-2025-06-26-20-53-25

I think we can now confidently say that continuing to rely on webviews will negatively impact performance and the overall user experience.

brishtibheja:
just fyi, dae once said

> While we load a separate HTML page for each feature, things are going to be a bit slower than native code. Later on once we can start reducing the number of separate webviews, loading performance should improve.

phu54321:
Hope one day we get a pure electron-based webapp of Anki that we can also access on https://ankiweb.net/

brishtibheja:
That'll probably never happen given AnkiWeb is strategically designed to not blow up AnkiMobile sales.

phu54321:
If this code moves all editor-related code to typescript, what API's will be exposed to allow custom buttons?

```py
def add_buttons(buttons: List[str], editor: Editor) -> None:
    shortcut = "Ctrl+Alt+Shift+H"

    def _(editor: Editor) -> None:
        editor.web.eval("setFormat('inserthtml', '<img src=_cha_cha-enable.png>');")

    buttons.append(
        editor.addButton(
            icon=None,
            cmd=f"add_cloze_hide_all_marker",
            func=_,
            tip="Make this note like 'Cloze (Hide All)'",
            label="█ CHA",
            keys=shortcut,
        )
    )


gui_hooks.editor_did_init_buttons.append(add_buttons)
```

For example, how should I port this code?

abdnh:
@phu54321 editor.addButton() will keep working. Most editor hooks are broken in this PR right now though, and we still need to think about reducing add-on breakages (and probably introduce a JS API).

dae:
Very excited to see the progress here @abdnh! Hope to have this properly reviewed by the end of the week.

> That'll probably never happen given AnkiWeb is strategically designed to not blow up AnkiMobile sales.

While there is an aspect of that, another significant concern is who will pay for the server resources. CPU or I/O-costly features cost us nothing to run on user's own computers, but that's not the case for a hosted service.

>> tatsumoto-ren posted the following in https://github.com/ankitects/anki/issues/3830#issuecomment-3006354480:

> Rewriting the editor in pure Qt will achieve the following:

I've moved that to https://github.com/ankitects/anki/issues/4142 so we keep this PR on topic.

qt/aqt/mediasrv.py
```diff
@@ -599,6 +602,210 @@ def handle_on_main() -> None:
     return b""
 
 
+def editor_update_note() -> bytes:
```
abdnh: mediasrv is a bit of a mess now with all this screen-specific logic. Will look into a cleaner solution, maybe by allowing screens to register their own handlers.

qt/aqt/mediasrv.py
```diff
@@ -779,3 +1070,6 @@ def _extract_dynamic_get_request(path: str) -> DynamicRequest | None:
         return legacy_page_data
     else:
         return None
+        return None
+        return None
+        return None
```
dae: ? :-)

qt/aqt/utils.py
```diff
@@ -125,6 +128,49 @@ def openLink(link: str | QUrl) -> None:
         QDesktopServices.openUrl(QUrl(link))
 
 
+def retrieve_url(url: str) -> tuple[str, str]:
+    "Download file into media folder and return local filename or None."
```
dae: It would be nice if we could move this into Rust at one point, so the clients fetch the files consistently.

qt/aqt/mediasrv.py
```diff
@@ -599,6 +602,242 @@ def handle_on_main() -> None:
...
+            cb=cast(Callable[[Any], None], cb),
+            filter=f"{req.filter_description} ({' '.join(f'*.{ext}' for ext in req.extensions)})",
+            key=req.key,
+        )
+
+    aqt.mw.taskman.run_on_main(on_main)
+
+    filename = await future
+
+    return generic_pb2.String(val=filename if filename else "").SerializeToString()
```
dae, line 825: Out of curiosity, what prompted you to reach for async here? I've used it with gRPC on AnkiWeb a few years back, and think it would improve the desktop code readability too, but integration into the Qt event look looked complicated when I looked into it in the past.

dae:
Sorry it's taken me a while to circle back to this Abdo. I'm going to start reading through the changes and start testing on iOS so I can get up to speed with things.

> The elephant in the room: Add-on compatibility.

We'll of course want to do what we can here, but I presume there'll be some workflows that can't be easily adapted. I think our best path forward here might be to have a transitionary period where the old and new screens are both available and can be switched between, like as talked about on https://github.com/ankitects/anki/issues/2474#issuecomment-1510775549. That will allow us to iterate on this after the initial merge without any immediate impact on users, and reduces the pressure on us to get this perfect from the start.

The changes in editor.py look fairly extensive, and it may be easiest to copy in a version of the file prior to the changes. For the other files like editcurrent.py, having separate code paths for the two cases might be simplest? For the TS side, you probably have a better idea of whether it would make sense to clone some of the files, the entire ts/editor, or add separate code paths to the existing files. WDYT?

It will be a bit of a pain to make both the old and new approach work at once, and will leave the codebase in a messier state. I'm hoping that it's something we'd have in place for <6 months, not for years like the old deck options/stats screens, as it's too much code to try and maintain multiple copies of for long periods.

Preserving the git history of your changes is more important than the legacy codepath, so I think it might make more sense to put any copies of the old code we make in a new file, rather than copying your changes into a new file (then importing them into the old location so add-ons still find things at the old location). But other ideas welcome.

Will trickle in some more comments as I go through this.

qt/aqt/mediasrv.py
dae: Also, I noticed a couple of changed files in python/ got left around.

qt/aqt/mediasrv.py
dae: AnkiMobile currently does this - each screen registers only the endpoints that are needed, so other endpoints give an error. As we get further along in our porting and we start transitioning from one 'page' to another in the same webview session, it'll probably become less practical to do that, though.

qt/aqt/mediasrv.py
abdnh: Blame Cursor suggestions 😅 It sometimes decides to duplicate the last line in the file for some reason.

qt/aqt/mediasrv.py
abdnh: It's required to get the return value of getFile() since we call that on the main thread? async looked like the most natural approach.

tatsumoto-ren:
Since these changes negatively impact the user experience on desktop, it would be better if they were implemented in an add-on instead. Alternatively, there should be a way to completely and permanently opt out of them. I'm not a fan of transition periods, as they don’t solve the problem, they only postpone it.

https://github.com/ankitects/anki/issues/4142#issuecomment-3059469064

abdnh:
> That will allow us to iterate on this after the initial merge without any immediate impact on users, and reduces the pressure on us to get this perfect from the start.

That's reasonable. I was worried figuring out add-on compatibility would take a lot of time and discussions.


> The changes in editor.py look fairly extensive, and it may be easiest to copy in a version of the file prior to the changes. For the other files like editcurrent.py, having separate code paths for the two cases might be simplest? For the TS side, you probably have a better idea of whether it would make sense to clone some of the files, the entire ts/editor, or add separate code paths to the existing files. WDYT?

I agree it will be simpler to have separate Python files for the new code. For the TS side, most of my changes consist of replacing bridgeCommand() calls with a backend call and some logic ported from Python. As there's not much change to the code flow, I feel the simplest option to add a legacy switch to existing code to make it go back to bridgeCommand handling.

phu54321:
This looks like the old Anki window.

image

I guess this PR is to resolve future technical debt, but I think this could be re-implementing some other WYSIWYG toolkits like [summernote](https://summernote.org/) or [tinymce](https://www.tiny.cloud/). 

I once used Summernote for making [kian](https://github.com/phu54321/kian/blob/develop/src/components/editor/Summernote/Summernote.vue), so maybe you can look at it. Better to avoid reinventing the wheels?

https://ankiweb.net/shared/info/805891399 this addon uses tinymce, but I heard it has some licensing issues.

abdnh:
> I guess this PR is to resolve future technical debt, but I think this could be re-implementing some other WYSIWYG toolkits like [summernote](https://summernote.org/) or [tinymce](https://www.tiny.cloud/).

This is out of the scope of this PR, but it's tracked here: https://github.com/ankitects/anki/issues/3836

dae:
This is an impressive amount of work Abdo, and I'm sorry I let this stall for so long.

Currently this uses the new editor by default, which is probably going to be disruptive. We might want to consider defaulting to the old editor instead (and making sure both addcards,editcurrent and browser all support the shift toggle).

Couple of minor issues I noticed while testing:
- editor_did_init is called twice (by __init__ and a later editorReady)
- "1 card added" got stuck for me when I added the second or third card, and did not go away.
Screenshot 2026-04-25 at 20 10 51
- Minor tweaks to layout/formatting to make things look neater - I think Luc may have brought that up above too.

tatsumoto-ren:
How about releasing this as an add‑on first, and if it gains enough popularity, consider adding it to the main program? I think it could cause too many breakages if it's merged directly. Also, you wouldn't gain anything from including it in Anki right now, since the functionality already exists in the current implementation.

Luc-Mcgrady:
The large margin is caused by the margin and spacing of the "verticalLayout_3" widget in the UI_Dialog 
https://github.com/ankitects/anki/blob/71ae5a6b673bc8943f048fed906751febfe585bd/qt/aqt/forms/addcards.ui#L26
QT:imageTypescript:imageWith the spacing removed it typescript looks like QT:
image

You'll have to fix it programmatically somehow  because this change also affects the QT reviewer :( 

Also if I open "edit current" (pressing e inside the reviewer) without holding shift it is broken. I get the error "unexpected editor type: review"

Hope this helps~

abdnh:
@tatsumoto-ren 

> I think it could cause too many breakages if it's merged directly.

It's disabled by default and will be put behind a preferences toggle (#4871) soon while we look into add-on compatibility and the long-term plan.

> Also, you wouldn't gain anything from including it in Anki right now, since the functionality already exists in the current implementation.

We can at least get some beta testing to catch regressions and maybe test it on AnkiMobile/AnkiDroid.

PR #4033, opened 2025-05-24
Align LaTeX with surrounding text

slotThe:
So far, LaTeX images were just rendered and included as pictures as-is. However, the bounding box of the image might not necessarily coincide with the actual baseline of the equation. By using preview.sty (which is hopefully not that big of an additional dependency), we can however access the "depth" -- the offset from the baseline -- of the generated image and correct for that.

---

The Anki manual sports a picture of the default PNG rendering:

I've generally found it to be a bit better with SVGs + some CSS that just vertically centres the pictures:

2025-05-24-132320_798x111_scrot

This PR gives a further improvement on that:

2025-05-24-132358_799x123_scrot

---

+ Alignment is not perfect all the time (at least I've spotted a few weird outliers, of which I however forgot to take screenshots), for which I have no clear explanation at this point.

+ I don't know the codebase at all, so it's very possible that I'm doing something very stupid -- feel free to shout at me :)

+ I've tested this manually with `./run`, but haven't managed to properly build the tests yet. I would imagine that `test_latex.py` might fail due to it checking the length of the media directory (where we now have more files). Since I couldn't actually test it, though, I refrained from making any changes here. My hope is that a possible CI will catch this

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -230,6 +230,7 @@ KolbyML <https://github.com/KolbyML>
 Adnane Taghi <dev@soleuniverse.me>
 Spiritual Father <https://github.com/spiritualfather>
 Emmanuel Ferdman <https://github.com/emmanuel-ferdman>
+Tony Zorman <mail@tony-zorman.com>
 
 ********************
 
--- a/pylib/anki/latex.py
+++ b/pylib/anki/latex.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import html
 import os
+import re
 from dataclasses import dataclass
 
 import anki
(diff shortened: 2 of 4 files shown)


dae:
Hi Tony,

It's a well-formed PR, and I appreciate you taking the time to make it, but I'm afraid I'm not very keen on this change, for the following reasons:

- It has the potential to break things for other users, which the outliers you encountered may be foreshadowing. If it does, it may need reverting or changes and extra releases.
- LaTeX remains low down on the list of priorities, as it can only be created on computers, and there is a steady trend towards mobile devices.
- It requires twice as many files in the media folder, and storing lots of little files adds overhead to AnkiWeb.
- If you referenced the add-on when making this, it could be argued that your code needs to fall under that add-on's license, and can't be declared as BSD.

Have you considered updating the add-on to work with modern Anki versions instead? I imagine it could still be made to work.

PR #4034, opened 2025-05-26
calculate accurate retrievability in card info by elapsed seconds

L-M-Sherlock:
Before:

image

After:

image

The new one is close to the R displayed in the forgetting curve:

image


user1823:
After https://github.com/open-spaced-repetition/fsrs-rs/pull/333 by Luc, the code in FSRS-rs looks much cleaner. So, you can disregard my suggestion above as it's probably cleaner to create a new function for handling elapsed time in seconds if we are going to use it in multiple places.

PR #4040, opened 2025-05-29
Add left margin to browser when sidebar is closed

### LINKED ISSUE #4037: Add padding to left side when browser sidebar is hidden
When the browser sidebar is hidden (https://github.com/ankitects/anki/pull/3953), there is no padding on the left side.

iamllama:
Closes #4037

After:

image

### DIFF
--- a/qt/aqt/browser/browser.py
+++ b/qt/aqt/browser/browser.py
@@ -170,7 +170,7 @@ class Browser(QMainWindow):
         if self.height() != 0:
             self.aspect_ratio = self.width() / self.height()
         self.set_layout(self.mw.pm.browser_layout(), True)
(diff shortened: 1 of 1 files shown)


iamllama:
> The sidebar stays empty.

Can't reproduce this on linux or win10

dae:
Thanks! Since we're looking at this area, I took the opportunity to fix the messy standard margin on macOS as well.

iamllama:
@dae after 18a3dbc, there's now a margin on the handle when the sidebar is visible, and none otherwise. I don't think the `not` before `visible` was intended

dae:
Whoops, that was silly. Thanks for letting me know.

PR #4046, opened 2025-06-02
Feat/Card ID special field

Luc-Mcgrady:
ref: https://forums.ankiweb.net/t/add-a-new-special-field-replacement/51725/12?u=sorata

image

### DIFF
--- a/rslib/src/notetype/render.rs
+++ b/rslib/src/notetype/render.rs
@@ -183,6 +183,8 @@ impl Collection {
             .or_insert_with(|| flag_name(card.flags).into());
         map.entry("Card")
             .or_insert_with(|| template.name.clone().into());
+        map.entry("CardID")
+            .or_insert_with(|| card.id.clone().to_string().into());
 
         Ok(())
     }


bryanboateng:
Hey @Luc-Mcgrady, I have a question.
In the code base above the `add_special_fields` function there is a comment that states:

```rust
/// Add special fields if they don't clobber note fields.
/// The fields supported here must coincide with SPECIAL_FIELDS in
/// notetype/mod.rs, apart from FrontSide which is handled by Python.
```

To me that reads that any field added to the `add_special_fields` function also needs to be in `SPECIAL_FIELDS`.
You didn't add `"CardID"` to `SPECIAL_FIELDS`. Is that intentional? If so, why?

Luc-Mcgrady:
Well spotted! Seems like an oversight to me. I'll try and add a test for the future.

bryanboateng:
I can also add it. Though writing the test will take me a bit longer than you.

Luc-Mcgrady:
Too late :sweat_smile:

bryanboateng:
No problem, thanks for the fix!

Issue #4036, opened 2025-05-27
Add a "web" translation module for docs & landing page


soleuniverse101:
As I understand, there are currently only 2 modules for translations : `core` and `qt`.

I'm suggesting creating a third one that would be named `web`, which would contain translation strings related to Anki's landing page (see this ) and for its docs, which are not always up to date AFAIK (for example the French one isn't even in the same format). Note that I thought of this 

I'm making this issue because I'm not sure if that's the best option, given the fact that docs are significant in size

soleuniverse101:
I made a proof of concept at https://github.com/soleuniverse101/anki-manual-i18n-poc using Google's Comprensive Rust book (https://github.com/google/comprehensive-rust) as inspiration. It's based on the packages made available by https://github.com/google/mdbook-i18n-helpers.

It parses the "sections/blocks" of each page and associates it to a locale string in a template used to generate translation files for each language. I think that it's a nice way of doing it because docs will stay more or less in the same structure whatever the language, though it still leaves translators some freedom to rephrase paragraphs nicely as long as they convey the same info.

You can check the result at https://soleuniverse101.github.io/anki-manual-i18n-poc/intro.html by changing the language to French using the top-right corner. (I only translated a few strings as it's meant to be an example of what it could be like at the end...)

For images, a cool thing is that mdbook-i18n-helpers also parses urls as translatable elements, therefore we could also specify different links for images (e.g. showing a page with Anki in the target language).

An obvious drawback with such an endeavor is the massive process of migrating the already available Anki manuals in other languages to the desired format. However I think that it could be done progressively, while keeping the current translated docs at the moment. Still, in the long run, having a single source of docs (the English version) upstream, with translated versions that follow the same structure, and whose translation strings are immediately updated when the original version changes would help keeping docs consistent and up-to-date as Anki keeps receiving new features. Also, it's not exactly a retranslation from zero, some of the docs roughly follow the structure of the https://docs.ankiweb.net/ version (such as the French one, which only has an outdated style, or the Chinese one, which is already a perfect replica). Therefore for some of them at least, the job will mainly be copy & paste. Though as mentioned in https://github.com/ankidroid/ankidroiddocs/pull/101#issuecomment-1943817712, it could still need proof reading and a non negligeable effort from translators to try and use the old docs to not have to rewrite everything from scratch while migrating entries one-by-one.

[...]

PR #4047, opened 2025-06-02
Feat/Health check

### LINKED ISSUE #3926: Improve feedback from "Evaluate" and potentially hide/move it (FSRS section)

Luc-Mcgrady:
- closes #3962
- closes #3926

https://github.com/user-attachments/assets/e3b6c561-f602-4089-8444-acbdf856abce


ts/routes/deck-options/FsrsOptions.svelte
```diff
@@ -41,6 +41,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     const defaults = state.defaults;
     const fsrsReschedule = state.fsrsReschedule;
     const daysSinceLastOptimization = state.daysSinceLastOptimization;
+    const logLossBadThreshold = 0.5;
```
Luc-Mcgrady: Any ideas what this value should be?

brishtibheja: what percentile of users see this?

user1823: @Expertium

rslib/src/scheduler/fsrs/params.rs
```diff
@@ -108,12 +111,13 @@ impl Collection {
...
+        let input = ComputeParametersInput {
             train_set: items.clone(),
             progress: Some(progress.clone()),
             enable_short_term: true,
             num_relearning_steps: Some(num_of_relearning_steps),
-        })?;
+        };
+        let mut params = fsrs.compute_parameters(input.clone())?;
         progress_thread.join().ok();
         if let Ok(fsrs) = FSRS::new(Some(current_params)) {
```
user1823:
```suggestion
        if let Ok(current_fsrs) = FSRS::new(Some(current_params)) {
```

rslib/src/scheduler/fsrs/params.rs
```diff
@@ -108,12 +111,13 @@ impl Collection {
...
             let current_log_loss = fsrs.evaluate(items.clone(), |_| true)?.log_loss;
```
user1823:
```suggestion
            let current_log_loss = current_fsrs.evaluate(items.clone(), |_| true)?.log_loss;
```

rslib/src/scheduler/fsrs/params.rs
user1823: Couple this with deleting line 130. (some cleanup since we are already in the area)

ts/routes/deck-options/FsrsOptions.svelte
Expertium: Please wait for a day or two, I'll make a correction function that corrects log loss based on n(reviews) and retention and we'll use the 95th percentile of the corrected value.

ts/routes/deck-options/FsrsOptions.svelte
```diff
@@ -187,12 +189,21 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
                         resp.params.length === 0;
 
-                    if (already_optimal) {
+                    const highLogLoss =
+                        resp.logLoss && resp.logLoss > logLossBadThreshold;
+
+                    if (resp.logLoss) {
+                        console.log(`FSRS-test-train-split-log-loss = ${resp.logLoss}`);
+                    }
+                    if (highLogLoss) {
```
user1823: I believe that we should still display a message if the user has enabled health check and the loss is in the normal range. Otherwise, the user will say "I enabled the health check, it took so long and then nothing happened".

user1823 (commented):
Left some minor comments. Rest LGTM.

ftl/core/deck-config.ftl
```diff
@@ -485,6 +486,14 @@ deck-config-percent-input = { $pct }%
 deck-config-optimizing-preset = Optimizing preset { $current_count }/{ $total_count }...
 deck-config-fsrs-must-be-enabled = FSRS must be enabled first.
 deck-config-fsrs-params-optimal = The FSRS parameters currently appear to be optimal.
+deck-config-fsrs-bad-fit-warning = Your memory is difficult for FSRS to predict. Recommendations:
+
+    - Suspend or reformulate leeches.
+    - Use the answer buttons consistently. Keep in mind that "Hard" is a passing grade, not a failing grade.
+    - Understand before you memorize.
+
+    It may take weeks or months before parameters become optimal.
```
user1823:
```suggestion
    After you follow these suggestions, it may take weeks or months before parameters become optimal.
```

Otherwise, it seems that the user just needs to do more reviews with FSRS enabled.

ftl/core/deck-config.ftl
```diff
@@ -485,6 +486,14 @@ deck-config-percent-input = { $pct }%
 deck-config-optimizing-preset = Optimizing preset { $current_count }/{ $total_count }...
 deck-config-fsrs-must-be-enabled = FSRS must be enabled first.
 deck-config-fsrs-params-optimal = The FSRS parameters currently appear to be optimal.
+deck-config-fsrs-bad-fit-warning = Your memory is difficult for FSRS to predict. Recommendations:
+
+    - Suspend or reformulate leeches.
```
user1823, line 491: Maybe we should link to https://docs.ankiweb.net/leeches.html here?

ftl/core/deck-config.ftl
```diff
@@ -485,6 +486,14 @@ deck-config-percent-input = { $pct }%
...
+    - Use the answer buttons consistently. Keep in mind that "Hard" is a passing grade, not a failing grade.
```
user1823, line 492: and to https://docs.ankiweb.net/studying.html#answer-buttons here

ftl/core/deck-config.ftl
Expertium: How will it work in terms of translations? Especially if there is no manual in the language that the user uses.

ts/routes/deck-options/FsrsOptions.svelte
```diff
@@ -323,17 +334,28 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
...
-        disabled={!checkingParams && computing}
-        on:click={() => checkParams()}
-    >
-        {#if checkingParams}
-            {tr.actionsCancel()}
-        {:else}
-            {tr.deckConfigEvaluateButton()}
-        {/if}
-    </button>
+    <label>
```
iamllama:
```suggestion
    <label style="margin-left: 0.5em;">
```

Purely a matter of aesthetics as the label can also be clicked on, this just makes it a tad less crowded

ftl/core/deck-config.ftl
Luc-Mcgrady: I thought this part was for the case that the user has done nothing wrong but have a small amount of reviews

> I'll make a correction function that corrects log loss based on n(reviews)

Would this solve this?

ftl/core/deck-config.ftl
Luc-Mcgrady: I don't think its simple to add links to alert boxes.

ts/routes/deck-options/FsrsOptions.svelte
Luc-Mcgrady: you can use `style:margin-left={"0.5em"}` in svelte in case you didn't know https://svelte.dev/docs/svelte/style :wink:

ftl/core/deck-config.ftl
Expertium: The correction is just something that I was planning to do anyway. It's better than using a fixed cutoff for logloss because it depends on n(reviews) and on retention (a lot, actually)

Luc-Mcgrady: I meant to ask if mentioning the number of reviews at all would be appropriate given that the number of reviews will be adjusted for.

Expertium: Ah. Honestly, I don't really like that part of the text, so I wouldn't mind if it was removed

user1823: This part of the message is not about the number of reviews. It's meant to tell the user that FSRS won't become good at predicting their memory immediately after they correct their rating/studying habits.

ftl/core/deck-config.ftl
GithubAnon0000: Regarding the manual language:

It wouldn't be an issue if https://github.com/ankitects/anki/issues/4036#issuecomment-2924058050 was implemented, as the user could just change the language using the globe icon.

Rct567:
Wouldn't it be better if the button displayed "And check health (slow)"?

GithubAnon0000:
> Wouldn't it be better if the button displayed "And check health (slow)"?

I tend to agree. Currently it's not clear whether the user needs to do anything in addition to checking that checkbox or if it will automatically health check whenever the user is e.g. studying.

I know this PR isn't supposed to be a bigger redesign, but I'd argue something like the following might make it more obvious:
anki

ts/routes/deck-options/FsrsOptions.svelte
Expertium: I analyzed the first 2000 users. Turns out, the correlation between logloss and n(reviews) is very weak, so we can just drop that part. But logloss is really strongly correlated with retention (well, we knew it already), so we definitely need a correction for that. It's going to take me 22-26 more hours to finish benchmarking: I can't reuse already existing files because we recently changed FSRS-6's default parameters and Jarrett hasn't updated the benchmark repo yet, so I have to run it myself.

ftl/core/deck-config.ftl
Expertium: I like "After you follow these suggestions, it may take weeks or months before parameters become optimal." more than the previous one, btw

ftl/core/deck-config.ftl
```diff
@@ -485,6 +485,14 @@ deck-config-percent-input = { $pct }%
 deck-config-optimizing-preset = Optimizing preset { $current_count }/{ $total_count }...
 deck-config-fsrs-must-be-enabled = FSRS must be enabled first.
 deck-config-fsrs-params-optimal = The FSRS parameters currently appear to be optimal.
+deck-config-fsrs-bad-fit-warning = Your memory is difficult for FSRS to predict. Recommendations:
```
Expertium, line 489: "Your memory is difficult for FSRS to adapt to" sounds better IMO
(I always try to make a Github suggestion instead of a comment and always forget how to do it)

Luc-Mcgrady: image

```suggestion
deck-config-fsrs-bad-fit-warning = FSRS is having difficulty adapting to your memory. Recommendations:
```
?

ftl/core/deck-config.ftl
```diff
@@ -395,6 +395,7 @@ deck-config-weights = FSRS parameters
 deck-config-compute-optimal-weights = Optimize FSRS parameters
 deck-config-compute-minimum-recommended-retention = Minimum recommended retention
 deck-config-optimize-button = Optimize Current Preset
+deck-config-health-check = Health check (Slow)
```
dae:
```suggestion
deck-config-health-check = Health check (slow)
```

ftl/core/deck-config.ftl
dae: Some thoughts:
- Keeping the grammar/vocab as simple as possible without compromising clarity makes it more accessible for non-native speakers
- We want to avoid implying the user should disable FSRS

Of the two suggestions above, Expertium's seems preferable due to the grammar.

Throwing out some other ideas:

"Your memory is difficult to predict optimally. Recommendations:" (implying that it still can be predicted to some degree)

Or maybe we should just launch straight into a suggestion?

"Some recommendations to help FSRS predict your memory better:"

"Some suggestions to help FSRS perform better:"

Not suggesting we go with those over your suggestions, just trying to discuss.

And for the end:

"After you follow these suggestions, it may take weeks or months before parameters become optimal." → would it make sense to say something like "before this message disappears" instead of "parameters become optimal"?

proto/anki/scheduler.proto
```diff
@@ -354,11 +354,13 @@ message ComputeFsrsParamsRequest {
   repeated float current_params = 2;
   int64 ignore_revlogs_before_ms = 3;
   uint32 num_of_relearning_steps = 4;
+  bool health_check = 5;
 }
 
 message ComputeFsrsParamsResponse {
   repeated float params = 1;
   uint32 fsrs_items = 2;
+  optional float logLoss = 3;
```
dae: snake_case

rslib/src/scheduler/fsrs/params.rs
```diff
@@ -55,6 +55,7 @@ impl Collection {
     /// Note this does not return an error if there are less than 400 items -
     /// the caller should instead check the fsrs_items count in the return
     /// value.
+    #[allow(clippy::too_many_arguments)]
```
dae: I'm guilty of doing this too, but let's do it properly and split the args into a struct instead.

ts/routes/deck-options/FsrsOptions.svelte
dae: Speaking of which, weren't we going to make it on by default?

dae: Would it make sense to reuse our 'your parameters appear to be optimal' message in the ok case?

dae:
> > Wouldn't it be better if the button displayed "And check health (slow)"?
> 
> I tend to agree. Currently it's not clear whether the user needs to do anything in addition to checking that checkbox or if it will automatically health check whenever the user is e.g. studying.
> 
> I know this PR isn't supposed to be a bigger redesign, but I'd argue something like the following might make it more obvious: anki

It looks good, but it's a bit inconsistent with the rest of the page. Maybe we should be moving params to the bottom (or even collapsing them by default)?

@Luc-Mcgrady, you've made this a preset option and not a global. Was that out of expediency, or because the flexibility is desireable? I'm a bit doubtful users would want to enable the check on some presets but not others, assuming we integrate it into the optimize

ts/routes/deck-options/FsrsOptions.svelte
Expertium: Technically, not quite. Health check isn't about current parameters or even *any* specific parameters; it's about whether FSRS is overall good at predicting the probability of recall for you.

I think we should add a new message specifically for the health check, like "FSRS can adapt to your memory well!"

Luc-Mcgrady:
> you've made this a preset option and not a global. Was that out of expediency, or because the flexibility is desireable? I'm a bit doubtful users would want to enable the check on some presets but not others, assuming we integrate it into the optimize

I'll change it, I think I took this too literally.

https://github.com/ankitects/anki/pull/3962#issuecomment-2927305958

dae:
Sorry Luc, that was my fault.

ts/routes/deck-options/FsrsOptions.svelte
brishtibheja: > Speaking of which, weren't we going to make it on by default?

Yeah, otherwise some newbs might never get a health check.

ftl/core/deck-config.ftl
brishtibheja: I have said this before that there will be some users who'll follow these guidelines and their log-loss can still come up high. So, at least we should avoid using negative language like "your memory is hard to predict".

> Some suggestions to help FSRS perform better

That sounds good to me.

> would it make sense to say something like "before this message disappears" instead of "parameters become optimal"?

Also, as expertium said it isn't about params but more about "your log-loss is very high compared to other users using FSRS with the same retention".

brishtibheja:
Folks, did anyone of you think about the usecase of performing the health check multiple times? I mean with evaluate, it makes sense to have a seperate option because it gets you precise numbers. But health check is literally just a bunch of recommendations. Once you see it, there's nothing else that you can do about it.

Let's think about it from the user's perspective when health check is turned off by default. If the user wants to improve the performance of FSRS, regardless of what health check says the user has to follow the same reccs. Health check doesn't change anything.

I'm basically having a hard time imagining the X in the statement "The user does a health check when they need X".

___

Maybe it makes more sense to automate it and have this run the first time you optimise and after that every 4 months or so?

ftl/core/deck-config.ftl
Expertium: How about "Your memory is difficult for FSRS to predict accurately" for the first sentence and "Follow these suggestions, and FSRS will typically adapt over the next few weeks or months." for the last sentence?

dae:
I'm having a bit of trouble following you, sorry. Firstly, didn't we decide to have it on by default? That would seem to make more sense for new users, who are more likely to have non-optimal params.

I am also assuming that:
1) If users see a warning when optimizing, they'll be disinclined to disable the check (or periodically re-enable it) until they get a clean bill of health.
2) A user is much more likely to transition from bad→good than vice versa, thus it's not a big deal if established users leave it off permanently.

ftl/core/deck-config.ftl
dae: Adding "accurately" softens it a bit, but it still feels a bit more serious than the "Some suggestions ..." one. Anyone else want to weigh in so we can figure out a majority?

For the latter, how about this tweak: "If you follow these suggestions, performance will usually improve over the next few weeks or months."

(the imperative + 'and' construct seems a bit difficult for some non-native speakers)

Or we could just overestimate and say "over the next few months" to make it simpler to read, and less likely to worry users if it's not happening quickly?

ts/routes/deck-options/FsrsOptions.svelte
dae: Yep, fair point - different message it is. Still have a minor preference for 'predict' over 'adapt to', though: https://github.com/ankitects/anki/pull/4047#discussion_r2122018611

brishtibheja:
> Firstly, didn't we decide to have it on by default?

My bad, that was a bad example.

> A user is much more likely to transition from bad→good than vice versa, thus it's not a big deal if established users leave it off permanently.

I think this thought experiment will help: imagine a scenario opposite to the one you described here. So, for some technically obscure reason, experienced users are very likely to transition from good to bad. In this case, what utility does the health check offer? The experienced user already knows the good practices she has to follow (good button usage, seperate content by presets, and so on). And knowing whether or not some preset has high log-loss doesn't change much of that.

It does come in use to tell you whether or not you have too many leeches in circulation for example, or when someone is a newb and doesn't know how to use answer button. My point is simply that "health check" isn't like evaluate and no one actually needs to use it again and again everytime. It's use case is much limited to justify having a checkbox for it.

___

Maybe not in this PR, but automating this in the future would be a good thing. So run this health check for newbs and after that, maybe everytime number of reviews doubles or by some other rule.

dae:
I appreciate the point you're trying to make, but what if the established user doesn't want unpredictable slowdowns when optimizing, or wants to check everything is ok more frequently than whatever automatic threshold we come up with?

Edit: we could automatically re-enable the checkbox after a certain period has elapsed, but I can almost guarantee some users will assume that's a bug and report it to us.

ts/routes/deck-options/FsrsOptions.svelte
Expertium: @Luc-Mcgrady I made two approximation functions, one for log loss and one for RMSE. The one for log loss is a function of one variable - retention, the one for RMSE is a function of two variables: retention and n(reviews). Also note that I divided n(reviews) by 1000.

```
    ## For Log Loss
    def func(a, b, c, x):
        # x: retention, excluding same-day reviews and first reviews
        return 0.623 * np.power(4 * x * (1 - x), 0.738)
```

```
    ## For RMSE
    def func2(a, b, c, d, e, f, g, x, y):
        # x: retention, excluding same-day reviews and first reviews
        # y: n(reviews)/1000, excluding same-day reviews and first reviews
        return 0.0135 / (np.power(x, 0.504) - 1.14) + 0.176 / (np.power(y, 0.825) + 2.22) + 0.101
```

Here's what they look like:
Figure_6

For RMSE it's a 2D plane
Figure_3
Figure_4
Figure_5


Then we take the real value of log loss/RMSE and divide it by the corresponding predicted value to obtain normalized log loss/RMSE. Here's what their distributions look like:
Figure_2
Figure_2


The medians of normalized values are very close to 1, which is exactly what I was hoping for. This means that predicted and real values match well.

Interpretation: if the normalized value is <=1, it's good. Otherwise it's bad. If it's <1, that means the actual metric is lower than what you would expect for a user in this "weight class" (retention and n(reviews)).

Finally, here's the whole process:
1) Calculate retention and n(reviews)/1000. **Both should exclude first reviews and same-day reviews**.
2) Use them to calculate predicted log loss and RMSE.
4) Run the 5-way split to obtain real log loss and RMSE values.
5) Divide real log loss and RMSE by predicted values to obtain normalized log loss and RMSE.
6) If norm. log loss >= **1.11** and norm. RMSE >= **1.53**, display the warning with recommendations, otherwise display a simple "FSRS can adapt to your memory well" or whatever message. I have tuned this in such a way that ~5% of users will get a warning. In other words, for 5% of users norm. log loss >= 1.11 and norm. RMSE >= 1.53.

Pseudocode:
```
predicted_log_loss = func1(retention)  # no first reviews and no same-day reviews
predicted_log_loss = predicted_log_loss.clip(min=0.0001)  # to avoid division by 0, just in case
predicted_rmse = func2(retention, n_reviews/1000)  # no first reviews and no same-day reviews 
observed_log_loss, observed_rmse = fsrs.5_way_split().results()  # or something, idk
norm_log_loss = observed_log_loss / predicted_log_loss
norm_rmse = observed_rmse / predicted_rmse
if norm_log_loss >= 1.11 and norm_rmse >= 1.53:
    message = warning_string
else:
    message = all_good_string
```

You (and other people) are likely wondering, "Why use two metrics and not just one?". It turns out that 2-4% of users are weird: their norm. log loss is <1, but their norm. RMSE is significantly greater than 1, or the other way around. That means that it's ambiguous: if one metric is good and the other one is bad, what should we do? Using two thresholds for two metrics instead of one solves the ambiguity issue. If both norm. log loss and norm. RMSE are >1, then it's *definitely* bad.

EDIT: I changed the first function
EDIT 2: updated the coefficients in approximation functions and the thresholds

user1823: Expertium's formula for predicted log loss can also be written as:

```py
    ## For Log Loss
    def func1(x):
        # x: retention, excluding same-day reviews and first reviews
        return 0.613 - 0.591 * np.power(x, 5.85)
```

This is easier to read and probably faster to calculate as it avoids floating point division.

Edit: Expertium changed the formula. So, see his edited comment above.

Expertium: Actually, wait. Theoretically, log loss should be 0 or close to 0 at retention = 0% and retention = 100%, with a maximum at 50%. So it must be like a like x*(1-x). Hold on, I'll make a better function that is better suited for users with <50% retention

Expertium: Much better. I'll update the comment above
Figure_6

Expertium: Alex wants to do more research into metrics and thresholds, and I need to run more benchmarks, sooo...time to wait for 100 more hours 😅

Expertium: @Luc-Mcgrady Update: while Alex has some concerns that these normalized metrics may not reflect FSRS's performance, we couldn't think of anything better, so let's go with what I described [above](https://github.com/ankitects/anki/pull/4047#discussion_r2126361203).

Luc-Mcgrady: @Expertium Is my implementation correct? https://github.com/ankitects/anki/pull/4047/commits/e642eb535efa661118023a0c1db70f385ab987f3

(ignore the `resp.healthCheckPassed && !resp.healthCheckPassed;`)

ftl/core/deck-config.ftl
```diff
@@ -485,6 +486,15 @@ deck-config-percent-input = { $pct }%
...
 deck-config-fsrs-must-be-enabled = FSRS must be enabled first.
 deck-config-fsrs-params-optimal = The FSRS parameters currently appear to be optimal.
+deck-config-fsrs-bad-fit-warning = Your memory is difficult for FSRS to predict. Recommendations:
+
+    - Suspend or reformulate leeches.
+    - Use the answer buttons consistently. Keep in mind that "Hard" is a passing grade, not a failing grade.
+    - Understand before you memorize.
+
+    After you follow these suggestions, it may take weeks or months before parameters become optimal.
+deck-config-fsrs-good-fit = FSRS is well adjusted to your memory.
```
Luc-Mcgrady, line 496: `deck-config-fsrs-good-fit ` is a placeholder. Suggestions welcome.

ts/routes/deck-options/FsrsOptions.svelte
Expertium: @Luc-Mcgrady I've updated the coefficients:
```
    def func(a, b, c, x):
        # x: retention, excluding same-day reviews and first reviews
        return 0.623 * np.power(4 * x * (1 - x), 0.738)
  
    def func2(a, b, c, d, e, f, g, x, y):
        # x: retention, excluding same-day reviews and first reviews
        # y: n(reviews)/1000, excluding same-day reviews and first reviews
        return 0.0135 / (np.power(x, 0.504) - 1.14) + 0.176 / (np.power(y, 0.825) + 2.22) + 0.101
```
And thresholds are 1.11 and 1.53.
Other than that, yes, seems good.
EDIT: there is a problem with your implementation, I explained it on Discord

Luc-Mcgrady: https://github.com/ankitects/anki/pull/4047/commits/eb38c7014bb639f9dddec221ee35359b347c27b5

ftl/core/deck-config.ftl
Expertium: "FSRS can adapt to your memory well."
It better conveys that the health check isn't about the current parameters

GithubAnon0000: Just to double check.

The health checks purpose is to find out whether FSRS predicts the users performance accurately? And if it doesn't, we assume the user has to change their behavior (like answer buttons, leeches, ect), whereas if FSRS does predict the performance accurately, then we want to tell the user that they're doing everything right?

Or is this actually about a theoretical adaptation to the users review habits were the user cannot drive any conclusions from it?

My line of thought here is this (pseudo code; just from the logic, I'm not proposing the actual strings):
```sh
if [ predicted_performance != actual_observed_performance ]; then
    show_message("User, you're doing something wrong. Try these things to improve your results: [list of things]");
else if [ predicted_performance == actual_observed_performance ]; then
    show_message("User, you're doing great. FSRS can predict your memory decay well.")
```

If I'm correct, then I'd propose that the "positive" string includes some mention of "user, you're doing great" (again, not proposing that actual string).

ftl/core/deck-config.ftl
Expertium: > Or maybe we should just launch straight into a suggestion?

I feel like we do need a sentence that says that FSRS isn't adapting very well, to make things clear. The "softest" that I can think of is "Your memory is somewhat unpredictable".

> Or we could just overestimate and say "over the next few months" to make it simpler to read, and less likely to worry users if it's not happening quickly?

Sure 👍

ftl/core/deck-config.ftl
Expertium: > And if it doesn't, we assume the user has to change their behavior (like answer buttons, leeches, ect), whereas if FSRS does predict the performance accurately, then we want to tell the user that they're doing everything right?

The thing is - we can't pinpoint exactly why FSRS is having trouble adapting to some user, so the best we can do is list all possible causes we can think of (leeches, inconsistent button usage, maybe something else?) and tell the user to work on that

ftl/core/deck-config.ftl
```diff
@@ -395,6 +395,7 @@ deck-config-weights = FSRS parameters
 deck-config-compute-optimal-weights = Optimize FSRS parameters
 deck-config-compute-minimum-recommended-retention = Minimum recommended retention
 deck-config-optimize-button = Optimize Current Preset
+deck-config-health-check = Perform a health check after optimization (slow)
```
dae: I feel this is a bit long. I'm going to tweak it directly so this PR doesn't drag out, but am open to further tweaks in follow-up PRs if anyone disagrees.

dae:
```suggestion
deck-config-health-check = Check health when optimizing (slow)
```

ftl/core/deck-config.ftl
```diff
@@ -485,6 +486,15 @@ deck-config-percent-input = { $pct }%
 deck-config-optimizing-preset = Optimizing preset { $current_count }/{ $total_count }...
 deck-config-fsrs-must-be-enabled = FSRS must be enabled first.
 deck-config-fsrs-params-optimal = The FSRS parameters currently appear to be optimal.
+deck-config-fsrs-bad-fit-warning = Your memory is difficult for FSRS to predict. Recommendations:
+
+    - Suspend or reformulate leeches.
+    - Use the answer buttons consistently. Keep in mind that "Hard" is a passing grade, not a failing grade.
+    - Understand before you memorize.
+
+    After you follow these suggestions, it may take weeks or months before parameters become optimal.
```
dae:
```suggestion
    If you follow these suggestions, performance will usually improve over the next few months.
```

PR #4048, opened 2025-06-02
Add IO mask colour fill tool

iamllama:
image

Forum post: https://forums.ankiweb.net/t/allow-having-multiple-io-mask-colours/61137

This pr proposes to add a new fill tool to the io editor, which would bring it closer to the IOE addon feature-wise

### DIFF
--- a/ts/routes/image-occlusion/tools/tool-buttons.ts
+++ b/ts/routes/image-occlusion/tools/tool-buttons.ts
@@ -6,6 +6,7 @@ import * as tr from "@generated/ftl";
 import {
     mdiCursorDefaultOutline,
     mdiEllipseOutline,
+    mdiFormatColorFill,
     mdiRectangleOutline,
     mdiTextBox,
     mdiVectorPolygonVariant,
@@ -14,6 +15,7 @@ import {
 import {
     cursorKeyCombination,
     ellipseKeyCombination,
+    fillKeyCombination,
     polygonKeyCombination,
     rectangleKeyCombination,
     textKeyCombination,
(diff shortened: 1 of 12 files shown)


ts/routes/image-occlusion/tools/tool-buttons.ts
```diff
@@ -50,4 +52,13 @@ export const tools = [
...
+    {
+        id: "fill-mask",
+        icon: mdiFormatColorFill,
+        iconSizeMult: 1.4,
+        tooltip: tr.editingImageOcclusionFillTool,
+        shortcut: fillKeyCombination,
+    },
+] as const;
+
+export type ActiveTool = typeof tools[number]["id"];
```
dae, line 64: TIL this is possible in TypeScript! Thank you for the typing improvements too. :-)

iamllama: > One minor suggestion: if fill is set to #000000, we could clear the explicit value.

Not sure if i understood correctly 🤔 The fill attr isn't saved in the cloze if its set to the default mask colour (#ffeba2), like the `angle` attr when its 0. If its changed to be cleared when set to #000000 then users' existing io notes using the default mask colour would have their notes bloated with it

dae: Sorry, that's a thinko on my part - from https://github.com/ankitects/anki/pull/4048/files#diff-a1adc784721911f5063bdb67703c3f2892d9fb58d5ee1738e0df305f7062f5caR361 I assumed #000 was the default. If our default is a color the user is unlikely to be able to pick easily, no point trying to remove it if it's the default.

iamllama: Ah, sorry, was mistaken as well. The text "clozes" behave differently, they don't save the fill attr when its the default text colour (#000000) and regular masks don't save it when its the default mask colour

iamllama: > If our default is a color the user is unlikely to be able to pick easily

The [`datalist`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/datalist#color_type) element should be able to provide preset options for the colour input, and seems to work in external/mobile browsers but not in anki itself, even though its chromium version should support it

dae: I'm guessing Qt overlooked something when trying to integrate it into their own color picker implementation.

Anyway, thank you as always for your contributions :-)

PR #4049, opened 2025-06-02
FIX diacritics being cutoff in input card templates

GithubAnon0000:
# Changes
This adds a bit of line-height to make the diacritics visible. It doesn't look out of place for latin letters in my opinion, even though the input field becomes slighter heigher due to the change in this PR.

# Rationale
While this can be done via templates as well, I think anki should support other languages without any additional, special setup.

# Tests
I asked a AI service to provide me a Dzongkha word and it provided _བཀྲ་ཤིས་_. I tested with this string and with a few random latin strings.

# Pictures
All these pictures are with the change from this PR.

_Dzongkha and latin next to each other:_
anki

_Only Dzongkha:_
anki

_Only Latin:_
anki

# Source
https://forums.ankiweb.net/t/type-to-answer-box-not-high-enough-to-display-diacritics/62073?u=anon_0000

### DIFF
--- a/ts/reviewer/reviewer.scss
+++ b/ts/reviewer/reviewer.scss
@@ -71,6 +71,7 @@ pre {
     width: 100%;
     // https://anki.tenderapp.com/discussions/beta-testing/1854-using-margin-auto-causes-horizontal-scrollbar-on-typesomething
     box-sizing: border-box;
+    line-height: 1.75;
 }
 
 code#typeans {


brishtibheja:
Try this too:`སྐྱེིད་`

FYI Indian and Tibetan scripts (same origin) stack consonants over each other and use diacritics for vowels so that's how it becomes larger.

dae:
@brishtibheja's example works for me. Thank you both!

PR #2935, opened 2024-01-09
Americanize "synchronisation"

hnvy:
Hello,

An extremely small PR as per https://github.com/ankitects/anki-manual/pull/192

Edit 1: I keep seeing "All checks have failed", I'm assuming that this is because of the `CONTRIBUTORS`?

Edit 2: Ah, I [learned something new today](https://buildkite.com/ankitects/anki-ci/builds). I had to  do [this](https://github.com/ankitects/anki/pull/2935/commits/3be9624f3234953019087ffdd8528d45c58c9e55). It should pass now.

PR #4050, opened 2025-06-02
Optimize All -> Optimize All Presets

Expertium:
The most minor PR ever


Expertium:
@Danika-Dakika is wondering if it's possible to dismiss this message. It doesn't seem like it's possible, but I agree that it would be a good addition. Might be worth making an issue for that

user1823:
Fun fact: https://github.com/ankitects/anki/pull/2935 was probably the PR having the smallest change possible.

dae:
It's not being dismissed when using the Optimize All Presets button?

Expertium:
> It's not being dismissed when using the Optimize All Presets button?

I believe Danika wants something like a cross icon to dismiss it without optimization. Btw: https://github.com/ankitects/anki/issues/4051

dae:
@Danika-Dakika could you elaborate? You want to be able to hide the message without clicking the suggested button?

PR #4052, opened 2025-06-02
Fix/Invalid parameter counts saveable

Luc-Mcgrady:
Before, if you set your parameters to a value which is not 21 numbers long, it would fallback to the previous versions parameter counts. These counts could be valid (blank for default for example) and allow the values to be saved.

I fixed this in svelte before I noticed the real bug, If you don't want that revert 75ccf58765d9b9cc7554fb69696155698ff054f4.


user1823 (approved):
Both changes LGTM and we need both, IMO.

dae:
Thank you both!

Issue #4053, opened 2025-06-03
Fix audio play functionality in Cross Profile Search and Import and AJT Japanese


tatsumoto-ren:
#4041 introduced a critical bug that prevents [Cross Profile Search and Import](https://ankiweb.net/shared/info/1772763629) from playing audio files stored in other profiles. 

## Hide the 25.02.6 release before people get a chance to download it! 

Release 25.02.7 after the issue is fixed.

GithubAnon0000:
> [#4041](https://github.com/ankitects/anki/pull/4041) introduced a critical bug

Is it really a bug though? Seems very much intended.


> Hide the 25.02.6 release before people get a chance to download it!

That's a bit too much to ask for and seems to be an unreasonable request in my opinion.

tatsumoto-ren:
>Is it really a bug though? Seems very much intended.

This bug was reported by users via our bug report channels. Of course, I don't want more people to run into this problem.

tatsumoto-ren:
Another issue was just discovered with patch #4041. It also breaks the audio search functionality in [AJT Japanese](https://ankiweb.net/shared/info/1344485230) when the found audio files are stored in offline audio sources on disk.

The buttons don't work anymore:

![Image](https://github.com/user-attachments/assets/98df1ba7-0027-4a4f-8e63-9532490dafe5)

PR #4054, opened 2025-06-04
Restore ability for add-ons to provide full paths to media

tatsumoto-ren:
This is a quick fix for #4053.

### DIFF
--- a/pylib/anki/sound.py
+++ b/pylib/anki/sound.py
@@ -51,7 +51,8 @@ class SoundOrVideoTag:
         "Prepend the media folder to the filename."
         if os.path.basename(self.filename) == self.filename:
             # Path in the current collection's media folder.
-            # Ensure filename doesn't reference parent folder.
+            # Turn it into a fully-qualified path so mpv can find it, and to
+            # ensure the filename doesn't get treated like a non-file scheme.
             head, tail = media_folder, self.filename
         else:
             # Add-ons can use absolute paths to play arbitrary files on disk.


dae:
Your change undoes the security fix, as mpv doesn't care if a file exists when deciding how to parse the filename.

While it would be possible to solve the original exploit by passing in a full or explicitly-relative path instead of just a filename, references outside the media folder are not something we've ever explicitly tried to support, nor want to encourage wide-spread use of, as it means decks are not self-contained and cannot be easily shared.

While I'm sorry this has caused disruption for your add-ons (this was certainly not intended), rather than making changes for everyone, I feel like your add-ons should be the ones relaxing this restriction rather than Anki itself, so that only the add-on users have that extra exposure/functionality. Presumably you would be able to work around the change with an update to your add-ons? If you need a new hook to do it, you could start with a monkey-patch until an update with the new hook is released.

A heads up: due to the severity of this issue, an update notification is planned for existing users tomorrow.

dae:
> Your change undoes the security fix

Sorry, I was wrong. As you've left the basename call in, and Windows does not allow colon in filenames, it would not cause a regression, nor allow fully-qualified paths. Let me think about it a bit more.

qt/aqt/sound.py
```diff
@@ -41,6 +42,15 @@
     tr,
 )
 
+# Utils
+##########################################################################
+
+
+def resolve_tag_path(tag: SoundOrVideoTag, media_folder: str) -> str:
+    file_path = tag.filename if os.path.isfile(tag.filename) else tag.path(media_folder)
+    return hooks.media_file_filter(file_path)
```
dae: - The media folder is not the CWD, so the isfile check looks wrong?
- Would it make sense to move this logic into .path() to avoid an extra helper function?

tatsumoto-ren: Add-ons create SoundOrVideoTag objects using full paths to files, enabling them to play arbitrary files on disk. For example, AJT Japanese plays files from external audio sources (i.e., a folder on a hard drive), and AJT CroPro plays files stored in currently closed Anki collections. I'd like to allow these full paths.

<details>
Example taken from AJT Japanese:

``` python
def _play_audio_file(self, file: FileUrlData) -> None:
    if os.path.isfile(file.url):
        return sound.av_player.play_tags([SoundOrVideoTag(filename=file.url)])
    elif mw.col.media.have(file.desired_filename):
        return sound.av_player.play_tags([SoundOrVideoTag(filename=file.desired_filename)])
```
</details>

Good idea. I refactored the code to use only the .path() method.

tatsumoto-ren:
> I feel like your add-ons should be the ones relaxing this restriction rather than Anki itself, so that only the add-on users have that extra exposure/functionality.

I don't want any add-ons that remove security fixes. And since many add-ons are affected, I'm not sure how they could all undo the same security fix without causing conflicts or duplicating code. A hard way would be to reimplement the media player functionality in each add-on, which would take a lot of time and seems like an even worse approach.

qt/aqt/sound.py
dae: If I've followed the code properly, this would allow things like 'anki.exe' to be accepted, since they'd be a valid file from the current working directory. And also allow fully-qualified paths inside fields, which I was trying to avoid. Are you able to do this in such a way that the fully qualified paths are allowed if the add-on has injected them, but not allowed inside fields? An earlier basename call for example.

dae:
To clarify, I was suggesting the add-on enable the "files outside media folder" functionality you want, but in a way that doesn't reintroduce the security flaw with paths starting with a scheme, such as making all paths absolute. But if you can make the tweaks I've requested above, I will put out another patch release instead.

qt/aqt/sound.py
tatsumoto-ren: > this would allow things like 'anki.exe' to be accepted, since they'd be a valid file from the current working directory. 

`[sound:anki.exe]` would be allowed, but `mpv` won't execute it, so I don't think it's a problem.

`[sound:/full/path/to/file]` has been allowed and worked before, and that's why add-ons use it. Since using them wasn't explicitly discouraged before, it's probably too strict to ban them right away.

> Are you able to do this in such a way that the fully qualified paths are allowed if the add-on has injected them, but not allowed inside fields? 

Since user-facing tags (e.g. `[sound:filename.ogg]`) and `SoundOrVideoTag()` objects are very different, maybe a different patch would deal with filtering inside user-facing tags. For example, we could trim paths down to just basenames before SoundOrVideoTag objects are created.

dae: > [sound:anki.exe] would be allowed, but mpv won't execute it, so I don't think it's a problem.

One of the motivations for not supporting full paths is reducing attack surface. mpv is a big, complicated program, and we've already been bitten multiple times. There may be future issues passing in certain paths like UNC strings, particular devices in /dev, etc that we haven't discovered yet, and I'd rather play it safe, even if it inconveniences the small subset of users who have used full paths in their fields.

I can't think of a way files in the default CWD could be exploited, but it feels a bit unclean, as we only avoid the exploit on Windows due to the fact that : and / are not allowed in filenames. Can we fix the add-on use case without depending on the CWD? e.g. after your other PR is merged and we know the path will be a basename if no add-ons are involved, what if you gated your add-on path on if the filename was already absolute, instead of whether it exists relative to the cwd?

tatsumoto-ren: One way would be to check if `os.path.abspath(filename) == filename` but I'm not sure if python can modify a path that is already absolute, using a different representation, making the check fail where it should pass.

dae: why not use isabs()?

dae: Yeah ok, reading its docs, that looks like it'd include raw/UNC paths which is not desirable. Hmm.

dae: Can't you just check if filename == basename(filename)?

tatsumoto-ren: > Can't you just check if filename == basename(filename)?

sounds good

dae:
Thank you for working with me. I will get another update out either tonight or tomorrow.

tatsumoto-ren:
Thank you for merging! This will help a lot of people!

PR #4056, opened 2025-06-04
improve the accuracy of the expected workload calculation

L-M-Sherlock:
**Key Changes:**

*   **Enhanced Workload Calculation:**
    *   The `fsrs::expected_workload` function now incorporates an `initial_pass_rate` (the pass rate of the first review for a card) and a `TERMINATION_PROB` for more precise estimations.
    *   The `LEARN_SPAN` constant, used in the workload calculation, has been significantly increased from `1000` to `100_000_000`. This change makes the workload curve smoother than before.
    *   The method for retrieving review costs (`get_costs_for_retention`) has been updated to fetch the `initial_pass_rate` from the database. This involved modifying the SQL query to calculate this new metric.

*   **Dependency Updates:**
    *   Updated `fsrs` to git revision `a7f7efc10f0a26b14ee348cc7402155685f2a24f`.
    *   Upgraded `priority-queue` from version `2.3.1` to `2.5.0`.
    *   Upgraded `snafu` and `snafu-derive` from version `0.8.5` to `0.8.6`.

**Detailed Changes by File:**

*   **`Cargo.lock` & `Cargo.toml`**:
    *   Reflect the version updates for `fsrs`, `priority-queue`, `snafu`, and `snafu-derive`.

*   **`rslib/src/deckconfig/service.rs`**:
    *   In `GetRetentionWorkloadRequest` handling:
        *   Changed `LEARN_SPAN` from `1000` to `100_000_000`.
        *   Introduced a new constant `TERMINATION_PROB` set to `0.01`.
        *   Modified the calls to `fsrs::expected_workload` to pass `learn_cost`, the newly fetched `initial_pass_rate`, and `TERMINATION_PROB` as arguments.

*   **`rslib/src/storage/card/get_costs_for_retention.sql`**:
    *   Added a Common Table Expression (CTE) named `initial_pass_rate`. This CTE calculates the average pass rate of the first review (where `ease > 1`) for cards included in `search_cids` by using `RANK() OVER (PARTITION BY cid ORDER BY id ASC)`.
    *   The final `SELECT` statement now includes this `initial_pass_rate`.

*   **`rslib/src/storage/card/mod.rs`**:
    *   Updated the function signature of `get_costs_for_retention` to return `Result<(f32, f32, f32, f32)>` to include the `initial_pass_rate`.
    *   Adjusted the implementation to retrieve and return this fourth value, defaulting to `0.5` if the database query doesn't return it.

### DIFF
--- a/rslib/src/storage/card/mod.rs
+++ b/rslib/src/storage/card/mod.rs
@@ -747,7 +747,7 @@ impl super::SqliteStorage {
             .get(0)?)
     }
 
-    pub(crate) fn get_costs_for_retention(&self) -> Result<(f32, f32, f32)> {
+    pub(crate) fn get_costs_for_retention(&self) -> Result<(f32, f32, f32, f32)> {
         let mut statement = self
             .db
             .prepare(include_str!("get_costs_for_retention.sql"))?;
@@ -758,6 +758,7 @@ impl super::SqliteStorage {
             row.get(0).unwrap_or(7000.),
             row.get(1).unwrap_or(23_000.),
             row.get(2).unwrap_or(30_000.),
+            row.get(3).unwrap_or(0.5),
         ))
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2295,7 +2295,7 @@ dependencies = [
 [[package]]
 name = "fsrs"
 version = "4.0.0"
-source = "git+https://github.com/open-spaced-repetition/fsrs-rs.git?rev=33ec3ee4d5d73e704633469cf5bf1a42e620a524#33ec3ee4d5d73e704633469cf5bf1a42e620a524"
+source = "git+https://github.com/open-spaced-repetition/fsrs-rs.git?rev=a7f7efc10f0a26b14ee348cc7402155685f2a24f#a7f7efc10f0a26b14ee348cc7402155685f2a24f"
 dependencies = [
  "burn",
  "itertools 0.14.0",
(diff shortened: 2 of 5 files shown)


Luc-Mcgrady (approved):
LGTM!

Expertium:
Does it use smoothing? I mean where we take the default cost/probability and the user-specific one, and calculate a weighted average with n(reviews) as weight? It definitely should, to ensure a smooth transition from default values to user-specific values for new users
Final cost = a * user cost + (1-a) * default cost,
a = reviews / (50 + reviews)

Expertium:
Does it use smoothing? I mean where we take the default cost/probability and the user-specific one, and calculate a weighted average with n(reviews) as weight? It definitely should, to ensure a smooth transition from default values to user-specific values for new users
Final cost = a * user cost + (1-a) * default cost,
a = reviews / (50 + reviews)

rslib/src/storage/card/mod.rs
```diff
@@ -747,7 +747,7 @@ impl super::SqliteStorage {
             .get(0)?)
     }
 
-    pub(crate) fn get_costs_for_retention(&self) -> Result<(f32, f32, f32)> {
+    pub(crate) fn get_costs_for_retention(&self) -> Result<(f32, f32, f32, f32)> {
```
dae: For future maintenance, the return value would be better broken into a separate struct.

dae:
No objections on my end save the one above. Let me know when any debate is complete and this is ready to merge.

Expertium (approved):

Expertium:
Actually @L-M-Sherlock, before we merge it, let's tune termination_prob. Try 3-4 different values and report how long it takes to run this on 30 values of desired retention for each value of termination_prob, please.
Something like this:
termination_prob = 0.01, time=0.01 s
termination_prob = 0.001, time=0.02 s
termination_prob = 0.0001, time=0.05 s

Expertium:
Jarrett DMed me
0.01: time:   [986.41 µs 1.0122 ms 1.0395 ms]
0.001: time:   [12.136 ms 12.880 ms 14.123 ms]
0.0001: time:   [112.33 ms 114.85 ms 117.55 ms]

This is for 30 values of DR
Let's go with 0.001 then

PR #4058, opened 2025-06-04
Fix absolute sound tags

abdnh:
Fix regression in https://github.com/ankitects/anki/pull/4041 that broke some TTS players and the 'Record Own Voice' feature.

### DIFF
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -25,6 +25,7 @@ from anki.scheduler.v3 import (
     SchedulingStates,
     SetSchedulingStatesRequest,
 )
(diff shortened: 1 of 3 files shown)


dae:
The current implementation feels a bit messy at the moment, as there's no clear distinction between internal calls that don't need validation, and calls that are provided user input. What do you think about the following instead? https://github.com/ankitects/anki/pull/4059

@tatsumoto-ren your input would be appreciated too

tatsumoto-ren:
#4059 Looks good to me. At least I don't see anything that would break existing code.

Additionally, I suggest trimming down paths to basenames inside [sound:xxx] tags. This could be done when a new deck is imported.

PR #4059, opened 2025-06-05
Treat play_file() and co as internal routines without protection

dae:
Our code and add-ons need a way to play audio from arbitrary locations. I propose we treat the _tag API as suitable for user input, and the _file API for internal use.

Closes #4058

### DIFF
--- a/qt/aqt/sound.py
+++ b/qt/aqt/sound.py
@@ -179,15 +179,15 @@ class AVPlayer:
     def play_file(self, filename: str) -> None:
         """Play the provided path.
 
-        SECURITY: Filename may be an arbitrary path. For content coming from a collection,
-        use play_tag() instead."""
+        SECURITY: Filename may be an arbitrary path. For filenames coming from a collection,
+        you should only ever use the os.path.basename(filename) as the filename."""
         self.play_tags([SoundOrVideoTag(filename=filename)])
 
     def play_file_with_caller(self, filename: str, caller: Any) -> None:
         """Play the provided path, noting down the caller.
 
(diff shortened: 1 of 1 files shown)


qt/aqt/sound.py
```diff
@@ -177,16 +177,28 @@ def clear_queue_and_maybe_interrupt(self) -> None:
             self._stop_if_playing()
 
     def play_file(self, filename: str) -> None:
-        self.play_tags([SoundOrVideoTag(filename=os.path.basename(filename))])
+        """Play the provided path.
+
+        SECURITY: Filename may be an arbitrary path. For content coming from a collection,
+        use play_tag() instead."""
```
abdnh: To me, this implies play_tags() does some validation and enforces a media path, while it's actually exactly the same under the hood.

abdnh:
This looks like a better approach for add-ons. If we want to leave validation to add-ons dealing with user input, a note about os.path.basename() in the docstrings would be nice.

dae:
If my change wasn't what you intended, please feel free to push a follow-up :-)

luc-vocab:
This was possibly the cause of https://github.com/Vocab-Apps/anki-hyper-tts/issues/282 as well ?

abdnh:
I see some av_player.play_file/insert_file calls in the add-on's code, so that's probably it if HyperTTS is playing audio files not in the media directory.

luc-vocab:
@abdnh can you elaborate on this ?

dae:
It was a bug in .6. It should be fixed in .7 already, so please test with that / ask the user to.

PR #4060, opened 2025-06-05
Update default styles

brishtibheja:
Added `line-height` to styling and margin to `hr`. 

Forums: https://forums.ankiweb.net/t/modernize-default-card-templates-for-readability-revised-proposal-and-technical-details/54669/26

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -148,7 +148,7 @@ user1823 <92206575+user1823@users.noreply.github.com>
 Gustaf Carefall <https://github.com/Gustaf-C>
 virinci <github.com/virinci>
 snowtimeglass <snowtimeglass@gmail.com>
-brishtibheja <sorata225yume@gmail.com>
+brishtibheja <136738526+brishtibheja@users.noreply.github.com>
 Ben Olson <github.com/grepgrok>
 Akash Reddy <akashreddy2003@gmail.com>
 Lucio Sauer <watermanpaint@posteo.net>
--- a/rslib/src/notetype/styling.css
+++ b/rslib/src/notetype/styling.css
@@ -1,6 +1,7 @@
 .card {
     font-family: arial;
     font-size: 20px;
+    line-height: 1.5;
     text-align: center;
     color: black;
     background-color: white;
--- a/ts/reviewer/reviewer.scss
+++ b/ts/reviewer/reviewer.scss
@@ -6,6 +6,7 @@
 
 hr {
     background-color: vars.palette(darkgray, 0);
+    margin: 1.5em 0;
 }
 
 body {
--- a/ts/reviewer/reviewer.scss
+++ b/ts/reviewer/reviewer.scss
@@ -6,7 +6,7 @@
 
 hr {
     background-color: vars.palette(darkgray, 0);
-    margin: 1.5em 0;
+    margin: 1em 0;
 }
 
 body {


brishtibheja:
edit: privated my email and edited contribs file. it worked. thanks to llama for the help.

uh... what?

```
Author sorata225yume, at the domain gmail.com NOT found in list
```

but here:


https://github.com/brishtibheja/anki/blob/9e65ed7ea51cd04daaa1e460b5d60bcf1c80af25/CONTRIBUTORS#L151

GithubAnon0000:
I don't really have an issue with these changes, since I know how to customize the templates anyways. Just asking though, so that it had been mentioned at least once:

* Is `line-height: 1.5` not a bit excessive? To me it seems like way to much (I use 1.25 in my templates which I think is reasonable).
* Same with `margin: 1.5em 0;`. It seems way to much (I use 1rem which seems reasonable).


Was there consensus on these specific values (maybe I'm one of the few that thinks it's too much spacing)?

brishtibheja:
I don't think anyone objected... actually, weren't you in that thread too?
I didn't see you object either. As for me, I think the values are fine.

GithubAnon0000:
# Objections raised
Now that I look at it again (the first half of the topic), someone did object: https://forums.ankiweb.net/t/modernize-default-card-templates-for-readability-revised-proposal-and-technical-details/54669/2?u=anon_0000. And three people upvoted it (4 with me), which I assume means "I agree".

I was sure I objected but apparently I did that in a previous thread. I remember that because I also claimed something about the font size, which was a false claim though because I misremembered (I had changed the value and thought it was the default); embarrassing moments stick, I guess.

# Let's make a Poll
Maybe we can make a poll? It would be a simple poll, just showing the card as it is now and the card as it will be with this PR merged in, then asking people which of those two options they prefer.

The poll should feature very short text (like those of some language learners) as well as longer text (like those that study with actual sentences – I do as an example).

# Side note
For reference, my spacing is much less to what you propose here and I'd argue it's actually plenty of spacing, see the pictures in my topic in the forums: https://forums.ankiweb.net/t/how-can-i-make-list-elements-li-more-readable-while-still-being-subtle/60144/14?u=anon_0000.

Obviously my template is heavily modified though, which is why I'd prefer a fair comparison as outlined in **Let's make a Poll** above.

dae:
Happy to wave through move conservative values, but given @GithubAnon0000 has objected here, it might be worth a poll if you'd like to keep the values as-is.

brishtibheja:
@GithubAnon0000 changed the margins for `hr`, the other one I feel somewhat strongly about. If you would like to have a different value (`1.25` is what you mentioned) can you make the poll yourself?

It'd be nice if you also explain to people that line height won't affect existing users.

GithubAnon0000:
I'll need a few days in that case. And I'll only be able to start the poll in the forums, as I do not use discord or reddit.

brishtibheja:
@dae it seems like 1.5 is a commonly accepted minimum (check W3) for A11Y.
having that in mind, now it seems only proper to me that we set line height
to at least that value. wdyt?

GithubAnon0000:
That's not true. First of all, here's the w3 link you're talking about: https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html

Second: It states the following:
> This SC does not dictate that authors must set all their content to the specified metrics, or provide a mechanism to do so. Rather, it specifies that an author's content has the ability to be set to those metrics without loss of content or functionality. 


That means the values must be modifyable and the cards must still work (e.g. everything must be visible, clickable, ect.). Anki already provides a way to change those values, so that's not an issue. And Anki doesn't have to use those values by default – rather, users should be able to use them if they need it (I assume the browser has accessibility features that can do that automatically or there is special software that users could use).

Ankis card templates are so simple by default, that there's basically no way that the cards break if the user were to set those values to the SC.

And for clarity, here are the success criterions (SC) copied from the w3 page:
> In content implemented using markup languages that support the following text style properties, no loss of content or functionality occurs by setting all of the following and by changing no other style property:
> 
>   *  Line height (line spacing) to at least 1.5 times the font size;
>   *  Spacing following paragraphs to at least 2 times the font size;
>   *  Letter spacing (tracking) to at least 0.12 times the font size;
>   *  Word spacing to at least 0.16 times the font size.
> 
> Exception: Human languages and scripts that do not make use of one or more of these text style properties in written text can conform using only the properties that exist for that combination of language and script.

GithubAnon0000:
# Images
1) Here's a demonstration with the values w3 has mentioned in their sc:
anki


2) And this is the default:
anki



# Code
```css
*:not(#divider) {
	line-height: 1.5em;
	margin-bottom: 2em;
	letter-spacing: 0.12em;
	word-spacing: 0.16em;
}
```


<hr>

Allowing these values is the right thing to do. But making it the default would upset a lot of users.

brishtibheja:
A lot of other websites do seem to recommend this value and so my point was less about meeting the SC from W3C (although I did initially misunderstood it).

brishtibheja:
> Allowing these values is the right thing to do. But making it the default would upset a lot of users.

I'm only changing the line height. And besides the fact that no one objected to this particular value in forums, the styling would only affect new users. Existing users are more likely to complain about changes but they won't get to see it.

GithubAnon0000:
> And besides the fact that no one objected to this particular value in forums

They did, as I pointed out above. At least 5 users objected (one commenter and 4 people giving hearts).

**Edit:** Actually, the objection was not about `line-height`, only regarding the other margin change you proposed.

brishtibheja:
> They did, as I pointed out above. At least 5 users objected (one commenter and 4 people giving hearts).

Vaibhav was talking about the margins though (personally don't trust likes in forums too but disregard that here).

dae:
You both make reasonable points, and it's not a simple call. Since I'm forced to choose, I guess the best option is to merge this in, as being more accessible by default makes sense as long as it's not too visually jarring, and the user can adjust anyway.

GithubAnon0000:
I know this was half a year ago; but I edited my card templates recently to support inline code in my cards (e.g. to support questions about the `git` command). I needed a visual that highlights the inline code block so that it is separated from the other text. I also have other use cases like a visual separation of paths from text. Essentially, the goal was to make the cards more easily skimable and readable for cards that reference code / commands or paths.

To prevent clipping of those styles that I created, I had to increase line-height from 1.25 to 1.5; and I now use it for every card, including those that do not use code or paths. It was a bit weird at first, but it made the cards so much more readable; I now support the change wholeheartedly.

I intially thought that the change to 1.5 has not been pushed (but I now checked and see that it indeed had been pushed) and wanted to ask to consider pushing that change. But since that had been done already, I'll just comment this to voice my support for that change, even though 6 months had passed since then.

PR #4061, opened 2025-06-06
refactor: split responsibilities between MainWindow and ProfileManager

marvinkopf:
refactor: split responsibilities between MainWindow and ProfileManager

- Extracted MainWindow and ProfileManager into separate classes
- Introduced use case classes (e.g. RestoreBackupUseCase, CleanupAndExitUseCase) to decouple logic from UI
- Use cases are callable and support external callback injection for integration
- Added global singleton instances for app-wide access

This refactoring lays the groundwork for another PR that adds dynamic routing to MainWindow.
[See also](https://forums.ankiweb.net/t/adding-dynamic-routing-to-the-ankiqt-class/61959)

WIP but mostly done. Is this something that could be merged? Or are the changes too much?

### DIFF
--- a/qt/aqt/__init__.py
+++ b/qt/aqt/__init__.py
@@ -778,6 +778,10 @@ def _run(argv: list[str] | None = None, exec: bool = True) -> AnkiApp | None:
             )
             sys.exit(1)
 
+    from aqt.utils import cleanup_and_exit
+
+    cleanup_and_exit.app = app
+
     # load the main window
     import aqt.main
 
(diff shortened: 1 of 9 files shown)


dae:
Hi Marvin,

I'm afraid I'm not really keen on these changes. For some background: https://github.com/ankitects/anki/blob/main/docs/contributing.md#refactoring

Aside from the potential regressions, this is likely to break some add-ons. And given that it results in no new functionality nor moves us closer to the greater use of Svelte goal, I would rather we not make this change at this time.

dae:
(that is not to say that the PR itself is bad - better code separation and tests are a good thing! I just feel this is not the best time to be making such changes)

PR #4066, opened 2025-06-07
Delay retention workload info display after FSRS optimization alerts

Sunrongguo2008:
When clicking the "Optimize current preset" button in FSRS settings, "The higher this value, the more frequently cards will be shown to you."(using `.two-line`) would experience layout shifts during async operations. The element would initially render with single-line height before the completion alert, then expand to proper two-line height after user dismisses the alert dialog.
图片
图片
Guess the reason is:
1. High computational issue: Currently using `min-height: calc(2ch + 30px)`, where the `ch` unit is based on the width of the character "0". Layout needs to be recalculated after text updates
2. Rendering Timing Issue: The alert pop-up blocks the main thread, leading to: After state update (`optimized=true`), it cannot immediately re-render, and the browser can only execute the rendering queue after the user clicks "OK".
I Modified CSS styles for `.two-line` class in FSRS settings component.
I haven't tested it, so I don't know if my changes are effective.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -230,6 +230,7 @@ KolbyML <https://github.com/KolbyML>
 Adnane Taghi <dev@soleuniverse.me>
 Spiritual Father <https://github.com/spiritualfather>
 Emmanuel Ferdman <https://github.com/emmanuel-ferdman>
+Sunong2008 <https://github.com/Sunrongguo2008>
 Marvin Kopf <marvinkopf@outlook.com>
 ********************
 
--- a/ts/routes/deck-options/FsrsOptions.svelte
+++ b/ts/routes/deck-options/FsrsOptions.svelte
@@ -405,7 +405,8 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     :global(.two-line) {
         white-space: pre-wrap;
-        min-height: calc(2ch + 30px);
+        height: calc(2 * 1.5em + 10px);
+        line-height: 1.5em;
         box-sizing: content-box;
         display: flex;
         align-content: center;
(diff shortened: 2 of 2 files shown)


dae:
Unfortunately, this doesn't seem to fix the problem. Changing the setTimeout from 200 to 2000 does, but delay-based "fixes" are bad, as there will always be slower devices that still trigger the problem. I tried inserting an await tick(), but it didn't seem to help. @Luc-Mcgrady in case you're interested

Luc-Mcgrady:
It's likely caused by the sliding fade in effect of the warning box taking longer than 200ms. A quick google tells me the default is 400ms which would explain it half showing.

dae:
I hadn't thought of that! It does sound probable.

Sunrongguo2008:
It seems that's indeed the case.  From the video below, the warning box's sliding fade-in takes about 10 frames before the alert appears. At 48fps, that's roughly 200ms.

https://github.com/user-attachments/assets/3cc28111-4e34-48b3-9ac0-44948a04ea76

Sunrongguo2008:
I think we can change WORKLOAD_UPDATE_DELAY_MS from 100 to 250 to ensure the warning box appears after alert popups. This resolves timing conflicts where:

Before:
100ms → Warning box starts rendering
200ms → Alert popup appears (interrupting the warning box render)

After:
200ms → Alert popup appears first
250ms → Warning box renders without interrupting

Luc-Mcgrady:
It might be better to add a `(201 - WORKLOAD_UPDATE_DELAY_MS)` ms delay to `optimized = true` (It's not used for anything besides the info box and changing `WORKLOAD_UPDATE_DELAY_MS` would as you could probably tell slow down the rate at which the values update when being changed).

https://github.com/ankitects/anki/blob/c1fc45928dfa3fd12050c2bf254c2e3f33463830/ts/routes/deck-options/FsrsOptions.svelte#L210

Sunrongguo2008:
You are right. I will change it.

user1823:
Because the issue was in the timing, do we still need to change the height?

Sunrongguo2008:
Done

dae:
Thanks all!

PR #4069, opened 2025-06-08
Feat/Message at 100% progress for optimize

Luc-Mcgrady:
image

Note this will show regardless of if it is stalling at 100% due to the health check or if it is stalling at 100% due to the 2 evaluates which determine if the parameters have improved since last optimization.


Expertium:
image

Can you change this to "FSRS can adapt to your memory well"? t better conveys that the health check is NOT about the current parameters. Dae approved this change, btw.

dae:
That's better done in the original PR, which hasn't been merged yet. (@Expertium)

PR #4072, opened 2025-06-10
Fix copied image not imported locally on 403 error.

phu54321:
This PR fixes issues described [here](https://forums.ankiweb.net/t/pasting-some-image-from-chrome-wont-import-the-image-to-the-media-database/62457). Maybe a partial reimplementation of https://github.com/ankitects/anki/commit/abd8759d502bea5707661d17b06787e395794156

## Problem description

When you copy images on browser, with the image, a html snippet is also copied.

```html
<html>
<body>
<!--StartFragment--><img src="https://www.researchgate.net/publication/372533503/figure/fig1/AS:11431281270911835@1723283890520/Carotid-sinus-and-carotid-body-The-carotid-sinus-nerve-a-branch-of-the-glossopharyngeal.jpg"/><!--EndFragment-->
</body>
</html>
```

If user tries to paste this to Anki, since HTML handler kicks in ahead of the image handler. When Anki sees ``, it tries to download that image, but due to user-agent or other unknown issues it fails with 403 forbidden error. So URL pointing to remote resource persists.

image
image

## Bug reproduction step

1. With Chrome browser, go to [this image](https://www.researchgate.net/publication/372533503/figure/fig1/AS:11431281270911835@1723283890520/Carotid-sinus-and-carotid-body-The-carotid-sinus-nerve-a-branch-of-the-glossopharyngeal.jpg) and copy the image.
3. Paste the image to the Anki.
4. Notice `403` forbidden error and image referenced remotely.

image
image

## Suggested solution

When Anki detects a single image paste,
 - First, try to download the image. If it works, use that. (preserving attributes like `alt`). I tried to make this procedure as backward-compliant as possible.
 - If downloading fails, use the image in the clipboard. This drops other attributes like `alt` as I tried to use existing function `EditorWebView._processImage`. Since download failure rarely happens, I hope this won't make much of the trouble.

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -1587,6 +1587,47 @@ class EditorWebView(AnkiWebView):
 
         self.editor.doDrop(html, internal, extended, cursor_pos)
 
+    def _processSingleImageHtml(self, mime: QMimeData, extended):
+        if not mime.hasHtml():
+            return None
+
+        html_content = mime.html()
+
+        # Unwraps fragment data.
+        try:
+            html_content = html_content[
+                html_content.index("<!--StartFragment-->")
+                + 20 : html_content.rindex("<!--EndFragment-->")
+            ]
+            if not html_content:
+                # maybe malformed html?
+                return None
+        except ValueError:
+            return None
+
+        bs = BeautifulSoup(html_content, "html.parser")
+        tags = bs.find_all()
+        if len(tags) == 1 and tags[0].name.lower() == "img":
+            img_tag = tags[0]
+
+            try:
+                src = img_tag["src"]
+            except KeyError:
+                return None
+
+            if self.editor.isURL(src):
+                fname = self.editor._retrieveURL(src)
+                if fname:
+                    img_tag["src"] = fname
+                    return str(img_tag)
+
+                else:
+                    image_html = self._processImage(mime, extended)
+                    if image_html:
+                        return image_html
+
+        return None
+
     # returns (html, isInternal)
     def _processMime(
         self, mime: QMimeData, extended: bool = False, drop_event: bool = False
@@ -1602,6 +1643,10 @@ class EditorWebView(AnkiWebView):
         mime = gui_hooks.editor_will_process_mime(
             mime, self, internal, extended, drop_event
         )
+
+        image_html = self._processSingleImageHtml(mime, extended)
+        if image_html:
+            return image_html, True
 
         # try various content types in turn
         if mime.hasHtml():


iamllama:
The 403 error is still present, along with a new one on wsl:
```
  File "/anki/qt/aqt/editor.py", line 1553, in onPaste
    self._onPaste(QClipboard.Mode.Clipboard)
  File "/anki/qt/aqt/editor.py", line 1548, in _onPaste
    html, internal = self._processMime(mime, extended)
  File "/anki/qt/aqt/editor.py", line 1647, in _processMime
    image_html = self._processSingleImageHtml(mime, extended)
  File "/anki/qt/aqt/editor.py", line 1625, in _processSingleImageHtml
    image_html = self._processImage(mime, extended)
  File "/anki/qt/aqt/editor.py", line 1729, in _processImage
    if not mime.hasImage():
RuntimeError: wrapped C/C++ object of type QMimeData has been deleted
```

phu54321:
403 error message box is expected. I think it could be a explicit visual cue that a fallback is working. The image content should be fetched from the clipboard though.

What are your OS? Wsl ubuntu? I probably should test this on other OSs too. (Only have tested on windows.)

dae:
I appreciate the contribution, but think this would be better done after #4029, as the copy/paste code needs to be reworked as part of that.

GithubAnon0000:
As a sidenode:

You mention that you're proposing to try the download first and only use the already available (and locally downloaded) clipboard content if the download fails. I have to admit that I never copy paste media into anki (I always download them with my webbrowser and then import them), but I'm quite unhappy with that proposal; I'd say it's much better to use the clipboard content first and only try the download if something fails.

Wouldn't it be much better to catch messages like the 403 and try to construct an img tag with the available info from the clipboard (while using the image from the clipboard)?

phu54321:
> you're proposing to try the download first and only use the already available (and locally downloaded) clipboard content if the download fails.

That's primarily a backward compatibility behavior, but it has some pros, too.

- Some OS doesn't respect image transparency on clipboard.
- Some images are in GIF or WebP format. We'd wanna preserve that format. (e.g. Animated GIF)

shut-up-and-take-my-money-money

> but think this would be better done after https://github.com/ankitects/anki/pull/4029, as the copy/paste code needs to be reworked as part of that.

I guess that rework would take quite long. Maybe I could make an addon for this PR, so it could be used until the update lands.

abdnh:
@phu54321 
> I guess that rework would take quite long. Maybe I could make an addon for this PR, so it could be used until the update lands.

I was actually working on porting copy & paste handling this week as part of that PR, and I hope to get the whole PR in a good state for review by the end of this month.

phu54321:
https://ankiweb.net/shared/info/934503660

Temporary addon for the issue, if someone wanna try.

Issue #3081, opened 2024-03-19
PyOxidizer has been abandoned and will need replacing


dae:
The tool we use to package binary builds hasn't received an update since Jan 2023, which I was nervous about. Yesterday, the project lead wrote a post admitting that they're unlikely to work again on it in the future. It's possible someone else will pick it up, but it was a complex project, so I suspect nobody will. It's a shame, because it's considerably nicer than the PyInstaller packaging tool we were using previously.

There's no immediate rush to switch away from PyOxidizer, but we will eventually need to, either when we want to upgrade past Python 3.9/10, or when PyOxidizer sufficiently bit-rots.

andrewsanchez:
I wonder if [briefcase](https://github.com/beeware/briefcase) could help out here. It's the tool built and used by [beeware](https://beeware.org/) which was created by the amazing [@ freakboy3742](https://github.com/freakboy3742), and now financially backed by the [support of Anaconda](https://beeware.org/news/buzz/exciting-news-for-the-future-of-beeware/).

I saw Russell Keith-Magee [present at Djangcon](https://www.youtube.com/watch?v=ewwDcRRE4z0). I think Beeware is an incredibly exciting and promising development in the Python world, especially now that Anaconda is funding (I think at least 2, atm) full-time developers to work on it. [This podcast interview](https://www.listennotes.com/podcasts/the-real-python/sharing-your-python-app-u-9A4bC6t_O/) with Russell was really interesting and informative.

dae:
There may be some good news here: https://github.com/indygreg/PyOxidizer/issues/751#issuecomment-2650409898

andrewsanchez:
Is a binary/embedded Python interpreter strictly required to build and distribute Anki Desktop? I'm curious because I was wondering if [uv](https://docs.astral.sh/uv/) could come into play here since `uv` makes it trivial to create a virtual environment. So rather than actually bundling the embedded interpreter binary (is that the right way to say that? 😆), one could bundle `uv` and use it to install Python and dependencies on the fly. You could still bundle wheels/CPython artifacts and install from the local wheels.

andrewsanchez:
It looks like Astral, the same organization behind `uv`, also has a repo with redistributable builds of Python: https://github.com/astral-sh/python-build-standalone

Found it on the PyOxidizer guy's pinned repos:

![Image](https://github.com/user-attachments/assets/bb4d214c-039d-4f5b-baf5-4fc270dcb3cd)

dae:
[...]

PR #4074, opened 2025-06-12
Migrate build system to uv

### LINKED ISSUE #3787: Consider using uv in our build system

dae:
Closes #3787, and is a step towards #3081 and #4022.

If playing with this branch, rm -rf out/pyenv when switching back.

This change breaks our PyOxidizer bundling process. While we probably could update it to work with the new venvs & lockfile, my intention is to use this as a base to try out a uv-based packager/installer.

See the first commit for some further notes.


dae:
@euank @knobix it would be great if you could check these changes don't break your PYTHON_BINARY / OFFLINE_BUILD workflows, and submit a patch here if they accidentally do

dae:
Associated forum thread: https://forums.ankiweb.net/t/new-online-installer-launcher/62745

dae:
Regardless of the outcomes of the forum discussion, I think we need to drop PyOxidizer with haste, as the high-severity CVEs on blocked dependencies are starting to build up, and Python 3.9's not long for this world. If the uv-based launcher approach ultimately proves unpopular, we can pivot to briefcase or pyinstaller, and the uv usage for building will be nice to keep in any case. If anyone has objections to this being merged in, please speak up soon.

(will be following up on the forum tomorrow about the launcher feedback)

dae:
@L-M-Sherlock this has updated us to fsrs 4.1.1 as we weren't pinning a specific commit - is that ok?

L-M-Sherlock:
It's OK if it doesn't introduce a compatible problem. In fsrs 4.1.1, I bump `rand` from `0.8` to `0.9` which caused a breaking change on the `StdRng` : (

dae:
It's ok, I updated the rand crate at the same time :-)

dae:
Another day, another Python CVE. Let's not wait any longer.

qt/pyproject.toml
```diff
@@ -0,0 +1,80 @@
...
+  "psutil; sys.platform == 'win32'",
+  "pywin32; sys.platform == 'win32'",
+  "pip-system-certs!=5.1",
+  "mock",
+  "types-decorator",
+  "types-flask",
+  "types-flask-cors",
+  "types-markdown",
+  "types-waitress",
+  "types-pywin32",
```
andrewsanchez, line 24: @dae Shouldn't `types-*` packages should be optional development dependencies? Looks like there are potentially some other packages in the `pyproject.toml` files that can be excluded in the installation (mock and werkzeug maybe?).

pylib/pyproject.toml
```diff
@@ -0,0 +1,35 @@
...
+
+[tool.hatch.build.targets.wheel]
+packages = ["anki"]
+
+[tool.hatch.version]
+source = "code"
+path = "../python/version.py"
+
+[tool.hatch.build.hooks.custom]
+path = "hatch_build.py"
```
andrewsanchez, line 35: @dae  Should `pylib/test/` and anything else be excluded? https://hatch.pypa.io/latest/build/#configuration

dae:
Yep, those look like oversights. Thanks!

build/ninja_gen/src/bin/update_uv.rs
```diff
@@ -0,0 +1,144 @@
...
+
+fn fetch_uv_release_info() -> Result<String, Box<dyn Error>> {
+    let client = Client::new();
+
+    println!("Fetching latest uv release info from GitHub...");
+    // Fetch latest release info
+    let response = client
+        .get("https://api.github.com/repos/astral-sh/uv/releases/latest")
+        .header("User-Agent", "Anki-Build-Script")
+        .send()?;
```
andrewsanchez, line 20: Probably not an issue since I don't see this being use programmatically anywhere, IIUC, but worth noting:

> The primary rate limit for unauthenticated requests is 60 requests per hour.

https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users

pylib/pyproject.toml
dae: Inspecting the wheel revealed unwanted `__pycache__` files, but no tests. I'll push a fix later.

build/ninja_gen/src/bin/update_uv.rs
dae: We're only likely to run it once every few months, so I think we'll be fine :-)

zanieb:
👋 Hi! 

I work on uv at Astral. Please feel free to ping me if there's anything we can do to help out!

Grinkers:
Thanks for reaching out, I have a 100% repro and write rust daily so I seem uniquely qualified to help out. I will circle back to Anki with the results and likely eventual code cleanup from our workaround #4162

Grinkers:
Thanks @zanieb for the pointers.

https://github.com/ankitects/anki/pull/4191
Proper fix.

In my case, I had a system uv config that was silently being merged in. The other user on the forum may have had either a system and/or user uv config.

dae:
Thank you both for getting to the bottom of this!

PR #4076, opened 2025-06-12
Ignore TaskManager's on_done callback if collection unloaded

abdnh:
AnkiHub runs sync-related background tasks before shutdown, and they often throw errors inside on_done callbacks depending on timing [1].

@RisingOrange heads-up in case you have feedback.

[1]: https://github.com/AnkiHubSoftware/ankihub_addon/pull/1128

### DIFF
--- a/qt/aqt/taskman.py
+++ b/qt/aqt/taskman.py
@@ -84,8 +84,15 @@ class TaskManager(QObject):
         fut = executor.submit(task, **args)
 
         if on_done is not None:
+
+            def wrapped_done(future: Future) -> None:
+                if uses_collection and not self.mw.col:
+                    print(f"Ignored on_done as collection unloaded: {repr(on_done)}")
+                    return
+                on_done(future)
+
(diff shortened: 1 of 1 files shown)


RisingOrange:
Looks good to me. 

One thing to note: the issue in https://github.com/AnkiHubSoftware/ankihub_addon/pull/1128 is that `aqt.mw.col.db` is `None`, not `aqt.mw.col`. So this fix wouldn't address that particular case. I'm not sure why `col.db` is `None` there, and whether it would make sense to also ignore `on_done` callbacks when `col.db` is `None`. I'd need to look into it more.

Any thoughts?

abdnh:
Both col and db are set to None in _unloadCollection() (db a bit earlier in the col.close() call). But db is also cleared in close_for_full_sync(). That might explain some errors.

I'll add an extra check.

dae:
Yeah, I think that's probably the best approach. Thanks!

PR #4077, opened 2025-06-12
Use abbreviations for days of week

brishtibheja:
Easy days looked a bit cramped on my phone so I shortened the strings and used abbr. instead of the full names.


dae:
Thanks!

PR #4078, opened 2025-06-12
Fix/Missing CardID special field

Luc-Mcgrady:
Also added a test so it doesn't happen again.

ref:
- https://github.com/ankitects/anki/pull/4046#issuecomment-2968047429 (thanks @bryanboateng)


dae:
The addition of a test is much appreciated :-)

PR #4084, opened 2025-06-14
Fix unescaped deck names potentially missing from overview

iamllama:
Fixes the same bug as:
- https://github.com/ankitects/anki/pull/3960

but for the deck overview page

### DIFF
--- a/qt/aqt/overview.py
+++ b/qt/aqt/overview.py
@@ -2,6 +2,7 @@
 # License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 from __future__ import annotations
 
+import html
 from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Any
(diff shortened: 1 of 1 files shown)


qt/aqt/overview.py
```diff
@@ -191,7 +192,7 @@ def _renderPage(self) -> None:
             self._show_finished_screen()
             return
         content = OverviewContent(
-            deck=deck["name"],
+            deck=html.escape(deck["name"]),
```
abdnh: This is passed to the overview_will_render_content hook and will break add-ons that do something like `mw.col.decks.by_name(content.deck)`.

Example: https://github.com/abdnh/anki-total-cards-matched-by-filtered-deck-search/blob/892319458cce1ee39f0b63217221f9b334e92acd/src/main.py#L17

dae:
Thank you both!

PR #4086, opened 2025-06-15
Limit study time to hours in reviews graph

user1823:
Relevant discussions:
- https://forums.ankiweb.net/t/reviews-graph-units-of-total-time-studied-suggestion/61237
- https://forums.ankiweb.net/t/why-does-anki-display-study-time-in-months/37722
- https://forums.ankiweb.net/t/poll-use-hours-in-total-time-stats/62076
- https://github.com/ankitects/anki/pull/3901#issuecomment-2973161601

I couldn't build Anki to test this but I am pretty confident that the changes would work.

### DIFF
--- a/ts/lib/tslib/time.ts
+++ b/ts/lib/tslib/time.ts
@@ -7,8 +7,8 @@ export const SECOND = 1.0;
 export const MINUTE = 60.0 * SECOND;
 export const HOUR = 60.0 * MINUTE;
 export const DAY = 24.0 * HOUR;
-export const MONTH = 30.417 * DAY; // 365/12 ≈ 30.417
 export const YEAR = 365.0 * DAY;
+export const MONTH = YEAR / 12;
 
 export enum TimespanUnit {
     Seconds,
@@ -36,21 +36,25 @@ export function unitName(unit: TimespanUnit): string {
     }
 }
 
-export function naturalUnit(secs: number): TimespanUnit {
+export function naturalUnit(secs: number, maxUnit: TimespanUnit = TimespanUnit.Years): TimespanUnit {
     secs = Math.abs(secs);
-    if (secs < MINUTE) {
-        return TimespanUnit.Seconds;
-    } else if (secs < HOUR) {
-        return TimespanUnit.Minutes;
-    } else if (secs < DAY) {
-        return TimespanUnit.Hours;
-    } else if (secs < MONTH) {
-        return TimespanUnit.Days;
-    } else if (secs < YEAR) {
-        return TimespanUnit.Months;
-    } else {
-        return TimespanUnit.Years;
+    const thresholds = [
+        { unit: TimespanUnit.Years, min: YEAR },
+        { unit: TimespanUnit.Months, min: MONTH },
+        { unit: TimespanUnit.Days, min: DAY },
+        { unit: TimespanUnit.Hours, min: HOUR },
+        { unit: TimespanUnit.Minutes, min: MINUTE },
+        { unit: TimespanUnit.Seconds, min: 0 },
+    ];
+    for (const { unit, min } of thresholds) {
+        if (unit > maxUnit) {
+            continue;
+        }
+        if (secs >= min) {
+            return unit;
+        }
     }
+    return TimespanUnit.Seconds;
 }
 
 /** Number of seconds in a given unit. */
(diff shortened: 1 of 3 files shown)


ts/routes/graphs/reviews.ts
```diff
@@ -340,7 +340,7 @@ export function renderReviews(
         averageAnswerTime: string,
         averageAnswerTimeLabel: string;
     if (showTime) {
-        totalString = timeSpan(total / 1000, false);
+        totalString = timeSpan(total / 1000, false, true, TimespanUnit.Hours);
```
user1823, line 343: IMO, we can set precise to False here because, for example, showing 52.26 hours of study time doesn't provide much additional information as compared to showing 52 hours of study time.

```suggestion
        totalString = timeSpan(total / 1000, false, false, TimespanUnit.Hours);
```

GithubAnon0000: I vote against this.

user1823:
Sorry, I am not sure how to fix the formatting error in `CardStats.svelte`. I am fine with dropping the changes in that file if the error is not trivial to fix.

ts/routes/graphs/reviews.ts
user1823: I understand your point. Perhaps, we should skip this for now and wait for a subsequent overhaul of the rounding system (https://forums.ankiweb.net/t/reduce-precision-of-time-studied-today/42309).

ts/routes/card-info/CardStats.svelte
```diff
@@ -58,7 +58,12 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         if (stats.memoryState) {
             let stability = timeSpan(stats.memoryState.stability * 86400, false, false);
             if (stats.memoryState.stability > 31) {
-                const nativeStability = stats.memoryState.stability.toFixed(0);
+                const nativeStability = timeSpan(
+                 stats.memoryState.stability * 86400,
+                 false,
+                 false,
+                 TimespanUnit.Days,
+                 );
```
GithubAnon0000:
```suggestion
                const nativeStability = timeSpan(
                    stats.memoryState.stability * 86400,
                    false,
                    false,
                    TimespanUnit.Days,
                );
```


The above should fix it I believe (your indentation was wrong)

user1823: Thanks Anon

ts/lib/tslib/time.ts
```diff
@@ -36,21 +36,25 @@ export function unitName(unit: TimespanUnit): string {
...
-        return TimespanUnit.Years;
+    const thresholds = [
+        { unit: TimespanUnit.Years, min: YEAR },
+        { unit: TimespanUnit.Months, min: MONTH },
+        { unit: TimespanUnit.Days, min: DAY },
+        { unit: TimespanUnit.Hours, min: HOUR },
+        { unit: TimespanUnit.Minutes, min: MINUTE },
+        { unit: TimespanUnit.Seconds, min: 0 },
+    ];
+    for (const { unit, min } of thresholds) {
```
dae: How about keeping things simple? You don't need to change this routine at all, and can just do Math.min(naturalUnit(seconds), maxUnit) in timeSpan().

dae (commented):
The initial proposal was about changing the 'amount of time studied' statistics to hours. I think the arguments for that made sense. But I don't think they apply to the stability. For intervals, I think "FSRS thinks you'll remember this for 3 months" is more useful/easily understandable than "..remember this for 2160 hours". If you feel strongly about that change, has it been discussed somewhere already?

user1823:
> For intervals, I think "FSRS thinks you'll remember this for 3 months" is more useful/easily understandable than "..remember this for 2160 hours"

Are you talking about the changes in CardInfo.Svelte? There, the stability is shown in days, not hours. And that's in addition to the stability in months. And that's the current behaviour too. I have just used the new TS method to do that because the new method also adds the unit (days) to the number.

ts/lib/tslib/time.ts
user1823: Makes sense. I will do that soon.

dae:
Yeah, the card info. I'd forgotten we also show days, but wasn't that mainly for FSRS devs/enthusiasts as it's easier to work with than months? How do the other FSRS regulars feel about switching days to hours? 

(the fact that we're showing both there makes me less concerned about changing this if that's what people want)

user1823:
> How do the other FSRS regulars feel about switching days to hours?

I repeat, I haven't changed days to hours. I have just added the unit "days" there.

> but wasn't that mainly for FSRS devs/enthusiasts as it's easier to work with than months?

Yes, but there shouldn't be any harm in adding a unit. 😅

dae:
I apologise for not reading carefully enough. 😞 In that case, looks good to go.

PR #4087, opened 2025-06-16
Add flip button to editor to swap Front/Back fields (closes #3911)

### LINKED ISSUE #3911: Flip card from browse menu
Related to this
https://forums.ankiweb.net/t/why-is-there-no-simple-way-to-flip-cards-back-to-front/41915/2

I think it would be easier to use if there was a "flip" button in the notes menu implementing the method described by nymvaline.

Image

~~By the way, I would like to implement this feature.~~ I'm letting it free if someone wants to work on it.

Mohammad-Alshabout:
This PR adds a flip button (↔) to the editor to swap the Front and Back fields of a note.

- Saves current content before flip
- Swaps Front and Back field text
- Updates the editor view
- Displays success tooltip

Closes #3911

### DIFF
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -1790,3 +1790,31 @@ def set_image_occlusion_button(editor: Editor) -> None:
 
 gui_hooks.editor_did_load_note.append(set_cloze_button)
 gui_hooks.editor_did_load_note.append(set_image_occlusion_button)
+
+def add_flip_button(buttons, editor: Editor):
+    # Callback-Funktion beim Klick: zuerst Änderungen speichern, dann Flip ausführen
+    def on_flip(editor: Editor):
+        editor.saveNow(lambda: flip_fields(editor))
+    # Felder tauschen und Editor aktualisieren
+    def flip_fields(editor: Editor):
+        front_text = editor.note["Front"]   # aktuellen Front-Text auslesen
+        back_text = editor.note["Back"]     # aktuellen Back-Text auslesen
+        editor.note["Front"] = back_text    # Front-Feld setzen
+        editor.note["Back"] = front_text    # Back-Feld setzen
+        editor.loadNote()                   # Editor-Ansicht mit neuen Werten aktualisieren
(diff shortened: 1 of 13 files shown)


qt/aqt/editor.py
```diff
@@ -1790,3 +1790,37 @@ def set_image_occlusion_button(editor: Editor) -> None:
...
+
+    def flip_fields(editor: Editor):
+        note = editor.note
+        if note is None:
+            return
+
+        front_text = note["Front"]
+        back_text = note["Back"]
+        note["Front"] = back_text
+        note["Back"] = front_text
```
GithubAnon0000, line 1808: I didn't test this (and am a novice programmer), but does this acutally reliably work? What if I rename my "Back" field to "Answer" as an example?

Or when the field names are different due to translations?

**Edit:** Also consider the case that there are more than two fields. How does it work in that case?

qt/aqt/editor.py
```diff
@@ -1790,3 +1790,37 @@ def set_image_occlusion_button(editor: Editor) -> None:
...
+        editor.loadNote()
+        if not editor.addMode:
+            note.flush()
+        tooltip("Fields flipped successfully.")
+
+    btn = editor.addButton(
+        icon=None,
+        cmd="flipFields",
+        func=on_flip,
+        tip="Front/Back Felder vertauschen",
```
GithubAnon0000, line 1818: Strings must be translatable. You can create an entry in the ftl/core/ folder and then reference that from code.
E.g. `browsing-add-notes` from [browsing.ftl](https://github.com/ankitects/anki/blob/main/ftl/core/browsing.ftl) can be accessed with `tr.browsing_add_notes()` as done in [tree.py](https://github.com/ankitects/anki/blob/5cb191d624ce8ea1f8bc8578cd2414df8a54904d/qt/aqt/browser/sidebar/tree.py#L954).

dae:
I'm sorry for the wasted effort @Mohammad-Alshabout 😞  , but I think this feature doesn't make sense. I've posted a comment on #3911.

PR #4088, opened 2025-06-16
Add streak display to the overview window

abdalrahman-alali:
I added a simple streak display to Anki's overview screen.
It shows on how many days the user has studied within the last 30 days.
The goal is to improve learning motivation by making progress more visible.

Screenshot of the streak feature


brishtibheja:
the sentiment around streaks is pretty consistent both here and among ankidroid devs, that it can be pretty toxic if not done right. things like leaderboards with seasons are much better than streaks.

(not anti-streaks but it should be an opt-in at the very least)

dae:
While I appreciate your enthusiasm, this PR is a mess, and I don't think we're ready to move to the implementation stage yet. We need to discuss it first, and decide whether it's a good idea, and if it is, how exactly it will be implemented. There's a bunch of other things going on at the moment, so it may be some time until we can do that I'm afraid.

PR #4089, opened 2025-06-17
Fix/seconds elapsed since last review in card stats shouldn't be negative

L-M-Sherlock:
The previous `timing.now` is the timestamp of opening the collection today, which is earlier than the real current time.


dae:
Let's fix the underlying problem instead, as it's confusing to have a 'now' that is not 'now' :-)

PR #4092, opened 2025-06-17
disable user selection

brishtibheja:
particularly is an issue on phones where scrolling is hindered by text selection, and it looks ridiculous like im using a website.

### DIFF
--- a/ts/routes/base.scss
+++ b/ts/routes/base.scss
@@ -21,6 +21,10 @@ textarea {
     background: var(--canvas-inset);
 }
 
+textarea {
+    user-select: none;
+}
+
 input {
     color: var(--fg);
 }


GithubAnon0000:
Were does this take effect?

brishtibheja:
This is loaded on every shared svelte screen. I can't build and test it
myself, can you do tha?

GithubAnon0000:
If you give me 2 – 3 days, yes.

Can you give me a few examples where exactly this should be applied? Would the FSRS params box be a place where you want to restrict the users from selecting stuff?


<hr>

As a side note: I'm not sure I'd love to see such a restriction. I think websites and apps should always allow the user to select and copy the strings (though I understand there are legitimate cases where it's better to disable it completely).

dae:
An argument could be made for disabling it in parts of our interface, but doing it globally like this is likely to cause problems once we move more user content into Svelte, like the review screen. And there may be areas where keeping it is convenient - think users copy+pasting card info, etc.

Issue #3833, opened 2025-02-22
Consider a cross-platform add-on system for templates


dae:
This is something I've mentioned in passing in the past, and I thought I should write it up for discussion/easy linking.

I should lede with a disclaimer: I don't think we're at a stage where we can start working on this yet. I'd like to see our Svelte migration further along first. After #3830, it will be about time to look into migrating the reviewing screen (which also needs a separate ticket). As part of that, we may want to look into isolating user content into an iframe so it can't modify the surrounding UI or access our private APIs, and the decisions we make there will influence what is possible below. It would also require extensive changes to our card layout screen, so I think we'd also want to have that in Svelte prior to implementing this.

Anyway, on to the topic itself. Shared decks have become a way for authors to distribute complex styling and behaviour, and lots of shared decks contain a bunch of CSS and JS in them. This presents a number of problems:

- If the styling or code ever needs tweaks for newer Anki versions, deck authors cannot have no way to push out updates
- Common functionality like a shortcut key to reveal hints ends up being duplicated across multiple templates/decks
- Providing such common functionality needs to be done via shared deck, with at least one dummy note
- Complex styling and JS makes it very hard for beginner/intermediate users to tweak the styling to their needs
- It's awkward to edit/test these changes inside the template editor

This got me thinking - what if we had a way to decouple generic styling and reusable scripts into an add-on like system that could be updated independently?

The way I imagine it working is that templates could list a number of 'components' that they require. These could include things like themes/styling tweaks. For example, a 'modern base' template that is closer to https://forums.ankiweb.net/t/modernize-default-card-template-for-readability-and-simplicity/53356, but also more fine-grained components like 'left-aligned text', 'cute dog background', etc. From the user's perspective, for simple things like this, they could browse the available components without having to deal with any HTML/JS themselves. And we could potentially highlight the ones a large number of users may wish to use.

The components could also be more functionality-based, e.g. one that enables an alternative Math renderer like katex, or something like a (pedagogically-suboptimal) multiple-choice function.

It might be helpful to have templates have a separate list of 'required' and 'recommended' components, so users know which ones they can optionally add/remove, and which ones are essential for functionality of that template.

Since this functionality would only alter the way Anki behaves during review/preview and wouldn't touch the UI, it would be cross-platform. The updating/downloading code could be implemented in our Rust layer, and management in a shared Svelte page.

[...]

PR #4093, opened 2025-06-17
add default styling for tables

brishtibheja:
changed default stylesheet to add border/margins/padding for table. the current defaults mean you don't even have borders for your table, which affects readability.

relevant forum post: https://forums.ankiweb.net/t/sane-css-defaults-for-tables/62803?u=sorata

### DIFF
--- a/ts/reviewer/reviewer.scss
+++ b/ts/reviewer/reviewer.scss
@@ -36,7 +36,6 @@ table {
 th, td {
     padding: 0.5em;
     border: 1px solid;
-    border-color: vars.palette(darkgray, 0);
 }
 
 img {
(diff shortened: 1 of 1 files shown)


Danika-Dakika:
Please see [the forum thread](https://forums.ankiweb.net/t/sane-css-defaults-for-tables/62803?u=danika_dakika) for my objections to this change.

dae:
Yeah, I'm afraid I think this will be too disruptive. We can't make fundamental changes to the implicit styles like this without breaking existing card templates, and that causes a lot of work for the community. Such changes, if done, are better made using some sort of opt-in/versioned styling, and I think that's going to be best done using something like a "pretty & modern" theme that users can opt into / new installs can default to, perhaps via #3833. We haven't even got the previous HR changes out into the hands of many users yet, and I wouldn't be surprised if someone somewhere is upset by that change (but when merging that change, I reasoned that it's probably not going to cause a lot of disruption - this change to table styling is much more invasive)

PR #4095, opened 2025-06-18
Clear desired retention and decay when changing decks

user1823:
These values are preset-specific and entries from previous deck may cause issues.

PS:
I created a somewhat related [bug report](https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/575) in the FSRS Helper add-on repo and the proposed solution ended up requiring creation of a new function in Anki (which will also solve a bug in Anki). Please share your thoughts on the proposed solution there.


dae:
Thank you for the fix! Your description in the other topic was a bit vague, but based on what you've said so far, it sounds ok.

PR #4096, opened 2025-06-18
Update the default value of FSRS-6 decay in forgetting curve

user1823:
Changed in https://github.com/open-spaced-repetition/fsrs-rs/commit/037345fd57472ea392a6086f217b1c73a9fa171a

I am not sure what's the best way to do this. Ideally, Anki should automatically use the value from FSRS-rs here.


dae:
If @L-M-Sherlock exposes the constant in fsrs's public API, we can just use that and no test will be necessary. No harm in merging this for now. 👍

L-M-Sherlock:
@dae, the constant has been exposed in fsrs-rs. But the affected code is in TypeScript.

https://github.com/open-spaced-repetition/fsrs-rs/blob/75cfeeaaf560bf9f30a714b42dccb3fd2a901696/src/lib.rs#L21

PR #4097, opened 2025-06-18
disable text selection in deck options

brishtibheja:
previous pr at https://github.com/ankitects/anki/pull/4092

here, disabled text selection for everything except the input/textareas.


GithubAnon0000:
Why do you prefer disabling text selection here? The way I see it is that it currently does no harm and also allows users to easily copy and paste the string if they feel like it (without having to use the dev console).

dae:
I get how keeping selection can look a bit unpolished/like a webpage sometimes. But it's also often times useful. Consider a user who wants to copy&paste something written in the deck options screen to Google it for example. I think we need to discuss this and get buy-in from other community members before proceeding.

(also, this PR doesn't appear to actually work - did you test it locally?)

brishtibheja:
It's probably less of an issue on desktop than it is on mobile. I keep selecting text when I don't want to and that interferes with what I'm actually trying to do.

> also, this PR doesn't appear to actually work - did you test it locally?

not really, thought this would be quite easy to do. @Luc-Mcgrady do you know what is wrong here? need your help.

Luc-Mcgrady:
Nothing in `deck-options-base.scss` appears to be applied to the deck options. If I were to guess its because it's mistakenly imported in the `.ts` file instead of the `.svelte` file which means it doesn't do anything. No ones noticed this so far because `base.scss` is still imported correctly, in a svelte file, and is identical to `deck-options-base.scss`. If i add an import

Here's a working diff

```diff
diff --git a/ts/routes/deck-options/DeckOptionsPage.svelte b/ts/routes/deck-options/DeckOptionsPage.svelte
index 2201493b2..82b6b8e59 100644
--- a/ts/routes/deck-options/DeckOptionsPage.svelte
+++ b/ts/routes/deck-options/DeckOptionsPage.svelte
@@ -3,6 +3,7 @@ Copyright: Ankitects Pty Ltd and contributors
 License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 -->
 <script lang="ts">
+    import "./deck-options-base.scss";
     import type { Writable } from "svelte/store";
 
     import "$lib/sveltelib/export-runtime";
diff --git a/ts/routes/deck-options/index.ts b/ts/routes/deck-options/index.ts
index 3b88c4d0c..ef2da7f56 100644
--- a/ts/routes/deck-options/index.ts
+++ b/ts/routes/deck-options/index.ts
@@ -6,7 +6,6 @@
  */
 
 import "$lib/sveltelib/export-runtime";
-import "./deck-options-base.scss";
 
 import { getDeckConfigsForUpdate } from "@generated/backend";
 import { ModuleName, setupI18n } from "@tslib/i18n";
```

Edit: This isn't limited to this screen I'll pr a fix.

user1823:
> I keep selecting text when I don't want to and that interferes with what I'm actually trying to do.

But, selecting text requires a long press. So, I don't understand how it can be triggered accidentally. If accidental text selection was so common, an argument can be made that every website should disable text selection because it would interfere with scrolling.

brishtibheja:
It's a habit on phones that we keep our thumbs glued on screen. And as for websites, they want people to copy things and share. It helps them. No app does this though.

brishtibheja:
Noticed something interesting. On websites that I tested, I can select text and still continue to scroll which is not what happens with Anki (scrolling is hindered and more text gets selected). 

So, seems like browsers are better able to handle touch events than regular apps. They've already considered people long-tapping out of habit.

To test, try accessing https://docs.ankiweb.net from a regular browser and from AnkiDroid webview (deck options > click links from tooltip).

Regardless of whether AnkiDroid can do something on their side, I think selectively disabling text selection on some screens will be a good choice.

GithubAnon0000:
> To test, try accessing https://docs.ankiweb.net from a regular browser and from AnkiDroid webview (deck options > click links from tooltip).

I can scroll just fine in AD if text is selected. If you're having an issue, I don't think it's because of AD, but rather your android system webview which renders that page.

Expertium:
I made a survey: https://forums.ankiweb.net/t/disable-text-selection-in-deck-options/62892/1
I'll report the results tomorrow

brishtibheja:
> I can scroll just fine in AD if text is selected. If you're having an issue, I don't think it's because of AD, but rather your android system webview which renders that page.

Maybe you misunderstood. I meant tap and hold on text, and continue to move your finger (this extends the selected text). I didn't mean, select text first, lift your finger and try scrolling then. 

Can you rest your thumb on screen without lifting it up, then try to scroll once text is selected? I actually tested two different devices, same behaviour.

brishtibheja:
@Expertium thanks. maybe explain to people why we are doing it? also, im actually not disabling selection inside input fields, like someone already said in the form.

GithubAnon0000:
> Can you rest your thumb on screen without lifting it up, then try to scroll once text is selected? I actually tested two different devices, same behaviour.

Yes, I can. Though obviously it selects the text while scrolling (which is expected behavior from webapps).

brishtibheja:
I think you're talking about scrolling as the selection range moves below the fold? If not, then idk (tested this with different device/webview ver).

Expertium:
@dae @brishtibheja results: https://docs.google.com/forms/d/1A4tgpIrMcbGgk84wixw-qLT4oYmkkz_Cd2Hw_vM34FE/viewanalytics
This survey was conducted on Discord, r/Anki and the forum, though I didn't add a question to keep track of who came from where.

Anyway, the majority are either against it or don't care. And yes, I did later add "(NOT in input fields, selection in input fields will remain)", though it's possible that the earlier responses would've been different if I had added it from the beginning.
image

dae:
Thanks for the survey. Given that the majority of users are opposed or ambivalent, I think it needs to be a more targeted change (e.g. only scoped to AnkiDroid, and only if the AD devs think this is actually worth doing).

chrislongros:
I personally think it is a useful feature and it should be kept.

brishtibheja:
You can just type it out right?

GithubAnon0000:
Why do all that typing if you could simply copy-paste? That seems to be something most users agreed with.

brishtibheja:
I wonder how many avenues there are for something like that to be put into use. Most of the text is just buttons and feature names/descrip. If you need to understand one particular feature, Anki already has built-in tooltips and we provide links to the manual. From a usage perspective, it doesn't seem that useful. At least I've never had to search anything from here.

GithubAnon0000:
> Anki already has built-in tooltips and we provide links to the manual.

I would agree with you. But as we can see on the forums and on reddit, a lot of users either ignore those resources or have no idea they exist.

I guess it just comes down to preference. And I'm wondering how useful it would be to disable text selection on mobile, considering https://github.com/ankitects/anki/pull/4097#issuecomment-2987732499 is quite convincing for me (and I find it hard to believe that users don't lift their thumb off the screen when using a smartphone).

brishtibheja:
> I would agree with you. But as we can see on the forums and on reddit, a lot of users either ignore those resources or have no idea they exist.

And those people rely on online guides and tutorials, they don't investigate each feature seperately. But I get your point and think we need a solution to that, rather than accepting it. FYI, there isn't much info on individual features online except for those two places.

> and I find it hard to believe that users don't lift their thumb off the screen when using a smartphone

If accidental long-touches weren't a thing, browsers wouldn't try to handle this case in the way they do now.

The main two ways this happens for me is:

1. I'm trying to read things slowly. So I touch and scroll, thumb still on screen, I might move back a little or continue forward.
2. I'm trying to find something fast. I do a quick swipe and then tap/hold the screen when I think I'm in the area I should be in.


(Also if I have dust on my screen and am trying to wipe it but that's secondary)

ts/routes/deck-options/deck-options-base.scss
```diff
@@ -28,6 +30,8 @@ input {
 // Setting 100% height causes the sticky element to hide as you scroll down on Safari.
 html {
     height: initial;
+    -webkit-user-select: none;
+    user-select: none;
 }
```
Luc-Mcgrady, line 38:
```suggestion
}

@media (pointer: coarse) {
  html {
      -webkit-user-select: none;
      user-select: none;
  }
}
```

Maybe something like this? (untested)

ref: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/pointer

david-allison:
> Thanks for the survey. Given that the majority of users are opposed or ambivalent, I think it needs to be a more targeted change (e.g. only scoped to AnkiDroid, and only if the AD devs think this is actually worth doing).

Might up for a discussion in the far future, but I can't see it being worthwhile any time soon on our side. Would take up tons of discussion time for marginal (if any) benefit, and we have numerous places in the app [and in this screen] where we can make more impactful UI/UX/accessibility changes.

<details>

<summary>Issues detected in Deck Options (AnkiDroid2.22alpha1 with anki 25.06b6)</summary>

</img>

</details>

ts/routes/deck-options/deck-options-base.scss
brishtibheja: ~~that would also select people using keyboard/voice assist, so people will see inconsistent behaviour in a laptop vs a PC.~~ it won't, i was wrong

(actually i still think disabling this everywhere is a better choice, mainly for consistency's sake)

GithubAnon0000:
> Issues detected in Deck Options (AnkiDroid2.22alpha1 with anki 25.06b6)
> [IMG]

Are you talking about inconsistent (and little) spacing? Not sure what those squares could mean otherwise.

ts/routes/deck-options/deck-options-base.scss
Luc-Mcgrady: maybe try it the same as its done here?

https://github.com/ankitects/anki/blob/bf46a5f08cea200dd5353066acebec030b33c816/ts/routes/deck-options/CardStateCustomizer.svelte#L46-L52

david-allison:
Accessibility issues using Google's scanner:

* Touch targets should be larger (aim for 48dp, most of which are around 22dp)
* Some of the 'reset' buttons are detected as being visible by the tool, therefore probably also by screen readers
* Contrast on the switches isn't ideal

My first change would be adding affordances so 'New cards/day' etc... appears clickable on mobile, since we don't have 'mouseover' events.

EDIT: unsubscribing, ping me if I'm needed

ts/routes/deck-options/deck-options-base.scss
brishtibheja: that only works for old safari browsers (@dae should we update that one?) 

here, maybe let's just go with your previous suggestion slightly modified

```suggestion
} 

@media (hover: none) and (pointer: coarse) {   
  -webkit-user-select: none;
  user-select: none;
 }
}
```

Luc-Mcgrady: Sounds good to me.

ts/routes/deck-options/deck-options-base.scss
```diff
@@ -30,8 +30,12 @@ input {
...
     height: initial;
-    -webkit-user-select: none;
-    user-select: none;
+} 
+
+@media (hover: none) and (pointer: coarse) {   
+  -webkit-user-select: none;
+  user-select: none;
+ }
 }
```
GithubAnon0000:
```suggestion
@media (hover: none) and (pointer: coarse) {   
    -webkit-user-select: none;
    user-select: none;
}
```

There was one stray `}` and intendation was two spaces instead of four (interessting that the tests passed though).

dae:
I'm afraid this doesn't feel like a priority right now.

Issue #4091, opened 2025-06-17
The interval is incorrect after apply `Set Due Date` on interday/intraday learning cards when `fsrs_enabled`.


L-M-Sherlock:
For interday learning cards, the `self.interval` is always equal to zero.

For example, if I have a 2d learning step and press `good` on a new card, the due will be two days later but the `interval` is still zero. Then this code will calculate the `new_interval` incorrectly:

https://github.com/ankitects/anki/blob/5cb191d624ce8ea1f8bc8578cd2414df8a54904d/rslib/src/scheduler/reviews.rs#L36-L38

I have no idea how to fix it without storing the last review timestamp in Card because the original `self.interval` is incorrect.

For intraday learning cards, `self.interval=0` is not a problem. However, the current behavior is that the `new_interval` is always equal to zero because the `self.original_or_current_due()` is a timestamp instead of a day offset. So `new_due - self.original_or_current_due()` is negative.

I could fix this case in this way:

```rust
            let due = self.original_or_current_due();
            let due_diff = if is_unix_epoch_timestamp(due) {
                let new_due_secs =
                    TimestampSecs::now().adding_secs(days_from_today as i64 * 86_400);
                ((new_due_secs.0 - due as i64) / 86_400) as i32
            } else {
                new_due - due
            };
            self.interval.saturating_add_signed(due_diff)
```

But the `due_diff` is still inaccurate because the code doesn't consider `rollover_hour`.

user1823:
The second case can be solved by something like this:

```rs
            let mut due = self.original_or_current_due();
            let due = if is_unix_epoch_timestamp(due) {
                let creation = storage.creation_stamp()?;
                due.elapsed_days_since(creation)
            } else {
                due
            };
            self.interval.saturating_add_signed(new_due - due)
```

L-M-Sherlock:
@dae, it's possible to fix the underlying problem before the next release? It requires a change on the DB schema to add a new field.

Or just apply user1823's patch to solve the second case?

user1823:
> It requires a change on the DB schema to add a new field.

A change in DB schema would be too disruptive. As a simpler alternative, we can pack the last review date in the data column, just like we do for FSRS memory states, desired retention and decay.

But, before we do any of this, we need to
[...]

PR #4101, opened 2025-06-19
Fix/set due date on intraday learning card

L-M-Sherlock:
A quick fix for the second case mentioned in #4091:

> For intraday learning cards, `self.interval=0` is not a problem. However, the current behavior is that the `new_interval` is always equal to zero because the `self.original_or_current_due()` is a timestamp instead of a day offset. So `new_due - self.original_or_current_due()` is negative.

Placing the last review info on card data is entangled with too much things, so I make this patch before I figure them out.


dae:
Thanks Jarrett ❤️

PR #4102, opened 2025-06-19
include decay in ComputeMemoryStateResponse

L-M-Sherlock:
prepare for https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/575


rslib/src/scheduler/fsrs/memory_state.rs
```diff
@@ -190,7 +190,15 @@ impl Collection {
             .or_not_found(conf_id)?;
         let desired_retention = config.inner.desired_retention;
         let historical_retention = config.inner.historical_retention;
-        let fsrs = FSRS::new(Some(config.fsrs_params()))?;
+        let params = config.fsrs_params();
+        let decay = if params.is_empty() {
```
user1823: Let's make a new function to obtain the decay from the given parameters. Something like

```rs
let decay = fsrs_decay(params);
```

```rs
pub fn fsrs_decay(params) -> Result<()> {
    if params.is_empty() {
        FSRS6_DEFAULT_DECAY // default decay for FSRS-6
    } else if params.len() < 21 {
        FSRS5_DEFAULT_DECAY // default decay for FSRS-4.5 and FSRS-5
    } else {
        params[20]
    }
}
```

and then use it here and in https://github.com/ankitects/anki/blob/344cac1ef4712822c78d8f96da8bf17144358213/rslib/src/scheduler/fsrs/memory_state.rs#L80

PR #4103, opened 2025-06-19
Fix/unapplied scss

Luc-Mcgrady:
Previously, a large number of scss files were unused.

ref:
- https://github.com/ankitects/anki/pull/4097#issuecomment-2987640698

There are still these imports in ts files but they don't seem as simple to fix so I'll leave them be for now:

image


GithubAnon0000:
Just curious: Did this change the way anki looks or behaves noticably?

Luc-Mcgrady:
Nothing major that I could see while testing it (I added `* {background-colour: red}` to check it was working so any changes would have been obscured by that)

ts/routes/change-notetype/change-notetype-base.scss
```diff
@@ -1,13 +1,13 @@
-@use "../lib/sass/bootstrap-dark";
+@use "../../lib/sass/bootstrap-dark";
```
abdnh: We can use the `$lib` alias.

Luc-Mcgrady:
```
✘ [ERROR] Can't find stylesheet to import.
  ╷
1 │ @use "$lib/sass/root-vars";
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
```
I think `$lib` is just for js/ts. :(

abdnh: `$lib/sass/root-vars` is already used in graphs-base.scss though and is working?

Luc-Mcgrady: Checking again this doesn't work for specifically `ts/routes/image-occlusion/image-occlusion-base.scss` for some reason. I've added it for the other ones now.

abdnh (approved):

dae:
The index.ts files are leftovers from our previous separate-page esbuilds, and the base files were used there too (I seem to recall Henrik wanted to do it that way because he didn't want the base styling to be loaded multiple times if we used the component multiple times, but I don't think we use multiple Page components at once currently anyway, and this is simpler). Thank you for the tidy-ups!

iamllama:
This has had unintended design changes: https://forums.ankiweb.net/t/anki-25-06-beta/62271/129

PR #4104, opened 2025-06-19
add night-mode to default templates

brishtibheja:
Forums: https://forums.ankiweb.net/t/add-nightmode-card-in-styling-css/62824/4?u=sorata

### DIFF
--- a/rslib/src/notetype/styling.css
+++ b/rslib/src/notetype/styling.css
@@ -6,3 +6,7 @@
     color: black;
     background-color: white;
 }
+
+.night-mode .card {
+    color: white;
+}


brishtibheja:
ah, this seems not to work with ankidroid's old reviewer. the work for new reviewer is almost done, maybe we should wait?

Or, use `nightMode` which is more extensively used in shared templates (and anki manual recommends it).

GithubAnon0000:
I'd also like to point out my comment regarding potentially unmet A11Y: https://forums.ankiweb.net/t/add-nightmode-card-in-styling-css/62824/5?u=anon_0000

brishtibheja:
I believe night mode is always some sort of black, at least in AD, any different for iOS?

GithubAnon0000:
Why is it different anyways? I'd expect my cards to look identical no matter which platform they are currently studied on. It would also make this PR easier. E.g. we could uniform it like this:
```
background-color: #2b2b2b;    /* or maybe a bit darker like #1a1a1a*/
color: rgba(250, 250, 250, 0.87);
```

# Pictures
Option with lighter gray:
anki


Option with darker gray:
anki


For reference, this is what anki desktop currently does (using css variables behind the scenes):
anki
My rgb value above is based on material you recommendations, but I do not feel strongly about it.
<details><summary>Source</summary>
<p>

https://m2.material.io/design/color/dark-theme.html#ui-application

anki


</p>
</details>

BrayanDSO:
I suggest using the CSS variables instead of that. The current code is designed in a way to follow those. And AnkiDroid has more than one dark theme, so this PR would break that.

If you want to solve the problem of beginners understanding CSS, don't make them need to use CSS (#3833). Thinking that using HEX codes is easier than using human readable variables for a non developer is naive.

Or just close this PR and wait for #3833

brishtibheja:
> I suggest using the CSS variables instead of that.

The syntax would be slightly confusing IMO.

> And AnkiDroid has more than one dark theme, so this PR would break that.

Only adding `color` which should be white in both dark and gray themes.

> Thinking that using HEX codes is easier than using human readable variables for a non developer is naive.

Agreed.

Re: cross platform add-on system for templates

Well that might take a decade who knows

BrayanDSO:
> Only adding color which should be white in both dark and gray themes.

As per GithubAnon0000's reasoning, the text color shouldn't white on both themes.

Either use variables or close this.

brishtibheja:
I checked what AnkiDroid currently does and in both of the dark themes, the text color is white (#FFFFFF), which is the same as what this PR does?

GithubAnon0000:
If the dark mode background is always some sort of dark gray or black, then white text would work (might maybe not be ideal, e.g. because of visual strain which affects A11Y, but it would work for most people).

It would still be weird if the background color changes depending on the platform used, like it currently apparently does. **Edit:** It would not be weird, of course, if the user specifically set it up this way (like AnkiDroid apparently allows to do).

Regarding templates, though, I don't really oppose changes much; I use my own template anyways and never get to see the default version (except during testing). So if the impact regarding too much / too little contrast and inconsistency is deemed negligible, then my opinion / comment shouldn't be blocking this.

BrayanDSO:
> , the text color is white (#FFFFFF)

It shouldn't. Up to improvements there.

> It would still be weird if the background color changes depending on the platform used

The apps are different. It would be weird if the background didn't match the app itself.

GithubAnon0000:
> The apps are different. It would be weird if the background didn't match the app itself.

I guess it depends on how much styled the cards are. With the current, default minimal styling, I think you're right.



> It shouldn't. Up to improvements there.

What would you propose?

According to the material you style guides a less contrast-rich white, like `color: rgba(250, 250, 250, 0.87);` would be better and could aid A11Y. But if the background changes for every platform, then #FFFFFF would be a safe bet, as it should always without doubt provide enough contrast with the background.

BrayanDSO:
> What would you propose?

Quoting myself above:

> Either use variables or close this.

Variables aren't that bad. While using the same color everywhere restricts accessibility and theming (in the native side).

dae:
My green light was based on the assumption that AnkiDroid is using #fff for all its dark modes, and wants to continue doing so. If that's not the case, I think we're better off leaving this as-is for now, as variables add an additional level of complexity for new users.

BrayanDSO:
> I think we're better off leaving this as-is for now

I'm okay with that.

PR #4105, opened 2025-06-19
Upgrade to nom 8.0.0

iamllama:
See https://web.archive.org/web/20250128194312/https://unhandledexpression.com/nom-8/

The main improvement with nom 8 is that it now recognises when a parser's output isn't used, and potentially avoids allocation in those cases, which lets us remove the hack introduced in #3886, along with catching a few other missed cases


dae:
That was a lovely surprise - thank you very much!

PR #4106, opened 2025-06-19
Save dr and decay in card even if item is None

user1823:
Functional changes:
- DR and decay are now stored in the card even if item is None because they don't depend on items
- Stored value of decay is now cleared when FSRS is disabled

Non-functional changes:
- Non-functional code from `compute_memory_state` was removed. This function is used by add-ons to fetch information. This function doesn't update the cards on its own.
- Document why we store DR and decay in the card
- In
```rs
let interval = fsrs.next_interval(Some(state.stability), card.desired_retention.unwrap(), 0);
```
`card.desired_retention.unwrap()` was replaced by `desired_retention.unwrap()`.

<details>
<summary>Original PR description</summary>
I think I had a very wrong idea about why desired retention and decay were stored in the card. (this doesn't affect my other PRs or requests though)

@L-M-Sherlock, please confirm that this documentation is correct. Also, please confirm whether the scheduler uses any of these stored values during reviews.
</details>

### DIFF
--- a/rslib/src/scheduler/fsrs/memory_state.rs
+++ b/rslib/src/scheduler/fsrs/memory_state.rs
@@ -105,8 +105,9 @@ impl Collection {
                 progress.update(true, |state| state.current_cards = idx as u32 + 1)?;
                 let mut card = self.storage.get_card(card_id)?.or_not_found(card_id)?;
                 let original = card.clone();
-                // store decay and desired retention in the card so that add-ons, card info and
-                // Stats don't need to access the deck config
+                // store decay and desired retention in the card so that add-ons, card info,
+                // Stats and browser search/sorts don't need to access the deck config.
+                // Values stored in the card are not used by the scheduler.
                 card.desired_retention = desired_retention;
                 card.decay = decay;
                 if let (Some(req), Some(item)) = (&req, item) {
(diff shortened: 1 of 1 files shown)


L-M-Sherlock (approved):
LGTM

user1823:
Hmm... Rust is confusing. Just breaking one if else block into nested if else blocks causes some problem.

Any idea how can I fix this issue?
```
expected `Option<FsrsItemForMemoryState>`, found `&FsrsItemForMemoryState`
```

Edit: Nvm. I did some silly mistake when copy.

rslib/src/scheduler/fsrs/memory_state.rs
```diff
@@ -105,77 +105,92 @@ impl Collection {
...
                                     }
                                 }
                             }
                         }
+                    } else {
+                        // clear memory states if item is None
+                        card.memory_state = None;
                     }
                 } else {
+                    // clear FSRS data if FSRS is disabled
```
user1823, line 190: To reduce code duplication, I think we should make a new function named `clear_fsrs_data_from_card` and use it here and in `set_deck`. But, I am not sure what's the best place to add the new function.

dae:
I'm in a rush to get a new beta out, so deferring to Jarrett's judgement here. Thank you both!

PR #4108, opened 2025-06-20
Fix repeated ticks in reviews graph

user1823:
In #4086, I tried to make the ticks looks cleaner because ticks like 16.67m, 33.33m, 1.11h, 1.39h, etc. look unclean. 

However, this resulted in a regression. In 25.06b4, the graph looks like this. Note the repeated 1h and 2h ticks.

This PR fixes that regression by setting precise=True again. I think we need a better way to control the rounding precision (https://forums.ankiweb.net/t/reduce-precision-of-time-studied-today/42309). But, that's out of scope of this PR.

### DIFF
--- a/ts/routes/graphs/reviews.ts
+++ b/ts/routes/graphs/reviews.ts
@@ -141,7 +141,7 @@ export function renderReviews(
 
     const yTickFormat = (n: number): string => {
         if (showTime) {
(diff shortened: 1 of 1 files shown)


dae:
These things happen. Appreciate the follow-up fix!

PR #4111, opened 2025-06-21
Make PYTHONPYCACHEPREFIX point to an absolute path

abdnh:
I noticed some cached data is now written to qt/out/pycache.


dae:
Thanks, that had been annoying me!

PR #4113, opened 2025-06-21
Fix/Add check_output_timestamps to PythonEnvironment

Luc-Mcgrady:
~~`restat` defaults to 1 (Though the docs don't make it seem that way?) meaning that `.check_output_timestamps()` doesn't do anything.~~

This, for me, fixes the python environment rebuilding every run.

### DIFF
--- a/build/ninja_gen/src/build.rs
+++ b/build/ninja_gen/src/build.rs
@@ -261,9 +261,10 @@ impl BuildStatement<'_> {
             panic!("{} must generate at least one output", action.name());
         }
         stmt.variables.push(("description".into(), group.into()));
-        if action.check_output_timestamps() {
-            stmt.rule_variables.push(("restat".into(), "1".into()));
-        }
+        stmt.rule_variables.push((
+            "restat".into(),
+            (action.check_output_timestamps() as u32).to_string(),
+        ));
         if action.generator() {
             stmt.rule_variables.push(("generator".into(), "1".into()));
         }
--- a/build/ninja_gen/src/build.rs
+++ b/build/ninja_gen/src/build.rs
@@ -261,10 +261,9 @@ impl BuildStatement<'_> {
             panic!("{} must generate at least one output", action.name());
         }
         stmt.variables.push(("description".into(), group.into()));
-        stmt.rule_variables.push((
-            "restat".into(),
-            (action.check_output_timestamps() as u32).to_string(),
-        ));
+        if action.check_output_timestamps() {
+            stmt.rule_variables.push(("restat".into(), "1".into()));
+        }
         if action.generator() {
             stmt.rule_variables.push(("generator".into(), "1".into()));
         }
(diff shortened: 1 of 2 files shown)


iamllama:
I can't seem to reproduce this, n2 (from `./tools/install-n2`) reports `no work to do` after a second `./run`

Luc-Mcgrady:
Here is the difference this pr makes for me:

https://github.com/user-attachments/assets/62b4ed56-5f8e-4f23-9f0b-1c6094c2a354

Is there something to do with `./tools/install-n2` that I'm meant to do to avoid this behaviour?

On main this is the output of  `./ninja -d explain -v`
```
    Finished `release` profile [optimized] target(s) in 0.15s
ninja explain: output /home/luc/Programming/forks/anki/out/pyenv/bin/black older than most recent input /home/luc/Programming/forks/anki/out/rust/release/runner (1750527715838813748 vs 1750546233608145463)
ninja explain: /home/luc/Programming/forks/anki/out/pyenv/bin/protoc-gen-mypy is dirty
ninja explain: /home/luc/Programming/forks/anki/out/pylib/anki/ankidroid_pb2.py is dirt
ninja explain: /home/luc/Programming/forks/anki/out/pylib/anki/ankidroid_pb2.pyi is dirty
...
```

iamllama:
> Is there something to do with ./tools/install-n2 that I'm meant to do to avoid this behaviour?

n2 is intended to be used instead of ninja (which doesn't support the `hide_success` rule var that ninja_gen emits, for example), and is installed via that script

Luc-Mcgrady:
using n2 fixes this for me but It's still worth fixing for ninja users right?

iamllama:
Dug into this a bit (discussed over discord), and from looking at ninja's source it seems to check for the `restat` rule var's presence (not its value) as the docs state
https://github.com/ninja-build/ninja/blob/656412538b6fc102b809a61e0efce422e5a20534/src/graph.cc#L514
https://github.com/ninja-build/ninja/blob/656412538b6fc102b809a61e0efce422e5a20534/src/build.cc#L912

Which means that every statement ends up getting `restat`'ed, which fixes the issue but not in an expected way 🤔

dae:
The build scripts try to support both n2 and ninja. @iamllama I'm not sure I'm following your comment - I thought you meant we were doing restat = 0 (which wouldn't work), but the code appears to be only conditionally writing it:

        if action.check_output_timestamps() {
            stmt.rule_variables.push(("restat".into(), "1".into()));
        }

Could you clarify?

dae:
Thanks Luc!

iamllama:
> I thought you meant we were doing restat = 0 (which wouldn't work), but the code appears to be only conditionally writing it:

@dae I was referring to the pr's original impl [ace2e5e](https://github.com/ankitects/anki/pull/4113/commits/ace2e5ef6a4a4c11bf2b5e77f746e958f3f20319) which would have always defined `restat`

dae:
Ah, gotcha. I was familiar with this issue, as I made the same mistake with assuming 0 = off with n2 in the past. 😅

PR #4116, opened 2025-06-22
Feat/Cmrr target selector

Luc-Mcgrady:
Allows you to select your preferred target for CMRR.

Regular with customisable loss aversion:
image

S*R:
image

Retention on a after day you stop reviewing at the end of the simulation:
image

Average retention of the days between when you stopped reviewing and a certain date:
image

ref: https://forums.ankiweb.net/t/anki-25-06-beta/62271/39?u=a_blokee

### DIFF
--- a/rslib/src/scheduler/fsrs/retention.rs
+++ b/rslib/src/scheduler/fsrs/retention.rs
@@ -25,7 +25,7 @@ impl Collection {
             };
         }
 
-        let target = req
+        let target_type = req
             .target
             .map(TryInto::try_into)
             .transpose()
@@ -33,8 +33,9 @@ impl Collection {
 
         let days_to_simulate = req.days_to_simulate as f32;
 
-        let target = match target {
+        let target = match target_type {
             Some(CmrrTarget::Memorized) => None,
+            Some(CmrrTarget::LossAversion) => None,
             Some(CmrrTarget::Stability) => {
                 wrap!(move |SimulationResult {
                                 cards,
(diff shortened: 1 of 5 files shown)


Expertium:
What's the difference between 1 and 3? Memorized vs Post Abandon Memorized. Is it just R(delta_t) vs R(delta_t+365)?

Note for Dae: this is for the beta only, the final release won't have any extra fields/options

Luc-Mcgrady:
I suppose memorised is just post abandon memorised with 0 days and with no loss aversion. I don't like the "Post Abandon" wording but its the best I could come up with.

> Note for Dae: this is for the beta only, the final release won't have any extra fields/options

It wont?

Expertium:
We don't want to overcomplicate things for users any further. My plan was to let beta-testers decide on the best way of defining "knowledge". Btw, I wasn't planning to make loss aversion customizable, but sure, why not.

rslib/src/scheduler/fsrs/retention.rs
```diff
@@ -1,7 +1,9 @@
+use anki_proto::scheduler::simulate_fsrs_review_request::cmrr_target::Kind;
```
Luc-Mcgrady: Should this break the check? (It's above the copyright notice)

iamllama: Looks like the added leeway was intended for cases like this?
https://github.com/ankitects/anki/blob/b250a2f7245cb41f10581bfca2f89106fc0e366c/qt/runanki.py#L1-L3

GithubAnon0000:
For reference, in case my future self comes back to this and is looking for the data...

A table with the results of those targets (from my own deck) can be found in the forums: https://forums.ankiweb.net/t/anki-25-06-beta/62271/108?u=anon_0000

<hr>

Also, @Expertium wrote in the forums (see link above):
> The idea is to ask beta-testers to try different options and see which one produces good results, “good” being subjective here.
> And if none of these options are good, we’ll have to remove CMRR.

What would you consider as being good? E.g. even if the CMRR targets would give me a MRR of .88, I wouldn't really know what to do with it. I'm not sure if I should consider this value at all, considering I'm aiming for .95 retention and don't really mind having a slighter higher workload than necessary.

Expertium:
> What would you consider as being good?

As I said, it's subjective. Basically, anything that doesn't make you think:
1) Why is CMRR always at 70%?
2) Why is CMRR always above 90%?
3) Why is CMRR always the same?

And makes you think "yeah, I guess that makes sense"

dae:
Thanks Luc!

PR #4122, opened 2025-06-23
FIX: Margin between icon was selectable but shouldn't be

GithubAnon0000:
Fixes: https://forums.ankiweb.net/t/bug-field-names-in-the-editor-are-selectable/63047?u=anon_0000

The space now cannot be selected anymore using the mouse or <kbd>Ctrl+A</kbd>. Collapsing fields still works fine.


ts/editor/CollapseLabel.svelte
```diff
@@ -34,5 +34,6 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 <style lang="scss">
     .collapse-label {
         cursor: pointer;
+        user-select: none;
```
GithubAnon0000, line 37:
```suggestion

```
Remove this since it doesn't have much use when text is selectable.

Making text selectable was requested in https://forums.ankiweb.net/t/bug-field-names-in-the-editor-are-selectable/63047/8?u=anon_0000

abdnh:
I like selectable text, but no strong preference here. I never had to copy any text in the UI.

Some parts of the editor can still be selected in this PR:
image

dae:
@abdnh's call here - editor-related changes at this point may make more work for him.

GithubAnon0000:
abdnh, if you're fine with it, I'd like to merge my suggestion above. This will make the text selectable.

ts/editor/CollapseLabel.svelte
GithubAnon0000: @abdnh

The PR you merged in doesn't allow for text selection. The suggestion above (removing line 37) would allow text selection.

Do I understand this correctly that you decided against text selection in this case? Or can I merge the above suggestion?

(If I can merge it, I'd have to open a new PR since GH doesn't allow me to commit it now that it's merged to main)

abdnh: Feel free to open a PR.

PR #4125, opened 2025-06-25
CHANGE right-click in the editor to show option to open folder in linux

GithubAnon0000:
Ref: https://github.com/ankitects/anki/pull/4029#issuecomment-3005226875 and comments below (start with the heading "Answering user1823" in the linked comment).

# Changes
1. With this PR the option "Show in folder" (found when right clicking on an image in the editor screen) will show for linux as well. Before, it was hidden away.
2. This PR uses `xdg-open` to open the file, which should work regardless of the linux distro and environment used.

> [!NOTE]
> #4029 modifies quite a few files reagrding the editor. Once #4029 is merged, the changes to editor.py done in this PR are irrelevant. The change in utils.py does stay relevant, though, because that code is still being used.

### DIFF
--- a/qt/aqt/utils.py
+++ b/qt/aqt/utils.py
@@ -938,7 +938,7 @@ def show_in_folder(path: str) -> None:
     else:
         # For linux, there are multiple file managers. Let's test if one of the
         # most common file managers is found and use it in case it is installed.
-        # If none of this list are installed, fallback to xdg-open. xdg-open
+        # If none of this list are installed, use a fallback. The fallback
         # might open the image in a web browser, image viewer or others,
         # depending on the users defaults.
         file_managers = [
@@ -952,8 +952,7 @@ def show_in_folder(path: str) -> None:
 
         available_file_manager = None
 
-        # Test if a file manager is installed. Fallback to xdg-open if none are
-        # found.
+        # Test if a file manager is installed and use it, fallback otherwise
         for file_manager in file_managers:
             if shutil.which(file_manager):
                 available_file_manager = file_manager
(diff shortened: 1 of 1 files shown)


iamllama:
> This PR uses xdg-open to open the file

On ubuntu 24 this just opens the file, which "Open image" already does

GithubAnon0000:
Maybe the defaults are different on your machine? I tried it on my debian oldstable and it also just opens the file.
`xdg-mime query default image/png` returns the gnome image viewer on my oldstable machine and gnome file manager on sid.

Another way to solve this would be to detect the desktop environment and try to open in the file manager, which is most commonly used by that desktop manager, like nautilus on gnome.

iamllama:
If it does the same thing as "Open image" it should ideally be hidden

GithubAnon0000:
I was trying to say that we could probably make it work like it does on other OSes, by detecting the DE used. I don't personally use this feature though and can close this PR if a "heuristic" is less prefered compared to hidding the feature.

GithubAnon0000:
(I'll try to create a heuristic version in the comming days, that works appropriately)

GithubAnon0000:
# Update
This code now searches for the most common file managers on linux and uses the first one it finds. If none are found, it uses `xdg-open`.

# Tests
I could only test it with debian, gnome. But commenting out installed file managers (e.g. `nautilus`) did act as expected.

# Comment
If there's a way to make the tests pass without having to format lines 945 – 950 like `ninja` did, please let me know. I initally had the comments directly beneath each other, making it more readable.

This is ready for review.

user1823:
> If none are found, it uses xdg-open.

What if xdg-open is not installed? Running `xdg-open --help` in the Terminal on a brand new Ubuntu (WSL) system gives me:
```bash
Command 'xdg-open' not found, but can be installed with:
sudo apt install xdg-utils
```

GithubAnon0000:
Thanks for testing! I didn't know that it's not a default app. I changed it so that the old (from current main) fallback is used, which shouldn't depend on a specific package. It works as expected for me on debian, sid.

**Edit:** Apparently it normally is installed by default on all major linux distros. But if you use a custom or minimal install (I guess WSL falls into that category), then it may not be available. The code from main (which I added back in c7b7721) should work regardless of the install method.

GithubAnon0000:
Pinging @abdnh, since this one affects your PR.

If the changes here are merged in, you no longer need to hide the option in the ts implementation.

PR #4130, opened 2025-06-28
FIX dropdown buttons having borders

GithubAnon0000:
Fixes: https://forums.ankiweb.net/t/anki-25-06-beta/62271/129?u=anon_0000

This might have been introduced by #4103.

# Changes
The file `base.scss` imported `_button-mixins.scss`. It didn't specify to remove the border and hover effect though. This PR fixes that.

### DIFF
--- a/ts/lib/sass/_button-mixins.scss
+++ b/ts/lib/sass/_button-mixins.scss
@@ -105,7 +105,12 @@
             cursor: not-allowed;
             color: var(--fg-disabled);
             box-shadow: none !important;
-            background-color: var(--button-gradient-end);
+            @if $with-dropdown {
+                background-color: transparent;
+            }
+            @else {
+                background-color: var(--button-gradient-end);
+            }
             border-bottom-color: var(--border-subtle);
         }
     }


GithubAnon0000:
Light Mode:
anki

Dark mode still has this issue. I'll un-draft once the issue is fixed.

GithubAnon0000:
The issue in dark mode was that `--button-bg` was applied, which is much lighter than the `canvas` variable the parent uses. I thus added a variable to check whether we're in a dropdown right now, and disable the background if true.

Maybe there's a better way to do this?

Result:
anki

GithubAnon0000:
The last commit fixes the disabled button state for dark mode. It now looks like this:

anki

chrislongros:
Works perfect!

user1823:
Since there were no issues before #4103, what if we just delete the files that were unused before?

abdnh:
This removes borders from add-on buttons.
image

> Since there were no issues before https://github.com/ankitects/anki/pull/4103, what if we just delete the files that were unused before?

I agree that's probably the best solution.

dae:
Given the regressions, I'll revert #4103 for the next beta, and we can clean up the files afterwards.

GithubAnon0000:
This PR is obsolete now and can be closed, correct?

abdnh:
> This PR is obsolete now and can be closed, correct?

Right.

PR #4131, opened 2025-06-28
Add disclaimer to workload tooltip

Luc-Mcgrady:
c.c. @Expertium 

Mildly off topic and not very important but `deck-config-desired-retention-tooltip` needs to be updated for `en-gb`
https://i18n.ankiweb.net/en-GB/core/core/templates/deck-config.ftl/?search=deck-config-desired-retention-tooltip&string=25342


sommerluk:
I'm confused. Where do we have a tooltip in the deck options that shows workload values? Isn't this string itself the tooltip? Or is this about the orange or blue boxes that pop up below the "desired retention" spinbox?

Luc-Mcgrady:
Its about the boxes. I wasn't sure of the best name for it.

PR #4139, opened 2025-06-30
Only run "empty_filtered_deck" on filtered decks

### LINKED ISSUE #4138: Pressing "E" on overview for a non-filtered deck marks every card as modified
Steps to reproduce:

* Launch Anki
* Click on the name of a regular (non-filtered) deck that contains some cards.
* Press the "E" key (shortcut for "empty filtered deck").

Expected results: Nothing happens.

Actual results: Every card in the deck has its modification date updated.  On the next sync, every card in the deck will need to be re-synced.

mbrubeck:
Fixes #4138.

### DIFF
--- a/rslib/src/decks/remove.rs
+++ b/rslib/src/decks/remove.rs
@@ -28,7 +28,7 @@ impl Collection {
         let card_count = match deck.kind {
             DeckKind::Normal(_) => self.delete_all_cards_in_normal_deck(deck.id)?,
             DeckKind::Filtered(_) => {
-                self.return_all_cards_in_filtered_deck(deck.id)?;
+                self.return_all_cards_in_filtered_deck(deck)?;
                 0
             }
         };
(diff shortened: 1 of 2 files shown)


mbrubeck:
Alternately, we could change the Rust backend to return `Err(FilteredDeckRequired)` when `empty_filtered_deck` is called on a non-filtered deck.  This would be slightly more intrusive, but it would match the behavior of `rebuild_filtered_deck`, and it might prevent this problem from recurring in other places.

mbrubeck:
> Alternately, we could change the Rust backend to return Err(FilteredDeckRequired) when empty_filtered_deck is called on a non-filtered deck.

This turned out to be pretty simple, so I updated the PR to use this approach instead.

If you prefer the original approach (just ignore the keypress in Python for non-filtered decks), I can revert to that version.

dae:
Doing it in Rust is the better approach, as it fixes the mobile clients too. Thank you for the fix!

PR #4145, opened 2025-07-01
Add search keyword to strip clozes beforehand

iamllama:
Ref: https://forums.ankiweb.net/t/make-browser-search-find-de-clozed-text/58348

This pr proposes to add a new search keyword `sc:...` such that `"sc:ive memorised it"` matches against cloze notes with `ive {{c1::memorised}} it` and `sc:aacdeilnoorstuyz` matches against cloze notes with `äčďéíľ{{c3::ňóôŕ}}šťúýž` (with `ignore accents` on) for example

On a col with ~50000 cloze notes and with bench code lifted from https://github.com/ankitects/anki/pull/2394#issuecomment-1455064835:
```
find_notes_with_sc          time:   [401.07 ms 401.40 ms 401.75 ms]
find_notes_with_ignore      time:   [758.06 ms 759.58 ms 761.29 ms]
find_notes_sc_with_ignore   time:   [1.1066 s 1.1071 s 1.1076 s]
```

> I’d recommend you start with a benchmark, as it may be slow compared to the accent search. If so, a similar approach to [consider extending sort field to all fields and/or changing HTML stripping · Issue #1085 · ankitects/anki · GitHub](https://github.com/ankitects/anki/issues/1085) may be required.

I've tried this (by updating the sfld in [Note::prepare_for_update](https://github.com/ankitects/anki/blob/7720c7de1aabb2cbcfb137df7d486e12eea28601/rslib/src/notes/mod.rs#L208)) and it works, but then it only gets updated when the note gets modified, and won't show up in the results until then. I don't know if addons rely on its existing behaviour either


dae (commented):
Thanks for the benchmark; that seems like it's reasonable compared to what we already have. Just one minor suggestion:

rslib/src/search/sqlwriter.rs
```diff
@@ -202,17 +215,34 @@ impl SqlWriter<'_> {
         self.args.push(text);
         let arg_idx = self.args.len();
 
-        let sfld_expr = if no_combining {
-            "coalesce(without_combining(cast(n.sfld as text)), n.sfld)"
+        let process_text_opt = (no_combining as u8) | ((strip_clozes as u8) << 1);
```
dae: How about we pull in https://docs.rs/bitflags/latest/bitflags/ and use that instead of magic numbers?

dae:
Thanks, that's easier to follow :-) 

> I've tried this (by updating the sfld in [Note::prepare_for_update](https://github.com/ankitects/anki/blob/7720c7de1aabb2cbcfb137df7d486e12eea28601/rslib/src/notes/mod.rs#L208)) and it works, but then it only gets updated when the note gets modified, and won't show up in the results until then.

We'd need to get the user to run 'check db' or force a recalculation on upgrade, which could cause conflicts if we bump the mtime. Thankfully your on-the-fly solution is fast enough for now anyway!

PR #4146, opened 2025-07-01
Add "Copy Debug Info" to `show_exception`'s dialog

iamllama:
Ref: https://forums.ankiweb.net/t/anki-25-06-beta/62271/94

`show_exception` is called to show an error dialog when an exception is raised in a `CollectionOp` or `QueryOp`, but there's no way to get at the debug info without a terminal, which this pr proposes to rectify

Before:
image
After:
image

The addon error's dialog's `Troubleshooting` button has also been changed to `Help`:
image


dae:
Much better - thank you! ❤️

PR #4147, opened 2025-07-01
Use system font for webviews instead of bootstrap font stack

GithubAnon0000:
Fixes: 
1. https://forums.ankiweb.net/t/inconsistent-chinese-japanese-ui-font-when-anki-is-in-a-different-language-than-the-system-language-on-windows/63261?u=anon_0000
2. https://forums.ankiweb.net/t/make-space-width-space-usage-in-front-of-units-consistent/53798?u=anon_0000

# Issue
Bootstrap applied a default style to the webviews. Due to that, the font that the user had set in their OS was ignored. This PR fixed that.

# Images
You can find more images in the second link above.

With this PR, the system font is applied correctly. I am using a weird font as a system font to make it very obvious.

anki


iamllama:
How does removing bootstrap's [native font stack](https://getbootstrap.com/docs/5.3/content/reboot/#native-font-stack) change the way pages look on mobile clients?

There doesn't seem to be consensus on whether `system-ui` should be removed or not either: https://github.com/twbs/bootstrap/pull/22377

GithubAnon0000:
> How does removing bootstrap's [native font stack](https://getbootstrap.com/docs/5.3/content/reboot/#native-font-stack) change the way pages look on mobile clients?

Good question. I do not know how to test that though (I'm assuming I'd need to install Android studio and compile the .apk, which is something I cannot do). In theory, though, it should take whatever font the mobile OS is using.

GithubAnon0000:
> There doesn't seem to be consensus on whether `system-ui` should be removed or not either: [twbs/bootstrap#22377](https://github.com/twbs/bootstrap/pull/22377)

I can try to use "system-ui" instead of "inherit" this evening and check whether everything still works as expected on my debian machine.

GithubAnon0000:
> I can try to use "system-ui" instead of "inherit"

As promised, I just tried that. The webview does not use the font that I selected, if `system-ui` is used. With `inherit`, it works fine as demonstrated above.

dae:
AnkiMobile already injects -apple-system into the webpage, so it should be unaffected, and AnkiDroid should be able to do similar if it needs to. Interesting how system-ui is still being debated, and it still seems to cause problems for Asian languages. Happy to give this a try to see how it works out.

PR #4149, opened 2025-07-02
Feat/Add legacy evaluate config bool

Luc-Mcgrady:
Allows advanced users to toggle the old evaluate functionality by running

```py
from anki.config import Config
mw.col.set_config_bool(Config.Bool.DECK_CONFIG_FSRS_LEGACY_EVALUATE, True)
```

in the console

image


rslib/src/backend/config.rs
```diff
@@ -39,6 +39,7 @@ impl From<BoolKeyProto> for BoolKey {
             BoolKeyProto::RenderLatex => BoolKey::RenderLatex,
             BoolKeyProto::LoadBalancerEnabled => BoolKey::LoadBalancerEnabled,
             BoolKeyProto::FsrsShortTermWithStepsEnabled => BoolKey::FsrsShortTermWithStepsEnabled,
+            BoolKeyProto::DeckConfigFsrsLegacyEvaluate => BoolKey::FsrsLegacyEvaluate,
```
Luc-Mcgrady: Should we keep this name consistent between proto and rust? I'm split between the explicitness of "DeckConfig" and the consistency between the FSRS related bool keys starting with `Fsrs` in the rust version e.g:

https://github.com/ankitects/anki/blob/da907053460e2b78c31199f97bbea3cf3600f0c2/rslib/src/config/bool.rs#L42-L43

Just `FsrsLegacyEvaluate` for both? Just splitting hairs.

qt/aqt/mediasrv.py
```diff
@@ -652,6 +652,7 @@ def handle_on_main() -> None:
     "compute_optimal_retention",
     "set_wants_abort",
     "evaluate_params",
```
Luc-Mcgrady:
```suggestion
```
?

rslib/src/backend/config.rs
dae: Let's keep it consistent with the existing entries - pushed an update.

dae:
Thanks Luc!

JSchoreels:
Thank you a lot ！

sunrucong:
Hi there, I was redirected here when asking about how to view RMSE and/or log loss in the new update but I am unsure how to make sense of the discussion here. What is the console where I would run that code mentioned at the top?

from anki.config import Config
mw.col.set_config_bool(Config.Bool.FSRS_LEGACY_EVALUATE, True)

Luc-Mcgrady:
https://docs.ankiweb.net/misc.html#debug-console

PR #4150, opened 2025-07-02
Fix unique blank decks being created per row with empty deck column when importing csv

iamllama:
Fixes the bug where if `#deck column:...` is used, each row with an empty field in the deck column would create its own new blank deck. After this pr, at most one new blank deck is created and used per import

image


dae:
Seems like an improvement (thanks!), but I'm left wondering whether this is a case we even want to support - would it make sense to have a missing deck name be an error?

iamllama:
Seeing as it might be easier to recover from within anki (by renaming the new blank deck) than outside (by fixing/editing the csv file, especially on mobile), perhaps a warning in the import log would be better than a hard error?

PR #4154, opened 2025-07-04
Allow creating deck via `#deck:...` if non-existent when importing

iamllama:
This pr proposes to allow creating the deck specified by `#deck:...` if it didn't already exist when importing csv files, which brings it in line with `#deck column:...`'s behaviour

image

ref: https://forums.ankiweb.net/t/feature-request-creating-new-decks-from-import-text-file-screen/62288/14

### DIFF
--- a/proto/anki/import_export.proto
+++ b/proto/anki/import_export.proto
@@ -176,29 +176,32 @@ message CsvMetadata {
   // to determine the number of columns.
   repeated string column_labels = 5;
   oneof deck {
+    // id of an existing deck
     int64 deck_id = 6;
     // One-based. 0 means n/a.
     uint32 deck_column = 7;
+    // name of new deck to be created
+    string deck_name = 8;
   }
   oneof notetype {
     // One notetype for all rows with given column mapping.
-    MappedNotetype global_notetype = 8;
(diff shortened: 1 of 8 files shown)


proto/anki/import_export.proto
```diff
@@ -176,29 +176,32 @@ message CsvMetadata {
...
-  uint32 guid_column = 14;
-  DupeResolution dupe_resolution = 15;
-  MatchScope match_scope = 16;
+  uint32 tags_column = 11;
+  bool force_delimiter = 12;
+  bool force_is_html = 13;
+  repeated generic.StringList preview = 14;
+  uint32 guid_column = 15;
+  DupeResolution dupe_resolution = 16;
+  MatchScope match_scope = 17;
```
dae: I'd recommend just using `<next biggest number>` for deck_name, so that the other fields don't need to be renumbered. Keeps the git blame tidy, and avoids a practice that would cause breakages if it was done on certain messages (think Notetypes that are persisted in the DB for example).

dae (commented):
Thank you as always! I'll circle back to this once things die down a bit with the new release. Just one minor point for now based on a quick skim:

dae:
Thank you!

PR #4156, opened 2025-07-04
Fix launcher glibc version check

iamllama:
Tried building the launcher, and got back `Error: Anki requires a modern Linux distro with glibc 2.36 or later.` even though my system's is 2.39, which this pr fixes


dae:
Thanks for the quick catch! I only just beat you to it: https://github.com/ankitects/anki/commit/cedece5cae0c3d3ba1347b31192b3b266e64edbf

PR #4159, opened 2025-07-04
Revert "Use system font for webviews instead of bootstrap font stack"

dae:
Reverts ankitects/anki#4147

This caused a regression in the appearance of editor fields - we'll need to address that if we re-apply this.

### DIFF
--- a/ts/lib/sass/base.scss
+++ b/ts/lib/sass/base.scss
@@ -44,7 +44,6 @@ html {
 }
 
 body {
-    font-family: inherit;
     overflow-x: hidden;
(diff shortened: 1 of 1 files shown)


GithubAnon0000:
Not sure why this happens. I could add the bs font stack back for editor-base.scss and reapply my fix for the other screens (which was reverted here).

It only seems to be the editor webview that's acting up (and in the templates editor, the source code still uses the wrong font. But my PR didn't affect that window anyways).

Is there a way to find out from what the note editor is inheriting the wrong font?

Issue #4158, opened 2025-07-04
25.07: Wrong font in Note Editor


BrayanDSO:
Wrong font used in the note editor:

<img width="506" height="1044" alt="Image" src="https://github.com/user-attachments/assets/eabb4641-9631-480c-970a-e01f609103da" />

Works alright in 25.02, even with the new installer

Debug info:
```
Anki 25.07 (2594dcb2) (ao)
Python 3.13.5 Qt 6.8.1 PyQt 6.8.0
Platform: Windows-11-10.0.26100-SP0

===Add-ons (active)===
(add-on provided name [Add-on folder, installed at, version, is config changed])
'' ['anki-launcher', 0, 'None', '']

===IDs of active AnkiWeb add-ons===


===Add-ons (inactive)===
(add-on provided name [Add-on folder, installed at, version, is config changed])
AnkiWebView Inspector ['31746032', 2023-06-27T16:26, 'None', '']
Edit Field During Review Cloze ['385888438', 2025-04-20T10:34, '6.22', mod]
Enhanced Cloze for Anki 21 ['1990296174', 2023-09-10T15:48, 'None', '']
Hint Hotkeys ['1844908621', 2023-04-14T11:13, 'None', '']
extended editor for field for tables searchreplace ['805891399', 2023-07-31T11:12, 'None', mod]
```
Also, is this supposed to use my system's python version?

BrayanDSO:
And I'm not the only one with the problem

<img width="579" height="55" alt="Image" src="https://github.com/user-attachments/assets/751f3da5-b455-4ad5-b2f5-7b579033d77d" />

dae:
Closed via https://github.com/ankitects/anki/pull/4159

BrayanDSO:
> Also, is this supposed to use my system's python version?
> 
> And what is the anki-launcher add-on? The View Add-on Page button for it is disabled. I had to check the code because I never installed that, never knew it was supposed to be installed, and instantly thought it could be something malicious.

@dae, could you answer that?

GithubAnon0000:
> And what is the anki-launcher add-on? The View Add-on Page button for it is disabled.

Weirdly, on my system I don't have this addon (I cannot see it in the beta or in the latest build from main).

You should be able to view the code in the `addons21/` folder though.

dae:
[...]

PR #4163, opened 2025-07-04
Use system font for webviews instead of bootstrap font stack and add exception for note editor

GithubAnon0000:
Reverts ankitects/anki#4159
Fixes: #4158

My previous fix works for every view I tested, except for the note editor, which inherits Noto Sans on my machine (or the bs font stack, if all `font-family` are removed from the code).

This PR adds back the fix to the webviews that work as expected, while adding an exception for the note editor. I couldn't figure out how to solve this, which is why I plan to open a new issue onces this PR is merged.

### DIFF
--- a/ts/editor/editor-base.scss
+++ b/ts/editor/editor-base.scss
@@ -10,5 +10,6 @@ $btn-disabled-opacity: 0.4;
 
 html,
 body {
+    font-family: var(--bs-font-sans-serif);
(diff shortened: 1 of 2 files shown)


dae:
I'd like to hold off on further changes here until the stable release has a chance to settle a bit - ideally we'll have another beta build soon that we can test this in.

PR #4162, opened 2025-07-04
Launcher: Run `uv python install` before running `uv sync`

Grinkers:
A potential fix for https://forums.ankiweb.net/t/bug-anki-25-07-fails-to-launch/63475/7.

I'm not exactly sure how this wasn't an issue before. It may be a system specific thing (I'm on Fedora 42 and had uv installed). I would be curious to know if `.local/share/AnkiProgramFiles/python` got properly populated for others before this.

However I was able to reproduce the issue even with trying to clean everything (`uv cache clean` and removing my uv binaries). It happened on both `uv 0.7.13` (bundled in the launcher release) and `uv 0.7.19 (e8bc3950e 2025-07-03)` (git).

"Minimal" repro
Try running `./anki` and have
```
Updating Anki...

error: No interpreter found for Python 3.13.5 in managed installations
Install failed: Failed with code Some(2): /home/grinkers/tools/anki-launcher/uv.amd64 sync --upgrade --managed-python --python 3.13.5
```

then
```
cd "/home/grinkers/.local/share/AnkiProgramFiles" && UV_PYTHON_INSTALL_DIR="/home/grinkers/.local/share/AnkiProgramFiles/python" "/home/grinkers/opt/anki-launcher/uv.amd64" "sync" "--upgrade" "--managed-python" "--python" "3.13.5"
error: No interpreter found for Python 3.13.5 in managed installations
```

PR #4166, opened 2025-07-04
Statically link MSVC runtime, removing the need to install the redistributable

Grinkers:
Fixes https://forums.ankiweb.net/t/windows-error-vcruntime140-dll-not-found-when-installing-with-the-launcher-possible-fix/63508.

before and after for the launcher exe.
20250705_05h50m26s_grim

Anki running on LTSC19 with nothing installed and no updates.
20250705_06h18m03s_grim

I don't have a convenient development setup with ninja, so didn't test full packaging; only launcher and anki-console from cargo. Binary size will increase by about 100kB or so.


Grinkers:
Sorry for the force spam. My email got mangled in https://github.com/ankitects/anki/pull/4162 by co-authorship. So my original commit passed, then a github domain ended in the authorship due to co-authorship.

abdnh (approved):
Working nicely on Windows Sandbox.

dae:
Thank you for this!

PR #4171, opened 2025-07-06
Fix/Help modal appears behind simulator modal

Luc-Mcgrady:
image

I think this comes with the caveat of making the background of the help modal everywhere else 75% opaque (as there will be 2 layers), but its quite un-noticeable.

### DIFF
--- a/ts/lib/components/HelpModal.svelte
+++ b/ts/lib/components/HelpModal.svelte
@@ -181,6 +181,11 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         margin-bottom: 1.5rem;
     }
 
+    .modal {
+        z-index: 1066;
+        background-color: rgba($color: black, $alpha: 0.5);
+    }
+
     .modal-title {
         margin-inline-end: 0.75rem;
     }
--- a/ts/routes/deck-options/SimulatorModal.svelte
+++ b/ts/routes/deck-options/SimulatorModal.svelte
@@ -296,7 +296,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                     min={0}
                     max={9999}
                 >
-                    <SettingTitle on:click={() => openHelpModal("simulateFsrsReview")}>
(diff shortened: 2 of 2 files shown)


iamllama:
Possible alternative to avoid setting z-index, but the backdrop gets left behind
```diff
diff --git a/ts/lib/components/TitledContainer.svelte b/ts/lib/components/TitledContainer.svelte
index 70e4a078c..a2f116427 100644
--- a/ts/lib/components/TitledContainer.svelte
+++ b/ts/lib/components/TitledContainer.svelte
@@ -41,11 +41,11 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                 {title}
             </h1>
         {/if}
-        <div class="help-badge position-absolute" class:rtl>
-            <slot name="tooltip" />
-        </div>
     </div>
     <slot />
+    <div class="help-badge position-absolute" class:rtl>
+        <slot name="tooltip" />
+    </div>
 </div>
 
 <style lang="scss">

```

iamllama:
The opened modal doesn't correspond to the simulator options, e.g. clicking on any simulator option opens up the modal at the last-opened section, so perhaps that should be changed here as well if its to be visible

Luc-Mcgrady:
> The opened modal doesn't correspond to the simulator options, e.g. clicking on any simulator option opens up the modal at the last-opened section, so perhaps that should be changed here as well if its to be visible

It seems like to get this to work properly we'd need to do some crazy stuff with the "settings" because the settings here are all it has access to. 

https://github.com/ankitects/anki/blob/42321857358247cea1fd9611c58262ed2c6eb54f/ts/routes/deck-options/FsrsOptionsOuter.svelte#L33-L78

so I'll just leave it as is for now. Maybe it would just be best to give up on the help modal for the simulator modal entirely?

Luc-Mcgrady:
An alternative could be creating a help modal solely for the simulator menu that copies the entries from the respective already existing modals?

dae:
While this is still being discussed, I'll merge in what we currently have, as it's still an improvement over what we had previously.

PR #4175, opened 2025-07-06
[A11Y] FIX: border ratio too low

GithubAnon0000:
Fixes: Planned To-Do number 3 from https://forums.ankiweb.net/t/reminder-for-myself-improve-a11y-in-the-deck-options/63506?u=anon_0000#p-174852-planned-todos-3

# Issue
The default border color `#c4c4c4` on white background fails WCAG AA with a ratio of 1.74:1.

# Proposed fix
1. Use `#858585` instead, which passes with a ratio of 3.69:1. This is achieved by editing _vars.scss and using the darkest lightgray from _color-palette.scss.
2. Let `.form-check-input` use `--border` instead of the bs border.
3. Don't use `--border-subtle` for `.select-container`, use `--border` instead.

# Images
With this PR:
anki

Without this PR (current main):
anki

### DIFF
--- a/ts/lib/components/SpinBox.svelte
+++ b/ts/lib/components/SpinBox.svelte
@@ -210,7 +210,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     .spin-box {
         width: 100%;
         background: var(--canvas-inset);
-        border: 1px solid var(--border);
+        border: 1px solid var(--border-ui);
         border-radius: var(--border-radius);
         overflow: hidden;
         position: relative;


brishtibheja:
are the borders that important? to me it looks less aesthetically appealing. perhaps we should focus on actual text and their contrast (look at preset/this deck/today for example, they seem to have contrast issues).

GithubAnon0000:
> are the borders that important?

Yes, they are. It's important that ui elements can be identified easily by everyone, including those that belong to the target group of A11Y changes.



> to me it looks less aesthetically appealing.

I think it's alright, though I'm open for suggestions (as long as there's proper contrast, spacing, ... for A11Y).



> perhaps we should focus on actual text and their contrast (look at preset/this deck/today for example, they seem to have contrast issues).

Yes, that can be improved as well. For A11Y, UI elements actually need a lower contrast compared to text (and the contrast I choose here is very close to the minimum). We should focus on both, though.

iamllama:
> Yes, they are. It's important that ui elements can be identified easily by everyone, including those that belong to the target group of A11Y changes.

Here it's used as a boundary between a huge section title and the relatively tiny entries, so there's already clearly visible contrast. Are there any places where it's a lot more subtle and could stand to benefit from a more visible border?

dae:
I too find the changed version a bit less aesthetically-pleasing, and wonder how much of a practical difference it will make for a11y.

GithubAnon0000:
> and wonder how much of a practical difference it will make for a11y.

It probably makes a big difference. This PR uses the minimum contrast for ui elements (see https://webaim.org/resources/contrastchecker/). The current contrast of 1.74:1 (main) is way too low for users with impaired vision.

<hr>

To those saying you dislike the way it looks with this PR: What exactly is the issue for you? Is it certain ui elements, the border used below the title, something else?

dae:
> It probably makes a big difference.

a11y checking tools provide useful guidelines, but I don't think we should be treating them as gospel. You're proposing a change that multiple people have said looks worse than before, so I think we need to at least hear from someone with poor vision who's in favour of this change. I tried taking off my glasses, but didn't find one more legible than the other, and was more bothered by the blurry text on both.

My earlier comment was focused on the change in color of the HRs. While I find the increased contrast around fields to be a bit harsh as well, I'm less bothered by it than the hr change.

GithubAnon0000:
> My earlier comment was focused on the change in color of the HRs. While I find the increased contrast around fields to be a bit harsh as well, I'm less bothered by it than the hr change.

Thanks for replying. With the latest commit, it looks like this:
anki


Is that better? I agree btw that this "hr" below the title doesn't have to meet minimal contrast. It's not something the user has to interact with, unlike the fields and switches.

dae:
I find it a mild regression in aesthetics, but could tolerate it if I knew it was making life easier for some people. Anyone else want to weigh in?

sommerluk:
That's subjective, but personally I like the new proposed design, especially because it makes UI elements as buttons and combo boxes easier to spot and recognize.

GithubAnon0000:
Just as a heads up: This PR caused a regression and needs a few changes. I wanted to update the PR and post a new image where the borders are less contrast-rich but still fullfill the minimal requirements. I'm very busy currently though, which is why this PR is on hold for now.

GithubAnon0000:
# Changes in recent commits
1. I reverted the changes I did earlier.
2. Changed interface borders to `#949494`.¹
3. Added interface border to a few widgets.

# Todo
1. Add changes to all widgets.
2. Check / update nightmode.
3. Get approval.²


<hr>

¹    User interfaces should have a minimum contrast of 3:1 according to [WebAIM](https://webaim.org/resources/contrastchecker/). Since the background is `#FFFFFF` and the chosen border is `#949494`, a ratio of 3.03:1 is reached.
²    As an alternative to changing this for all people: maybe a high-contrast toggle like in #4268 would be better?

GithubAnon0000:
I remember this had unintended side effects which I wanted to solve. But I totally lost track of what exactly that was. I want to come back to this, as I think it's quite important to fix the low contrast issue. But I also want to make it look good and work without side effects. I think it's better if I close this current PR and open a new one once I created something better.

PR #4176, opened 2025-07-06
Reorder CardInfo to show Retention Chart before a potentially long Review Log

Grinkers:
Before, even with this huge window, the useful Retention Chart ends up out of view if the card has a long review history. Some of my cards are over a decade of history. It feels silly having to scroll through potentially decades of logs over and over to see the fixed size chart.

Before:
20250707_03h14m40s_grim

After:
20250707_03h11m24s_grim

### DIFF
--- a/ts/routes/card-info/CardInfo.svelte
+++ b/ts/routes/card-info/CardInfo.svelte
@@ -37,14 +37,14 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
             <CardStats {stats} />
         </Row>
 
-        {#if showRevlog}
-            <Row>
-                <Revlog revlog={stats.revlog} {fsrsEnabled} />
-            </Row>
-        {/if}
         {#if fsrsEnabled && showCurve}
             <Row>
                 <ForgettingCurve revlog={stats.revlog} {desiredRetention} {decay} />
+            </Row>
+        {/if}
+        {#if showRevlog}
+            <Row>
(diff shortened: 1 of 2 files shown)


brishtibheja:
Nice change. There was a relevant discussion about a redesign in [the forums](https://forums.ankiweb.net/t/i-want-to-redesign-the-card-info-window-looking-for-opinions/63318/3?u=sorata), that might interest you.

iamllama:
For my usecase, i'd like to be able to quickly see the most recent revlog entries to gauge performance and on mobile this would likely mean having to scroll every single card info page

> Some of my cards are over a decade of history. It feels silly having to scroll through potentially decades of logs over and over to see the fixed size chart.

How does the forgetting curve look like for those cards? My understanding was that the longer the revlog, the more dense and less legible/useful the chart (given fixed-width), which pairs with it being further down as well

Also, previously brought up in https://forums.ankiweb.net/t/redesign-card-info/53438

dae:
Unfortunately this is likely to upset some other users, so we'd need a different approach. One option might be to make the sections collapsible, and remember the previous state?

Grinkers:
Sorry, I did a search for the charts on the forums but didn't search very well.

> Unfortunately this is likely to upset some other users, so we'd need a different approach. One option might be to make the sections collapsible, and remember the previous state?

I agree. I mentioned this issue [on the forum](https://forums.ankiweb.net/t/i-want-to-redesign-the-card-info-window-looking-for-opinions/63318/13?u=grinkers). This is a sloppy fix to more general problem.

Issue #4152, opened 2025-07-04
Consider a simple GUI for the launcher


dae:
The current launcher implementation relies on a terminal/console window, and some users find a text interface daunting. We might want to consider a lightweight Rust GUI implementation instead of a terminal - preferably something that doesn't increase the launcher size too much, or make the build complicated. egui? tauri? Other ideas?

glutanimate:
I don't have much to contribute technically at the moment (though tauri or something similar definitely sounds promising), but I just wanted to really strongly second this. I've been loosely following the discussions and totally understand the constraints that are pushing us away from PyOxidizer. That said, I'd really love to see us find a way to hide the current launcher before 25.07 gains broader adoption and people associate the current experience with Anki.

The current UX feels pretty unexpected by today’s standards, and given that Anki already has a reputation for being a bit intimidating to new users, this adds another visible friction point that we shouldn’t underestimate. I would honestly be very hesitant recommending non-techy people to give this a try in the current state, and with school starting back up again (or fairly soon, not sure on US schedules), right now probably is a period where a lot of users would be gaining their first impression of Anki. And that would actually make me a bit nervous with this launcher.

In any case, sorry for the slightly rambly and not particularly helpful comment in terms of actually getting to a solution, just wanted to heavily +1 this.

Grinkers:
A short term suggestion: if the env is empty (no lock file?), then auto select 1. At least then the First Time User Experience will just have anki show up. Maybe change to Tools->Upgrade just select 1 too?

It's quite rare for software to optimize downgrades, if they support it at all.

abdnh:
Tauri does look like a good choice. Even though not it's not as lightweight as egui, it has a lot of options for [distribution](https://tauri.app/distribute/) and [code signing](https://tauri.app/distribute/sign/macos/) that Anki can probably make use of to reduce custom code.

It'll also solve [the task manager issue on Windows](https://github.com/ankitects/anki/issues/4144). Uv can be embedded as an external binary: https://tauri.app/develop/sidecar/

phinjensen:
I definitely agree that an automatic download of the latest version would be best. Maybe there's a case for a special build that doesn't auto-install the latest version, but I feel like that's definitely the outlier in user needs.

dae:
The next launcher update will install the latest version on first run, but I don't think it makes sense to do so when the user choose Tools>Upgrade/Downgrade, as we do need an easy way for users to switch between versions due to the add-on ecosystem.

[...]

PR #4177, opened 2025-07-06
Try unix terminals in roughly most specific to least specific.

Grinkers:
I found this while looking into why the launcher wasn't working on one of my systems (no issue with Anki, an old gnome-terminal was throwing a fit with my wayland DE). 

I figured since I was already here, I should make it a bit more pleasant for people who might have installed linux with gnome or kde, but have since installed new terminals.

Feel free to close if we think https://github.com/ankitects/anki/issues/4152 will be done before a new release. I don't think it's worth going into being able to define terminals just for the temporary launcher.

### DIFF
--- a/qt/launcher/src/platform/unix.rs
+++ b/qt/launcher/src/platform/unix.rs
@@ -9,17 +9,23 @@ use anyhow::Result;
 pub fn relaunch_in_terminal() -> Result<()> {
     let current_exe = std::env::current_exe().context("Failed to get current executable path")?;
 
-    // Try terminals in order of preference
+    // Try terminals in roughly most specific to least specific.
+    // First, try commonly used terminals for riced systems.
+    // Second, try the minimalist/compatibility terminals.
+    // Finally, try terminals usually installed by default.
     let terminals = [
-        ("x-terminal-emulator", vec!["-e"]),
-        ("gnome-terminal", vec!["--"]),
-        ("konsole", vec!["-e"]),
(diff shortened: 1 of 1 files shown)


dae:
Sounds reasonable, thanks.

pkubowicz:
Why is xterm before konsole?

I use pretty standard Kubuntu (i.e., KDE), and I have both xterm and konsole (without ever manually installing xterm). I think every Ubuntu (or Debian) installation contains xterm as a fallback, but it's just it – a really ugly app that looks nothing like the surrounding desktop environment. To a GNOME or KDE user, xterm launching will most probably cause raising eyebrows, as it is not something you see often, it feels like doing a time travel to a distant past.

I think the order should be the opposite: gnome-terminal and konsole before, and xterm as the last resort.

Grinkers:
> Why is xterm before konsole?
> 
> I use pretty standard Kubuntu (i.e., KDE), and I have both xterm and konsole (without ever manually installing xterm). I think every Ubuntu (or Debian) installation contains xterm as a fallback, but it's just it – a really ugly app that looks nothing like the surrounding desktop environment. To a GNOME or KDE user, xterm launching will most probably cause raising eyebrows, as it is not something you see often, it feels like doing a time travel to a distant past.
> 
> I think the order should be the opposite: gnome-terminal and konsole before, and xterm as the last resort.

Sorry, thought it was only installed opt-in for compatibility, but I guess it's installed by default for compatibility.

PR #4179, opened 2025-07-06
Feat/Replace easy day table with display:grid

Luc-Mcgrady:
Apparently the previous implementation ran into issues on small screens (@brishtibheja).

# After

image
## Simulator:
image

# Before

image
## Simulator:
image


brishtibheja:
Nice, don't you think you should have a max-width though? For larger screens.

dae:
Thanks as always Luc!

PR #4180, opened 2025-07-07
aqt wheel: fix unintended inclusion of qt/aqt/data

gonsoos:
The inclusion of files under `qt/aqt` is handled by `qt/pyproject.toml`, not `qt/hatch_build.py`, which works with the files under `out/qt/_aqt`.

Excluding qt/aqt/data and all files ending in `.ui` in `qt/pyproject.toml` fixes the issue, since the other unwanted files (**/*.scss, **/*.ts, and **/tsconfig*) all reside under `qt/aqt/data`.

Fixes: 04996c77f3866ff23441ed357f075424e583f99a


dae:
Thank you!

PR #4186, opened 2025-07-08
Re-order terminals, again, for better UX.

Grinkers:
See https://github.com/ankitects/anki/pull/4177. A better ordering.


Grinkers:
@pkubowicz Does this seem better?

dae:
On Debian/Ubuntu, x-terminal-emulator should be a symlink to the user's preferred terminal program, so it might be worth putting up higher? Not sure how many users have it accurately set though.

Grinkers:
> On Debian/Ubuntu, x-terminal-emulator should be a symlink to the user's preferred terminal program, so it might be worth putting up higher? Not sure how many users have it accurately set though.

I don't have a convenient debian based system to check, but if setting x-terminal-emulator is opt-in then it should certainly be higher. I guess if we're being more complete, we should probably look for $XDG_CURRENT_DESKTOP so even if a user has both kde and gnome, we pick the right one. Not sure if this is all worth it, instead of just working on the gui though.

At the very least, I do think this ordering is better. Essentially riced -> wayland -> x11

pkubowicz:
Thank you for doing the tweaks.

In my Kubuntu, I have x-terminal-emulator installed by default and running it starts Konsole.

But the order looks good enough now.

dae:
Fingers crossed this doesn't cause another regression!

GithubAnon0000:
> On Debian/Ubuntu, x-terminal-emulator should be a symlink to the user's preferred terminal program

I can confirm this is true (tested on debian).

PR #4187, opened 2025-07-08
Fix/FSRS simulator fallback to memory_state_from_sm2 for after setting “Ignore cards reviewed before”

L-M-Sherlock:
source: https://forums.ankiweb.net/t/strange-fsrs-simulator-result-after-setting-ignore-cards-reviewed-before/59876

It's a following patch for #3940

PR #4189, opened 2025-07-08
Fix/FSRS simulator fallback to memory_state_from_sm2 when converting cards

L-M-Sherlock:
previous discussion: #4187

### DIFF
--- a/rslib/src/scheduler/fsrs/simulator.rs
+++ b/rslib/src/scheduler/fsrs/simulator.rs
@@ -132,7 +132,7 @@ impl Collection {
         fn is_included_card(c: &Card) -> bool {
             c.queue != CardQueue::Suspended
                 && c.queue != CardQueue::PreviewRepeat
-                && c.queue != CardQueue::New
+                && c.ctype != CardType::New
         }
         // calculate any missing memory state
         for c in &mut cards {


rslib/src/scheduler/fsrs/simulator.rs
```diff
@@ -141,15 +144,28 @@ impl Collection {
             }
         }
         let days_elapsed = self.timing_today().unwrap().days_elapsed as i32;
-        let new_cards = cards
-            .iter()
-            .filter(|c| c.memory_state.is_none() || c.queue == CardQueue::New)
-            .count()
-            + req.deck_size as usize;
+        let new_cards =
+            cards.iter().filter(|c| c.ctype == CardType::New).count() + req.deck_size as usize;
```
Luc-Mcgrady: https://github.com/ankitects/anki/pull/4187#discussion_r2192403041 continued:
@user1823 new_cards is used to determine how many cards to add; so shouldn't suspended cards be excluded?
```suggestion
            cards.iter().filter(|c| c.queue == CardQueue::New).count() + req.deck_size as usize;
```

user1823: But, many people (especially those who use pre-made decks) keep most of their new cards suspended and then unsuspend them when they want to introduce those cards.

Secondly, even before this PR, the simulator was counting these suspended new cards in `new_cards` because of the `c.memory_state.is_none()` part.

Luc-Mcgrady: > But, many people (especially those who use pre-made decks) keep most of their new cards suspended and then unsuspend them when they want to introduce those cards.

I think the simulator should stay true to what would happen without manual intervention. To assume that the user will always plan to un-suspend all the cards seems weird to me. If they do plan to then they can use `aditional cards to simulate` but this will not work in reverse for people who don't plan to include their suspended cards. (Like myself 😉)

> Secondly, even before this PR, the simulator was counting these suspended new cards in new_cards because of the c.memory_state.is_none() part.

This is not true. I've tested it. Try suspending all your cards in a deck with new cards and you will get `no new cards to simulate` (at least I do). I'd look to try and find why but am on holiday 🏝️ (I'm using an out of date version of Anki in case you can't reproduce this).

image

user1823: > I think the simulator should stay true to what would happen without manual intervention. 

I was tending to shift my opinion towards that. But then, I found the following search query, which makes most of the above arguments invalid, IMO. 😅
https://github.com/ankitects/anki/blob/208729fa3e3ecb261c359c9a75e83291e80b499d/ts/routes/deck-options/FsrsOptions.svelte#L94

But still, I think that we should use `type` so that the buried new cards are included in the count.

Also, do you agree that `is_included_card` should also be modified to match new cards based on their `type` so that Anki doesn't try to use memory_state_from_sm2 on buried new cards?

Luc-Mcgrady: Sounds good to me. I don't want to delve too deeply into anything right now.

Luc-Mcgrady:
```suggestion
            cards.iter().filter(|c| c.ctype == CardType::New && c.queue != CardQueue::Suspended).count() + req.deck_size as usize;
```
Just in case suspended cards are ever included in the search for whatever reason?

user1823:
@L-M-Sherlock, as discussed, we need to change line 135 in  `fn is_included_card` to

```rs
                && c.ctype != CardType::New
```

dae:
Thanks everyone!

PR #4190, opened 2025-07-08
Fix/Rename `Retention rate` to `Retention Rate`

ShanaryS:
`Title Case` is used for all other titles.


dae:
Thank you for taking the time to contribute! I'll need to hold off on this for a little bit until the outcome of https://forums.ankiweb.net/t/rename-true-retention-retention-rate/63446/9 has been decided

PR #4191, opened 2025-07-08
set UV_PYTHON_DOWNLOADS=auto when doing `uv sync`

Grinkers:
This is a proper fix for https://github.com/ankitects/anki/pull/4162

uv scans many paths. At least the System, User, and Workspace. It silently merges the options. This PR is a revert of our workaround and explicitly overrides the `UV_PYTHON_DOWNLOADS` flag to always download python.

### DIFF
--- a/build/runner/src/pyenv.rs
+++ b/build/runner/src/pyenv.rs
@@ -34,8 +34,9 @@ pub fn setup_pyenv(args: PyenvArgs) {
 
     run_command(
         Command::new(args.uv_bin)
+            .env_clear()
             .env("UV_PROJECT_ENVIRONMENT", args.pyenv_folder.clone())
-            .args(["sync", "--locked"])
+            .args(["sync", "--locked", "--no-config"])
             .args(args.extra_args),
     );
 
--- a/qt/launcher/src/main.rs
+++ b/qt/launcher/src/main.rs
@@ -255,27 +255,13 @@ fn handle_version_install_or_update(state: &State, choice: MainMenuChoice) -> Re
         None
     };
 
-    // `uv sync` sometimes does not pull in Python automatically
-    // This might be system/platform specific and/or a uv bug.
-    let mut command = Command::new(&state.uv_path);
-    command
-        .current_dir(&state.uv_install_root)
-        .env("UV_CACHE_DIR", &state.uv_cache_dir)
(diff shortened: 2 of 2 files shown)


zanieb:
You could also set `UV_NO_CONFIG` or `--no-config`, I'm not sure if you want to respect other user uv settings.

Grinkers:
https://github.com/astral-sh/uv/issues/3420
uv follows the precedence that cargo has set with silently unifying. I think we should probably let users have their settings? But also maybe we just want to do things for the user.

I'll let @dae decide.

dae:
We want Anki to function correctly regardless of the customization a user's done to the system-wide uv, so I think we might be best off with a two-pronged approach: clearing out the existing env vars, and disabling the use of system config. WDYT of the change I've pushed?

Grinkers:
> We want Anki to function correctly regardless of the customization a user's done to the system-wide uv, so I think we might be best off with a two-pronged approach: clearing out the existing env vars, and disabling the use of system config. WDYT of the change I've pushed?

I did a clean build of 7760fd253 and it works on my system, so I think we're good.

For future reference as of Fedora 25-07, 
```
cat /etc/uv/uv.toml
# The RPM-packaged uv deviates from the default configuration in two ways.
#
# First, we set "python-downloads" to "manual" in order to avoid unintended
# Python downloads. We suggest using RPM-packaged (system) Pythons that benefit
# from distribution maintenance and integration. Use "uv python install" to
# manually install managed Pythons.
#
# Second, we set "python-preference" to "system" instead of "managed".
# Otherwise, any managed Python would be used for uv operations where no
# particular Python is specified, even if the only available managed Python
# were much older than the primary system Python.
#
# No choices can be appropriate for all users and applications. To restore the
# default behavior, comment out settings in this file or override them in a
# configuration file with higher precedence, such as a user-level configuration
# file. See https://docs.astral.sh/uv/configuration/files/ for details on the
# interaction of project-, user-, and system-level configuration files.
#
# https://docs.astral.sh/uv/reference/settings/#python-downloads
python-downloads = "manual"
# https://docs.astral.sh/uv/reference/settings/#python-preference
python-preference = "system"
```

dae:
Thanks again for your help here :-)

Edit: I forgot to rename the PR title, oh well.

Edit 2: nothing's ever easy 😅  the removal of things like TEMP is breaking Windows

PR #4193, opened 2025-07-09
FIX Graph Tooltip uses wrong font

GithubAnon0000:
Fixes the tooltip issue mentioned in https://forums.ankiweb.net/t/inconsistent-chinese-japanese-ui-font-when-anki-is-in-a-different-language-than-the-system-language-on-windows/63261/9?u=anon_0000

Basically, the tooltip continued to use the bootstrap font instead of using the system font.


iamllama:
> Basically, the tooltip continued to use the bootstrap font instead of using the system font.

Any reason why this happens? `ts/routes/+layout.svelte` already indirectly imports `ts/lib/sass/base.scss`

EDIT: ah, bootstrap defines its own tooltip class that this one shadows

dae:
Thank you both!

PR #4194, opened 2025-07-10
Feat/per-deck desired retention

L-M-Sherlock:
Source: https://forums.ankiweb.net/t/allow-configuring-dr-for-each-deck-in-a-preset/61103

TODOs:
- [x] support pre-deck desired retention when rescheduling


proto/anki/deck_config.proto
```diff
@@ -219,6 +219,8 @@ message DeckConfigsForUpdate {
       bool review_today_active = 5;
       // Whether new_today applies to today or a past day.
       bool new_today_active = 6;
+      // Deck-specific desired retention override
+      optional float desired_retention = 7;
```
dae, line 223: While you've modelled this off the old limits code, DR is not really a 'limit', and it probably makes more sense to put in CurrentDeck instead?

rslib/src/scheduler/answering/mod.rs
```diff
@@ -444,6 +444,17 @@ impl Collection {
...
+        // Get deck-specific desired retention if available, otherwise use config
+        // default
+        let desired_retention = if let Ok(normal_deck) = deck.normal() {
+            normal_deck
+                .desired_retention
+                .unwrap_or(config.inner.desired_retention)
+        } else {
+            config.inner.desired_retention
+        };
+
```
dae: A little more DRY:

```
let desired_retention = deck
    .normal()
    .ok()
    .and_then(|d| d.desired_retention)
    .unwrap_or(config.inner.desired_retention);
```

rslib/src/scheduler/answering/mod.rs
```diff
@@ -662,6 +673,46 @@ pub(crate) mod test {
...
+        // Enable FSRS
+        col.set_config_bool(BoolKey::Fsrs, true, false)?;
+
+        // Create a deck with specific desired retention
+        let deck_id = DeckId(1);
+        let deck = col.get_deck(deck_id)?.unwrap();
+        let mut deck_clone = (*deck).clone();
+        if let DeckKind::Normal(ref mut normal) = deck_clone.kind {
+            normal.desired_retention = Some(0.85); // Set deck-specific desired
```
dae: *deck_clone.normal_mut().unwrap() = ...

rslib/src/scheduler/fsrs/memory_state.rs
```diff
@@ -205,7 +205,17 @@ impl Collection {
             .storage
             .get_deck_config(conf_id)?
             .or_not_found(conf_id)?;
-        let desired_retention = config.inner.desired_retention;
+
+        // Get deck-specific desired retention if available, otherwise use config
+        // default
+        let desired_retention = if let Ok(normal_deck) = deck.normal() {
```
dae: again

ts/routes/deck-options/FsrsOptions.svelte
```diff
@@ -85,6 +88,29 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
     $: newCardsIgnoreReviewLimit = state.newCardsIgnoreReviewLimit;
 
+    // Create tabs for desired retention
```
dae, line 91: These looks a bit awkward without 'today only', but someone with more UI experience can tweak them later.

proto/anki/deck_config.proto
L-M-Sherlock: `CurrentDeck` is not writable in the frontend. It seems to induce a lot of changes if we make it writable? And we should also add a new field to `UpdateDeckConfigsRequest` and modify the related code.

dae: It's not a huge amount of changes, but it's not a tiny amount either, so I'm ok with you leaving things as they are currently.

L-M-Sherlock:
After reading some texts about daily limits in the [manual](https://docs.ankiweb.net/deck-options.html#daily-limits) and [FAQ](https://faqs.ankiweb.net/the-2021-scheduler.html#daily-limits), I run into a problem: 

> The limits of each deck will control the number of cards that are gathered from that deck and its subdecks. Limits are applied from the deck you select, so if you select a child deck, its parents' limits will not apply.

So, the per-deck daily limits would affect subdecks. But the current per-deck desired retention design of this PR doesn't.

After checking the related code, we need something like `LimitTreeMap` for the per-deck desired retention. But it seems not a minor change.

@dae, any thoughts?

ts/routes/deck-options/SpinBoxFloatRow.svelte
```diff
@@ -23,9 +23,12 @@
         <slot />
     </Col>
     <Col --col-size={6} breakpoint="xs">
-        <ConfigInput>
-            <SpinBox bind:value {min} {max} {step} {percentage} bind:focused />
-            <RevertButton slot="revert" bind:value {defaultValue} />
-        </ConfigInput>
+        <Row class="flex-grow-1">
```
abdnh, line 26: Not something that needs to be done in this PR, but just a note that the SpinBoxFloatRow and SpinBoxRow components are almost identical and can probably be refactored into a single component.

ts/routes/deck-options/FsrsOptions.svelte
abdnh: Maybe space-evenly/space-around looks a bit better here.

abdnh:
> So, the per-deck daily limits would affect subdecks. But the current per-deck desired retention design of this PR doesn't.
> 
> After checking the related code, we need something like `LimitTreeMap` for the per-deck desired retention. But it seems not a minor change.

Something like this maybe?

```diff
diff --git a/rslib/src/scheduler/answering/mod.rs b/rslib/src/scheduler/answering/mod.rs
index 53f16c753..e875f245a 100644
--- a/rslib/src/scheduler/answering/mod.rs
+++ b/rslib/src/scheduler/answering/mod.rs
@@ -447,12 +447,20 @@ impl Collection {

         // Get deck-specific desired retention if available, otherwise use config
         // default
+        let config_desired_retention = if let Some(config_id) = self.get_current_deck()?.config_id()
+        {
+            self.get_deck_config(config_id, true)?
+                .unwrap_or_default()
+                .inner
+                .desired_retention
+        } else {
+            config.inner.desired_retention
+        };
         let desired_retention = deck
             .normal()
             .ok()
             .and_then(|d| d.desired_retention)
-            .unwrap_or(config.inner.desired_retention);
-
+            .unwrap_or(config_desired_retention);
         let fsrs_enabled = self.get_config_bool(BoolKey::Fsrs);
         let fsrs_next_states = if fsrs_enabled {
             let params = config.fsrs_params();
```

L-M-Sherlock:
> ```diff
> +        let config_desired_retention = if let Some(config_id) = self.get_current_deck()?.config_id()
> ```

Sorry, I don't know what the `current_deck` means here. What is it different from `storage.get_deck(card.deck_id)`?

abdnh:
It's different from card.deck_id when the card is in a subdeck of the deck you selected, so the parent deck's desired retention will be used, similar to the case with daily limits. 

This code is more correct (the current deck's specific value is checked first):
```diff
diff --git a/rslib/src/scheduler/answering/mod.rs b/rslib/src/scheduler/answering/mod.rs
index 53f16c753..9f04436d7 100644
--- a/rslib/src/scheduler/answering/mod.rs
+++ b/rslib/src/scheduler/answering/mod.rs
@@ -447,11 +447,15 @@ impl Collection {

         // Get deck-specific desired retention if available, otherwise use config
         // default
-        let desired_retention = deck
+        let current_deck = self.get_current_deck()?;
+        let current_deck_config = self
+            .get_deck_config(current_deck.config_id().unwrap_or_default(), true)?
+            .unwrap_or_default();
+        let desired_retention = current_deck
             .normal()
             .ok()
             .and_then(|d| d.desired_retention)
-            .unwrap_or(config.inner.desired_retention);
+            .unwrap_or(current_deck_config.inner.desired_retention);

         let fsrs_enabled = self.get_config_bool(BoolKey::Fsrs);
         let fsrs_next_states = if fsrs_enabled {
```

dae:
Apologies if it's me here who's misreading things, but I think you guys might be talking about two different things?

1) Jarrett seems to be asking if we should fall back on a parent deck's setting if the current deck hasn't been customized. I think the answer is ... probably not? This new setting is not really a 'limit', but we put it there for convenience. We only propagate the new/review limits down the tree because each of the limits affects its children, which doesn't apply here.
2) I think Abdo's comments were more about the filtered deck case? Or was there something else you're referring to Abdo - maybe the deck that the user has selected? Is your thinking that the currently-selected deck should apply the setting to child decks? That seems inconsistent with other scheduling settings, which are drawn from the current card's deck/config.

abdnh:
> Is your thinking that the currently-selected deck should apply the setting to child decks? That seems inconsistent with other scheduling settings, which are drawn from the current card's deck/config

I was thinking maybe it should work similar to Display Order settings, which only use the selected deck. But yes, that's inconsistent with most other settings.

L-M-Sherlock:
Let me elaborate it via a hypothetical collection with decks and presets:

Preset A's desired retention = 90%
Preset B's desired retention = 80%

Preset B: Parent: {Card 1}
Preset A: Parent::Child1: {Card 2} (with deck-specific desired retention 70%)
Preset A: Parent::Child1::Grandchild: {Card 3}

When I review Card 2, the possible values of desired retention to use are:
1. 70% (just respect the deck-specific desired retention of the deck which the card belong to)
2. 70% or 80% or 90% (depends on which deck I selected)

When I review Card 3, the possible values of desired retention to use are:
1. 90% (when the deck-specific DR is missing, respect to the preset's DR)
2. 70% (the deck-specific DR will override all its subdecks' DR)
3. 70% or 80% or 90% (depends on which deck I selected)

In my opinion, `70% or 80% or 90%` is too confusing. And the current implementation is fallback to preset's DR. But I prefer the override version.

Gardengul:
> 2\. 70% (the deck-specific DR will override all its subdecks' DR)

If someone wants to redefine all the subdecks in the deck, they will use the "Save to All Subdecks".

> 1\. 70% (just respect the deck-specific desired retention of the deck which the card belong to)
> 1\. 90% (when the deck-specific DR is missing, respect to the preset's DR)

It seems this is exactly the option users expect. 
You can ask the users on the forum.
https://forums.ankiweb.net/t/allow-configuring-dr-for-each-deck-in-a-preset/61103

user1823:
IMO, the expected behaviour would be:

1. 70% (just respect the deck-specific desired retention of the deck which the card belong to)

2. 70% (the deck-specific DR will override all its subdecks' DR) [unless the subdecks have their own deck-specific DR]

Note: I have added a caveat to the second case in square brackets.

Expertium:
Poll: https://forums.ankiweb.net/t/allow-configuring-dr-for-each-deck-in-a-preset/61103/20?u=expertium

Expertium:
~~It seems like there is no consensus~~ Actually, maybe there is
image

After reading replies, I think a good plan would be:
1) Rename "Limits start from top" to something that doesn't use the word "limits". We could create another setting like that specifically for DR, but idk if that's really necessary
2) If that setting is enabled, then per-deck DR of deck A also applies to its subdeck B and to its subsubdeck C, etc.
3) If disabled, per-deck DR only applies to deck A, but not to its subdecks

Thoughts, everyone?

Or just go with the "no subdecks" approach since that's what the majority have voted for ¯\\_(ツ)_/¯

EDIT: ok, let's just go with "no subdecks" approach, seems like that's what most people want

L-M-Sherlock:
Sounds good to me, lol. Let's keep the current implementation.

user1823:
There's duplicated logic for obtaining a deck's effective desired retention value in compute_memory_state() and answering/mod.rs. This logic can be extracted to a helper function to avoid code duplication.

```rs
fn effective_desired_retention(deck: &Deck, config: &DeckConfig) -> f32 {
    deck
        .normal()
        .ok()
        .and_then(|d| d.desired_retention)
        .unwrap_or(config.inner.desired_retention)
}
```
```rs
let desired_retention = effective_desired_retention(&deck, &config);
```

ts/routes/deck-options/FsrsOptions.svelte
dae: Seems to be a bit better, but some people might find the standard limits are now a bit worse?

dae:
Thanks for your work on this Jarrett, and for your patience.

tmujir:
Hi there! I am just a user so apologies if this is not the right place to post. But what is the point of this feature if it doesn't allow for subdecks to have their own individual DRs separate from the parent deck DR? I was dismayed to find out that it doesn't work that way. It would be really great if we had a toggle for "Subdeck DR overrides parent deck DR" because otherwise this feature only works for parent level decks, no? Please correct me if I am mistaken.

L-M-Sherlock:
Please check the previous discussion.

tmujir:
Thanks for your reply! I saw that the majority voted for the "no subdecks" approach, is it possible that we can get a toggle option to have the subdeck DR override the parent DR so everyone can be satisfied?

L-M-Sherlock:
I think it's feasible but very complicated. I recommend creating an issue to request that feature.

tmujir:
Thanks, I made a thread!

https://forums.ankiweb.net/t/request-option-for-deck-specific-dr-to-apply-subdecks/67215

PR #4197, opened 2025-07-12
fix: set cursor to pointer when on range

### LINKED ISSUE #4195: Hovering the easy days sliders should use the `Hand` cursor
When you hover the `Easy days` sliders, the `Hand` cursor should show up (the one below), like when you hover a switch.

Image

As of now, using the default `Arrow` cursor is an accessibility issue.

```text
Anki 25.07.2 (3adcf05c) (ao)
Python 3.13.5 Qt 6.8.1 PyQt 6.8.0
Platform: Windows-11-10.0.26100-SP0
```

Somewhat related: https://forums.ankiweb.net/t/deck-options-easy-days-switch-triggered-when-scrolling/63854

bradleyszoke:
Fixes an accessibility issue where the cursor does not change from default to pointer when hovering the easy days slider. 

Closes #4195


abdnh:
Thank you!

PR #4201, opened 2025-07-13
Add margin to QPushButton to prevent clipping

### LINKED ISSUE #4183: Text is partly cut in QPushButton:focus
Testing on Anki 25.07.2 for Windows. The issue has been present since around version 2.1.55, I believe.
Edit: also reproducible in version ⁨25.08b1.

See, for example, this report: https://forums.ankiweb.net/t/anki-25-01-02-beta-rc/54490/50

Notice how the word `Yes` is partially cut off on both the left and right sides:

Image

This could be related to the following code:

https://github.com/ankitects/anki/blob/c57b7c496d26bac0ecaaa4d18825a9c6b93ce532/qt/aqt/stylesheets.py#L177-L183

- Changing the `QPushButton:focus` `border` to `1px` seems to prevent the issue, but also alters the appearance of buttons, e.g., in the Editor.
- Adding any `padding` to `QPushButton:focus`, even `padding: 0px;`, also appears to prevent the issue. However, the underline below the "Y" in "Yes" changes slightly in length when the button is focused versus unfocused. Also, it breaks some styles, e.g., in some add-on created buttons.
- Changing `QPushButton:default`'s `border` to `2px` produces a similar effect as the previous solution but also breaks some styles (e.g., the `Reset Window Sizes` button in Preferences).

Here is how the `Yes` button appears after applying `QPushButton:focus {{ padding: 0px; }}` (the text is no longer cut off, but the underline under the "Y" is slightly longer):

Image

jcznk:
Closes #4183 

Took inspiration from: https://stackoverflow.com/questions/66637388/is-there-a-way-to-change-the-external-border-of-a-qpushbutton

This trick should prevent border growth on QPushButtons on `focus` and `hover` from shrinking the space available for the button’s label. Previously, this could cause text (such as `Yes` in confirmation dialogs) to be clipped on the left and right.

Before: 
image

After:
image

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -234,6 +234,7 @@ Emmanuel Ferdman <https://github.com/emmanuel-ferdman>
 Sunong2008 <https://github.com/Sunrongguo2008>
 Marvin Kopf <marvinkopf@outlook.com>
 Kevin Nakamura <grinkers@grinkers.net>
+jcznk <https://github.com/jcznk>
 
 ********************
 
--- a/qt/aqt/stylesheets.py
+++ b/qt/aqt/stylesheets.py
@@ -177,9 +177,13 @@ class CustomStyles:
     QPushButton:default {{
         border: 1px solid {tm.var(colors.BORDER_FOCUS)};
     }}
+    QPushButton {{
+        margin: 1px;
+    }}
     QPushButton:focus {{
         border: 2px solid {tm.var(colors.BORDER_FOCUS)};
         outline: none;
+        margin: 0px;
     }}
     QPushButton:hover,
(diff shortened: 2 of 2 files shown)


user1823:
@jcznk, thank you for the fix. I think that this change has largely fixed the issue. But, I still noticed some clipping.

In the choose deck window opened from the Add note window in Anki 25.09,
Before hover:
image

On hover: (The screenshot tool didn't capture the mouse pointer)
image

It would be great if you could fix this too.

jcznk:
@user1823 Thanks for the report! I could also reproduce this. I pushed a fix: https://github.com/ankitects/anki/pull/4323

PR #4202, opened 2025-07-13
Fix default-coloured IO masks not following css var

iamllama:
ref: https://forums.ankiweb.net/t/io-color-variables-are-broken/63925

#4048 introduced a regression where default-coloured IO masks no longer followed `--inactive-shape-color`, which this pr proposes to fix


abdnh:
Not sure if this was also introduced by your original PR, but the `highlight-shape-color` variable is also not working properly?

image

(Transparent shape on answer side).

iamllama:
@abdnh isn't that because of the alpha in the default highlight colour (#ff8e8e00)? If i remove it by changing it to #ff8e8e it now shows up

abdnh:
Oh I see 👍

PR #4208, opened 2025-07-15
Feat/Use cached workload values

Luc-Mcgrady:
Makes the info box update faster after a longer initial delay.

https://github.com/user-attachments/assets/2bd7b145-a0bf-40be-bcd4-fa1ffe0e762a


Expertium:
For Dae: we might have to make this feature slower to make it more accurate, hence why caching can be useful

dae:
How does the combined method perform compared to the sum of the individual requests? Is the goal here to minimize latency or total wallclock time? I'm guessing you're trying to optimize for the "hold down the spinner" case, at the expense of the 'quick single edit' case?

Luc-Mcgrady:
I might be mistaken (I haven't tested it or anything) but I was under the impression that a lot of the delay regarding the values changing was produced by the overhead from API call itself (?) rather than the the rust code. This also allows us to use one single SQL call for all the values which should further speed it up.

The current code can also cause race conditions such as (I believe) this:
https://forums.ankiweb.net/t/anki-25-06-beta/62271/8?u=a_blokee

Would it be worth me using `console.time` or something to check these assertions?

dae:
It's not a no at this point, but a rough approximation based on the expected speed of the new proposed code would be good. Approx 30 requests over localhost shouldn't have much in the way of latency, so I'd expect the combined call to be roughly equivalent. 

Currently the calculation is done when the user selects that field, so it does some useless calculation as the user tabs through fields. If we go down this path, it would be nicer if the calculation were deferred until something had actually changed.

Luc-Mcgrady:
This PR:
image
Before:
image

Both seem to take approximately the same time per request. I'm assuming this is because of the SQL statement.

dae:
Thanks, with those results the combined approach does look preferable.

Luc-Mcgrady:
The tests failing is just a CI bug.

image

dae:
🚀

PR #4213, opened 2025-07-19
Fix list undo error on card editor

### LINKED ISSUE #4010: If a list is un-done, then the text gets appended after the list
# Steps to Reproduce
1. Click **Add**.
2. Type a random string into a field, like `Hello World`.
3. Press <kbd>Ctrl+Z</kbd> to undo.
4. The list persists and the now re-done entry gets appended (html becomes `<ol><li>Hello World</li></ol>Hello World`).

# Expected Behavior
The list should be replaced by the previous content of the field, i.e. the content that is now re-done.

# Actual Behavior
The list stays and the previous content gets appended.

# Versions
It goes back to at least 24.11 (87ccd24e), but is still present with the current build from main from today: Anki 25.05 (d3d6bd8ce006f178e2271fd8d317fdc8832095df)

<details><summary>_</summary>
<p>

Anki 25.05 (8fc822a6) (src) (ao) → My git is at d3d6bd8ce006f178e2271fd8d317fdc8832095df, not sure why Anki reports the wrong commit. **Edit:** Deleting `out/buildhash` solved the incorrect hash issue. No idea why that was neccessary, though. **Edit 2:** [Here](https://forums.ankiweb.net/t/the-buildhash-isnt-updating-properly-e-g-after-checking-out-another-branch-is-this-intended-behavior-or-a-bug/61066/2?u=anon_0000)'s why that was neccessary.

</p>
</details> 

Originially reported in https://forums.ankiweb.net/t/anki-25-05-beta/59710/52?u=anon_0000.

fcolona:
Fix #4010

### Issues
If a list is un-done, the text gets appended after the list.

### Proposed Fix
Apparently, the issue comes from the usage of the default undo implementation and it's interaction with rich text. In that sense, the proposed fix is to use a custom implementation of the undo/redo actions in order to have finer control of the action's behavior and prevent current and future bugs.

### To do
The caret restoration after the undo/redo is not working properly. It seems to be related to doubly attaching event listeners to the contenteditable field (local and global input handlers). Some help with figuring it out would be appreciated.


fcolona:
Another related issue is the undoing of superscripts and subscripts, that does not work as expected and does not remove the superscript/subscript, but instead removes the normal text. That, too, is fixed with this custom implementation.

abdnh:
Appears to be working well for lists.

I noticed the cursor position behaves weirdly, e.g., if you bold the word “hello” then undo that, the cursor is placed after the h.

dae:
Do you plan to return to this?

fcolona:
Hi @dae, sorry for the lack of feedback. I couldn't find a way to consistently restore the cursor in all cases. And now that I am back from vacation, I won't be able to continue working on this. Closing this PR for now, but please feel free to build on it if it’s still useful.

PR #4222, opened 2025-07-23
Refactor IO fill tool target check logic

iamllama:
I've only recently found out about https://fabric5.fabricjs.com/docs/fabric.Group.html#subTargetCheck, which when set, causes event handlers to do what `findTargetInGroup` does, but without the z-index bug. This pr proposes to use that instead


dae:
Nice!

PR #4231, opened 2025-07-28
Feat/use current_retrievability_seconds in SQL

L-M-Sherlock:
fix this: https://forums.ankiweb.net/t/inconsistency-in-same-day-reviews-retrievability/64315

### DIFF
--- a/rslib/src/browser_table.rs
+++ b/rslib/src/browser_table.rs
@@ -143,6 +143,21 @@ impl Card {
             })
         }
     }
+
+    pub(crate) fn seconds_since_last_review(&self, timing: &SchedTimingToday) -> Option<u32> {
+        if let Some(last_review_time) = self.last_review_time {
+            Some(timing.now.elapsed_secs_since(last_review_time) as u32)
+        } else if !self.is_due_in_days() {
+            let last_review_time =
+                TimestampSecs(self.original_or_current_due() as i64 - self.interval as i64);
+            Some(timing.now.elapsed_secs_since(last_review_time) as u32)
+        } else {
+            self.due_time(timing).map(|due| {
+                (due.adding_secs(-86_400 * self.interval as i64)
+                    .elapsed_secs()) as u32
+            })
+        }
+    }
 }
 
 impl Note {
@@ -543,12 +558,12 @@ impl RowContext {
         self.cards[0]
             .memory_state
             .as_ref()
-            .zip(self.cards[0].days_since_last_review(&self.timing))
+            .zip(self.cards[0].seconds_since_last_review(&self.timing))
             .zip(Some(self.cards[0].decay.unwrap_or(FSRS5_DEFAULT_DECAY)))
-            .map(|((state, days_elapsed), decay)| {
-                let r = FSRS::new(None).unwrap().current_retrievability(
+            .map(|((state, seconds), decay)| {
+                let r = FSRS::new(None).unwrap().current_retrievability_seconds(
                     (*state).into(),
-                    days_elapsed,
+                    seconds,
                     decay,
                 );
                 format!("{:.0}%", r * 100.)
--- a/rslib/src/search/mod.rs
+++ b/rslib/src/search/mod.rs
@@ -378,9 +378,10 @@ fn card_order_from_sort_column(column: Column, timing: SchedTimingToday) -> Cow<
         Column::Stability => "extract_fsrs_variable(c.data, 's') asc".into(),
         Column::Difficulty => "extract_fsrs_variable(c.data, 'd') asc".into(),
         Column::Retrievability => format!(
-            "extract_fsrs_retrievability(c.data, case when c.odue !=0 then c.odue else c.due end, c.ivl, {}, {}) asc",
+            "extract_fsrs_retrievability(c.data, case when c.odue !=0 then c.odue else c.due end, c.ivl, {}, {}, {}) asc",
             timing.days_elapsed,
-            timing.next_day_at.0
+            timing.next_day_at.0,
+            timing.now.0,
         )
         .into(),
     }
--- a/rslib/src/search/sqlwriter.rs
+++ b/rslib/src/search/sqlwriter.rs
@@ -418,13 +418,13 @@ impl SqlWriter<'_> {
                 write!(self.sql, "extract_fsrs_variable(c.data, 'd') {op} {d}").unwrap()
             }
             PropertyKind::Retrievability(r) => {
-                let (elap, next_day_at) = {
+                let (elap, next_day_at, now) = {
                     let timing = self.col.timing_today()?;
-                    (timing.days_elapsed, timing.next_day_at)
+                    (timing.days_elapsed, timing.next_day_at, timing.now)
                 };
                 write!(
                     self.sql,
-                    "extract_fsrs_retrievability(c.data, case when c.odue !=0 then c.odue else c.due end, c.ivl, {elap}, {next_day_at}) {op} {r}"
+                    "extract_fsrs_retrievability(c.data, case when c.odue !=0 then c.odue else c.due end, c.ivl, {elap}, {next_day_at}, {now}) {op} {r}"
                 )
                 .unwrap()
             }
--- a/rslib/src/storage/card/filtered.rs
+++ b/rslib/src/storage/card/filtered.rs
@@ -14,6 +14,8 @@ pub(crate) fn order_and_limit_for_search(
 ) -> String {
     let temp_string;
     let today = timing.days_elapsed;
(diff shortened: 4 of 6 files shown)


user1823:
I won't be able to do a proper review for the next 1-2 days. But, a quick feedback for now: using timing.now instead of timing.next_day_at will cause Anki to ignore the user's next day starts at setting. Also see: https://github.com/ankitects/anki/pull/4124#discussion_r2163960490

L-M-Sherlock:
> using timing.now instead of timing.next_day_at will cause Anki to ignore the user's next day starts at setting

I use `timing.now` because it's more accurate. If I use `timing.next_day_at`, the retrievability of learning cards will be too inaccurate.

And these changes don't affect the scheduling, so I think they are safe.

user1823:
Maybe use a hybrid method? Calculate the elapsed time using both timing.now and timing.next_day_at and use whichever is larger?

So, for example,
- if elapsed_days = 0 (because of `as u32`) and elapsed_seconds = 10800 (3 hours), then calculate R using 3 hours
- but if elapsed_days = 1 and elapsed_seconds = 10800 (3 hours), then calculate R using 1 day.

L-M-Sherlock:
> Maybe use a hybrid method? Calculate the elapsed time using both timing.now and timing.next_day_at and use whichever is larger?

`timing.next_day_at` is always larger than `timing.now`. The `timing.next_day_at` is the timestamp of the `next_day_at`.

user1823:
But, we use as u32 with it, which is equivalent to `math.floor`

L-M-Sherlock:
> * but if elapsed_days = 1 and elapsed_seconds = 10800 (3 hours), then calculate R using 1 day.

So you think `elapsed_days = 1` is better than `elapsed_seconds = 10800`, right? Could you elaborate it with a case?

The forgetting curve is continuous, and the PR's implementation is consistent with it:

image

dae:
YOLO'ing this for the next beta, but further discussion is welcome.

rslib/src/browser_table.rs
```diff
@@ -105,7 +105,8 @@ impl Card {
 
     /// Returns true if the card has a due date in terms of days.
     fn is_due_in_days(&self) -> bool {
-        matches!(self.queue, CardQueue::DayLearn | CardQueue::Review)
+        self.original_or_current_due() <= 365_000 // keep consistent with SQL
```
user1823, line 108: This includes new cards too, which is not intended.

rslib/src/browser_table.rs
```diff
@@ -127,19 +128,17 @@ impl Card {
 
     /// This uses card.due and card.ivl to infer the elapsed time. If 'set due
```
user1823, line 129: This comment should be moved below the first if block.

rslib/src/browser_table.rs
```diff
@@ -127,19 +128,17 @@ impl Card {
...
         if let Some(last_review_time) = self.last_review_time {
-            Some(timing.next_day_at.elapsed_days_since(last_review_time) as u32)
+            Some(timing.now.elapsed_secs_since(last_review_time) as u32)
         } else if !self.is_due_in_days() {
-            Some(
-                (timing.next_day_at.0 as u32).saturating_sub(self.original_or_current_due() as u32)
-                    / 86_400,
-            )
+            let last_review_time =
+                TimestampSecs(self.original_or_current_due() as i64 - self.interval as i64);
```
user1823, line 136: The interval in seconds is stored as a negative integer. So, is this equation correct?

L-M-Sherlock: The type of `Card`'s `interval` is `u32`, so it cannot be negative. I guess you referred to a pretty old version.

user1823:
> So you think elapsed_days = 1 is better than elapsed_seconds = 10800, right? Could you elaborate it with a case?

My main concern was that this is inconsistent with R calculation in the scheduler. But, that consistency is not very important, I guess. Now, I think it's fine to calculate R differently in the scheduler and browser tables/searching, etc. until we develop a proper model for short-term memory.

Quite related, @Expertium, I know that you are trying to develop a model for short-term memory. For that, one must consider the seconds elapsed between the two reviews. But, as demonstrated in some examples above, the rollover time should also be considered. To elaborate, there should be a way for your neural network to know whether the two reviews that were separated by 8 hours were in the same day or on consecutive days (sleep consolidates memory).

Coming back to this PR, I pointed out 2 possible bugs in the comments above.

rslib/src/browser_table.rs
L-M-Sherlock: It's OK because new cards don't have memory state. So the code doesn't calculate retrievability for them.

Expertium:
> But, as demonstrated in some examples above, the rollover time should also be considered. To elaborate, there should be a way for your neural network to know whether the two reviews that were separated by 8 hours were in the same day or on consecutive days (sleep consolidates memory).

I don't plan to incorporate the rollover time into FSRS. I'll just use fractional interval lengths.
Btw, what's interesting is that after using a neural net as a part of FSRS, I found that short-term memory follows the same patterns as long-term memory, just with a steep drop-off of R in the t << S region. So we won't need entirely new formulas, we can just reuse the main formulas for SInc and post-lapse S, maybe with an extra correction for <1d intervals.

rslib/src/browser_table.rs
L-M-Sherlock: I think it's OK because it means the function is inaccurate in some cases. If we move it below the if block, it will not appear in the function doc:

image

rslib/src/browser_table.rs
user1823: Well, this is still a problem because someone might use this function for another purpose in the future. At least, there should be a comment that this function returns true for new cards too. It won't be an ideal solution as the name of the function would be misleading, but it's still better than the current situation.

rslib/src/browser_table.rs
user1823: The documentation in https://github.com/ankidroid/Anki-Android/wiki/Database-Structure looks outdated.

But, there is still an issue. The interval of relearning cards is the interval (in days) that the card will get after it passes through all the steps. It is not the interval between two steps. I didn't check the interval of learning cards, but it should behave similarly.

rslib/src/browser_table.rs
user1823: In that case, I think that the doc should be reworded as follows:
```
If last_review_date isn't stored in the card, this uses card.due and card.ivl to infer the elapsed time, which won't be accurate if 'set due date' or an add-on has changed the due date.
```

rslib/src/browser_table.rs
L-M-Sherlock: I have checked the codebase just now. Emmm, the interval of learning card is zero... So, we need other solution for this case. Here is my proposal: https://forums.ankiweb.net/t/fill-the-card-last-review-time-field-when-check-database/64792

rslib/src/browser_table.rs
L-M-Sherlock: Feel free to open a PR for these suggestions! I'm working on another issue now.

rslib/src/browser_table.rs
user1823: That looks like a good idea. But, we still need to account for the case when last_review_date is None. Would it make sense to use the revlog in that situation? The performance hit won't be very high because it's unlikely that a user has more than 100 learning or relearning cards at the same time.

rslib/src/browser_table.rs
user1823: https://github.com/ankitects/anki/pull/4235

rslib/src/browser_table.rs
L-M-Sherlock: > because it's unlikely that a user has more than 100 learning or relearning cards at the same time

It's hard to say. Only @dae knows the overall situations. What do you think of this idea?

user1823: @dae, what do you think about this proposed solution? For (re)learning cards, if last_review_time is None, there is no way to get an estimate the time of last review without consulting the revlog.

So, would it be acceptable to use the revlog for that purpose while sorting cards in the browser (or searching cards)? 

The suggestion is based on the presumption that the performance hit won't be very high because it's unlikely that a user has more than 100 learning or relearning cards at the same time.

The alternative is to consider the due time to be the time of last review. This is obviously incorrect, but, IMO, the best option available to us if the revlog can't be used in this case.

dae: Sorry, just saw your question. I'm afraid I'm not particularly fond of trying to populate the missing data during a sort, especially as this issue will diminish over time as users update their clients/check their database.

PR #4236, opened 2025-07-29
Feat/Card stats update review time

Luc-Mcgrady:
Updates the `last_review_time` of a card when opening the stats menu as the value is calculated anyway.

I'm not sure if you're comfortable having `card_stats` modify the database but if you are here you go :)

### DIFF
--- a/rslib/src/stats/card.rs
+++ b/rslib/src/stats/card.rs
@@ -30,14 +30,26 @@ impl Collection {
 
         let (average_secs, total_secs) = average_and_total_secs_strings(&revlog);
         let timing = self.timing_today()?;
-        let seconds_elapsed = if let Some(last_review_time) = card.last_review_time {
-            timing.now.elapsed_secs_since(last_review_time) as u32
+
+        let last_review_time = if let Some(last_review_time) = card.last_review_time {
+            last_review_time
         } else {
-            self.storage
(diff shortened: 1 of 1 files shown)


rslib/src/stats/card.rs
```diff
@@ -30,14 +30,26 @@ impl Collection {
...
         } else {
-            self.storage
-                .time_of_last_review(card.id)?
-                .map(|ts| timing.now.elapsed_secs_since(ts))
-                .unwrap_or_default() as u32
+            let mut new_card = card.clone();
+            let last_review_time = TimestampSecs(
+                self.storage
+                    .time_of_last_review(card.id)?
+                    .map(|ts| timing.now.elapsed_secs_since(ts))
```
user1823:
```suggestion

```

The code is currently storing the elapsed seconds since the last review as the last review time, rather than the actual timestamp. The correct approach is to save the result of time_of_last_review (the timestamp) directly into last_review_time.

dae:
I'm worried this could lead to sync conflicts. Consider the case where a user has the card info screen open, and scrolls down the browser window. Perhaps it would be better to fill the missing info, but not write it out? Or do a 'silent update', where the property is written, but mtime/mod remain unchanged yet.

dae:
Thanks Luc!

PR #4237, opened 2025-07-30
Fix Cards with Missing Last Review Time During Database Check

L-M-Sherlock:
### Problem

It takes a long time to fill all existing cards' `last_review_time` because the user needs to review them all.

Source: https://forums.ankiweb.net/t/fill-the-card-last-review-time-field-when-check-database/64792

### Solution
This PR adds functionality to the database check process to automatically fix cards with missing last review times by:

1. **Identifying affected cards**: During database check, find cards that are not new (have been reviewed) but have no `last_review_time` recorded
2. **Recovery from revlog**: Use the revision log entries to determine the actual last review time for these cards
3. **Automatic repair**: Update the cards with the correct `last_review_time` based on their review history
4. **User feedback**: Report the number of fixed cards to the user through the database check results

### Key Changes

- **New database check category**: Added `card_last_review_time_empty` counter to track and report fixed cards
- **Improved revlog filtering**: Refactored review entry filtering logic into a reusable `has_rating_and_affect_scheduling()` method that properly excludes manual reschedules and cramming sessions
- **Enhanced card repair**: Extended `fix_card_properties()` to also repair missing last review times by cross-referencing with revlog data
- **Localization support**: Added user-facing messages to inform users about the repair process

### Benefits

- **Better scheduling accuracy**: Cards with proper last review times will be scheduled more accurately
- **Improved statistics**: Review statistics will be more accurate when all cards have correct last review times  
- **Automatic repair**: Users don't need to manually fix corrupted card data - it's handled automatically during database check
- **Transparency**: Users are informed about what was fixed during the database check process

The fix is conservative and only updates cards where the data is clearly missing, ensuring no valid data is overwritten.

### DIFF
--- a/ftl/core/database-check.ftl
+++ b/ftl/core/database-check.ftl
@@ -5,6 +5,11 @@ database-check-card-properties =
         [one] Fixed { $count } invalid card property.
        *[other] Fixed { $count } invalid card properties.
     }
+database-check-card-last-review-time-empty =
+    { $count ->
+        [one] Fixed { $count } card with no last review time.
+       *[other] Fixed { $count } cards with no last review time.
+    }
 database-check-missing-templates =
     { $count ->
         [one] Deleted { $count } card with missing template.
--- a/ftl/core/database-check.ftl
+++ b/ftl/core/database-check.ftl
@@ -7,8 +7,8 @@ database-check-card-properties =
     }
 database-check-card-last-review-time-empty =
     { $count ->
-        [one] Filled missing last review time in { $count } card.
-       *[other] Filled missing last review time in { $count } cards.
+        [one] Added last review time to { $count } card.
+       *[other] Added last review time to { $count } cards.
     }
 database-check-missing-templates =
     { $count ->
--- a/ftl/core/database-check.ftl
+++ b/ftl/core/database-check.ftl
@@ -7,8 +7,8 @@ database-check-card-properties =
     }
 database-check-card-last-review-time-empty =
     { $count ->
-        [one] Fixed { $count } card with no last review time.
-       *[other] Fixed { $count } cards with no last review time.
+        [one] Filled missing last review time in { $count } card.
+       *[other] Filled missing last review time in { $count } cards.
     }
 database-check-missing-templates =
     { $count ->
--- a/rslib/src/revlog/mod.rs
+++ b/rslib/src/revlog/mod.rs
@@ -85,6 +85,13 @@ impl RevlogEntry {
         .unwrap()
     }
 
+    /// Returns true if this entry represents a reset operation.
+    /// These entries are created when a card is reset using
+    /// [`Collection::reschedule_cards_as_new`].
+    pub(crate) fn is_reset(&self) -> bool {
+        self.review_kind == RevlogReviewKind::Manual && self.ease_factor == 0
+    }
+
     /// Returns true if the review entry is not manually rescheduled and not
     /// cramming. Used to filter out entries that shouldn't be considered
     /// for statistics and scheduling.
--- a/rslib/src/revlog/mod.rs
+++ b/rslib/src/revlog/mod.rs
@@ -88,19 +88,21 @@ impl RevlogEntry {
     /// Returns true if this entry represents a reset operation.
     /// These entries are created when a card is reset using
     /// [`Collection::reschedule_cards_as_new`].
-    /// The `ease_factor` should be 0 because
-    /// [`Collection::set_due_date`] also sets created `RevlogEntry` with
-    /// `RevlogReviewKind::Manual` but the `ease_factor` is not 0.
+    /// The 0 value of `ease_factor` differentiates it
+    /// from entry created by [`Collection::set_due_date`] that has
+    /// `RevlogReviewKind::Manual` but non-zero `ease_factor`.
     pub(crate) fn is_reset(&self) -> bool {
         self.review_kind == RevlogReviewKind::Manual && self.ease_factor == 0
     }
 
     /// Returns true if this entry represents a cramming operation.
-    /// These entries are created when a card is previewed using
+    /// These entries are created when a card is reviewed in a
+    /// filtered deck with "Reschedule cards based on my answers
+    /// in this deck" disabled.
     /// [`crate::scheduler::answering::CardStateUpdater::apply_preview_state`].
-    /// The `ease_factor` should be 0 because
-    /// [`crate::scheduler::states::ReviewState::revlog_kind`] returns
-    /// `RevlogReviewKind::Filtered` when `days_late() < 0`.
+    /// The 0 value of `ease_factor` distinguishes it from the entry
+    /// created when a card is reviewed before its due date in a
+    /// filtered deck with reschedule enabled or using Grade Now.
     pub(crate) fn is_cramming(&self) -> bool {
         self.review_kind == RevlogReviewKind::Filtered && self.ease_factor == 0
     }
--- a/rslib/src/revlog/mod.rs
+++ b/rslib/src/revlog/mod.rs
@@ -88,7 +88,7 @@ impl RevlogEntry {
     /// Returns true if the review entry is not manually rescheduled and not
     /// cramming. Used to filter out entries that shouldn't be considered
     /// for statistics and scheduling.
-    pub(crate) fn has_rating_and_affect_scheduling(&self) -> bool {
+    pub(crate) fn has_rating_and_affects_scheduling(&self) -> bool {
         // not rescheduled/set due date/reset
         self.button_chosen > 0
             // not cramming
--- a/rslib/src/revlog/mod.rs
+++ b/rslib/src/revlog/mod.rs
@@ -102,12 +102,16 @@ impl RevlogEntry {
         self.review_kind == RevlogReviewKind::Filtered && self.ease_factor == 0
     }
 
+    pub(crate) fn has_rating(&self) -> bool {
+        self.button_chosen > 0
+    }
+
     /// Returns true if the review entry is not manually rescheduled and not
     /// cramming. Used to filter out entries that shouldn't be considered
     /// for statistics and scheduling.
     pub(crate) fn has_rating_and_affects_scheduling(&self) -> bool {
         // not rescheduled/set due date/reset
-        self.button_chosen > 0
+        self.has_rating()
             // not cramming
             && !self.is_cramming()
     }
--- a/rslib/src/revlog/mod.rs
+++ b/rslib/src/revlog/mod.rs
@@ -92,6 +92,16 @@ impl RevlogEntry {
         self.review_kind == RevlogReviewKind::Manual && self.ease_factor == 0
     }
 
+    /// Returns true if this entry represents a cramming operation.
+    /// These entries are created when a card is previewed using
+    /// [`crate::scheduler::answering::CardStateUpdater::apply_preview_state`].
+    /// The `ease_factor` should be 0 because
+    /// [`crate::scheduler::states::ReviewState::revlog_kind`] returns
+    /// `RevlogReviewKind::Filtered` when `days_late() < 0`.
+    pub(crate) fn is_cramming(&self) -> bool {
+        self.review_kind == RevlogReviewKind::Filtered && self.ease_factor == 0
+    }
+
     /// Returns true if the review entry is not manually rescheduled and not
     /// cramming. Used to filter out entries that shouldn't be considered
     /// for statistics and scheduling.
@@ -99,7 +109,7 @@ impl RevlogEntry {
         // not rescheduled/set due date/reset
         self.button_chosen > 0
             // not cramming
-            && (self.review_kind != RevlogReviewKind::Filtered || self.ease_factor != 0)
+            && !self.is_cramming()
     }
 }
 
--- a/rslib/src/revlog/mod.rs
+++ b/rslib/src/revlog/mod.rs
@@ -88,6 +88,9 @@ impl RevlogEntry {
     /// Returns true if this entry represents a reset operation.
     /// These entries are created when a card is reset using
     /// [`Collection::reschedule_cards_as_new`].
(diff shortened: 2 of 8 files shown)


rslib/src/revlog/mod.rs
```diff
@@ -84,6 +84,16 @@ impl RevlogEntry {
         })
         .unwrap()
     }
+
+    /// Returns true if the review entry is not manually rescheduled and not
+    /// cramming. Used to filter out entries that shouldn't be considered
+    /// for statistics and scheduling.
+    pub(crate) fn has_rating_and_affect_scheduling(&self) -> bool {
```
Luc-Mcgrady:
```suggestion
    pub(crate) fn has_rating_and_affects_scheduling(&self) -> bool {
```

rslib/src/stats/graphs/retention.rs
```diff
@@ -53,10 +53,7 @@ impl GraphsContext {
         self.revlog
             .iter()
             .filter(|review| {
-                // not rescheduled/set due date/reset
-                review.button_chosen > 0
-                    // not cramming
-                    && (review.review_kind != RevlogReviewKind::Filtered || review.ease_factor != 0)
+                review.has_rating_and_affect_scheduling()
```
Luc-Mcgrady:
```suggestion
                review.has_rating_and_affects_scheduling()
```

rslib/src/scheduler/fsrs/memory_state.rs
```diff
@@ -323,7 +323,7 @@ fn get_last_revlog_info(revlogs: &[RevlogEntry]) -> HashMap<CardId, LastRevlogIn
         .for_each(|(card_id, group)| {
             let mut last_reviewed_at = None;
             for e in group.into_iter() {
-                if e.button_chosen >= 1 {
+                if e.has_rating_and_affect_scheduling() {
```
Luc-Mcgrady:
```suggestion
                if e.has_rating_and_affects_scheduling() {
```

ftl/core/database-check.ftl
```diff
@@ -5,6 +5,11 @@ database-check-card-properties =
         [one] Fixed { $count } invalid card property.
        *[other] Fixed { $count } invalid card properties.
     }
+database-check-card-last-review-time-empty =
+    { $count ->
+        [one] Fixed { $count } card with no last review time.
```
user1823:
```suggestion
        [one] Filled missing last review time in { $count } card.
```

ftl/core/database-check.ftl
```diff
@@ -5,6 +5,11 @@ database-check-card-properties =
...
+       *[other] Fixed { $count } cards with no last review time.
```
user1823:
```suggestion
       *[other] Filled missing last review time in { $count } cards.
```

rslib/src/scheduler/fsrs/memory_state.rs
```diff
@@ -306,15 +306,15 @@ pub(crate) fn fsrs_items_for_memory_states(
...
-struct LastRevlogInfo {
+pub(crate) struct LastRevlogInfo {
     /// Used to determine the actual elapsed time between the last time the user
     /// reviewed the card and now, so that we can determine an accurate period
     /// when the card has subsequently been rescheduled to a different day.
-    last_reviewed_at: Option<TimestampSecs>,
+    pub(crate) last_reviewed_at: Option<TimestampSecs>,
 }
 
 /// Return a map of cards to info about last review/reschedule.
```
user1823:
```suggestion
/// Return a map of cards to info about last review.
```

rslib/src/scheduler/fsrs/memory_state.rs
```diff
@@ -323,7 +323,7 @@ fn get_last_revlog_info(revlogs: &[RevlogEntry]) -> HashMap<CardId, LastRevlogIn
         .for_each(|(card_id, group)| {
             let mut last_reviewed_at = None;
             for e in group.into_iter() {
-                if e.button_chosen >= 1 {
+                if e.has_rating_and_affects_scheduling() {
                     last_reviewed_at = Some(e.id.as_secs());
                 }
```
user1823, line 330: I am assuming that this loop goes from earlier revlogs to later revlogs. So, without the following code, if a card is Reset and doesn't have any subsequent rating, the last_reviewed_at will be the timestamp of the revlog before the Reset entry, which is not desired.

```suggestion
                } else if e.button_chosen = 0 && e.ease_factor = 0 {
                    last_reviewed_at = None;
                }
```

L-M-Sherlock: If the card is Reset, its type will be New and skipped in the filling process.

rslib/src/storage/card/mod.rs
```diff
@@ -390,7 +391,20 @@ impl super::SqliteStorage {
...
             .prepare(include_str!("fix_ordinal.sql"))?
             .execute(params![mtime, usn])?;
-        Ok((new_cnt, other_cnt))
+        let mut last_review_time_cnt = 0;
+        let revlog = self.get_all_revlog_entries_in_card_order()?;
+        let last_revlog_info = get_last_revlog_info(&revlog);
+        for (card_id, last_revlog_info) in last_revlog_info {
+            let card = self.get_card(card_id)?;
+            if let Some(mut card) = card {
+                if card.ctype != CardType::New && card.last_review_time.is_none() {
```
L-M-Sherlock, line 407: @user1823 you can check this code.

rslib/src/scheduler/fsrs/memory_state.rs
user1823: Yes, but it still makes sense to correct the output of the function because we might use it for another purpose later.

rslib/src/storage/card/mod.rs
```diff
@@ -390,7 +391,20 @@ impl super::SqliteStorage {
             .db
             .prepare(include_str!("fix_ordinal.sql"))?
             .execute(params![mtime, usn])?;
-        Ok((new_cnt, other_cnt))
+        let mut last_review_time_cnt = 0;
+        let revlog = self.get_all_revlog_entries_in_card_order()?;
```
user1823, line 402: Rather than fetching all revlogs and calculating last review time for each card, wouldn't it be faster to fetch the revlogs of only those cards which lack last_review_time?

Something like this: (untested)
```sql
SELECT max(id), cid GROUP BY cid FROM revlog WHERE ease >= 1 AND cid IN (SELECT id FROM cards WHERE type >= 1 AND json_extract(data, '$.lrt') IS NULL)
```

rslib/src/scheduler/fsrs/memory_state.rs
L-M-Sherlock: OK. I find it's better to wrap this condition in a new function because this condition is also used in other cases.

rslib/src/storage/card/mod.rs
L-M-Sherlock: The current code is fast enough in some large collections. But I hope @dae could test it in some huge collections.

user1823:
The tests can be fixed by adding `button_chosen: 0,` to those revlogs in the tests that have RevlogReviewKind::Manual.

Or, you can modify `revlog()` so that it automatically sets `button_chosen` to 0 if review_kind is Manual or Rescheduled.

rslib/src/storage/card/mod.rs
```diff
@@ -365,7 +366,7 @@ impl super::SqliteStorage {
         mtime: TimestampSecs,
         usn: Usn,
         v1_sched: bool,
-    ) -> Result<(usize, usize)> {
+    ) -> Result<(usize, usize, usize)> {
```
dae: Struct would be better than unnamed tuple.

rslib/src/storage/card/mod.rs
```diff
@@ -390,7 +391,20 @@ impl super::SqliteStorage {
...
-        Ok((new_cnt, other_cnt))
+        let mut last_review_time_cnt = 0;
+        let revlog = self.get_all_revlog_entries_in_card_order()?;
+        let last_revlog_info = get_last_revlog_info(&revlog);
+        for (card_id, last_revlog_info) in last_revlog_info {
+            let card = self.get_card(card_id)?;
+            if let Some(mut card) = card {
+                if card.ctype != CardType::New && card.last_review_time.is_none() {
+                    card.last_review_time = last_revlog_info.last_reviewed_at;
+                    self.update_card(&card)?;
```
dae, line 409: This has the potential to cause data loss if run before changes on other devices have been synced. Like in https://github.com/ankitects/anki/pull/4236#issuecomment-3149180936 we could write the changes without mtime/usn bump, but I don't think that's a good idea, as then last_review_time doesn't propagate. I think our best approach here may be calling set_schema_modified() if last_review_time_cnt > 0 (dbcheck.rs:310), forcing the user to either do a one-way sync of the changes to other devices, or revert their changes.

rslib/src/revlog/mod.rs
```diff
@@ -84,6 +84,37 @@ impl RevlogEntry {
...
+        self.review_kind == RevlogReviewKind::Manual && self.ease_factor == 0
+    }
+
+    /// Returns true if this entry represents a cramming operation.
+    /// These entries are created when a card is previewed using
+    /// [`crate::scheduler::answering::CardStateUpdater::apply_preview_state`].
+    /// The `ease_factor` should be 0 because
+    /// [`crate::scheduler::states::ReviewState::revlog_kind`] returns
+    /// `RevlogReviewKind::Filtered` when `days_late() < 0`.
+    pub(crate) fn is_cramming(&self) -> bool {
```
dae, line 106: These helpers are a nice touch.

rslib/src/storage/card/mod.rs
dae: I need to give this a bit more thought. In the past, a forced one-way sync was appropriate for fixing some issues, but this is probably going to happen fairly frequently until people have updated all of their clients, so I'm a bit worried about the amount of one way syncs it'll cause. But perhaps I'm overthinking things. Anyone want to weigh in?

rslib/src/storage/card/mod.rs
L-M-Sherlock: Done on 86bbc2c

rslib/src/storage/card/mod.rs
L-M-Sherlock: Done on 6947afd

user1823: > revert their changes.

I am assuming this means using "Download from AnkiWeb" on the next sync. Isn't this more likely to cause data loss? If there are unsynced changes on both devices and the user uses Check Database, they won't get any option to preserve their changes because Anki will force a one-way sync.

rslib/src/revlog/mod.rs
```diff
@@ -84,6 +84,40 @@ impl RevlogEntry {
         })
         .unwrap()
     }
+
+    /// Returns true if this entry represents a reset operation.
+    /// These entries are created when a card is reset using
+    /// [`Collection::reschedule_cards_as_new`].
+    /// The `ease_factor` should be 0 because
```
user1823:
```suggestion
    /// The 0 value of `ease_factor` differentiates it
```

rslib/src/revlog/mod.rs
```diff
@@ -84,6 +84,40 @@ impl RevlogEntry {
...
+    /// [`Collection::set_due_date`] also sets created `RevlogEntry` with
```
user1823:
```suggestion
    /// from entry created by [`Collection::set_due_date`] that has
```

rslib/src/revlog/mod.rs
```diff
@@ -84,6 +84,40 @@ impl RevlogEntry {
...
+    /// `RevlogReviewKind::Manual` but the `ease_factor` is not 0.
```
user1823:
```suggestion
    /// `RevlogReviewKind::Manual` but non-zero `ease_factor`.
```

rslib/src/revlog/mod.rs
```diff
@@ -84,6 +84,40 @@ impl RevlogEntry {
...
+    pub(crate) fn is_reset(&self) -> bool {
+        self.review_kind == RevlogReviewKind::Manual && self.ease_factor == 0
+    }
+
+    /// Returns true if this entry represents a cramming operation.
+    /// These entries are created when a card is previewed using
```
user1823:
```suggestion
    /// These entries are created when a card is reviewed in a
    /// filtered deck with "Reschedule cards based on my answers
    /// in this deck" disabled.
```

rslib/src/revlog/mod.rs
```diff
@@ -84,6 +84,40 @@ impl RevlogEntry {
...
+    /// [`crate::scheduler::answering::CardStateUpdater::apply_preview_state`].
+    /// The `ease_factor` should be 0 because
```
user1823:
```suggestion
    /// The 0 value of `ease_factor` distinguishes it from the entry
```

rslib/src/revlog/mod.rs
```diff
@@ -84,6 +84,40 @@ impl RevlogEntry {
...
+    /// [`crate::scheduler::states::ReviewState::revlog_kind`] returns
```
user1823:
```suggestion
    /// created when a card is reviewed before its due date in a
```

rslib/src/revlog/mod.rs
```diff
@@ -84,6 +84,40 @@ impl RevlogEntry {
...
+    /// `RevlogReviewKind::Filtered` when `days_late() < 0`.
```
user1823:
```suggestion
    /// filtered deck with reschedule enabled or using Grade Now.
```

dae:
Thank you both. I'm still a bit worried about the extra full syncs this may cause, but I can't see a better way to handle this right now. Bumping usn without mtime won't help - the sync protocol will still treat that as newer, resulting in other changes being overwritten, so we still need the full sync so that users are aware of the conflict.

PR #4246, opened 2025-08-04
Fix show_exception's messagebox always formatting as plaintext

iamllama:
Fixes https://forums.ankiweb.net/t/anki-25-08-beta-3/64738/25

The [default behaviour](https://doc.qt.io/qt-6/qmessagebox.html#textFormat-prop), which this pr proposes to use, is to format text as richtext if detected as such, and plaintext otherwise

### DIFF
--- a/qt/aqt/errors.py
+++ b/qt/aqt/errors.py
@@ -179,7 +179,6 @@ def _init_message_box(
     _mbox.setWindowTitle("Anki")
     _mbox.setText(user_text)
     _mbox.setIcon(QMessageBox.Icon.Warning)
-    _mbox.setTextFormat(Qt.TextFormat.PlainText)
 
     def show_help():
         openHelp(help_page)


dae:
We've been bitten by Qt's heuristics not doing what we expected in the past, so in recent years I try to be explicit about the rendering mode instead of having it auto-detected. This change here may fix this particular case, but might cause other messages to be rendered incorrectly if they're assumed to be HTML when they're not.

_backend.py:249 is where this error gets converted from markdown (which looks ok in plain text as well) to HTML. Currently the search error is the only one that's handled this way. Maybe an alternative solution would be to stop converting to HTML there, and only do so in the browser when showing search errors? Other ideas welcome too.

iamllama:
> We've been bitten by Qt's heuristics not doing what we expected in the past, so in recent years I try to be explicit about the rendering mode instead of having it auto-detected.

We're already doing a lot of that whenever `showWarning`, `showCritical`, and `showInfo` are called without specifying the textformat though. Grepping shows 104 calls, with only 5 forcing a particular format and thus not relying on qt's heuristic

EDIT: ive implemented the suggestion so that the browser converts it to markdown before showing the error instead. This it how it looks in plaintext when triggered from the filtered deck dialog:
image

dae:
The more modern show_warning() etc variants do default to text-only mode, but agreed we still have a lot of code using the old approach. We'll likely break more things if we do a naive search&replace though, so we'd need to update them on a case-by-case basis. At least our errors are consistent now. :-)

PR #4248, opened 2025-08-05
Fix simulator graph not visible when viewport height < 400px

iamllama:
ref: https://forums.ankiweb.net/t/landscape-mode-doesnt-work-with-simulator/65089

Its max height is currently `calc(100vh - 400px)`, which can go negative when on landscape mode. The fix proposed is to set a lower bound of 200px


dae:
Sounds good - thanks!

PR #4250, opened 2025-08-06
IO: change colours + use named colours

brishtibheja:
(see https://forums.ankiweb.net/t/poll-change-default-image-occlusion-mask-colours/65193/6 for poll)

I set active colour to `lightgray` and inactive to `royalblue`. The poll showed people preferred gray/blue more than yellow/red.

___

Also, changed the other colours from hex codes to special HTML names. I think that makes it feel less tech-y to people.


rslib/src/image_occlusion/notetype.css
```diff
@@ -1,10 +1,10 @@
...
-    --active-shape-color: #ff8e8e;
-    --inactive-shape-border: 1px #212121;
-    --active-shape-border: 1px #212121;
-    --highlight-shape-color: #ff8e8e00;
-    --highlight-shape-border: 1px #ff8e8e;
+    --inactive-shape-color: lightgray;
+    --active-shape-color: royalblue;
+    --inactive-shape-border: 1px black;
+    --active-shape-border: 1px black;
+    --highlight-shape-color: transparent;
```
brishtibheja, line 6: this used to show the usage of alpha values but it was probably never clear to users, let's just document how to set the opacity.

rslib/src/image_occlusion/notetype.css
```diff
@@ -1,10 +1,10 @@
 #image-occlusion-canvas {
-    --inactive-shape-color: #ffeba2;
-    --active-shape-color: #ff8e8e;
-    --inactive-shape-border: 1px #212121;
-    --active-shape-border: 1px #212121;
-    --highlight-shape-color: #ff8e8e00;
-    --highlight-shape-border: 1px #ff8e8e;
+    --inactive-shape-color: lightgray;
+    --active-shape-color: royalblue;
```
solusyus: I suggest using `dodgerblue` instead of `royalblue`. It's less muddy and provides better contrast with `lightgray`.
I'll provide an image so that others can compare these two:

The left one is the `dodgerblue` and the right one is the `royalblue`.

dae:
(no objections on my end, and just waiting for color decisions to be finalized)

rslib/src/image_occlusion/notetype.css
brishtibheja: right one is prettier actually (soothing), but right is brighter so went with it.

solusyus (approved):

glutanimate:
Thanks for this proposal @brishtibheja. I left some [thoughts](https://forums.ankiweb.net/t/poll-change-default-image-occlusion-mask-colours/65193/15) on past and new color choices over at the forum thread.

A couple of additional implementation thoughts, should we end up changing the colors:

- As you raised in the thread, we would have to figure something out in terms of either keeping the mask editor in sync with template colors or consciously deciding that it is OK to have a desync between what users see in the editor and what they get when reviewing cards. Depending on which colors we end up with, this could be smaller or larger, which could factor into our decision here. The editor mask colors are currently hardcoded here: https://github.com/ankitects/anki/blob/5462d99255c41ed47b7953ca6487594ba7cabd5d/ts/routes/image-occlusion/tools/lib.ts#L10-L11
- We also have color fallbacks in place that take effect when the card template does not contain the corresponding IO color variables. We'll likely want to update these as well: https://github.com/ankitects/anki/blob/5462d99255c41ed47b7953ca6487594ba7cabd5d/ts/routes/image-occlusion/review.ts#L466-L480 One small caveat here: This could be one scenario where our changes actualy could affect how existing cards are rendered (e.g. a user having an invalid IO card template, using current defaults, would see different colors past the next Anki update). But this seems very edge-casey, so I think OK to ignore.

iamllama:
> As you raised in the thread, we would have to figure something out in terms of either keeping the mask editor in sync with template colors

It shouldn't be too involved, but the changes that i reckon will need to be made for this to work on desktop may likely conflict with Abdo's editor pr

dae:
Someone pointed out on the forums how this would be disruptive at the moment as the editor colours aren't controlled by the template.

https://forums.ankiweb.net/t/poll-change-default-image-occlusion-mask-colours/65193/28?u=dae

brishtibheja:
closing this PR

@dae Can I ask you to add an issue so that it doesn't get lost? Editor not reading the template is really a big issue IMO.

PR #4251, opened 2025-08-06
Feat/Neaten dr graph x-axis

Luc-Mcgrady:
image

Adds a percentage symbol and removes "plotted on x-axis" from the setting list.

ref: 
- https://github.com/ankitects/anki/pull/4199#issuecomment-3126222254


dae:
Thanks as always Luc!

PR #4252, opened 2025-08-06
Use space-around for tabbed values

Luc-Mcgrady:
ref:
- https://github.com/ankitects/anki/pull/4194#discussion_r2214379398

before/after
image
image

### DIFF
--- a/ts/routes/deck-options/TabbedValue.svelte
+++ b/ts/routes/deck-options/TabbedValue.svelte
@@ -55,6 +55,9 @@
         width: 100%;
         display: flex;
         flex-wrap: nowrap;
+        &:has(li:nth-child(3)) {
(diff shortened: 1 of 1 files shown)


ts/routes/deck-options/TabbedValue.svelte
```diff
@@ -55,7 +55,7 @@
         width: 100%;
         display: flex;
         flex-wrap: nowrap;
-        justify-content: space-between;
+        justify-content: space-around;
```
Luc-Mcgrady, line 61:
```suggestion
        justify-content: space-evenly;
```
space-around/space-evenly

image
image

brishtibheja: around looks better to me

PR #4253, opened 2025-08-07
Fix/Retention help button bounds

Luc-Mcgrady:
Forum link: https://forums.ankiweb.net/t/retention-stat-click-area/64220?u=a_blokee

Restricts the area which you click to activate the help button to the (?) circular question mark (as opposed to the entire title).

### DIFF
--- a/ts/lib/components/TitledContainer.svelte
+++ b/ts/lib/components/TitledContainer.svelte
@@ -12,7 +12,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     export { className as class };
 
     export let title: string;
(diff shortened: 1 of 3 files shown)


abdnh:
This broke help modals in Deck Options. I noticed the issue while reviewing this PR but disabling HMR made it momentarily go away so I dismissed it. This is the error:
> Uncaught TypeError: Cannot read properties of undefined (reading 'show')

For some reason, I can only reproduce this in dev build. The [latest prerelease](https://github.com/ankitects/anki/releases/tag/25.08b5) doesn't have the issue.

Svelte bug? 🤔

dae:
For me, the beta seems to be broken too. Reverting for now.

PR #4262, opened 2025-08-10
Hi res microhpone icon for the recording dialog

Madoshakalaka:
The recording icon is very choppy:

image

I replaced it with an svg microphone icon and preserved the original dimensions of 60x60

Screenshot From 2025-08-11 01-09-42

The icon comes from fontawesome which has a [really permissive licence](https://fontawesome.com/license/free) that allows it to be used in any commercial project as long as we preserve the comment string in the svg file


abdnh (approved):

dae:
Thank you!

PR #4264, opened 2025-08-12
A small clarification in deck-config.ftl

Expertium:
Just in case someone is not familiar with the term "leech"


brishtibheja:
Just don't use the term leech, instead say "cards that you fail repeatedly".

PR #4267, opened 2025-08-13
Replace activeWindow() with activeModalWidget()

abdnh:
https://forums.ankiweb.net/t/the-x-in-a-minimized-deck-options-window-doesnt-close-it/65561

Confirmed to work on Linux and macOS as well.


dae:
Wasn't aware of that one - nice!

PR #4269, opened 2025-08-15
Make simulator fill missing values of DR and decay too

user1823:
If a card has missing memory states, it will likely have missing DR and decay too. So, it makes sense to simultaneously update them as well.


dae:
@user1823 thank you, seems reasonable

@Luc-Mcgrady I missed the fact that you're bumping the USN here. I'll update this so we bypass that like in #4236

dae:
Thanks @user1823! If possible, it would be a little more convenient if you could pick a new branch name each time instead of always using patch-1.

PR #4270, opened 2025-08-15
Change FSRS button placement

Luc-Mcgrady:
People seem to prefer something like this. 

Forum post: https://forums.ankiweb.net/t/move-optimize-current-preset-on-the-same-level-as-optimize-all-presets/65740?u=a_blokee

image
image
image
image

### DIFF
--- a/ts/routes/deck-options/FsrsOptions.svelte
+++ b/ts/routes/deck-options/FsrsOptions.svelte
@@ -436,17 +436,18 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                 {/if}
             </button>
         {/if}
-        <div style:margin-left="0.5rem">
-            {#if computingParams || checkingParams}
-                {computeParamsProgressString}
-            {:else if totalReviews !== undefined}
-                {tr.statisticsReviews({ reviews: totalReviews })}
-            {/if}
-        </div>
 
         <button class="btn btn-primary float-right" on:click={() => computeAllParams()}>
             {tr.deckConfigSaveAndOptimize()}
         </button>
+    </div>
+
+    <div style:min-height="1.5em">
+        {#if computingParams || checkingParams}
+            {computeParamsProgressString}
+        {:else if totalReviews !== undefined}
+            {tr.statisticsReviews({ reviews: totalReviews })}
+        {/if}
     </div>
 
     <Warning warning={lastOptimizationWarning} className="alert-warning" />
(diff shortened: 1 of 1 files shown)


Expertium:
I made that post because I thought that "Optimize Current Preset" and "Optimize All Presets" not being on the same level is ugly. But this is even more ugly when optimization is running 😭
Can you move the optimization text (reviews and %) below the optimize buttons? Like this

Optimize Current Preset
Optimize All Presets
29.1% out of 197,948 reviews

brishtibheja:
In the forums, I suggested a checkbox (or a toggle button like health check) labelled as "optimise only this preset".

Luc-Mcgrady:
> Optimize Current Preset
Optimize All Presets
29.1% out of 197,948 reviews

I'm assuming the first 2 are supposed to remain on the same line?

image

image

image

Personally, I think these options are fine as they are in Anki currently. People want "A change" as indicated by the poll but regarding if it's a change for the better I don't have any strong opinions.

image

I also leave a gap to prevent the layout shift when the text appears.

Expertium:
@Luc-Mcgrady can you make it like this?
image
Some distance between the two buttons. Just enough to avoid misclicks

Expertium:
@Luc-Mcgrady "thumbs down" as in "I can't" or as in "I can, but I don't think it's a good idea"?

Luc-Mcgrady:
I think that if the objective of this pr is to make it more visually appealing then that change would be counter productive.

Expertium:
@dae according to the poll, most people are in favor of this change. But it's unclear how to make it work on mobile devices, and there is still disagreement on the finer details
image
Your opinion is welcome
(well, ultimately your opinion is what matters when merging or not merging this PR)

dae:
Personally, I mildly prefer the existing vertical layout:
- Both buttons start with "Optimize", which the vertical layout makes easier to see at a glance
- It looks better on narrow screens, especially after you changed my earlier terser wording to add "Preset"
I'll freely admit I'm not great with UI, so my opinions here may be in the minority.

That said, I find this PR and ones like it to be rather frustrating. I feel like we spend too much time discussing and making changes which barely move the usability/aesthetic needle, and we spend too much time changing things that (while perhaps not ideal) aren't broken.

Expertium:
If you want a meaningful change, there is no short supply of ideas here: https://forums.ankiweb.net/t/how-to-prevent-users-from-misusing-hard-ideas-are-welcome/49092

Anyway, feel free to close this PR. I don't have very strong feelings on this, unlike the Hard misuse stuff

Luc-Mcgrady:
I will try to avoid PR's like this in the future. I didn't like the change myself.

Issue #4263, opened 2025-08-11
[BUG] Learning cards are shown as buried cards on deck overview page


user1823:
Steps to reproduce:
- Study some new cards. Don't complete the reviews yet. Exit the reviewer.
- Wait for the newly introduced cards to become due again (the next learning step)
- Click on the deck. The Decks overview page shows the learning cards as buried, but they are not buried.

<img width="240" src="https://github.com/user-attachments/assets/e1e37e63-d6d0-4b21-b29b-26e6edce95f3"><br><img width="145" src="https://github.com/user-attachments/assets/5dd76e20-d4c3-4d6e-815b-e357cc476946">

None of these cards is buried.

Closing and relaunching Anki resolves the issue. (This also means that sharing an apkg for reproduction isn't possible.)

<img width="115" src="https://github.com/user-attachments/assets/fd53deb0-1d1d-42c8-b694-bfdfe294744e" />

Anki 25.08 beta5

First reported in
- https://github.com/ankidroid/Anki-Android/issues/17110

Relevant code:
https://github.com/ankitects/anki/blob/f3b4284afbb38b894164cd4de3e7b690f2bc62a5/qt/aqt/overview.py#L228-L241

https://github.com/ankitects/anki/blob/f3b4284afbb38b894164cd4de3e7b690f2bc62a5/rslib/src/scheduler/queue/mod.rs#L88-L137

As the deck overview page shows 5 buried learning cards, it means that the `get_queued_cards` function in the Rust backend is returning `0` as the `learning_count` for some reason.

AmandaSternberg-creator:
Hi! I’d like to work on this issue. Is it available?

dae:
Do you have a plan for how to solve this without the problem mentioned in the previous PR? If not, and you're more looking for something to get started with open source contributions, this might not be the best repo for a first contribution.

AmandaSternberg-creator:
Hi! Thanks for the feedback. I understand that this issue may be tricky, especially since a previous PR attempted to solve it but introduced other problems. I am willing to go through that PR in detail, try to understand why it did not work out and explore a possible solution. If you think this is reasonable I would be happy to give it a try, otherwise I would appreciate any suggestions for another issue I could take on instead?

PR #4271, opened 2025-08-16
Fix learning cards incorrectly shown as buried in deck overview

escribapetrus:
Addresses #4263 

When learning cards become due while queue counts are cached, they were incorrectly shown as 'buried' in the deck overview. This happened because the counts() method only updated the learning cutoff when all counts were zero, missing newly due learning cards.

- Modified counts() to also check for newly due learning cards
- Added has_newly_due_learning_cards() to detect timing changes
- Added test case to prevent regression

Fixes issue where learning cards appear buried after exiting reviewer and waiting for next learning step, resolved by restarting Anki.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/rslib/src/scheduler/queue/mod.rs
+++ b/rslib/src/scheduler/queue/mod.rs
@@ -184,15 +184,32 @@ impl CardQueues {
         }
     }
 
-    /// Return the current due counts. If there are no due cards, the learning
-    /// cutoff is updated to the current time first, and any newly-due learning
-    /// cards are added to the counts.
+    /// Return the current due counts. If there are no due cards or if learning
+    /// cards have become due since the last cutoff update, the learning cutoff
+    /// is updated to the current time first, and any newly-due learning cards
+    /// are added to the counts.
     pub(crate) fn counts(&mut self) -> Counts {
-        if self.counts.all_zero() {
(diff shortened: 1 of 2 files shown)


user1823:
What if we just remove the `if` condition and make it something like this?

```rs
    pub(crate) fn counts(&mut self) -> Counts {
        self.update_learning_cutoff_and_count();
        self.counts
    }
```

Also, this comment must be updated:
https://github.com/ankitects/anki/blob/f3b4284afbb38b894164cd4de3e7b690f2bc62a5/rslib/src/scheduler/queue/mod.rs#L37-L39

user1823:
I am not sure how this change affects the inclusion/exclusion from the counts of those learning cards that aren't due yet, but are due within the learn_ahead limit. Such learning cards should be added to the counts only when no other card is due, but this PR is altering/removing that check. So, proper testing is required.

dae:
Thank you for identifying the cause of that bug!

Unfortunately, I fear this approach may have unintended side-effects. If my memory serves me correctly, the reason the current code avoids fetching newly-due learning cards until the counts hit zero is to avoid the current card changing. For example, imagine the user has a review card shown in the review screen, and they edit it. When they return to study, they expect the same card to appear. If we're always looking for newly-due cards, I suspect that would result in the newly-due card appearing instead.

Issue #4085, opened 2025-06-14
Consider a user-friendly 'streak'


dae:
Some users like to use third-party add-ons to track their 'streak' to provide motivation/keep them honest. I've resisted adding such a feature up until now, because the existing implementations can be toxic - users get overly disappointed when they miss a day for some reason, and feel inclined to do things like set the date back in time and/or do single-card reviews to maintain their streak.

I think a better approach would be something that averages performance out over time, and allows for low/off days. Preferably where the user decides the bar they want to set, perhaps guided by their daily workload amount. And maybe a rolling average, so that bad past performance doesn't lead to users getting stuck in a hole?

Alhaj-nit:
Hi @dae,

I'm a student contributor working on a university assignment. I'd like to experiment with a friendly streak tracker prototype, based on a 7-day rolling average with a daily review goal set by the user (e.g. 50 cards/day).

Would it be helpful if I created a small side-widget (possibly Qt) that summarizes "6 of 7 days met", with customizable goal?

Thanks!

dae:
If this is implemented, it wouldn't be in Qt, as we'd want it in the mobile clients too. But we need to discuss it first: https://github.com/ankitects/anki/pull/4088#issuecomment-2982709761

Gardengul:
You can see what other habit tracker apps offer instead of a streak.

https://github.com/iSoron/uhabits/discussions/689
<details>
  <summary>Example</summary>

> How is the score calculated?
To compute the score of a habit, the app uses a statistical method known as exponential smoothing. Basically, it computes a weighted average that takes into consideration every repetition of the habit, from the very first day you started your habit until today. Recent repetitions, however, are considered more important than older ones, and have a larger impact on the score. This method has many nice features, including:
Every repetition counts, even way back in the past. This is not true for other simpler formulas, such as counting how many times have you performed the habit in the last week/month/year and then dividing by the number of days in that interval.
If you have a poor habit score, then a few repetitions can bring your score up very quickly. As your score improves, however, the reward for each repetition gets smaller and smaller, so you have to keep at it if you want to see further increases.
If you have a high score for a very long time and you take a short break, it's quite easy to restore your score back to what it was. If you start taking frequent breaks, however, then your score will suffer.
How long will it take to reach the highest score?
[...]

PR #4273, opened 2025-08-17
Congratulations Screen with 7-Day future due graph

thomasrixen:
I felt that the current congratulations screen was a bit too minimal and didn’t provide much useful feedback after finishing a session. To make it more informative and visual, I implemented a 7-day review forecast and integrated it directly into the congrats page.

Here was what it looked like before:
Screenshot from 2025-08-17 23-24-17

and here is with the graph:
Screenshot from 2025-08-17 23-27-28

This gives users a quick overview of what’s ahead, and helps to plan study habits more effectively.

I had shared the idea earlier on the [forum](https://forums.ankiweb.net/t/enhanced-congratulations-screen-with-7-day-review-forecast/65610), but there wasn’t much feedback, so I decided to go ahead and try implementing it. I’d love to hear your thoughts. Is this something you see as useful, or should it be adjusted before being considered for merging?

### DIFF
--- a/rslib/src/scheduler/mod.rs
+++ b/rslib/src/scheduler/mod.rs
@@ -31,6 +31,15 @@ pub struct SchedulerInfo {
     pub timing: SchedTimingToday,
 }
 
+#[derive(Debug, Clone)]
+pub struct ReviewForecastDay {
+    pub day_offset: u32,
+    pub total: u32,
+    pub review: u32,
+    pub learn: u32,
+    pub new: u32,
+}
+
 impl Collection {
     pub fn scheduler_info(&mut self) -> Result<SchedulerInfo> {
         let now = TimestampSecs::now();
@@ -131,5 +140,37 @@ impl Collection {
     pub(crate) fn set_creation_stamp(&mut self, stamp: TimestampSecs) -> Result<()> {
         self.state.scheduler_info = None;
         self.storage.set_creation_stamp(stamp)
+    }
+
+    /// Return forecast data for the next `days` days (capped at 7).
+    pub(crate) fn sched_forecast(
+        &mut self,
+        days: u32,
+    ) -> Result<Vec<anki_proto::scheduler::ReviewForecastDay>> {
+        use anki_proto::scheduler::ReviewForecastDay as PbDay;
+        let timing = self.timing_for_timestamp(TimestampSecs::now())?;
+        let today = timing.days_elapsed;
+        let mut out = Vec::new();
+        let want = days.min(7);
+
+        for offset in 0..want {
+            let target_day = today + offset;
+            let rev_cnt = self
+                .storage
+                .db
+                .prepare_cached("SELECT COUNT(*) FROM cards WHERE queue = 2 AND due = ?")
+                .and_then(|mut stmt| {
+                    stmt.query_row([(target_day as i64)], |row| row.get::<_, u32>(0))
+                })
+                .unwrap_or(0);
+            out.push(PbDay {
+                day_offset: offset,
+                total: rev_cnt,
+                review: rev_cnt,
+                learn: 0,
+                new: 0,
+            });
+        }
+        Ok(out)
     }
 }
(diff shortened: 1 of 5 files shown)


thomasrixen:
It looks like every pull request shows the same error, _"Translations" has no attribute "preferences_color_blind"_. I’m not sure how to make the tests pass.

brishtibheja:
The stats page already shows data like this, so I'm a bit concerned about duplication. But more than that, I'm not sure future due is something people would find helpful on the congrats page?

I've seen some other app(s) where at the end of study session you're shown stats regarding **that very session.** That feels like a direction we can try to investigate. But again, Anki shows "Today" stats on the stats screen, so either:

1. We duplicate the info.
2. We keep it at only at one place and remove it from stats screen.

___

Apart from what Anki already has, I have some ideas for charts that we can show here:

1. Breakdown of how fast/slow you were per-time.
2. Breakdown of pass-rate per-time.
3. Breakdown of per-card times for every card state.

thomasrixen:
The congratulations screen is the last thing a user sees after finishing their reviews and closing the app. That makes it an important moment: right now it only displays text, which in my experience is easy to ignore.

I think this page could be made more engaging by adding some visualizations. In particular, showing the short-term review schedule (e.g., the upcoming week) gives the user a sense of when they should come back, which can help with motivation.

While the statistics screen already includes future due information, my intention with this change is different: it’s about making the congratulations screen more attractive and meaningful, rather than duplicating the stats. Long-term numbers like “reviews in 55 days” are less relevant here. The focus is on the immediate next steps and encouraging users to return soon.

brishtibheja:
Makes sense, but then you should probably just add the stat about the next day with non-zero reviews due instead of a graph like here.

We can still add graphs for nicing this page up, but future due like you did in this PR seems unnecessary.

thomasrixen:
I see, thanks for the clarification. Just to make sure I understand correctly: you’re suggesting that instead of showing a graph of future due cards, I should only display a simple stat about the next day with non-zero reviews due.

For the graphs, do you have specific ideas of what kind of visualizations would add value to this page, aside from the next-day stat?

ZornHadNoChoice:
If streaks (https://github.com/ankitects/anki/issues/4085) get implemented then something related to them might fit in the congratulations screen.

sommerluk:
If showing statistics here, I would suggest using the same color palette as in the statistics page, so that everything looks more familar.

dae:
I appreciate you taking the time to contribute Thomas, but I'm afraid I'm not super-keen on this change at this time. I fear it will lead to users wanting to hide the graph, alter it, show other graphs, etc, and we have our hands full with other things at the moment. There's plenty of scope to improve this screen, but I'd like to see us in a better technical place before we start making more tweaks here, and the review/congrats flow should probably be on a single page in the future.

A simple text message that says the next time a card will become due is easier to wave through, though I wonder how useful it would be for the typical user who has cards due every day.

rslib/src/scheduler/mod.rs
```diff
@@ -132,4 +141,36 @@ impl Collection {
...
+        let today = timing.days_elapsed;
+        let mut out = Vec::new();
+        let want = days.min(7);
+
+        for offset in 0..want {
+            let target_day = today + offset;
+            let rev_cnt = self
+                .storage
+                .db
+                .prepare_cached("SELECT COUNT(*) FROM cards WHERE queue = 2 AND due = ?")
```
dae, line 161: For future reference, SQL should generally live in src/storage, with the collection method calling into the relevant storage function.

ts/routes/congrats/CongratsPage.svelte
```diff
@@ -77,6 +88,16 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
                     {@html info.deckDescription}
                 </div>
             {/if}
+
+            {#if forecastData.length > 0 && forecastData.some((d) => d.total > 0)}
+                <div class="graph-section">
+                    <p class="graph-description">
+                        Below is your study forecast for the upcoming week. This shows
```
dae, line 95: Also for future reference, text would need to be made translatable.

PR #4275, opened 2025-08-18
Fix/Simulator intervals decending overflows

Luc-Mcgrady:
Maybe the new cards having -infinity as an interval has something to do with the root cause? I think this fixes it regardless. Anki should panic when using descending intervals before this pr:

image


dae:
saturating_neg() would a better way to handle this

dae:
Thanks Luc!

Issue #4226, opened 2025-07-24
Provide a simpler way for users to enable a pypi mirror?


dae:
An env var or config file would be easier than users needing to manually modify files.

> 以下内容来自知乎 沢见荧的文章
> Anki 新版启动器（Anki Launcher）更新下载缓慢解决办法（uv 镜像设置） - 沢见荧的文章 - 知乎
> https://zhuanlan.zhihu.com/p/1927728002831216855
> 
> 在地址栏输入并打开 %APPDATA% 目录，或打开 C:\Users(你的用户名)\AppData\Roaming 目录
> 创建 uv 目录（有的话不用创建）并打开
> 创建 uv.toml 文件（有的话不用创建）
> 打开文件添加以下内容并保存：
> 
> python-install-mirror = "https://registry.npmmirror.com/-/binary/python-build-standalone/"
> [[index]]
> url = "https://mirrors.tuna.tsinghua.edu.cn/pypi/web/simple/"
> default = true
> 
> 即可成功设置镜像。

Originally reported on https://forums.ankiweb.net/t/anki-anki-launcher-uv/64485

jdorn-eu:
If there are a handful of known good mirrors, the launcher could check latency to them at startup and automatically choose the one located closest to the user. I assume packages from pip are cryptographically signed (?)

cloudfish:
It appears to me that this workaround is no longer working, probably because of the `--no-config` option.

```
Anki Launcher

Checking for updates...
Unable to check for Anki versions. Please check your internet connection.

Error: Failed to run (2): /home/xxxxx/anki-launcher-25.07.5-linux/uv.amd64 run --no-project --no-config --managed-python --with pip-system-certs --python 3.13.5 /home/xxxxx/anki-launcher-25.07.5-linux/versions.py: error: Failed to download https://github.com/astral-sh/python-build-standalone/releases/download/20250612/cpython-3.13.5%2B20250612-x86_64-unknown-linux-gnu-install_only_stripped.tar.gz
  Caused by: Request failed after 3 retries
  Caused by: error sending request for url (https://github.com/astral-sh/python-build-standalone/releases/download/20250612/cpython-3.13.5%2B20250612-x86_64-unknown-linux-gnu-install_only_stripped.tar.gz)
  Caused by: operation timed out

Press enter to close...
```

dae:
[...]

PR #4280, opened 2025-08-19
Fix mirror configuration not working during launcher download

aldlss:
I’m the author of the post referenced in #4226. Thanks for noticing and addressing this issue.

However, the current fix doesn’t work on my side, likely because:

+ Both `uv sync` https://github.com/ankitects/anki/blob/68bc4c02cfa87356e74c04cf025a28d590f55539/qt/launcher/src/main.rs#L301  and `uv run versions.py` 
https://github.com/ankitects/anki/blob/68bc4c02cfa87356e74c04cf025a28d590f55539/qt/launcher/src/main.rs#L663 are invoked with `--no-config`, so the mirror settings in the config file are ignored (main cause).
+ `uv run versions.py` downloads the Python interpreter before the mirror settings have been written to the config file.

Using environment variables to configure the mirror would work better in this flow. I’ve tested this locally and can confirm it resolves the issue for me.


dae:
Thank you for looking into this. Unfortunately, there's a conflict on the CONTRIBUTORS file. Could I trouble you to rebase over the latest main, so I can merge this in?

aldlss:
> Thank you for looking into this. Unfortunately, there's a conflict on the CONTRIBUTORS file. Could I trouble you to rebase over the latest main, so I can merge this in?

Rebased onto the latest main and resolved the CONTRIBUTORS conflict. Thanks!

dae:
Thank you!

PR #4281, opened 2025-08-19
Fix sync errors not being reported in some cases

abdnh:
The lack of an explicit parent arg for show_waring() in handle_sync_error() was causing the dialog to be closed immediately in case the parent happens to be the progress dialog spawned by taskman.with_progress(). I was able to confirm this by triggering an error in col.sync_collection().

This should fix a commonly reported issue in AnkiHub, where many new users (that probably didn't confirm their emails yet) reported AnkiHub showing "AnkiHub sync cancelled" at startup (the add-on hooks AnkiWeb sync to handle cases where a full sync is triggered by AnkiHub changes).

Related PR from last year: https://github.com/ankitects/anki/pull/3222 (I noticed fewer reports after this one, so I assume it did fix the issue for many users).

### DIFF
--- a/qt/aqt/sync.py
+++ b/qt/aqt/sync.py
@@ -73,7 +73,7 @@ def handle_sync_error(mw: aqt.main.AnkiQt, err: Exception) -> None:
     elif isinstance(err, Interrupted):
         # no message to show
         return
-    show_warning(str(err))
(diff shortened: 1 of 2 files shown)


dae:
Nice, thank you. We've dealt with a number of similar issues over the years.

PR #4283, opened 2025-08-19
Fix/Workload simulator config deck_size unset

Luc-Mcgrady:
fixes https://github.com/open-spaced-repetition/fsrs-rs/issues/359


TheBB:
Thanks, can confirm this fixes the issue.

PR #4284, opened 2025-08-20
Fix LRT database check for cards with no usable reviews

user1823:
Fixes https://forums.ankiweb.net/t/anki-25-08-beta-3/64738/62

### DIFF
--- a/rslib/src/storage/card/mod.rs
+++ b/rslib/src/storage/card/mod.rs
@@ -403,7 +403,9 @@ impl super::SqliteStorage {
         let last_revlog_info = get_last_revlog_info(&revlog);
         for (card_id, last_revlog_info) in last_revlog_info {
             let card = self.get_card(card_id)?;
-            if let Some(mut card) = card {
+            if last_revlog_info.last_reviewed_at.is_none() {
+                continue;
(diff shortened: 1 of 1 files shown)


user1823:
Here's a more thorough fix, but I am not sure if it would break something.

```rs
        for (card_id, last_revlog_info) in last_revlog_info {
            let card = self.get_card(card_id)?;
            if let Some(mut card) = card {
                if card.ctype != CardType::New && card.last_review_time.is_none() {
                    if last_revlog_info.last_reviewed_at.is_some() {
                        card.last_review_time = last_revlog_info.last_reviewed_at;
                        self.update_card(&card)?;
                        last_review_time_cnt += 1;
                    } else if card.reps == 0 {
                        let mut position = col.get_next_card_position();
                        card.schedule_as_new(position, True, True);
                        self.update_card(&card)?;
                        position += 1;
                        other_cnt += 1;
                    }
                }
            }
        }
```

Edit: This code will cause unexpected behaviour when the user introduces cards by using `Set Due Date`. Introducing cards with SDD is not recommended when using FSRS. But still, we shouldn't add a check like this as many users don't know the best practice.

dae:
Thank you!

PR #4286, opened 2025-08-24
Increase randomness in random sorting of new cards

user1823:
Currently, the new cards appear roughly in the same order on consecutive days (if they are skipped by burying). This change aims to increase randomness by spreading out the salt across the hash space.


dae:
Seems reasonable - thank you!

PR #4291, opened 2025-08-26
Fix ninja BuildAction name sanitization

memchr:
rust commit [8296ad0](https://github.com/rust-lang/rust/commit/8296ad04568597fbfab4460530032cd54f5cd26b#diff-d43cdd86cb96902ecb316c5956a0769c5de03808dd81512c6d916697c9ab058eR24) changes the output of std::any::type_name to include
regions such as lifetime and generic arguments, which results in invalid
Ninja rule names being generated, such as `CargoBuild<_>`.

This commit removes the substring starting at (and including) the first `<`.


dae:
Thank you for this!

PR #4292, opened 2025-08-27
Add desired_retention field to NormalDeckSchema11

L-M-Sherlock:
To allow add-ons to read the deck-specific desired retention, I add `desired_retention` to `NormalDeckSchema11`.
- https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/602

But `f32` is not allowed due to the trait `Eq`, so I convert it to `u32`. Is it the best practice?


dae:
Given that this is our legacy/compatibility layer, and it doesn't have any existing floating points, I think your approach of keeping things the same is probably best. Thanks Jarret.

Issue #4279, opened 2025-08-19
Margin above “Burying” is too big


dae:
> Screenshot_2025-08-13-23-12-52-169_com.ichi2.anki.a-edit1440×1566 136 KB

Originally reported on https://forums.ankiweb.net/t/margin-above-burying-is-too-big/65648

I suspect it's a byproduct of the two separate divs in DeckOptionsPage that's intended to cause the page to divide into two roughly-equal columns, and might be fixable?

PR #4298, opened 2025-08-30
FIX gap above bury

GithubAnon0000:
Fixes: #4279

The issue was that the `gap` was set to 20px by default, instead of only setting the gap when we use a dual-column layout.


dae:
Thank you for getting to the bottom of that one!

PR #4310, opened 2025-09-01
Show the number of cards added when adding

### LINKED ISSUE #4309: Show the number of cards added when adding
> For example, say I’m adding a new multi-cloze card “Anki was created in {{c1::2006}} by {{c2::Damien Elmes}}.” When I add the card, it displays ‘Added’, but I want it to say ‘Added 2 new cards’. The benefit of this is that I can double-check all the cloze was formatted correctly and will be added as expected. Sometimes I make cards with 10+ cloze so knowing the number would be a nice quality of life.
> Screenshot 2025-09-01 at 07.53.201532×461 36 KB
> Screenshot 2025-09-01 at 07.52.281514×912 57 KB

Originally reported on https://forums.ankiweb.net/t/new-multi-cloze-cards-should-display-number-of-new-cards-added-in-pop-up/66214

iamllama:
Closes #4309

image

This might require updating the corresponding fn in ankidroid's [libanki](https://github.com/ankidroid/Anki-Android/blob/b3e1cdfb6970d97a87e05381186ca4bbc1e49aed/libanki/src/main/java/com/ichi2/anki/libanki/Collection.kt#L654)

### DIFF
--- a/ftl/core/adding.ftl
+++ b/ftl/core/adding.ftl
@@ -1,5 +1,10 @@
 adding-add-shortcut-ctrlandenter = Add (shortcut: ctrl+enter)
 adding-added = Added
+adding-added-cards =
+    Added { $count ->
+        [one] { $count } card
+       *[other] { $count } cards
+    }
 adding-discard-current-input = Discard current input?
 adding-keep-editing = Keep Editing
 adding-edit = Edit "{ $val }"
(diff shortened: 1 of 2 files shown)


dae:
Can't fault it. Thank you! ❤️

ftl/core/adding.ftl
```diff
@@ -1,5 +1,10 @@
 adding-add-shortcut-ctrlandenter = Add (shortcut: ctrl+enter)
 adding-added = Added
+adding-added-cards =
```
abdnh, line 3: Minor note: importing-cards-added can be used here instead of adding a new message. It's currently used for image occlusion on mobile.

iamllama: I've opened #4316 with this change 👍

PR #4312, opened 2025-09-02
Enable nc: to only search in a specific field (#4276)

### LINKED ISSUE #4276: Enable `nc:` to only search in a specific field?
> Currently, when using `nc:` to search for words/characters of another language, the search is always global. In other words, the search is done on all fields contents. This is very useful, but often not precise enough to find the desired card.
> It would be great to have the ability to search in the no-combining way in a specific field only.

Originally reported on https://forums.ankiweb.net/t/enable-nc-to-only-search-in-a-specific-field/65489

maxr777:
Closes #4276

image
image
image
image

### DIFF
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -3,6 +3,7 @@
 
 use std::sync::LazyLock;
 
+use anki_proto::search::search_node::FieldSearchMode as FieldSearchModeProto;
 use nom::branch::alt;
 use nom::bytes::complete::escaped;
 use nom::bytes::complete::is_not;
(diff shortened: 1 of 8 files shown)


maxr777:
Different language:

image

rslib/src/search/parser.rs
```diff
@@ -57,6 +57,7 @@ pub enum SearchNode {
         field: String,
         text: String,
         is_re: bool,
+        is_nc: bool,
```
dae: Our data structure makes it appear like is_re=1 && is_nc=1 is possible, but we don't support that when parsing/writing. An enum field of normal/re/nc would better convey what we actually support.

dae:
Thank you for contributing Max! I've made a suggestion below, and if you could fix the conflict at that same time, that would be appreciated.

rslib/src/search/parser.rs
maxr777: I changed to an enum - this should be it unless I missed something. Ty for the feedback

dae:
Thank you!

PR #4317, opened 2025-09-03
Fix/Invalid memory states in simulator after parameters changed

### LINKED ISSUE #4064: Using FSRS simulator after updating params (without saving) uses old memory states
If I update the FSRS parameters (by optimization or manually) and then use the FSRS simulator before saving the preset options, the simulator uses the memory states saved in the cards (i.e., the states generated from the previous parameters).

Possible solutions:
- If the parameters have been updated, ask the user to save the preset options when they try to use the simulator.
- If the parameters are not the same as the saved parameters, don't use the memory states stored in the cards. Rather, calculate the memory states on the fly (but don't save them to the card).

Luc-Mcgrady:
fixes #4064

This will prevent users from running the simulator after optimizing, before saving the preset.

### DIFF
--- a/ftl/core/deck-config.ftl
+++ b/ftl/core/deck-config.ftl
@@ -498,6 +498,7 @@ deck-config-desired-retention-below-optimal = Your desired retention is below op
 # cards that can be recalled or retrieved on a specific date.
 deck-config-fsrs-simulator-experimental = FSRS Simulator (Experimental)
 deck-config-fsrs-simulate-desired-retention-experimental = FSRS Desired Retention Simulator (Experimental)
+deck-config-fsrs-simulate-save-preset = After optimizing, please save your config before running the simulator.
 deck-config-fsrs-desired-retention-help-me-decide-experimental = Help Me Decide (Experimental)
 deck-config-additional-new-cards-to-simulate = Additional new cards to simulate
 deck-config-simulate = Simulate
--- a/ts/routes/deck-options/FsrsOptions.svelte
+++ b/ts/routes/deck-options/FsrsOptions.svelte
@@ -340,9 +340,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     }
 
     function showSimulatorModal(modal: Modal) {
-        if (
-            fsrsParams($config).toString() === initialParams.toString()
-        ) {
+        if (fsrsParams($config).toString() === initialParams.toString()) {
             modal?.show();
         } else {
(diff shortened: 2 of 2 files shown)


ts/routes/deck-options/FsrsOptions.svelte
```diff
@@ -53,6 +53,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     let desiredRetentionFocused = false;
     let desiredRetentionEverFocused = false;
     let optimized = false;
+    const initialParmas = [...fsrsParams($config)];
```
user1823: Typo: initialParams

ts/routes/deck-options/FsrsOptions.svelte
```diff
@@ -338,6 +339,17 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
         state.save(UpdateDeckConfigsMode.COMPUTE_ALL_PARAMS);
     }
 
+    function showSimulatorModal(modal: Modal) {
+        if (
+            initialParmas.length === 0 ||
```
user1823: If initial params are empty and current params are not, this will return `True`, which is not desired.

ftl/core/deck-config.ftl
```diff
@@ -498,6 +498,7 @@ deck-config-desired-retention-below-optimal = Your desired retention is below op
 # cards that can be recalled or retrieved on a specific date.
 deck-config-fsrs-simulator-experimental = FSRS Simulator (Experimental)
 deck-config-fsrs-simulate-desired-retention-experimental = FSRS Desired Retention Simulator (Experimental)
+deck-config-fsrs-simulate-save-preset = After optimizing, please save your config before running the simulator.
```
user1823, line 501:
```suggestion
deck-config-fsrs-simulate-save-preset = Please save the updated parameters before running the simulator.
```

user1823 (changes requested):

ts/routes/deck-options/FsrsOptions.svelte
Luc-Mcgrady:
```suggestion
```

ftl/core/deck-config.ftl
Luc-Mcgrady: I think its better if we keep the focus on saving the config, as there is no way to save the parameters separately.

dae:
Thanks!

PR #4318, opened 2025-09-03
Fix undo queue being cleared on every error

### LINKED ISSUE #4311: Filtered Deck errors break Undo/Redo
Tested on Anki 25.08b5, Windows 11. I was only able to reproduce the issue with Filtered Deck-related errors.

Steps to reproduce

1. Perform an action (e.g., add a new note).  
2. Trigger a *Filtered Deck* error. For example:  
   - Go to a non-filtered deck and press R (rebuild), triggering:  
     `This action can only be used on a filtered deck`  
   - Or create a filtered deck that matches 0 cards, triggering:  
     `No cards matched the provided search. Some cards may have been excluded because they are in a different filtered deck, or suspended.`  
3. Press `Ctrl+Z` or go to `Edit > Undo`.  
   - Expected result: The last action is undone.  
   - Actual result: Nothing happens.

mcostn:
fixes #4311 

This only discards the undo and study queues when the error is not a `FilteredDeckError`


dae:
Unfortunately, I'm not sure this is going to be correct in all cases. Consider the case where some operation has already mutated the database, and it then encounters a filtered deck where it expected a normal one. We'd end up rolling back the mutations, but leaving the undo state and study queues in an inconsistent state.

mcostn:
I see. I think I understand the problem. Are you saying that in the case a `FilteredDeckError` occurs after a mutation has been done to the database, the undo state and study queues won't be in the same state as the database?

In that case, can't I just move the rollback in the if?
```rust
if !matches!(err, AnkiError::FilteredDeckError { .. }) {
    self.discard_undo_and_study_queues();
    if autocommit {
        self.storage.rollback_trx()?;
    } else {
        self.storage.rollback_rust_trx()?;
    }
}
Err(err)
```

dae:
No, because then any changes made to the database prior to the error would not be rolled back. 

```
for deck in decks {
  if !deck.normal().is_ok() { return Err(...) }
  col.update_deck(deck)
}
```

PR #4320, opened 2025-09-04
Decode images asynchronously

abdnh:
This sets the `decoding` property for `img` elements to `async` to avoid blocking rendering of card content while images are being loaded, which is a common issue with notes with a lot of images or high resolutions. I found setting this noticeably reduces lag.


dae:
Interestingly, there are some reports that it can actually slow down the loading of large images:

https://github.com/vercel/next.js/discussions/50524

No personal experience with this, and no strong feelings. If you'd like to merge this so we can get feedback from beta testers, please feel free.

dae:
Sigh - I'd planned to update the custom color PR to use the collection prior to building the release, but managed to forget about it, so I'll need to build .2 now. Just wanted to check your intentions here: I presume you wanted this to be merged after the patch release, not before it?

abdnh:
> I presume you wanted this to be merged after the patch release, not before it?

Oh, I didn't notice you made a release actually. I wanted to merge it before that.

dae:
Ok 👍

jcznk:
It seems to perform slightly worse on my computer (Windows 11) when there are multiple images:

Anki 25.09rc1:

https://github.com/user-attachments/assets/b0b1cff8-68cb-4169-a670-50bc3b16fb25


Anki 25.09.2:

https://github.com/user-attachments/assets/0be9f483-f728-4c39-8b0f-adb05a24814d

abdnh:
@jcznk Did you notice any rendering speed improvements as you flip through cards with images?

jcznk:
@abdnh Yes, the cards do seem to load a bit faster, especially when they contain large pictures. But the images themselves often lag, and the overall effect is somewhat "flickery".

Chilaiditi:
Hello, I think this change is causing massive image flickering on cards that utilize CT/MRI scrolling.

dae:
I've created #4373

caleblee789:
I have noticed that images tend to load slower on cards that have a large number of images, but faster when there are only 1 or 2 images.

docdd6:
> This sets the `decoding` property for `img` elements to `async` to avoid blocking rendering of card content while images are being loaded, which is a common issue with notes with a lot of images or high resolutions. I found setting this noticeably reduces lag.

The images take too much time to load now on the new version

atonofclaygit:
Hi all, I've also noticed massive flickering on my cards that utilize CT/MRI scrolling. The javascript for the cards is based on the code graciously provided by user kleinerpirat in [https://forums.ankiweb.net/t/how-to-create-scrollable-image-gallery-for-ct-mris/28442](https://github.com/ankitects/anki/pull/url)

abdnh:
Tracked here: https://github.com/ankitects/anki/issues/4373

A potential fix was implemented but it's not released yet.

PR #4321, opened 2025-09-05
Updated Dockerfile to use Ninja build system

e-hamrin:
Hi! First PR so hope I get this right. 

## Changes
Updated the old Dockerfile that used Blaze build, to using Ninja. 

Wheels are not included in the final image to make it as small as possible.


docs/docker/Dockerfile
```diff
@@ -1,35 +1,76 @@
...
-RUN curl -fsSL https://github.com/bazelbuild/bazelisk/releases/download/v1.7.4/bazelisk-linux-amd64 \
-    > /usr/local/bin/bazel \
-    && chmod +x /usr/local/bin/bazel \
-    # Bazel expects /usr/bin/python
-    && ln -s /usr/local/bin/python /usr/bin/python
+FROM ubuntu:24.04 AS build
 WORKDIR /opt/anki
-COPY . .
-# Build python wheels.
+ENV PYTHON_VERSION="3.9"
```
abdnh: Recent Anki versions use Python 3.13: https://github.com/ankitects/anki/blob/main/.python-version

e-hamrin:
Hey @abdnh thanks for the input, missed that! 😀 I have updated the python version now, builds as expected.

dae:
Thank you both!

docs/docker/Dockerfile
```diff
@@ -1,35 +1,78 @@
...
+RUN uv python install ${PYTHON_VERSION} --default
+
+COPY . . 
+
 RUN ./tools/build
 
+
 # Install pre-compiled Anki.
-FROM python:${PYTHON_VERSION}-slim as installer
+FROM python:3.13-slim AS installer
```
MarkMoretto, line 65: I'd be interested to know more about hardcoding the vesion here vs. using  the PYTHON_VERSION environment variable.

PR #4323, opened 2025-09-06
Prevent clipping for QPushButton:default:hover

jcznk:
Should fix the issue reported at: https://github.com/ankitects/anki/pull/4201#issuecomment-3262286135

Before:

image

After:

image

<hr>

EDIT: Note that this change causes the border to "grow" outwards rather than inwards. Whether this constitutes an aesthetic regression is probably a matter of personal taste. I slightly prefer the border growing inwards, but as I understand it, this is necessary to prevent the clipping problem. (Unless, of course, we want to make more radical changes to the stylesheet, which I would not be against. I don’t particularly like the 2px border and could live without it entirely.)

The topic I took inspiration from (https://stackoverflow.com/questions/66637388/is-there-a-way-to-change-the-external-border-of-a-qpushbutton) was opened by a user explicitly looking for a way to have the border grow outwards. So, maybe, people like it that way. My suggestion is to test the changes and see how they are received, and adjust later if needed.

### DIFF
--- a/qt/aqt/stylesheets.py
+++ b/qt/aqt/stylesheets.py
@@ -180,7 +180,7 @@ class CustomStyles:
     QPushButton {{
         margin: 1px;
     }}
-    QPushButton:focus {{
+    QPushButton:focus, QPushButton:default:hover {{
         border: 2px solid {tm.var(colors.BORDER_FOCUS)};
         outline: none;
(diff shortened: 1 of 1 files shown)


abdnh:
Looking good. Thank you!

PR #4326, opened 2025-09-09
fix: persist colour picker's custom palette

### LINKED ISSUE #4233: Custom Colors Not Sticky Anymore
> Having updated to the latest desktop version the custom colors are no longer sticky. I use this feature all the time - please fix or tell me what I am not doing right. Thanks in advance!

Originally reported on https://forums.ankiweb.net/t/custom-colors-not-sticky-anymore/64662/2

Appears to be a regression in Qt 6.9.

iamllama:
Closes #4233

Custom palette colours are currently reset when anki is restarted. This pr proposes to save them to the current profile. Tested on win 10 and wsl ubuntu

I understand its a qt regression, but without a timeline on when/if it'll get fixed, i reckon its worth handling in anki in the meantime

A quirk with this impl is that (afaik) there is no event for when the picker is closed without a colour being selected. So i've chosen to save the palette when its both opened and when a colour is selected

### DIFF
--- a/proto/anki/frontend.proto
+++ b/proto/anki/frontend.proto
@@ -27,6 +27,9 @@ service FrontendService {
   rpc deckOptionsRequireClose(generic.Empty) returns (generic.Empty);
   // Warns python that the deck option web view is ready to receive requests.
   rpc deckOptionsReady(generic.Empty) returns (generic.Empty);
+
+  // Save colour picker's custom colour palette
+  rpc SaveCustomColours(generic.Empty) returns (generic.Empty);
 }
 
 service BackendFrontendService {}
--- a/qt/aqt/editor.py
+++ b/qt/aqt/editor.py
@@ -151,6 +151,7 @@ class Editor:
         self.add_webview()
         self.setupWeb()
         self.setupShortcuts()
+        self.setupColourPalette()
         gui_hooks.editor_did_init(self)
 
     # Initial setup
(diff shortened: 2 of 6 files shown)


abdnh:
I think the RPC approach is good. Better to save the colors immediately than potentially lose them if Anki doesn't exit cleanly.

I gave this a test on macOS and it appears to be working well.

abdnh (approved):

iamllama:
Do you think it's worth saving to the collection itself? This'd anticipate the note editor getting ported to mobile clients where the native browser's colour picker might not allow changing the palette (pic below is from ankidroid, for example):
image

This way users can save their favourite colours on desktop and use them on mobile

related: https://forums.ankiweb.net/t/add-an-input-field-in-io-fill-colour-to-type-out-the-colours/63579/2

dae:
> I understand its a qt regression, but without a timeline on when/if it'll get fixed, i reckon its worth handling in anki in the meantime

It's bothering quite a few users, so I think that's the right call. Thanks for figuring out a workaround! I'll get this into 25.09.1.

(edit: whoops: https://github.com/ankitects/anki/pull/4320#issuecomment-3301481182)

> Do you think it's worth saving to the collection itself? 

Yeah, good call. It's not a lot of data, and it will avoid work for users in the future if we decide to switch.

campbellgunn-j:
This fix did not work for me. Please investigate - thanks!

campbellgunn-j:
I now understand the problem - the link to the latest version 25.09 for MacOS installs 25.07 NOT 25.09 even though the filename says 25.09. If, however, you then upgrade from within 25.07 to 25.09 using the Tools>Upgrade launcher then 25.09 installs and the custom colors are now sticky - thanks for the fix but please update the initial download link to avoid confusion!

PR #4327, opened 2025-09-09
CHANGE launcher strings to clarify installing for novices

GithubAnon0000:
Aims to solve questions like
- https://forums.ankiweb.net/t/trouble-installing-latest-upgrade/66354?u=anon_0000
- https://forums.ankiweb.net/t/mac-downloading-issue/66346?u=anon_0000
- https://www.reddit.com/r/Anki/comments/1nemoye/sooo_i_just_installed_anki_on_windows_11_and_when/

# Important

I ran `./qt/launcher/lin/build.sh` to build the launcher. It seems to download the dependency and then still fails to build with
```sh
info: downloading component 'rust-std' for 'aarch64-unknown-linux-gnu'
info: installing component 'rust-std' for 'aarch64-unknown-linux-gnu'
cat: ../../../.version: No such file or directory
   Compiling proc-macro2 v1.0.101
   Compiling unicode-ident v1.0.18
   Compiling libc v0.2.175
   Compiling heck v0.5.0
   Compiling getrandom v0.3.3
   Compiling rustix v1.0.8
   Compiling cfg-if v1.0.3
   Compiling bitflags v2.9.3
   Compiling camino v1.1.12
   Compiling serde v1.0.219
   Compiling linux-raw-sys v0.9.4
   Compiling either v1.15.0
   Compiling once_cell v1.21.3
   Compiling quote v1.0.40
   Compiling anyhow v1.0.99
   Compiling syn v2.0.106
   Compiling fastrand v2.3.0
   Compiling serde_json v1.0.143
   Compiling option-ext v0.2.0
   Compiling dirs-sys v0.5.0
   Compiling itertools v0.14.0
   Compiling itoa v1.0.15
   Compiling ryu v1.0.20
   Compiling tempfile v3.21.0
   Compiling memchr v2.7.5
   Compiling launcher v1.0.0 (/home/anon/Downloads/anki/Source_Code/anki/qt/launcher)
   Compiling dirs v6.0.0
   Compiling snafu-derive v0.8.8
   Compiling snafu v0.8.8
   Compiling anki_io v0.0.0 (/home/anon/Downloads/anki/Source_Code/anki/rslib/io)
   Compiling anki_process v0.0.0 (/home/anon/Downloads/anki/Source_Code/anki/rslib/process)
    Finished `release` profile [optimized] target(s) in 6.02s
   Compiling libc v0.2.175
   Compiling rustix v1.0.8
   Compiling getrandom v0.3.3
   Compiling cfg-if v1.0.3
   Compiling linux-raw-sys v0.9.4
   Compiling bitflags v2.9.3
   Compiling serde v1.0.219
   Compiling snafu v0.8.8
   Compiling camino v1.1.12
   Compiling option-ext v0.2.0
   Compiling fastrand v2.3.0
   Compiling either v1.15.0
   Compiling once_cell v1.21.3
   Compiling itertools v0.14.0
   Compiling anyhow v1.0.99
   Compiling serde_json v1.0.143
   Compiling dirs-sys v0.5.0
   Compiling memchr v2.7.5
   Compiling ryu v1.0.20
   Compiling tempfile v3.21.0
   Compiling itoa v1.0.15
   Compiling launcher v1.0.0 (/home/anon/Downloads/anki/Source_Code/anki/qt/launcher)
   Compiling anki_process v0.0.0 (/home/anon/Downloads/anki/Source_Code/anki/rslib/process)
   Compiling anki_io v0.0.0 (/home/anon/Downloads/anki/Source_Code/anki/rslib/io)
   Compiling dirs v6.0.0
error: linker `aarch64-linux-gnu-gcc` not found
  |
  = note: No such file or directory (os error 2)

error: could not compile `launcher` (bin "anki-console") due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `launcher` (bin "build_win") due to 1 previous error
error: could not compile `launcher` (bin "launcher") due to 1 previous error
```

I, thus, couldn't build and check if it works. It's only string changes though, so shouldn't cause any issues.

I also ran `./ninja format` and `./ninja fix` – I expected my `printf()` in `qt/launcher/src/main.rs` to be broken into multiple lines (due to 80 column limit), though that didn't happen. Not sure if I'm supposed to manually do that or if something else is going wrong here.


dae:
The build script expects to be run from its containing folder, and you'll need to have a cross-compiling linker installed to run it. You should be able to 'cargo build' to build just the current arch, or modify the script temporarily.

As for the changes, I'm not a huge fan I'm afraid. The first one is a bunch of extra text that people will always have to see. The others are ok, though 'specific' feels like it it'll just make things harder for non-native speakers.

How about we take a more selective approach instead? The launcher knows if it's been run before, and can treat the first-run case differently.

GithubAnon0000:
> The others are ok, though 'specific' feels like it it'll just make things harder for non-native speakers.

I wanted to have it more explicit because I thought it was a bit too ambiguous. But looking at it now I do agree with you. The "specific" should be removed.



> As for the changes, I'm not a huge fan I'm afraid. The first one is a bunch of extra text that people will always have to see.
> […]
> How about we take a more selective approach instead? The launcher knows if it's been run before, and can treat the first-run case differently.

Fair enough. The issue though is that users don't know what to do with that screen. There are surprisingly a lot of users that get worried or even a bit scared when they see a terminal window. And if there are no instructions, it makes it worse (and explains all the posts on the forum and on reddit). I thus think handling the selective approach won't solve the issue, as users will have to see the launcher whenever they want to up- or downgrade.

Ideally there would be a way to have the official anki logo shown and then short instructions (maybe even additionally using tooltips for more info). But the current terminal approach doesn't support that (though I think it should be possible to add e.g. images, though it would probably be very difficult to make it cross platform and would increase dependencies).

I don't have a better idea on how to solve this issue (except for writing a gui launcher which – I believe – is already on gh issue tracker). I'm happy to make changes to the current PR, especially if you or someone else has a good idea on how to do this!

dae:
> I thus think handling the selective approach won't solve the issue, as users will have to see the launcher whenever they want to up- or downgrade.

But they'll have seen the terminal on first install. Surely it's not going to be as surprising to them on subsequent updates, especially since they'd be invoking them from within Anki?

If you want something that is always displayed, I think it should be less visually noisy, as people will always see it. Something like "To finish installing, press the Enter/Return key on your keyboard".

GithubAnon0000:
> But they'll have seen the terminal on first install. Surely it's not going to be as surprising to them on subsequent updates, especially since they'd be invoking them from within Anki?

Not sure. If you think that subsequent updates won't be as much of a problem, then we can go for a selective approach as you suggested before. I do share your goal to remove unneccessary clutter, so would like to improve my PR in that regard if possible.

dae:
How about the following?

```diff
diff --git a/qt/launcher/src/main.rs b/qt/launcher/src/main.rs
index f22f374ba..54976183f 100644
--- a/qt/launcher/src/main.rs
+++ b/qt/launcher/src/main.rs
@@ -161,7 +161,7 @@ fn run() -> Result<()> {
 
     print!("\x1B[2J\x1B[H"); // Clear screen and move cursor to top
     println!("\x1B[1mAnki Launcher\x1B[0m\n");
-    println!("This is the Anki Launcher. If you're running it for the first time, press Enter to install the latest version. Otherwise, pressing Enter will check for updates and then open Anki.\n");
+    println!("Press the Enter/Return key on your keyboard to install or update Anki.\n");
 
     ensure_os_supported()?;
 
@@ -441,8 +441,8 @@ fn file_timestamp_secs(path: &std::path::Path) -> i64 {
 
 fn get_main_menu_choice(state: &State) -> Result<MainMenuChoice> {
     loop {
-        println!("1) Install Latest Anki (press Enter)");
-        println!("2) Choose a specific version");
+        println!("1) Install Latest Anki (default)");
+        println!("2) Choose version");
 
         if let Some(current_version) = &state.current_version {
             let normalized_current = normalize_version(current_version);
```

GithubAnon0000:
Looks good, I think your suggestion is much better! I'll need a few days to come back to it, solve the conflicts and then merge your suggested changes, though.

dae:
Thank you!

GithubAnon0000:
Thank you as well!

Btw, in case you didn't manually check it: I let it be as a draft because I didn't yet build the launcher.

**Edit:** Ah, as I can see it has been resolved by user1823.

dae:
You'd marked it as a draft; the oversight was my fault.

GithubAnon0000:
No worries. I could have mentioned that I have yet to test the changes, but it worked out in the end.

Issue #4198, opened 2025-07-13
Stats - Retention rate help box links to Deck Options manual page


user1823:
The help box for the "Retention rate" part of the Stats screen links to the Deck Options page of the manual, which doesn't have any description for the Retention rate.

abdnh:
The FSRS subsection describes retention though? This should be fixed in the manual first if you feel the description can be improved.

user1823:
The deck options page explains desired retention, not retention rate. This is expected because retention rate is not an "option".

IMO, the ideal fix for the issue is to remove the "For more information, see Retention rate in the manual." part from the help box.

Hannanilsenn:
Hi! I’m new to contributing and this will be my first issue. It looks straightforward, and I’d like to give it a try. I was thinking of removing the line in the help box that references “Retention rate in the manual.” Does this still need to be fixed?

user1823:
Yes, it still needs to be fixed and removing that line is the ideal solution.

PR #4329, opened 2025-09-10
Stats - Retention rate help box links to Deck Options manual page #4198

Hannanilsenn:
This PR removes the help box link for the "Retention rate setting," which previously pointed to the Deck Options manual page. The corresponding section no longer exists in the manual, so keeping the link would be misleading.

If the manual is updated with relevant information in the future, the link can be re-added easily.


user1823:
This removes the links from the Deck Options page too. We want to remove the link from "Retention rate" only.

Hannanilsenn:
Thanks for the feedback! I understand, I’ll adjust the PR so that only the link for “Retention rate” is removed and leave the other links on the Deck Options page intact.

user1823:
You must revert https://github.com/ankitects/anki/pull/4329/commits/c83324bf816c716728aa828b4706f9076b019452 and then build Anki to check whether this change actually works. 

Also, add "Fixes #4198" to the top post so that the issue is auto-closed when the PR is merged.

dae:
How about a different related link instead, like https://docs.ankiweb.net/deck-options.html#desired-retention? You make URLs optional, and future PRs will be less likely to bother.

user1823:
> You make URLs optional, and future PRs will be less likely to bother.

That's a fair point. 

But, if the text is going to say "For more information, see ⁨Retention⁩ in the manual." and link to the desired retention section, I think that it would be confusing. 

If we are linking to the desired retention, the text should say "For more information, see ⁨**Desired retention**⁩ in the manual."

Hannanilsenn:
I agree with @user1823 , if the link goes to Desired retention, the help text should also say Desired retention to avoid confusion. I can update the PR so that both the text and the link say Desired retention. Does that sound good?

Hannanilsenn:
I see the CI failed because the branch is out of date with main. I’ll rebase onto the latest main and push an update shortly.

ts/lib/components/HelpModal.svelte
```diff
@@ -22,7 +22,8 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     import { type HelpItem, HelpItemScheduler } from "./types";
 
     export let title: string;
-    export let url: string;
+    export let url: string | undefined = undefined;
```
user1823:
```suggestion
    export let url: string;
```

ts/routes/graphs/TrueRetention.svelte
```diff
@@ -72,12 +71,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     >
         <HelpModal
             title={tr.statisticsTrueRetentionTitle()}
-            url={HelpPage.DeckOptions.fsrs}
+            url="https://docs.ankiweb.net/deck-options.html#desired-retention"
```
user1823:
```suggestion
            url={HelpPage.DeckOptions.desiredRetention}
```

Combine the above with adding this
```ts
        desiredRetention: "http://docs.ankiweb.net/deck-options.html#desired-retention",
```

just after

https://github.com/ankitects/anki/blob/3890e12c9e48c028c3f12aa58cb64bd9f8895e30/ts/lib/tslib/help-page.ts#L30

Hannanilsenn:
Thanks for the suggestions! I’ve applied them, and the on:mount change is included again after I reverted it. Hopefully everything looks good now.

ts/routes/graphs/TrueRetention.svelte
```diff
@@ -72,7 +72,8 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
     >
         <HelpModal
             title={tr.statisticsTrueRetentionTitle()}
-            url={HelpPage.DeckOptions.fsrs}
+            url={HelpPage.DeckOptions.desiredRetention}
+            linkLabel="Desired retention"
```
user1823:
```suggestion
            linkLabel={tr.deckConfigDesiredRetention()}
```

I am sorry that I missed this in the previous review. Looks good to me otherwise.

ts/lib/tslib/help-page.ts
```diff
@@ -28,6 +28,7 @@ export const HelpPage = {
         dailyLimits: "https://docs.ankiweb.net/deck-options.html#daily-limits",
         audio: "https://docs.ankiweb.net/deck-options.html#audio",
         fsrs: "http://docs.ankiweb.net/deck-options.html#fsrs",
+        desiredRetention: "http://docs.ankiweb.net/deck-options.html#desired-retention",
```
user1823:
```suggestion
        desiredRetention: "https://docs.ankiweb.net/deck-options.html#desired-retention",
```

ts/lib/tslib/help-page.ts
```diff
@@ -28,6 +28,7 @@ export const HelpPage = {
         dailyLimits: "https://docs.ankiweb.net/deck-options.html#daily-limits",
         audio: "https://docs.ankiweb.net/deck-options.html#audio",
         fsrs: "http://docs.ankiweb.net/deck-options.html#fsrs",
```
user1823:
```suggestion
        fsrs: "https://docs.ankiweb.net/deck-options.html#fsrs",
```

user1823 (commented):
Two other minor suggestions. We should link to the HTTPS version.

Hannanilsenn:
Thanks for the suggestions @user1823 ! I’ve updated the links to use HTTPS.

dae:
Thank you both for this!

PR #4330, opened 2025-09-10
Add safe_rename to handle cross-device file moves (OS error 18)

AmandaSternberg-creator:
I am a student currently taking an open source course and I wanted to contribute with a small change first and get an understanding for how Rust works. The purpose of this PR is to provide a safer file move operation and prevent crashes when files are on different devices. This PR includes the following:

- Added a safe_rename function that handles OS error 18 (Cross-device link) when moving files across different filesystems
- Updated remove_files to use safe_rename instead of fs::rename
- Added a test for safe_rename to ensure the file moves correctly and the content remains intact

I ran into an issue where I couldn’t run the existing tests in the Anki workspace due to dependency conflicts, so I verified the functionality in a separate minimal Rust project to ensure safe_rename works as expected. If you have any suggestions on why these dependency conflicts occur or tips for properly running tests in the workspace, I would greatly appreciate your guidance.

I would also welcome any feedback or suggestions to improve my contribution and learn more. Thank you! :)

### DIFF
--- a/rslib/src/media/files.rs
+++ b/rslib/src/media/files.rs
@@ -565,18 +565,18 @@ mod test {
 
     #[test]
     fn safe_rename_moves_file() {
-        let dir = tempdir().unwrap();
-        let src = dir.path().join("test_src.txt");
-        let dst = dir.path().join("test_dst.txt");
+    let dir = tempdir().unwrap();
+    let src = dir.path().join("myfile.txt");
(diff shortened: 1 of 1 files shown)


AmandaSternberg-creator:
Supersedes [#4330](https://github.com/ankitects/anki/pull/4330). This is a cleaner branch with the same changes rebuilt due to environment issues.

PR #4331, opened 2025-09-11
latex: remove blank lines inside math blocks (fixes #3466)

### LINKED ISSUE #3466: Generating tmp.tex from [$$] block adds blank lines, breaking some environments
> A card with the following text:
> [$$]
> \begin{tikzcd}
> PB \ar[r, "g^{-1}"] \ar[d, "k_*"']
> & PD \ar[d, "h_*"] \\
> PC \ar[r, "f^{-1}"']
> & PA
> \end{tikzcd}
> [/$$]
> 
> generates the following tmp.tex file:
> [... preamble, including the needed \usepackage etc commands, snipped ...]
> \begin{displaymath}
> \begin{tikzcd}
> PB \ar[r, "g^{-1}"] \ar[d, "k_*"']
> 
> & PD \ar[d, "h_*"] \\
> 
> PC \ar[r, "f^{-1}"']
> 
> & PA
> \end{tikzcd}
> 
> \end{displaymath}
> \end{document}
> 
> which fails to compile, because of the blank lines.  I don’t know why Anki is introducing these blank lines, and I don’t know how to stop it doing so.
> Anki version:
> Anki 24.06.3 (d678e393)
> Platform: macOS-14.7-arm64-arm-64bit

Originally reported on https://forums.ankiweb.net/t/generating-tmp-tex-from-block-adds-blank-lines-breaking-some-environments/50169/2

matbe766:
Removes `<br>`/`<div>` tags inside math blocks (`[$]`, `[$$]`) instead of turning them into newlines.  
This prevents blank lines that broke environments like `tikzcd`.  
Standard `[latex]` blocks still keep newlines.

Fixes #3466

### DIFF
--- a/rslib/src/latex.rs
+++ b/rslib/src/latex.rs
@@ -79,10 +79,7 @@ pub fn extract_latex(text: &str, svg: bool) -> (Cow<'_, str>, Vec<ExtractedLatex
 
         let fname = fname_for_latex(&latex, svg);
         let img_link = image_link_for_fname(&latex, &fname);
-        extracted.push(ExtractedLatex {
-            fname,
-            latex: latex.into(),
-        });
+        extracted.push(ExtractedLatex { fname, latex });
 
         img_link
     });
--- a/rslib/src/latex.rs
+++ b/rslib/src/latex.rs
@@ -64,18 +64,24 @@ pub fn extract_latex(text: &str, svg: bool) -> (Cow<'_, str>, Vec<ExtractedLatex
     let mut extracted = vec![];
 
     let new_text = LATEX.replace_all(text, |caps: &Captures| {
+        // [latex] blocks preserve newlines between <br>/<div>
+        // [$] and [$$] blocks remove <br>/<div> completely so that
+        // no blank lines are introduced inside math environments.
         let latex = match (caps.get(1), caps.get(2), caps.get(3)) {
-            (Some(m), _, _) => m.as_str().into(),
-            (_, Some(m), _) => format!("${}$", m.as_str()),
-            (_, _, Some(m)) => format!(r"\begin{{displaymath}}{}\end{{displaymath}}", m.as_str()),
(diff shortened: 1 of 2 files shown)


rslib/src/latex.rs
```diff
@@ -146,4 +163,18 @@ mod test {
             }]
         );
     }
+
+    /// Ensures that math blocks do not contain unintended blank lines
+    /// when <br> or <div> elements are present in the HTML.
+    #[test]
+    fn no_blank_lines_in_math() {
```
abdnh, line 170: This doesn't seem to test the relevant behavior.

rslib/src/latex.rs
```diff
@@ -96,6 +99,20 @@ fn strip_html_for_latex(html: &str) -> Cow<'_, str> {
     out
 }
 
+/// Removes HTML breaks (<br>, <div>) from math blocks instead of
+/// converting them to newlines. This prevents LaTeX environments
+/// from being broken by unintended blank lines.
+fn strip_html_for_latex_math(html: &str) -> Cow<'_, str> {
```
abdnh, line 105: Nitpick: we can add an extra arg to strip_html_for_latex() instead of a new function to reduce duplication.

dae:
I don't remember the reason why we originally did this, but I'm pretty sure there was one. I think we need to know that first.

https://fs.blog/chestertons-fence/

matbe766:
I understand your concern. From my side, I only noticed the issue in math blocks ([$], [$$]) where the extra newlines broke environments like tikzcd.

This is actually my first issue/PR, so I appreciate the feedback and I’d be grateful for any tips on how you’d prefer me to handle this. I’m happy to adjust the approach if you can point me in the right direction.

dae:
Trying to find the answer to the question above is the first step, and I'm afraid I don't have time at the moment. If you don't have the time or inclination either, I'd recommend picking a different issue that is less likely to cause regressions, and will impact more people.

Issue #4288, opened 2025-08-25
[BUG/Android]: trash_media_files - Cross-device link (os error 18)


david-allison:
Still diagnosing on our side, I'll relay relevant information here. 

At a quick glance, it feels the `fs::rename` would be the issue, but I don't see how the cross-device link would be produced via the Rust.

* https://github.com/ankidroid/Anki-Android/issues/19134 

```
net.ankiweb.rsdroid.BackendException$BackendCardTypeException: Cross-device link (os error 18)
	at net.ankiweb.rsdroid.exceptions.BackendIoException.<init>(BackendIoException.java:23)
	at net.ankiweb.rsdroid.BackendException$Companion.fromError(BackendException.kt:197)
	at net.ankiweb.rsdroid.BackendKt.unpackResult(Backend.kt:317)
	at net.ankiweb.rsdroid.BackendKt.access$unpackResult(Backend.kt:1)
	at net.ankiweb.rsdroid.Backend.runMethodRaw$lambda$1(Backend.kt:133)
	at net.ankiweb.rsdroid.Backend.withBackend(Backend.kt:145)
	at net.ankiweb.rsdroid.Backend.runMethodRaw(Backend.kt:132)
	at anki.backend.GeneratedBackend.trashMediaFilesRaw(GeneratedBackend.kt:2889)
	at anki.backend.GeneratedBackend.trashMediaFiles(GeneratedBackend.kt:2896)
	at com.ichi2.anki.libanki.Media.trashFiles(Media.kt:156)
```

`deckPath: /storage/emulated/0/Android/data/com.ichi2.anki/files/AnkiDroid`

<details>

<summary>Relevant code</summary>

https://github.com/ankitects/anki/blob/3d0a408a2b1f86a647e2f3e6eee20893fa0dc706/rslib/src/media/service.rs#L43-L45
https://github.com/ankitects/anki/blob/3d0a408a2b1f86a647e2f3e6eee20893fa0dc706/rslib/src/media/mod.rs#L92-L108
https://github.com/ankitects/anki/blob/3d0a408a2b1f86a647e2f3e6eee20893fa0dc706/rslib/src/media/files.rs#L323-L362

</details>

dae:
trash_folder() was designed to create the trash folder in the same folder parent folder as the media folder, to avoid such issues, so the error is surprising. Is it only one user affected?

david-allison:
Only one user affected, and I haven't heard back from them

PR #4332, opened 2025-09-12
Add safe_rename for cross-device file moves (OS error 18)

AmandaSternberg-creator:
This PR replaces my previous PR [#4330](https://github.com/ankitects/anki/pull/4330) with a clean branch and updated commits to solve the issue https://github.com/ankitects/anki/issues/4288. I started fresh because I had issues with my local development environment, so this is a new push containing the same fix, now tested and ready for review.

fs::rename fails when moving files between different filesystems. This change prevents crashes by falling back to a safe copy-and-remove strategy. Summary of changes:

- Added a new safe_rename function to handle OS error 18 (cross-device file move errors) by copying and removing the file if fs::rename fails
- Updated remove_files to use safe_rename instead of fs::rename
- Added a test for safe_rename to verify correct behavior and file integrity after moving

This is a just a cleaner branch and updated PR because of earlier environment setup issues. I am still new to Rust, so I would really appreciate any feedback on how to make this code more idiomatic and maintainable, especially regarding the testing approach.

### DIFF
--- a/qt/aqt/stylesheets.py
+++ b/qt/aqt/stylesheets.py
@@ -180,7 +180,7 @@ class CustomStyles:
     QPushButton {{
         margin: 1px;
     }}
-    QPushButton:focus {{
+    QPushButton:focus, QPushButton:default:hover {{
         border: 2px solid {tm.var(colors.BORDER_FOCUS)};
         outline: none;
         margin: 0px;
@@ -198,9 +198,6 @@ class CustomStyles:
                 tm.var(colors.BUTTON_GRADIENT_END),
             )
         };
-    }}
-    QPushButton:default:hover {{
-        border-width: 2px;
     }}
     QPushButton:pressed,
     QPushButton:checked,


abdnh:
I assume this is tryring to solve https://github.com/ankitects/anki/issues/4288

rslib/src/media/files.rs
```diff
@@ -320,6 +320,18 @@ pub(crate) fn mtime_as_i64<P: AsRef<Path>>(path: P) -> io::Result<i64> {
         .as_millis() as i64)
 }
 
+pub(crate) fn safe_rename(src: &Path, dst: &Path) -> io::Result<()> {
+    match fs::rename(src, dst) {
+        Ok(_) => Ok(()),
+        Err(e) if e.raw_os_error() == Some(18) => {
```
abdnh: What does 18 mean here? A comment would be nice.

I'm actually getting this instead on Windows when I try to move a file from C to D:
```
Error: Os { code: 17, kind: CrossesDevices, message: "The system cannot move the file to a different disk drive." }
```

Probably not worthing handling that.

rslib/src/media/files.rs
```diff
@@ -550,4 +564,20 @@ mod test {
             Cow::<str>::Owned(format!("{}_", " ".repeat(MAX_MEDIA_FILENAME_LENGTH - 2)))
         );
     }
+    #[test]
+    fn safe_rename_moves_file() {
```
abdnh, line 570: This test is not really confirming the function works across disks. But I'm not sure how difficult it is to test that.

rslib/src/media/files.rs
iamllama: ERROR_NOT_SAME_DEVICE (17) on windows, EXDEV (18) on posix

AmandaSternberg-creator: @abdnh @iamllama Thank you very much for the feedback! I have updated the code to use io::ErrorKind::CrossesDevices instead of raw error codes, which makes the implementation a lot clearer and would solve the issue!

rslib/src/media/files.rs
AmandaSternberg-creator: @abdnh The test only checks the copy+remove fallback logic and not an actual cross-device move. Testing a true cross-device scenario would require either a mocked filesystem or integration tests with multiple mount points, which felt out of scope for a unit test here.

dae:
I'm afraid I'm not keen to rush to fix the problem, as we currently only have a single report of it happening, and the trash folder should already be on the same device. I think we need to understand what exactly is happening for this user before we attempt to fix it.

PR #4333, opened 2025-09-12
Add syntax for multi-card cloze deletions

tfpgh:
Hi everyone! This was a feature I found myself wanting, so I figured I would implement it. I've never contributed to Anki before, so let me know if feature PRs like this are appreciated.

This PR adds support for **multi-card cloze deletions**: allowing one deletion to appear on multiple cards from the same note. Not sure if this is the best possible name for this feature.

An example note with the new syntax:
`Words ending with {{c1::-té}}, {{c2::-sion}}, or {{c3::-tion}} are typically {{c1,2,3,4::feminine}}.`

This creates 4 cards, with feminine deleted on all four:
1. `Words ending with [...], -sion, or -tion are typically [...].`
2. `Words ending with -té, [...], or -tion are typically [...].`
3. `Words ending with -té, -sion, or [...] are typically [...].`
4. `Words ending with -té, -sion, or -tion are typically [...].`

There seems to be some desire for this feature on Reddit and the forum. I didn't edit the documentation, but I'm open to adding some or making any other changes if this gets approval to be pulled in.

Thanks!

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -245,6 +245,7 @@ Max Romanowski <maxr777@proton.me>
 Aldlss <ayaldlss@gmail.com>
 Hanna Nilsén <hanni614@student.liu.se>
 Elias Johansson Lara <elias.johanssonlara@gmail.com>
+Toby Penner <tobypenner01@gmail.com>
 
 ********************
 


GithubAnon0000:
As an aside: That can be already achieved with a string like this one:
> Words ending with {{c1::-té}}, {{c2::-sion}}, or {{c3::-tion}} are typically {{c4::{{c3::{{c2::{{c1::feminine}}.}}}}}}

Though, arguably, your version is more readable.

tfpgh:
> As an aside...

Not an aside at all! Should we introduce a cleaner syntax which adds a bit of complexity? I don't know.

Your version didn't intuitively pop into my head as an option. Maybe we add something to the cloze docs that explains how to delete something on multiple cards?

dae:
Thank you for contributing! I can't fault your PR - it looks well written, and I love that you've added a bunch of tests. There are a few things that give me pause however:

- I don't recall anyone asking for this before, which tells me it's rather niche. Every niche feature we add has a maintenance burden.
- We've turned a simple scalar value into a vector, which adds some overhead, and could result in large slowdowns if users add a lot of cloze numbers (inevitably there will be a few users who use scripts to generate such things without considering the performance implications)

That said, the code is already written, and this probably isn't going to impact maintenance a great deal (though in some regards, these decisions are death by a thousand cuts), so in this case please feel free to merge it in if you'd like Abdo.

@tfpgh recommend rebasing this in a few hours to clear the conflict so Abdo can process if he chooses

tfpgh:
@dae Thank you for your comment. I totally understand everything you wrote.

@abdnh Should be ready to pull in. If you decide to proceed, but need any changes, let me know.

abdnh:
No strong opinions here. I also thought it's well-written and unlikely to cause issues so it's fine to merge.

Issue #4215, opened 2025-07-20
Browser search bar doesn't convert line breaks/new lines


jcznk:
Tested on version 25.07.3 rc1. I was able to reproduce the same behavior with every other version I tested.

Steps to reproduce:

1. Go to the Card Browser.
2. Copy some text containing line breaks/newlines from any source.
3. Paste it into the search bar and perform a search → no search results.
4. Edit the pasted text in the search bar by deleting the "spaces" created by the line breaks/newlines and replacing them with normal spaces → the search now works as expected.

<img src="https://github.com/user-attachments/assets/8c78363c-bbd8-48bd-a566-f1872a1c125b" height=400>

eliasjlara:
Hello, I would like to contribute to Anki! Is this something I can work on?

PR #4336, opened 2025-09-15
Add support for newlines in search

eliasjlara:
Fixes this issue: https://github.com/ankitects/anki/issues/4215

Currently, newlines in search queries aren’t treated as breaks, so a query like:
```
THIS
IS
A
TEST
```
fails to match cards. This query would be represented as `Search(UnqualifiedText("THIS\nIS\nA\nTEST"))` in search/parser.rs, and given that cards store newline characters as `<br>`, it yields no results. 

This PR updates the parser to treat newline characters `\n` as breaks during parsing. This ensures multi-line queries match card content correctly.

An alternative would be to replace `\n` with spaces, but handling them as breaks preserves the original input more accurately.

### DIFF
--- a/pylib/tests/test_find.py
+++ b/pylib/tests/test_find.py
@@ -309,3 +309,12 @@ def test_findDupes():
     assert not r
     # front isn't dupe
     assert col.find_dupes("Front") == []
+
+
+def test_find_cards_with_newlines():
+    col = getEmptyCol()
+    note = col.newNote()
+    note["Front"] = "foo"
+    note["Back"] = "foo<br>bar"
+    col.addNote(note)
+    assert len(col.find_cards("foo\nbar")) == 1
--- a/rslib/src/search/parser.rs
+++ b/rslib/src/search/parser.rs
@@ -204,7 +204,7 @@ fn group_inner(input: &str) -> IResult<'_, Vec<Node>> {
 }
 
 fn whitespace0(s: &str) -> IResult<'_, Vec<char>> {
-    many0(one_of(" \u{3000}")).parse(s)
+    many0(one_of(" \n\u{3000}")).parse(s)
 }
 
 /// Optional leading space, then a (negated) group or text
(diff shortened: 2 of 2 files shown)


abdnh:
With this, `foo\nbar` will also match fields with `foo bar`. This can be unexpected.

Additionally, If you paste this text to the browser search field and add double quotes (`"foo bar"`), it'll be normalized to `foo bar`, which is not equivalent.

I'm not sure about the right solution here. Maybe it's better to handle this at the browser's level, so that `\n` is converted to `<br>`. Any thought @jcznk?

pylib/tests/test_find.py
```diff
@@ -309,3 +309,12 @@ def test_findDupes():
     assert not r
     # front isn't dupe
     assert col.find_dupes("Front") == []
+
+
+def test_find_cards_with_newlines():
```
abdnh: Most new tests are added to the Rust layer instead.

jcznk:
I also am not sure what the "objectively" correct solution would be. For my personal use case, simply stripping `\n` and replacing them with spaces would be appropriate. If I need exact fidelity with breaks and newlines, I can either copy the HTML (for `<br>`) or perform a regex search (for `\n`).

As a side note, from my testing, newlines seem to never work in regular (non-regex) browser searches, even when the note's fields actually contain them. E.g., if I have this span

```
<span title="FirstLine
SecondLine">Hello World</span>
```

and I copy-paste `FirstLine\nSecondLine` to the browser search bar, no search result is yielded. 

(I tried testing the current PR, but I couldn't make it work. Browser searches seem to behave as usual. This is my first time building Anki from source, so maybe I'm doing something wrong)

abdnh:
> For my personal use case, simply stripping \n and replacing them with spaces would be appropriate.

That's reasonable.
@eliasjlara If you'd like to implement this solution, it only requires a .replace() here: https://github.com/ankitects/anki/blob/c56e6e55eca2463f70d27a17d953a3218d65978e/qt/aqt/browser/browser.py#L523

eliasjlara:
Alright, this seems like a resonable approach. Thanks for the help! I have implemented the fix but I'm getting some merge errors, I will try to fix this asap.

dae:
You've accidentally included changes to the i18n submodules in this PR. Happy to merge once they're removed.

eliasjlara:
> You've accidentally included changes to the i18n submodules in this PR. Happy to merge once they're removed.

These should be removed now :)

dae:
Thank you both!

PR #4337, opened 2025-09-16
(UI polish) Improved margins in Card Browser's "Previewer"

jcznk:
This PR slightly increases the distance between the `Replay`/`Back Only`/`Previous`/`Next` buttons and the lower/right border of the Preview pane. The buttons were previously glued to the window frame. I tried to make the change subtle, yet noticeable.

Tested on Windows 11. 

Before:
Before

After:
After


dae:
Thanks! I had to make some tweaks for Macs.

PR #4338, opened 2025-09-17
Make timebox message translatable with flexible variable order

snowtimeglass:
An example of the timebox message:
image

Currently, the timebox dialog message is built from two separate strings (and a space), each containing one variable:
`{ $count } cards studied in` + ` ` + `{ $count } minutes.`
image
image
As a result, translators cannot flexibly reorder the words of the variables in their translations. 

(In contrast, they can do so for `Studied x cards in y seconds today (z s/card)` on the deck list and the statistics screen.)
image
image
image

This PR's commit introduces a single string with both variables for the timebox message, allowing translators to adjust the order for more natural expressions in their languages.

### DIFF
--- a/ftl/core/studying.ftl
+++ b/ftl/core/studying.ftl
@@ -46,15 +46,14 @@ studying-type-answer-unknown-field = Type answer: unknown field { $val }
 studying-unbury = Unbury
 studying-what-would-you-like-to-unbury = What would you like to unbury?
 studying-you-havent-recorded-your-voice-yet = You haven't recorded your voice yet.
-studying-card-studied-in =
-    { $count ->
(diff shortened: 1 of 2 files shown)


dae:
Thanks!

sommerluk:
It's a small change, but one that makes translators' lives a lot easier.

Word puzzles composed of two or more strings that combine into a single final message are extremely difficult to translate. Unfortunately, Anki contains many of them. I would love to see them eliminated.

@snowtimeglass  Thanks so much for your work and this pull request!

snowtimeglass:
@sommerluk Thanks a lot for your encouraging words!

PR #4339, opened 2025-09-17
Improve performance of `reviews_for_fsrs` function

L-M-Sherlock:
Before:

image

After:

image

2x faster

### DIFF
--- a/rslib/src/scheduler/fsrs/params.rs
+++ b/rslib/src/scheduler/fsrs/params.rs
@@ -481,24 +481,22 @@ pub(crate) fn reviews_for_fsrs(
     let skip = if training { 1 } else { 0 };
     // Convert the remaining entries into separate FSRSItems, where each item
     // contains all reviews done until then.
-    let items: Vec<(RevlogId, FSRSItem)> = entries
-        .iter()
-        .enumerate()
-        .skip(skip)
-        .map(|(outer_idx, entry)| {
-            let reviews = entries
-                .iter()
-                .take(outer_idx + 1)
-                .enumerate()
-                .map(|(inner_idx, r)| FSRSReview {
-                    rating: r.button_chosen as u32,
-                    delta_t: delta_ts[inner_idx],
-                })
-                .collect();
-            (entry.id, FSRSItem { reviews })
-        })
-        .filter(|(_, item)| !training || item.reviews.last().unwrap().delta_t > 0)
-        .collect_vec();
+    let mut items = Vec::with_capacity(entries.len());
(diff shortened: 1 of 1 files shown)


L-M-Sherlock:
I tested it with 761852 revlog entries.

Before:

fsrs_items_for_memory_states took 98.398208ms

After:

fsrs_items_for_memory_states took 28.536041ms

user1823:
> When not training, only create the final FSRS item

I think this breaks the `Evaluate` feature. Even though it's not exposed in the UI, Anki still uses `Evaluate` to decide whether to keep the newly optimized parameters or not.

You can test this by enabling Evaluate using the debug console command and comparing the outputs before and after the change.

L-M-Sherlock:
> I think this breaks the `Evaluate` feature. Even though it's not exposed in the UI, Anki still uses `Evaluate` to decide whether to keep the newly optimized parameters or not.

It doesn't because `training` is true in this case:

https://github.com/ankitects/anki/blob/3890e12c9e48c028c3f12aa58cb64bd9f8895e30/rslib/src/scheduler/fsrs/params.rs#L303-L328

user1823:
> It doesn't because training is true in this case

Oh, I see. But, this means that evaluate doesn't process the revlog entries in the same way as the scheduler because of the two `if training` blocks in the `reviews_for_fsrs` function. 

I said this about 1.5 years ago:
> … evaluation (should) exactly match the scheduler behaviour. This is because the main purpose of the evaluation is to determine the effectiveness of scheduling, which would be most accurate when both treat the revlogs in the same way.

But, probably not so important because we never noticed this in the past 2 years.

L-M-Sherlock:
> Oh, I see. But, this means that evaluate doesn't process the revlog entries in the same way as the scheduler because of the two `if training` blocks in the `reviews_for_fsrs` function.

Here are three differences:

https://github.com/ankitects/anki/blob/3890e12c9e48c028c3f12aa58cb64bd9f8895e30/rslib/src/scheduler/fsrs/params.rs#L434-L451

https://github.com/ankitects/anki/blob/3890e12c9e48c028c3f12aa58cb64bd9f8895e30/rslib/src/scheduler/fsrs/params.rs#L457-L459

https://github.com/ankitects/anki/blob/3890e12c9e48c028c3f12aa58cb64bd9f8895e30/rslib/src/scheduler/fsrs/params.rs#L481-L501

We could discuss about them in a new issue.

dae:
Thank you as both as always - your reviews of these PRs @user1823 are appreciated.

Issue #4278, opened 2025-08-19
UX: Improve ‘invalid params provided’ error


dae:
> Users don’t know how to resolve this error, and we don’t auto-fix it:
> 
> Maybe using something similar to deck-config-invalid-parameters (if the error is similar):
> The provided FSRS parameters are invalid. Leave them blank to use the default parameters.
> 

Originally reported on https://forums.ankiweb.net/t/ux-improve-invalid-params-provided-error/65557/2

Linking to a new FAQ would be good too.

AmandaSternberg-creator:
Hi! I’d like to work on this issue. Is it available?

user1823:
Does this really need a fix on Anki's side? I think that fsrs-rs is returning a vague error and it should ideally determine what is wrong and return a specific error message. 

For example, in the following code, it isn't clear whether the revlog is abnormal or the parameters are abnormal.

https://github.com/open-spaced-repetition/fsrs-rs/blob/58b0fb23b589bb99eed13eb481fc164dde81a1e3/src/inference.rs#L176-L177

@L-M-Sherlock

dae:
If this error is sometimes triggered by bad revlog data, agreed that we need a better error to make it clearer. But I was under the impression the common case is the user has managed to get invalid params into their deck presets. The current error message was lazy/brief because the assumption was that Anki would be validating these params prior to saving, and users wouldn't hit this unless they were manually modifying the collection. I presume we don't currently know how the invalid params are getting into the collection, and what they look like?

Given the number of reports, perhaps we should consider falling back on the default params instead of throwing an error if the params are invalid?

user1823:
> But I was under the impression the common case is the user has managed to get invalid params into their deck presets.

That's probably the most common cause. 
But, just by looking at the code, having bad revlog data can also trigger the issue. I don't have any confirmed reports of such cases, though.

> I presume we don't currently know how the invalid params are getting into the collection, and what they look like?

Just a few versions ago, Anki wasn't validating the parameters before saving them into the collection and due to a "bug" in accessing the saved params, they were not used.

- https://github.com/ankitects/anki/pull/4052

~~So, these reports are likely from those people who saved invalid params with previous versions and have now updated to the latest version.~~

But, such issues will be picked by `check_and_fill_parameters` in https://github.com/open-spaced-repetition/fsrs-rs/blob/0187b45608ec9470b74f289aaaf0069b458220df/src/model.rs#L316 and return `FSRSError::InvalidParameters`

[...]

PR #4335, opened 2025-09-14
Try to speedup update memory state

DanielPechersky:
First commit is a refactor for code clarity
Second commit is my attempted speedup, assuming `card.set_memory_state` is much slower than everything else.
My change will allocate an extra 3.51 MB of scratch space for every 10,000 cards the function processes.

PR #4342, opened 2025-09-17
Improve FSRS invalid parameters error message with FAQ link

AmandaSternberg-creator:
This PR improves the error message shown when invalid FSRS parameters are provided. It provides a solution to [#4278](https://github.com/ankitects/anki/issues/4278). The PR includes:

- Updated text to explain what users should do
- Added a link to the Anki FAQ for more detailed guidance
- Updated both Rust and FTL strings to ensure consistency in CLI logs and GUI

### DIFF
--- a/rslib/src/scheduler/fsrs/error.rs
+++ b/rslib/src/scheduler/fsrs/error.rs
@@ -13,13 +13,7 @@ impl From<FSRSError> for AnkiError {
             FSRSError::OptimalNotFound => AnkiError::FsrsUnableToDetermineDesiredRetention,
             FSRSError::Interrupted => AnkiError::Interrupted,
             FSRSError::InvalidParameters => AnkiError::FsrsParamsInvalid,
-            FSRSError::InvalidInput => AnkiError::InvalidInput {
-                source: InvalidInputError {
-                    message: "invalid params provided".to_string(),
-                    source: None,
-                    backtrace: None,
-                },
-            },
+            FSRSError::InvalidInput => AnkiError::FsrsParamsInvalid,
             FSRSError::InvalidDeckSize => AnkiError::InvalidInput {
                 source: InvalidInputError {
                     message: "no cards to simulate".to_string(),
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -248,6 +248,7 @@ Elias Johansson Lara <elias.johanssonlara@gmail.com>
 Toby Penner <tobypenner01@gmail.com>
 Danilo Spillebeen <spillebeendanilo@gmail.com>
 Matbe766 <matildabergstrom01@gmail.com>
+Amanda Sternberg <mandis.sternberg@gmail.com>
 
 
 ********************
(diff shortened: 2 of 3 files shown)


rslib/src/scheduler/fsrs/error.rs
```diff
@@ -15,7 +15,7 @@ impl From<FSRSError> for AnkiError {
             FSRSError::InvalidParameters => AnkiError::FsrsParamsInvalid,
             FSRSError::InvalidInput => AnkiError::InvalidInput {
                 source: InvalidInputError {
-                    message: "invalid params provided".to_string(),
```
abdnh, line 18: We should translate this string if it's a common occurrence.

AmandaSternberg-creator: Thanks for the review! I switched the hard-coded "invalid params provided" to go through AnkiError::FsrsParamsInvalid (localized) and updated the FTL message to mention using default values + a FAQ reference.

I am currently blocked by CI: check:minilints says “cargo/licenses.json is out of date”. Locally I rebased onto upstream/main, regenerated on Linux using the project container (./run linux ./ninja fix:minilints) and check:minilints passes with no changes (my Linux-generated file matches the one in the branch byte-for-byte).

Do you have any suggestions on how I can resolve this problem? Thanks!

Luc-Mcgrady: https://github.com/ankitects/anki/pull/4367

The CI problem should be fixed by this pr. Try and rebase again.

AmandaSternberg-creator: Ahh, that makes sense now! I hadn’t realized the cargo license version mismatch was the reason my local build passed while CI failed. Thanks for explaining! After rebasing onto the latest main with #4367 the issue is gone and CI is green :)

user1823:
Also see https://github.com/ankitects/anki/issues/4278#issuecomment-3342813873

dae:
We usually try to keep things like URLs out of translation strings, as they're hard to update if each language has copied the text. Including the link would require adding a variable to the string. Happy to merge if you either do that or drop the link part, as the rest is an improvement over what we currently have, thank you.

AmandaSternberg-creator:
Dropped the FAQ URL as suggested, thanks! 🙂

dae:
I'm afraid there are some problems with the commits in this PR at the moment - they're changing files unrelated to this change. They will squash down to nothing like in https://github.com/ankitects/anki/pull/4335#issuecomment-3360217714, but they make the PR a bit harder to review, and as there's currently a conflict on the CONTRIBUTORS file, this is a good opportunity for you to practice your git skills. Something like:

```
# Make sure your local info about origin/main is up to date
git fetch

# Save the diff between your branch and origin/main into a patch (handles binaries too)
git diff --binary origin/main...HEAD > /tmp/changes.patch

# Throw away your branch’s history and reset it to match origin/main
git reset --hard origin/main

# Re-apply the patch and stage the changes
git apply --index /tmp/changes.patch

# Commit all changes
git commit 

# Push and overwrite the remote branch safely (fails if someone else pushed meanwhile)
git push --force-with-lease
```

AmandaSternberg-creator:
Thanks and sorry for that! I have cleaned up the branch as suggested. PR now only includes the intended changes :)

dae:
Thanks Amanda!

PR #4344, opened 2025-09-17
feat: add support for `tag:nc:...` searches

iamllama:
Noticed while looking at #4312 that support for this was missing, which this pr proposes to add

An integration test seemed out of place in sqlwriter.rs, so i opted to add one in pylib instead


abdnh (approved):

abdnh:
We can add a parsing test to parser.rs, but it's not a big deal.

PR #4347, opened 2025-09-17
fix(import): show message when importing empty CSV file

### LINKED ISSUE #4346: CSV import error message is not ideal
How to reproduce:
* Have a empty file empty.csv
* do file>import and import empty.csv

Actual result:
* a view titled "Import file", with content "500: The file you selected is empty". and nothing else

Image

Expected:
* 500 should not be here. It's not user relevant.
* there should at least be a button to close the view. In practice it can be closed with esc, and I assume other standard shortcut, but it's not discoverable

josod827:
A fix to the message "500: The file you selected is empty." when importing an empty csv-file. Now it shows a popup instead, with an OK button: "The file you selected is empty and cannot be imported."

The "Import File" window is still shown along with the popup. It might make sense to avoid showing the import window at all when the file is empty, and only display the error message. I think it would be reasonable for further work.

Fixes #4346

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -26,10 +26,10 @@ GitHub's online interface.
 For users who previously confirmed the license of their contributions on the
 support site, it would be great if you could add your name below as well.
 
-********************
+---
 
 AMBOSS MD Inc. <https://www.amboss.com/>
-Aristotelis P.  <https://glutanimate.com/contact>
+Aristotelis P. <https://glutanimate.com/contact>
 Erez Volk <erez.volk@gmail.com>
 zjosua <zjosua@hotmail.com>
 Yngve Hoiseth <yngve@hoiseth.net>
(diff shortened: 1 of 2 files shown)


abdnh:
Ideally, we should handle all possible errors. The import pages intentionally display backend errors inline like this, but we can certainly improve styling and hide the HTTP code at least. Here are the files you need to modify:

https://github.com/ankitects/anki/blob/main/ts/lib/components/ErrorPage.svelte
https://github.com/ankitects/anki/blob/main/ts/routes/import-page/ImportPage.svelte

josod827:
> Ideally, we should handle all possible errors. The import pages intentionally display backend errors inline like this, but we can certainly improve styling and hide the HTTP code at least. Here are the files you need to modify:
> 
> https://github.com/ankitects/anki/blob/main/ts/lib/components/ErrorPage.svelte https://github.com/ankitects/anki/blob/main/ts/routes/import-page/ImportPage.svelte

Thank you so much for your feedback! I will definitely check on it.

dae:
Yep, we definitely want to be handling that error in the Svelte side, so the mobile clients handle it too.

dae:
Also, matching based on string values is fragile, and won't work when the interface is in a language other than English. Since you'll need to go back to the drawing board on this PR, please open a new PR if you want to make another attempt.

josod827:
> Also, matching based on string values is fragile, and won't work when the interface is in a language other than English. Since you'll need to go back to the drawing board on this PR, please open a new PR if you want to make another attempt.

Thanks for the feedback! I’ll close this PR and look into adding a structured error code instead of matching on strings.

PR #4349, opened 2025-09-19
Snapshot test update_memory_state

DanielPechersky:
Todo:

- [ ] Pass test
- [ ] Make sure we cover both cards with and without items
- [ ] Add a test for FSRS disabled
- [ ] Make sure snapshot isn't too large
- [ ] I currently use a file "collection.anki21b" shared on the Anki Discord. Need to host this somewhere and download it the first time it's needed.


DanielPechersky:
Note: "collection.anki21b" is taken from here currently https://drive.google.com/file/d/1z7ocz_AyHZsF6jMRgvMlHNU8ytF4nU11/view?usp=sharing

dae:
Jarrett has confirmed a manual check which will suffice, as we don't want to be committing large files into the repo.

PR #4354, opened 2025-09-22
Avoid QProgressBar overflow by normalizing to kilobytes (fixes #4341)

### LINKED ISSUE #4341: Full sync fails with collection >2GB
> Problem Description
> After adding a new field (which triggers a schema change) and choosing the “Upload to AnkiWeb / one-way sync (upload)” option, Anki consistently crashes during the sync progress phase.
> Observed Behavior
> Anki exits with an OverflowError raised from aqt/progress.py while updating the progress bar.
> Hypothesis (for maintainers):
> The stack points to QProgressBar.setMaximum(self._max) receiving a value larger than a 32-bit signed int (INT_MAX = 2,147,483,647). If the progress bar is using raw byte counts for “maximum/ value,” a transfer > ~2 GiB would overflow and crash.

Normalizing current and total to kilobytes would probably be a quick fix.

Originally reported on https://forums.ankiweb.net/t/crashes-on-one-way-sync-sync-server-upload-size-2-gb/66319

matbe766:
When a full one-way upload exceeds 2,147,483,647 bytes, QProgressBar.setMaximum() overflows (int32) and Anki exits with OverflowError.

Fix: If max > INT32_MAX, scale both value and max to kilobytes (ceil division). Small totals keep raw bytes; unknown totals use indeterminate 0..0. Visible labels/UX unchanged—only the internal range/value is scaled.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -243,14 +243,11 @@ Lee Doughty <32392044+leedoughty@users.noreply.github.com>
 memchr <memchr@proton.me>
 Max Romanowski <maxr777@proton.me>
 Aldlss <ayaldlss@gmail.com>
-<<<<<<< HEAD
 Hanna Nilsén <hanni614@student.liu.se>
 Elias Johansson Lara <elias.johanssonlara@gmail.com>
 Toby Penner <tobypenner01@gmail.com>
 Danilo Spillebeen <spillebeendanilo@gmail.com>
-Matbe766 <matildabergstrom01@gmail.com
-
->>>>>>> fcd53376f (qt: normalize large progress totals to kilobytes to avoid QProgressBar overflow (#4341); docs: update CONTRIBUTORS)
+Matbe766 <matildabergstrom01@gmail.com>
 
 
 ********************
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -247,6 +247,9 @@ Hanna Nilsén <hanni614@student.liu.se>
 Elias Johansson Lara <elias.johanssonlara@gmail.com>
 Toby Penner <tobypenner01@gmail.com>
 Danilo Spillebeen <spillebeendanilo@gmail.com>
+Matbe766 <matildabergstrom01@gmail.com
+
+>>>>>>> fcd53376f (qt: normalize large progress totals to kilobytes to avoid QProgressBar overflow (#4341); docs: update CONTRIBUTORS)
 
 ********************
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -243,6 +243,7 @@ Lee Doughty <32392044+leedoughty@users.noreply.github.com>
 memchr <memchr@proton.me>
 Max Romanowski <maxr777@proton.me>
 Aldlss <ayaldlss@gmail.com>
+<<<<<<< HEAD
 Hanna Nilsén <hanni614@student.liu.se>
 Elias Johansson Lara <elias.johanssonlara@gmail.com>
 Toby Penner <tobypenner01@gmail.com>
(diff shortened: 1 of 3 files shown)


abdnh:
For context: #4341

We can't assume the progress dialog will only be used with byte ranges and automatically scale that. That can potentially be unexpected for some callers or add-ons.
 I think this is better handled for full sync specifically:
https://github.com/ankitects/anki/blob/436590f4c2930e0a6d150d2866f83623d9c76d79/qt/aqt/sync.py#L206

matbe766:
> For context: #4341
> 
> We can't assume the progress dialog will only be used with byte ranges and automatically scale that. That can potentially be unexpected for some callers or add-ons. I think this is better handled for full sync specifically:
> 
> https://github.com/ankitects/anki/blob/436590f4c2930e0a6d150d2866f83623d9c76d79/qt/aqt/sync.py#L206

Thank you for feedback! I will look into that asap.

matbe766:
@abdnh I’ve updated the solution with your feedback in mind, I hope I understood your point correctly. Please let me know if I should refine it further.

qt/aqt/sync.py
```diff
@@ -209,11 +209,30 @@ def on_full_sync_timer(mw: aqt.main.AnkiQt, label: str) -> None:
         return
     sync_progress = progress.full_sync
 
+    # If we've reached total, show the "checking" label
     if sync_progress.transferred == sync_progress.total:
         label = tr.sync_checking()
+
+    INT32_MAX = 2_147_483_647
```
abdnh: Sorry, forgot to mention that last time, but we can simplify the fix by just dividing `transferred` and `total` by 1024 and passing that to `mw.progress.update()`?

CONTRIBUTORS
```diff
@@ -243,9 +243,13 @@ Lee Doughty <32392044+leedoughty@users.noreply.github.com>
 memchr <memchr@proton.me>
 Max Romanowski <maxr777@proton.me>
 Aldlss <ayaldlss@gmail.com>
+<<<<<<< HEAD
```
abdnh: Please remove this.

qt/aqt/sync.py
```diff
@@ -209,11 +209,20 @@ def on_full_sync_timer(mw: aqt.main.AnkiQt, label: str) -> None:
...
+    # If we've reached total, show the "checking" label
     if sync_progress.transferred == sync_progress.total:
         label = tr.sync_checking()
+
+    total = sync_progress.total
+    transferred = sync_progress.transferred
+
+    # Always scale to KB to avoid int32 overflow
+    max_for_bar = (total + 1023) // 1024 if total > 0 else 0
+    value_for_bar = (transferred + 1023) // 1024 if total > 0 else None
```
dae: The PR is mostly there, but I have some questions about this part:

- Why do we round up 0-1023 to 1kb?
- Why do we special-case the 0 case?

matbe766: Thanks for the feedback!

I rounded up (total + 1023) // 1024 to avoid progress bars getting “stuck” at 0 KB when there are a few bytes, since a max=0 would display as indeterminate. 

I added the special-case for total == 0 because I thought it represented an “unknown/indeterminate” size, and wanted to show the spinner instead of a full bar.

That said, if using plain total // 1024 and letting 0 be 0 is the expected behavior, I’m happy to adjust the code to match that.

dae: Not really fond of the former - it lies about progress being made, and makes the code harder to read at a glance. The latter also seems unnecessary, as I believe Qt will treat max=0 as indeterminate - but perhaps I'm misremembering?

matbe766: Thanks for clarifying! That makes sense. I’ll simplify the code to use floor division and remove the 0 special-case, so Qt can handle indeterminate progress itself.

dae:
Thanks for bearing with me!

PR #4360, opened 2025-09-23
Fix/first and latest review dates should only  consider entries with a rating.

L-M-Sherlock:
The previous logic may use the date of a rescheduling entry as the latest review, which is not a review.

Before:

image

After:

image

### DIFF
--- a/rslib/src/stats/card.rs
+++ b/rslib/src/stats/card.rs
@@ -76,8 +76,14 @@ impl Collection {
             note_id: card.note_id.into(),
             deck: deck.human_name(),
             added: card.id.as_secs().0,
-            first_review: revlog.first().map(|entry| entry.id.as_secs().0),
-            latest_review: revlog.last().map(|entry| entry.id.as_secs().0),
+            first_review: revlog
+                .iter()
+                .find(|entry| entry.has_rating())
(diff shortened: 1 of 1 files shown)


user1823:
What about this?

```rs
            latest_review: last_review_time,
```

L-M-Sherlock:
>             latest_review: last_review_time,

It also excludes reviews which didn't affect scheduling.

rslib/src/stats/card.rs
```diff
@@ -76,8 +76,14 @@ impl Collection {
             note_id: card.note_id.into(),
             deck: deck.human_name(),
             added: card.id.as_secs().0,
-            first_review: revlog.first().map(|entry| entry.id.as_secs().0),
-            latest_review: revlog.last().map(|entry| entry.id.as_secs().0),
+            first_review: revlog
+                .iter()
+                .find(|entry| entry.has_rating())
+                .map(|entry| entry.id.as_secs().0),
+            latest_review: revlog
```
user1823, line 84: Then, a comment may be better to avoid confusion in the future.
```suggestion
            // last_review_time is not used to ensure cram revlogs are included.
            latest_review: revlog
```

Issue #4153, opened 2025-07-04
Add i18n to launcher


dae:
The launcher currently is all in English. To make it more accessible, we should make the strings translatable. I'd recommend an approach based off our existing anki/i18n module, so we get static type checking like in rslib.

One consideration is that users sometimes have a Windows system language that doesn't match the language they're comfortable with. If we just default to the system language, that can cause some users to struggle (e.g. when installing Anki on a work/school computer in Japan). A button/option to choose a different language might be nice - but is probably too complicated for our existing text interface, so that might be better done after #4152

PR #4361, opened 2025-09-23
feat: add i18n to launcher

iamllama:
Makes progress on #4153. There are still a number of strings left untouched, but i wanted to check if this was acceptable first

I opted to use the `locale_config` crate to get the user's locale, which looks at the `HTTP_ACCEPT_LANGUAGE`, `LANG` and `LC_*` env vars before calling system-specific apis. Perhaps one of those could be set in case the situation mentioned in the issue applies? Or we could just default to `en` for the time being

EDIT: not sure how to fix ci here, builds and runs fine locally


dae:
Thank you very much for looking into this!

The only notable problem I see is it's a bit wasteful - it includes all the core strings as well, turning the 1.6M binary into a 19M one. Maybe we could solve that by having the build script check for an env var, and alter the included files? It will mean a recompilation each time we build rslib vs the launcher, but that's not a big deal. A cleaner but more complicated approach might be to have the crate provide macros to shift the include!()s to the crate that uses anki_i18n rather than i18n itself, but I don't have a strong preference.

Re the lang selection, it's admittedly an uncommon issue, and we could tuck a bunch of the current launcher options behind an advanced screen, which would make it easier to include this too.

I'm guessing CI has unearthed a bug/corner case with our build scripts that is not triggering the recompilation of the anki_i18n crate. When new files are added, they're not picked up by the previously-generated build.ninja, which only lists previously-available files, so the configure process needs to be re-run.

iamllama:
> A cleaner but more complicated approach might be to have the crate provide macros to shift the include!()s to the crate that uses anki_i18n rather than i18n itself

Not sure i understood what you meant here, would the macros include the code to init `I18n` as well? anki_i18n is a workspace package, it's built once and reused for both anki and the launcher (so the codegen only happens once)

Initially tried checking for an env to only process launcher.ftl, but anki failed to rebuild after building the launcher, as the python translations weren't rebuilt. Also tried adding a feature to the i18n crate, but couldn't due to feature unification and features being additive

So i thought to generate a separate `strings_launcher.rs` and conditionally import it via `RUSTFLAGS="--cfg launcher"`. Not fond of the approach ive taked here, but besides duplicating the i18n crate for the launcher, it's the only one ive tried that plays nice with `./run`, `./check` and building the launcher 😓

iamllama:
Pushed a commit (9f90f1c...df743d2) that uses generics to select which set of translations (all or launcher-specfic) to use. This also removes the need to rebuild or pass in envs/rustcfgs when building the launcher

dae:
That's a great solution too - thank you! Out of curiosity, if you run fix:minilints after merging in the latest commit/PR, does your machine still generate licenses in a different order?

> Initially tried checking for an env to only process launcher.ftl, but anki failed to rebuild after building the launcher, as the python translations weren't rebuilt. Also tried adding a feature to the i18n crate, but couldn't due to feature unification and features being additive

Generics were a better solution here anyway, but for future reference, it probably would have been possible to get this working by 1) emitting a pragma in the build script to ensure it gets recompiled when the env var changes (I presume you already knew that part), and 2) maybe_update_env_file() to teach the launcher about it

iamllama:
I ran ./check and ./ninja fix:minilints before pushing, and neither raised issues. Before updating cargo-license, fix:minilints tried to revert the order swap, but now there's no change afterwards 👍

EDIT: @dae sorry, it was fine earlier, but now even with cargo-licence@0.7.0 it's still trying to swap the order again on main 😓

Ran fix:minilints, which reported no work to do, but running check:minilints afterwards still failed with `cargo/licenses.json is out of date; run ./ninja fix:minilints`

Terminal output:
```console
❯ git diff
❯ ./ninja check:minilints
    Finished `release` profile [optimized] target(s) in 0.13s
failed: check:minilints
cargo/licenses.json is out of date; run ./ninja fix:minilints
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.22s
failed: check:minilints
cargo/licenses.json is out of date; run ./ninja fix:minilints

Build failed.
❯ ./ninja fix:minilints
    Finished `release` profile [optimized] target(s) in 0.13s
n2: no work to do

Build succeeded in 0.06s.
❯ git diff
❯
```

dae:
Great - thanks again!

Issue #2694, opened 2023-10-01
Min limit of fuzz should respect the interval scheduled at last review


user1823:
In FSRS, at high difficulties, the stability grows very slowly. When fuzz is enabled, this can mean that after answering a card correctly, the next interval stays the same, or even get reduced. Note that this is not just a theory. Rather, some people (including me) actually faced this issue before a fix was added to the FSRS repo.

What the fix aims to do?
If the `ivl` determined by FSRS is larger than the `scheduledDays` and the fuzz is making the new interval smaller than the `scheduledDays`, we intervene and make the interval greater than scheduledDays.

But, if the ivl determined by FSRS is smaller than the scheduledDays (for any reason such as a change in desired retention), we don't change anything and let the fuzzing work just like it was working previously.

Code:
```js
function apply_fuzz(ivl) {
  if (!enable_fuzz || ivl < 2.5) return ivl;
  ivl = Math.round(ivl);
  let min_ivl = Math.max(2, Math.round(ivl * 0.95 - 1));
  let max_ivl = Math.round(ivl * 1.05 + 1);
  if (is_review()) {
    const scheduledDays = states.current.normal?.review.scheduledDays ? states.current.normal.review.scheduledDays : states.current.filtered.rescheduling.originalState.review.scheduledDays;
    if (ivl > scheduledDays) {
      min_ivl = Math.max(min_ivl, scheduledDays + 1);
    }
  }
```
Link: https://github.com/open-spaced-repetition/fsrs4anki/blob/dfaf660e5aa163402272a8d8b148869a45f09e52/fsrs4anki_scheduler.js#L172-L182

Note: Here, the value of scheduledDays should not be read from the revlog. Reason: It is possible that the card was reviewed on AnkiDroid (or on Desktop before FSRS was enabled). In this case, the revlog entry will correspond to the interval given by SM-2, not the actual interval after which the card was reviewed.

If it is not possible to know the actual scheduled interval of the card, elapsed days can be a reasonable approximation. Reason: The code applies only when the `ivl` determined by FSRS is greater than the previous interval. So, using `elapsedDays` instead of `scheduledDays` just means that the code is not effective in a few cases (where the `ivl` is between the previously scheduled days and the days elapsed since the previous review).

Code of the helper add-on (using elapsedDays):
```py
    last_elapsed_days = (
        int((revlogs[0].time - revlogs[1].time) / 86400) if len(revlogs) >= 2 else 0
    )
[...]

PR #4364, opened 2025-09-25
Fix/Ensure fuzz doesn't go backward during rescheduling

user1823:
Fixes https://github.com/ankitects/anki/issues/2694

### DIFF
--- a/.buildkite/linux/entrypoint
+++ b/.buildkite/linux/entrypoint
@@ -16,6 +16,7 @@ if [ "$CLEAR_RUST" = "1" ]; then
     rm -rf $BUILD_ROOT/rust
 fi
 
+rm -f out/build.ninja
 ./ninja pylib qt check
 
 echo "--- Ensure libs importable"
--- a/rslib/src/scheduler/fsrs/memory_state.rs
+++ b/rslib/src/scheduler/fsrs/memory_state.rs
@@ -137,7 +137,7 @@ impl Collection {
                                             // reschedule it
                                             let original_interval = card.interval;
                                             let greater_than_last = |interval: u32| {
-                                                let previous_interval = last_info.previous_interval as u32
+                                                let previous_interval = last_info.previous_interval.unwrap_or(0);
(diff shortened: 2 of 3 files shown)


Luc-Mcgrady (changes requested):
I get
```
thread '<unnamed>' panicked at rslib/src/scheduler/fsrs/rescheduler.rs:89:17:
attempt to subtract with overflow
```

user1823:
This error doesn't seem related to my changes. Do you encounter this when rescheduling on the `main` branch too?

This error suggests that the `count` on the `due_before` is 0, which is not possible because at least the current card is due on that date, i.e., the count should be at least 1. Pinging @L-M-Sherlock as he is the author of this code.

Luc-Mcgrady (approved):
I do also get this error on main but this prevents me from testing this PR.

Luc-Mcgrady (changes requested):
It runs fine on main :sweat_smile:

Luc-Mcgrady (approved):
Ok it is broken for me on main while reschedule is active, sorry for the spam pings :sweat_smile:

user1823:
CI failure is due to a CI bug. This probably needs a fix too as it frequently appears in PRs and causes confusion.

L-M-Sherlock:
Wait, I have a question: what if the user increases their desired retention drastically? The interval is highly possible to be shorter than the last interval. And it is intended, right?

Edit: nvm, the code only works when the new interval is larger than previous interval.

user1823:
> what if the user increases their desired retention drastically? The interval is highly possible to be shorter than the last interval. And it is intended, right?

Yes, you are right. That's why we have this condition (if interval > previous_interval).

Also, this logic was added to the scheduler way back in 2023 (https://github.com/ankitects/anki/commit/41bddfc6cb17f04383bf486f1cd46036dd6790e9). It's just that it didn't work during rescheduling.

dae:
Thank you @user1823!

PR #4367, opened 2025-09-27
fix(ci): pin cargo-license to 0.7.0

iamllama:
I unknowingly updated to cargo-license@0.7.0 a few days ago, which now normalises licences properly by [sorting ANDs as well](https://github.com/onur/cargo-license/blob/6ec6d98f31cbe58240b64ee486aa1393660bf6d2/src/lib.rs#L538)

https://github.com/ankitects/anki/pull/4361/commits/33299bc052832d098e1fd352ef29ed1fca869af1 made it so that cargo-license@0.7.0 is only installed if not already present, which means that CI is stuck on an older version (that only sorts ORs) while devs might locally run the latest version when `./check`ing. 

This mismatch means one or the other would complain about licenses.json, which this pr proposes to fix

EDIT: @dae the check updated the ci's version of cargo-licenses, so this pr is essentially a fix for the breakage it caused 😭

### DIFF
--- a/cargo/licenses.json
+++ b/cargo/licenses.json
@@ -2226,7 +2226,7 @@
   {
     "authors": "Ibraheem Ahmed <ibraheem@ibraheem.ca>",
     "description": "A high performance, zero-copy URL router.",
-    "license": "MIT AND BSD-3-Clause",
+    "license": "BSD-3-Clause AND MIT",
     "license_file": null,
     "name": "matchit",
     "repository": "https://github.com/ibraheemdev/matchit"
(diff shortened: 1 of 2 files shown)


dae:
Thanks for digging into this ❤️

dae:
Didn't have time to investigate why Windows failed to update to 0.7.0, but I've manually updated its version as well and that's fixed the remaining failure. Does the main branch pass check:minilints for you @abdnh without manually updating it?

abdnh:
Yes, works without a manual update.

PR #4370, opened 2025-09-29
Colored button borders to prevent Hard misuse

Expertium:
Hard misuse is when users think that Hard = fail even though Hard = pass. I focused on this solution out of many for the following reasons:
1) Works on all devices, no need to worry about desktop vs mobile
2) Doesn't require adding entirely new UI elements
3) Doesn't require bikeshedding wording
4) Doesn't disrupt the existing workflow (uh, studyflow?) of users who are used to 4 buttons 

For the sake of accessibility for colorblind people, in the future a second color scheme can be added in Tools -> Preferences, but that's above my paygrade.

Credit for the code goes to rossgb on Discord (https://github.com/rbrownwsws)

### DIFF
--- a/qt/aqt/data/web/css/reviewer-bottom.scss
+++ b/qt/aqt/data/web/css/reviewer-bottom.scss
@@ -28,6 +28,34 @@ button {
     white-space: nowrap;
     margin: 9px;
     position: relative;
+    border-top-color: color(border-subtle);
+}
+
+.answerButton,
+.answerButton:hover {
+    border-style: solid;
+}
+
+.answerButton:focus {
+    border-style: dashed;
+}
+
+.answerButton,
+.answerButton:hover,
+.answerButton:focus {
+    border-width: 2px;
+}
+
+.answerIncorrect,
+.answerIncorrect:hover,
+.answerIncorrect:focus {
+    border-color: #e01b24;
+}
+
+.answerCorrect,
+.answerCorrect:hover,
+.answerCorrect:focus {
+    border-color: #2ec27e;
 }
 
 .hitem {
--- a/qt/aqt/reviewer.py
+++ b/qt/aqt/reviewer.py
@@ -919,6 +919,12 @@ timerStopped = false;
                 extra = """id="defease" """
             else:
                 extra = ""
+
+            if i == 1:
+                button_class = "answerIncorrect"
+            else:
+                button_class = "answerCorrect"
+
             due = self._buttonTime(i, v3_labels=labels)
             key = (
                 tr.actions_shortcut_key(val=aqt.mw.pm.get_answer_key(i))
@@ -926,9 +932,10 @@ timerStopped = false;
                 else ""
             )
             return """
-<td align=center><button %s title="%s" data-ease="%s" onclick='pycmd("ease%d");'>\
+<td align=center><button %s class="answerButton %s" title="%s" data-ease="%s" onclick='pycmd("ease%d");'>\
 %s%s</button></td>""" % (
                 extra,
+                button_class,
                 key,
                 i,
                 i,


Expertium:
This is what it looks like (WSL, Ubuntu, can't build jack shit on my Windows itself)
image

mikehardy:
I think there are reasonable objections on color choice for people with colorblindness but, if there were no borders before, it's progress not a regression so putting it off waiting for colorblind handling seems unreasonable ("prefer good vs a perfect-or-nothing" bias)

So to be clear, I think this is a nifty enhancement

> a second color scheme

But mostly commenting to say I'm lucky to not suffer from any colorblindness so I just recently learned that "adding a second theme" with goal of accessibility is insufficient, there are apparently multiple distinct types of colorblindess which need to be handled with different adaptations. So ideally in future there would be general color choice availability to set a custom theme, and/or some presets targeted at the most common cases. Here's just the first seemingly on-target link I dredged up which adds some color (dad joke) to my comment https://uxdesign.cc/color-blindness-in-user-interfaces-66c27331b858

GithubAnon0000:
The colors look way too attention grabbing; they would certainly distract me. The user cannot use templates to turn those borders off or at least change their colors. Which means the user would have to either get an addon or build Anki from source with the changes included.

Wasn't the goal to have this optional? Adding a way to turn off the colors would be much preffered, maybe in the preferences screen.

Expertium:
Ideally, it should be optional but enabled by default.

ghost:
As someone who inadvertently used "Hard" as "failed" before, even after reading parts of the manual , I think this change would have been helpful. 

It's a small UI change but conveys the meaning beautifully.

Danika-Dakika:
> Ideally, it should be optional but enabled by default.

* But does this PR do that? Or does it impose the colors on everyone -- even existing users? At the very least, it should be disable-able. 

----

* If this is merged, will it trigger as enabled for everyone when they update? Or only for new installs?
* How does it look with dark theme? Same colors? 
* Is it compatible with "Minimalist mode"? I don't know if just the shape of the buttons changes, or there are other differences.

GithubAnon0000:
> * But does this PR do that? Or does it impose the colors on everyone -- even existing users? At the very least, it should be disable-able.
> * If this is merged, will it trigger as enabled for everyone when they update? Or only for new installs?

Everyone will get the colors with the current PR. There is no way to disable it.

> * How does it look with dark theme? Same colors?

Judging by [reviewer-bottom.scss](https://github.com/ankitects/anki/blob/aaa9f8c11eeb18454bbc3990fbdcf938542a4faa/qt/aqt/data/web/css/reviewer-bottom.scss), yes, same colors for dark theme.

jcznk:
> * How does it look with dark theme? Same colors?
> * Is it compatible with "Minimalist mode"? I don't know if just the shape of the buttons changes, or there are other differences.

image

This is how it looks with dark theme and minimalist mode. I personally find it quite distracting and a bit rough-looking (for example, I’d prefer a 1px border instead of 2px). That said, I can see its value for new users. I’d be fine with it as long as it’s easily toggleable.

image

When you focus on a button (“Good” in this example), the border becomes dashed. This feels a bit out of place (I don’t recall this style being used elsewhere in the interface). A subtler approach might look more elegant.

Expertium:
I suppose I'll close the PR then. There is nobody who both _can and wants to_ implement a dropdown menu with Normal, Colorblind-friendly and Off in Tools -> Preferences.
The people who want to don't have the technical skills. The people who have the skills don't want to do it.

jcznk:
I think this is a worthwhile idea and deserves further discussion. Perhaps a good compromise could be to show colored buttons only in the "default" mode, and remove the colors when the interface is set to "minimalist" mode. This would be very easy to implement and could be refined later on.

Expertium:
Sure, but *someone* has to do that work. There are no people who both can and will, as I said.
And due to...uh, certain geopolitical events I can't do international bank transfers and can only pay in crypto, so while I would gladly pay a few hundred bucks for a proper implementation, the number of people who would accept crypto is limited.

GithubAnon0000:
I can help with the preferences part. I already did most of it and only need to find out how to remove the bordered color based on the checkbox value.

jcznk:
@Expertium The solution I suggested would actually be quite trivial to implement. We'd just need to add `.fancy` to the CSS selectors.

Here is an example:

https://github.com/user-attachments/assets/bdf3afb6-146f-4a0f-8dac-d26225c84a93

Expertium:
Thank you, but that's not what I have in mind. I have in mind a dropdown menu in Tools -> Preferences with the following three options:
1) Normal
2) Colorblind-friendly
3) Off

Normal would make borders red/green (like now), Colorblind-friendly would be golden/blue, and Off would be grey

GithubAnon0000:
> I have in mind a dropdown menu in Tools -> Preferences with the following three options:
> 
>     1. Normal
> 
>     2. Colorblind-friendly
> 
>     3. Off

Unless the colorblind PR is merged, I don't think we should do that. But if a checkbox would do, I can probably help, as stated above. The checkbox works as it should, I only need to find out to to access the value to change the colors.

PR

Danika-Dakika:
Thanks for the screenshots! 

I thought of more color questions -- 

1. Does this use the same red as for the Learn/Relearn counter -- and the same green as for the Due counter? (Those seem appropriate, don't they?) I believe those change between dark/light theme. Piggy-backing on those might be more efficient.

> Works on all devices, no need to worry about desktop vs mobile
2. Does it though? Does this integrate with the by-default-already-colored buttons [red, orange/gray, green, blue] on AnkiDroid and AnkiMobile? The blue buttons on AnkiWeb? Or are those study screens built separately, and this feature wouldn't reach them (and would have to be implemented for each separately)? This isn't an all-platforms-or-nothing proposition -- I just think we should know what's going to happen across the ecosystem.

----

> But does this PR do that?
> > No, but the choice is "merge this as is" or "never get colored button borders", because nobody wants to be the one to do it.

I understand why you put this forward, and I know this has been your passion-project for a while -- but that's such a short-sighted view. This very well might be a good way to _teach_ this lesson, but once taught, it doesn't need to be permanent for every user. 

I agree with mikehardy that a lack of colorblindness support doesn't need to block it. I can probably even get on board with imposing it on all users on upgrade [^1] , since we can't tell **_who_** doesn't know this. But I can't see how we could release this with no off-switch. 

[^1]: (wouldn't it be great if it came with some educational messaging too though?)

GithubAnon0000:
Okay, I have implemented it and it seems to work fine on my machine. I can open a new PR if desired, or somehow merge into this one (which I might not have permissions to).

Danika-Dakika:
[Oh, it looks like there was a whole party while I was thinking about colors!]

> The checkbox works as it should

Idea: move that checkbox down adjacent to the other "answer buttons" one. Probably ... after it?

Expertium:
@Danika-Dakika 

> Does it though? Does this integrate with the by-default-already-colored buttons [red, orange/gray, green, blue] on AnkiDroid and AnkiMobile? The blue buttons on AnkiWeb? Or are those study screens built separately, and this feature wouldn't reach them (and would have to be implemented for each separately)? This isn't an all-platforms-or-nothing proposition -- I just think we should know what's going to happen across the ecosystem.

Of course it has to be implemented separately, I was just saying that there are no _fundamental_, PC-specific reasons why it can't be done.

> I understand why you put this forward, and I know this has been your passion-project for a while -- but that's such a short-sighted view.

Why? At the time of writing it there weren't any people willing and capable of implementing a toggle, so my hail mary PR wouldn't be merged because I can't implement a toggle myself. I don't understand what's short-sighted here. And btw, I think that this PR would still be a net improvement without the on/off toggle.

@GithubAnon0000 feel free to make a new PR!

GithubAnon0000:
New PR opened in #4371.

PR #4289, opened 2025-08-26
Feat/Svelte port for reviewer

Luc-Mcgrady:
closes #3871

> [!NOTE]  
> While testing this, make sure to enable "Use new reviewer" in the preferences:
> image

image

I'm not sure if I have the right idea here. If I don't please tell me so I can give up and leave it to the people who probably should be left to do this.

- [x] It doesn't appear until you enter and exit back out, I'm assuming because the height is unknown. This didn't happen when I initially tried it with the ts page rather than the sveltekit one (7e92c4016938be6964d3d85e89bdf8f4c19b873b last good commit). 
- [x] The due dates aren't shown yet (although "due" is passed to the element so it should be quick)
- [x] The down arrow for the more button is a placeholder
- [x] Buttons look weird, I don't think width is set correctly

I'm not sure how faithful to the original html I need to be. If I should use the old table methods, for addon support reasons perhaps, let me know.

Reviewer

- [x] Main window API access
- [X] Mathjax
- - [x] MathJax Font's
- [x] Typed answers
- [X] Audio buttons
- - [x] Autoplay
- [x] Flags display
- [x] "More" menu
- - [x] Flags don't un-set
- - [x] Shortcuts
- - [x] Audio control buttons
- - [x] Add vertical scroll to menu when it overflows the screen(?)
- [x] Timer
- [x] Undo
- [x] Image occlusion is broken for `./tools/runopt`
- [x] Congratulations screen
- [x] Spacebar rates cards option
- [x] Changing flag resets timer
- [ ] Auto Advance Wait for audio
- [x] Preload resources
- [ ] "Processing" equivalent for when `next_card_data` takes a long time.
- [ ] Hooks
- [x] Clicking the toolbar causes enter to do the toolbar action rather than progress the card.
- [x] Holding enter rushes through the cards
- [x] Clicking the edit or more button, then hitting enter, both reviews the card and triggers the button
- [ ] "Card is a leech" tooltip
- [ ] Custom Scheduling
- [ ] Light mode tooltip background colour
- [ ] Finalise field numbers for `NextCardData` protobuf message (Maybe group some fields into separate messages?)
- [ ] Iframes

> [!NOTE]  
> I'll repeatedly edit this comment with any queries I have about how things should be implemented. For the time being.
> - https://github.com/ankitects/anki/pull/4289#issuecomment-3518166035

PR #4371, opened 2025-09-29
FEAT/ADD: colored answer buttons (with option to disable them)

GithubAnon0000:
This is based on #4370. I just took their code and added the preferences toggle.

The preference takes effect after restarting Anki, just like the other ones.

By default, the colors will be shown.

# Images

With option off:
OFF

With option on:
ON


GithubAnon0000:
Actually, for a new profile this bugs out. I'll investigate.

**Edit:** Fixed. I used `true` where I should have used `false` instead. It workes perfectly fine for me now.

jcznk:
Just a nit: as mentioned earlier, I suggest changing

```
.answerButton,
.answerButton:hover,
.answerButton:focus {
    border-width: 2px;
}

```

to use 1px instead. I think this looks more refined and consistent with the rest of the interface, where a 2px border is only used when buttons are focused.

GithubAnon0000:
Done. It now looks like this:

ONE_PX

ghost:
Will it be enabled by default? I am of the opinion that it would be most helpful if it was. The tinkerers amongst us will easily find the option to disable it.

GithubAnon0000:
> Will it be enabled by default?

Yes.

ghost:
Fantastic. Thank you @Expertium and @GithubAnon0000 for this!

Expertium:
Yeah, it doesn't make sense to have it disabled by default since it's supposed to help exactly the kind of person who is unlikely to tinker with settings or read the manual.

rslib/src/config/schema11.rs
```diff
@@ -25,6 +25,7 @@ pub(crate) fn schema11_config_as_string(creation_offset: Option<i32>) -> String
         "schedVer": 2,
         "creationOffset": creation_offset,
         "sched2021": true,
+        "showColor": false,
```
iamllama: This shouldn't be necessary, as per the comments above

GithubAnon0000: I'll remove it. Weirdly, after https://github.com/ankitects/anki/pull/4371/commits/84d887b831cdcf00853f41e2f7a732bd66b088c9 for new and old (i.e. imported collection packages) I get this when I try to review a card and show the backside:

```
Anki 25.09.2 (d1793550) (ao)
Python 3.13.5 Qt 6.9.1 PyQt 6.9.1
Platform: Linux-6.12.19-amd64-x86_64-with-glibc2.41

  File "/home/anon/Downloads/anki/Source_Code/anki/pylib/anki/config.py", line 43, in get_immutable
    return from_json_bytes(self.col._backend.get_config_json(key))
                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/home/anon/Downloads/anki/Source_Code/anki/out/pylib/anki/_backend_generated.py", line 585, in get_config_json
    raw_bytes = self._run_command(9, 0, message.SerializeToString())
  File "/home/anon/Downloads/anki/Source_Code/anki/pylib/anki/_backend.py", line 171, in _run_command
    raise backend_exception_to_pylib(err)
anki.errors.NotFoundError: Your database appears to be in an inconsistent state. Please use the Check Database action. No such value: 'showColor'

The above exception was the direct cause of the following exception:

  File "/home/anon/Downloads/anki/Source_Code/anki/qt/aqt/webview.py", line 745, in handler
    cb(val)
    ~~^^^^^
  File "/home/anon/Downloads/anki/Source_Code/anki/qt/aqt/reviewer.py", line 804, in _onTypedAnswer
    self._showAnswer()
    ~~~~~~~~~~~~~~~~^^
  File "/home/anon/Downloads/anki/Source_Code/anki/qt/aqt/reviewer.py", line 482, in _showAnswer
    self._showEaseButtons()
    ~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/anon/Downloads/anki/Source_Code/anki/qt/aqt/reviewer.py", line 862, in _showEaseButtons
    middle = self._answerButtons()
  File "/home/anon/Downloads/anki/Source_Code/anki/qt/aqt/reviewer.py", line 961, in _answerButtons
    buf += but(ease, label)
           ~~~^^^^^^^^^^^^^
  File "/home/anon/Downloads/anki/Source_Code/anki/qt/aqt/reviewer.py", line 935, in but
    if self.mw.col.conf["showColor"]:
       ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/home/anon/Downloads/anki/Source_Code/anki/pylib/anki/config.py", line 62, in __getitem__
    val = self.get_immutable(key)
  File "/home/anon/Downloads/anki/Source_Code/anki/pylib/anki/config.py", line 45, in get_immutable
    raise KeyError from exc
KeyError

===Add-ons (active)===
(add-on provided name [Add-on folder, installed at, version, is config changed])
AnkiWebView Inspector ['31746032', 2023-06-27T21:26, 'None', '']
Hide Cursor When Idle ['638875964', 2025-04-28T13:58, 'None', '']

===IDs of active AnkiWeb add-ons===
31746032 638875964

===Add-ons (inactive)===
(add-on provided name [Add-on folder, installed at, version, is config changed])
Image Occlusion Enhanced ['1374772155', 2022-04-09T09:15, 'None', '']
Review Heatmap ['1771074083', 2022-06-30T03:43, 'None', '']
Study Time Stats ['1247171202', 2024-02-24T17:59, 'None', '']

```

It claims `showColor` doesn't exist. Tools > Check Database doesn't fix it either. Opening the preferences screen and then closing it solves this though. Do you have an idea why and how to solve it?

**Edit:** Also, the checkbox isn't checked by default anymore, which is unintended.

qt/aqt/preferences.py
```diff
@@ -137,6 +137,7 @@ def setup_collection(self) -> None:
         form.showEstimates.setChecked(reviewing.show_intervals_on_buttons)
         form.showProgress.setChecked(reviewing.show_remaining_due_counts)
         form.showPlayButtons.setChecked(not reviewing.hide_audio_play_buttons)
+        form.showColoredButtons.setChecked(not reviewing.show_colored_buttons)
```
rbrownwsws: If you are inverting the property shouldn't this `not` be removed?
```suggestion
        form.showColoredButtons.setChecked(reviewing.show_colored_buttons)
```

qt/aqt/preferences.py
```diff
@@ -172,6 +173,7 @@ def update_collection(self, on_done: Callable[[], None]) -> None:
         reviewing.show_intervals_on_buttons = form.showEstimates.isChecked()
         reviewing.time_limit_secs = form.timeLimit.value() * 60
         reviewing.hide_audio_play_buttons = not self.form.showPlayButtons.isChecked()
+        reviewing.show_colored_buttons = not self.form.showColoredButtons.isChecked()
```
rbrownwsws: Here too
```suggestion
        reviewing.show_colored_buttons = self.form.showColoredButtons.isChecked()
```

qt/aqt/reviewer.py
```diff
@@ -919,22 +919,42 @@ def but(i: int, label: str) -> str:
...
+
             due = self._buttonTime(i, v3_labels=labels)
             key = (
                 tr.actions_shortcut_key(val=aqt.mw.pm.get_answer_key(i))
                 if aqt.mw.pm.get_answer_key(i)
                 else ""
             )
-            return """
+
+            if self.mw.col.conf["showColor"]:
```
rbrownwsws: This is inverted too?
```suggestion
            if not self.mw.col.conf["showColor"]:
```

rslib/src/config/schema11.rs
iamllama: I reckon you'd have to use `get_config_bool` as it provides a default if the value isn't present, since `col.conf[...]` calls `get_config_json` which expects a value and errors out otherwise
https://github.com/ankitects/anki/blob/a842ba14981ed11664664808887617c42bb186e2/pylib/anki/collection.py?plain=1#L584-L587

This means you can also get rid of `#[strum(to_string = "showColor")]`

qt/aqt/data/web/css/reviewer-bottom.scss
```diff
@@ -28,6 +28,34 @@ button {
     white-space: nowrap;
     margin: 9px;
     position: relative;
+    border-top-color: color(border-subtle);
```
rbrownwsws:
```suggestion
```

I think this shouldn't be here and is from Expertium misusing a diff (it's part of the diff context copied from the `.nightMode` block at the bottom)

rslib/src/config/schema11.rs
GithubAnon0000: Isn't that what I have done in bool.rs and preferences.rs?

The `show_intervals_on_buttons` one gets automatically added to the db but I seem to be doing everything like `show_intervals_on_buttons` did. I'm missing something here. I cannot add `self.get_config_bool()` into collection.py as I'd get

```
  File "/home/anon/Downloads/anki/Source_Code/anki/qt/aqt/webview.py", line 745, in handler
    cb(val)
    ~~^^^^^
  File "/home/anon/Downloads/anki/Source_Code/anki/qt/aqt/reviewer.py", line 804, in _onTypedAnswer
    self._showAnswer()
    ~~~~~~~~~~~~~~~~^^
  File "/home/anon/Downloads/anki/Source_Code/anki/qt/aqt/reviewer.py", line 482, in _showAnswer
    self._showEaseButtons()
    ~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/anon/Downloads/anki/Source_Code/anki/qt/aqt/reviewer.py", line 862, in _showEaseButtons
    middle = self._answerButtons()
  File "/home/anon/Downloads/anki/Source_Code/anki/qt/aqt/reviewer.py", line 961, in _answerButtons
    buf += but(ease, label)
           ~~~^^^^^^^^^^^^^
  File "/home/anon/Downloads/anki/Source_Code/anki/qt/aqt/reviewer.py", line 935, in but
    if not self.get_config_bool(Config.Bool.SHOW_COLORED_BUTTONS):
           ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Reviewer' object has no attribute 'get_config_bool'

Qt warning: QObject::disconnect: wildcard call disconnects from destroyed signal of QPushButton::unnamed 
Qt warning: QObject::disconnect: wildcard call disconnects from destroyed signal of QPushButton::unnamed
```

iamllama: `get_config_bool` is defined on `Collection`, so instead of `self.mw.col.conf["showColor"]` it'd be `self.mw.col.get_config_bool(Config.Bool.SHOW_COLORED_BUTTONS)`

You'll also need to add SHOW_COLORED_BUTTONS to config.proto and modify this:
https://github.com/ankitects/anki/blob/a842ba14981ed11664664808887617c42bb186e2/rslib/src/backend/config.rs?plain=1#L13-L17

GithubAnon0000: Thank you, that seems to have solved the issue completely!

GithubAnon0000:
Regarding https://github.com/ankitects/anki/pull/4370#issuecomment-3348919522 (@Danika-Dakika):

With the last commit, I re-use the colors that ankis color palette is already using. I set the color for light and dark mode to the same color for now, though it can be easily changed later on if desired.

I chose not to use the colors that relearn and learn use. My reasoning is that those colors could change in the future. In that case, the color of the grading button borders should probably still be using green and red. Also: With the latest commit it should be easy to account for color blindness modes as well.


LATEST

dae:
I'm afraid this will need to wait until we get the reviewer ported to Svelte so we avoid duplicate effort. @Luc-Mcgrady has already started to look into it in #4289, and some of this work might be able to be reused there.

Expertium:
> I'm afraid this will need to wait until we get the reviewer ported to Svelte

I could agree if there was a clear deadline for when the migration to Svelte will be done AND the deadline was less than a year away. That is **not** the case.
@dae I'm sure you're getting tired of me complaining, but I want to say: "let's not do any UI-related work for an indefinite amount of time" does not strike me as a reasonable stance.

Btw, Dae, I was hoping you would be more willing to merge this PR since it's not you who had to do this work. I was under the impression that you would be ok with colored button borders but just didn't want to work on that on your own.

Luc-Mcgrady:
It would be me doing the work again of implementing it in the svelte PR, hence duplicate work. If there are any problems with this implementation then any fixes would be temporary and a waste of time. The amount of time until the reviewer is  done is not indefinite as I am actively working on the svelte reviewer.

Porting the reviewer to svelte is difficult as it is. I don't want extra work keeping up with changes that happen outside the pr :sweat_smile:

ghost:
@Luc-Mcgrady Would you be open to this PR including the necessary code to replicate this UI change in Svelte as well?

Maltesaa:
If this is going to be implemented in the Svelte reviewer anyways, and the work for this one is already done how is it duplicate work?

ghost:
@Luc-Mcgrady I can see why it would feel frustrating to put a lot of work to replicate/port an interface that is also a moving target. Separately from this PR, would you be open to (as part of the great work you are doing in #4289) add this UI change to your planned tasks as part the work in the reviewer port, to be worked on when you prefer?

Luc-Mcgrady:
> and the work for this one is already done

Any changes / fixes for this will have to be replicated twice.

> add this UI change to your planned tasks as part the work in the reviewer port, to be worked on when you prefer?

It would be better in a separate PR after the reviewer is ported.

GithubAnon0000:
Maybe there is a way to make both sides happy? What if we merge this PR and you, @Luc-Mcgrady, ignore this feature in your porting efforts? Once your port is finalized, I could open a new PR to reimplement this feature. I assume I should be capable of that (even though I'm a novice), as the most difficult one (the porting itself) would be done by people like you that know what they're doing.

Of course, my suggestion might be a bad idea. I cannot really tell due to limited knowledge. So if it's bad, feel free to openly call it out.

Luc-Mcgrady:
> (even though I'm a novice)

Liar :wink: 

There's a lot of code in this PR which can be re-used by my PR (notably the config bool). I checked and currently the 2 branches have no merge conflicts (yet). It's not only my/your time to waste though. People who are not us will have to review this PR and as I said previously keep up with any associated bugs. Give me time to complete the svelte migration and then you can do whatever you like :).

dae:
> I could agree if there was a clear deadline for when the migration to Svelte will be done AND the deadline was less than a year away. That is not the case.

It's dragged on much longer than I would have liked. A significant factor there is because I've been prioritizing correspondence and PRs over development work in the limited time I've had available. Can you think how your own contributions, however well intentioned (and in most cases useful/appreciated), might have also played a small part here? But the blame ultimately lies on me as the project manager, and I think I need to to shake things up so that I can do more development and spend less time on the other stuff, which I don't really enjoy. That includes getting better at saying no.

GithubAnon0000:
> Liar 😉

It is true, but thank you for the complement!

> It's not only my/your time to waste though. People who are not us will have to review this PR and as I said previously keep up with any associated bugs.

That makes a lot of sense. I'll gradly wait for your PR.

encodemaster:
Now, this can be added to the Labs section.
WDYT @abdnh ?

GithubAnon0000:
We are waiting for https://github.com/ankitects/anki/pull/4289 to be ready first. Then, this PR could be implemented as part of https://github.com/ankitects/anki/pull/4289.

At least that's how I understood it. Though you are right that this could in theory be pushed already (though it would be incompatible with the svelte port).

abdnh:
Yes, this needs to wait until #4289 is ready.

PR #4374, opened 2025-10-02
Improved error message for importing empty CSV-file

### LINKED ISSUE #4346: CSV import error message is not ideal

josod827:
Improves error handling on the import CSV page. It removes the HTTP status code prefix from error messages and makes them more suitable for users, and also adds a Close button.

Fixes #4346

errormsg_anki

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -26,10 +26,10 @@ GitHub's online interface.
 For users who previously confirmed the license of their contributions on the
 support site, it would be great if you could add your name below as well.
 
-********************
+---
 
 AMBOSS MD Inc. <https://www.amboss.com/>
-Aristotelis P.  <https://glutanimate.com/contact>
+Aristotelis P. <https://glutanimate.com/contact>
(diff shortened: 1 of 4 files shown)


dae:
I'm afraid this has some problems:

- pycmd is desktop-specific. All clients have bridgeCommand, but the close handler hasn't been implemented yet. I recommend dropping that change from this PR to make it easier to merge.
- Instead of munging the error text, we should probably just change what we output in post.ts
- This PR contains other unrelated changes

dae:
Actually, it looks like https://github.com/ankitects/anki/pull/4383 is closer to mergable, so I wouldn't recommend spending more time on this.

PR #4375, opened 2025-10-02
Fix launcher strings

user1823:
https://github.com/ankitects/anki/pull/4327 accidentally changed the wrong string. This PR fixes that.

FYI @GithubAnon0000


dae:
Thanks for catching this @user1823

GithubAnon0000:
Thank you @user1823 !

PR #4380, opened 2025-10-03
UPDATE(docs): Add build dependencies and instructions for building / running the launcher

GithubAnon0000:
This is mainly for people like me that experience problems like this one: https://github.com/ankitects/anki/pull/4327. I was confused because I thought ankis build script auto downloaded the dependency for me; turns out, I had to `apt install` it. I'm pretty sure anki does download several build dependencies already, so didn't expect that that wasn't the case here as well.

The building and executing part was quite easy to understand once the dependency issue was solved. I prefer to have actual examples in the docs nonetheless.

This is linux focused (possible even x86_64 focused) as I have no other OS and architecture to test this on.

### DIFF
--- a/.buildkite/linux/entrypoint
+++ b/.buildkite/linux/entrypoint
@@ -23,7 +23,7 @@ echo "--- Ensure libs importable"
 SKIP_RUN=1 ./run
 
 echo "--- Check Rust libs"
-cargo install cargo-deny --version 0.14.24
+cargo install cargo-deny@0.19.0
 cargo deny check
 
 echo "--- Cleanup"
--- a/.deny.toml
+++ b/.deny.toml
@@ -7,6 +7,17 @@ db-urls = ["https://github.com/rustsec/advisory-db"]
 ignore = [
   # burn depends on an unmaintained package 'paste'
   "RUSTSEC-2024-0436",
+  # bincode is unmaintained (via burn). Alternatives: postcard, bitcode, rkyv, wincode
+  "RUSTSEC-2025-0141",
+  # rustls-pemfile is unmaintained. Alternative: use rustls-pki-types directly (PemObject trait)
+  "RUSTSEC-2025-0134",
+  # unic-* crates are unmaintained (used for Unicode category detection).
+  # Alternative: icu_properties
+  "RUSTSEC-2025-0081", # unic-char-property
+  "RUSTSEC-2025-0075", # unic-char-range (or use native Rust char ranges since 1.45.0)
+  "RUSTSEC-2025-0080", # unic-common
+  "RUSTSEC-2025-0094", # unic-ucd-category
+  "RUSTSEC-2025-0098", # unic-ucd-version
 ]
 
 [licenses]
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -255,6 +255,7 @@ Ranjit Odedra <ranjitodedra.dev@gmail.com>
 Eltaurus <https://github.com/Eltaurus-Lt>
 jariji
 Francisco Esteva <fr.esteva@duocuc.cl>
+SelfishPig <https://github.com/SelfishPig>
 
 ********************
 
--- a/qt/aqt/deckbrowser.py
+++ b/qt/aqt/deckbrowser.py
@@ -234,7 +234,7 @@ class DeckBrowser:
         if node.collapsed:
             prefix = "+"
         else:
-            prefix = "-"
+            prefix = "−"
 
         def indent() -> str:
             return "&nbsp;" * 6 * (node.level - 1)
--- a/qt/aqt/errors.py
+++ b/qt/aqt/errors.py
@@ -14,7 +14,7 @@ from markdown import markdown
 
 import aqt
 from anki.collection import HelpPage
-from anki.errors import BackendError, Interrupted
+from anki.errors import BackendError, CardTypeError, Interrupted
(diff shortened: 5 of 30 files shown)


docs/linux.md
```diff
@@ -41,6 +41,14 @@ On some distros such as Arch Linux and Fedora, you may need to install the
 error while loading shared libraries: libcrypt.so.1: cannot open shared object file: No such file or directory
 ```
 
+## Dependencies for building the launcher
+
+If you want to build the launcher, you will need to install the following dependency:
+
+```
+sudo apt install gcc-aarch64-linux-gnu
```
abdnh, line 49: gcc is already listed in the Requirements section, so this is possibly redundant?

docs/development.md
```diff
@@ -195,6 +195,22 @@ If ANKI_PROFILE_CODE is set, Python profiling data will be written on exit.
 - The anki-release package is created/published with the scripts in qt/release.
 - The installer/launcher is created with the build scripts in qt/launcher/{platform}.
 
+On Linux, you can build the launcher by following these steps:
```
abdnh: Would be nice to list other platforms too, since this is not something specific to Linux. There are `mac/build.sh` and `win/build.bat`. The output path is `out/launcher_exe/anki-launcher-25.09.2-windows.exe` on Windows though. We can maybe omit the full path and only mention that the launcher executable can found under `out/launcher`.

docs/linux.md
GithubAnon0000: No, gcc and gcc-aarch64-linux-gnu are different packages. I had gcc installed, of course, but it specifically required gcc-aarch64-linux-gnu to build.

docs/development.md
GithubAnon0000: > We can maybe omit the full path and only mention that the launcher executable can found under out/launcher.

Yes, that's very reasonable! I'll change it. I'll add that the dev has to find and run a launcher.amd64 / launcher.arm64 depending on their architecture; probably only needed for novices though, as more experienced programmers probably figure this out pretty fast.

> Would be nice to list other platforms too, since this is not something specific to Linux. There are mac/build.sh and win/build.bat.

Agreed! I only didn't do it (as mentioned in my PR description) because I cannot test if it actually works the way I wrote. But I did somewhat forget that reviewers can test it themselves. I'll add the mac and win commands.

GithubAnon0000:
@abdnh Please especially check the windows and macos commands / instructions; I know that the Linux instructions and commands are correct because I tried them; but I have no way of testing on a windows or macos machine.

abdnh:
I had to disable code signing on Windows/MacOS to build, but I don't think we need to mention these details. Users advanced enough to build the launcher will figure that out.

docs/development.md
```diff
@@ -195,7 +195,47 @@ If ANKI_PROFILE_CODE is set, Python profiling data will be written on exit.
...
+(on Windows).
+
+For example, on Linux, you can build the launcher by following these steps:
+
+```
+cd ./qt/launcher/lin/
+./build.sh
+```
+
+## Issues during Building
```
dae:
```suggestion
## Issues During Building
```

docs/development.md
```diff
@@ -195,7 +195,47 @@ If ANKI_PROFILE_CODE is set, Python profiling data will be written on exit.
...
+
+If you are experiencing issues building the launcher, make sure that all dependencies
+are installed. See [Building from source](#building-from-source) for more info.
+
+## Running
+
+Once the launcher is built, you can find the executable under `out/launcher` (located
+in the project root). In that folder, you will find a `launcher.amd64` and a
+`launcher.arm64` binary file. Select the one matching your architecture and run
+it to test your changes.
```
dae: This is Linux-specific - Windows and Mac don't have separate binaries per architecture (and Windows has only 1).

docs/linux.md
```diff
@@ -41,6 +41,14 @@ On some distros such as Arch Linux and Fedora, you may need to install the
 error while loading shared libraries: libcrypt.so.1: cannot open shared object file: No such file or directory
 ```
 
+## Dependencies for building the launcher
```
dae: Title Case

GithubAnon0000:
> On Mac, NODMG=1 will skip the slow bundling/code signing, and on Windows set NOCOMP=1 will skip code signing and compression. It might be too optimistic to assume they'll read the build script ^_^;

Dae, are you saying I should add this to the instructions? If it's in the build script then it's done automatically for the dev, right? I'd need a pointer on where / after which instruction to add this info and whether it's optional or necessary to do that.

abdnh:
> If it's in the build script then it's done automatically for the dev, right? I'd need a pointer on where / after which instruction to add this info and whether it's optional or necessary to do that.

NODMG/NOCOMP are environment variables that need to be set to `1` before running the build script (required for anyone other than dae).

GithubAnon0000:
Thank you dae and abdnh! I made some changes.

I have implemented the env var change as text in https://github.com/ankitects/anki/pull/4380/commits/d040148b866896fa12985df0752cd8d1fe5ea86e and then in the table in https://github.com/ankitects/anki/pull/4380/commits/efb1b1a4bcc452438d6d084144356d245f0a3ca3. If you prefer the text one, the last table commit can be reverted. Both commits can stand on their own, they aren't dependent on each other. I'll let you decide.

dae:
Thank you, and apologies for the long wait.

Issue #4358, opened 2025-09-23
Launcher update should open the launcher


dae:
> So, you might want to reconsider that decision.
> Responding to one of the arguments in the commit message:
> 
> But this causes problems when the launcher is updated automatically by things like homebrew.
> 
> If the launcher is updated (even if automatically), the user would want Anki to be updated too. Otherwise, they would have disabled the automatic updates.

Originally reported on https://forums.ankiweb.net/t/launcher-update-should-open-the-launcher/66597

We could perhaps handle both cases by having some flag package managers can set to alter the default behaviour.

dae:
Closed by #4381

PR #4381, opened 2025-10-05
feat: show launcher if different version was installed

iamllama:
ref: #4358

This pr proposes to show the launcher whenever its buildhash is different from what was seen previously

A caveat of this approach is that anki must be built in release mode to update the hash even for launcher-only releases. As for package managers, automatic updates used to mean always getting the latest anki, and if users want to continue doing so, it'd have to be via the launcher in any case

But if that's not acceptable, a possible alternative would be to bake `.version` into the launcher and only show it if the installed anki is older and the user hasn't explicitly chosen to keep/use that version

### DIFF
--- a/qt/launcher/build.rs
+++ b/qt/launcher/build.rs
@@ -7,4 +7,7 @@ fn main() {
             .manifest_required()
             .unwrap();
     }
+    println!("cargo:rerun-if-changed=../../out/buildhash");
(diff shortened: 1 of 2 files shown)


dae:
Thank you for this!

> A caveat of this approach is that anki must be built in release mode to update the hash even for launcher-only releases

It's a potential footgun, but currently my release scripts build wheels then the launcher, so it should be fine unless they're changed down the line (and we could address it then).

PR #4383, opened 2025-10-06
solved issue: error code removed and error page styled and close butt…

medProgAyat:
Screenshot 1404-07-14 at 18 39 06

I solved issue from post.ts where error code added to error agenda for debugging, instead i used this expression: 
```typescript
throw new Error(process.env.NODE_ENV === "production"?`${msg}`:`${result.status}: ${msg}`); 
// which is much more proper that hard coding like removing 500: using regex 
// which can failed in future commits like by producing another code other than 500 
// or even using dash instead of `:` after error code
```
that is according to profile that is used to run program, if producion then status code will be removed, that is more accurate i think.

the main error message is changed to 'The selected file is empty. Please choose a valid file.' that is more user friendly as issued in issue tracker

in implementing error message and its box and button i tried to use _base.scss classes to make error page compatible with theme changes.

and in +page.ts i used @josod827 PR experience which was about producing error from t=other than ErrorPage.svelte and change in that file doesnt shown so i used it commits to implement correct behavoir

### DIFF
--- a/.idea.dist/repo.iml
+++ b/.idea.dist/repo.iml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$">
+      <sourceFolder url="file://$MODULE_DIR$/out/pylib" isTestSource="false" />
+      <sourceFolder url="file://$MODULE_DIR$/pylib" isTestSource="false" />
+      <sourceFolder url="file://$MODULE_DIR$/qt" isTestSource="false" />
+      <excludeFolder url="file://$MODULE_DIR$/extra" />
+      <excludeFolder url="file://$MODULE_DIR$/out/pyenv" />
+    </content>
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
--- a/.vscode.dist/tasks.json
+++ b/.vscode.dist/tasks.json
@@ -12,8 +12,7 @@
                 "command": "tools/ninja.bat",
                 "args": [
                     "pylib",
-                    "qt",
-                    "extract:win_amd64_audio"
+                    "qt"
                 ]
             }
         }
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -189,7 +189,7 @@ Christian Donat <https://github.com/cdonat2>
 Asuka Minato <https://asukaminato.eu.org>
 Dillon Baldwin <https://github.com/DillBal>
 Voczi <https://github.com/voczi>
-Ben Nguyen <105088397+bpnguyen107@users.noreply.github.com>     
+Ben Nguyen <105088397+bpnguyen107@users.noreply.github.com>
 Themis Demetriades <themis100@outlook.com>
 Luke Bartholomew <lukesbart@icloud.com>
 Gregory Abrasaldo <degeemon@gmail.com>
@@ -252,6 +252,9 @@ Amanda Sternberg <mandis.sternberg@gmail.com>
 arold0 <arold0@icloud.com> 
 nav1s <nav1s@proton.me>
 medProgAyat <hosseinmehriyari1383@gmail.com>
+Ranjit Odedra <ranjitodedra.dev@gmail.com>
+Eltaurus <https://github.com/Eltaurus-Lt>
+jariji
 
 ********************
 
--- a/docs/editing.md
+++ b/docs/editing.md
@@ -46,10 +46,14 @@ see and install a number of recommended extensions.
 
 ## PyCharm/IntelliJ
 
-If you decide to use PyCharm instead of VS Code, there are somethings to be
-aware of.
+### Setting up Python environment
 
-### Pylib References
+To make PyCharm recognize `anki` and `aqt` imports, you need to add source paths to _Settings > Project Structure_.
+You can copy the provided .idea.dist directory to set up the paths automatically:
 
-You'll need to use File>Project Structure to tell IntelliJ that pylib/ is a
-sources root, so it knows references to 'anki' in aqt are valid.
+```
+mkdir .idea && cd .idea
+ln -sf ../.idea.dist/* .
+```
+
+You also need to add a new Python interpreter under _Settings > Python > Interpreter_ pointing to the Python executable under `out/pyenv` (available after building Anki).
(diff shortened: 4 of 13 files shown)


ts/lib/components/ErrorPage.svelte
```diff
@@ -2,17 +2,42 @@
...
+<script context="module" lang="ts">
+    declare function pycmd(cmd: string);
+</script>
+
 <script lang="ts">
     export let error: Error;
+
+    function closePage() {
+        try {
+            pycmd("close");
```
abdnh: We can use bridgeCommand() instead:
```diff
diff --git a/ts/lib/components/ErrorPage.svelte b/ts/lib/components/ErrorPage.svelte
index c17f7d273..9230ee2b3 100644
--- a/ts/lib/components/ErrorPage.svelte
+++ b/ts/lib/components/ErrorPage.svelte
@@ -2,16 +2,15 @@
 Copyright: Ankitects Pty Ltd and contributors
 License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 -->
-<script context="module" lang="ts">
-    declare function pycmd(cmd: string);
-</script>
+
 
 <script lang="ts">
+    import { bridgeCommand } from "@tslib/bridgecommand";
     export let error: Error;
 
     function closePage() {
         try {
-            pycmd("close");
+            bridgeCommand("close");
         } catch {
             history.back();
         }
```

ts/lib/components/ErrorPage.svelte
```diff
@@ -2,17 +2,42 @@
...
+            history.back();
+        }
+    }
 </script>
 
-<div class="message">
-    {error.message}
+<div class="error-box">
+    <p class="error-text">{error.message}</p>
+    <button class="btn btn-primary" on:click={closePage}>Got it</button>
```
abdnh: We can use an existing translation message here:
```diff
diff --git a/ts/lib/components/ErrorPage.svelte b/ts/lib/components/ErrorPage.svelte
index c17f7d273..5560d0eba 100644
--- a/ts/lib/components/ErrorPage.svelte
+++ b/ts/lib/components/ErrorPage.svelte
@@ -7,6 +7,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 </script>

 <script lang="ts">
+    import * as tr from "@generated/ftl"
     export let error: Error;

     function closePage() {
@@ -20,7 +21,7 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html

 <div class="error-box">
     <p class="error-text">{error.message}</p>
-    <button class="btn btn-primary" on:click={closePage}>Got it</button>
+    <button class="btn btn-primary" on:click={closePage}>{tr.actionsClose()}</button>
 </div>

 <style lang="scss">
```

ts/lib/components/ErrorPage.svelte
```diff
@@ -2,17 +2,42 @@
...
+</script>
+
 <script lang="ts">
     export let error: Error;
+
+    function closePage() {
+        try {
+            pycmd("close");
+        } catch {
+            history.back();
```
abdnh, line 16: I wonder if this works on mobile. Maybe a better approach is to replace pycmd() with a protobuf RPC. Check deckOptionsRequireClose/deck_options_require_close for an example.

ts/lib/components/ErrorPage.svelte
```diff
@@ -18,7 +20,9 @@ License: GNU AGPL, version 3 or later; http://www.gnu.org/licenses/agpl.html
 
 <div class="error-box">
     <p class="error-text">{error.message}</p>
-    <button class="btn btn-primary" on:click={closePage}>Got it</button>
+    <button class="btn btn-primary" on:click={closePage}>
+        {tr.actionsClose() || "Got it"}
```
abdnh: actionsClose() will always return something meaningful, so no need for the second part.

ts/lib/components/ErrorPage.svelte
abdnh: Yes, I think we need to avoid `history.back()` here and implement the RPC approach.

medProgAyat:
Hi 
I want to use rpc but i need to have subject cleared. So please explain me wat should i do. I mean i used bridging "pycmd jn last commits and command bridge in new one" to close window and if backend isnt a windowed one like qt or other desktop apps then we should try browser api like history.back. then if i want to add rpc so, wats next? I think you mean i should replace command bridge with rpc and then close window from a function inside python, but wat about history.back? It still needed because if we arent at app there isnt any rpc , we are at website browser and it is best practice to send browser history.back()

abdnh:
>  I think you mean i should replace command bridge with rpc and then close window from a function inside python, but wat about history.back? It still needed because if we arent at app there isnt any rpc , we are at website browser and it is best practice to send browser history.back()

Most communication with the backend is happening via Protobuf methods and the page will not function at all without that, so we don't have to fall back to different behavior here. If we implement a method such as deckOptionsRequireClose, a different implementation can be written for each client. AnkiDroid for example has its own implementation: https://github.com/ankidroid/Anki-Android/blob/32a1b44ee14c7e8915ecd427a9fa70f07f93a2ad/AnkiDroid/src/main/java/com/ichi2/anki/pages/DeckOptions.kt#L316C22-L316C45

If the importing page is integrated in a website eventually (AnkiWeb maybe), the website will have to provide its own implementations of some Protobuf methods (specifically, the ones in frontend.proto).

ts/lib/generated/post.ts
```diff
@@ -41,7 +43,12 @@ async function postProtoInner(url: string, body: Uint8Array): Promise<Uint8Array
         } catch {
             // ignore
         }
-        throw new Error(`${result.status}: ${msg}`);
+        // it is ugly to show internall server client status to user, we should just show error's to user
```
dae:
```suggestion
        // hide HTTP status in production builds
```

medProgAyat:
> Mostly looks good, but please either implement the RPC as suggested, or remove the close button.

I’m still a bit confused — it seems like gRPC is mainly used for communication between Python and Rust, but in this case, we’re trying to communicate from Svelte to Python.
Could you please tell me exactly what I should do? You know, this is actually the first time I’ve ever worked with Protobuf, and to make things more complicated, I’ve realized that AQT is also involved here.
Where can I look to better understand the source code so I can write it the way you want?

abdnh:
@medProgAyat Have you checked `deckOptionsRequireClose` and `deck_options_require_close` in the source as I suggested? That should give you a good idea.

dae:
Happy to re-open if you plan to return to this.

PR #4385, opened 2025-10-09
feat: add tooltips for sidebar items in the browse cards window 

nav1s:
This pr makes it easier to view hidden items in the sidebar, as shown here:

anki-pr

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -250,7 +250,7 @@ Danilo Spillebeen <spillebeendanilo@gmail.com>
 Matbe766 <matildabergstrom01@gmail.com>
 Amanda Sternberg <mandis.sternberg@gmail.com>
 arold0 <arold0@icloud.com> 
-
+nav1s <nav1s@proton.me>
 
 ********************
 


GithubAnon0000:
> This pr makes it easier to view hidden items in the sidebar, as shown here:

What do you mean? What are "hidden items in the sidebar" and how did it look before?

**Edit:** Actually, looking at your code, I think I understand. If the sidebar is too small then the text of sidebar children might be cut off. You propose a fix where the text gets shown in the on_hover popup window.

nav1s:
> > This pr makes it easier to view hidden items in the sidebar, as shown here:
> 
> What do you mean? What are "hidden items in the sidebar" and how did it look before?
> 
> **Edit:** Actually, looking at your code, I think I understand. If the sidebar is too small then the text of sidebar children might be cut off. You propose a fix where the text gets shown in the on_hover popup window.

Yeah, that's exactly what I meant, should have explained it more clearly in the pr description.
Edit: also currently we have similar behavior for the note descriptions but not for the sidebar as shown here:

anki-pr2

PR #4386, opened 2025-10-10
fix: make uv use system certstore

iamllama:
ref: https://forums.ankiweb.net/t/invalid-peer-certificate-issue-with-latest-version/66945?u=llama

uv normally relies on the `webpki-root` crate when loading certs ([docs](https://docs.astral.sh/uv/concepts/authentication/certificates/#tls-certificates)), which doesn't necessarily include custom certs that might be needed when behind a proxy


dae:
Thank you!

PR #4388, opened 2025-10-12
Fix(launcher)/Exclude pre-release if a newer major_minor exists

user1823:
The current main produces a list like "25.09.2, 25.08b5, 25.07.5, 25.06b7, 25.02.7" Here, 25.08b5 and 25.06b7 should be filtered out for the same reason this code filters out the older patch releases.


dae:
Thank you!

PR #4389, opened 2025-10-13
Update PyCharm docs

abdnh:
This updates the PyCharm docs about making PyCharm recognize `anki` and `aqt` imports. I also added a basic `idea.dist` for that.


dae:
Thanks Abdo :-)

PR #4391, opened 2025-10-14
Enable FSRS by default

### LINKED ISSUE #3616: Make FSRS the default?
In the next non-trivial (not 24.11.x) update, I think it's about time we enable FSRS out of the box. Any objections?

L-M-Sherlock:
Make FSRS (Free Spaced Repetition Scheduler) enabled by default for new users and users who haven't explicitly configured it.

Changes:
- Add BoolKey::Fsrs to default-true boolean keys in Rust backend
- Set defaultValue to true for FSRS toggle in deck options UI

close #3616

For AO, please create a new issue to keep track of it.

### DIFF
--- a/proto/anki/config.proto
+++ b/proto/anki/config.proto
@@ -57,6 +57,7 @@ message ConfigKey {
     LOAD_BALANCER_ENABLED = 26;
     FSRS_SHORT_TERM_WITH_STEPS_ENABLED = 27;
     FSRS_LEGACY_EVALUATE = 28;
+    FSRS = 29;
   }
   enum String {
     SET_DUE_BROWSER = 0;
--- a/pylib/tests/shared.py
+++ b/pylib/tests/shared.py
@@ -9,6 +9,7 @@ import tempfile
 import time
 
 from anki.collection import Collection as aopen
+from anki.config import Config
 
 # Between 2-4AM, shift the time back so test assumptions hold.
 lt = time.localtime()
@@ -49,6 +50,8 @@ def getEmptyCol():
     (fd, path) = tempfile.mkstemp(suffix=".anki2")
     shutil.copy(_emptyCol, path)
     col = aopen(path)
+    # Disable FSRS for legacy scheduler tests
+    col.set_config_bool(Config.Bool.FSRS, False, undoable=False)
     return col
 
 
--- a/pylib/tests/test_schedv3.py
+++ b/pylib/tests/test_schedv3.py
@@ -10,6 +10,7 @@ from typing import Dict
 import pytest
 
 from anki import hooks
+from anki.config import Config
 from anki.consts import *
 from anki.lang import without_unicode_isolation
 from anki.scheduler import UnburyDeck
@@ -19,6 +20,8 @@ from tests.shared import getEmptyCol as getEmptyColOrig
 
 def getEmptyCol():
     col = getEmptyColOrig()
+    # Disable FSRS for legacy scheduler tests
+    col.set_config_bool(Config.Bool.FSRS, False, undoable=False)
     return col
 
 
--- a/rslib/src/backend/config.rs
+++ b/rslib/src/backend/config.rs
@@ -40,6 +40,7 @@ impl From<BoolKeyProto> for BoolKey {
             BoolKeyProto::LoadBalancerEnabled => BoolKey::LoadBalancerEnabled,
             BoolKeyProto::FsrsShortTermWithStepsEnabled => BoolKey::FsrsShortTermWithStepsEnabled,
             BoolKeyProto::FsrsLegacyEvaluate => BoolKey::FsrsLegacyEvaluate,
+            BoolKeyProto::Fsrs => BoolKey::Fsrs,
         }
     }
 }
--- a/rslib/src/config/bool.rs
+++ b/rslib/src/config/bool.rs
@@ -79,6 +79,7 @@ impl Collection {
             | BoolKey::RestorePositionReviewer
             | BoolKey::LoadBalancerEnabled
             | BoolKey::FsrsHealthCheck
+            | BoolKey::Fsrs
             | BoolKey::NormalizeNoteText => self.get_config_optional(key).unwrap_or(true),
 
             // other options default to false
--- a/rslib/src/config/schema11.rs
+++ b/rslib/src/config/schema11.rs
@@ -25,6 +25,7 @@ pub(crate) fn schema11_config_as_string(creation_offset: Option<i32>) -> String 
         "schedVer": 2,
         "creationOffset": creation_offset,
         "sched2021": true,
+        "fsrs": true,
     });
     serde_json::to_string(&obj).unwrap()
 }
--- a/rslib/src/deckconfig/update.rs
+++ b/rslib/src/deckconfig/update.rs
@@ -501,7 +501,7 @@ mod test {
             limits: Limits::default(),
             new_cards_ignore_review_limit: false,
             apply_all_parent_limits: false,
-            fsrs: false,
+            fsrs: true,
             fsrs_reschedule: false,
             fsrs_health_check: true,
         };
--- a/rslib/src/scheduler/answering/mod.rs
+++ b/rslib/src/scheduler/answering/mod.rs
@@ -706,6 +706,8 @@ pub(crate) mod test {
     #[test]
     fn state_application() -> Result<()> {
         let mut col = Collection::new();
+        // Disable FSRS for this legacy scheduler test
+        col.set_config_bool(BoolKey::Fsrs, false, false)?;
         if col.timing_today()?.near_cutoff() {
             return Ok(());
         }
--- a/rslib/src/scheduler/queue/builder/mod.rs
+++ b/rslib/src/scheduler/queue/builder/mod.rs
@@ -431,6 +431,8 @@ mod test {
     #[test]
     fn review_queue_building() -> Result<()> {
         let mut col = Collection::new();
+        // Disable FSRS for this legacy scheduler test
+        col.set_config_bool(BoolKey::Fsrs, false, false)?;
 
         let mut deck = col.get_or_create_normal_deck("Default").unwrap();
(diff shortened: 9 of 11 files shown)


JSchoreels:
Should we include some kind of banner the first time that option will be by default used ? Just in case some people where on SM2 by default but who'd still want to switch back to SM2 if they really want it

jakeprobst:
its worth noting that with this change you will be enabling fsrs for everyone that has not explicitly turned fsrs on then back off.
https://github.com/ankitects/anki/blob/04a0b10a15abc61c409d0073eb856fe75db3f6f2/rslib/src/deckconfig/update.rs#L212-L214 There is logic that keeps the variable unsaved in the configuration unless it has changed. And since its unsaved the default-to-true gets triggered even if it is a preexisting deck.
I can see an argument that this might be a desireable interaction but I think this case should be known and discussed.


(I played around with this a few weeks back as I was gonna make this exact PR next april fools day. I thought things would move slowly enough that it would still be a relevant PR to make then)

L-M-Sherlock:
> Should we include some kind of banner the first time that option will be by default used ? Just in case some people where on SM2 by default but who'd still want to switch back to SM2 if they really want it

I'm not good at UI/UX, so PR is welcome.

> I can see an argument that this might be a desireable interaction but I think this case should be known and discussed.

Sure.

ghost:
FSRS has clear advantages over SM-2, and because of the "Default Effect" most people won't be likely to change the default unless something has gone wrong.

iamllama:
We could perhaps enable it for new collections only by adding a default here: https://github.com/ankitects/anki/blob/362d09e31f17c234c9635a6836501820dc2b1215/rslib/src/config/schema11.rs?plain=1#L27-L28

L-M-Sherlock:
> We could perhaps enable it for new collections only by adding a default here:

I have tested the PR's code. It has enabled FSRS by default for new collections.

Edit: oh, I know what happened. For a new collection, the config doesn't have the `fsrs` field. The PR will return true for `fsrs` by default in this case.

user1823:
This seems to be a good opportunity to make some changes similar to https://github.com/ankitects/anki/pull/3860 so that we don't need to read the config on every card answer.

https://github.com/ankitects/anki/blob/321e23acb2b02d657ca08727546c016d7ef27381/rslib/src/scheduler/answering/mod.rs#L449

Please feel free to ask me to create a separate issue for this.

L-M-Sherlock:
> Please feel free to ask me to create a separate issue for this.

Sure, please do that, thank you!

tatsumoto-ren:
@L-M-Sherlock   if you want more users to test FSRS, consider shipping Anki with no default scheduler. When a new collection (new Anki profile) is created, force the user to choose one or the other.

Also, a look through the Anki Forums shows many users confused or having problems with FSRS, which indicates it isn't ready yet.

L-M-Sherlock:
> if you want more users to test FSRS, consider shipping Anki with no default scheduler. When a new collection (new Anki profile) is created, force the user to choose one or the other.

To be honest, @maimemo has been doing something similar. We shipped different schedulers all the time for A/B tests.

> Also, a look through the Anki Forums shows many users confused or having problems with FSRS, which indicates it isn't ready yet.

If there are a lot of users confused or having problems, FSRS will get fixed soon. Just like SM-2 as the default, the community has invented tons of methods to fix it.

tatsumoto-ren:
A less intrusive option is to have a popup on a new collection asking the user which scheduler they want to choose.

501581178-dba09644-7f23-4e2a-84e3-6b002fe35fad

Example how to make a message like this: https://github.com/ankitects/anki/blob/f94d05bcbe17c8d1eacad3178c77f58c9fb3c3bd/qt/aqt/deckbrowser.py#L407C1-L429C4

GithubAnon0000:
> A less intrusive option is to have a popup on a new collection asking the user which scheduler they want to chose.

In that case the message would need to be neutral, without bias. Your message is biased. And inaccurate for that matter. You do not *have* to optimize FSRS at all, as I already covered in the linked issue. It will improve retention though, if you do. Which is a huge plus compared to a non-dynamic, never adjusting algorithm like SM-2.

Also: The common user has no idea which algorithm is better for them; and probably will not read even a short explaination section. So I reccon it's better to just have a default (which might be FSRS if there are no real disadvatages and if its superior results reported by users are convincing dae) and allow users to change it in the preset options

tatsumoto-ren:
> In that case the message would need to be neutral, without bias

Yeah, I think it can be adjusted. The important part is to warn users that FSRS is experimental and that there's no solid data on how it affects retention given the same number of daily reviews. Some users report higher retention, others lower. Including a link to an article that explains the differences would also be helpful.

nezmaka:
> Yeah, I think it can be adjusted. The important part is to warn users that FSRS is experimental and that there's no solid data on how it affects retention given the same number of daily reviews. Some users report higher retention, others lower. Including a link to an article that explains the differences would also be helpful.
 
Iirc, FSRS is based on an algorithm that has had several papers published about it, showing improved retention. Users' so called anecdotal reports about their own retentions are not exactly reliable.  

For what it's worth, I find that even without the rather misleading note, the mere calling of the one ‘Anki Scheduler’ and the other ‘FSRS’ biases towards the former. Had I not known anything and just installed Anki for the first time, without the time to look into the different schedulers, I'd have chosen Anki Scheduler while I wish I'd have chosen FSRS.

dae:
@L-M-Sherlock normally we perform a bunch of computations when the user enables FSRS. Changing the default value of this bool will mean users that were on SM2 will suddenly have the code treating their deck like an FSRS one, despite those computations not having taken place. I know we sometimes generate memory state on the fly, but I wonder if this approach might lead to problems?

Perhaps we should be considering a more conservative approach that enables FSRS for new users, but does not change the behaviour for existing users? That's going to generate a lot less confusion/support requests when people upgrade.

Given how the Hard button still confuses some people, perhaps we should consider having this PR wait until we have button colors or a 2 button mode in place?

L-M-Sherlock:
Since I’ve also made several breaking changes in the fsrs-rs crate, I plan to test the latest code in Anki once #4335 is merged. Let's hold off on this PR until that testing is complete.

L-M-Sherlock:
Due to my recent experiments, I think it's better to close this PR at first because FSRS-6 with default parameters doesn't outperform SM-2 in scheduling. I will reopen it until I find a better method to solve this problem

related post: https://x.com/JarrettYe/status/2021539338958610607

image

GithubAnon0000:
@L-M-Sherlock Approximately half a year has passed: in your last post you said FSRS might not outperform SM-2 and that you're looking for a better method to solve this problem. I'm wondering if things have changed and if you, managed to make progress on this issue.

tatsumoto-ren:
@GithubAnon0000 FSRS appears to be a dead end. The best solution would be to deprecate it and invent a completely new scheduling algorithm. I don't think FSRS will ever be "ready", given that even its creators have admitted it.

Expertium:
I don't particularly care since I plan to make my own fork of Anki, but just to be fair, Jarrett's simulations were with **default**, not optimized parameters. Also, FSRS-6 is not the latest version, FSRS-7 is.

GithubAnon0000:
@tatsumoto-ren It still seems to be the best option right now, for me. At least based on my study success and statistics, I am happy with the results. I did not have the same results with SM-2 when I was using it; FSRS feels like a big upgrade to me.

Out of interest: You would like to create a completely new scheduling algorithm? What would this algorithm do differently, that makes it more effective? It would also need to be validated with studies, which FSRS already is doing.

GithubAnon0000:
@Expertium What would make your fork different? Are you just going to update the default scheduler or also other things?

And yes, I do know it was about default params, not optimized params.

tatsumoto-ren:
@Expertium, making a fork might be a good idea if you can maintain it. You could fork the last decent version (around 2.1.20) and then backport only the necessary bug fixes and improvements.

> What would this algorithm do differently, that makes it more effective?

@GithubAnon0000 I would use the so-called  [low-key Anki setup](https://ajatt.top/blog/setting-up-anki.html#starting-ease) (based on SM-2, but with a fixed ease), then make the intervals grow less over time so they cannot go up to 10+ years. The curve should look something like this.

govno

Expertium:
I would integrate FSRS-7 (no idea when Anki will switch from 6 to 7) into my fork, as well as a neural network. That neural network will use not only interval lengths and grades, like FSRS, but also a lot of other input features, such as:
1. Card->note->deck->preset hierarchy
2. Review duration, like 7 seconds or 15 seconds
3. Time of day, like 2:15 PM
4. Day of the week/month/year/decade/century 
5. Number of reviews/new cards done today (what the heatmap add-on shows)
6. Card age (sum of all intervals)
7. Collection age (time since the first review of the first card)
8. Card creation date, and maybe deck and preset creation dates, if they prove to be useful

And others that are basically "let's look at the same data from a different angle".
srs-benchmark has the RWKV (pronounced 'rwa-kuv') neural net, that's what I'm using as the foundation. I'll make it more efficient and more accurate.

In terms of user experience, there will be FSRS-7, RWKV-Curve (what is currently called RWKV in srs-benchmark) and RWKV-Instant (what is currently called RWKV-P).
With RWKV you won't have to click "Optimize" since parameters are trained on 5k users and then frozen.
RWKV parameter optimization

Curve will be similar to FSRS. Instant will be...anarchy. No such thing as due dates or interval lengths. **Number of due cards changes every time you do a review.** Desired retention is still a meaningful knob, it controls what cards are considered due, but the relationship between workload and DR will no longer be straightforward. Unlike Curve, RWKV-Instant doesn't have a forgetting curve, it only predicts probability of recall right now and says nothing about what p(recall) will be tomorrow or in 10 years. And according to our benchmark, it beats FSRS-7 on 100.0% of users. Though the FSRS-7 that is currently in srs-benchmark is slightly worse than the final version (I'll PR it later), RWKV-Instant beats both versions anyway, on 100.0% of users. RWKV-Curve beats the final version of FSRS-7 on 93.3% of users (that is with same-day reviews included).

No SM-2 btw.

PR #4393, opened 2025-10-16
Feat/Prioritize previously attempted cards in intraday learning queue

user1823:
The Problem
The "learn now" add-on can bulk-add many new cards to the learn queue with due == now. When a user answers some of those and fails them, the failed cards are scheduled for a few minutes later, but the current queue-building behavior requires the user to finish the remaining "learn now" cards (due == now) before the failed cards are shown again. This increases the chance of repeated failures and reduces learning efficiency.

The Change
The queue builder now partitions intraday learning cards into two groups: previously-attempted (reps > 0) and never-attempted (reps == 0). Within each group cards remain ordered by due time. The builder concatenates the groups with previously-attempted cards first.

Why this fixes the problem
Failed learning cards (which now have reps > 0 and a later due timestamp) will be prioritized ahead of never-attempted cards added by "learn now". This ensures the user sees the failed cards as soon as they are due, reducing repeated failures.

Why changes to Anki rather than the add-on
As far as I know, there is nothing that the add-on can do to improve this situation. Secondly, I think that Anki should, at some point, natively include the "Learn now" feature as it is quite useful when the user wants to introduce specific cards (for e.g., for an upcoming exam) without creating a temporary deck for that purpose. "Learn now" is also a good way to solve one aspect of the Set Due Date misuse problem (see https://github.com/open-spaced-repetition/fsrs4anki/issues/675#issuecomment-2548839346).


abdnh (approved):
Looking good. Thank you!

PR #4398, opened 2025-10-18
(UI polish) Filtered decks: replace maximumSize with sizePolicy in "Limit to" QSpinBox

jcznk:
**Context**

This PR updates the `Limit` and `Limit_2` `QSpinBox` widgets in the filtered deck creation dialog.
It’s an updated version of a previous PR.

(Note: this still includes the first commit from the original version of the PR. I wasn’t sure whether it’s necessary to remove it or if it just remains for history, since the “Files changed” section now shows only the intended modifications.)

Tested on: Windows 11

**Problem**

Previously, Anki set a fixed maximum width of 60 px for these spin boxes.
This could prevent the full content from being displayed, depending on input length, system font size, or UI scaling.

**Solution**
1. Remove maximumSize. This allows Qt to automatically determine an appropriate width based on the maximum input value.
2. Apply a fixed-size policy. This ensures the spin box doesn’t stretch horizontally when the dialog is resized.

Before:

image

After: 

image

### DIFF
--- a/qt/aqt/forms/filtered_deck.ui
+++ b/qt/aqt/forms/filtered_deck.ui
@@ -85,11 +85,11 @@
       </item>
       <item row="2" column="2">
        <widget class="QSpinBox" name="limit">
-        <property name="maximumSize">
-         <size>
-          <width>75</width>
-          <height>16777215</height>
-         </size>
+        <property name="sizePolicy">
+         <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
+          <horstretch>0</horstretch>
+          <verstretch>0</verstretch>
+         </sizepolicy>
         </property>
         <property name="minimum">
          <number>1</number>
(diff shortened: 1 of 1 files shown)


GithubAnon0000:
Does this scale if the user has a big font size set, e.g. because of accessibility? Let's say the font size is 24 instead of 12, does the space still suffice?

I would assume the answer is yes, since I'd expect QT to handle that properly. But if not, maybe a new issue should be opened (or this PR extended).

jcznk:
@GithubAnon0000 Good catch! It actually doesn’t scale when the system font size changes.

I opted for a different approach (tested only on Windows 11):

1. Remove any references to `maximumSize` (or `minimumSize`). This lets Qt determine an appropriate minimum size based on the maximum input length and other factors.
2. Apply a size policy that fixes the QSpinBox’s width, so it doesn’t expand when the parent window is resized.

GithubAnon0000:
I tested https://github.com/ankitects/anki/pull/4398/commits/b29c4678e8cc6e73d12e8518d4565455a62a7fed on debian linux (sid) using gnome DE. It appears to be working great for me, too.

dae:
Thank you both!

PR #4404, opened 2025-10-19
fix(launcher): apply cache settings to all uv invocations

### LINKED ISSUE #4402: launcher reading from system-wide cache?
I was under the impression that the flags we pass to uv keep all of its files inside AnkiProgramData, but the following report seems to imply that's not the case:

https://forums.ankiweb.net/t/new-anki-update-not-working/66664

We'll want to try reproduce the issue, and if confirmed it's not just a read-only check, see if we can avoid it, so that an uninstall leaves the system in the same state it was prior to install.

iamllama:
Closes #4402

Currently, we only set cache settings (`UV_NO_CACHE` or `UV_CACHE_DIR`) when syncing, but not when checking the existing version or getting releases (which do write to the cache)

uv's [default caching behaviour](https://docs.astral.sh/uv/concepts/cache/#cache-directory) is to write to an external cache directory, and this pr proposes to make it cache in `AnkiProgramFiles` if applicable

### DIFF
--- a/.idea.dist/repo.iml
+++ b/.idea.dist/repo.iml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$">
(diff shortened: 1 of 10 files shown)


dae:
Thanks for digging into this!

PR #4407, opened 2025-10-25
Fix unescaped HTML in correct type-in answers

Eltaurus-Lt:
This is to patch a regression that occurred in [dc5fa60](https://github.com/ankitects/anki/commit/dc5fa60c8b5fa561b7a9d45faf106e646acc3a60): by skipping tokenization of correct inputs, it made them bypass html escaping  that was part of `render_tokens`, which caused the following issue: https://forums.ankiweb.net/t/basic-type-in-the-answer-strips-text-inside-when-answer-is-correct/67112

This PR restores the escaping and adds a unit test to catch such issues in the future. I can add more tests to cover non-combining comparison, and also incorrect and empty answers, if that's appropriate (the last one had [an issue before](https://github.com/ankitects/anki/pull/2658/commits/d73c09c38004b9dd3249379eca426bebcdac5072) as well). 

PS: I'm still figuring out how to build the repo, so the PR wasn't ninja-checked yet, only the functions themselves were tested in a standalone script.

### DIFF
--- a/rslib/src/typeanswer.rs
+++ b/rslib/src/typeanswer.rs
@@ -58,7 +58,7 @@ trait DiffTrait {
         if self.get_typed() == self.get_expected() {
             format_typeans!(format!(
                 "<span class=typeGood>{}</span>",
-                htmlescape::encode_minimal(self.get_expected_original().into())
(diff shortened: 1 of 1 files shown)


dae:
Thank you!

PR #4409, opened 2025-10-29
tweak the hard delay for first step

L-M-Sherlock:
Currently, it's the average of the first step (again interval) and the second step (good interval).

According to my preliminary analysis on Anki-revlog-10k dataset, the hard stability is closer to again stability instead of good stability.

I use the formula, S(hard) = a * S(again) + (1-a) * S(good), to fit the data, and the average coefficient a is 0.777729. For the sake of elegance, I arbitrarily pick 0.75 for this PR.

For more details, please see:

https://github.com/open-spaced-repetition/short-term-memory-research/blob/2782cf80d6c002acd10c65143fea2662ed804ff3/short_term_analysis.ipynb

### DIFF
--- a/pylib/tests/test_schedv3.py
+++ b/pylib/tests/test_schedv3.py
@@ -490,14 +490,14 @@ def test_nextIvl():
     ##################################################
     ni = col.sched.nextIvl
     assert ni(c, 1) == 30
-    assert ni(c, 2) == (30 + 180) // 2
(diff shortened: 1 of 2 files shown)


L-M-Sherlock:
Someone complained that this PR will cause more `Hard` abuse problem, so I close it.

Here is my alternative solution based on custom scheduling: https://github.com/open-spaced-repetition/anki-deck-specific-first-step

PR #4410, opened 2025-10-29
Fix/Exclude BackendFrontendService from write_python_interface

Luc-Mcgrady:
Ref:
- https://github.com/ankitects/anki/pull/4289#issuecomment-3442123498

c.c. @iamllama (I fixed it :wink:)

### DIFF
--- a/rslib/proto/python.rs
+++ b/rslib/proto/python.rs
@@ -22,7 +22,7 @@ pub(crate) fn write_python_interface(services: &[BackendService]) -> Result<()> 
     write_header(&mut out)?;
 
     for service in services {
-        if service.name == "BackendAnkidroidService" {
+        if ["BackendAnkidroidService", "BackendFrontendService"].contains(&service.name.as_str()) {
             continue;
         }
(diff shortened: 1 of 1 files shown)


abdnh:
Frontend RPCs don't work here anyway and they are excluded from the Rust interface. Try this in the debug console:
```python
mw.col._backend.get_scheduling_states_with_context()
```

I think we should exclude them instead:
```diff
diff --git a/rslib/proto/python.rs b/rslib/proto/python.rs
index a5adb4179..5c245de1d 100644
--- a/rslib/proto/python.rs
+++ b/rslib/proto/python.rs
@@ -22,7 +22,7 @@ pub(crate) fn write_python_interface(services: &[BackendService]) -> Result<()>
     write_header(&mut out)?;

     for service in services {
-        if service.name == "BackendAnkidroidService" {
+        if ["BackendAnkidroidService", "BackendFrontendService"].contains(&service.name.as_str()) {
             continue;
         }
         for method in service.all_methods() {
```

PR #4413, opened 2025-11-01
Fix/Per deck retention not used when card is moved to filtered deck

Luc-Mcgrady:
forum link:
- https://forums.ankiweb.net/t/custom-decks-ignoring-subdeck-s-desired-retention/67282/2

### DIFF
--- a/rslib/src/scheduler/answering/mod.rs
+++ b/rslib/src/scheduler/answering/mod.rs
@@ -443,10 +443,14 @@ impl Collection {
             .storage
             .get_deck(card.deck_id)?
             .or_not_found(card.deck_id)?;
-        let home_deck = self
-            .storage
-            .get_deck(card.original_or_current_deck_id())?
-            .or_not_found(card.original_or_current_deck_id())?;
+        let home_deck = if card.original_deck_id.0 == 0 {
+            &deck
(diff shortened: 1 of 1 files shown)


rslib/src/scheduler/answering/mod.rs
```diff
@@ -443,9 +443,20 @@ impl Collection {
             .storage
             .get_deck(card.deck_id)?
             .or_not_found(card.deck_id)?;
-        let config = self.home_deck_config(deck.config_id(), card.original_deck_id)?;
+        let home_deck = if card.original_deck_id.0 == 0 {
```
abdnh, line 446: Minor suggestion: Change `home_deck_config()` to `home_deck()` and use it instead to reduce some duplication.

Luc-Mcgrady: https://github.com/ankitects/anki/blob/acdf486b290bd47d13e2e880fbb1c14773899091/rslib/src/scheduler/states/fuzz.rs#L54

I think the reason I didn't do this is because the other instance where `home_deck_config` is used, you don't need the full home_deck info and only need the home info's deckId to get the config; so it would be an extra unused read.

PR #4414, opened 2025-11-01
feat(launcher): embed python

### LINKED ISSUE #4151: Consider integrating Python in the launcher

iamllama:
Closes #4151

Looked into why embedding statically/dynamically python would need version-scoped installers, and found that [PyConfig's layout](https://github.com/PyO3/pyo3/blob/67da00d07da5a5755344d101663e69e9209554bb/pyo3-ffi/src/cpython/initconfig.rs#L82) depends on the targetted version. If we went that route, we could use pyembed (of the dead pyoxidizer family) or pyo3 and link against a specific version

Instead, i've opted to link against libpython at runtime, using the venv bin* to get the libpath for the currently selected version (3.9/3.13 etc.), using the [venv activation script](https://virtualenv.pypa.io/en/legacy/userguide.html#using-virtualenv-without-bin-python) to setup up `sys.path`, `sys.prefix` etc, and only relying on the 3.x stable api and [The Very High Level Layer](https://docs.python.org/3/c-api/veryhigh.html#the-very-high-level-layer) to setup and start the interpreter. If it fails, it'll fallback to the usual launch method of spawning the venv bin

Wasn't able to find any examples of run-time linking python, but i did see a mention in the [python manual](https://github.com/python/cpython/blob/b3a38438d83245e52dff80f348c7fde539333cea/Doc/faq/windows.rst#how-can-i-embed-python-into-a-windows-application) so we're not doing anything untoward here (hopefully)

Portability: this Just Works™ on my wsl ubuntu and win10 machines (after wrestling with the quirks of 3.9 and 3.13 on each platform), but i've not been able to test on macs

Overhead: once on first-startup to get and cache libpython's path, and none as far as i can tell on runtime

*this may seem weird but there's precedence: pyo3 [invokes](https://github.com/PyO3/pyo3/blob/67da00d07da5a5755344d101663e69e9209554bb/pyo3-build-config/src/impl_.rs#L210) python in build scripts

### DIFF
--- a/qt/launcher/src/platform/windows.rs
+++ b/qt/launcher/src/platform/windows.rs
@@ -286,7 +286,7 @@ impl Drop for PyFfi {
 }
 
 macro_rules! load_sym {
-    ($lib:expr, $name:literal) => {
+    ($lib:expr, $name:expr) => {
         std::mem::transmute(
             GetProcAddress($lib, PCSTR::from_raw($name.as_ptr().cast()))
                 .ok_or_else(|| anyhow!("failed to load {}", $name.to_string_lossy()))?,
@@ -294,9 +294,16 @@ macro_rules! load_sym {
     };
 }
 
+macro_rules! ffi {
+    ($lib:expr, $exec:expr, $($field:ident),* $(,)?) => {
+        #[allow(clippy::missing_transmute_annotations)] // they're not missing
+        PyFfi { exec: $exec, $($field: load_sym!($lib, ::std::ffi::CString::new(stringify!($field)).unwrap()),)* lib: $lib.0, }
+    };
+}
+
 impl PyFfi {
     #[allow(non_snake_case)]
-    pub fn load(path: impl AsRef<std::path::Path>) -> Result<Self> {
+    pub fn load(path: impl AsRef<std::path::Path>, exec: CString) -> Result<Self> {
         unsafe {
             let wide_filename: Vec<u16> = path
                 .as_ref()
@@ -311,13 +318,17 @@ impl PyFfi {
                 LOAD_LIBRARY_FLAGS::default(),
             )?;
 
-            #[allow(clippy::missing_transmute_annotations)] // they're not missing
-            Ok(PyFfi {
-                Py_InitializeEx: load_sym!(lib, c"Py_InitializeEx"),
-                Py_IsInitialized: load_sym!(lib, c"Py_IsInitialized"),
-                PyRun_SimpleString: load_sym!(lib, c"PyRun_SimpleString"),
-                Py_FinalizeEx: load_sym!(lib, c"Py_FinalizeEx"),
-                lib: lib.0,
+            Ok(ffi! {
+                lib,
+                exec,
+                Py_IsInitialized,
+                PyRun_SimpleString,
+                Py_FinalizeEx,
+                PyConfig_InitPythonConfig,
+                PyConfig_SetBytesString,
+                Py_InitializeFromConfig,
+                PyConfig_SetBytesArgv,
+                PyStatus_Exception,
             })
         }
     }
--- a/qt/launcher/src/platform/windows.rs
+++ b/qt/launcher/src/platform/windows.rs
@@ -6,7 +6,6 @@ use std::io::stdin;
 use std::os::windows::ffi::OsStrExt;
 use std::process::Command;
 
-use anki_io::ToUtf8Path;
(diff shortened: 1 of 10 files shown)


qt/launcher/src/platform/windows.rs
```diff
@@ -261,3 +275,107 @@ pub fn prepare_to_launch_normally() {
...
+        .to_str()
+        .ok_or_else(|| anyhow!("failed to get venv activation script path"))?
+        .replace('\\', "\\\\");
+
+    // NOTE: without windows_subsystem=console or pythonw,
+    // we need to reconnect stdin/stdout/stderr within the interp
+    // reconnect_stdio_to_console doesn't make a difference here
+    let console_snippet = if console {
+        r#" sys.stdout = sys.stderr = open("CONOUT$", "w"); sys.stdin = open("CONIN$", "r");"#
+    } else {
```
iamllama: This is very jank, but apart from using the console subsystem, i've not been able to find another solution for this. It also potentially doesn't show output between starting the interpreter and running this snippet

Perhaps anki-console could be changed from a wrapper that just spawns anki.exe with env vars to a stripped-down version of the launcher but with the console subsystem

iamllama:
I've just come across this nugget on https://docs.python.org/3/library/site.html
> Changed in version 3.14: [site](https://docs.python.org/3/library/site.html#module-site) is no longer responsible for updating [sys.prefix](https://docs.python.org/3/library/sys.html#sys.prefix) and [sys.exec_prefix](https://docs.python.org/3/library/sys.html#sys.exec_prefix) on [Virtual Environments](https://docs.python.org/3/library/sys_path_init.html#sys-path-init-virtual-environments). This is now done during the [path initialization](https://docs.python.org/3/library/sys_path_init.html#sys-path-init). As a result, under [Virtual Environments](https://docs.python.org/3/library/sys_path_init.html#sys-path-init-virtual-environments), [sys.prefix](https://docs.python.org/3/library/sys.html#sys.prefix) and [sys.exec_prefix](https://docs.python.org/3/library/sys.html#sys.exec_prefix) no longer depend on the [site](https://docs.python.org/3/library/site.html#module-site) initialization, and are therefore unaffected by [-S](https://docs.python.org/3/using/cmdline.html#cmdoption-S).

Which means that the venv activation script (`activate_this.py`) will no longer work from 3.14 onwards. Made a venv with cpython-3.14.0b2-linux-x86_64-gnu to check, and it turns out the script isn't even available anymore, which is disappointing to say the least.

~~Will look into using pyo3/PyConfig instead~~

EDIT: I've used bindgen on the 3.9/3.13 lin/win cpython headers to codegen `PyConfig`. This way we can avoid using pyo3 which, as far as i can tell, needs a concrete python version to configure itself and link against. I'd hoped to avoid version-scoping, but with this we'll still have installers that support <=3.X instead of one fixed version per. In any case it falls back to the usual launch so there's no difference

NB: The pythonistas have realised this is very inconvenient ([PEP 741](https://peps.python.org/pep-0741/)), and so from 3.14 onwards we can set `sys.argv` and `sys.executable` without knowing any layouts. ~~But this still isn't headache-free as we'd need to manipulate `PyObject`s. And having to support 3.9/3.13 won't allow us to actually use it~~

dae:
This is some impressive technical work Llama, and ties in to the decision we need to make as mentioned on https://github.com/ankitects/anki/pull/4390#issuecomment-3883021358. I might create a new issue to discuss, and respond there.

PR #4416, opened 2025-11-03
Sync tooltip focus fix

### LINKED ISSUE #4188: [BUG] “Collection Sync Complete” Popup Stays On Top of Other Applications
> Description:
> When performing a sync in recent versions of Anki, a popup message appears stating “Collection Sync complete.” However, if I switch to another application while this message is displayed, the popup remains visible and floats above other application windows, rather than staying within the Anki window or dismissing itself. This behavior is distracting and not consistent with standard UI practices.

Originally reported on https://forums.ankiweb.net/t/bug-collection-sync-complete-popup-stays-on-top-of-other-applications/63680

We may end up addressing this as we get more of the UI in Svelte, as we could overlay a pop-up over the page instead of in a separate window.

JMannervik:
Prevents the "Collection Sync complete" tooltip from appearing 
     when the Anki window is minimized or when switching to another 
     application. The tooltip now automatically closes if the window 
     loses focus, becomes invisible, or is minimized.
     
     Closes #4188

### DIFF
--- a/qt/aqt/utils.py
+++ b/qt/aqt/utils.py
@@ -1109,10 +1109,13 @@ def tooltip(
     def close_if_parent_inactive() -> None:
         if not _tooltipLabel or window is None:
             return
+        # Check if window is still active and application has focus
+        app_has_focus = aqt.mw.app.focusWindow() is not None
+        window_is_active = window.isActiveWindow()
         if (
             not window.isVisible()
             or (window.windowState() & Qt.WindowState.WindowMinimized)
-            or (was_active and not window.isActiveWindow())
+            or (was_active and (not window_is_active or not app_has_focus))
         ):
             closeTooltip()
 


iamllama:
It'd be better to move this into `tooltip()` as Abdo originally said, to address the issue for all tooltips

JMannervik:
Thanks for the feedback! I have centralized it now

abdnh:
For reference: https://github.com/ankitects/anki/issues/4188

If I run this in the debug console, the tooltip immediately disappears due to the active window check:

```python
from aqt.utils import tooltip

tooltip("hello", parent=mw, period=5000)
```

I assume this can potentially happen in some places in the code or in add-ons, so we need to handle it somehow.

JMannervik:
I think i found a solution for this @abdnh, I updated the focus check to only start monitoring if the window was active when the tooltip was created, so tooltips from the debug console or other tooltips (when Anki isn't active) still appear

abdnh:
There is a case where the tooltip persists despite Anki not being on top. On Windows for example, if I switch to another app without minimizing Anki, the tooltip will stay. I don't know if this is easy to fix.

JMannervik:
Hmm, yes it seems hard to get a fix for all tooltips...

JMannervik:
I dont have windows so i cant really check, but when i looked online it seems like app.focusWindow() is more reliable on windows. I can try this as a last attemt perhaps :)

abdnh:
Doesn't seem to make a difference.

> Hmm, yes it seems hard to get a fix for all tooltips...

Yeah, maybe it's not worth spending more time on it, as the main window will be moved to Svelte in the future.

dae:
Closing as this seems to have stalled.

PR #4426, opened 2025-11-09
Fix FindDuplicates dialog size exceeding screen bounds

### LINKED ISSUE #4181: `find duplicate` window way too large after large query
> I have a very long search query in the search history, and it seems the window resize depending on how long the longest query is.
> The find duplicate window is twice as large as my screen, it can be resized vertically but not horizontally, and there’s no way to clear the search history.
> I searched the code to find a solution, it seems the search history is in the profile prefs21.db table, but I can’t easily read and modify the values there. It the second time I encounter this problem, I was able to fix it in the past but I think the db structure was different.

Originally reported on https://forums.ankiweb.net/t/find-duplicate-window-way-too-large/63513

Nechaiter:
Description
Fixes an issue where the Find Duplicates dialog could open larger than the available screen size and prevent manual horizontal resizing.

After a large query, the dialog could be set to a minimum capped size based on the query text's length. This often made the dialog wider than the screen, preventing it from being resized manually or correctly constrained by the ensureWidgetInScreenBoundaries function.

Changes
-Modified qt/aqt/forms/finddupes.ui to set an appropriate size adjustment policy on the search QComboBox
-Changed the resizing logic/conditions within the ensureWidgetInScreenBoundaries function.

Before:

https://github.com/user-attachments/assets/5567d4a9-45bf-42b7-b1cd-26d6166aab31

After

https://github.com/user-attachments/assets/db1f1edc-fac0-4eb6-a82e-addcfebbb82b

Fixes #4181

### DIFF
--- a/rslib/src/scheduler/answering/mod.rs
+++ b/rslib/src/scheduler/answering/mod.rs
@@ -443,9 +443,20 @@ impl Collection {
             .storage
             .get_deck(card.deck_id)?
             .or_not_found(card.deck_id)?;
-        let config = self.home_deck_config(deck.config_id(), card.original_deck_id)?;
+        let home_deck = if card.original_deck_id.0 == 0 {
+            &deck
+        } else {
+            &self
+                .storage
+                .get_deck(card.original_deck_id)?
+                .or_not_found(card.original_deck_id)?
+        };
+        let config = self
+            .storage
+            .get_deck_config(home_deck.config_id().or_invalid("home deck is filtered")?)?
+            .unwrap_or_default();
 
-        let desired_retention = deck.effective_desired_retention(&config);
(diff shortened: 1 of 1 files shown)


qt/aqt/utils.py
```diff
@@ -809,7 +809,7 @@ def ensureWidgetInScreenBoundaries(widget: QWidget) -> None:
     wsize = widget.size()
     cappedWidth = min(geom.width(), wsize.width())
     cappedHeight = min(geom.height(), wsize.height())
-    if cappedWidth > wsize.width() or cappedHeight > wsize.height():
+    if cappedWidth < wsize.width() or cappedHeight < wsize.height():
```
abdnh, line 812: Good catch!

PR #4430, opened 2025-11-15
Fixed overlapping ranges when all history is selected

### LINKED ISSUE #2961: 366 of 365 days studied

### LINKED ISSUE #3932: Overlapping ranges in Reviews graph when All history is selected
When All history is selected, the Reviews graph shows bars with overlapping ranges.

The first bar from right shows "0-19 days"

The second bar from right shows "1-20 days ago"

JMannervik:
Fixes #3932
 Fixes #2961 

Fix Calendar diagram: prevents dates from other years from appearing when "All history" is selected

Reviews graph: corrects "days ago" range labels (e.g., "0-19 days" and "1-20 days ago" no longer overlap)


user1823:
I'm on `0abac06c` and overlapping ranges are still there:
image
image

user1823:
> I do not have that much history so it was a bit hard to test, would love some help testing and getting feedback if you have time

I guess that you can use the collection shared [here](https://github.com/ankitects/anki/pull/4349#issuecomment-3311014628) for testing purposes.

Pinging @Expertium to ask his permission for the same.

JMannervik:
Thanks for the feedback and the help, i am starting to work on a new solution that I can test more before updating this PR:)

JMannervik:
Thanks for the feedback and the help, i am starting to work on a new solution that i will be able to test

JMannervik:
Thanks! I have been able to test the new fix now as well, if you could try it aswell it would be very helpful @user1823. And thank you @Luc-Mcgrady for the great tips!

user1823:
I am testing `14b88a6f2`.

The ranges don't overlap now. But, I noticed some issues:
- The running total of the newest bin doesn't match the total below the graph, which means that the graph misses some reviews.
image

- The oldest bin is very thin. This is mathematically correct because it represents a smaller number of days, but is not a good UX, IMO.
image
image

- The changes in time.ts and calendar.ts are probably not needed.

JMannervik:
good point, i can remove the changes made in time and calendar, any suggestions for the ux for oldest bin or should i let that be?

user1823:
> any suggestions for the ux for oldest bin

I think that you can make it as wide as the rest of the bins.

JMannervik:
I can't seem to recreate the error where the running total does not match the overall total. For me it matches for this version of the PR
 
image
image

Luc-Mcgrady:
> The running total of the newest bin doesn't match the total below the graph, which means that the graph misses some reviews.

> I can't seem to recreate the error where the running total does not match the overall total. For me it matches for this version of the PR

1-20 days ago doesn't include today. Maybe this is the reason.

JMannervik:
> > The running total of the newest bin doesn't match the total below the graph, which means that the graph misses some reviews.
> 
> > I can't seem to recreate the error where the running total does not match the overall total. For me it matches for this version of the PR
> 
> 1-20 days ago doesn't include today. Maybe this is the reason.

good point, I think it is logic to include today either case, i will make that adjustment!

user1823:
I am not sure which one is correct, but this shows a different number of reviews as compared to Anki 25.09.

This PR (`e3546ea49`):
image
image
image
image


Anki 25.09:
image
image
image
image

JMannervik:
> I am not sure which one is correct, but this shows a different number of reviews as compared to Anki 25.09.


I cant really recreate this error, even if I switch beween main and this PR I get the same running total for both of them. It seems confusing since I did not change where the data is collected from for the total.  In short, the sums are computed from the same source in both branches, so the difference probably means either the input data differs or one of the environments still has stale build artifacts.

user1823:
Sorry for that. The difference was because of incorrect timezone on my WSL system.

This looks very good but there is a small issue:
If I select the 1 year option, the graph shows reviews from "0 days ago" to "365 days ago", i.e., it includes reviews of 366 days instead of the expected 365 days.

The correct bins should be "0 to 4 days ago", "5 to 9 days ago", "10 to 14 days ago" and so on upto "360 to 364 days ago".

JMannervik:
Okay! No problem at all, great that we could figure it out. I have made the necessary changes for the bins to be correctly calculated and for the checks to pass now! Thanks for all the help @user1823!

user1823:
Functionally, it looks perfect to me now. Thank you!
You can add Fixes #3932 and Fixes https://github.com/ankitects/anki/issues/2961 to your original post so that GitHub will auto-close these issues when this PR is merged.

The changes in config/bool.rs and revlog/mod.rs are unrelated and should be reverted.

JMannervik:
No problem at all, I have made the revertions and added the info to the original post aswell. Thanks again for the help!

ts/routes/graphs/reviews.ts
```diff
@@ -91,37 +91,61 @@ export function renderReviews(
...
             xMin = -89;
+            xMax = 0;
             break;
         case GraphRange.Year:
             xMin = -364;
+            xMax = 0;
             break;
         case GraphRange.AllTime:
             xMin = min(sourceData.reviewCount.keys())!;
             break;
```
Luc-Mcgrady: It would probably be a little neater if you inverted this to add an "xMax = 1" to the all time case

ts/routes/graphs/reviews.ts
```diff
@@ -91,37 +91,61 @@ export function renderReviews(
...
-    if (range === GraphRange.AllTime) {
-        x.nice(desiredBars);
+    let x = scaleLinear().domain([xMin!, xMax]);
+    let thresholds = x.ticks(desiredBars);
+    if (thresholds.length >= 2) {
+        const spacing = thresholds[1] - thresholds[0];
+        const partial = thresholds[0] - x.domain()[0];
+        if (spacing > 0 && partial > 0 && partial < spacing) {
+            const adjustedMin = thresholds[0] - spacing;
+            x = scaleLinear().domain([adjustedMin, xMax]);
```
Luc-Mcgrady: Are you adjusting the xMin twice, once here and once above? (Actual question, I can't tell). Maybe there's some way to improve this.

Luc-Mcgrady (changes requested):
For the not all time graphs today is cut off.

image

Main branch:
image

ts/routes/graphs/reviews.ts
JMannervik: Thing is that I tried to do that but I got the error that I couldnt assign 1 to a type 0, therefore i solved it like this. An alternative would be to "let xMax = number;" and then assign it different values for each case aswell

ts/routes/graphs/reviews.ts
JMannervik: Thats okay, right now im not adjusting xMin twice. Im creating a new scale with an adjusted minimum after computing ticks.

I looked into a way to simplify it, but I cant really figure out how to make it shorter, but i have a solution, with some comments included so that I set xMin in the switch statement based on the range, then I create an initial scale to compute tick spacing. Then after that I adjust xMin once to align with tick spacing. I can push that up shortly, as soon as ive tested it

JMannervik:
I looked over the comments you left @Luc-Mcgrady! I resolved the first one, and for the second one i refactored it and aded some comments so that it might be easier to understand what it is that is actually happening, I hope that is enough :))

I dont know why the buildkite checks fail now though, should I do anything about it? The files it is complaining about are unrelated files that i have not touched in the latest commits.

ts/routes/graphs/reviews.ts
```diff
@@ -90,37 +90,37 @@ export function renderReviews(
 ): TableDatum[] {
     const svg = select(svgElem);
     const trans = svg.transition().duration(600) as any;
-
-    let xMax = 1;
+
```
user1823: Remove this extra whitespace and the checks will pass.

JMannervik:
> For the not all time graphs today is cut off.

This is fixed now aswell, It was that the xMax values for 1 month and 3 months needed to be 1 and not 0!

user1823:
Why does the year require xMax to be 0 while the rest require it to be 1?

JMannervik:
> Why does the year require xMax to be 0 while the rest require it to be 1?

Good question, I dont think that is necessary now actually I had not concidered it!

 I changed it when I tried to fix the range of the year.  (i.e. graph shows reviews from "0 days ago" to "365 days ago", i.e., it includes reviews of 366 days instead of the expected 365 days.)  and then i made changes so that the range 0-4 days, 5-9 days and so on worked. Together it worked, but now when i test xmax = 1 for the year it makes no difference.

user1823:
I haven't yet tried to build it again. But, I reviewed the code and some parts look over-engineered. It is always the best practice to keep the code as simple as possible.

For example, this code looks functionally equivalent to what is proposed in this PR, but is much easier to read.
```ts
export function dayLabel(daysStart: number, daysEnd: number): string {
    // Convert bin boundaries [start, end) to display range
    // For past days: bin [-10, -5) should display "5-9 days ago"
    // For future days: bin [5, 10) should display "In 5-9 days"
    
    const larger = Math.max(Math.abs(daysStart), Math.abs(daysEnd));
    const smaller = Math.min(Math.abs(daysStart), Math.abs(daysEnd));
    
    if (larger - smaller <= 1) {
        // singular
        if (daysStart >= 0) {
            return tr.statisticsInDaysSingle({ days: daysStart });
        } else {
            return tr.statisticsDaysAgoSingle({ days: -daysStart });
        }
    } else {
        // range
        if (daysStart >= 0) {
            // Future: [5, 10) -> "In 5-9 days"
            return tr.statisticsInDaysRange({
                daysStart,
                daysEnd: daysEnd - 1,
            });
        } else if (daysEnd <= 0) {
            // Past: [-10, -5) -> "5-9 days ago"
            return tr.statisticsDaysAgoRange({
                daysStart: Math.abs(daysEnd),
                daysEnd: Math.abs(daysStart) - 1,
            });
        } else {
            // Crosses zero: [-5, 1) -> "0-4 days ago"
            return tr.statisticsDaysAgoRange({
                daysStart: 0,
                daysEnd: Math.abs(daysStart) - 1,
            });
        }
    }
}
```

But more importantly, do we really need to modify the dayLabel function? Why not fix the values before inputting them into dayLabel?

JMannervik:
> I haven't yet tried to build it again. But, I reviewed the code and some parts look over-engineered. It is always the best practice to keep the code as simple as possible.
> 
> But more importantly, do we really need to modify the dayLabel function? Why not fix the values before inputting them into dayLabel?

I see, yes ofcourse I can spend some time trying to simple it down, and minimize the code as much as possible. I will try this solution of modfying it in review ts aswell, and see if i can make that work!

JMannervik:
I was able to simplyfy it quite a bit thanks to you suuggestion @user1823! I dont know how I could simplify it more, but i am open to suggestions! Thanks again for all the help and feedback :)))

ts/lib/tslib/time.ts
```diff
@@ -177,10 +177,17 @@ export function dayLabel(daysStart: number, daysEnd: number): string {
                 daysStart,
                 daysEnd: daysEnd - 1,
             });
+        } else if (daysEnd <= 0) {
+            // Past: [-10, -5) -> "5-9 days ago"
```
user1823: If I recall my high-school math right, [-10, -5) means including the day -10 and not including the day -5. So, isn't 5-9 days ago incorrect for this? Shouldn't it be 6 to 10 days ago?

In my code example above, I just reorganized your code and didn't really pay attention to the brackets. But now that I look at them, the original logic in dayLabel, where it subtracted 1 from daysEnd rather than daysStart, was actually correct. So, this change is introducing another bug while trying to fix one. The original dayLabel function logic is correct, and the real issue must be elsewhere.

JMannervik: Alright, I will spend some time today and tomorrow trying to figure this out completely

JMannervik:
Okay, now I have managed to fix both of the issues #3932 and #2961 while only making changes in review. The solution is simpler than before and I dont touch the actual daylabel function now. 

I think this is enough for the issues, there is no overlap for any of the ranges and the data is correct in the ranges aswell! 

As usual, all and any feedback are appreciated!

ts/routes/graphs/reviews.ts
```diff
@@ -109,19 +108,46 @@ export function renderReviews(
...
+    if (thresholds.length >= 2) {
+        const spacing = thresholds[1] - thresholds[0];
+        const partial = thresholds[0] - xMin!;
+        if (spacing > 0 && partial > 0 && partial < spacing) {
+            xMin = Math.max(thresholds[0] - spacing, shouldCapRange ? originalXMin! : -Infinity);
+            x = scaleLinear().domain([xMin, xMax]);
+            thresholds = x.ticks(desiredBars);
+        }
+    }
+    // For Year, shift thresholds forward by one day to make first bin 0-4 instead of 0-5
```
user1823: Why does the "Year" case need special handling?

ts/routes/graphs/reviews.ts
```diff
@@ -109,19 +108,46 @@ export function renderReviews(
...
-        .value((m) => {
-            return m[0];
-        })
+    let bins = bin()
+        .value((m) => m[0])
         .domain(x.domain() as any)
-        .thresholds(x.ticks(desiredBars))(sourceMap.entries() as any);
+        .thresholds(thresholds)(sourceMap.entries() as any);
+
+    if (bins.length > 1 && thresholds.length > 1) {
```
user1823: What does this code do? Adding a small comment will help.

ts/routes/graphs/reviews.ts
JMannervik: Because if it not handeled specially the last bin becomes 0-5 and then the first bin is 361-364, which is just 4 days, so the bins are not the same sizes. This was the most simple way for me to solve this that i could find.

JMannervik:
I added some comments and could remove some of the logic that adressed issues that were not related to the code but the enviroment! If anyone else could test this aswell to make sure it works well for everyone that would be greatly appreciated!

ts/routes/graphs/reviews.ts
```diff
@@ -109,19 +108,36 @@ export function renderReviews(
             break;
     }
     const desiredBars = Math.min(70, Math.abs(xMin!));
-
-    const x = scaleLinear().domain([xMin!, xMax]);
-    if (range === GraphRange.AllTime) {
-        x.nice(desiredBars);
+    const shouldCapRange = range !== GraphRange.AllTime;
+    const originalXMin = shouldCapRange ? xMin! : undefined;
```
user1823:
```suggestion
    const originalXMin = xMin!;
```

The previous code was redundant because there's already a check for shouldCapRange in line 125.

ts/routes/graphs/reviews.ts
```diff
@@ -109,19 +108,36 @@ export function renderReviews(
...
+        const partial = thresholds[0] - xMin!;
+        if (spacing > 0 && partial > 0 && partial < spacing) {
+            // Extend xMin backward to align with tick spacing, but cap at originalXMin for fixed ranges
+            xMin = Math.max(thresholds[0] - spacing, shouldCapRange ? originalXMin! : -Infinity);
+            x = scaleLinear().domain([xMin, xMax]);
+            thresholds = x.ticks(desiredBars);
+        }
+    }
+    // For Year, shift thresholds forward by one day to make first bin 0-4 instead of 0-5
+    if (range === GraphRange.Year) {
```
user1823:
```suggestion
    if (range === GraphRange.Year || range === GraphRange.AllTime) {
```

This fixes the following issue:
In "all", the latest bin says "0-20 days" ago (i.e., it includes 21 days) instead of the expected "0-19 days" (because all the other bins include 20 days).

ts/routes/graphs/reviews.ts
```diff
@@ -109,19 +108,36 @@ export function renderReviews(
...
+        thresholds = [...new Set(thresholds.map(t => Math.min(t + 1, 1)))].sort((a, b) => a - b);
```
Luc-Mcgrady, line 130: I'm not sure using a spread on a Set is needed here. I don't think thresholds will ever have duplicate elements.

ts/routes/graphs/reviews.ts
Luc-Mcgrady: > // For Year, shift thresholds forward by one day to make first bin 0-4 instead of 0-5

Are you removing the "today" bar for this year on purpose?

JMannervik:
@[Luc-Mcgrady](https://github.com/Luc-Mcgrady)
Luc-Mcgrady
[3 hours ago](https://github.com/ankitects/anki/pull/4430#discussion_r2611448036)
// For Year, shift thresholds forward by one day to make first bin 0-4 instead of 0-5

Are you removing the "today" bar for this year on purpose?

No i dont think so, Day 0 (today) is still included: the last bin will be [last_negative, 1) which includes day 0

Luc-Mcgrady:
I mean that for the all bar, unlike the year bar, today has its own bar:
image
On second look it is weird to me that it has 0-5 days, with the today bar also there. I assume this is a quirk of the labeling? (You can see the pixel wide bin for today on the right)
image
(The year bar does not have the today bar as shown here:)
image

CONTRIBUTORS
```diff
@@ -255,6 +255,7 @@ Ranjit Odedra <ranjitodedra.dev@gmail.com>
 Eltaurus <https://github.com/Eltaurus-Lt>
 jariji
 Francisco Esteva <fr.esteva@duocuc.cl>
+Junia Mannervik <junia.mannervik@gmail.com>
```
Luc-Mcgrady, line 258: The CLI is flagging this as not your email. Perhaps there is something wrong with your git history:
image

image

JMannervik:
Regarding the today bar @Luc-Mcgrady I think the problem is fixed now, when i made the changes that were suggested, could you check if you can still se the bar with these changes?

JMannervik:
[user1823](https://github.com/user1823)
left a comment
Left two suggestions.

There's one more problem:

The graph shows up to "1839 days ago" (i.e., a total of 1840 days) but the days studied part says "of 1841 days". One of them has to be wrong.


Regarding this, when i try to recreate it the oldest bin includes the same day span as all other bins (20 or 50 in my case depending on if i select all history or not) For both of them my total amount of days is included in the span. for example: my total of is 3613 days, had the oldest bin show 3600-3649, which is because the bin should be the same sizes. 

Is this a neccessary fix still to make sure the oldest bin instead (in my case) displays 3600-3612 days? 

And is do you still have the same issue now? is it still one day to "little" included in the graph"?

user1823:
Functionally, this looks good to me,

> Is this a necessary fix still to make sure the oldest bin instead (in my case) displays 3600-3612 days?

No. In my example, the problem was that the days studied part showed more days than the graph. You have already fixed this issue in the latest commits.

> is it still one day to "little" included in the graph"?

It's not included now (as expected).

I suggest the following patch to remove some redundancies and minimize the diff relative to the main.

```diff
diff --git a/ts/routes/graphs/reviews.ts b/ts/routes/graphs/reviews.ts
index d66496687..3f3975f64 100644
--- a/ts/routes/graphs/reviews.ts
+++ b/ts/routes/graphs/reviews.ts
@@ -90,6 +93,7 @@ export function renderReviews(
 ): TableDatum[] {
     const svg = select(svgElem);
     const trans = svg.transition().duration(600) as any;
+
     const xMax = 1;
     let xMin = 0;
     // cap max to selected range
@@ -108,36 +112,34 @@ export function renderReviews(
             break;
     }
     const desiredBars = Math.min(70, Math.abs(xMin!));
-    const shouldCapRange = range !== GraphRange.AllTime;
+    const unboundRange = range == GraphRange.AllTime;
     const originalXMin = xMin!;

     // Create initial scale to determine tick spacing
     let x = scaleLinear().domain([xMin!, xMax]);
-    // Generate thresholds: these define where bins are split (e.g., [-5, -10, -15, ...])
-    // D3's bin() will create bins between consecutive thresholds: [threshold[i], threshold[i+1])
     let thresholds = x.ticks(desiredBars);
-    // Adjust xMin to align with tick spacing so the oldest bin has the same width as others
-    if (thresholds.length >= 2) {
+    // For unbound ranges, extend xMin backward so that the oldest bin has the same width as others
+    if (unboundRange && thresholds.length >= 2) {
         const spacing = thresholds[1] - thresholds[0];
         const partial = thresholds[0] - xMin!;
         if (spacing > 0 && partial > 0 && partial < spacing) {
-            // Extend xMin backward to align with tick spacing, but cap at originalXMin for fixed ranges
-            xMin = Math.max(thresholds[0] - spacing, shouldCapRange ? originalXMin! : -Infinity);
+            xMin = thresholds[0] - spacing;
             x = scaleLinear().domain([xMin, xMax]);
             thresholds = x.ticks(desiredBars);
         }
     }
-    // For Year, shift thresholds forward by one day to make first bin 0-4 instead of 0-5
+    // For Year & All Time, shift thresholds forward by one day to make first bin 0-4 instead of 0-5
     if (range === GraphRange.Year || range === GraphRange.AllTime) {
        thresholds = [...new Set(thresholds.map(t => Math.min(t + 1, 1)))].sort((a, b) => a - b);
     }

     const sourceMap = showTime ? sourceData.reviewTime : sourceData.reviewCount;
-    // Create bins using the thresholds: each bin contains data points between consecutive thresholds
     const bins = bin()
-        .value((m) => m[0]) // Extract the day number from each map entry [day, data]
-        .domain(x.domain() as any) // Bins are constrained to this domain [xMin, xMax]
-        .thresholds(thresholds)(sourceMap.entries() as any); // Use computed thresholds to split data into bins
+        .value((m) => m[0])
+        .domain(x.domain() as any)
+        .thresholds(thresholds)(sourceMap.entries() as any);

     // empty graph?
     const totalDays = sum(bins, (bin) => bin.length);
@@ -232,9 +240,7 @@ export function renderReviews(
         // If bin ends at 0, treat it as crossing zero so day 0 is included
         // For the first (oldest) bin, use the original xMin to ensure labels match the intended range
         const isFirstBin = bins.length > 0 && d.x0 === bins[0].x0;
-        const startDay = isFirstBin
-            ? originalXMin
-            : Math.floor(d.x0!);
+        const startDay = isFirstBin ? originalXMin : Math.floor(d.x0!);
         const endDay = d.x1! === 0 ? 1 : d.x1!;
         const day = dayLabel(startDay, endDay);
         const totals = totalsForBin(d);
@@ -351,9 +360,7 @@ export function renderReviews(
         })
         .on("mouseout", hideTooltip);

-    // Calculate periodDays from the actual data range, not the adjusted xMin
-    // For AllTime, xMin might be extended for bin alignment, so use the original xMin (actual oldest data)
-    // For fixed ranges, use the original xMin before adjustment (which matches the displayed range)
+    // The xMin might be extended for bin alignment, so use the original xMin
     const periodDays = -originalXMin + 1;
     const studiedDays = sum(bins, (bin) => bin.length);
     const studiedPercent = (studiedDays / periodDays) * 100;
```

JMannervik:
@user1823 Now I have implemented the final suggested changes you made aswell! Thanks for the help :)

user1823:
It looks like you have applied my patch manually. You could have simply copied it into a new patch file (say user1823.patch) and then used `git apply user1823.patch` to apply the suggested changes.

Trying to apply the patch manually caused two minor issues:
- We are now creating a new constant (`actualXMin`), which is not really necessary. We can directly use `originalXmin` in `periodDays`.
- There is an unnecessary whitespace change in line 93 (just before `const xMax = 1;`). You can fix it by adding an empty line (not having any indentation) just before `const xMax = 1;`

JMannervik:
yes i did haha, i will try to fix it now :)

JMannervik:
> I have verified that these changes fix both of the linked issues and the diff looks clean to me. Thank you for the contribution.

Thank you for all the help! I feel like I have learned alot thanks to you :)

user1823:
This approach seems to work but it looked a bit hacky (e.g., why do we need to shift the thresholds?). After studying the problem a bit, I realized that the actual cause of the issue is this:

D3 creates bins in the ascending order of thresholds (e.g., [-20, -15), [-15, -10), [-10, -5), , [-5, 0)). When working with negative day values (past dates), this causes two fundamental issues:
- Semantic mismatch: For past dates, we want bins like "0-4 days ago", "5-9 days ago", but D3's bin [-5, 0) contains days -5, -4, -3, -2, -1, which semantically represents "1-5 days ago", not "0-4 days ago".
- Day 0 (today) gets isolated: D3's automatic threshold generation places day 0 in a separate bin [0, 1), separating it from the other past day bins.

Theoretically, this issue can be easily solved by transforming negative day values to their absolute values before binning. This allows D3 to naturally create bins like [0, 5), [5, 10), [10, 15) which directly map to "0-4 days ago", "5-9 days ago", "10-14 days ago". So, we only need to negate the boundaries when displaying the labels. 

I tried to implement this. It seems to work, but the graph renders in the opposite direction. 

image


Here's a patch (relative to `main`) of my attempt. If this approach of transforming the data looks better than shifting the thresholds, someone can find the error in this and fix it.

```patch
diff --git a/ts/lib/tslib/time.ts b/ts/lib/tslib/time.ts
index 25d70eef3..9f6bbb5c6 100644
--- a/ts/lib/tslib/time.ts
+++ b/ts/lib/tslib/time.ts
@@ -161,8 +161,13 @@ export function timeSpan(
 }

 export function dayLabel(daysStart: number, daysEnd: number): string {
+    // Convert bin boundaries [start, end) to display range
+    // For future days: bin [5, 10) should display "In 5-9 days"
+    // For past days: bin [-5, -10) should display "5-9 days ago"
+
     const larger = Math.max(Math.abs(daysStart), Math.abs(daysEnd));
     const smaller = Math.min(Math.abs(daysStart), Math.abs(daysEnd));
+
     if (larger - smaller <= 1) {
         // singular
         if (daysStart >= 0) {
@@ -172,15 +177,15 @@ export function dayLabel(daysStart: number, daysEnd: number): string {
         }
     } else {
         // range
-        if (daysStart >= 0) {
+        if (daysStart >= 0 && daysEnd > 0) {
             return tr.statisticsInDaysRange({
                 daysStart,
                 daysEnd: daysEnd - 1,
             });
         } else {
             return tr.statisticsDaysAgoRange({
-                daysStart: Math.abs(daysEnd - 1),
-                daysEnd: -daysStart,
+                daysStart: Math.abs(daysStart),
+                daysEnd: Math.abs(daysEnd) - 1,
             });
         }
     }

diff --git a/ts/routes/graphs/reviews.ts b/ts/routes/graphs/reviews.ts
index ed0e07f8c..59fc29396 100644
--- a/ts/routes/graphs/reviews.ts
+++ b/ts/routes/graphs/reviews.ts
@@ -91,37 +94,52 @@ export function renderReviews(
     const svg = select(svgElem);
     const trans = svg.transition().duration(600) as any;

-    const xMax = 1;
-    let xMin = 0;
+    const xMin = -1;
+    let xMax = 0;
     // cap max to selected range
     switch (range) {
         case GraphRange.Month:
-            xMin = -30;
+            xMax = 30;
             break;
         case GraphRange.ThreeMonths:
-            xMin = -89;
+            xMax = 89;
             break;
         case GraphRange.Year:
-            xMin = -364;
+            xMax = 364;
             break;
         case GraphRange.AllTime:
-            xMin = min(sourceData.reviewCount.keys())!;
+            xMax = Math.abs(min(sourceData.reviewCount.keys())!);
             break;
     }
-    const desiredBars = Math.min(70, Math.abs(xMin!));
+    const desiredBars = Math.min(70, xMax);
+    const unboundRange = range == GraphRange.AllTime;
+    const originalXMax = xMax;

-    const x = scaleLinear().domain([xMin!, xMax]);
-    if (range === GraphRange.AllTime) {
-        x.nice(desiredBars);
+    // Transform data to use absolute values (days ago)
+    const sourceMap = showTime ? sourceData.reviewTime : sourceData.reviewCount;
+    const transformedData = Array.from(sourceMap.entries()).map(
+        ([day, reviews]) => [Math.abs(day), reviews] as [number, Reviews],
+    );
+
+    let x = scaleLinear().domain([xMin, xMax]);
+    let thresholds = x.ticks(desiredBars);
+
+    // For unbound ranges, extend xMax forward so that the last bin has the same width as others
+    if (unboundRange && thresholds.length >= 2) {
+        const spacing = thresholds[1] - thresholds[0];
+        const lastThreshold = thresholds[thresholds.length - 1];
+        const partial = xMax - lastThreshold;
+        if (spacing > 0 && partial > 0 && partial < spacing) {
+            xMax = lastThreshold + spacing;
+            x = scaleLinear().domain([xMin, xMax]);
+            thresholds = x.ticks(desiredBars);
+        }
     }

-    const sourceMap = showTime ? sourceData.reviewTime : sourceData.reviewCount;
     const bins = bin()
-        .value((m) => {
-            return m[0];
-        })
+        .value((m) => m[0])
         .domain(x.domain() as any)
-        .thresholds(x.ticks(desiredBars))(sourceMap.entries() as any);
+        .thresholds(thresholds)(transformedData as any);

     // empty graph?
     const totalDays = sum(bins, (bin) => bin.length);
@@ -212,7 +236,10 @@ export function renderReviews(
     }

     function tooltipText(d: BinType, cumulative: number): string {
-        const day = dayLabel(d.x0!, d.x1!);
+        // Bins are [x0, x1) where x0 and x1 are absolute "days ago"
+        // Convert back to negative days for dayLabel
+        // e.g., bin [5, 10) means days -9 to -5, which displays as "5-9 days ago"
+        const day = dayLabel(-d.x0!, -d.x1!);
         const totals = totalsForBin(d);
         const dayTotal = valueLabel(sum(totals));
         let buf = `<table><tr><td>${day}</td><td align=end>${dayTotal}</td></tr>`;
@@ -327,7 +357,7 @@ export function renderReviews(
         })
         .on("mouseout", hideTooltip);

-    const periodDays = -xMin + 1;
+    const periodDays = originalXMax + 1;
     const studiedDays = sum(bins, (bin) => bin.length);
     const studiedPercent = (studiedDays / periodDays) * 100;
     const total = yCumMax;
```

Luc-Mcgrady:
> Regarding the today bar @Luc-Mcgrady I think the problem is fixed now, when i made the changes that were suggested, could you check if you can still se the bar with these changes?

Maybe it would be better to include the today bar on the all time graph instead of removing it for the 1 year graph? This is an actual question it's not really my call.

user1823:
Today being in its own bin was a bug. See https://github.com/ankitects/anki/issues/2961#issuecomment-2124393966

JMannervik:
> > Regarding the today bar @Luc-Mcgrady I think the problem is fixed now, when i made the changes that were suggested, could you check if you can still se the bar with these changes?
> 
> Maybe it would be better to include the today bar on the all time graph instead of removing it for the 1 year graph? This is an actual question it's not really my call.

I think that it makes more sense to in the year graph to just have a bar that includes today, (0-4), Since you can still see the today bar on one and three months. I think it looks more clean to not have a bar for today on either year or alltime. But that is just my opinion :))))

JMannervik:
Is there something else I can do here to get this PR accepted? I think the solution that this provides is better than the current state of the review graph. I belive it would be best to implement this for now to close the issues that this concerns.

Luc-Mcgrady (approved):
The maintainer is away right now. Please be patient.

dae:
Apologies for keeping you waiting so long. Thank you for your efforts to fix this issue - if Abdo's happy with it too, happy to see it merged.

PR #4432, opened 2025-11-17
Docs/Add more required packages to Linux build guide

user1823:
Updated package installation instructions for Debian/Ubuntu to include gcc-12 and libxkbfile1.

### DIFF
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -34,7 +34,13 @@ sudo apt install libxcb-icccm4 libxcb-image0 libxcb-keysyms1 \
   libxcb-randr0 libxcb-render-util0 libxkbfile1
 ```
 
-On some distros such as Arch Linux and Fedora, you may need to install the
+The libraries that might be required on Arch Linux:
+
+```
+sudo pacman -S nss libxkbfile
+```
+
+On some distros such as Fedora, you may need to install the
 `libxcrypt-compat` package if you get an error like this:
 
 ```


docs/linux.md
```diff
@@ -14,7 +14,7 @@ official releases.
 **Ensure some basic tools are installed**:
 
 ```
-$ sudo apt install bash grep findutils curl gcc g++ make git rsync
+$ sudo apt install bash grep findutils curl gcc gcc-12 g++ make git rsync
```
GithubAnon0000, line 17: Why do we need `gcc-12` in addition to `gcc` (my debian actually has gcc-10, 12 and 14 installed though I'm not sure if `gcc` pulled all of them)?

docs/linux.md
```diff
@@ -31,7 +31,7 @@ for example:
 
 ```
 sudo apt install libxcb-icccm4 libxcb-image0 libxcb-keysyms1 \
-  libxcb-randr0 libxcb-render-util0
+  libxcb-randr0 libxcb-render-util0 libxkbfile1
```
GithubAnon0000, line 34: Interessting. I wonder why `libxkbfile1` is needed (and if it's needed for X11 only or also for wayland). From the man pages:

> Description: X11 keyboard file manipulation library
> libxkbfile provides an interface to read and manipulate description files for XKB, the X11 keyboard configuration extension.

That being said, my debian also had this installed. Maybe I just never noticed its needed.

user1823:
I don't know why exactly we need them but I installed the packages mentioned in linux.md and then ran `./run` and the build failed because of these missing packages. So, I installed them and ran `./run` again and the build succeeded.

dae:
Thank you!

PR #4435, opened 2025-11-18
fix(Import): case-fold media filenames when checking uniqueness

### LINKED ISSUE #4433: Importer overwrites existing media files with different letter cases
> I’ve read posts about Anki’s case sensitivity so I’m reporting this bug here.
> Suppose you have two decks with a0.jpg and A0.JPG each. When you import the a0.jpg deck then the A0.JPG deck, a0.jpg will be replaced with A0.JPG without any notice. The actual image within the card will be displayed as A0.JPG too.
> If you add a third deck with a0.jpg, Anki will add hash to the filename from this point despite not having the actual a0.jpg inside the media folder.
> I had this issue after importing some decks with images, which contained files like 0.jpg and 0.JPG in different decks.
> Bonus sync bug: Add a forth deck with A0.JPG. Close the program to sync it, then open it again. Delete all four decks. Close the program and you’ll get a sync error briefly flashing before it crashes.
> I made the 4 decks experimenting with this problem, you can get it here:
>   
> 
>       drive.google.com
>   
> 
>   
>       
> 
> 
> 
> anki case variation cards.zip
> 
> Google Drive file.
> 
>   
> 
>   
>     
>     
>   
> 
>   
> 
> 
> Version info:
> Version ⁨25.09.2 (3890e12c)⁩
> Python 3.13.5 Qt 6.9.1 Chromium 122
> Windows 11 25H2 26200.7171

Originally reported on https://forums.ankiweb.net/t/media-files-case-sensitivity-ignored-on-the-first-collision/67460

iamllama:
Fixes #4433

The fix proposed is to compare media filenames case-insensitively on windows only. Initially set it to case-fold all the time as i thought case-sensitivity was a filesystem property, but it seems windows allows [per-directory case-sensitivity](https://devblogs.microsoft.com/commandline/per-directory-case-sensitivity-and-wsl/) (still problematic as win32 programs usually expect case insensitive filenames)

Perhaps we could test for this at startup by touching two files whose names only differ by case and using [same-file](https://github.com/BurntSushi/same-file) to check if they're considered the same in the media folder

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -254,6 +254,7 @@ nav1s <nav1s@proton.me>
 Ranjit Odedra <ranjitodedra.dev@gmail.com>
 Eltaurus <https://github.com/Eltaurus-Lt>
 jariji
+Francisco Esteva <fr.esteva@duocuc.cl>
 
 ********************
 
--- a/qt/aqt/forms/finddupes.ui
+++ b/qt/aqt/forms/finddupes.ui
@@ -47,6 +47,9 @@
        <property name="insertPolicy">
         <enum>QComboBox::NoInsert</enum>
        </property>
+       <property name="sizeAdjustPolicy">
+        <enum>QComboBox::SizeAdjustPolicy::AdjustToMinimumContentsLengthWithIcon</enum>
+       </property>
       </widget>
      </item>
     </layout>
--- a/qt/aqt/tts.py
+++ b/qt/aqt/tts.py
@@ -94,8 +94,15 @@ class TTSPlayer:
 
             rank -= 1
 
-        # if no preferred voices match, we fall back on language
-        # with a rank of -100
+        # if no requested voices match, use a preferred fallback voice
+        # (for example, Apple Samantha) with rank of -50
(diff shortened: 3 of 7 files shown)


yclee126:
MacOS APFS is also case-insensitive by default, wouldn't it be better to always case-fold? Syncing between different filesystems could be a problem too.

iamllama:
> Syncing between different filesystems could be a problem too.

Doesn't seem like it'd be easy to handle this retroactively (there's already some normalisation when syncing but it doesn't update media references afaict), will look into it. Maybe it can be done during `Check Database`

For now i've added a change to case-fold when adding new media for consistency

rslib/src/media/files.rs
```diff
@@ -173,7 +173,9 @@ pub fn add_data_to_folder_uniquely<'a, P>(
 where
     P: AsRef<Path>,
 {
-    let normalized_name = normalize_filename(desired_name);
+    // case-fold when to account for case-insensitive filesystems
```
abdnh: Nitpick: "case-fold" suggests something like unicase is used here.

PR #4442, opened 2025-11-20
fix(reviewer/card): Implement night mode styles in reviewer.scss

iusx:
Group 590

---

On macOS, dark mode isn’t supported in some places. I’ve put up with this issue for a long time, so I decided to try fixing it. (Maybe this is the reason—at least it works fine on my local setup and now supports dark mode.)

(I like Anki,

### DIFF
--- a/ts/reviewer/reviewer.scss
+++ b/ts/reviewer/reviewer.scss
@@ -23,7 +23,7 @@ body {
 
 // explicit nightMode definition required
 // to override default .card styling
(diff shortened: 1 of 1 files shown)


iusx:
It was a problem with my flashcard deck (not Anki). I switched to another deck and it supports dark mode now, Sorry

telegram-cloud-photo-size-2-5258412491987947356-y

PR #4444, opened 2025-11-24
Implement some Flexible Grading features as a Reviewer

### LINKED ISSUE #4443: Implement some Flexible Grading features as a Reviewer
https://github.com/ankitects/anki/pull/4289 breaks add-on compatibility, and I'm afraid it means Flexible Grading might stop working someday. I want to add an alternative code path that implements some Flexible Grading features without using WebViews.

Since https://github.com/ankitects/anki/pull/4289 implements a new reviewer by subclassing the default, I think it makes sense to implement some Flexible Grading features the same way: as an alternative reviewer. This would follow up on https://github.com/ankitects/anki/pull/2502 since the ability to configure answer keys was originally unique to Flexible Grading too.

https://github.com/ankitects/anki/pull/4289 goes the wrong direction: it is overly complicated, wastes screen space, and is slower than a native Qt UI. I'd like to be able to opt out permanently, as I suggested in [this comment](https://github.com/ankitects/anki/issues/4142#issuecomment-3059469064):

> I think it's better to tackle the issue from a different angle. Add preferences to "opt out of Webview screens". So, when you enable this option (or a subset of options), you would get a native UI instead.

tatsumoto-ren:
Implement Flexible bottom bar (which has Flexible Grading features).

Fixes: #4443

I also thought about completely removing `mw.bottomWeb` when Flexible Grading is enabled, so that it doesn't stay in memory and waste resources, but it requires a lot of refactoring and messing with other code, so I left it out for now.

## Screenshots:

- Enable flexible reviewer in settings:
  screenshot-2025-11-24-04-32-29
- Deck browser:
  screenshot-2025-11-24-07-55-50
- Overview:
  screenshot-2025-11-24-07-57-30
- Question side:
  screenshot-2025-11-24-07-57-57
- Answer side:
  screenshot-2025-11-24-07-58-01


drehak:
Why is it monospace?

user1823:
Users may want to enable some of these features but not all. So, it would probably be better to add more granular options in the preferences instead of forcing the user to choose one of the two reviewers. Also, the word "flexible" doesn't clearly describe what this option does.

Also, this is probably better done after the svelte reviewer PR is merged, to prevent merge conflicts.

tatsumoto-ren:
> Why is it monospace?

It fits the terminal/vim-like theme.

> Users may want to enable some of these features but not all. 

Which specific features would you like to toggle?

> Also, the word "flexible" doesn't clearly describe what this option does.

"Flexible grading" is the [add-on](https://ankiweb.net/shared/info/1715096333)'s name, but feel free to suggest a clearer name.

>  this is probably better done after the svelte reviewer PR is merged

They do not conflict, so it should not matter. The only likely merge conflict would be Preferences code. My plan is to let users choose between the Svelte reviewer, this reviewer, and the default reviewer. We do not want to use the Svelte reviewer and prefer native Qt widgets for menus, so the solution is to add an alternative code path.

tatsumoto-ren:
screenshot-2025-11-30-08-56-44

I think it's time for Anki to stop using mypy and replace it with beartype. This custom class is a subclass of QWidget, it's not an error.

user1823:
> "Flexible grading" is the [add-on](https://ankiweb.net/shared/info/1715096333)'s name, but feel free to suggest a clearer name.

> Which specific features would you like to toggle?

Can you list what features does this PR introduce? I was assuming that it introduces most of the features from your add-on. However, I built it today and realized that it is a watered-down version.

The things I noticed:
- It replaces graphical answer buttons with text-only buttons.
- Similar is the case with other buttons like Edit, More, Options, Custom Study, Description, Get Shared, Create Deck and Import File button.
- Added reps done today to the Front side of the bottom bar

Are there any other changes?

abdnh:
I've not tested #4289 yet, but given that the existing reviewer will stay as an option for a while, users will still be able to use Flexible Grading or any other reviewer add-on. I don't see much benefit to including this add-on to Anki. Instead, I think we should discuss ways to support add-ons with the new reviewer (e.g. JS API)

tatsumoto-ren:
>Can you list what features does this PR introduce? I was assuming that it introduces most of the features from your add-on. However, I built it today and realized that it is a watered-down version.

The code was written from scratch. It implements the bottom bar in a different way and copies some features from the flexible grading add-on, namely the minimalist, narrow bottom bar. Other features can be ported over in separate PRs.

>Are there any other changes?

That's the bulk of it.

>Instead, I think we should discuss ways to support add-ons with the new reviewer (e.g. JS API)

The users (of the flexible grading add-on) do not want to use the Svelte reviewer, or any interfaces made with CSS/HTML/JS. They prefer native Qt widgets for menus. So I think it's a good idea to give the users a way to disable the bloatware.

I can implement these changes in an add-on, but I need a reliable and supported way to disable all HTML-based menus (the bottom bar in this case) because they get in the way.

tatsumoto-ren:
I was tired of waiting for the features to be merged into Anki, so I released them as an add-on.

See the announcement: https://freesoftwareextremist.com/notice/B6s5soob8JqUNhMMng

PR #4449, opened 2025-11-26
feat: add actual date in tooltip for Reviews graph

selinakwok:
Added the actual date above the "days ago" line in the tooltip on hover for the Reviews graph, based on this [forum post](https://forums.ankiweb.net/t/also-show-the-date-for-each-histogram-bar-in-stats-reviews-mouseover-popup/42655).

- When the graph bin is a single day (1 month/3 months/current day in year), the long date format is displayed, following the format in the Calendar section:

Screenshot 2025-11-26 192237
Screenshot 2025-11-26 192340
Screenshot 2025-11-26 192403

- When the bin covers a range of days (year), a shorter date format is used instead (eg. Nov 16, 2025 - Nov 20, 2025): 

Screenshot 2025-11-26 192414

_This is my first contribution, happy to adjust anything :)_


ts/routes/graphs/reviews.ts
```diff
@@ -212,10 +218,43 @@ export function renderReviews(
     }
 
     function tooltipText(d: BinType, cumulative: number): string {
+        let dateStr: string;
+        const now = new Date();
+        const larger = Math.max(Math.abs(d.x0!), Math.abs(d.x1!));
+        const smaller = Math.min(Math.abs(d.x0!), Math.abs(d.x1!));
+        if (larger - smaller > 1) {
```
abdnh: We can simplify this to `d.x1! - d.x0!`?

abdnh (approved):

ts/routes/graphs/reviews.ts
selinakwok: Yes, this works and is clearer, thanks for pointing this out! Just pushed the change.

abdnh (approved):

dae:
Thank you Selina! Happy to merge, but there's currently a conflict on the CONTRIBUTORS file - would you mind rebasing?

selinakwok:
Thanks dae! Merged with main and resolved the conflicts in the contributors file and reviews.ts (conflicted with recent changes from #4430).

However, it seems like there is a bug related to #4430, where the Reviews chart tooltip day label for the most recent bin is excluding the current day (currently showing "Yesterday" instead of "0-1 days ago" or similar), screenshot below:

Screenshot 2026-02-12 235746 <br>

I haven't had the time to investigate this issue; should I submit another issue, create another pr, or try to fix it in this pr? Thank you!

user1823:
> where the Reviews chart tooltip day label for the most recent bin is excluding the current day

Can you reproduce the issue on `main`? Asking because I didn't encounter this issue when testing that PR.

abdnh:
I also don't see the issue in your PR after you merged the latest changes.

user1823:
This issue should be fixed by the following patch (not tested, though):

```diff
diff --git a/ts/routes/graphs/reviews.ts b/ts/routes/graphs/reviews.ts
index 06e02e9a7..f30bd4231 100644
--- a/ts/routes/graphs/reviews.ts
+++ b/ts/routes/graphs/reviews.ts
@@ -132,8 +132,11 @@ export function renderReviews(
             thresholds = x.ticks(desiredBars);
         }
     }
-    // For Year & All Time, shift thresholds forward by one day to make first bin 0-4 instead of 0-5
-    if (range === GraphRange.Year || range === GraphRange.AllTime) {
+    // When working with negative values, D3's bin generation is unintuitive. As a workaround,
+    // shift thresholds forward by one day when we have grouped data (bins containing >1 day)
+    const binSpacing = thresholds.length >= 2 ? thresholds[1] - thresholds[0] : 1;
+    const hasGroupedBins = binSpacing > 1;
+    if (hasGroupedBins) {
         thresholds = [...new Set(thresholds.map((t) => Math.min(t + 1, 1)))].sort(
             (a, b) => a - b,
         );

```

It occurred because there were only about 100 days in the All time graph, so the data was not grouped.

Just FYI, @JMannervik 

Edit:
The above patch is required, but it's not a complete fix because in this specific case, we do have grouped bins. So, the issue is something else. I recommend creating a new issue.

selinakwok:
I applied the patch but it didn't resolve the issue. I think the problem is related to the following code:
- ts/routes/graphs/reviews.ts (`tooltipText` function, L271-273)
```ts
const startDay = isFirstBin ? originalXMin : Math.floor(d.x0!);
const endDay = d.x1! === 0 ? 1 : d.x1!;
const day = dayLabel(startDay, endDay);
```
- ts/routes/graphs/time.ts 
```ts
export function dayLabel(daysStart: number, daysEnd: number): string {
    const larger = Math.max(Math.abs(daysStart), Math.abs(daysEnd));
    const smaller = Math.min(Math.abs(daysStart), Math.abs(daysEnd));
    if (larger - smaller <= 1) {
        // singular
        if (daysStart >= 0) {
            return tr.statisticsInDaysSingle({ days: daysStart });
        } else {
            return tr.statisticsDaysAgoSingle({ days: -daysStart });
        }
    } else {
```
In this case, for the latest bin, `startDay = d.x0 = -1` and `endDay = d.x1 = 0 -> 1`. When passed into the `dayLabel` function, `larger=smaller=1` and so it returns a singular label using -1 (yesterday).

user1823:
Thanks for finding out the cause. This patch should fix it:
```diff
diff --git a/ts/lib/tslib/time.ts b/ts/lib/tslib/time.ts
index 25d70eef3..787658d1f 100644
--- a/ts/lib/tslib/time.ts
+++ b/ts/lib/tslib/time.ts
@@ -161,9 +161,7 @@ export function timeSpan(
 }

 export function dayLabel(daysStart: number, daysEnd: number): string {
-    const larger = Math.max(Math.abs(daysStart), Math.abs(daysEnd));
-    const smaller = Math.min(Math.abs(daysStart), Math.abs(daysEnd));
-    if (larger - smaller <= 1) {
+    if (Math.abs(daysEnd - daysStart) <= 1) {
         // singular
         if (daysStart >= 0) {
             return tr.statisticsInDaysSingle({ days: daysStart });
diff --git a/ts/routes/graphs/reviews.ts b/ts/routes/graphs/reviews.ts
index f30bd4231..caaede142 100644
--- a/ts/routes/graphs/reviews.ts
+++ b/ts/routes/graphs/reviews.ts
@@ -274,11 +274,12 @@ export function renderReviews(
             });
         }
         // Convert bin boundaries [x0, x1) for dayLabel
-        // If bin ends at 0, treat it as crossing zero so day 0 is included
+        // If the last bin ends at 0, treat it as crossing zero so day 0 is included
         // For the first (oldest) bin, use the original xMin to ensure labels match the intended range
         const isFirstBin = bins.length > 0 && d.x0 === bins[0].x0;
+        const isLastBin = bins.length > 0 && d.x0 === bins[bins.length - 1].x0;
         const startDay = isFirstBin ? originalXMin : Math.floor(d.x0!);
-        const endDay = d.x1! === 0 ? 1 : d.x1!;
+        const endDay = d.x1! === 0 && isLastBin ? 1 : d.x1!;
         const day = dayLabel(startDay, endDay);
         const totals = totalsForBin(d);
         const dayTotal = valueLabel(sum(totals));
```

I noticed one remaining issue. It's not introduced by your PR. So, we can create a new issue to track it. But, here it is, if you want to fix it.

After searching for `deck:current introduced:100` in the Stats page and selecting all history, the first two bins seem to be overlapping.

Issue #4448, opened 2025-11-26
"br" appears in error message


Arthur-Milchior:
Version ⁨25.09.2 (3890e12c)⁩

See image:
<img width="500" height="101" alt="Image" src="https://github.com/user-attachments/assets/16558415-263d-43db-ab8f-29a25a3bf84f" />

PR #4451, opened 2025-11-26
fix: prevent <br> from appearing as text in error message

leedoughty:
Fixes: https://github.com/ankitects/anki/issues/4448

An issue was found where a `<br>` tag appeared as literal text in an error message when editing the card template. After investigating, I discovered that the error text was being set as PlainText. Changing this to RichText allows the HTML `<br>` tag to render as an actual line break, so the error message now displays as intended.

**Before:**
before

**After:**
after

### DIFF
--- a/tools/minilints/src/main.rs
+++ b/tools/minilints/src/main.rs
@@ -202,7 +202,7 @@ fn sveltekit_temp_file(path: &str) -> bool {
 }
 
 fn check_cargo_deny() -> Result<()> {
-    Command::run("cargo install cargo-deny@0.18.3")?;
+    Command::run("cargo install cargo-deny@0.18.6")?;
     Command::run("cargo deny check")?;
     Ok(())
 }


iamllama:
This change was proposed before, but rejected: https://github.com/ankitects/anki/pull/4246#issuecomment-3158136028

abdnh:
Maybe we can get away with adding a special case for CardTypeError: Check the type of the exception in show_exception() and pass a text format to _init_message_box() accordingly.

leedoughty:
> Maybe we can get away with adding a special case for CardTypeError: Check the type of the exception in show_exception() and pass a text format to _init_message_box() accordingly.

Thanks for the suggestion, @abdnh. I’ve updated the code so that show_exception() now checks for CardTypeError and passes the RichText text format to _init_message_box() just for that case. I tested it out and confirmed that HTML in CardTypeError messages now displays correctly, while other errors still use plain text. Let me know if you have any other thoughts or suggestions.

qt/aqt/errors.py
```diff
@@ -36,6 +36,14 @@ def show_exception(*, parent: QWidget, exception: Exception) -> None:
     global _mbox
     error_lines = []
     help_page = HelpPage.TROUBLESHOOTING
+
+    # default to PlainText
+    text_format = Qt.TextFormat.PlainText
+
+    # set CardTypeError messages as rich text to allow HTML formatting
+    if type(exception).__name__ == "CardTypeError":
```
abdnh: We should use the CardTypeError type from the `anki.errors` module instead.

abdnh (approved):

qt/aqt/errors.py
leedoughty: I've made the change but noticed CI failing afterwards, whereas the checks were passing on my machine locally. I'm currently looking into this to see the cause and whether it's related to my change.

GithubAnon0000: might not be your fault. after abdnh merged main into my pr, the tests failed as well but they passed before. It might be that main needs an update, not your pr.

iamllama: The CI failure is caused by #4455. Will be fixed by #4439 (pushed a commit there, CI passes now)

GithubAnon0000: How did #4455 cause this? It hasn't been merged to main as far as I can tell and thus shouldn't be able to cause the CI to fail.

iamllama: See #4439's description. It doesn't matter whether a pr was merged, as long as it can run CI it can cause changes to it. #4367 was one such example.

PR #4453, opened 2025-12-02
Fix/don't reschedule suspended cards

L-M-Sherlock:


iamllama:
For context, this fixes an `attempt to subtract with overflow` panic at https://github.com/ankitects/anki/blob/5614d20bedcc4dd268136d389ad796b404a69d2c/rslib/src/scheduler/fsrs/rescheduler.rs?plain=1#L89 when run via `./run` (went unnoticed in release as rust doesn't panic during overflows then)

L-M-Sherlock:
In addition to @iamllama's context:

https://github.com/ankitects/anki/blob/2d4de33cf3160342c4c704c294e643c3e11071b1/rslib/src/storage/card/deck_due_counts.sql#L1-L14

The `count` doesn't include suspended cards.

rslib/src/scheduler/fsrs/memory_state.rs
```diff
@@ -128,8 +129,10 @@ impl Collection {
                                         timing.next_day_at.elapsed_days_since(*last_review) as i32;
                                     // and the card's not new
                                     if let Some(state) = &card.memory_state {
-                                        // or in (re)learning
-                                        if card.ctype == CardType::Review {
+                                        // or in (re)learning and suspended
```
Luc-Mcgrady, line 132:
```suggestion
                                        // and in review and suspended
```
This comment is confusing because it's merged with the top one 
```
// and the card's not new
```
and I didn't realise untill after I had made the review :sweat_smile:

dae:
Thank you all!

PR #4454, opened 2025-12-03
Add headers to deleted.txt to detect Note Type and GUID

hnvy:
Hi

A very minor change.

I was reviewing the Anki Manual and noted that there is a somewhat outdated [section](https://docs.ankiweb.net/backups.html#deletion-log):
> [...] please note the import feature only supports a single note type at one time [...]

This PR adds headers to `deleted.txt` to allow for easy import:
```
#guid column:1
#notetype column:2
```

If undesired, will close this PR.

If accepted, I can update https://docs.ankiweb.net/backups.html#deletion-log

Other option is to explain to the user to add the said headers to the top of `deleted.txt`.

Whichever you feel is better.

Harvey

### DIFF
--- a/qt/aqt/main.py
+++ b/qt/aqt/main.py
@@ -1643,7 +1643,9 @@ title="{}" {}>{}</button>""".format(
         existed = os.path.exists(path)
         with open(path, "ab") as f:
             if not existed:
-                f.write(b"nid\tmid\tfields\n")
+                f.write(b"#guid column:1\n")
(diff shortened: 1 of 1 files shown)


abdnh (approved):
This looks reasonable.

brishtibheja:
> If undesired, will close this PR.

Here's a thread with this topic https://forums.ankiweb.net/t/deletion-log-support-note-type-column/52721?u=sorata

dae:
Thanks!

PR #4455, opened 2025-12-03
feat(proto): extend AddMediaFile to accept file paths for streaming s…

shivraj1182:
Modify AddMediaFileRequest message to support both inline byte data and file path references
Use oneof construct to ensure exactly one data source is provided (backward compatible)
This enables memory-efficient streaming uploads for large media files on mobile clients
Addresses issue #608: AnkiDroid OutOfMemoryError when adding large media files


david-allison:
Requesting this is closed: 

* https://github.com/ankidroid/Anki-Android-Backend/issues/608#issuecomment-3605282567

PR #4458, opened 2025-12-07
fix(io): remove incorrect error toast shown when saving twice

iamllama:
ref: https://github.com/ankidroid/Anki-Android/issues/19678#issuecomment-3621096241

Saving an IO note twice on mobile is a no-op, which is incorrectly considered an error

### DIFF
--- a/ts/routes/image-occlusion/add-or-update-note.svelte.ts
+++ b/ts/routes/image-occlusion/add-or-update-note.svelte.ts
@@ -37,7 +37,9 @@ export const addOrUpdateNote = async function(
             backExtra,
             tags,
         });
-        showResult(mode.noteId, result, noteCount);
+        if (result.note) {
+            showResult(mode.noteId, result, noteCount);
+        }
     } else {
         const result = await addImageOcclusionNote({
             // IOCloningMode is not used on mobile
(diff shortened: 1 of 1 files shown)


ts/routes/image-occlusion/add-or-update-note.svelte.ts
```diff
@@ -37,7 +37,9 @@ export const addOrUpdateNote = async function(
             backExtra,
             tags,
         });
-        showResult(mode.noteId, result, noteCount);
+        if (result.note) {
```
david-allison, line 40: (may be misunderstanding): This no longer appears to notify the user of unexpected errors

iamllama: > Is this confirmed to be current backend behaviour?

Tested by [calling `anki.imageOcclusion.save()`](https://github.com/ankidroid/Anki-Android/blob/0fd4125d96550168175fd0dcea464fa2b2bfafd7/AnkiDroid/src/main/java/com/ichi2/anki/pages/ImageOcclusion.kt#L92) via the vite dev server


> This no longer appears to notify the user of unexpected errors

If an error occurred, the call to `updateImageOcclusionNote` wouldve thrown (showing an alert instead). We could perhaps detect that the correct number of cards weren't generated/updated but 1) that count isn't returned by the backend and 2) if it were we'd only get back the number of new cards confirmed to be generated, not updated

david-allison: This is my unfamiliarity with the TypeScript showing. Looks good to me. Thanks for your patience.

